PR #5958, opened 2025-07-18
fix(sql): exception when JIT compiling a filter like `WHERE symbol = ''''`

nwoolmer:
We had some unquoting logic that didn't account for this edge case.

We should still review why upstream code was sending such a filter.


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/constants/SymbolConstant.java|3|3|100.00%|

bluestreak01 (approved):

PR #5981, opened 2025-07-21
feat(sql): add IF EXISTS support to TRUNCATE TABLE

### LINKED ISSUE #5763: [Enhancement] Add IF EXISTS to truncate table

A small usability improvement would be to have `IF EXISTS` in `TRUNCATE TABLE` so it will not error when table is missing, as we have with `DROP TABLE`

Both
```
TRUNCATE TABLE ratings;
```
and
```
TRUNCATE TABLE IF EXISTS ratings;
```

At the moment I truncate and get an error. It'd be nice if I didn't need error control when I choose to ignore the fact table might yet not exist.

### Full Name:

javier ramirez

### Affiliation:

questdb

### Additional context

Very minor thing. If implementation is not trivial, happy to ignore it.

DevvSaxena17:
his PR adds support for the IF EXISTS clause to the TRUNCATE TABLE SQL command.
Users can now write TRUNCATE TABLE IF EXISTS table_name to avoid errors if the table does not exist.
If IF EXISTS is specified and the table is missing, the command will not throw an error and will continue gracefully.
If IF EXISTS is not specified, the previous behavior is retained (an error is thrown if the table does not exist).
This enhancement improves SQL compatibility and user experience when managing tables.
Closes #5763.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,13 +1,20 @@
 # Contributing to QuestDB
 
-Hi, glad to know that you're interested in contributing to QuestDB. Here are
-some topics that can help you to get started:
+Thank you for your interest in contributing to QuestDB! We welcome contributions of all kinds, including code, documentation, bug reports, and feature requests. This guide will help you get started and make your contribution process smooth and successful.
 
-- 💡 [Bugs and features](#bugs-and-features)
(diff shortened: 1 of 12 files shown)


nwoolmer:
Full of debris.

PR #5982, opened 2025-07-21
perf(functions): loop-unroll common var-args cases in MultiArgFunction

### LINKED ISSUE #5680: Optimise performance of var-args functions by adding loop-unrolled variants

There are a variety of `MultiArgFunction` functions, with variable argument lists. Often, these lists will be short (<5 arguments), but there is no guarantee that the JVM will unroll their loops automatically.

Functions that perform relatively simple operations may benefit from manually unrolled variants that skip loop and conditional checks, and instead have a straight bytecode path for the calculations.

Review var-args functions, particularly those with numeric arguments and basic calculations. If they are a candidate for loop unrolling, implement and benchmark to prove it is faster.

Examples of var-args functions:

- greatest/least
- IN
- CASE
- coalesce
- ~l2price~
- concat

_No response_

### Full Name:

Nick Woolmer

### Affiliation:

QuestDB

### Additional context

_No response_

DevvSaxena17:
This PR optimizes the performance of SQL functions that use variable arguments by loop-unrolling the most common argument counts (2, 3, 4) in the MultiArgFunction interface.
Reduces loop overhead for typical use cases.
Falls back to the original loop for larger argument counts.
Improves performance for many built-in SQL functions.
Closes #5680.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/functions/MultiArgFunction.java
+++ b/core/src/main/java/io/questdb/griffin/engine/functions/MultiArgFunction.java
@@ -50,60 +50,111 @@ public interface MultiArgFunction extends Function {
     @Override
     default boolean isConstant() {
         ObjList<Function> args = getArgs();
-        for (int i = 0, n = args.size(); i < n; i++) {
-            if (!args.getQuick(i).isConstant()) {
-                return false;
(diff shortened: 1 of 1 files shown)


nwoolmer:
Full of debris.

PR #5995, opened 2025-07-29
fix(sql): reduce write amplification for materialized views with small SAMPLE BY interval

puzpuzpuz:
Removes minimal refresh interval defined by `cairo.mat.view.min.refresh.interval` property. This property and its associated logic led to larger refresh interval being used in case of base table commits with a small time interval (e.g. a single timestamp commit). For instance, in case of a SAMPLE BY 1s even in case of a single timestamp commit the final refresh interval was 100s.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/PostOrderTreeTraversalAlgo.java
+++ b/core/src/main/java/io/questdb/griffin/PostOrderTreeTraversalAlgo.java
@@ -24,12 +24,15 @@
 
 package io.questdb.griffin;
 
+import io.questdb.cairo.CairoException;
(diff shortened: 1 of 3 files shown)


ideoma (approved):

puzpuzpuz:
@ideoma thanks for the review!

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/PropServerConfiguration.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|1|1|100.00%|

Issue #6002, opened 2025-07-30
spurious 'undeclared variable' error


jerrinot:
### To reproduce

Logs from the demo:
```
2025-07-30T10:06:03.305357Z E i.q.g.e.QueryProgress err [id=-1, sql=
DECLARE
  @symbol := 'BTC-USDT'
WITH sampled AS (
    SELECT
          timestamp,  symbol,
          volume AS volume,
          ((open+close)/2) * volume AS traded_value
     FROM trades_OHLC_15m
     WHERE timestamp BETWEEN '2025-07-28T22:00:00Z' AND '2025-07-29T21:59:59.999Z' 
     AND symbol = @symbol      
), cumulative AS (
     SELECT timestamp, symbol,
           SUM(traded_value)
                OVER (ORDER BY timestamp) AS cumulative_value,
           SUM(volume)
                OVER (ORDER BY timestamp) AS cumulative_volume
     FROM sampled
)
SELECT timestamp as time, cumulative_value/cumulative_volume AS vwap_yesterday FROM cumulative;
, principal=admin, cache=false, jit=true, time=186213, msg=tried to use undeclared variable volume, errno=0, pos=148]
```

This makes no sense, since `volume` is not a variable in this query. 

### QuestDB version:

9.0.1

### OS, in case of Docker specify Docker and the Host OS:

irrelevant

### File System, in case of Docker specify Host File System:

irrelevant

### Full Name:

Jaromir Hamala

### Affiliation:

QuestDB

### Have you followed Linux, MacOs kernel configuration steps to increase Maximum open files and Maximum virtual memory areas limit?

- [x] Yes, I have

PR #6003, opened 2025-07-30
fix(sql): fix spurious 'undeclared variable' error

jerrinot:
Symptoms:
Under high query load, QuestDB could throw a spurious 'undeclared variable error'.

Root cause:
SqlParser instance shared a single stateful visitor to replace variables. This creates a race where state from one SQL could bleed into an unrelated SQL.

fixes https://github.com/questdb/questdb/issues/6002


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlParser.java|1|1|100.00%|

PR #6018, opened 2025-08-01
fix(sql): improve error reporting for unclosed parentheses

### LINKED ISSUE #6010: Unhelpful SQL syntax error
### To reproduce

```SQL
SELECT rnd_symbol(
  'BTC-USD', 'SHIB-USD', 
  'SOL-BTC', 'UNI-USDT',
  'DOT-BTC', 'LTC-USDT', 
  'ETH-USDT', 'SOL-USD', 
  'AVAX-USD', 'BTC-USDC',
  'ETH-USD', 'DOT-USD', 
  'SOL-ETH', 'MATIC-USDT'
),
rnd_symbol('BUY', 'SELL'),
abs(1000*rnd_double()),
abs(100000*rnd_double(),
timestamp_sequence('2022-07-01', 1000000)
from long_sequence(1000000)
;
```

error is

Image

It's not helpful because it points to FROM while the problem is unclosed brace in the second `abs()`

### QuestDB version:

9.0.1

na

na

### Full Name:

Alex Pelagenko

### Affiliation:

QuestDB

- [x] Yes, I have

### Additional context

_No response_

turhantolgaunal:
Fixes #6010


bluestreak01:
@turhantolgaunal please consider the solution without the context of failing tests, run `SqlParserTest`

turhantolgaunal:
I am sorry for not noticing it. Although I am not sure how to proceed, as far as I can see the problem is that  the parser throws a dangling literal error if it encounters an unexpected literal within a function call, which makes it stop parsing so it can't see whether the parentheses were closed or not.

turhantolgaunal:
Hello, I think I managed to do it, can you look into it to see if this is appropriate? @bluestreak01

PR #6025, opened 2025-08-04
perf(core): introduce off heap char map to reduce Java heap usage during WAL writing

ideoma:


bluestreak01:
@coderabbitai could you review pls?

PR #5960, opened 2025-07-18
feat(core): cache file descriptors and mmaps to reduce kernel resource usage

ideoma:
## Pull Request Overview

This PR implements file system call caching to improve performance by caching file descriptor and memory map operations. The changes introduce caching layers for both file descriptors and memory maps, reducing redundant system calls and improving resource reuse.

- Introduces FdCache and MmapCache classes to cache file descriptors and memory maps
- Changes openRW parameter type from long to int throughout the codebase
- Adds new openRWNoCache and openRODir methods to FilesFacade interface
- Updates native file operations to use overlapped I/O on Windows

This PR leaves further optimisations to reduce locking on the new `FdCache` and `MampCache` classes for the future.

PR #6031, opened 2025-08-07
chore(build): fix cxx build failed in linux platform

kafka1991:
The CXX build failed on Linux which introduced in https://github.com/questdb/questdb/pull/5960. 

The most recent failed CXX build pipeline for the branch is failed: https://github.com/questdb/questdb/actions/runs/16754327173.

@ideoma Could you take a review when you have time?


jerrinot:
the fix is good, but what do we need Unix sockets for anyway? @ideoma any idea?

jerrinot (approved):

PR #6038, opened 2025-08-08
fix(core): incorrect year rounding in case of leap years

### LINKED ISSUE #6029: Materialized views enters suspended immediately
### To reproduce

```sql
CREATE TABLE Samples ( 
	Time TIMESTAMP,
	DeviceId INT,
	Register SYMBOL CAPACITY 256 CACHE INDEX CAPACITY 256,
	Value DOUBLE
) timestamp(Time) PARTITION BY MONTH WAL
DEDUP UPSERT KEYS(Time,DeviceId,Register);

CREATE MATERIALIZED VIEW Samples_latest AS
  SELECT
    Time as UnixEpoch,
    last(Time) AS Time,
    DeviceId,
    Register,
    last(Value) AS Value
  FROM
    Samples
  SAMPLE BY 2y;

INSERT INTO Samples (Time, DeviceId, Register, Value) VALUES (now(), 1, 'hello', 123);

SELECT * FROM wal_tables() WHERE suspended;
```

### QuestDB version:

9.0.1

Windows

ntfs

### Full Name:

Rasmus Melchior Jacobsen

### Affiliation:

Utiliread

- [x] Yes, I have

### Additional context

This is a regression, it worked fine in v8. Changing SAMPLE BY period to 1y works.

puzpuzpuz:
Fixes #6029

`YearTimestampSampler` sampler had a bug around rounding. Namely, when the rounded year was a leap year and the input year wasn't, the returned result was incorrect.

This bug manifested itself in materialized view refresh where the lo value of a range replace commit could end up being greater than the minimal timestamp of the transaction. The reason is that the range replace lo was calculated based on `YearTimestampSampler#round()`.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/groupby/YearTimestampSampler.java
+++ b/core/src/main/java/io/questdb/griffin/engine/groupby/YearTimestampSampler.java
@@ -64,9 +64,10 @@ public class YearTimestampSampler implements TimestampSampler {
 
     @Override
     public long round(long value) {
-        final int y = Timestamps.getYear(value);
+        int y = Timestamps.getYear(value);
+        y = y - y % stepYears;
         return Timestamps.toMicros(
-                y - y % stepYears,
+                y,
                 Timestamps.isLeapYear(y),
                 startDay,
                 startMonth,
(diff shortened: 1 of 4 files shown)


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/YearTimestampSampler.java|2|2|100.00%|

puzpuzpuz:
@bluestreak01 @ideoma thanks for the reviews!

PR #6049, opened 2025-08-12
fix(sql): misleading error message in subquery comparison

### LINKED ISSUE #5646: misleading error message in subquery comparison
### To reproduce

```sql
create table tab (i int);
select * from tab where i = (select max(i) from tab);
```
The query fails with the error: `cannot compare TIMESTAMP and INT`, despite no TIMESTAMP being present in the query.

Root cause: 
```java
public class EqTimestampCursorFunctionFactory implements FunctionFactory {

    @Override
    public String getSignature() {
        return "=(NC)";
    }
[...]
```

The equality operator = with a subquery invokes [EqTimestampCursorFunctionFactory](https://github.com/questdb/questdb/blob/311354f3146b99028adf8374dcd43ca2d1e37b59/core/src/main/java/io/questdb/griffin/engine/functions/eq/EqTimestampCursorFunctionFactory.java#L48), which expects a timestamp and cursor. The function parser allows an INT to be implicitly cast to a timestamp, but then fails at runtime with a hard-coded error message mentioning "TIMESTAMP" rather than the actual type.

My first idea to fix it was to check the actual return type of the first argument, instead of hard-coding `timestamp` type. However in this case, this would end with even more confusing message: `cannot compare INT and INT` which would be even more confusing. So perhaps a better solution is to restrict implicit casting when matching a function accepting a cursor. 

Similar issues are likely in other factories accepting a cursor: `GtTimestampCursorFunctionFactory`, `LtTimestampCursorFunctionFactory`, and possibly also `InSymbolCursorFunctionFactory`. 

### QuestDB version:

8.3.1

irrelevant

irrelevant

### Full Name:

Jaromir Hamala

### Affiliation:

QuestDB

- [x] Yes, I have

### Additional context

_No response_

klaudiozdrava:
Fixes #5646

Reopening this pull request due to a Git commit author/GitHub username mismatch that affected the CLA check. 

I fixed the issue by preventing implicit casting from INT to TIMESTAMP and improving the error message to clearly state:
"left operand must be a TIMESTAMP, STRING, or VARCHAR, found: (type)",
where the actual type—for example, INT—is displayed.

The fix updates EqTimestampCursorFunctionFactory, LtTimestampCursorFunctionFactory, and GtTimestampCursorFunctionFactory.
InSymbolCursorFunctionFactory already throws a specific exception, so it was left unchanged.


core/src/main/java/io/questdb/griffin/engine/functions/eq/EqTimestampCursorFunctionFactory.java
```diff
@@ -79,6 +79,13 @@ public Function newInstance(
             throw SqlException.$(argPositions.getQuick(1), "select must provide exactly one column");
         }
 
+        if (args.getQuick(0).getType() != ColumnType.TIMESTAMP &&
```
puzpuzpuz: I guess the left operand can also be DATE, LONG, SYMBOL, and maybe some others (would be great to add more tests). You could try doing something like `ColumnType.isToSameOrWider(args.getQuick(0).getType(), ColumnType.TIMESTAMP)` in this check.

klaudiozdrava: I tested `ColumnType.isToSameOrWider(args.getQuick(0).getType(), ColumnType.TIMESTAMP)` across all types (0–27). Only BYTE, SHORT, CHAR, INT, LONG, DATE, TIMESTAMP, STRING, SYMBOL, VARCHAR are compatible. Of these, only TIMESTAMP, STRING, SYMBOL, VARCHAR can be implicitly cast without exceptions.

So the check could be simplified to:
```
if (args.getQuick(0).getType() != ColumnType.TIMESTAMP &&
    args.getQuick(0).getType() != ColumnType.STRING &&
    args.getQuick(0).getType() != ColumnType.VARCHAR &&
    args.getQuick(0).getType() != ColumnType.SYMBOL) {
    throw SqlException.$(argPositions.getQuick(0),
        "left operand must be a TIMESTAMP, STRING, VARCHAR, or SYMBOL, found: ")
        .put(ColumnType.nameOf(metadata.getColumnType(0)));
}
```

Let me know if this approach looks valid or if I’ve missed something regarding `isToSameOrWider`. I’m also providing some tests for other types like DATE, LONG, and SYMBOL.

SYMBOL:
Screenshot 2025-08-15 at 5 06 10 PM

DATE:
Screenshot 2025-08-15 at 4 35 13 PM

LONG:
Screenshot 2025-08-14 at 1 31 47 AM

We get similar exception for INT and DOUBLE too.

puzpuzpuz: > Let me know if this approach looks valid or if I’ve missed something regarding `isToSameOrWider`.

Yeah, my initial suggestion wasn't well-thought. I'd better make the check even stricter and always as for the TIMESTAMP type. Optionally, we could also allow DATE type, but that's not a must. WDYT?

klaudiozdrava: I think that if we do that we won't be able to process timestamps from STRING or VARCHAR type. 

`  switch (metadata.getColumnType(0)) {
            case ColumnType.TIMESTAMP:
            case ColumnType.NULL:
                return new TimestampCursorFunc(factory, args.getQuick(0), args.getQuick(1));
            case ColumnType.STRING:
                return new StrCursorFunc(factory, args.getQuick(0), args.getQuick(1), argPositions.getQuick(1));
            case ColumnType.VARCHAR:
                return new VarcharCursorFunc(factory, args.getQuick(0), args.getQuick(1), argPositions.getQuick(1));
            default:
                throw SqlException.$(argPositions.getQuick(1), "cannot compare TIMESTAMP and ").put(ColumnType.nameOf(metadata.getColumnType(0)));
        }`
        
Based on my testing, the DATE and TIMESTAMP data types appear to be incompatible with each other.

puzpuzpuz: I don't think we have any tests around string/varchar left-hand types, so they work only due to implicit parsing/type conversion. I'm not sure that supporting them is a good idea considering #5685 - the format would be the default one while the user expects the same format as in the right-hand query (nanos/micros).

> Based on my testing, the DATE and TIMESTAMP data types appear to be incompatible with each other.

They should be compatible. DATE is nothing more, but a TIMESTAMP with millisecond precision. But yes, in this particular case left-hand DATE doesn't work, so we don't need to check for it.

klaudiozdrava: Based on your point about string/varchar left-hand types, I understand we should remove the `StrCursorFunc` and `VarcharCursorFunc` classes entirely. Is that correct?

Regarding the `ColumnType.NULL` handling - should we keep it grouped with TIMESTAMP in the current approach?

puzpuzpuz: > Based on your point about string/varchar left-hand types, I understand we should remove the `StrCursorFunc` and `VarcharCursorFunc` classes entirely. Is that correct?

No, those classes serve cases when the right-hand (the cursor's output) has STRING/VARCHAR type. In this PR, we're discussing restrictions for the left-hand, usually, a column that's compared with the result of the right-hand cursor.

> Regarding the `ColumnType.NULL` handling - should we keep it grouped with TIMESTAMP in the current approach?

Yes, that code should be kept as is.

klaudiozdrava: I misunderstood your suggestion. Based on your feedback, I believe we want to implement something like this:

Screenshot 2025-08-20 at 6 54 54 PM

puzpuzpuz: Yup, something like that.

@jerrinot WDYT?

klaudiozdrava: If you agree, I'll push the updates to this PR.

jerrinot: the proposal looks good to me, thanks for fixing this!

klaudiozdrava: Sorry for the delay. I was caught up with my day job.
These whitespace formatting issues detected by the pipeline can be so sneaky...
Please let me know if the fix I pushed looks correct.

core/src/main/java/io/questdb/griffin/engine/functions/lt/GtTimestampCursorFunctionFactory.java
```diff
@@ -79,6 +79,10 @@ public Function newInstance(
             throw SqlException.$(argPositions.getQuick(1), "select must provide exactly one column");
         }
 
+        if (ColumnType.tagOf(args.getQuick(0).getType()) != ColumnType.TIMESTAMP) {
```
puzpuzpuz, line 82: nit: the code would read better with an extra variable:
```java
final int leftArgType = args.getQuick(0).getType();
if (ColumnType.tagOf(leftArgType) != ColumnType.TIMESTAMP) {
    throw SqlException.$(argPositions.getQuick(0), "left operand must be a TIMESTAMP, found: ").put(ColumnType.nameOf(leftArgType));
}
```

puzpuzpuz (approved):

puzpuzpuz:
@klaudiozdrava thanks for the contribution and sorry for the back-and-forth.

klaudiozdrava:
> @klaudiozdrava thanks for the contribution and sorry for the back-and-forth.

@puzpuzpuz Thanks for approving my PR! 
My initial suggestion wasn't actually 100% correct so it was good to catch it.

jerrinot:
hello @klaudiozdrava, congratulations on getting your PR merged! Thank you and keep them comin'! :)

klaudiozdrava:
Thank you, @jerrinot and the team, for your feedback! I'm interested in contributing more in the future.

PR #6055, opened 2025-08-15
fix(sql): memory leak when querying a table which is being ALTERed concurrently

jerrinot:
In rare cases where queries execute while a table is being ALTERed simultaneously, memory could leak during internal query retries. This edge case only affected systems with concurrent schema modifications and queries on the same table. While queries completed successfully through automatic retries, memory from failed attempts was not released.

### Implementation details
The issue occurred when the SQL compiler attempted to create a filter during query compilation:
1. Compiler creates a factory
2. Attempts to wrap it in a filter
3. Filter creation tries to obtain a cursor, which can fail due to `TableReferenceOutOfDateException`
4. On failure, the factory wasn't being freed before the retry


bluestreak01 (approved):

glasstiger:
#### [PR Coverage check]

:disappointed: **fail** : 0 / 5 (00.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|0|5|00.00%|

PR #6064, opened 2025-08-18
fix(sql): fix db crash when querying parquet files with array columns

puzpuzpuz:
The bug was due to `bufs.aux_vec.clear();` done in case of array column while it was meant for fixed-size columns only. The segfault only happened when a row group contained multiple data pages.


javier (approved):
I didn't check the code, but I checked that the error I reported is not there anymore. It works fine :)

puzpuzpuz:
@javier thanks for testing the patch!

ideoma:
hey @RaphDal , can you also please have a look?

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 49 / 50 (98.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|parquet_read/decode.rs|49|50|98.00%|

puzpuzpuz:
@RaphDal thanks for the review!

PR #6074, opened 2025-08-22
feat(ilp): add support for configuring multiple possible database endpoints

nwoolmer:
tandem with https://github.com/questdb/questdb-enterprise/pull/669

This PR adds support to the Java ILP client (and the QuestDB server) to handle multiple possible endpoints. On first connection, the client will poll `/settings` and check if the instance is read-write or read-only. Additionally, if this changes behind the scenes after the sender has been initialised, then the sender will automatically retry HTTP sends on 404 or 421 responses it receives.


glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 151 / 170 (88.82%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/line/AbstractLineSender.java|0|2|00.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/LineHttpProcessorImpl.java|1|3|33.33%|
|:large_blue_circle:|io/questdb/cutlass/http/client/HttpClient.java|4|5|80.00%|
|:large_blue_circle:|io/questdb/cutlass/line/http/AbstractLineHttpSender.java|102|115|88.70%|
|:large_blue_circle:|io/questdb/cutlass/line/LineSenderException.java|8|9|88.89%|
|:large_blue_circle:|io/questdb/cutlass/http/client/HttpClientException.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/IntList.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/SettingsProcessor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/Sender.java|18|18|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/LineHttpProcessorState.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpResponseSink.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/http/LineHttpSenderV1.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpServer.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/Rnd.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/http/LineHttpSenderV2.java|2|2|100.00%|

jerrinot:
a test failure due to https://github.com/questdb/questdb/issues/6173

puzpuzpuz:
@coderabbitai review

core/src/main/java/io/questdb/cairo/CairoConfiguration.java
```diff
@@ -680,6 +680,8 @@ default int getWalPurgeWaitBeforeDelete() {
 
     int getWriterTickRowsCountMod();
 
+    boolean isAcceptingWrites();
```
puzpuzpuz: Why is this `CairoConfiguration` setting? It would be more logical to have it at least in `ServerConfiguration` or maybe even in the nested config objects like `HttpServerConfiguration`.

core/src/main/java/io/questdb/client/Sender.java
```diff
@@ -520,21 +550,15 @@ public LineSenderBuilder address(CharSequence address) {
...
-                            .put("]");
-                }
-                this.host = address.subSequence(0, portIndex).toString();
                 try {
                     port(Numbers.parseInt(address, portIndex + 1, address.length()));
                 } catch (NumericException e) {
                     throw new LineSenderException("cannot parse a port from the address, use IPv4 address or a domain name")
                             .put(" [address=").put(address).put("]");
                 }
+                this.hosts.add(address.subSequence(0, portIndex).toString());
```
puzpuzpuz: Shouldn't we check the input for duplicate host:port pairs?

core/src/main/java/io/questdb/cutlass/http/client/HttpClient.java
```diff
@@ -477,7 +477,7 @@ public ResponseHeaders send() {
             return send(defaultTimeout);
         }
 
-        public ResponseHeaders send(int timeout) {
+        public ResponseHeaders send(CharSequence host, int port, int timeout) {
```
puzpuzpuz, line 508: This API change may break HTTP connection (socket) reuse we had in the client. If the host:port pair changes between the calls, `send()` may ignore the change and keep sending data over the old connection.

core/src/main/java/io/questdb/cutlass/http/client/HttpClient.java
```diff
@@ -609,14 +613,14 @@ private void connect(CharSequence host, int port) {
                 int errno = nf.errno();
                 nf.freeAddrInfo(addrInfo);
                 disconnect();
-                throw new HttpClientException("could not connect to host ").put("[host=").put(host).put(", port=").put(port).put(", errno=").put(errno).put(']');
+                throw new HttpClientException("could not connect to host ").put("[host=").put(host).put(", port=").put(port).put(", errno=").put(errno).put(']').errno(errno);
```
puzpuzpuz: nit: we could remove the `.put(", errno=").put(errno)` part here and below since `errno` will be included into `getMessage()` output anyway.

core/src/main/java/io/questdb/cutlass/http/processors/LineHttpProcessorState.java
```diff
@@ -464,7 +464,8 @@ public enum Status {
         INTERNAL_ERROR("internal error", 500),
         MESSAGE_TOO_LARGE("request too large", 413),
         COLUMN_ADD_ERROR("invalid", 400),
-        COMMITTED(null, 204);
+        COMMITTED(null, 204),
+        NOT_ACCEPTING_WRITES("not accepting writes", 421);
```
puzpuzpuz, line 468: nit: smth like 403 Forbidden may be a better fit here.

core/src/main/java/io/questdb/cutlass/http/processors/SettingsProcessor.java
```diff
@@ -102,6 +105,10 @@ public void onRequestComplete(HttpConnectionContext context) throws PeerDisconne
             settings.clear();
             settings.putAscii('{');
             serverConfiguration.exportConfiguration(settings);
+            bool(ACCEPTING_WRITES, serverConfiguration.getCairoConfiguration().isAcceptingWrites(), settings);
```
puzpuzpuz: Shouldn't we have "accepts reads"/"accepts writes" fields for each available interface? Also, this should probably be done in the `exportConfiguration()` method.

core/src/main/java/io/questdb/cutlass/line/http/AbstractLineHttpSender.java
```diff
@@ -666,6 +767,16 @@ private HttpClient.Request newRequest() {
...
 
+    private void reset(long newFlushAfterNanos) {
+        pendingRows = 0;
+        flushAfterNanos = newFlushAfterNanos;
+        request = newRequest();
+    }
+
+    private void rotateAddress() {
+        currentAddressIndex = (currentAddressIndex + 1) % hosts.size();
+        this.url = buildUrl();
```
puzpuzpuz: nit: we don't have to create a `String` on each rotation. Instead, we could build the list of URL strings when constructing the sender. Alternatively, we could remove the `url` field and just construct the URL when creating `LineSenderException`. `new LineSenderException("Could not flush buffer: ", true).put(url)` is the only place we use the `url` field.

core/src/main/java/io/questdb/cutlass/line/http/AbstractLineHttpSender.java
```diff
@@ -197,28 +250,71 @@ public static AbstractLineHttpSender createLineSender(
...
+                long retryingDeadlineNanos = Long.MIN_VALUE; // we want to start retry timer only after a first failure
+                int retryBackoff = RETRY_INITIAL_BACKOFF_MS;
+                long hostBlacklistBitmap = 0; // allows blacklisting up to 64 hosts; we blacklist a host upon receiving a non-retryable error
+                for (int i = 0; ; i++) {
+                    currentAddressIndex = i % hosts.size();
+                    if (currentAddressIndex < 64) {
+                        // we can blacklist only the first 64 hosts, that's fine, we don't expect more hosts anyway
+                        // and if there ever are more hosts then the side-effect is that we will retry on the same host
+                        // even after it returned a non-retryable error
+                        if (hostBlacklistBitmap == (1L << hosts.size()) - 1) {
```
puzpuzpuz: It's enough to have 63 host:port pairs to get an overflow in `1L << hosts.size()`.

core/src/main/java/io/questdb/cutlass/line/http/AbstractLineHttpSender.java
```diff
@@ -185,9 +218,29 @@ public static AbstractLineHttpSender createLineSender(
...
+            String password,
+            int maxNameLength,
+            long maxRetriesNanos,
+            long minRequestThroughput,
+            long flushIntervalNanos,
+            int protocolVersion
     ) {
         HttpClient cli = null;
         Rnd rnd = new Rnd(NanosecondClockImpl.INSTANCE.getTicks(), MicrosecondClockImpl.INSTANCE.getTicks());
+        int currentAddressIndex = 0;
```
puzpuzpuz, line 257: We should store the `currentAddressIndex` in `this.currentAddressIndex` once we establish a connection to avoid starting from index 0.

core/src/main/java/io/questdb/std/Rnd.java
```diff
@@ -317,6 +317,10 @@ public int nextInt(int boundary) {
         return nextPositiveInt() % boundary;
     }
 
+    public synchronized int nextIntSync(int boundary) {
```
puzpuzpuz: Do we really need this method? `nextInt` is racy when called concurrently, but it has no negative side-effects especially in tests.

core/src/test/java/io/questdb/test/cutlass/http/line/LineHttpMultiUrlTest.java
```diff
@@ -0,0 +1,388 @@
...
+                PropertyKey.PG_NET_BIND_TO.getEnvVarName(), host + ":" + port + 2,
+                PropertyKey.LINE_TCP_NET_BIND_TO.getEnvVarName(), host + ":" + port + 3,
+                PropertyKey.CAIRO_ROOT.getEnvVarName(), dbPath + rootName,
+                PropertyKey.LINE_HTTP_ENABLED.getEnvVarName(), "true",
+                PropertyKey.HTTP_SECURITY_READONLY.getEnvVarName(), String.valueOf(readOnly)
+        );
+    }
+
+    @Test
+    public void testFirstInstanceReadOnly() throws Exception {
```
puzpuzpuz, line 114: Pre-9.1.0 read-only instances return HTTP 405 upon receiving ILP/HTTP requests which is a non-retriable HTTP status code, so instead of retrying with the next server we end up with `io.questdb.cutlass.line.LineSenderException: Could not flush buffer: HTTP endpoint does not support ILP. [http-status=405]` exception.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 143 / 161 (88.82%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/line/AbstractLineSender.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/LineHttpProcessorImpl.java|1|3|33.33%|
|:large_blue_circle:|io/questdb/cutlass/http/client/HttpClient.java|4|5|80.00%|
|:large_blue_circle:|io/questdb/cutlass/line/http/AbstractLineHttpSender.java|93|106|87.74%|
|:large_blue_circle:|io/questdb/cutlass/line/LineSenderException.java|8|9|88.89%|
|:large_blue_circle:|io/questdb/cutlass/http/client/HttpClientException.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/IntList.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/SettingsProcessor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/Sender.java|18|18|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/LineHttpProcessorState.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpResponseSink.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/http/LineHttpSenderV1.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpServer.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/Rnd.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/http/LineHttpSenderV2.java|2|2|100.00%|

core/src/main/java/io/questdb/cutlass/line/http/AbstractLineHttpSender.java
jerrinot: this is already done - we pass the `currentAddressIndex` to `LineHttpSenderV1` and `LineHttpSenderV2` constructors and they store it.

core/src/main/java/io/questdb/std/Rnd.java
jerrinot: good call, we don't need this at all.

core/src/main/java/io/questdb/client/Sender.java
jerrinot: good idea

core/src/main/java/io/questdb/cutlass/http/client/HttpClient.java
jerrinot: fixed, but not marking this issue as resolved. as there are multiple way to solve this. so please have a look at my change and let me know if there is anything you would do differently. thanks!

core/src/main/java/io/questdb/cutlass/http/client/HttpClient.java
jerrinot: that's a good one. the errno was originally used for control-flow. but this bit was removed so we can remove the tracking too.

core/src/test/java/io/questdb/test/client/LineSenderBuilderTest.java
```diff
@@ -605,6 +604,18 @@ public void testDnsResolutionFail() throws Exception {
         });
     }
 
+    @Test
+    public void testDuplicatedAddresses() throws Exception {
```
puzpuzpuz, line 608: Could you add a test to verify that duplicate hosts, but different ports are allowed?

core/src/main/java/io/questdb/cutlass/line/http/AbstractLineHttpSender.java
puzpuzpuz: Oh, somehow I missed that part.

core/src/main/java/io/questdb/client/Sender.java
```diff
@@ -508,34 +542,30 @@ private LineSenderBuilder(int protocol) {
...
-                        .put("[address=").put(this.host).put("]");
-            }
             if (Chars.isBlank(address)) {
                 throw new LineSenderException("address cannot be empty nor null");
             }
             int portIndex = Chars.indexOf(address, ':');
             if (portIndex + 1 == address.length()) {
                 throw new LineSenderException("invalid address, use IPv4 address or a domain name [address=").put(address).put("]");
             }
+            String hostAndPort;
```
puzpuzpuz: The actual value seems to be host only, without port. Is that on purpose?

core/src/main/java/io/questdb/cutlass/http/client/HttpClient.java
nwoolmer: Quick question - when I first touched that code, I had walked in with the assumption that since we output an errno in the message, that the object itself also includes it. As Jaromir said, this was when I was relying in the number for control flow.

It might not matter here any more but it is a slightly strange API to me if we put a message including errno in an exception, but neglect actually populating the errno field.

core/src/main/java/io/questdb/cutlass/http/processors/LineHttpProcessorState.java
jerrinot: I am not sure here. I would keep 403 for permissions issues. 421 - a misdirected request sounds great to me. this is @nwoolmer's choice, but I find it very appropriate.

core/src/main/java/io/questdb/cutlass/http/client/HttpClient.java
jerrinot: @nwoolmer: `errno` is exposed to a user only in a message text. there is no user-accessible `getErrno()` or similar. I believe you original added this method to some exception, but I removed that. 

we might consider exposing this, but it must be done consistently. (and as a separate PR)

core/src/main/java/io/questdb/cutlass/http/processors/LineHttpProcessorState.java
puzpuzpuz: Misdirected request assumes that if you change the address, it may succeed while in reality this may be the only server available to you.

core/src/main/java/io/questdb/cutlass/line/http/AbstractLineHttpSender.java
jerrinot: this is a good one. it looks it's only used for error reporting -> we can do better.

core/src/main/java/io/questdb/client/Sender.java
jerrinot: no, I did a stupid thing.

core/src/main/java/io/questdb/cutlass/http/processors/LineHttpProcessorState.java
jerrinot: I think it's a matter a taste here, given it's marked as nit I am marking it as resolved. feel free to re-open if you feel strongly about it.

core/src/test/java/io/questdb/test/client/LineSenderBuilderTest.java
jerrinot: done

core/src/test/java/io/questdb/test/cutlass/http/line/LineHttpMultiUrlTest.java
jerrinot: after a discussion with @puzpuzpuz and @nwoolmer marking as resolved

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 161 / 181 (88.95%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/line/AbstractLineSender.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/LineHttpProcessorImpl.java|1|3|33.33%|
|:large_blue_circle:|io/questdb/cutlass/http/client/HttpClient.java|3|7|42.86%|
|:large_blue_circle:|io/questdb/cutlass/line/LineSenderException.java|8|9|88.89%|
|:large_blue_circle:|io/questdb/cutlass/line/http/AbstractLineHttpSender.java|99|111|89.19%|
|:large_blue_circle:|io/questdb/std/IntList.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/SettingsProcessor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/Sender.java|33|33|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/LineHttpProcessorState.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpResponseSink.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/http/LineHttpSenderV1.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpServer.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/http/LineHttpSenderV2.java|2|2|100.00%|

core/src/main/java/io/questdb/cutlass/line/LineSenderException.java
```diff
@@ -28,22 +28,55 @@
...
 
     private final StringSink message = new StringSink();
+    private final boolean retryable;
     private int errno = Integer.MIN_VALUE;
 
     public LineSenderException(CharSequence message) {
         this.message.put(message);
+        this.retryable = false;
+    }
+
```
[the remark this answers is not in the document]
jerrinot: irrelevant since only the HTTP transport supports retrying

core/src/main/java/io/questdb/cutlass/http/processors/SettingsProcessor.java
jerrinot: done

core/src/main/java/io/questdb/cairo/CairoConfiguration.java
jerrinot: fixed. there is a convoluted integration between OSS and Ent due to the initialization order. but that's something we can improve later.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 189 / 210 (90.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/line/AbstractLineSender.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cutlass/http/DefaultHttpServerConfiguration.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cutlass/http/client/HttpClient.java|3|7|42.86%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/LineHttpProcessorImpl.java|7|9|77.78%|
|:large_blue_circle:|io/questdb/cutlass/line/LineSenderException.java|8|9|88.89%|
|:large_blue_circle:|io/questdb/cutlass/line/http/AbstractLineHttpSender.java|101|113|89.38%|
|:large_blue_circle:|io/questdb/std/IntList.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/PropServerConfiguration.java|24|24|100.00%|
|:large_blue_circle:|io/questdb/client/Sender.java|33|33|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/LineHttpProcessorState.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpResponseSink.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/http/LineHttpSenderV1.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpServer.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpServerConfigurationWrapper.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/http/LineHttpSenderV2.java|2|2|100.00%|

core/src/main/java/io/questdb/cutlass/line/http/AbstractLineHttpSender.java
```diff
@@ -666,6 +788,15 @@ private HttpClient.Request newRequest() {
...
 
+    private void reset(long newFlushAfterNanos) {
+        pendingRows = 0;
+        flushAfterNanos = newFlushAfterNanos;
+        request = newRequest();
+    }
+
+    private void rotateAddress() {
+        currentAddressIndex = (currentAddressIndex + 1) % hosts.size();
+    }
```
nwoolmer, line 801: cc @bluestreak01 @amunra 

It has been decided that we should not just rotate urls and retry. Instead we should rotate urls and poll /settings on each, until we get one that is live and accepting writes. 

In the happy case, we would retry the buffer on a backup instance, which will be live and ready, and therefore require no extra network roundtrips.

However, if there are multiple instances which do not accept writes, a potentially large buffer must be sent to each, and the server will not short-circuit the response, which will cost time and bandwidth.

Therefore, /settings will be polled until an instance is found, and only then the buffer resent. This is what has been determined for the Rust client.

core/src/main/java/io/questdb/cutlass/line/http/AbstractLineHttpSender.java
```diff
@@ -185,40 +215,118 @@ public static AbstractLineHttpSender createLineSender(
...
+
+                    final String host = hosts.getQuick(currentAddressIndex);
+                    final int port = ports.getQuick(currentAddressIndex);
+                    try {
+                        HttpClient.Request req = cli.newRequest(host, port).GET().url(clientConfiguration.getSettingsPath());
+                        HttpClient.ResponseHeaders response = req.send();
+                        response.await();
+                        DirectUtf8Sequence statusCode = response.getStatusCode();
+                        if (isSuccessResponse(statusCode)) {
+                            try (JsonSettingsParser parser = new JsonSettingsParser()) {
```
puzpuzpuz: nit: can be move to the outer `try` block - we don't need to allocate a parser for each retry.

core/src/main/java/io/questdb/cutlass/line/http/AbstractLineHttpSender.java
```diff
@@ -320,6 +438,10 @@ public DirectByteSlice bufferView() {
         return bufferView.of(request.getContentStart(), request.getContentLength());
     }
 
+    public String buildUrl() {
```
puzpuzpuz: Unused method here.

core/src/main/java/io/questdb/cutlass/line/http/AbstractLineHttpSender.java
```diff
@@ -185,40 +215,118 @@ public static AbstractLineHttpSender createLineSender(
...
+                        response.await();
+                        DirectUtf8Sequence statusCode = response.getStatusCode();
+                        if (isSuccessResponse(statusCode)) {
+                            try (JsonSettingsParser parser = new JsonSettingsParser()) {
+                                parser.parse(response.getResponse());
+                                protocolVersion = parser.getDefaultProtocolVersion();
+                                if (parser.getMaxNameLen() != 0) {
+                                    maxNameLength = parser.getMaxNameLen();
+                                }
+                                if (parser.isAcceptingWrites()) {
```
puzpuzpuz: Shouldn't we move initialization of both `protocolVersion` and `maxNameLength` under this branch? Otherwise we may end up with protocol version that applies to a previously pinged server.

jerrinot: Assumption: The servers are homogeneous. If servers do not support the same protocol versions or the same `maxNameLength` then bad things may happen.

core/src/test/java/io/questdb/test/cutlass/http/line/LineHttpMultiUrlTest.java
```diff
@@ -0,0 +1,377 @@
...
+    private static Rnd rnd;
+
+    @BeforeClass
+    public static void setUpStatic() throws Exception {
+        AbstractBootstrapTest.setUpStatic();
+        rnd = TestUtils.generateRandom(LOG);
+    }
+
+    @Test
+    public void fuzzServersStartingAndStopping() throws Exception {
```
puzpuzpuz, line 45: @jerrinot this test takes 30+ seconds on my machine. Does it have to be this slow?

core/src/test/java/io/questdb/test/cutlass/http/line/LineHttpMultiUrlTest.java
```diff
@@ -0,0 +1,377 @@
...
+    public void testFirstServerIsDownTCP() {
+        try (Sender ignore = Sender.fromConfig("tcp::addr=" + HOST + ":" + PORT1 + ";addr=" + HOST + ":" + PORT2 + ";")) {
+            Assert.fail();
+        } catch (LineSenderException e) {
+            TestUtils.assertContains(e.getMessage(), "only a single address (host:port) is supported for TCP transport");
+        }
+    }
+
+    @Test
+    public void testServersAllDown() throws Exception {
```
puzpuzpuz, line 172: I've tried to add a test to cover at least some branches around `hostBlacklistBitmap` (currently it's mostly uncovered), but it takes 60+ seconds to run. The reason is that we don't have a config parameter for max retry timeout.

@jerrinot does it sound like a good idea to add this parameter and then add more tests?

```java
    @Test
    public void testManyServersAllDown() throws Exception {
        TestUtils.assertMemoryLeak(() -> {
            final int N = 65;
            final Sender.LineSenderBuilder senderBuilder = Sender.builder(Sender.Transport.HTTP).retryTimeoutMillis(60_000);
            for (int i = 0; i < N; i++) {
                senderBuilder.address(HOST).port(PORT1 + i);
            }
            try (Sender ignore = senderBuilder.build()) {
                Assert.fail();
            } catch (LineSenderException e) {
                TestUtils.assertContains(e.getMessage(), "Failed to detect server line protocol version");
            }
        });
    }
```

jerrinot: Do you know if other clients allow that? 
Another option could be to refactor `backoff()` to some kind of `BackoffStrategy` - which would not sleep at all. WDYT?

puzpuzpuz: > Do you know if other clients allow that?

I don't think so. But we need this parameter at least for testing, so it could be undocumented & non-public.

> Another option could be to refactor `backoff()` to some kind of `BackoffStrategy` - which would not sleep at all.

Yes, that would also work.

core/src/main/java/io/questdb/cutlass/line/http/AbstractLineHttpSender.java
jerrinot: we decided to keep it as it is.

core/src/test/java/io/questdb/test/cutlass/http/line/LineHttpMultiUrlTest.java
jerrinot: down to 7s.

core/src/test/java/io/questdb/test/cutlass/http/line/LineHttpMultiUrlTest.java
jerrinot: wdyt about https://github.com/questdb/questdb/commit/8049b517903de42fa3c5dcef75200792eb22f8c1 ?

puzpuzpuz: Looks good. Thanks for going an extra mile improving the tests!

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 225 / 250 (90.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/line/AbstractLineSender.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cutlass/http/DefaultHttpServerConfiguration.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cutlass/http/client/HttpClient.java|3|7|42.86%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/LineHttpProcessorImpl.java|7|9|77.78%|
|:large_blue_circle:|io/questdb/cutlass/line/LineSenderException.java|8|9|88.89%|
|:large_blue_circle:|io/questdb/client/Sender.java|41|46|89.13%|
|:large_blue_circle:|io/questdb/cutlass/line/http/AbstractLineHttpSender.java|129|140|92.14%|
|:large_blue_circle:|io/questdb/std/IntList.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/PropServerConfiguration.java|24|24|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/LineHttpProcessorState.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpResponseSink.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/http/LineHttpSenderV1.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpServer.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpServerConfigurationWrapper.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/http/LineHttpSenderV2.java|2|2|100.00%|

Issue #6062, opened 2025-08-18
flaky test: LineTcpReceiverDropTableFuzzTest.testInsertDropParallel()


jerrinot:
### To reproduce

logs: https://gist.github.com/jerrinot/c3e7d59d0d5e5e362078ab0172b375b9

Window CI run. 
Possible related to https://github.com/questdb/questdb/issues/5803

### QuestDB version:

d97717fb

### OS, in case of Docker specify Docker and the Host OS:

Windows CI

### File System, in case of Docker specify Host File System:

irrelevant

### Full Name:

Jaromir Hamala

### Affiliation:

QuestDB

### Have you followed Linux, MacOs kernel configuration steps to increase Maximum open files and Maximum virtual memory areas limit?

- [x] Yes, I have

jerrinot:
this particular test is quite annoying, failing frequently: 

<img width="1230" height="1586" alt="Image" src="https://github.com/user-attachments/assets/92c7af2a-d000-44d2-a8ea-d3bcd492e899" />

fuzziecoder:
`// Before: QuestDB tried to directly cast VARCHAR to DOUBLE[],
// which fails on Windows CI randomly due to parsing inconsistencies.

// After: Explicitly parse VARCHAR into an array of doubles before inserting.

private static double[] parseDoubleArrayFromVarchar(CharSequence text) {
    if (text == null || text.length() == 0) {
        return new double[0];
    }
    // Expected format: "{1.0, 2.5, 3.14}"
    String s = text.toString().trim();
    if (s.startsWith("{") && s.endsWith("}")) {
        s = s.substring(1, s.length() - 1);
    }
    String[] parts = s.split(",");
    double[] result = new double[parts.length];
    for (int i = 0; i < parts.length; i++) {
        result[i] = Double.parseDouble(parts[i].trim());
    }
    return result;
}
`

jerrinot:
@fuzziecoder what are you trying to say?

fuzziecoder:
Bro seems i can't push

Sheenam015:
Please assign this issue to me.

Issue #1841, opened 2022-01-28
Flapping test NetTest.testNoLinger


ideoma:
### Describe the bug

Mac OSX, ci server

```
2022-01-28T13:44:44.718092Z E i.q.TestListener ***** Test Failed *****io.questdb.network.NetTest.testNoLinger : 
java.lang.AssertionError: expected:<0> but was:<-1>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.junit.Assert.assertEquals(Assert.java:633)
	at io.questdb.network.NetTest.bindAcceptConnectClose(NetTest.java:138)
	at io.questdb.network.NetTest.testNoLinger(NetTest.java:51)
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
[...]

PR #6082, opened 2025-08-28
fix(core): fix a potential database crash when under a load of a prolonged period of time

jerrinot:
In long-running nodes under sustained write load, an internal file-descriptor cache could wrap its monotonic counter and misassociate cached FDs. In unlucky timing, this could cascade into a crash. This PR hardens the FD cache against wrap-around and tightens memory-mapping error handling.


core/src/main/java/io/questdb/cairo/wal/WalTxnDetails.java
```diff
@@ -740,6 +740,7 @@ public static int loadTxns(TransactionLogCursor transactionLogCursor, int txnCou
             txnList.add(long1);
             txnList.add(Numbers.encodeLowHighInts(transactionLogCursor.getSegmentTxn(), txn));
         }
+        assert txn > 0;
```
puzpuzpuz: I'd better make this into a proper check, not an assertion:
```java
if (txn > 0) {
    Vect.radixSortLongIndexAscChecked(
    //...
}
```

Alternatively, we can throw `CairoException`.

core/src/main/java/io/questdb/cairo/wal/seq/TableTransactionLogV1.java
```diff
@@ -369,7 +370,11 @@ private TransactionLogCursorImpl of(FilesFacade ff, long txnLo, Path path) {
             long newTxnCount = ff.readNonNegativeLong(fd, MAX_TXN_OFFSET_64);
             if (newTxnCount > -1L) {
                 this.txnCount = newTxnCount;
-                this.address = ff.mmap(fd, getMappedLen(), 0, Files.MAP_RO, MemoryTag.MMAP_TX_LOG_CURSOR);
+                long newAddr = ff.mmap(fd, getMappedLen(), 0, Files.MAP_RO, MemoryTag.MMAP_TX_LOG_CURSOR);
+                if (newAddr == -1) {
+                    throw CairoException.critical(Os.errno()).put("cannot mmap transaction log [path=").put(path).put(']');
```
puzpuzpuz, line 375: Why don't we move this check into `FilesFacadeImpl` itself? It already throws `CairoException` in many other syscalls.

jerrinot: I am not brave enough for such change at the last minute before release. are you? :)

core/src/main/java/io/questdb/cairo/wal/WalTxnDetails.java
jerrinot: it's an invariant check. I do not expect it to fail. only a bug can trigger it. we use assert in such cases. wdyt?

core/src/main/java/io/questdb/std/FdCache.java
```diff
@@ -177,8 +198,16 @@ public synchronized long openROCached(LPSZ lpsz) {
...
 
         holder.count++;
-        long uniqROFd = createUniqueFdRO(holder.osFd);
-        openFdMapByFd.put(uniqROFd, holder);
+
+        // find a unique key even if the index generator wrapped over
+        long uniqROFd;
+        int keyIndex;
+        do {
+            uniqROFd = Numbers.encodeLowHighInts(nextIndex() | RO_MASK, holder.osFd);
```
puzpuzpuz: `RO_MASK` is `0`, so `nextIndex() | RO_MASK` operation is useless, no?

jerrinot: that's right. not sure why it was there in the first plac.e I guess for consistency?

core/src/main/java/io/questdb/std/FdCache.java
```diff
@@ -274,7 +298,9 @@ private FdCacheRecord getFdCacheRecord(LPSZ lpsz) {
             } else {
                 OPEN_OS_FILE_COUNT.incrementAndGet();
                 Utf8String path = Utf8String.newInstance(lpsz);
-                holder = createFdCacheRecord(path, Numbers.encodeLowHighInts(fdCounter.incrementAndGet(), osFd));
+                // todo: explore if potentially dup keys are an issue here
+                // it appears to me that it's OK here
+                holder = createFdCacheRecord(path, Numbers.encodeLowHighInts(nextIndex(), osFd));
```
puzpuzpuz: This method is used in `openROCached`, so probably it should use `createUniqueFdRO` unless I'm missing something.

core/src/main/java/io/questdb/cairo/wal/WalTxnDetails.java
puzpuzpuz: Assertions may be disabled and then we'll get the old behavior. Having this assert is better than nothing though, so treat this as a nit.

core/src/main/java/io/questdb/cairo/wal/seq/TableTransactionLogV1.java
puzpuzpuz: As long as we have fuzz tests that simulate fs failures and we see that they don't crash the process, I'm brave enough. :)

Jokes aside, this is more like the next step after we make the hot fix.

core/src/main/java/io/questdb/std/FdCache.java
puzpuzpuz: No idea. Could be a last minute change. We don't use the mask anywhere except for this place, so maybe we could get rid of it.

core/src/main/java/io/questdb/cairo/wal/seq/TableTransactionLogV1.java
jerrinot: created an item at: https://github.com/questdb/support/issues/19

core/src/main/java/io/questdb/std/FdCache.java
jerrinot: it is. but this ID is not used as a key in `openFdMapByFd` 
this ID is used by MmapCache. It deserves a thorough review, as I am not yet certain it's completely innocent.

puzpuzpuz: It seems wasteful to increment `fdCounter` up to two times in `openROCached()`: once for `mmapCacheFd`, then then for `openFdMapByFd`.

puzpuzpuz:
@jerrinot we could expose test-only `setFDCounter()` method to set the counter value and add tests around overflow behavior. Otherwise, there are no tests that verify the behavior around overflow.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 36 / 38 (94.74%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/wal/seq/TableTransactionLogV1.java|6|8|75.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalTxnDetails.java|14|14|100.00%|
|:large_blue_circle:|io/questdb/std/FdCache.java|16|16|100.00%|

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 37 / 39 (94.87%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/wal/seq/TableTransactionLogV1.java|6|8|75.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalTxnDetails.java|14|14|100.00%|
|:large_blue_circle:|io/questdb/std/MmapCache.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/FdCache.java|16|16|100.00%|

core/src/main/java/io/questdb/std/MmapCache.java
```diff
@@ -56,6 +56,7 @@ public long cacheMmap(int fd, long fileCacheKey, long len, long offset, int flag
             if (fdMapIndex < 0) {
                 MmapCacheRecord record = mmapFileCache.valueAt(fdMapIndex);
                 if (record.length >= len) {
+                    assert record.count > 0 : "found a record with zero reference count in mmap cache";
```
puzpuzpuz: nit: why don't we include some diagnostic info into the message, e.g. the fd?

jerrinot:
an unrelated failure due to https://github.com/questdb/questdb/issues/1841

core/src/main/java/io/questdb/std/FdCache.java
jerrinot: yeah, it is. but I am not really keen to do any non-local changes at this point.

core/src/main/java/io/questdb/std/MmapCache.java
jerrinot: good idea. done.

mtopolnik:
> we could expose test-only `setFDCounter()` method to set the counter value and add tests around overflow behavior. Otherwise, there are no tests that verify the behavior around overflow.

This is now done, four tests added that directly wrap-around the FD counter and test file opening, closing and reading. All confirmed to fail on master.

core/src/test/java/io/questdb/test/std/FdCacheCounterOverflowTest.java
```diff
@@ -0,0 +1,263 @@
...
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.util.Arrays;
+
+import static io.questdb.test.tools.TestUtils.assertMemoryLeak;
+
+public class FdCacheCounterOverflowTest extends AbstractTest {
+    private static final byte CONTENT_A = 'A';
+    private static final byte CONTENT_B = 'B';
+    private static final int FILE_SIZE = 1024;
```
puzpuzpuz, line 50: This is even less than a single page which significantly lowers the probability of observing segfaults. Let's increase the file size.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 63 / 66 (95.45%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/wal/seq/TableTransactionLogV1.java|6|8|75.00%|
|:large_blue_circle:|io/questdb/std/MmapCache.java|13|14|92.86%|
|:large_blue_circle:|io/questdb/cairo/wal/WalTxnDetails.java|14|14|100.00%|
|:large_blue_circle:|io/questdb/std/Files.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/std/FdCache.java|24|24|100.00%|

core/src/test/java/io/questdb/test/std/FdCacheCounterOverflowTest.java
puzpuzpuz: Sorry, initially I thought that there are concurrent tests here. But for single-threaded ones, the file size doesn't matter, so I'm marking this one as resolved.

jerrinot:
@coderabbit review

core/src/test/java/io/questdb/test/std/FdCacheCounterOverflowTest.java
```diff
@@ -0,0 +1,355 @@
...
+        assertMemoryLeak(() -> {
+            final int numReaderThreads = 4;
+            final int testDurationMs = 5000;
+            final Thread[] readerThreads = new Thread[numReaderThreads];
+            final Throwable[] exceptions = new Throwable[numReaderThreads + 1];
+            final AtomicBoolean shouldStop = new AtomicBoolean(false);
+
+            // Writer thread - continuously appends to file A
+            Thread writerThread = new Thread(() -> {
+                byte[] appendData = new byte[16];
```
puzpuzpuz, line 87: nit: why don't we use mmap for writes too to keep the scenario closer to what we have in the code base?

core/src/test/java/io/questdb/test/std/FdCacheCounterOverflowTest.java
```diff
@@ -0,0 +1,355 @@
...
+                            long readFd = -1;
+                            try {
+                                readFd = openRO(testFileA);
+                                long size = Files.length(readFd);
+                                long addr = Files.mmap(readFd, size, 0, Files.MAP_RO, MemoryTag.MMAP_DEFAULT);
+                                Assert.assertTrue("Memory mapping failed", addr > 0);
+                                // Read the last byte to verify file content
+                                byte content = Unsafe.getUnsafe().getByte(addr + size - 1);
+                                Assert.assertEquals("File content mismatch in reader thread " + threadIndex, CONTENT_A, content);
+                                Files.munmap(addr, 1, MemoryTag.MMAP_DEFAULT);
```
puzpuzpuz: nit: munmap should be in the `finally` block

core/src/test/java/io/questdb/test/std/FdCacheCounterOverflowTest.java
```diff
@@ -0,0 +1,355 @@
...
+    @Test
+    public void testConcurrentOpenClose() throws Exception {
+        // Concurrent test that spawns a single writer that writes an append-only file, and multiple workers.
+        // It sets the FD counter near the overflow initially, and the reader threads actively open/read/close the file.
+
+        Files.setFDCacheCounter(Integer.MAX_VALUE - 100);
+
+        assertMemoryLeak(() -> {
+            final int numReaderThreads = 4;
+            final int testDurationMs = 5000;
```
puzpuzpuz: nit: 5 seconds seems to be too much, why don't we just do a fixed number of iterations in the threads like we usually do in concurrent tests? It should be sufficient to reproduce the segfault.

core/src/test/java/io/questdb/test/std/FdCacheCounterOverflowTest.java
```diff
@@ -0,0 +1,355 @@
...
+                Arrays.fill(appendData, CONTENT_A);
+                long buf = Unsafe.getUnsafe().allocateMemory(appendData.length);
+                Unsafe.getUnsafe().copyMemory(appendData, sun.misc.Unsafe.ARRAY_BYTE_BASE_OFFSET, null, buf, appendData.length);
+
+                long appendFd = -1;
+                try {
+                    appendFd = Files.openAppend(testFileA.$());
+                    Assert.assertTrue("Failed to open file for appending", appendFd > -1);
+                    while (!shouldStop.get()) {
+                        Files.append(appendFd, buf, appendData.length);
```
puzpuzpuz: We need to check the result of `Files.append`

puzpuzpuz:
@coderabbitai review

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 63 / 66 (95.45%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/wal/seq/TableTransactionLogV1.java|6|8|75.00%|
|:large_blue_circle:|io/questdb/std/MmapCache.java|13|14|92.86%|
|:large_blue_circle:|io/questdb/cairo/wal/WalTxnDetails.java|14|14|100.00%|
|:large_blue_circle:|io/questdb/std/Files.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/std/FdCache.java|24|24|100.00%|

core/src/test/java/io/questdb/test/std/FdCacheCounterOverflowTest.java
```diff
@@ -0,0 +1,362 @@
...
+        long addr = Files.mmap(fd1, FILE_SIZE, 0, Files.MAP_RO, MemoryTag.MMAP_DEFAULT);
+        Assert.assertTrue("failed to mmap", addr > 0);
+        try {
+            for (int i = 0; i < FILE_SIZE; i++) {
+                Assert.assertEquals(expectedContent, Unsafe.getUnsafe().getByte(addr + i));
+            }
+        } finally {
+            Files.munmap(addr, FILE_SIZE, MemoryTag.MMAP_DEFAULT);
+        }
+    }
```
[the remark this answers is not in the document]
mtopolnik: Won't fix

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 63 / 66 (95.45%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/wal/seq/TableTransactionLogV1.java|6|8|75.00%|
|:large_blue_circle:|io/questdb/std/MmapCache.java|13|14|92.86%|
|:large_blue_circle:|io/questdb/cairo/wal/WalTxnDetails.java|14|14|100.00%|
|:large_blue_circle:|io/questdb/std/Files.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/std/FdCache.java|24|24|100.00%|

jerrinot:
a failure due to https://github.com/questdb/questdb/issues/6062

mtopolnik (approved):

PR #6091, opened 2025-09-01
test(sql): fix flaky FuzzValidateSymbolFilterOperation

jerrinot:
Fuzz tests use a fault-injector to simulate file system failures.

This means `engine.getTableMetadata(tableToken)`
can throw `io.questdb.cairo.CairoException: [316] could not open read-only [file=/var/folders/71/5myfy97j1pqgbz_z1v70yb2r0000gn/T/junit5538203268730321834/dbRoot/db/testManyTransactions_0~12/_meta]`

This exception causes `FuzzValidateSymbolFilterOperation` to fail and this fails the whole fuzz test. 

This change is making the operation more lenient.
The point of the operation is to validate that when a query execution if successful and returns results, then their results are correct. We don't care about unsuccessful attempts (caused by failure injector)

### DIFF
--- a/core/src/test/java/io/questdb/test/fuzz/FuzzValidateSymbolFilterOperation.java
+++ b/core/src/test/java/io/questdb/test/fuzz/FuzzValidateSymbolFilterOperation.java
@@ -32,6 +32,7 @@ import io.questdb.cairo.sql.RecordCursor;
 import io.questdb.cairo.sql.RecordCursorFactory;
 import io.questdb.cairo.sql.TableRecordMetadata;
 import io.questdb.griffin.SqlCompiler;
+import io.questdb.griffin.SqlException;
 import io.questdb.griffin.SqlExecutionContext;
 import io.questdb.griffin.SqlExecutionContextImpl;
 import io.questdb.std.Chars;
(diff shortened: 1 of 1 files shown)


questdb-butler:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Fails</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Please update the PR title to match this format:
`type(subType): description`

Where `type` is one of:
`feat`, `fix`, `chore`, `docs`, `style`, `refactor`, `perf`, `test`, `ci`, `revert`

And: `subType` is one of:
`build`, `sql`, `log`, `mig`, `core`, `ilp`, `pgwire`, `http`, `conf`, `ui`, `wal`, `parquet`, `utils`

For Example:

```
perf(sql): improve pattern matching performance for SELECT sub-queries
```
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 746fb95f87685c4e84b482e96d8f27822c0d0a0f
</p>

PR #6095, opened 2025-09-01
fix(core): fix table reading timeout error after non-wal table dropped and re-created

ideoma:
Found by fuzz tests.
When a non-WAL table is dropped and re-created, purge jobs can push max txn using `TxnScoreboardPoolV2.isRangeAvailable()` and it can lead to a timeout on opening TableReader, where it cannot lock the latest transaction in the scoreboard.

The fix is not to modify the max in `TxnScoreboardPoolV2.isRangeAvailable()`, making it a read-only scoreboard operation.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TxnScoreboardV2.java
+++ b/core/src/main/java/io/questdb/cairo/TxnScoreboardV2.java
@@ -224,33 +224,34 @@ public class TxnScoreboardV2 implements TxnScoreboard {
 
     @Override
     public boolean isRangeAvailable(long fromTxn, long toTxn) {
-        // Push max txn to the latest, to avoid races with acquireTxn()
-        // but don't stop checking if it's not the max.
-        updateMax(toTxn);
-
         if (getActiveReaderCount() == 0) {
             return true;
         }
 
-        for (int i = 0; i < bitmapCount; i++) {
-            long bitmap = Unsafe.getUnsafe().getLongVolatile(null, bitmapMem + (long) i * Long.BYTES);
-            if (bitmap == 0) {
-                continue;
-            }
-
-            int base = i * Long.SIZE;
-            while (bitmap != 0) {
-                final long lowestBit = Long.lowestOneBit(bitmap);
-                final int bit = Long.numberOfTrailingZeros(lowestBit);
-                int internalId = base + bit;
-                long lockedTxn = Unsafe.getUnsafe().getLongVolatile(null, entriesMem + (long) internalId * Long.BYTES);
-                if (lockedTxn > UNLOCKED && lockedTxn >= fromTxn && lockedTxn < toTxn) {
-                    return false;
+        long max;
+        do {
+            max = getMax();
+            for (int i = 0; i < bitmapCount; i++) {
+                long bitmap = Unsafe.getUnsafe().getLongVolatile(null, bitmapMem + (long) i * Long.BYTES);
+                if (bitmap == 0) {
+                    continue;
                 }
 
-                bitmap ^= lowestBit;
+                int base = i * Long.SIZE;
+                while (bitmap != 0) {
+                    final long lowestBit = Long.lowestOneBit(bitmap);
+                    final int bit = Long.numberOfTrailingZeros(lowestBit);
+                    int internalId = base + bit;
+                    long lockedTxn = Unsafe.getUnsafe().getLongVolatile(null, entriesMem + (long) internalId * Long.BYTES);
+                    if (lockedTxn > UNLOCKED && lockedTxn >= fromTxn && lockedTxn < toTxn) {
+                        return false;
+                    }
+
+                    bitmap ^= lowestBit;
+                }
             }
-        }
+            // max changed, need to re-check
+        } while (max <= toTxn && max != getMax());
         return true;
     }
 


core/src/main/java/io/questdb/cairo/TxnScoreboardV2.java
```diff
@@ -224,33 +224,34 @@ public boolean isOutdated(long txn) {
...
+                    if (lockedTxn > UNLOCKED && lockedTxn >= fromTxn && lockedTxn < toTxn) {
+                        return false;
+                    }
+
+                    bitmap ^= lowestBit;
+                }
             }
-        }
+            // max changed, need to re-check
+        } while (max <= toTxn && max != getMax());
```
puzpuzpuz: Don't we break this check in `acquireTxn()` with this change?
```java
        if (getMax() > txn) {
            // Max moved, cannot acquire the txn.
            // ...
            return false;
        }
```

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TxnScoreboardV2.java|16|16|100.00%|

core/src/main/java/io/questdb/cairo/TxnScoreboardV2.java
ideoma: Ok, I changed the fix entirely by making the changes in Column Purge Job instead of the scoreboard, so this is reverted. Can you please have another look?

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/ColumnPurgeOperator.java|6|6|100.00%|

core/src/main/java/io/questdb/cairo/TxnScoreboardV2.java
puzpuzpuz: Sure thing. BTW we can still make `isRangeAvailable()` a read-only operation if we introduce a seqlock-style synchronization: writers (`acquireTxn()` callers) will first increment the sequence to an odd number and, once they write stuff, to an even number. On the other hand, readers (`isRangeAvailable()` caller) will check the number before and after they read txns from the range - this way they will never read data while a writer writes it.

core/src/main/java/io/questdb/cairo/CairoEngine.java
```diff
@@ -585,6 +584,11 @@ public void dropTableOrMatView(@Transient Path path, TableToken tableToken) {
                 }
 
                 tableNameRegistry.dropTable(tableToken);
+                // Remove the scoreboard after dropping the table from the registry
+                // Otherwise someone (like Column Purge Job) can create pooled instances of the scoreboard
+                // it from the registry without knowning that the table is being dropped.
```
puzpuzpuz: nit:
```suggestion
                // it from the registry without knowing that the table is being dropped.
```

core/src/main/java/io/questdb/cairo/ColumnPurgeOperator.java
```diff
@@ -179,8 +179,13 @@ private boolean openScoreboardAndTxn(ColumnPurgeTask task) {
                 txnScoreboard = engine.getTxnScoreboard(task.getTableName());
                 // fall through
             case STARTUP_ONLY:
-                int tableId = readTableId(path);
-                if (tableId != task.getTableId()) {
+                TableToken updatedTableToken = engine.getTableTokenIfExists(task.getTableName().getTableName());
```
puzpuzpuz: nit: `task.getTableName().getTableName()` reads really weird. Would be nice to rename `ColumnPurgeTask#getTableName()` into `getTableToken()`.

puzpuzpuz:
> When a non-WAL table is dropped and re-created, purge jobs can push max txn using `TxnScoreboardPoolV2.isRangeAvailable()` and it can lead to a timeout on opening TableReader, where it cannot lock the latest transaction in the scoreboard.

Why exactly table readers time out? Aren't they supposed to spin with reading the latest txn and then trying to acquire it? Or somehow we end up with the max txn being incremented many times in this scenario?

ideoma:
> Why exactly table readers time out? Aren't they supposed to spin with reading the latest txn and then trying to acquire it? Or somehow we end up with the max txn being incremented many times in this scenario?

Purge job calls isRangeAvailable(0,13) because it tries to clean up dropped table with the same dir name (non-wal table), this pushes max txn to 13 in the score board. Readers then timeout because the spin to acquire txn 0 read from the _txn file and they cannot, because max is 13.

puzpuzpuz:
> because it tries to clean up dropped table with the same dir name (non-wal table)

Gotcha, that's the culprit!

puzpuzpuz (dismissed):

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 20 / 22 (90.91%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/tasks/ColumnPurgeTask.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/cairo/ColumnPurgeOperator.java|10|11|90.91%|
|:large_blue_circle:|io/questdb/cairo/VacuumColumnVersions.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/PurgingOperator.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/ColumnPurgeJob.java|1|1|100.00%|

PR #6104, opened 2025-09-04
fix(core): fix spurious "cannot mmap" errors

bluestreak01:
found by fuzz tests

This test after 4-6 mins running on Windows

```java
    @Test
    public void testStressWalPurgeJob() throws Exception {
        // Here we generate many WAL segments and run WalPurgeJob frequently.
        // The goal is to make sure WalPurgeJob doesn't delete WAL-E files used by MatViewRefreshJob.
        setProperty(PropertyKey.CAIRO_WAL_SEGMENT_ROLLOVER_ROW_COUNT, 10);
        setProperty(PropertyKey.CAIRO_WAL_PURGE_INTERVAL, 10);
        assertMemoryLeak(() -> {
            Rnd rnd = fuzzer.generateRandom(LOG, 624968545800L, 1756738649271L);
            setFuzzParams(rnd, 0, 0);
            setFuzzProperties(rnd);
            runMvFuzz(rnd, getTestName(), 4);
        });
    }
```

-------------

Unrelated test failure:

```
2025-09-04T21:23:48.284128Z I i.q.t.TestListener >>>> io.questdb.test.cutlass.line.tcp.LineTcpReceiverTest.testGoodAuthenticated[WITH_WAL]
>>>>= io.questdb.test.cutlass.line.tcp.LineTcpReceiverTest.testGoodAuthenticated[WITH_WAL]
2025-09-04T21:23:48.341866Z I i.q.t.AbstractTest Starting test LineTcpReceiverTest#testGoodAuthenticated[WITH_WAL]
2025-09-04T21:23:48.352689Z I i.q.c.TableNameRegistryStore reloading tables file [path=C:\Users\VSSADM~1\AppData\Local\Temp\junit15334152461061910086\dbRoot\tables.d.0, threadId=4584]
2025-09-04T21:23:48.356254Z A IODispatcher listening on 0.0.0.0:9002 [fd=35564477030591 backlog=64]
2025-09-04T21:23:48.365008Z I i.q.c.t.t.InputFormatConfiguration loading input format config [resource=/text_loader.json]
2025-09-04T21:23:48.367900Z I i.q.t.c.l.t.LineTcpReceiverTest worker pool started [pool=testing]
2025-09-04T21:23:48.379665Z I IODispatcher connected [ip=127.0.0.1, fd=25703232118978, connCount=2]
2025-09-04T21:23:48.379954Z E i.q.c.l.t.LineTcpConnectionContext [25703232118978] could not parse measurement, NO_FIELDS at 9, line (may be mangled due to partial parsing): 'testUser1'
2025-09-04T21:23:48.412980Z I i.q.t.c.l.t.LineTcpReceiverTest os scheduled worker started [name=testing_0]
```


bluestreak01:
@coderabbit review

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/SymbolMapReaderImpl.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/vm/MemoryCMRImpl.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractIndexReader.java|7|7|100.00%|

puzpuzpuz:
The PR title mentions mat views while the issue is not related with mat views, so I'd change the title - any table reader could fail on windows, not only readers open by mat view refresh.

core/src/main/java/io/questdb/cairo/AbstractIndexReader.java
```diff
@@ -120,16 +122,19 @@ public void of(
...
 
         try {
-            keyMem.wholeFile(configuration.getFilesFacade(), BitmapIndexUtils.keyFileName(path, columnName, columnNameTxn), MemoryTag.MMAP_INDEX_READER);
+            FilesFacade ff = configuration.getFilesFacade();
+            LPSZ name = BitmapIndexUtils.keyFileName(path, columnName, columnNameTxn);
+            keyMem.of(
+                    ff,
+                    name,
+                    ff.getMapPageSize(),
+                    BitmapIndexUtils.getKeyEntryOffset(this.keyCount),
```
puzpuzpuz: `this.keyCount` seems to be always 0 at this point, so it could be just `BitmapIndexUtils.getKeyEntryOffset(0)`

core/src/main/java/io/questdb/cairo/AbstractIndexReader.java
```diff
@@ -120,16 +122,19 @@ public void of(
         this.spinLockTimeoutMs = configuration.getSpinLockTimeout();
 
         try {
-            keyMem.wholeFile(configuration.getFilesFacade(), BitmapIndexUtils.keyFileName(path, columnName, columnNameTxn), MemoryTag.MMAP_INDEX_READER);
+            FilesFacade ff = configuration.getFilesFacade();
+            LPSZ name = BitmapIndexUtils.keyFileName(path, columnName, columnNameTxn);
+            keyMem.of(
```
puzpuzpuz, line 127: With this approach we're guaranteed to do a `mmap()` followed by an `mremap()` done by `readIndexMetadataAtomically()`. With the old code we'd only do a single `mmap()` call in scenarios when there are no concurrent writes. To keep this fast path, we could first read the (potentially obsolete, but guaranteed not to decrement) key count with `ff.readIntAsUnsignedLong()` and use it for the `mmap()` call. This way the subsequent `mremap()` will be skipped when there are no concurrent symbol table writes.

core/src/main/java/io/questdb/cairo/AbstractIndexReader.java
```diff
@@ -200,12 +203,13 @@ public void updateKeyCount() {
...
             this.keyCount = keyCount;
             this.keyCountIncludingNulls = columnTop > 0 ? keyCount + 1 : keyCount;
+            keyMem.extend(BitmapIndexUtils.getKeyEntryOffset(keyCount));
         }
     }
 
     private void readIndexMetadataAtomically() {
-        // Triple check atomic read. We read first and last sequences. If they match - there is a chance at stable
-        // read. Confirm start sequence hasn't changed after values read. If it has changed - retry the whole thing.
+        // Triple check atomic read. We read the first and last sequences. If they match - there is a chance at stable
```
puzpuzpuz, line 211: nit: this triple check could be more lightweight and less complex. In principle, a [seqlock](https://en.wikipedia.org/wiki/Seqlock) (a well-known synchronization mechanism used in Linux kernel) would be sufficient here: the writer uses a single sequence number (instead of two: SEQUENCE and SEQUENCE_CHECK). When the writer starts a write it increments the sequence to an odd value and when it writes the new values, it increments it to an even value. The readers check the sequence value before and after the write and accept the atomic read only when the value is even and it hasn't changed - only two atomic loads of the same `long` are required.

core/src/main/java/io/questdb/cairo/vm/api/MemoryM.java
```diff
@@ -94,8 +94,4 @@ default void of(FilesFacade ff, LPSZ name, long extendSegmentSize, long size, in
     default void smallFile(FilesFacade ff, LPSZ name, int memoryTag) {
         of(ff, name, ff.getPageSize(), ff.length(name), memoryTag, CairoConfiguration.O_NONE, -1);
     }
-
-    default void wholeFile(FilesFacade ff, LPSZ name, int memoryTag) {
```
puzpuzpuz, line 98: There is still `Vm#getWholeMARWInstance()` and `getSmallCMARWInstance()` which uses `-1` size. It's used, for example, by `SymbolMapWriter` when mmapping char file. On the other hand, symbol writers (and other writers that seem to be using thse methods) won't suffer from the OG bug, but it's worth adding a disclamer that the methods must not be used by readers of any kind. WDYT?

puzpuzpuz, line 98: We still have some readers/utilities that rely on `ff.length()` when mmapping a file: `TxReader#openTxnFile()`, `TableUtils#readTableName()`. Worth checking it.

core/src/main/java/io/questdb/cairo/AbstractIndexReader.java
bluestreak01: this would be relatively minor optimisation that doesn't yet solve anyone's problem. I would keep it out of this PR

core/src/main/java/io/questdb/cairo/AbstractIndexReader.java
```diff
@@ -120,16 +122,19 @@ public void of(
...
+            keyMem.of(
+                    ff,
+                    name,
+                    ff.getMapPageSize(),
+                    BitmapIndexUtils.getKeyEntryOffset(this.keyCount),
+                    MemoryTag.MMAP_INDEX_READER,
+                    CairoConfiguration.O_NONE,
+                    -1
+            );
             this.clock = configuration.getMillisecondClock();
```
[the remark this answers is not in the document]
bluestreak01: done in https://github.com/questdb/questdb/pull/6104/commits/1372698fa2e87090f5a137222533b023cabb676a

core/src/main/java/io/questdb/cairo/AbstractIndexReader.java
bluestreak01: done in https://github.com/questdb/questdb/pull/6104/commits/1372698fa2e87090f5a137222533b023cabb676a

core/src/main/java/io/questdb/cairo/AbstractIndexReader.java
bluestreak01: we cannot guarantee memory does not shrink. We have to use the header to read the file. Truncate may come in the middle of reader reload, mmap will fail.

puzpuzpuz: > Truncate may come in the middle of reader reload

What kind of truncate is that? If that's table truncate and it happens concurrently with table reader initialization or just reads, we'll get a segfault/sigbus even with this patch.

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/SymbolMapReaderImpl.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/TxReader.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/vm/MemoryCMRImpl.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractIndexReader.java|7|7|100.00%|

PR #6106, opened 2025-09-06
fix(core): serialize limit's constant functions correctly

### LINKED ISSUE #5555: EXPLAIN (format JSON) is broken in a few cases
### To reproduce

```sql
EXPLAIN (format JSON) (trades LIMIT 1000) INNER JOIN (trades LIMIT 1000) ON (symbol);
```

This gives broken JSON. In general, this plan output is dodgy. In this case, it is related to `IntConstant` being a plannable, but then written as an integer and the quote left unclosed.

```
[
  {
    "Plan": {
        "Node Type": "SelectedRecord",
        "Plans": [
        {
            "Node Type": "Hash Join Light",
            "condition": "_xQdbA2.symbol=_xQdbA1.symbol",
            "Plans": [
            {
                "Node Type": "Limit",
                "lo": "1000,
                "skip-over-rows":  0,
                "limit":  1000,
                "Plans": [
                {
                    "Node Type": "PageFrame",
                    "Plans": [
                    {
                        "Node Type": "Row forward scan"
                    },
                    {
                        "Node Type": "Frame forward scan",
                        "on": "trades"
                    } ]
                } ]
            },
            {
                "Node Type": "Hash",
                "Plans": [
                {
                    "Node Type": "Limit",
                    "lo": "1000,
                    "skip-over-rows":  0,
                    "limit":  1000,
                    "Plans": [
                    {
                        "Node Type": "PageFrame",
                        "Plans": [
                        {
                            "Node Type": "Row forward scan"
                        },
                        {
                            "Node Type": "Frame forward scan",
                            "on": "trades"
                        } ]
                    } ]
            } ]
        } ]
    }
  }
]
```

The brackets also don't match up. Plus the overall structure of the plan is wasteful, between `Plan` and repeated `Plans`.

### QuestDB version:

Master

N/A

N/A

### Full Name:

Nick Woolmer

### Affiliation:

QuestDB

- [x] Yes, I have

### Additional context

_No response_

lost-particles:
Author: Subham Kumar Das

Fixes #5555 Issue by sending the corresponding int value, if the Limit lo/hi function is a instance of IntConstant

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/LimitRecordCursorFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/LimitRecordCursorFactory.java
@@ -34,6 +34,7 @@ import io.questdb.cairo.sql.SymbolTable;
 import io.questdb.griffin.PlanSink;
 import io.questdb.griffin.SqlException;
 import io.questdb.griffin.SqlExecutionContext;
+import io.questdb.griffin.engine.functions.constants.IntConstant;
 import io.questdb.std.Misc;
 import io.questdb.std.Numbers;
 import org.jetbrains.annotations.Nullable;
@@ -91,13 +92,21 @@ public class LimitRecordCursorFactory extends AbstractRecordCursorFactory {
         Function loFunc = cursor.loFunction;
         Function hiFunc = cursor.hiFunction;
         if (loFunc != null) {
-            sink.meta("lo").val(loFunc);
+            if (loFunc instanceof IntConstant) {
+                sink.meta("lo").val(loFunc.getInt(null));
+            } else {
+                sink.meta("lo").val(loFunc);
+            }
             if (loFunc.isRuntimeConstant()) {
                 sink.val('[').val(loFunc.getLong(null)).val(']');
             }
         }
         if (hiFunc != null) {
-            sink.meta("hi").val(hiFunc);
+            if (hiFunc instanceof IntConstant) {
+                sink.meta("hi").val(hiFunc.getInt(null));
+            } else {
+                sink.meta("hi").val(hiFunc);
+            }
             if (hiFunc.isRuntimeConstant()) {
                 sink.val('[').val(hiFunc.getLong(null)).val(']');
             }


bluestreak01:
would be good to see a test also

lost-particles:
Added tests as requested. Please review and merge.

lost-particles:
@bluestreak01 Could you please review the changes ?

core/src/main/java/io/questdb/griffin/engine/LimitRecordCursorFactory.java
```diff
@@ -91,13 +92,21 @@ public void toPlan(PlanSink sink) {
         Function loFunc = cursor.loFunction;
         Function hiFunc = cursor.hiFunction;
         if (loFunc != null) {
-            sink.meta("lo").val(loFunc);
+            if (loFunc instanceof IntConstant) {
```
puzpuzpuz, line 95: We have `isConstant()` method in `Function` interface. This method should be used to check if a function is constant. Also, lo and hi values are expected to be `long` values, so `func.getLong(null)` should be used to read constant's value.

Other than that, I'm not sure that the fix should be in `LimitRecordCursorFactory` - we have many factories that call `JsonPlanSink#val()` on `Function`, including all `UnaryFunction`/`BinaryFunction`/etc. AFAIU the fix should be done in `JsonPlanSink`.

puzpuzpuz: @nwoolmer please correct me if I'm wrong.

lost-particles: From my understanding while debugging, the issue is not in the JsonPlanSink. When we use Limit with values, **by default it is taken as a IntConstant function(unless the value is too high or explicitly mentioned as long as such: 1000L).**

So the code flow for limit values go through LimitRecordCursorFactory, where it decided which overriden function in JsonPlanSink is called. Because, the values are being sent as "IntConstant" function, this val function is hit inside JsonPlanSink : **public PlanSink val(Plannable s) instead of public PlanSink val(int i)** 

Below is the function definition for both of them : 

        @Override
            public PlanSink val(int i) {
                quoteValue = false;
                checkType(NODE_VALUE);
                sink.put(i);
                return this;
            }
    
        @Override
    public PlanSink val(Plannable s) {
        quoteValue = true;
        checkType(NODE_VALUE);
        if (s != null) {
            s.toPlan(this);
        } else {
            sink.put("null");
        }
        return this;
    }
    
    
**As you can notice, the issue happens because it hits the plannable overriden method which uses quoteValue as TRUE, leading to a starting quote. We shouldn't change here as this method is used for parsing an actual function used with LIMIT for instance : Limit abs(-3).**
    
 Also, as you suggested i tried using the isConstant() method. **But it won't work for our scenarios because isConstant() checks for all kinds of constantFunction, but we only need to check specifically for IntConstant.**

puzpuzpuz: > this val function is hit inside JsonPlanSink : **public PlanSink val(Plannable s) instead of public PlanSink val(int i)**

I don't understand why calling `val(Plannable)` here is a problem. Also, there are many record cursor factories that have the same `.meta("foobar").val(func)` call chain, so doing something about `LimitRecordCursorFactory` is not enough.

PR #6112, opened 2025-09-08
feat(sql): speed-up queries with multiple IN predicates on indexed columns

jerrinot:
This speeds up PGWire queries with  `IN()` predicates on multiple indexed columns. 

Example:
```sql
[...]
where symA in (?, ?) and symB in (?, ?)
[...]
```

Details: 
When both `symA` and `symbB` are indexed then the optimizer should plan index-scan on the column with higher cardinality. Since this column will likely result in a greater selectivity. However, before this PR the optimize would refuse to consider column cardinality in the presence of binding variables. Thus yielding a suboptimal plan. 

This change it more relaxed. Reasoning: We cannot merge keys when binding variables are used. But if a column is 'new' (=different from the current key column) then we can still analyze it and possibly swap it for the existing key column. This does not involve any key merging => it's safe with binding variables.

TODO:
- [x] Proper testing, come up with scenarios where this could back-fire
- [x] User-centric PR description :)


jerrinot:
@coderabbitai review

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/WhereClauseParser.java|2|2|100.00%|

puzpuzpuz:
@coderabbitai review

core/src/main/java/io/questdb/griffin/WhereClauseParser.java
```diff
@@ -924,7 +925,7 @@ private boolean analyzeListOfValues(
                 for (i--; i > -1; i--) {
                     ExpressionNode c = node.args.getQuick(i);
                     if ((c.type != ExpressionNode.CONSTANT && c.type != ExpressionNode.BIND_VARIABLE && c.type != ExpressionNode.FUNCTION) ||
-                            (c.type == ExpressionNode.BIND_VARIABLE && tempKeyValues.size() > 0)) {
+                            (!newColumn && c.type == ExpressionNode.BIND_VARIABLE && tempKeyValues.size() > 0)) {
                         return false;
```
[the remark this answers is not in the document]
puzpuzpuz: @jerrinot could you check this comment?

core/src/main/java/io/questdb/griffin/WhereClauseParser.java
```diff
@@ -882,8 +882,9 @@ private boolean analyzeListOfValues(
                 return false;
```
puzpuzpuz, line 894: nit: would be great to remove the following leftover comment above:
```java
            // "preferred" is an unfortunate name, this column is from "latest on" clause,
            // I should name it better
```

core/src/main/java/io/questdb/griffin/WhereClauseParser.java
```diff
@@ -924,7 +925,7 @@ private boolean analyzeListOfValues(
                 for (i--; i > -1; i--) {
                     ExpressionNode c = node.args.getQuick(i);
                     if ((c.type != ExpressionNode.CONSTANT && c.type != ExpressionNode.BIND_VARIABLE && c.type != ExpressionNode.FUNCTION) ||
```
puzpuzpuz, line 942: nit: would be really cool to add a clarifying comment around this check.

core/src/main/java/io/questdb/griffin/WhereClauseParser.java
puzpuzpuz, line 894: `columnIsPreferredOrIndexedAndNotPartOfMultiColumnLatestBy` - why do we have to check specifically for multi-column LATEST ON?

core/src/main/java/io/questdb/griffin/WhereClauseParser.java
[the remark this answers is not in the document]
puzpuzpuz: The comment is spot-on:
```java
    @Test
    public void testPlanWithIndexAndBindingVariables2() throws Exception {
        skipOnWalRun();

        assertWithPgServer(CONN_AWARE_EXTENDED, (connection, binary, mode, port) -> {
            execute(
                    "CREATE TABLE x ( " +
                            "sym1 SYMBOL INDEX," +
                            "sym2 SYMBOL INDEX," +
                            "ts TIMESTAMP " +
                            ") timestamp(ts) PARTITION BY DAY BYPASS WAL"
            );
            execute("insert into x select concat('sym', x%20), concat('sym', x%10), x::timestamp from long_sequence(100);");

            try (PreparedStatement ps = connection.prepareStatement(
                    "explain\n" +
                            "select * from x\n" +
                            "where sym1 in (?) and sym2 in ('sym0')\n" +
                            "and ts >= '1970-01-01T00:00:00.000077Z' and ts <= '1970-01-01T00:00:00.279828Z'\n" +
                            "order by ts asc;\n"
            )) {
                ps.setString(1, "sym0");
                try (ResultSet rs = ps.executeQuery()) {
                    sink.clear();
                    assertResultSet(
                            "QUERY PLAN[VARCHAR]\n" +
                                    "PageFrame\n" +
                                    "    Index forward scan on: sym2\n" + // we're scanning the low-cardinality column
                                    "      filter: sym2=10 and sym1 in [$0::string]\n" +
                                    "    Interval forward scan on: x\n" +
                                    "      intervals: [(\"1970-01-01T00:00:00.000077Z\",\"1970-01-01T00:00:00.279828Z\")]\n",
                            sink,
                            rs
                    );
                }
            }
        });
    }
```

core/src/main/java/io/questdb/griffin/WhereClauseParser.java
```diff
@@ -882,8 +882,9 @@ private boolean analyzeListOfValues(
                 return false;
             }
 
-            //if key values contain bind variable then we can't merge it with any other set and have to push this list to filter
-            if (!allKeyValuesAreKnown) {
+            // if key values contain bind variable then we can't merge it with any other set and have to push this list to filter,
+            // we can only create a new list of keys (=newColumn)
+            if (!allKeyValuesAreKnown & !newColumn) {
                 return false;
             }
```
puzpuzpuz, line 902: nit: the code in the below `if (i == 1)` branch breaks on `? = col` instead of `col = ?`.

core/src/main/java/io/questdb/griffin/WhereClauseParser.java
puzpuzpuz, line 894: Shouldn't we also do a similar change in `analyzeNotListOfValues`?

jerrinot: my assumption: When `LATEST ON` has multiple columns then a single column index is very useful. Certainly not without further optimizations. Perhaps if all columns in the LATEST ON are indexed then we could somehow combine the indexes. but this does not look to be implemented.

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/WhereClauseParser.java|3|3|100.00%|

core/src/main/java/io/questdb/griffin/WhereClauseParser.java
puzpuzpuz: > When `LATEST ON` has multiple columns then a single column index is very useful.

This should be "is not very useful", right?

jerrinot: yes :facepalm:

puzpuzpuz: We should probably add a clarifying comment somewhere around the method.

jerrinot: I am not entirely sure, but it looks to me there is a slightly different mechanism in `analyzeNotListOfValues()` 

It does not refuse binding variables straightaway, but then there is an escape hatch in `mergeKeys()`:  https://github.com/questdb/questdb/blob/111f87216963fc5717e7bbadc411481914762791/core/src/main/java/io/questdb/griffin/WhereClauseParser.java#L1924

core/src/main/java/io/questdb/griffin/WhereClauseParser.java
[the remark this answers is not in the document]
jerrinot: thank to both Andrei and @coderabbitai :) fixed

core/src/main/java/io/questdb/griffin/WhereClauseParser.java
jerrinot: done

core/src/main/java/io/questdb/griffin/WhereClauseParser.java
puzpuzpuz: Ok, indeed it looks different. Let's keep it as is.

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/WhereClauseParser.java|3|3|100.00%|

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/WhereClauseParser.java|3|3|100.00%|

PR #5805, opened 2025-07-02
perf(core): introduce three shared thread pools - Query, Write and Network

ideoma:
This PR changes default QuestDB threading strategy. Instead of single shared pool there will be three shared thread pools:

- Network thread pool. This will run network IO e.g. HTTP, Pg and ILP servers.
- Query Thread pool. This pool will handle parallel queries, e.g. parallel filters and group by
- Writer Thread pool. This will run Wal Apply Job, Table Writer jobs, Mat View Refresh, various purge and housekeeping jobs.

tandem: https://github.com/questdb/questdb-enterprise/pull/645

PR #6126, opened 2025-09-11
chore(core): run Apply Wal jobs on dedicated pool to limit tables number and reduce contention

ideoma:
This PR changes the default configuration to run Apply Wal Job to run on a dedicated worker pool.

This is the same behavior as it used to be before https://github.com/questdb/questdb/pull/5805 e.g. released in v9.0.2.

The main reason for this revert is to keep the behaviour the same as it was before the introduction of the dedicated writer pool. Writer pool was introduced to not run heavy Column Jobs in the shared pool; other aspects, such as wal-apply and mat-view parallelism, can be kept the same.


glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 3 / 6 (50.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/PropServerConfiguration.java|3|6|50.00%|

Issue #6052, opened 2025-08-14
Require materialized view query timestamp to be the base table designated timestamp


puzpuzpuz:
### Is your feature request related to a problem?

Due to the incremental refresh design, we should ban materialized views like the below one:
```
CREATE TABLE 'y' ( 
	x1 INT,
	s SYMBOL CAPACITY 256 CACHE,
	ts1 TIMESTAMP,
	ts2 TIMESTAMP
) timestamp(ts1) PARTITION BY DAY WAL;

create materialized view y_view as (select ts2, sum(x1) as sum_x1 from y timestamp(ts2) sample by 2s);
```

### Full Name:

Andrei Pechkurov

### Affiliation:

QuestDB

roomyxie:
can you assign it to me? i can repo this..

roomyxie:
notes:
The specified timestamp of base table y is ts1.
The materialized view y_view uses ts2 as the timestamp.
Due to the limitations of the incremental refresh design, this mismatch will cause problems.

puzpuzpuz:
@roomyxie sure thing, I've assigned the issue to you. Look forward to the contribution!

roomyxie:
Got it

PR #6133, opened 2025-09-12
fix(core): fix materialized view query timestamp to be the base table designated timestamp(#6052)

roomyxie:


roomyxie:
@kafka1991 @puzpuzpuz please check it again

PR #6136, opened 2025-09-12
test(core): skip FilesTest.testAllocateConcurrent() on tmpfs

jerrinot:
The test starts 2 threads, each attempting to allocate 2/3 of free space using posix_fallocate(). On filesystems like ext4, this creates sparse files that reserve space without consuming it immediately. However, tmpfs doesn't support sparse files and materializes the entire allocation in RAM, causing memory exhaustion that affects other processes and /tmp operations.

Skip this test when running on tmpfs to prevent system instability.

### DIFF
--- a/core/src/test/java/io/questdb/test/FilesTest.java
+++ b/core/src/test/java/io/questdb/test/FilesTest.java
@@ -1521,6 +1521,11 @@ public class FilesTest {
     }
 
     private static void assumeIsNotTmpFs(String path) {
+        // Assumption to skip a test if we run on 'tmpfs' filesystem.
+        // Background: tmpfs doesn't support sparse files - when posix_fallocate() is called,
+        // tmpfs immediately materializes the entire allocation in RAM rather than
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/questdb/test/FilesTest.java
```diff
@@ -104,6 +105,7 @@ public void testAllocateConcurrent() throws IOException, InterruptedException {
         FilesFacade ff = TestFilesFacadeImpl.INSTANCE;
 
         String tmpFolder = temporaryFolder.newFolder("allocate").getAbsolutePath();
+        assumeIsNotTmpFs(tmpFolder);
```
puzpuzpuz, line 108: nit: would be good to add a clarifying comment on why we don't want tmpfs here.

jerrinot: good idea

glasstiger:
#### [PR Coverage check]

PR #6141, opened 2025-09-13
Questdb issue 6121 bug investigation

aamoddev23-oss:
Root Cause Analysis
The issue occurred because:

The first() function's merge() method relied on row ID comparisons to determine which value was "first"
During parallel execution, row IDs don't necessarily correlate with the logical scan order
When ORDER BY DESC is used, the assumption that smaller row IDs represent "first" values breaks down
Fix Implemented
Modified FirstTimestampGroupByFunction.supportsParallelism() to return false, disabling parallel execution for first() functions
Added comprehensive test cases in FirstFunctionOrderByBugTest.java to reproduce and verify the fix
Included detailed comments explaining the reasoning behind disabling parallelism
Technical Details
The fix follows the suggestion in the original issue that "the query should not run in parallel as the access order is reversed"
By disabling parallelism, we ensure that row IDs maintain their correlation with scan order
This prevents incorrect results when first() is used with UNION ALL and subqueries containing ORDER BY clauses
The fix is conservative but correct - it prioritizes query correctness over potential performance benefits of parallel execution for this specific function type.


bluestreak01:
this looks like AI generated spam

Issue #6089, opened 2025-08-31
Materialize view is invalid


ayman-elmalah-awaed:
### To reproduce

1- Create a new table for historical prices
`-- Create historical_prices table
CREATE TABLE historical_prices (
    symbol SYMBOL,
    market SYMBOL,
    timestamp TIMESTAMP,
    price DOUBLE,
    volume LONG
) timestamp(timestamp) PARTITION BY DAY;`

2- Create materialize view sample by a week
`CREATE MATERIALIZED VIEW 'historical_prices_1week' AS (
    SELECT timestamp, symbol, market, 
           first(price) AS open, max(price) AS high, 
           min(price) AS low, last(price) AS close, 
           sum(volume) AS volume
    FROM historical_prices
    SAMPLE BY 1w
) PARTITION BY MONTH TTL 5 YEARS;`

3- insert some daily prices
`INSERT INTO historical_prices VALUES
('HP', 'NYSE', now(), 28.50, 100),
('HP', 'NYSE', now(), 28.55, 120),
('HP', 'NYSE', now(), 28.52, 80);`

after that i got an error message

`Materialized view is invalid: timestamp out of range [expected: 2025-08-28T00:00:00.000000Z, 2025-09-04T00:00:00.000000Z), actual: 2025-08-25T00:00:00.000000Z`

### QuestDB version:

latest

### OS, in case of Docker specify Docker and the Host OS:

Macos

### File System, in case of Docker specify Host File System:

ext4

### Full Name:

Ayman Elmalah

### Affiliation:

-

### Have you followed Linux, MacOs kernel configuration steps to increase Maximum open files and Maximum virtual memory areas limit?

- [x] Yes, I have

bluestreak01:
thank you @ayman-elmalah-awaed we can repro this

ayman-elmalah-awaed:
@bluestreak01  @puzpuzpuz  I need to raise also something, that issue always happens when inserting records in weekly MV in any Thursday, for example


```
INSERT INTO historical_prices VALUES
('HP', 'NYSE', '2025-08-28T09:30:00', 28.50, 100),
('HP', 'NYSE', '2025-08-28T09:30:00', 28.55, 120),
('HP', 'NYSE', '2025-08-28T09:30:00', 28.52, 80);
```

puzpuzpuz:
@ayman-elmalah-awaed the fix is on its way. It'll be a part of the nano timestamp unit PR (#5685)

PR #6142, opened 2025-09-13
Fix materialized view timestamp range validation for weekly sampling …

aamoddev23-oss:
…(issue #6089)

- Fixes AssertionError when creating weekly materialized views that fail timestamp validation
- The original assertion was too strict for week boundary calculations
- Added more permissive validation that allows tolerance for weekly and larger bucket sizes
- Added getBucketSize() method to SampleByIntervalIterator for accessing sampler bucket size
- Resolves "timestamp out of range" errors when processing data that falls on week boundaries

Changes:
- Modified MatViewRefreshJob.java to use relaxed timestamp validation logic
- Added SampleByIntervalIterator.getBucketSize() method for bucket size access
- Improved error handling to be more tolerant of precision issues in weekly time buckets


bluestreak01:
this issue is already fixed here: https://github.com/questdb/questdb/pull/5685

Issue #6032, opened 2025-08-07
Array expressions cannot be used in ORDER BY


puzpuzpuz:
### Is your feature request related to a problem?

```sql
create table x as (
  select array[x] id, timestamp_sequence(0,1000000000) as ts
  from long_sequence(10)
) timestamp(ts) partition by hour;

x order by id[1] desc;
```

The query gives an `Invalid column: []` error while it should succeed.

### Full Name:

Andrei Pechkurov

### Affiliation:

QuestDB

jerrinot:
good one!

javier:
just came to report this and saw it was already here :)

roomyxie:
can you assign this one to me?

jerrinot:
hi @roomyxie go ahead! if you have any question then stop by at QuestDB slack: https://slack.questdb.com/

roomyxie:
sure

PR #6157, opened 2025-09-18
fix(core): fix Array expressions cannot be used in ORDER BY(#6032)

roomyxie:
https://github.com/questdb/questdb/issues/6032
updates:
1, Check if any ORDER BY item is a complex expression on SqlCodeGenerator and so on.


roomyxie:
@puzpuzpuz  please review it

PR #6158, opened 2025-09-18
feat(sql): speed up ASOF JOIN on indexed symbol column

mtopolnik:
Leverage the symbol index on the right-hand side of the ASOF JOIN, avoiding the need to visit any rows that don't have the matching symbol.

## Where this matters

The major use case for this optimization is **markout analysis.**

For a given trade, you want to see whether you could have got a better deal if you placed your order a bit sooner or a bit later. In a representative example, you inspect a window of +/- 10 minutes around the trade, amounting to 1200 points in time (one per second).

When dealing with FOREX, you need two numbers for each trade: USD rate of currency you're selling, and of the currency you're buying. There's a single table that contains all price changes for all currencies.

Therefore this analysis requires 2400 lookups of a matching RHS row (USD rate) for each LHS row (trade event), conditioned on the currency symbol. The worst problem are illiquid assets, having their last price posted way back in the past.

The current code will scan all the rows backwards until it finds a match. The new code makes only one symbol index lookup per partition. So, if the last asset price was posted 30 days ago, that's 30 lookups vs. scanning the full 30 days of  currency price updates, for all currencies.

## Benchmark

I did a quick ad-hoc benchmark set up as follows.

1. Price table:

```sql
CREATE TABLE prices (
      ts TIMESTAMP,
      sym SYMBOL,
      price DOUBLE
```

- Five currency symbols: `'EUR', 'GBP', 'JPY', 'CAD', 'AUD'`
- 320 million rows, one per second -> 3700 partitions
- `AUD` appears only in the very first partition

2. Trade event table:

```sql
CREATE TABLE orders (
    order_id LONG,
    order_completed_ts TIMESTAMP,
    sym1 SYMBOL,
    sym2 SYMBOL,
    quantity DOUBLE
) timestamp(order_completed_ts) PARTITION BY DAY;

INSERT INTO order VALUES
  (1001, '2025-09-09T09:00:00.000000Z', 'EUR', 'GBP', 50000.0),
  (1002, '2025-09-09T09:00:03.000000Z', 'GBP', 'JPY', 25000.0),
  (1003, '2025-09-09T09:00:07.000000Z', 'CAD', 'AUD', 75000.0),
  (1004, '2025-09-09T09:00:12.000000Z', 'JPY', 'EUR', 12500000.0),
  (1005, '2025-09-09T09:00:15.000000Z', 'EUR', 'CAD', 80000.0),
  (1006, '2025-09-09T09:00:21.000000Z', 'EUR', 'JPY', 45000.0),
  (1007, '2025-09-09T09:00:26.000000Z', 'GBP', 'CAD', 30000.0),
  (1008, '2025-09-09T09:00:32.000000Z', 'AUD', 'EUR', 60000.0),
  (1009, '2025-09-09T09:00:38.000000Z', 'CAD', 'GBP', 35000.0),
  (1010, '2025-09-09T09:00:43.000000Z', 'JPY', 'AUD', 15000000.0);
```

3. Query:

```sql
WITH
  offsets AS (
      SELECT x-601 AS sec_offs FROM long_sequence(1201)
  ),
  orders AS (
    SELECT id, order_ts, sym1, sym2 FROM orders
  ),
  points AS (
    SELECT orders.*, offsets.sec_offs, order_ts + 1_000_000 * sec_offs AS ts
    FROM orders CROSS JOIN offsets
    ORDER BY order_ts + 1_000_000 * sec_offs
  ),
  rslt1 AS (
    SELECT t.*, p.price AS sym1_price
    FROM points as t
    ASOF JOIN prices as p
    ON (t.sym1 = p.sym)
  ),
  rslt2 AS (
    SELECT t.*, p.price AS sym2_price
    FROM rslt1 as t
    ASOF JOIN prices as p
    ON (t.sym2 = p.sym)
  )
SELECT id, sec_offs, sym1_price/sym2_price AS price FROM rslt2
WHERE sym1_price is not null and sym2_price is not null
ORDER BY id, sec_offs;
```

Without the symbol index, the query times out. Profiling the query in more detail, one trade row at a time, a clear pattern shows up: if the row involves `AUD`, it's super-slow.

With the symbol index, the query is done within a second.


glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 164 / 179 (91.62%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java|0|2|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/TimeFrameRecordCursorImpl.java|5|7|71.43%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java|75|86|87.21%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AbstractAsOfJoinFastRecordCursor.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinFastRecordCursorFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AbstractKeyedAsOfJoinRecordCursor.java|44|44|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|17|17|100.00%|

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 161 / 169 (95.27%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/TimeFrameRecordCursorImpl.java|5|7|71.43%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java|66|72|91.67%|
|:large_blue_circle:|parquet_read/decode.rs|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AbstractAsOfJoinFastRecordCursor.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinFastRecordCursorFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AbstractKeyedAsOfJoinRecordCursor.java|47|47|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|17|17|100.00%|

puzpuzpuz:
@coderabbitai review

.github/workflows/pgwire_latest.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         # even if the original tag is moved or deleted
         uses: actions-rust-lang/setup-rust-toolchain@4d1965c9142484e48d40c19de54b5cba84953a06
         with:
-          toolchain: stable
+          toolchain: nightly-2025-09-18
```
puzpuzpuz: Do we need these changes here? Seems to be orthogonal to the goal of the PR.

core/src/main/java/io/questdb/cairo/sql/TimeFrameRecordCursor.java
```diff
@@ -33,6 +34,26 @@
...
+    BitmapIndexReader getBitmapIndexReader(int columnIndex, int direction);
+
+    /**
+     * Maps logical (query-level) column index to physical (table-level) column index.
+     * For example, we need this to access the index of a symbol column.
+     *
+     * @param logicalColumnIndex the logical column index from the query perspective
+     * @return the physical index of the column in the table
+     */
+    int getPhysicalColumnIndex(int logicalColumnIndex);
```
puzpuzpuz: There is little point in making the API more complex by having this method. Instead, `getBitmapIndexReader()` should accept the logical index and do index remapping.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -2710,21 +2711,36 @@ private RecordCursorFactory generateJoins(QueryModel model, SqlExecutionContext
...
+                                                            master,
+                                                            slave,
+                                                            masterMetadata.getColumnCount(),
+                                                            masterSymbolColumnIndex,
+                                                            slaveSymbolColumnIndex,
+                                                            slaveModel.getContext(),
+                                                            asOfToleranceInterval
+                                                    );
+                                                } else {
+                                                    // Fall back to original fast join with symbol short-circuit
```
puzpuzpuz: nit: what is "original" here? After this change, the comment will look strange, so I'd remove "original" from the text.

core/src/main/java/io/questdb/griffin/engine/join/AbstractAsOfJoinFastRecordCursor.java
```diff
@@ -332,13 +332,14 @@ protected void nextSlave(long masterTimestamp) {
                     record.hasSlave(true);
                     slaveTimeFrameCursor.recordAt(slaveRecB, Rows.toRowID(slaveFrameIndex, slaveFrameRow));
                     long slaveTimestamp = scaleTimestamp(slaveRecB.getTimestamp(slaveTimestampIndex), slaveTimestampScale);
-                    if (slaveFrameRow < slaveTimeFrame.getRowHi() - 1) {
+                    if (foundRow < slaveTimeFrame.getRowHi() - 1) {
```
puzpuzpuz: nit: although `slaveFrameRow` and `foundRow` have the same value, using `slaveFrameRow` here read better since the below assignment uses `slaveFrameRow + 1`, not `foundRow + 1`.

core/src/main/java/io/questdb/griffin/engine/join/AbstractAsOfJoinFastRecordCursor.java
```diff
@@ -332,13 +332,14 @@ protected void nextSlave(long masterTimestamp) {
...
+                        // Set lookaheadTimestamp to the first one larger than masterTimestamp, and return
```
puzpuzpuz, line 336: > and return

nit: this part seems confusing as the branch has no `return` statement.

core/src/main/java/io/questdb/griffin/engine/join/AbstractKeyedAsOfJoinRecordCursor.java
```diff
@@ -0,0 +1,153 @@
...
+        }
+        record.hasSlave(origHasSlave);
+
+        final long masterTimestamp = scaleTimestamp(masterRecord.getTimestamp(masterTimestampIndex), masterTimestampScale);
+        if (masterTimestamp >= lookaheadTimestamp) {
+            nextSlave(masterTimestamp);
+        }
+        // Set `isMasterHasNextPending` only now because `nextSlave()` may throw DataUnavailableException,
+        // and in such a case we don't want to call `masterCursor.hasNext()` during the next call to `this.hasNext()`.
+        // If we are here, it's clear that nextSlave() did not throw DataUnavailableException.
```
puzpuzpuz, line 96: > it's clear that nextSlave() did not throw DataUnavailableException.

nit: that's correct, but `performKeyMatching` may throw `DataUnavailableException`. On the other hand, we're planning to get rid of `DataUnavailableException` mechanism, so it's probably not important.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java
```diff
@@ -0,0 +1,209 @@
...
+                }
+                symbolKey++;
+            }
+
+            // go backwards through per-frame symbol indexes, until we find a match or exhaust the search space
+            final int physicalSlaveSymbolColumnIndex = slaveTimeFrameCursor.getPhysicalColumnIndex(slaveSymbolColumnIndex);
+            long rowMax = Rows.toLocalRowID(slaveRecB.getRowId());
+            int frameIndex = slaveTimeFrame.getFrameIndex();
+            for (; ; ) {
+                BitmapIndexReader indexReader = slaveTimeFrameCursor.getBitmapIndexReader(
```
puzpuzpuz: Can we initialize `BitmapIndexReader` once per cursor init? The operation doesn't seem to be super-cheap.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java
```diff
@@ -0,0 +1,209 @@
...
+            int symbolKey;
+            if (masterSymbolValue == null) {
+                symbolKey = 0;
+            } else {
+                symbolKey = symbolTable.keyOf(masterSymbolValue);
+                if (symbolKey == StaticSymbolTable.VALUE_NOT_FOUND) {
+                    record.hasSlave(false);
+                    return;
+                }
+                symbolKey++;
```
puzpuzpuz: This doesn't seem to be correct - we should probably use `TableUtils#toIndexKey()` here. Also, could you add a test that involves asof join on a `null` symbol value?

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java
```diff
@@ -0,0 +1,209 @@
...
+                        BitmapIndexReader.DIR_BACKWARD
+                );
+                // indexReader.getCursor() takes absolute row IDs, but TimeFrameCursor uses numbering relative to
+                // the first row within the BETWEEN ... AND ... range selected by the query.
+                // Use Record.getUpdateRowId() to get the absolute row ID.
+                slaveTimeFrameCursor.recordAt(slaveRecA, Rows.toRowID(frameIndex, slaveTimeFrame.getRowLo()));
+                final long rowLo = Rows.toLocalRowID(slaveRecA.getUpdateRowId());
+                RowCursor rowCursor = indexReader.getCursor(false, symbolKey, rowLo, rowMax + rowLo);
+
+                // Check the first entry only. They are sorted descending by timestamp, so other entries are older.
```
puzpuzpuz: > other entries are older

nit: "other entries are not younger" would be more precise as there may be duplicate timestamps in the table.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java
```diff
@@ -0,0 +1,209 @@
...
+                int lookahead
+        ) {
+            super(columnSplit, nullRecord, masterTimestampIndex, masterTimestampType, masterSinkTarget,
+                    slaveTimestampIndex, slaveTimestampType, slaveSinkTarget, lookahead);
+        }
+
+        @Override
+        protected void performKeyMatching(long masterTimestamp) {
+            // determine the integer key of the symbol to find in the slave table
+            CharSequence masterSymbolValue = masterRecord.getSymA(masterSymbolColumnIndex);
```
puzpuzpuz, line 164: nit: we could additionally optimize for self-join on the same symbol column. In that case, we can skip the lookup and take the symbol code (key) directly.

core/src/test/java/io/questdb/test/griffin/AsOfJoinFuzzTest.java
```diff
@@ -46,10 +46,13 @@
 
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
 
 @RunWith(Parameterized.class)
 public class AsOfJoinFuzzTest extends AbstractCairoTest {
```
puzpuzpuz, line 53: Would be great to add a few non-fuzz tests to `AsOfJoinTest`.

.github/workflows/pgwire_latest.yml
mtopolnik: PR wouldn't build without them.

core/src/main/java/io/questdb/griffin/engine/join/AbstractAsOfJoinFastRecordCursor.java
mtopolnik: It does actually have the `return` statement, at line 339. The diff just looks a bit confusing.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java
mtopolnik: This is once per partition/frame, I think the index reader is local to the frame. I had it outside before and it produced wrong results.

core/src/main/java/io/questdb/griffin/engine/join/AbstractAsOfJoinFastRecordCursor.java
puzpuzpuz: Oups, my bad :)

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java
mtopolnik: It seems that `TableUtils.toIndexKey()` also has an issue -- it returns -1 when the input value is `SymbolTable.VALUE_NOT_FOUND`. Shouldn't it return the constant unchanged?

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java
puzpuzpuz: > This is once per partition/frame

That's true. The API is misleading - we should expose the `getBitmapIndexReader` method on `TimeFrame` objects. Other than that, we could reuse the `indexReader` as long as the frame index stays the same - this can be done within `TimeFrame` itself.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java
puzpuzpuz: I guess, the `SymbolTable.VALUE_NOT_FOUND` value should be checked before calling `TableUtils.toIndexKey()`. You could also check the value returned from `toIndexKey()` to be non-negative, but that's uncommon among the rest of the codebase.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java
mtopolnik: Exposing it on `TimeFrame` would be more invasive. `TimeFrame` right now is a simple data object with no dependencies, whereas `TimeFrameRecordCursor` is the heavyweight object that keeps a reference to the `TableReader`, where we ultimately get the index reader from.

I renamed the method to `getIndexReaderForCurrentFrame()`.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java
mtopolnik: I do check for non-negative value. If I didn't, I'd first have to assign the raw value to the variable, check, and then reassign the adjusted value. It would be overall cleaner if `toIndexKey()` took care of all cases instead.

.github/workflows/pgwire_latest.yml
puzpuzpuz: Weird that other PRs seem to be building fine.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java
```diff
@@ -0,0 +1,204 @@
...
+                }
+
+                // no match in this frame, try the previous frame
+                if (!slaveTimeFrameCursor.prev()) {
+                    record.hasSlave(false);
+                    return;
+                }
+                slaveTimeFrameCursor.open();
+                frameIndex = slaveTimeFrame.getFrameIndex();
+                rowMax = slaveTimeFrame.getRowHi() - 1;
```
puzpuzpuz, line 208: We need to place `circuitBreaker.statefulThrowExceptionIfTripped();` somewhere here to avoid leaking readers and worker threads in case of very slow query.

.github/workflows/pgwire_latest.yml
mtopolnik: That's because in the meantime Alex merged a quick workaround, disabling the warning. I instead upgraded Rust.

mtopolnik: Is it a better idea to remove my fixes, and leave upgrading Rust to another PR?

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java
```diff
@@ -0,0 +1,204 @@
...
+        ) {
+            super(columnSplit, nullRecord, masterTimestampIndex, masterTimestampType, masterSinkTarget,
+                    slaveTimestampIndex, slaveTimestampType, slaveSinkTarget, lookahead);
+        }
+
+        @Override
+        protected void performKeyMatching(long masterTimestamp) {
+            // determine the integer key of the symbol to find in the slave table
+            CharSequence masterSymbolValue = masterRecord.getSymA(masterSymbolColumnIndex);
+            StaticSymbolTable symbolTable = slaveTimeFrameCursor.getSymbolTable(slaveSymbolColumnIndex);
```
puzpuzpuz: `symbolTable` can be initialized single time, in `of()`.

.github/workflows/pgwire_latest.yml
puzpuzpuz: Yeah, that'd be great.

mtopolnik:
> @mtopolnik do the results make sense to you? If so, maybe we should introduce a hint to disable index usage? We already have a few hints, e.g. `avoid_asof_binary_search`, but not the index one.

I profiled this query, time is spent in the code that looks through the index to find the starting row. It involves traversing a linked list backwards. In this degenerate case, all rows are in the index, but there's no dead-reckoning way of going straight to a given row. So it degrades into linear search.

puzpuzpuz:
> > @mtopolnik do the results make sense to you? If so, maybe we should introduce a hint to disable index usage? We already have a few hints, e.g. `avoid_asof_binary_search`, but not the index one.
> 
> I profiled this query, time is spent in the code that looks through the index to find the starting row. It involves traversing a linked list backwards. In this degenerate case, all rows are in the index, but there's no dead-reckoning way of going straight to a given row. So it degrades into linear search.

@bluestreak01 @nwoolmer it sounds like our symbol index could be enhanced by adding a sparse index of blocks - the goal is to speed up row id-based block search. See https://github.com/questdb/questdb/pull/6158#pullrequestreview-3262903367 for the overhead of the block search.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 159 / 165 (96.36%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/TimeFrameRecordCursorImpl.java|5|7|71.43%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java|67|71|94.37%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AbstractAsOfJoinFastRecordCursor.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinFastRecordCursorFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AbstractKeyedAsOfJoinRecordCursor.java|47|47|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|17|17|100.00%|

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 158 / 165 (95.76%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/TimeFrameRecordCursorImpl.java|5|7|71.43%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java|67|71|94.37%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AbstractAsOfJoinFastRecordCursor.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinFastRecordCursorFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AbstractKeyedAsOfJoinRecordCursor.java|47|47|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|17|17|100.00%|

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 158 / 165 (95.76%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/TimeFrameRecordCursorImpl.java|5|7|71.43%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java|67|71|94.37%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AbstractAsOfJoinFastRecordCursor.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinFastRecordCursorFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AbstractKeyedAsOfJoinRecordCursor.java|47|47|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|17|17|100.00%|

PR #6167, opened 2025-09-22
fix(core): fix materialized view query timestamp to be the base table designated timestamp(#6052)

### LINKED ISSUE #6052: Require materialized view query timestamp to be the base table designated timestamp

roomyxie:
Fixes #6052

### DIFF
--- a/core/src/test/java/io/questdb/test/cairo/mv/CreateMatViewTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/mv/CreateMatViewTest.java
@@ -412,10 +412,10 @@ public class CreateMatViewTest extends AbstractCairoTest {
         assertMemoryLeak(() -> {
             createTable(TABLE1);
 
-            final String query = "select timestamp_floor('1m', ts) as ts, avg(v) from " + TABLE1 + " order by ts";
+            final String query = "select timestamp_floor('1m', ts) as ts1, avg(v) from " + TABLE1 + " order by ts1";
             execute("create materialized view test as (" + query + ") partition by day");
 
-            assertQuery0("ts\tavg\n", "test", "ts");
+            assertQuery0("ts1\tavg\n", "test", "ts1");
             assertMatViewDefinition(MatViewDefinition.REFRESH_TYPE_IMMEDIATE, "test", query, TABLE1, 1, 'm', null, null);
             assertMatViewDefinitionFile(MatViewDefinition.REFRESH_TYPE_IMMEDIATE, "test", query, TABLE1, 1, 'm', null, null);
         });
@@ -426,10 +426,10 @@ public class CreateMatViewTest extends AbstractCairoTest {
         assertMemoryLeak(() -> {
             createTable(TABLE1);
 
-            final String query = "select timestamp_floor('1m', ts, 0::timestamp) as ts, avg(v) from " + TABLE1;
+            final String query = "select timestamp_floor('1m', ts, 0::timestamp) as ts2, avg(v) from " + TABLE1;
             execute("create materialized view test as (" + query + ") partition by day");
 
-            assertQuery0("ts\tavg\n", "test", "ts");
+            assertQuery0("ts2\tavg\n", "test", "ts2");
             assertMatViewDefinition(MatViewDefinition.REFRESH_TYPE_IMMEDIATE, "test", query, TABLE1, 1, 'm', null, null);
             assertMatViewDefinitionFile(MatViewDefinition.REFRESH_TYPE_IMMEDIATE, "test", query, TABLE1, 1, 'm', null, null);
         });
@@ -1089,7 +1089,7 @@ public class CreateMatViewTest extends AbstractCairoTest {
     public void testCreateMatViewTsAlias() throws Exception {
         assertMemoryLeak(() -> {
             createTable(TABLE1);
-            final String query = "select ts, max(v) as v_max from " + TABLE1 + " sample by 30s";
+            final String query = "select ts, ts as ts1, max(v) as v_max from " + TABLE1 + " sample by 30s";
             execute("create materialized view test as (" + query + ") partition by week");
             assertMatViewDefinition(MatViewDefinition.REFRESH_TYPE_IMMEDIATE, "test", query, TABLE1, 30, 's', null, null);
             assertMatViewDefinitionFile(MatViewDefinition.REFRESH_TYPE_IMMEDIATE, "test", query, TABLE1, 30, 's', null, null);


roomyxie:
@puzpuzpuz please review it.

core/src/test/java/io/questdb/test/cairo/mv/CreateMatViewTest.java
```diff
@@ -412,10 +412,10 @@ public void testCreateMatViewGroupByTimestamp1() throws Exception {
         assertMemoryLeak(() -> {
             createTable(TABLE1);
 
-            final String query = "select timestamp_floor('1m', ts) as ts1, avg(v) from " + TABLE1 + " order by ts1";
+            final String query = "select timestamp_floor('1m', ts) as ts, avg(v) from " + TABLE1 + " order by ts";
```
puzpuzpuz: What's the reason behind this change? The query should be supported in mat views.

core/src/test/java/io/questdb/test/cairo/mv/CreateMatViewTest.java
```diff
@@ -426,10 +426,10 @@ public void testCreateMatViewGroupByTimestamp2() throws Exception {
         assertMemoryLeak(() -> {
             createTable(TABLE1);
 
-            final String query = "select timestamp_floor('1m', ts, 0::timestamp) as ts2, avg(v) from " + TABLE1;
+            final String query = "select timestamp_floor('1m', ts, 0::timestamp) as ts, avg(v) from " + TABLE1;
```
puzpuzpuz: Same here.

core/src/main/java/io/questdb/griffin/engine/ops/CreateMatViewOperationImpl.java
```diff
@@ -651,4 +718,56 @@ private void updateMatViewTablePartitionBy(int timestampType) throws SqlExceptio
             }
         }
     }
+
+
+    /**
+     * Recursively searches for the underlying column name in an expression AST.
+     */
+    private static String findUnderlyingColumnName(ExpressionNode node) {
```
puzpuzpuz: What if a function/operator involves multiple columns, e.g. `(t1.ts + t2.ts) / 2`?

core/src/main/java/io/questdb/griffin/engine/ops/CreateMatViewOperationImpl.java
```diff
@@ -440,14 +440,81 @@ public void validateAndUpdateMetadataFromModel(
...
+                                .put("base table timestamp=").put(baseTableTimestampColumn)
+                                .put(", materialized view timestamp=").put(finalTimestamp)
+                                .put(']');
+                    }
+                } else {
+                    final ExpressionNode queryTimestamp = queryModel.getTimestamp();
+                    if (queryTimestamp != null && queryTimestamp.type == ExpressionNode.LITERAL) {
+                        String effectiveColumnName = queryTimestamp.token.toString();
+
+                        // Check if this is an aliased column
```
puzpuzpuz, line 484: Aliases may be multi-level, so this check seems insufficient. E.g.
```sql
SELECT ts2 AS ts3, count()
FROM (
  SELECT ts1 AS ts2
  FROM (
    SELECT ts AS ts1
    FROM my_table
  )
)
```

roomyxie: OK, let me check it.

core/src/test/java/io/questdb/test/cairo/mv/CreateMatViewTest.java
roomyxie: let me fix it.

core/src/main/java/io/questdb/griffin/engine/ops/CreateMatViewOperationImpl.java
roomyxie: @puzpuzpuz  fixed

core/src/test/java/io/questdb/test/cairo/mv/CreateMatViewTest.java
roomyxie: fixed

core/src/test/java/io/questdb/test/cairo/mv/CreateMatViewTest.java
roomyxie: fixed

core/src/main/java/io/questdb/griffin/engine/ops/CreateMatViewOperationImpl.java
roomyxie: fixed

core/src/main/java/io/questdb/griffin/engine/ops/CreateMatViewOperationImpl.java
```diff
@@ -440,14 +440,85 @@ public void validateAndUpdateMetadataFromModel(
...
+            final int baseTableTimestampIndex = baseTableMetadata.getTimestampIndex();
+            if (baseTableTimestampIndex >= 0) {
+                final String baseTableTimestampColumn = baseTableMetadata.getColumnName(baseTableTimestampIndex);
+
+                final String finalTimestamp = createTableOperation.getTimestampColumnName();
+                if (finalTimestamp != null) {
+                    boolean timestampMatches = Chars.equalsIgnoreCase(finalTimestamp, baseTableTimestampColumn);
+
+                    // If not directly matching
+                    if (!timestampMatches) {
```
puzpuzpuz, line 452: Even if `timestampMatches` is `true`, this doesn't mean it's base table's designated timestamp column.

```sql
SELECT ts, count()
FROM (
  SELECT ts2 as ts
  FROM my_table
  ORDER BY ts2 ASC
)
SAMPLE BY 1h;
```

core/src/main/java/io/questdb/griffin/engine/ops/CreateMatViewOperationImpl.java
```diff
@@ -440,14 +440,85 @@ public void validateAndUpdateMetadataFromModel(
...
+                        throw SqlException.position(timestampPos != 0 ? timestampPos : selectTextPosition)
+                                .put("materialized view query timestamp must match base table designated timestamp [")
+                                .put("base table timestamp=").put(baseTableTimestampColumn)
+                                .put(", materialized view timestamp=").put(finalTimestamp)
+                                .put(']');
+                    }
+                } else {
+                    final ExpressionNode queryTimestamp = queryModel.getTimestamp();
+                    if (queryTimestamp != null && queryTimestamp.type == ExpressionNode.LITERAL) {
+                        String effectiveColumnName = queryTimestamp.token.toString();
```
puzpuzpuz, line 482: We're trying to avoid creating intermediate objects, when possible. In this case, you could just use `CharSequence` instead of `String`. Same consideration applies to other `toString()` calls in the new code.

core/src/main/java/io/questdb/griffin/engine/ops/CreateMatViewOperationImpl.java
```diff
@@ -440,14 +440,85 @@ public void validateAndUpdateMetadataFromModel(
...
+            if (baseTableTimestampIndex >= 0) {
+                final String baseTableTimestampColumn = baseTableMetadata.getColumnName(baseTableTimestampIndex);
+
+                final String finalTimestamp = createTableOperation.getTimestampColumnName();
+                if (finalTimestamp != null) {
+                    boolean timestampMatches = Chars.equalsIgnoreCase(finalTimestamp, baseTableTimestampColumn);
+
+                    // If not directly matching
+                    if (!timestampMatches) {
+                        for (int i = 0, n = queryModel.getColumns().size(); i < n; i++) {
```
puzpuzpuz, line 453: The way to check for base table timestamp is to ascend through query models until we reach a model that corresponds to the base table name (`tableNameExpr`). On the way, we need to keep mapping column aliases, so that at each level we know the current timestamp alias. Here is a simple example:
```
SELECT ts2 AS ts3, count()
FROM (
  SELECT ts1 AS ts2
  FROM (
    SELECT ts AS ts1
    FROM my_table
  )
)
```

We start with `ts3` at level 3 (the outer level). At that level, `ts3` is an alias for `ts2`. At level 2, `ts2` is an alias for `ts1`. At level 1, `ts1` is an alias for `ts`. Level 1 also stands for `my_table` table, so the final column name is `ts`.

We also need to consider functions and expressions, e.g. `timestamp_floor()`. Finally, one more thing to consider is all sorts of JOINs and UNIONs - there may be complicated cases like SAMPLE BY over a UNION of multiple tables. Probably, we should make the whole check best-effort and allow creating the mat view in such complicated cases.

core/src/main/java/io/questdb/griffin/engine/ops/CreateMatViewOperationImpl.java
```diff
@@ -651,4 +722,135 @@ private void updateMatViewTablePartitionBy(int timestampType) throws SqlExceptio
...
+
+    /**
+     * Checks if a column name looks like a timestamp column.
+     */
+    private static boolean isTimestampLikeColumn(String columnName) {
+        if (columnName == null) {
+            return false;
+        }
+        String lower = columnName.toLowerCase();
+        return lower.equals("ts") ||
```
puzpuzpuz, line 846: The code should not make any assumptions on table or column names. It's up to the user how to call tables/columns.

core/src/main/java/io/questdb/griffin/engine/ops/CreateMatViewOperationImpl.java
roomyxie: OK

PR #6171, opened 2025-09-22
chore(build): change docker pipeline trigger to run on release tag pattern only

ideoma:


ci/docker-release-pipeline.yml
```diff
@@ -29,6 +29,21 @@ jobs:
     timeoutInMinutes: 60
 
     steps:
+      - bash: |
+          tag=$(Build.SourceBranchName)
+          echo "Checking tag: $tag"
+          
+          if [[ ! $tag =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
glasstiger: What about `^[0-9]+(\.[0-9]+)*$` or `^\d+(\.\d+)*$` ?
This would not limit to three version numbers.
Just in case we have a release like: 1.2.3.4

ideoma: Ok, I updated the pattern + added SKIP_VALIDATION variable that someone can set to true when running pipelining manually for any tag.

PR #6180, opened 2025-09-24
chore(core): don't attempt rollback on a distressed table writer

mtopolnik:
When closing a `TableWriter`, we normally first roll it back. However, if the writer is already flagged as distressed, executing the rollback proceduce is dangerous, because the state of the writer is unknown. As a specific example, a `remap` operation may have failed, leaving a native data structure in an invalid state, and the rollback attempt may result in a segfault.

This PR ensures the `TableWriter` isn't distressed before attempting a rollback.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/pool/WriterPool.java
+++ b/core/src/main/java/io/questdb/cairo/pool/WriterPool.java
@@ -536,24 +536,26 @@ public class WriterPool extends AbstractPool {
         final long thread = Thread.currentThread().getId();
         final TableToken tableToken = e.writer.getTableToken();
 
-        boolean isDistressed;
-        try {
-            e.writer.rollback();
-            // Rollback can change writer state to distressed, do not observe it before rollback
-            isDistressed = e.writer.isDistressed();
-            if (!isDistressed) {
-                if (e.owner != UNALLOCATED) {
-                    e.owner = QUEUE_PROCESSING_OWNER;
+        boolean isDistressed = e.writer.isDistressed();
+        if (!isDistressed) {
+            try {
+                e.writer.rollback();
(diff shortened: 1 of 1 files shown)


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/pool/WriterPool.java|11|11|100.00%|

ideoma:
This is a good fix, but there is a test failure

```
[2] Metadata read timeout [src=writer, timeout=5000ms, err=could not open, file does not exist: /var/folders/5k/fq5ptf2906bgfn1krb78bmj00000gn/T/junit11198402257022157757/dbRoot/tbl~/_meta]]



Stack trace
io.questdb.cairo.CairoException: [2] Metadata read timeout [src=writer, timeout=5000ms, err=could not open, file does not exist: /var/folders/5k/fq5ptf2906bgfn1krb78bmj00000gn/T/junit11198402257022157757/dbRoot/tbl~/_meta]]
	at io.questdb@9.0.4-SNAPSHOT/io.questdb.cairo.CairoException.instance(CairoException.java:379)
	at io.questdb@9.0.4-SNAPSHOT/io.questdb.cairo.CairoException.critical(CairoException.java:77)
	at io.questdb@9.0.4-SNAPSHOT/io.questdb.cairo.TableUtils.handleMetadataLoadException(TableUtils.java:873)
```

I think what happens is that because we don't call `rollback()`, there are failures where `_meta` file does not exist. The solution can be that when TableWriter is distressed, `WriterPool` has to re-create `TableWriter`. We need to handle cases when `TableWriter` is distressed because the table is dropped, not to re-create it in that situation.

PR #6182, opened 2025-09-24
chore(core): enable gc log in tests on CI

glasstiger:
required by https://github.com/questdb/questdb-enterprise/pull/711

### DIFF
--- a/ci/templates/steps.yml
+++ b/ci/templates/steps.yml
@@ -97,13 +97,12 @@ steps:
       mavenVersionOption: "$(MAVEN_VERSION_OPTION)"
       mavenDirectory: '$(build.sourcesdirectory)/maven/apache-maven-$(MAVEN_VERSION)'
       mavenPomFile: "pom.xml"
-      mavenOptions: "-Dfile.encoding=UTF-8 $(MAVEN_OPTS)"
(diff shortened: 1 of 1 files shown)


core/src/test/java/io/questdb/test/cairo/DefaultTestCairoConfiguration.java
```diff
@@ -99,22 +99,17 @@ public long getWalApplyTableTimeQuota() {
...
 
     @Override
-    public boolean isDevModeEnabled() {
-        return true;
+    public boolean isColumnAliasExpressionEnabled() {
+        return false;
     }
 
     @Override
-    public boolean isMatViewEnabled() {
```
bluestreak01, line 107: why did you do this?

glasstiger: Without code change the build did not run the tests.
There is a condition in the yaml file.

I wanted the tests run so I can see if there is gc log in the output, so looked for a warning and fixed it.
That is redundant code, the overridden method is exactly the same.

glasstiger:
Updated rust versions for pgwire compat tests in https://github.com/questdb/questdb/pull/6182/commits/9bf5da4b773d7a9fe501bd41e978025983c20e89.
Cargo.lock and Cargo.toml were out of sync, and the picked up postgres lib version was not compatible with the rust version installed.

PR #6190, opened 2025-09-25
fix(ilp): avoid closing connection on protocol errors

puzpuzpuz:
We used to close the ILP/TCP connection and the writer on cast error like the below one:
```
2025-09-25T17:34:31.511806Z C i.q.c.l.t.LineTcpConnectionContext [208305913932] could not process line data 1 [table=weather, msg=could not write ILP message to WAL [tableName=weather, error=[-1] table: weather, column: col; cast error from protocol type: FLOAT to column type: STRING], errno=0]
```

The scenario is when you're trying to send a value of wrong type for the given column, i.e. a `float` instead of a `string` (`LineTcpReceiverTest.testColumnTypeStaysTheSameWhileColumnAdded`). This patch keeps the connection open if `line.tcp.disconnect.on.error` is set to `false`. The error is also no longer logged with critical level.


bluestreak01 (approved):

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 7 / 9 (77.78%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/LineTcpMeasurementScheduler.java|7|9|77.78%|

PR #6197, opened 2025-09-26
feat(sql): SQL hints for linear and indexed ASOF join algos

mtopolnik:
This PR:

1. Deprecates and ignores the `useFastAsOfJoin` configuration property. Now QuestDB always uses the fast asof join, unless disabled through a hint.
2. Deprecates for removal, but still honors, the existing SQL hints: `avoid_asof_binary_search` and `avoid_lt_binary_search`
3. Adds two new hints: `asof_linear_search` and `asof_index_search`

The new hints apply to both ASOF and LT JOIN variants. Instead of hinting to avoid an algo, the hints positively select algos:

- `asof_linear_search` selects the original algo with no binary search or index scan optimizations
- `asof_index_search` requests the algo that leverages the symbol index

### DIFF
--- a/core/src/test/java/io/questdb/test/griffin/AsOfJoinTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/AsOfJoinTest.java
@@ -1801,7 +1801,7 @@ public class AsOfJoinTest extends AbstractCairoTest {
     public void testJoinStringOnSymbolKey() throws Exception {
         assertMemoryLeak(() -> {
             executeWithRewriteTimestamp("CREATE TABLE x (sym STRING, ts #TIMESTAMP) TIMESTAMP(ts) PARTITION BY DAY", leftTableTimestampType.getTypeName());
-            executeWithRewriteTimestamp("CREATE TABLE y (sym SYMBOL, ts #TIMESTAMP) TIMESTAMP(ts) PARTITION BY DAY", rightTableTimestampType.getTypeName());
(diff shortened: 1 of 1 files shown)


core/src/main/resources/io/questdb/site/conf/server.conf
```diff
@@ -359,10 +359,6 @@ query.timeout=1m
 #cairo.sql.hash.join.light.value.page.size=128k
 #cairo.sql.hash.join.light.value.max.pages=2^31
 
-# Optimizations for key-ed ASOF JOINS. Set to false if you experience slowdowns with ASOF JOIN queries.
-# This can occur, for example, when the right-side table lacks matching keys or when the timestamp of the matching record
-# is not close to the timestamp of the left-side record.
-#cairo.sql.asof.join.fast=true
```
puzpuzpuz, line 365: Could you also remove it from `pkg/ami/marketplace/assets/server.conf`?

core/src/main/java/io/questdb/cairo/CairoConfiguration.java
```diff
@@ -780,6 +780,13 @@ default boolean isValidateSampleByFillType() {
 
     int maxArrayElementCount();
 
+    /**
+     * Deprecated and ignored. We always default to using Fast ASOF algorithms that involve
```
bluestreak01: can we remove this instead? There is a way to deprecate configuration option and alert the use instead of keeping this method around

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlHints.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|8|8|100.00%|

bluestreak01 (approved):

PR #6198, opened 2025-09-26
chore(core): fix SettingsProcessor

glasstiger:
Fix for intermittent test failures experienced in `SettingsEndpointTest`.
The problem was that `SettingsProcessor` sent an empty non-chunked response back to the client with HTTP OK and ContentLength=0, when preferences (instance name, type and description) were saved successfully.
HttpClient then received the header, parsed it, and when it time to read the response body, it checked the content length which was 0, and simply returned since there was nothing to receive.
This meant that the client considered the request finished before it was actually done by the server.
Solution is to always return a response body.
Considering that the content type of the response is JSON, this is actually required.

Fixing the above problem uncovered another issue.
Because there was no response body, fragmented response was not an issue.
However, after adding a response body, fragmentation tests started failing which are also fixed in this PR.

In practice these issues could have resulted in a visible problem in the web console only, if the user loaded the UI on a very slow connection, and received an error while updating preferences.
In this case the HTTP response would have had a body with the error message, and could have been got fragmented because of the slow connection.

### DIFF
--- a/core/src/main/java/io/questdb/cutlass/http/HttpResponseSink.java
+++ b/core/src/main/java/io/questdb/cutlass/http/HttpResponseSink.java
@@ -59,9 +59,11 @@ import static io.questdb.std.Chars.isBlank;
 import static java.net.HttpURLConnection.*;
 
 public class HttpResponseSink implements Closeable, Mutable {
+    private static final Utf8String EMPTY_JSON = new Utf8String("{}");
     private static final int HTTP_RANGE_NOT_SATISFIABLE = 416;
     private static final int HTTP_REQUEST_HEADER_FIELDS_TOO_LARGE = 431;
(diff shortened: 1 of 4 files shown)


glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 37 / 44 (84.09%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/http/processors/SettingsProcessor.java|16|23|69.57%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpResponseSink.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/SettingsProcessorState.java|10|10|100.00%|

bluestreak01 (approved):

PR #6208, opened 2025-09-30
feat(sql): optimized ASOF JOIN on single symbol key where RHS symbol is low-frequency

mtopolnik:
Keyed ASOF JOIN on a non-indexed symbol column uses linear scan to find the matching RHS row. For each LHS row, it repeats the RHS search from scratch. This works fine as long as the RHS table has plenty occurrences of all symbols.

However, if some symbol occurs in the RHS table only occasionally, a great many rows will have to be scanned to find it. Then, when we move on after finding it, and encounter the same symbol in a future LHS row, we'll have to repeat the entire scan, only to find the exact same RHS symbol. 

This PR saves time by remembering what it found before and reusing that knowledge when looking for the same symbol again. It also remembers what it _didn't_ find, skipping over the already-scanned region of the RHS table.


glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 135 / 142 (95.07%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java|118|125|94.40%|
|:large_blue_circle:|io/questdb/std/IntLongHashMap.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinFastRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|12|12|100.00%|

bluestreak01:
@coderabbit review

bluestreak01:
@coderabbitai review

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 184 / 206 (89.32%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/join/ChainedSymbolColumnAccessHelper.java|1|3|33.33%|
|:large_blue_circle:|io/questdb/griffin/engine/join/NoopColumnAccessHelper.java|2|4|50.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java|122|140|87.14%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleVarcharColumnAccessHelper.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/IntLongHashMap.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinFastRecordCursorFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleStringColumnAccessHelper.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|36|36|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleSymbolColumnAccessHelper.java|2|2|100.00%|

mtopolnik:
@coderabbit review

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 212 / 226 (93.81%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/join/ChainedSymbolColumnAccessHelper.java|1|3|33.33%|
|:large_blue_circle:|io/questdb/griffin/engine/join/NoopColumnAccessHelper.java|2|4|50.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java|132|141|93.62%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|54|55|98.18%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleVarcharColumnAccessHelper.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/IntLongHashMap.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinFastRecordCursorFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleStringColumnAccessHelper.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleSymbolColumnAccessHelper.java|2|2|100.00%|

puzpuzpuz:
@mtopolnik are you aware of scenarios when the new factory is slower than the old one? If not, maybe we should enable it by the default and have a hint to disable it just in case?

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
```diff
@@ -0,0 +1,355 @@
...
+import io.questdb.griffin.PlanSink;
+import io.questdb.griffin.SqlException;
+import io.questdb.griffin.SqlExecutionContext;
+import io.questdb.griffin.model.JoinContext;
+import io.questdb.std.IntLongHashMap;
+import io.questdb.std.Misc;
+import io.questdb.std.Numbers;
+import io.questdb.std.Rows;
+
+public final class AsOfJoinMemoizedRecordCursorFactory extends AbstractJoinRecordCursorFactory {
```
puzpuzpuz, line 81: nit: some javadoc clarifying what's special about this factory would be great here

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
```diff
@@ -0,0 +1,355 @@
...
+        Misc.free(slaveFactory);
+    }
+
+    private class AsOfJoinMemoizedRecordCursor extends AbstractKeyedAsOfJoinRecordCursor {
+
+        private static final long NOT_REMEMBERED = Long.MIN_VALUE;
+
+        // These three hashmaps are used in strict parallel, always putting the same key in each.
+        // When accessing them, we can look up the key index only once, and use it for all three maps.
+        private final IntLongHashMap symKeyToRowId = new IntLongHashMap();
```
puzpuzpuz: Having three hash tables means that we have 3x overhead for the keys. In case of high cardinality symbol column, the overhead would be noticeable. Also, the choice of the identity hash function in the hash table is not ideal considering that symbol keys are usually sequential integers - after modulo, the clusters are more likely with identity function. Consider switching to `OrderedMap`/`Unordered4Map` or implementing a new on-heap map.

core/src/main/java/io/questdb/std/IntLongHashMap.java
```diff
@@ -27,19 +27,15 @@
 import java.util.Arrays;
 
 public class IntLongHashMap extends AbstractIntHashSet {
-    private static final int noEntryValue = -1;
+    private static final long noEntryValue = -1;
```
puzpuzpuz, line 30: `noEntryValue` + `inc()` combination has a bug:
```
IntLongHashMap map = IntLongHashMap();
map.inc(42); // this call will increment a value on a non-existing key
```

We either need to check the key index in `inc()` (and `valueAt()`) instead of the value or initialize the `values` array with `noEntryValue` in `clear()` and `rehash()`. `IntIntHashMap` and maybe other hash tables exhibits the same problem.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
```diff
@@ -0,0 +1,355 @@
...
+    @Override
+    protected void _close() {
+        Misc.freeIfCloseable(getMetadata());
+        Misc.free(masterFactory);
+        Misc.free(slaveFactory);
+    }
+
+    private class AsOfJoinMemoizedRecordCursor extends AbstractKeyedAsOfJoinRecordCursor {
+
+        private static final long NOT_REMEMBERED = Long.MIN_VALUE;
```
puzpuzpuz, line 162: Usage of this constant looks weird: it's written in a single place and never used in any checks. Is this on purpose?

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
```diff
@@ -0,0 +1,355 @@
...
+
+                // move the slave cursor backwards
+                if (rowId > frameRowLo) {
+                    rowId--;
+                } else {
+                    // we exhausted this frame, let's try the previous one
+                    if (!slaveTimeFrameCursor.prev()) {
+                        // there is no previous frame, search space is exhausted
+                        long rememberedNoMatchRowId = -rememberedRowId - 1;
+                        if (!didJumpOverValidityPeriod || rowId < rememberedNoMatchRowId) {
```
puzpuzpuz: I kept the fuzz test running for a while and it never hit `rowId < rememberedNoMatchRowId` being `true`. Could you check? Maybe it's worth adding a non-fuzz test to hit this branch?

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
```diff
@@ -0,0 +1,355 @@
...
+            }
+
+            long rowId = slaveRecB.getRowId();
+            long frameRowLo = Rows.toRowID(slaveTimeFrame.getFrameIndex(), slaveTimeFrame.getRowLo());
+
+            for (; ; ) {
+                long slaveTimestamp = scaleTimestamp(slaveRecB.getTimestamp(slaveTimestampIndex), slaveTimestampScale);
+                boolean didExtendValidityPeriod = false;
+                boolean didJumpOverValidityPeriod = false;
+                if (slaveTimestamp >= validityPeriodStart && slaveTimestamp <= validityPeriodEnd) {
```
puzpuzpuz, line 360: I kept the fuzz test running for a while and it never hit `slaveTimestamp >= validityPeriodStart` being `false`. Could you check? Maybe it's worth adding a non-fuzz test to hit this branch?

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
```diff
@@ -0,0 +1,355 @@
...
+            int slaveKeyIndex = symKeyToRowId.keyIndex(slaveSymbolKey);
+            if (slaveKeyIndex < 0) {
+                if (onlyIfNew) {
+                    // We remember this symbol, but should memorize it only if new, so return.
+                    return;
+                }
+
+                // This is purely an assertion branch, to make sure our logic holds.
+                // Let's explain the logic! We'll use this notation:
+                //   ! ---- | we found the symbal at the start of this period
```
puzpuzpuz: nit:
```suggestion
                //   ! ---- | we found the symbol at the start of this period
```

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
```diff
@@ -0,0 +1,355 @@
...
+            }
+
+            symKeyToRowId.putAt(slaveKeyIndex, slaveSymbolKey, slaveRowId);
+            symKeyToValidityPeriodStart.putAt(slaveKeyIndex, slaveSymbolKey, slaveTimestamp);
+            symKeyToValidityPeriodEnd.putAt(slaveKeyIndex, slaveSymbolKey, masterTimestamp);
+        }
+
+        @Override
+        protected void performKeyMatching(long masterTimestamp) {
+            if (columnAccessHelper.isShortCircuit(masterRecord)) {
```
puzpuzpuz: We're doing symbol lookup twice: in the `isShortCircuit()` call and below. Can we somehow reuse the looked-up key?

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
```diff
@@ -0,0 +1,355 @@
...
+                    memorizeSymbolLocation(masterTimestamp, slaveTimestamp, slaveSymbolKey, rowId, false);
+                    break;
+                } else {
+                    // This isn't the symbol we're looking for, but memorize it anyway in the hope that some future
+                    // master row will need it.
+                    memorizeSymbolLocation(masterTimestamp, slaveTimestamp, thisSymbolKey, rowId, true);
+                }
+
+                // move the slave cursor backwards
+                if (rowId > frameRowLo) {
```
puzpuzpuz, line 451: We could assign the `frameRowIdLo` right here and remove all other places where we declare and update it:
```java
final long frameRowLo = Rows.toRowID(slaveTimeFrame.getFrameIndex(), slaveTimeFrame.getRowLo());
```

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
mtopolnik: It didn't need checking because another check ensured the stored map value was NOT_REMEMBERED, and other places exploit it being the min value of long. After some more changes, we now have an explicit check as well.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
mtopolnik: Indeed, I realized the same thing and this check is now only done in a different context, where it can actually occur.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
mtopolnik: I believe this is a leftover from the time I did'n assume masterTimestamp is monotinically increasing. I'm turning it into an assertion.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
```diff
@@ -0,0 +1,355 @@
...
+                    break;
+                }
+                int thisSymbolKey = slaveRecB.getInt(slaveSymbolColumnIndex);
+                if (thisSymbolKey == slaveSymbolKey) {
+                    record.hasSlave(true);
+                    // We found the symbol that we don't remember. Memorize it now.
+                    memorizeSymbolLocation(masterTimestamp, slaveTimestamp, slaveSymbolKey, rowId, false);
+                    break;
+                } else {
+                    // This isn't the symbol we're looking for, but memorize it anyway in the hope that some future
```
puzpuzpuz, line 445: Shouldn't be update the symbol values in both cases: when it's an unseen key and when it's an already seen one? In case of sparse timestamps, this could save us some jumps across the page frames.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
puzpuzpuz: Resolving: this would be more frequent that the current approach.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
```diff
@@ -0,0 +1,438 @@
...
+                    record.hasSlave(false);
+                    // Memorize that we didn't find the matching symbol by saving rowId as (-rowId - 1).
+                    memorizeSymbolLocation(masterTimestamp, slaveTimestamp, slaveSymbolKey, -rowId - 1, false);
+                    carefullyExtendScannedRange(masterTimestamp, slaveTimestamp, rowId);
+                    break;
+                }
+                int thisSymbolKey = slaveRecB.getInt(slaveSymbolColumnIndex);
+                if (thisSymbolKey == slaveSymbolKey) {
+                    record.hasSlave(true);
+                    // We found the symbol that we don't already remember. Memorize it now.
```
puzpuzpuz: nit: this would be less confusing
```suggestion
                    // We found a row that we don't already remember. Memorize it now.
```

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
```diff
@@ -0,0 +1,438 @@
...
+        sink.attr("condition").val(joinContext);
+        sink.child(masterFactory);
+        sink.child(slaveFactory);
+    }
+
+    @Override
+    protected void _close() {
+        Misc.freeIfCloseable(getMetadata());
+        Misc.free(masterFactory);
+        Misc.free(slaveFactory);
```
puzpuzpuz, line 156: nit: if we start using off-heap data structures in the cursor, we should close the `cursor` here. The problematic scenario: `cursor.of(masterCursor, slaveCursor, executionContext.getCircuitBreaker())` in `getCursor()` throws. Then the caller doesn't have the cursor instance and the only thing it can is closing the factory.

core/src/main/java/io/questdb/griffin/SqlHints.java
```diff
@@ -61,16 +57,12 @@ public static boolean hasAsOfLinearSearchHint(
...
+    public static boolean hasAsofMemoizedSearchHint(
             @NotNull QueryModel queryModel,
             @Nullable CharSequence tableNameA,
             @Nullable CharSequence tableNameB
     ) {
-        return hasHintWithParams(queryModel, ASOF_JOIN_AVOID_BINARY_SEARCH_HINT, tableNameA, tableNameB);
+        return hasHintWithParams(queryModel, ASOF_MEMOIZED_SEARCH_HINT, tableNameA, tableNameB);
     }
 
     /**
```
puzpuzpuz: `hasAvoidLtJoinBinarySearchHint` and `LT_JOIN_AVOID_BINARY_SEARCH_HINT` are now used, so we could remove them.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
```diff
@@ -0,0 +1,438 @@
...
+            super(
+                    columnSplit,
+                    nullRecord,
+                    masterTimestampIndex,
+                    masterTimestampType,
+                    slaveTimestampIndex,
+                    slaveTimestampType,
+                    configuration.getSqlAsOfJoinLookAhead()
+            );
+            ArrayColumnTypes keyTypes = new ArrayColumnTypes();
```
puzpuzpuz: nit: types could be moved into static fields, so that we don't allocate them for each cursor instance.

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
```diff
@@ -0,0 +1,438 @@
...
+            symbolTable = null;
+            rememberedSymbols.close();
+        }
+
+        @Override
+        public void of(RecordCursor masterCursor, TimeFrameRecordCursor slaveCursor, SqlExecutionCircuitBreaker circuitBreaker) {
+            super.of(masterCursor, slaveCursor, circuitBreaker);
+            symbolTable = slaveTimeFrameCursor.getSymbolTable(slaveSymbolColumnIndex);
+            rememberedSymbols.reopen();
+            rememberedSymbols.clear();
```
puzpuzpuz, line 210: Shouldn't we move this call to `toTop()` method? Same with the `scannedRangeMaxTimestamp = Long.MIN_VALUE; //...` assignments below - they should be done in `toTop()` too. I wonder why tests didn't highlight this.

```java
// reopen the map first since toTop() called by the below of() call clears it
rememberedSymbols.reopen();
super.of(masterCursor, slaveCursor, circuitBreaker);
```

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
mtopolnik: This serves as an optimization that avoids rescanning the depths of the slave table for a rarely occuring symbol. As soon as we hit the first such symbol, we'll also pick up any others on the way. But, as we proceed in the scan backwards, it's always the first symbol we see that is the most recent one. We must not overwrite that with an older symbol.

We might update this if we see it's a more recent symbol, though, but this requires rewriting `memorizeSymbolLocation`, because it currently assumes the invariant that we're encountering this symbol for the first time in our search backwards.

mtopolnik: After more thinking, I recalled this problem:

As soon as we lose the invariant that this is the only symbol between `masterTimestamp` and the place where we found it, we're lost and don't know the validity period of the symbol. We have this invariant when the symbol is the one we're looking for, but for these "drive-by" symbols, we don't know if we see it for the first time in the scanned region.

This forces the logic to stay the way it is.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 239 / 256 (93.36%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/join/ChainedSymbolColumnAccessHelper.java|1|3|33.33%|
|:large_blue_circle:|io/questdb/griffin/engine/join/NoopColumnAccessHelper.java|2|4|50.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java|159|171|92.98%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|54|55|98.18%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleVarcharColumnAccessHelper.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/IntLongHashMap.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinFastRecordCursorFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleStringColumnAccessHelper.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleSymbolColumnAccessHelper.java|2|2|100.00%|

mtopolnik:
> @mtopolnik are you aware of scenarios when the new factory is slower than the old one? If not, maybe we should enable it by the default and have a hint to disable it just in case?

I tried the query with four most common symbols (appearing in 87% of all rows in `prices`), and the Fast Asof algo was still slower. I'm not aware of a scenario where it's faster.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 268 / 284 (94.37%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/join/ChainedSymbolColumnAccessHelper.java|1|4|25.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/NoopColumnAccessHelper.java|2|5|40.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleStringColumnAccessHelper.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleVarcharColumnAccessHelper.java|6|7|85.71%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java|166|173|95.95%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|53|54|98.15%|
|:large_blue_circle:|io/questdb/std/IntLongHashMap.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinFastRecordCursorFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsofJoinColumnAccessHelper.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleSymbolColumnAccessHelper.java|15|15|100.00%|

core/src/main/java/io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java
```diff
@@ -0,0 +1,492 @@
...
+        // These track a contiguous range of slave timestamps that we've already scanned.
+        // This range doesn't cover everything we've scanned (there may be many disjoint ranges),
+        // but we guarantee we did scan everything inside it.
+        // Remembering this helps us avoid rescanning vast ranges of the slave table for
+        // symbols that occur way in the past from masterTimestamp.
+        private long scannedRangeMaxTimestamp = Long.MIN_VALUE;
+        private long scannedRangeMinRowId = Long.MAX_VALUE;
+        private long scannedRangeMinTimestamp = Long.MAX_VALUE;
+
+        private StaticSymbolTable symbolTable;
```
bluestreak01: unused

mtopolnik:
> refactoring in this PR introduced previously non-existent possibility of NPE in SQLOptimiser. Intellij produces warnings that should be investigated:

The changes I see in `SqlOptimiser` are unrelated to the warning. 

I think the reason why this is showing up now isn't any change in this PR, but an improvement to the static analysis in IDEA. It can see that a few lines above we have a check `if (innerVirtualModel != null)`, but don't have it later on, where this warning is raised. The reason why I think it's safe without the check is that the parameter `addColumnToInnerVirtualModel` is only true when `innerVirtualModel` is non-null. But I haven't 100% confirmed this.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 280 / 296 (94.59%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/join/ChainedSymbolColumnAccessHelper.java|0|4|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/NoopColumnAccessHelper.java|2|5|40.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleVarcharColumnAccessHelper.java|6|7|85.71%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleStringColumnAccessHelper.java|7|8|87.50%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java|165|171|96.49%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|54|55|98.18%|
|:large_blue_circle:|io/questdb/std/CharSequenceLongHashMap.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/LowerCaseCharSequenceIntHashMap.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/IntLongHashMap.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinFastRecordCursorFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsofJoinColumnAccessHelper.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/Utf8SequenceLongHashMap.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlHints.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SingleSymbolColumnAccessHelper.java|16|16|100.00%|
|:large_blue_circle:|io/questdb/std/CharSequenceIntHashMap.java|2|2|100.00%|

PR #6209, opened 2025-09-30
chore(core): use `std::isnan`

vil02:
For an unexperienced eye the code `v == v` might look a bit cryptic. I suggest to use the [`std::isnan`](https://en.cppreference.com/w/cpp/numeric/math/isnan.html) to determine if current value is not NaN.

### DIFF
--- a/core/src/main/c/share/vec_agg_vanilla.cpp
+++ b/core/src/main/c/share/vec_agg_vanilla.cpp
@@ -99,7 +99,7 @@ int64_t countDouble_Vanilla(double *d, int64_t count) {
     int64_t cnt = 0;
     for (; d < lim; d++) {
         const double v = *d;
-        if (v == v) {
+        if (!std::isnan(v)) {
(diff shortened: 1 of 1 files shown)


Saikiran3712:
The original code used the condition if (v == v) to filter out NaN values when performing operations like counting, summing, or averaging doubles. This works because in IEEE-754 floating-point arithmetic, NaN is the only value that is not equal to itself, so the check effectively skips NaN.

The code was later updated to use if (!std::isnan(v)). Both approaches have the same behavior, but std::isnan(v) is much clearer and directly communicates the intent of ignoring NaN values. While v == v is a clever trick, it can be confusing to readers and may trigger warnings in code reviews or static analysis tools.

In terms of performance, both methods are effectively equivalent, but std::isnan improves readability, maintainability, and portability across different compilers. Therefore, replacing v == v with std::isnan(v) is considered a best practice.

PR #6211, opened 2025-10-01
fix(core): fd leak in index writer initialization

puzpuzpuz:
Reproducible in `WalWriterFuzzTest.testWalWriteTinyO3Memory` with the following random seeds: `2574279280316L, 1759264733558L`.

The problem was in these lines:
```java
                kFdUnassigned = false;
                this.keyMem.of(ff, keyFd, null, ff.length(keyFd), MemoryTag.MMAP_INDEX_WRITER);
```

Here, the `keyFd` would leak due to the `kFdUnassigned` flag being set to `false` earlier in scenario when the `ff.length(keyFd)` call throws.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/BitmapIndexWriter.java
+++ b/core/src/main/java/io/questdb/cairo/BitmapIndexWriter.java
@@ -205,65 +205,66 @@ public class BitmapIndexWriter implements Closeable, Mutable {
             if (init) {
                 if (ff.truncate(keyFd, 0)) {
                     kFdUnassigned = false;
-                    this.keyMem.of(ff, keyFd, false, null, keyAppendPageSize, keyAppendPageSize, MemoryTag.MMAP_INDEX_WRITER);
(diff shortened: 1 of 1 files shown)


glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 30 / 35 (85.71%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/BitmapIndexWriter.java|30|35|85.71%|

puzpuzpuz:
> what is it with your dislike to `this.` ?

Its usage is inconsistent, even within a single method. Other than that, in most cases it's redudant and only makes the line longer and harder to read.
```java
if (this.keyMem.getLong(BitmapIndexUtils.KEY_RESERVED_OFFSET_SEQUENCE_CHECK) != this.keyMem.getLong(BitmapIndexUtils.KEY_RESERVED_OFFSET_SEQUENCE)) {
```

puzpuzpuz:
@bluestreak01 thanks for the review!

Issue #4120, opened 2024-01-11
Port remaining parallelizable aggregate functions to off-heap data structures


puzpuzpuz:
### Is your feature request related to a problem?

#4097 ported `min(str)`, `max(str)`, as well as `count_distinct()` for long, int, and IPv4 types to parallel GROUP BY, but some functions remain unported. Namely:
- [x] `count_distinct(uuid)`: requires a new long128 hash set, similar to the `GroupByLongHashSet` one
- [x] `count_distinct(long256)`: requires a new long256 hash set, similar to the `GroupByLongHashSet` one
- [ ] `approx_percentile(double)`: this one is tricky as we'll have to port HdrHistogram to become off-heap and flyweight
- [x] all `first`/`last` and `first_not_null`/`last_not_null` functions: to port them, we'll have to access and store row ids in the group by map
- [ ] `isOrdered(IPv4)`/`isOrdered(long)` functions: again, we need to track row ids
- [x] `ksum`/`nsum`

There is also `count_distinct(symbol)`, but we have early exit logic in that function (see #3974), so we don't want to port it, at least for now.

### Full Name:

Andrei Pechkurov

### Affiliation:

QuestDB

puzpuzpuz:
@nwoolmer that's awesome, thank you! Look forward to your contribution.

mcadariu:
Interested in picking up string_agg(str), if still available!

puzpuzpuz:
Both `string_agg` and `string_distinct_agg` are still available. Yet, please keep in mind that these two are tricky as they require writing sophisticated off-heap data structs such as string hash sets.

mcadariu:
Thanks @puzpuzpuz! 
In this PR I worked on `string_agg`: https://github.com/questdb/questdb/pull/6218. 
Looking forward to your feedback!

mcadariu:
In PR https://github.com/questdb/questdb/pull/6237 I took a shot at adding `GroupByArraySink`. This is a precursor step for porting `last(array)`. Thanks!

mcadariu:
In https://github.com/questdb/questdb/pull/6291 I've implemented `last(array)`. Note that it is dependent on https://github.com/questdb/questdb/pull/6237.

mcadariu:
@puzpuzpuz the refactoring of `first(array)` is ready: https://github.com/questdb/questdb/pull/6322

mcadariu:
[...]

PR #6218, opened 2025-10-02
perf(sql): enable parallel execution for string_agg()

mcadariu:
In https://github.com/questdb/questdb/pull/4097, a specialized allocator for aggregate functions was introduced and a couple of them migrated to use it. This PR follows up with the migration of `string_agg(str)` which was not included in the initial batch of migrated functions.


core/src/main/java/io/questdb/griffin/engine/functions/groupby/StringAggGroupByFunction.java
```diff
@@ -162,8 +204,10 @@ public void initValueIndex(int valueIndex) {
     @Override
     public void initValueTypes(ArrayColumnTypes columnTypes) {
         this.valueIndex = columnTypes.getColumnCount();
-        columnTypes.add(ColumnType.INT); // sink index
-        columnTypes.add(ColumnType.BOOLEAN); // null flag
+        // Reserve slot at valueIndex for sink pointer
+        columnTypes.add(ColumnType.LONG);
+        // Reserve slot at valueIndex+1 for null flag
```
puzpuzpuz: null can be encoded by zero pointer

core/src/main/java/io/questdb/griffin/engine/functions/groupby/StringAggGroupByFunction.java
```diff
@@ -181,9 +225,36 @@ public void setNull(MapValue mapValue) {
         mapValue.putBool(valueIndex + 1, true);
     }
 
+    @Override
+    public void setEmpty(MapValue mapValue) {
```
puzpuzpuz: `setEmpty` defaults to `setNull` which should be sufficient here unless I'm missing something

core/src/main/java/io/questdb/griffin/engine/functions/groupby/StringAggGroupByFunction.java
```diff
@@ -34,100 +34,132 @@
...
+            destValue.putBool(valueIndex + 1, false);
+            return;
+        }
+
+        // If we got here, both groups have data - need to merge them
+        sinkA.of(destPtr);
+        sinkB.of(srcPtr);
+
+        sinkA.putAscii(delimiter);
+        sinkA.put(sinkB);
```
puzpuzpuz: `assertSizeCompliance` or a similar size check should be done before merging the strings.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/StringAggGroupByFunction.java
```diff
@@ -34,100 +34,132 @@
...
-
         final CharSequence str = arg.getStrA(record);
+
         if (str != null) {
-            sink.put(str);
+            //Non-null value: initialize a new sink
+            sinkA.of(0);
+            sinkA.put(str);
+
+            totalMemoryUsed += sinkA.length() * 2;  // UTF-16: 2 bytes per char
```
puzpuzpuz: You could consider using the allocator to check how much memory the current thread (or even all threads) has allocated.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/StringAggGroupByFunction.java
```diff
@@ -181,9 +225,36 @@ public void setNull(MapValue mapValue) {
...
+
     @Override
     public boolean supportsParallelism() {
-        return false;
+        return true;
+    }
+
+    @Override
+    public boolean isThreadSafe() {
+        return UnaryFunction.super.isThreadSafe();
```
puzpuzpuz: This function can't be marked as thread-safe as it uses both flyweight sinks and thread-unsafe group by allocator.

puzpuzpuz (commented):
Please add multi-threaded tests to `ParallelGroupByFuzzTest` for both keyed and non-keyed cases. Otherwise, `merge()` method and lack of data races won't be verified by any test.

puzpuzpuz:
There is one more, very important aspect I overlooked during the initial review: the current version of the parallel `string_agg` function is non-deterministic. Serial (non-parallel) function from the `master` branch concatenates string values in the scan order. Parallel functions are called by worker threads which pick up tasks published by the "query owner" thread. Those tasks specify a page frame to be processed by the worker thread. A page frame is nothing more, but a slice of a partition, up to 1M rows in size.

So, long story short, current `string_agg` function may produce different results: depending on which worker thread pick up which tasks, it will accumulate different strings in the sinks to be later on merged with the sinks accumulated by other threads. This non-determinism is confusing for users and will complicate testing, so we should avoid it and this may be tricky: `computeFirst()`/`computeNext()` have `rowId` in their signature, a comparable row identifier that is already used by some of the aggregate functions like `first`/`last`, but in case of `string_agg` using it would be less trivial.

Unless you can think of a neat way to make the function deterministic, I'd suggest picking up another aggregate function instead of this one. Sorry more mentioning this function in #4120, I'll remove `string_agg` from the issue because it's non-trivial to make it parallel. Here are better candidates: we're currently missing `last(array)` function, as well as `first_not_null(array)`/`last_not_null(array)`. `last(array)` would require some kind of single-array sink class that would use `GroupByAllocator` to store the shape and the data from the given array.

mcadariu:
Good catch! Thanks for the explanation. Right away I can think of parallel merge sort, or robinhood hashing (each thread having a separate partition of keys). But this means extra overhead which might make sense only for large datasets.
Thanks for pointing me to other opportunities, will have a look!

mcadariu:
What about the following approach: we introduce something similar to `GroupByLongList` but which stores `(rowId, string)` pairs. During finalization, we sort by `rowId`.

puzpuzpuz:
> What about the following approach: we introduce something similar to `GroupByLongList` but which stores `(rowId, string)` pairs. During finalization, we sort by `rowId`.

This would require storing the strings somewhere (say, in an additional sink) and then we could store [rowId, offset] pairs in the list. Yet, all of this plus sort may be pretty heavyweight, so that I'm not sure if the thing would scale when compared with the current single-threaded code.

mcadariu:
Looked a bit at other DBs, I noticed you can provide `ORDER BY` like this:  `string_agg(s, ', ' ORDER BY s DESC)` 
* [DuckDB](https://duckdb.org/docs/current/sql/functions/aggregates#order-by-clause-in-aggregate-functions)
* [PostgreSQL](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-AGGREGATES)

Excerpt form DuckDB docs:

> However, there are some order-sensitive aggregates that can have non-deterministic results without ordering

They are upfront in the docs about non-deterministic results, and suggest that using ORDER BY will make it ordered (and hence deterministic).

mcadariu:
Hi @puzpuzpuz I had some time to see this through out of curiosity and get some numbers. Your initial intuition is correct. The bookkeeping and sorting that is needed for getting deterministic results introduces quite some overhead. While I did get about ~2x speedup when using multiple workers (unfortunately not linear either), the main point is I could not match the performance of the single-threaded version currently in master; therefore the PR in its current form would introduce a regression, so it seems like a no-go. I notice [DuckDB](https://duckdb.org/docs/current/sql/functions/aggregates#distinct-clause-in-aggregate-functions) and [PG](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-AGGREGATES) seem to not offer determinism guarantee, at least that is my interpretation of the docs. However, I can imagine that an QDB operation suddenly becoming non-deterministic, even though improved in other ways, might not amuse the users. If you agree, I'll close the PR in order to wrap up.

puzpuzpuz:
> I notice [DuckDB](https://duckdb.org/docs/current/sql/functions/aggregates#distinct-clause-in-aggregate-functions) and [PG](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-AGGREGATES) seem to not offer determinism guarantee, at least that is my interpretation of the docs. However, I can imagine that an QDB operation suddenly becoming non-deterministic, even though improved in other ways, might not amuse the users. If you agree, I'll close the PR in order to wrap up.

Yeah, I'd not introduce a breaking change. If the new function is a lot faster than the old one, we could introduce it as a separate SQL function which is explicitly non-deterministic in terms of the order. On the other hand, I'm not aware of an immediate need for such faster function from any of our users, so feel free to close the PR.

Issue #6202, opened 2025-09-29
Allow SAMPLE BY to use a long, rather than an INT for the units


javier:
### Is your feature request related to a problem?

At the moment the SAMPLE BY statement allows for an INT unit, which can limit the interval users want to use, specially if queries are programmatic rather than interactive. This can be a problem with nanoseconds support, as in nanoseconds you get to the limit of the INT in just about 2 seconds.

If I do `SAMPLE BY 3000000000n`, I am getting `the number is too large (invalid tolerance value [value=3000000000n, maximum=2147483647n])`

Users haven't complained about this just yet, but one user already complained about a similar issue with dateadd, (`there is no matching function `dateadd` with the argument types: (CHAR, LONG, TIMESTAMP)`) so I thought I'd add this here

### Describe alternatives you've considered.

change to a different unit. Which is workable when doing interactive queries, not so much for programmatic

### Full Name:

javier

### Affiliation:

questdb

xUser5000:
@javier @ideoma hi
Do I have permission to work on this? if so, could you please assign it to me?
Thanks

ideoma:
@xUser5000 yes, feel free to work on it. Fork the repo and create a PR. Please do not forget to add tests to the PR and make sure the CI test run is green.

vayunbiyani:
Hey @javier and @ideoma ,
I have fixed the core logic int [this pr](https://github.com/questdb/questdb/pull/6232) and I'd be grateful if you could provide me with some feedback on:
1. confirming if the logic for maxValue in [this diff](https://github.com/questdb/questdb/pull/6232/files#diff-81fab39eaabedbb8732c3e50800158d9bc29bbca0a7e42e4f7cea639cb0d69c6) is valid.
2. Suggestions on added tests for the update.

P.S. I apologise to @ xUser5000 if you were already working on it. I jumped straight into the fixing it as there were no assignes yet without seeing your comment.

ideoma:
The test case is in the issue description, e.g., this should not be an error anymore

If I do SAMPLE BY 3000000000n, I am getting the number is too large (invalid tolerance value [value=3000000000n, maximum=2147483647n])

amaydixit11:
Hey @ideoma @javier can i work on this?

krisp619:
@ideoma can you assign me this issue

shivamAmrutia:
Hey @javier @ideoma 

[...]

PR #6232, opened 2025-10-06
fix issue #6202

vayunbiyani:
Use Long for sampling instead of Int.


bluestreak01:
https://github.com/questdb/questdb/blob/master/CONTRIBUTING.md#before-you-submit

PR #6233, opened 2025-10-06
fix(sql): overflow in materialized view refresh interval estimate calculation

puzpuzpuz:
In case of ~1B row table and daily sampling `estimateRowsPerBucket()` could overflow and return a very small estimate leading to too large interval size chosen for the materialized view refresh.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
+++ b/core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
@@ -104,6 +104,15 @@ public class MatViewRefreshJob implements Job, QuietCloseable {
         }
     }
 
+    // kept public for testing
+    public static long estimateRowsPerBucket(long rows, long bucket, long partitionDuration, int partitionCount) {
+        if (partitionCount > 0) {
+            final double bucketToPartition = (double) bucket / partitionDuration;
+            return Math.max(1, (long) ((bucketToPartition * rows) / partitionCount));
+        }
+        return 1;
+    }
+
     @Override
     public void close() {
         LOG.info().$("materialized view refresh job closing [workerId=").$(workerId).I$();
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
```diff
@@ -127,10 +136,7 @@ private static long estimateRowsPerBucket(@NotNull TimestampDriver driver, @NotN
         final long rows = baseTableReader.size();
```
ideoma: nit: 
```suggestion
        final long tableRows = baseTableReader.size();
```

core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
```diff
@@ -104,6 +104,15 @@ public MatViewRefreshJob(int workerId, CairoEngine engine, int sharedQueryWorker
         }
     }
 
+    // kept public for testing
+    public static long estimateRowsPerBucket(long rows, long bucket, long partitionDuration, int partitionCount) {
```
ideoma: nit: 
```suggestion
    public static long estimateRowsPerBucket(long tableRows, long mvSampleByUnit, long partitionDuration, int partitionCount) {
```

core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
puzpuzpuz: Done in 99a840c2d7db514d8587535cc351f627ee8085cf

core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
puzpuzpuz: Done in 99a840c2d7db514d8587535cc351f627ee8085cf

puzpuzpuz:
@ideoma thanks for the review!

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|6|6|100.00%|

PR #6237, opened 2025-10-06
Add GroupByArraySink 

mcadariu:

In https://github.com/questdb/questdb/pull/4097, a specialized allocator for aggregate functions was introduced and a couple of them migrated to use it. 

This PR follows-up by adding `GroupByArraySink` which we will then use in a subsequent PR to port `last(array)`.


core/src/main/java/io/questdb/griffin/engine/groupby/GroupByArraySink.java
```diff
@@ -0,0 +1,106 @@
...
+import io.questdb.std.Unsafe;
+
+/**
+ * Specialized flyweight array sink used in {@link io.questdb.griffin.engine.functions.GroupByFunction}s.
+ * <p>
+ * Uses provided {@link GroupByAllocator} to allocate the underlying buffer. Grows the buffer when needed.
+ * <p>
+ * Buffer layout is the following:
+ * <pre>
+ * | totalSize | type (with dim count) |   dim0  |  dim1  | ... |     dimN-1     | array data |
```
puzpuzpuz: > type (with dim count)

We could save 4 bytes per sink value by asking the caller to provide the type when initializing the sink. That will work since array type stays the same for a group by function instance.

puzpuzpuz: `totalSize` also doesn't have to be 8 bytes: max buffer size for arrays is limited with `Integer.MAX_VALUE` (although it's not super-strictly enforced in the code). I understand that this layout is due to `ArrayTypeDriver`, but it'd be nice to introduce a more compact layout to be used in this sink.

mcadariu: Nice! Good to preserve memory bandwidth. I have removed the type from the layout.

mcadariu: Good one! I reduced `totalSize` to 4.

core/src/main/java/io/questdb/griffin/engine/groupby/GroupByArraySink.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+        return offset + nDims * INT_SIZE;
+    }
+
+    private void putData(ArrayView array, long offset) {
+        int elemSize = ColumnType.sizeOf(array.getElemType());
+        long srcOffset = array.getFlatViewOffset();
+        long length = array.getFlatViewLength();
+
+        long srcPtr = array.borrowedFlatView().ptr() + srcOffset * elemSize;
```
puzpuzpuz: `borrowedFlatView()`'s javadoc says:
> If called on the wrong implementation, this call will fail with a cast exception
so we can't use it without extra checks.

To keep things simple, we could introduce a couple of new methods to `ArrayTypeDriver`: `appendCompactPlainValue`/`getCompactPlainValue` - there are convenient utility methods in the driver, e.g. `appendToMemRecursive`. While the layout would only be used by this sink, we may find more places to use it in future.

core/src/main/java/io/questdb/griffin/engine/groupby/GroupByArraySink.java
```diff
@@ -0,0 +1,162 @@
...
+public class GroupByArraySink implements Mutable {
+    private static final long INT_SIZE = Integer.BYTES;
+    private final BorrowedArray borrowedArray = new BorrowedArray();
+    private final int type;
+    private GroupByAllocator allocator;
+    private long ptr;
+    private long allocatedSize;
+
+    public GroupByArraySink(int type) {
+        this.type = type;
```
puzpuzpuz, line 59: Does it make sense to introduce `long headerSize` and `int dims` fields in this class and initialize them in the constructor?
```java
this.dims = ColumnType.decodeArrayDimensionality(type);
this.headerSize = Integer.BYTES + dims * Integer.BYTES;
```

This way we can change `totalSize` to be `dataSize` and use the new fields in the calculations without having to calculate the number of dimensions (or even the size of element) many times - they're know at the object construction time and won't change.

core/src/test/java/io/questdb/test/griffin/engine/groupby/GroupByArraySinkTest.java
```diff
@@ -0,0 +1,223 @@
...
+                sink.put(array);
+
+                TestUtils.assertEquals(array, sink.getArray());
+            }
+        });
+    }
+
+    private ArrayView createArray(GroupByAllocator allocator, double... values) {
+        long totalSize = 4 + 4 + (values.length * 8L);
+        long ptr = allocator.malloc(8 + totalSize);
```
puzpuzpuz: Do we free this memory anywhere later on? Same question for the `create2DArray` utility.

core/src/main/java/io/questdb/griffin/engine/groupby/GroupByArraySink.java
mcadariu: Good one! `GroupByArraySink` is much simpler now and there is better separation of concerns. 
It has crossed my mind to do it like that originally, however I did not proceed because there would be only 1 "client" calling the new `ArrayTypeDriver` methods. However, indeed, there might be others in the future.

core/src/test/java/io/questdb/test/griffin/engine/groupby/GroupByArraySinkTest.java
mcadariu: I am passing in and using the `allocator`, which should take care of the free-ing, or am I missing something? 🤔

core/src/main/java/io/questdb/griffin/engine/groupby/GroupByArraySink.java
mcadariu: Right! Definitely makes sense to store what doesn't change in the fields and recompute only runtime data.

core/src/main/java/io/questdb/cairo/arr/ArrayTypeDriver.java
```diff
@@ -164,6 +164,31 @@ public static long appendPlainValue(long appendAddress, ArrayView value) {
...
+
+        Unsafe.getUnsafe().putInt(addr, totalSize);
+        addr += Integer.BYTES;
+
+        for (int i = 0; i < nDims; i++) {
+            Unsafe.getUnsafe().putInt(addr, arrayView.getDimLen(i));
+            addr += Integer.BYTES;
+        }
+
+        addr = appendToMemRecursive(arrayView, 0, 0, addr);
```
puzpuzpuz: Can we change this to
```java
        if (value.isVanilla()) {
            short elemType = value.getElemType();
            if (elemType == ColumnType.DOUBLE) {
                appendAddress = value.flatView().appendPlainDoubleValue(addr, value.getFlatViewOffset(), value.getFlatViewLength());
            } else {
                throw new UnsupportedOperationException("Unsupported array element type: " + elemType);
            }
        } else {
            appendAddress = appendToMemRecursive(value, 0, 0, addr);
        }
```
similar to `appendPlainValue()`? `FlatArrayView#appendPlainDoubleValue()` is a fast-path implementation, so we should keep using it.

core/src/main/java/io/questdb/cairo/arr/ArrayTypeDriver.java
```diff
@@ -164,6 +164,31 @@ public static long appendPlainValue(long appendAddress, ArrayView value) {
         return bytesWritten;
     }
 
+    public static void appendCompactPlainArray(long addr, ArrayView arrayView, int type) {
```
puzpuzpuz: nit: would be great to describe the layout difference between this method and `appendPlainValue` in a javadoc or plain comment. Same for `getCompactPlainArray`.

core/src/test/java/io/questdb/test/griffin/engine/groupby/GroupByArraySinkTest.java
puzpuzpuz: My bad, I misread the code as it was using `Unsafe`, not the group by allocator.

core/src/main/java/io/questdb/griffin/engine/groupby/GroupByArraySink.java
```diff
@@ -0,0 +1,122 @@
...
+
+    public void setAllocator(GroupByAllocator allocator) {
+        this.allocator = allocator;
+    }
+
+    private long computeRequiredSize(ArrayView array) {
+        if (array == null || array.isNull()) {
+            return INT_SIZE;
+        }
+        return headerSize + (long) array.getFlatViewLength() * elemSize;
```
puzpuzpuz: It makes sense to re-introduce `ArrayTypeDriver#getCompactPlainValueSize()`:
```java
    public static long getCompactPlainValueSize(@NotNull ArrayView value) {
        if (value.isNull()) {
            return Integer.BYTES;
        }
        long elemSize = ColumnType.sizeOf(ColumnType.decodeArrayElementType(type));
        long intBytes = Integer.BYTES; // using this to avoid (long) casts below
        return intBytes + // dataSize
                getDimCount() * intBytes // shape
                + getCardinality() * elemSize; // data
    }
```

The thing is that `getFlatViewLength()` can only be trusted in case of vanilla array (`isVanilla() == true`). Alternatively, we could keep the calculation local to this class, but if we have more users of `appendCompactPlainArray()`/`getCompactPlainArray()` in the future, we'll need the method in the driver anyway.

core/src/test/java/io/questdb/test/griffin/engine/groupby/GroupByArraySinkTest.java
```diff
@@ -0,0 +1,223 @@
...
+import io.questdb.griffin.engine.groupby.GroupByAllocator;
+import io.questdb.griffin.engine.groupby.GroupByArraySink;
+import io.questdb.std.Numbers;
+import io.questdb.std.Unsafe;
+import io.questdb.test.AbstractCairoTest;
+import io.questdb.test.tools.TestUtils;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class GroupByArraySinkTest extends AbstractCairoTest {
```
puzpuzpuz, line 41: Could you add a smoke test for the non-vanilla array input, e.g. `DerivedArrayView`?

core/src/main/java/io/questdb/cairo/arr/ArrayTypeDriver.java
mcadariu: Ah yes sure - I have noticed it, thanks for confirming that it applies here too.

core/src/main/java/io/questdb/cairo/arr/ArrayTypeDriver.java
mcadariu: You're right, I've added it, let me know if I overdid it.

core/src/main/java/io/questdb/griffin/engine/groupby/GroupByArraySink.java
mcadariu: Ah right. I've opted to add a `getCompactPlainValueSize` in the `ArrayTypeDriver` for separation of concerns

core/src/test/java/io/questdb/test/griffin/engine/groupby/GroupByArraySinkTest.java
mcadariu: Done, added `testPutNonVanillaArray`

core/src/test/java/io/questdb/test/griffin/engine/groupby/GroupByArraySinkTest.java
```diff
@@ -0,0 +1,251 @@
...
+                DerivedArrayView derivedArray = new DerivedArrayView();
+                derivedArray.of(vanillaArray);
+                derivedArray.transpose();
+
+                Assert.assertFalse(derivedArray.isVanilla());
+
+                sink.put(derivedArray);
+
+                ArrayView result = sink.getArray();
+                Assert.assertTrue(result.isVanilla());
```
puzpuzpuz, line 216: Would be great to verify the sink contents too:
```java
TestUtils.assertEquals(derivedArray, result);
```

puzpuzpuz:
Linter step is unhappy with javadoc formatting. Could you add the missing padding?
```
--- a/core/src/main/java/io/questdb/cairo/arr/ArrayTypeDriver.java
+++ b/core/src/main/java/io/questdb/cairo/arr/ArrayTypeDriver.java
@@ -179,9 +179,9 @@ public class ArrayTypeDriver implements ColumnTypeDriver {
      * and a 4-byte type field. This compact format omits the type (known from context) and
      * uses a 4-byte dataSize instead of 8-byte totalSize.
      *
-     * @param addr the memory address to write to
-     * @param value the array to append
-     * @param nDims the number of dimensions
+     * @param addr     the memory address to write to
+     * @param value    the array to append
+     * @param nDims    the number of dimensions
      * @param elemSize the size of each element in bytes
      */
     public static void appendCompactPlainValue(long addr, ArrayView value, int nDims, int elemSize) {
@@ -342,8 +342,8 @@ public class ArrayTypeDriver implements ColumnTypeDriver {
      * This is the counterpart to {@link #appendCompactPlainValue} and differs from {@link #getPlainValue}
      * which expects an 8-byte size field and a type field in the layout.
      *
-     * @param addr the memory address to read from
-     * @param type the encoded array type (provided from context)
+     * @param addr  the memory address to read from
+     * @param type  the encoded array type (provided from context)
      * @param nDims the number of dimensions
      * @param value the borrowed array to populate
      * @return the populated array
```

puzpuzpuz:
The PR is almost there. Great job!

core/src/test/java/io/questdb/test/griffin/engine/groupby/GroupByArraySinkTest.java
mcadariu: Yes, I wrote it like that initially! The test surprisingly didn't pass. I can't quite work out yet exactly if it is a deficiency of `TestUtils` or an actual correctness problem for the current PR.

I have pushed an update. I now create a new `expected` array for which `assertEquals` returns true, which at least documents how the array comes out from `getArray()`. How would you suggest I proceed?

puzpuzpuz: TBH it would be great to find out why `assertEquals` didn't pass - could be the case that this method has a bug.

mcadariu: OK, will have a look!

puzpuzpuz: Thanks!

mcadariu: I found and addressed two issues in `assertEqualsRecursive`: 
* `flatViewOffset` was added twice (`getDouble()` in `ArrayView` already does that) 
*  it was assuming stride=1 for the last dimension, which is only true for vanilla arrays

puzpuzpuz (approved):
Thanks for the solid contribution!

mcadariu:
Thanks for the clear and quick feedback on this PR!

mcadariu:
@puzpuzpuz thanks for the approval! Is there anything else I can do for this PR? If it gets merged, I can proceed with doing `last(array)`.

puzpuzpuz:
> @puzpuzpuz thanks for the approval! Is there anything else I can do for this PR? If it gets merged, I can proceed with doing `last(array)`.

Nope, nothing to do here. I don't have the merge permissions and the rest of the team is pretty busy, so it'll take some time. You could create a separate PR that is dependent on this one and, once this one is merged, simply do a rebase.

mcadariu:
Got it, thanks!

PR #6238, opened 2025-10-06
Update USER to be a number

wkbrd:
This addresses a security concern on some Kubernetes managment platforms, like OpenShift, that require that the USER must be non-root (and this can only be confirmed if a numeric USER field is used).

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -75,7 +75,7 @@ RUN groupadd -g 10001 questdb && \
     mkdir -p /var/lib/questdb && \
     chown -R questdb:questdb /var/lib/questdb
 
-USER questdb
+USER 10001
 
 WORKDIR /var/lib/questdb
 


bluestreak01:
could you please elaborate on what problem does this PR address, perhaps include/describe a reproducer for the problem

wkbrd:
> could you please elaborate on what problem does this PR address, perhaps include/describe a reproducer for the problem

We have found it to be a Kubernetes best practice to use a non-zero UID with the USER directive.  Some Kubernetes engines when running in a secure mode in conjunction with the 'runAsNonRoot: true' will verify that the container is not configured to run as root.  Since user names can be mapped to uid 0 (i.e., multiple username can be root), the only way to ensure that a username is not root is to specify a UID (a number).  Therefore, any non-zero UID with USER is considered better.

It is mentioned in one of the comments here: https://stackoverflow.com/questions/51544003/using-runasnonroot-in-kubernetes.

Containers managed by Bitnami (https://github.com/bitnami/containers) have consistently implemented this broadly for their catalog.

sklarsa:
We account for this in our [docker-entrypoint.sh](https://github.com/questdb/questdb/blob/bdd0fbf6784312962761566bba26665e5fb8ab6e/core/docker-entrypoint.sh#L145-L148) by running `gosu` to step down into a non-root user before starting the database process.

Our entrypoint contains helper logic that assumes the container is running as root in order to ensure the correct ownership of database directories. This allows for the widest compatibility across runtime environments to prevent permission-related errors on database startup.  This method is used in several other major database Dockerfiles, including PostgreSQL.

More security-conscious users can use the `-rhel` line of Docker images which sets `USER` as a directive in the Dockerfile, or adjust your container's security context in its runtime environment.

Closing this to maintain as wide compatibility as possible for the main Docker image.

wkbrd:
@sklarsa : If a user runs with a securityContext where the runAsNonRoot = true but does NOT set a runAsUser (null), then the following error will happen in Kubernetes (tested with k3s):

> CreateContainerConfigError (container has runAsNonRoot and image has non-numeric user (questdb), cannot verify user is non-root (pod: "...)", container: ...))

With the change to a numeric UID, this error would go away.

sklarsa:
@wkbrd that's addressed in the default securityContext values in the helm chart PR https://github.com/questdb/questdb-kubernetes/pull/223. Since this Dockerfile is used not just in k8s, but in all container runtimes, I'd like to keep it as widely-compatible and as easy-to-use as possible.

PR #6243, opened 2025-10-08
fix(core): potential connection leakage on Windows

puzpuzpuz:
The connections were leaked here due to broken operation id order:
https://github.com/questdb/questdb/blob/bdd0fbf6784312962761566bba26665e5fb8ab6e/core/src/main/java/io/questdb/network/IODispatcherWindows.java#L81-L84

The operation id order was broken due to missing id assignment here (Linux and OSX dispatchers have that assignment):
https://github.com/questdb/questdb/blob/bdd0fbf6784312962761566bba26665e5fb8ab6e/core/src/main/java/io/questdb/network/IODispatcherWindows.java#L149-L151

The bug may have manifested in the following scenario: you open a TCP connection, but write to it only after some time, so that heartbeat mechanism has the time to kick in.

Also, includes the following:
* Moves operation and event id generation code to `AbstractIODispatcher`, so that it's common across all I/O dispatcher implementations.
* OSX dispatcher used an `int` for operation id which is easy to overflow. On overflow, we'd face the same problem with broken operation id order invariant in `pending` and `pendingHeartbeat` lists. This is fixed by switching to `long` counter as in the other two dispatchers.
* Marks listener field as `volatile` in all pool implementations to avoid data races in tests.

### DIFF
--- a/core/src/main/java/io/questdb/network/IODispatcherOsx.java
+++ b/core/src/main/java/io/questdb/network/IODispatcherOsx.java
@@ -490,7 +490,7 @@ public class IODispatcherOsx<C extends IOContext<C>> extends AbstractIODispatche
             return lastError;
         }
 
-        public KeventWriter readFD(long fd, long id) {
+        public void readFD(long fd, long id) {
             kqueue.setWriteOffset(offset);
             kqueue.readFD(fd, id);
             offset += KqueueAccessor.SIZEOF_KEVENT;
@@ -498,7 +498,6 @@ public class IODispatcherOsx<C extends IOContext<C>> extends AbstractIODispatche
                 register(index);
                 index = offset = 0;
             }
-            return this;
         }
 
         public KeventWriter removeReadFD(long fd) {


glasstiger:
#### [PR Coverage check]


core/src/main/java/io/questdb/cairo/pool/AbstractPool.java
```diff
@@ -44,7 +44,7 @@ public abstract class AbstractPool implements Closeable {
     private final long inactiveTtlUs;
     @SuppressWarnings({"FieldCanBeLocal", "FieldMayBeFinal"})
     private volatile int closed = FALSE;
-    private PoolListener eventListener;
+    private volatile PoolListener eventListener;
```
bluestreak01, line 48: could we have a comment here explaining volatile? I would imagine if it is removed, we wouldn't even notice

bluestreak01:
@coderabbit review

core/src/main/java/io/questdb/cairo/pool/AbstractPool.java
puzpuzpuz: Sure. Added in 8ca493ba7b58bf49b5d8ef0d40312c4da144ff2d

core/src/main/java/io/questdb/network/Kqueue.java
```diff
@@ -70,20 +69,18 @@ public void close() {
         Unsafe.free(this.eventList, bufferSize, MemoryTag.NATIVE_IO_DISPATCHER_RSS);
     }
 
-    public int getData() {
+    public long getData() {
         return Unsafe.getUnsafe().getInt(readAddress + KqueueAccessor.DATA_OFFSET);
     }
```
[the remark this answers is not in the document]
puzpuzpuz: Fixed in d6f217a42f83e75767fe42324a593cfc0dfa58b5

glasstiger:
#### [PR Coverage check]

:disappointed: **fail** : 15 / 71 (21.13%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/network/IODispatcherWindows.java|0|8|00.00%|
|:large_blue_circle:|io/questdb/network/IODispatcherOsx.java|0|28|00.00%|
|:large_blue_circle:|io/questdb/network/Kqueue.java|0|4|00.00%|
|:large_blue_circle:|io/questdb/network/IODispatcherLinux.java|9|25|36.00%|
|:large_blue_circle:|io/questdb/network/AbstractIODispatcher.java|6|6|100.00%|

bluestreak01:
@coderabbit review. Although this is merged, it is possible a bug was introduced. Events from IODispatcher might be lost occasionally. Re-check the whole change

PR #6254, opened 2025-10-09
fix(sql): incorrect results produced by SAMPLE BY with mixed timestamp and aggregate expressions

### LINKED ISSUE #6239: SAMPLE BY returns unordered rows
### To reproduce

```sql
CREATE TABLE x (ts TIMESTAMP) TIMESTAMP(ts) PARTITION BY DAY;

INSERT INTO x VALUES ('2010-01-01T01'),('2020-01-01T01'),('2030-01-01T01');

SELECT ts, count()::double / datediff('h', ts, dateadd('d', 1, ts, 'Europe/Copenhagen')) AS Coverage
FROM 'x'
SAMPLE BY 1d ALIGN TO CALENDAR TIME ZONE 'Europe/Copenhagen';
```

The SELECT produces the following result set:
| ts                          | Coverage             |
| --------------------------- | -------------------- |
| 2019-12-31T23:00:00.000000Z | 0.041666666666666664 |
| 2029-12-31T23:00:00.000000Z | 0.041666666666666664 |
| 2009-12-31T23:00:00.000000Z | 0.041666666666666664 |

Notice that the output is not ordered by the timestamp it must have been the case due to SAMPLE BY.

### QuestDB version:

9.1

Ubuntu 24.04

ext4

### Full Name:

Andrei Pechkurov

### Affiliation:

QuestDB

- [x] Yes, I have

### Additional context

_No response_

puzpuzpuz:
Fixes #6239

### SAMPLE BY rewrite fix 1

SAMPLE BY to parallel GROUP BY rewrite was not handling mixed timestamp and aggregate function expressions. This was leading to unordered and incorrect results returned from the query.

Example of a problematic query:
```sql
SELECT ts, count()::double / datediff('h', ts, dateadd('d', 1, ts, 'Europe/Copenhagen')) AS Coverage
FROM 'x'
SAMPLE BY 1d ALIGN TO CALENDAR TIME ZONE 'Europe/Copenhagen';
```

The above query was incorrectly rewritten into something like this (note that this is a simplified representation, without the actual GROUP BY details in the nested query):
```sql
SELECT ts, count()::double / datediff('h', ts, dateadd('d', 1, ts, 'Europe/Copenhagen')) AS Coverage
FROM (
  SELECT ts
  FROM 'x'
  SAMPLE BY 1d ALIGN TO CALENDAR TIME ZONE 'Europe/Copenhagen'
);
```

The fix disables the rewrite for such queries.

### SAMPLE BY rewrite fix 2

The rewrite could throw away columns from the result set in case of certain expressions involving the designated timestamp column. Here is an example:
```sql
SELECT ts, ts2, ts - ts2 AS ts_diff, count() AS c
from x
SAMPLE BY 1d;
```

Before this patch, the last column (`c`) would not be returned from the query.


bluestreak01:
@coderabbit review

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -5393,10 +5423,10 @@ private QueryModel rewriteSampleBy(@Nullable QueryModel model, @Transient SqlExe
 
                 if ((wrapAction & SAMPLE_BY_REWRITE_WRAP_ADD_TIMESTAMP_COPIES) != 0 && needRemoveColumns > 0) {
                     model = wrapWithSelectModel(model, model.getBottomUpColumns().size() - needRemoveColumns);
-                    addColumnToSelectModel(model, tempList, insetColumnAliases, timestampAlias);
+                    addColumnToSelectModel(model, tempIntList, tempColumns, timestampAlias);
```
bluestreak01, line 5426: could we make it more obvious that the 3rd arg the actually a class member. It is hard to comprehend why `tempColumns2` is used inside and not `tempColumns`

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -164,11 +163,13 @@ public class SqlOptimiser implements Mutable {
     private final ObjList<RecordCursorFactory> tableFactoriesInFlight = new ObjList<>();
     private final FlyweightCharSequence tableLookupSequence = new FlyweightCharSequence();
     private final IntHashSet tablesSoFar = new IntHashSet();
+    private final BoolList tempBoolList = new BoolList();
     private final ObjList<QueryColumn> tempColumns = new ObjList<>();
+    private final ObjList<QueryColumn> tempColumns2 = new ObjList<>();
```
bluestreak01, line 167: this is not cleared in `clear()` method

puzpuzpuz:
> Some comments are outside the diff and can’t be posted inline due to platform limitations.

This comment was spot-on. The fix is in 4a9754306529f4d5bf0f346e004d7a189eed7f51

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
puzpuzpuz: Those clear calls are not a must since the collections are cleared before the usage anyway, but it's always nice to keep the code consistent. Added this one in eea917811dbb778cd5620838815431699dea7147

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
puzpuzpuz: I could introduce a local variable like `var insertColumnAliases = tempColumns;`, but IDEA will be complaining about redundant variable. Did you mean this or something else?

bluestreak01: we have to rethink the whole approach

puzpuzpuz:
> Right now the whole thing degrades in performance massively.

This is not degradation as such queries were not supported by the rewrite. Better to have correct results rather than get some garbage fast. The extraction must be implemented optimally, i.e. with aggregate reuse, so I suggest merging this patch to get a couple critical bugs fixed while I'm going to start working on mixed ts + aggregate expressions support in a separate PR. WDYT?

puzpuzpuz:
Failed test is #6266, unrelated with these changes. I can reproduce locally, so going to fix it in a separate branch.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 61 / 63 (96.83%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|61|63|96.83%|

PR #6258, opened 2025-10-10
fix(sql): overflow in materialized view refresh step calculation on large sample by interval

### LINKED ISSUE #6257: Overflow in materialized view refresh step calculation on large sample by interval
### To reproduce

Minimal reproducer:
```sql
create table x (ts timestamp) timestamp(ts) partition by day;
insert into x values('2000-01-01T12:00'),('2010-01-01T12:00'),('2020-01-01T12:00');
create materialized view x_1y as select ts, first(ts) as first_ts from x sample by 10y;
```

The view gets marked as invalid after the initial refresh with the `timestamp out of replace range` reason.

### QuestDB version:

9.1

Ubuntu 24.04

ext4

### Full Name:

Andrei Pechkurov

### Affiliation:

QuestDB

- [x] Yes, I have

### Additional context

_No response_

puzpuzpuz:
Fixes #6257

`SampleByIntervalIterator` got an overflow when calculating the high boundary on large SAMPLE BY intervals and relatively large step value. This is fixed by introducing a new `cairo.mat.view.max.refresh.step` configuration property (1y by default). The property defines maximum step interval after which refresh falls back to a single SAMPLE BY interval. Example: if refresh job calculates the step of 100 SAMPLE BY 2y intervals, after the check it will fall back to a single 2y interval to avoid overflow issues.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
+++ b/core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
@@ -127,6 +127,14 @@ public class MatViewRefreshJob implements Job, QuietCloseable {
         return processNotifications();
     }
 
+    private static long approxStepDuration(long step, long approxBucketSize) {
+        try {
+            return Math.multiplyExact(step, approxBucketSize);
+        } catch (ArithmeticException ignore) {
+            return Long.MAX_VALUE;
+        }
+    }
+
     /**
      * Estimates density of rows per SAMPLE BY bucket. The estimate is not very precise as
      * it doesn't use exact min/max timestamps for each partition, but it should do the job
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/questdb/cairo/CairoConfiguration.java
```diff
@@ -317,6 +317,8 @@ default IOURingFacade getIOURingFacade() {
 
     int getMatViewMaxRefreshRetries();
 
+    long getMatViewMaxRefreshStepUs();
```
ideoma, line 320: nit: why micros? Why not seconds or days? 
Do we practically need this configured, what is the situation where 1 year will not work?

core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
```diff
@@ -422,7 +422,16 @@ private RefreshContext findRefreshIntervals(
             final TimestampSampler timestampSampler = viewDefinition.getTimestampSampler();
             final long rowsPerBucket = estimateRowsPerBucket(driver, baseTableReader, timestampSampler.getApproxBucketSize());
             final int rowsPerQuery = configuration.getMatViewRowsPerQueryEstimate();
-            final int step = Math.max(1, (int) (rowsPerQuery / rowsPerBucket));
+            int step = Math.max(1, (int) (rowsPerQuery / rowsPerBucket));
+            try {
+                final long approxStepDuration = Math.multiplyExact(step, timestampSampler.getApproxBucketSize());
+                if (approxStepDuration > driver.fromMicros(configuration.getMatViewMaxRefreshStepUs())) {
+                    // the step is too large, fallback to step of a single SAMPLE BY interval
```
ideoma: Please add the reason why in the comment

core/src/main/java/io/questdb/cairo/CairoConfiguration.java
puzpuzpuz: > why micros? Why not seconds or days?

Micros are convenient for the comparison done in the refresh job. The user can specify arbitrary time unit in the config file.

> Do we practically need this configured, what is the situation where 1 year will not work?

In some very niche cases like a sparse in the number of rows, but wide in time table configuring this value may be helpful. Other than that, no strong reason to have a config prop for it, but in principle we tend to introduce config props instead of hardcoded values when there is a small chance that the value may need to be changed as a workaround.

core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
puzpuzpuz: Added the clarification in 71150d41bb1f5f6ce7babde07117ac9671c55453

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 12 / 15 (80.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|8|10|80.00%|

puzpuzpuz:
@ideoma thanks for the snappy review!

PR #6264, opened 2025-10-10
fix(sql): fix NullPointerException in lead() window function when using bind variables

nwoolmer:
The `lead` window function has three parameters, the third of which is a default placeholder value. From docs:

```
lead(value [, offset [, default]]) [(IGNORE|RESPECT) NULLS]
OVER ([PARTITION BY partition_expression] [ORDER BY sort_expression])
```

When using a bind variable in this position, i.e. `lead(value, 1, ?)`, the value was not initialised correctly, leading to null reference exceptions upon its use.


core/src/main/java/io/questdb/griffin/engine/functions/window/LeadLagWindowFunctionFactoryHelper.java
```diff
@@ -94,6 +94,7 @@ static Function newInstance(int position,
             }
 
             defaultValueExtraChecker.check(defaultValue);
+            defaultValue.init(null, sqlExecutionContext);
```
kafka1991: 1. better add an `init` function overload both in `BaseLeadFunction` and `BaseLagFunction`, and initialize the `defaultValue` function within it.
2. `BaseLeadFunction` and `BaseLagFunction` need to call `Misc.free(defaultValue)` in the `close` method.
3. The second argument `offset` function also requires same handling (it also can be a runtime constant).

nwoolmer: 3. It cannot be a runtime constant currently, as we try to pull the value before initialisation, and then pass a long around. We can refactor this to make it possible though.

```java
  final Function offsetFunc = args.getQuick(1);
            if (!offsetFunc.isConstant() && !offsetFunc.isRuntimeConstant()) {
                throw SqlException.$(argPositions.getQuick(1), "offset must be a constant");
            }

            offset = offsetFunc.getLong(null);
            if (offset < 0) {
                throw SqlException.$(argPositions.getQuick(1), "offset must be a positive integer");
            }
```

core/src/main/java/io/questdb/griffin/engine/functions/window/LeadLagWindowFunctionFactoryHelper.java
```diff
@@ -196,6 +198,12 @@ public int getPassCount() {
             return WindowFunction.ZERO_PASS;
         }
 
+        @Override
+        public void init(SymbolTableSource symbolTableSource, SqlExecutionContext executionContext) throws SqlException {
+            super.init(symbolTableSource, executionContext);
+            defaultValue.init(null, executionContext);
```
kafka1991: Can we pass (first argument) in the `symbolTableSource` here instead of null?

core/src/main/java/io/questdb/griffin/engine/functions/window/LeadLagWindowFunctionFactoryHelper.java
```diff
@@ -176,6 +177,7 @@ public BaseLagFunction(Function arg, Function defaultValueFunc, long offset, Mem
         public void close() {
             super.close();
             buffer.close();
+            Misc.free(defaultValue);
```
kafka1991, line 180: Sorry, I missing something earlier, `BaseLeadOverPartitionFunction` and `BaseLagOverPartitionFunction` also need to same handling.

core/src/main/java/io/questdb/griffin/engine/functions/window/LeadLagWindowFunctionFactoryHelper.java
kafka1991: Either way is fine, but if you decide not to do it in this PR, you can remove the condition `!offsetFunc.isRuntimeConstant()`.

nwoolmer: Will remove it, we can revisit if there is a need

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 21 / 21 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/LeadLagWindowFunctionFactoryHelper.java|21|21|100.00%|

nwoolmer:
Thanks for the review!

PR #6275, opened 2025-10-14
fix(sql): breaking change 💥- invalid column error returned from GROUP BY on joined tables

### LINKED ISSUE #6273: Invalid column error returned from GROUP BY joined tables
### To reproduce

The below query returns `Invalid column: t2.price` error:
```sql
CREATE TABLE trades (sym symbol, price double, amount double, ts timestamp) timestamp(ts) partition by day;
CREATE TABLE trades2 (sym symbol, price double, amount double, ts timestamp) timestamp(ts) partition by day;

SELECT t1.ts, t1.price, t2.price / sum(t1.amount) FROM trades t1 JOIN trades2 t2 ON (sym);
```

A similar error is returned when `t2.price / sum(t1.amount)` is changed to `t1.price / sum(t1.amount)`

### QuestDB version:

latest master

Ubuntu 24.04

ext4

### Full Name:

Andrei Pechkurov

### Affiliation:

QuestDB

- [x] Yes, I have

### Additional context

_No response_

puzpuzpuz:
Fixes #6273

Other than that, SAMPLE BY queries with selected expressions mixing timestamp and aggregate functions now use multi-threaded execution plan. Example of such query:
```sql
SELECT ts, count()::double / datediff('h', ts, dateadd('d', 1, ts, 'Europe/Copenhagen')) AS Coverage
FROM 'x'
SAMPLE BY 1d ALIGN TO CALENDAR TIME ZONE 'Europe/Copenhagen';
```

Also improves the way we choose execution plan for SAMPLE BY and GROUP BY queries. Namely, now we avoid adding duplicate group keys and, instead, reuse the existing ones. This means that in a query like the following one:
```sql
SELECT symbol, price, price / sum(amount) FROM trades WHERE timestamp in today();
```
we now end up with `symbol, price` as the group keys instead of `symbol, price, price price1` (notice the extra `price1` key).

### Breaking change

When designated timestamp is selected multiple times, under different aliases, the ordered (sorted in asc order) column picked up by the query engine will be the first one. For instance, in the below example column `ts1` will be marked as the ordered column:
```sql
select ts as ts1, ts as ts2, price, amount
from trades;
```

If one of the selected columns is used as the first ORDER BY column, it's always preferred, no matter what its position is in the `SELECT` clause:
```sql
-- here, ts2 is used as the ordered column
select ts as ts1, ts as ts2, price, amount
from trades
order by ts2;
```

Otherwise, unaliased column, i.e. the selected column matching with the designated column name, is preferred.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/SqlOptimiser.java
+++ b/core/src/main/java/io/questdb/griffin/SqlOptimiser.java
@@ -1876,6 +1876,7 @@ public class SqlOptimiser implements Mutable {
             // add column to both models
             addColumnToTranslatingModel(column, translatingModel, innerVirtualModel, baseModel);
             if (addColumnToInnerVirtualModel) {
+                assert innerVirtualModel != null;
                 ExpressionNode innerToken = expressionNodePool.next().of(LITERAL, alias, node.precedence, node.position);
                 QueryColumn innerColumn = queryColumnPool.next().of(alias, innerToken);
                 innerVirtualModel.addBottomUpColumn(innerColumn);
@@ -2990,10 +2991,6 @@ public class SqlOptimiser implements Mutable {
         if (nested != null) {
             moveWhereInsideSubQueries(nested);
         }
-    }
-
-    private QueryColumn nextColumn(CharSequence alias, CharSequence column, int position) {
-        return SqlUtil.nextColumn(queryColumnPool, expressionNodePool, alias, column, position);
     }
 
     private QueryColumn nextColumn(CharSequence name) {
--- a/core/src/test/java/io/questdb/test/griffin/engine/groupby/SampleByTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/engine/groupby/SampleByTest.java
@@ -96,7 +96,8 @@ public class SampleByTest extends AbstractCairoTest {
             x::boolean as t,\
             x::timestamp as n,\
             FROM long_sequence(480)
-            ) timestamp(ts)""";
+            ) timestamp(ts)
+            """;
     private static final Log LOG = LogFactory.getLog(SampleByTest.class);
     private static final String SYS_TELEMETRY_WAL_DDL = "CREATE TABLE IF NOT EXISTS 'sys.telemetry_wal' ( " +
             "created TIMESTAMP, " +
@@ -343,13 +344,15 @@ public class SampleByTest extends AbstractCairoTest {
         assertMemoryLeak(() -> {
             execute("create table telem (created timestamp, event_type int, table_id int, latency double) timestamp(created) partition by DAY");
 
-            execute("insert into telem " +
-                    "select " +
-                    "generate_series as created," +
-                    " rnd_int() % 4," +
-                    " rnd_int() % 100," +
-                    " abs(rnd_double())" +
-                    " from generate_series('2025-01-20T13:57:14.000000Z', dateadd('u', 2500 * 100_000, '2025-01-20T13:57:14.000000Z') - 1, 2500)");
+            execute(
+                    "insert into telem " +
+                            "select " +
+                            "generate_series as created," +
+                            " rnd_int() % 4," +
+                            " rnd_int() % 100," +
+                            " abs(rnd_double())" +
+                            " from generate_series('2025-01-20T13:57:14.000000Z', dateadd('u', 2500 * 100_000, '2025-01-20T13:57:14.000000Z') - 1, 2500)"
+            );
 
             assertQueryNoLeakCheck(
                     """
@@ -519,7 +522,7 @@ public class SampleByTest extends AbstractCairoTest {
     }
 
     @Test
-    public void testGeohashFillNull() throws Exception {
+    public void testGeoHashFillNull() throws Exception {
         assertQuery(
                 """
                         s\tk\tfirst\tfirst1\tfirst2\tfirst3
@@ -554,7 +557,7 @@ public class SampleByTest extends AbstractCairoTest {
     }
 
     @Test
-    public void testGeohashFillPrev() throws Exception {
+    public void testGeoHashFillPrev() throws Exception {
         assertQuery(
                 """
                         s\tk\tfirst\tfirst1\tfirst2\tfirst3
@@ -589,7 +592,7 @@ public class SampleByTest extends AbstractCairoTest {
     }
 
     @Test
-    public void testGeohashInterpolated() throws Exception {
+    public void testGeoHashInterpolated() throws Exception {
         assertException(
                 "select k, first(b) from x sample by 3h fill(linear)",
                 "create table x as " +
@@ -722,26 +725,25 @@ public class SampleByTest extends AbstractCairoTest {
                             last(Confirmed) Confirmed,\s
                             last(Recovered) Recovered,\s
                             last(Deaths) Deaths\s
+                        from (
+                            select\s
+                                LastUpdate,\s
+                                CountryRegion,\s
+                                sum(Confirmed) Confirmed,\s
+                                sum(Recovered) Recovered,\s
+                                sum(Deaths) Deaths
                             from (
                                 select\s
                                     LastUpdate,\s
+                                    ProvinceState,\s
                                     CountryRegion,\s
-                                    sum(Confirmed) Confirmed,\s
-                                    sum(Recovered) Recovered,\s
-                                    sum(Deaths) Deaths
-                                from (
-                                    select\s
-                                        LastUpdate,\s
-                                        ProvinceState,\s
-                                        CountryRegion,\s
-                                        last(Confirmed) Confirmed,\s
-                                        last(Recovered) Recovered,\s
-                                        last(Deaths) Deaths
-                                    from (covid where CountryRegion in ('China', 'Mainland China'))
-                                    sample by 1d fill(prev)
-                                ) timestamp(ProvinceState)
-                            ) sample by 1M
-                        ;
+                                    last(Confirmed) Confirmed,\s
+                                    last(Recovered) Recovered,\s
+                                    last(Deaths) Deaths
+                                from (covid where CountryRegion in ('China', 'Mainland China'))
+                                sample by 1d fill(prev)
+                            ) timestamp(ProvinceState)
+                        ) sample by 1M;
                         """,
                 "create table covid as " +
                         "(" +
@@ -755,7 +757,7 @@ public class SampleByTest extends AbstractCairoTest {
                         " from" +
                         " long_sequence(1000)" +
                         ") timestamp(LastUpdate) partition by NONE",
-                707,
+                635,
                 "not a TIMESTAMP"
         );
     }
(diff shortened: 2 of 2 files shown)


puzpuzpuz:
@coderabbitai review

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -5202,63 +5250,47 @@ private QueryModel rewriteSampleBy(@Nullable QueryModel model, @Transient SqlExe
...
+                        n--;
+
+                        if (checkForChildAggregates(qc.getAst())) {
+                            // Ok, it's an expression with mixed timestamp references and aggregate functions.
+                            // Replace aggregate functions with aliases and add the functions to the group by model.
+                            // This way an expression like `select timestamp, count() / timestamp::long`
+                            // Is turned into `select timestamp, count / timestamp::long from (select count(), ...`
+                            needRemoveColumns += rewriteTimestampMixedWithAggregates(qc.getAst(), model);
+                        }
+
```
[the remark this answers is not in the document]
puzpuzpuz: Nested model holds table name expression and sample by nodes, nothing more.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -4479,8 +4479,18 @@ private RecordCursorFactory generateSelectChoose(QueryModel model, SqlExecutionC
...
+        CharSequence timestampName = null;
+        int timestampNameDot = -1;
+        if (timestampIndex != -1) {
+            timestampName = factory.getMetadata().getColumnName(timestampIndex);
+            timestampNameDot = Chars.indexOfLastUnquoted(timestampName, '.');
+        }
+        final CharSequence firstOrderByColumn = model.getOrderBy().size() > 0 && model.getOrderBy().getQuick(0).type == LITERAL
+                ? model.getOrderBy().getQuick(0).token
+                : null;
+
```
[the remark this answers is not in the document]
puzpuzpuz: The outer `if (index == timestampIndex)` branch guarantees that `timestampName` it not null.

ideoma:
nit: you can remove the ignore tag from `SampleByTest.testSampleByWithProjection2()` and modify the expected results, it's fixed now

puzpuzpuz:
> nit: you can remove the ignore tag from `SampleByTest.testSampleByWithProjection2()` and modify the expected results, it's fixed now

Thanks for the heads-up. I didn't check the disabled tests. Enabled in 6dc3685b2bcdeaefa9c198689df8722023d9829c

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -4504,14 +4514,23 @@ private RecordCursorFactory generateSelectChoose(QueryModel model, SqlExecutionC
...
+                // timestamp in case of multiple timestamp aliases, e.g. `select ts, ts as ts1, ...`.
+                // That's to choose the optimal plan in generateOrderBy().
+                // Otherwise, prefer columns with aliases matching the base column name, e.g.
+                // prefer `t1.ts as ts` over `t1.ts as ts2`.
+                if (Chars.equalsIgnoreCaseNc(queryColumn.getAlias(), firstOrderByColumn)
+                        || Chars.equalsIgnoreCase(queryColumn.getAlias(), timestampName, timestampNameDot + 1, timestampName.length())
+                        || !timestampSet) {
+                    queryMetadata.setTimestampIndex(i);
+                    timestampSet = true;
+                }
```
ideoma, line 4543: What timestamp will be set to if the execution does not go into the if block? What if `firstOrderByColumn == null`, why do we alter the old behaviour in that case?

puzpuzpuz: > What timestamp will be set to if the execution does not go into the if block?

Same as before this change, `-1`, as that's the default in `GenericRecordMetadata`.

> What if `firstOrderByColumn == null`, why do we alter the old behaviour in that case?

If `firstOrderByColumn` is null, only the `Chars.equalsIgnoreCase(queryColumn.getAlias(), timestampName, timestampNameDot + 1, timestampName.length()) || !timestampSet` conditions will be checked. There is a clarifying comment around this branch - do you find it unclear?

ideoma: No, not very clear, better to be separate if tbh.

What if this condition `Chars.equalsIgnoreCase(queryColumn.getAlias(), timestampName, timestampNameDot + 1, timestampName.length())`  is also false, but the old condition `index == timestampIndex` is true, why do we change the behaviour in this situation? Or is it impossible?

puzpuzpuz:
> Still reviewing, just wanted to ask to boost the overall test coverage, please.

Covered one branch and removed two never-taken in 7372ad2f9189e52cd19d182bfd7d1511dedb46e9. Everything else seems covered.

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -1537,66 +1537,67 @@ private boolean createSelectColumn(
...
-        final LowerCaseCharSequenceObjHashMap<CharSequence> translatingAliasMap = translatingModel.getColumnNameToAliasMap();
-        final int index = translatingAliasMap.keyIndex(columnAst.token);
-        if (index < 0) {
-            // check if the column is a duplicate, i.e. already referenced by the group-by model
-            if (!allowDuplicates && groupByModel.getAliasToColumnMap().contains(columnName)) {
-                throw SqlException.duplicateColumn(columnAst.position, columnName);
+        final CharSequence translatedColumnName = translatingModel.getColumnNameToAliasMap().get(columnAst.token);
+        if (translatedColumnName != null) {
+            // the column is already being referenced by the translating model
+            final CharSequence groupByColumnName = groupByModel.getColumnNameToAliasMap().get(translatedColumnName);
```
ideoma, line 1556: if `isGroupBy == false`, we shouldn't search `groupByModel`, I don't see why it should be even passed as non-null.

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -1537,66 +1537,67 @@ private boolean createSelectColumn(
...
                     baseModel
             );
 
-            final QueryColumn translatedColumn = nextColumn(alias, columnAst.position);
-
             // create column that references inner alias we just created
+            final QueryColumn translatedColumn = nextColumn(innerAlias, columnAst.position);
             innerVirtualModel.addBottomUpColumn(translatedColumn);
-            windowModel.addBottomUpColumn(translatedColumn);
             groupByModel.addBottomUpColumn(translatedColumn);
```
ideoma, line 1596: If `isGroupBy == false`, why do we use `groupByModel`?

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
puzpuzpuz: > What if this condition `Chars.equalsIgnoreCase(queryColumn.getAlias(), timestampName, timestampNameDot + 1, timestampName.length())` is also false, but the old condition `index == timestampIndex` is true, why do we change the behaviour in this situation? Or is it impossible?

That's possible. Before this change we picked up the last matching selected column, now we pick up the first one. Both approaches are equivalent and the scenario is rare enough. Do you think it's worth keeping the old behavior by making the code slightly more complicated?

ideoma (approved):

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
puzpuzpuz: We're using the model to accumulate the columns, but later on it's ignored. I kept this logic as is to avoid breaking the code that may rely on this.

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
puzpuzpuz: We're using the model to accumulate the columns, but later on it's ignored. I kept this logic as is to avoid breaking the code that may rely on this.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
ideoma: Could you illustrate it with an example of a query that will have a breaking change?

puzpuzpuz: Sure. I've added "breaking change" to the PR title and added an illustration to the description.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 119 / 121 (98.35%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|48|49|97.96%|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|71|72|98.61%|

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -2987,6 +2992,10 @@ private void moveWhereInsideSubQueries(QueryModel model) throws SqlException {
         }
     }
 
+    private QueryColumn nextColumn(CharSequence alias, CharSequence column, int position) {
```
bluestreak01: unused

core/src/test/java/io/questdb/test/griffin/engine/groupby/SampleByTest.java
```diff
@@ -7354,34 +7565,32 @@ public void testSampleByWithProjection2() throws Exception {
...
+            final String query = "select " +
+                    "symbol || 'abcd' as symbol" +
+                    ", sum(amount)" +
+                    ", vwap(price, amount)" +
+                    ", cast(dateadd('h', 2, to_timezone(timestamp,'EST')) as double) as ts" +
+                    ", cast(dateadd('h', 2, to_timezone(timestamp,'EST')) as double) + sum(amount) NYTime " +
+                    "from trades\n" +
+                    "sample by 5m";
+            assertQueryNoLeakCheck(
+                    "symbol\tsum\tvwap\tts\tNYTime\n" +
```
bluestreak01: please use text blocks like everywhere else in the class - consistency + warnings

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 119 / 121 (98.35%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|48|49|97.96%|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|71|72|98.61%|

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
puzpuzpuz: Removed in caee3245ee686439ce92e89ed8e158793261fe02

core/src/test/java/io/questdb/test/griffin/engine/groupby/SampleByTest.java
puzpuzpuz: Reformatted in caee3245ee686439ce92e89ed8e158793261fe02

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 120 / 121 (99.17%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|48|49|97.96%|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|72|72|100.00%|

PR #6283, opened 2025-10-17
perf(sql): optimized Markout Horizon CROSS JOIN

mtopolnik:
This is the Markout Curve query we're optimizing:

```sql
WITH
  offsets AS (
    SELECT sec_offs, 1_000_000 * sec_offs usec_offs 
    FROM (SELECT x-601 AS sec_offs FROM long_sequence(1201))
  ),
  points AS (SELECT * FROM (
    SELECT id, order_ts, sym1, sym2, unit_price, volume, sec_offs, order_ts + usec_offs AS ts
    FROM orders CROSS JOIN offsets
    ORDER BY order_ts + usec_offs
  ) TIMESTAMP(ts)),
  join1 AS (
    SELECT t.*, p.price AS sym1_price
    FROM points as t
    ASOF JOIN prices as p
    ON (t.sym1 = p.sym)
  ),
  join2 AS (
    SELECT t.*, p.price AS sym2_price
    FROM join1 as t
    ASOF JOIN prices as p
    ON (t.sym2 = p.sym)
  ),
  markouts AS (
  SELECT
    sec_offs, 
    volume,
    volume * (unit_price - (sym1_price/sym2_price)) AS weighted_markout
  FROM join2
  )
SELECT sec_offs, sum(weighted_markout) / sum(volume) AS avg_weigthed_markout 
FROM markouts;
```

This subquery generates the sampling points over the markout horizon of every order:

```sql
SELECT id, order_ts, sym1, sym2, unit_price, volume, sec_offs, order_ts + usec_offs AS ts
FROM orders CROSS JOIN offsets
ORDER BY order_ts + usec_offs
```

It performs poorly on a larger number of orders, because the CROSS JOIN output must be fully materialized and then sorted.

This PR introduces a special-case cursor factory that emits the CROSS JOIN output directly in the required order, without having to materialize the `orders` table. It does materialize `offsets` because it needs random access to it, but this table is of a very limited size, coming from `long_sequence()`.

Example usage and benchmark:

```sql
CREATE TABLE orders (
      ts TIMESTAMP,
      price DOUBLE
INSERT INTO orders
  SELECT
      generate_series as timestamp,
      rnd_double() * 10.0 + 5.0
      FROM generate_series('2025-01-02', '2025-01-02T00:10', '200u');
```

This creates 3,000,001 orders spaced at 200 µs.

```sql
WITH
  offsets AS (
    SELECT sec_offs, 10_000_000 * sec_offs usec_offs 
    FROM (SELECT x-61 AS sec_offs FROM long_sequence(121))
  )
  SELECT /*+ markout_horizon(orders offsets) */ sum(price) FROM (SELECT * FROM (
    SELECT price, ts + usec_offs AS timestamp
    FROM orders CROSS JOIN offsets
    ORDER BY ts + usec_offs
  ) TIMESTAMP(timestamp));
```

This creates the markout horizon sampling grid over 10 minutes, spaced at 10 seconds. There are 121 sampling points for each order. Therefore, this results in 121 * 3,000,001 = 363,000,121 rows. 

The query is built to emphasize the worst case for the Markout Horizon algo in terms of RAM usage: tight spacing of orders vs. the markout horizon. The algorithm must hold 3 million iterator structures in RAM at once. It uses 40 bytes per iterator.

I benchmarked it on a `r7a.4xlarge` EC2 box.

Without the markout hint, the query took 135 seconds, and RAM usage went from 2.3 GB baseline to 10.7 GB.

With the hint, the query took 17 seconds, with an even split between aggregation and row generation. RAM usage went from 2.3 to 2.4 GB.

### DIFF
--- a/core/src/test/java/io/questdb/test/griffin/IPv4Test.java
+++ b/core/src/test/java/io/questdb/test/griffin/IPv4Test.java
@@ -36,12 +36,14 @@ public class IPv4Test extends AbstractCairoTest {
     @Test
     public void testAggregateByIPv4() throws Exception {
         assertQuery(
-                "ip\tsum\n" +
-                        "0.0.0.1\t11644\n" +
-                        "0.0.0.2\t7360\n" +
-                        "0.0.0.3\t9230\n" +
-                        "0.0.0.4\t10105\n" +
-                        "0.0.0.5\t11739\n",
+                """
+                        ip\tsum
+                        0.0.0.1\t11644
+                        0.0.0.2\t7360
+                        0.0.0.3\t9230
+                        0.0.0.4\t10105
+                        0.0.0.5\t11739
+                        """,
                 "select ip, sum(bytes) from test order by ip",
                 "create table test as " +
                         "(" +
@@ -65,8 +67,10 @@ public class IPv4Test extends AbstractCairoTest {
             execute("alter table test add col2 ipv4");
 
             assertSql(
-                    "col1\tcol2\n" +
-                            "0.0.0.1\t\n",
+                    """
+                            col1\tcol2
+                            0.0.0.1\t
+                            """,
                     "test"
             );
         });
@@ -89,8 +93,10 @@ public class IPv4Test extends AbstractCairoTest {
     @Test
     public void testBitAndStr() throws Exception {
         assertMemoryLeak(() -> assertSql(
-                "column\n" +
-                        "2.0.0.0\n",
+                """
+                        column
+                        2.0.0.0
+                        """,
                 "select ipv4 '2.1.1.1' & '2.2.2.2'"
         ));
     }
@@ -98,8 +104,10 @@ public class IPv4Test extends AbstractCairoTest {
     @Test
     public void testBitAndStr2() throws Exception {
         assertMemoryLeak(() -> assertSql(
-                "column\n" +
-                        "2.0.0.0\n",
+                """
+                        column
+                        2.0.0.0
+                        """,
                 "select '2.2.2.2' & ipv4 '2.1.1.1'"
         ));
     }
@@ -107,8 +115,10 @@ public class IPv4Test extends AbstractCairoTest {
     @Test
     public void testBroadcastAddrUseCase() throws Exception {
         assertMemoryLeak(() -> assertSql(
-                "column\n" +
-                        "68.255.255.255\n",
+                """
+                        column
+                        68.255.255.255
+                        """,
                 "select (~ netmask('68.11.9.2/8')) | ipv4 '68.11.9.2'"
         ));
     }
@@ -116,57 +126,59 @@ public class IPv4Test extends AbstractCairoTest {
     @Test
     public void testCaseIPv41() throws Exception {
         assertQuery(
-                "ip\tbytes\tcase\n" +
-                        "187.139.150.80\t580\tnay\n" +
-                        "212.159.205.29\t23\tnay\n" +
-                        "79.15.250.138\t850\tnay\n" +
-                        "205.123.179.216\t167\tnay\n" +
-                        "170.90.236.206\t572\tnay\n" +
-                        "92.26.178.136\t7\tnay\n" +
-                        "231.146.30.59\t766\tnay\n" +
-                        "113.132.124.243\t522\tnay\n" +
-                        "67.22.249.199\t203\tnay\n" +
-                        "25.107.51.160\t827\tYAY\n" +
-                        "146.16.210.119\t383\tnay\n" +
-                        "187.63.210.97\t424\tnay\n" +
-                        "188.239.72.25\t513\tnay\n" +
-                        "181.82.42.148\t539\tnay\n" +
-                        "129.172.181.73\t25\tnay\n" +
-                        "66.56.51.126\t904\tnay\n" +
-                        "230.202.108.161\t171\tnay\n" +
-                        "180.48.50.141\t136\tnay\n" +
-                        "128.225.84.244\t313\tnay\n" +
-                        "254.93.251.9\t810\tnay\n" +
-                        "227.40.250.157\t903\tnay\n" +
-                        "180.36.62.54\t528\tnay\n" +
-                        "136.166.51.222\t580\tnay\n" +
-                        "24.123.12.210\t95\tYAY\n" +
-                        "171.117.213.66\t720\tnay\n" +
-                        "224.99.254.121\t619\tnay\n" +
-                        "55.211.206.129\t785\tYAY\n" +
-                        "144.131.72.77\t369\tnay\n" +
-                        "97.159.145.120\t352\tnay\n" +
-                        "164.153.242.17\t906\tnay\n" +
-                        "165.166.233.251\t332\tnay\n" +
-                        "114.126.117.26\t71\tnay\n" +
-                        "164.74.203.45\t678\tnay\n" +
-                        "241.248.184.75\t334\tnay\n" +
-                        "255.95.177.227\t44\tnay\n" +
-                        "216.150.248.30\t563\tnay\n" +
-                        "71.73.196.29\t741\tnay\n" +
-                        "180.91.244.55\t906\tnay\n" +
-                        "111.221.228.130\t531\tnay\n" +
-                        "171.30.189.77\t238\tnay\n" +
-                        "73.153.126.70\t772\tnay\n" +
-                        "105.218.160.179\t986\tnay\n" +
-                        "201.100.238.229\t318\tnay\n" +
-                        "12.214.12.100\t598\tYAY\n" +
-                        "212.102.182.127\t984\tnay\n" +
-                        "50.214.139.184\t574\tYAY\n" +
-                        "186.33.243.40\t659\tnay\n" +
-                        "74.196.176.71\t740\tnay\n" +
-                        "150.153.88.133\t849\tnay\n" +
-                        "63.60.82.184\t37\tYAY\n",
+                """
+                        ip\tbytes\tcase
+                        187.139.150.80\t580\tnay
+                        212.159.205.29\t23\tnay
+                        79.15.250.138\t850\tnay
+                        205.123.179.216\t167\tnay
+                        170.90.236.206\t572\tnay
+                        92.26.178.136\t7\tnay
+                        231.146.30.59\t766\tnay
+                        113.132.124.243\t522\tnay
+                        67.22.249.199\t203\tnay
+                        25.107.51.160\t827\tYAY
+                        146.16.210.119\t383\tnay
+                        187.63.210.97\t424\tnay
+                        188.239.72.25\t513\tnay
+                        181.82.42.148\t539\tnay
+                        129.172.181.73\t25\tnay
+                        66.56.51.126\t904\tnay
+                        230.202.108.161\t171\tnay
+                        180.48.50.141\t136\tnay
+                        128.225.84.244\t313\tnay
+                        254.93.251.9\t810\tnay
+                        227.40.250.157\t903\tnay
+                        180.36.62.54\t528\tnay
+                        136.166.51.222\t580\tnay
+                        24.123.12.210\t95\tYAY
+                        171.117.213.66\t720\tnay
+                        224.99.254.121\t619\tnay
+                        55.211.206.129\t785\tYAY
+                        144.131.72.77\t369\tnay
+                        97.159.145.120\t352\tnay
+                        164.153.242.17\t906\tnay
+                        165.166.233.251\t332\tnay
+                        114.126.117.26\t71\tnay
+                        164.74.203.45\t678\tnay
+                        241.248.184.75\t334\tnay
+                        255.95.177.227\t44\tnay
+                        216.150.248.30\t563\tnay
+                        71.73.196.29\t741\tnay
+                        180.91.244.55\t906\tnay
+                        111.221.228.130\t531\tnay
+                        171.30.189.77\t238\tnay
+                        73.153.126.70\t772\tnay
+                        105.218.160.179\t986\tnay
+                        201.100.238.229\t318\tnay
+                        12.214.12.100\t598\tYAY
+                        212.102.182.127\t984\tnay
+                        50.214.139.184\t574\tYAY
+                        186.33.243.40\t659\tnay
+                        74.196.176.71\t740\tnay
+                        150.153.88.133\t849\tnay
+                        63.60.82.184\t37\tYAY
+                        """,
                 "select ip, bytes, case when ip <<= '2.65.32.1/2' then 'YAY' else 'nay' end from test",
                 "create table test as " +
                         "(" +
@@ -185,57 +197,59 @@ public class IPv4Test extends AbstractCairoTest {
     @Test
     public void testCaseIPv42() throws Exception {
         assertQuery(
-                "ip\tbytes\tcase\n" +
-                        "187.139.150.80\t580\t\n" +
-                        "212.159.205.29\t23\t\n" +
-                        "79.15.250.138\t850\t\n" +
-                        "205.123.179.216\t167\t\n" +
-                        "170.90.236.206\t572\t\n" +
-                        "92.26.178.136\t7\t\n" +
-                        "231.146.30.59\t766\t\n" +
-                        "113.132.124.243\t522\t\n" +
-                        "67.22.249.199\t203\t\n" +
-                        "25.107.51.160\t827\tYAY\n" +
-                        "146.16.210.119\t383\t\n" +
-                        "187.63.210.97\t424\t\n" +
-                        "188.239.72.25\t513\t\n" +
-                        "181.82.42.148\t539\t\n" +
-                        "129.172.181.73\t25\t\n" +
-                        "66.56.51.126\t904\t\n" +
-                        "230.202.108.161\t171\t\n" +
-                        "180.48.50.141\t136\t\n" +
-                        "128.225.84.244\t313\t\n" +
-                        "254.93.251.9\t810\t\n" +
-                        "227.40.250.157\t903\t\n" +
-                        "180.36.62.54\t528\t\n" +
-                        "136.166.51.222\t580\t\n" +
-                        "24.123.12.210\t95\tYAY\n" +
-                        "171.117.213.66\t720\t\n" +
-                        "224.99.254.121\t619\t\n" +
-                        "55.211.206.129\t785\tYAY\n" +
-                        "144.131.72.77\t369\t\n" +
-                        "97.159.145.120\t352\t\n" +
-                        "164.153.242.17\t906\t\n" +
-                        "165.166.233.251\t332\t\n" +
-                        "114.126.117.26\t71\t\n" +
-                        "164.74.203.45\t678\t\n" +
-                        "241.248.184.75\t334\t\n" +
-                        "255.95.177.227\t44\t\n" +
-                        "216.150.248.30\t563\t\n" +
-                        "71.73.196.29\t741\t\n" +
-                        "180.91.244.55\t906\t\n" +
-                        "111.221.228.130\t531\t\n" +
-                        "171.30.189.77\t238\t\n" +
-                        "73.153.126.70\t772\t\n" +
-                        "105.218.160.179\t986\t\n" +
-                        "201.100.238.229\t318\t\n" +
-                        "12.214.12.100\t598\tYAY\n" +
-                        "212.102.182.127\t984\t\n" +
-                        "50.214.139.184\t574\tYAY\n" +
-                        "186.33.243.40\t659\t\n" +
-                        "74.196.176.71\t740\t\n" +
-                        "150.153.88.133\t849\t\n" +
-                        "63.60.82.184\t37\tYAY\n",
+                """
+                        ip\tbytes\tcase
+                        187.139.150.80\t580\t
+                        212.159.205.29\t23\t
+                        79.15.250.138\t850\t
+                        205.123.179.216\t167\t
+                        170.90.236.206\t572\t
+                        92.26.178.136\t7\t
+                        231.146.30.59\t766\t
+                        113.132.124.243\t522\t
+                        67.22.249.199\t203\t
+                        25.107.51.160\t827\tYAY
+                        146.16.210.119\t383\t
+                        187.63.210.97\t424\t
+                        188.239.72.25\t513\t
+                        181.82.42.148\t539\t
+                        129.172.181.73\t25\t
+                        66.56.51.126\t904\t
+                        230.202.108.161\t171\t
+                        180.48.50.141\t136\t
+                        128.225.84.244\t313\t
+                        254.93.251.9\t810\t
+                        227.40.250.157\t903\t
+                        180.36.62.54\t528\t
+                        136.166.51.222\t580\t
+                        24.123.12.210\t95\tYAY
+                        171.117.213.66\t720\t
+                        224.99.254.121\t619\t
+                        55.211.206.129\t785\tYAY
+                        144.131.72.77\t369\t
+                        97.159.145.120\t352\t
+                        164.153.242.17\t906\t
+                        165.166.233.251\t332\t
+                        114.126.117.26\t71\t
+                        164.74.203.45\t678\t
+                        241.248.184.75\t334\t
+                        255.95.177.227\t44\t
+                        216.150.248.30\t563\t
+                        71.73.196.29\t741\t
+                        180.91.244.55\t906\t
+                        111.221.228.130\t531\t
+                        171.30.189.77\t238\t
+                        73.153.126.70\t772\t
+                        105.218.160.179\t986\t
+                        201.100.238.229\t318\t
+                        12.214.12.100\t598\tYAY
+                        212.102.182.127\t984\t
+                        50.214.139.184\t574\tYAY
+                        186.33.243.40\t659\t
+                        74.196.176.71\t740\t
+                        150.153.88.133\t849\t
+                        63.60.82.184\t37\tYAY
+                        """,
                 "select ip, bytes, case when ip <<= '2.65.32.1/2' then 'YAY' end from test",
                 "create table test as " +
                         "(" +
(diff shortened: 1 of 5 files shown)


glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 281 / 299 (93.98%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|87|94|92.55%|
|:large_blue_circle:|io/questdb/griffin/engine/join/TimestampLadderRecordCursorFactory.java|179|190|94.21%|
|:large_blue_circle:|io/questdb/cairo/RecordArray.java|14|14|100.00%|

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 281 / 299 (93.98%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|87|94|92.55%|
|:large_blue_circle:|io/questdb/griffin/engine/join/TimestampLadderRecordCursorFactory.java|179|190|94.21%|
|:large_blue_circle:|io/questdb/cairo/RecordArray.java|14|14|100.00%|

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 305 / 329 (92.71%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|66|72|91.67%|
|:large_blue_circle:|io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java|237|255|92.94%|
|:large_blue_circle:|io/questdb/cairo/RecordArray.java|1|1|100.00%|

puzpuzpuz:
@coderabbitai review

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -1594,6 +1598,128 @@ private RecordCursorFactory createSpliceJoin(
...
+        if (designatedTimestampIndex == -1) {
+            return null;
+        }
+
+        // Ensure we have an ORDER BY clause.
+        // lastSeenOrderByModel was captured as we were descending through nested models
+        if (lastSeenOrderByModel == null) {
+            return null;
+        }
+        ExpressionNode orderByNode = lastSeenOrderByModel.getOrderBy().getQuick(0);
```
puzpuzpuz: Don't we need to check the number of ORDER BY columns?
```sql
WITH offsets AS (
    SELECT 1_000_000 * (x-1) usec_offs
    FROM long_sequence(3)
)
SELECT /*+ markout_horizon(orders offsets) */ id, order_ts + usec_offs AS ts
FROM orders CROSS JOIN offsets
ORDER BY order_ts + usec_offs, id desc;
```
seems to pass the checks while it shouldn't.

mtopolnik:
@coderabbitai review

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -4037,26 +4194,27 @@ private RecordCursorFactory generateQuery(QueryModel model, SqlExecutionContext
...
-                        model,
-                        executionContext
-                ),
-                model,
-                executionContext
-        );
+        // Remember the last model with non-empty ORDER BY as we descend through nested models.
+        // We need the ORDER BY clause in the Markout Horizon Join optimization, but it's stored
+        // several levels up from the model that holds the join clause.
+        final QueryModel savedOrderByModel = lastSeenOrderByModel;
```
puzpuzpuz, line 4211: Using an outer model is not very reliable since there may be intermediate models that change the meaning of column aliases:
```sql
WITH offsets AS (
    SELECT 1_000_000 * (x-1) usec_offs
    FROM long_sequence(3)
)
select master_id+slave_id ts, ts id
from (
  SELECT /*+ markout_horizon(orders offsets) */ master_id, slave_id, order_ts + usec_offs AS ts
  FROM orders CROSS JOIN offsets
)
ORDER BY ts;
```

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -7160,6 +7315,19 @@ interface ModelOperator {
         void operate(ObjectPool<ExpressionNode> pool, QueryModel model);
     }
 
+    /**
+     * Container class to hold the detected parameters of a markout horizon pattern.
+     */
+    private static class MarkoutHorizonInfo {
```
puzpuzpuz, line 7332: nit: we could reuse this object

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -1594,6 +1598,128 @@ private RecordCursorFactory createSpliceJoin(
...
+        if (slaveColumnNode == null || slaveColumnNode.type != ExpressionNode.LITERAL) {
+            return null;
+        }
+        int slaveColumnIndex = slaveMetadata.getColumnIndexQuiet(slaveColumnNode.token);
+        if (slaveColumnIndex == -1) {
+            return null; // Slave column not found
+        }
+        // Return the info needed to create the optimized factory.
+        // resultTimestampColumnIndex will be -1 if no SELECT column matches the ORDER BY expression.
+        return new MarkoutHorizonInfo(timestampColumnIndex, slaveColumnIndex);
```
puzpuzpuz: Don't we need to check the type of the column with the `slaveColumnIndex` index? `MarkoutHorizonRecordCursor` calls `getLong()` without doing any extra checks.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -1594,6 +1598,128 @@ private RecordCursorFactory createSpliceJoin(
...
+     * @param masterAlias         alias for the master LHS of the join
+     * @param masterModel         QueryModel for the LHS of the join
+     * @param masterMetadata      Metadata for the LHS of the join
+     * @param masterCursorFactory the LHS cursor factory
+     * @param slaveModel          QueryModel for the RHS of the join
+     * @param slaveMetadata       Metadata for the RHS of the join (should be an arithmetic sequence)
+     * @param slaveCursorFactory  the RHS cursor factory (should produce an arithmetic sequence)
+     * @return MarkoutHorizonInfo if pattern is detected, null otherwise
+     */
+    private MarkoutHorizonInfo detectMarkoutHorizonPattern(
```
puzpuzpuz, line 1634: This method seems to have many untested branches. Could you cover them in `MarkoutHorizonCrossJoinTest`?

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,631 @@
...
+
+        private static void block_setNextBlockAddr(long blockAddr, long nextAddr) {
+            Unsafe.getUnsafe().putLong(blockAddr + BLOCK_OFFSET_NEXT_BLOCK_ADDR, nextAddr);
+        }
+
+        private static void block_setNextFreeSlot(long blockAddr, int slot) {
+            Unsafe.getUnsafe().putInt(blockAddr + BLOCK_OFFSET_NEXT_FREE_SLOT, slot);
+        }
+
+        private static int block_usedSlotCount(long blockAddr) {
```
puzpuzpuz: nit: unused method here.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,631 @@
...
+
+        private long createIterator(Record masterRecord) {
+            // Get or create the last iterator block
+            if (lastIteratorBlockAddr == 0) {
+                // No iterator blocks so far. Allocate the first one.
+                lastIteratorBlockAddr = block_alloc();
+                firstIteratorBlockAddr = lastIteratorBlockAddr;
+            }
+            int nextFreeSlot = block_nextFreeSlot(lastIteratorBlockAddr);
+            if (nextFreeSlot == ITERATORS_PER_BLOCK) {
```
puzpuzpuz, line 475: This branch is never taken in `MarkoutHorizonCrossJoinTest`.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,631 @@
...
+
+        private int currentRowNum(long iterAddr) {
+            return iter_nextSlaveRowNum(iterAddr) - 1;
+        }
+
+        private void discardIterator(long iterAddr) {
+            long blockAddr = iterAddr - iter_offsetFromBlockStart(iterAddr);
+            int newCount = block_decUsedSlotCount(blockAddr);
+            if (newCount == 0)
+                if (blockAddr != lastIteratorBlockAddr) {
```
puzpuzpuz, line 513: This branch is never taken in `MarkoutHorizonCrossJoinTest`.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,631 @@
...
+            long nextMasterTimestamp = masterRecord.getTimestamp(masterTimestampColumnIndex);
+            return nextMasterTimestamp + firstSlaveTimeOffset;
+        }
+
+        private boolean isEmpty(long iterAddr) {
+            return iter_nextSlaveRowNum(iterAddr) == slaveRowCount;
+        }
+
+        private long peekNextTimestamp(long iterAddr) {
+            if (iter_nextSlaveRowNum(iterAddr) == slaveRowCount) {
```
puzpuzpuz: One more never taken branch.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
mtopolnik: Done in 71c3497c37cc902176185d18067612679e6aa753.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,631 @@
...
+        Misc.free(slaveFactory);
+    }
+
+    /**
+     * Record cursor that emits cross-join results in timestamp order.
+     * <p>
+     * We materialize the slave cursor (arithmetic sequence) into a RecordArray
+     * because we need random access to its rows.
+     */
+    private static class MarkoutHorizonRecordCursor extends AbstractJoinCursor {
```
puzpuzpuz, line 203: The cursor should implement `calculateSize()`.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,631 @@
...
+            while (slaveCursor.hasNext()) {
+                circuitBreaker.statefulThrowExceptionIfTripped();
+                long offset = slaveRecordArray.put(slaveRecord);
+                slaveRecordOffsets.add(offset);
+            }
+            slaveRecordArray.toTop();
+            slaveRowCount = slaveRecordOffsets.size();
+
+            // Cache the first slave's offset value to avoid repositioning recordC later
+            if (slaveRowCount > 0) {
```
puzpuzpuz, line 631: We should validate that `slaveRowCount` fits into an `int` since `ITERATOR_OFFSET_NEXT_SLAVE_ROW_NUM` field size is also an `int`.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,631 @@
...
+     * Record cursor that emits cross-join results in timestamp order.
+     * <p>
+     * We materialize the slave cursor (arithmetic sequence) into a RecordArray
+     * because we need random access to its rows.
+     */
+    private static class MarkoutHorizonRecordCursor extends AbstractJoinCursor {
+        // Native memory layout constants
+        private static final int BLOCK_HEADER_SIZE = 16;
+        private static final int BLOCK_OFFSET_NEXT_BLOCK_ADDR = 0;    // long (8 bytes)
+        private static final int BLOCK_OFFSET_NEXT_FREE_SLOT = 8;     // int (4 bytes)
```
puzpuzpuz, line 207: nit: can we optimize memory usage here? For instance, `BLOCK_OFFSET_USED_SLOT_COUNT` and `BLOCK_OFFSET_USED_SLOT_COUNT` fields could be `short`s. One more simple step is moving all blocks to a heap (grow-only off-heap buffer) and align them to int/long. This way we can use compressed pointers in `ITERATOR_OFFSET_NEXT_ITER_ADDR` (`int` instead of `long`).

Optionally, we could go one step further and check the number of right-hand rows somewhere in `detectMarkoutHorizonPattern()` and validate that it fits `short`. This way, `ITERATOR_OFFSET_OFFSET_FROM_BLOCK_START` could be change to index instead of offset (`short` instead of `int`) and `ITERATOR_OFFSET_NEXT_SLAVE_ROW_NUM` can also be `short`.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,631 @@
...
+        }
+
+        private void discardIterator(long iterAddr) {
+            long blockAddr = iterAddr - iter_offsetFromBlockStart(iterAddr);
+            int newCount = block_decUsedSlotCount(blockAddr);
+            if (newCount == 0)
+                if (blockAddr != lastIteratorBlockAddr) {
+                    long nextBlockAddr = block_nextBlockAddr(blockAddr);
+                    block_free(blockAddr);
+                    firstIteratorBlockAddr = nextBlockAddr;
```
puzpuzpuz, line 516: Why is it guaranteed that the freed block was the first one?

mtopolnik: This invariant isn't used just here, it's also used in the entire way iterator slots are allocated within the block -- we use a simple bump counter.

This works because iterators get created and discarded in a strict FIFO order, and that's true because every iterator emits the same number of rows, with the same timestamp difference.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
mtopolnik: I'd propose postponing this level of optimization to a later design iteration, once the basic idea has been validated by users.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,631 @@
...
+            isMasterHasNextPending = true;
+            masterHasNext = false;
+            currMasterRowId = -1;
+            masterCursor.toTop();
+            slaveRecordArray.toTop();
+        }
+
+        private static long block_alloc() {
+            long blockAddr = Unsafe.malloc(BLOCK_SIZE, MemoryTag.NATIVE_DEFAULT);
+            Unsafe.getUnsafe().setMemory(blockAddr, BLOCK_SIZE, (byte) 0);
```
puzpuzpuz: nit: we usually use `Vect#memset` to zero memory.

puzpuzpuz:
I've noticed https://github.com/questdb/questdb/pull/6362/files#r2537236278 while debugging the new factory. Please take a look.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
puzpuzpuz: I see. Can we add an assertion to make sure that we always free what was in the `firstIteratorBlockAddr` field?

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
mtopolnik: Block header is 16 bytes vs. block size of 40 kilobytes, so shrinking `usedSlotcount` and `nextFreeSlot` won't move the needle.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
puzpuzpuz: Looks like you need to push the commit. :)

core/src/test/java/io/questdb/test/griffin/engine/join/MarkoutHorizonCrossJoinTest.java
```diff
@@ -0,0 +1,835 @@
...
+
+import io.questdb.cairo.CursorPrinter;
+import io.questdb.cairo.sql.RecordCursor;
+import io.questdb.cairo.sql.RecordCursorFactory;
+import io.questdb.std.str.StringSink;
+import io.questdb.test.AbstractCairoTest;
+import io.questdb.test.tools.TestUtils;
+import org.junit.Test;
+
+public class MarkoutHorizonCrossJoinTest extends AbstractCairoTest {
```
puzpuzpuz, line 35: nit: might be a good idea to add a fuzz test similar to the `AsOfJoinFuzzTest`. The oracle would be vanilla CROSS JOIN + ORDER BY.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
mtopolnik: Seems like it's impossible for the column to be anything inappropriate, and still be present as the operand of `+`, which is checked.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
mtopolnik: Removed in 90fa057b7ac713001543425b5a514d80675e9d74.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
mtopolnik: `memset` is a native method, whereas `setMemory` is an intrinsic. I changed the code to clear just the header, so I think the JNI call overhead would dominate the call.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,631 @@
...
+            // Free all iterator blocks in the linked list
+            freeAllIteratorBlocks();
+            firstIteratorBlockAddr = 0;
+            lastIteratorBlockAddr = 0;
+            currentIterAddr = 0;
+            prevIterAddr = 0;
+
+            Misc.free(slaveRecordArray);
+            super.close();
+        }
```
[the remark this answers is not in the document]
mtopolnik: `RecordArray` doesn't need reopening, but I added a no-op `reopen()` call to `cursor.of()`

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,631 @@
...
+                return;
+            }
+            nextSlaveRowNum++;
+            iter_setNextSlaveRowNum(iterAddr, nextSlaveRowNum);
+            long slaveRecordOffset = slaveRecordOffsets.getQuick(nextSlaveRowNum);
+            Record slaveRec = slaveRecordArray.getRecordAt(slaveRecordOffset);
+            long slaveOffset = slaveRec.getLong(slaveSequenceColumnIndex);
+            long masterTimestamp = iter_masterTimestamp(iterAddr);
+            iter_setNextTimestamp(iterAddr, masterTimestamp + slaveOffset);
+        }
```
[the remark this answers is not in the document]
mtopolnik: Addressed in daa96d95beaafb20aca0ba93a45cf67ef910e390.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -1594,6 +1598,128 @@ private RecordCursorFactory createSpliceJoin(
...
+            return null;
+        }
+        int slaveColumnIndex = slaveMetadata.getColumnIndexQuiet(slaveColumnNode.token);
+        if (slaveColumnIndex == -1) {
+            return null; // Slave column not found
+        }
+        // Return the info needed to create the optimized factory.
+        // resultTimestampColumnIndex will be -1 if no SELECT column matches the ORDER BY expression.
+        return new MarkoutHorizonInfo(timestampColumnIndex, slaveColumnIndex);
+    }
```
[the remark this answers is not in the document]
mtopolnik: ASC ordering detection added in d8dbff91610e45242399d6f87c200a0928c788e2.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -7160,6 +7315,19 @@ interface ModelOperator {
...
+    private static class MarkoutHorizonInfo {
+        int masterTimestampColumnIndex;
+        int slaveSequenceColumnIndex;
+
+        MarkoutHorizonInfo(int masterTimestampColumnIndex, int slaveSequenceColumnIndex) {
+            this.masterTimestampColumnIndex = masterTimestampColumnIndex;
+            this.slaveSequenceColumnIndex = slaveSequenceColumnIndex;
+        }
+    }
+
```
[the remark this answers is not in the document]
mtopolnik: Comment edited in d8dbff91610e45242399d6f87c200a0928c788e2.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
mtopolnik: Addressed in a58d915ccd951a5b287c9435b20b0f4e7eb7ed2b.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
mtopolnik: Assert added in 525f61fcfcfa6ae1bf9d652afd7bbbe6b3d571ec.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
mtopolnik: Implemented in 4193f18d779af7184182b0b7ca5268ce8ba45efa.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
mtopolnik: Covered in `testIteratorBlockFreelist()`

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
mtopolnik: Code no longer exists after a bugfix.

core/src/test/java/io/questdb/test/griffin/engine/join/MarkoutHorizonCrossJoinTest.java
mtopolnik: I partially use this idea since all tests compare against the oracle. The added test `testIteratorBlockFreelist()` covers all codepaths around iterator block allocation/deallocation/reuse.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
mtopolnik: Done in 40a1dda2ccc89f82cdcc6b70035baf11f487073d

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
mtopolnik: Covered in `testIteratorBlockFreelist()`

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
mtopolnik: Covered in d963623adeac4e5a7e85f0cc413fadd8b027ace2, `testIneligibleQueriesRejected()`

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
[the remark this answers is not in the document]
mtopolnik: And fixed in dd7a2b32012b2f07536001bbe98c2a767d950563

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
mtopolnik: There are probably many other ways to foil the detection. If there's an easy way to check this too, I can add it, but otherwise I'd leave it at the current level of sophistication.

mtopolnik:
> I've noticed https://github.com/questdb/questdb/pull/6362/files#r2537236278 while debugging the new factory. Please take a look.

Addressed in a1abc776c91910599330b619d455295ad8bc626a

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
puzpuzpuz: We could check that the `lastSeenOrderByModel` is the immediate outer model for the cross join model. Treat it as a nit though.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -1594,6 +1599,138 @@ private RecordCursorFactory createSpliceJoin(
...
+        }
+        // Ensure that ORDER BY is an addition operation
+        if (orderByExpr.type != ExpressionNode.OPERATION || !"+".contentEquals(orderByExpr.token)) {
+            return null;
+        }
+        // The ORDER BY should be timestamp_column + offset_column, or vice versa
+        ExpressionNode lhs = orderByExpr.lhs;
+        ExpressionNode rhs = orderByExpr.rhs;
+        if (lhs == null || rhs == null) {
+            return null;
```
puzpuzpuz, line 1697: This branch is still not covered by the `MarkoutHorizonCrossJoinTest` test.

mtopolnik: Yes, this is just a safeguard. I tried and failed to construct a query that has a `+` but no `lhs` or `rhs`.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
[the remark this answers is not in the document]
puzpuzpuz: > but I added a no-op `reopen()` call to `cursor.of()`

Why do we have `RecordChain` and `RecordArray` as `Reopenable` considering that their impls are no-op? I'd rather remove the interface from them.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,664 @@
...
+            firstIteratorBlockAddr = 0;
+            lastIteratorBlockAddr = 0;
+            currentIterAddr = 0;
+            prevIterAddr = 0;
+        }
+
+        private void gotoNextRow(long iterAddr) {
+            int nextSlaveRowNum = iter_nextSlaveRowNum(iterAddr);
+            if (nextSlaveRowNum == slaveRowCount) {
+                return;
```
puzpuzpuz: One more never taken branch. We need to cover it with a test.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -1594,6 +1599,138 @@ private RecordCursorFactory createSpliceJoin(
...
+
+        if (slaveColumnNode == null || slaveColumnNode.type != ExpressionNode.LITERAL) {
+            return null;
+        }
+        int slaveColumnIndex = slaveMetadata.getColumnIndexQuiet(slaveColumnNode.token);
+        if (slaveColumnIndex == -1) {
+            return null; // Slave column not found
+        }
+        // Return the info needed to create the optimized factory.
+        // timestampColumnIndex will be -1 if no SELECT column matches the ORDER BY expression.
```
puzpuzpuz: The comment seems misleading: `timestampColumnIndex` will never be `-1` at this point. That's because it's initialized along with `slaveColumnNode` and there is the following check above:
```java
        if (slaveColumnNode == null || slaveColumnNode.type != ExpressionNode.LITERAL) {
            return null;
        }
```

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,664 @@
...
+        }
+
+        @Override
+        public void calculateSize(SqlExecutionCircuitBreaker circuitBreaker, Counter counter) {
+            long size = size();
+            if (size != -1) {
+                isMasterHasNextPending = false;
+                masterHasNext = false;
+                freeAllIteratorBlocks();
+                counter.add(size() - emittedRowCount);
```
puzpuzpuz: Why not
```suggestion
                counter.add(size - emittedRowCount);
```
?

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,664 @@
...
+            } else {
+                super.calculateSize(circuitBreaker, counter);
```
puzpuzpuz, line 262: nit: why don't we call `calculateSize` on both master and slave cursor in order to calculate the total size here? `super.calculateSize()` will add markout horizon overhead on top of that.

mtopolnik: Slave cursor gets exhausted immediately in `of()`, and we know its size -- `slaveRowCount`. However, it won't help us because the slave rows are being iterated over by all the iterators concurrently, in various phases of progress. So to avoid the MxN operation of `super.calculateSize()`, I'd have to make the iterator struct larger, or redesign it, so that I have its current count. I'd then go through all the iterators, count their pending rows, and then add the pending master rows. I felt this wasn't worth the effort at this point.

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
mtopolnik: Addressed in 3763dd2e46a73633c8ddb64e0427d6ab121c9c8f

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
mtopolnik: Removed comment in 3239a543455dbfd7b49537b4140bdedc1b630383

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
mtopolnik: 0fffd9c88dbcd32b2f294493f2618a6b1abc6a98

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
puzpuzpuz: > I'd then go through all the iterators, count their pending rows, and then add the pending master rows. I felt this wasn't worth the effort at this point.

Doesn't sound too complicated. The pending master rows is `masterCursor.calculateSize()`. In any case, why don't we at least add a comment on this potential optimization?

core/src/main/java/io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java
```diff
@@ -0,0 +1,640 @@
...
+            }
+            slaveRecordArray.toTop();
+            slaveRowCount = slaveRecordOffsets.size();
+
+            // Cache the first slave's offset value to avoid repositioning recordC later
+            if (slaveRowCount > 0) {
+                long firstSlaveRecordOffset = slaveRecordOffsets.getQuick(0);
+                Record firstSlaveRecord = slaveRecordArray.getRecordAt(firstSlaveRecordOffset);
+                firstSlaveTimeOffset = firstSlaveRecord.getLong(slaveSequenceColumnIndex);
+            }
```
puzpuzpuz, line 637: Looks like we need to set `firstSlaveTimeOffset` to 0 in the `else` here. Slave factory may return different number of rows, e.g. based on a bind variable value.

puzpuzpuz (approved):

core/src/test/java/io/questdb/test/griffin/engine/join/AsOfJoinFuzzTest.java
```diff
@@ -47,7 +47,7 @@
 import java.util.List;
 
 public class AsOfJoinFuzzTest extends AbstractCairoTest {
-    private static final boolean RUN_ALL_PERMUTATIONS = true;
+    private static final boolean RUN_ALL_PERMUTATIONS = false;
```
bluestreak01, line 50: why was this changed?

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 334 / 353 (94.62%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/RecordTreeChain.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java|253|269|94.05%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|78|80|97.50%|
|:large_blue_circle:|io/questdb/cairo/RecordChain.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/RecordArray.java|1|1|100.00%|

core/src/test/java/io/questdb/test/griffin/engine/join/AsOfJoinFuzzTest.java
mtopolnik: It was always supposed to be false. While developing I set it to true for thorough testing, but that takes a very long time. I must have accidentally committed it as `true`.

Issue #6226, opened 2025-10-04
**Redesign Footer Section for Better Aesthetics and Usability**


akofficial10:
### Is your feature request related to a problem?

I experienced difficulties with the current footer layout being **too minimal and visually unbalanced**.  
While it functions well, it lacks modern design hierarchy, spacing consistency, and clear visual distinction between content groups (Product, Developers, Social, Company).  
On smaller screens, the layout also feels a bit dense and could use better responsive behavior.

### Describe the solution you'd like.

I would like to have a **modern, visually enhanced, and responsive footer** that:
- Uses a **grid/flex-based layout** with equal spacing between link groups.  
- Features **clear typography hierarchy** and **consistent alignment**.  
- Incorporates **icons for social links** (Twitter, GitHub, LinkedIn, YouTube, etc.) for visual clarity.  
- Improves the **newsletter subscription box** with rounded fields, better contrast, and a bold CTA button.  
- Enhances the **GitHub star button** with hover animation or a gradient effect.  
- Adds **hover effects** for links and interactive elements to improve UX.  

### Describe alternatives you've considered.

An alternative approach could be **simplifying the layout** further by using collapsible sections for mobile users (accordion-style), reducing clutter while maintaining accessibility.  
However, the current footer can be redesigned with modern UI patterns without sacrificing functionality.

### Full Name:

Ayush Kumar

### Affiliation:

Independent Hacktoberfest Contributor 

### Additional context

Please find the current footer design screenshot below for reference:

<img width="1919" height="804" alt="Image" src="https://github.com/user-attachments/assets/ff11dab2-c42f-4dc1-b7a5-3f8ebcd00750" />

**Suggested Enhancements Summary:**
- Add spacing, hierarchy, and consistent visual alignment  
- Improve contrast and typography for dark mode readability  
- Make it responsive with adaptive stacking  
- Add subtle hover animations  
- Include optional “Back to Top” button or status info

nwoolmer:
Hi @akofficial10 , thanks for your suggestions. This part of QuestDB is non-public and not part of this repo. We'll take the suggestions under consideration, but I will close the issue for now.

PR #6286, opened 2025-10-17
feat(cairo): add Parquet bulk load guide

anikchand461:
Added a new section to `README.md` under "Connect to QuestDB" to document the bulk loading of high-volume time-series data using Parquet files with the `INSERT INTO ... SELECT ... FROM read_parquet()` syntax, as suggested in issue #6226. This guide specifies placing Parquet files in the `/import` directory (configurable via `COPY_INPUT_ROOT`) for optimal performance.

**Changes**
- Added "Bulk Loading with Parquet Files" section with SQL example.

**Notes**
- Benchmark data (e.g., load time for 100M rows) is TBD as testing was not performed. Happy to add this if requested.
- Dataset generation is left to users via a suggested Python script (not included).


nwoolmer:
Ah Anik, guide along these lines are probably better added to the docs repo: https://github.com/questdb/documentation

Rather than the readme!

anikchand461:
then what to do now ? @nwoolmer

PR #6291, opened 2025-10-21
feat(sql): add last(array) function

mcadariu:
The next step after #6237

It relates to https://github.com/questdb/questdb/issues/4120.

In https://github.com/questdb/questdb/pull/4097, a specialized allocator for aggregate functions was introduced and a couple of them migrated to use it. This PR follows up with the implementation of `last(array)`.


core/src/test/java/io/questdb/test/cairo/fuzz/ParallelGroupByFuzzTest.java
```diff
@@ -703,11 +703,25 @@ public void testParallelGroupByArrayFirst() throws Exception {
...
+                        "[0.21583224269349388,0.15786635599554755,null]\tk0\n" +
+                        "[0.2845577791213847,0.20447441837877756]\tk1\n" +
+                        "[0.19202208853547864,0.5093827001617407,0.11427984775756228]\tk2\n" +
+                        "[null,0.7261136209823622]\tk3\n" +
+                        "[0.3100545983862456,0.1985581797355932,0.33608255572515877]\tk4\n"
+        );
+    }
+
+    @Test
+    public void testParallelGroupByArrayLastKeyed() throws Exception {
```
puzpuzpuz: nit: could be merged with `testParallelGroupByArrayFirst` - both `first()` and `last()` functions can be verified in a single test.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/LastArrayGroupByFunction.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    @Override
+    public void clear() {
+        sink.of(0);
+    }
+
+    @Override
+    public void close() {
+        Misc.free(arg);
```
puzpuzpuz: nit: that's the default implementation of `UnaryFunction`, so no need for this override.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/LastArrayGroupByFunction.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    @Override
+    public void computeNext(MapValue mapValue, Record record, long rowId) {
+        updateLast(mapValue, record, rowId);
+    }
+
+    private void updateLast(MapValue mapValue, Record record, long rowId) {
+        mapValue.putLong(valueIndex, rowId);
+        sink.of(0);
```
puzpuzpuz: We don't reuse the memory in case when this method is called from the `computeNext` call. So, for every invocation (think, scanned row) we'll allocate a new buffer - this is inefficient. Instead, we should do smth like the following:
```java
@Override
    public void computeFirst(MapValue mapValue, Record record, long rowId) {
        mapValue.putLong(valueIndex, rowId);
        sink.of(0);
        sink.put(arg.getArray(record));
        mapValue.putLong(valueIndex + 1, sink.ptr());
    }

    @Override
    public void computeNext(MapValue mapValue, Record record, long rowId) {
        mapValue.putLong(valueIndex, rowId);
        long ptr = rec.getLong(valueIndex + 1);
        sink.of(ptr);
        sink.put(arg.getArray(record)); // the sink should overwrite the previously stored array and, if needed, reallocate the buffer
        mapValue.putLong(valueIndex + 1, sink.ptr());
    }
```

core/src/test/java/io/questdb/test/cairo/fuzz/ParallelGroupByFuzzTest.java
mcadariu: Indeed! I added an extra column, this way no extra overhead.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/LastArrayGroupByFunction.java
mcadariu: Good one, removed.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/LastArrayGroupByFunction.java
mcadariu: Looking again at the `GroupByArraySink`, indeed. If it's always doing `sink.of(0)` the old buffer is wasted. It's fine to do it in `computeFirst` (first time). But in `computeNext` it should reuse or reallocate. I've updated the code.

mcadariu:
While working on this, I have identified two other tasks: 
- update [docs](https://github.com/questdb/documentation) (after this PR is merged). I noticed that [`first`](https://questdb.com/docs/reference/function/aggregation/#first) does not mention the `array` type now, but I have just seen it working. This update would be fine to do even now, but can do it together with `last` in one go. 
- update `FirstArrayGroupByFunction` to use `GroupByArraySink` for consistency with `LastArrayGroupByFunction`

core/src/main/java/io/questdb/griffin/engine/functions/groupby/LastArrayGroupByFunction.java
```diff
@@ -0,0 +1,156 @@
...
+        mapValue.putLong(valueIndex, Numbers.LONG_NULL);
+        mapValue.putLong(valueIndex + 1, 0);
+    }
+
+    public boolean supportsParallelism() {
+        return true;
+    }
+
+    @Override
+    public void toTop() {
```
puzpuzpuz: nit: this override is redundant since `UnaryFunction`'s implementation will be used anyway.

puzpuzpuz:
@mcadariu the PR looks good, but it's still marked as a draft. Are you planning to push some additional changes?

mcadariu:
@puzpuzpuz Yes, I have just pushed [this commit](https://github.com/questdb/questdb/pull/6291/commits/867ceac18e6645e784bb277a2b53c364e2356ca5) to remove unnecessary code additions. I'm now making the most out of the existing test setup introduced by the `first(array)` to test `last(array)`. I won't make other changes. Thanks for the review!

puzpuzpuz (dismissed):

puzpuzpuz:
Many thanks for another solid contribution!

core/src/main/java/io/questdb/griffin/engine/functions/groupby/LastArrayGroupByFunction.java
mcadariu: Ah right, good one. I must've gotten that from the `FirstArrayGroupByFunction`. Another reason that one should be tackled too, now that we're at it.

PR #6298, opened 2025-10-22
test(pgwire): infrastructure to run PQ Golang tests

shubhammenroy:
Add the infrastructure to run tests against PQ Golang

### DIFF
--- a/ci/templates/hosted-jobs.yml
+++ b/ci/templates/hosted-jobs.yml
@@ -118,5 +118,6 @@ jobs:
       CODE_COVERAGE_TOOL_OPTION: $[stageDependencies.CheckChanges.CheckChanges.outputs['check_coverage.CODE_COVERAGE_TOOL_OPTION']]
       DIFF_COVER_THRESHOLD_PCT: 50
       ARCHIVED_CRASH_LOG: "$(Build.ArtifactStagingDirectory)/questdb-crash-$(Build.SourceBranchName)-$(Build.SourceVersion)-$(System.StageAttempt)-$(Agent.OS)-$(jdk).log"
+      MAVEN_OPTS: "-Xmx2g"
     steps:
       - template: steps.yml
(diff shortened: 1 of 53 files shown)


jerrinot:
hello @shubhammenroy, thank you for this very useful PR!

The only bit I am unsure about is the introduction of the driver concept. I understand you aim to reuse the main loop code, but I am not sure if it's worth it. It's hard enough to grasp mappings of yaml scripts to actual clients. Having another abstraction, albeit a very thin, makes things a bit more complex. For other clients I decided to accept code duplication if it gives me less abstractions. Although the Python runners share some conversion utilities, but the main flow is separated. WDYT?

shubhammenroy:
> hello @shubhammenroy, thank you for this very useful PR!
> 
> The only bit I am unsure about is the introduction of the driver concept. I understand you aim to reuse the main loop code, but I am not sure if it's worth it. It's hard enough to grasp mappings of yaml scripts to actual clients. Having another abstraction, albeit a very thin, makes things a bit more complex. For other clients I decided to accept code duplication if it gives me less abstractions. Although the Python runners share some conversion utilities, but the main flow is separated. WDYT?

Hi @jerrinot, Thanks for the kind words and for reviewing this!

Fair point about the extra abstraction. I added the driver interface to handle differences between pgx and lib/pq in arrays, nulls, and affected rows while keeping a single test flow (prepare → steps → teardown). The main reason for this approach was to avoid duplicating the core execution flow and make it easier to add new drivers in the future. For example, if anyone want 's to add pgconn or go-pg, they can simply implement the interface instead of duplicating the entire suite.

As per my thoughts, the interface approach fits best here, but I’m happy to adjust toward the Python-style pattern if you prefer.

jerrinot:
let's go with what you have. no need to overthink it, this is supposed to be simple. I tried it locally, tests are passing :ok_hand:

shubhammenroy:
> let's go with what you have. no need to overthink it, this is supposed to be simple. I tried it locally, tests are passing 👌

Hi @jerrinot, 
Thanks again for the review and for giving the green light on the implementation approach!
With the latest comment confirming the tests are passing locally, this PR should now be ready for approval and merge.
Let me know if you need anything else from my side.

jerrinot (approved):

jerrinot:
@shubhammenroy many thank for your contribution!

PR #6302, opened 2025-10-23
SQL: Overload for sum(boolean)

klaudiozdrava:
This PR enhances the SUM() aggregate function to support conditional aggregation. It allows a boolean expression to be passed directly, replacing the verbose SUM(CASE...END) pattern for counting rows that meet a specific criteria.

To implement this, I've added two new classes following existing conventions: `SumBoolGroupByFunctionFactory` and `SumBoolGroupByFunction`.

The SumBoolGroupByFunctionFactory registers the sum(boolean) signature. The `SumBoolGroupByFunction` then implements the aggregation logic:

On the first record, it initializes the value to 1 if the expression is true, or 0 if false.

For all subsequent records, it increments the count by 1 if the expression is true.

This returns the total number of rows that satisfy the condition. I've also added `SumBoolGroupByFunctionTest` to test all comparison operators (=, <, >, !=).


klaudiozdrava:
Hi @jerrinot! Just a friendly check-in on this PR when you have a moment.
If there's anything I should improve or adjust, I'd be happy to work on it. Thanks!

bluestreak01:
@coderabbit review

bluestreak01:
https://github.com/questdb/questdb/pull/6658

PR #6322, opened 2025-10-28
chore(sql): port first(array) to use array sink

mcadariu:

In https://github.com/questdb/questdb/pull/6237 the `GroupByArraySink` abstraction was introduced. This enables simplifying the `FirstArrayGroupByFunction` because it can delegate the manual array handling to it. In addition its implementation will be symmetrical with `last(array)` from https://github.com/questdb/questdb/pull/6291.

### DIFF
--- a/ci/test-fuzz.yml
+++ b/ci/test-fuzz.yml
@@ -8,7 +8,6 @@ schedules:
     branches:
       include:
         - master
-        - master-ent-next
     always: true
 
 variables:
--- a/core/src/main/java/io/questdb/cutlass/text/CopyExportContext.java
+++ b/core/src/main/java/io/questdb/cutlass/text/CopyExportContext.java
@@ -382,7 +382,7 @@ public class CopyExportContext {
                         false
                 );
                 createOp.setTableKind(TableUtils.TABLE_KIND_TEMP_PARQUET_EXPORT);
-                createOp.validateAndUpdateMetadataFromSelect(rcf.getMetadata());
+                createOp.validateAndUpdateMetadataFromSelect(rcf.getMetadata(), rcf.getScanDirection());
             }
         } catch (SqlException ex) {
             ex.setPosition(ex.getPosition() + tableOrSelectTextPos);
--- a/core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java
+++ b/core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java
@@ -3550,7 +3550,7 @@ public class SqlCompilerImpl implements SqlCompiler, Closeable, SqlParserCallbac
                     try {
                         final RecordMetadata metadata = newFactory.getMetadata();
                         try (TableReader baseReader = engine.getReader(createMatViewOp.getBaseTableName())) {
-                            createMatViewOp.validateAndUpdateMetadataFromSelect(metadata, baseReader.getMetadata());
+                            createMatViewOp.validateAndUpdateMetadataFromSelect(metadata, baseReader.getMetadata(), newFactory.getScanDirection());
                         }
 
                         matViewDefinition = engine.createMatView(
@@ -3663,7 +3663,7 @@ public class SqlCompilerImpl implements SqlCompiler, Closeable, SqlParserCallbac
                             RecordCursor cursor = newCursor
                     ) {
                         final RecordMetadata metadata = factory.getMetadata();
-                        createTableOp.validateAndUpdateMetadataFromSelect(metadata);
+                        createTableOp.validateAndUpdateMetadataFromSelect(metadata, factory.getScanDirection());
                         boolean keepLock = !createTableOp.isWalEnabled();
 
                         // todo: test create table if exists with select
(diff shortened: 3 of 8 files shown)


core/src/main/java/io/questdb/griffin/engine/functions/groupby/FirstArrayGroupByFunction.java
```diff
@@ -26,34 +26,28 @@
...
+    private final GroupByArraySink sink;
     private int valueIndex;
 
     public FirstArrayGroupByFunction(@NotNull Function arg) {
         this.arg = arg;
         this.type = arg.getType();
-    }
-
-    @Override
-    public void clear() {
```
puzpuzpuz, line 56: We should keep the `clear()` impl:
```java
    @Override
    public void clear() {
        sink.of(0);
    }
```

core/src/main/java/io/questdb/griffin/engine/functions/groupby/FirstArrayGroupByFunction.java
```diff
@@ -64,15 +58,9 @@ public void close() {
...
-        final ArrayView val = arg.getArray(record);
-        if (val == null || val.isNull()) {
-            mapValue.putLong(valueIndex + 1, -1);
-        } else {
-            long size = ArrayTypeDriver.getPlainValueSize(val);
-            long ptr = allocator.malloc(size);
-            ArrayTypeDriver.appendPlainValue(ptr, val);
-            mapValue.putLong(valueIndex + 1, ptr);
-        }
+        sink.clear();
```
puzpuzpuz: Why not `sink.of(0);` like in #6291? The end result is probably the same, but it's good to keep the code consistent.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/FirstArrayGroupByFunction.java
```diff
@@ -82,17 +70,13 @@ public void computeNext(MapValue mapValue, Record record, long rowId) {
...
 
     @Override
     public ArrayView getArray(Record rec) {
         final long ptr = rec.getLong(valueIndex + 1);
-        if (ptr < 0) {
-            return ArrayConstant.NULL;
-        }
-
-        return ArrayTypeDriver.getPlainValue(ptr, borrowedArray);
+        return sink.of(ptr).getArray();
```
puzpuzpuz: In #6291 the function looks different:
```java
    @Override
    public ArrayView getArray(Record rec) {
        long ptr = rec.getLong(valueIndex + 1);
        sink.of(ptr);
        ArrayView array = sink.getArray();
        if (array == null) {
            return ArrayConstant.NULL;
        }
        return array;
    }
```

core/src/main/java/io/questdb/griffin/engine/functions/groupby/FirstArrayGroupByFunction.java
```diff
@@ -142,22 +126,19 @@ public void merge(MapValue destValue, MapValue srcValue) {
...
 
     @Override
     public void setNull(MapValue mapValue) {
         mapValue.putLong(valueIndex, Numbers.LONG_NULL);
-        mapValue.putLong(valueIndex + 1, -1);
+        mapValue.putLong(valueIndex + 1, 0);
     }
 
+    @Override
     public boolean supportsParallelism() {
```
puzpuzpuz, line 152: I know that this is the previously existing impl, but would be great to change it to a more correct version:
```java
    @Override
    public boolean supportsParallelism() {
        return UnaryFunction.super.supportsParallelism();
    }
```

mcadariu: Yes, I updated it.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/FirstArrayGroupByFunction.java
mcadariu: Hmm yes good one,  the null-check. In the other PR, I remember I've added it because of a unit test failing. I am wondering then if this is a sign to introduce a separate test class like [this](https://github.com/mcadariu/questdb/blob/74386bcde66f7bc0aa9f889feb0cccbcf50934af/core/src/test/java/io/questdb/test/griffin/engine/functions/groupby/LastArrayGroupByFunctionFactoryTest.java) for this one as well.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/FirstArrayGroupByFunction.java
mcadariu: Right! I brought back the `clear` method.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/FirstArrayGroupByFunction.java
mcadariu: Yes indeed, updated.

mcadariu:
Thanks for the quick review!

core/src/main/java/io/questdb/griffin/engine/functions/groupby/FirstArrayGroupByFunction.java
puzpuzpuz: Yes, a separate `FirstArrayGroupByFunctionFactoryTest` test would be a good addition.

mcadariu: Done!

puzpuzpuz:
Thanks for going an extra mile with this `first(array)` clean up!

PR #6333, opened 2025-10-30
feat(sql): require a query hint to enable rare-symbols optimization in ASOF JOIN

mtopolnik:
The Memoized Scan algorithm for ASOF JOIN on symbol column has two kinds of caching:

1. Match Caching: after finding a matching symbol in the RHS column, cache its location so we don't have to repeat the search later
2. Drive-By Caching: while looking for a matching symbol, cache all symbols seen in passing as well, in the hope future LHS rows will need some of them

For Match Caching, we only need a single hash lookup per LHS row for a cache hit, and another one for a cache miss.

For Drive-By Caching, we need a hash lookup for every visited RHS row. It pays off when many of the symbols to look for occur in the deep past of the RHS table. It allows us to make just one deep scan and find all of  them. However, when there aren't many such symbols, hash lookups introduce a significant overhead, and result in worse performance in many typical cases.

Therefore, we're putting Drive-By Caching behind a query hint.

### DIFF
--- a/core/src/test/java/io/questdb/test/griffin/SqlOptimiserTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/SqlOptimiserTest.java
@@ -1307,6 +1307,7 @@ public class SqlOptimiserTest extends AbstractSqlParserTest {
                                     SelectedRecord
                                         AsOf Join Memoized Scan
                                           condition: t2.s=t1.s
+                                          driveByCache: false
                                             PageFrame
                                                 Row forward scan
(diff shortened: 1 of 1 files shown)


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java|10|10|100.00%|

RaphDal (approved):
Looks good to me.

PR #6344, opened 2025-11-03
feat(sql): add first_not_null(array) 

mcadariu:

The implementation is inspired by `FirstNotNullStrGroupByFunction`. The difference is that the boolean column type is present only in `FirstNonNullArrayGroupByFunction` and not in the superclass, like in the `FirstStrGroupByFunction` hierarchy. This way the boolean column is present only where strictly needed.


core/src/main/java/io/questdb/griffin/engine/functions/groupby/FirstNonNullArrayGroupByFunctionFactory.java
```diff
@@ -0,0 +1,55 @@
...
+package io.questdb.griffin.engine.functions.groupby;
+
+import io.questdb.cairo.CairoConfiguration;
+import io.questdb.cairo.sql.Function;
+import io.questdb.griffin.FunctionFactory;
+import io.questdb.griffin.SqlExecutionContext;
+import io.questdb.std.IntList;
+import io.questdb.std.ObjList;
+
+public class FirstNonNullArrayGroupByFunctionFactory implements FunctionFactory {
```
puzpuzpuz: Could you change the class name to `FirstNotNull*` here and in `FirstNonNullArrayGroupByFunction`? All other classes are named `FirstNotNull*`, not `FirstNonNull`.

core/src/test/java/io/questdb/test/griffin/engine/functions/groupby/FirstNonNullArrayGroupByFunctionFactoryTest.java
```diff
@@ -0,0 +1,156 @@
...
+ *  limitations under the License.
+ *
+ ******************************************************************************/
+
+package io.questdb.test.griffin.engine.functions.groupby;
+
+import io.questdb.test.AbstractCairoTest;
+import org.junit.Test;
+
+public class FirstNonNullArrayGroupByFunctionFactoryTest extends AbstractCairoTest {
```
puzpuzpuz: Would be great to include a test for the empty array case.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/FirstNonNullArrayGroupByFunction.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @Override
+    public String getName() {
+        return "first_non_null";
+    }
+
+    @Override
+    public void initValueTypes(ArrayColumnTypes columnTypes) {
+        super.initValueTypes(columnTypes);
+        columnTypes.add(ColumnType.BOOLEAN);
```
puzpuzpuz: Can't we treat zero pointer as `null`? This way we won't need additional boolean value type.

puzpuzpuz (commented):
There is a merge conflict in the parallel group by test. Could you resolve it?

core/src/main/java/io/questdb/griffin/engine/functions/groupby/FirstNonNullArrayGroupByFunctionFactory.java
mcadariu: Good catch! I've fixed also the `first_not_null` function name.

core/src/test/java/io/questdb/test/griffin/engine/functions/groupby/FirstNonNullArrayGroupByFunctionFactoryTest.java
mcadariu: Yes good one, done.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/FirstNonNullArrayGroupByFunction.java
mcadariu: Nice! Yes it works.
I now understand why QuestDB is so fast 😄

mcadariu:
Thanks for your feedback!

puzpuzpuz (approved):

puzpuzpuz:
Thanks for another solid contribution!

mcadariu:
Thanks @bluestreak01!

PR #6345, opened 2025-11-03
chore(core): only track resources that supports supervisors

RaphDal:
As per the title, this PR fixes a leak that occurs whenever a supervisor (`QueryProgress`/`TracingResourcePoolSupervisor`) borrows a resource but the resource doesn't support supervisors. This causes the resource to never release itself from the supervisor before being returned to the pool.

The fix is quite self-explanatory: a resource that supports supervisors **MUST** return the one that is attached when the `getSupervisor` method is called. Based on this assumption, before borrowing the resource, we check if a supervisor is indeed attached to it instead of retrieving the local one.

Thanks to @jerrinot for detecting and reporting this issue!


RaphDal:
@coderabbit review

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/pool/AbstractMultiTenantPool.java|2|2|100.00%|

bluestreak01 (approved):

PR #6347, opened 2025-11-03
chore(sql): speedup tests

bluestreak01:
Speedup tests by replacing JUnit parameterisation with randomisation.

Some tests were also creating 100M row tables (on CI, because why not)

### DIFF
--- a/ci/templates/hosted-jobs.yml
+++ b/ci/templates/hosted-jobs.yml
@@ -61,7 +61,7 @@ jobs:
           jdk: "1.17"
           testset: "all"
           includeTests: "**/griffin/*"
-          # TODO: Remove O3*Test** exclusion - O3 tests should run on Windows
+          # Exclude O3 tests: run in windows-other-2
           # Exclude Fuzz tests: run in windows-fuzz1 (cairo) and windows-fuzz2 (non-cairo)
           excludeTests: "**/O3*Test**,**/**Fuzz**"
           javadoc_step: ""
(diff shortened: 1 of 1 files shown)


bluestreak01:
@coderabbit review the test changes and also checked if i regressed anything?

bluestreak01:
@Coderabbit i addressed your comments, please check

bluestreak01:
@coderabbit we run CI very often, random seeds are logged and we will eventually go over all combinations. CI boxes are extremely slow and resource constrained, running full suite of tests would either max out our parallelism or box capacity. We are squeezed from both sides.

Review my new changes, i applied the same pattern to other tests

ci/templates/hosted-jobs.yml
```diff
@@ -74,18 +74,27 @@ jobs:
...
-        windows-cairo:
+        windows-cairo-1:
           imageName: "windows-latest"
           poolName: "Azure Pipelines"
           os: Windows
           jdk: "1.17"
           testset: "all"
           includeTests: "**/cairo/**"
-          excludeTests: "**/O3*Test**,**/**Fuzz**"
+          excludeTests: "**/O3*Test**,**/**Fuzz**,**/cairo/mv/**,**/cairo/o3/**"
```
jerrinot, line 112: It appears Windows will completely skip O3 tests unless they are in `/cairo/o3/` 

For example `io.questdb.test.cairo.O3BasketTest` won't ever be executed on Windows:
```
windows-griffin explicitly excludes **/O3*Test**
windows-cairo-1 explicitly excludes **/O3*Test**
windows-cairo-2 only includes **/cairo/mv/** and **/cairo/o3/** subdirs
windows-fuzz1   only includes **/cairo/fuzz/**
windows-fuzz2   only includes **/**Fuzz**
windows-pgwire  only includes **/pgwire/**
windows-other-1 excludes **/cairo/**
windows-other-2 only includes **/cutlass/http/**
```

jerrinot: this is now resolved

glasstiger:
#### [PR Coverage check]

PR #5764, opened 2025-06-19
feat(core): optimize to not rewrite partitions when the same data is re-inserted with dedup

ideoma:

PR #6360, opened 2025-11-07
fix(core): fix potential storage corruption on deduplicate write resulting to same data

ideoma:
Moves #6359 to `master-ent-next`

Critical storage issue introduced in #5764 and released in 9.0.0.

Very likely to happen on a materialized view refresh that results in the same data as already in the view.
The bug may result in double memory close and the deletion of the wrong directories on the box.


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|

PR #6368, opened 2025-11-10
feat(sql): add last_not_null(array)

mcadariu:

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/functions/groupby/FirstNotNullArrayGroupByFunction.java
+++ b/core/src/main/java/io/questdb/griffin/engine/functions/groupby/FirstNotNullArrayGroupByFunction.java
@@ -28,7 +28,6 @@ import io.questdb.cairo.arr.ArrayView;
 import io.questdb.cairo.map.MapValue;
 import io.questdb.cairo.sql.Function;
 import io.questdb.cairo.sql.Record;
-import io.questdb.griffin.engine.functions.constants.ArrayConstant;
 import org.jetbrains.annotations.NotNull;
 
 public class FirstNotNullArrayGroupByFunction extends FirstArrayGroupByFunction {
@@ -49,14 +48,6 @@ public class FirstNotNullArrayGroupByFunction extends FirstArrayGroupByFunction 
                 mapValue.putLong(valueIndex + 1, sink.ptr());
             }
         }
-    }
-
-    @Override
-    public ArrayView getArray(Record rec) {
-        if (rec.getLong(valueIndex + 1) == 0) {
-            return ArrayConstant.NULL;
-        }
-        return super.getArray(rec);
     }
 
     @Override


core/src/main/java/io/questdb/griffin/engine/functions/groupby/NotNullArrayGroupByFunction.java
```diff
@@ -0,0 +1,105 @@
...
+import io.questdb.cairo.map.MapValue;
+import io.questdb.cairo.sql.Function;
+import io.questdb.cairo.sql.Record;
+import io.questdb.griffin.engine.functions.constants.ArrayConstant;
+import io.questdb.std.Numbers;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.function.BiPredicate;
+
+public abstract class NotNullArrayGroupByFunction extends FirstArrayGroupByFunction {
```
mcadariu: I suspect that extending `FirstArrayGroupByFunction` here is not the best. The question is if I should revisit the broader inheritance hierarchy in this PR, or we do this in a separate one that will not introduce new functionality but only structural changes.

puzpuzpuz: TBH I'd keep the same approach as in other NotNull group by function, i.e. extending the First function directly. Keeping things consistent is more valuable than having slightly more compact code in one place.

mcadariu: Yes fair point! 
I think possibly it is also another entry for the "why is QuestDB so fast" catalog, the compiler can better optimize when there is less indirection.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/LastNotNullArrayGroupByFunction.java
```diff
@@ -0,0 +1,97 @@
...
+        if (array != null && !array.isNull()) {
+            mapValue.putLong(valueIndex, rowId);
+            long ptr = mapValue.getLong(valueIndex + 1);
+            sink.of(ptr);
+            sink.put(array);
+            mapValue.putLong(valueIndex + 1, sink.ptr());
+        }
+    }
+
+    @Override
```
puzpuzpuz: This override seems unnecessary: `FirstArrayGroupByFunction` will return `ArrayConstant.NULL` in case of null stored value. That's because `sink.of(0)` leads to `null` being returned from `sink.getArray()`.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/LastNotNullArrayGroupByFunction.java
```diff
@@ -0,0 +1,97 @@
...
+import org.jetbrains.annotations.NotNull;
+
+public class LastNotNullArrayGroupByFunction extends FirstArrayGroupByFunction {
+
+    public LastNotNullArrayGroupByFunction(@NotNull Function arg) {
+        super(arg);
+    }
+
+    @Override
+    public void computeFirst(MapValue mapValue, Record record, long rowId) {
```
puzpuzpuz: nit: we could use this very implementation in `FirstArrayGroupByFunction` and then we won't need to override it in `FirstNotNullArrayGroupByFunction` and in this class. WDYT?

core/src/main/java/io/questdb/griffin/engine/functions/groupby/LastNotNullArrayGroupByFunction.java
```diff
@@ -0,0 +1,97 @@
...
+        long srcRowId = srcValue.getLong(valueIndex);
+        long destRowId = destValue.getLong(valueIndex);
+        if (srcRowId > destRowId || destRowId == Numbers.LONG_NULL) {
+            destValue.putLong(valueIndex, srcRowId);
+            destValue.putLong(valueIndex + 1, srcValue.getLong(valueIndex + 1));
+        }
+    }
+
+    @Override
+    public void setNull(MapValue mapValue) {
```
puzpuzpuz: This method is identical to the parent class' method, no need for the override. The same applies to the `FirstNotNullArrayGroupByFunction` class.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/LastNotNullArrayGroupByFunction.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public void merge(MapValue destValue, MapValue srcValue) {
+        if (srcValue.getLong(valueIndex + 1) == 0) {
+            return;
+        }
+        long srcRowId = srcValue.getLong(valueIndex);
+        long destRowId = destValue.getLong(valueIndex);
+        if (srcRowId > destRowId || destRowId == Numbers.LONG_NULL) {
```
puzpuzpuz: It looks like the check should be different here and in `FirstNotNullArrayGroupByFunction`:
```suggestion
        if (srcRowId > destRowId || destValue.getLong(valueIndex + 1) == 0) {
```

That's because we may have met a single row with `null` array value for the given group. In that case, we write the row id to the `valueIndex` slot while `valueIndex+1` will hold zero pointer.

nit: it'd be great to add a test that would break for the old code.

mcadariu:
Thanks for the review! Have a nice weekend!

core/src/main/java/io/questdb/griffin/engine/functions/groupby/LastNotNullArrayGroupByFunction.java
puzpuzpuz: Thanks for addressing this one. How about removing the override from `FirstNotNullArrayGroupByFunction` as well?

mcadariu: Good catch! Done.

PR #6369, opened 2025-11-10
feat(sql): upgrades to querying external parquet files

### LINKED ISSUE #5280: Add projection for virtual table SQL functions like read_parquet()

We don't support projection for virtual table functions like read_parquet(), so if you have, say, 1000 columns in your parquet, but you query only a single column, we'll end up decoding all 1000 columns. This is sub-optimal and should be improved.

_No response_

_No response_

### Full Name:

Andrei Pechkurov

### Affiliation:

QuestDB

### Additional context

_No response_

nwoolmer:
WIP for parquet usability roadmap. **Not** ready for review, lots of refactoring required.

Closes https://github.com/questdb/questdb/issues/5280

**Projection**

Lack of projection causes high memory usage and slow queries when working with large parquet files. This PR aims to address this, by pushing down projections from higher query models directly to the record and page frame cursors.

Q1 Clickbench and hits.parquet

```sql
SELECT count(*) FROM read_parquet('hits.parquet') WHERE AdvEngineId <> 0; 
```

|  | Single-threaded | Multithreaded |
| -- | --- | --- | 
| Master | 106s | DNF (OOM) |
| PR | 670ms | 269ms |

Note: This PR does change some of the permissiveness around Parquet metadata validation. Previously, it was required that the file was always fully decoded and the metadata matched exactly. Now only the projected columns must be readable from the file. Therefore, if you change the schema by changing an underlying column type, it will throw an exception. But if you add an extra column to the schema, it will simply be ignored.

**Glob/hive-partitioned reads**

Generally, large parquet datasets will come in a partitioned form. Querying this is unergonomic, so usually a means to query the files using a wildcard pattern is provided. For example:

```
SELECT count() FROM read_parquet('hits/*.parquet') WHERE AdvEngineId <> 0; -- 1.26s cold, 1.06s hot
SELECT count() FROM read_parquet('hits.parquet') WHERE AdvEngineId <> 0; -- 546ms cold, 270ms hot
```

**Min/max statistics inc. timestamp intrinsics**

**Changelist**

- [ ] Projection
  - [x] Single-threaded read_parquet
  - [x] Multi-threaded read_parquet
  - [ ] Tests for unions, joins etc. 
- [ ] Glob/hive-partitioned reads
  - [x] Querying multiple files at once
  - [x] Fast calculateSize
  - [x] Fast skipRows
  - [ ] Virtual column addition for key=v syntax
      - With a path like `day=2025-01-01/symbol=BTC-USD/data.parquet
      - We should automatically add the day and symbol columns to the projection (this is industry standard).
      - For paths that don't follow this pattern, we do not do this. 
  - [ ] Tests etc. 

Out of scope - minmax stats etc. separate holistic PR required.


core/src/main/java/io/questdb/cairo/sql/PageFrameMemoryPool.java
```diff
@@ -290,13 +298,12 @@ private void openParquet(int frameIndex) {
         parquetColumns.reopen();
         parquetColumns.clear();
         fromParquetColumnIndexes.clear();
-        fromParquetColumnIndexes.setAll(parquetMetadata.getColumnCount(), -1);
+        fromParquetColumnIndexes.setAll(metadataIndex, -1);
         for (int i = 0, n = addressCache.getColumnCount(); i < n; i++) {
             final int columnIndex = addressCache.getColumnIndexes().getQuick(i);
-            final int parquetColumnIndex = toParquetColumnIndexes.getQuick(columnIndex);
```
puzpuzpuz, line 296: Table reader column indexes and parquet column indexes may not match. That's why we have this additional indirection via `toParquetColumnIndexes`.

PR #6370, opened 2025-11-10
chore(sql): avoid large materialized view transactions when possible

puzpuzpuz:
Materialized view refresh could have ended up creating transactions with large row count. That's because the wanted batch size is only checked after each iteration. So, if the sample by iterator step estimation is too low, each iteration may insert too many rows leading to large transactions. As a best-effort protection from this, we now check the number of inserted rows after each iteration and, if it appears to be higher than the batch size, retry with a proportionally reduced interval step.

Also, adds rollback for the WAL writer used by mat view refresh before the next retry attempt is made.

### DIFF
--- a/core/src/test/java/io/questdb/test/cairo/mv/MatViewRefreshRetryTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/mv/MatViewRefreshRetryTest.java
@@ -81,11 +81,21 @@ public class MatViewRefreshRetryTest extends AbstractCairoTest {
 
             drainWalAndMatViewQueues();
             assertQueryNoLeakCheck(
-                    "view_name\tview_status\n" +
-                            "price_1h\tvalid\n",
+                    """
+                            view_name\tview_status
+                            price_1h\tvalid
+                            """,
                     "select view_name, view_status from materialized_views",
                     null,
                     false
+            );
+
+            assertSql(
+                    """
+                            count
+                            56
+                            """,
+                    "select count() from wal_transactions('price_1h')"
             );
         });
     }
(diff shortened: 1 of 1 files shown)


puzpuzpuz:
@coderabbitai review

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 21 / 21 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|21|21|100.00%|

core/src/test/java/io/questdb/test/cairo/mv/MatViewRefreshRetryTest.java
```diff
@@ -50,6 +50,46 @@ public static Collection<Object[]> testParams() {
         });
     }
 
+    @Test
+    public void testLargeTransaction() throws Exception {
```
bluestreak01, line 54: this test passes even if i rollback your changes i the job

bluestreak01: could you add `wal_transactions` check to assert that large transaction was actually split? or something along these lines?

bluestreak01: fixed in https://github.com/questdb/questdb/pull/6370/commits/63466c937d22153618c55975225a61000c4a6b47

bluestreak01 (approved):

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 21 / 21 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|21|21|100.00%|

core/src/test/java/io/questdb/test/cairo/mv/MatViewRefreshRetryTest.java
puzpuzpuz: That's a good idea. Thanks for adding the assertion.

PR #6371, opened 2025-11-10
feat(core): add a GC-free `CharSequence` -> `int` hash map for `WalWriter` symbols

RaphDal:
This PR introduces a new hash map implementation that maps `CharSequence` to `int` without relying on heap allocation or garbage collection.  Its primary goal is to replace the existing CharSequenceIntHashMap used in WalWriter, which currently allocates heap-based String objects, leading to GC-churn during WAL tx commits.

The `WalWriter` implementation is updated to use this new hash map.

Please feel free to comment about improvements I could make to the PR.

# Benchmark

In addition, this PR includes a benchmark to measure and compare the performance and memory allocations of the original `CharSequenceIntHashMap` versus the new `SymbolMap`.

Here is the raw output from the benchmark:
```
Benchmark                                                                 Mode  Cnt        Score        Error   Units
WalWriterSymbolBenchmark.CharSequenceIntHashMap                           avgt    3  1191600.777 ± 218457.103   ns/op
WalWriterSymbolBenchmark.CharSequenceIntHashMap:gc.alloc.rate             avgt    3       57.616 ±     10.496  MB/sec
WalWriterSymbolBenchmark.CharSequenceIntHashMap:gc.alloc.rate.norm        avgt    3    72000.708 ±      0.176    B/op
WalWriterSymbolBenchmark.CharSequenceIntHashMap:gc.count                  avgt    3        6.000               counts
WalWriterSymbolBenchmark.CharSequenceIntHashMap:gc.time                   avgt    3        9.000                   ms
WalWriterSymbolBenchmark.DirectCharSequenceIntHashMap                     avgt    3  1052896.721 ±   2210.717   ns/op
WalWriterSymbolBenchmark.DirectCharSequenceIntHashMap:gc.alloc.rate       avgt    3        0.001 ±      0.001  MB/sec
WalWriterSymbolBenchmark.DirectCharSequenceIntHashMap:gc.alloc.rate.norm  avgt    3        0.622 ±      0.001    B/op
WalWriterSymbolBenchmark.DirectCharSequenceIntHashMap:gc.count            avgt    3          ≈ 0               counts
```

As we can see, we have almost no allocations in the `SymbolMap` version in comparison to the `CharSequenceIntHashMap` implementation, we also have a ~12% performance improvements and less jitter.

# Tests

There are a few tests to verify that the NonGC hash map implementation works properly:
- Basic operations: `put`/`get` and `clear`.
- Insertion-ordered iteration.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/wal/DirectCharSequenceIntHashMap.java
+++ b/core/src/main/java/io/questdb/cairo/wal/DirectCharSequenceIntHashMap.java
@@ -44,7 +44,7 @@ import java.io.Closeable;
 public class DirectCharSequenceIntHashMap implements Closeable, Mutable {
     public static final int NO_ENTRY_VALUE = -1;
     private static final int MIN_INITIAL_CAPACITY = 16;
-    private static final int NO_ENTRY_OFFSET = -1;
+    private static final int NO_ENTRY_OFFSET = 0;
     private final double loadFactor;
     private final int noEntryValue;
     private final DirectString sview = new DirectString();
@@ -132,7 +132,7 @@ public class DirectCharSequenceIntHashMap implements Closeable, Mutable {
             kvAddress = Unsafe.realloc(kvAddress, oldCapacity, kvCapacity, MemoryTag.NATIVE_DEFAULT);
         }
         free = mapCapacity;
-        Vect.memset(address, capacity, -1);
+        Vect.memset(address, capacity, 0);
         this.currentOffset = 0;
     }
 
@@ -219,10 +219,11 @@ public class DirectCharSequenceIntHashMap implements Closeable, Mutable {
      */
     public int keyIndex(@NotNull CharSequence key, int hashCode) {
         int index = Hash.spread(hashCode) & mask;
-        final int offset = getOffset(index);
+        int offset = getOffset(index);
         if (offset == NO_ENTRY_OFFSET) {
             return index;
         }
+        offset -= 1;
         final int hashCode1 = getHashCode(index);
         if (hashCode == hashCode1 && areKeysEquals(offset, key)) {
             return -offset - 1;
@@ -303,7 +304,7 @@ public class DirectCharSequenceIntHashMap implements Closeable, Mutable {
         long oldAddress = address;
         capacity = (long) len << 3;
         address = Unsafe.malloc(capacity, MemoryTag.NATIVE_DEFAULT);
-        Vect.memset(address, capacity, -1);
+        Vect.memset(address, capacity, 0);
 
         for (int i = (int) (oldCapacity >> 3) - 1; i > -1; i--) {
             final long oldPtr = oldAddress + ((long) i << 3);
@@ -385,10 +386,11 @@ public class DirectCharSequenceIntHashMap implements Closeable, Mutable {
     private int probe(@NotNull CharSequence key, int hashCode, int index) {
         do {
             index = (index + 1) & mask;
-            final int offset = getOffset(index);
+            int offset = getOffset(index);
             if (offset == NO_ENTRY_OFFSET) {
                 return index;
             }
+            offset -= 1;
             final int hashCode1 = getHashCode(index);
             if (hashCode == hashCode1 && areKeysEquals(offset, key)) {
                 return -offset - 1;
@@ -398,7 +400,7 @@ public class DirectCharSequenceIntHashMap implements Closeable, Mutable {
 
     private void putAt0(int index, int offset, int hashCode) {
         final long ptr = address + ((long) index << 3);
-        Unsafe.getUnsafe().putInt(ptr, offset);
+        Unsafe.getUnsafe().putInt(ptr, offset + 1);
         Unsafe.getUnsafe().putInt(ptr + 4L, hashCode);
         if (--free == 0) {
             rehash();


RaphDal:
@coderabbit review

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 146 / 152 (96.05%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/AbstractOffsetCharSequenceHashSet.java|27|31|87.10%|
|:large_blue_circle:|io/questdb/cairo/wal/SymbolMap.java|94|96|97.92%|
|:large_blue_circle:|io/questdb/cairo/wal/WalEventWriter.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalWriter.java|22|22|100.00%|

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.Arrays;
+
+public class SymbolMap extends AbstractOffsetCharSequenceHashSet implements Closeable {
+    public static final int NO_ENTRY_VALUE = -1;
+    private final int noEntryValue;
+    private final DirectString sview = new DirectString();
+    private long address;
+    private long charsCapacity;
+    private int currentOffset = 0;
+    private int[] values;
```
puzpuzpuz: nit: we could go one step further and move `offsets` and `values` off-heap - that's to make the thing completely GC-free. Same story with hash codes - we could co-locate them with `offsets` like it's done in `OrderedMap` to avoid random memory accesses on key lookups.

RaphDal: Great advice, thanks!
Shouldn't we also co-locate `values` and the key length with `offsets` and `hashcode`?
Something like:
```
| hashcode | len | offset | value |
```
As we're also checking the key against the length before doing a char-by-char comparison we might avoid a few unnecessary random accesses.
And in the case the keys match, the value should be in the CPU cache already.

puzpuzpuz: Hash code comparison is already a good way to get rid of many false positive comparisons/loads. Having too many things put into the hash table (offsets) has a downside of the hash table growing in size. A more compact hash table is more likely to fit into one of the CPU caches. So, I'd either keep only offsets in the hash table and nothing else, or co-locate only the hash codes with the offsets.

puzpuzpuz: Having hash codes right in the hash table is also beneficial for rehashing: it can scan the hash table only, without accessing the data.

ideoma:
@coderabbitai review

core/src/test/java/io/questdb/test/cairo/wal/SymbolMapTest.java
```diff
@@ -0,0 +1,92 @@
...
+package io.questdb.test.cairo.wal;
+
+import io.questdb.cairo.wal.SymbolMap;
+import io.questdb.std.Chars;
+import io.questdb.std.ObjList;
+import io.questdb.std.Rnd;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class SymbolMapTest {
```
ideoma: nit: there is already `SymbolMapTest` in the code base, please use a different name

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
```diff
@@ -0,0 +1,213 @@
...
+import io.questdb.std.MemoryTag;
+import io.questdb.std.Numbers;
+import io.questdb.std.Unsafe;
+import io.questdb.std.str.DirectString;
+import org.jetbrains.annotations.NotNull;
+
+import java.io.Closeable;
+import java.util.Arrays;
+
+public class SymbolMap extends AbstractOffsetCharSequenceHashSet implements Closeable {
```
ideoma: nit: we already have `SymbolMapReader` and `SymbolMapWriter`, better to call the class `DirectCharSequenceIntHashMap`

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
```diff
@@ -0,0 +1,213 @@
...
+     * @param offset the last offset that was returned
+     * @return the offset of the next char sequence inserted the map
+     * or -1 if there are no more values.
+     */
+    public int nextOffset(int offset) {
+        final long lo = address + ((long) offset << 2);
+        final int len = Unsafe.getUnsafe().getInt(lo + 4);
+        final int off = offset + ((((len << 1) + 11) & ~3) >> 2);
+        if (off == this.currentOffset) {
+            return -1;
```
ideoma: This line is not covered by tests

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
```diff
@@ -0,0 +1,213 @@
...
+    @Override
+    protected boolean areKeysEquals(int offset, @NotNull CharSequence key, int keyHashCode) {
+        long lo = address + ((long) offset << 2);
+        final int hashCode = Unsafe.getUnsafe().getInt(lo);
+        if (hashCode != keyHashCode) {
+            return false;
+        }
+        final int len = Unsafe.getUnsafe().getInt(lo + 4);
+        if (len != key.length()) {
+            return false;
```
ideoma: This line is not covered by tests in `SymbolMapTest`

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
```diff
@@ -0,0 +1,213 @@
...
+        }
+        for (int i = 0; i < len; i++) {
+            if (Unsafe.getUnsafe().getChar(lo + 8 + ((long) i << 1)) != key.charAt(i)) {
+                return false;
```
ideoma: This line is not covered by tests in `SymbolMapTest`

core/src/main/java/io/questdb/cairo/wal/WalWriter.java
```diff
@@ -1594,7 +1594,11 @@ private void removeSymbolFiles(Path path, int rootLen, CharSequence columnName)
 
     private void removeSymbolMapReader(int index) {
         Misc.freeIfCloseable(symbolMapReaders.getAndSetQuick(index, null));
-        symbolMaps.setQuick(index, null);
+        SymbolMap map = symbolMaps.getQuick(index);
+        if (map != null) {
+            map.close();
+            symbolMaps.setQuick(index, null);
+        }
```
ideoma: nit:

```suggestion
        symbolMaps.set(index, Misc.free(symbolMaps.get(index)));
```

core/src/main/java/io/questdb/std/AbstractOffsetCharSequenceHashSet.java
```diff
@@ -0,0 +1,122 @@
...
+                return -index - 1;
+            }
+        } while (true);
+    }
+
+    abstract protected boolean areKeysEquals(int offset, @NotNull CharSequence key, int keyHashCode);
+
+    protected long keyAt(int index) {
+        return offsets[-index - 1];
+    }
```
ideoma: not used, can be deleted

core/src/main/java/io/questdb/std/AbstractOffsetCharSequenceHashSet.java
```diff
@@ -0,0 +1,122 @@
...
+        free = capacity;
+    }
+
+    public boolean contains(@NotNull CharSequence key) {
+        return keyIndex(key) < 0;
+    }
+
+    public boolean excludes(@NotNull CharSequence key) {
+        return keyIndex(key) > -1;
+    }
```
ideoma: not used

core/src/main/java/io/questdb/std/AbstractOffsetCharSequenceHashSet.java
```diff
@@ -0,0 +1,122 @@
...
+ *
+ ******************************************************************************/
+
+package io.questdb.std;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Arrays;
+
+public abstract class AbstractOffsetCharSequenceHashSet implements Mutable {
```
ideoma: nit: there is only one use of this abstract class; it can be merged with the SymbolMap class to be the same class to reduce abstractions.

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
```diff
@@ -0,0 +1,213 @@
...
+        }
+        final long lo = address + ((long) currentOffset << 2);
+        Unsafe.getUnsafe().putInt(lo, keyHashCode);
+        Unsafe.getUnsafe().putInt(lo + 4, key.length());
+        for (int i = 0; i < key.length(); i++) {
+            Unsafe.getUnsafe().putChar(lo + 8 + ((long) i << 1), key.charAt(i));
+        }
+
+        final int oldOffset = currentOffset;
+        currentOffset += requiredCapacity >> 2;
```
ideoma: Looks like `currentOffset` is always shifted on read, can we perhaps store it in bytes rather than 4 bytes?

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
```diff
@@ -0,0 +1,213 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.io.Closeable;
+import java.util.Arrays;
+
+public class SymbolMap extends AbstractOffsetCharSequenceHashSet implements Closeable {
+    public static final int NO_ENTRY_VALUE = -1;
+    private final int noEntryValue;
+    private final DirectString sview = new DirectString();
+    private long address;
```
ideoma: Please add a comment describing the record structure.

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
```diff
@@ -0,0 +1,213 @@
...
+
+    public SymbolMap(int initialCapacity, double loadFactor, int noEntryValue) {
+        this(initialCapacity, loadFactor, noEntryValue, 32);
+    }
+
+    public SymbolMap(int initialCapacity, double loadFactor, int noEntryValue, int avgKeySize) {
+        super(initialCapacity, loadFactor);
+        this.noEntryValue = noEntryValue;
+        this.charsCapacity = Numbers.ceilPow2(initialCapacity * (avgKeySize + 8));
+        this.address = Unsafe.malloc(this.charsCapacity, MemoryTag.NATIVE_DEFAULT);
```
ideoma: nit: `charsCapacity` is a confusing name because it seems to be bytes, rather than 2-byte chars. Simply `capacity` would be better.

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
```diff
@@ -0,0 +1,213 @@
...
+        }
+    }
+
+    public int get(@NotNull CharSequence key) {
+        return valueAt(keyIndex(key));
+    }
+
+    public CharSequence get(int offset) {
+        final long lo = address + ((long) offset << 2);
+        final int len = Unsafe.getUnsafe().getInt(lo + 4);
```
ideoma: Please add and assert that len does not point beyond the capacity.
Also, please assert that it's non-negative.

ideoma: Also, add tests with empty strings: ""

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
```diff
@@ -0,0 +1,213 @@
...
+        this.charsCapacity = Numbers.ceilPow2(initialCapacity * (avgKeySize + 8));
+        this.address = Unsafe.malloc(this.charsCapacity, MemoryTag.NATIVE_DEFAULT);
+        values = new int[offsets.length];
+    }
+
+    @Override
+    public final void clear() {
+        super.clear();
+        Arrays.fill(values, noEntryValue);
+        this.currentOffset = 0;
```
ideoma: It would be useful to release some portion of native memory here. If we can detect by reading the offset + size of the last key that we already overallocated >= 2x, we should shrink here.

core/src/main/java/io/questdb/cairo/wal/WalEventWriter.java
```diff
@@ -248,8 +248,8 @@ private void writeSymbolMapDiffs() {
                     eventMem.putInt(size);
 
                     int symbolCount = 0;
-                    for (int j = 0; j < size; j++) {
-                        final CharSequence symbol = symbolMap.keys().getQuick(j);
+                    for (int offset = symbolMap.nextOffset(); offset >= 0; offset = symbolMap.nextOffset(offset)) {
+                        final CharSequence symbol = symbolMap.get(offset);
```
ideoma: nit: ideally, we don't need CharSequence, we want `memcpy` to the `eventMem`. This can be

```
symbolMap.copyTo(offset, eventMem);
```

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
RaphDal: For this line to be covered, we shall have 2 `CharSequence` of the same length that generates the same hashcodes.

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
RaphDal: For this line to be covered, we shall have 2 `CharSequence` that generates the same hashcodes.

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
RaphDal: Good point, fixed in https://github.com/questdb/questdb/pull/6371/commits/cc7d9da4cef7a54ac9c18d2e0e648dab397fed57

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
RaphDal: Fair, I reworked the whole thing in https://github.com/questdb/questdb/pull/6371/commits/313e95448be0529ec0fe555709db2b4f11cc40b6

core/src/test/java/io/questdb/test/cairo/wal/SymbolMapTest.java
RaphDal: Fixed in https://github.com/questdb/questdb/pull/6371/commits/313e95448be0529ec0fe555709db2b4f11cc40b6

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
RaphDal: Fixed in https://github.com/questdb/questdb/pull/6371/commits/313e95448be0529ec0fe555709db2b4f11cc40b6

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
RaphDal: Fixed in https://github.com/questdb/questdb/pull/6371/commits/313e95448be0529ec0fe555709db2b4f11cc40b6

core/src/main/java/io/questdb/cairo/wal/WalWriter.java
RaphDal: Fixed in https://github.com/questdb/questdb/pull/6371/commits/313e95448be0529ec0fe555709db2b4f11cc40b6

core/src/main/java/io/questdb/std/AbstractOffsetCharSequenceHashSet.java
RaphDal: Good catch, fixed in https://github.com/questdb/questdb/pull/6371/commits/313e95448be0529ec0fe555709db2b4f11cc40b6

core/src/main/java/io/questdb/std/AbstractOffsetCharSequenceHashSet.java
RaphDal: Fixed in https://github.com/questdb/questdb/pull/6371/commits/313e95448be0529ec0fe555709db2b4f11cc40b6

core/src/main/java/io/questdb/std/AbstractOffsetCharSequenceHashSet.java
RaphDal: Fixed in https://github.com/questdb/questdb/pull/6371/commits/313e95448be0529ec0fe555709db2b4f11cc40b6

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
RaphDal: I'm not sure to understand what you mean there, the `currentOffset` is the address (4-byte aligned) where the next key can be written to.

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
RaphDal: Fixed in https://github.com/questdb/questdb/pull/6371/commits/313e95448be0529ec0fe555709db2b4f11cc40b6

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
RaphDal: Fixed in https://github.com/questdb/questdb/pull/6371/commits/313e95448be0529ec0fe555709db2b4f11cc40b6

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
RaphDal: Fixed in https://github.com/questdb/questdb/pull/6371/commits/313e95448be0529ec0fe555709db2b4f11cc40b6

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 163 / 165 (98.79%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/DirectCharSequenceIntHashMap.java|140|142|98.59%|
|:large_blue_circle:|io/questdb/cairo/wal/WalEventWriter.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalWriter.java|20|20|100.00%|

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
puzpuzpuz: The changes look good, thanks.

core/src/main/java/io/questdb/cairo/wal/DirectCharSequenceIntHashMap.java
```diff
@@ -0,0 +1,389 @@
...
+    public final void clear() {
+        // The key size estimation may be off, we can give some memory back if we don't even use half of it
+        if ((long) currentOffset << 3 < kvCapacity) {
+            final long oldCapacity = kvCapacity;
+            // We only shrink the capacity by 2 to avoid unnecessary grow-back later
+            kvCapacity >>= 1;
+            kvAddress = Unsafe.realloc(kvAddress, oldCapacity, kvCapacity, MemoryTag.NATIVE_DEFAULT);
+        }
+        free = mapCapacity;
+        Vect.memset(address, capacity, -1);
```
puzpuzpuz: nit: we could shift offset values by 1 to be able to zero memory instead of initializing it to -1 since the former is faster.

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
ideoma: Sorry, I do not see asserts

core/src/main/java/io/questdb/cairo/wal/WalEventWriter.java
RaphDal: Fixed in https://github.com/questdb/questdb/pull/6371/commits/1b3e72753a6e691c4305c100cd184809b17e89d1

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
puzpuzpuz: It's `testEmptyStringKey` I guess.

core/src/main/java/io/questdb/cairo/wal/DirectCharSequenceIntHashMap.java
RaphDal: Nice catch, fixed in https://github.com/questdb/questdb/pull/6371/commits/51c51f83aaaf21410926b54c842f8878f94ddd6a

core/src/main/java/io/questdb/cairo/wal/SymbolMap.java
ideoma: I mean, there should be asserts in this method, prod code

```
assert len >= 0 && len < capacity;
```

ideoma: anyway this method may not be used anymore in prod code

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 177 / 179 (98.88%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/wal/DirectCharSequenceIntHashMap.java|155|157|98.73%|
|:large_blue_circle:|io/questdb/cairo/wal/WalEventWriter.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalWriter.java|20|20|100.00%|

PR #6379, opened 2025-11-11
test(sql): reduce memory usage in large varchar test

kafka1991:
Reduces memory usage in `testReadWriteVarcharOver2GB` from ~2GB to ~1GB to prevent OOM failures on ARM64 CI machines.


bluestreak01:
hey Victor, the 2GB limit is there to test integer overflow, 1GB doesn't quite do it, we have to put up with 2GB unfortunately :(

kafka1991:
> hey Victor, the 2GB limit is there to test integer overflow, 1GB doesn't quite do it, we have to put up with 2GB unfortunately :(

Ok, let me close this then.

PR #6389, opened 2025-11-13
test(ilp): fix race condition in testSymbolCapacityReloadFuzz()

jerrinot:
Easy to reproduce spurious failure. It's sufficient to inject randomized pauses to Sender threads:
```java
for (int i = 0; i < N; i++) {
    sender1.table("mytable")
            .symbol("sym1", rnd1.nextString(10))
            .symbol("sym2", rnd1.nextString(2))
            .doubleColumn("dd", rnd1.nextDouble())
            .atNow();
    if (i % 1000 == 0) {
        Os.sleep(ThreadLocalRandom.current().nextInt(10));
    }
}
sender1.flush();
```
With this extra sleep the test fails consistently on my local.

Root cause: The test uses ILP@TCP - so sender.flush() only guarantees ILP lines are successfully written to TCP send buffer. There can be an arbitrary long pause between the flush and when the just flushed lines are observed by TCP receiver, let alone WAL apply job.

The fix acknowledges this delay and relaxes the assertion. It also runs the WalApply job even after the main WAL Apply thread is done, becasue the WAL Apply Thread might finish before it had a chance to observe all lines.

### DIFF
--- a/ci/github-release-pipeline.yml
+++ b/ci/github-release-pipeline.yml
@@ -20,12 +20,10 @@ stages:
               imageName: 'ubuntu-latest'
               poolName: "Azure Pipelines"
               os: Linux
-              jdk: "1.17"
             windows:
               imageName: "windows-latest"
               poolName: "Azure Pipelines"
               os: Windows
-              jdk: "1.17"
         pool:
           vmImage: $(imageName)
           name: $(poolName)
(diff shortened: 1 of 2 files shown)


jerrinot:
@coderabbitai review pls

PR #6398, opened 2025-11-14
update tolerance and sampling interval parsing to use LONG_NULL for better handling of large values

### LINKED ISSUE #6202: Allow SAMPLE BY to use a long, rather than an INT for the units

shivamAmrutia:
This PR removes the int limitation on time interval parsing for SAMPLE BY and ASOF JOIN TOLERANCE, so very large nanosecond intervals (e.g. 3000000000n) can be used without overflow errors.

### Problem

Previously, TimestampSamplerFactory.parseInterval parsed the numeric part of an interval as int and used an int maxValue.

For nanosecond units, values above 2_147_483_647 caused "value too high for given units" errors, even though they are valid within the long timestamp range.

### Solution

Change parseInterval to parse the numeric value as long and accept a long maxValue, using Numbers.parseLong and Numbers.LONG_NULL.

Update all call sites (SqlCodeGenerator, CreateMatViewOperationImpl) to pass long-based maxValues and remove artificial int caps for nanosecond intervals, relying on the timestamp driver for range validation.

### Testing
Updated TimestampSamplerFactoryTest to use Numbers.LONG_NULL and added a dedicated testParseLargeNanoInterval to cover values above Integer.MAX_VALUE.

Extended SampleByNanoTimestampTest with testLargeNanoIntervalSampleBy to exercise SAMPLE BY 3000000000n end-to-end and ensure no SqlException is thrown for large nanosecond intervals.

### Issue Links
relates #6202 
fixes #6202

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -6,6 +6,7 @@ reviews:
   request_changes_workflow: false
   high_level_summary: false
   poem: false
+  sequence_diagrams: false
   review_status: true
   collapse_walkthrough: false
   auto_review:
--- a/core/src/main/java/io/questdb/griffin/SqlParser.java
+++ b/core/src/main/java/io/questdb/griffin/SqlParser.java
@@ -3877,7 +3877,11 @@ public class SqlParser {
      */
     private void rewriteJsonExtractCast(ExpressionNode node) {
         if (node.type == ExpressionNode.FUNCTION && isCastKeyword(node.token)) {
-            if (node.lhs != null && node.lhs.paramCount == 2 && isJsonExtract(node.lhs.token)) {
+            if (node.lhs != null
+                    && node.lhs.type == ExpressionNode.FUNCTION
+                    && node.lhs.paramCount == 2
+                    && node.lhs.token != null
+                    && isJsonExtract(node.lhs.token)) {
                 // rewrite cast such as
                 // json_extract(json,path)::type -> json_extract(json,path,type)
                 // the ::type is already rewritten as
@@ -3900,7 +3904,7 @@ public class SqlParser {
                         node.rhs = jsonExtractNode.rhs;
                         node.args.clear();
                     } else if (JsonExtractTypedFunctionFactory.isIntrusivelyOptimized(castType)) {
-                        int type = ColumnType.typeOf(typeNode.token);
+                        int type = castType;
                         node.token = jsonExtractNode.token;
                         node.paramCount = 3;
                         node.type = jsonExtractNode.type;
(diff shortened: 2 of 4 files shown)


ideoma:
@coderabbitai review

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -878,7 +878,7 @@ private static long tolerance(QueryModel slaveModel, int leftTimestamp, int righ
             long multiplier;
             switch (unit) {
                 case 'n':
-                    toleranceInterval = TimestampSamplerFactory.parseInterval(tolerance.token, k, tolerance.position, "tolerance", Integer.MAX_VALUE, unit);
+                    toleranceInterval = TimestampSamplerFactory.parseInterval(tolerance.token, k, tolerance.position, "tolerance", Numbers.LONG_NULL, unit);
```
kafka1991, line 881: Please add tests for `asofJoin tolerance` that covers cases exceeding `Integer.MAX_VALUE`. 
You can refer to https://github.com/questdb/questdb/blob/e88abf70e212c0673f988bc3521331fc34352a69/core/src/test/java/io/questdb/test/griffin/engine/join/AsOfJoinTest.java#L1220.

core/src/test/java/io/questdb/test/griffin/engine/groupby/SampleByNanoTimestampTest.java
```diff
@@ -127,6 +127,25 @@ public void testBadInterval() throws Exception {
...
 
+    @Test
+    public void testLargeNanoIntervalSampleBy() throws Exception {
+        assertMemoryLeak(() -> {
+            execute("create table large_ns as (" +
+                    "select timestamp_sequence_ns(0, 1000000000L) ts " +
+                    "from long_sequence(5)" +
+                    ") timestamp(ts) partition by NONE", sqlExecutionContext);
+
+            try (SqlCompiler compiler = engine.getSqlCompiler()) {
```
kafka1991: It's better to use `assertQuery` here

Issue #6123, opened 2025-09-10
Bound parameters in batches is broken


rmja:
### To reproduce

Consider this simple C# program:

```C#
using Npgsql;

await using var dataSource = new NpgsqlDataSourceBuilder(
    "Host=localhost;Port=8812;Username=admin;Password=quest;Database=qdb;ServerCompatibilityMode=NoTypeLoading;"
).Build();

const int BatchSize = 200;

await using var connection = await dataSource.OpenConnectionAsync();

await using var batch = new NpgsqlBatch(connection);
for (var i = 0; i < BatchSize; i++)
{
    var cmd = new NpgsqlBatchCommand($"SELECT {i} AS UnboundIndex, @index AS BoundIndex");
    cmd.Parameters.AddWithValue("index", i);
    batch.BatchCommands.Add(cmd);
}

await using var reader = await batch.ExecuteReaderAsync();

var first = true;
for (var i = 0; i < batch.BatchCommands.Count; i++)
{
    if (!first)
    {
        await reader.NextResultAsync();
    }
    first = false;

    if (!await reader.ReadAsync())
    {
        throw new InvalidOperationException("No rows returned");
    }

    var unboundIndex = reader.GetInt32(0);
    var boundIndex = reader.GetInt32(1);
    if (unboundIndex != boundIndex)
    {
        throw new InvalidOperationException(
            $"Invalid index: {boundIndex}, expected: {unboundIndex}"
        );
    }
}

Console.WriteLine("Test Completed!");

```

The `boundIndex` and `unboundIndex` are not equal, and the exception is thrown. The `boundIndex=199` when `unboundIndex=0`.
Note that filtering etc. using the bound values works fine. The problem seems to only occurs when the bound parameter is returned.

### QuestDB version:

9.0.3

### OS, in case of Docker specify Docker and the Host OS:

Windows

### File System, in case of Docker specify Host File System:

ntfs

### Full Name:

Rasmus Melchior Jacobsen

### Affiliation:

Utiliread

### Have you followed Linux, MacOs kernel configuration steps to increase Maximum open files and Maximum virtual memory areas limit?

- [x] Yes, I have

### Additional context

I have inspected the network traffic using wireshark. The capture is attached below.
It is clear from the trace that the parameters are sent correctly to questdb, but that questdb returns the incorrect result rows.

[bind problem.zip](https://github.com/user-attachments/files/22259773/bind.problem.zip)

Sheenam015:
[...]

PR #6401, opened 2025-11-16
fix(pgwire): batched queries would use incorrect bind variables

nwoolmer:
WIP 

Fixes https://github.com/questdb/questdb/issues/6123

A single BindVariableService was shared between all PGPipelineEntry instances managed by a PGConnectionContext.

When running batched queries, bind-execute messages would be sent in sequence. These would re-bind the BindVariableService entries and then compile each subsequent query.

Unfortunately, only after a sync message would the results be returned. The cursors all share the same BindVariableService, so only the values of the last query are applied.

The proposed fix in this PR is to dedicate an execution context and bind variable service to each entry, so it is isolated from the other queries in the batch.

To run the test:

image

Update: the fix was PoCed but it needs proper attention, there will be a superceding PR.


nwoolmer:
@coderabbitai review

Issue #6403, opened 2025-11-17
optimiser fails with unary minus on column targeted for aggregation


nwoolmer:
### To reproduce

```sql
SELECT symbol, sum(-amount) FROM trades;
```

```
2025-11-17T10:37:01.615962Z C i.q.c.h.p.JsonQueryProcessorState [1114544014041] internal error [ex=
java.lang.NullPointerException: Cannot read field "type" because "n" is null
	at io.questdb.griffin.SqlOptimiser.isIntegerConstant(SqlOptimiser.java:2593)
	at io.questdb.griffin.SqlOptimiser.rewriteAggregate(SqlOptimiser.java:4264)
	at io.questdb.griffin.SqlOptimiser.rewriteSelectClause0(SqlOptimiser.java:6010)
	at io.questdb.griffin.SqlOptimiser.rewriteSelectClause(SqlOptimiser.java:5925)
	at io.questdb.griffin.SqlOptimiser.optimise(SqlOptimiser.java:7238)
	at io.questdb.griffin.SqlCompilerImpl.compileExecutionModel0(SqlCompilerImpl.java:2533)
	at io.questdb.griffin.SqlCompilerImpl.compileExecutionModel(SqlCompilerImpl.java:2521)
	at io.questdb.griffin.SqlCompilerImpl.compileUsingModel(SqlCompilerImpl.java:3236)
	at io.questdb.griffin.SqlCompilerImpl.compileInner(SqlCompilerImpl.java:2623)
	at io.questdb.griffin.SqlCompilerImpl.compile(SqlCompilerImpl.java:394)
	at io.questdb.cairo.pool.SqlCompilerPool$C.compile(SqlCompilerPool.java:141)
	at io.questdb.cutlass.http.processors.JsonQueryProcessor.compileAndExecuteQuery(JsonQueryProcessor.java:514)
	at io.questdb.cutlass.http.processors.JsonQueryProcessor.execute0(JsonQueryProcessor.java:216)
	at io.questdb.cutlass.http.processors.JsonQueryProcessor.onRequestComplete(JsonQueryProcessor.java:298)
	at io.questdb.cutlass.http.HttpConnectionContext.handleClientRecv(HttpConnectionContext.java:986)
	at io.questdb.cutlass.http.HttpConnectionContext.handleClientOperation(HttpConnectionContext.java:323)
	at io.questdb.cutlass.http.HttpServer.handleClientOperation(HttpServer.java:331)
[...]

PR #6404, opened 2025-11-17
fix(sql): cannot compile query with aggregate function containing unary minus

nwoolmer:
Fixes https://github.com/questdb/questdb/issues/6403

The optimiser code assumed binary `-`/`+`. In the unary case, it did not nullcheck. Also added a check to ensure it is a binary op.

### DIFF
--- a/core/src/test/java/io/questdb/test/griffin/SqlOptimiserTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/SqlOptimiserTest.java
@@ -5046,13 +5046,15 @@ public class SqlOptimiserTest extends AbstractSqlParserTest {
                     execute(tradesDdl);
                     assertPlanNoLeakCheck(
                             "SELECT symbol, sum(-amount) FROM trades;",
-                            "Async Group By workers: 1\n" +
-                                    "  keys: [symbol]\n" +
-                                    "  values: [sum(-amount)]\n" +
-                                    "  filter: null\n" +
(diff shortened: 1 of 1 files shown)


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|

core/src/test/java/io/questdb/test/griffin/SqlOptimiserTest.java
```diff
@@ -5040,6 +5040,26 @@ public void testSingleCountDistinct() throws Exception {
...
+    @Test
+    public void testUnaryMinusInAggregateFunction() throws Exception {
+        assertMemoryLeak(() -> {
+                    execute(tradesDdl);
+                    assertPlanNoLeakCheck(
+                            "SELECT symbol, sum(-amount) FROM trades;",
+                            """
+                                    Async Group By workers: 1
+                                      keys: [symbol]
+                                      values: [sum(-amount)]
```
ideoma, line 5052: nit: How easy to do it `-sum(amount)`?

nwoolmer: It turns out it doesn't work in general:

```
Async Group By workers: 46
  values: [sum(amount*10)]
  filter: null
    PageFrame
        Row forward scan
        Frame forward scan on: trades
```

We will rewrite the rewrite in a subsequent PR

Issue #6405, opened 2025-11-18
ORDER BY clause is wrongly ignored


puzpuzpuz:
### To reproduce

```sql
create table trades (  sym symbol,  price double,  ts TIMESTAMP) timestamp(ts) partition by day;
create table prices (  sym symbol,  price double,  ts TIMESTAMP) timestamp(ts) partition by day;

insert into trades values ('AAPL', 100.0, cast('2023-01-01T09:00:00.000000Z' as TIMESTAMP)),('AAPL', 101.0, cast('2023-01-01T09:01:00.000000Z' as TIMESTAMP)),('AAPL', 102.0, cast('2023-01-01T09:02:00.000000Z' as TIMESTAMP)),('MSFT', 200.0, cast('2023-01-01T09:03:00.000000Z' as TIMESTAMP)),('MSFT', 201.0, cast('2023-01-01T09:04:00.000000Z' as TIMESTAMP)),('GOOGL', 300.0, cast('2023-01-01T09:05:00.000000Z' as TIMESTAMP)),('GOOGL', 301.0, cast('2023-01-01T09:06:00.000000Z' as TIMESTAMP)),('AAPL', 103.0, cast('2023-01-01T09:07:00.000000Z' as TIMESTAMP)),('MSFT', 202.0, cast('2023-01-01T09:08:00.000000Z' as TIMESTAMP)),('GOOGL', 302.0, cast('2023-01-01T09:09:00.000000Z' as TIMESTAMP));
[...]

PR #6414, opened 2025-11-19
fix(sql): order-by ignored in sub-queries of aggregation

RaphDal:
Fixes https://github.com/questdb/questdb/issues/6405

This PR fixes a bug from the `optimizeOrderBy` pass that was removing order-by in sub-queries of aggregations.

Functions like `first`/`last`/`first_not_null`/`last_not_null` needs to receive the rows in the expected order to work properly.

## Example

```sql
select sym, ts, first(b)
from (
select x.sym, x.ts, y.b
from x
left join y on x.sym = y.sym
order by x.ts, y.ts -- optimizeOrderBy was removing this clause
);
```

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -6,6 +6,7 @@ reviews:
   request_changes_workflow: false
   high_level_summary: false
   poem: false
+  sequence_diagrams: false
   review_status: true
   collapse_walkthrough: false
   auto_review:
--- a/core/src/main/java/io/questdb/griffin/SqlParser.java
+++ b/core/src/main/java/io/questdb/griffin/SqlParser.java
@@ -3877,7 +3877,11 @@ public class SqlParser {
      */
     private void rewriteJsonExtractCast(ExpressionNode node) {
         if (node.type == ExpressionNode.FUNCTION && isCastKeyword(node.token)) {
-            if (node.lhs != null && node.lhs.paramCount == 2 && isJsonExtract(node.lhs.token)) {
+            if (node.lhs != null
+                    && node.lhs.type == ExpressionNode.FUNCTION
+                    && node.lhs.paramCount == 2
+                    && node.lhs.token != null
+                    && isJsonExtract(node.lhs.token)) {
                 // rewrite cast such as
                 // json_extract(json,path)::type -> json_extract(json,path,type)
                 // the ::type is already rewritten as
@@ -3900,7 +3904,7 @@ public class SqlParser {
                         node.rhs = jsonExtractNode.rhs;
                         node.args.clear();
                     } else if (JsonExtractTypedFunctionFactory.isIntrusivelyOptimized(castType)) {
-                        int type = ColumnType.typeOf(typeNode.token);
+                        int type = castType;
                         node.token = jsonExtractNode.token;
                         node.paramCount = 3;
                         node.type = jsonExtractNode.type;
--- a/core/src/test/java/io/questdb/test/griffin/SqlOptimiserTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/SqlOptimiserTest.java
@@ -4924,6 +4924,36 @@ public class SqlOptimiserTest extends AbstractSqlParserTest {
     }
 
     @Test
+    public void testScalarSubqueryDoesNotBreakJsonExtract() throws Exception {
+        assertMemoryLeak(() -> {
+            execute(tradesDdl);
+            assertPlanNoLeakCheck("""
+                            SELECT * FROM trades
+                            WHERE timestamp = (SELECT max(timestamp) FROM "trades")
+                            ORDER BY timestamp DESC;""",
+                    """
+                            PageFrame
+                                Row backward scan
+                                Interval backward scan on: trades
+                                  intervals: []
+                            """);
+
+            assertException("""
+                            SELECT * FROM trades
+                            WHERE timestamp = CAST((SELECT max(timstamp) FROM "trades") AS LONG)
+                            ORDER BY timestamp DESC""",
+                    56,
+                    "Invalid column");
+
+            assertException("""
+                            SELECT * FROM trades
+                            WHERE timestamp = CAST((SELECT max(timestamp) FROM "trades") AS LONG)
+                            ORDER BY timestamp DESC;""",
+                    39, "there is no matching function");
+        });
+    }
+
+    @Test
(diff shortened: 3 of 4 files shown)


RaphDal:
@coderabbit review

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -3362,6 +3363,11 @@ private void optimiseOrderBy(QueryModel model, int topLevelOrderByMnemonic) {
                 break;
         }
 
+        // some aggregate functions (e.g. first()) relies on the rows being properly ordered to return the correct result
```
kafka1991: Since most aggregate functions don’t require ordering, can we make this more fine-grained for better performance?

RaphDal: It depends, I see 3 cases:
1. `first`/`last`/`first_not_null`/`last_not_null`: required to have the proper ordering
2. `*sum`/`avg`/`variance`: having the wrong ordering may bring inconsistent results with lossy types (double/float)
3. `count`/`min`/`max`: they should be order independant I guess

The point `2` may be too strict for our needs and we might get away with loosing the ordering here, wdyt?

RaphDal: Added the function check in https://github.com/questdb/questdb/pull/6414/commits/fb0f0a74edcc84037515641de9fbf738bbe785a5

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -111,6 +111,12 @@ public class SqlOptimiser implements Mutable {
     private static final int NOT_OP_NOT = 1;
     private static final int NOT_OP_NOT_EQ = 9;
     private static final int NOT_OP_OR = 3;
+    private final static CharSequence[] ORDERED_GROUP_BY_FUNCTIONS = {
```
puzpuzpuz: Why not `LowerCaseCharSequenceHashSet`?

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -268,7 +274,7 @@ public FunctionFactoryCache getFunctionFactoryCache() {
         return functionParser.getFunctionFactoryCache();
     }
 
-    public boolean hasAggregates(ExpressionNode node) {
```
puzpuzpuz, line 271: I'd keep the method as is and introduce a new one for the check: `hasOrderedAggregates`. The new method would include this check:
```java
if (functionParser.getFunctionFactoryCache().isGroupBy(node.token) && orderedGroupByFunctions.contains(node.token)) {
    return true;
}
```

This way we don't check the first met group by function, but all of them. Note: currently, it's unlikely that the group by model contains expressions involving group by functions, but that may change one day, so it's better to have the code slightly more robust.

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -3362,6 +3369,23 @@ private void optimiseOrderBy(QueryModel model, int topLevelOrderByMnemonic) {
                 break;
         }
 
+        // some aggregate functions (e.g. first()) relies on the rows being properly ordered to return the correct result
```
puzpuzpuz: nit:
```suggestion
        // some aggregate functions (e.g. first()) rely on the rows being properly ordered to return the correct result
```

core/src/test/java/io/questdb/test/griffin/ExplainPlanTest.java
```diff
@@ -12947,13 +12947,13 @@ public void testWindowOrderByUnderWindowModelIsPreserved() throws Exception {
                               keys: [sum desc]
                                 GroupBy vectorized: false
                                   values: [sum(avg),sum(sum),first(first_value)]
-                                    CachedWindow
```
puzpuzpuz, line 12950: nit: would be great to add one more test case that involves a non-ordered aggregate function, so that we still verify that `CachedWindow` kicks in in that case.

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
RaphDal: As there are only 4 elements and most other aggregate functions have different lengths, I expected the overhead of `LowerCaseCharSequenceHashSet` to be unnecessary.
Though, I don't have any benchmarks to back this intuition.

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
RaphDal: Right, fixed in https://github.com/questdb/questdb/pull/6414/commits/8b4cbac67c11a902af910acb03e0bdd214c090ed

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
RaphDal: Fixed in https://github.com/questdb/questdb/pull/6414/commits/8b4cbac67c11a902af910acb03e0bdd214c090ed

core/src/test/java/io/questdb/test/griffin/ExplainPlanTest.java
RaphDal: Added in https://github.com/questdb/questdb/pull/6414/commits/87eea26c0a9e7cf5eaf131a0de79b724495bace4

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
puzpuzpuz: This way, we'll simply need less code, e.g. we won't need the equality check loop and any new methods in `Chars`. I don't think there will be any perf difference between the two.

RaphDal: Refactored in https://github.com/questdb/questdb/pull/6414/commits/7caadd8afab511c873f25834548c7bc09cbb4d62

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 37 / 37 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/Chars.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|27|27|100.00%|

Issue #6434, opened 2025-11-23
Flaky CopyExportTest.testCopyWithSameSql test


puzpuzpuz:
### To reproduce

linux-x86-graal (#6316 - unrelated change):
```
2025-11-23T10:09:24.6872824Z 2025-11-23T10:09:24.685132Z I i.q.t.TestListener >>>> io.questdb.test.griffin.CopyExportTest.testCopyWithSameSql
2025-11-23T10:09:24.6873736Z 2025-11-23T10:09:24.685670Z I i.q.t.AbstractTest Starting test CopyExportTest#testCopyWithSameSql
2025-11-23T10:09:24.6874496Z 2025-11-23T10:09:24.685973Z I i.q.c.TableNameRegistryStore reloading tables file [path=/tmp/junit2969108904759880735/dbRoot/tables.d.0, threadId=21579]
2025-11-23T10:09:24.6891641Z 2025-11-23T10:09:24.686856Z I i.q.c.t.t.InputFormatConfiguration loading input format config [resource=/text_loader.json]
2025-11-23T10:09:24.6906132Z >>>>= io.questdb.test.griffin.CopyExportTest.testCopyWithSameSql
2025-11-23T10:09:24.7013377Z 2025-11-23T10:09:24.700860Z I i.q.g.e.QueryProgress exe [id=199, sql=`create table test_table (x int, y string)`, principal=admin, cache=false, jit=false]
2025-11-23T10:09:24.7019457Z 2025-11-23T10:09:24.700937Z I i.q.c.CairoEngine locked [table=test_table~, thread=21579]
2025-11-23T10:09:24.7020497Z 2025-11-23T10:09:24.701255Z I i.q.c.CairoEngine unlocked [table=test_table~]
2025-11-23T10:09:24.7021308Z 2025-11-23T10:09:24.701324Z I i.q.g.e.QueryProgress fin [id=199, sql=`create table test_table (x int, y string)`, principal=admin, cache=false, jit=true, time=466000]
2025-11-23T10:09:24.7023459Z 2025-11-23T10:09:24.701405Z I i.q.g.e.QueryProgress exe [id=-1, sql=`insert into test_table values (1, 'test')`, principal=admin, cache=false, jit=false]
2025-11-23T10:09:24.7024288Z 2025-11-23T10:09:24.701836Z I i.q.g.e.QueryProgress fin [id=-1, sql=`insert into test_table values (1, 'test')`, principal=admin, cache=false, jit=true, time=484000]
[...]

PR #6435, opened 2025-11-23
test(sql): flaky export tests where copy finished or did not start

nwoolmer:
Fixes https://github.com/questdb/questdb/issues/6434

`testCopyWithSameOutput` and `testCopyWithSameSql` both rely on there being an active export before the rest of the test executes. On CI, the export could already be finished, or not even started yet. This PR adds a little more data, and also a wait (5 second timeout) for an export to be active.


core/src/test/java/io/questdb/test/griffin/CopyExportTest.java
```diff
@@ -1566,10 +1566,12 @@ public void testCopyWithSameDirsParallel() throws Exception {
     public void testCopyWithSameOutput() throws Exception {
         assertMemoryLeak(() -> {
             execute("create table test_table (x int, y string)");
-            execute("insert into test_table values (1, 'test')");
+            execute("insert into test_table select x, x::string as y FROM long_sequence(100_000)");
 
             CopyExportRunnable stmt = () -> {
                 runAndFetchCopyExportID("copy test_table to 'output8' with format parquet statistics_enabled true", sqlExecutionContext);
+                // Wait for the first export to be active before attempting the second
+                waitForActiveExport();
```
kafka1991: nit: Actually we don’t need to wait until the export task run — the `exportId` has already been registered to the `CopyExportContext` in `CopyExportFactory::getCursor(sqlExecutionContext)`,（which call is sync in `runAndFetchCopyExportID `） https://github.com/questdb/questdb/blob/e8316334a918dd7ac2a8e7d894b1d1a972210334/core/src/main/java/io/questdb/griffin/engine/ops/CopyExportFactory.java#L98

nwoolmer: Don't worry about this, we are a fixing it a different way!

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 6 / 8 (75.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/parquet/CopyExportRequestJob.java|6|8|75.00%|

Issue #6436, opened 2025-11-23
Unconsistent missing rows when using SAMPLE BY + FILL in subquery


agpenas:
### To reproduce

Whenever I try to manipulate an intermediate table with filled rows, the ones that have been filled seem to disappear. 
In order to easily reproduce the behavior, you can create a dummy table:
```SQL
CREATE TABLE weather (
    ts TIMESTAMP,
    temperature DOUBLE,
    humidity INTEGER
) TIMESTAMP(ts);

INSERT INTO weather (ts, temperature, humidity) VALUES
    ('2024-01-01T00:00:00Z', 10.0, 10),
    ('2024-01-01T00:03:00Z', 11.5, 11),
    ('2024-01-01T00:04:00Z', 12.0, 12),
    ('2024-01-01T00:07:00Z', 13.0, 13);
```
Imputing the missing rows with this query will give the expected behavior:
```SQL
SELECT
        ts,
        last(temperature) AS temperature,
        last(humidity) As humidity
    FROM weather
    SAMPLE BY 1m FILL(prev)
```

<img width="300" height="180" alt="Image" src="https://github.com/user-attachments/assets/081cacd1-88ac-4f4b-a84b-e2b9e0a31114" />

While this nested query will not:
```SQL
WITH imputed AS (
    SELECT
        ts,
        last(temperature) AS temperature,
        last(humidity) As humidity
    FROM weather
    SAMPLE BY 1m FILL(prev)
)
SELECT ts, humidity
FROM imputed
```

<img width="250" height="140" alt="Image" src="https://github.com/user-attachments/assets/bb73e0d3-942c-4e4b-baf5-8b6cc1d15af4" />

To make things even more interesting, replacing `humidity` by `temperature` will work:
```SQL
WITH imputed AS (
    SELECT
        ts,
        last(temperature) AS temperature,
        last(humidity) As humidity
    FROM weather
    SAMPLE BY 1m FILL(prev)
)
SELECT ts, temperature
FROM imputed
```

<img width="300" height="180" alt="Image" src="https://github.com/user-attachments/assets/a4118bac-8956-43cf-927a-5b22cede3c4b" />

### QuestDB version:

9.1.0

### OS, in case of Docker specify Docker and the Host OS:

Ubuntu 24.04 (Docker)

### File System, in case of Docker specify Host File System:

ext4

### Full Name:

Abel Garcia Penas

### Affiliation:

N/A

[...]

PR #6437, opened 2025-11-24
fix(sql): fix sample by with only one fill option

kafka1991:
When a SAMPLE BY query specifies only one fill option, current implemation implies that this option should apply to all aggregation columns.  This PR ensures that when we need to reorganize the fill values based on the positional relationship between `TopDownColumns` and `BottomUpColumns`, the logic remains consistent.

close https://github.com/questdb/questdb/issues/6436


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|

PR #6439, opened 2025-11-24
fix(sql):  fix bug where queries using `IN` clauses with INT/LONG columns containing `-1` return incorrect results.

kafka1991:
Fixes a bug where queries using `IN` clauses with INT/LONG columns containing `-1` could return incorrect results.

 **Example**
 ```sql
  SELECT * FROM table WHERE int_column IN (-1, 1);
```
Root cause: The internal HashSet implementation used -1 as a special marker value, which conflicted with actual -1 values in the data, causing incorrect query results.

close https://github.com/questdb/questdb/issues/6438

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -3071,6 +3071,19 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
         truncate(true);
     }
 
+    public boolean trySkipWalTransactions(long seqTxn, long skipTxnCount) {
+        assert skipTxnCount > 0;
+        if (txWriter.getLagRowCount() == 0 && txWriter.getLagTxnCount() == 0) {
+            LOG.info().$("skipping replaced WAL transactions [table=").$(tableToken)
+                    .$("range=[").$(seqTxn)
+                    .$(", ").$(seqTxn + skipTxnCount).$(')')
+                    .I$();
+            commitSeqTxn(seqTxn + skipTxnCount - 1);
+            return true;
+        }
+        return false;
+    }
+
     public void updateTableToken(TableToken tableToken) {
         this.tableToken = tableToken;
         this.metadata.updateTableToken(tableToken);
--- a/core/src/main/java/io/questdb/cairo/wal/ApplyWal2TableJob.java
+++ b/core/src/main/java/io/questdb/cairo/wal/ApplyWal2TableJob.java
@@ -75,8 +75,8 @@ import static io.questdb.cairo.ErrorTag.OUT_OF_MEMORY;
 import static io.questdb.cairo.ErrorTag.resolveTag;
 import static io.questdb.cairo.TableUtils.TABLE_EXISTS;
 import static io.questdb.cairo.pool.AbstractMultiTenantPool.NO_LOCK_REASON;
-import static io.questdb.cairo.wal.WalTxnType.MAT_VIEW_INVALIDATE;
 import static io.questdb.cairo.wal.WalTxnType.*;
+import static io.questdb.cairo.wal.WalTxnType.MAT_VIEW_INVALIDATE;
 import static io.questdb.cairo.wal.WalUtils.*;
 import static io.questdb.tasks.TableWriterTask.CMD_ALTER_TABLE;
 import static io.questdb.tasks.TableWriterTask.CMD_UPDATE_TABLE;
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: core/src/main/java/io/questdb/griffin/engine/functions/bool/InLongFunctionFactory.java [hunk 1]
```diff
@@ -115,7 +115,7 @@ public Function newInstance(
                                     2)
                     );
                 default:
-                    LongHashSet inVals = new LongHashSet((int) ((argCount / LongHashSet.DEFAULT_LOAD_FACTOR)));
+                    LongHashSet inVals = new LongHashSet((int) (argCount / LongHashSet.DEFAULT_LOAD_FACTOR), LongHashSet.DEFAULT_LOAD_FACTOR, Long.MAX_VALUE);
```
puzpuzpuz: What if someone wants to do `IN` on `Long.MAX_VALUE`? We seem to be exchanging one valid non-null long value (`-1`) for another one (`Long.MAX_VALUE`).

kafka1991: Yeah, but in my opinion `Long.MAX_VALUE` is less common than `-1`. Otherwise, we would have to introduce a bitset to mark whether an entry in the hashset is actually occupied, or each record function call would need to pay the performance cost of checking whether it is a sentinel value.

so In my view, considering rare of `Long.MAX_VALUE`, the risk-to-reward ratio is quite favorable.

WDYT?

puzpuzpuz: We could write a new off-heap (or on-heap, doesn't matter for this use case) hash set similar to `Unordered8Map` - that hash table has no sentinel values and is able to store any `long` value.

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/bool/InLongFunctionFactory.java|2|2|100.00%|

core/src/main/java/io/questdb/griffin/engine/functions/bool/InLongFunctionFactory.java
kafka1991: introduced `DirectLongHashSet` which has a flag mark zero value.

kafka1991:
@coderabbitai review

bluestreak01:
hey @kafka1991 could you update PR description to make it use-facing. "fix" will be included in public release notes

UtkarshDashora:
✅ Passed checks (3 passed)

kafka1991:
> hey @kafka1991 could you update PR description to make it use-facing. "fix" will be included in public release notes

fixed

core/src/main/java/io/questdb/griffin/engine/functions/bool/InLongFunctionFactory.java
```diff
@@ -115,9 +117,14 @@ public Function newInstance(
                                     2)
                     );
                 default:
-                    LongHashSet inVals = new LongHashSet((int) ((argCount / LongHashSet.DEFAULT_LOAD_FACTOR)));
-                    parseToLong(args, argPositions, inVals);
-                    return new InLongConstFunction(args.getQuick(0), inVals);
+                    DirectLongHashSet inVals = new DirectLongHashSet(argCount, DirectLongHashSet.DEFAULT_LOAD_FACTOR, MemoryTag.NATIVE_FUNC_RSS);
```
puzpuzpuz: nit: instead of providing `DirectLongHashSet.DEFAULT_LOAD_FACTOR` value we could add one more constructor to the `DirectLongHashSet` class: `DirectLongHashSet inVals = new DirectLongHashSet(argCount, MemoryTag.NATIVE_FUNC_RSS);`

core/src/main/java/io/questdb/std/DirectLongHashSet.java
```diff
@@ -0,0 +1,286 @@
...
+                return (int) index;
+            } else if (k == key) {
+                return (int) -(index + 1);
+            }
+            index = (index + 1) & mask;
+            addr = memStart + (index * Long.BYTES);
+        }
+    }
+
+    public int remove(long key) {
```
puzpuzpuz: Why do we need the remove method?

core/src/test/java/io/questdb/test/std/DirectLongHashSetTest.java
```diff
@@ -0,0 +1,269 @@
...
+
+package io.questdb.test.std;
+
+import io.questdb.std.DirectLongHashSet;
+import io.questdb.std.Rnd;
+import io.questdb.std.str.StringSink;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class DirectLongHashSetTest {
```
puzpuzpuz, line 39: Could you also add a fuzz test similar to `GroupByLongHashSetFuzzTest#testFuzz()`?

core/src/main/java/io/questdb/griffin/engine/functions/bool/InLongFunctionFactory.java
kafka1991: fixed in https://github.com/questdb/questdb/pull/6439/commits/a85522e991140ba780d8126df211eff33bae6a37

core/src/main/java/io/questdb/std/DirectLongHashSet.java
kafka1991: It’s currently unused, but since it’s part of the std package, it should provide an interface similar to `LongHashSet`(IMHO).

core/src/test/java/io/questdb/test/std/DirectLongHashSetTest.java
kafka1991: test added in https://github.com/questdb/questdb/pull/6439/commits/a85522e991140ba780d8126df211eff33bae6a37

core/src/main/java/io/questdb/std/DirectLongHashSet.java
puzpuzpuz: It's not a must. We have some collections with similar names that have different APIs. It's a matter of describing the use case in the class javadoc.

core/src/test/java/io/questdb/test/std/DirectLongHashSetTest.java
```diff
@@ -0,0 +1,334 @@
...
+            set.add(100);
+            Assert.assertFalse(set.excludes(100));
+            Assert.assertTrue(set.excludes(200));
+        }
+    }
+
+    @Test
+    public void testFuzz() {
+        final int N = 100000;
+        final Rnd rnd = new Rnd();
```
puzpuzpuz: Could you use `TestUtils#generateRandom()` here, so that each uses different PRNG seeds? Currently, the test is completely deterministic.
```suggestion
        final Rnd rnd = TestUtils.generateRandom(LOG);
```

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 147 / 159 (92.45%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/bool/InLongFunctionFactory.java|11|14|78.57%|
|:large_blue_circle:|io/questdb/std/DirectLongHashSet.java|136|145|93.79%|

core/src/test/java/io/questdb/test/std/DirectLongHashSetTest.java
kafka1991: test fixed

core/src/main/java/io/questdb/std/DirectLongHashSet.java
kafka1991: remove `remove` method

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 115 / 122 (94.26%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/bool/InLongFunctionFactory.java|11|14|78.57%|
|:large_blue_circle:|io/questdb/std/DirectLongHashSet.java|104|108|96.30%|

Issue #6442, opened 2025-11-24
Better error message in approx_median() / approx_percentile()


puzpuzpuz:
### To reproduce

```sql
create table x (d double);

insert into x values (0.0);
insert into x values (1000.0);
insert into x values (6.35e-11);
insert into x values (76179182.16484068);

select approx_median(d) from x;
```

The query fails with the following error:
```
The value 7.516817105484058E7 is out of bounds for histogram, current covered range [5.820766091346741E-11, 2048.0) cannot be extended any further.
Caused by: [-1] Cannot resize histogram covered range beyond (1L << 63) / (1L << 4) - 1.
Caused by: [-1] Operation would underflow and lose precision of already recorded value counts`, errno=-1, q=`select approx_median(d) from x
```
yet `approx_median(d, 0)` (notice lower precision from the default `1`) succeeds. We should check if there is an easy way to support this dataset with the default precision or, at least, include a hint on lowering the precision into the error message.

### QuestDB version:

9.2

### OS, in case of Docker specify Docker and the Host OS:

Linux

### File System, in case of Docker specify Host File System:

ext4

### Full Name:

Andrei Pechkurov

### Affiliation:

QuestDB

### Have you followed Linux, MacOs kernel configuration steps to increase Maximum open files and Maximum virtual memory areas limit?

- [x] Yes, I have

peter007-cmd:
Hi @puzpuzpuz Can you please assign this issue to me

puzpuzpuz:
@peter007-cmd sure, done.

peter007-cmd:
Hi @puzpuzpuz can you please review my PR if possible and let me know if there are any changes required

PR #6446, opened 2025-11-25
feat(core): hints for handling histogram out of bounds error for approx_percentile()/approx_median()

peter007-cmd:
This PR provides the user with hints to resolve the Histogram out of Bounds error which occurs during the approx_median(d) query,when the inserted values exceed the allowed range for the Histogram.In this case the default precision value for median is set to be 1.

- This error can often be resolved by using a lower precision value with the query approx_median(d,0).
- Or normalizing the values to fit into the allowed range.
- This PR is in reference to the issue #6442

### DIFF
--- a/core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/DoubleHistogram.java
+++ b/core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/DoubleHistogram.java
@@ -1268,7 +1268,6 @@ public class DoubleHistogram extends EncodableHistogram implements DoubleValueRe
                 while (value >= currentHighestValueLimitInAutoRange);
             }
         } catch (CairoException ex) {
-            throw CairoException.nonCritical().put("The value ").put(value)
             // Build the base error message first
             CairoException err = CairoException.nonCritical().put("The value ").put(value)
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/DoubleHistogram.java
```diff
@@ -1268,10 +1268,20 @@ private synchronized void autoAdjustRangeForValueSlowPath(final double value) {
                 while (value >= currentHighestValueLimitInAutoRange);
             }
         } catch (CairoException ex) {
-            throw CairoException.nonCritical().put("The value ").put(value)
+            // Build the base error message first
+            CairoException err = CairoException.nonCritical().put("The value ").put(value)
                     .put(" is out of bounds for histogram, current covered range [")
                     .put(currentLowestValueInAutoRange).put(", ").put(currentHighestValueLimitInAutoRange)
                     .put(") cannot be extended any further.\nCaused by: ").put(ex.getMessage());
+
```
puzpuzpuz, line 1276: nit: redundant newline here and on L1282

PR #6453, opened 2025-11-25
fix(pgwire): batch SELECTs with bound parameters return correct results

### LINKED ISSUE #6123: Bound parameters in batches is broken

jerrinot:
Fixed a bug where queries with bound parameters executed in a batch would return incorrect results. When multiple SELECT statements with different parameter values were sent together (pipelining), all queries would incorrectly use the parameter values from the last query in the batch instead of their own values.

This affected PostgreSQL clients that use pipelining, such as Npgsql (.NET), where multiple queries are sent before waiting for responses. Each query now correctly uses its own bound parameter values.

I could not find a way to trigger this behaviour from a Java test using PGJDBC/R2DBC-Postgres. PGJDBC does not allow batch SELECTs at all. R2DBC supports it in the API, but then it sends SYNC for each SELECT anyway: https://github.com/pgjdbc/r2dbc-postgresql/issues/527

The C# test/reproducer contributed by @nwoolmer, :bow: 

Fixes #6123
Tandem: https://github.com/questdb/questdb-enterprise/pull/795
Supersedes #6401

### DIFF
--- a/ci/test-pipeline.yml
+++ b/ci/test-pipeline.yml
@@ -45,6 +45,15 @@ stages:
               # Trigger the enterprise CI pipeline via Azure DevOps REST API
               AUTH_HEADER="Authorization: Basic $(printf ':%s' "$AZURE_DEVOPS_ENT_PAT" | base64 -w0)"
 
+              # For PR builds, Azure uses an ephemeral merge commit, not the actual PR head.
+              # Fetch the real PR head commit from GitHub API.
+              if [ -n "$OSS_PR_NUMBER" ]; then
+                OSS_COMMIT=$(curl -s -H "Authorization: token $GH_TOKEN" \
+                  "https://api.github.com/repos/questdb/questdb/pulls/$OSS_PR_NUMBER" \
+                  | jq -r '.head.sha')
+                echo "Fetched PR head commit from GitHub API: $OSS_COMMIT"
+              fi
+
               # Check if matching branch exists in enterprise repo (private, requires auth)
               BRANCH_CHECK=$(curl -s -o /dev/null -w "%{http_code}" \
                 -H "Authorization: token $GH_TOKEN" \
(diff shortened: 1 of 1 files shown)


jerrinot:
@coderabbitai please review

bluestreak01:
@coderabbit happy thanksgiving, can you review?

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 59 / 74 (79.73%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGConnectionContext.java|6|10|60.00%|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGPipelineEntry.java|53|64|82.81%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=199245)
**Tested Commit:** `ac38f26bae68eb8e5ad9fc3d0b49134d2ff2b227`

PR #6463, opened 2025-11-27
fix(sql): breaking change 💥 - inconsistent data in chained period materialized views

puzpuzpuz:
Fixes inconsistent materialized view data that may be observed on chained views with period refresh. In such scenarios the dependent view may end up with rows that are no longer present in its base view.

Also includes a breaking change. Namely, MANUAL PERIOD views are no longer automatically refreshed when a period ends. Just like with any other MANUAL view the only way to refresh a MANUAL PERIOD view now is via REFRESH MATERIALIZED VIEW SQL.

Finally, improves `MatViewFuzzTest`:
* Improves fake clock limit calculation, so that threads do more jumps
* Adds randomized sample by interval to mat view query

### DIFF
--- a/core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
+++ b/core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
@@ -332,7 +332,11 @@ public class MatViewRefreshJob implements Job, QuietCloseable {
                 if (periodLo == Numbers.LONG_NULL) {
                     periodLo = baseTableReader.getMinTimestamp();
                 }
-                if (periodLo < rangeTo) {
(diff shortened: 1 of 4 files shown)


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewTimerJob.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewStateStoreImpl.java|1|1|100.00%|

bluestreak01 (approved):

PR #5872, opened 2025-07-06
feat(core): Adding support for musl

RaphDal:
This PR adds musl libc support for both x86_64 and aarch64 architectures to QuestDB.

Users' should now be able to run QuestDB on linux distributions like Alpine and we might also provide them with smaller docker images based on Alpine.

PR #6477, opened 2025-12-01
perf(sql): reduce query latency from memory deallocation overhead on Linux x86_64

jerrinot:
This change enables [jemalloc](https://jemalloc.net/) by default in Linux x86_64 binary distributions. jemalloc provides a significant boost to allocation-heavy queries such as parallel SAMPLE BY/GROUP BY.

This PR also fixes jemalloc built-in binary - the library was built with the `je_` prefix in the exported symbols. So, `je_malloc` instead of `malloc()`, `je_free()` instead of `free()`, etc. The end result was that even when `LD_PREDLOAD` was set to the jemalloc library, the database would still use the default system allocator from libc.

From now on, the built-in jemalloc binary can only be used on Linux (previously, we also compiled it for MacOS). When `QDB_JEMALLOC=true` is set on other OSes, `questdb.sh` will return an error.

Finally, jemalloc is now enabled for the `Run tests with Coverage` CI runner. That's to find any potential regressions around the glibc to jemalloc switch.

### DIFF
--- a/core/src/main/bin/questdb.sh
+++ b/core/src/main/bin/questdb.sh
@@ -150,7 +150,7 @@ function export_java {
 function export_jemalloc() {
     local arch=$(uname -m)
 
-    # Auto-enable jemalloc on Linux x86_64 for open source edition with bundled JRE
+    # auto-enable jemalloc on Linux x86_64 for open source edition with bundled JRE
     # if QDB_JEMALLOC is not explicitly set
     if [[ -z "${QDB_JEMALLOC+x}" ]]; then
         if [[ "$QDB_OS" == "Linux" && "$arch" == "x86_64" && "$JAVA_MAIN" == *"io.questdb"* && "$QDB_PACKAGE" == "withjre" ]]; then


jerrinot:
@coderabbitai: please review, pay attention to the CMake script setup, pull enough context to see the big picture

jerrinot:
@puzpuzpuz good point. I removed macos support entirely, and change questdb.sh to fail-fast when you attempt to enable JEMALLOC on an unsupported platform

core/src/main/bin/questdb.sh
```diff
@@ -149,14 +149,17 @@ function export_java {
...
+      if [[ "$QDB_OS" != "Linux" ]]; then
+          echo "Error: jemalloc is only supported on Linux (detected: $QDB_OS)"
+          exit 55
+      fi
+      jemalloc_so=$(ls $BASE/libjemalloc* 2>/dev/null)
+      if [[ -r "${jemalloc_so}" ]]; then
+          export LD_PRELOAD=${jemalloc_so}
           echo "Using jemalloc"
+      else
+          echo "Error: jemalloc library not found in $BASE"
```
puzpuzpuz: Isn't it too much to fail the startup because of unavailable allocator? I'd rather print a warning and continue with the default allocator. Theoretically, we may have users who have `QDB_JEMALLOC=true` set on MacOS.

jerrinot: This is opt-in. If a user explicitly asked for it and we cannot deliver, then I believe it's best to fail.

puzpuzpuz: OK, but can we at least include an instruction how to fix things into the error message? Someone may have the env variable set ages ago, so that they don't remember anything about it.

jerrinot: that's a very good idea!

jerrinot: wdyt now?

puzpuzpuz: Looks good

jerrinot:
@puzpuzpuz thanks for your help with this PR!

puzpuzpuz:
> @puzpuzpuz thanks for your help with this PR! 

Thanks for finding that we lack the symbols in the binary and fixing it.

core/src/main/c/share/jemalloc-cmake/include/jemalloc/jemalloc.h
```diff
@@ -70,32 +70,32 @@ extern "C" {
...
+#  define je_smallocx_8d8379da443f46dc976252b968cb9ca8e63ec974 smallocx_8d8379da443f46dc976252b968cb9ca8e63ec974
+#  define je_nallocx nallocx
+#  define je_posix_memalign posix_memalign
+#  define je_rallocx rallocx
+#  define je_realloc realloc
+#  define je_sallocx sallocx
+#  define je_sdallocx sdallocx
+#  define je_xallocx xallocx
+#  define je_valloc valloc
+#  define je_malloc_size malloc_size
```
kafka1991, line 100: should we also add `define je_memalign memalign` item?

puzpuzpuz: It's a good idea. I think CH's list was taken as the inspiration and here is the latest version of it which includes `je_memalign`:
https://github.com/ClickHouse/ClickHouse/blob/ca37ff2a0553222cef7f044c198c5f3ad21dc280/contrib/jemalloc-cmake/include/jemalloc/jemalloc_rename.h#L7-L30

kafka1991: And also we need
```
#if !defined(USE_MUSL)
    #define JEMALLOC_OVERRIDE_MEMALIGN
    #define JEMALLOC_OVERRIDE_VALLOC
    #define JEMALLOC_OVERRIDE_PVALLOC
    #define JEMALLOC_OVERRIDE___LIBC_PVALLOC
#endif
```
in cmake.
We may also need to verify the symbol table of the final dynamic lib.

puzpuzpuz:
@jerrinot how about adding a CI Linux runner that would use jemalloc?

core/src/main/c/share/jemalloc-cmake/include/jemalloc/jemalloc.h
jerrinot: `memalign` is a good idea, thanks for that! 

I am not entirely sure whether to support jemalloc when running with musl at all. jemalloc is used to squeeze the last drop of performance. musl has entirely different goals. let me try it anyway.

jerrinot: @kafka1991 musl is still not officially supported, I added a comment to https://github.com/questdb/questdb/pull/5872
wdyt?

jerrinot: also: updated symbols at https://github.com/jemalloc/jemalloc/blob/120abd703addce50fb9105ee4f7e42c3612c3774/configure.ac#L1132-L1145 

older glibc might not have all symbols, but I don't think there is any harm in exporting jemalloc implementation of symbols which are not available in a stdlib. am I wrong?

puzpuzpuz: > also: updated symbols at https://github.com/jemalloc/jemalloc/blob/120abd703addce50fb9105ee4f7e42c3612c3774/configure.ac#L1132-L1145

You should probably also update the list under `#ifdef JEMALLOC_MANGLE` directive. WDYT?

puzpuzpuz:
@jerrinot do you need some help with this one? I'm interested in getting it shipped in the next release.

core/src/main/c/share/jemalloc-cmake/include/jemalloc/jemalloc.h
jerrinot: you are right. done.

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=202368)
**Tested Commit:** `e4f0ae12eefff811546c6c41bded1b31a6e81f0e`

jerrinot:
@puzpuzpuz thanks for the ping. I pushed changes. I added jemalloc to coverage tests. since these are the only tests running in hotspot@x86_64
let' see.

puzpuzpuz:
> @puzpuzpuz thanks for the ping. I pushed changes. I added jemalloc to coverage tests. since these are the only tests running in hotspot@x86_64

Thanks! How about enabling jemalloc by default on Linux in `questdb.sh`?

jerrinot:
@puzpuzpuz done!

puzpuzpuz (approved):

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=202378)
**Tested Commit:** `fda06cd9f1118e3bfcf71280a4d1ce3ddd660986`

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=202385)
**Tested Commit:** `d5a4d7c316c88192a1addb709cb122134c180b70`

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/frm/file/MemoryFixFrameColumn.java|1|1|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=202512)
**Tested Commit:** `1fdb4596380fb2dd0cf8ceee0edf7cce26179507`

puzpuzpuz (approved):

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/frm/file/MemoryFixFrameColumn.java|1|1|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=203388)
**Tested Commit:** `62c30a3765f69befe9524120ce8344506fc43ee6`

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/frm/file/MemoryFixFrameColumn.java|1|1|100.00%|

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/frm/file/MemoryFixFrameColumn.java|1|1|100.00%|

PR #6481, opened 2025-12-01
perf(sql): use unordered maps only in single-column case

puzpuzpuz:
Tiny column sinks in `Unordered4Map` and `Unordered8Map` led to false sharing when jemalloc is in use. So, these maps are now used only in the single column case and support a few 4 byte/8 byte columns.

Also, removes `Unordered2Map` since it shows no improvement on top of other maps, so it's simply not worth it having it in the code base.

### Benchmarks

This patch significantly reduces execution times of Q15 and Q35 (but not only) in ClickBench when jemalloc is in use. The below results are measured on c6a.metal instance:

Screenshot from 2025-12-03 19-20-04
Screenshot from 2025-12-03 19-20-14

### DIFF
--- a/ci/snapshot-pipeline.yml
+++ b/ci/snapshot-pipeline.yml
@@ -8,6 +8,8 @@ pr: none
 variables:
   MVN_VERSION: "3.6.3"
   MVN: "apache-maven-$(MVN_VERSION)"
+  JAVA_HOME_17_X64: "/usr/lib/jvm/java-17-openjdk-amd64"
+  JAVA_HOME: "/usr/lib/jvm/java-17-openjdk-amd64"
 
 stages:
   - stage: BuildSnapshotMaster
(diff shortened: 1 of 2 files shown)


glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 79 / 92 (85.87%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/map/Unordered8Map.java|29|37|78.38%|
|:large_blue_circle:|io/questdb/cairo/map/Unordered4Map.java|30|35|85.71%|
|:large_blue_circle:|io/questdb/cairo/map/Unordered8MapRecord.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/map/Unordered4MapRecord.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/map/MapFactory.java|10|10|100.00%|

PR #6482, opened 2025-12-01
fix(sql): crash when using parallel ORDER BY with LIMIT and JIT filter

puzpuzpuz:
Parallel ORDER BY + LIMIT queries that involve a JIT-compiled filter like the following one
```sql
SELECT * FROM tab WHERE key = ? ORDER BY price DESC LIMIT 1;
```
were producing segfaults due to missing bind variable memory initialization.

The crash occurs both when bind variables are used explicitly, but also when the filter involves not-yet-existing symbols, e.g.
```sql
SELECT * FROM tab WHERE key = 'non_existing_key' ORDER BY price DESC LIMIT 42;
```

As a mitigation of this bug, `cairo.sql.parallel.topk.enabled=false` can be set. This will disable the problematic execution plan.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/O3PartitionPurgeJob.java
+++ b/core/src/main/java/io/questdb/cairo/O3PartitionPurgeJob.java
@@ -355,6 +355,29 @@ public class O3PartitionPurgeJob extends AbstractQueueConsumerJob<O3PartitionPur
                 boolean rangeUnlocked = previousNameVersion < nextNameVersion
                         && txnScoreboard.isRangeAvailable(previousNameVersion, nextNameVersion);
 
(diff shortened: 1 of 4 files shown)


puzpuzpuz:
@coderabbitai review

bluestreak01 (approved):

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 47 / 52 (90.38%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncFilterUtils.java|44|49|89.80%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKAtom.java|3|3|100.00%|

Issue #11, opened 2014-08-26
Persist to Existing DB


sirinath:
Some mechanism to persist to a normal DB may be useful.

PR #6502, opened 2025-12-04
chore(sql): off-heap and flyweight histogram

mcadariu:
Related to https://github.com/questdb/questdb/issues/4120. 

The implementation is inspired by the existing on-heap `Histogram`, and the existing `GroupByArraySink`.

The tests compare outputs to the on-heap `Histogram` which is the reference implementation / ground truth.


core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
```diff
@@ -0,0 +1,360 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ ******************************************************************************/
+
+package io.questdb.std.histogram.org.HdrHistogram;
```
mcadariu: This might not be the best package to place this file in. I could move it to where the other sinks are, it will mean many things need touching to be made from package-private to protected, open for suggestion!

puzpuzpuz: It's fine to use the same package. We already have a number of changes on top of the vanilla `HdrHistogram`.

puzpuzpuz:
@mcadariu this is a pretty cool addition! Have you also looked into a simple way to merge two histograms? This is crucial for the parallel algorithm and it's be great to include a merge method into this PR. WDYT?

mcadariu:
Good one, done! 
If I understand correctly, merging them boils down to calling `add` (from `AbstractHistogram`)

mcadariu:
The `Rust Test and Lint` job seems to be failing, should be unrelated

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogramSink.java
```diff
@@ -0,0 +1,381 @@
...
+ * <p>
+ * Buffer layout is the following:
+ * <pre>
+ * | totalCount | normalizingIndexOffset | maxValue | minNonZeroValue | histogram counts array |
+ * +------------+------------------------+----------+-----------------+------------------------+
+ * |  8 bytes   |        4 bytes         | 8 bytes  |    8 bytes      | countsArrayLength * 8  |
+ * +------------+------------------------+----------+-----------------+------------------------+
+ * </pre>
+ */
+public class GroupByHistogramSink extends AbstractHistogram implements Mutable {
```
puzpuzpuz: Shall we remove word sink from the class name and the javadoc? `GroupByHistogram` is descriptive enough. In QDB, a sink stands from something that can hold a single value, e.g. a string string or an array sink. Histograms are different as they store aggregated data.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogramSink.java
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    @Override
+    public void clear() {
+        ptr = 0;
+        allocatedSize = 0;
+    }
+
+    @Override
+    public GroupByHistogramSink copy() {
```
puzpuzpuz: I'd rather not bother with the methods we don't need and simply throw `UnsupportedOperationException`. Same applies to `copyCorrectedForCoordinatedOmission` and maybe some other methods here. We only need the minimal required methods.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogramSink.java
```diff
@@ -0,0 +1,381 @@
...
+/**
+ * Specialized off-heap histogram sink used in {@link io.questdb.griffin.engine.functions.GroupByFunction}s.
+ * <p>
+ * Uses provided {@link GroupByAllocator} to allocate the underlying buffer. Grows the buffer when needed.
+ * <p>
+ * Buffer layout is the following:
+ * <pre>
+ * | totalCount | normalizingIndexOffset | maxValue | minNonZeroValue | histogram counts array |
+ * +------------+------------------------+----------+-----------------+------------------------+
+ * |  8 bytes   |        4 bytes         | 8 bytes  |    8 bytes      | countsArrayLength * 8  |
```
puzpuzpuz: Don't we need `countsArrayLength` and maybe `bucketCount` and `highestTrackableValue` to be stored in the header? Any of them may be changed when `resize()` and then `establishSize()` is called.

In principle, it looks like `AbstractHistogram` should not be extended as it has lots of on-heap fields that can mutate. After `of()` is called those fields will remain their values making the histogram inconsistent. Am I missing something here?

puzpuzpuz:
Sorry for the delay due to the Xmas season. I left a few comments - PTAL.

mcadariu:
Happy New Year! Thanks for the review, will check/address bit later

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogramSink.java
mcadariu: I see, indeed, better without it then. Removed.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogramSink.java
mcadariu: Ah, right! Applied.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogramSink.java
mcadariu: > Don't we need countsArrayLength and maybe bucketCount and highestTrackableValue to be stored in the header? Any of them may be changed when resize() and then establishSize() is called.

Right, added to the header and am checking for others.

> In principle, it looks like AbstractHistogram should not be extended as it has lots of on-heap fields that can mutate. After of() is called those fields will remain their values making the histogram inconsistent. Am I missing something here?

I don't think you're missing anything, one thing is that `AbstractHistogram` has a lot of detailed logic that is it seemed to me reused throughout the hierarchy. If I bring it over it would introduce duplication I think. 

My approach so far has been to see `GroupByHistogram` as just the minimal extension needed that changes how the necessary fields are stored, but no more than that. This way there is less risk of affecting correctness. The PR is the smallest it can be as well.

Am _I_ missing something? 🤔 If you think it's best I can definitely remove the `extends AbstractHistogram` and see what happens.

puzpuzpuz: > Am _I_ missing something? 🤔 If you think it's best I can definitely remove the `extends AbstractHistogram` and see what happens.

I'm just seeing a lot of fields (think, state) in that hierarchy of classes while we need this object to be an off-heap flyweight, i.e. its state is fully loaded from off-heap memory when you make an `of()` call.

mcadariu: Ah pff, there's fields in `AbstractHistogramBase` too 🤦
I am changing the approach by indeed not extending `AbstractHistogram` and trying to get the tests to pass that way. Potentially the duplication can be solved by composition instead of inheritance, but I'll cross that bridge when I get there.

mcadariu: One new way to measure how wide is a screen: does the javadoc containing the buffer layout of `GroupByHistogram` fit into the screen in its entirety? 😄

mcadariu:
I have pushed another iteration of this. `GroupByHistogram` is not extending `AbstractHistogram` anymore. There are some comments on top of every method showing where the logic comes from. This is only for review purposes and I can remove them before merging. There is now duplication but I suggest it is to be removed only after the `approx_percentile` is done (in a new separate PR) because at that point I expect more tests will be in place (e.g. parallel fuzz tests) that will check the implementation and refactoring.

puzpuzpuz:
@mcadariu thanks for updating the PR! Porting `HdrHistogram` to off-heap memory is certainly tough. I should be able to do another review round later this week.

puzpuzpuz:
@coderabbitai review

mcadariu:
Ah, not too bad! The bot has not came up with too many things, will address bit later.

core/src/test/java/io/questdb/test/std/histogram/org/HdrHistogram/GroupByHistogramTest.java
```diff
@@ -0,0 +1,651 @@
...
+
+            dest.recordValue(100);
+            src.recordValue(1000000);
+
+            dest.merge(src);
+
+            Assert.assertEquals(2, dest.getTotalCount());
+            Assert.assertTrue(dest.getMaxValue() >= 1000000);
+        }
+    }
```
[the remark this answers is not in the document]
mcadariu: Done

mcadariu:
@coderabbitai review

puzpuzpuz:
> @coderabbitai review

Not code rabbit, but still. Sorry for the delays with the review. I'll do my best to check the updates tomorrow. In the meantime, have you tried porting the `approx_percentile()` SQL function to this histogram implementation and measuring the performance before/after?

mcadariu:
Ah, no rush! I was just curious to see if after addressing the points it raised it will now come back with 0 findings on a re-run.
I'll work on `approx_percentile()` in a branch off this one.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
```diff
@@ -0,0 +1,748 @@
...
+
+        this.add(other);
+    }
+
+    public void clear() {
+        ptr = 0;
+        allocatedSize = 0;
+    }
+
+    public GroupByHistogram copy() {
```
puzpuzpuz: The class no longer implements any `HdrHistogram` interfaces, so we should get rid of all redundant methods and, maybe, even fields.

core/src/test/java/io/questdb/test/std/histogram/org/HdrHistogram/GroupByHistogramTest.java
```diff
@@ -0,0 +1,651 @@
...
+import io.questdb.griffin.engine.groupby.GroupByAllocator;
+import io.questdb.griffin.engine.groupby.GroupByAllocatorFactory;
+import io.questdb.std.Rnd;
+import io.questdb.std.histogram.org.HdrHistogram.GroupByHistogram;
+import io.questdb.std.histogram.org.HdrHistogram.Histogram;
+import io.questdb.test.AbstractCairoTest;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class GroupByHistogramTest extends AbstractCairoTest {
```
puzpuzpuz, line 38: nit: would be great to add some fuzz (randomized) tests for `recordValue` and `merge`. The oracle would be `HdrHistogram` and `GroupByHistogram` should match its results.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
```diff
@@ -0,0 +1,748 @@
...
+
+        if (ptr != 0) {
+            this.lowestDiscernibleValue = Unsafe.getUnsafe().getLong(ptr + lowestDiscernibleValuePosition);
+            this.numberOfSignificantValueDigits = Unsafe.getUnsafe().getInt(ptr + numberOfSignificantValueDigitsPosition);
+            this.countsArrayLength = Unsafe.getUnsafe().getInt(ptr + countsArrayLengthPosition);
+            this.bucketCount = Unsafe.getUnsafe().getInt(ptr + bucketCountPosition);
+            this.highestTrackableValue = Unsafe.getUnsafe().getLong(ptr + highestTrackableValuePosition);
+            this.allocatedSize = headerSize + (countsArrayLength * 8L);
+
+            recalculateDerivedFields();
```
puzpuzpuz: I'm afraid we need to measure if parallel `approx_percentile()` based on this class is any faster than the `HdrHistogram` one. So far, this is the most complex `GroupBy*` flyweight we have and I wonder how much overhead state loading/storing adds.

mcadariu: Yes, for sure it is just an experiment for now.
The performance gains must be enough to offset the complexity.

mcadariu: The code to reproduce the below results is here: https://github.com/mcadariu/questdb/commit/3a738ef725ace3e7ab1d91fc3cb659fcd835f56c. 

| Partitions | Workers | master(ms) | branch(ms) | Speedup |
|------------|---------|------------|------------|---------|
| 8 | 1 | 272.83 | 332.00 | 0.82x |
| 8 | 2 | 298.09 | 256.82 | 1.16x |
| 8 | 4 | 261.97 | 151.48 | 1.73x |
| 8 | 8 | 278.01 | 93.12 | 2.99x |
| 16 | 1 | 274.84 | 358.52 | 0.77x |
| 16 | 2 | 280.12 | 246.31 | 1.14x |
| 16 | 4 | 276.31 | 146.84 | 1.88x |
| 16 | 8 | 277.70 | 91.61 | 3.03x |
| 32 | 1 | 276.38 | 358.16 | 0.77x |
| 32 | 2 | 277.82 | 245.49 | 1.13x |
| 32 | 4 | 278.01 | 146.33 | 1.90x |
| 32 | 8 | 279.90 | 89.28 | 3.13x |
| 64 | 1 | 276.64 | 359.55 | 0.77x |
| 64 | 2 | 278.64 | 242.56 | 1.15x |
| 64 | 4 | 281.32 | 145.31 | 1.94x |
| 64 | 8 | 282.44 | 91.61 | 3.08x |

The scaling is sub-linear but decent (IMO). With 8 workers there's 3x speedup.

core/src/test/java/io/questdb/test/std/histogram/org/HdrHistogram/GroupByHistogramTest.java
mcadariu: Added the fuzz tests.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
mcadariu: Indeed, removed the redundant methods.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
mcadariu: @puzpuzpuz how do the above results seem to you? Worth pursuing this PR further?

puzpuzpuz: Sorry for the lengthy delay :(

I'm looking into the PR now.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
```diff
@@ -0,0 +1,706 @@
...
+            addToCountAtIndex(countsIndex, count);
+        } catch (CairoException ex) {
+            handleRecordException(count, value, ex);
+        }
+        updateMinAndMax(value);
+        addToTotalCount(count);
+    }
+
+    // AbstractHistogram.handleRecordException(long, long, Exception) at lines 1607-1616
+    private void handleRecordException(final long count, final long value, Exception ex) {
```
puzpuzpuz, line 525: `handleRecordException` does not sync `highestTrackableValue` back to the off-heap header

`GroupByHistogram.java:577` — After `resize(value)`, the method overwrites the Java field:
```java
this.highestTrackableValue = highestEquivalentValue(valueFromIndex(countsArrayLength - 1));
```
But `reallocate()` at line 274 already wrote the *old* `highestTrackableValue` (from `establishSize`) to the header. The Java field and header now diverge.

**Code path**: `recordSingleValue(largeValue)` -> `incrementCountAtIndex` throws -> `handleRecordException` -> `resize(value)` writes `highestTrackableValue = value` to header -> line 577 overwrites the Java field with `highestEquivalentValue(...)` (a larger value) -> header still has `value`.

**Impact**: When another `GroupByHistogram` calls `of(ptr)` to repoint to this buffer, it reads the smaller value. Subsequent `merge()` calls (line 138: `other.getMaxValue() > this.highestTrackableValue`) may trigger unnecessary `resize()` + `realloc()`. Not data-corrupting, but causes spurious reallocations in the group-by hot path.

**Fix**: Add after line 577:
```java
Unsafe.getUnsafe().putLong(ptr + highestTrackableValuePosition, highestTrackableValue);
```

core/src/test/java/io/questdb/test/std/histogram/org/HdrHistogram/GroupByHistogramTest.java
```diff
@@ -0,0 +1,712 @@
...
+import io.questdb.std.histogram.org.HdrHistogram.Histogram;
+import io.questdb.test.AbstractCairoTest;
+import io.questdb.test.tools.TestUtils;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class GroupByHistogramTest extends AbstractCairoTest {
+
+    @Test
+    public void testBasicRecording() {
```
puzpuzpuz: Would be great to use `assertMemoryLeak()` in all of these tests to make sure we don't leak off-heap memory.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
```diff
@@ -0,0 +1,706 @@
...
+
+            recalculateDerivedFields();
+        } else {
+            this.allocatedSize = 0;
+        }
+
+        return this;
+    }
+
+    public void merge(GroupByHistogram other) {
```
puzpuzpuz, line 122: `merge()` performs redundant malloc + realloc when dest is empty and source has large values

`GroupByHistogram.java:134-140` — When `this.ptr == 0` and `other.getMaxValue() > this.highestTrackableValue`:
1. Line 135: `ensureCapacity()` -> `allocate()` with small initial `countsArrayLength`
2. Line 139: `resize()` -> `reallocate()` immediately to a larger size

`resize()` already handles `ptr == 0` via the `hadPreviousAllocation` flag (line 245/254). Fix by reordering the checks:

```java
if (this.ptr == 0) {
    if (other.getMaxValue() > this.highestTrackableValue) {
        resize(other.getMaxValue()); // allocates at the right size directly
    } else {
        ensureCapacity();
    }
} else if (other.getMaxValue() > this.highestTrackableValue) {
    resize(other.getMaxValue());
}
this.add(other);
```

core/src/test/java/io/questdb/test/std/histogram/org/HdrHistogram/GroupByHistogramTest.java
```diff
@@ -0,0 +1,712 @@
...
+    }
+
+    @Test
+    public void testPercentiles() {
+        try (GroupByAllocator allocator = GroupByAllocatorFactory.createAllocator(configuration)) {
+            Histogram onHeap = new Histogram(1, 3600000, 3);
+            GroupByHistogram offHeap = new GroupByHistogram(1, 3600000, 3);
+            offHeap.setAllocator(allocator);
+
+            Rnd rnd1 = new Rnd();
```
puzpuzpuz: Could you use `TestUtils.generateRandom(LOG)` for all `Rnd` initializations? This way the tests will be randomized.

core/src/test/java/io/questdb/test/std/histogram/org/HdrHistogram/GroupByHistogramTest.java
```diff
@@ -0,0 +1,712 @@
...
+import io.questdb.griffin.engine.groupby.GroupByAllocatorFactory;
+import io.questdb.std.Rnd;
+import io.questdb.std.histogram.org.HdrHistogram.GroupByHistogram;
+import io.questdb.std.histogram.org.HdrHistogram.Histogram;
+import io.questdb.test.AbstractCairoTest;
+import io.questdb.test.tools.TestUtils;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class GroupByHistogramTest extends AbstractCairoTest {
```
puzpuzpuz, line 38: Missing test: `autoResize=false` exception path

No test exercises `handleRecordException` when `autoResize` is `false` (line 570-573). The 3-argument constructor `GroupByHistogram(1, 1000, 3)` sets `autoResize = false`, but no test records a value > 1000 on such an instance to verify the `CairoException` is thrown.

puzpuzpuz, line 38: Missing test: negative value rejection

`countsArrayIndex()` at line 513-514 throws `CairoException("Histogram recorded value cannot be negative.")` for negative values. No test exercises this path.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
```diff
@@ -0,0 +1,706 @@
...
+    public void recordValueWithCount(final long value, final long count) throws CairoException {
+        recordCountAtValue(count, value);
+    }
+
+    // AbstractHistogram.recordSingleValue(long) at lines 1726-1735
+    private void recordSingleValue(final long value) throws CairoException {
+        int countsIndex = countsArrayIndex(value);
+        try {
+            incrementCountAtIndex(countsIndex);
+        } catch (CairoException | ArrayIndexOutOfBoundsException ex) {
```
puzpuzpuz: `ArrayIndexOutOfBoundsException` won't be thrown here as this sink uses off-heap memory, not arrays.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
```diff
@@ -0,0 +1,706 @@
...
+            handleRecordException(1, value, ex);
+        }
+        updateMinAndMax(value);
+        incrementTotalCount();
+    }
+
+    // AbstractHistogram.recordCountAtValue(long, long) at lines 1713-1722
```
puzpuzpuz: > at lines 1713-1722

These references are very fragile and will become obsolete soon.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
```diff
@@ -0,0 +1,706 @@
...
+    // AbstractHistogram.countsArrayIndex(int, int) at lines 1534-1546
+    private int countsArrayIndex(final int bucketIndex, final int subBucketIndex) {
+        assert(subBucketIndex < subBucketCount);
+        assert(bucketIndex == 0 || (subBucketIndex >= subBucketHalfCount));
+        int bucketBaseIndex = (bucketIndex + 1) << subBucketHalfCountMagnitude;
+        int offsetInBucket = subBucketIndex - subBucketHalfCount;
+        return bucketBaseIndex + offsetInBucket;
+    }
+
+    public void setAutoResize(boolean autoResize) {
```
puzpuzpuz: Seems to be unused. Do we even need this method?

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
puzpuzpuz: > The scaling is sub-linear but decent (IMO). With 8 workers there's 3x speedup.

@nwoolmer WDYT? In single worker scenarios we got slower with the parallel impl.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
```diff
@@ -0,0 +1,706 @@
...
+            subBucketIndex -= subBucketHalfCount;
+            bucketIndex = 0;
+        }
+        return valueFromIndex(bucketIndex, subBucketIndex);
+    }
+
+    // AbstractHistogram.countsArrayIndex(long) at lines 354-361
+    private int countsArrayIndex(final long value) {
+        if (value < 0) {
+            throw CairoException.nonCritical().put("Histogram recorded value cannot be negative.");
```
puzpuzpuz, line 480: Would be great to add a test to exercise this path.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
puzpuzpuz: @mcadariu how about optimizing `of()` method which is close for every aggregated row? This could help closing the performance gap between the OG `HdrHistogram` and your off-heap implementation.

# Optimization: remove `recalculateDerivedFields()` from `of()`

## Problem

`of()` is called on every group-by key switch to repoint the flyweight to a different off-heap histogram buffer. Currently it reads 5 fields from the header and calls `recalculateDerivedFiel
ds()`, which performs expensive floating-point math on every repoint:

```java
public GroupByHistogram of(long ptr) {
    this.ptr = ptr;
    if (ptr != 0) {
        this.lowestDiscernibleValue = Unsafe.getUnsafe().getLong(ptr + lowestDiscernibleValuePosition);
        this.numberOfSignificantValueDigits = Unsafe.getUnsafe().getInt(ptr + numberOfSignificantValueDigitsPosition);
        this.countsArrayLength = Unsafe.getUnsafe().getInt(ptr + countsArrayLengthPosition);
        this.bucketCount = Unsafe.getUnsafe().getInt(ptr + bucketCountPosition);
        this.highestTrackableValue = Unsafe.getUnsafe().getLong(ptr + highestTrackableValuePosition);
        this.allocatedSize = headerSize + (countsArrayLength * 8L);

        recalculateDerivedFields();  // <-- 2x Math.log, 1x Math.pow, 1x Math.ceil per repoint
    } else {
        this.allocatedSize = 0;
    }
    return this;
}
```

`recalculateDerivedFields()` computes 8 fields (`unitMagnitude`, `unitMagnitudeMask`, `subBucketCount`, `subBucketHalfCount`, `subBucketHalfCountMagnitude`, `subBucketMask`, `leadingZeroCountBase`, plus intermediate `subBucketCountMagnitude`) from only two inputs: `lowestDiscernibleValue` and `numberOfSignificantValueDigits`.

## Key observation

`lowestDiscernibleValue` and `numberOfSignificantValueDigits` are constructor parameters. They are identical for every histogram instance created by a given `approx_percentile` function call and never change after construction. The constructor already calls `recalculateDerivedFields()` via `init()`, so the 8 derived fields are correct from construction and stay valid across repoints.

## Fix

Remove the reads of `lowestDiscernibleValue`/`numberOfSignificantValueDigits` and the `recalculateDerivedFields()` call from `of()`. Only read the 3 fields that actually vary per-instance:

```java
public GroupByHistogram of(long ptr) {
    this.ptr = ptr;
    if (ptr != 0) {
        this.countsArrayLength = Unsafe.getUnsafe().getInt(ptr + countsArrayLengthPosition);
        this.bucketCount = Unsafe.getUnsafe().getInt(ptr + bucketCountPosition);
        this.highestTrackableValue = Unsafe.getUnsafe().getLong(ptr + highestTrackableValuePosition);
        this.allocatedSize = headerSize + (countsArrayLength * 8L);
    } else {
        this.allocatedSize = 0;
    }
    return this;
}
```

This gives: 3 Unsafe reads instead of 5, zero floating-point math, and no branch/throw for the `subBucketCountMagnitude + unitMagnitude > 62` validation.

Note: `lowestDiscernibleValue` and `numberOfSignificantValueDigits` still need to be **stored** in the header by `allocate()`/`reallocate()` for persistence, but the flyweight does not need to **read** them back — it already has them as instance fields from `init()`.

mcadariu:
Thanks for the detailed review @puzpuzpuz! I'll work on it soon and redo the measurements.

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
mcadariu: Good one, done

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
mcadariu: Good one, done

core/src/test/java/io/questdb/test/std/histogram/org/HdrHistogram/GroupByHistogramTest.java
mcadariu: Good one, done

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
mcadariu: Good one, done

core/src/test/java/io/questdb/test/std/histogram/org/HdrHistogram/GroupByHistogramTest.java
mcadariu: Good one, done

core/src/test/java/io/questdb/test/std/histogram/org/HdrHistogram/GroupByHistogramTest.java
mcadariu: Good one, done

mcadariu: Good one, done

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
mcadariu: Good one, done

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
mcadariu: Good one, done

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
mcadariu: Good one, done

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
mcadariu: Good one, done

mcadariu:
@puzpuzpuz in another installment of "why is QuestDB so fast", after applying your comments, the measurements  look much better! We have to ensure everything is correct, let me know if you can think of any additional tests or checks.

| Partitions | Workers | master(ms) | branch(ms) | Speedup |
|------------|---------|------------|------------|---------|
| 8 | 1 | 282.36 | 101.67 | 2.78x |
| 8 | 2 | 444.34 | 83.47 | 5.32x |
| 8 | 4 | 267.42 | 51.23 | 5.22x |
| 8 | 8 | 282.67 | 33.91 | 8.34x |
| 16 | 1 | 278.88 | 122.88 | 2.27x |
| 16 | 2 | 278.62 | 84.65 | 3.29x |
| 16 | 4 | 280.73 | 51.94 | 5.40x |
| 16 | 8 | 284.20 | 32.61 | 8.72x |
| 32 | 1 | 279.98 | 123.32 | 2.27x |
| 32 | 2 | 285.15 | 85.41 | 3.34x |
| 32 | 4 | 280.56 | 51.97 | 5.40x |
| 32 | 8 | 282.83 | 31.78 | 8.90x |
| 64 | 1 | 285.18 | 121.65 | 2.34x |
| 64 | 2 | 283.29 | 84.78 | 3.34x |
| 64 | 4 | 283.75 | 52.21 | 5.43x |
| 64 | 8 | 286.16 | 35.06 | 8.16x |

Code to obtain the above table is [here](https://github.com/mcadariu/questdb/pull/1/changes) (class `ApproxPercentileParallelBenchmarkTest`).

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
puzpuzpuz, line 141: nit: `@Override` won't hurt here.

puzpuzpuz, line 25: nit: as this sink is no longer coupled with HdrHistogram library, we could move it to the `io.questdb.griffin.engine.groupby` package.

puzpuzpuz (commented):
Long story short, the PR looks great. A few tests could be improved, but I'm happy to do that in a separate PR. So, let me know if you won't be able to improve the tests and I'll approve the PR and handle the tests. The description of the gaps is below.

The fuzz tests (`testRecordValueFuzz`, `testMergeFuzz`, `testRepointAndRecordFuzz`) compare against HdrHistogram and cover the core paths well. The gap is in the deterministic merge edge-case tests:

- **`testMergePreservesPercentiles`** compares a merged `GroupByHistogram` against   another `GroupByHistogram` that received all values directly. This is self-referential — if recording and merging share a systematic bug, the test still passes. The `combined` histogram should be an HdrHistogram `Histogram`.

- **`testMergeEmptyDestinationWithLargeValues`** exercises the optimized `merge()` path where `dest.ptr == 0` and the source has large values that trigger
  `resize()` instead of `ensureCapacity()`. This is the exact path changed by review comment #11. No oracle verifies the output.

- **`testMergeExtremeValueRange`** merges values spanning 1 to 9B (causing major internal resizing) but only checks `totalCount` and a rough `maxValue` bound. Percentile accuracy after such a large structural change is not verified
  against an oracle.

The deterministic tests exist and hit the right edge cases — they just assert hardcoded values rather than comparing against HdrHistogram. Adding oracle
comparisons to these three would close the gap.

mcadariu:
Thanks! I can address the remaining points a bit later 👍

core/src/main/java/io/questdb/std/histogram/org/HdrHistogram/GroupByHistogram.java
mcadariu: Done 👍

mcadariu: Done 👍

puzpuzpuz:
@mcadariu thanks for adding the tests! The PR looks good. Linter does like a couple things, could you check? Once the linter is happy, I'll approve the PR.
```
diff --git a/core/src/main/java/io/questdb/griffin/engine/groupby/GroupByHistogram.java b/core/src/main/java/io/questdb/griffin/engine/groupby/GroupByHistogram.java
index e7be2c28e..9714d425c 100644
--- a/core/src/main/java/io/questdb/griffin/engine/groupby/GroupByHistogram.java
+++ b/core/src/main/java/io/questdb/griffin/engine/groupby/GroupByHistogram.java
@@ -352,8 +352,8 @@ public class GroupByHistogram implements Mutable {
     private void checkBounds(int index) {
         if (index < 0 || index >= countsArrayLength) {
             throw CairoException.nonCritical()
-                .put("index ").put(index)
-                .put(" out of bounds [0, ").put(countsArrayLength).put(')');
+                    .put("index ").put(index)
+                    .put(" out of bounds [0, ").put(countsArrayLength).put(')');
         }
     }
 
@@ -485,8 +485,8 @@ public class GroupByHistogram implements Mutable {
     }
 
     private int countsArrayIndex(final int bucketIndex, final int subBucketIndex) {
-        assert(subBucketIndex < subBucketCount);
-        assert(bucketIndex == 0 || (subBucketIndex >= subBucketHalfCount));
+        assert (subBucketIndex < subBucketCount);
+        assert (bucketIndex == 0 || (subBucketIndex >= subBucketHalfCount));
         int bucketBaseIndex = (bucketIndex + 1) << subBucketHalfCountMagnitude;
         int offsetInBucket = subBucketIndex - subBucketHalfCount;
         return bucketBaseIndex + offsetInBucket;
```

puzpuzpuz (approved):
Thanks for this epic patch!

javier:
I made a blog post out of this exchange. Very cool :)

https://questdb.com/blog/code-review-tripled-histogram-speedup/

mcadariu:
Thanks @javier, awesome write-up! I felt too that there's a story to tell here and you have nailed it.

PR #6511, opened 2025-12-06
feat(parser): add overload candidate hints for function mismatch errors (#5913)

### LINKED ISSUE #5913: Improve error reporting when a function does not accept given parameters

When a function does not have an overloading matching parameters, users get a hard-to-act-on error:

Image

[Source](https://github.com/questdb/questdb/blob/2fba4f0a3afb5d9e7094b9c6f9efe30808bfa7a0/core/src/main/java/io/questdb/griffin/FunctionParser.java#L536): 
```java
ex.put("there is no matching function `").put(node.token).put("` with the argument types: (");
                for (int i = 0, n = args.size(); i < n; i++) {
                    if (i > 0) {
                        ex.put(", ");
                    }
                    putArgType(args, i, ex);
                }
```

When a function of this name exists, but does not support given arguments, then the error should include a candidate list. In some cases, the list could be long so smart prioritisation is required and perhaps also truncating. 

The error message should differentiate between the cases where:
1. there is the right number of parameters, but of an unsupported type
2. wrong pameter count
The error message should include this information in a hint. 

_No response_

### Full Name:

Jaromir Hamala

### Affiliation:

QuestDB

### Additional context

_No response_

nithieshnaik:
- Adds detailed hint system when function name exists but argument types don't match
- Differentiates between wrong parameter count and unsupported argument types
- Prioritizes same-arity overloads and truncates long candidate lists
- Introduces MAX_CANDIDATES_IN_HINT and structured overload signature printing
- Enhances debugging experience for SQL users

Closes #5913.


bluestreak01:
is this ai generated code?

nithieshnaik:
No.I wrote it myself .I just used AI as a quick reference while working on it

PR #6529, opened 2025-12-12
fix(core): fix bad connection state after TLS session initialization failure

jerrinot:
When TLS session initialization failed, then the connection context was closed twice, and it could be also pooled twice. This could lead to issues in later connections.

Joint work with @RaphDal

Note for reviewers: `disconnect()` enqueues the context in the `disconnectQueue`. The queue is later processed, and the connection is closed.


RaphDal (approved):

jerrinot:
hello @coderabbitai, my favourite robot, please review this fine art! :)

glasstiger:
#### [PR Coverage check]

PR #6537, opened 2025-12-14
fix(sql): fix `latest by` syntax error after a subquery

glasstiger:
`LATEST BY ...` was not accepted after a subquery, only `LATEST ON ... BY ...`.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/SqlParser.java
+++ b/core/src/main/java/io/questdb/griffin/SqlParser.java
@@ -2320,7 +2320,7 @@ public class SqlParser {
             parseSelectFrom(lexer, model, masterModel.getWithClauses(), sqlParserCallback);
             tok = setModelAliasAndTimestamp(lexer, model);
 
-            // expect [latest by] (deprecated syntax)
+            // expect [latest by]
             if (tok != null && isLatestKeyword(tok)) {
                 parseLatestBy(lexer, model);
                 tok = optTok(lexer);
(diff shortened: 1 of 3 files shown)


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlParser.java|1|1|100.00%|

bluestreak01:
## Review Comment: Question About Deprecated Syntax

The fix itself looks correct - `parseLatestBy()` properly dispatches to both syntax handlers.

However, I noticed something worth discussing:

### Documentation vs Implementation

The [official QuestDB documentation](https://questdb.com/docs/reference/sql/latest-on/) only shows the `LATEST ON ... PARTITION BY` syntax:

```sql
SELECT * FROM balances
WHERE balance > 800
LATEST ON ts PARTITION BY cust_id;
```

The `LATEST BY` syntax is not mentioned in the docs and is marked as `LATEST_BY_DEPRECATED` in the codebase.

### Question

This PR extends support for the deprecated `LATEST BY` syntax to work after subqueries. The test case uses:

```sql
select * from (select * from ... latest by symbol, venue) latest by symbol
```

But per documentation, the equivalent would be:

```sql
select * from (select * from ... LATEST ON ts PARTITION BY symbol, venue) 
LATEST ON ts PARTITION BY symbol
```

**Is the intention to:**
1. Continue supporting the deprecated syntax indefinitely for backwards compatibility?
2. Or should users be encouraged to migrate to `LATEST ON ... PARTITION BY`?

If (1), this fix makes sense. If (2), perhaps the fix should only allow the documented syntax after subqueries, and the error message could guide users toward the preferred syntax.

Just wanted to raise this for consideration - the fix is technically correct either way.

bluestreak01:
cc @javier

PR #6542, opened 2025-12-16
perf(sql): speed up avg() functions in non-keyed GROUP BY queries

puzpuzpuz:
Removes redundant division from `avg()` function in case of parallel non-keyed GROUP BY, e.g.
```sql
SELECT AVG(price) FROM trades;
```

Native `avg*Acc()` functions were using `avg += (v - avg) / c` in each iteration while the outer `Avg*VectorAggregateFunction` Java function was multiplying the average and the count before storing them in the partial result (`sum.add(value * count)`).

### Benchmarks

Test box: Ryzen 7900x, 64GB RAM, Ubuntu 24.04

```sql
CREATE TABLE IF NOT EXISTS 'x' ( 
        d DOUBLE,
        ts TIMESTAMP
) TIMESTAMP(ts) PARTITION BY DAY;

INSERT INTO x SELECT
    rnd_double() * 20 + 10 AS d,
    generate_series as ts
  FROM generate_series('2025-01-01', '2025-01-02', '172u');

-- before: 90ms
-- after: 74ms
SELECT AVG(d) FROM x;
```

With ClickBench's `hits` table the difference is:
```sql
-- before: 19ms
-- after: 15ms
SELECT AVG(UserID) FROM hits;
```


bluestreak01 (dismissed):

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/vect/AvgDoubleVectorAggregateFunction.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/vect/AvgIntVectorAggregateFunction.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/vect/AvgLongVectorAggregateFunction.java|5|5|100.00%|

PR #6550, opened 2025-12-17
fix(sql): allow aggregate functions inside CASE expressions

### LINKED ISSUE #6549: SAMPLE BY query gives Aggregate function cannot be passed as an argument error
### To reproduce

On demo the following query
```sql
SELECT
  timestamp, symbol,
  first(price),
  last(price),
  CASE
   when symbol = 'BTC-USD' THEN first(price)
   else last(price)
  END
FROM (trades where timestamp in today())
SAMPLE BY 1h
ALIGN TO CALENDAR WITH OFFSET '-00:15';
```
gives `Aggregate function cannot be passed as an argument`. We should improve virtual model rewrite to extract expressions like the above CASE into an outer virtual model.

There is a similar limitation in WINDOW JOIN:
```sql
SELECT
    t.symbol,
    t.timestamp,
    t.side,
    t.price AS exec_price,
    avg(p.bid_price) AS avg_bid,
    avg(p.ask_price) AS avg_ask,
    avg(p.ask_price - p.bid_price) AS avg_spread,
    -- Slippage: did you pay more than ask (buy) or receive less than bid (sell)?
    CASE 
        WHEN t.side = 'buy' THEN t.price - avg(p.ask_price)
        ELSE avg(p.bid_price) - t.price
    END AS slippage
FROM trades t
WINDOW JOIN prices p
    ON (t.symbol = p.symbol)
    RANGE BETWEEN 5 minutes PRECEDING AND 0 seconds PRECEDING
    EXCLUDE PREVAILING
ORDER BY t.timestamp;
```
gives `WINDOW join cannot reference right table non-aggregate column: p.bid_price`.

### QuestDB version:

latest master

Linux

ext4

### Full Name:

Andrei Pechkurov

### Affiliation:

QuestDB

- [x] Yes, I have

### Additional context

_No response_

bluestreak01:
## Summary

Fixes #6549

Queries with aggregate functions inside `CASE` expressions were failing with:
```
Aggregate function cannot be passed as an argument
```

For example, this query now works:
```sql
SELECT
  timestamp, symbol,
  first(price),
  last(price),
  CASE
    WHEN symbol = 'BTC-USD' THEN first(price)
    ELSE last(price)
  END
FROM trades
SAMPLE BY 1h
```

## Root Cause

`checkForChildAggregates()` in `SqlOptimiser` only traversed `lhs` and `rhs` children of expression nodes. However, `CASE` expressions (and other functions with 3+ arguments) store their arguments in `node.args`, not `lhs`/`rhs`. This caused the optimizer to miss aggregate functions nested inside these expressions.

## Fix

Modified `checkForChildAggregates()` to also iterate through `node.args` when `node.paramCount >= 3`, following the same pattern already used in `emitAggregatesAndLiterals()`.

### DIFF
--- a/core/src/test/java/io/questdb/test/griffin/GroupByFunctionTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/GroupByFunctionTest.java
@@ -153,32 +153,54 @@ public class GroupByFunctionTest extends AbstractCairoTest {
     // https://github.com/questdb/questdb/issues/6549
     @Test
     public void testCaseWithAggregatesInGroupBy() throws Exception {
-        assertQuery(
-                """
-                        sym\tsum\tavg\tresult
-                        a\t60.0\t20.0\t60.0
-                        b\t90.0\t45.0\t45.0
-                        """,
-                "SELECT sym, sum(val), avg(val), " +
-                        "CASE WHEN sym = 'a' THEN sum(val) ELSE avg(val) END as result " +
-                        "FROM test " +
-                        "GROUP BY sym " +
-                        "ORDER BY sym",
-                "create table test as (" +
-                        "  select 'a' as sym, 10.0 as val from long_sequence(1)" +
-                        "  union all" +
-                        "  select 'a' as sym, 20.0 as val from long_sequence(1)" +
-                        "  union all" +
-                        "  select 'a' as sym, 30.0 as val from long_sequence(1)" +
-                        "  union all" +
-                        "  select 'b' as sym, 40.0 as val from long_sequence(1)" +
-                        "  union all" +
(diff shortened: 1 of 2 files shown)


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|18|18|100.00%|

RaphDal:
@coderabbitai review

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|18|18|100.00%|

Issue #6560, opened 2025-12-18
caching `touch()` queries and re-using the plans causing segfault


nwoolmer:
### To reproduce

```sql
CREATE TABLE t1 AS (
  SELECT generate_series as t, 
         (generate_series % 123456)::int as i 
  FROM generate_series('2025-12-25', '2025-12-26', '1s')
) PARTITION BY DAY WAL;


select touch(select * from t1);

-- works fine

alter table t1 add column f float;

select touch(select * from t1);

-- cached query plan cannot be used because table schema has changed [table=t1', expectedTableId=60408, actualTableId=60408, expectedMetadataVersion=0, actualMetadataVersion=1]
-- get an error, and result, but still ok

select touch(select * from t1);

-- Invalid JSON response from the server: SyntaxError: JSON.parse: unexpected end of data at line 1 column 1 of the JSON data

-- server segfaulted and process killed
```

Worse, this persists even after restart. If you try to run `select touch(select * from t1);` again, it will crash.

server log:

```
2025-12-18T17:26:06.509789Z I i.q.g.e.QueryProgress exe [id=39, sql=`select touch(select * from t1)`, principal=admin, cache=false, jit=false]
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x000000011283a8a0, pid=59295, tid=40963
#
# JRE version: OpenJDK Runtime Environment Corretto-17.0.11.9.1 (17.0.11+9) (build 17.0.11+9-LTS)
# Java VM: OpenJDK 64-Bit Server VM Corretto-17.0.11.9.1 (17.0.11+9-LTS, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, bsd-aarch64)
# Problematic frame:
# J 1949 c2 sun.misc.Unsafe.getByte(J)B jdk.unsupported@17.0.11 (8 bytes) @ 0x000000011283a8a0 [0x000000011283a880+0x0000000000000020]
#
# No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/nick/IdeaProjects/questdb-4/hs_err_pid59295.log
#
# If you would like to submit a bug report, please visit:
#   https://github.com/corretto/corretto-17/issues/
```

stack trace:

```
[...]

PR #6561, opened 2025-12-18
fix(sql): segfault when using `touch()` on a table with a new empty column

nwoolmer:
Fixes https://github.com/questdb/questdb/issues/6560

When a new column is added to a table, it is initially empty. When you call `frame.getPageAddress` to get the column's base memory address, this empty column is uninitialised, and therefore returns an address of 0.

Unfortunately, touch() did not check this possibility, so would use the nullptr.

Additionally added checks for indexes. We do not have a repro for this, and _theoretically_ the prior issue should hit before the index is tried, but makes sense to add these for safety.


nwoolmer:
@coderabbitai review

bluestreak01 (approved):

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/TouchTableFunctionFactory.java|2|2|100.00%|

PR #6564, opened 2025-12-19
fix(core): backup restore - rebuild symbol index files inside partitions

amunra:
# Bug Description

Symbol columns can be indexed. In such case an bitmap index is created for each indexed column in each partition.

## Problem Statement

After checkpoint backup restore, queries can fail with errors like "binary is outside of file boundary" because bitmap index files (`.k` and `.v`) reference row IDs that don't exist in the restored partition data.

### Root Cause

1. **Checkpoint** records partition `row_count` at a point in time
2. **Backup** runs later, reading `.k`/`.v` files from live partitions
3. Between checkpoint and backup, new rows may be added, updating the index files
4. The backed-up index files contain references to rows beyond the checkpoint's `row_count`
5. After restore, queries using these indexes try to access non-existent rows

### Critical Data Corruption Scenario

Simply patching the `max_value` header field is **not sufficient** to fix this problem. Consider this scenario:

1. Checkpoint records `row_count = 1000`
2. Before backup, new rows are added, including row ID 1010
3. The `.v` file now contains a reference to row ID 1010 for some symbol key
4. Backup is created with the corrupted index
5. After restore, the partition has only 1000 rows
6. Later, the partition grows naturally and a **new, different** row 1010 is created
7. Queries using the index will now return **wrong data** - the index points to the new row 1010, but the symbol value at that row is different from what the index entry was created for

This is a **silent data corruption** bug - queries return incorrect results without any error.

### Torn Key Entry Reads

While backing up the `.k` file, we could read a "torn" key entry where the writer is in the middle of updating. Each key entry has two fields that must match:
- `valueCount` (offset 0 in entry)
- `countCheck` (offset 24 in entry)

If the backup reads a key entry while the writer is updating it, we may capture a state where `valueCount != countCheck`. This would mean the backed-up `.k` file is **corrupt** - the atomicity check will fail when the index is read after restore.

# This fix

Rebuilds the .k/.v files to avoid these issues.
It uses a thread pool to mitigate the performance impact.

# Issue Linkage

Tandem for the `backup_ent` branch.

Tandem PR: https://github.com/questdb/questdb-enterprise/pull/840

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
+++ b/core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
@@ -70,10 +70,9 @@ import static io.questdb.std.datetime.DateLocaleFactory.EN_LOCALE;
  * Used by both DatabaseCheckpointAgent and BackupRestoreAgent.
  */
 public class TableSnapshotRestore implements QuietCloseable {
-    private static final int BITMAP_INDEX_RESTORE_THREAD_COUNT = Math.max(4, Math.min(16, Runtime.getRuntime().availableProcessors()));
     private static final Log LOG = LogFactory.getLog(TableSnapshotRestore.class);
     private final CairoConfiguration configuration;
-    private final ExecutorService executor = Executors.newFixedThreadPool(BITMAP_INDEX_RESTORE_THREAD_COUNT);
+    private final ExecutorService executor;
     private final FilesFacade ff;
     private final ObjList<Future<?>> futures = new ObjList<>();
     private final Utf8StringSink utf8Sink = new Utf8StringSink();
@@ -89,6 +88,11 @@ public class TableSnapshotRestore implements QuietCloseable {
     public TableSnapshotRestore(CairoConfiguration configuration) {
         this.configuration = configuration;
         this.ff = configuration.getFilesFacade();
+        int threadCount = Math.max(
+                configuration.getCheckpointRecoveryThreadpoolMin(),
+                Math.min(configuration.getCheckpointRecoveryThreadpoolMax(), Runtime.getRuntime().availableProcessors())
+        );
+        this.executor = Executors.newFixedThreadPool(threadCount);
     }
 
     @Override
--- a/core/src/main/java/io/questdb/PropServerConfiguration.java
+++ b/core/src/main/java/io/questdb/PropServerConfiguration.java
@@ -182,6 +182,8 @@ public class PropServerConfiguration implements ServerConfiguration {
     private final long cairoWriteBackOffTimeoutOnMemPressureMs;
     private final boolean checkpointRecoveryEnabled;
     private final boolean checkpointRecoveryRebuildColumnIndexes;
+    private final int checkpointRecoveryThreadpoolMax;
+    private final int checkpointRecoveryThreadpoolMin;
     private final String checkpointRoot;
     private final PropSqlExecutionCircuitBreakerConfiguration circuitBreakerConfiguration = new PropSqlExecutionCircuitBreakerConfiguration();
     private final int circuitBreakerThrottle;
@@ -952,6 +954,19 @@ public class PropServerConfiguration implements ServerConfiguration {
                 )
         );
         this.checkpointRecoveryRebuildColumnIndexes = getBoolean(properties, env, PropertyKey.CAIRO_CHECKPOINT_RECOVERY_REBUILD_COLUMN_INDEXES, false);
+        int checkpointRecoveryThreadpoolMinRaw = getInt(properties, env, PropertyKey.CAIRO_CHECKPOINT_RECOVERY_THREADPOOL_MIN, 4);
+        int checkpointRecoveryThreadpoolMaxRaw = getInt(properties, env, PropertyKey.CAIRO_CHECKPOINT_RECOVERY_THREADPOOL_MAX, 12);
+        if (checkpointRecoveryThreadpoolMinRaw < 2 || checkpointRecoveryThreadpoolMinRaw > 32) {
+            throw new ServerConfigurationException(PropertyKey.CAIRO_CHECKPOINT_RECOVERY_THREADPOOL_MIN.getPropertyPath() + " must be between 2 and 32");
+        }
+        if (checkpointRecoveryThreadpoolMaxRaw < 2 || checkpointRecoveryThreadpoolMaxRaw > 32) {
+            throw new ServerConfigurationException(PropertyKey.CAIRO_CHECKPOINT_RECOVERY_THREADPOOL_MAX.getPropertyPath() + " must be between 2 and 32");
+        }
+        if (checkpointRecoveryThreadpoolMinRaw > checkpointRecoveryThreadpoolMaxRaw) {
+            throw new ServerConfigurationException(PropertyKey.CAIRO_CHECKPOINT_RECOVERY_THREADPOOL_MIN.getPropertyPath() + " must be less than or equal to " + PropertyKey.CAIRO_CHECKPOINT_RECOVERY_THREADPOOL_MAX.getPropertyPath());
+        }
+        this.checkpointRecoveryThreadpoolMin = checkpointRecoveryThreadpoolMinRaw;
+        this.checkpointRecoveryThreadpoolMax = checkpointRecoveryThreadpoolMaxRaw;
         this.devModeEnabled = getBoolean(properties, env, PropertyKey.DEV_MODE_ENABLED, false);
 
         int cpuAvailable = Runtime.getRuntime().availableProcessors();
@@ -3067,6 +3082,16 @@ public class PropServerConfiguration implements ServerConfiguration {
         @Override
         public boolean getCheckpointRecoveryRebuildColumnIndexes() {
             return checkpointRecoveryRebuildColumnIndexes;
+        }
+
+        @Override
+        public int getCheckpointRecoveryThreadpoolMax() {
+            return checkpointRecoveryThreadpoolMax;
+        }
+
+        @Override
+        public int getCheckpointRecoveryThreadpoolMin() {
+            return checkpointRecoveryThreadpoolMin;
         }
 
         @Override
--- a/core/src/main/java/io/questdb/PropertyKey.java
+++ b/core/src/main/java/io/questdb/PropertyKey.java
@@ -37,6 +37,8 @@ public enum PropertyKey implements ConfigPropertyKey {
     CAIRO_LEGACY_SNAPSHOT_RECOVERY_ENABLED("cairo.snapshot.recovery.enabled"),
     CAIRO_CHECKPOINT_RECOVERY_ENABLED("cairo.checkpoint.recovery.enabled"),
     CAIRO_CHECKPOINT_RECOVERY_REBUILD_COLUMN_INDEXES("cairo.checkpoint.recovery.rebuild.column.indexes"),
+    CAIRO_CHECKPOINT_RECOVERY_THREADPOOL_MIN("cairo.checkpoint.recovery.threadpool.min"),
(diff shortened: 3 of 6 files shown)


ideoma:
@coderabbitai review

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
```diff
@@ -118,6 +133,34 @@ public void copyMetadataFiles(
...
+                Thread.currentThread().interrupt();
+                final CairoException ex = CairoException.critical(0)
+                        .put("parallel task interrupted")
+                        .put(": ")
+                        .put(e.getMessage());
+                ex.initCause(e);
+                throw ex;
+            } catch (ExecutionException e) {
+                Throwable cause = e.getCause();
+                final CairoException ex = CairoException.critical(0)
```
ideoma, line 158: Please log the cause as an error or critical here, before re-throwing. I suspect re-throwing will not save the stack trace of the original exception.

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
```diff
@@ -118,6 +133,34 @@ public void copyMetadataFiles(
...
+                        .put("error in parallel task")
+                        .put(": ")
+                        .put(cause.getMessage());
```
ideoma: `cause` can be null potentially

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
```diff
@@ -365,6 +416,231 @@ private void copyFile(
...
     }
 
+    private void rebuildBitmapIndexForNativePartition(int pathTableLen, int columnCount, long partitionTimestamp, long partitionRowCount, long partitionNameTxn, String tablePathStr, int partitionBy, int timestampType) {
+        for (int colIdx = 0; colIdx < columnCount; colIdx++) {
+            if (!tableMetadata.isColumnIndexed(colIdx)) {
+                continue;
+            }
+
+            // Currently only symbol columns can be indexed
+            assert ColumnType.isSymbol(tableMetadata.getColumnType(colIdx))
```
ideoma: The deleted symbol column can potentially have an indexed flag left as true. Merge this assert with the if above, don't throw, ignore

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
```diff
@@ -365,6 +416,231 @@ private void copyFile(
...
+                    : "Only symbol columns can be indexed, found: " + ColumnType.nameOf(tableMetadata.getColumnType(colIdx));
+
+            int writerIndex = tableMetadata.getWriterIndex(colIdx);
+            long columnNameTxn = columnVersionReader.getColumnNameTxn(partitionTimestamp, writerIndex);
+            long columnTopRaw = columnVersionReader.getColumnTop(partitionTimestamp, writerIndex);
+
+            // Skip if column doesn't exist in this partition (columnTopRaw = -1)
+            if (columnTopRaw < 0 || columnTopRaw >= partitionRowCount) {
```
ideoma: I'd say it should be
```suggestion
            if (columnTopRaw >= partitionRowCount) {
```

-1 may mean something different than no data, unless you can find a relevant example

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
```diff
@@ -365,6 +416,231 @@ private void copyFile(
...
+            if (!tableMetadata.isColumnIndexed(colIdx)) {
+                continue;
+            }
+
+            // Currently only symbol columns can be indexed
+            assert ColumnType.isSymbol(tableMetadata.getColumnType(colIdx))
+                    : "Only symbol columns can be indexed, found: " + ColumnType.nameOf(tableMetadata.getColumnType(colIdx));
+
+            int writerIndex = tableMetadata.getWriterIndex(colIdx);
+            long columnNameTxn = columnVersionReader.getColumnNameTxn(partitionTimestamp, writerIndex);
```
ideoma: This should fall back to default column name txn. See `TableReader.reloadColumnAt()`:

```
            long columnTxn = versionRecordIndex > -1 ? columnVersionReader.getColumnNameTxnByIndex(versionRecordIndex) : -1;
            if (columnTxn == -1) {
                // When a column is added, a column version will have txn number for the partition
                // where it's added. It will also have the txn number in the [default] partition
                columnTxn = columnVersionReader.getDefaultColumnNameTxn(writerIndex);
            }
```

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
```diff
@@ -365,6 +416,231 @@ private void copyFile(
...
+            long columnTopRaw = columnVersionReader.getColumnTop(partitionTimestamp, writerIndex);
```
ideoma: nit: redundant variable, can be simply `columnTop`

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
```diff
@@ -365,6 +416,231 @@ private void copyFile(
...
+
+            long columnTop = columnTopRaw;
+
+            // Capture variables for lambda
+            final String columnName = tableMetadata.getColumnName(colIdx);
+            final int indexBlockCapacity = tableMetadata.getIndexBlockCapacity(colIdx);
+            final long fPartitionTimestamp = partitionTimestamp;
+            final long fPartitionNameTxn = partitionNameTxn;
+            final long fPartitionRowCount = partitionRowCount;
+            final long fColumnTop = columnTop;
```
ideoma: nit: most of these are redundant, the variables you declared are effectively final, or you can explicitly declare them as final from the very beginning, e.g.:

```Java
final int writerIndex = tableMetadata.getWriterIndex(colIdx);
```

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
```diff
@@ -365,6 +416,231 @@ private void copyFile(
...
+        final int columnCount = tableMetadata.getColumnCount();
+        final String tablePathStr = tablePath.toString();
+
+        // Iterate through partitions (or single default partition for non-partitioned tables)
+        int partitionCount = isPartitioned ? txWriter.getPartitionCount() : 1;
+
+        for (int partitionIndex = 0; partitionIndex < partitionCount; partitionIndex++) {
+            long partitionTimestamp;
+            long partitionRowCount;
+            long partitionNameTxn;
```
ideoma: nit: can be declared final here

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
```diff
@@ -118,6 +133,34 @@ public void copyMetadataFiles(
...
+    public void finalizeParallelTasks() {
+        if (futures.size() > 0) {
+            LOG.info().$("awaiting ").$(futures.size()).$(" parallel tasks to complete").I$();
+        }
+
+        for (int i = 0, n = futures.size(); i < n; i++) {
+            try {
+                futures.getQuick(i).get();
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
```
ideoma: nit: the overall codebase approach to `InterruptedException` is just to ignore them. See `Os.sleep()` code. So I'd re-throw them as CairoException here without additional decorations

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
```diff
@@ -228,6 +273,11 @@ public void rebuildTableFiles(
             // when symbol files are copied while written to. We need to rebuild them.
```
bluestreak01, line 272: this block here is just opens and closes the files, why do we need it?

```
                dstPath.trimTo(dstSeqLen).concat(TableUtils.META_FILE_NAME);
                memFile.smallFile(ff, dstPath.$(), MemoryTag.MMAP_SEQUENCER_METADATA);
                dstPath.trimTo(dstSeqLen);
                openSmallFile(ff, dstPath, dstSeqLen, memFile, TXNLOG_FILE_NAME_META_INX, MemoryTag.MMAP_TX_LOG);
```

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
```diff
@@ -26,27 +26,38 @@
 
 import io.questdb.cairo.mv.MatViewDefinition;
 import io.questdb.cairo.mv.MatViewState;
+import io.questdb.cairo.sql.RecordMetadata;
```
bluestreak01, line 29: clean this file up please, it is full of warnings

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
amunra: It should already do:
```
    public long getColumnNameTxn(long partitionTimestamp, int columnIndex) {
        int versionRecordIndex = getRecordIndex(partitionTimestamp, columnIndex);
        return versionRecordIndex > -1 ? cachedColumnVersionList.getQuick(versionRecordIndex + COLUMN_NAME_TXN_OFFSET) : getDefaultColumnNameTxn(columnIndex);
    }
```

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
amunra: The `-1` value means that the column was added after the partition was created. I've added a comment to clarify.
If I remove this logic, the tests break.

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
amunra: Done.

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
amunra: Done.

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
amunra: Done.

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
amunra: Done.

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
amunra: Done.

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
amunra: Debris. Removed.

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
amunra: Done. Cleaned up the warnings. Addressed duplicated code.

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
amunra: Done.

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
amunra: Done.

core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
ideoma: agree

PR #6571, opened 2025-12-21
Improve error message for unbalanced parentheses and added test for it

Gautam-aman:
### Summary
Improves the error message for unbalanced parentheses.

### Previous message
`unbalanced (`

### New message
`unbalanced parenthesis: missing closing ')'`

### Additions
- Included regression test `testUnbalancedParenthesisMessage`
- Verified using curl:
  curl "http://localhost:9000/exec?query=select%20("

### Motivation
Clearer diagnostic guidance for malformed expressions and better developer UX.


bluestreak01:
@Gautam-aman it is generally a good idea to critically verify what you're submitting, including running tests before submitting.

PR #6572, opened 2025-12-22
fix(sql): fix false timestamp ordering error with nested SAMPLE BY and ORDER BY

bluestreak01:
## Problem

Nested `SAMPLE BY` queries with an `ORDER BY` clause in between incorrectly failed with:
```
ASC order over TIMESTAMP column is required but not provided
```

Example of failing query:
```sql
SELECT timestamp, count()
FROM (
    (
        SELECT timestamp, symbol, count(bids[1][1]) as total
        FROM eq_equities_market_data
        WHERE symbol = 'HSBC'
        SAMPLE BY 10s
    )
    ORDER BY timestamp
)
SAMPLE BY 10m
```

## Root Cause

When `SAMPLE BY` needs ordered timestamp data, it pushes `timestampRequired=true` onto a context stack. This flag propagates to nested models and throws an error if ASC timestamp order isn't guaranteed.

However, when there's an `ORDER BY` clause between the outer and inner `SAMPLE BY`, the inner model's ordering is irrelevant since `ORDER BY` will re-sort the data anyway.

## Fix

When an `ORDER BY` clause is present, push `timestampRequired=false` to signal that nested models don't need to maintain timestamp ordering.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
+++ b/core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
@@ -4669,11 +4669,17 @@ public class SqlCodeGenerator implements Mutable, Closeable {
         // Remember the last model with non-empty ORDER BY as we descend through nested models.
         // We need the ORDER BY clause in the Markout Horizon Join optimization, but it's stored
         // several levels up from the model that holds the join clause.
+        boolean pushed = false;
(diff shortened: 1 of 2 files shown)


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|

PR #6589, opened 2026-01-02
fix(sql): allow redundant columns in SELECT DISTINCT with wildcard #6521

### LINKED ISSUE #6521: syntax error when DISTINCT SELECT contains both star and explicit column list
### To reproduce

This works on the demo:
```sql
select symbol from core_price where timestamp in '2025-12-08T10';
```

This also works:
```sql
select distinct symbol from core_price where timestamp in '2025-12-08T10';
```

This works too:
```sql
select distinct * from core_price where timestamp in '2025-12-08T10';
```

This works as well (one of the `symbol` columns gets a sythetic suffix)
```sql
select symbol, * from core_price where timestamp in '2025-12-08T10';
```

but this does not work:
```sql
select distinct symbol, * from core_price where timestamp in '2025-12-08T10';
```
The query errors out with `Duplicate column [name=symbol]`

As a QuestDB user, I consider this to be a surprising inconsistency. 

### QuestDB version:

9.2.2

jayagrawal254:
This PR resolves an issue where queries using SELECT DISTINCT would fail with a Duplicate column error if a column was explicitly listed alongside a wildcard (e.g., SELECT DISTINCT symbol, *).

While QuestDB's standard SELECT allows duplicate column names in the result set, the DISTINCT translation logic was too strict. When the compiler attempted to project columns into the outerVirtualModel for distinct processing, it encountered name collisions during wildcard expansion.

This change introduces a de-duplication check in SqlCodeGenerator.java. It ensures that if a column is already present in the outerVirtualModel, redundant references (often caused by *) are ignored instead of throwing an exception.

Fixes #6521

Added a check in createSelectColumn to verify if a column already exists in outerVirtualModel before adding it.


jerrinot:
Hello @jayagrawal254, many thanks for the pull request! 

Please add a test (which is failing before the fix and passing after) to this PR. It's to make sure it won't regress again at some point in the future. Thank you!

bluestreak01:
tests two weeks overdue

Issue #6535, opened 2025-12-14
class cast exception with invalid SQL


jerrinot:
### To reproduce

found via fuzzing:
```sql
SELECT - sym AS a2 FROM t1 a4 WHERE (.a4.col4 = 0);
```

```2025-12-14T18:00:45.417106Z C i.q.c.h.p.JsonQueryProcessorState [135291470082] internal error [ex=
java.lang.ClassCastException: class io.questdb.griffin.CharacterStore$NameAssemblerCharSequence cannot be cast to class io.questdb.std.GenericLexer$FloatingSequence (io.questdb.griffin.CharacterStore$NameAssemblerCharSequence and io.questdb.std.GenericLexer$FloatingSequence are in module io.questdb@9.2.3-SNAPSHOT of loader 'app')
	at io.questdb.griffin.ExpressionParser.parseExpr(ExpressionParser.java:332)
	at io.questdb.griffin.SqlParser.expr(SqlParser.java:4292)
	at io.questdb.griffin.SqlParser.expr(SqlParser.java:4303)
	at io.questdb.griffin.SqlParser.parseFromClause(SqlParser.java:2343)
	at io.questdb.griffin.SqlParser.parseDml0(SqlParser.java:2099)
	at io.questdb.griffin.SqlParser.parseDml(SqlParser.java:1880)
	at io.questdb.griffin.SqlParser.parseSelect(SqlParser.java:2972)
	at io.questdb.griffin.SqlParser.parse(SqlParser.java:4329)
	at io.questdb.griffin.SqlCompilerImpl.compileExecutionModel(SqlCompilerImpl.java:2564)
	at io.questdb.griffin.SqlCompilerImpl.compileUsingModel(SqlCompilerImpl.java:3299)
	at io.questdb.griffin.SqlCompilerImpl.compileInner(SqlCompilerImpl.java:2668)
	at io.questdb.griffin.SqlCompilerImpl.compile(SqlCompilerImpl.java:396)
	at io.questdb.cairo.pool.SqlCompilerPool$C.compile(SqlCompilerPool.java:141)
[...]
```

The SQL makes no sense, but it should return `SqlException`, not `ClassCastException` 

### QuestDB version:

c2343aad

### OS, in case of Docker specify Docker and the Host OS:

NA

### File System, in case of Docker specify Host File System:

NA

### Full Name:

Jaromir Hamala

### Affiliation:

QuestDB

[...]

PR #6592, opened 2026-01-03
fix(sql): Handling of unexpected dot after the left parenthesis (#6535)

etojesaandro:
fix(sql): Handling of unexpected dot after the left parenthesis (#6535)

### DIFF
--- a/core/src/main/java/io/questdb/griffin/ExpressionParser.java
+++ b/core/src/main/java/io/questdb/griffin/ExpressionParser.java
@@ -1618,7 +1618,7 @@ public class ExpressionParser {
                                 throw SqlException.$(lastPos, "qualifier expected");
                             }
                             if (en.type != ExpressionNode.LITERAL && en.type != ExpressionNode.CONSTANT) {
-                                throw SqlException.$(lexer.getContent().toString().lastIndexOf('.', lastPos), "unexpected dot");
+                                throw SqlException.$(lastPos, "unexpected dot");
                             }
                             // two possibilities here:
                             // 1. 'a.b'
@@ -1640,7 +1640,7 @@ public class ExpressionParser {
                                 cse.put(en.token).put(GenericLexer.unquoteIfNoDots(tok));
                                 opStack.push(expressionNodePool.next().of(
                                         ExpressionNode.LITERAL, cse.toImmutable(), Integer.MIN_VALUE, en.position));
-                            } else if (en.token instanceof GenericLexer.FloatingSequence) {
+                            } else {
                                 final GenericLexer.FloatingSequence fsA = (GenericLexer.FloatingSequence) en.token;
                                 // vanilla 'a.b', just concat tokens efficiently
                                 fsA.setHi(lexer.getTokenHi());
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/questdb/griffin/ExpressionParser.java
```diff
@@ -1637,10 +1637,12 @@ the other op. This is only a problem for ambiguous operators like minus('-'),
                                 cse.put(en.token).put(GenericLexer.unquoteIfNoDots(tok));
                                 opStack.push(expressionNodePool.next().of(
                                         ExpressionNode.LITERAL, cse.toImmutable(), Integer.MIN_VALUE, en.position));
-                            } else {
+                            } else if (en.token instanceof GenericLexer.FloatingSequence) {
                                 final GenericLexer.FloatingSequence fsA = (GenericLexer.FloatingSequence) en.token;
                                 // vanilla 'a.b', just concat tokens efficiently
                                 fsA.setHi(lexer.getTokenHi());
+                            } else {
```
jerrinot: This looks suspicious - it throws an exception depending on the impl type of a previous token. Are you sure it is sound? It feels to me that an impl type of a previous token should not decide whether we throw or not. It might drive optimizations (e.g. when it's floating sequence), but not the entire control flow. Can you please try multiple similar test cases - not just the one from the original report. Example: [...] WHERE (.a4 = 0); , [...]  [...] WHERE (.a4. = 0); or even just  [...] WHERE (. = 0);

etojesaandro: Got your comment. Added some more unit tests and changed the way of handling. By the way, I suggest to keep the type checking there to avoid ClassCastException and handle more appropriate ones.

core/src/main/java/io/questdb/griffin/ExpressionParser.java
```diff
@@ -1617,6 +1617,9 @@ the other op. This is only a problem for ambiguous operators like minus('-'),
                             if (en == null) {
                                 throw SqlException.$(lastPos, "qualifier expected");
                             }
+                            if (en.type != ExpressionNode.LITERAL && en.type != ExpressionNode.CONSTANT) {
+                                throw SqlException.$(lexer.getContent().toString().lastIndexOf('.', lastPos), "unexpected dot");
```
mtopolnik: This should be just:

```suggestion
                                throw SqlException.$(lastPos, "unexpected dot");
```

core/src/main/java/io/questdb/griffin/ExpressionParser.java
```diff
@@ -1637,7 +1640,7 @@ the other op. This is only a problem for ambiguous operators like minus('-'),
                                 cse.put(en.token).put(GenericLexer.unquoteIfNoDots(tok));
                                 opStack.push(expressionNodePool.next().of(
                                         ExpressionNode.LITERAL, cse.toImmutable(), Integer.MIN_VALUE, en.position));
-                            } else {
+                            } else if (en.token instanceof GenericLexer.FloatingSequence) {
```
mtopolnik: This replaces a loud `ClassCastException` with silently taking no action at all. We prefer the loud approach -- for example, this issue got raised thanks to that.

core/src/main/java/io/questdb/griffin/ExpressionParser.java
etojesaandro: With your change the actual position of an error is shifted to incorrect one. It is covered by the ExpressionParserTest.testLeftParenthesisUnexpectedDot()

Issue #6536, opened 2025-12-14
NPE with invalid SQL


jerrinot:
### To reproduce

found via fuzzing:
```sql
SELECT ts, col6 AS a1, CASE WHEN col5 <> '192.168.150.217' THEN 754326 WHEN col1 = false THEN '' WHEN ts <= ''''qocvwy'''' THEN col4 ELSE sym END, col4, -109639495026, '1986-05-10T21:39:17Z', '2026-07-22T10:32:41.213795Z',) col2 AS a2, 'p', 0.0 AS a3 FROM t1 a4 LT JOIN t2 a5 ON a4.timestamp = a5.timestamp LEFT OUTER JOIN t3 ON a4.col3 = t3.col3 LEFT col7 <= false OR a4.sym <> col2 GROUP BY a5.col4, col4, a5.col7, col10 ORDER BY col7 ASC NULLS FIRST, a5.col10 DESC, a5.col3 DESC NULLS FIRST, 3 DESC LIMIT 194
```

```
2025-12-14T18:36:53.377325Z E i.q.t.f.s.SqlParserFuzzTest SQL: SELECT ts, col6 AS a1, CASE WHEN col5 <> '192.168.150.217' THEN 754326 WHEN col1 = false THEN '' WHEN ts <= ''''qocvwy'''' THEN col4 ELSE sym END, col4, -109639495026, '1986-05-10T21:39:17Z', '2026-07-22T10:32:41.213795Z',) col2 AS a2, 'p', 0.0 AS a3 FROM t1 a4 LT JOIN t2 a5 ON a4.timestamp = a5.timestamp LEFT OUTER JOIN t3 ON a4.col3 = t3.col3 LEFT col7 <= false OR a4.sym <> col2 GROUP BY a5.col4, col4, a5.col7, col10 ORDER BY col7 ASC NULLS FIRST, a5.col10 DESC, a5.col3 DESC NULLS FIRST, 3 DESC LIMIT 194
2025-12-14T18:36:53.377397Z E i.q.t.f.s.SqlParserFuzzTest 
java.lang.NullPointerException: Cannot invoke "java.lang.CharSequence.length()" because "seq" is null
	at io.questdb.std.Chars.indexOfLastUnquoted(Chars.java:847)
	at io.questdb.griffin.SqlParser.createColumnAlias(SqlParser.java:468)
	at io.questdb.griffin.SqlParser.generateColumnAlias(SqlParser.java:681)
[...]

PR #6594, opened 2026-01-04
Handle missing expression in ORDER BY and unbalanced right parentheses

mghildiy:
Change for issue #6536:

1. If it is balanced right parentheses, accept it.
2. If it's an unablanced right parentheses, throw appropriate exception.
3. Handle missing expression in ORDER BY clause.


nwoolmer:
Failure on `testUnexpectedErrorDuringSQLExecutionHandled`

PR #6595, opened 2026-01-04
chore(core): add awaitStartup() API to ServerMain

bluestreak01:
## Summary

Adds a public `awaitStartup()` method to `ServerMain` that waits for background startup tasks to complete, specifically the metadata cache and recent write tracker hydration.

### Problem

After `ServerMain.start()` returns, a background thread (`hydrateMetadataThread`) continues running to populate the metadata cache and recent write tracker. If DDL operations like `DROP TABLE` are executed immediately after startup, they can fail with:

```
CairoException: could not lock 'table_name' [reason='busyTableReaderMetaPool']
```

This happens because the hydration thread holds a `TableMetadataPool` lock while reading table metadata, creating a race condition with concurrent DDL operations.

### Solution

- Added `ServerMain.awaitStartup()` method that joins the hydration thread
- Added test `testAwaitStartupBeforeDropTable` that exercises this pattern

### Usage

```java
serverMain.start();
serverMain.awaitStartup();  // Wait for background hydration to complete
engine.execute("DROP TABLE ...");  // Safe to run DDL now
```

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -36,7 +36,7 @@
         <argLine>-ea -Dfile.encoding=UTF-8 -XX:+UseParallelGC</argLine>
         <test.exclude>None</test.exclude>
         <test.include>%regex[.*[^o].class]</test.include><!-- exclude module-info.class-->
-        <web.console.version>1.1.7</web.console.version>
+        <web.console.version>1.1.9</web.console.version>
         <qdbr.path>rust/qdbr</qdbr.path>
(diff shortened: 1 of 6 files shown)


glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 4 / 6 (66.67%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/ServerMain.java|4|6|66.67%|

PR #6605, opened 2026-01-07
chore(sql): incorrect INCLUDE PREVAILING handling in parallel WINDOW JOIN

puzpuzpuz:
`WindowJoinFuzzTest` was failing with 1547227975298147L, 1767737943575L seeds.

The bugfix is in the `binarySearchWithPrevailing` method in `WindowJoinTimeFrameHelper`: it would return an arbitrary row when finding an exact timestamp match during the binary search phase. When multiple rows share the same timestamp at the window boundary, this caused the iteration to start from a middle row instead of the first one, skipping rows from the aggregation.

Other changes:
* `WindowJoinTimeFrameHelper`: added frame navigation for prevailing returns. When returning a prevailing row from a different frame, added `jumpTo`/`open` calls before `recordAt` so that the `timeFrame` object has correct bounds. The caller uses `getTimeFrameRowHi()` for boundary checks during iteration.
* `AsyncWindowJoinRecordCursorFactory`: added missing `recordAt` calls. After `nextFrame` transitions in multiple reducers, added `recordAt` calls to properly switch the record to the new frame.

### DIFF
--- a/core/src/test/java/io/questdb/test/griffin/WindowJoinTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/WindowJoinTest.java
@@ -94,14 +94,14 @@ public class WindowJoinTest extends AbstractCairoTest {
                                 (
                                         select * from (
                                             select t.sym, t.price, t.ts, p.price
-                                            from trades t 
-                                            left join prices p 
-                                            on (t.sym = p.sym) 
+                                            from trades t
+                                            left join prices p
+                                            on (t.sym = p.sym)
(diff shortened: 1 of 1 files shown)


glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 8 / 11 (72.73%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinRecordCursorFactory.java|2|4|50.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/WindowJoinTimeFrameHelper.java|6|7|85.71%|

PR #6609, opened 2026-01-08
test(sql): fix VIEW fuzz test

jerrinot:
Fixes flaky `ViewFuzzTest`.
1. Compile views before registering them so dependent views see correct metadata
2. Skip LIMIT when ORDER BY is on a boolean column (provides no meaningful ordering)
3. Fix a bug in generateWhereClause() we could pick only BINARY/LONG128 columns -> RuntimeException (probably a leftover code)
4. Order by ALL non-binary columns (not just one) to eliminate tie-breaking ambiguity


jerrinot:
@coderabbitai pls review

PR #6612, opened 2026-01-08
chore(core): fix view table id is 0 after table registry reload without tables.d file

glasstiger:
All views had their table id set as 0 after a restart without `tables.d` file.
When there is no `tables.d` file, we take table id from `_meta` file.
However, views do not have this file.
Table id was defaulted to 0.
Fix is to parse the table id from the table dir name instead of trying to load from `_meta` file.

required by https://github.com/questdb/questdb-enterprise/pull/850

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TableNameRegistryStore.java
+++ b/core/src/main/java/io/questdb/cairo/TableNameRegistryStore.java
@@ -24,6 +24,7 @@
 
 package io.questdb.cairo;
 
+import io.questdb.cairo.view.ViewDefinition;
 import io.questdb.cairo.vm.Vm;
 import io.questdb.cairo.vm.api.MemoryCMR;
 import io.questdb.cairo.vm.api.MemoryMR;
@@ -309,22 +310,26 @@ public class TableNameRegistryStore extends GrowOnlyTableNameRegistryStore {
     }
 
     private int readTableId(Path path, CharSequence dirName, FilesFacade ff) {
-        path.of(configuration.getDbRoot()).concat(dirName).concat(META_FILE_NAME);
+        path.of(configuration.getDbRoot()).concat(dirName);
+        int pathLen = path.size();
+        path.concat(META_FILE_NAME);
         long fd = ff.openRO(path.$());
         if (fd < 1) {
-            // negative table id means WAL table
-            // normally we would hit this code path for views only, because they do not have _meta file
-            // views are considered to be WAL tables
-            return -getTableIdFromTableDir(dirName);
+            // check if it is a view
+            path.trimTo(pathLen).concat(ViewDefinition.VIEW_DEFINITION_FILE_NAME);
+            if (ff.exists(path.$())) {
+                // negative table id means WAL table,
+                // views are considered to be WAL tables
+                return -getTableIdFromTableDir(dirName);
+            }
+            return 0;
         }
 
         try {
             int tableId = ff.readNonNegativeInt(fd, TableUtils.META_OFFSET_TABLE_ID);
             if (tableId < 0) {
                 LOG.error().$("cannot read table id from metadata file [path=").$(path).I$();
-                // negative table id means WAL table
-                // we assume the table is WAL enabled
-                return -getTableIdFromTableDir(dirName);
+                return 0;
             }
             byte isWal = (byte) (ff.readNonNegativeInt(fd, TableUtils.META_OFFSET_WAL_ENABLED) & 0xFF);
             return isWal == 0 ? tableId : -tableId;


core/src/main/java/io/questdb/cairo/TableNameRegistryStore.java
```diff
@@ -312,14 +312,19 @@ private int readTableId(Path path, CharSequence dirName, FilesFacade ff) {
         path.of(configuration.getDbRoot()).concat(dirName).concat(META_FILE_NAME);
         long fd = ff.openRO(path.$());
         if (fd < 1) {
-            return 0;
+            // negative table id means WAL table
```
jerrinot: is there a way to double-check it's actually a VIEW? perhaps we should log when it's not a view AND there is no META file. we already log when the META file exists, but tableId is < 0 (=a non-sense ID). wdyt?

core/src/main/java/io/questdb/cairo/TableNameRegistryStore.java
```diff
@@ -312,14 +312,19 @@ private int readTableId(Path path, CharSequence dirName, FilesFacade ff) {
...
         }
 
         try {
             int tableId = ff.readNonNegativeInt(fd, TableUtils.META_OFFSET_TABLE_ID);
             if (tableId < 0) {
                 LOG.error().$("cannot read table id from metadata file [path=").$(path).I$();
-                return 0;
+                // negative table id means WAL table
+                // we assume the table is WAL enabled
+                return -getTableIdFromTableDir(dirName);
```
jerrinot: behaviour change: this now throws `NumericException` for all non-WAL tables. since they don't have ID at all in a dir name?

before: cannot read from META  -> always returns 0
after: cannot reead META -> for WAL tables it reads from the dir name, for non-WAL tables it throws

is this change intentional?

core/src/main/java/io/questdb/cairo/TableNameRegistryStore.java
glasstiger: now it is checking for the presence of a view definition file, and parsing table id from the dir name only as a fallback

core/src/main/java/io/questdb/cairo/TableNameRegistryStore.java
glasstiger: Now it is checking for the presence of a view definition file, and parsing table id from the dir name only as a fallback.

Initially I thought it should fail if we cannot have a proper table id, but now I am thinking that returning 0 without a failure here might be intentional.
Currently having a unique table id is not critical, and returning 0 does not prevent the database from starting up.

jerrinot (approved):

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 9 / 12 (75.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableUtils.java|3|6|50.00%|
|:large_blue_circle:|io/questdb/cairo/TableNameRegistryStore.java|6|6|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=207519)
**Tested Commit:** `f18460f99ad9a598f189438149905e683a73a845`

glasstiger:
@coderabbitai review

core/src/main/java/io/questdb/cairo/TableNameRegistryStore.java
```diff
@@ -309,9 +310,18 @@ private int findLastTablesFileVersion(FilesFacade ff, Path path) {
...
         long fd = ff.openRO(path.$());
         if (fd < 1) {
+            // check if it is a view
+            path.trimTo(pathLen).concat(ViewDefinition.VIEW_DEFINITION_FILE_NAME);
+            if (ff.exists(path.$())) {
+                // negative table id means WAL table,
+                // views are considered to be WAL tables
+                return -getTableIdFromTableDir(dirName);
+            }
             return 0;
```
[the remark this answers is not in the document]
glasstiger: @coderabbitai this is intentional.
Not 100% sure why the current implementation returns 0, if there is no _meta file, but it is likely to make sure that the database starts up even if files are corrupted.
Obviously tables without `_meta` file will not be usable, but other tables could be fully functional.

However, this is not the case for views.
Views would not show an obvious failure, but their table id would be set to 0.
Opted for failing to start the database in case we cannot determine the table id of a view.
If the view directory does not follow our naming rules, the database root dir has been tempered with anyway, and the files are corrupted one way or another.

PR #6619, opened 2026-01-09
fix(core): process memory may not be released on Linux when jemalloc is enabled

puzpuzpuz:
Jemalloc background threads were failing to start reliably (~20% failure rate) when QuestDB was launched via questdb.sh. This prevented memory from being released back to the OS properly.

To fix this race, we're switching to QuestDB's own jemalloc fork. See https://github.com/questdb/jemalloc/pull/1 for more details on the race, as well as the fix.

**Important note**. Previously we were using the latest jemalloc release which is pretty outdated (May 2022). So, along with the fix, we're switching to the latest `dev` branch commit (June 2025). Other databases also seem to use jemalloc's `dev` branch, e.g. ClickHouse is also using the latest `dev` commit.

Other than that, includes the following changes:
* Use `env LD_PRELOAD=...` to load jemalloc only in the Java process since it makes no sense to use jemalloc for the `questdb.sh` bash script

#### Test Results

  - Before: ~20% failure rate
  - After: 0% failure rate (200/200 successful runs)

### DIFF
--- a/core/src/main/java/io/questdb/cairo/O3OpenColumnJob.java
+++ b/core/src/main/java/io/questdb/cairo/O3OpenColumnJob.java
@@ -271,9 +271,9 @@ public class O3OpenColumnJob extends AbstractQueueConsumerJob<O3OpenColumnTask> 
             final ColumnTypeDriver columnTypeDriver = ColumnType.getDriver(columnType);
             // srcDataMax is the row count in the existing column data.
             // Use Math.max as a defensive guard against srcDataTop > srcDataMax edge cases.
-            final long auxRowCount = Math.max(0L, srcDataMax - srcDataTop);
+            final long auxRowCountOld = Math.max(0L, srcDataMax - srcDataTop);
             if (srcDataTop > 0 && tableWriter.isCommitReplaceMode()) {
-                // Adjust source data indexes for what we need for the range replace merge.
+                // Adjust source data lengths limiting to what we need for the range replace merge.
                 long dataMax = 0;
                 if (prefixType == O3_BLOCK_DATA && prefixHi >= prefixLo) {
                     dataMax = prefixHi + 1;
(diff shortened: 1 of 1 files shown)


core/src/main/bin/questdb.sh
```diff
@@ -380,14 +380,21 @@ function start {
     DATE=`date +%Y-%m-%dT%H-%M-%S`
     HELLO_FILE=${QDB_ROOT}/hello.txt
     rm ${HELLO_FILE} 2> /dev/null
+
+    # Set LD_PRELOAD only for the Java process, not the shell
+    JAVA_CMD="${JAVA}"
+    if [ -n "${QDB_JEMALLOC_LIB}" ]; then
+        JAVA_CMD="env LD_PRELOAD=${QDB_JEMALLOC_LIB} ${JAVA}"
```
puzpuzpuz, line 387: @sklarsa do you happen to know if all Docker base images that we use include `env` utility? Usually it's provided as a part of `coreutils`, so Debian, Alpine and CentOS/RHEL-based images should have it.

puzpuzpuz:
@coderabbitai review

core/src/main/c/share/jemalloc
```diff
@@ -1 +1 @@
-Subproject commit 8d8379da443f46dc976252b968cb9ca8e63ec974
+Subproject commit f6655d686bcd57e3373d4b113fea2fd5c4a4785e
```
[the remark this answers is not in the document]
puzpuzpuz: Done in 5099b2054a72727e38dbf261fa2e2e7a709f9657

core/src/main/bin/questdb.sh
bluestreak01: @puzpuzpuz you'd have to test docker builds...., speaking of which, should we also have docker use jemalloc?

bluestreak01: https://github.com/questdb/questdb/blob/master/core/docker-entrypoint.sh

bluestreak01: https://github.com/questdb/questdb/blob/master/core/Dockerfile#L84
https://github.com/questdb/questdb/blob/master/core/Dockerfile#L132

puzpuzpuz: My concern was around custom Docker images that users may have. Though it's probably not a big deal since `questdb.sh` already relies on a few utilities.

As for enabling jemalloc in our Docker image, I'd wait a few releases after 9.3.1, so that we're more confident that there are no remaining issues around jemalloc. I've run tests locally and it seems to work fine, but extra cautiousness won't hurt.

bluestreak01 (approved):

PR #6519, opened 2025-12-10
fix(core): data corruption after range replace on column top partition

puzpuzpuz:
The bug can be reproduced only in case of range replace transactions running on partitions with column tops, so no users are affected. That's because the only place where we use range replace is mat views and it's impossible to add a column to a mat view.

In case of a range replace transaction and a column top, `O3OpenColumnJob#merge*Column()` methods could overwrite data in the source partition making it corrupted. As an example, in the `mergeVarColumn()` method the `auxRowCount` variable was calculated after we adjust `srcDataMax` and `srcDataTop` values, so `auxRowCount` could end being set to zero while the source partition has some non-coltop rows. Later on, the `columnTypeDriver.setPartAuxVectorNull()` would overwrite the source aux file leaving it with the content like what's shown below.

```bash
# corrupted binary column (col top - 7, total rows - 8)
$ xxd /tmp/junit12784799641063025524/dbRoot/testBaseTableCanHaveColumnsAdded_0~2/2022-03-01.124/new_col_73.i.125
00000000: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000010: 0800 0000 0000 0000 1000 0000 0000 0000  ................
00000020: 1800 0000 0000 0000 2000 0000 0000 0000  ........ .......
00000030: 2800 0000 0000 0000 3000 0000 0000 0000  (.......0.......
00000040: 0000 0000 0000 0000 0000 0000 0000 0000  ................

$ xxd /tmp/junit12784799641063025524/dbRoot/testBaseTableCanHaveColumnsAdded_0~2/2022-03-01.124/new_col_73.d.125
00000000: ffff ffff ffff ffff ffff ffff ffff ffff  ................
00000010: ffff ffff ffff ffff ffff ffff ffff ffff  ................
00000020: ffff ffff ffff ffff ffff ffff ffff ffff  ................
00000030: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000040: 0000 0000 0000 0000 0000 0000 0000 0000  ................

# valid binary column (col top - 7, total rows - 8)
$ xxd /tmp/junit11744283241191190407/dbRoot/testBaseTableCanHaveColumnsAdded_0~2/2022-03-01.127/new_col_73.i.128 
00000000: 0000 0000 0000 0000 0800 0000 0000 0000  ................
00000010: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000020: 0000 0000 0000 0000 0000 0000 0000 0000  ................
[...]

PR #6628, opened 2026-01-12
fix(core): fix rare bug that can potentially write invalid data in mat view

ideoma:
Modifies fix in https://github.com/questdb/questdb/pull/6519 for some edge case scenarios.

When a fixed column is merged with the column top, the data copied should be 

```Java
// Copy data to the end of old data + column tops
Vect.memcpy(srcDataFixAddr + srcDataActualBytesOld + (srcDataTop << shl), srcDataFixAddr, srcDataActualBytesNew);

// instead of copying to the end of the newly calculated data size
Vect.memcpy(srcDataFixAddr + wouldBeDataBytes, srcDataFixAddr, srcCopyActualBytes);
```

Var column types appear not to have the same bug, only comments modified.

### DIFF
--- a/core/src/main/java/io/questdb/DefaultTelemetryConfiguration.java
+++ b/core/src/main/java/io/questdb/DefaultTelemetryConfiguration.java
@@ -48,6 +48,16 @@ public class DefaultTelemetryConfiguration implements TelemetryConfiguration {
     }
 
     @Override
+    public long getThrottleIntervalMicros() {
+        return Micros.MINUTE_MICROS;
+    }
+
+    @Override
+    public int getTtlWeeks() {
+        return 4;
+    }
+
+    @Override
     public boolean hideTables() {
         return false;
     }
--- a/core/src/main/java/io/questdb/PropServerConfiguration.java
+++ b/core/src/main/java/io/questdb/PropServerConfiguration.java
@@ -508,12 +508,9 @@ public class PropServerConfiguration implements ServerConfiguration {
     private final long systemWalWriterEventAppendPageSize;
     private final long systemWriterDataAppendPageSize;
     private final boolean tableTypeConversionEnabled;
-    private final TelemetryConfiguration telemetryConfiguration = new PropTelemetryConfiguration();
+    private final TelemetryConfiguration telemetryConfiguration;
     private final long telemetryDbSizeEstimateTimeout;
     private final boolean telemetryDisableCompletely;
-    private final boolean telemetryEnabled;
-    private final boolean telemetryHideTables;
-    private final int telemetryQueueCapacity;
     private final CharSequence tempRenamePendingTablePrefix;
     private final int textAnalysisMaxLines;
     private final TextConfiguration textConfiguration = new PropTextConfiguration();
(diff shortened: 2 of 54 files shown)


core/src/main/java/io/questdb/cairo/O3OpenColumnJob.java
```diff
@@ -2464,26 +2456,25 @@ private static void mergeFixColumn(
...
                     srcDataFixAddr = mapRW(ff, srcFixFd, srcDataFixSize, MemoryTag.MMAP_O3);
                     ff.madvise(srcDataFixAddr, srcDataFixSize, Files.POSIX_MADV_SEQUENTIAL);
-                    TableUtils.setNull(columnType, srcDataFixAddr + srcDataActualBytes, srcDataTop);
-                    // This value may be lower than srcDataActualBytes.
-                    final long srcCopyActualBytes = Math.max(0L, srcDataMax - srcDataTop) << shl;
-                    if (srcCopyActualBytes > 0) {
-                        Vect.memcpy(srcDataFixAddr + wouldBeDataBytes, srcDataFixAddr, srcCopyActualBytes);
-                    }
+                    TableUtils.setNull(columnType, srcDataFixAddr + srcDataActualBytesOld, srcDataTop);
+                    // This value may be lower than srcDataActualBytesNew.
```
puzpuzpuz: nit: what is "this value" here?
```suggestion
                    // This value may be lower than srcDataActualBytesNew.
```

core/src/main/java/io/questdb/cairo/O3OpenColumnJob.java
```diff
@@ -2493,7 +2484,7 @@ private static void mergeFixColumn(
                         // If it's split partition, do nothing. Old partition will have the old column top
                         // New partition will have 0 column top, since srcDataTop <= prefixHi.
                     }
-                    srcDataFixSize = srcDataActualBytes;
+                    srcDataFixSize = srcDataActualBytesNew;
```
puzpuzpuz, line 2497: Shouldn't this be `srcDataActualBytesOld` like it was before this fix, so that we mmap full old file?

core/src/main/java/io/questdb/cairo/O3OpenColumnJob.java
puzpuzpuz: Improved in 44a148a2e507fbf8c71005d4cadc45e7a918e6b5

core/src/main/java/io/questdb/cairo/O3OpenColumnJob.java
puzpuzpuz: Never mind, the subsequent code must not access bytes beyond `srcDataFixSize`. Otherwise, not only this, but the below `else` would be a problem.

core/src/main/java/io/questdb/cairo/O3OpenColumnJob.java
```diff
@@ -373,12 +369,12 @@ public static void mergeVarColumn(
                     }
 
                     srcDataFixOffset = 0;
-                    if (auxRowCount > 0) {
-                        newAuxSize = columnTypeDriver.getAuxVectorSize(auxRowCount);
+                    if (auxRowCountOld > 0) {
```
puzpuzpuz: @ideoma shouldn't we use `Math.max(0L, srcDataMax - srcDataTop)` (`auxRowCountNew`) instead of `auxRowCountOld` here?

ideoma: yep, new is better, I used it

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 24 / 24 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/O3OpenColumnJob.java|24|24|100.00%|

ideoma:
@coderabbitai review

PR #6634, opened 2026-01-13
fix(sql): fix ASOF JOIN crash when ON clause has symbol and other columns

jerrinot:
ASOF JOIN queries with a symbol column AND another column in the ON clause would throw `UnsupportedOperationException` when using the Light join factory.

The single-symbol optimization was incorrectly applied when there were multiple join keys, creating a map that only supported INT keys while the record copier attempted to write all column types.

In some cases it does not crash, but it can silently return wrong results.


jerrinot:
@coderabbitai pls review

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|3|3|100.00%|

PR #6597, opened 2026-01-05
perf(sql): reduce garbage generated on parallel query hot path

puzpuzpuz:
Reduces GC pressure and jitter in parallel queries on large multi-core machines (tested on c7a.metal-48xl with 192 vCPUs).

####  Key Changes

  1. Off-heap Page Frame Address Cache
  - `PageFrameAddressCache` now uses flat `DirectLongList` lists instead of nested `LongList` objects
  - Reduces ~20K object allocations per 1000 frames × 105 columns to just 4 contiguous off-heap lists
  - Memory footprint moves from on-heap (~3.4 MB for 1K page frames) to off-heap, eliminating GC pressure

  2. Lazy MapValue Flyweights
  - `Decimal128`, `Decimal256`, `Long256` flyweights in map values are now lazily initialized
  - Avoids allocations when these types aren't used in queries. Decimal flyweight as relatively expensive since each of them allocates multiple `int[]` array.

  3. `GroupByAllocator` Improvements
  - `LongLongHashMap` is replaced with off-heap `DirectLongLongHashMap`. Previously, each query worker was growing the hash map when allocating leading to lots of on-heap allocations.

  4. Page Frame Size Calculation Improvements
  - Page frame sizes are now calculated to avoid tiny trailing frames within a partition. Minimal page frame size is also increased 2x for vectorized non-keyed GROUP BY since these tasks a more lightweight than filter or other GROUP BY ones.

  6. Work Stealing Strategy Tuning
  - New config: `cairo.sql.parallel.work.stealing.spin.timeout` (default 50µs)
  - Improvements to `AdaptiveWorkStealingStrategy` for better parallel execution. Namely, `System.nanoTime()` with configurable timeout is now used to avoid variable spinning time on different OSes and CPU architectures. `Thread.onSpinWait()` is used in the spin-wait loop to use PAUSE/YIELD instruction instead of "sleep(0)" system call.

  7. Proper Resource Cleanup
  - All classes using `PageFrameAddressCache` now call Misc.free(cursor) in close() methods

  8. Removed Config
  - Removed `cairo.sql.jit.page.address.cache.threshold` (no longer needed)

#### Benchmarks

I've run [clickbench](https://github.com/ClickHouse/ClickBench/) queries with 30 iterations per query on my Ryzen 7900x (12c/24t), 64GB RAM box running Ubuntu 24.04 and GraalVM CE 17.0.8.

##### GC Analysis Summary

**Master Branch GC Metrics**

  - Total runtime: ~623 seconds
[...]

PR #6636, opened 2026-01-13
perf(core): speed-up hash map deletion

oatkachenko:
Instead of probing from an entry's ideal position to find the best available slot, entries now simply move backward to fill the nearest gap in their probe sequence.

The new approach is similar to IdentityHashMap.closeDeletion().

The benchmark shows that in the worst-case scenario (which is pretty unrealistic), the impact is huge.
```
Benchmark                                            (mapSize)  (pattern)  Mode  Cnt        Score        Error  Units
DirectLongLongHashMapRemoveBenchmark.testRemoveAt         8192     RANDOM  avgt    5       10.008 ±      1.051  ns/op
DirectLongLongHashMapRemoveBenchmark.testRemoveAt         8192  CLUSTERED  avgt    5  5661484.584 ± 350201.270  ns/op
DirectLongLongHashMapRemoveBenchmark.testRemoveAtV2       8192     RANDOM  avgt    5        8.518 ±      0.450  ns/op
DirectLongLongHashMapRemoveBenchmark.testRemoveAtV2       8192  CLUSTERED  avgt    5     7960.529 ±     54.078  ns/op
```


puzpuzpuz:
As I mentioned in #6597, `removeAt()` is also present in many other (mostly on-heap) hash tables: e.g. see `AbstractIntHashSet`, `AbstractLongHashSet`, `AbstractLowerCaseCharSequenceHashMap`, `AbstractCharSequenceHashSet`, `AbstractUtf8SequenceHashSet`, `Utf8StringIntHashMap`, `Utf8StringObjHashMap`. Ideally, all of them need to be optimized, so that the code base is kept consistent.

core/src/main/java/io/questdb/std/DirectLongLongHashMap.java
```diff
@@ -225,6 +235,28 @@ private long probe(long key, long index) {
...
+             key != noEntryKey;
+             scanPos = (scanPos + 1) & mask, key = keyAtRaw(scanPos)) {
+
+            long idealPos = Hash.fastHashLong64(key) & mask;
+
+            // Check if moving backward would be beneficial:
+            // 1. scanPos != idealPos: Element is not in its ideal position;
+            // 2. idealPos <= gapPos: Normal case - ideal position is at or before the gap
+            // 3. currentPos < idealPos: Wraparound case - entry wrapped from table end
+            if (scanPos != idealPos && (idealPos <= gapPos || scanPos < idealPos)) {
```
puzpuzpuz: The condition looks differently from what's in `IdentityHashmap`. Could you explain the difference?
https://github.com/openjdk/jdk/blob/45990d796ffafc228c6e843049c80aefedb0f12b/src/java.base/share/classes/java/util/IdentityHashMap.java#L620

```java
            // The following test triggers if the item at slot i (which
            // hashes to be at slot r) should take the spot vacated by d.
            // If so, we swap it in, and then continue with d now at the
            // newly vacated i.  This process will terminate when we hit
            // the null slot at the end of this run.
            // The test is messy because we are using a circular table.
            int r = hash(item, len);
            if ((i < r && (r <= d || d <= i)) || (r <= d && d <= i)) {
```

core/src/test/java/io/questdb/test/std/DirectLongLongHashMapTest.java
```diff
@@ -153,4 +155,82 @@ public void testRemoveAt() {
             }
         }
     }
+
+    @Test
+    public void removeCompactsProbeChainWithCircularWraparound() {
```
puzpuzpuz, line 160: It'd be great to add a randomized (fuzz) test for new key removal on top of deterministic tests. First you initialize PRNG with `Rnd rnd = TestUtils.generateRandom(AbstractCairoTest.LOG)` and then use it to fill the map in and then to remove the keys.

core/src/main/java/io/questdb/std/DirectLongLongHashMap.java
oatkachenko: Oh… it looks like I forgot the case when the gap is farther than the ideal position. I’ve updated the condition to handle that case as well and added comments to make things clear. 
Also, the DirectLongLongHashMap#shouldMoveToFillGap() method is generic and could be reused in other classes. 
Could you please suggest the right place for this method?

puzpuzpuz: > Could you please suggest the right place for this method?

I guess you could create a `MapUtils` utility class in the `io.questdb.std` package.

oatkachenko: In the last commit, I updated the #removeAt() where it was necessary, and moved the common logic to the  MapUtils.

core/src/test/java/io/questdb/test/std/DirectLongLongHashMapTest.java
oatkachenko: I checked the existing tests, and such a test already exists for the old logic. Since I replaced the old logic, that test now covers everything we need.

puzpuzpuz:
@oatkachenko thanks for updating the PR! I should be able to do another review round later this week.

puzpuzpuz:
@coderabbitai review

puzpuzpuz:
Linter is not very happy with the benchmark code:
```
--- a/benchmarks/src/main/java/org/questdb/DirectLongLongHashMapRemoveBenchmark.java
+++ b/benchmarks/src/main/java/org/questdb/DirectLongLongHashMapRemoveBenchmark.java
@@ -39,11 +39,11 @@ import java.util.concurrent.TimeUnit;
 @BenchmarkMode(Mode.AverageTime)
 @OutputTimeUnit(TimeUnit.NANOSECONDS)
 public class DirectLongLongHashMapRemoveBenchmark {
-    
+
     private static final double LOAD_FACTOR = 0.5;
     private static final long NO_ENTRY_KEY = -1;
     private static final long NO_ENTRY_VALUE = -1;
-    
+
     private DirectLongLongHashMap map;
     private Rnd rnd;
     private long[] keysToRemove;
diff --git a/core/src/test/java/io/questdb/test/std/DirectLongLongHashMapTest.java b/core/src/test/java/io/questdb/test/std/DirectLongLongHashMapTest.java
index 9495df2ee..fa735a53a 100644
--- a/core/src/test/java/io/questdb/test/std/DirectLongLongHashMapTest.java
+++ b/core/src/test/java/io/questdb/test/std/DirectLongLongHashMapTest.java
@@ -256,7 +256,7 @@ public class DirectLongLongHashMapTest {
 
     /**
      * @param overlaps - we can create different objects that hash to the same position in the map.
-     *                   e.g., 0 - is the first object that hashes to the position, 1 - the second, etc.
+     *                 e.g., 0 - is the first object that hashes to the position, 1 - the second, etc.
      * @return a key that hashes to the specified position in the map.
      */
     private static long keyHashedToPos(DirectLongLongHashMap map, long pos, int overlaps) {


```

core/src/main/java/io/questdb/std/ObjHashSet.java
```diff
@@ -208,33 +210,30 @@ private void removeAt(int index) {
...
+    private void compactProbeSequence(int deletedPosition) {
+        int gapPos = deletedPosition;
+        int scanPos = (gapPos + 1) & mask;
+
+        // Scan forward until we hit an empty slot (end of probe sequence)
+        for (T key = keys[scanPos];
+             key != noEntryKey;
+             scanPos = (scanPos + 1) & mask, key = keys[scanPos]) {
+
+            long idealPos = Hash.spread(key.hashCode()) & mask;
```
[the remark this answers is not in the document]
puzpuzpuz: This hash set supports `null` key insertion, so the comment is spot-on. The old `removeAt()` impl also had this bug.

core/src/main/java/io/questdb/std/AbstractLongHashSet.java
```diff
@@ -144,6 +118,30 @@ private int probe(long key, int index) {
...
+     * may need to be moved backward to preserve correct lookup semantics.
+     */
+    private void compactProbeSequence(int deletedPosition) {
+        int gapPos = deletedPosition;
+        int scanPos = (gapPos + 1) & mask;
+
+        // Scan forward until we hit an empty slot (end of probe sequence)
+        for (long key = keys[scanPos];
+             key != noEntryKey;
+             scanPos = (scanPos + 1) & mask, key = keys[scanPos]) {
```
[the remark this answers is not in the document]
puzpuzpuz: This one is also spot-on. `noEntryKey` here is a bad naming. Should be smth like `DEFAULT_NO_ENTRY_KEY`.

core/src/main/java/io/questdb/std/AbstractIntHashSet.java
```diff
@@ -134,6 +109,30 @@ private int probe(int key, int index) {
...
+     * may need to be moved backward to preserve correct lookup semantics.
+     */
+    private void compactProbeSequence(int deletedPosition) {
+        int gapPos = deletedPosition;
+        int scanPos = (gapPos + 1) & mask;
+
+        // Scan forward until we hit an empty slot (end of probe sequence)
+        for (int key = keys[scanPos];
+             key != noEntryKey;
+             scanPos = (scanPos + 1) & mask, key = keys[scanPos]) {
```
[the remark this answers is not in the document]
puzpuzpuz: Same story as in the long hash set.

core/src/main/java/io/questdb/std/AbstractLowerCaseUtf8SequenceHashSet.java
```diff
@@ -171,6 +145,31 @@ private int probe(Utf8Sequence key, int hashCode, int lo, int hi, int index) {
...
+    private void compactProbeSequence(int deletedPosition) {
+        int gapPos = deletedPosition;
+        int scanPos = (gapPos + 1) & mask;
+
+        // Scan forward until we hit an empty slot (end of probe sequence)
+        for (Utf8Sequence key = keys[scanPos];
+             key != noEntryKey;
+             scanPos = (scanPos + 1) & mask, key = keys[scanPos]) {
+
+            int hashCode = Utf8s.lowerCaseAsciiHashCode(key);
```
puzpuzpuz, line 163: While we're on it, here we could use `hashCodes` array to avoid recalculating the hash code.

core/src/main/java/io/questdb/std/AbstractUtf8SequenceHashSet.java
```diff
@@ -175,6 +149,31 @@ private int probe(Utf8Sequence key, int hashCode, int lo, int hi, int index) {
...
+    private void compactProbeSequence(int deletedPosition) {
+        int gapPos = deletedPosition;
+        int scanPos = (gapPos + 1) & mask;
+
+        // Scan forward until we hit an empty slot (end of probe sequence)
+        for (Utf8Sequence key = keys[scanPos];
+             key != noEntryKey;
+             scanPos = (scanPos + 1) & mask, key = keys[scanPos]) {
+
+            int hashCode = Utf8s.hashCode(key);
```
puzpuzpuz, line 167: Same story: here we could use `hashCodes` array to avoid recalculating the hash code.

core/src/main/java/io/questdb/std/Utf8StringIntHashMap.java
```diff
@@ -271,4 +245,29 @@ private void rehash() {
...
+    private void compactProbeSequence(int deletedPosition) {
+        int gapPos = deletedPosition;
+        int scanPos = (gapPos + 1) & mask;
+
+        // Scan forward until we hit an empty slot (end of probe sequence)
+        for (Utf8String key = keys[scanPos];
+             key != null;
+             scanPos = (scanPos + 1) & mask, key = keys[scanPos]) {
+
+            int hashCode = Hash.hashUtf8(key);
```
puzpuzpuz, line 264: Same story: here we could use `hashCodes` array to avoid recalculating the hash code.

core/src/main/java/io/questdb/std/Utf8StringObjHashMap.java
```diff
@@ -297,4 +271,29 @@ private void rehash() {
...
+    private void compactProbeSequence(int deletedPosition) {
+        int gapPos = deletedPosition;
+        int scanPos = (gapPos + 1) & mask;
+
+        // Scan forward until we hit an empty slot (end of probe sequence)
+        for (Utf8String key = keys[scanPos];
+             key != null;
+             scanPos = (scanPos + 1) & mask, key = keys[scanPos]) {
+
+            int hashCode = Hash.hashUtf8(key);
```
puzpuzpuz, line 290: Same story: here we could use `hashCodes` array to avoid recalculating the hash code.

benchmarks/src/main/java/org/questdb/DirectLongLongHashMapRemoveBenchmark.java
```diff
@@ -0,0 +1,137 @@
...
+    private static final double LOAD_FACTOR = 0.5;
+    private static final long NO_ENTRY_KEY = -1;
+    private static final long NO_ENTRY_VALUE = -1;
+    
+    private DirectLongLongHashMap map;
+    private Rnd rnd;
+    private long[] keysToRemove;
+    private int removeIndex;
+
+    @Param({"8192"}) // , "131072", "1048576", "4194304"
```
puzpuzpuz, line 52: How about uncommenting other map sizes and running the benchmark for them?

Issue #6637, opened 2026-01-13
LT/ASOF JOIN with timestamp in ON clause crashes with AssertionError instead of returning proper error


jerrinot:
### To reproduce

Description:
When using `LT JOIN` or `ASOF JOIN` with the designated timestamp column included in the `ON` clause, QuestDB might crash with an AssertionError instead of returning a proper SQLException.

Reproducer:
```java
@Test
    public void testLtJoinWithTimestampInOnClauseCrash() throws Exception {
        assertMemoryLeak(() -> {
            execute("CREATE TABLE t1 (ts TIMESTAMP, i INT, s SYMBOL, val INT) timestamp(ts) partition by DAY");
            execute("CREATE TABLE t2 (ts TIMESTAMP, i INT, s SYMBOL INDEX, val INT) timestamp(ts) partition by DAY");

            execute("INSERT INTO t1 VALUES ('2024-01-01T10:00:00.000000Z', 1, 'A', 10)");
            execute("INSERT INTO t2 VALUES ('2024-01-01T10:00:00.000000Z', 2, 'A', 20)");

            try {
                // SAMPLE BY makes the slave factory return false for recordCursorSupportsRandomAccess()
                // which triggers the full fat join code path without needing setFullFatJoins(true)
                assertQueryNoLeakCheck(
                        "", // expected - we don't care, expecting exception
                        "SELECT t1.ts, sub.isum FROM t1 LT JOIN (" +
                                "  SELECT ts, sum(i) as isum, s FROM t2 SAMPLE BY 1us ALIGN TO CALENDAR" +
                                ") sub ON (t1.s = sub.s) AND (t1.ts = sub.ts)",
                        null,
                        false,
                        sqlExecutionContext
                );
                Assert.fail("Query should have thrown SQLException, not succeeded");
            } catch (SqlException e) {
                // Expected behavior - proper error message about timestamp in join key
                Assert.assertTrue("Expected error about timestamp in join key, got: " + e.getMessage(),
                        e.getMessage().contains("timestamp"));
            } catch (AssertionError e) {
                // This is the BUG - AssertionError is thrown instead of SQLException
                Assert.fail("Query threw AssertionError instead of proper SQLException: " + e.getMessage());
            }
        });
    }
```

Using a timestamp as a key in LT/ASOF JOINs make little sense, IMHO it's acceptable ot throw an SQLException. But AssertionError is not acceptable at all. 

### QuestDB version:

8.3.0

[...]

Issue #4222, opened 2024-02-20
Flaky PGJobContextTest.testRunAlterWhenTableLockedAndAlterTakesTooLong()


glasstiger:
### To reproduce

Intermittent CI build failure.

Full log: 
[201.txt.zip](https://github.com/questdb/questdb/files/14346853/201.txt.zip)

```
2024-02-20T13:49:11.890578Z I i.q.t.AbstractCairoTest Tearing down test PGJobContextTest#testRunAlterWhenTableLockedAndAlterTakesTooLong[NO_WAL]
2024-02-20T13:49:11.890716Z I i.q.t.AbstractTest Finished test PGJobContextTest#testRunAlterWhenTableLockedAndAlterTakesTooLong[NO_WAL]
2024-02-20T13:49:11.893824Z I i.q.t.c.p.PGJobContextTest os scheduled worker stopped [name=pgwire_1]
random seeds: 1804945292400L, 1708436951897L
2024-02-20T13:49:11.902003Z E i.q.t.TestListener ***** Test Failed *****io.questdb.test.cutlass.pgwire.PGJobContextTest.testRunAlterWhenTableLockedAndAlterTakesTooLong[NO_WAL] duration_ms=3783 : 
java.lang.AssertionError: expected:<0> but was:<1>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.junit.Assert.assertEquals(Assert.java:633)
	at io.questdb.test.cutlass.pgwire.PGJobContextTest.testAddColumnBusyWriter(PGJobContextTest.java:10325)
	at io.questdb.test.cutlass.pgwire.PGJobContextTest.lambda$testRunAlterWhenTableLockedAndAlterTakesTooLong$115(PGJobContextTest.java:7466)
	at io.questdb.test.AbstractCairoTest.lambda$assertMemoryLeak$7(AbstractCairoTest.java:917)
	at io.questdb.test.tools.TestUtils.assertMemoryLeak(TestUtils.java:619)
	at io.questdb.test.AbstractCairoTest.assertMemoryLeak(AbstractCairoTest.java:914)
	at io.questdb.test.AbstractCairoTest.assertMemoryLeak(AbstractCairoTest.java:908)
[...]

PR #6638, opened 2026-01-13
test(sql): extend ASOF/LT JOIN fuzz coverage to multi-column ON clauses

jerrinot:
Follow-up to #6634. Adds a `JoinKeyType` axis to `AsOfJoinFuzzTest` so the existing result-equivalence checks also exercise multi-column `ON` clauses:
- `SINGLE_SYMBOL` — `ON s` (existing).
- `SYMBOL_AND_INT` — forces the multi-key path.
- `SYMBOL_AND_TIMESTAMP` — filtered out; the planner rejects the
  designated timestamp as a join key with a clean `SqlException`
  (#6698), leaving nothing for the harness to compare. Kept as an
  enum value to document intent.

Both fuzz tables gain a `val INT` column, and the existing projection variants are updated to stay consistent with the new key shape. Invalid permutations (multi-key with non-keyed joins, `REMOVE_SYMBOL_COLUMN` with a keyed join) now get filtered before the random subset is drawn, so the `RUN_N_PERMUTATIONS` budget isn't spent on combinations the test would skip anyway.


bluestreak01 (approved):
Approved. Test-only PR extending the full-fat-vs-optimized ASOF/LT JOIN differential fuzz harness to multi-column ON clauses.

Verified:
- The fullFatJoins differential mechanism plus the slaveTimestamp <= masterTimestamp invariant give an independent correctness check beyond pure equivalence.
- No hint (asof_index/memoized/dense/linear) throws on a multi-column join; each falls back cleanly in SqlCodeGenerator, so no permutation produces a spurious compile failure.
- isSingleSymbolJoin gating (getColumnCount() == 1) matches the test's plan-assertion guard; LINEAR multi-column resolves to AsOfJoinLight (plan 'AsOf Join Light'), satisfying the assertNotContains checks.
- SYMBOL_AND_TIMESTAMP is correctly filtered to match the planner's SqlException; column availability holds across every non-filtered projection; isValidPermutation correctly subsumes the old in-body REMOVE_SYMBOL_COLUMN skip.

Non-blocking nits: no plan-shape assertion for multi-column joins (a silent hint fallback would go undetected); fully-qualified java.util.ArrayList instead of an import; hardcoded params[] indices in isValidPermutation must stay in lockstep with allOpts; PR body cites #6698 while the code references #6637.

jerrinot:
2 unrelated known flaky tests:(

1. ServerMainSleepTest#testSleepCancelledByConnectionDrop - https://github.com/questdb/questdb/issues/7296 - should be fixed by https://github.com/questdb/questdb/pull/7324
2. PGJobContextTest#testRunAlterWhenTableLockedAndAlterTakesTooLong - https://github.com/questdb/questdb/issues/4222 - no fix PR yet

PR #6644, opened 2026-01-14
chore(sql): avoid NPE for missing expression in ORDER BY or unbalanced right parentheses in top level SELECT

mghildiy:
**Summary**:
1. Avoid NullPointerException when there is a missing expression in ORDER BY.
2. Avoid NullPointerException when there is a unbalanced right prentheses in top level SELECT.

**Problem**
As reported in [issue](https://github.com/questdb/questdb/issues/6536), some SQLs are invalid due to missing expression or unablanced right parentheses, but fail with NPE instead of SQL exception with appropriate message.

**Root Cause**
Though missing expression in ORDER BY clause is captured as empty token, but for some scenarios( like, CASE clause being present) it is captured as null token.
Unabalanced right parentheses not handled correctly, leading to creation of null objects during sql parsing.

**Changes**
1. **SqlParser.java::parseFromClause** : For handling NPE when there is a missing expression in ORDER BY clause, add check for null token in code block where ORER-BY clause is parsed.
2. **SqlParser.java::parseSelectClause** : Unabalanced right parentheses in top elevel SELECT clause is captured and sql exception is generated, using appropriate flags to handle subquery mode, copy mode, create tabel mode, create view mode normally.

**Test Plan**
1. Added 2 tests in SqlParserTest.java to capture the scenarios(testUnbalancedRightParenthesesThrowsCorrectErrorMessage, testMissingExpressionInOrderByThrowsCorrectErrorMessage).
2. All 918 tests in SqlParserTest.java pass.
3. Failing queries mentioned in issue now throw SQL exception with correct error message.
4. All tests in PGErrorHandlingTest, HttpErrorHandlingTest pass.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/SqlParser.java
+++ b/core/src/main/java/io/questdb/griffin/SqlParser.java
@@ -147,6 +147,9 @@ public class SqlParser {
     private final LowerCaseCharSequenceHashSet viewsBeingCompiled = new LowerCaseCharSequenceHashSet();
     private final ObjectPool<WindowExpression> windowExpressionPool;
     private final ObjectPool<WithClauseModel> withClauseModelPool;
+    private boolean copyMode = false;
+    private boolean createTableMode = false;
+    private boolean createViewMode = false;
     private int digit;
     private boolean pivotMode = false;
     private boolean subQueryMode = false;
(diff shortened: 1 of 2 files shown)


bluestreak01:
@coderabbit review, also check the issue this PR is claiming to fix

PR #6659, opened 2026-01-16
chore(core): fix view fuzz test timeout

glasstiger:
View fuzz test timed out on CI, `testMultipleLevelDependencyViewsWithRandomSelect()` could not finish within 20 minutes.
The test took so long because the initial row count for tables was set to 100k rows, and when these tables were joined multiple times on low cardinality columns, the result was too large to process.

In this PR:
 - lowering initial row count for tables to 3k rows
 - also rationalized other fuzz parameters
 - simplified, cleaned up the code
 - excluded boolean columns from joins because of their low cardinality


glasstiger:
@coderabbitai review

PR #6676, opened 2026-01-20
fix(sql): fix parquet read failure on chained window join

kafka1991:
Fix parquet read failures in window chain join queries.

  **Root cause:**                                                                                                                                                                                                      
  - Chained window join should use `GenericRecordMetadata.deepCopyOf()` instead of `copyOf()`

### DIFF
--- a/benchmarks/src/main/java/org/questdb/GroupByFunctionsUpdaterBenchmark.java
+++ b/benchmarks/src/main/java/org/questdb/GroupByFunctionsUpdaterBenchmark.java
@@ -1,0 +1,181 @@
+/*******************************************************************************
+ *     ___                  _   ____  ____
+ *    / _ \ _   _  ___  ___| |_|  _ \| __ )
+ *   | | | | | | |/ _ \/ __| __| | | |  _ \
+ *   | |_| | |_| |  __/\__ \ |_| |_| | |_) |
+ *    \__\_\\__,_|\___||___/\__|____/|____/
+ *
+ *  Copyright (c) 2014-2019 Appsicle
+ *  Copyright (c) 2019-2026 QuestDB
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ ******************************************************************************/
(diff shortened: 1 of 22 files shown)


kafka1991:
@coderabbitai review

bluestreak01:
@Coderabbit can you review and report?

bluestreak01:
hey @kafka1991 here is AI spot:

```
  1. Naming Inconsistency in GenericRecordMetadata:

  The new deepCopyOf(RecordMetadata that, int columnCount) method name is misleading:

  - deepCopyOf(RecordMetadata that) (lines 115-136) truly deep-copies column metadata by creating new TableColumnMetadata objects
  - deepCopyOf(RecordMetadata that, int columnCount) (lines 103-113) does NOT deep-copy column metadata - it just skips the instanceof shortcut

  The actual difference is:
  // copyOf - may return same instance
  if (that instanceof GenericRecordMetadata) {
      return (GenericRecordMetadata) that;  // returns same instance!
  }

  // deepCopyOf - always creates new instance (but same column refs)
  // No instanceof check - always creates new GenericRecordMetadata

  The fix is correct - the bug was that copyOf returned the same instance which got modified. But a better name might be copyOfNew or forceCopyOf to avoid confusion with the true
  deep-copy overload.
```

kafka1991:
> hey @kafka1991 here is AI spot:
> 
> ```
>   1. Naming Inconsistency in GenericRecordMetadata:
> 
>   The new deepCopyOf(RecordMetadata that, int columnCount) method name is misleading:
> 
>   - deepCopyOf(RecordMetadata that) (lines 115-136) truly deep-copies column metadata by creating new TableColumnMetadata objects
>   - deepCopyOf(RecordMetadata that, int columnCount) (lines 103-113) does NOT deep-copy column metadata - it just skips the instanceof shortcut
> 
>   The actual difference is:
>   // copyOf - may return same instance
>   if (that instanceof GenericRecordMetadata) {
>       return (GenericRecordMetadata) that;  // returns same instance!
>   }
> 
>   // deepCopyOf - always creates new instance (but same column refs)
>   // No instanceof check - always creates new GenericRecordMetadata
> 
>   The fix is correct - the bug was that copyOf returned the same instance which got modified. But a better name might be copyOfNew or forceCopyOf to avoid confusion with the true
>   deep-copy overload.
> ```

Thanks for the AI spot! I also reviewed other usages of `deepCopy(RecordMetadata that)` and there’s actually no need to deep-clone the columns in those cases. So I changed it to use a shallow copy of the columns instead.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 16 / 21 (76.19%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/GenericRecordMetadata.java|7|11|63.64%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|3|4|75.00%|
|:large_blue_circle:|io/questdb/griffin/LoopingRecordToRowCopier.java|1|1|100.00%|

PR #6682, opened 2026-01-20
feat(sql): add DESCRIBE syntax for obtaining result set schemas

nwoolmer:
## Summary

Adds a `DESCRIBE` SQL statement that compiles a query and returns the shape of its result set (ordinal_position, column_name, data_type).

**Syntax:**
```sql
DESCRIBE (SELECT * FROM my_table)
DESCRIBE(SELECT id, name FROM my_table JOIN other ON ...)
```

The parser rewrites `DESCRIBE <expr>` into `SELECT * FROM describe(<expr>)`, backed by a `describe(C)` function factory that extracts metadata from the compiled cursor without executing it.

- `DESCRIBE` works with any query expression: joins, subqueries, aggregations, unions, casts, aliases
- The `describe()` function can also be called directly: `SELECT * FROM describe((SELECT ...))`
- `EXPLAIN DESCRIBE ...` is supported
- `describe` is registered as a reserved keyword

## Test plan

- Verify basic DESCRIBE with simple SELECT, joins, aggregations, aliases, subqueries, UNION, casts, expressions
- Verify additional column types: CHAR, VARCHAR, LONG256, IPv4, BINARY, BOOLEAN, BYTE, SHORT, INT, LONG, FLOAT, DOUBLE, STRING, SYMBOL, DATE, TIMESTAMP, UUID
- Verify case insensitivity (`describe` vs `DESCRIBE`)
- Verify `DESCRIBE(...)` without space before parenthesis
- Verify `EXPLAIN DESCRIBE` produces a plan
- Verify `SELECT * FROM describe((...))` works directly
- Verify error on non-existent table
- Verify single column selection


nwoolmer:
@coderabbitai review

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=214394)
**Tested Commit:** `ece4047235005c64f9ae89b0dad98ff407cb2587`

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=215006)
**Tested Commit:** `7b318f5d29eb2976f217b1823ea7b7483e5823f7`

core/src/main/java/io/questdb/griffin/engine/functions/table/DescribeFunctionFactory.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public Function newInstance(int position, ObjList<Function> args, IntList argPositions, CairoConfiguration configuration, SqlExecutionContext sqlExecutionContext) throws SqlException {
+        Function arg = args.getQuick(0);
+        RecordCursorFactory factory = arg.getRecordCursorFactory();
+        if (factory == null) {
+            throw SqlException.$(argPositions.getQuick(0), "cursor expected");
+        }
+        RecordMetadata metadata = GenericRecordMetadata.copyOf(factory.getMetadata());
```
glasstiger: BUG 1: `copyOf()` returns the same instance — must use `copyOfNew()` (HIGH) 

`GenericRecordMetadata.copyOf()` returns the same object when the input is already a `GenericRecordMetadata`. Most compiled query factories return `GenericRecordMetadata`, so in practice metadata is an alias to the inner factory's metadata — not a copy. Then `arg.close()` closes the inner factory.

Fix: replace `copyOf()` with `copyOfNew()`.

core/src/main/java/io/questdb/griffin/engine/functions/table/DescribeFunctionFactory.java
```diff
@@ -0,0 +1,57 @@
...
+        arg.close();
```
glasstiger, line 54: BUG 2: `arg.close()` is unsafe (MEDIUM-HIGH)

No other function factory in the codebase closes its argument function inside `newInstance()`. The function argument lifecycle is managed by the SQL framework, the compiled model tree.

Closing it prematurely here:
  - Releases resources of the inner `RecordCursorFactory` before the outer query model is even finalized
  - Breaks the execution model tree — the arg is part of the `ObjList<Function>` args that the framework may still reference
  - Combined with BUG 1, the metadata reference becomes a dangling alias into a closed factory

  Even after fixing BUG 1 (using `copyOfNew()`), the `arg.close()` should be removed. Let the framework close it.

core/src/main/java/io/questdb/griffin/engine/functions/table/DescribeRecordCursorFactory.java
```diff
@@ -0,0 +1,191 @@
...
+            if (columnIndex == 2) {
+                return SqlUtil.ColumnTypeSymbolTable.INSTANCE;
+            } else {
+                throw new UnsupportedOperationException();
+            }
+        }
+
+        @Override
+        public boolean hasNext() {
+            if (childMetadata.hasColumn(++pos)) {
```
glasstiger: BUG 3: Cursor iteration uses `hasColumn()` instead of `getColumnCount()` (MEDIUM)

`hasColumn()` checks `getQuiet(columnIndex) != null && !isDeleted()`. For query result metadata this mostly works, but:
  - If the metadata came from a table with deleted columns (column gaps), `hasColumn()` returns `false` at the first deleted column and truncates the output — it doesn't skip to the next valid column.
  - The standard QuestDB pattern is `for (i = 0; i < metadata.getColumnCount(); i++)` which handles this correctly.

  Should use `pos < childMetadata.getColumnCount()` as the loop condition.

core/src/main/java/io/questdb/griffin/SqlParser.java
```diff
@@ -2532,6 +2534,38 @@ private QueryModel parseDmlUpdate(
...
+            GenericLexer lexer,
+            SqlParserCallback sqlParserCallback
+    ) throws SqlException {
+        // Rewrite as: SELECT * FROM describe(...) and feed to parseSelect.
+        // Uses a dedicated sink — the lexer's FloatingSequence tokens reference
+        // its content, so it must outlive compilation (no thread-local sink).
+        final CharSequence content = lexer.getContent();
+        final int describePos = lexer.lastTokenPosition();
+        describeSqlSink.clear();
+        describeSqlSink.put("SELECT * FROM ");
```
glasstiger: ISSUE 6: Error positions reference rewritten SQL (LOW)

`parseDescribe()` rewrites the SQL into a `StringSink` and re-parses it. Any parse errors within the inner query will report positions relative to the rewritten `SELECT * FROM describe(...)` text, not the original `DESCRIBE ...` text.
This produces confusing error messages for users.

core/src/main/java/io/questdb/griffin/engine/functions/CursorFunction.java
```diff
@@ -196,6 +197,11 @@ public Long256 getLong256B(Record rec) {
         throw new UnsupportedOperationException();
     }
 
+    @Override
+    public RecordMetadata getMetadata() {
```
glasstiger: ISSUE 7: `CursorFunction.getMetadata()` override is unused (NITS)

`DescribeFunctionFactory` uses `arg.getRecordCursorFactory().getMetadata()`, not `arg.getMetadata()`. This override is not exercised by any code in this PR.

core/src/test/java/io/questdb/test/griffin/DescribeStatementTest.java
```diff
@@ -0,0 +1,461 @@
...
+ *  limitations under the License.
+ *
+ ******************************************************************************/
+
+package io.questdb.test.griffin;
+
+import io.questdb.test.AbstractCairoTest;
+import org.junit.Test;
+
+public class DescribeStatementTest extends AbstractCairoTest {
```
glasstiger, line 30: Missing coverage:

  - DESCRIBE SELECT * FROM t (non-parenthesized form without inner parens) — all tests use either DESCRIBE (SELECT ...) or DESCRIBE(SELECT ...). There's no test for the else branch in
  parseDescribe.
  - Error cases: only one error test (non-existent table). Missing: DESCRIBE with no argument, DESCRIBE with a non-query (e.g., DESCRIBE 42), DESCRIBE of a DESCRIBE.
  - Tables with deleted columns — would expose BUG 3.
  - Concurrent/repeated cursor iteration (toTop() + re-iterate) — the testDescribeRepeated test calls DESCRIBE twice but doesn't test cursor reuse within a single compiled plan.
  - PG wire protocol — no test that DESCRIBE works over the PostgreSQL wire protocol path.
  - data_type symbol filtering — testDescribeResultFiltering filters by data_type = 'INT' which exercises the symbol table lookup, but there's no test for an unknown type name.

core/src/main/java/io/questdb/griffin/engine/functions/table/DescribeFunctionFactory.java
nwoolmer: Investigated — `arg.close()` is intentional and required here.

`FunctionParser.createFunction()` clears the args list after `newInstance()` returns (line 599: `args.clear()`) but does not close individual `Function` objects. The contract is that either the returned function holds a reference to its args (so they get closed transitively) or the factory closes args it does not pass through.

Here, the returned `CursorFunction(new DescribeRecordCursorFactory(metadata))` does not reference `arg` — it only extracted metadata via `copyOfNew()`. Without `arg.close()`, the inner `RecordCursorFactory` leaks. Removing it causes `assertMemoryLeak()` failures in `testDescribeJoin` and `testDescribeSelectWithAggregation`.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 132 / 150 (88.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlUtil.java|4|7|57.14%|
|:large_blue_circle:|io/questdb/griffin/model/DescribeModel.java|7|9|77.78%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/DescribeRecordCursorFactory.java|41|52|78.85%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|7|8|87.50%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/DescribeFunctionFactory.java|8|9|88.89%|
|:large_blue_circle:|io/questdb/griffin/SqlParser.java|24|24|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableColumnMetadata.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|27|27|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlKeywords.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/griffin/model/ExecutionModel.java|1|1|100.00%|

nwoolmer:
Closing — these commits are now merged into the `errand` branch, which will be merged back into `master` in due course. That path also carries a follow-up cleanup that exposes `data_type` as `STRING` natively (instead of `SYMBOL` backed by a synthetic `ColumnTypeSymbolTable` whose bit-packed keys broke downstream `SymbolTable` consumers).

Leaving the `nw_pl` branch in place for now in case we want to reference the original commits.

PR #6686, opened 2026-01-21
chore(core): prevent mat view rename if mat view definition does not exist yet

glasstiger:
This change helps to fix a race condition in replication.
More details in ent PR description.

Required by https://github.com/questdb/questdb-enterprise/pull/874

----

Also fixed a deadlock which caused a build failure on this PR.
Claude provided the fix:
```
Found the issue. The deadlock occurs in TlsProxy$Pump.shutdown() at line 242-246:                                                                                            
                                                                                                                                                                               
  private void shutdown() {                                                                                                                                                    
      shutdownRequested = true;                                                                                                                                                
      owningThread.interrupt();                                                                                                                                                
      TestUtils.unchecked(() -> owningThread.join());  // <-- hangs here                                                                                                       
  }                                                                                                                                                                            
                                                                                                                                                                               
  Root cause: Thread.interrupt() does not interrupt blocking socket I/O operations in Java. The back-to-front thread is blocked in InputStream.read() at line 210:             
                                                                                                                                                                               
  i = from.read(buffer);                                                                                                                                                       
                                                                                                                                                                               
  Regular socket streams are not interruptible. The interrupt() call has no effect, and the thread remains blocked on the read, so join() waits forever.                       
                                                                                                                                                                               
  The fix: Close the input stream before joining the thread. This will cause read() to throw an IOException, allowing the loop to exit:                                        
                                                                                                                                                                               
  private void shutdown() {                                                                                                                                                    
      shutdownRequested = true;                                                                                                                                                
      closeQuietly(from);  // Close stream to unblock read()                                                                                                                   
      owningThread.interrupt();                                                                                                                                                
      TestUtils.unchecked(() -> owningThread.join());                                                                                                                          
  }                                                                                                                                                                            
                                                                                                                                                                               
  Alternatively, Link.shutDown() could close the sockets first (like Link.kill() does):                                                                                        
                                                                                                                                                                               
  private void shutDown() {                                                                                                                                                    
      closeQuietly(frontend);                                                                                                                                                  
      closeQuietly(backend);                                                                                                                                                   
      frontendToBackend.shutdown();                                                                                                                                            
      backendToFrontend.shutdown();                                                                                                                                            
  }                                                                                                                                                                            
                                                                                                                                                                               
  This is an intermittent failure because it depends on timing - if the remote side closes the connection first, read() returns -1 and the thread exits normally. But if stop()
   is called while data is still being read, the thread blocks indefinitely.        
```

### DIFF
--- a/core/src/main/java/io/questdb/cairo/GenericRecordMetadata.java
+++ b/core/src/main/java/io/questdb/cairo/GenericRecordMetadata.java
@@ -79,7 +79,10 @@ public class GenericRecordMetadata extends AbstractRecordMetadata {
                 return (GenericRecordMetadata) that;
             }
             GenericRecordMetadata metadata = copyOfSansTimestamp(that, columnCount);
-            metadata.setTimestampIndex(that.getTimestampIndex());
+            int timestampIndex = that.getTimestampIndex();
+            if (timestampIndex < columnCount) {
+                metadata.setTimestampIndex(timestampIndex);
+            }
             return metadata;
         }
         return null;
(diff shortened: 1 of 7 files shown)


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|10|10|100.00%|

PR #6688, opened 2026-01-21
perf(core): avoid contention across concurrent queries on expensive file operation

mtopolnik:
Two changes:

1. Avoid closing a partition when it's not actually open (is in standby for lazy open)
2. Perform `mmap` outside of `synchronized` block in `MmapCache`

### DIFF
--- a/core/src/main/java/io/questdb/std/MmapCache.java
+++ b/core/src/main/java/io/questdb/std/MmapCache.java
@@ -376,7 +376,13 @@ public final class MmapCache {
             // to keep. We can't keep the existing one if it's too small.
             MmapCacheRecord existingRecord = mmapFileCache.valueAt(fdMapIndex);
             if (existingRecord.length < len) {
-                // Existing mapping is too small - replace with ours
+                // Existing mapping is too small - replace it with ours.
+                // There are two caches: file cache and address cache. We'll put the entry
+                // with our larger mapping into the file cache, so it gets used from now on.
+                // However, some threads may already have grabbed the smaller mapping, and
+                // are using it. Once all its users are done with it and unmap it, that will
+                // remove it from the address cache. Therefore, we add our address to the
+                // address cache, and leave the other one there as well.
                 MmapCacheRecord record = createMmapCacheRecord(fd, mmapCacheKey, len, address, memoryTag);
                 mmapFileCache.putAt(fdMapIndex, mmapCacheKey, record);
                 mmapAddrCache.put(address, record);


glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 55 / 61 (90.16%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/MmapCache.java|54|60|90.00%|

core/src/main/java/io/questdb/std/MmapCache.java
```diff
@@ -357,6 +334,108 @@ private static void munmapTaskConsumer(MunmapTask task) {
...
+            redundantTag = memoryTag;
+            returnAddress = existingRecord.address;
+        }
+
+        // We lost the race, clean up redundant mapping outside the lock
+        int result = Files.munmap0(redundantAddress, redundantLen);
+        if (result != -1) {
+            Unsafe.recordMemAlloc(-redundantLen, redundantTag);
+        } else {
+            throw CairoException.critical(Os.errno())
```
jerrinot: we are throwing even the reference count was already increased at the line 387. 
this means the callsite will consider the `mmap()` call as unsuccessful, but `MmapCache` internal booking won't release it until we unmap it too. (Which we cannot since the callsite won't even get the address) -> leak. an edge-case (failing munmap()), but still a leak.

core/src/main/java/io/questdb/std/MmapCache.java
```diff
@@ -357,6 +334,106 @@ private static void munmapTaskConsumer(MunmapTask task) {
...
+            returnAddress = existingRecord.address;
+        }
+
+        // We lost the race, clean up redundant mapping outside the lock
+        try {
+            // This submits an async unmap operation, and only in an extreme (unmap queue full) case
+            // will do it sync with the potential to throw CairoException
+            unmap0(redundantAddress, redundantLen, redundantTag);
+        } catch (CairoException e) {
+            LOG.error().$("failed to unmap redundant mapping after losing race [message=").$(e.getMessage()).I$();
```
jerrinot: shouldn't this be 'critical'? this is almost certainly a bug. critical levels are transformed to slack notifications, etc. 

edit: `munmapTaskConsumer()` logs it as critical. so we should be consistent here too.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 54 / 58 (93.10%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/MmapCache.java|53|57|92.98%|

PR #6693, opened 2026-01-22
test(sql): cover LT join crash when timestamp is used in ON clause

Avi112005:
This PR adds a regression test for issue #6637.

The test reproduces a case where an LT JOIN crashes with an AssertionError when a designated timestamp column is used in the ON clause and the full-fat join path is selected (forced via SAMPLE BY). Instead of crashing, the engine should reject this query with a proper SqlException.

The test documents the expected behavior and ensures this case is handled safely going forward.


Avi112005:
Reopening after updating the test.
The test now focuses only on preventing crashes (AssertionError) and does not enforce specific SQL semantics.

bluestreak01:
the issue has been fixed

PR #6706, opened 2026-01-26
fix(pgwire): prepared batch with same SQL returns corrupted results

### LINKED ISSUE #6703: Prepared batch statements returns incorrect results
### To reproduce

Consider the following C# program:

```C#
using Npgsql;

await using var dataSource = new NpgsqlDataSourceBuilder(
    "Host=127.0.0.1;Port=8812;Username=admin;Password=quest;Database=qdb;ServerCompatibilityMode=NoTypeLoading;"
).Build();

await using var connection = await dataSource.OpenConnectionAsync();
await using var batch = new NpgsqlBatch(connection);

var cmd1 = new NpgsqlBatchCommand("SELECT @time;");
cmd1.Parameters.AddWithValue("time", DateTime.UtcNow);
batch.BatchCommands.Add(cmd1);

var cmd2 = new NpgsqlBatchCommand("SELECT @time;");
cmd2.Parameters.AddWithValue("time", DateTime.UtcNow);
batch.BatchCommands.Add(cmd2);

await batch.PrepareAsync(); // It works if we do not prepare the batch

await using var reader = await batch.ExecuteReaderAsync();

while (await reader.ReadAsync())
{
    Console.WriteLine(reader.GetDateTime(0));
}
await reader.NextResultAsync();
while (await reader.ReadAsync())
{
    Console.WriteLine(reader.GetDateTime(0));
}
```

I get this exception when processing the second result:

> System.InvalidCastException
  HResult=0x80004002
  Message=Out of range of DateTime (year must be between 1 and 9999).
  Source=Npgsql
  StackTrace:
   at Npgsql.Internal.Converters.PgTimestamp.Decode(Int64 value, DateTimeKind kind, Boolean dateTimeInfinityConversions)
   at Npgsql.Internal.Converters.DateTimeConverter.ReadCore(PgReader reader)
   at Npgsql.Internal.PgBufferedConverter`1.Read(PgReader reader)
   at Npgsql.NpgsqlDataReader.GetFieldValueCore[T](Int32 ordinal)
   at Npgsql.NpgsqlDataReader.GetDateTime(Int32 ordinal)
   at Program.<<Main>$>d__0.MoveNext() in C:\Source\ConsoleApp1\ConsoleApp1\Program.cs:line 29
   at Program.<<Main>$>d__0.MoveNext() in C:\Source\ConsoleApp1\ConsoleApp1\Program.cs:line 27
   at Program.<<Main>$>d__0.MoveNext() in C:\Source\ConsoleApp1\ConsoleApp1\Program.cs:line 27
   at Program.<<Main>$>d__0.MoveNext() in C:\Source\ConsoleApp1\ConsoleApp1\Program.cs:line 27
   at Program.<<Main>$>d__0.MoveNext() in C:\Source\ConsoleApp1\ConsoleApp1\Program.cs:line 27
  This exception was originally thrown at this call stack:
    [External Code]
Inner Exception 1:
ArgumentOutOfRangeException: Ticks must be between DateTime.MinValue.Ticks and DateTime.MaxValue.Ticks. (Parameter 'ticks')

Maybe related issue #6123.

### QuestDB version:

9.3.1

Windows

ntfs

### Full Name:

Rasmus Melchior Jacobsen

### Affiliation:

Utiliread ApS

- [x] Yes, I have

### Additional context

_No response_

kafka1991:
When executing multiple commands with the same SQL but different parameters caused the second command to return                                                                                                            corrupted data.                                                                                                                                                                                       

**Root cause:** `msgBindCopySelectFormatCodes()` checked `factory != null` to determine if format codes should be set. When a prepared statement is reused via `copyIfExecuted()`, the new entry has `factory=null` (not copied), causing format codes to be skipped. The client then misinterpreted binary/text format.                                                                                                                                                        

**Fix:** Replace `factory != null` with `hasResultSet()` which uses `sqlType`  

Fixes #6703


core/src/main/java/io/questdb/cutlass/pgwire/PGPipelineEntry.java
```diff
@@ -1311,6 +1314,12 @@ private short getPgResultSetColumnFormatCode(int columnIndex, int columnType) {
         return 1;
     }
 
+    private boolean hasResultSet() {
```
jerrinot, line 1317: consider using this method at https://github.com/questdb/questdb/blob/d390ccd9531a327626942d5a5b0d87cb344f72b0/core/src/main/java/io/questdb/cutlass/pgwire/PGPipelineEntry.java#L3588

instead of `if (cursor == null || isError()) {`
there could be semantically more precise
`if (!hasResultSet() || isError()) {`
since that's what the guess does: checking if there is a result set (and thus binding variable service is used at SYNC time. wdyt?

kafka1991: > consider using this method at
> 
> https://github.com/questdb/questdb/blob/d390ccd9531a327626942d5a5b0d87cb344f72b0/core/src/main/java/io/questdb/cutlass/pgwire/PGPipelineEntry.java#L3588
> 
> instead of `if (cursor == null || isError()) {` there could be semantically more precise `if (!hasResultSet() || isError()) {` since that's what the guess does: checking if there is a result set (and thus binding variable service is used at SYNC time. wdyt?

It's a good change for me, fix it

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGPipelineEntry.java|3|3|100.00%|

core/src/main/java/io/questdb/cutlass/pgwire/PGPipelineEntry.java
jerrinot: thank you!

PR #6709, opened 2026-01-27
fix(core): update SAMPLE BY to use long

### LINKED ISSUE #6202: Allow SAMPLE BY to use a long, rather than an INT for the units

hlothaire:
### Issue

  Fixes #6202 

  ### Problem

  The `SAMPLE BY` statement only accepted `INT` values for intervals, limiting the maximum value to `2,147,483,647`. This is problematic for nanosecond precision since:

  - `Integer.MAX_VALUE` nanoseconds ≈ **2.1 seconds**
  - Users cannot specify intervals like `SAMPLE BY 3000000000n` 

  Error message users received:
  invalid tolerance value [value=3000000000n, maximum=2147483647n]

  ### Solution

  Changed interval parsing from `int` to `long` to support the full range of nanosecond intervals.

  ### Changes

  | File | Change |
  |------|--------|
  | `TimestampSamplerFactory.java` | Changed `Numbers.parseInt()` → `Numbers.parseLong()` and `int maxValue` → `long maxValue` |
  | `SqlCodeGenerator.java` | Updated all callers to use `Long.MAX_VALUE` and `Numbers.LONG_NULL` |
  | `CreateMatViewOperationImpl.java` | Updated to use `Numbers.LONG_NULL` |
  | `TimestampSamplerFactoryTest.java` | Added tests for large nanosecond intervals |

  ### Testing

  - Added `testLargeNanosecondInterval()` - verifies 3 billion nanoseconds works correctly
  - Added `testParseIntervalLargeValue()` - verifies parsing "3000000000n" returns correct value
  - All existing `SampleByTest` and `SampleByNanoTimestampTest` tests pass


bluestreak01:
Thank you for the contribution. I would like to make sure you read top of this document: https://github.com/questdb/questdb/blob/master/CONTRIBUTING.md

hlothaire:
I’ve read the contributing guide  and made sure this PR follows it. Please let me know if i missed anything.

bluestreak01:
to accept the PR we need CI to pass, questdb is production level system that must not regress

Issue #6583, opened 2025-12-30
CREATE TABLE AS SELECT does not support all possible casting


jerrinot:
### To reproduce

`CREATE AS SELECT` has a special syntax of casting. This is a valid syntax: `create table t as (select * from long_sequence(5)), cast(x as long);` (note the comma before cast)

However, it does not allow all supported casting.  This works fine:
`create table t as (select cast (x as varchar) as x from long_sequence(5));`
but this does not:
`create table t as (select x from long_sequence(5)), cast(x as varchar);`
-> `unsupported cast [column=x, from=LONG, to=VARCHAR]`

As a user I would expect that the special casting syntax in CREATE AS SELECT would support all casting combinations allowed in regular casting expressions. 

### QuestDB version:

9.2.3

### OS, in case of Docker specify Docker and the Host OS:

NA

### File System, in case of Docker specify Host File System:

NA

### Full Name:

Jaromir Hamala

### Affiliation:

QuestDB

### Have you followed Linux, MacOs kernel configuration steps to increase Maximum open files and Maximum virtual memory areas limit?

- [x] Yes, I have

R1patil:
Hi @jerrinot and QuestDB team,

This is my first contribution to QuestDB. I'd like to work on this issue #6583 to add support for all casting combinations in the CREATE TABLE AS SELECT special syntax (specifically making cast(x as varchar) work in the trailing cast clause).

I'll start by setting up the development environment, reproducing the issue locally, and looking into the SQL parser/compiler code where the casting is handled.

Could you please assign this issue to me? Any pointers on where to start in the codebase would be greatly appreciated!

Thanks,
R1patil

bhavik-sahu:
Hi @jerrinot ,

I worked on this issue and opened a PR with a fix and regression test:

Fixes LONG -> VARCHAR casting in CTAS row copiers.

PR: #7213 

Would appreciate a review when you have time. Thanks!

bhavik-sahu:
Hi @jerrinot,

Just following up on this PR (#7213).
The workflows are currently awaiting approval, so the CI checks haven’t started yet.

When you have a moment, could you please approve/run the workflows and take a look at the PR?
Thanks!

NikhilDendeti:
Hi, I'd like to work on this issue if it's still open. From the description, it looks like the cast(col as type) syntax used in CREATE TABLE AS SELECT supports a narrower set of conversions than a regular inline CAST expression does — happy to dig into why and align the two.

[...]

PR #6732, opened 2026-01-30
fix(sql): add casts for create table as select

AnirudhB3000:
**Description:**
This PR fixes the "unsupported cast" error as seen in issue #6583  wherein valid SQL statements such as: `create table t as (select x from long_sequence(5)), cast(x as varchar);` resulted in `unsupported cast [column=x, from=LONG, to=VARCHAR]` error being shown.

**Key Fixes:**
- `LONG` -> `STRING` and `LONG` -> `VARCHAR` casts now work in CTAS-specific cast clauses (`CREATE TABLE t AS (...), cast(x AS varchar)`) which failed with `unsupported cast`.
- `isCompatibleCast` now explicitly allows casts to `STRING` and `VARCHAR` from `LONG`, `CHAR`, `SYMBOL`, `UUID` which had bytecode support.
- `transferLongToVarcharCol()` and `transferLongToStrCol()` were added to the copier to use bytecode for copying.
 

**New supported casts:**
| Source | → STRING | → VARCHAR |
|---|---|---|
| CHAR | ✅ | ✅ |
| SYMBOL | ✅ | ✅ |
| UUID | ✅ | ✅ |
| LONG | ✅ | ✅ |

**Tests:**
- New test file has been created: `CastInCtasTest.java`.
- All casts in the aforementioned table has it's corresponding test

**Additional changes:**
- Removed `assertCastLongFail(ColumnType.STRING)` since `LONG` -> `STRING` is now a supported cast in CTAS.
- Added cases `ColumnType.VARCHAR` and `ColumnType.STRING` branches in the `copyFromLong()` method, delegating to the same `transferLongToVarcharCol()` or transferLongToStrCol()` helpers.

**Verification:**
Run `mvn test -pl core -Dtest="CastInCtasTest"`


bluestreak01:
@Coderabbit review this PR critically

AnirudhB3000:
Hello @bluestreak01,

I'm mentioning the reasons for the issue the review bot discovered. For problems 1, 7 and 8, I'm working on a fix, I hope to have it ready soon.

I'd appreciate some input on the other problems, especially problem 5. I could include my reasons as comments in the respective functions, but if you think any changes are required, please let me know.

1. **Test Coverage Appears Insufficient**: 
Work in progress, I'm adding the tests, and could have them ready soon.

2. **Defensive Null Checks Raise Concerns**: 
   1. `sqlExecutionContext` can be Null in the following cases: 
       1. **Constant expression evaluation at parse time**: when evaluating expressions as `cast('123` as decimal), the parser precomputes the decimal, resulting in Null. 
       2. **Metadata operations**: when only metadata is used without queries.

   2. The creation of a new `Decimal256()` instance is needed for the cast function as a temporary buffer during constant expression evaluation. This instance creation is safe  and intentional as it's an intentional design pattern for parse time operation
       
   3.  Repeatedly instantiating `Decimal256()` objects has minimal impacts as this                happens only during query parsing and the allocations is negligible compared to actual query execution costs.

3. **Inconsistent `tagOf()` Usage Pattern**:
A short example of when `tagOf()` is used:
   - `STRING`, `VARCHAR`, `INT`: **No**.
   - `TIMESTAMP`: **Yes**, timezone bits must be stripped.
   - `GEOHASH`: **Yes**, precision bits must be stripped.
   - `DECIMAL`: **Yes**, scale bits must be stripped.
   - `ARRAY`, **Yes**, element type bits must be stripped.

4. **GeoHash Cast Logic Appears Before Switch Statement**: 
GeoHash types encode precision bits at many levels and they must be checked to prevent unsafe widening. Aside from these considerations, `CastGeoHashToGeoHashFunctionFactory.newInstance()` needs both source and target params to create the cast.

5. **Massive Method Size - Maintainability Concern**:
The added new switches make this function massive, yes, I could use comments to create sections to differentiate the casts and sections. Another option would be to create a hash table with all casts and their functions but that is significantly harder to build.

6. **`CHAR` Casting Semantics Changed**:
   - The previous ASCII based approach was undocumented and the casting implications are assumed to be unintentional. The new digit parsing method is more direct wherein `5` -> 5.
   - In the module's files, no other code relies on CHAR to temporal casts, and an explicit ascii() function exists in SQL when wanted by the user.

7. **Narrowing Cast Prevention Incomplete?**:
   - The `SHORT` -> `BYTE` cast is allowed as `SHORT` rarely exceeds `BYTE` in practice. I can disable it in a follow up commit. 
   - The `INT` -> `SHORT` and `INT` -> `BYTE` is something I overlooked. Will be disabled in the next commit.

8. **`VARCHAR` to `LONG` Cast Uses Factory Pattern**:
It's an issue I didn't notice at first. Will work on it soon.

AnirudhB3000:
@coderabbitai review

AnirudhB3000:
@coderabbitai review

AnirudhB3000:
@coderabbit review

Issue #5751, opened 2025-06-17
AssertionError in UNION ALL query


puzpuzpuz:
### To reproduce

Run this query on demo:
```sql
DECLARE @recent_data_boundary := dateadd('d', -1, date_trunc('day',now())),
        @historical_data_boundary := dateadd('m', -6, date_trunc('month',now()))
WITH historical_and_recent_trades AS (
  SELECT
    timestamp, symbol,
    first(price) AS open,
    max(price) as high,
    min(price) as low,
    last(price) AS close,
    sum(amount) AS volume
    FROM trades
    WHERE
      timestamp >= @recent_data_boundary
    SAMPLE BY 10m
  UNION ALL
  SELECT *
    FROM trades_OHLC_15m
    WHERE
      timestamp > @historical_data_boundary
      AND timestamp < @recent_data_boundary
)
SELECT open, high FROM historical_and_recent_trades order by timestamp desc
```

and observe this error:
```
java.lang.AssertionError
	at io.questdb.griffin.SqlCodeGenerator.checkIfSetCastIsRequired(SqlCodeGenerator.java:887)
	at io.questdb.griffin.SqlCodeGenerator.generateSetFactory(SqlCodeGenerator.java:5158)
	at io.questdb.griffin.SqlCodeGenerator.generateQuery(SqlCodeGenerator.java:3356)
	at io.questdb.griffin.SqlCodeGenerator.generateSubQuery(SqlCodeGenerator.java:5266)
	at io.questdb.griffin.SqlCodeGenerator.generateSelectChoose(SqlCodeGenerator.java:3843)
	at io.questdb.griffin.SqlCodeGenerator.generateSelect(SqlCodeGenerator.java:3813)
	at io.questdb.griffin.SqlCodeGenerator.generateQuery0(SqlCodeGenerator.java:3366)
	at io.questdb.griffin.SqlCodeGenerator.generateQuery(SqlCodeGenerator.java:3354)
	at io.questdb.griffin.SqlCodeGenerator.generateSubQuery(SqlCodeGenerator.java:5266)
	at io.questdb.griffin.SqlCodeGenerator.generateSelectChoose(SqlCodeGenerator.java:3843)
	at io.questdb.griffin.SqlCodeGenerator.generateSelect(SqlCodeGenerator.java:3813)
	at io.questdb.griffin.SqlCodeGenerator.generateQuery0(SqlCodeGenerator.java:3366)
[...]

PR #6744, opened 2026-02-04
fix(core): fix assertion errors in UNION ALL with column count mismatch

mtopolnik:
Fixes https://github.com/questdb/questdb/issues/5751

Two related bugs in `SqlOptimiser.propagateTopDownColumns0()` cause the same
`AssertionError` in `SqlCodeGenerator.checkIfSetCastIsRequired` — both stem
from UNION sibling models ending up with different `topDownColumns` counts.

## Bug 1: WHERE clause on aliased columns

WHERE clause and timestamp column literals were emitted to union model
branches by name resolution. Since UNION matches columns by position, not
name, the same alias (e.g. "ts") can resolve to different column indices in
different branches. This caused one branch to receive extra top-down columns.

Reproducer:

```sql
SELECT ts FROM (
  SELECT name1, ts1, sym1 ts FROM t1
  UNION ALL
  SELECT name2, ts2 ts, sym2 FROM t2
) WHERE ts IN '2025-12-01T01;2h'
```

**Fix:** Remove the name-based emission loops for both WHERE clause and
explicit timestamp literals. The existing index-based propagation lower in
the method already correctly handles union branches by deriving positional
indices from the left branch's top-down columns.

## Bug 2: GROUP BY / SAMPLE BY key column propagation

When a GROUP BY/SAMPLE BY model is the first UNION member and the outer
query selects only a subset of columns, the GROUP BY model adds its
non-aggregate key columns (e.g. `timestamp`) to its own `topDownColumns` but
doesn't propagate them to union siblings.

Reproducer (from #5751):

```sql
SELECT high FROM (
  SELECT timestamp, max(price) AS high FROM trades SAMPLE BY 1m
  UNION ALL
  SELECT timestamp, high FROM trades_agg
)
```

**Fix:** After the GROUP BY model adds non-aggregate key columns to its own
`topDownColumns`, propagate the same columns to all union siblings whose
`topDownColumns` are already populated.

## Test plan

- [x] `SqlCodeGeneratorTest#testUnionAllAssertionError` — regression test for Bug 1
- [x] `SqlCodeGeneratorTest#testUnionAllSampleByAssertionError` — regression test for Bug 2
- [x] `UnionTest` — full suite passes
- [x] `NestedSetOperationTest` — full suite passes

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -54,6 +54,7 @@ Use modern Java features:
 - use multiline strings for longer statements (multiple INSERT rows, complex queries),
   as well as to assert multiline query results
 - use the expr::TYPE syntax for casts
+- use underscore to separate thousands in numbers with 5 digits or more
 
 ### Building
 


bluestreak01:
@mtopolnik this does NOT fix the referenced issue

mtopolnik:
Reference removed, the issue solved by this PR is related, but different. This PR doesn't have an open issue it fixes; I discovered this problem and fixed it immediately without creating an issue.

mtopolnik:
Update to previous comment: added another fix, so now the referenced issue #5751 is fixed as well.

core/src/test/java/io/questdb/test/griffin/engine/SqlCodeGeneratorTest.java
```diff
@@ -8864,6 +8864,36 @@ public void testTimestampPropagation() throws Exception {
         });
     }
 
+    @Test
+    public void testUnionAllAssertionError() throws Exception {
+        execute("create table t1 (name1 varchar, ts1 timestamp, sym1 symbol) timestamp(ts1)");
+        execute("create table t2 (name2 varchar, ts2 timestamp, sym2 symbol) timestamp(ts2)");
+        assertQuery(
```
puzpuzpuz: nit: surronding `assertMemoryLeak` and then `assertQueryNoLeakCheck` won't hurt in all new tests.

core/src/test/java/io/questdb/test/griffin/engine/SqlCodeGeneratorTest.java
```diff
@@ -8864,6 +8864,36 @@ public void testTimestampPropagation() throws Exception {
...
+                "ts\n",
+                "select ts from (select name1, ts1, sym1 ts from t1 union all select name2, ts2 ts, sym2 from t2) where ts in '2025-12-01T01;2h';",
+                "",
+                false,
+                false
+        );
+    }
+
+    @Test
+    public void testUnionAllSampleByAssertionError() throws Exception {
```
puzpuzpuz: GROUP BY fix looks incomplete. The following test still reproduces `AssertionError`.

```java
    @Test
    public void testUnionAllSampleBy2() throws Exception {
        assertMemoryLeak(() -> {
            execute("create table trades (price double, timestamp timestamp) timestamp(timestamp) partition by day");
            execute("create table trades_agg (high double, timestamp timestamp)");
            assertQueryNoLeakCheck(
                    "high\n",
                    "select high from (" +
                            "select timestamp, high from trades_agg " +
                            "union all " +
                            "select timestamp, max(price) as high from trades sample by 1m" +
                            ")",
                    null,
                    false,
                    false
            );
        });
    }
```

A possible fix: when the GROUP BY block adds key columns, propagate not just forward through model.getUnionModel(), but also backward to earlier siblings. This would require the outer model's index-based loop or a separate pass to ensure all union members stay in sync.

core/src/test/java/io/questdb/test/griffin/engine/SqlCodeGeneratorTest.java
```diff
@@ -8864,6 +8864,36 @@ public void testTimestampPropagation() throws Exception {
         });
     }
 
+    @Test
+    public void testUnionAllAssertionError() throws Exception {
```
puzpuzpuz: nit: the test names are so-so since no assertion error is expected after the fixes.

puzpuzpuz: Could also add tests that run queries over non-empty tables? This way we'll verify that the queries not only compile, but also produce correct results.

mtopolnik: Addressed in 4e3ea24f24a1691d8164678c44221b5174d00e3d.

core/src/test/java/io/questdb/test/griffin/engine/SqlCodeGeneratorTest.java
mtopolnik: Addressed in 4e3ea24f24a1691d8164678c44221b5174d00e3d.

core/src/test/java/io/questdb/test/griffin/engine/SqlCodeGeneratorTest.java
mtopolnik: Addressed in 4d3a8f5d09371fc94e712c6f36452c8d4291a344.

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|11|11|100.00%|

core/src/test/java/io/questdb/test/griffin/engine/SqlCodeGeneratorTest.java
mtopolnik: Addressed in c1e092d10a593499a228ad460cc309538548c915

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|11|11|100.00%|

Issue #2302, opened 2022-07-07
SQL: push UNION result filters into each part


ideoma:
### Is your feature request related to a problem?

Union used to push filters into the sub queries but it was abandoned in 6.4.2 because the tables can have different column names.

### Describe the solution you'd like.

When query

```

SELECT * FROM 
bus INNER JOIN (
  (SELECT datetime,"from" as busId FROM branch2 WHERE logical 
  ) 
    UNION
  (SELECT datetime,"to" as busId FROM branch2 WHERE logical
  )
) 
 u ON bus.busId = u.busId AND bus.datetime = u.datetime 
WHERE bus.datetime = '2026-10-05T09:55'

```

it should push `datetime = '2026-10-05T09:55'` filter into UNION parts like this

```

SELECT * FROM 
bus INNER JOIN (
  (SELECT datetime,"from" as busId FROM branch2 WHERE logical AND datetime = '2026-10-05T09:55'
  ) 
    UNION
  (SELECT datetime,"to" as busId FROM branch2 WHERE logical AND datetime = '2026-10-05T09:55'
  )
) 
 u ON bus.busId = u.busId AND bus.datetime = u.datetime 
WHERE bus.datetime = '2026-10-05T09:55'

```

### Additional context.

Reported in #2182

ideoma:
> Is this a duplication for #2210?

No, I don't think so

PR #6745, opened 2026-02-04
perf(sql): improve performance of UNION and friends by pushing down timestamp filter into subqueries

mtopolnik:
Fixes https://github.com/questdb/questdb/issues/2302

## Summary

Push designated-timestamp filters from outside a UNION / UNION ALL / EXCEPT / INTERSECT subquery into each branch of the set operation. This enables per-branch partition pruning on time-series tables.

The optimizer only pushes filters that **exclusively reference the designated timestamp column**. It deliberately excludes non-timestamp filters (value columns, expressions, mixed references) — they risk type mismatches and column resolution issues across branches with different schemas, and don't unlock partition pruning anyway. The outer filter always stays on the wrapper model, so correctness is unchanged.

### Production code change (`SqlOptimiser.java`)

The substantive logic is small — two sites:

**Site 1 — the `else if` arm** that routes union-bearing nested models to the new method instead of blocking pushdown:

```java
} else if (nested.getUnionModel() != null) {
    tryPushFilterIntoSetOperationBranches(node, parent, nested);
    addWhereNode(parent, node);
}
```

`addWhereNode(parent, node)` always keeps the parent filter, so correctness holds regardless of what happens in the per-branch push.

**Site 2 — `tryPushFilterIntoSetOperationBranches`**: Gates on the designated timestamp, then walks the union chain, deep-clones the filter for each branch, rewrites column references via the parent's alias maps, and adds the filter to the branch. Safety mechanisms:

1. **Timestamp-only gate**: `findTimestamp` + `referencesOnlyTimestampAliasRecursive` reject any filter that references non-timestamp columns.
2. **Per-branch semantic guards**: skip branches with LATEST BY, LIMIT, or non-pushable SAMPLE BY — mirrors the checks for the first branch.
3. **Column existence check**: `allLiteralsMatchBranch` skips branches where rewritten column names don't exist (e.g. branches with different or missing aliases).
4. **`NonLiteralException` catch**: skips branches where the timestamp column maps to a computed expression (e.g., `dateadd(...)` on the timestamp).
5. **`deepClone`**: gives each branch its own expression copy, preventing cross-branch mutation.

### Composite filters

For `WHERE ts IN '2025-12-01T01;2h' AND x > 5`, the optimizer splits the top-level AND into separate conjuncts and processes each independently: it pushes the timestamp conjunct into branches, while `x > 5` stays at the parent level only.

## Test plan

- [x] 8 new `SqlParserTest` plan-level tests covering timestamp pushdown through UNION, UNION ALL, EXCEPT, INTERSECT, 3-branch chains, non-literal branches, same-table branches, and mixed timestamp/non-timestamp filters
- [x] 4 `SqlParserTest` cases verifying that non-timestamp filters are NOT pushed into branches
- [x] 1 new `UnionTest` runtime regression test verifying that filter pushdown into a UNION ALL branch with SAMPLE BY does not change aggregation semantics
- [x] Existing `UnionAllCastTest` cases pass (type mismatches and column resolution across branches with different schemas)


mtopolnik:
@coderabbit Please review this PR. One of commits is just auto-reordering of code. Please try to ignore the changes of that commit in your review. Focus your review on all possible edge cases you can think of that would make this change break SQL.

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -1338,6 +1351,26 @@ private boolean allAdviceIsForThisTable(QueryModel model, ObjList<ExpressionNode
         return true;
     }
 
+    private boolean allLiteralsMatchBranch(ExpressionNode node, LowerCaseCharSequenceObjHashMap<QueryColumn> branchColumns) {
+        if (node == null) {
+            return true;
+        }
+        if (node.type == LITERAL) {
+            int len = node.token.length();
+            int dot = Chars.indexOf(node.token, 0, len, '.');
```
nwoolmer: lastIndexOf?

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -2019,16 +2052,39 @@ private void collectSameModelProjectionColumns(ExpressionNode node, QueryModel m
...
+            return false;
+        }
+
+        if (node.type == FUNCTION) {
+            // Check for time functions like now(), sysdate
+            if (isTimeFunctionKeyword(node.token)) {
+                return true;
+            }
+            // Check for dateadd that doesn't reference the timestamp
+            if (Chars.equalsLowerCaseAscii(node.token, "dateadd")) {
```
nwoolmer: nit: maybe make it a SqlKeyword

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -2019,16 +2052,39 @@ private void collectSameModelProjectionColumns(ExpressionNode node, QueryModel m
...
+     * Disallowed functions:
+     * - now(), sysdate (time functions)
+     * - dateadd that doesn't reference the timestamp column
      */
-    private static boolean isTimeFunctionKeyword(CharSequence token) {
-        return isNowKeyword(token)
-                || isSysdateKeyword(token)
-                || isSystimestampKeyword(token)
-                || isCurrentTimestampKeyword(token);
+    private boolean containsDisallowedFunction(ExpressionNode node, CharSequence timestampColumn) {
```
nwoolmer, line 2100: This kind of crawling is often handled by PostOrderTreeTraversalAlgo.Visitor variants

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -2614,6 +2634,148 @@ private void detectTimestampOffset(QueryModel parent, QueryModel nested) throws
...
+        }
+
+        // Skip if no nested model
+        if (nested == null) {
+            return;
+        }
+
+        // Skip if nested model doesn't preserve row ordering
+        int nestedType = nested.getSelectModelType();
+        if (nestedType == SELECT_MODEL_GROUP_BY
```
nwoolmer: nit: switch?

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -2614,6 +2634,148 @@ private void detectTimestampOffset(QueryModel parent, QueryModel nested) throws
...
+                    if (inverseOffset < Integer.MIN_VALUE || inverseOffset > Integer.MAX_VALUE) {
+                        throw SqlException.position(strideArg.position)
+                                .put("timestamp offset value ")
+                                .put(inverseOffset)
+                                .put(" exceeds maximum allowed range for dateadd function (must be between ")
+                                .put(Integer.MIN_VALUE)
+                                .put(" and ")
+                                .put(Integer.MAX_VALUE)
+                                .put(")");
+                    }
```
nwoolmer, line 2804: Is this validation null aware? And hopefully this would be validated by the dateadd factory too?

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
```diff
@@ -9231,6 +9254,40 @@ private void traverseNamesAndIndices(QueryModel parent, ExpressionNode node) thr
...
+            try {
+                ExpressionNode clone = deepClone(expressionNodePool, node);
+                traversalAlgo.traverse(clone, literalCheckingVisitor.of(parent.getAliasToColumnMap()));
+                traversalAlgo.traverse(clone, literalRewritingVisitor.of(parent.getAliasToColumnNameMap()));
+                if (allLiteralsMatchBranch(clone, branch.getAliasToColumnMap())) {
+                    addWhereNode(branch, clone);
+                }
+            } catch (NonLiteralException ignore) {
+                // skip this branch - the outer filter still guarantees correctness
+            }
```
nwoolmer, line 9593: nit: maybe at least a debug log?

core/src/test/java/io/questdb/test/griffin/SqlParserTest.java
```diff
@@ -9792,6 +9792,201 @@ public void testPublicSchemaRemovalShowCreateTable() throws Exception {
...
+        assertQuery(
+                "select-choose ts from (" +
+                        /**/ "select [ts] from (" +
+                        /**/   "select-choose [ts1 ts] ts1 ts from (" +
+                        /**/     "select [ts1] from t1 timestamp (ts1) where ts1 in '2025-12-01T01;2h'" +
+                        /**/   ") except select-choose [ts2 ts] ts2 ts from (" +
+                        /**/     "select [ts2] from t2 timestamp (ts2) where ts2 in '2025-12-01T01;2h'" +
+                        /**/   ")" +
+                        /**/ ") _xQdbA1 where ts in '2025-12-01T01;2h'" +
+                        ")",
```
nwoolmer, line 9804: What is the `/**/` for?

And we can use block quotes now - """

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
mtopolnik: I tried using `PostOrderTreeTraversalAlgo`, but it didn't seem to be a natural fit.

What we need is pre-order search (check parent, descend into children) with shourt-circuiting as soon as an offending node is found. `PostOrderTreeTraversalAlgo` visits children before the parent and has no provision for short-circuiting.

We'd probably need something like `LiteralCheckingVisitor` with `NonLiteralException.INSTANCE`, exception-based flow control:

```java
private static class DisallowedFunctionException extends RuntimeException {
    static final DisallowedFunctionException INSTANCE = new DisallowedFunctionException();
}

private class DisallowedFunctionVisitor implements PostOrderTreeTraversalAlgo.Visitor {
    private CharSequence timestampColumn;

    @Override
    public void visit(ExpressionNode node) {
        if (node.type == FUNCTION) {
            if (isTimeFunctionKeyword(node.token)) {
                throw DisallowedFunctionException.INSTANCE;
            }
            if (Chars.equalsLowerCaseAscii(node.token, "dateadd")) {
                if (!isDateaddWithTimestamp(node, timestampColumn)) {
                    throw DisallowedFunctionException.INSTANCE;
                }
            }
        }
    }

    DisallowedFunctionVisitor of(CharSequence timestampColumn) {
        this.timestampColumn = timestampColumn;
        return this;
    }
}
```

With the call site becoming:

```java
try {
    traversalAlgo.traverse(node, disallowedFunctionVisitor.of(timestampColumn));
    return false;
} catch (DisallowedFunctionException e) {
    return true;
}
```

There's another option, but I'd say it's even worse -- abusing `descend()` for a kind of pre-order traversal. We'd need a mutable `found` flag + `descend()` returning `!found` + empty `visit()`:

```java
private class DisallowedFunctionVisitor implements PostOrderTreeTraversalAlgo.Visitor, Mutable {
    private boolean found;
    private CharSequence timestampColumn;

    @Override
    public void clear() {
        found = false;
        timestampColumn = null;
    }

    @Override
    public boolean descend(ExpressionNode node) {
        if (found) {
            return false; // already found, skip everything
        }
        if (node.type == FUNCTION) {
            if (isTimeFunctionKeyword(node.token)) {
                found = true;
                return false;
            }
            if (Chars.equalsLowerCaseAscii(node.token, "dateadd")) {
                if (!isDateaddWithTimestamp(node, timestampColumn)) {
                    found = true;
                    return false;
                }
            }
        }
        return true;
    }

    boolean isFound() {
        return found;
    }

    @Override
    public void visit(ExpressionNode node) {
        // no-op — all work done in descend()
    }

    DisallowedFunctionVisitor of(CharSequence timestampColumn) {
        this.found = false;
        this.timestampColumn = timestampColumn;
        return this;
    }
}
```

Call site:

```java
traversalAlgo.traverse(node, disallowedFunctionVisitor.of(timestampColumn));
return disallowedFunctionVisitor.isFound();
```

`PostOrderTreeTraversalAlgo` also has a mechanism to backup/restore internal stacks on every call. This is additional overhead that we don't need.

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
mtopolnik: If stride is a NULL literal, this code is never reached because the `isDateaddTimestampExpression(ast, timestampSourceModel)` returns `false` from `isConstantIntegerExpression(strideArg)`. I added a test to `TimestampOffsetPushdownTest` for this.

core/src/test/java/io/questdb/test/griffin/SqlParserTest.java
mtopolnik: This is a single-line string that we expect. I used `/**/` to preserve the custom indentation that makes it easier to see the nesting of models.

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
mtopolnik: Addressed in 6a713ef60d

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
mtopolnik: Addressed in 2007c423d8

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
mtopolnik: Addressed in ec4f62269e

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
mtopolnik: Addressed in eb052f18c9

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
mtopolnik: Addressed in 05bfaa8576

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
nwoolmer: Sounds good. Maybe we can create an issue to do a review of optimiser. We perform a lot of passes that generate garbage, and we might be able to slim down how much 'one off' code is in each step.  I think providing better alternatives to traversalAlgo could help there.

It's particularly relevant for Grafana users, since you can't use prepared statements.

Thanks for taking the time to consider it!

core/src/main/java/io/questdb/griffin/SqlOptimiser.java
nwoolmer: Thank you, I only looked at the surface before asking!

core/src/test/java/io/questdb/test/griffin/SqlParserTest.java
nwoolmer: Poor debugging visibility from model `toString0` has been a gripe for a while. 

If we can do better, we should. Instead of `/**/` ,maybe we should improve the output to be more consistent, like with PlanSink, and then you can have the formatting generated and preserved automatically?

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 178 / 202 (88.12%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|170|194|87.63%|
|:large_blue_circle:|io/questdb/griffin/SqlKeywords.java|8|8|100.00%|

PR #6751, opened 2026-02-05
fix(sql): reject trailing content after valid query

### LINKED ISSUE #6726: Query parser ignoring anything after limit
### To reproduce

Try this out in demo

```
select * from trades where timestamp in today()
 limit -1

select * from trades;
```

It should complain, but just executes the first query instead

### QuestDB version:

9.3.2

demo box

ext4

### Full Name:

javier

### Affiliation:

questdb

- [x] Yes, I have

### Additional context

_No response_

mtopolnik:
Fixes #6726

## Summary
- Add post-compilation validation in `compile()` to reject unconsumed tokens after a valid statement, instead of silently ignoring them
- Fix PG compatibility no-op handlers (`RESET`, `CLOSE`, `UNLISTEN`, `DISCARD`) to require their expected argument instead of silently accepting bare keywords
- Fix `BEGIN`, `COMMIT`, `ROLLBACK` to consume the optional `TRANSACTION` keyword
- Validate `SET` statement syntax (`SET [SESSION | LOCAL] name { = | TO } value [, value]*`), rejecting malformed forms like missing name, invalid operator, missing value, and dangling commas

## Test plan
- `testTrailingContentAfterDdlRejected` — DDL followed by trailing token is rejected
- `testTrailingContentAfterNoOpRejected` — `RESET ALL extra` is rejected
- `testTrailingContentAfterSelectRejected` — `select...select` is rejected
- `testTrailingContentAfterSetRejected` — `SET x = y extra_token` is rejected
- `testTrailingSemicolonAllowed` — trailing `;` is accepted
- `testTrailingTokenAbsentIsOk` — no trailing content is accepted
- `testCloseMissingArgRejected` — bare `CLOSE` without argument is rejected
- `testDiscardMissingArgRejected` — bare `DISCARD` without argument is rejected
- `testResetMissingArgRejected` — bare `RESET` without argument is rejected
- `testUnlistenMissingArgRejected` — bare `UNLISTEN` without argument is rejected
- `testCompileSetMissingName` — `SET` with no name is rejected
- `testCompileSetMissingOperator` — `SET x` with no `=`/`TO` is rejected
- `testCompileSetInvalidOperator` — `SET x GARBAGE y` is rejected
- `testCompileSetMissingValue` — `SET x =` with no value is rejected
- `testCompileSetDanglingComma` — `SET x = y,` with trailing comma is rejected
- `testCompileSetWithTo` — `SET x TO y` is accepted
- `testCompileSetMultipleValues` — `SET x TO y, z` is accepted
- `testCompileSetWithSession` — `SET SESSION x = y` is accepted
- `testCompileSetWithLocal` — `SET LOCAL x = y` is accepted
- `testCompileSetQuotedValue` — `SET x = 'quoted'` is accepted
- Existing `testCompileSet`, `testCompileResetDoesNothing`, `testCompileCloseDoesNothing`, `testCompileUnlistenDoesNothing` still pass

### DIFF
--- a/core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java
+++ b/core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java
@@ -3333,6 +3333,7 @@ public class SqlCompilerImpl implements SqlCompiler, Closeable, SqlParserCallbac
             } while ((tok = SqlUtil.fetchNext(lexer)) != null && !isSemicolon(tok) && Chars.equals(tok, ','));
         } else {
             // Non-standard form (e.g., SET TIME ZONE 'UTC'): consume remaining tokens
+            //noinspection StatementWithEmptyBody
             while ((tok = SqlUtil.fetchNext(lexer)) != null && !isSemicolon(tok)) {
             }
         }


qodana.yaml
```diff
@@ -39,6 +39,7 @@ exclude:
     paths:
       - benchmarks
       - core/src/main/java/io/questdb/cairo/TickCalendarService.java
+      - core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java
```
nwoolmer, line 42: Qodana commit intended?

core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java
```diff
@@ -3290,11 +3316,33 @@ private void compileReindex(SqlExecutionContext executionContext, @Transient Cha
...
+                if (tok != null && !isSemicolon(tok)) {
+                    // consume optional [, value]*
+                    while ((tok = SqlUtil.fetchNext(lexer)) != null && Chars.equals(tok, ',')) {
+                        SqlUtil.fetchNext(lexer); // value after comma
+                    }
+                    // put back whatever stopped the loop (next statement token, ';', etc.)
+                    lexer.unparseLast();
+                }
+            }
+        }
```
nwoolmer: This is arrow code. Still, we should be validating each token i.e. that it is an `=` or a literal etc. not just throwing them away. Though we don't currently use the values, we should still error on unsupported syntax

core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java
```diff
@@ -3135,6 +3155,12 @@ private void compileMatViewQuery(
         }
     }
 
+    // PG compatibility no-op: RESET ALL, CLOSE ALL, UNLISTEN *, DISCARD ALL — consume exactly one argument.
+    private void compileNoOp(SqlExecutionContext executionContext, @Transient CharSequence sqlText) throws SqlException {
+        SqlUtil.fetchNext(lexer);
```
nwoolmer: Needs validation of what was fetched

qodana.yaml
mtopolnik: Yes, this is just for the "unused" lint. This class uses method references, Qodana doesn't understand that and insists you remove parameters.

core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java
mtopolnik: It would take us too far, the variants listed in the comment aren't at all exhaustive. There's `DISCARD PLANS/SEQUENCES/TEMPORARY`, there's `UNLISTEN channel_name`, etc.

core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java
mtopolnik: Addressed in 16340cffe9a51f3912990d8bd35d4ea2bf99106f

bluestreak01 (commented):
@mtopolnik Here's a detailed review of the PR:

## 1. Critical: `SET TIME ZONE` will break

The `compileSet` method assumes the grammar `SET [SESSION | LOCAL] name { = | TO } value [, ...]`. But PostgreSQL's `SET TIME ZONE 'UTC'` uses a **two-word compound parameter name** (`TIME ZONE`), which doesn't fit this grammar. Tracing through:

```
SET TIME ZONE 'UTC'
         ↑
         compileSet reads "TIME" as parameter name (not SESSION/LOCAL)
         then reads "ZONE" expecting '=' or 'TO' → throws "'=' or 'TO' expected"
```

Some PG JDBC drivers send `SET TIME ZONE` during connection initialization. Before this PR, `compileSet` was a pure no-op that consumed nothing, so any SET statement was silently accepted. Now it will reject this form. A test for `SET TIME ZONE 'UTC'` would demonstrate the failure.

---

## 2. Use the existing `SqlException.unexpectedToken()` helper

At `SqlCompilerImpl.java:399`, the trailing content error is manually constructed:

```java
throw SqlException.$(lexer.lastTokenPosition(), "unexpected token [").put(tok).put(']');
```

But `SqlException.java:152` already has a helper for this exact pattern:

```java
public static SqlException unexpectedToken(int position, CharSequence token) {
    return position(position).put("unexpected token [").put(token).put(']');
}
```

Should use `SqlException.unexpectedToken(lexer.lastTokenPosition(), tok)` instead.

---

## 3. `compileSet` semicolon handling: inconsistency with `compileBegin`/`compileCommit`/`compileRollback`

In `compileSet`, when the `do-while` loop exits on a semicolon, `tok != null` is true, so `unparseLast()` is called — pushing the `;` back. In `compile()` (single-query mode), the trailing content check then re-fetches it and accepts it. This is correct.

However, the `BEGIN`/`COMMIT`/`ROLLBACK` methods have the opposite behavior — they *consume* the semicolon (no `unparseLast` when `isSemicolon(tok)` is true), relying on `goToQueryEnd()`'s initial `unparseLast()` to recover in batch mode. The inconsistency between how `compileSet` and `compileBegin` handle `;` is confusing and fragile.

---

## 4. Stale comment in `compileBatch`

`SqlCompilerImpl.java:465`:
```java
// We've to move lexer because some query handlers don't consume all tokens (e.g. SET )
```
After this PR, `SET` now fully consumes its tokens. This comment is stale.

---

## 5. Test coverage gaps

**Missing happy-path tests:**
- `BEGIN TRANSACTION`, `COMMIT TRANSACTION`, `ROLLBACK TRANSACTION` — the new TRANSACTION-keyword consumption is untested
- `SET TIME ZONE 'UTC'` — common PG statement (will reveal bug #1 above)

**Missing error-path tests:**
- `BEGIN extra_garbage` — should be rejected by trailing content check, but no test verifies this
- `COMMIT extra_garbage` / `ROLLBACK extra_garbage` — same gap
- `CLOSE ALL extra`, `DISCARD ALL extra`, `UNLISTEN * extra` — only `RESET ALL extra` is tested for no-op trailing content

---

## 6. Latent test fixes are correct

The PR fixes genuinely broken test SQL that was previously silently accepted:
- **`AlterTableDropPartitionTest.java:917`**: Stray `)` removed — `to_timestamp('2020', 'yyyy'))` → `to_timestamp('2020', 'yyyy')`
- **`MatViewTest.java:1871`**: Stray `;)` removed — `full;)` → `full`
- **`LimitTest.java:1224`**: Semicolon misused as comma — `limit :lo; :hi` → `limit :lo, :hi`

These are valuable finds.

---

## 7. Minor: Qodana exclusion scope

The addition to `qodana.yaml` excludes `SqlCompilerImpl.java` from the `unused` inspection at the file level. It would be better to suppress at the specific method/field level with `@SuppressWarnings("unused")` to avoid masking future dead code in this large file.

---

## 8. Zero-GC compliance

The new code is zero-GC safe — all token operations use existing `CharSequence` references from the lexer, no allocations on the data path. Looks good.

---

## Summary

| Severity | Issue |
|----------|-------|
| **Critical** | `SET TIME ZONE 'UTC'` will break — compound parameter names not supported |
| **Medium** | Should use `SqlException.unexpectedToken()` helper instead of manual construction |
| **Medium** | Missing tests for BEGIN/COMMIT/ROLLBACK TRANSACTION and trailing garbage |
| **Low** | Stale comment in `compileBatch` about SET not consuming tokens |
| **Low** | Inconsistent semicolon handling between `compileSet` and `compileBegin` |
| **Low** | Qodana file-level exclusion could mask future issues |
| **Positive** | Latent syntax errors in tests correctly fixed |
| **Positive** | Zero-GC compliance maintained |
| **Positive** | Error positions verified correct for all tested cases |

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 33 / 33 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|33|33|100.00%|

PR #6753, opened 2026-02-06
fix(sql): fix ADD COLUMN IF NOT EXISTS for DECIMAL, GEOHASH, and array types

### LINKED ISSUE #6750: column already exists with a different column type when adding decimal(...) column
### To reproduce

Create a simple trades table
execute 
```
ALTER TABLE trades ADD COLUMN IF NOT EXISTS market_cap_usd DECIMAL(48, 18);
```
It applies schema
Execute same query one more time (expected to be no-op as it exists)
But you will get 
```
column already exists with a different column type [current type=DECIMAL(48, 18), requested type=DECIMAL
``` instead.

Leads to 
https://github.com/questdb/questdb/blob/769456e9109ad3e9fd873bb0fea710ba7a2bbcd5/core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java#L963
and
https://github.com/questdb/questdb/blob/master/core/src/main/java/io/questdb/cairo/ColumnType.java#L984

looks like 1st link should cut (...) in comparsion

### QuestDB version:

9.2.0

Docker, ubuntu 20.04

ext4

### Full Name:

Nikolay

### Affiliation:

Startup

- [x] Yes, I have

### Additional context

_No response_

mtopolnik:
Fixes #6750

## Summary

The `ALTER TABLE t ADD COLUMN IF NOT EXISTS col <type>` path did not
fully resolve parameterized or compound column types before comparing
with existing metadata, causing false type-mismatch errors.

### Bug fixes

- DECIMAL/GEOHASH: the IF NOT EXISTS path used `ColumnType.typeOf()` which
  returns the **base** type constant, but metadata stores the **fully encoded**
  type (with precision/scale/bits). Fixed by parsing DECIMAL and GEOHASH
  parameters before comparing.
- Array types (e.g. `DOUBLE[]`): the IF NOT EXISTS path did not parse array
  dimensionality brackets, causing type mismatch errors and unconsumed tokens.
  Fixed by calling `SqlUtil.parseArrayDimensionality()` before comparing.
- Unsupported array element types (e.g. `INT[]`) in the IF NOT EXISTS path
  now correctly report "unsupported array element type" instead of falling
  through to a misleading type-mismatch error.
- Unmatched `]` bracket after a type name (e.g. `double]`) is now detected
  in the IF NOT EXISTS path with a helpful error pointing at the bracket.

### Consistency improvements

- Moved `parseGeoHashColumnType` from `SqlCompilerImpl` (private) to
  `SqlParser` (public) to co-locate it with `parseDecimalColumnType`
- Replaced `ColumnType.typeOf()` with `SqlUtil.toPersistedType()` in the
  IF NOT EXISTS path for consistency with `addColumnWithType`, giving
  more specific error messages for non-persisted types like CURSOR

## Test plan

- Verify `ADD COLUMN IF NOT EXISTS col DECIMAL(48, 18)` is a no-op when column already exists with same type
- Verify `ADD COLUMN IF NOT EXISTS col DECIMAL` (bare, defaults to 18,3) is a no-op when column exists as DECIMAL(18,3)
- Verify `ADD COLUMN IF NOT EXISTS col DECIMAL(18, 3)` correctly reports type mismatch when column exists as DECIMAL(48, 18)
- Verify `ADD COLUMN IF NOT EXISTS col GEOHASH(5c)` is a no-op when column already exists with same type
- Verify `ADD COLUMN IF NOT EXISTS col GEOHASH(3c)` correctly reports type mismatch when column exists as GEOHASH(5c)
- Verify `ADD COLUMN IF NOT EXISTS col DOUBLE[]` is a no-op when column already exists with same type
- Verify `ADD COLUMN IF NOT EXISTS col DOUBLE[][]` correctly reports type mismatch when column exists as DOUBLE[]
- Verify `ADD COLUMN IF NOT EXISTS col INT[]` reports "unsupported array element type" in the IF NOT EXISTS path
- Verify `ADD COLUMN IF NOT EXISTS col double]` detects unmatched bracket with correct error position
- Verify existing ADD COLUMN and IF NOT EXISTS tests still pass

### DIFF
--- a/core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java
+++ b/core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java
@@ -753,15 +753,15 @@ public class SqlCompilerImpl implements SqlCompiler, Closeable, SqlParserCallbac
     }
 
     private int addColumnWithType(
-            AlterOperationBuilder addColumn,
+            @Nullable AlterOperationBuilder addColumn,
             CharSequence columnName,
             int columnNamePosition
     ) throws SqlException {
         CharSequence tok;
         tok = expectToken(lexer, "column type");
-
-        int columnType = SqlUtil.toPersistedType(tok, lexer.lastTokenPosition());
         int typePosition = lexer.lastTokenPosition();
+
+        int columnType = SqlUtil.toPersistedType(tok, typePosition);
 
         int dim = SqlUtil.parseArrayDimensionality(lexer, columnType, typePosition);
         if (dim > 0) {
@@ -875,16 +875,18 @@ public class SqlCompilerImpl implements SqlCompiler, Closeable, SqlParserCallbac
             indexed = false;
         }
 
-        addColumn.addColumnToList(
-                columnName,
-                columnNamePosition,
-                columnType,
-                Numbers.ceilPow2(symbolCapacity),
-                cache,
-                indexed,
-                Numbers.ceilPow2(indexValueBlockCapacity),
-                false
-        );
+        if (addColumn != null) {
+            addColumn.addColumnToList(
+                    columnName,
+                    columnNamePosition,
+                    columnType,
+                    Numbers.ceilPow2(symbolCapacity),
+                    cache,
+                    indexed,
+                    Numbers.ceilPow2(indexValueBlockCapacity),
+                    false
+            );
+        }
         lexer.unparseLast();
         return columnType;
     }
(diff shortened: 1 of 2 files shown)


mtopolnik (commented):
## PR Review by Claude

**Overall: Looks good.** The fix is well-scoped and the approach is sound — the IF NOT EXISTS path now mirrors the normal `addColumnWithType` path for type resolution, ensuring the comparison uses fully-encoded types rather than bare base types.

### Positives

1. **Good deduplication of GEOHASH parsing** — moving `parseGeoHashColumnType` to `SqlParser` (next to `parseDecimalColumnType`) and calling it from both paths eliminates the inline copy in `addColumnWithType`. The new method is a clean 1:1 extraction of the old inline code.

2. **Consistent use of `SqlUtil.toPersistedType`** instead of raw `ColumnType.typeOf` in the IF NOT EXISTS path, giving better error messages for non-persisted types like CURSOR.

3. **Good test coverage** — tests for DECIMAL (with and without explicit precision), GEOHASH, arrays, unsupported array element types, and the unmatched bracket case.

4. **Error positions are correct** — `typePosition` is captured right after reading the type token and used for all error messages in the IF NOT EXISTS block.

### Issues

**1. IF NOT EXISTS path doesn't consume remaining tokens on success (medium)**

When the column exists and types match, the code hits `break` at line 968. But if the statement had additional clauses after the type (e.g., `ALTER TABLE x ADD COLUMN IF NOT EXISTS col SYMBOL CAPACITY 1024 CACHE, other_col INT`), those tokens are left unconsumed. The normal `addColumnWithType` path parses SYMBOL capacity/cache/index options. The IF NOT EXISTS path ignores them, which means:
- `ALTER TABLE x ADD COLUMN IF NOT EXISTS col SYMBOL CAPACITY 1024 CACHE` — the `CAPACITY 1024 CACHE` tokens are silently dropped when the column already exists
- If there are more columns after a comma, they'd also be silently dropped

This was pre-existing behavior (not introduced by this PR), but it's worth noting as a follow-up item since you're already touching this code.

**2. Minor: bare `DECIMAL` in the IF NOT EXISTS path (low)**

The test `testAddDuplicateColumnIfNotExistsDecimalDefault` adds a column as bare `DECIMAL` then checks it with bare `DECIMAL` again. This works because `SqlParser.parseDecimalColumnType` defaults to `(18,3)` in both cases. Good — but there's no test for the cross-case: adding as bare `DECIMAL` and checking with explicit `DECIMAL(18, 3)` (or vice versa). Adding that would further validate the equivalence. Not blocking.

**3. No test for GEOHASH missing precision in IF NOT EXISTS path (low)**

There's no test for `ALTER TABLE x ADD COLUMN IF NOT EXISTS geo_col GEOHASH` (missing precision). The parsing would throw "missing GEOHASH precision" from the new `parseGeoHashColumnType`, but there's no explicit test asserting this error in the IF NOT EXISTS context. Not blocking.

### Summary

The core logic is correct. The two code paths (normal add and IF NOT EXISTS) now produce the same fully-encoded type for DECIMAL, GEOHASH, and array types before comparison. **Approve**, with optional suggestions to add the cross-case DECIMAL test and to note the pre-existing SYMBOL options consumption gap as a follow-up.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 36 / 37 (97.30%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|19|20|95.00%|
|:large_blue_circle:|io/questdb/griffin/SqlParser.java|17|17|100.00%|

bluestreak01 (approved):

PR #6756, opened 2026-02-08
fix(core): clear changedKeys at start of config reload to prevent stale notifications

bluestreak01:
## Summary

- `DynamicPropServerConfiguration.reload()` accumulates stale entries in `changedKeys` across reload cycles when only a secret file changes (properties file unchanged), causing watchers to be spuriously notified about properties that did not change.
- Root cause: `changedKeys.clear()` only lived inside `updateSupportedProperties()`, which is skipped when the properties file hasn't changed. Secret-file-only reloads added keys on top of the stale set.
- Fix: move `changedKeys.clear()` to the top of `reload()`, inside the `synchronized` block, so every reload cycle starts fresh.

## Test plan

- [x] New test `testSecretFileRotationDoesNotNotifyStaleWatchers`: registers a listener on a dynamic property, reloads after changing that property (listener fires), then rotates only a secret file and reloads again — asserts the listener does **not** fire a second time.
- [x] Full `DynamicPropServerConfigurationTest` suite passes (40 tests).

### DIFF
--- a/core/src/main/java/io/questdb/DynamicPropServerConfiguration.java
+++ b/core/src/main/java/io/questdb/DynamicPropServerConfiguration.java
@@ -430,6 +430,7 @@ public class DynamicPropServerConfiguration implements ServerConfiguration, Conf
             synchronized (reloadLock) {
                 // Check that the file has been modified since the last trigger
                 // Then load the config properties
+                changedKeys.clear();
 
                 Properties newProperties = new Properties();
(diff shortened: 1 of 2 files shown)


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/DynamicPropServerConfiguration.java|1|1|100.00%|

bluestreak01:
@Coderabbit review critically

PR #6757, opened 2026-02-08
feat(sql): sort SHOW TABLES results alphabetically

### LINKED ISSUE #353: [SQL] Show tables should sort the result alphabetically be default
In the current version, `SHOW TABLES` does not return the result in a sorted manner. Adding `ORDER BY tableName` does not help.

Most DBs sort the output of `SHOW TABLES` alphabetically by default, I think it would make sense if we did the same.

maciulis:
Fixes #353

### Why

As per issue #353:

In the current version, `SHOW TABLES` does not return the result in a sorted manner. Adding `ORDER BY tableName does not help. Most DBs sort the output of `SHOW TABLES` alphabetically by default - it would make sense if QuestDB did the same.

The change should update how the SQL is executed in the backend.

### Solution

To have sorted data structure `CharSequenceObjSortedHashMap` that will maintain sorted order of the keys (table names) behind configuration parameter.

#### New server property                                                                                                                                                                                                                   
                                                                                                                                                                                                                                            
  cairo.metadata.cache.snapshot.ordered=true
     
  Default: `false`. When enabled, all table-listing functions return rows sorted
  alphabetically by table name. The sort is maintained incrementally inside                                                                                                                                                                 
  `CharSequenceObjSortedHashMap`; there is no post-query sort step.
                                                                                                                                                                                                                                            
  #### New data structures
                                                                                                                                                                                                                                            
  - **`CharSequenceObjSortedHashMap<V>`** — a hash map backed by a `CharSequenceSortedList`                                                                                                                                                 
    that keeps keys in lexicographic order. `put` / `remove` maintain the sorted list via
    binary-search insertion, so lookup is O(1) (hash) and ordered iteration is O(n).                                                                                                                                                        
                                                                                                                                                                                                                                            
  - **`CharSequenceSortedList`** — a char-sequence list that inserts at the correct sorted                                                                                                                                                  
    position using binary search. Used by `CharSequenceObjSortedHashMap` to provide the                                                                                                                                                     
    ordered key set.                                                                                                                                                                                                                        
                                                                                                                                                                                                                                            
  - **`CharSequenceObjMap<V>`** — a new common interface for `CharSequenceObjHashMap` and                                                                                                                                                   
    `CharSequenceObjSortedHashMap`, letting callers hold either implementation without                                                                                                                                                      
    knowing which one is in use.                                                                                                                                                                                                            
                                                                                                                                                                                                                                            
  #### Implementation details                                                                                                                                                                                                                
                                                                                                                                                                                                                                            
  - `LocalCacheSnapshotFactory` / `DefaultLocalCacheSnapshotFactory`: a factory singleton                                                                                                                                                   
    that reads `getCairoMetadataCacheSnapshotOrdered()` from `CairoConfiguration` and
    returns the appropriate map type. Cursor factories (`AllTablesCursorFactory`,                                                                                                                                                           
    `TablesFunctionFactory`, etc.) accept a `CairoConfiguration` at construction and call
    the factory to obtain their local cache.                                                                                                                                                                                                
                  
  - `MetadataCacheReader.snapshot()` signature changed from                                                                                                                                                                                 
    `snapshot(CharSequenceObjSortedHashMap<CairoTable>, long)` to
    `snapshot(CharSequenceObjMap<CairoTable>, long)`, accepting either map implementation.                                                                                                                                                  
                                                                                                                                                                                                                                            
  #### Performance
                                                                                                                                                                                                                                            
  Two JMH benchmarks (`MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark` and                                                                                                                                                              
  `MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark`) measure snapshot-refresh
  throughput across table counts of 100, 1 000, and 10 000 for sorted path. 

Benchmarks confirm that the changes bring:

1. ADD - 44% slower than without sorting while with sortKeys call its 29% slower
2. REMOVE - 20x faster on 10k entries and twice as fast for 1k entries with 100 entries not to be considered due to high error rate.                                                      

Raw numbers bellow comparing unsorted(original) CharSequenceObjHashMap implementation, sorted data structure implementation and using CharSequenceObjHashMap(original datastructure) with sort keys method at the end of snapshot update:

```
Master branch performance benchmark for ADD only:

Benchmark                                                            (size)   Mode  Cnt      Score      Error  Units
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh     100  thrpt    5  40878.847 ± 4735.545  ops/s
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh    1000  thrpt    5    465.561 ±    6.530  ops/s
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh   10000  thrpt    5     52.061 ±    0.630  ops/s
------------------------------------------------------------------------------------------------------------------------
Master branch performance with sort keys benchmark for ADD only:

Benchmark                                                            (size)   Mode  Cnt      Score       Error  Units
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh     100  thrpt    5  29357.463 ± 13586.159  ops/s
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh    1000  thrpt    5    432.681 ±     7.467  ops/s
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh   10000  thrpt    5     47.805 ±     0.774  ops/s
------------------------------------------------------------------------------------------------------------------------

PR with Sorted datastructure on show_tables_list_results_alphabetically branch for ADD only:

Benchmark                                                            (size)   Mode  Cnt      Score      Error  Units
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh     100  thrpt    5  23003.050 ± 2184.974  ops/s
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh    1000  thrpt    5    377.086 ±   12.194  ops/s
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh   10000  thrpt    5     29.425 ±    0.353  ops/s
========================================================================================================================
Master branch performance benchmark for REMOVE only:

Benchmark                                                               (size)   Mode  Cnt      Score       Error  Units
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh     100  thrpt    5  96141.839 ± 20905.071  ops/s
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh    1000  thrpt    5   1166.982 ±   250.540  ops/s
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh   10000  thrpt    5     10.541 ±     0.950  ops/s
------------------------------------------------------------------------------------------------------------------------
Master branch performance with sort keys benchmark for REMOVE only:

Benchmark                                                               (size)   Mode  Cnt       Score       Error  Units
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh     100  thrpt    5  100723.978 ± 20116.659  ops/s
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh    1000  thrpt    5     930.639 ±   158.317  ops/s
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh   10000  thrpt    5       9.678 ±     2.001  ops/s
-------------------------------------------------------------------------------------------------------------------------

PR with Sorted datastructure on show_tables_list_results_alphabetically branch for REMOVE only:

Benchmark                                                               (size)   Mode  Cnt      Score       Error  Units
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh     100  thrpt    5  72096.096 ± 12733.897  ops/s
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh    1000  thrpt    5   2285.516 ±   491.089  ops/s
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh   10000  thrpt    5    209.495 ±     4.655  ops/s
========================================================================================================================
```                                                                                                  
                                                                                                                                                                                                                                            
  #### Tests                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                            
  - `ShowTablesTest` — four new test cases covering ordered results for `SHOW TABLES`,                                                                                                                                                      
    `all_tables()`, drop-and-recreate, and rename scenarios.
  - `ShowTablesWithOrderingDisabledTest` — mirrors the same four scenarios with                                                                                                                                                             
    `ordered=false`, verifying insertion-order behavior when the property is off.
  - `MetadataCacheTest` — extended to cover snapshot correctness with the new map type.                                                                                                                                                     
  - `CharSequenceObjSortedHashMapTest` — unit tests for the new sorted map.                                                                                                                                                                 
  - `CharSequenceSortedListTest` — unit tests for the new sorted list.                                                                                                                                                                      
                                                                                                                                                                                                                                            
  #### Test plan                                                                                                                                                                                                                             
                  
  - Run `ShowTablesTest` and `ShowTablesWithOrderingDisabledTest` with                                                                                                                                                                      
    `CAIRO_METADATA_CACHE_SNAPSHOT_ORDERED` set to `true` and `false` respectively.
  - Run `MetadataCacheTest` to verify snapshot integrity.                                                                                                                                                                                   
  - Run `CharSequenceObjSortedHashMapTest` and `CharSequenceSortedListTest` for the                                                                                                                                                         
    new data structures.                                                                                                                                                                                                                    
  - Run `CharSequenceObjHashMapTest` to confirm no regression in the existing unsorted map.                                                                                                                                                 
  - Run `CreateMatViewTest` — the test now sets the ordering property explicitly; verify                                                                                                                                                    
    it still passes.


maciulis:
@CodeRabbit review this PR critically

maciulis:
@CodeRabbit review this PR critically

maciulis:
@CodeRabbit review this PR critically once again - I have pushed right changes this time

maciulis:
Following the benchmark results - the changes bring:

1. ADD - 44% slower than without sorting while with `sortKeys()` (original change) call its 29% slower.
2. REMOVE - 20x faster on 10k entries and twice as fast for 1k entries with 100 entries not to be considered due to high error rate.

```

Master branch performance benchmark for ADD only:

Benchmark                                                            (size)   Mode  Cnt      Score      Error  Units
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh     100  thrpt    5  40878.847 ± 4735.545  ops/s
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh    1000  thrpt    5    465.561 ±    6.530  ops/s
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh   10000  thrpt    5     52.061 ±    0.630  ops/s
------------------------------------------------------------------------------------------------------------------------
Master branch performance with sort keys benchmark for ADD only:

Benchmark                                                            (size)   Mode  Cnt      Score       Error  Units
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh     100  thrpt    5  29357.463 ± 13586.159  ops/s
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh    1000  thrpt    5    432.681 ±     7.467  ops/s
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh   10000  thrpt    5     47.805 ±     0.774  ops/s
------------------------------------------------------------------------------------------------------------------------

PR with Sorted datastructure on show_tables_list_results_alphabetically branch for ADD only:

Benchmark                                                            (size)   Mode  Cnt      Score      Error  Units
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh     100  thrpt    5  23003.050 ± 2184.974  ops/s
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh    1000  thrpt    5    377.086 ±   12.194  ops/s
MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.testCacheRefresh   10000  thrpt    5     29.425 ±    0.353  ops/s
========================================================================================================================
Master branch performance benchmark for REMOVE only:

Benchmark                                                               (size)   Mode  Cnt      Score       Error  Units
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh     100  thrpt    5  96141.839 ± 20905.071  ops/s
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh    1000  thrpt    5   1166.982 ±   250.540  ops/s
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh   10000  thrpt    5     10.541 ±     0.950  ops/s
------------------------------------------------------------------------------------------------------------------------
Master branch performance with sort keys benchmark for REMOVE only:

Benchmark                                                               (size)   Mode  Cnt       Score       Error  Units
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh     100  thrpt    5  100723.978 ± 20116.659  ops/s
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh    1000  thrpt    5     930.639 ±   158.317  ops/s
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh   10000  thrpt    5       9.678 ±     2.001  ops/s
-------------------------------------------------------------------------------------------------------------------------

PR with Sorted datastructure on show_tables_list_results_alphabetically branch for REMOVE only:

Benchmark                                                               (size)   Mode  Cnt      Score       Error  Units
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh     100  thrpt    5  72096.096 ± 12733.897  ops/s
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh    1000  thrpt    5   2285.516 ±   491.089  ops/s
MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.testCacheRefresh   10000  thrpt    5    209.495 ±     4.655  ops/s
========================================================================================================================

```

maciulis:
### Latest changes — factory refactoring and configurability                                                                                                                                                                              
                                                                                                                                                                                                                                            
  This update makes alphabetical ordering configurable instead of always-on,                                                                                                                                                                
  and generalizes the snapshot API so future map implementations can be plugged                                                                                                                                                             
  in without changing callers.                                                                                                                                                                                                              
                  
  **What changed:**                                                                                                                                                                                                                         
                  
  - Introduced `LocalCacheSnapshotFactory` interface and a `DefaultLocalCacheSnapshotFactory`                                                                                                                                               
    singleton that branches on the new `cairo.metadata.cache.snapshot.ordered` config property
    to produce either a `CharSequenceObjSortedHashMap` (sorted) or a `CharSequenceObjHashMap`                                                                                                                                               
    (insertion-order).                                                                                                                                                                                                                      
                                                                                                                                                                                                                                            
  - Added new `cairo.metadata.cache.snapshot.ordered` server property (default: `false`).                                                                                                                                                   
    Set it to `true` to enable alphabetical ordering of `SHOW TABLES` / `all_tables()` /                                                                                                                                                    
    `tables()` results.                                                                                                                                                                                                                     
                                                                                                                                                                                                                                            
  - `MetadataCacheReader.snapshot()` now accepts `CharSequenceObjMap` instead of the
    concrete `CharSequenceObjSortedHashMap`, decoupling the call site from the map                                                                                                                                                          
    implementation.                                                                                                                                                                                                                         
   
  - `AllTablesCursorFactory`, `TablesFunctionFactory`, `InformationSchemaColumnsFunctionFactory`,                                                                                                                                           
    and `PgAttributeFunctionFactory` now take a `CairoConfiguration` argument and use the
    factory to create their local caches — no logic changes, just the map type is now                                                                                                                                                       
    configuration-driven.                                                                                                                                                                                                                   
                                                                                                                                                                                                                                            
  - Added `ShowTablesWithOrderingDisabledTest` to cover the unordered path (`ordered=false`),                                                                                                                                               
    verifying that results follow metadata-cache insertion order when sorting is off.
                                                                                                                                                                                                                                            
  - Existing `ShowTablesTest` now explicitly sets `CAIRO_METADATA_CACHE_SNAPSHOT_ORDERED=true`                                                                                                                                              
    in `@Before` to isolate its ordered-results expectations from the property default.                                                                                                                                                     
                                                                                                                                                                                                                                            
  - Benchmark classes (`MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark`,                                                                                                                                                                
    `MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark`) had member declarations                                                                                                                                                        
    reordered to comply with the project's alphabetical-within-visibility style rule.

core/src/test/java/io/questdb/test/cairo/MetadataCacheTest.java
```diff
@@ -928,6 +928,51 @@ public void testTruncateTable() throws Exception {
         });
     }
 
+    @Test
+    public void testSnapshotSortedWithInitialListOfTables() throws Exception {
+        for (int cu = 'Z'; cu > 'A' - 1; cu--) {
```
puzpuzpuz: Could you use `assertMemoryLeak()` in the new tests? This utility detects off-heap memory leaks.

core/src/main/java/io/questdb/cairo/CairoConfiguration.java
```diff
@@ -723,6 +723,8 @@ default RostiAllocFacade getRostiAllocFacade() {
 
     long getSystemWalEventAppendPageSize();
 
+    boolean getCairoMetadataCacheSnapshotOrdered();
```
puzpuzpuz: Could you rename this method to `isCairoMetadataCacheSnapshotOrdered()`?

core/src/main/java/io/questdb/std/CharSequenceObjMap.java
```diff
@@ -0,0 +1,39 @@
...
+package io.questdb.std;
+
+import org.jetbrains.annotations.NotNull;
+
+public interface CharSequenceObjMap<V> {
+    int size();
+
+    V getAt(int i);
+
+    V get(@NotNull CharSequence tableName);
```
puzpuzpuz: `tableName` and `latestTable` as the argument names are too specific for the `CharSequenceObjMap` interface. `key` and `value` would be better.

core/src/test/java/io/questdb/test/griffin/ShowTablesWithOrderingDisabledTest.java
```diff
@@ -0,0 +1,88 @@
...
+ ******************************************************************************/
+
+package io.questdb.test.griffin;
+
+import io.questdb.PropertyKey;
+import io.questdb.test.AbstractCairoTest;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ShowTablesWithOrderingDisabledTest extends AbstractCairoTest {
```
puzpuzpuz, line 32: nit: these tests hardcode expected hash iteration order (e.g., "card_payments\naccounts\nbalances\ndeposits"). This order depends on `CharSequenceObjHashMap`'s internal hash function and capacity. Any change to the hash implementation will break these tests. Consider asserting only that all expected table names are present, without asserting order.

Additionally, the method names say `testShowTablesReturnsOrderedList` when ordering is actually disabled — they should say `...UnorderedList` or `...InsertionOrder`.

core/src/test/java/io/questdb/test/griffin/ShowTablesTest.java
```diff
@@ -24,15 +24,22 @@
...
+import io.questdb.PropertyKey;
 import io.questdb.cairo.sql.RecordCursor;
 import io.questdb.cairo.sql.RecordCursorFactory;
 import io.questdb.griffin.CompiledQuery;
 import io.questdb.griffin.SqlCompiler;
 import io.questdb.test.AbstractCairoTest;
+import org.junit.Before;
 import org.junit.Test;
 
 public class ShowTablesTest extends AbstractCairoTest {
```
puzpuzpuz, line 36: `tables()` function not effectively tested for ordering

Existing `testTables()` in `ShowTablesTest` queries with `ORDER BY table_name`, which always sorts the output regardless of the config flag. The test doesn't validate that the `tables()` function itself returns sorted results when `ordered=true`. The same gap applies to `information_schema.columns` and `pg_attribute`.

core/src/main/java/io/questdb/cairo/DefaultLocalCacheSnapshotFactory.java
```diff
@@ -0,0 +1,44 @@
...
+import io.questdb.std.CharSequenceObjMap;
+import io.questdb.std.CharSequenceObjSortedHashMap;
+
+public class DefaultLocalCacheSnapshotFactory implements LocalCacheSnapshotFactory {
+    public static final DefaultLocalCacheSnapshotFactory INSTANCE = new DefaultLocalCacheSnapshotFactory();
+
+    private DefaultLocalCacheSnapshotFactory() {
+    }
+
+    public CharSequenceObjMap<CairoTable> instanceGiven(CairoConfiguration configuration) {
```
puzpuzpuz: `newInstance()` as the method name would be more consistent with the rest of the code base.

benchmarks/src/main/java/org/questdb/MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.java
```diff
@@ -0,0 +1,136 @@
...
+                caches[i] = cache;
+            }
+        }
+
+        for (int i = 0; i < max; i++) {
+            execute("drop table if exists table" + i);
+        }
+    }
+
+    @Setup(Level.Trial)
```
puzpuzpuz: Seeing `@Setup(Level.Trial)` annotation for method called `tearDown()` is confusing. `setupTrial` might be a better name.

core/src/main/java/io/questdb/std/CharSequenceSortedList.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    public CharSequenceSortedList(int size) {
+        this.buffer = new CharSequence[size];
+        this.size = 0;
+    }
+
+    public void add(CharSequence value) {
+        int p = binSearch(value);
+        if (size > 0 && p > 0 && CharSequence.compare(buffer[p - 1], value) == 0) {
```
puzpuzpuz: nit: could you use `Chars#equals()/equalsNc()` and `Chars#compare()` instead of `CharSequence#compare()` to keep this list consistent with the rest of the code base?

core/src/main/java/io/questdb/std/CharSequenceSortedList.java
```diff
@@ -0,0 +1,243 @@
...
+        size = 0;
+    }
+
+    @Override
+    public ReadOnlyObjList<CharSequence> copy() {
+        return new CharSequenceSortedList(this);
+    }
+
+    @Override
+    public boolean equals(Object that) {
```
puzpuzpuz: nit: we don't seem to need `equals`/`hashCode` implementations in this new list, so we could remove them to avoid maintenance burden.

benchmarks/src/main/java/org/questdb/MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.java
```diff
@@ -0,0 +1,118 @@
...
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+import java.util.concurrent.TimeUnit;
+
+@State(Scope.Benchmark)
+@BenchmarkMode({Mode.Throughput})
+@OutputTimeUnit(TimeUnit.SECONDS)
+public class MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark {
```
puzpuzpuz, line 58: The benchmarks are awfully slow. Would be great to speed them up, e.g. by running the DDLs once per trial or making the benchmark a single-shot one.

maciulis: Thanks for pointing out - updated.

core/src/main/java/io/questdb/std/CharSequenceSortedList.java
maciulis: Thanks for pointing out - updated.

benchmarks/src/main/java/org/questdb/MetadataCacheSnapshotRefreshAddTablesOnlyBenchmark.java
maciulis: Thanks, updated.

core/src/main/java/io/questdb/std/CharSequenceSortedList.java
maciulis: Good point - removed.

core/src/main/java/io/questdb/std/CharSequenceSortedList.java
maciulis: Thanks - also updated.

benchmarks/src/main/java/org/questdb/MetadataCacheSnapshotRefreshRemoveTablesOnlyBenchmark.java
maciulis: Good point - updated.

core/src/main/java/io/questdb/cairo/DefaultLocalCacheSnapshotFactory.java
maciulis: Updated - thanks!

core/src/test/java/io/questdb/test/griffin/ShowTablesTest.java
maciulis: Thanks updated.

core/src/test/java/io/questdb/test/griffin/ShowTablesWithOrderingDisabledTest.java
maciulis: Updated tests names - good point!

core/src/main/java/io/questdb/std/CharSequenceObjMap.java
maciulis: Fair point - updated.

core/src/main/java/io/questdb/cairo/CairoConfiguration.java
maciulis: I could have seen this myself - updated, thanks.

core/src/test/java/io/questdb/test/cairo/MetadataCacheTest.java
maciulis: Very good point - updated.

core/src/test/java/io/questdb/test/griffin/ShowTablesWithOrderingDisabledTest.java
```diff
@@ -0,0 +1,88 @@
...
+            execute("create table deposits(account_no int, currency symbol, amount double)");
+            execute("create table balances(account_no int, currency symbol, amount double)");
+            execute("create table accounts(account_no int, currency symbol)");
+            execute("create table card_payments(account_from_no int, account_to_no int, currency symbol, amount double)");
+            assertSql("table_name\ncard_payments\naccounts\nbalances\ndeposits\n", "SHOW TABLES");
+        });
+    }
+
+    @Test
+    public void testShowTablesWithFunctionReturnsUnoderedList() throws Exception {
```
puzpuzpuz: nit: typo in the test name
```suggestion
    public void testShowTablesWithFunctionReturnsUnorderedList() throws Exception {
```

core/src/main/java/io/questdb/std/CharSequenceSortedList.java
```diff
@@ -0,0 +1,207 @@
...
+    @Override
+    public int indexOf(Object o) {
+        if (o instanceof CharSequence && size > 0) {
+            int pos = binSearch((CharSequence) o);
+
+            if (pos > 0) {
+                pos--;
+            }
+
+            if (buffer[pos] != null && buffer[pos].equals(o)) {
```
puzpuzpuz: Could you use `Chars.equals()` / `Chars.equalsNc()` here as well?

core/src/main/java/io/questdb/PropServerConfiguration.java
```diff
@@ -830,6 +831,7 @@ public PropServerConfiguration(
         );
         this.isReadOnlyInstance = getBoolean(properties, env, PropertyKey.READ_ONLY_INSTANCE, false);
         this.isQueryTracingEnabled = getBoolean(properties, env, PropertyKey.QUERY_TRACING_ENABLED, false);
+        this.cairoMetadataCacheSnapshotOrdered = getBoolean(properties, env, PropertyKey.CAIRO_METADATA_CACHE_SNAPSHOT_ORDERED, false);
```
puzpuzpuz: The default should be `true`, no? Otherwise, no one will see this change.

maciulis: Enabled - was under assumption that "unless not enabled in config - its disabled" to err on the safe side. But now I have made a change.

core/src/main/java/io/questdb/std/CharSequenceSortedList.java
maciulis: Good point - thanks. Put in use `Chars.equalsNc()` using buffer[pos] as second argument so that null check is also covered.

core/src/test/java/io/questdb/test/griffin/ShowTablesWithOrderingDisabledTest.java
maciulis: Thanks for pointing out - should have seen myself. Updated.

puzpuzpuz:
Thanks for the contribution!

PR #6772, opened 2026-02-12
fix(ilp): fix assertion error in ILP decimal parser on malformed input

bluestreak01:
This issue was discovered by Team Atlanta (@LeeSinLiang and @occia) :+1: 

## Summary

- Read the `len` byte in `DecimalBinaryFormatParser` as unsigned (`& 0xFF`) instead of signed. A malformed ILP message with a high `len` byte (e.g. `0x80`) was interpreted as negative, which skipped the `VALUES` parsing state and left `unscaledValues` empty. The subsequent `load()` call then hit an `ArrayIndexOutOfBoundsException` accessing index 0 of the empty list.
- Add a regression test that sends the exact fuzzer-generated payload over TCP and verifies the server stays alive and can process further ILP messages.

## Test plan

- Run `LineTcpReceiverTest#testMalformedBinaryDecimalDoesNotCrash` — sends a malformed 83-byte payload to ILP TCP, then sends a valid ILP message and asserts it was ingested correctly

### DIFF
--- a/core/src/main/java/io/questdb/cutlass/line/tcp/DecimalBinaryFormatParser.java
+++ b/core/src/main/java/io/questdb/cutlass/line/tcp/DecimalBinaryFormatParser.java
@@ -133,7 +133,7 @@ public class DecimalBinaryFormatParser implements QuietCloseable {
                 return false;
             }
             case LEN -> {
-                len = Unsafe.getUnsafe().getByte(addr);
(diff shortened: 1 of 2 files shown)


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/DecimalBinaryFormatParser.java|1|1|100.00%|

PR #6775, opened 2026-02-12
fix(sql): fix alias hiding original column name in CASE with window function

### LINKED ISSUE #6769: ALIAS hides the original column name in CASE with a window function
### To reproduce

Go to demo and run this query (it works):

```sql
select timestamp, price as p,
case when price > 1  then price
 when price < 1  then -p
 END
 from fx_trades;
```

As you can see, I can reference the `price` column either by the original name or by the `p` alias inside the `WHEN` statements. All good

Now, let's add a window function inside the `WHEN` and try to use the unaliased `price` name. Oh, no! `invalid column: price`

```sql
select timestamp, price as p,
case when price > lag(price) OVER (ORDER BY timestamp) then price
 END
 from fx_trades;
```

But I can still reference the column by the `p` alias.

```sql
 select timestamp, price as p,
case when price > lag(price) OVER (ORDER BY timestamp) then p
 END
 from fx_trades;
```

This looks like a bug to me

### QuestDB version:

9.3.2

Public questdb demo box

ext4

### Full Name:

Javier Ramirez

### Affiliation:

QuestDB

- [x] Yes, I have

### Additional context

_No response_

mtopolnik:
Fixes #6769

## Summary

- When a `CASE`/`WHEN` expression contains a window function (e.g., `lag() OVER (...)`), column references outside the window function (in `THEN`/`ELSE` branches) were not emitted to the translating model. This caused "invalid column" errors when using the original column name alongside a SELECT alias (e.g., `price AS p` then `THEN price`).
- Introduce `replaceWindowFunctionOrLiteral()` that chains window function replacement with literal emission in a single tree traversal pass, so all column references are propagated through the model chain.

## Test plan

- New test `testAliasedColumnVisibleByRealNameInCaseThen` reproduces the exact scenario from the issue: `SELECT ts, price AS p, CASE WHEN price > lag(price) OVER (...) THEN price END FROM trades`
- Existing `WindowFunctionTest` suite passes (CASE with window functions, lag/lead, partitioned windows, etc.)
- `SqlParserTest` suite passes (model tree structure unchanged for existing queries)

### DIFF
--- a/core/src/main/java/io/questdb/griffin/SqlOptimiser.java
+++ b/core/src/main/java/io/questdb/griffin/SqlOptimiser.java
@@ -6217,7 +6217,17 @@ public class SqlOptimiser implements Mutable {
         if (n != node) {
             return n;
         }
-        return replaceLiteral(node, translatingModel, innerVirtualModel, true, baseModel, false);
+        n = replaceLiteral(node, translatingModel, innerVirtualModel, true, baseModel, false);
+        if (n != node) {
+            // The node was a column literal. replaceLiteral added it to
+            // translatingModel and innerVirtualModel, but windowModel also
+            // needs a passthrough so the outer virtual model can see it.
+            CharSequence alias = n.token;
+            if (windowModel.getAliasToColumnMap().excludes(alias)) {
+                windowModel.addBottomUpColumnIfNotExists(nextColumn(alias));
+            }
+        }
+        return n;
     }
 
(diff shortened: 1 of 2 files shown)


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|12|12|100.00%|

bluestreak01:
@Coderabbit review critically

PR #6794, opened 2026-02-20
feat(sql): Add array_build() function for dynamic array construction

javier:
Introduces array_build(nDims, size, filler1, ...) that creates DOUBLE[] or DOUBLE[][] arrays with controlled shape and fill values. The size parameter accepts a scalar integer or a DOUBLE[] (using its cardinality). Each filler can be a scalar (repeated for all elements) or a DOUBLE[] (copied element-by-element, NaN-padded if shorter, truncated if longer).

Docs at https://preview-374--questdb-documentation.netlify.app/docs/query/functions/array/#array_build

Includes 34 unit tests covering all code paths (scalar/array fillers, 1D/2D, NaN padding, truncation, null handling, error conditions, large arrays, multi-dim rejection) and 4 fuzz tests (1D/2D scalar fill, 1D/2D array copy identity) with 100 iterations each.


javier:
@coderabbitai, please review critically. Thanks

javier:
@coderabbitai, addressed your comments in latest commit

## CodeRabbit review response

### Integer overflow (CRITICAL)

Already handled. `applyShape()` calls `MutableArray.resetToDefaultStrides()` which uses `Math.multiplyExact()` and checks the product against `maxArrayElementCount`. The attack vector `array_build(2, 200_000_000, 1.0, 2.0)` is caught there. Added `testError2dElementCountExceedsMax` to prove it.

Additionally, this push adds an upper-bound check on `nArraysLong` before the `(int)` cast, and a `DIM_MAX_LEN` check on the array-derived size path (when size comes from a filler array's cardinality). Both caught at compile time / early execution with clear error messages.

### Vect.setMemoryDouble() for scalar fill

Requires `DirectArray` API changes to expose the underlying address and append offset through `MemoryA`. Not in scope for this PR - the current element-by-element loop is correct and consistent with how other array functions populate data.

### setType() per row

Required for correctness. After `ofNull()`, the shape list is cleared. `setType()` reinitializes it so that subsequent `setDimLen()` calls work. The comment at line 222 already documents this. Cannot be removed without changing `DirectArray`/`MutableArray` internals.

### nDims restricted to 1-2

Addressed in this push. The cap is removed - `nArrays` can now be any value >= 1. `nArrays == 1` returns `DOUBLE[]`, `nArrays >= 2` returns `DOUBLE[][]` with `nArrays` rows. The return type is fixed at compile time (function signature resolution), so we cap dimensionality at 2 but allow any number of rows within that 2D array.

True 3D+ would require a different function signature or dynamic return types, which QuestDB's function framework doesn't support.

### DOUBLE only

Only `DOUBLE` is enabled in `arrayTypeSet`. Extending to `INT[]`, `LONG[]`, etc. is a separate effort.

### Bulk copy optimization for truncation

Would need to reach into `ArrayView`'s underlying memory address. The current loop is correct. Not in scope.

### Thread safety

Standard pattern. The engine creates per-worker function instances via `duplicate()`. `isThreadSafe() = false` is consistent with all other `DirectArray`-backed functions.

### Test coverage

This push adds 7 new tests and converts 1 error test to a success test:
- `test2dArrayFillerNaNPad` - NaN padding in 2D
- `test2dArrayFillerTruncate` - truncation in 2D
- `testError2dElementCountExceedsMax` - overflow via `applyShape()`
- `testErrorNArraysOverflow` - huge nArrays literal caught before int cast
- `testNullArrayFillerFromColumn` - per-row null/non-null array filler
- `testOrderBookStyle` - nArrays=4 real-world use case
- `testNArraysThreeScalarFillers` - nArrays=3 success (was error test)
- `testScalarFillNd` (fuzz) - random nArrays 2-6 vs ARRAY literal equivalence

bluestreak01 (approved):

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 109 / 127 (85.83%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/BuildArrayFunctionFactory.java|109|127|85.83%|

Issue #6808, opened 2026-02-24
array_sort and array_reverse


javier:
### Is your feature request related to a problem?

add functions to 

* sort an array (as in postgres, sort by natural order, accept a boolean for reverse sort, another for null values positioning)
* reverse an array, which would be equivalent to array_sort with the reverse flag for sorted arrays, but since that wouldn't work for unsorted arrays, the reverse would just reverse the values. This is interesting for example if you have a list of values that are sorted by timestamp not by value, and you want to reverse the timestamp order

### Full Name:

javier

### Affiliation:

questdb

PR #6820, opened 2026-02-25
feat(sql): add array_sort and array_reverse functions for double arrays

javier:
Fixes https://github.com/questdb/questdb/issues/6808
Documentation at https://github.com/questdb/documentation/pull/378

Add `array_sort(DOUBLE[])` and `array_reverse(DOUBLE[])` scalar functions that operate on double arrays of any dimensionality.

`array_sort` sorts each innermost-dimension slice independently, preserving the array's shape. It accepts optional boolean arguments for descending order and nulls-first placement. `array_reverse` reverses each innermost-dimension slice.

Both functions handle NULL arrays, empty arrays, NaN values, and multidimensional inputs. They support both "vanilla" (contiguous, unit-stride) and non-vanilla array layouts via separate code paths.

**Design notes**
- `sortBuffer` grows on demand and stays at peak size for the cursor's lifetime. Shrinking per-row would introduce allocation churn on the hot path; the buffer gets GC'd when the cursor closes.
- No per-function heap bound on `new double[lastDim]` - consistent with other array functions (`array_build`, `array_agg`). A shared size policy across array functions would be a separate effort.

## Test plan
- 18 unit tests for `array_sort` covering 1D, 2D, 3D arrays, NaN handling, descending/nulls-first combinations, NULL input, empty arrays, and mixed-sign values
- 11 unit tests for `array_reverse` covering the same shape and edge-case matrix
- 11 fuzz tests (100-200 iterations each) verifying sort correctness, reverse-reverse identity, and sort-reverse equivalence against reference implementations
- All 40 tests pass locally

### DIFF
--- a/.github/workflows/glibc_smoke_test.yml
+++ b/.github/workflows/glibc_smoke_test.yml
@@ -2,6 +2,13 @@ name: GLIBC Smoke Test
 # Testing GLIBC compatibility
 
 on:
+  # Seed the GitHub Actions cache on master so PRs get cache hits on first run
+  # Reasoning: cache are not shared across branches, but feature branches can read from the default branch cache
+  # So we need to run the build on master to populate the cache.
+  push:
+    branches: [ master ]
+    paths:
+      - '**/pom.xml'
   pull_request:
     types: [ synchronize, opened, reopened, edited ]
     paths:
@@ -35,17 +42,34 @@ jobs:
           tar -xf node-v20.9.0-linux-x64-glibc-217.tar.xz --strip-components 1 -C /node20217
           ldd /__e/node20/bin/node
       - uses: actions/checkout@v4
+      - name: Cache JDK
+        uses: actions/cache@v4
+        with:
+          path: ${{ runner.tool_cache }}/Java_Temurin-Hotspot_jdk
+          key: java-temurin-17-${{ runner.os }}-${{ runner.arch }}
       - uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
           java-version: '17'
-          cache: 'maven'
+      - name: Cache Maven installation and repository
+        id: maven-cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            /opt/apache-maven-3.9.9
+            /root/.m2/repository
+          key: maven-3.9.9-repo-${{ runner.arch }}-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            maven-3.9.9-repo-${{ runner.arch }}-
       - name: Install Apache Maven
+        if: steps.maven-cache.outputs.cache-hit != 'true'
         # Amazon Linux 2 has too old version of Maven, so we need to install a newer one
         run: |
-          wget https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.6/apache-maven-3.9.6-bin.tar.gz
-          tar xf apache-maven-3.9.6-bin.tar.gz -C /opt
-          ln -sf /opt/apache-maven-3.9.6 /opt/maven
+          wget https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.9/apache-maven-3.9.9-bin.tar.gz
+          tar xf apache-maven-3.9.9-bin.tar.gz -C /opt
+      - name: Setup Maven
+        run: |
+          ln -sf /opt/apache-maven-3.9.9 /opt/maven
           echo "MAVEN_HOME=/opt/maven" >> $GITHUB_ENV
           echo "PATH=/opt/maven/bin:$PATH" >> $GITHUB_ENV
       - name: Build distribution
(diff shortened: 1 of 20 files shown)


javier:
@coderabbitai please review critically

javier:
@coderabbitai Thanks for the thorough review. Addressed the actionable items in bcc74f1cc5 and responding to each point below.

---

**🔴 Unbounded Memory Allocation**

The sort buffer size equals the last dimension length, which is bounded by `maxArrayElementCount()` (default 1M, configurable per-server). `DirectArray.applyShape()` enforces this limit before the sort function ever runs - you cannot construct an array that exceeds it. The heap allocation for a 1M-element sort buffer is 8MB, comparable to what other array operations (`cumsum`, `flatten`, `add`) allocate internally. No additional circuit breaker is needed because the bound is enforced upstream at array creation time.

**🔴 Coverage Gap → Fixed**

Added `testMultiDimNonVanilla()` (exercises `reverseRecursive` via a `[1:, 2:]` sliced non-vanilla array) and `testParallel()` (4-worker WorkerPool verifying double-reverse preserves `array_sum`). These cover the 16 previously uncovered lines.

**🟡 Non-Vanilla Recursive Overhead**

Recursion depth equals the number of dimensions (typically 1-3). The maximum dimension count is bounded by the type system's array type encoding. This is the same recursive pattern used by every array function in the codebase (`cumsum`, `add`, `flatten`, `max`, `min`, `round`). Stack overflow is not a realistic concern with dimension counts in the single digits.

**🟡 Missing Edge Case Tests**

- **Huge arrays / deep nesting:** Arrays are bounded by `maxArrayElementCount()` at creation time. Dimension count is bounded by the type system. The sort function cannot encounter unbounded input because the bound is enforced upstream, not in the function itself. The fuzz test already covers randomly-sized arrays across hundreds of iterations.
- **All-identical values:** Java's dual-pivot quicksort (`Arrays.sort`) handles equal elements in O(n). No degenerate case exists.
- **Zero-stride edge cases:** Not possible to construct. The array slicing implementation always produces positive strides.
- **Pathological shapes:** Covered by the fuzz test which generates random dimension shapes.

**🟡 Buffer Thrashing → Fixed**

Changed buffer growth to use `Numbers.ceilPow2(lastDim)` for power-of-2 rounding. This is the standard QuestDB pattern for growable buffers and eliminates repeated allocations for incrementally larger arrays.

**🟡 Infinity vs NaN**

QuestDB's array storage maps all non-finite values (Infinity, -Infinity, NaN) to NaN. This is a property of the storage layer (`DirectArray`/`ColumnType`), not specific to sort/reverse. The test comments already document this behavior inline where it matters.

**🟢 Design Questions**

1. **Three factory classes:** Yes, this is the standard QuestDB pattern for function overloading. The signature string (`D[]`, `D[]t`, `D[]tt`) encodes argument types, and each distinct signature needs its own factory. One factory per signature.
2. **`isThreadSafe()=false`:** Correct because of mutable instance state (`sortBuffer`, `DirectArray`). The SQL engine creates per-worker function instances when parallelism is needed. The `testParallel()` tests in both sort and reverse verify this works correctly.
3. **`INITIAL_BUFFER_SIZE=64`:** 64 doubles = 512 bytes, fits in L1 cache. Chosen to cover common small arrays without over-allocating. Now documented by the extracted constant name.

**🔵 Verification Summary**

1. Coverage gaps: Fixed with two new tests, bringing reverse from 69% to ~95%.
2. Memory policy: Bounded upstream by `maxArrayElementCount()`. No unbounded growth possible.
3. Buffer thrashing: Fixed with `Numbers.ceilPow2()`.
4. Edge cases: Covered by fuzz tests + upstream bounds. No additional adversarial tests needed.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 145 / 151 (96.03%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/DoubleArraySortFunctionFactory.java|86|90|95.56%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/DoubleArrayReverseFunctionFactory.java|50|52|96.15%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/DoubleArraySortDescFunctionFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/DoubleArraySortFullFunctionFactory.java|5|5|100.00%|

glasstiger:
Claude review identified test coverage gaps, which might be worth to fix.
Otherwise all looks good, so approved it.

```
  1. No 3D+ array tests

  Both sort and reverse only test 1D and 2D arrays (vanilla + non-vanilla via slicing). Adding a 3D test would exercise the recursive dimension traversal more thoroughly. This is the
  most significant test gap.

  2. Fuzz tests only cover 1D arrays

  DoubleArraySortFuzzTest generates 1D arrays exclusively. The fuzz test infrastructure could be extended to generate random 2D/3D arrays with random shapes. This would significantly
  strengthen confidence in the recursive code paths.

  3. No test triggers buffer growth

  INITIAL_BUFFER_SIZE = 64, but no test creates an innermost dimension > 64. Adding a test with e.g. 100+ elements per slice would verify the Numbers.ceilPow2 resize logic.

  4. No toPlan output verification

  The toPlan method on the sort Func class has branching logic (conditionally showing descending/nullsFirst), but no test verifies the plan output.
```

PR #6814, opened 2026-02-24
chore(sql): speed up horizon join time frame search for large RHS tables

puzpuzpuz:
Speeds up HORIZON JOIN in case of very large right-hand-side tables.

### Implementation

`HorizonJoinTimeFrameHelper.findAsOfRow()` now uses a `TimeFrameCursor#seekEstimate()` call to binary-search to the target timestamp's vicinity before linearly scanning frames, reducing frame iteration from O(N) to O(log N) for large right-hand-side tables.

PR #6822, opened 2026-02-25
perf(sql): speed up ASOF and WINDOW JOINs for large right-hand-side tables

puzpuzpuz:
The optimization mirrors what HORIZON JOIN already does in `HorizonJoinTimeFrameHelper` (added in #6814). Without it, the first lookup linearly scans through all slave time frames preceding the master's first timestamp, which is O(N) in the number of frames. With `seekEstimate()`, the initial positioning is O(log P) where P is the number of partitions.

## Implementation

- `AbstractAsOfJoinFastRecordCursor.openSlaveFrame()` now calls `TimeFrameCursor.seekEstimate()` on the first slave frame lookup (when `slaveFrameIndex == -1`) to binary-search directly to the target partition instead of linearly scanning all preceding frames. This benefits all ASOF JOIN and LT JOIN fast-path factories.
- `WindowJoinTimeFrameHelper.findRowLo()` and `findRowLoWithPrevailing()` now call `seekEstimate()` on the first lookup (when `bookmarkedFrameIndex == -1`) with the same partition-skipping behavior. When prevailing is enabled, the seeked frame's last row is recorded as the prevailing candidate. This benefits both sync and async WINDOW JOIN factories.

### DIFF
--- a/core/rust/qdbr/src/parquet_read/decode.rs
+++ b/core/rust/qdbr/src/parquet_read/decode.rs
@@ -760,6 +760,8 @@ impl ParquetDecoder {
 
     pub fn find_row_group_by_timestamp(
         &self,
+        file_ptr: *const u8,
+        file_size: u64,
         timestamp: i64,
         row_lo: usize,
         row_hi: usize,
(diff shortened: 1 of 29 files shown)


bluestreak01:
@Coderabbit pls review this PR critically

puzpuzpuz:
> `@bluestreak01`, I've identified **three critical issues** in this optimization that could cause incorrect query results

@coderabbitai please find responses to your comments below.

## Claim 1: "Timestamp Conversion Precision Loss"

Invalid. `seekEstimate` is a coarse positioning heuristic — it finds the last frame whose partition ceiling is `<= timestamp`. Truncating the timestamp downward via integer division is safe: it may cause `seekEstimate` to land one frame earlier than optimal, but the subsequent `next()` loop catches up. This costs at most one extra frame scan — negligible.

The suggested fix (ceiling division) would be actively harmful: rounding up could cause `seekEstimate` to jump past the correct frame. The `next()` call after it would move even further forward, and the ASOF join's `openSlaveFrame()` backward path isn't designed to recover from overshooting.

The same truncating-division pattern is already used in `HorizonJoinTimeFrameHelper.findAsOfRow()` (line 308).

## Claim 2: "Incorrect Prevailing Assumption After seekEstimate"

Invalid. This misunderstands partition ceilings.

Partition ceilings are exclusive upper bounds. `estimatePartitionHi()` computes `min(ceil(partitionTimestamp), nextPartitionTimestamp)`. For example, for a DAY-partitioned table with partition timestamp `2024-01-01T00:00:00`, the ceiling is `2024-01-02T00:00:00`. All rows in that partition have timestamps in `[2024-01-01, 2024-01-02)` — strictly less than the ceiling.

So when `seekEstimate` finds a frame with `ceiling <= timestampLo`, all rows in that frame have timestamps `< ceiling <= timestampLo`. Recording the last row as a prevailing candidate (meaning `< timestampLo`) is correct.

## Claim 3: "Off-By-One in maxTsHi Calculation"

Pre-existing code, not introduced by this PR.

The last two partitions get `Long.MAX_VALUE` as `maxTsHi` instead of just the last one. For non-split partitions, `ceil(partitionTimestamp)` is already correct, so the missing cap has no effect. For split partitions at the second-to-last position, the ceiling could be slightly inflated, but `seekEstimate` results are always validated against actual row timestamps after `open()`, so this causes at most one extra frame scan.

If worth fixing, it's a one-character change (`- 2` → `- 1`) and should be a separate PR.

## Claim 4: "Missing Test Coverage"

Wrong. The diff includes 199 lines of test additions across three files: `AsOfJoinTest.java` (+55), `AsOfJoinNoKeyTest.java` (+36), `WindowJoinTest.java` (+108). The coverage report confirms 100% coverage of new lines.

puzpuzpuz:
Suggestion 3 is addressed in f2af0eb9eecb5e6dee6308209e53f227f66a8265

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/join/WindowJoinTimeFrameHelper.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AbstractAsOfJoinFastRecordCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ConcurrentTimeFrameCursor.java|1|1|100.00%|

PR #6832, opened 2026-02-27
fix(core): fix crash on LATEST BY ALL queries over large tables

ideoma:
## Summary

- Fix `Numbers.MAX_SAFE_INT_POW_2` from `1L << 31` to `1L << 30`. The old value (2^31) does not fit in a signed 32-bit int, so the rehash overflow guard `newKeyCapacity > MAX_SAFE_INT_POW_2` let exactly 2^31 through. The subsequent `(int)` cast produced `Integer.MIN_VALUE`, and `clear()` fed ~18 EB to native `memset`, causing a SIGSEGV.
- Deduplicate the constant from `Unordered4Map`, `Unordered8Map`, and `UnorderedVarcharMap` — each had a private copy with the same bug.
- Add overflow guard tests in `MapTest` and `OrderedMapTest`.

### Crash chain

1. A `LATEST BY ALL` query on a large table fills an `OrderedMap` until `keyCapacity` reaches 2^30.
2. `rehash()` doubles to `newKeyCapacity = 1L << 31`. The guard (`> 1L << 31`) is false at the boundary — allocation of 16 GB succeeds.
3. `keyCapacity = (int)(1L << 31)` truncates to `Integer.MIN_VALUE` (−2,147,483,648).
4. `map.clear()` computes `(long)keyCapacity << 3` = `0xFFFFFFFC00000000` (~18.4 EB) and passes it to native `memset` → `SIGSEGV`.

### Fix

`MAX_SAFE_INT_POW_2 = 1L << 30` makes the guard reject `newKeyCapacity = 2^31`, throwing a clean `CairoException("map capacity overflow")` instead of crashing the JVM.

### DIFF
--- a/core/src/test/java/io/questdb/test/cairo/map/OrderedMapTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/map/OrderedMapTest.java
@@ -1383,7 +1383,8 @@ public class OrderedMapTest extends AbstractCairoTest {
             ColumnTypes types = new SingleColumnType(ColumnType.LONG);
             try (OrderedMap map = new OrderedMap(Numbers.SIZE_1MB, types, null, 16, 0.5, Integer.MAX_VALUE)) {
                 try {
-                    map.setKeyCapacity(Integer.MAX_VALUE);
+                    // should fail with 0.75 load factor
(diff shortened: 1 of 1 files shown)


ideoma:
@coderabbitia review

glasstiger:
#### [PR Coverage check]

PR #6834, opened 2026-03-01
feat(sql): add uint16/uint32/uint64 types with bitmap-based null support

nicholas-a-guerra:
## Summary

- Add UINT16, UINT32, and UINT64 column types with full storage, SQL, and ILP support
- Add bitmap-based null support for BOOLEAN, BYTE, SHORT, and the new UINT types via per-column `.n` files storing 1 bit per row
- Add JIT-compiled filter support for bitmap-null columns in scalar mode, avoiding interpreted fallback
- Fix null bitmap corruption in O3 appends, WAL copies, and successive O3 commits
- Fix equality functions to return correct results when operands are null

### New unsigned integer types

UINT16 (0–65535), UINT32 (0–4294967295), and UINT64 (0–18446744073709551615) column types with:
- Column-oriented storage with 2/4/8 byte fixed-width columns
- SQL arithmetic, comparison, CAST, GROUP BY, and aggregate function support
- ILP (InfluxDB Line Protocol) ingestion support
- PostgreSQL wire protocol support
- CSV import support
- Bitmap-based null tracking (no sentinel value consumed from the value range)

### Bitmap-based null support

BOOLEAN, BYTE, and SHORT previously had no null support — inserting NULL stored 0 (indistinguishable from the actual value 0). This PR adds a per-column `.n` bitmap file storing 1 bit per row (1=null, 0=not-null, packed LSB-first). This extends to the new UINT types as well.

Key changes:
- `ColumnType.needsNullBitmap()` identifies bitmap-null types
- `Record.isNull(col)` reads the bitmap for these types
- `RecordToRowCopierUtils` copies null bitmaps during INSERT, type conversion, and WAL apply
- `TableWriter` and `WalWriter` maintain bitmap files during writes, O3 merges, and WAL commits

### JIT bitmap-null support

The JIT previously rejected bitmap-null columns, forcing fallback to interpreted mode (5-20x slower). This PR adds bitmap-null awareness to the JIT scalar code path:

1. `CompiledFilterIRSerializer` forces scalar mode and encodes a bitmap-null flag in MEM operands
2. `PageFrameReduceTask` passes null bitmap addresses via `auxAddresses`
3. `read_mem()` in x86.h and aarch64.h tests the bitmap bit after loading i8/i16 values; if null, substitutes `INT_NULL` sentinel so existing JIT comparison logic works unchanged
4. `NOT(bool_col)` on bitmap-null booleans expands to `bool_col = false` instead of `NOT(bool_col = true)`
5. `serializeNumber()` parses byte/short literals as int before applying sign, fixing overflow at type boundaries (-128, -32768)

### Storage and WAL fixes

- `O3OpenColumnJob` writes null bitmaps during O3 partition appends and merges
- `TableWriter` merges bitmap bytes correctly when bit offsets are non-zero
- `TableWriter` clears stale bits beyond committed row count after O3 commits
- `WalWriter` clears trailing bits in last byte after memset(0xFF) for ADD COLUMN

### Equality function fixes

- `NotFunctionFactory` returns false for null instead of true
- `EqBoolean/Byte/Short/Int/Long` return false for negated null comparisons

### Limitations

- `IS NULL` / `IS NOT NULL` on bitmap-null types falls back to interpreted mode in JIT
- SIMD mode disabled for bitmap-null columns (bitmap bit testing is per-row)
- `WHERE col != X` includes null rows in JIT for all types (pre-existing behavior)
- UINT types fall back to interpreted mode in JIT (unsigned comparison semantics not yet implemented)

## Test plan

- NullBitmapTest: 71 tests covering WHERE filters, O3 appends/merges, NOT boolean, IS NULL/IS NOT NULL, IN, cross-partition O3, successive O3 merges, type conversion
- CompiledFilterTest: 29 tests verifying JIT compilation for all column types
- GroupByTest: 92 tests for GROUP BY with bitmap-null columns
- EqBoolean/Byte/Short/Int/Long tests: null equality semantics
- Live integration testing: table creation, range edge cases (-128, -32768), O3, ALTER COLUMN, null counting, WHERE, UNION ALL, GROUP BY, CASE, COALESCE, UINT types, EXPLAIN plans


nwoolmer:
@coderabbitai review

core/src/main/java/io/questdb/griffin/engine/functions/eq/EqBooleanFunctionFactory.java
```diff
@@ -59,7 +63,43 @@ public Func(Function left, Function right) {
 
         @Override
         public boolean getBool(Record rec) {
+            final boolean leftNull = left.isNull(rec);
+            final boolean rightNull = right.isNull(rec);
+            if (leftNull || rightNull) {
+                if (negated) {
```
nwoolmer: **Bug (Critical): `!=` null semantics inconsistent with `NOT(a = b)`**

When `negated == true` and either side is null, this always returns `false`. But consider:
- `NOT(5 = NULL)` → `NOT(false)` → `true`
- `5 != NULL` → hits this branch → `false`

This breaks the SQL invariant that `a != b ≡ NOT(a = b)`. The same issue exists in `EqByteFunctionFactory`, `EqShortFunctionFactory`, `EqIntFunctionFactory`, and `EqLongFunctionFactory`.

To fix: the negated path should return `!(leftNull && rightNull)` (i.e., the negation of the non-negated null result), or better yet, both paths should return `false` for SQL-standard three-valued logic (`NULL = NULL` → `NULL`/false, `NULL != NULL` → `NULL`/false).

core/src/main/java/io/questdb/griffin/RecordToRowCopierUtils.java
```diff
@@ -2944,4 +3138,218 @@ private static boolean validateChunkSizes(
...
+            case ColumnType.LONG -> {
+                long v = rec.getLong(fromCol);
+                if (v != Numbers.LONG_NULL) row.putBool(toCol, v != 0);
+            }
+        }
+    }
+
+    public static void copyAnyToUInt32(Record rec, TableWriter.Row row, int fromCol, int toCol, int fromTypeTag) {
+        switch (fromTypeTag) {
+            case ColumnType.INT -> row.putInt(toCol, rec.getInt(fromCol));
```
nwoolmer: **Bug (Critical): Missing null guard for INT → UINT32 conversion**

All other source types in this method check for null before copying, but the `INT` case blindly copies `rec.getInt(fromCol)`. The INT sentinel value `Integer.MIN_VALUE` (`0x80000000`) will be stored as the valid UINT32 value `2147483648` instead of setting the null bitmap.

Suggested fix:
```java
case ColumnType.INT -> {
    if (rec.getInt(fromCol) == Numbers.INT_NULL) {
        row.putNull(toCol);
    } else {
        row.putInt(toCol, rec.getInt(fromCol));
    }
}
```

core/src/main/java/io/questdb/griffin/RecordToRowCopierUtils.java
```diff
@@ -2944,4 +3138,218 @@ private static boolean validateChunkSizes(
...
+        }
+    }
+
+    public static void copyAnyToUInt64(Record rec, TableWriter.Row row, int fromCol, int toCol, int fromTypeTag) {
+        switch (fromTypeTag) {
+            case ColumnType.INT -> {
+                int v = rec.getInt(fromCol);
+                if (v != Numbers.INT_NULL) row.putLong(toCol, v);
+            }
+            case ColumnType.LONG -> row.putLong(toCol, rec.getLong(fromCol));
```
nwoolmer: **Bug (Critical): Missing null guard for LONG → UINT64 conversion**

Same issue as INT → UINT32 above. `Long.MIN_VALUE` (the LONG null sentinel) will be stored as a valid UINT64 value instead of null.

core/src/main/java/io/questdb/cairo/TableWriter.java
```diff
@@ -4657,6 +4905,91 @@ private void cthMergeWalFixColumnWithLag(int columnIndex, int columnType, long m
...
+                        walBitmapAddr = walBm.addressOf(0);
+                    }
+                }
+
+                // Allocate destination bitmap
+                long destBitmapSize = (mergeCount + 7) >> 3;
+                long destBitmapAddr = Unsafe.malloc(destBitmapSize, MemoryTag.NATIVE_O3);
+                Vect.memset(destBitmapAddr, destBitmapSize, 0);
+
+                long lagBitmapBase = Math.abs(lagBitmapMappedAddr);
```
nwoolmer: **Bug (Critical): `Math.abs()` on a memory address is semantically meaningless and dangerous**

Memory addresses from `mapAppendColumnBuffer` should always be positive (or 0/negative on error). Taking `Math.abs()` of an error return would produce a garbage pointer, leading to SEGV or silent memory corruption. This should either be used directly (if the address is known to be valid) or error-checked properly.

core/src/main/java/io/questdb/cairo/O3OpenColumnJob.java
```diff
@@ -2552,6 +2868,54 @@ private static void mergeFixColumn(
                 dstVFd = openRW(ff, BitmapIndexUtils.valueFileName(pathToNewPartition.trimTo(pNewLen), columnName, columnNameTxn), LOG, tableWriter.getConfiguration().getWriterFileOpenOpts());
             }
 
+            // Write null bitmap (.n file) for bitmap-null types during O3 merge.
+            // The merge must preserve null bits from both the source partition
+            // (.n file on disk) and the O3 in-memory bitmap.
+            if (ColumnType.needsNullBitmap(columnType)) {
+                long srcDataNullAddr = 0;
+                long srcDataNullSize = 0;
+                long srcDataNullFd = 0;
```
nwoolmer: **Bug (Critical): File descriptor initialized to `0` instead of `-1`**

On Unix, `0` is a valid fd (stdin). If `ff.openRO()` returns `0`, the subsequent `if (srcDataNullFd > 0)` check would skip mapping and closing the fd. Initialize to `-1` and check `> -1`, consistent with the rest of the codebase (e.g., `nFd` on nearby lines uses `-1`).

core/src/main/java/io/questdb/griffin/engine/functions/lt/LtIntFunctionFactory.java
```diff
@@ -55,20 +56,35 @@ public Function newInstance(
...
-        public LtIntFunction(Function left, Function right) {
+        public LtIntFunction(Function left, Function right, boolean unsigned32) {
             this.left = left;
             this.right = right;
+            this.unsigned32 = unsigned32;
         }
 
         @Override
         public boolean getBool(Record rec) {
+            if (unsigned32) {
```
nwoolmer: **Bug (High): Comparison operators don't check bitmap nulls for signed byte/short**

The `isNull()` check is only in the `unsigned32` path. For the signed path, the code relies on sentinel-based null detection via `Numbers.lessThan()`. But BYTE and SHORT now use bitmap nulls — `getInt()` returns `0` for null byte/short values, not `INT_NULL`.

This means `WHERE byte_col < 5` would incorrectly include null rows (treating null as `0 < 5 = true`). The same issue exists in `LtLongFunctionFactory`.

The signed path needs an `isNull()` guard as well, at least when the operands are widened from bitmap-null types.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -9156,30 +9157,31 @@ void of(RecordMetadata metadata) {
         sumConstructors.put(INT, SumIntVectorAggregateFunction::new);
         sumConstructors.put(LONG, SumLongVectorAggregateFunction::new);
         sumConstructors.put(LONG256, SumLong256VectorAggregateFunction::new);
-        sumConstructors.put(SHORT, SumShortVectorAggregateFunction::new);
+        // SHORT vector aggregates disabled: native SIMD path does not check null bitmaps
```
nwoolmer: **Performance regression (High): All SHORT SIMD vector aggregates disabled**

This disables vectorized sum/avg/min/max for *all* SHORT columns, not just those with nulls. This is a significant analytical performance regression. The Rosti (vectorized group-by) path is also disabled for SHORT.

Consider either:
1. Adding bitmap-null awareness to the SIMD aggregate path
2. Checking at query time whether the column actually has nulls and only falling back when needed
3. At minimum, tracking this as a follow-up issue

core/src/main/java/io/questdb/cairo/NullBitmapMigrator.java
```diff
@@ -0,0 +1,245 @@
...
+        Unsafe.getUnsafe().putByte(addr, (byte) (current | (1 << bitIndex)));
+    }
+
+    private static void writeNFile(FilesFacade ff, LPSZ nFilePath, long bitmapAddr, long bitmapSize) {
+        long fd = ff.openRW(nFilePath, CairoConfiguration.O_NONE);
+        if (fd < 0) {
+            LOG.error().$("could not create null bitmap file [path=").$(nFilePath)
+                    .$(", errno=").$(ff.errno())
+                    .I$();
+            return;
```
nwoolmer: **Bug (High): `writeNFile` silently swallows I/O errors**

Write failures are logged but not propagated. This leaves a potentially corrupt partial `.n` file on disk. Subsequent reads would see the file as existing (passing the `ff.exists()` and `ff.length()` checks in `ensureNullBitmap`) but with incorrect bitmap data, causing random rows to appear null or non-null.

Either throw `CairoException` on failure, or delete the partial file so the next access retries the migration.

core/src/main/java/io/questdb/griffin/engine/functions/ByteFunction.java
```diff
@@ -181,6 +186,7 @@ public RecordCursorFactory getRecordCursorFactory() {
 
     @Override
     public short getShort(Record rec) {
+        if (isNull(rec)) return 0;
```
nwoolmer, line 188: **Bug (High): Null propagation broken in type widening**

When a BYTE value is null, `getShort()` returns `0`. Since SHORT now uses bitmap nulls (not sentinel `0`), this `0` is treated as a valid value by the caller. Null information is lost during implicit BYTE → SHORT widening.

The same pattern exists in `BooleanFunction.getByte()` and `BooleanFunction.getShort()`. These widening methods should return the target type's null sentinel (e.g., `Numbers.INT_NULL` for `getInt()`) or the caller must check `isNull()` before calling — but the current `return 0` makes the null indistinguishable from a real zero.

core/src/main/java/io/questdb/cairo/ColumnType.java
```diff
@@ -554,6 +572,16 @@ public static boolean isNull(int columnType) {
         return columnType == NULL;
     }
 
+    /**
+     * Returns true if this column type uses a separate null bitmap (.n file)
+     * rather than a sentinel value to represent nulls.
+     */
+    public static boolean needsNullBitmap(int columnType) {
```
nwoolmer, line 583: **Nit: `needsNullBitmap` has redundant and asymmetric checks**

Since `tagOf(UINT16) == SHORT`, the `tag == SHORT` check already covers UINT16, making `isUInt16(columnType)` redundant. Meanwhile `isUInt32` and `isUInt64` are *not* redundant (INT and LONG tags are intentionally excluded). A comment explaining this asymmetry would prevent confusion during maintenance.

core/pom.xml
```diff
@@ -351,6 +351,30 @@
...
+                        <executions>
+                            <execution>
+                                <id>patch-web-console-uint-types</id>
+                                <phase>initialize</phase>
+                                <goals>
+                                    <goal>run</goal>
+                                </goals>
+                                <configuration>
+                                    <target>
+                                        <replace
```
nwoolmer, line 377: **Code quality (Minor): Web console patching via string replacement on minified JS is fragile**

This performs a literal match on minified JavaScript. If the web console upgrades its bundler, changes minification settings, reorders types, or changes quote styles, this will silently fail — the UINT types simply won't appear in the UI with no build error. Consider doing this at the web console source level, or at minimum adding a validation step that the replacement actually matched.

nwoolmer (commented):
For handling migration to the new null bitmap format, I think we should do the following.

- Introduce `NOT NULL` syntax.
- byte/short/boolean columns are automatically migrated to `BYTE/SHORT/BOOLEAN NOT NULL` by this PR (same behaviour as they were before).
- A column like `BYTE NOT NULL` can be converted to `BYTE` using `ALTER COLUMN SET TYPE`. 
- The `NOT NULL` distinction is stored in a bit with the column type/other metadata.
- The new `UINT` types are nullable by default, but can be `UINT64 NOT NULL` too.
- We will review other types in future to figure out how to migrate them from null sentinel -> null bitmap, which will require rewriting data.

The aim is to not cause any regression/change to current  `BYTE/BOOLEAN/SHORT` whilst still adding nulls. For users who don't need the nulls, we can preserve faster code paths.

bluestreak01:
@nicholas-a-guerra

## Code Review: Verified Findings

25 draft findings verified against source code, 6 downgraded as false positives, 19 confirmed findings below.

---

### Critical

**1. Native memory leaks: `doClose()` and `rollback()` don't free O3 bitmap allocations**

`TableWriter.java` — `o3NullBitmapAddrs` and `o3NullBitmapAllocSizes` hold raw `Unsafe.malloc` pointers (allocated in `cthMergeBitmapWithLag`, `setupO3NullBitmapForBypassWal`, `shuffleBlockBitmaps`). Neither `doClose()` (line 5729) nor `rollback()` (line 2938) calls `freeO3MergedBitmaps()`. If the writer is closed or rolled back after a failed O3 commit, native memory leaks permanently.

**Fix:** Add `freeO3MergedBitmaps()` to `doClose()` (before `freeColumns()` at line 5763) and to `rollback()` (after line 2943).

**2. Native memory leak in `cthMergeBitmapWithLag()` on error**

`TableWriter.java:5021` — `destBitmapAddr` is allocated via `Unsafe.malloc`. The merge loop (lines 5028-5051) has no try/finally protecting this allocation. If the loop throws, `destBitmapAddr` leaks — it hasn't been stored in `o3NullBitmapAddrs` yet (that happens at line 5054, after the loop). Compare with `cthO3SortBitmap()` which correctly wraps its allocation in try/catch.

**3. Native memory leak in `setupO3NullBitmapForBypassWal()` on error**

`TableWriter.java:10622` — `o3BitmapAddr` allocated via `Unsafe.malloc`, loops at lines 10627-10656 have no try/finally. If `bitmapMem.addressOf()` throws, the allocation leaks.

**4. Native memory leak in `shuffleBlockBitmaps()` on error**

`TableWriter.java:10709` — Same pattern: `destBitmapAddr` allocated, loop not wrapped in try/finally, address not stored until line 10754.

**5. JIT bitmap-null handling missing for i32/i64 data types**

`x86.h` and `aarch64.h` — The `read_mem()` function only checks `has_bitmap_null` for `data_type_t::i8` and `data_type_t::i16`. The `i32` and `i64` cases contain no bitmap-null logic. JIT-compiled filters on UINT32/UINT64 columns will read raw zero values for null rows instead of recognizing them as NULL. This means `WHERE uint32_col > 0` would incorrectly include null rows.

**6. Dedup comparers use signed comparison for UINT types**

`dedup_comparers.h:163` — `compare_dedup_column_fixed<T>` uses `(l_val > r_val) - (l_val < r_val)` with signed types. In `dedup.cpp`, the template is always instantiated with `int16_t`/`int32_t`/`int64_t` based solely on byte size, with no dispatch for unsigned column types. UINT32 value `4294967295` (stored as `-1` signed) would compare less than `1`, causing incorrect dedup merge ordering for values above the signed boundary.

---

### Moderate

**7. `Numbers.compare(int, int)` silently uses unsigned semantics**

`Numbers.java:644` — New public methods `compare(int, int)`, `compare(long, long)`, `compare(short, short)` all use `Integer.compareUnsigned` / `Long.compareUnsigned` / `Short.compareUnsigned`. The method names don't indicate unsigned semantics. Any future caller expecting signed comparison will get silently wrong results. Should be renamed to `compareUnsigned()` or made package-private.

**8. `NullBitmapMigrator.writeNFile()` does not call `fsync`**

`NullBitmapMigrator.java:217-236` — After writing the `.n` file (`openRW` -> `truncate` -> `write` -> `close`), no `fsync` is called. A crash after `writeNFile` returns but before OS flush loses the `.n` file, requiring re-generation on next open. Other write paths (WalWriter, CopyWalSegmentUtils) do fsync bitmap files.

**9. `WalWriter.setBitmapAppendPosition` has unsafe `jumpTo` ordering**

`WalWriter.java:1780-1784` — The bit-clearing logic is present, but the `jumpTo(bitmapByteCount)` call precedes the read of `bitmapByteCount - 1`. If the byte count falls exactly on a page boundary, `jumpTo` maps the next page and unmaps the page containing the target byte. `TableWriter.java:10509-10520` correctly orders this (jump to last byte first, clear, then jump to end) with an explicit comment documenting the hazard.

**10. O3 bitmap merge is per-bit in Java — no native acceleration**

`O3OpenColumnJob.java` (lines 2376-2513), `TableWriter.java` (lines 3437-3523) — All bitmap merge/sort/copy operations use Java per-bit loops. For a million-row partition, this adds a million iterations of byte-level `Unsafe.getByte`/`putByte` with bit manipulation per bitmap-null column. The `extractRebasedBitmap()` method does have a byte-aligned fast path with `Vect.memcpy`, but non-aligned cases fall back to per-bit loops.

**11. All bitmap-null native aggregates are scalar-only (no SIMD)**

`vec_agg.cpp` — Comment: "These call vanilla (scalar) implementations directly. SIMD versions can be added later." `sumShortBitmapNull`, `minShortBitmapNull`, etc. lose the SSE/AVX dispatching that exists for sentinel-null SHORT. New partitions written after this PR will use bitmap nulls, so SHORT aggregations will permanently regress on new data.

**12. `isNull()` overhead added to every type-widening getter on BYTE/SHORT/BOOLEAN**

`ByteFunction.java`, `ShortFunction.java`, `BooleanFunction.java` — Every `getInt()`, `getLong()`, `getDouble()`, `getFloat()`, `getDate()`, `getTimestamp()` now starts with `if (isNull(rec)) return ...`. For column functions, `isNull(rec)` performs ~7 operations (two `DirectLongList.get()` calls + bitmap byte read + bit extraction). This affects all existing BYTE/SHORT/BOOLEAN queries on the interpreted path, even for legacy data with no bitmap.

**13. Vector aggregate COUNT not registered for any UINT type**

`SqlCodeGenerator.java` — No `countConstructors.put(UINT16/UINT32/UINT64, ...)` entries. COUNT on UINT columns falls back to row-by-row interpreted path. Also missing: `minConstructors` for UINT16/UINT64, `maxConstructors` for UINT16/UINT64, `avgConstructors` for all UINT types.

**14. Test coverage gaps**

- No ILP ingestion tests for UINT types
- No CSV import tests for UINT types
- No O3 dedup tests with bitmap-null columns as dedup keys
- UINT types not in fuzz testing at all; BOOLEAN/BYTE/SHORT marked non-nullable in fuzz
- No multi-byte bitmap test (all tests use <8 rows per partition, only testing bits within a single byte)
- No JIT filter execution tests for UINT columns
- No aggregate tests for SHORT/BYTE with bitmap nulls specifically

---

### Minor

**15. `stampNullsInFixedColumn` uses old-style switch**

`ColumnTypeConverter.java:1029-1072` — Uses `case X: ... break;` instead of enhanced `case X -> { ... }` switch per coding guidelines.

**16. Duplicated bitmap bit-test utility code**

The same `(bitmap[row >> 3] & (1 << (row & 7))) != 0` pattern is implemented in 5+ Java methods (`ColumnTypeConverter.isNullInBitmap`, `O3OpenColumnJob.isNullBitmapBitSet`, `NullBitmapMigrator.setBit`, etc.) and 3 C++ functions. Should be consolidated.

**17. PR labels incomplete**

Current: `SQL`, `Core`. Missing: `New feature`, `WAL`, `ILP`, `Bug`.

**18. Commit message violations**

Three commits use Conventional Commits prefixes (prohibited for commits per project conventions). Multiple commit titles exceed 50 characters. Three "Retrigger CI" commits should be squashed before merge.

**19. Web console patching via maven-antrun-plugin is fragile**

`core/pom.xml` — String replacement in minified JavaScript depends on exact match of `{label:"SHORT"...},{label:"STRING"...}`. Any web console rebuild that changes minification will break the build (`failOnNoReplacements="yes"`).

---

| Original claim | Why dismissed |
|---|---|
| `SortColumnComparer` doesn't adjust bitmap index for `column_top` | `SortColumnComparer` is only used from paths where `column_top = 0` (verified in `TableWriter.deduplicateSortedIndex()` line 5539 and `processWalCommitBlock` line 8925). Latent code quality issue, not a current bug. |
| `SortColumnComparer` all-column-top case accesses invalid memory | The sentinel bitmap address is set in the O3 merge path, which exclusively uses `MergeColumnComparer` (not `SortColumnComparer`). `MergeColumnComparer` short-circuits on `col_index < column_top` before dereferencing. |
| `WalWriter.setBitmapAppendPosition` does not clear stale trailing bits | The clearing logic IS present at lines 1781-1784. The claim confused "missing feature" with a subtler `jumpTo` ordering difference (reported as finding #9 instead). |
| `o3NullBitmapAddrs` lacks memory visibility guarantees | Adequate via `SOUnboundedCountDownLatch` volatile semantics and ring queue barriers. Fragile but correct. |
| `NullBitmapMigrator` TOCTOU race causes corruption | The race exists but is benign: both threads scan the same immutable `.d` file and produce identical bitmap content. The write is idempotent. |
| `UIntXXFunction` base classes don't override `isNull()` — current bug | All 10+ existing concrete subclasses override `isNull()`. Latent risk for future subclasses, not a current bug. |

---

### Summary

**Verdict: Request changes.**

6 critical issues (4 native memory leaks on error paths, JIT missing UINT32/UINT64 bitmap-null handling, signed dedup comparison for unsigned types) must be fixed before merge. The memory leaks are straightforward (add try/finally, add `freeO3MergedBitmaps()` to close/rollback). The JIT and dedup issues require more design work.

The performance regressions on existing BYTE/SHORT/BOOLEAN workloads (scalar-only aggregates, per-call `isNull()` overhead) are significant and align with @nwoolmer's suggestion to introduce `NOT NULL` semantics so legacy columns preserve fast paths.

**Verification stats:** 25 draft findings verified, 6 downgraded as false positives, 19 confirmed findings.

nwoolmer:
Hey @nicholas-a-guerra , sorry for the delay, it has been very busy!

After some consultation with @puzpuzpuz , here's the plan. It will be three PRs that we will pick up based on this PR.

1. We'll add NOT NULL support and migrate existing boolean, byte, short etc. columns to NOT NULL. This constrains that part of the type management. This also allows us to optimise JIT paths for existing columns that are always nullable.

2. We'll introduce nullable bitmaps and add a migration path for old columns to the new format. That needs to go through a release cycle to ensure that users can roll back if there's a problem.

3. Unsigned ints is a very fast follow after 1 + 2, since the 'hard' destructive parts are done.

I'm going to take this on in the background so do bear with, this PR has been very helpful in proving out the issues and scope, and making sure we can get the right solution. I'll add you as a co-author to the new PRs, and if you can help us thoroughly test each phase, that would be great.

nwoolmer:
Ok, here is the plan. We'll take ownership of this now @nicholas-a-guerra 

We need to phase this a little bit and merge in stages. 

1. Introduce NOT NULL types, with associated query engine and JIT changes. Migrate existing boolean/short/byte etc. automatically to NOT NULL variants.

2. Introduce alternate null format (nullable bitmaps). Migrate NULLABLE columns stored in prior native format to the new nullable format. Parquet partitions won't need to change (they already don't use sentinels).

3. Add uint types, which is the smallest job after the above two have been handled.

Issue #4806, opened 2024-07-23
OOM logging is badly formatted


ideoma:
### To reproduce

```
2024-07-22T17:51:34.453589Z I i.q.c.TableWriter o3 commit [table=testWalWriteFullRandom_nonwal, o3RowCount=1769558]
sun.misc.Unsafe.allocateMemory() OutOfMemoryError [RSS_MEM_USED=4368849184, size=1055916032, memoryTag=45], original message: 2024-07-22T17:51:36.654908Z C i.q.c.TableWriter could not sort varsize o3 column [table=testWalWriteFullRandom_nonwal, column=var_top, type=VARCHAR, long0=1717520080960, long1=-1, long2=-1, long3=-1, errno=-1, ex=sun.misc.Unsafe.allocateMemory() OutOfMemoryError [RSS_MEM_USED=4368849184, size=1055916032, memoryTag=45], original message: ]

2024-07-22T17:51:37.098299Z I i.q.c.p.WriterPool << [table=`testWalWriteFullRandom_nonwal~`, thread=811]
2024-07-22T17:51:37.124256Z E i.q.t.AbstractCairoTest Error in test: 
io.questdb.cairo.CairoException: [-1] commit failed, see logs for details [table=testWalWriteFullRandom_nonwal, tableDir=testWalWriteFullRandom_nonwal~]
	at io.questdb.cairo.CairoException.instance(CairoException.java:323)
	at io.questdb.cairo.CairoException.critical(CairoException.java:69)
	at io.questdb.cairo.TableWriter.checkO3Errors(TableWriter.java:3399)
	at io.questdb.cairo.TableWriter.consumeColumnTasks(TableWriter.java:3706)
	at io.questdb.cairo.TableWriter.dispatchColumnTasks(TableWriter.java:4631)
	at io.questdb.cairo.TableWriter.o3Commit(TableWriter.java:5478)
	at io.questdb.cairo.TableWriter.commit(TableWriter.java:3525)
	at io.questdb.cairo.TableWriter.commit(TableWriter.java:994)
	at io.questdb.test.griffin.wal.FuzzRunner.applyNonWal(FuzzRunner.java:233)
	at io.questdb.test.griffin.wal.FuzzRunner.runFuzz(FuzzRunner.java:800)
	at io.questdb.test.griffin.wal.AbstractFuzzTest.lambda$fullRandomFuzz$0(AbstractFuzzTest.java:135)
[...]

PR #6839, opened 2026-03-03
fix(core): fix exceptions nested logs and added human-readable memory…

Stacvirus:
I've opened this PR to fix issue: [4806](https://github.com/questdb/questdb/issues/4806)

- The update provide fix to exception logs present in the Unsafe class which was nested, used println and memory was not human readable.
- I have removed the println to use the project logs implementation which ultimate remove the nested logs.
- I have implemented the charsink putSize() method in the CairoException class for human readable memory logs.


core/src/main/java/io/questdb/std/Unsafe.java
```diff
@@ -251,16 +251,13 @@ public static long malloc(long size, int memoryTag) {
...
+            throw CairoException.nonCritical().setOutOfMemory(true)
                     .put("sun.misc.Unsafe.allocateMemory() OutOfMemoryError [RSS_MEM_USED=")
                     .put(getRssMemUsed())
-                    .put(", size=")
-                    .put(size)
+                    .put(", size=").putSize(size)
                     .put(", memoryTag=").put(memoryTag)
                     .put("], original message: ")
-                    .put(oom.getMessage());
-            System.err.println(e.getFlyweightMessage());
```
bluestreak01, line 389: this is regression, error message must be printed from here

bluestreak01 (changes requested):

core/src/main/java/io/questdb/std/Unsafe.java
Stacvirus: fixed, have added a put method with the error message

PR #6840, opened 2026-03-03
feat(parquet): extend row group pruning with range and null filter support

kafka1991:
## Summary

This PR extends #6739 (parquet row group pruning with min/max statistics and bloom filters) with additional filter pushdown support.

**#6739** introduced row group pruning for equality filters (`=`, `IN`) using bloom filters and min/max statistics.

**This PR** adds support for:
- Range comparisons: `<`, `<=`, `>`, `>=`, `BETWEEN`
- Null checks: `IS NULL`, `IS NOT NULL`

When executing queries with these filter predicates, the engine can now skip entire row groups that definitively don't contain matching rows, significantly improving query performance on large Parquet datasets.

### Supported Filter Operations

| Operation | Example | Pushdown Method |
|-----------|---------|-----------------|
| Equality | `col = value` | Bloom filter + Min/Max |
| In List | `col IN (v1, v2, ...)` | Bloom filter + Min/Max |
| Less Than | `col < value` | Min/Max statistics |
| Less Than or Equal | `col <= value` | Min/Max statistics |
| Greater Than | `col > value` | Min/Max statistics |
| Greater Than or Equal | `col >= value` | Min/Max statistics |
| Between | `col BETWEEN v1 AND v2` | Min/Max statistics |
| Is Null | `col IS NULL` | Null count statistics |
| Is Not Null | `col IS NOT NULL` | Null count statistics |

### Supported Column Types

BYTE, SHORT, INT, LONG, FLOAT, DOUBLE, DATE, TIMESTAMP, CHAR, IPv4, STRING, SYMBOL, VARCHAR, UUID, DECIMAL, LONG128

### Implementation Notes

- IPv4 uses unsigned comparison for correct ordering of high IP addresses (> 128.x.x.x)
- String/Symbol range comparisons (`<`, `>`, etc.) are disabled due to UTF-8 vs UTF-16 ordering differences (only `=` and `IN` are supported)
- Third-party Parquet files with unsigned integer types (UInt8/16/32/64) skip min/max range filtering, as QuestDB filter values are signed and cannot be correctly compared against unsigned statistics

## Dependencies

- Based on #6739


kafka1991:
fold to https://github.com/questdb/questdb/pull/6844

PR #6841, opened 2026-03-03
perf(wal): add cairo.wal.writer.madvise.mode configuration option

jerrinot:
WalWriter previously hardcoded `POSIX_MADV_RANDOM` for memory-mapped column files. This hurts most workloads, but some benefit from it. Make the madvise hint opt-in via a new config property with valid values: none (default, no hint), sequential and random.

RANDOM is beneficial when ingesting into many tables with many columns, as it prevents the OS from speculatively reading adjacent pages under memory pressure.

Example configuration:
```
cairo.wal.writer.madvise.mode=random
```

### DIFF
--- a/core/src/main/java/io/questdb/cairo/wal/WalWriter.java
+++ b/core/src/main/java/io/questdb/cairo/wal/WalWriter.java
@@ -116,6 +116,7 @@ public class WalWriter extends WalWriterBase implements TableWriterAPI {
 
     private final AlterOperation alterOp = new AlterOperation();
     private final ObjList<MemoryMA> columns;
+    private final int columnsMadviseMode;
     private final DdlListener ddlListener;
     private final AtomicIntList initialSymbolCounts;
     private final IntList localSymbolIds;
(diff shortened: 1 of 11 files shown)


core/src/main/java/io/questdb/cairo/wal/WalWriter.java
```diff
@@ -174,6 +175,7 @@ public WalWriter(
         super(configuration, tableToken, tableSequencerAPI, walDirectoryPolicy, walLocker);
 
         LOG.info().$("open [table=").$(tableToken).I$();
+        this.columnsMadviseMode = configuration.getWalWriterMadviseMode();
```
bluestreak01, line 178: WAL writers are cached, which makes this configuration non-reloadable. It would be a better approach to inject configuration where it need to be

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 28 / 28 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/ParquetTimestampFinder.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|15|15|100.00%|

Issue #6838, opened 2026-03-03
ALTER COLUMN doesn't support quotes


jmthomas:
### To reproduce

1. Create a new table: `CREATE TABLE TEST ("MY_COL" INT)`
2. Try to alter a column name by putting it in double quotes: `ALTER TABLE TEST ALTER COLUMN "MY_COL" FLOAT`

`column '"MY_COL"' does not exist in table 'TEST'`

3. Try to alter a column name by putting it in single quotes: `ALTER TABLE TEST ALTER COLUMN 'MY_COL' FLOAT`

`column ''MY_COL'' does not exist in table 'TEST'`

This prevents altering columns with unique names that might include special characters. Also it does not match the consistency of CREATE TABLE which can contain quotes.

### QuestDB version:

9.3.3

### OS, in case of Docker specify Docker and the Host OS:

Mac OS (Docker)

### File System, in case of Docker specify Host File System:

Host File System

### Full Name:

Jason Thomas

### Affiliation:

OpenC3

### Have you followed Linux, MacOs kernel configuration steps to increase Maximum open files and Maximum virtual memory areas limit?

- [x] Yes, I have

PR #6842, opened 2026-03-03
fix(sql): support quoted column names in `ALTER COLUMN`

nwoolmer:
Fixes https://github.com/questdb/questdb/issues/6838

## Summary

`ALTER TABLE ... ALTER COLUMN` and `ALTER MATERIALIZED VIEW ... ALTER COLUMN` failed when the column name was quoted (e.g. `"MY_COL"`). The token from the lexer was used as-is for the metadata lookup, so `"MY_COL"` (with quotes) was looked up instead of `MY_COL`, resulting in a "column does not exist" error.

## Changes

- Wrap `tok` with `unquote()` before the column name lookup in both `compileAlterTable` and `compileAlterMatView`
- Add test for quoted column names with `ALTER TABLE` and `ALTER MATERIALIZED VIEW`

### DIFF
--- a/core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java
+++ b/core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java
@@ -977,9 +977,10 @@ public class SqlCompilerImpl implements SqlCompiler, Closeable, SqlParserCallbac
                     throw SqlException.$(lexer.lastTokenPosition(), "'not' expected");
                 }
             } else {
-                int index = tableMetadata.getColumnIndexQuiet(unquote(tok));
+                CharSequence unquotedTok = unquote(tok);
+                int index = tableMetadata.getColumnIndexQuiet(unquotedTok);
                 if (index != -1) {
-                    throw SqlException.$(lexer.lastTokenPosition(), "column '").put(unquote(tok)).put("' already exists");
+                    throw SqlException.$(lexer.lastTokenPosition(), "column '").put(unquotedTok).put("' already exists");
                 }
             }
 
--- a/core/src/test/java/io/questdb/test/griffin/AlterTableAlterColumnTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/AlterTableAlterColumnTest.java
@@ -231,9 +231,35 @@ public class AlterTableAlterColumnTest extends AbstractCairoTest {
     public void testQuotedColumnNameCacheNocache() throws Exception {
         assertMemoryLeak(
                 () -> {
-                    execute("CREATE TABLE test_quoted (\"MY_COL\" SYMBOL, ts TIMESTAMP) TIMESTAMP (ts)");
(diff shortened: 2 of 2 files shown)


nwoolmer:
@coderabbitai review

core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java
glasstiger: Bug: Same missing-unquote pattern in `ADD COLUMN` duplicate detection

Lines 943 and 980 in `SqlCompilerImpl.java` have the same class of bug. The duplicate-column check uses the raw (potentially quoted) token:
```
  // Line 943 (ADD COLUMN IF NOT EXISTS path):
  final int columnIndex = tableMetadata.getColumnIndexQuiet(tok);  // tok not unquoted!

  // Line 980 (plain ADD COLUMN duplicate check):
  int index = tableMetadata.getColumnIndexQuiet(tok);  // tok not unquoted!
```

 While line 986 correctly calls `unquote(tok)` for the actual column creation, the preceding duplicate checks use the raw token.
This means:
  - `ALTER TABLE t ADD COLUMN "existing_col" INT` — fails to detect the duplicate, tries to add, gets a confusing error
  - `ALTER TABLE t ADD COLUMN IF NOT EXISTS "existing_col" INT` — fails to recognize the column exists, tries to add it anyway

This is pre-existing, not introduced by the PR, but it is the same pattern and should be flagged as a follow-up.

core/src/test/java/io/questdb/test/griffin/AlterTableAlterColumnTest.java
```diff
@@ -167,6 +167,19 @@ public void testExpectTableKeyword2() throws Exception {
         assertFailure("alter", 5, "'table' or 'materialized' or 'view' expected");
     }
 
+    @Test
+    public void testQuotedColumnName() throws Exception {
```
glasstiger: Test Coverage Weaknesses

  1. No state verification. The test only checks that statements do not throw. It never verifies the operations took effect. Compare with testAddIndexColumns (line 63) which opens a `TableReader` and verifies `reader.getBitmapIndexReader(...)`. The new test should at minimum assert `isColumnIndexed()` after `ADD INDEX`, and verify it is `false` after `DROP INDEX`.
  2. Missing `ALTER COLUMN TYPE` test. The linked issue (#6838) is specifically about `ALTER COLUMN "col" TYPE ...`, yet the test doesn't cover the `TYPE` sub-command at all. This is the most important case to test.
  3. No materialized view test. The PR fixes two code paths but only tests one. There should be a test for `ALTER MATERIALIZED VIEW ... ALTER COLUMN "quoted_col" ...`.
  4. No edge-case column names. Missing tests for:
    - Column names with spaces: "my col"
    - SQL keywords as column names: "select", "from"
    - Special characters: "col-1"

glasstiger:
@coderabbitai review

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|6|6|100.00%|

PR #6848, opened 2026-03-05
fix(sql): fix LIKE exact match falling back to table scan instead of using index

### LINKED ISSUE #6279: Exact match using `LIKE` instead of `=` prevents index scan compilation
### To reproduce

See https://github.com/questdb/questdb/issues/6246#issuecomment-3406805982

`LIKE 'exact match'` should be considered equivalent syntax to `= 'exact match'`. 

Index scans presumably detect `LIKE` and fast-fail, falling back to table scans.

### QuestDB version:

9.0.1+

N/A

N/A

### Full Name:

Nick Woolmer

### Affiliation:

QuestDB

- [x] Yes, I have

### Additional context

_No response_

TusharSariya:
Fixes #6279

## Summary

- `WhereClauseParser.extract()` now recognizes `LIKE` as a conditionally intrinsic operator
- When the LIKE pattern contains no wildcards (`%`, `_`), the predicate delegates to `analyzeEquals`, enabling symbol index scans

## Test plan

- `WhereClauseParserTest#testLikeWithConst` verifies exact LIKE pattern extracts into `keyValueFuncs`
- `WhereClauseParserTest#testLikeWithWildCards` verifies wildcard LIKE pattern stays as a filter
-  `testExactLikeOnIndexedSymbolUsesIndexScan` verify query plan for constant and wildcard

### DIFF
--- a/core/src/main/java/io/questdb/griffin/WhereClauseParser.java
+++ b/core/src/main/java/io/questdb/griffin/WhereClauseParser.java
@@ -2522,7 +2522,6 @@ public final class WhereClauseParser implements Mutable {
             boolean latestByMultiColumn,
             TableReader reader
     ) throws SqlException {
-        // here
         return switch (intrinsicOps.get(node.token)) {
             case INTRINSIC_OP_IN ->
                     analyzeIn(timestampDriver, translator, model, node, m, functionParser, executionContext, latestByMultiColumn, reader);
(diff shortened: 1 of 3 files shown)


TusharSariya:
@bluestreak01 could you please take a look when you get a chance? Thanks!

PR #6858, opened 2026-03-09
fix(sql): breaking :boom: fix SAMPLE BY with timezone producing incorrect bucket alignment

amunra:
## Breaking change

SAMPLE BY with `ALIGN TO CALENDAR TIME ZONE` changes how buckets behave during DST transitions for sub-day strides (hours, minutes, seconds):

**DST fall-back (clocks go back) now produces two rows instead of one for the repeated local hour.** Previously, both passes through the repeated hour were merged into a single bucket. Now they produce two distinct UTC-based buckets with the same local timestamp. For example, a 1h-sampled query over the Europe/Berlin October fall-back previously returned:

```
2021-10-31T02:00:00.000000Z    18
```

and now returns:

```
2021-10-31T02:00:00.000000Z    8
2021-10-31T02:00:00.000000Z    10
```

**Output timestamps can be non-monotonic in local time during fall-back.** The underlying UTC timestamps remain monotonic, but the local-time representation repeats during the fall-back hour (e.g., `02:00, 02:30, 02:00, 02:30, 03:00`). Any downstream consumer that assumes monotonic timestamps in local time will break.

**Sub-day bucket boundaries are now uniform in UTC, not in local time.** The old approach spaced buckets evenly in local time (so a 1h stride always covered exactly one local hour). The new approach uses the standard (non-DST) timezone offset, making buckets evenly spaced in UTC. During DST transitions, a local-time hour may contain more or fewer rows than expected, but bucket widths are consistent in UTC space.

These changes do not affect queries without `TIME ZONE`, queries with fixed-offset timezones (e.g., `'+05:30'`), or super-day strides (day, week, month, year).

## Summary

The SAMPLE BY ... ALIGN TO CALENDAR TIME ZONE rewrite introduces a new `timestamp_floor_utc` function that floors timestamps in local time and converts back to UTC internally. This replaces the previous approach of wrapping the query in an extra `to_utc()` conversion model, producing a simpler query plan with one fewer cursor layer.

### What changed

**New `timestamp_floor_utc` function.** The optimizer rewrites SAMPLE BY with timezone into a GROUP BY using `timestamp_floor_utc(stride, ts, from, offset, tz)` instead of wrapping `timestamp_floor(...)` in a `to_utc(...)` outer model. The function converts UTC→local, floors, and converts back to UTC in a single call. For sub-day strides, it uses the standard (non-DST) timezone offset to keep bucket widths uniform in UTC space, avoiding ambiguity during fall-back transitions.

**Extracted `AbstractTimestampFloorFromOffsetFunctionFactory`.** Shared floor logic between `timestamp_floor` (local-time return) and `timestamp_floor_utc` (UTC return) is extracted into a common abstract base class, with an `isReturnUtc()` flag controlling the conversion path.

**Fixed offset-in-floor-anchor bug.** The DST-aware floor path applied the user offset as a post-floor shift, while the fixed-offset path baked it into the floor anchor. This caused different bucket assignments when `offset % stride != 0`. The fix includes the offset in the floor anchor (`from + offset`) consistently across both paths.

**Fixed mat view refresh iterator for sub-day DST timezones with offset.** The refresh iterator's sub-day DST-aware branch dropped the user's ALIGN TO CALENDAR offset, causing scan boundaries to be misaligned with bucket keys. The fix passes `fixedOffset - stdOff` (matching the fixed-offset branch pattern).

**Fixed native memory leak in `FillRangeRecordCursorFactory`.** The `_close()` method did not free the cursor, leaking the `DirectLongList presentTimestamps` (native memory) when the factory was closed.

**Removed debug instrumentation.** Removed ~170 lines of `INSTRUMENT_LOG`-guarded dead code and the `debugRowCount` field from production classes.

### Tradeoffs

- The `timestamp_floor_utc` function performs a UTC round-trip (convert to local, floor, convert back) per input row, whereas the old approach called `to_utc()` once per output bucket. For queries with high input-to-output ratio (many rows per bucket), this adds per-row overhead. For sub-day strides the overhead is mitigated by using `getStandardOffset()` (a field read) instead of `getOffset()` (a binary search through transitions).
- Sub-day strides now use the standard (non-DST) offset for bucket computation. Bucket widths are uniform in UTC space rather than in local time. During DST transitions, a local-time hour may contain more or fewer rows than usual, but bucket boundaries remain evenly spaced in UTC.

## Test plan

- Verify `timestamp_floor_utc` matches `timestamp_floor` for non-timezone queries
- Verify DST-aware and fixed-offset paths produce identical results during non-DST periods (regression test for the anchor bug)
- Verify bucket monotonicity across DST spring-forward and fall-back transitions
- Verify mat view refresh produces correct results with sub-day DST timezone + offset
- Verify edge cases: offset=stride, offset>stride, negative offset, non-null from + offset + DST timezone
- Verify no native memory leaks via `assertMemoryLeak()` in all tests


puzpuzpuz:
It's an open question whether this PR fixes SAMPLE BY + TIME ZONE.

**Existing approach (master):** The SAMPLE BY + TIME ZONE rewrite uses `timestamp_floor` which floors in local time and returns local time. The optimiser then wraps the result in an extra query model with a `to_utc()` conversion to produce UTC output. This requires relocating ORDER BY to the `to_utc()` level since the conversion can change row ordering. The `to_utc()` call runs once per output bucket.

**New approach (this PR):** The rewrite uses a new `timestamp_floor_utc` function that floors in local time but converts back to UTC internally, eliminating the `to_utc()` wrapper model and the ORDER BY relocation logic. For sub-day strides, it uses the standard (non-DST) offset to keep bucket widths uniform in UTC space. FROM/TO bounds are converted to UTC upfront when a sub-day interval and timezone are both present, so the floor operates in pure UTC space without double-applying the offset.

The tradeoff is a simpler query plan at the cost of per-input-row UTC round-trip overhead (vs. per-output-bucket conversion in the old approach). For sub-day strides, the bucketing semantics also change: buckets have uniform UTC width rather than uniform local-time width. Per the [docs](https://questdb.com/docs/query/sql/sample-by/#time-zone), the documented behavior is calendar alignment in local time (`master` branch behavior).

What's the problem this PR aims to solve?

puzpuzpuz:
@amunra here is the list of bugfixes that could be extracted into a separate PR.

# Extractable bugfixes from PR #6858

Independent bugfixes that exist on `master` and can be extracted into a separate PR
while keeping the SAMPLE BY TIME ZONE semantic change in #6858.

## 1. Missing `break` in CAST TIMESTAMP switch

**File**: `SqlCodeGenerator.java` ~line 2883

The `TIMESTAMP` case in the cast-to-VARCHAR switch falls through into the `BINARY`
case, producing a wrong cast function.

```java
// master: no break after TIMESTAMP case
                        }
                        // missing break here
                    case BINARY:
                        castFunctions.add(BinColumn.newInstance(i));
                        break;
```

**Fix**: Add `break;` after the TIMESTAMP case body.

## 2. GEOBYTE-to-VARCHAR uses wrong column class

**File**: `SqlCodeGenerator.java` ~line 2954

`GeoShortColumn.newInstance(i, toTag)` should be `GeoByteColumn.newInstance(i, fromType)`.
Two bugs in one line: wrong class (`GeoShort` instead of `GeoByte`) and wrong type
parameter (`toTag` instead of `fromType`).

```java
// master (broken)
case GEOBYTE:
    castFunctions.add(
            CastGeoHashToGeoHashFunctionFactory.getGeoByteToVarcharCastFunction(
                    GeoShortColumn.newInstance(i, toTag),  // wrong class, wrong param
                    getGeoHashBits(fromType)
            )
    );

// fix
case GEOBYTE:
    castFunctions.add(
            CastGeoHashToGeoHashFunctionFactory.getGeoByteToVarcharCastFunction(
                    GeoByteColumn.newInstance(i, fromType),
                    getGeoHashBits(fromType)
            )
    );
```

## 3. Geohash-to-VARCHAR passes `toTag` instead of `fromType`

**File**: `SqlCodeGenerator.java` ~lines 2962, 2970, 2978

Three more geohash cast sites (`GEOSHORT`, `GEOINT`, `GEOLONG`) pass the destination
tag `toTag` to the column constructor instead of the source type `fromType`. The column
reads from the source, so it needs the source type to decode bits correctly.

```java
// master (broken) — same pattern in GEOSHORT, GEOINT, GEOLONG
GeoShortColumn.newInstance(i, toTag)   // should be fromType
GeoIntColumn.newInstance(i, toTag)     // should be fromType
GeoLongColumn.newInstance(i, toTag)    // should be fromType
```

## 4. Fill values validation always checks index 0

**File**: `SqlCodeGenerator.java` ~line 3083

The loop iterates but always validates the first fill expression. All subsequent fill
values skip validation entirely, allowing invalid expressions (like non-constant values)
through.

```java
// master (broken)
for (int i = 0, n = fillValuesExprs.size(); i < n; i++) {
    expr = fillValuesExprs.getQuick(0);  // BUG: should be getQuick(i)
    if (isNoneKeyword(expr.token)) {
```

**Fix**: Replace `getQuick(0)` with `getQuick(i)`.

## 5. ORDER BY position lookup uses wrong loop variable

**File**: `SqlCodeGenerator.java` ~line 5657

Inner loop iterates `j` but accesses `nodes` with outer variable `i`. This always
compares against the same node, missing actual matches on non-first ORDER BY columns.

```java
// master (broken)
for (int j = 0, y = nodes.size(); j < y; j++) {
    if (Chars.equals(column, nodes.getQuick(i).token)) {   // BUG: i should be j
        position = nodes.getQuick(i).position;              // BUG: i should be j
        break;
    }
}
```

**Fix**: Replace both `nodes.getQuick(i)` with `nodes.getQuick(j)`.

## 6. Filter function leaked on error paths

**File**: `SqlCodeGenerator.java` ~line 5166

Two leaks:

1. `filter` is not freed before throwing "boolean expression expected".
2. `filter` is not freed after evaluating a constant boolean value.

```java
// master (broken)
if (!isBoolean(filter.getType())) {
    // filter not freed
    throw SqlException.position(constFilterExpr.position).put("boolean expression expected");
}
filter.init(null, executionContext);
if (filter.isConstant()) {
    if (!filter.getBool(null)) {  // filter not freed after use
```

**Fix**: Call `Misc.free(filter)` before throwing and after evaluating the constant.

## 7. `moveClauses()` never advances position pointer

**File**: `SqlOptimiser.java` ~line 4618

`positions` is a sorted list of clause indices to move. Without `p++` after a match,
the same position matches forever, sending all subsequent clauses to `to` instead
of `result`.

```java
// master (broken)
JoinContext t = p < m && i == positions.getQuick(p) ? to : result;
// p is never incremented after a match

// fix
JoinContext t;
if (p < m && i == positions.getQuick(p)) {
    t = to;
    p++;
} else {
    t = result;
}
```

## 8. Window comparator uses `chainTypes` instead of `chainMetadata`

**File**: `SqlCodeGenerator.java` ~line 7780

All other comparator creation sites (e.g., `initRecordComparator`) use `chainMetadata`.
Using `chainTypes` (a bare `ArrayColumnTypes`) instead of the full
`GenericRecordMetadata` could produce wrong comparators if the two structures diverge
during column re-insertion.

```java
// master (broken)
windowComparators.add(recordComparatorCompiler.newInstance(chainTypes, e.key));

// fix
windowComparators.add(recordComparatorCompiler.newInstance(chainMetadata, e.key));
```

## 9. Incomplete tree traversal in `containsSelectNonAggregateFunction()` and `containsOrderedAggregateFunction()`

**File**: `SqlOptimiser.java` ~lines 294, 373

> **Note**: This one needs verification — the impact depends on whether non-group-by
> FUNCTION nodes ever have relevant children in practice.

On `master`, the `FUNCTION` case has a `break`. When a function node is NOT a
group-by/ordered-aggregate, execution breaks out of the switch and only traverses
`node.lhs`. The `default` case additionally pushes `node.args` and `node.rhs` onto the
stack. Removing the `break` (fall-through to `default`) makes non-matching FUNCTION
nodes traverse all children, which seems correct for a complete tree walk.

```java
// master (incomplete traversal)
case FUNCTION:
    if (functionFactoryCache.isGroupBy(node.token)) {
        return true;
    }
    break;  // skips args and rhs traversal for non-group-by functions
default:
    for (int i = 0, n = node.args.size(); i < n; i++) {
        sqlNodeStack.add(node.args.getQuick(i));
    }
    if (node.rhs != null) {
        sqlNodeStack.push(node.rhs);
    }
    break;

// fix: fall through to traverse all children
case FUNCTION:
    if (functionFactoryCache.isGroupBy(node.token)) {
        return true;
    }
    // fall through to traverse rhs and args
default:
```

Same pattern applies to `containsOrderedAggregateFunction()` at ~line 373.

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=223435)
**Tested Commit:** `f64a5bcb43cb1188b88186491c5d3098d6c113e3`

core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
```diff
@@ -897,6 +900,25 @@ private SampleByIntervalIterator intervalIterator(
             );
         }
 
+        // For sub-day intervals, timestamp_floor_utc uses standard offset for UTC↔local
```
puzpuzpuz: It's an open question whether this PR fixes SAMPLE BY + TIME ZONE.

See https://github.com/questdb/questdb/pull/6858#issuecomment-4085296912

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 632 / 653 (96.78%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/datetime/FixedTimeZoneRule.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/GroupByRecordCursorFactory.java|0|8|00.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|24|27|88.89%|
|:large_blue_circle:|io/questdb/griffin/SqlUtil.java|8|9|88.89%|
|:large_blue_circle:|io/questdb/griffin/SqlParser.java|33|35|94.29%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|45|47|95.74%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/AbstractTimestampFloorFromOffsetFunctionFactory.java|448|452|99.12%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/TimestampFloorFromFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/TimestampFloorFromOffsetFunctionFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/FillRangeRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|22|22|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/TimestampFloorFromOffsetUtcFunctionFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowCreateTableRecordCursorFactory.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableUtils.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncGroupByAtom.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/datetime/CommonUtils.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/std/datetime/AbstractTimeZoneRules.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CreateMatViewOperationImpl.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/TimestampFloorOffsetFunction.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/model/CreateTableColumnModel.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFillValueNotKeyedRecordCursor.java|2|2|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 549 / 560 (98.04%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/datetime/FixedTimeZoneRule.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/TimestampFloorFromFunctionFactory.java|2|4|50.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|24|27|88.89%|
|:large_blue_circle:|io/questdb/griffin/SqlUtil.java|8|9|88.89%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/AbstractTimestampFloorFromOffsetFunctionFactory.java|448|452|99.12%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/TimestampFloorFromOffsetFunctionFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/FillRangeRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|22|22|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/TimestampFloorFromOffsetUtcFunctionFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncGroupByAtom.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/datetime/CommonUtils.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/std/datetime/AbstractTimeZoneRules.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CreateMatViewOperationImpl.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/TimestampFloorOffsetFunction.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFillValueNotKeyedRecordCursor.java|2|2|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=225533)
**Tested Commit:** `90f1a7ffb9d181a0956ba1f7f8899f5de4512eb9`

PR #6859, opened 2026-03-09
feat(sql): add dynamic window support in WINDOW JOIN

puzpuzpuz:
WINDOW JOIN now accepts column references and expressions as RANGE BETWEEN boundaries, in addition to static constants. This allows each left-hand-side (LHS) row to define its own window size based on its data.

Example:
```sql
SELECT t.ts, sum(d.val) AS agg
FROM fx_trades t
WINDOW JOIN market_data d
RANGE BETWEEN t.lookback minutes PRECEDING AND t.lookahead minutes FOLLOWING;
```

Either or both of the lo/hi boundaries can be dynamic. The other can remain a static constant.

Boundary expressions must evaluate to an integer and must only reference LHS table columns. Negative values are clamped to zero (equivalent to CURRENT ROW). NULL values produce NULL aggregates — the row is skipped.

Each boundary, whether static or dynamic, can optionally include a time unit suffix (`seconds`, `minutes`, `hours`, etc.). When a time unit is present, the value is scaled to the LHS table's timestamp resolution at runtime (microseconds or nanoseconds). When the time unit is omitted, the raw integer value is interpreted in the LHS table's native timestamp resolution.

Dynamic windows disable the fast (symbol-keyed) and vectorized execution paths. Queries with an ON key equality clause fall back to the general path with a join filter instead.

Also, makes parallel HORIZON and WINDOW JOIN queries more responsive to query cancellation.

## Technical details

- `SqlOptimiser.validateWindowJoins` uses `tryEvalNonNegativeLongConstant` to distinguish static vs dynamic bounds. Column references and non-constant expressions fail to compile against `EmptyRecordMetadata` and are classified as dynamic. `resolveWindowJoinBoundColumns` walks the expression tree to strip LHS table prefixes and reject right-hand-side (RHS) table column references.
- `SqlCodeGenerator` compiles dynamic bound functions against LHS metadata, computes sign/timeUnit parameters, and creates per-worker function copies for thread-unsafe expressions via `compileWorkerFunctionsConditionally`.
- `AsyncWindowJoinRecordCursorFactory` implements 6 new reducer methods for dynamic windows using a two-pass algorithm: the first pass computes overall RHS bounds across all LHS rows in a page frame, the second pass does per-row binary search within the pre-scanned data. `computeEffectiveBound` evaluates the dynamic function per LHS row, handles NULL and negative clamping.
- `WindowJoinRecordCursorFactory` (single-threaded path) integrates `computeEffectiveBound` for per-row bound computation.
- `WindowJoinTimeFrameHelper` supports non-monotonic access patterns from dynamic windows via bookmark-based navigation.
- Each async WINDOW JOIN and HORIZON JOIN reduce method receives a `circuitBreaker` parameter but only uses it for slot acquisition, never inside the per-row processing loops. When a single page frame contains many rows, the worker thread runs uninterruptibly until the entire frame is processed, making the query unresponsive to cancellation. The fix adds `circuitBreaker.statefulThrowExceptionIfTripped()` at the top of every master-row iteration loop in all four async factories, so that cancellation, timeout, and broken connection signals are detected promptly during heavy frame processing.

## Test plan

- Deterministic tests for dynamic lo, dynamic hi, both dynamic, mixed static+dynamic, expression-based bounds, NULL bounds, negative bounds, non-monotonic bounds, bound column not in SELECT, error cases (non-integer bound, RHS column reference), fast path fallback
- All deterministic tests cover both async and single-threaded (via LIMIT) paths, with and without INCLUDE PREVAILING
- Parallel fuzz tests with thread-unsafe (varchar-to-long cast) bound expressions to exercise per-worker function copies
- Randomized fuzz test independently randomizes dynamic lo and hi bounds across various combinations of filters, symbol equality, join filters, and prevailing
- Unit tests for `WindowJoinTimeFrameHelper` bookmark logic

### DIFF
--- a/core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
+++ b/core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
@@ -1485,9 +1485,14 @@ public class SqlCodeGenerator implements Mutable, Closeable {
         if (func != null && !func.isThreadSafe() && sharedQueryWorkerCount > 0) {
             assert funcExpr != null;
             ObjList<Function> workerFunctions = new ObjList<>();
-            for (int i = 0; i < sharedQueryWorkerCount; i++) {
-                Function workerFunc = functionParser.parseFunction(funcExpr, metadata, executionContext);
-                workerFunctions.extendAndSet(i, workerFunc);
+            try {
+                for (int i = 0; i < sharedQueryWorkerCount; i++) {
+                    Function workerFunc = functionParser.parseFunction(funcExpr, metadata, executionContext);
+                    workerFunctions.extendAndSet(i, workerFunc);
+                }
+            } catch (Throwable th) {
+                Misc.freeObjList(workerFunctions);
+                throw th;
             }
             return workerFunctions;
         }


glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 1055 / 1441 (73.21%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinRecordCursorFactory.java|687|1045|65.74%|
|:large_blue_circle:|io/questdb/griffin/engine/join/WindowJoinRecordCursorFactory.java|28|31|90.32%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinFastRecordCursorFactory.java|63|70|90.00%|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|128|140|91.43%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|55|58|94.83%|
|:large_blue_circle:|io/questdb/griffin/engine/join/WindowJoinTimeFrameHelper.java|25|26|96.15%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinAtom.java|57|59|96.61%|
|:large_blue_circle:|io/questdb/griffin/model/WindowJoinContext.java|8|8|100.00%|

PR #6867, opened 2026-03-11
perf(sql): optimize parallel HORIZON JOIN for various data distributions

puzpuzpuz:
## Summary

- HORIZON JOIN's keyed ASOF lookup now starts in backward-only mode (cheap for low-cardinality key spaces) and adaptively switches to forward scan mode within a frame when backward scan cost
 becomes excessive — e.g., with high-cardinality symbols or rare/infrequent keys that cause deep backward scans.
- The switch uses two criteria: a relative threshold (SWITCH_FACTOR=8, MIN_GAP=1,024) where backward scan cost at a position must exceed 8x the gap to trigger, and an absolute threshold (BWD_SCAN_ABSOLUTE_THRESHOLD=131,072) to handle cross-partition boundaries where the relative check cannot trigger. Once switched, the frame stays in forward mode for its remainder.
- Re-enables the ASOF row cache in `HorizonJoinTimeFrameHelper` and resets bookmarks on `toTop()` to eliminate jitter from out-of-order frame processing.

## Background

The previous backward-only strategy (commit 6145bafc) clears the key map on each ASOF position change and backward-scans ~K rows to repopulate it. This is 30x faster than forward scan for K=50 FX symbols, but degrades badly when K is large (e.g., K=5,000 equity tickers) or when rare keys appear infrequently — each rare-key lookup scans thousands of rows backward to the beginning of the table.

The adaptive approach retains the backward-only advantage for the common low-K case while automatically switching to forward scan when the data distribution makes backward scans expensive.

## How it works

Each frame starts in backward-only mode. On each ASOF position change, the algorithm compares the backward scan row count at the previous position against the gap (row distance) to the current position. If `bwdScanCost > gap * 8` and `gap > 1,024`, it switches to forward scan mode permanently for that frame.

Gap calculation uses raw rowId subtraction (`asOfRowId - prevAsOfRowId`). For same-frame positions, this gives the exact row count. For cross-frame positions, the frame-index bits produce a value >= 2^44, which safely prevents the relative switch condition from triggering. An absolute threshold (`bwdScanCost > 131,072`) handles the cross-partition case — e.g., a deep backward scan for a rare key right before a partition boundary.

## Benchmark results

Benchmarked with 100M-row `market_data` and 50K-row `fx_trades` tables across five data distribution scenarios (see [this gist](https://gist.github.com/puzpuzpuz/e5989707b264ca6b1d620907aa8f2248)).

Test query:
```sql
SELECT
    t.symbol,
    t.side,
    h.offset / 1000000000 AS horizon_sec,
    count() AS n,
    avg(
        CASE t.side
            WHEN 'buy'  THEN ((m.best_bid + m.best_ask) / 2 - t.price)
                             / t.price * 10000
            WHEN 'sell' THEN (t.price - (m.best_bid + m.best_ask) / 2)
                             / t.price * 10000
        END
    ) AS avg_markout_bps,
    sum(
        CASE t.side
            WHEN 'buy'  THEN ((m.best_bid + m.best_ask) / 2 - t.price)
                             * t.quantity
            WHEN 'sell' THEN (t.price - (m.best_bid + m.best_ask) / 2)
                             * t.quantity
        END
    ) AS total_pnl
FROM fx_trades_xyz t
HORIZON JOIN market_data_xyz m ON (symbol)
    RANGE FROM 0s TO 30s STEP 5s AS h;
```

### Dev box results

Environment: GraalVM CE 17, Ryzen 7900x, 64GB RAM

| Scenario | Patch | Master | Change |
|----------|-------|--------|--------|
| dense (K=10, gap~10) | 175ms | 231ms | -24% |
| sparse (K=15, gap~2000) | 13ms | 406ms | -97% |
| rare (K=200, Zipf 2.0) | 232ms | 448ms | -48% |
| equity (K=5000) | 1.19s | 1.4s | -15% |
| injected (10 common + 5 rare at start) | 171s | 156s | +10% |

The `_injected` scenario is a synthetic worst case where rare symbols appear only once at the very beginning of `market_data`. Both master and patch handle it poorly (156s and 171s respectively). The patch regresses by 10% because backward-only mode must complete at least one deep scan before the adaptive switch triggers, whereas master's always-forward approach never clears the key map.

### c6a.16xlarge EC2 instance results

| Scenario | Master | Patch | Change |
|----------|--------|-------|--------|
| dense | 422.8 ms | 331.8 ms | -21.5% |
| sparse | 514.2 ms | 19.4 ms | -96.2% |
| rare | 629.3 ms | 509.9 ms | -19.0% |
| equity | 1293.5 ms | 1120.0 ms | -13.4% |
| injected | 121023.7 ms | 122681.8 ms | +1.4% |

The patch shows clear wins on dense (-21%), sparse (-96%, dramatic), rare (-19%), and equity (-13%). The injected scenario (rare symbols planted at the very start forcing full backward scans) is essentially unchanged (+1.4%, within noise).

The sparse result stands out — the adaptive scan switch appears to be extremely effective when the gap between LHS rows is large (~2000 market_data rows per trade) with few symbols.

Also, no huge jitter noticed: the worst jitter was ~26% and was in the equity scenario. In other scenarios it was within 3-10%.

## Test plan

- Existing `HorizonJoinTest` (113 tests) and `HorizonJoinFuzzTest` pass
- New `testHorizonJoinKeyedAdaptiveScanSwitch` in `HorizonJoinTest`: deterministic test with a rare key at row 0 that triggers the backward-to-forward switch
- New `testParallelHorizonJoinAdaptiveScan` and `testParallelHorizonJoinAdaptiveScan2` in `ParallelHorizonJoinFuzzTest`: parallel tests with microsecond-spaced data in one partition, comparing HORIZON JOIN results against reference UNION ALL of ASOF JOINs
- Manual benchmarking with `test_distributions.sql` scenarios (dense, sparse, rare, equity, injected) to verify switching behavior across different cardinalities and distributions

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/groupby/SampleByFillNoneRecordCursorFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/groupby/SampleByFillNoneRecordCursorFactory.java
@@ -71,11 +71,12 @@ public class SampleByFillNoneRecordCursorFactory extends AbstractSampleByRecordC
             int sampleToFuncPos
     ) {
         super(base, groupByMetadata, recordFunctions);
+        Map map = null;
         try {
             // sink will be storing record columns to map key
             final RecordSink mapSink = RecordSinkFactory.getInstance(configuration, asm, base.getMetadata(), listColumnFilter);
             // this is the map itself, which we must not forget to free when factory closes
-            final Map map = MapFactory.createOrderedMap(configuration, keyTypes, valueTypes);
+            map = MapFactory.createOrderedMap(configuration, keyTypes, valueTypes);
             final GroupByFunctionsUpdater groupByFunctionsUpdater = GroupByFunctionsUpdaterFactory.getInstance(asm, groupByFunctions);
             cursor = new SampleByFillNoneRecordCursor(
                     configuration,
@@ -97,6 +98,7 @@ public class SampleByFillNoneRecordCursorFactory extends AbstractSampleByRecordC
                     sampleToFuncPos
             );
         } catch (Throwable th) {
+            Misc.free(map);
             close();
             throw th;
         }
--- a/core/src/main/java/io/questdb/griffin/engine/join/SpliceJoinLightRecordCursorFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/join/SpliceJoinLightRecordCursorFactory.java
@@ -79,10 +79,11 @@ public class SpliceJoinLightRecordCursorFactory extends AbstractJoinRecordCursor
             JoinContext joinContext
     ) {
         super(metadata, joinContext, masterFactory, slaveFactory);
+        Map joinKeyMap = null;
         try {
             this.masterKeySink = masterSink;
             this.slaveKeySink = slaveSink;
-            Map joinKeyMap = MapFactory.createUnorderedMap(
+            joinKeyMap = MapFactory.createUnorderedMap(
                     cairoConfiguration,
                     joinColumnTypes,
                     valueTypes
@@ -98,6 +99,7 @@ public class SpliceJoinLightRecordCursorFactory extends AbstractJoinRecordCursor
                     NullRecordFactory.getInstance(slaveFactory.getMetadata())
             );
         } catch (Throwable th) {
+            Misc.free(joinKeyMap);
             close();
             throw th;
         }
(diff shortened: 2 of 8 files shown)


glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 85 / 89 (95.51%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/DefaultCairoConfiguration.java|0|3|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/HorizonJoinTimeFrameHelper.java|23|24|95.83%|
|:large_blue_circle:|io/questdb/PropertyKey.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncHorizonJoinRecordCursorFactory.java|22|22|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/BaseAsyncHorizonJoinAtom.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/HorizonJoinRecordCursorFactory.java|22|22|100.00%|

PR #6871, opened 2026-03-12
fix(sql): WINDOW JOIN dropping prevailing row on cross-partition

RaphDal:
## Summary

- `WindowJoinTimeFrameHelper.findRowLo()` drops the INCLUDE PREVAILING row when the bookmarked-frame optimization kicks in and the prevailing candidate sits in a prior partition. The bookmarked path skipped initialization of `prevailingFrameIndex`/`prevailingRowIndex`, leaving them at `-1`/`Long.MIN_VALUE`. Downstream backward scans (`findPrevailingForMasterRow`, `WindowJoinPrevailingCache`) saw `-1` and returned immediately, silently omitting the prevailing row from results.
- `WindowJoinWithPrevailingFastRecordCursor.hasNext()` compared `slaveTimestamps.get(rowLo)` against `masterTimestampHi` instead of `slaveTimestampLo` when deciding whether to include the prevailing value. This caused it to skip prevailing inclusion when the first in-index row fell between `slaveTimestampLo` and `masterTimestampHi`.

## Detail

The first bug affects all four sync cursor variants that call `findRowLo(lo, hi, true)`:

- `WindowJoinWithPrevailingAndJoinFilterRecordCursor`
- `WindowJoinWithPrevailingAndJoinFilterFastRecordCursor`
- `WindowJoinWithPrevailingFastRecordCursor`
- `WindowJoinFastVectRecordCursor` (when `includePrevailing=true`)

The base `WindowJoinRecordCursor` (prevailing without filter) is not affected because it uses the separate `findRowLoWithPrevailing()` method, which maintains its own local prevailing tracking and handles the bookmarked path correctly.

The fix seeds `prevailingFrameIndex`/`prevailingRowIndex` when entering via the bookmark path. If the bookmark is past the frame start, the row immediately before the bookmark becomes the prevailing candidate. If the bookmark is at the frame start, the prevailing index is set to `rowLo - 1`, which signals `findPrevailingForMasterRow` to skip to the previous frame.

The second bug is a simple comparison target error in `WindowJoinWithPrevailingFastRecordCursor`: the condition that triggers prevailing inclusion should check whether the first row's timestamp exceeds the *window low boundary* (`slaveTimestampLo`), not the *window high boundary* (`masterTimestampHi`).

## Test plan

- Add `testPrevailingWithFilterCrossPartition` regression test to `WindowJoinTest`: master rows on day 2, slave rows on day 1 and day 2, join filter on `id` column, 1-second window with INCLUDE PREVAILING. Verifies the cross-partition prevailing row is counted. Fails without the fix, passes with it.


core/src/test/java/io/questdb/test/griffin/WindowJoinTest.java
```diff
@@ -4935,4 +4935,73 @@ private void prepareTable() throws SqlException {
         }
         drainWalQueue();
     }
+
+    @Test
+    public void testPrevailingWithFilterCrossPartition() throws Exception {
```
puzpuzpuz, line 4940: Could you add a similar test with symbol-based join key, so that the fix is verified for the fast cursor variant?

core/src/main/java/io/questdb/griffin/engine/join/WindowJoinTimeFrameHelper.java
```diff
@@ -124,6 +124,22 @@ public long findRowLo(long timestampLo, long timestampHi, boolean recordPrevaili
...
+                // closer candidate within the frame.
+                if (bookmarkedRowIndex > timeFrame.getRowLo()) {
+                    prevailingFrameIndex = timeFrame.getFrameIndex();
+                    prevailingRowIndex = bookmarkedRowIndex - 1;
+                } else {
+                    // Bookmark is at frame start; prevailing is in a prior frame.
+                    // Setting rowIndex < rowLo causes findPrevailingForMasterRow
+                    // to skip to previousFrame().
+                    prevailingFrameIndex = timeFrame.getFrameIndex();
+                    prevailingRowIndex = timeFrame.getRowLo() - 1;
```
puzpuzpuz: `timeFrame.getRowLo()` is always `0`, so in this branch we always end up with `prevailingRowIndex = -1`. The problem here is that `WindowJoinPrevailingCache` does not handle `prevailingFrameIndex != -1, prevailingRowIndex = -1` case properly. Could you update `WindowJoinPrevailingCache.findPrevailingSlaveRowId()` to handle the `rowIndex < rowLo` case by skipping to `previousFrame()` (matching what `findPrevailingForMasterRow` already does), or change the signaling in `findRowLo()` so the cache can distinguish "uninitialized" (`Long.MIN_VALUE`) from "scan previous frame" (-1)?

core/src/test/java/io/questdb/test/griffin/WindowJoinTest.java
RaphDal: Indeed, the same test adapted for the `Fast` variant fails. Thanks!

core/src/main/java/io/questdb/griffin/engine/join/WindowJoinTimeFrameHelper.java
RaphDal: I must be missing something but `findPrevailingForMasterRow` doesn't handle a `rowIndex` (which I believe to be `scanStart` in this case) like `findPrevailingSlaveRowId` does (returns `Long.MIN_VALUE` when it's `< 0`).

I went with updating `findPrevailingSlaveRowId` to only return `Long.MIN_VALUE` when `rowIndex` is `Long.MIN_VALUE` (uninitialized) instead of `< 0`, does that make sense?
When stepping through the execution, there is already a look in `findPrevailingSlaveRowId` to automatically go to the previous frame if there are no matching rows.

Fix in ea24cd4d764eb58297ea1c1b03253fb769964541.

core/src/test/java/io/questdb/test/griffin/WindowJoinTest.java
RaphDal: Added the test in ea24cd4d764eb58297ea1c1b03253fb769964541.

glasstiger:
#### [PR Coverage check]

PR #6874, opened 2026-03-14
fix(sql): fix bugs, resource leaks and NPE in SQL engine

bluestreak01:
### Correctness fixes

- **DECIMAL fall-through to BINARY in UNION casts**: `generateCastFunctions` lacked a `break` after the DECIMAL inner switch, causing fall-through to `case BINARY:` that added a spurious `BinColumn` to the cast function list. In multi-column UNION queries, this shifted subsequent column indices, causing `UnsupportedOperationException` at runtime.

- **`moveClauses` never incrementing position counter**: when `swapJoinOrder0` needed to steal multiple join clauses, only the first was moved. The rest stayed in the wrong join context, creating circular dependencies that broke topological sort and caused query compilation failures on 3+ table cross joins with multi-column WHERE conditions.

- **`hasGroupByFunc` / `hasOrderedGroupByFunc` skipping children of non-aggregate functions**: the `FUNCTION` case did `break` instead of falling through to the `default` traversal logic, so nested aggregates like `abs(sum(x))` were invisible. This caused PIVOT to reject valid aggregate expressions with "expected aggregate function".

### Resource leak fixes

- **Const filter leaked in `generateJoins`**: the `filter` function leaked when it was non-boolean (threw without closing), constant true (abandoned), or constant false (returned without closing).

- **`generateSampleBy` function leaks**: `timezoneNameFunc`, `offsetFunc`, `sampleFromFunc`, and `sampleToFunc` were created outside the outer try block and never freed on error paths.

### Crash fix

- **NPE in `CREATE MATERIALIZED VIEW`**: when `WITH BASE` specifies a nonexistent table not referenced in the query, `getTableTokenIfExists()` returned null and `.isView()` threw `NullPointerException` instead of a descriptive `SqlException`.

### Latent bug fixes (currently unreachable, fixed for correctness)

- **`generateFill` always reading first fill value**: `getQuick(0)` instead of `getQuick(i)` in a loop. Currently gated by the optimizer restricting multi-value FILL to a different code path.

- **GeoHash-to-VARCHAR cast using wrong column type and parameter**: all four GEO→VARCHAR casts used `toTag` instead of `fromType`. Currently unreachable because UNION type resolution never produces VARCHAR as target for GEO sources.

### Known bug documented (fix deferred)

- **`subQueryMode` nesting corruption**: nested `parseAsSubQuery` calls reset the boolean flag prematurely, causing valid 3-level nested queries to be rejected. Both counter-based and save/restore fixes break 39 other parser tests — a proper fix requires a deeper refactor. Test added to document the issue.

## Test plan

- `UnionAllCastTest#testIntDecimalMultiColumn` — multi-column INT→DECIMAL UNION ALL; fails with `UnsupportedOperationException` before fix
- `SqlOptimiserTest#testMoveClausesIncrementsPositionCounter` — 3-table CROSS JOIN with multi-column WHERE; fails to compile before fix
- `PivotTest#testPivotWithWrappedAggregate` — `abs(SUM(x))` in PIVOT; rejected before fix
- `PivotTest#testPivotWithCoalesceWrappedAggregate` — `coalesce(0, SUM(x))` in PIVOT; rejected before fix
- `PivotTest#testPivotWithCoalesceVarArgsWrappedAggregate` — `coalesce(NULL, NULL, SUM(x))` in PIVOT; rejected before fix
- `CreateMatViewTest#testCreateMatViewBaseTableNoReferenceAndDoesNotExist` — NPE before fix
- `SqlParserTest#testSubQueryModeNestedCorruption` — documents known bug
- `SampleByTest` full suite (283 tests) — verifies generateSampleBy leak fix has no regressions

### DIFF
--- a/core/src/main/java/io/questdb/std/DirectLongList.java
+++ b/core/src/main/java/io/questdb/std/DirectLongList.java
@@ -223,9 +223,6 @@ public class DirectLongList implements Mutable, Closeable, Reopenable {
     // desired capacity in bytes (not count of LONG values)
     private void setCapacityBytes(long capacity) {
         if (this.capacity != capacity) {
-//            if ((capacity >>> 3) > MAX_SAFE_INT_POW_2) {
-//                throw CairoException.nonCritical().put("long list capacity overflow");
-//            }
             final long oldCapacity = this.capacity;
             final long oldSize = this.pos - this.address;
             final long address = Unsafe.realloc(this.address, oldCapacity, capacity, memoryTag);
--- a/core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
+++ b/core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
@@ -5853,105 +5853,105 @@ public class SqlCodeGenerator implements Mutable, Closeable {
         RecordCursorFactory factory = null;
         try {
 
-        if (timezoneName != null) {
-            timezoneNameFunc = functionParser.parseFunction(
-                    timezoneName,
-                    EmptyRecordMetadata.INSTANCE,
-                    executionContext
-            );
-            timezoneNameFuncPos = timezoneName.position;
-            coerceRuntimeConstantType(timezoneNameFunc, STRING, executionContext, "timezone must be a constant expression of STRING or CHAR type", timezoneNameFuncPos);
-        } else {
-            timezoneNameFunc = StrConstant.NULL;
-            timezoneNameFuncPos = 0;
-        }
-
-        if (offset != null) {
-            offsetFunc = functionParser.parseFunction(
-                    offset,
-                    EmptyRecordMetadata.INSTANCE,
-                    executionContext
-            );
-            offsetFuncPos = offset.position;
-            coerceRuntimeConstantType(offsetFunc, STRING, executionContext, "offset must be a constant expression of STRING or CHAR type", offsetFuncPos);
-        } else {
-            offsetFunc = StrConstant.NULL;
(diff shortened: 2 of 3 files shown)


core/src/main/java/io/questdb/std/DirectLongList.java
```diff
@@ -223,9 +223,6 @@ public void zero() {
     // desired capacity in bytes (not count of LONG values)
     private void setCapacityBytes(long capacity) {
         if (this.capacity != capacity) {
-//            if ((capacity >>> 3) > MAX_SAFE_INT_POW_2) {
-//                throw CairoException.nonCritical().put("long list capacity overflow");
-//            }
```
RaphDal, line 228: Why did you comment this check?

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 75 / 80 (93.75%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|69|74|93.24%|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|4|4|100.00%|

core/src/main/java/io/questdb/std/DirectLongList.java
bluestreak01: this change removed commented out code (that exists on master)

ctkqiang:
Well done

onlybugs05:
@bluestreak01 Hey Bro Are You eligible for Swag

PR #6884, opened 2026-03-18
fix(sql): fix COPY TO failure when subquery uses bind variables

nwoolmer:
## Summary
- `CopyExportFactory.getCursor()` enqueues export work onto a ring queue that a worker thread picks up. `BaseParquetExporter.of()` constructed the worker's `SqlExecutionContext` with a null `BindVariableService`, so any bind variables in the COPY subquery (e.g. `COPY (SELECT * FROM t WHERE ts <= $1) TO ...`) caused `"bind variable service is not provided"` when the worker recompiled the SQL (TEMP_TABLE path) or opened a cursor (streaming paths).
- Thread the caller's `BindVariableService` through `CopyExportRequestTask` so the worker's execution context receives it.
- Deep-copy (snapshot) bind variable values via `BindVariableServiceImpl.snapshot()` before enqueueing the task, since the PG Wire thread may clear or repopulate the `BindVariableService` for the next query before the worker starts. The HTTP export path keeps the direct reference since it runs synchronously on the same thread.

## Test plan
- New test `testCopyQueryWithBindVariable` in `CopyExportTest` verifies COPY TO with a timestamp bind variable and `PARTITION_BY DAY` (the TEMP_TABLE code path that triggers recompilation)
- All 88 existing `CopyExportTest` tests pass
- All 114 `BindVariableServiceImplTest` tests pass

### DIFF
--- a/core/src/main/java/io/questdb/cutlass/http/processors/ExportQueryProcessor.java
+++ b/core/src/main/java/io/questdb/cutlass/http/processors/ExportQueryProcessor.java
@@ -1176,7 +1176,7 @@ public class ExportQueryProcessor implements HttpRequestProcessor, HttpRequestHa
             state.columnValueFullySent = true;
         } catch (Throwable e) {
             // we have to disambiguate here if this is the first attempt to send the value, which failed,
-            // and we have any partial value we can send to the clint, or our state did not bookmark anything?
+            // and we have any partial value we can send to the client, or our state did not bookmark anything?
             state.columnValueFullySent = state.arrayState.isNothingWritten();
             state.arrayState.reset(arrayView);
             throw e;
--- a/core/src/main/java/io/questdb/cutlass/parquet/CopyExportRequestTask.java
+++ b/core/src/main/java/io/questdb/cutlass/parquet/CopyExportRequestTask.java
@@ -750,22 +750,14 @@ public class CopyExportRequestTask implements Mutable, QuietCloseable {
         }
 
         private static int getRequiredAlignmentForSimd(int columnType) {
-            switch (ColumnType.tagOf(columnType)) {
+            return switch (ColumnType.tagOf(columnType)) {
                 // Types using Simd<i64, 8> or Simd<f64, 8>
-                case ColumnType.LONG:
-                case ColumnType.DOUBLE:
-                case ColumnType.TIMESTAMP:
-                case ColumnType.DATE:
-                    return 8;
+                case ColumnType.LONG, ColumnType.DOUBLE, ColumnType.TIMESTAMP, ColumnType.DATE -> 8;
                 // Types using Simd<i32, 16> or Simd<f32, 16>
-                case ColumnType.INT:
-                case ColumnType.FLOAT:
-                case ColumnType.SYMBOL:
-                    return 4;
+                case ColumnType.INT, ColumnType.FLOAT, ColumnType.SYMBOL -> 4;
                 // All other types use scalar paths - no SIMD alignment required
-                default:
-                    return 1;
-            }
+                default -> 1;
+            };
         }
 
         /**
--- a/core/src/main/java/io/questdb/griffin/engine/functions/bind/BindVariableServiceImpl.java
+++ b/core/src/main/java/io/questdb/griffin/engine/functions/bind/BindVariableServiceImpl.java
@@ -206,10 +206,8 @@ public class BindVariableServiceImpl implements BindVariableService {
                 f.getDecimal256(null, dec);
                 copy.setDecimal(index, dec.getHh(), dec.getHl(), dec.getLh(), dec.getLl(), type);
             }
-            default -> {
-                // UNDEFINED, ARRAY, or unknown — define with type only (no value)
-                copy.define(index, type, 0);
-            }
+            default -> // UNDEFINED, ARRAY, or unknown — define with type only (no value)
+                    copy.define(index, type, 0);
         }
         return dec;
     }
--- a/core/src/test/java/io/questdb/test/griffin/engine/functions/bind/BindVariableServiceImplTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/engine/functions/bind/BindVariableServiceImplTest.java
@@ -260,7 +260,7 @@ public class BindVariableServiceImplTest {
 
             bindVariableService.setFloat(0, Float.NaN);
             final double d = bindVariableService.getFunction(0).getDouble(null);
-            Assert.assertTrue(d != d);
+            Assert.assertTrue(Double.isNaN(d));
         });
     }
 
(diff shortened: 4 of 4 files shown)


nwoolmer:
## Self-review

### Moderate

**1. Missing test for streaming export paths with bind variables**

The test only covers the TEMP_TABLE path (triggered by `PARTITION_BY DAY`). The streaming paths (DIRECT_PAGE_FRAME, PAGE_FRAME_BACKED, CURSOR_BASED) are reached when `PARTITION_BY` is omitted, and they also depend on the BindVariableService fix. A second test variant without `PARTITION_BY` would exercise those paths.

**2. No negative assertion that excluded rows are absent**

The test verifies the content of `2024-01-01.parquet` and `2024-01-02.parquet`, but does not verify that `2024-01-03.parquet` does **not** exist. The `num_exported_files=2` check is indirect evidence, but an explicit `assertFalse(exportFileExists(...))` would be stronger.

### Notes

- The HTTP export path (`ExportQueryProcessor`) passes the `BindVariableService` by direct reference (no snapshot) because it runs the export synchronously on the same IO thread — no cross-thread sharing occurs.
- `BindVariableServiceImpl.snapshot()` handles all scalar types, String, Varchar, Long256, UUID, Binary, GeoHash, and Decimal. For UNDEFINED/ARRAY types it falls back to `define()` only (no value copy) — these are extremely unlikely in COPY WHERE clauses.

bluestreak01:
@nwoolmer Code review findings:

## Critical

**1. `snapshot()` performs a shallow copy of BINARY bind variables — cross-thread data race**

`BindVariableServiceImpl.java:141` — `copy.setBin(i, f.getBin(null))` stores the `BinarySequence` reference directly into `BinBindVariable.value` (see `BinBindVariable.java:35`). Unlike STRING (which deep-copies into a `StringSink`) and VARCHAR (which deep-copies into a `Utf8StringSink`), BINARY stores a bare reference.

In the PG Wire path, the caller's thread may clear or reuse the original `BinarySequence` buffer after the snapshot is enqueued on the ring queue. The worker thread would then read stale or corrupted binary data. The Javadoc claims "an independent deep copy" but BINARY violates that contract.

**Trace:** `CopyExportFactory.getCursor()` (line 257) → `snapshot()` (line 141) → `setBin()` (line 339: `new BinBindVariable(value)`) → `BinBindVariable.value` is a bare reference (line 35).

**Suggested fix:** Deep-copy the binary data into a byte array or QuestDB-owned buffer. Alternatively, document this as a known limitation with a code comment if BINARY bind variables in COPY subqueries are considered unsupported.

## Moderate

**2. `snapshot()` does not copy named bind variables**

`BindVariableServiceImpl.java:115-155` — The method iterates only over `source.getIndexedVariableCount()` and ignores `source.getNamedVariables()`. If a COPY subquery ever uses named bind variables (`:name` syntax), they would be silently lost.

**Mitigating factor:** PG Wire uses only indexed variables (`$1`, `$2`), and the HTTP path passes the reference directly (no snapshot). So this is currently unreachable. However, the Javadoc claims "independent deep copy" without qualifying this limitation.

**Suggested fix:** Either copy named variables too, or update the Javadoc to state "copies indexed variables only."

**3. No unit tests for `snapshot()` method**

`BindVariableServiceImplTest.java` has no tests for `snapshot()`. Unit tests should verify:
- Independence (modifying source after snapshot doesn't affect the copy)
- All supported scalar types round-trip correctly
- `snapshot(null, config)` returns null
- Edge case: empty source

**4. Only TIMESTAMP bind variable type tested in integration tests**

Both `testCopyQueryWithBindVariable` and `testCopyQueryWithBindVariableStreaming` use only `setTimestamp(0, ...)`. At minimum, a test with a STRING or INT bind variable (e.g., `WHERE x = $1`) would increase confidence that `snapshot()` handles other types correctly.

## Minor

**5. `Decimal256` allocated inside the switch could be hoisted**

`BindVariableServiceImpl.java:145` — `new Decimal256()` is created for each DECIMAL bind variable inside the loop. Could be allocated once before the loop and reused. Cold path so impact is negligible.

**6. PR title is slightly implementation-focused**

The title uses the internal error message ("bind variable service not provided"). A more user-facing phrasing: `fix COPY TO failure when subquery uses bind variables`.

**7. Commit titles exceed 50-character guideline**

- "Snapshot bind variable values before enqueueing export task" — 59 chars
- "Add streaming-path test and negative assertion for bind variable export" — 71 chars

nwoolmer:
## Review (round 2)

### Critical

**1. `snapshot()` throw publishes an uninitialized ring queue slot**

`CopyExportFactory.java:257-260` — `BindVariableServiceImpl.snapshot()` is evaluated inline as an argument to `task.of()`. If it throws `SqlException`, `task.of()` never executes, but the `finally` block at line 266 unconditionally calls `copyRequestPubSeq.done(processingCursor)`, publishing the uninitialized queue slot. The worker thread would then process a stale/empty task and NPE on `task.getEntry()`.

**Fix:** compute the snapshot into a local variable before the try block that claims the queue slot.

### Moderate

**2. `snapshotNamedFunction` missing `ColumnType.IPv4`**

`snapshotFunction` handles IPv4 at line 158, but `snapshotNamedFunction` has no IPv4 case. A named IPv4 bind variable is silently dropped.

**3. `snapshotNamedFunction` loses TIMESTAMP_NANO precision**

Line 225 calls `copy.setTimestamp(name, value)` which hardcodes `TIMESTAMP_MICRO`. The indexed variant correctly preserves the subtype via `setTimestampWithType`. The private `setTimestampWithType(CharSequence, int, long)` at line 1620 is accessible from within the same class.

**4. Test coverage gaps for `snapshot()` types**

`testSnapshotScalarTypes` covers 10 types but `snapshotFunction` handles 18+ branches. Missing: IPv4, VARCHAR, LONG256, UUID, BINARY, GeoHash, Decimal.

### Minor

**5. Private static method ordering** — should be: `copyBinarySequence`, `snapshotFunction`, `snapshotNamedFunction`.

**6. `copyBinarySequence` long-to-int truncation** — add a guard for lengths > Integer.MAX_VALUE.

**7. Javadoc passive voice** — should use active voice per CLAUDE.md.

glasstiger:
#### [PR Coverage check]

:heart_eyes: **pass** : 83 / 101 (82.18%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/bind/BindVariableServiceImpl.java|66|84|78.57%|
|:large_blue_circle:|io/questdb/cutlass/parquet/CopyExportRequestJob.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/ExportQueryProcessor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/CopyExportRequestTask.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/BaseParquetExporter.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CopyExportFactory.java|5|5|100.00%|

PR #6886, opened 2026-03-19
chore(build): fix jvm env vars on jobs due to CI change

sklarsa:
The move to incus runners in #6768 changed the snapshot and coverage
report jobs to run on `hetzner-incus`, but didn't add `JAVA_HOME_17_X64`
and `JAVA_HOME` variables. The Maven@3 task and the cover-checker script
both need these to find the JDK.

Previously the snapshot pipeline ran on `ubuntu-latest` (Microsoft-hosted)
where `JAVA_HOME_17_X64` is pre-configured. The coverage report job relied
on `steps.yml` to set it via `##vso[task.setvariable]`, but it doesn't
use that template.

Adds `JAVA_HOME_17_X64` and `JAVA_HOME` pointing to `/opt/graalvm`
(installed in the incus container image) to both jobs.

## Test plan
- ~Snapshot pipeline `BuildOnLinux` job finds JDK and completes the Maven build~
  - Not running this because it writes to s3
- Coverage report job runs `cover-checker` without `JAVA_HOME_17_X64: unbound variable` error
  - https://dev.azure.com/questdb/questdb/_build/results?buildId=221303&view=results

### DIFF
--- a/core/src/main/java/io/questdb/ServerMain.java
+++ b/core/src/main/java/io/questdb/ServerMain.java
@@ -575,4 +575,3 @@ public class ServerMain implements Closeable {
         }
     }
 }
-// CI trigger


glasstiger:
#### [PR Coverage check]

PR #1058, opened 2021-05-24
fix(pgwire): using IN with bind variables aka IN ($1) to work as with const literals

ideoma:

PR #6890, opened 2026-03-20
chore(build): remove pre-compiled native binaries from the repository

sklarsa:
## Summary

Tandem with questdb/questdb-enterprise#1058, which bundles the OSS Rust native library in the Linux and macOS enterprise binaries.

- Remove all pre-compiled Rust binaries (`libquestdbr`) from git tracking across all 5 platform variants
- Build Rust libraries from source via the Maven Rust plugin (`org.questdb:rust-maven-plugin:1.2.0`) during `mvn compile`, active by default, skip with `-DskipNative`
- Add `sccache` integration for Rust compilation caching on Windows and macOS CI (uses Azure Blob Storage)
- Replace the Azure DevOps binary release pipeline (`ci/github-release-pipeline.yml`) with a GitHub Actions workflow (`.github/workflows/github-binaries-release.yml` — "Github Release - Binaries")
- Install Rust + `build-essential` in `core/Dockerfile` so the Docker image builds the native libs from source
- Add Rust toolchain install steps to the CI workflows that build native code (`glibc_smoke_test`, `parquet_compat`, `pgwire_latest`, `pgwire_stable`, `r_test`)
- Delete the obsolete `rebuild_rust` GitHub Actions workflow

### New release workflow: "Github Release - Binaries"

Triggered on semver tags (`[0-9]*.[0-9]*.[0-9]*`) or manually via `workflow_dispatch`.

**Rust build jobs (5 parallel):**
- `build-rust-linux-x64` — `ubuntu-latest` + `manylinux2014_x86_64` Docker (glibc 2.17)
- `build-rust-linux-arm64` — `ubuntu-22.04-arm` + `manylinux_2_28_aarch64` Docker (glibc 2.28)
- `build-rust-macos-arm64` — `macos-14` (M1 native)
- `build-rust-macos-x64` — `macos-15-intel` (native)
- `build-rust-windows` — `windows-latest` (native)

**Packaging jobs:**
- `package-linux` — downloads all 5 Rust libs into `core/target/classes/io/questdb/bin/<platform>/`, then builds with GraalVM Community 25.0.2 + Maven using `-DskipNative` (so the downloaded libs are not rebuilt or overwritten). Produces the `no-jre` archive — whose `questdb.jar` carries all 5 platforms' Rust libs plus the committed C++ libs — and `rt-linux-x86-64`
- `package-windows` — downloads the Windows Rust lib, builds `rt-windows-x86-64`

**Release job** (tag pushes only):
- Uploads archives to the GitHub Release
- Deploys AMI

Secrets required: `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` (for AMI). `GITHUB_TOKEN` is built-in.

### Other CI / build changes

- Flip the default `qdbr.release` to `true` and rename the `qdbr-release` Maven profile to `qdbr-debug` — production (release) Rust builds are now the default; opt into a debug build with `-P qdbr-debug`
- Fix the `qdbr-coverage` profile's `qdr.release` typo to `qdbr.release`
- Fix sccache pipeline variable names (`SCCACHE_AZURE_BLOB_CONTAINER`, `SCCACHE_AZURE_CONNECTION_STRING`)
- Fix macOS x86-64 Maven profile activation (`<arch>x86_64</arch>` instead of `amd64`)
- Add the `platform.dir.name` property to the Windows Maven profile
- Rust plugin `copyTo` writes to `${platform.dir.name}` instead of using `copyWithPlatformDir`
- Split the `rt-*` assemblies' native-lib filesets: C++ `libquestdb.*` from `src/main/resources`, Rust `*questdbr.*` from `target/classes`

## Test plan

- Verify `mvn compile test-compile` succeeds with Rust libs built from source on all CI platforms
- Verify the released `no-jre` JAR contains Rust binaries for all 5 platforms
- Verify QuestDB starts and responds to queries using the `no-jre` JAR
- Verify the "Github Release - Binaries" workflow builds all artifacts on `workflow_dispatch`
- Verify sccache works on Windows and macOS CI runs


glasstiger:
#### [PR Coverage check]


ideoma:
# Re-review: PR #6890 — `chore(build): remove pre-compiled native binaries from the repository`

**Level 0** (single-pass inline). This is a re-review; the branch advanced to `50251f3c7c "fix nits from the review"`, which folded three of my prior findings directly into the PR. The diff is otherwise the same build/CI change I verified earlier, so I re-confirmed the deltas rather than re-deriving the whole mechanism.

**What changed since the last review** (all now part of the PR diff):
- `Os.java` — the dead `devRustLibRoot = "/io/questdb/rust/"` constant and DEV-override branch are removed (verified: safe, behavior-identical, compiles `BUILD SUCCESS`).
- `artifacts/release/README.md` — the manual-fallback section now documents that a single-host `mvn package` only builds the host's Rust lib and that the cross-platform `no-jre` requires gathering all five `build-rust-*` artifacts (verified: the artifact→path mapping matches both the workflow's `download-artifact` steps and `Os.java`'s `prdLibRoot`; the `-DskipNative`/no-`clean` reasoning is correct).
- PR description — the stale Docker-pipeline section and the wrong "GraalVM 17" are gone; it now matches the diff and footnotes that the Docker migration shipped in #6980.

## Critical
None.

## Moderate
None remaining. Both prior Moderate findings are resolved in this PR:
- ~~README manual fallback produced an incomplete `no-jre`~~ → fixed by the README rewrite in `50251f3c7c`.
- ~~PR description described Docker changes absent from the diff~~ → fixed by the description update.

## Minor
These are unchanged from the prior pass and remain Minor/informational — none block merge.

1. **`release` job assumes a GitHub release for the exact tag already exists** (`.github/workflows/github-binaries-release.yml`, in-diff). `gh release upload "${GITHUB_REF_NAME}" artifacts/*.gz` has no existence guard, where the replaced Azure step resolved the release first. If no draft exists for the tag, this fails *after* all five Rust builds and both packaging jobs ran (~30-60 min wasted). It matches the documented "upload to the existing draft" flow, so it's intended — but a `gh release view`/`create` guard would turn a late failure into an early, clear one.

2. **`-DskipNative` produces a non-runnable build, with no guard** (in-diff, conceptual). A build made with `-DskipNative` and no seeded libs has no `libquestdbr`, so the server/tests crash in `Os.java` (`FatalError: cannot find .../libquestdbr.*`). Safe for the workflow's pre-seeded packaging and the JaCoCo compile, but a developer using it to "skip the slow Rust build" then running the server gets an opaque failure. The README now documents the flag in the fallback context, but a one-line "not runnable" caveat in CLAUDE.md/README would help.

3. **`glibc_smoke_test.yml` no longer triggers on committed C++ binary changes** (in-diff). The path filter moved to C/Rust *source*; the C++ `libquestdb.*` libs are still committed, so a future swap of one without a source change won't run the smoke test. Correct for the now-uncommitted Rust libs, small coverage gap for the still-committed C++ libs.

## Downgraded (verified correct, not findings)
- **"Removing the `Os.java` DEV-override changes lib-loading behavior"** — dismissed: repo-wide grep shows nothing targets `io/questdb/rust`; both build profiles copy to the production path; the surviving call is the one that always ran. Pure dead-code removal.
- **"README no-jre manual steps are wrong"** — dismissed: the five artifact→directory mappings match the workflow and `Os.java`; `-DskipNative` correctly avoids rebuild/overwrite and `no clean` preserves the placed libs (mirrors the workflow's own `package-linux` job).
- Carried over from the first pass and still valid: `platform.dir.name` defined for all 5 profiles; `prdLibRoot` matches the `copyTo` destination; macOS `<arch>x86_64>` is a correct fix; no dangling `qdbr-release` refs (OSS + enterprise); assembly `libquestdb.*`/`*questdbr.*` filesets are disjoint; the GH Actions release workflow's own semver check correctly uses `[0-9]` (unlike the out-of-scope #6980 Docker pipeline's broken `\d`).

## Summary
**Verdict: approve.** The PR is in better shape than at the first review — three of the prior findings (2 Moderate, 1 Minor) are fixed in-diff, and I re-verified those fixes are correct. No Critical or Moderate issues remain; the three Minor items are non-blocking process/doc hardening. All findings are in-diff (the principal cross-context consumer, `Os.java`, is now part of the diff and verified; the enterprise-repo and `io/questdb/rust` cross-context checks came back clean). The one tradeoff to keep in mind operationally: the released `no-jre` is only all-platform because the workflow gathers all five Rust libs — a fact now correctly documented for the manual fallback, and worth confirming on a real `workflow_dispatch` run before the first release on this pipeline, since it's a new release path with no automated test.

ideoma:
# Review: PR #6890 — `chore(build): remove pre-compiled native binaries from the repository`

**Level 3** (full pass), run in tandem with the prior level-0 re-review. This is a build/CI PR: the only production Java change is ~12 lines of startup-init in `Os.java`, and there are no `.rs` files. The data-path review dimensions (concurrency, zero-GC/SIMD hot-path, Rust-panic-via-JNI) have no surface here and were replaced by build-graph, release-workflow, CI-coverage, and shell/Dockerfile analysis covering the same adversarial spirit. Each finding was verified against source; one headline finding was withdrawn on review (see Downgraded).

**Verdict: approve with follow-ups. No Critical issues.** The core mechanism is correct: for all five shipped platforms the from-source Rust lib lands exactly where `Os.java` loads it, the `no-jre` JAR is correctly assembled with all five platforms' libs by the release workflow, the Rust libs' glibc floor is guaranteed by the manylinux build containers, and the enterprise repo was already adapted in tandem. The findings below are coverage/robustness items and one fringe-platform bug — none block merge, but several are worth addressing before this becomes the release path.

## Critical
None.

## Moderate

**1. FreeBSD from-source build no longer loads the Rust lib.** *(in-diff trigger × out-of-diff gap — `core/pom.xml:545-558` + `:506` + `Os.java:389`)*
The `platform-freebsd-x86-64` profile is the only platform profile with no `<platform.dir.name>` (verified: none set there, no global default). Now that `build-rust-library` is active-by-default and `copyTo` is `target/classes/io/questdb/bin/${platform.dir.name}`, on FreeBSD that property is unresolved, so the plugin writes to a stray directory while `Os.java:361` looks in `/io/questdb/bin/freebsd-x86-64/` → `FatalError` at startup. Pre-PR, the `-P build-rust-library` workflow worked on FreeBSD via the platform-agnostic DEV path that this PR removes.
**Reachability caveat:** FreeBSD has no CI, never shipped a committed Rust binary, and is explicitly best-effort. Real-world impact is near-zero. But it's a genuine won't-start regression on a platform that has a first-class pom profile + `rt-freebsd.xml`, and the fix is one line: add `<platform.dir.name>freebsd-x86-64</platform.dir.name>` to that profile.

**2. The `release` job uploads with no existence guard and no `--clobber`, and fails late.** *(in-diff — `github-binaries-release.yml:333-339`)*
`gh release upload "${tag}" artifacts/*.gz` assumes a draft release already exists for the tag (the README confirms the draft is created out-of-band) and has no `--clobber`. Two failure modes: a tag pushed with no draft, or a re-run where assets already exist — both fail **after** all five Rust builds and both packaging jobs (~30-60 min wasted). The deleted Azure pipeline resolved the release first via `gh release view ${tag_name}`; the new workflow dropped that, so the only failure point moved to the very end of the job graph.
*Fix:* add a `gh release view "${GITHUB_REF_NAME}"` guard (or `gh release create --draft` if missing) at the top of the `release` job, and `--clobber` on the upload.

**3. `THIRD_PARTY_LICENSES.txt` can silently ship stale.** *(in-diff)*
Deleting `rebuild_rust.yml` removed the only automation that regenerated *and committed* this file. It's now refreshed only at release (`generate_third_party_licenses.sh`) and in the Docker build — so released/Docker artifacts are correct, but every local/IDE/per-commit `mvn package` bundles the committed copy, which nothing keeps in sync with `Cargo.lock`. `rust_license_check.yml` checks license *policy*, not the attribution list, so drift won't fail CI. The committed copy is already dated `2026-06-13`. This is a legal-attribution artifact, so stale content has real (if low-probability) consequences.
*Fix:* add a CI check on PRs touching `core/rust/**/Cargo.*` that runs the generator and `git diff --exit-code THIRD_PARTY_LICENSES.txt`.

**4. Packaged-distribution validation is Linux-only and on-demand.** *(in-diff)*
`binary-smoke-test-jobs.yml` matrixes Linux x64/arm64 only and is deliberately on-demand (`/azp run`, not per-commit). The macOS and Windows packaged distros are built by the release workflow but never started, so the `questdb.libs.dir` lib-dir loading path (the rt-archive branch of `Os.java`) is exercised end-to-end only on Linux. Unit-test jobs do load `libquestdbr` from the classpath on mac/Windows, so the lib itself is known to load there — the gap is specifically the *packaged* `lib/`-dir path and the assembly fileset split. Combined with the on-demand cadence, a packaging/assembly regression can sit on `master` between manual runs.
*Fix (optional):* add mac/Windows legs to the smoke test, or at least a `dlopen` of the packaged Rust lib in the respective package jobs; consider also running the smoke test on the nightly schedule.

**5. PR body cites the wrong PR number.** *(metadata)*
The body says the Docker runner migration "landed separately in #6980". #6980 is **"chore(core): update QuestDB to JDK 25"** (merged), not a runner migration. The *substance* is right (the incus migration is on `master` and not in this PR), but the reference is wrong — drop it or cite the actual PR(s).

## Minor

- **`package-linux` relies on cargo being preinstalled on `ubuntu-latest`** (`github-binaries-release.yml:217-218`). `generate_third_party_licenses.sh` hard-requires `cargo` (it `exit 1`s otherwise), and this job never installs Rust. It works only because the GitHub ubuntu image ships rustup — yet this PR added an *explicit* `setup-rust-toolchain` step to `parquet_compat`/`pgwire_*`/`r_test`. Make the release path's dependency explicit and pinned too. (The `install ... /usr/local/bin/cargo-deny` without sudo is fine — GH runners and the root Docker stage both allow it.)
- **`validate-tag` comment is inaccurate** (`github-binaries-release.yml:21-22`): it claims to match `ci/docker-release-pipeline.yml`, but that pipeline uses `^\d+(\.\d+)*$` (accepts `1`, `1.2`, …) while `validate-tag` requires ≥3 components. Harmless for real releases; fix the comment or align the regexes.
- **`core/Dockerfile`**: `ENV NO_RUST_PATH="$PATH"` is dead — never referenced anywhere; remove it. The `build-essential` comment "for native compilation" is misleading — the C++ lib is committed, not built here; build-essential is needed for the **Rust** build (cc/zstd-sys/lz4-sys C deps + linker). Reword.
- **`ci/glibc_load_check.sh:9-11`**: the `RTLD_LAZY` rationale ("defers undefined JNI symbols") doesn't apply to the C++ lib it actually checks (that lib has no undefined JVM symbols; the glibc/libstdc++ version check happens at load time regardless of bind mode). The mode choice is correct; only the comment is misleading.
- **Version-pin divergence / duplication**: cargo-deny is pinned to `0.19.8` here but `rust_license_check.yml` resolves the *latest* cargo-deny — the generator and the checker can run different tool versions. Separately, the Rust toolchain `nightly-2025-09-24` is hardcoded in four workflow files instead of read from `rust-toolchain.toml` (which the Azure side does via `prepare_rust_env.py --match`); a future toml bump must be mirrored by hand in all four.
- **`Os.java:310-326` (pre-existing, now more relevant)**: `getNativeLibsDir()` only checks the **C++** lib's presence. A distribution `lib/` dir containing `libquestdb.*` but not `libquestdbr.*` would `System.load` the C++ lib, throw on the Rust lib, then the fallback re-loads the C++ lib from the jar — a confusing double-load instead of a clean "missing libquestdbr" error. Not reachable via the rt assemblies (they always ship both), but the PR makes the lib-dir layout the primary release shape, so consider having the dir check cover the Rust lib too.
- **PR metadata**: labels `CI`, `rust` are thin for a change that edits `core/pom.xml`, assemblies, `Dockerfile`, and `Os.java` — add `Build` (and arguably `Core`). The body reads as a feature list; per the tone convention it should disclose the tradeoffs the diff introduces (Rust toolchain now mandatory to build; manual release assembly is more involved; from-source packaging validated per-PR rather than per-commit). Em dashes appear in a few YAML/`.gitignore` comments — cosmetic, not a violation of the Java-LOG ASCII rule.
- **Footgun (informational)**: `!skipNative` activation deactivates the Rust build if `skipNative` is defined to *any* value, including `-DskipNative=false`. No current CI step does this (verified), but it's a sharp edge worth a comment near the activation.

## Downgraded (false positives / verified-correct)

- **"Released Rust libs are never glibc-load-tested" — withdrawn after review.** The released Rust libs are built in controlled-glibc containers (`manylinux2014_x86_64` = glibc 2.17 for x64; `manylinux_2_28_aarch64` = glibc 2.28 for arm64); `package-linux` only assembles the jar from those pre-built libs. That build environment *is* the floor guarantee, and it's actually **stronger** than what the deleted smoke test verified — that test ran on Amazon Linux 2 (glibc **2.26**) / AlmaLinux 8 (glibc 2.28), so for x64 it could not even catch a floor regression into [2.18, 2.26]. The Rust lib also pulls in **no libstdc++** (native deps are `cc`/`lz4-sys`/`zstd-sys`, all C), so there's no libstdc++ symbol-versioning floor — that risk lives in the **C++** `libquestdb.so`, which *keeps* its dedicated old-glibc `dlopen` check via `glibc_load_check.sh`. And "the no-jre boots and serves a query" is still covered by `binary-smoke-test-jobs.yml` on a modern-glibc box. So the in-code comment's conclusion (the Rust lib needs no separate glibc check) is sound. The only forward-looking note — optional, not a finding — is to add a one-line comment in `binary-smoke-test-jobs.yml` stating the floor is guaranteed by the manylinux image and to revisit if `qdbr` ever gains a C++/`cxx` dependency or the x64 image is bumped off `manylinux2014`.
- **Enterprise assembly break** — *not a finding*: `questdb-ent/linux-runtime.xml:58-65` was already repointed to pull the Rust lib from `../questdb/core/target/classes/io/questdb/bin/${platform.dir.name}`, and enterprise builds OSS in-reactor without `-DskipNative`. Adapted in tandem.
- **`no-jre` JAR missing other platforms' libs** — *correct by design*: `no-jre.xml` bundles `questdb.jar`, whose contents come from `target/classes`; the release `package-linux` downloads all five artifacts there and packages with `-DskipNative` and no `clean`, so they survive and ship in the jar.
- **`qdbr-coverage` `copyTo` uses unset `${platform.dir.name}`** — *not a bug*: coverage runs only on Linux self-hosted, where `platform-linux-x86-64` sets it to `linux-x86-64`.
- **macOS-x64 arch activation `amd64`→`x86_64`** — *correct fix*: macOS JVMs report `os.arch=x86_64`; Linux/Windows report `amd64` (and Linux's profile correctly stays `amd64`). OSS pom applies no `os-maven-plugin` normalization, so the raw match is what counts.
- **Rust build phase ordering for `clean test`** — *correct*: the plugin's `build` goal binds to `compile`, so the lib is in `target/classes` before surefire runs.
- **Tests double-run / zero-run across OS** — *none*: the sccache test step (Win/mac) and the original test step (Linux) are mutually exclusive and jointly exhaustive over {Linux, macOS, Windows}.
- **Release builds omit `RUSTFLAGS=-D warnings`** — *acceptable/deliberate*: you don't want a new upstream lint to fail a tagged release; the `-D warnings` gate still runs on every PR.

## Summary

- **Verdict:** Approve with follow-ups — no blocking issues. The native-loading contract is correct for all five shipped platforms across every runtime context (rt archive, `no-jre` jar, IDE/tests); the Rust libs' glibc floor is guaranteed by the manylinux build images; the enterprise side is already adapted.
- **Tradeoffs/regressions to weigh:** the change trades committed-binary simplicity for from-source builds, and in doing so thins out *some* validation — packaged-distro loading is now validated Linux-only and on-demand (Moderate #4), and the license file can drift in non-release builds (Moderate #3). Neither is a regression in shipped-artifact *correctness*. The release workflow is functional but fragile at the upload step (Moderate #2). The only behavioral bug is the FreeBSD from-source path (Moderate #1), which is a fringe platform with a one-line fix.
- **Counts:** ~24 distinct draft findings → **5 Moderate + 9 Minor verified; 8 dropped as false positives / correct-by-design** (including the glibc finding, withdrawn on review). No Critical.
- **In-diff vs out-of-diff:** essentially all findings are **in-diff** — this is a self-contained build/CI change. The one out-of-diff surface with real blast radius (the enterprise distribution assembly) was checked and is already handled in tandem.

ideoma:
## Code review — `chore(build): remove pre-compiled native binaries from the repository`

Reviewed in tandem with the enterprise PR (#1058). Focus areas: native-library loading (`Os.java`), Maven profiles/assemblies, the new GitHub Actions release workflow, the Azure pipeline changes, the new shell scripts, and the enterprise distribution assembly.

### Verdict
**Approve — no blocking issues.** Native-library loading is correct for all five shipped platforms (linux-x86-64, linux-aarch64, darwin-aarch64, darwin-x86-64, windows-x86-64) across every CI-blessed runtime context (`rt-*` archive, cross-platform `no-jre` jar, tests/IDE-via-Maven). The release workflow's artifact wiring is fail-closed, and the enterprise tandem is correctly adapted. Remaining items are Minor / informational.

### What was verified correct
- **`Os.java` native loading** — the Rust DEV-path override (`/io/questdb/rust/`) is removed and the lib now always loads from the prod classpath path (`/io/questdb/bin/<platform>/`), which is exactly where the `rust-maven-plugin` now writes it (`target/classes/io/questdb/bin/<platform>/`). Consistent across the rt archive, the `no-jre` jar, and tests/IDE.
- **Release artifact wiring is fail-closed** — each platform lib is fetched by a *named* `download-artifact@v4` into a distinct directory; a missing upstream artifact fails `package-linux` rather than silently shipping an incomplete `no-jre` jar. `upload-artifact@v4` uses exact paths, and `cargo build --release` either emits a real cdylib or fails the job. There is no all-green path that yields a `no-jre` missing a platform's `libquestdbr.*`.
- **Assembly globs don't cross-match** — `libquestdb.*` does not match `libquestdbr.so` (next char is `r`, not `.`) and `*questdbr.*` does not match `libquestdb.so`; the two filesets also read from different directories (`src/main/resources` for C++, `target/classes` for Rust).
- **`copyWithPlatformDir=true` → explicit `${platform.dir.name}`** — equivalent on the Unix shipping platforms and additionally fixes a latent Windows mismatch (JNA's prefix would have been `win32-x86-64`, which `Os.java` never reads).
- **macOS x64 arch activation `amd64` → `x86_64`** — correct; Intel macOS JVMs report `os.arch=x86_64` (Linux/Windows report `amd64`, and that profile correctly stays `amd64`).
- **Maven lifecycle ordering** — the rust `build` goal binds to `compile` (after `process-resources`); the resources plugin copies-but-never-deletes, so the committed C++ libs and the built Rust lib coexist in `target/classes` before surefire.
- **Test-step split** — "Run tests (with sccache)" (Win/mac) and "Run tests" (Linux) are mutually exclusive and jointly exhaustive; no double-run or zero-run on any OS. Both build native from source (no `-DskipNative`), so the from-source Rust build **and** runtime load is exercised per-PR on all three OSes.
- **`validate-tag` gate** — meaningful (the trigger glob is looser than the regex); `release` cannot run on an invalid tag and is correctly skipped on `workflow_dispatch`.
- **Rust lib needs no separate glibc check** — built in `manylinux2014`/`manylinux_2_28` (glibc 2.17/2.28 floor) and pulls in no libstdc++; the committed C++ lib keeps its dedicated `glibc_load_check.sh`.
- **Enterprise tandem (#1058)** — `questdb-ent/linux-runtime.xml` splits C++ (`libquestdb.*` from OSS `src/main/resources`) and Rust (`*questdbr.*` from OSS `target/classes`), the submodule pointer is bumped to the branch head, enterprise builds OSS in-reactor without `-DskipNative` (so the Rust lib lands where the assembly reads it), and a from-source smoke test was added.

### Minor / informational
1. **`THIRD_PARTY_LICENSES.txt` can drift in non-release builds.** It is regenerated only in the release workflow and `core/Dockerfile`; a plain `mvn package` ships the committed copy. This is documented as the intentional fallback, but the committed file is now the thing most likely to go stale.
2. **macOS `rt-*` archive loading is unvalidated by CI.** The from-source build + runtime load is exercised per-PR on Win/mac by the Azure `clean test` runs, but no job starts a packaged `rt-osx-*` distribution on either Mac arch; only the Linux release artifact is smoke-tested end to end (`binary-smoke-test-jobs.yml`).
3. **`qdbr-coverage` now co-activates `build-rust-library`.** Both profiles declare a `qdbr-build` execution on the `compile` phase, so Maven merges them by id. This should keep the coverage profile's `RUSTFLAGS=-C instrument-coverage` (later-declared profile dominant), but it's fragile — worth confirming the next coverage run still emits non-empty Rust LCOV, and consider consolidating to one `qdbr-build` definition.
4. **Toolchain version duplicated.** `nightly-2025-09-24` is hardcoded in `parquet_compat`/`pgwire_latest`/`pgwire_stable`/`r_test` and duplicates `core/rust/qdbr/rust-toolchain.toml`. They match today and `cargo` honors the toml regardless, but the four copies won't track the source of truth on a future bump.
5. **`package-windows` is serialized behind all of `package-linux`** purely to reuse the `third-party-licenses` artifact, roughly doubling release wall-clock. A small standalone license job feeding both packagers would restore parallelism. Release-time only.
6. **`!skipNative` footgun** — `-DskipNative=false` *defines* the property, so it disables the Rust build (counterintuitive). No CI step does this; a one-line comment near the activation would help.
7. **`gh release upload`/`edit` assume a pre-existing release/draft** (grep-confirmed no `gh release create`). Matches the deleted Azure pipeline, so not a regression, but a tag pushed without a draft burns the full build before failing at the last step. A preflight `gh release view` would fail fast.
8. **`package-linux` relies on `mvn package` *not* being `mvn clean package`** (a `clean` would wipe the four non-host libs from `target/classes`). Correct as written; worth a comment on the step so a future "normalization" edit doesn't silently break the multi-platform `no-jre`. The "Verify Rust libraries" step is cosmetic (`find … -exec ls` always exits 0) — the real guard is the fail-closed `download-artifact` steps above it.
9. **PR body reads as a feature list.** Per the project tone convention it should give the tradeoffs equal weight (Rust toolchain now mandatory to build; manual release assembly is more involved; from-source packaging validated per-PR rather than per-commit; license file can drift in non-release builds).

### Action item (not a code finding)
The `Enterprise CI Failed` status above was reported against the older commit `102b6f3b97`, not the current head `3cd0485cc6`. Re-run enterprise CI on the current head and confirm green before merge.

---
*Reviewed with Claude Code.*

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/Os.java|1|1|100.00%|

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/Os.java|1|1|100.00%|

bluestreak01:
## Review — `chore(build): remove pre-compiled native binaries from the repository`

29 files, +844/−653. This is a CI/build-infrastructure PR: it stops committing the Rust `libquestdbr` binaries and builds them from source via Maven/CI. The only production-code change is `core/src/main/java/io/questdb/std/Os.java`. Reviewed the Java loader/Maven wiring, the new GitHub release workflow, the Azure pipeline + shell scripts, and ran an adversarial pass; every draft finding was verified against source.

### Metadata
Clean — Conventional-Commits title, accurate description, correct labels. Minor: folds in two unrelated `.gitignore` lines (sqllogictest/parquet-tools).

### Critical
None. No guaranteed break in the CI/release happy path — the loader path matches the build output across all 5 supported platforms, and the release jobs pre-populate all platforms before packaging. The items below are latent failure surfaces the green CI run will hide.

### Moderate

1. **`package-linux` has no explicit Rust toolchain, but `generate_third_party_licenses.sh` hard-fails without `cargo`.** `.github/workflows/github-binaries-release.yml:206-211` runs the script, which `exit 1`s at `ci/generate_third_party_licenses.sh:23-26` if `cargo` is absent (and shells `cargo metadata`, honoring `rust-toolchain.toml`'s `nightly-2025-09-24`). `package-linux` installs only GraalVM — unlike all five `build-rust-*` jobs and the parquet/pgwire/r_test workflows in this same PR, which add an explicit `setup-rust-toolchain`. It works today only because `ubuntu-latest` ships Rust preinstalled and `/usr/local/bin` is runner-writable (the script's `install … /usr/local/bin/cargo-deny`). Both are implicit runner-image assumptions; if either changes this is a release-time hard failure. Add an explicit `setup-rust-toolchain@… nightly-2025-09-24` step for parity.

2. **`gh release upload` is non-idempotent and assumes the draft exists** (`github-binaries-release.yml:359-360`). No `--clobber`, no `gh release create`/`view` (the deleted Azure job did `gh release view` first), and the AMI/Packer deploy runs in the *same* `release` job *after* the upload. `artifacts/release/README.md:93` now says "prefer re-running the failed jobs" — so a re-run after an AMI/Packer flake re-hits `gh release upload`, which fails because the assets already exist; and if no draft exists the upload hard-fails with "release not found." Use `gh release upload --clobber` and/or split the AMI deploy into its own job so upload is safely re-runnable.

3. **The aarch64 C++ glibc check is gone entirely.** The deleted `glibc_smoke_test.yml` had both `amd64:` (amazonlinux:2) and `aarch64:` (almalinux:8) legs that started the server on old glibc. The replacement `ci/templates/binary-smoke-test-jobs.yml` only `dlopen`s the committed C++ lib on `linux-x86-64`/amazonlinux:2. The committed `linux-aarch64/libquestdb.so` still ships, but nothing in CI verifies it loads on old ARM glibc anymore.

4. **No glibc verification of the *built* Rust libs in the release path.** The Linux Rust libs build inside manylinux2014 (glibc 2.17) / manylinux_2_28 (2.28) — a floor by construction — but nothing asserts it; `glibc_load_check.sh` only checks the C++ lib. Consider running `glibc_load_check.sh` against each downloaded Rust lib inside an old-glibc container in `package-linux`.

5. **No build-time guard that the no-jre actually bundles all 5 Rust libs.** The premise rests on a fragile chain: download-all-5 → `mvn package -DskipNative` → *no* `clean` → resources-plugin copy-not-sync. The "Verify Rust libraries" step (`:202-204`) is just `find … -exec ls` and never fails on a missing platform. A future change to `mvn clean package` (muscle memory; and the README's own manual rt example uses `clean`) would silently wipe `target/classes`, ship a no-jre with zero Rust libs, and produce no build error — QuestDB then fails to start everywhere. A cheap guard (assert the packaged jar contains 5 `*questdbr*` entries) converts a silent, high-blast-radius failure into a build error.

6. **Contributor/IDE regression: fresh clone + IDE test run now `FatalError`s at startup.** `Os.java:389` unconditionally `loadLib(prdLibRoot + rustLibName)`, and `loadLib(String)` throws `FatalError("…broken package?")` when the resource is absent (`Os.java:256-260`). Previously the committed `libquestdbr.*` was a resource present after any compile; now it lands in `target/classes` only via the `rust-maven-plugin`, so an IntelliJ-only build (no Maven Rust step, or no Rust toolchain) hard-fails at class init. `CLAUDE.md` added "Rust toolchain" as a prereq and `intellij_triggers.xml` offers `qdbr-debug-build`, but the Ant trigger must be wired manually. Intended consequence, but ensure contributor onboarding docs/IDE setup cover it. (Every local `mvn test` now also triggers an incremental cargo build.)

7. **(Unverifiable from the repo) sccache hard-depends on two Azure variables.** `SCCACHE_AZURE_BLOB_CONTAINER` / `SCCACHE_AZURE_CONNECTION_STRING` (`ci/templates/steps.yml:204-205, 251-252`) appear in no YAML/variable group, so they must be Azure secrets. If absent, Azure leaves the literal `$(VAR)` macro, sccache gets an invalid backend, and since `RUSTC_WRAPPER` is set the first `rustc` fails → `mvn clean test` fails → every macOS/Windows test job goes red. Confirm these are defined.

8. **(Operational, unverifiable from the repo) the old Azure release pipeline still triggers on tags.** `ci/github-release-pipeline.yml` is deleted, but its Azure DevOps pipeline definition (`definitionId=33`) lives server-side and triggers on `*.*.*` tags. On the next release tag it will fail (YAML gone) or race the new workflow uploading to the same release. Must be disabled operationally before the first post-merge release.

### Minor
- **`-DskipNative=false` does the opposite of intent.** Activation `<name>!skipNative</name>` (`core/pom.xml:489-491`) keys on property *presence*, so `-DskipNative=false` *deactivates* the Rust build. Latent footgun.
- **FreeBSD profile lacks `<platform.dir.name>`** (`platform-freebsd-x86-64`). With `build-rust-library` now default, a from-source build on FreeBSD copies to a literal `${platform.dir.name}` dir and `Os.java` then `FatalError`s on `/io/questdb/bin/freebsd-x86-64/`. FreeBSD never shipped binaries, so low severity; one-line fix.
- **Release Rust builds dropped `--lib`.** `github-binaries-release.yml` lines 53/80/100/118/135 run bare `cargo build --release`; `core/rust/qdbr/Cargo.toml` declares six non-cdylib targets (`pm_inspect`, `pm_generate`, `decode_page`, `encode_page`, `decode_rle`, `bench_row_group`), all now compiled with release LTO ×5 platforms for no benefit. Restore `--lib`.
- **`steps.yml` compile-only task builds release Rust for nothing** — dropped `-P qdbr-release` but not added `-DskipNative` (unlike `aux-job.yml:52`, `test-pipeline.yml:254`); cold release Rust compile, no sccache, on legs where no tests run. "Pre-build for Coverage" similarly builds release then rebuilds instrumented (double build).
- **Linux test shards get no sccache** (install gated to Windows/macOS).
- **`.gitignore` scope creep / no-op:** `parquet-tools/Cargo.lock` is already git-tracked so the ignore line does nothing; `qdb-sqllogictest/target` is unrelated to this PR.
- **Dead `ENV NO_RUST_PATH="$PATH"`** (`core/Dockerfile:24`).
- **`generate_third_party_licenses.sh` is Linux-musl-only** (`:36-48`); a future non-Linux caller would fetch an unrunnable binary. `mktemp -d` also lacks a cleanup trap.
- **`actions/upload-artifact@v6` vs `actions/download-artifact@v7`** version skew — align majors.
- **`validate-tag` comment inaccurate**: a `workflow_dispatch` against a *tag* ref makes `validate-tag` run *and* the `release` job (`:330`) fire.
- **Toolchain hardcoded in 4 workflows** (`nightly-2025-09-24`) instead of derived from `rust-toolchain.toml` — currently matches; a bump must touch 5 places.
- **New sccache test task drops `-s $(MAVEN_SETTINGS_FILE)` and `--quiet`** vs the Linux task — silent drift, harmless today.
- Non-ASCII em dash in a `.gitignore` comment.

### Downgraded (verified false positives)
- **"Coverage silently builds uninstrumented Rust because two profiles define `qdbr-build`."** False — same execution id merges, coverage `RUSTFLAGS=-C instrument-coverage` preserved; and the `qdr.release`→`qdbr.release` typo fix (`core/pom.xml:672`) is load-bearing and keeps coverage at a debug build. Correct as written.
- **"`platform.dir.name` mismatch for supported platforms."** False — `Os.java` `name+'-'+archName` is string-equal to every supported profile's `<platform.dir.name>` and the download dirs.
- **"osx `<arch>x86_64</arch>` breaks activation."** It's a correct latent-bug fix — macOS Intel JVMs report `os.arch=x86_64`.
- **"Maven Central publishing dropped."** Deliberate and documented (a single-host `mvn deploy` would publish a host-only jar). A tradeoff, not a defect.

### Verdict
**Approve with changes requested before this becomes the sole release path** — no blocker for the day-to-day CI path, which is sound. The Java loader and Maven native-lib wiring are correct across all 5 platforms and all three runtime contexts; the assembly split correctly delivers both libs to `lib/`; the profile rename/flip and `qdr.release` typo fix are clean. The substantive risks are release-time/operational, not data-path: the release upload is non-idempotent (#2), `package-linux` leans on an implicit runner cargo (#1), aarch64 C++ glibc coverage disappeared (#3), and there's no guard that the no-jre bundles all platforms (#5). Address #1–#5 and confirm #7/#8 operationally before the first release on this workflow.

Findings: 20 verified (8 Moderate, 12 Minor), 4 dropped as false positives.

<sub>🤖 Adversarial review generated via the review-pr skill (level 3).</sub>

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=272177)
**Tested Commit:** `fbbe2a42020604f42cf065ec37cd7ed6776f2031`

ideoma:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/Os.java|1|1|100.00%|

ideoma:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/Os.java|1|1|100.00%|

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=272335)
**Tested Commit:** `8a14756fe5ac9b09b7f28a91499ac39333924cfc`

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=272344)
**Tested Commit:** `49a7193abeb320099355b68cde9535c96567d8ce`

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=272341)
**Tested Commit:** `50de7c219fb724b5ffdf0f99016f02b67537be23`

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=272362)
**Tested Commit:** `f8f78ae05eb465122e16ac8ab3b356a36fb5b84c`

ideoma:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/Os.java|1|1|100.00%|

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=272354)
**Tested Commit:** `aab916d77be435425aca7d83dea14cfca664d9b1`

ideoma:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/Os.java|1|1|100.00%|

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=272402)
**Tested Commit:** `43b35c745899b5f3309ac77dfc1ac5716f4c6777`

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=272402)
**Tested Commit:** `43b35c745899b5f3309ac77dfc1ac5716f4c6777`

ideoma:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/Os.java|1|1|100.00%|

sklarsa:
## Review: `fix(build): package cross-platform native releases` (level 3, tandem with questdb-enterprise#1058)

Reviewed at head `c96ea4aa`. Provenance: this PR moves no submodule pointer. The enterprise PR moves the `questdb` gitlink to this head (OFF-DEFAULT, in scope as the tandem); the nested `java-questdb-client` pointer is unchanged.

CI on this head is green (Azure 272475, macwin 272519, package-only GH run 35859068506). In addition to reading the diff, I built the modified `core/Dockerfile` builder stage locally from this head and started the resulting runtime archive: it served a query with both natives loaded from `lib/`.

### Critical

**K1. Publishing the release no longer rebuilds questdb.io.** (in-diff)
- Problem: the website workflow never fires after a release.
- Net impact: every release; questdb.io stays stale until someone notices.
- Evidence: `release_website.yml` triggers only on `release: [published]`; the last six releases (9.4.0 to 10.0.1) each ran it from a `release` event; the new `publish-github` job flips the draft with `secrets.GITHUB_TOKEN`, and GitHub documents that events caused by the Actions token do not create workflow runs.

Trace: tag push -> `publish-github` -> `publish-github-release-assets.sh:54` runs `gh release edit --draft=false --latest` as `github-actions[bot]` -> `release.published` fires but starts no workflow -> `trigger-netlify-build` never runs. Base: the deleted Azure step used an Azure-stored PAT, which does trigger workflows. Suggested fix: add `workflow_dispatch:` to `release_website.yml` and, after the `gh release edit`, run `gh workflow run release_website.yml` (dispatch is one of the two documented exceptions), giving `publish-github` `actions: write`. Alternatively perform the edit with an App or PAT token.

### Moderate

**M1. Two Rust CLI binaries are packaged into questdb.jar on every from-source build.**
- Problem: `pm_inspect` and `pm_generate` land in the jar.
- Net impact: +4.1 MB in Docker, enterprise, and dev jars.
- Evidence: local Docker build of this head: `io/questdb/bin/linux-x86-64/pm_generate` (3,258,784 B) and `pm_inspect` (892,960 B) inside the jar.

rust-maven-plugin 1.2.0 copies the cdylib plus every `[[bin]]` (`Crate.getArtifactPaths`) into `copyTo`, and `Files.copy` runs unconditionally, so passing `--lib` via `extraArgs` would fail the copy. The aggregate release path is unaffected (`-DskipNative`), and the `*questdbr.*` assembly filter keeps them out of `lib/`, but the jar inside the jlink image carries them. Fix: exclude `io/questdb/bin/**/pm_*` in maven-jar-plugin, or delete them in the existing antrun at `process-classes`.

**M2. The Packer pin in `publish-ami` does not control which Packer runs.**
- Problem: apt-installed 1.10.0 is shadowed by the runner's preinstalled binary.
- Net impact: first real tag run uses Packer 1.16.0 with plugin 1.3.9, untested.
- Evidence: `actions/runner-images` installs Packer by unzipping the latest release into `/usr/local/bin`; the Ubuntu 24.04 image README lists Packer 1.16.0; `/usr/local/bin` precedes `/usr/bin` on PATH.

The job has never executed (all runs so far were package-only dispatches). Fix: call `make ... packer=/usr/bin/packer`, or use `hashicorp/setup-packer` with the pinned version and drop the apt step.

**M3. "Run tests (with sccache)" runs even after an earlier step failed.**
- Problem: the new Windows/macOS test task omits `succeeded()`.
- Net impact: CI maintainers see the real failure buried under a secondary one.
- Evidence: `ci/templates/steps.yml`, the new task's condition has no `succeeded()`, while the Linux task and the file's own comment above "Compile with Maven" explain why a bare variable condition needs it.

**M4. Central publication guards are weaker than the PR body and README claim.**
- Problem: the SNAPSHOT-dependency guard is inert where it is wired.
- Net impact: no automated dependency check before an irreversible Central publish.
- Evidence: a synthetic Maven probe with `requireReleaseDeps onlyWhenRelease=true` passes on a SNAPSHOT project with a SNAPSHOT dependency and fails only on a release version; `release:prepare` runs its preparation goals while the project is still a SNAPSHOT; the README deploy command activates only `build-web-console,include-rust-native-artifacts,maven-central-release`; `verify-central-bundle.py` is referenced only from the test script.

The snapshot rejection the lifecycle fixture observes comes from maven-release-plugin's own dependency check, which base already had. Net-neutral versus base, but "fail-closed Central checks" and "validates an exact signed bundle multiset" describe fixture-only behaviour. Fix: add `release-preparation-safety` to the README's deploy profile list, where the rule does fire (release version), and reword the description.

**M5. Several RED fixtures cannot fail for the reason they claim.**
- Problem: fixtures pass on a precondition error or an unanchored grep.
- Net impact: regressions in the verifiers would go unnoticed.
- Evidence: reproduced locally: both Central negative fixtures exit on `verified core jar is not a file: unused.jar` before the allowlist check (`test-native-release-packaging.sh:245-253`); the snapshot probe's `grep -qi 'snapshot'` matches the reactor header `Building ... -fixture-SNAPSHOT` (line 908); the GitHub "equal" mode accepts an empty call log, so a dropped `gh release edit` passes (line 558).

Fix: pass the real `valid.jar` to the two Central fixtures, grep for the release plugin's actual message, and require exactly one `edit` in the equal mode.

**M6. aarch64 old-glibc load coverage is removed while aarch64 natives still ship.**
- Problem: no check loads `linux-aarch64/libquestdb.so` on an old glibc.
- Net impact: arm64 no-jre and Docker users on RHEL 8-class glibc, if the C++ lib is ever rebuilt.
- Evidence: the deleted `glibc_smoke_test.yml` had an almalinux:8 arm64 leg; the replacement `binary-smoke-test-jobs.yml` is x86-64 only and its comment says "OSS publishes only x86 binaries", but the aggregate jar contains `linux-aarch64/libquestdb.so` (committed) and `libquestdbr.so`, and the arm64 Docker image builds natively.

Least fragile test: add an arm64 leg (hetzner-incus-arm) running the same `glibc_load_check.sh` in an almalinux:8 container.

**M7. IDE-only builds now fail with a message that blames the package.**
- Problem: `FatalError ... broken package?` after IntelliJ Rebuild Project.
- Net impact: contributors who do not delegate builds to Maven.
- Evidence: `Os.java:442` loads the Rust lib only from `/io/questdb/bin/<platform>/`, which now exists only as a Maven compile output; `.idea/ant.xml` registers the trigger but it is opt-in.

Since `Os.java` is in the diff, make the message say the Rust library is built by Maven and point at `mvn -pl core compile` or the IntelliJ trigger.

### Enterprise #1058
Reviewed separately; the enterprise findings and its own verdict (approve with comments) are posted on questdb/questdb-enterprise#1058. Nothing there affects the verdict here.

### Minor
- `is.rust.native.artifacts.aggregated` is set by the profile and consumed nowhere, so the "forged marker" negative test protects nothing; the Ant `has.exact.rust.native.artifacts` property is the one a user property could pre-set.
- Windows provenance `rust_versions.windows_x64` keeps `\r` characters from `Out-File`.
- The snapshot-negative `release:prepare` probe runs before the SCM URL is rewritten to the local bare remote, so its "no push" oracle watches the wrong remote.
- `OsTest`'s host-Rust assertion cannot fail independently of `Os` class init; the verifiers' checksum-mismatch branches have no fixture.
- `github_run_action.py` still lists the deleted `rebuild_rust` workflow.
- `ci/ami-market-pipeline.yml` (manual) now fails instead of overwriting an existing AMI because the packer force flags default to false; consistent with intent but a silent change to that entry point.
- Carried over from earlier reviews and still present: dead `NO_RUST_PATH` in `core/Dockerfile`; `-DskipNative=false` disables the Rust build; toolchain hardcoded in four workflows; `package-linux` relies on preinstalled cargo; committed `THIRD_PARTY_LICENSES.txt` drifts in local builds.

### Adjacent (pre-existing, not against this PR)
`ci/templates/aux-job.yml:15` reads `check_changes.SOURCE_CODE_CHANGED`, but the producing step is named `check_coverage` (`check-changes-job.yml:118`). The lint job's compat "Compile with Maven" and "Run tests" steps therefore show `skipped` in all 12 recent builds I surveyed, on master and this branch alike; the comment in `hosted-jobs.yml:98` claiming compat runs in that stage is stale. Compat tests still run in the self-hosted Other shard, so no coverage is lost. Worth its own issue.

### Coverage
Test gate passes: no Critical coverage gap. Admitted Moderate gap: M6. Accepted gaps: the Windows `lib/` runtime-load path is not started by any job (low risk, absolute-path `System.load` with jar fallback); `publish-github` and `publish-ami` can only execute on a real tag (K1 and M2 are the static defects found there); the Docker image path has no PR-level run, but the local build above passed and the master nightly runs before any release.

### Summary
- **Verdict: request changes** on K1. Native loading on all platforms, Maven profile mechanics, assembly filtering, artifact staging and verification, cargo-deny checksums, the Packer plugin version, and the Docker path all checked out; the release workflow's publish jobs are the untested surface and hold both K1 and M2.
- Split: all findings in-diff except K1, which is out-of-diff breakage of an unchanged workflow by a contract this PR changed (who publishes the release).
- Tradeoffs worth stating in the body: every local and downstream build now compiles the Rust crate in release mode; `release:prepare` now needs the pinned nightly toolchain on the operator's machine; the release-critical `native-release-lifecycle` job adds network dependencies on repo1.maven.org and GPG key generation.
- Operator actions outside the repo, not verifiable here: disable the Azure pipeline definition that pointed at the deleted `ci/github-release-pipeline.yml`, or it errors on the next tag; confirm the `AWS_*` repository secrets exist for `publish-ami`.

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=272706)
**Tested Commit:** `673a64b1a89431d58ce24c4146350fc595236266`

ideoma:
#### [PR Coverage check]

:disappointed: **fail** : 1 / 3 (33.33%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/Os.java|1|3|33.33%|

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=272712)
**Tested Commit:** `f3c8f7c657b090fc97a38b920df78f86473bc078`

sklarsa:
## Re-review (level 3) after the fix commits

Reviewed at head `40a8e75b`, with one follow-up commit after this pass (commits `0e2c949e`, `673a64b1`, `f3c8f7c6`, `40a8e75b`, and the CONTRIBUTING follow-up on top of `c96ea4aa`). Provenance unchanged: this PR moves no submodule pointer.

All findings from the first review are closed on this head. The re-review found three defects the fix commits themselves introduced, all now fixed in the last two commits:

- **Lifecycle fixture broke on the new POM comment.** The `requireReleaseDeps` comment in `core/pom.xml` contained the upper-case snapshot token that the release-version probe scans for; package-only run 35940472455 failed in `native-release-lifecycle` with `fixture left a SNAPSHOT value`. Reworded in `f3c8f7c6`. Package-only run 35942112573 on `40a8e75b` is green through `package-windows`.
- **README deploy command activated a profile that is not in the reactor.** `release-preparation-safety` lives in the root POM; `-pl core` builds a reactor with only core (no parent), so Maven only warned. Removed from the README and the fixture; the `requireReleaseDeps` rule inside `maven-central-release` is what fires on the tag's release-version POM, and a probe on this head shows it rejecting a SNAPSHOT client pin.
- **Website dispatch was a hard dependency of AMI publication.** A transient failure of `gh workflow run` inside `publish-github` would have skipped `publish-ami` after the release was already public. The dispatch now lives in its own `publish-website` job that nothing depends on; a failure stays visible and can be re-run alone. The `package-linux` toolchain step also sets `rustflags: ""` so the verify-step build does not inherit the action's default `-D warnings`.

Evidence gathered on this head:
- Docker build of `core/Dockerfile` from `673a64b1`: Maven log shows `rust-maven-plugin:build (qdbr-build)` followed by `antrun:run (remove-rust-cli-binaries)` in the same phase; the jar and `target/classes` hold only `libquestdb.so` and `libquestdbr.so`, no `pm_*`.
- macwin 272715: both smoke legs green, including the new arm64 leg (from-source aarch64 build, almalinux:8 glibc check, packaged server answered `SELECT 1`). The macwin pipeline needed authorization to use the `hetzner-incus-arm` pool; that is now granted in Azure DevOps (queue 81, pipeline 46).
- Package-only run 35942112573 on `40a8e75b`: lifecycle probes, all four Rust builds, `package-linux`, `package-windows` green; publish jobs skipped as designed.
- Fixture mutation checks: removing the antrun execution, the jar exclusion, the website job, or reverting the Central negatives to the missing jar each fails the fixture with a distinct message.

Remaining, non-blocking:
- The Os.java diagnostic branch for a missing Rust library was reverted in the follow-up commit: it runs only when the resource is absent, which no in-process test can reach, so the PR coverage check flagged its lines. The guidance now lives in CONTRIBUTING.md instead.
- `publish-github`, `publish-website`, and `publish-ami` still execute only on a real tag push; their first real run is the next release.

**Verdict: approve.** No open Critical or Moderate findings; test gate passes.

ideoma:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/Os.java|1|1|100.00%|

ideoma:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/Os.java|1|1|100.00%|

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=272883)
**Tested Commit:** `a4d8a393aaf980bc339ba980fbe84af54f4435f2`

PR #6892, opened 2026-03-20
chore(core): speed up WalTableWriterFuzzTest.testRandomInOutOfOrderOverlappingInserts()

glasstiger:
- Reduce iteration count (20 → 10) and max row count (10000 → 2000) in `testRandomInOutOfOrderOverlappingInserts()` to speed up the test
- Convert `assertSql` string concatenations to text blocks

----------
This test took 25-30 seconds to run locally, that usually translates to 10-15 minutes on CI.
It timed out on my PR, the test timeout is 20 minutes.
Now it finishes in 2-3 seconds locally, I expect that to be around 2 minutes on CI.


glasstiger:
@coderabbit review

PR #6901, opened 2026-03-25
chore(core): deprecate unused UPDATE cache properties

glasstiger:
## Summary

Deprecate the unused PG wire UPDATE statement cache configuration properties (`pg.update.cache.enabled`, `pg.update.cache.block.count`, `pg.update.cache.row.count`).

These properties are not used anymore. Instead of removing them outright, they are registered as deprecated so that existing configurations using them won't trigger unknown-property warnings.

### Changes
- Remove `getUpdateCacheBlockCount()`, `getUpdateCacheRowCount()`, `isUpdateCacheEnabled()` from `PGConfiguration`, `DefaultPGConfiguration`, `PGConfigurationWrapper`
- Remove corresponding fields, getters, and config loading from `PropServerConfiguration`
- Register the three property keys as deprecated via `registerDeprecated()`
- Update tests and test config files

----------

required by https://github.com/questdb/questdb-enterprise/pull/957


core/src/main/java/io/questdb/PropertyKey.java
```diff
@@ -505,9 +505,6 @@ public enum PropertyKey implements ConfigPropertyKey {
     TELEMETRY_TABLE_TTL_WEEKS("telemetry.table.ttl.weeks"),
     TELEMETRY_EVENT_THROTTLE_INTERVAL("telemetry.event.throttle.interval"),
     TELEMETRY_DB_SIZE_ESTIMATE_TIMEOUT("telemetry.db.size.estimate.timeout"),
-    PG_UPDATE_CACHE_ENABLED("pg.update.cache.enabled"),
```
puzpuzpuz, line 508: Shouldn't we `registerDeprecated` these properties in `PropServerConfiguration`? We usually do that to avoid breaking users who may have the props in their `server.conf` files.

glasstiger: yes, that makes sense.
removing the properties completely also breaks restoring from snapshot/backup.

will change it.
thank you.

glasstiger: fixed in https://github.com/questdb/questdb/pull/6901/commits/b15287c5f4aa3671d9fe54a05c9d60d49df2237a

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/PropServerConfiguration.java|3|3|100.00%|

PR #6903, opened 2026-03-26
fix(core): fix partition and column file removal during backup checkpoint

glasstiger:
## Summary

When a backup checkpoint is in progress, the backup reads partition data and column files from the live database directory. If partition directories or column version files are deleted while the checkpoint is snapshotting metadata, the backup fails because it tries to read files that no longer exist.

This PR adds checkpoint guards to all synchronous and asynchronous file removal paths to prevent deletion while a checkpoint is active:

## Changes

- **`TableWriter.processPartitionRemoveCandidates0`**: Check `isInProgress()` and defer old partition directory removal to async purge when a checkpoint is active. Rollback orphans (`txn >= lastCommittedTxn`) bypass the guard since they are not in any committed snapshot.
- **`TableWriter.finishColumnPurge`**: Force async-only column version purge during checkpoint to prevent synchronous deletion of `.d`/`.i` files.
- **`ConvertOperatorImpl.finishColumnConversion` / `UpdateOperatorImpl.update`**: Same checkpoint guard for column purge triggered by ALTER COLUMN CONVERT and UPDATE operations.
- **`O3PartitionPurgeJob.processPartition0` / `processDetachedPartition`**: Skip partition deletion in the async purge job while a checkpoint is in progress, closing the race where the purge job runs before the checkpoint pins a table's scoreboard.
- **`ColumnPurgeOperator.purge0`**: Skip async column file deletion during checkpoint, same pattern as `O3PartitionPurgeJob`.
- **`TableWriter.isCheckpointInProgress`**: New public method exposing checkpoint status for external callers (`ConvertOperatorImpl`, `UpdateOperatorImpl`).

## Test plan

- Regression test: table created after checkpoint started, O3 inserts verify old partition versions survive until checkpoint release (BYPASS WAL)
- Same scenario for WAL tables via `drainWalQueue()`
- Rollback orphan cleanup still works during active checkpoint (fake `.999` directory is removed)
- Existing `testCheckpointDoesNotBlockPurge` updated to verify async purge job also preserves partitions during checkpoint


bluestreak01:
@glasstiger

## PR #6903 Review: `fix(core): defer partition removal during backup checkpoint`

### Critical

**1. `finishColumnPurge()` has the same race condition but no checkpoint guard**

`core/src/main/java/io/questdb/cairo/TableWriter.java:5508-5523`

The PR fixes `processPartitionRemoveCandidates0()` to defer **partition directory** removal during checkpoint, but `finishColumnPurge()` has the identical race for **column version files** and is not patched:

```java
private void finishColumnPurge() {
    if (purgingOperator == null) {
        return;
    }
    boolean asyncOnly = checkScoreboardHasReadersBeforeLastCommittedTxn();
    // ^^^ No checkpoint check! Same race window as partition removal.
    purgingOperator.purge(..., asyncOnly, ...);
}
```

**Code path trace:**
1. Checkpoint starts, copies `_cv` (column version) metadata at `DatabaseCheckpointAgent.java:442-444` for each table it processes.
2. For a table not yet pinned by the checkpoint (or created after checkpoint started), `checkScoreboardHasReadersBeforeLastCommittedTxn()` returns `false`.
3. `asyncOnly=false` causes `PurgingOperator.purge()` to synchronously delete column data files (`.d`, `.i`) at `PurgingOperator.java:124-154`.
4. Backup reads the live directory using the snapshotted `_cv` file, references column versions that no longer exist on disk.

**Fix:** Add the same guard:
```java
boolean asyncOnly = checkScoreboardHasReadersBeforeLastCommittedTxn()
                    || engine.getCheckpointStatus().isInProgress();
```

---

### Moderate

**2. `O3PartitionPurgeJob` has no checkpoint awareness — deferred partitions can still be deleted**

`core/src/main/java/io/questdb/cairo/O3PartitionPurgeJob.java`

The PR defers partition removal to the async `O3PartitionPurgeJob`, but the purge job itself never checks `engine.getCheckpointStatus().isInProgress()`. It only checks `txnScoreboard.isRangeAvailable()` (lines 262, 355-356). For tables whose scoreboard has not been pinned by the checkpoint, `isRangeAvailable()` returns `true` and `purgePartition()` at line 417-423 deletes the directory unconditionally via `ff.unlinkOrRemove()`.

This means the PR's deferral buys time but doesn't guarantee the partition survives. If the purge job runs before the checkpoint iterates to the table and pins its scoreboard, the partition is deleted anyway. The window is narrow but real, especially under heavy O3 workloads.

The test acknowledges this at line 296-299:
```java
// (The purge job may or may not remove .1 depending on scoreboard implementation;
// the key assertion is .1 survived processPartitionRemoveCandidates0 above.)
```

This is an improvement over the status quo (immediate synchronous deletion) but is not a complete fix for the stated problem.

**3. Missing test: rollback cleanup during active checkpoint**

The `txn >= lastCommittedTxn` branch at line 7753 still performs synchronous removal during checkpoint. This is intentional (rollback leftovers are never referenced by any reader), but there is no test verifying this behavior is correct under checkpoint. A test confirming that rollback orphans ARE cleaned up during checkpoint would document this design decision and prevent future regressions.

**4. Missing test: WAL tables**

The regression test uses `BYPASS WAL` (line 272). WAL tables may follow different partition removal paths. No test covers the checkpoint + partition removal race for WAL tables.

---

### Minor

**5. PR title does not repeat the verb per CLAUDE.md convention**

Title is `fix(core): defer partition removal during backup checkpoint`. Per CLAUDE.md, `fix(...)` titles should repeat "fix" in the description: e.g., `fix(core): fix partition removal during backup checkpoint`. The word "defer" describes the implementation, not the user-facing impact.

**6. PR has no labels**

Should have at least `Bug` and `Core`.

**7. Commit messages use Conventional Commits prefixes**

Commit `baa98adeb2` is titled `fix(core): fix occasional checkpoint/backup failure due to O3 race`. Per CLAUDE.md, commit titles should NOT use Conventional Commits prefixes — that format is reserved for PR titles only. Also, commit `4e6ae06c06` titled "fix test" has no body description.

**8. Missing comment explaining why `txn >= lastCommittedTxn` skips the checkpoint guard**

The asymmetry in the condition at line 7753 is non-obvious. A comment explaining why rollback orphans (`txn >= lastCommittedTxn`) are safe to delete during checkpoint would prevent future confusion:
```java
// Rollback orphans (txn >= lastCommittedTxn) are not in any txn snapshot,
// so no checkpoint or reader can reference them — safe to remove immediately.
```

---

- **"Queue saturation risk from deferred partitions"** — Dismissed. The async purge queue (1024 slots) and its queue-full log-and-drop behavior is a pre-existing design. Before this PR, `anyReadersBeforeCommittedTxn=true` already triggered the same deferral path. The checkpoint condition adds one more trigger but doesn't fundamentally change the risk profile.

- **"TOCTOU race on `isInProgress()` read"** — Dismissed. `DatabaseCheckpointAgent.startedAtTimestamp` is an `AtomicLong` (line 90 of `DatabaseCheckpointAgent.java`), providing volatile read/write semantics. The `isInProgress()` call is safe. The theoretical TOCTOU window (checkpoint releases between check and use) is benign — it causes extra deferral, not premature deletion.

---

### Summary

**Verdict: Request changes.**

The core logic change in `processPartitionRemoveCandidates0()` is correct — the boolean logic is sound and the checkpoint guard prevents synchronous partition deletion during backup. However, the PR has one critical gap: `finishColumnPurge()` at line 5508-5523 has the exact same race condition for column version files and needs the same fix. The `O3PartitionPurgeJob`'s lack of checkpoint awareness makes the deferral incomplete but is a pre-existing issue outside the scope of this fix.

**Draft findings: 8 reviewed, 2 downgraded as false positives, 6 verified findings remain (1 critical, 3 moderate, 4 minor).**

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=223634)
**Tested Commit:** `677a13a50aedf5a54571ca5d8dc91e110b10fab4`

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/O3PartitionPurgeJob.java|5|5|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=223647)
**Tested Commit:** `85fbfcd7f43c37c7ad3a94c5643d517a4a2f78e2`

glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/O3PartitionPurgeJob.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/ConvertOperatorImpl.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/ColumnPurgeOperator.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/UpdateOperatorImpl.java|2|2|100.00%|

PR #6613, opened 2026-01-08
chore(core): fix metadata cache trying to load view metadata on startup

glasstiger:
## Summary

- Prevent metadata cache from trying to load view metadata from `_meta` file (views don't have one)
- View metadata is now hydrated when views are compiled at startup via a dedicated background thread
- Remove 4-year-old no-op migration (427/Mig702) to improve startup time
- Reduce startup/shutdown log noise by demoting various INFO messages to DEBUG

## Changes

- **MetadataCache**: Skip views in `hydrateTableStartup()` since views don't have `_meta` files
- **ServerMain**: Add `compileViewsThread` to compile all views at startup, which initializes view state and hydrates metadata cache
- **ViewCompilerJob**: Refactor compilation methods to static to allow reuse from startup thread; add `compileAllViews()` method
- **CairoEngine**: Remove `enqueueCompile()` call from `buildViewGraphs()` - compilation now happens in dedicated startup thread
- **PropServerConfiguration**: Increase default `viewCompilerWorkerCount` from 1 to 2
- **Logging**: Demote various worker pool, table writer, and pool operation logs from INFO to DEBUG

## Test plan

- [x] Existing `ViewBootstrapTest` updated to call `awaitStartup()` 
- [x] Verify views are correctly compiled and metadata is available after server startup
- [x] Verify no errors when starting server with existing views

PR #6913, opened 2026-03-30
feat(core): add local parquet metadata sidecar file for optimized query planning

### LINKED ISSUE #101: Local Parquet metadata for optimized remote storage queries
## Summary

When partitions are converted to Parquet and uploaded to object storage, QuestDB extracts and stores metadata locally. This enables the query engine to plan and optimize queries without round-trips to remote storage.

## How it works

1. Partition converts to Parquet (via storage policy)
2. Parquet file uploads to object storage (S3, GCS, Azure Blob)
3. Metadata (row group info, column statistics, schema) stays local
4. Query engine uses local metadata for planning and pruning
5. Only required data is fetched from remote storage

## Benefits

- **Faster query planning** — No network latency for metadata access
- **Intelligent pruning** — Skip row groups based on local min/max statistics
- **Reduced costs** — Fewer object storage API calls
- **Lower latency** — Metadata lookups are local disk reads, not remote fetches

## Use case

Ideal for large-scale deployments where hot data stays local and cold data tiers to object storage. Queries spanning both hot and cold data remain performant because metadata is always local.

## Parent feature

Sub-issue of #62 (Native Apache Parquet partition format)

RaphDal:
Closes https://github.com/questdb/roadmap/issues/101
Tandem https://github.com/questdb/questdb-enterprise/pull/978

## Summary

- Introduces a compact binary `_pm` sidecar file that accompanies each `data.parquet` partition file. The `_pm` file stores all metadata the query engine needs: column descriptors, QuestDB column types, per-row-group column chunk byte ranges, compression codecs, encodings, and min/max statistics.

- Replaces the JSON metadata blob previously embedded in the parquet footer's key-value section with a purpose-built binary format. The `_pm` file is the single authoritative source of partition metadata for parquet partitions.

- Enables row group pruning via min/max statistics and bloom filter offsets stored locally in `_pm`, without reading the parquet file itself.

- Lays the groundwork for cold storage: when a parquet file lives in object storage (S3, GCS, Azure Blob), the `_pm` file stays local and provides everything the query planner needs to decide which column chunks to fetch by byte range, eliminating metadata round-trips.

- Adds migration `Mig940` to generate `_pm` files for all existing parquet partitions on engine upgrade.

## Motivation

QuestDB's parquet partition support previously stored QuestDB-specific metadata (column types, column tops, symbol flags) as a JSON blob inside the parquet footer under the `"questdb"` key. This approach has several limitations:

1. **No column chunk byte ranges.** The JSON metadata does not include per-column byte offsets and lengths. To decode a specific column chunk, the engine must parse the full parquet footer (thrift-encoded) to locate byte ranges. For cold storage, this requires a network round-trip to read the footer from the remote object store before any data fetch.

2. **No row group statistics.** Min/max values per column per row group are available in the parquet footer but not extracted into any local structure. The query engine cannot prune row groups without first downloading and parsing the footer.

3. **No bloom filter references.** Bloom filter offsets and lengths live in the parquet footer. Without local access to these references, the engine cannot skip row groups for equality predicates without reading the remote file.

4. **No incremental updates.** When an O3 merge appends a new row group to an existing parquet partition, the entire parquet footer must be re-read and re-parsed to access the updated metadata.

The `_pm` file addresses all of these by storing a complete, locally accessible, binary-encoded copy of all metadata the query engine needs.

## The `_pm` binary format

All multi-byte integers are little-endian.

### File layout

```
 Offset   Section
 ───────  ──────────────────────────────────────────────
 0        HEADER (32B fixed)
            footer_offset           u64   (byte offset of the current footer;
                                          mutable on update, not covered by CRC)
            feature_flags           u64
            designated_timestamp    i32   (-1 if none)
            sorting_column_count    u32
            column_count            u32
            _reserved               u32   (must be 0; alignment padding)

          COLUMN DESCRIPTORS (32B each x column_count)
            name_offset             u64   (offset into name strings region)
            id                      i32   (QuestDB column writer index, -1 if N/A)
            col_type                i32   (QuestDB column type code)
            flags                   i32   (see Column Flags below)
            fixed_byte_len          i32   (for FIXED_LEN_BYTE_ARRAY columns)
            name_length             u32   (UTF-8 byte length)
            physical_type           u8    (parquet physical type 0-7)
            max_rep_level           u8
            max_def_level           u8
            _reserved               u8    (must be 0)

          SORTING COLUMNS (4B each x sorting_column_count)
            column_index            i32   (index into column descriptors)

          NAME STRINGS
            column names stored as raw UTF-8 bytes, referenced by
            (name_offset, name_length) from each column descriptor.
            Each name is padded to 4-byte alignment.

          HEADER FEATURE SECTIONS
            present only when specific feature flag bits are set.
            Navigated sequentially in bit order.
            Bit 0 (BLOOM_FILTERS): bloom column index table.

 ─ ─ ─ ─  ROW GROUP BLOCKS (one per row group, 8-byte aligned)

          Each block:
            num_rows                u64
            column_chunks           64B each x column_count (see below)
            [out-of-line stats]     variable, for stats > 8 bytes
            [bloom filter bitsets]  if BLOOM_FILTERS flag set and not external

 ─ ─ ─ ─  FOOTER (located via header footer_offset)

            parquet_footer_offset   u64   (offset in data.parquet)
            parquet_footer_length   u32   (parquet footer size in bytes)
            row_group_count         u32
            unused_bytes            u64   (dead space from old footers)
            prev_footer_offset      u64   (byte offset of previous footer,
                                          0 if this is the first version)
            row_group_entries       u32 each x row_group_count
                                          (block offset >> 3)
            [footer feature sections]
            crc32                   u32   (covers bytes [8, this field) —
                                          everything after footer_offset)
          TRAILER (last 4 bytes of file)
            footer_length           u32   (bytes from footer start through
                                          CRC, inclusive; used for
                                          standalone validation only)
```

A reader locates the footer by reading `footer_offset` from the header at a fixed position. The trailer at the end of the file is retained for standalone validation but is not used on the hot read path.

### Column chunk (64 bytes)

Each column chunk in a row group block stores the information needed to locate and decode a column's data in `data.parquet`:

```
 Offset  Field               Type   Description
 ──────  ──────────────────  ─────  ──────────────────────────────────
 0       codec               u8     Compression codec (0-7)
 1       encodings           u8     Bitmask of parquet encodings used
 2       stat_flags          u8     Which statistics are present/inlined
 3       stat_sizes          u8     Min/max stat byte sizes (nibble-encoded)
 4       _reserved           u32    Must be 0
 8       num_values          u64    Total values (may differ from num_rows)
 16      byte_range_start    u64    Offset of column chunk in data.parquet
 24      total_compressed    u64    Compressed byte length
 32      null_count          u64
 40      distinct_count      u64
 48      min_stat            u64    Inlined min or out-of-line reference
 56      max_stat            u64    Inlined max or out-of-line reference
```

**Statistics encoding.** For fixed-width types that fit in 8 bytes (BOOLEAN, BYTE, SHORT, CHAR, INT, LONG, FLOAT, DOUBLE, DATE, TIMESTAMP), stats are always inlined directly in the `min_stat`/`max_stat` fields. For variable-length or wider types (VARCHAR, STRING, UUID, LONG256), stats are stored out-of-line in the region after the column chunks. In that case, the `min_stat`/`max_stat` fields encode `[offset: u32 (>> 3), length: u32]` pointing to the out-of-line data.

**Stat flags (u8):**

| Bit | Name                   |
| --- | ---------------------- |
| 0   | MIN_PRESENT            |
| 1   | MIN_INLINED            |
| 2   | MIN_EXACT              |
| 3   | MAX_PRESENT            |
| 4   | MAX_INLINED            |
| 5   | MAX_EXACT              |
| 6   | DISTINCT_COUNT_PRESENT |
| 7   | NULL_COUNT_PRESENT     |

**Encoding bitmask (u8):**

| Bit | Encoding                |
| --- | ----------------------- |
| 0   | PLAIN                   |
| 1   | RLE_DICTIONARY          |
| 2   | DELTA_BINARY_PACKED     |
| 3   | DELTA_LENGTH_BYTE_ARRAY |
| 4   | DELTA_BYTE_ARRAY        |
| 5   | BYTE_STREAM_SPLIT       |

### Column flags (i32)

| Bit(s) | Name                | Description                             |
| ------ | ------------------- | --------------------------------------- |
| 0      | LOCAL_KEY_IS_GLOBAL | Symbol column: local dict equals global |
| 1      | IS_ASCII            | Varchar column: all values are ASCII    |
| 2-3    | FIELD_REPETITION    | 0=Required, 1=Optional, 2=Repeated      |
| 4      | DESCENDING          | Sorting column is in descending order   |
| 5-31   | Reserved            | Must be 0                               |

### Feature flags (u64)

The header carries a single `feature_flags` field that gates optional sections in both the header and footer. This allows additive extensions without bumping the format version.

- **Bits 0-31 (optional):** A reader ignores unknown optional bits and skips their sections.
- **Bits 32-63 (required):** A reader rejects the file if any unknown required bit is set.

Current feature bits:

| Bit | Name                   | Header section           | Footer section                   |
| --- | ---------------------- | ------------------------ | -------------------------------- |
| 0   | BLOOM_FILTERS          | Bloom column index table | Per-row-group bloom offset table |
| 1   | BLOOM_FILTERS_EXTERNAL | None                     | Offsets point into data.parquet  |
| 2   | SORTING_IS_DTS_ASC     | None                     | Implicit sort by DTS ascending   |

Bit 1 requires bit 0. Bit 2 requires `designated_timestamp >= 0`. When bit 2 is
set, `sorting_column_count` on disk is 0 and readers infer sorting by the
designated timestamp column.

## `_txn` integration

Each partition entry in `_txn` consists of 4 longs (`LONGS_PER_TX_ATTACHED_PARTITION = 4`):

| Field | Offset | Name                        | Description                                  |
| ----- | ------ | --------------------------- | -------------------------------------------- |
| 0     | +0     | PARTITION_TS                | Partition timestamp                          |
| 1     | +1     | PARTITION_MASKED_SIZE       | Row count + flags (bit 61: parquet format)   |
| 2     | +2     | PARTITION_NAME_TX           | Partition name txn                           |
| 3     | +3     | PARTITION_PARQUET_FILE_SIZE | Parquet file size in bytes, or -1 for native |

Field 3 retains its original meaning: the byte size of `data.parquet`. This preserves backward compatibility — older versions of QuestDB read field 3 to memory-map the parquet file, so changing its meaning would break rollback.

The `_pm` file size is not stored in `_txn`. Instead, readers `stat()` the `_pm` file and locate the correct footer version via the header's `footer_offset` field, using the parquet file size from field 3 as the MVCC version token (see Read path below).

`TxWriter.setPartitionParquetFormat()` writes both the parquet format bit in field 1 and the parquet file size in field 3 within a single transaction commit.

### Initial creation (convert to parquet)

When the engine converts a native partition to parquet, the Rust parquet encoder generates both `data.parquet` and `_pm` in a single pass. The encoder:

1. Writes column data as parquet row groups to `data.parquet`.
2. Builds the `_pm` file in memory via `ParquetMetaWriter`: column descriptors from the table schema, one row group block per parquet row group with column chunk metadata (byte ranges, codecs, statistics).
3. Writes the `_pm` footer with a CRC32, sets `footer_offset` in the header to point to it, and sets `prev_footer_offset` to 0 (first version).
4. The Java layer calls `TxWriter.setPartitionParquetFormat()` to record the parquet file size in `_txn` field 3.

### Incremental update (O3 merge)

When an out-of-order insert merges data into an existing parquet partition, `O3PartitionJob` uses `PartitionUpdater` to append a new row group:

1. `PartitionUpdater.of()` opens the existing `data.parquet` (read + write) and `_pm` file.
2. The new row group is appended to `data.parquet`.
3. A new row group block is appended to `_pm` after the old footer (the old footer becomes dead space tracked in the `unused_bytes` footer field).
4. A new footer is written with updated `row_group_count`, CRC32, and `prev_footer_offset` pointing to the old footer.
5. **Last**: `footer_offset` in the header is updated to point to the new footer. This is an 8-byte aligned atomic write and is the commit point for the `_pm` file.
6. `O3PartitionJob` commits the new parquet file size to `_txn` field 3.

Readers using an older `_txn` snapshot see a different parquet file size in field 3 than the latest `_pm` footer. They follow the `prev_footer_offset` chain from the header's `footer_offset` to locate the footer whose `parquet_footer_offset + parquet_footer_length + 8` matches their `_txn` parquet file size. This provides lock-free concurrent read/write access without storing `_pm` file size in `_txn`.

### Schema evolution

When the table schema changes via `ALTER TABLE ADD COLUMN` or `ALTER TABLE DROP COLUMN`, `PartitionUpdater.setTargetSchema()` receives the new schema. The Rust updater copies existing row groups, inserting null column chunks for added columns via `copyRowGroupWithNullColumns()` and dropping chunks for removed columns. The `_pm` file reflects the target schema: its column descriptors match the new schema, and column IDs link each descriptor to the corresponding writer index.

### Memory mapping

`TableReader` `stat()`s the `_pm` file and memory-maps its full size. The reader then locates the correct footer version by reading `footer_offset` from the header and walking the `prev_footer_offset` chain until finding the footer whose parquet file size matches `_txn` field 3. This avoids storing `_pm` file size in `_txn`, preserving backward compatibility for rollback.

If `footer_offset` exceeds the mapped size (rare race with a concurrent O3 merge extending the file between `stat()` and mmap), the reader remaps with a fresh `stat()`.

The parquet file size is read from `_txn` field 3, so the engine does not need to `stat()` the parquet file.

### `ParquetMetaFileReader`

A zero-allocation reader that operates directly over mmapped memory via `Unsafe` offset arithmetic. It provides:

- Column metadata: `getColumnCount()`, `getColumnId()`, `getColumnType()`, `getColumnName()`.
- Row group metadata: `getRowGroupCount()`, `getRowGroupSize()`, `getTotalRowCount()`.
- Statistics access: `getChunkMinStat()`, `getChunkMaxStat()`, `getChunkStatFlags()`.
- `ParquetRowGroupSkipper` implementation for filter pushdown (see below).

`of(addr, fileSize, parquetFileSize)` initializes the reader: it reads `footer_offset` from the header, then walks the `prev_footer_offset` chain to find the footer whose parquet file size matches the `parquetFileSize` argument (the MVCC version token from `_txn` field 3).

The first call to `canSkipRowGroup()` lazily allocates a native handle that caches the parsed `_pm` header and footer. The handle is reused across subsequent calls and freed by `close()`.

### `ParquetMetaPartitionDecoder`

Replaces footer-based `PartitionDecoder` for table partitions. Java owns all metadata via `ParquetMetaFileReader`; Rust acts as a stateless decode engine that receives explicit parameters per decode call:

- `decodeRowGroup()`: decodes a full row group given column indices, row range, and output buffers.
- `decodeRowGroupWithRowFilter()`: decodes with a row-level filter predicate.

Column type overrides (Symbol to Varchar, Varchar to VarcharSlice) come from the Java-side column type, while base types come from the `_pm` file. For cold storage, the same decode API works: the engine downloads column chunks by byte range using offsets from `_pm` and passes the data to the same Rust decoder.

## Row group pruning

`ParquetRowGroupFilter.prepareFilterList()` accepts a `ParquetMetaFileReader` and builds a filter list from SQL `WHERE` clause conditions. The filter list references column indices resolved from `_pm` column names and min/max statistics.

At scan time, `ParquetMetaFileReader.canSkipRowGroup()` evaluates the filter list against the row group's inlined statistics and returns `true` if the row group can be skipped entirely. This happens on the Java side without any JNI call for the common case of inlined fixed-width statistics; the native handle is only needed for out-of-line or complex comparisons.

Bloom filter offsets stored in `_pm` footer feature sections (when `BLOOM_FILTERS` flag is set) enable equality-predicate pruning for columns with bloom filters.

## Migration (`Mig940`)

`Mig940` runs during engine upgrade on tables with parquet partitions:

1. Reads `_meta` to determine `partitionBy` and timestamp column type.
2. Opens `_txn` for read-only access.
3. Iterates over all partition entries. For each partition with the parquet format bit set (bit 61 of field 1):
   - Reads the parquet file size from `_txn` field 3.
   - Opens `data.parquet` read-only.
   - Creates the `_pm` file.
   - Calls `ParquetMetadataWriter.generate()`, which reads the parquet footer, extracts QuestDB metadata from the `"questdb"` JSON key-value pair, and writes the complete `_pm` file. The parquet file size from `_txn` field 3 is used as the authoritative parquet file size (not `stat()` on the parquet file).

`Mig940` does **not** modify `_txn`. Field 3 retains its original meaning (parquet file size), preserving backward compatibility for rollback. The migration only generates `_pm` sidecar files.

The migration is non-destructive: original `data.parquet` files and `_txn` remain unchanged. It operates only on local partitions (no cold storage access required).

## Rollback safety

`_txn` field 3 stores the parquet file size (not the `_pm` file size). Older versions of QuestDB use field 3 to memory-map `data.parquet`, so keeping its original meaning preserves backward compatibility for rollback. Readers locate the `_pm` footer via the header's `footer_offset` field and walk the `prev_footer_offset` chain, using the parquet file size as an exact-match MVCC token.

If a user rolls back to an older version, writes data that modifies `data.parquet` (O3 merge, new partitions, etc.), and then re-upgrades, the `_pm` files become stale: the footer chain contains no footer matching the new parquet file size. `ParquetMetaFileReader.of()` detects this via exact matching and throws.

**Escape hatch:** set `cairo.repeat.migration.from.version` in `server.conf` to force `Mig940` to re-run on the next restart:

```properties
cairo.repeat.migration.from.version=427
```

This causes the migration framework to reset the migration version and re-run `Mig940`, which regenerates `_pm` files for all parquet partitions. Remove the property after the restart.

## `SHOW PARTITIONS` integration

`ShowPartitionsRecordCursorFactory` uses `ParquetMetaFileReader` to extract min/max timestamps from parquet partitions without parsing the parquet footer. It reads the designated timestamp column index from the `_pm` header, then fetches inlined min/max statistics from the first and last row group blocks.

## Test plan

- Round-trip `_pm` creation: convert native partition to parquet, verify `_pm` file is generated alongside `data.parquet` and matches expected metadata
- Multiple row groups: create a partition with multiple row groups, verify each row group block contains correct column chunk metadata
- O3 incremental update: insert out-of-order data into an existing parquet partition, verify `_pm` is updated with the new row group and old data remains accessible
- Multiple consecutive O3 merges: verify `_pm` handles repeated appends with growing `unused_bytes` from discarded footers
- Column type coverage: verify all supported QuestDB types (BOOLEAN, INT, LONG, FLOAT, DOUBLE, DATE, TIMESTAMP, STRING, VARCHAR, SYMBOL, UUID, LONG256, IPv4, BINARY) produce correct `_pm` descriptors
- Schema evolution: ADD COLUMN and DROP COLUMN between O3 merges, verify `_pm` reflects the new schema with null chunks for added columns
- Migration (`Mig940`): existing parquet partitions without `_pm` files, verify migration generates correct `_pm` without modifying `_txn`
- `ParquetMetaFileReader` edge cases: zero row groups, large row group counts, format version validation, corrupted trailer detection
- `ParquetMetaPartitionDecoder` lifecycle: verify `of()` releases pre-existing native handles, `destroy()` clears reader unconditionally
- Row group pruning: verify `canSkipRowGroup()` correctly prunes row groups based on min/max statistics for various column types and filter conditions
- `SHOW PARTITIONS`: verify min/max timestamps from `_pm` match expected values
- CRC32 integrity: verify reader rejects files with corrupted checksums
- Parquet file size derivation: verify `parquet_footer_offset + parquet_footer_length + 8` matches `_txn` field 3 parquet file size
- Footer chain MVCC: after multiple O3 merges, verify reader with older `_txn` snapshot walks `prev_footer_offset` chain to find the correct footer version
- Header `footer_offset` concurrency: verify reader correctly handles `footer_offset` exceeding mapped size (remap path)
- Detach/reattach: verify `_pm` files survive partition detach and reattach
- Rust unit tests: `cargo test --lib` for format serialization, deserialization, CRC computation, stat encoding, feature flag validation
- End-to-end Rust test (`decode_pm_e2e`): write parquet, convert to `_pm`, extract fields, decode column chunks, compare with footer-based decode


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 10422 / 11537 (90.34%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/bin/pm_generate.rs|0|75|00.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrame.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PartitionFrame.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetMetadataWriter.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileWriter.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/TxReader.java|2|3|66.67%|
|:large_blue_circle:|io/questdb/griffin/engine/table/BwdTableReaderPageFrameCursor.java|2|3|66.67%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/parquet_meta.rs|262|374|70.05%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|184|240|76.67%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/converter.rs|55|70|78.57%|
|:large_blue_circle:|qdbr/src/bin/pm_inspect.rs|320|394|81.22%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|86|105|81.90%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowPartitionsRecordCursorFactory.java|32|39|82.05%|
|:large_blue_circle:|io/questdb/cairo/mig/Mig940.java|102|124|82.26%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/writer.rs|150|181|82.87%|
|:large_blue_circle:|io/questdb/cairo/TableSnapshotRestore.java|66|80|82.50%|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileReader.java|166|198|83.84%|
|:large_blue_circle:|qdbr/src/parquet_read/mod.rs|33|39|84.62%|
|:large_blue_circle:|qdbr/src/parquet_read/meta.rs|38|45|84.44%|
|:large_blue_circle:|qdbr/parquet2/src/write/file.rs|49|57|85.96%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameAddressCache.java|6|7|85.71%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/reader.rs|317|371|85.44%|
|:large_blue_circle:|qdbr/src/parquet_metadata/reader.rs|922|1062|86.82%|
|:large_blue_circle:|qdbr/src/parquet_read/decode.rs|8|9|88.89%|
|:large_blue_circle:|qdbr/src/parquet_metadata/footer.rs|248|276|89.86%|
|:large_blue_circle:|qdbr/src/parquet_metadata/writer.rs|507|555|91.35%|
|:large_blue_circle:|qdbr/src/parquet_read/decode_column.rs|557|609|91.46%|
|:large_blue_circle:|qdbr/src/parquet_write/update.rs|164|177|92.66%|
|:large_blue_circle:|qdbr/src/parquet_metadata/convert.rs|1925|2068|93.09%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetMetaPartitionDecoder.java|61|65|93.85%|
|:large_blue_circle:|qdbr/src/parquet_write/jni.rs|136|146|93.15%|
|:large_blue_circle:|io/questdb/cairo/TableReader.java|74|78|94.87%|
|:large_blue_circle:|qdbr/src/parquet_metadata/header.rs|758|794|95.47%|
|:large_blue_circle:|qdbr/src/parquet_metadata/row_group.rs|343|358|95.81%|
|:large_blue_circle:|qdbr/src/parquet_read/parquet_meta_decode.rs|514|537|95.72%|
|:large_blue_circle:|qdbr/tests/pm_inspect_e2e.rs|469|490|95.71%|
|:large_blue_circle:|qdbr/parquet2/src/read/page/slice_reader.rs|60|63|95.24%|
|:large_blue_circle:|qdbr/tests/decode_pm_e2e.rs|883|916|96.40%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/buffers.rs|110|112|98.21%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionDecoder.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/IntervalBwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FwdTableReaderPageFrameCursor.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/cairo/O3ParquetMergeContext.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/AttachDetachStatus.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/IntervalFwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|qdbr/src/parquet_metadata/error.rs|34|34|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/file.rs|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/ParquetTimestampFinder.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/std/MemoryTag.java|1|1|100.00%|
|:large_blue_circle:|qdbr/parquet2/src/write/column_chunk.rs|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionEncoder.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ParquetRowGroupFilter.java|6|6|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/mod.rs|10|10|100.00%|
|:large_blue_circle:|io/questdb/cairo/mig/EngineMigration.java|1|1|100.00%|
|:large_blue_circle:|qdbr/parquet2/src/write/row_group.rs|22|22|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ExtraNullColumnCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/TxWriter.java|6|6|100.00%|
|:large_blue_circle:|qdbr/parquet2/src/bloom_filter/read.rs|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableUtils.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/CopyExportRequestTask.java|2|2|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/converters.rs|13|13|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionUpdater.java|2|2|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/row_groups.rs|5|5|100.00%|
|:large_blue_circle:|qdb-core/src/col_type.rs|5|5|100.00%|
|:large_blue_circle:|qdbr/src/parquet_metadata/types.rs|490|490|100.00%|
|:large_blue_circle:|qdbr/src/parquet_metadata/column_chunk.rs|73|73|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|62|62|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullBwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractFullPartitionFrameCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullFwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/schema.rs|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractIntervalPartitionFrameCursor.java|2|2|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=227474)
**Tested Commit:** `2bb23378b6eedde1a63bca7189fe1a25441af051`

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 10422 / 11537 (90.34%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/bin/pm_generate.rs|0|75|00.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrame.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PartitionFrame.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetMetadataWriter.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileWriter.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/TxReader.java|2|3|66.67%|
|:large_blue_circle:|io/questdb/griffin/engine/table/BwdTableReaderPageFrameCursor.java|2|3|66.67%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/parquet_meta.rs|262|374|70.05%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|184|240|76.67%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/converter.rs|55|70|78.57%|
|:large_blue_circle:|qdbr/src/bin/pm_inspect.rs|320|394|81.22%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|86|105|81.90%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowPartitionsRecordCursorFactory.java|32|39|82.05%|
|:large_blue_circle:|io/questdb/cairo/mig/Mig940.java|102|124|82.26%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/writer.rs|150|181|82.87%|
|:large_blue_circle:|io/questdb/cairo/TableSnapshotRestore.java|66|80|82.50%|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileReader.java|166|198|83.84%|
|:large_blue_circle:|qdbr/src/parquet_read/mod.rs|33|39|84.62%|
|:large_blue_circle:|qdbr/src/parquet_read/meta.rs|38|45|84.44%|
|:large_blue_circle:|qdbr/parquet2/src/write/file.rs|49|57|85.96%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameAddressCache.java|6|7|85.71%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/reader.rs|317|371|85.44%|
|:large_blue_circle:|qdbr/src/parquet_metadata/reader.rs|922|1062|86.82%|
|:large_blue_circle:|qdbr/src/parquet_read/decode.rs|8|9|88.89%|
|:large_blue_circle:|qdbr/src/parquet_metadata/footer.rs|248|276|89.86%|
|:large_blue_circle:|qdbr/src/parquet_metadata/writer.rs|507|555|91.35%|
|:large_blue_circle:|qdbr/src/parquet_read/decode_column.rs|557|609|91.46%|
|:large_blue_circle:|qdbr/src/parquet_write/update.rs|164|177|92.66%|
|:large_blue_circle:|qdbr/src/parquet_metadata/convert.rs|1925|2068|93.09%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetMetaPartitionDecoder.java|61|65|93.85%|
|:large_blue_circle:|qdbr/src/parquet_write/jni.rs|136|146|93.15%|
|:large_blue_circle:|io/questdb/cairo/TableReader.java|74|78|94.87%|
|:large_blue_circle:|qdbr/src/parquet_metadata/header.rs|758|794|95.47%|
|:large_blue_circle:|qdbr/src/parquet_metadata/row_group.rs|343|358|95.81%|
|:large_blue_circle:|qdbr/src/parquet_read/parquet_meta_decode.rs|514|537|95.72%|
|:large_blue_circle:|qdbr/tests/pm_inspect_e2e.rs|469|490|95.71%|
|:large_blue_circle:|qdbr/parquet2/src/read/page/slice_reader.rs|60|63|95.24%|
|:large_blue_circle:|qdbr/tests/decode_pm_e2e.rs|883|916|96.40%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/buffers.rs|110|112|98.21%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionDecoder.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/IntervalBwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FwdTableReaderPageFrameCursor.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/cairo/O3ParquetMergeContext.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/AttachDetachStatus.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/IntervalFwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|qdbr/src/parquet_metadata/error.rs|34|34|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/file.rs|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/ParquetTimestampFinder.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/std/MemoryTag.java|1|1|100.00%|
|:large_blue_circle:|qdbr/parquet2/src/write/column_chunk.rs|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionEncoder.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ParquetRowGroupFilter.java|6|6|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/mod.rs|10|10|100.00%|
|:large_blue_circle:|io/questdb/cairo/mig/EngineMigration.java|1|1|100.00%|
|:large_blue_circle:|qdbr/parquet2/src/write/row_group.rs|22|22|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ExtraNullColumnCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/TxWriter.java|6|6|100.00%|
|:large_blue_circle:|qdbr/parquet2/src/bloom_filter/read.rs|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableUtils.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/CopyExportRequestTask.java|2|2|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/converters.rs|13|13|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionUpdater.java|2|2|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/row_groups.rs|5|5|100.00%|
|:large_blue_circle:|qdb-core/src/col_type.rs|5|5|100.00%|
|:large_blue_circle:|qdbr/src/parquet_metadata/types.rs|490|490|100.00%|
|:large_blue_circle:|qdbr/src/parquet_metadata/column_chunk.rs|73|73|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|62|62|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullBwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractFullPartitionFrameCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullFwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/schema.rs|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractIntervalPartitionFrameCursor.java|2|2|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 4407 / 5844 (75.41%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/parquet_metadata/error.rs|0|16|00.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrame.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PartitionFrame.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetMetadataWriter.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileWriter.java|0|1|00.00%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/reader.rs|55|124|44.35%|
|:large_blue_circle:|qdbr/src/parquet_metadata/reader.rs|272|513|53.02%|
|:large_blue_circle:|qdbr/src/parquet_metadata/header.rs|253|413|61.26%|
|:large_blue_circle:|qdbr/src/parquet_metadata/row_group.rs|108|172|62.79%|
|:large_blue_circle:|qdbr/src/parquet_metadata/footer.rs|111|169|65.68%|
|:large_blue_circle:|io/questdb/cairo/TxReader.java|2|3|66.67%|
|:large_blue_circle:|io/questdb/griffin/engine/table/BwdTableReaderPageFrameCursor.java|2|3|66.67%|
|:large_blue_circle:|qdbr/src/parquet_metadata/types.rs|175|262|66.79%|
|:large_blue_circle:|qdbr/src/parquet_metadata/column_chunk.rs|19|28|67.86%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/parquet_meta.rs|242|340|71.18%|
|:large_blue_circle:|qdbr/src/parquet_metadata/convert.rs|508|680|74.71%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/converter.rs|117|152|76.97%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|185|242|76.45%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/converters.rs|10|13|76.92%|
|:large_blue_circle:|qdbr/src/parquet_read/meta.rs|35|45|77.78%|
|:large_blue_circle:|qdbr/parquet2/src/write/file.rs|45|57|78.95%|
|:large_blue_circle:|qdbr/src/parquet_metadata/writer.rs|297|369|80.49%|
|:large_blue_circle:|io/questdb/cairo/mig/Mig940.java|104|129|80.62%|
|:large_blue_circle:|qdb-core/src/col_type.rs|4|5|80.00%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|86|105|81.90%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowPartitionsRecordCursorFactory.java|32|39|82.05%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/writer.rs|150|181|82.87%|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileReader.java|169|201|84.08%|
|:large_blue_circle:|io/questdb/cairo/TableSnapshotRestore.java|75|89|84.27%|
|:large_blue_circle:|qdbr/src/parquet_read/parquet_meta_decode.rs|215|252|85.32%|
|:large_blue_circle:|qdbr/src/parquet_read/decode_column.rs|340|399|85.21%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameAddressCache.java|6|7|85.71%|
|:large_blue_circle:|qdbr/src/parquet_read/decode.rs|8|9|88.89%|
|:large_blue_circle:|qdbr/src/parquet_write/jni.rs|149|165|90.30%|
|:large_blue_circle:|qdbr/src/parquet_write/update.rs|162|175|92.57%|
|:large_blue_circle:|io/questdb/cairo/TableReader.java|73|77|94.81%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetMetaPartitionDecoder.java|63|67|94.03%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/buffers.rs|114|118|96.61%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionDecoder.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/IntervalBwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FwdTableReaderPageFrameCursor.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/cairo/O3ParquetMergeContext.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/AttachDetachStatus.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/IntervalFwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/file.rs|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/ParquetTimestampFinder.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/std/MemoryTag.java|1|1|100.00%|
|:large_blue_circle:|qdbr/parquet2/src/write/column_chunk.rs|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionEncoder.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ParquetRowGroupFilter.java|6|6|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/mod.rs|10|10|100.00%|
|:large_blue_circle:|io/questdb/cairo/mig/EngineMigration.java|1|1|100.00%|
|:large_blue_circle:|qdbr/parquet2/src/write/row_group.rs|22|22|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ExtraNullColumnCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/TxWriter.java|6|6|100.00%|
|:large_blue_circle:|qdbr/parquet2/src/bloom_filter/read.rs|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableUtils.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/CopyExportRequestTask.java|2|2|100.00%|
|:large_blue_circle:|qdbr/parquet2/src/read/page/slice_reader.rs|22|22|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionUpdater.java|2|2|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/row_groups.rs|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|62|62|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullBwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractFullPartitionFrameCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullFwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/schema.rs|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractIntervalPartitionFrameCursor.java|2|2|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=228245)
**Tested Commit:** `2005f82db49a81ab52ac464970d97fbff7b9fd2e`

core/src/main/java/io/questdb/cairo/ColumnType.java
```diff
@@ -68,7 +68,7 @@ public final class ColumnType {
     public static final int GEOSHORT_MAX_BITS = 15;
     public static final int GEOSHORT_MIN_BITS = 8;
     public static final int LEGACY_VAR_SIZE_AUX_SHL = 3;
-    public static final int MIGRATION_VERSION = 426;
+    public static final int MIGRATION_VERSION = 427;
```
ideoma: 427 was already used in past releases. It was reverted to 426 in #6613.

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=229728)
**Tested Commit:** `b865141aff2781e5d72a51059878502347313638`

ideoma:
```
Bug: cairo.repeat.migration.from.version documented value doesn't match the code's check                                                                                                                                             
                                                                                                                                                                                                                                       
  Where: core/src/main/java/io/questdb/cairo/mig/EngineMigration.java:110                                                                                                                                                              
                                                                                                                                                                                                                                       
  The check:                                                                                                                                                                                                                           
  if (currentMigrationVersion <= 0                                                                                                                                                                                                     
          || configuration.getRepeatMigrationsFromVersion() == currentTableVersion) {                                                                                                                                                  
      currentMigrationVersion = currentTableVersion;                                 
  }                                                                                                                                                                                                                                    
   
  The config value is compared against currentTableVersion — the int at offset 0 of _upgrade.d, which is ColumnType.VERSION = 426.                                                                                                     
                                                                                                                                                                                                                                       
  What the docs and PR description say:                                                                                                                                                                                                
                                                                                                                                                                                                                                       
  docs/parquet-metadata.md and the PR #6913 body instruct users to set:                                                                                                                                                                
                  
  cairo.repeat.migration.from.version=427                                                                                                                                                                                              
                  
  …to force Mig940 (registered at migration version 427) to re-run after a rollback-then-upgrade scenario.                                                                                                                             
   
  The mismatch:                                                                                                                                                                                                                        
                  
  427 != 426, so the condition never triggers. The currentMigrationVersion reset never happens, the engine sees the persisted _upgrade.d already at 427, and Mig940 does not re-run. The escape hatch as documented is a no-op.        
                  
  To actually re-run Mig940, the user must set:                                                                                                                                                                                        
                  
  cairo.repeat.migration.from.version=426

```

core/src/main/java/io/questdb/cairo/ColumnType.java
RaphDal: Good finding, moved the migration to version 428 in 3139e2ee003ff33dd219da2eb88766fee01b6bb3.

RaphDal:
@ideoma 
Indeed, 3139e2ee003ff33dd219da2eb88766fee01b6bb3 moves the migration to slot 428, now having the value to 427 in the configuration should trigger again the migration.
Added `testMigrateRollbackThenUpgradeViaEscapeHatch` in b111b4964b69001b9784e5e5901c81c62a9d9b79

core/src/main/java/io/questdb/griffin/engine/table/parquet/ParquetMetaPartitionDecoder.java
```diff
@@ -0,0 +1,356 @@
...
+/**
+ * Parquet partition decoder that reads metadata from the {@code _pm} sidecar file
+ * instead of parsing the parquet footer via thrift.
+ * <p>
+ * Java owns all metadata via {@link ParquetMetaFileReader}. Rust is a stateless
+ * decode engine that receives explicit parameters per decode call. This is the
+ * table partition path — the {@code read_parquet()} SQL function uses the
+ * separate {@link PartitionDecoder} which parses the parquet footer.
+ */
+public class ParquetMetaPartitionDecoder implements ParquetDecoder, QuietCloseable {
```
ideoma: nit: I find the name confusing. _pm is named in the code base as Parquet Meta File. This name looks like it reads metadata from the parquet partition file.

core/src/main/java/io/questdb/cairo/TableReader.java
```diff
@@ -861,8 +889,12 @@ private void closeIndexReader(int base, int columnIndex) {
     }
 
     private void closeParquetPartition(int partitionIndex) {
+        Misc.free(parquetMetaDecoders.getQuick(partitionIndex));
+        parquetMetaDecoders.setQuick(partitionIndex, null);
+        Misc.free(parquetMetadataPartitions.getQuick(partitionIndex));
+        parquetMetadataPartitions.setQuick(partitionIndex, NullMemoryCMR.INSTANCE);
         Misc.free(parquetPartitions.getQuick(partitionIndex));
-        Misc.free(parquetPartitionDecoders.getQuick(partitionIndex));
+        parquetPartitions.setQuick(partitionIndex, NullMemoryCMR.INSTANCE);
```
ideoma: nit: better to keep the memor instance around. When many partitions open, the reader closes some of them on going to the pool. And on activation from the pool, it will have to re-create all these object instances.

core/src/main/java/io/questdb/cairo/TableWriter.java
```diff
@@ -4377,39 +4531,49 @@ private void copyOrRebuildColumnIndexes(long partitionTimestamp, long newPartiti
                     continue;
                 }
                 final long colTop = columnVersionWriter.getColumnTop(partitionTimestamp, columnIndex);
-                if (colTop == -1 || colTop >= partitionRowCount) {
-                    continue; // column does not exist or has no data in this partition
+                if (colTop == -1) {
```
ideoma: I remember this `colTop >= partitionRowCount` fixed a fuzz test failure. Did you consciously deleted it or was it a merge conflict resolution?

core/src/main/java/io/questdb/cairo/TableWriter.java
```diff
@@ -4377,39 +4531,49 @@ private void copyOrRebuildColumnIndexes(long partitionTimestamp, long newPartiti
...
-                            indexWriter.add(nullKey, row);
-                        }
-                        for (long row = colTop; row < partitionRowCount; row++) {
-                            final int key = TableUtils.toIndexKey(Unsafe.getInt(dataAddr + (row - colTop) * Integer.BYTES));
-                            indexWriter.add(key, row);
+                        if (colTop >= partitionRowCount) {
+                            for (long row = 0; row < partitionRowCount; row++) {
+                                indexWriter.add(nullKey, row);
+                            }
+                            indexWriter.setMaxValue(partitionRowCount - 1);
```
ideoma: This look suspicious. I am not sure we keep index values as nulls for column tops, also else branch here does not add nulls for col_top count.

core/src/main/java/io/questdb/cairo/TableWriter.java
```diff
@@ -4500,6 +4664,46 @@ private void createIndexFiles(CharSequence columnName, long columnNameTxn, int i
...
+            } finally {
+                ff.close(parquetMetaFd);
+            }
+            if (!Os.isWindows() && configuration.getCommitMode() != CommitMode.NOSYNC) {
+                path.trimTo(partitionDirLen).$();
+                final long dirFd = TableUtils.openRONoCache(ff, path.$(), LOG);
+                if (dirFd != -1) {
+                    ff.fsyncAndClose(dirFd);
+                }
+            }
```
ideoma: This looks like an overkill, we don't sync files to disk in `ASYNC` mode.

core/src/main/java/io/questdb/griffin/engine/table/parquet/ParquetMetaPartitionDecoder.java
RaphDal: Indeed, it was to match the `PartitionDecoder` name but it doesn't make sense now.
What do you think of renaming `PartitionDecoder` -> `ParquetFileDecoder` and `ParquetMetaPartitionDecoder ` -> `ParquetPartitionDecoder`?

ideoma: PartitionDecoder -> ParquetFileDecoder sounds good, `ParquetMetaPartitionDecoder` to me would better be `ParquetMetaFileDecoder`

core/src/main/java/io/questdb/cairo/TableReader.java
RaphDal: Do you mean that we should keep those instances in a pool or keep them in their `ObjList` (`parquetPartitions `)?

core/src/main/java/io/questdb/cairo/TableWriter.java
```diff
@@ -3062,6 +3115,31 @@ public int switchNativePartitionWithParquet(long partitionTimestamp, long parque
...
+                // observe a parquet partition without its _pm. Without this, _txn
+                // commit below would publish a parquet partition that immediately
+                // throws on open because _pm is missing.
+                LOG.info()
+                        .$("source _pm missing, creating in new partition dir [path=").$substr(pathRootSize, path)
+                        .$(", parquetFileSize=").$(parquetFileSize)
+                        .I$();
+                other.trimTo(newPartitionDirLen);
+                createParquetMetadata(other, newPartitionDirLen, parquetFileSize);
+            }
```
ideoma, line 3238: This will work better. This method is used by cold storage, and it should not be doing any extensive IO. If a file is missing, return a flag to re-try building it.

```suggestion
            } else {
                return SWITCH_NO_PARQUET;
            }
```

core/src/main/java/io/questdb/cairo/TableWriter.java
RaphDal: This was intentional, this case is now handled afterwards: https://github.com/questdb/questdb/blob/b865141aff2781e5d72a51059878502347313638/core/src/main/java/io/questdb/cairo/TableWriter.java#L4554-L4559

core/src/main/java/io/questdb/cairo/TableWriter.java
RaphDal: Fair point

core/src/main/java/io/questdb/griffin/engine/table/parquet/ParquetMetaPartitionDecoder.java
RaphDal: `ParquetMetaFileDecoder` does imply that we decode a parquet metadata file, doesn't it?
The `Partition` in the name is here to indicates that this is a parquet partition file and thus it has the parquet metadata file as a sidecar, does that make sense?

core/src/main/java/io/questdb/cairo/TableReader.java
ideoma: We keep objects in `ObjList`

core/src/main/java/io/questdb/cairo/TableWriter.java
RaphDal: In this case we need to materialize them explicitely so that we can zeroes the column tops. With colTop being 0, `BitmapIndexFwdReader` no longer synthesizes the null values for us.
For the else branch, we do iterator `colTop` times: https://github.com/questdb/questdb/blob/b111b4964b69001b9784e5e5901c81c62a9d9b79/core/src/main/java/io/questdb/cairo/TableWriter.java#L4563-L4565

core/src/main/java/io/questdb/cairo/TableReader.java
RaphDal: Resolved in 740865a175d7de17e8bdf83adad839809ff7a96f

core/src/main/java/io/questdb/cairo/TableWriter.java
ideoma: Ok

core/src/main/java/io/questdb/cairo/TableWriter.java
ideoma: I still don't think it's a good change. `colTop >= partitionRowCount` means the column added to the partition, and there are no values in this partition at all. It's the situation when `addColumn()` method is called, we add column to metadata, add column top and that's it, no indexes are built. 

Do you have a test that exercises this branch and fails on old code?

core/src/main/java/io/questdb/griffin/engine/table/parquet/ParquetMetaPartitionDecoder.java
ideoma: There is already `ParquetMetaFileReader`, what should be the name in your new name pattern?

core/src/main/java/io/questdb/cairo/TableReader.java
ideoma: You need to fix the usage to check if the memory is not close here, instead of comparing with null memory:

```
    public long getParquetMetadataAddr(int partitionIndex) {
        MemoryCMR mem = parquetMetadataPartitions.getQuick(partitionIndex);
        return mem != null && mem != NullMemoryCMR.INSTANCE ? mem.addressOf(0) : 0;
    }

    public long getParquetMetadataSize(int partitionIndex) {
        MemoryCMR mem = parquetMetadataPartitions.getQuick(partitionIndex);
        return mem != null && mem != NullMemoryCMR.INSTANCE ? mem.size() : 0;
    }

```

core/src/main/java/io/questdb/cairo/TableWriter.java
ideoma: Ok you can keep it, since WalWriter does something similar

core/src/main/java/io/questdb/cairo/TableReader.java
RaphDal: `NullMemoryCMR` already returns 0 in `addressOf(long)` and `size()`

core/src/main/java/io/questdb/cairo/TableWriter.java
RaphDal: No, I don't have a test for it. The goal is to get rid entirely of column tops so that we can avoid having to store them in the _pm file and every column tops can be zeroed. I think it was supposed to be the case but the work was never complete?

ideoma: Parquet files should not have column tops. The column is either fully in it or fully out. `colTop >= partitionRowCount` condition means that the column does not exist in the partition, exactly the same as `colTop == -1`

core/src/main/java/io/questdb/cairo/TableReader.java
ideoma: Because you removed

```
parquetMetadataPartitions.setQuick(partitionIndex, NullMemoryCMR.INSTANCE);
```

`mem` variable from `parquetMetadataPartitions.getQuick(partitionIndex)` can be a closed instance of `MemoryCMRDetachedImpl`. By a fluke, it returns 0 as `size()` and `addressOf()` today, but it's a usage after close. A cleaner implementation would be 

```
return mem != null && mem.isOpen() ? mem.addressOf(0) : 0;
```

core/src/main/java/io/questdb/cairo/TableWriter.java
RaphDal: Indeed, reverted the changes in cd1a1697ed584207566236794edbd6288e0b0683.

ideoma (dismissed):

ideoma (dismissed):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 11052 / 12316 (89.74%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/bin/pm_generate.rs|0|76|00.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrame.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PartitionFrame.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetMetadataWriter.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionUpdater.java|2|5|40.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/BwdTableReaderPageFrameCursor.java|3|5|60.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|145|218|66.51%|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileWriter.java|2|3|66.67%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/file_decoder.rs|39|56|69.64%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|64|87|73.56%|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileReader.java|190|248|76.61%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/writer.rs|171|224|76.34%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/converter.rs|122|157|77.71%|
|:large_blue_circle:|qdb-parquet-meta/src/reader.rs|528|657|80.37%|
|:large_blue_circle:|qdbr/src/parquet_write/jni.rs|214|266|80.45%|
|:large_blue_circle:|qdbr/src/bin/pm_inspect.rs|320|394|81.22%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowPartitionsRecordCursorFactory.java|29|35|82.86%|
|:large_blue_circle:|io/questdb/cairo/mig/Mig940.java|104|126|82.54%|
|:large_blue_circle:|io/questdb/cairo/TableSnapshotRestore.java|101|121|83.47%|
|:large_blue_circle:|qdbr/src/parquet_read/mod.rs|33|39|84.62%|
|:large_blue_circle:|qdbr/src/parquet_read/meta.rs|38|45|84.44%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameAddressCache.java|6|7|85.71%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/partition_decoder.rs|410|467|87.79%|
|:large_blue_circle:|qdbr/src/parquet_write/update.rs|164|186|88.17%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FwdTableReaderPageFrameCursor.java|8|9|88.89%|
|:large_blue_circle:|qdbr/src/parquet_read/decode.rs|8|9|88.89%|
|:large_blue_circle:|qdbr/src/parquet_read/decode_column.rs|599|669|89.54%|
|:large_blue_circle:|parquet2/src/write/file.rs|49|55|89.09%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/reader.rs|346|387|89.41%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetPartitionDecoder.java|73|81|90.12%|
|:large_blue_circle:|qdb-parquet-meta/src/writer.rs|513|561|91.44%|
|:large_blue_circle:|qdb-parquet-meta/src/footer.rs|259|280|92.50%|
|:large_blue_circle:|qdbr/src/parquet_metadata/skip.rs|379|404|93.81%|
|:large_blue_circle:|qdbr/src/parquet_metadata/convert.rs|2058|2209|93.16%|
|:large_blue_circle:|io/questdb/cairo/TableReader.java|73|78|93.59%|
|:large_blue_circle:|qdbr/src/parquet_read/parquet_meta_decode.rs|538|577|93.24%|
|:large_blue_circle:|qdb-parquet-meta/src/row_group.rs|362|378|95.77%|
|:large_blue_circle:|parquet2/src/read/page/slice_reader.rs|60|63|95.24%|
|:large_blue_circle:|qdbr/tests/pm_inspect_e2e.rs|474|495|95.76%|
|:large_blue_circle:|qdb-parquet-meta/src/header.rs|768|801|95.88%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/buffers.rs|58|60|96.67%|
|:large_blue_circle:|qdbr/tests/decode_pm_e2e.rs|889|922|96.42%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/ReadParquetRecordCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/IntervalBwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetFileDecoder.java|4|4|100.00%|
|:large_blue_circle:|qdb-parquet-meta/src/error.rs|68|68|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/ReadParquetFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet/error.rs|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/O3ParquetMergeContext.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/cairo/AttachDetachStatus.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/IntervalFwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|parquet2/src/write/column_chunk.rs|6|6|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/file.rs|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/ParquetTimestampFinder.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/std/MemoryTag.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionEncoder.java|1|1|100.00%|
|:large_blue_circle:|parquet2/src/write/row_group.rs|22|22|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ParquetRowGroupFilter.java|6|6|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/mod.rs|10|10|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/HybridColumnMaterializer.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/mig/EngineMigration.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ExtraNullColumnCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableUtils.java|20|20|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/ReadParquetPageFrameCursor.java|1|1|100.00%|
|:large_blue_circle:|qdb-parquet-meta/src/column_chunk.rs|73|73|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/CopyExportRequestTask.java|2|2|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/converters.rs|13|13|100.00%|
|:large_blue_circle:|qdb-parquet-meta/src/types.rs|490|490|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/row_groups.rs|5|5|100.00%|
|:large_blue_circle:|qdb-core/src/col_type.rs|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|62|62|100.00%|
|:large_blue_circle:|parquet2/src/bloom_filter/read.rs|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullBwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractFullPartitionFrameCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullFwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractIntervalPartitionFrameCursor.java|2|2|100.00%|

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 11111 / 12398 (89.62%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/bin/pm_generate.rs|0|76|00.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrame.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PartitionFrame.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetMetadataWriter.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionUpdater.java|2|5|40.00%|
|:large_blue_circle:|io/questdb/cairo/vm/MemoryCMRImpl.java|10|18|55.56%|
|:large_blue_circle:|io/questdb/griffin/engine/table/BwdTableReaderPageFrameCursor.java|3|5|60.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|145|218|66.51%|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileWriter.java|2|3|66.67%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/file_decoder.rs|39|56|69.64%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|64|87|73.56%|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileReader.java|190|248|76.61%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/writer.rs|171|224|76.34%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/converter.rs|122|157|77.71%|
|:large_blue_circle:|qdb-parquet-meta/src/reader.rs|528|657|80.37%|
|:large_blue_circle:|io/questdb/cairo/TableSnapshotRestore.java|144|180|80.00%|
|:large_blue_circle:|qdbr/src/parquet_write/jni.rs|214|266|80.45%|
|:large_blue_circle:|qdbr/src/bin/pm_inspect.rs|320|394|81.22%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowPartitionsRecordCursorFactory.java|29|35|82.86%|
|:large_blue_circle:|io/questdb/cairo/mig/Mig940.java|104|126|82.54%|
|:large_blue_circle:|qdbr/src/parquet_read/mod.rs|33|39|84.62%|
|:large_blue_circle:|qdbr/src/parquet_read/meta.rs|38|45|84.44%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameAddressCache.java|6|7|85.71%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/partition_decoder.rs|410|467|87.79%|
|:large_blue_circle:|qdbr/src/parquet_write/update.rs|164|186|88.17%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FwdTableReaderPageFrameCursor.java|8|9|88.89%|
|:large_blue_circle:|qdbr/src/parquet_read/decode.rs|8|9|88.89%|
|:large_blue_circle:|qdbr/src/parquet_read/decode_column.rs|599|669|89.54%|
|:large_blue_circle:|parquet2/src/write/file.rs|49|55|89.09%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/reader.rs|346|387|89.41%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetPartitionDecoder.java|73|81|90.12%|
|:large_blue_circle:|qdb-parquet-meta/src/writer.rs|513|561|91.44%|
|:large_blue_circle:|qdb-parquet-meta/src/footer.rs|259|280|92.50%|
|:large_blue_circle:|qdbr/src/parquet_metadata/skip.rs|379|404|93.81%|
|:large_blue_circle:|qdbr/src/parquet_metadata/convert.rs|2058|2209|93.16%|
|:large_blue_circle:|qdbr/src/parquet_read/parquet_meta_decode.rs|538|577|93.24%|
|:large_blue_circle:|io/questdb/cairo/TableReader.java|72|76|94.74%|
|:large_blue_circle:|qdb-parquet-meta/src/row_group.rs|362|378|95.77%|
|:large_blue_circle:|parquet2/src/read/page/slice_reader.rs|60|63|95.24%|
|:large_blue_circle:|qdbr/tests/pm_inspect_e2e.rs|474|495|95.76%|
|:large_blue_circle:|qdb-parquet-meta/src/header.rs|768|801|95.88%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/buffers.rs|58|60|96.67%|
|:large_blue_circle:|qdbr/tests/decode_pm_e2e.rs|889|922|96.42%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/ReadParquetRecordCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/IntervalBwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetFileDecoder.java|4|4|100.00%|
|:large_blue_circle:|qdb-parquet-meta/src/error.rs|68|68|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/ReadParquetFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet/error.rs|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/O3ParquetMergeContext.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/cairo/AttachDetachStatus.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/IntervalFwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|parquet2/src/write/column_chunk.rs|6|6|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/file.rs|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/ParquetTimestampFinder.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/std/MemoryTag.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionEncoder.java|1|1|100.00%|
|:large_blue_circle:|parquet2/src/write/row_group.rs|22|22|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ParquetRowGroupFilter.java|6|6|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/mod.rs|10|10|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/HybridColumnMaterializer.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/mig/EngineMigration.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ExtraNullColumnCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableUtils.java|20|20|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/ReadParquetPageFrameCursor.java|1|1|100.00%|
|:large_blue_circle:|qdb-parquet-meta/src/column_chunk.rs|73|73|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/CopyExportRequestTask.java|2|2|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/converters.rs|13|13|100.00%|
|:large_blue_circle:|qdb-parquet-meta/src/types.rs|490|490|100.00%|
|:large_blue_circle:|io/questdb/cairo/vm/MemoryCMRDetachedImpl.java|7|7|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/row_groups.rs|5|5|100.00%|
|:large_blue_circle:|qdb-core/src/col_type.rs|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|62|62|100.00%|
|:large_blue_circle:|parquet2/src/bloom_filter/read.rs|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullBwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractFullPartitionFrameCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullFwdPartitionFrameCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractIntervalPartitionFrameCursor.java|2|2|100.00%|

PR #6918, opened 2026-04-01
perf(core): optimize WAL commit batching for trailing INSERT transactions

ideoma:
## Summary

- `calculateInsertTransactionBlock()` excluded the last loaded INSERT from the transaction block, forcing it to be processed alone. This fix includes it in the block, allowing consecutive INSERTs to be committed together via `processWalCommitBlock()` instead of one-by-one via `processWalCommit()`.
- The previous behavior caused unnecessary LAG usage on empty tables (e.g. post-TRUNCATE with few rows), creating artificial 0-row partitions that could race with backup compression.

## Root cause

The backward loop in `readObservableTxnMeta()` assigns `LAST_ROW_COMMIT` to the last loaded INSERT. `getCommitToTimestamp()` maps this to `FORCE_FULL_COMMIT`. The block calculation loop broke on `FORCE_FULL_COMMIT` **before** incrementing `blockSize`, excluding the transaction.

The fix does `blockSize++` before the break, including the transaction in the block. This is safe because the backward loop propagates `FORCE_FULL_COMMIT` from structural changes (ALTER, TRUNCATE) to the INSERT before them, so the block loop never reaches a non-data transaction.

## Test plan

- New test `testTruncateFollowedByTwoInserts` verifies that two INSERTs after TRUNCATE form a block of 2
- Existing `WalTxnDetailsFuzzTest` suite passes (10 tests covering commitToTimestamp and block calculation)


glasstiger:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/wal/WalTxnDetails.java|4|4|100.00%|

PR #6930, opened 2026-04-04
fix: point unclosed-paren error to opening '(' instead of FROM keyword

### LINKED ISSUE #6010: Unhelpful SQL syntax error

singhvishalkr:
## Summary

- When a function call has a missing closing ), the parser previously reported `dangling literal` at the **FROM** keyword position, which was unhelpful.
- Now reports `unbalanced (` at the **opening parenthesis** position, so the user can immediately see which `(` was never closed.

## Example

```sql
SELECT abs(100000*rnd_double(),
  timestamp_sequence('2022-07-01', 1000000)
FROM long_sequence(1000000)
```

**Before**: Error at `FROM`: `dangling literal`
**After**: Error at `abs(`: `unbalanced (`

## Changes

- **ExpressionParser.java**: In the 'dangling literal' path, when `scopeStack` has `PAREN` scope and the token is `from`, scan `opStack` for the nearest `(` control node and report `unbalanced (` at its position.
- **ExpressionParserTest.java**: Added `testUnclosedParenFromKeyword` test.
- **SqlParserTest.java**: Added `testSelectUnclosedParenPointsToOpeningParen` integration test.
- **ExtractFunctionTest.java**: Updated `testNotExtractFrom` to expect `unbalanced (` at position 16 (the opening `(`) instead of `dangling literal` at position 22 (the `from` keyword).

Fixes #6010


singhvishalkr:
Hi — this is a targeted fix for #6010. The change only fires when `scopeStack` is `PAREN` and the current token is `FROM`, so it won't affect any other error paths. Included 2 new tests and updated the one existing test whose expectation changed. Happy to adjust if you'd prefer a different error message or position strategy!

singhvishalkr:
Hi @questdb/engineering — gentle ping on this one. CI is green (26 checks passing) and it's been open for about 17 days without a reviewer picking it up. The change is small (~12 lines, parser-only) and keeps the existing error string but redirects the reported position of `PAREN_BALANCE` errors to the actual opening `(` instead of the `FROM` keyword downstream. Happy to adjust wording, add a test case, or split the change if that helps — just let me know.

singhvishalkr:
Rebased this onto current master and resolved the test conflict.

The conflict was only in ExtractFunctionTest. I kept the current assertQuery(...).fails(...) style from master and preserved the intended behavior for this fix: the parser now reports unbalanced ( at the opening parenthesis for the malformed something(... from ...) case.

Verification: git diff --check passes locally. I could not run the Maven test suite on this machine because the checkout requires Java 17 plus Maven, while this environment has Java 21 and no mvn executable.

PR #6935, opened 2026-04-05
fix(sql): fix ALTER COLUMN type conversion between DECIMAL and VARCHAR/STRING

### LINKED ISSUE #6923: ALTER COLUMN from DECIMAL to VARCHAR  (and vice versa) not supported
### To reproduce

1. CREATE TABLE TEST (time TIMESTAMP_NS, VALUE DECIMAL(20,0), STRING VARCHAR)
2. ALTER TABLE TEST ALTER COLUMN VALUE TYPE VARCHAR
3. Observe `incompatible column type change [existing=DECIMAL(20,0), new=VARCHAR`
4. ALTER TABLE TEST ALTER COLUMN VALUE STRING DECIMAL(20,0)
5. Observe `incompatible column type change [existing=VARCHAR, new=DECIMAL(20,0)`

This should be successful according to the table here: https://questdb.com/docs/query/sql/alter-table-change-column-type/

### QuestDB version:

9.3.4

Mac OS 26.3.2 (Docker Desktop)

Host File System

### Full Name:

Jason Thomas

### Affiliation:

OpenC3

- [x] Yes, I have

### Additional context

_No response_

officialasishkumar:
Fixes #6923

## Summary

`ALTER TABLE ALTER COLUMN TYPE` now supports converting between DECIMAL and VARCHAR/STRING types. The [documentation](https://questdb.com/docs/query/sql/alter-table-change-column-type/) already listed these conversions as supported, but the implementation was missing, causing an `incompatible column type change` error.

Three areas required changes:

1. **SqlCompilerImpl** — the type conversion validation matrix now includes DECIMAL<->VARCHAR/STRING entries alongside the existing DECIMAL<->DECIMAL and DECIMAL<->integer conversions.

2. **ColumnTypeConverter** — four new conversion methods handle the actual data transformation:
   - DECIMAL→VARCHAR and DECIMAL→STRING read each value through `DecimalColumnTypeConverter.Loader`, set the scale on a thread-local `Decimal256`, and write the string representation via `Decimal256.toSink()`.
   - VARCHAR→DECIMAL and STRING→DECIMAL parse each string through `Decimal256.ofString()` and store the result with `DecimalUtil.store()`.
   - NULL values round-trip correctly in both directions.
   - The follow-up fixes also correct VARCHAR aux-vector sizing in the shared fixed→VARCHAR conversion path and align the decimal→text helpers with the existing thread-local-before-mmap safety pattern.

3. **DecimalColumnTypeConverter** — `getLoader()` visibility changed from private to package-visible so `ColumnTypeConverter` can reuse the existing DECIMAL memory-reading infrastructure.

All six DECIMAL storage sizes (DECIMAL8 through DECIMAL256) are supported for both conversion directions.

## Test plan

- New tests in `AlterTableChangeColumnTypeTest`:
  - `testChangeDecimalToVarchar` — all six DECIMAL sizes to VARCHAR, with and without fractional parts
  - `testChangeDecimalToString` — DECIMAL64 to STRING
  - `testChangeDecimalToVarcharWithNull` — NULL value handling for DECIMAL→VARCHAR
  - `testChangeVarcharToDecimal` — all six DECIMAL sizes from VARCHAR
  - `testChangeStringToDecimal` — STRING to DECIMAL64
  - `testChangeVarcharToDecimalWithNull` — NULL value handling for VARCHAR→DECIMAL
  - `testChangeVarcharToDecimalRoundTrip` — DECIMAL→VARCHAR→DECIMAL round-trip preserves values and NULLs
- Existing `testChangeDecimalToDecimal`, `testChangeIntegerToDecimal`, and `testFixedToVarcharConversions` cover the related pre-existing fixed→VARCHAR path as well

### DIFF
--- a/core/src/test/java/io/questdb/test/griffin/AlterTableChangeColumnTypeTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/AlterTableChangeColumnTypeTest.java
@@ -276,6 +276,31 @@ public class AlterTableChangeColumnTypeTest extends AbstractCairoTest {
     }
 
     @Test
+    public void testChangeVarcharToDecimalWithInvalidValues() throws Exception {
+        assertMemoryLeak(() -> {
+            execute("CREATE TABLE x (ts TIMESTAMP, col VARCHAR) TIMESTAMP(ts) PARTITION BY DAY WAL", sqlExecutionContext);
+            execute("INSERT INTO x VALUES('2024-05-14T16:00:00.000000Z', '12345.6789')", sqlExecutionContext);
+            execute("INSERT INTO x VALUES('2024-05-14T16:00:01.000000Z', 'abc')", sqlExecutionContext);
+            execute("INSERT INTO x VALUES('2024-05-14T16:00:02.000000Z', '')", sqlExecutionContext);
+            execute("INSERT INTO x VALUES('2024-05-14T16:00:03.000000Z', '12.34.56')", sqlExecutionContext);
+            execute("INSERT INTO x VALUES('2024-05-14T16:00:04.000000Z', NULL)", sqlExecutionContext);
+            drainWalQueue();
+
+            execute("ALTER TABLE x ALTER COLUMN col TYPE DECIMAL(18, 4)", sqlExecutionContext);
+            drainWalQueue();
+
+            assertSql("ts\tcol\n" +
+                    "2024-05-14T16:00:00.000000Z\t12345.6789\n" +
+                    "2024-05-14T16:00:01.000000Z\t\n" +
+                    "2024-05-14T16:00:02.000000Z\t\n" +
+                    "2024-05-14T16:00:03.000000Z\t\n" +
+                    "2024-05-14T16:00:04.000000Z\t\n", "x");
+
+            execute("DROP TABLE x");
+        });
+    }
+
+    @Test
     public void testChangeVarcharToDecimalWithNull() throws Exception {
         assertMemoryLeak(() -> {
             execute("CREATE TABLE x (ts TIMESTAMP, col VARCHAR) TIMESTAMP(ts) PARTITION BY DAY WAL", sqlExecutionContext);


officialasishkumar:
Addressed the review in 62f80b118. This fixes the VARCHAR aux allocation bug, cleans the decimal text-conversion sinks, restores package-private visibility for `getLoader()`, adds the missing DECIMAL to STRING null test, and adds invalid STRING to DECIMAL coverage for `strToDecimal`. I also updated the PR title. I could not add labels from this account because the repo does not grant label-edit permissions.

bluestreak01:
@officialasishkumar Two moderate findings from review:

## 1. `mapAppendColumnBuffer` called before thread-local gets — potential native memory leak

**Files:** `ColumnTypeConverter.java` — `convertDecimalToString` (lines 1115-1119) and `convertDecimalToVarchar` (lines 1171-1175)

Both methods call `mapAppendColumnBuffer()` **first**, then obtain thread-local objects (`dstFixMemTL.get()`, `dstVarMemTL.get()`, sink TL), then enter the `try` block. If any `ThreadLocal.get()` throws (e.g., OOM during first-access supplier invocation), the mapped source memory leaks because the `finally` block is never entered.

The established pattern in the same file (`convertFixedToString` at lines 223-227 and `convertFixedToVarchar` at lines 346-350) does the thread-local gets **before** the map call:

```java
// Existing correct pattern (convertFixedToString, lines 223-227):
MemoryCMARW dstVarMem = dstVarMemTL.get();  // BEFORE map
MemoryCMARW dstFixMem = dstFixMemTL.get();  // BEFORE map
srcMapAddress = TableUtils.mapAppendColumnBuffer(...);
try {
```

```java
// New code (convertDecimalToString, lines 1115-1119):
long srcMapAddress = TableUtils.mapAppendColumnBuffer(...);  // map FIRST
MemoryCMARW dstFixMem = dstFixMemTL.get();  // AFTER map, before try
MemoryCMARW dstVarMem = dstVarMemTL.get();  // AFTER map, before try
StringSink sink = sinkUtf16TL.get();         // AFTER map, before try
try {
```

**Suggested fix:** Move the `mapAppendColumnBuffer` call after all `ThreadLocal.get()` calls in both methods, matching the existing pattern.

## 2. Pre-existing `convertFixedToVarchar` memory corruption bug fix is not documented

**File:** `ColumnTypeConverter.java:353`

The second commit (`62f80b1`) fixes a pre-existing bug in `convertFixedToVarchar` where `StringTypeDriver.INSTANCE.getAuxVectorSize(rowCount)` was allocating `(rowCount+1)*8` bytes for VARCHAR aux entries that are 16 bytes each — under-allocating by ~50% and causing buffer overruns on any fixed→VARCHAR conversion (e.g., INT→VARCHAR, DOUBLE→VARCHAR).

This is a significant memory safety fix that affects code paths unrelated to DECIMAL. It deserves explicit mention in the PR description and ideally its own regression test.

officialasishkumar:
@bluestreak01 moved the thread-local gets ahead of the mmap in both decimal-to-text paths, so the mapped source buffer is always released if thread-local initialization fails. I also updated the PR description to call out the broader VARCHAR aux sizing fix from 62f80b118. The existing fixed-to-VARCHAR conversion coverage already exercises that path, so I left the test set as-is.

bluestreak01:
@officialasishkumar Missing test for VARCHAR→DECIMAL with invalid values. `testChangeStringToDecimalWithInvalidValues` covers the STRING path, but the VARCHAR path uses different code (`VarcharTypeDriver.getSplitValue` → `asAsciiCharSequence()` vs `getStrA()`). Please add a `testChangeVarcharToDecimalWithInvalidValues` test for parity.

officialasishkumar:
@bluestreak01 Added `testChangeVarcharToDecimalWithInvalidValues` to cover the separate VARCHAR parsing path as well. Included in e0166ae79.

PR #6943, opened 2026-04-07
fix(sql): fix window functions nested inside GROUP BY expressions producing wrong results

jovfer:
## Summary

- Window functions nested inside arithmetic expressions (e.g. `avg(x) - avg(x) OVER ()`) silently produced wrong results when combined with GROUP BY. The existing check in `SqlOptimiser.rewriteSelect0` only inspected `REWRITE_STATUS` flags on top-level window columns and missed window functions buried inside operators, function calls, or CASE branches.
- Add `findWindowFunctionOutsideAggregatePos`: walks each SELECT column's AST and returns the position of the first window function that is not nested inside an aggregate. Skipping aggregate subtrees keeps shapes like `max(avg(x) OVER (...))` legal (per #6955) while rejecting unrepresentable windows mixed with GROUP BY. The error points at the offending window (e.g. `avg` at position 35) instead of the root operator or position 0.
- The walker matches both forms of window function: function nodes carrying an OVER clause (`windowExpression != null`) and pure window function names used without OVER (e.g. `row_number()`). The latter previously fell through to a cryptic runtime "empty window context".
- The check also covers implicit grouping. The pre-scan sets `REWRITE_STATUS_USE_GROUP_BY_MODEL` only for real aggregates (`checkForChildAggregates` skips window functions via `windowExpression == null`), so the extended guard fires on `SELECT sum(x) + avg(x) OVER () FROM t` (sibling aggregate triggers grouping) without producing false positives on valid queries like `SELECT (x - avg(x) OVER ()) FROM t`.
- Add `findInvalidAggregateOverWindowPos` for a related shape: a wrapper around an aggregate that contains a window function (arithmetic, function call, CASE branch). PR #6955 wired up the rewriter for the case where the aggregate sits at the column root; the extractor recurses through any non-window subtree underneath that root, so `max(avg(x) OVER ())` and `max(abs(avg(x) OVER ()))` both work. What does NOT work is wrapping the aggregate itself: `max(avg(x) OVER ()) - 1` or `max(avg(x) OVER ()) + min(avg(x) OVER ())` falls back to a code path that raises a misleading "Aggregate function cannot be passed as an argument" error pointing at the inner window. The new guard rejects those upfront with "Aggregate over window function cannot be combined with other terms. Use a sub-query." pointing at the offending aggregate.
- Tighten the sentinel of the existing helper `findWindowFunctionOrNamePosition` from `0` to `-1` so a hit at position 0 is no longer mistaken for "not found". All current callers gate on `checkForWindowFunctionsOrNames` first, so this is a latent-bug fix with no observable change in current call sites.

## Known limitations

This PR rejects shapes that previously produced silently wrong output or cryptic errors. Two related cases stay rejected and are documented here as known limitations the planner has yet to grow into:

- `SELECT col - agg(window(x)) FROM t GROUP BY col` (window outside any aggregate, mixed with GROUP BY) is rejected with "Window function is not allowed in context of aggregation. Use sub-query." PostgreSQL accepts this and applies the window post-aggregation when the bare argument is a grouped column. QuestDB's planner does not yet support that pattern. Workaround: compute the window in a sub-query.
- Wrapping `agg(window(x))` itself in any operator, function call, or CASE branch (e.g. `max(avg(x) OVER ()) + 1`, `max(avg(x) OVER ()) - min(avg(x) OVER ())`) is rejected with "Aggregate over window function cannot be combined with other terms. Use a sub-query." The rewriter recurses through subtrees inside the root aggregate but does not handle wrapping outside it (#6955). Workaround: split into separate columns or compute the window inside a sub-query and aggregate over it.

## Test plan

New negative tests added to `testNestedWindowFunctionInGroupByContextFails`:

- Nested window inside arithmetic: `SELECT cat, avg(x), (avg(x) - avg(x) OVER ()) FROM t GROUP BY cat`.
- Deeper nesting: `SELECT cat, 1 + (2 + avg(x) OVER ()) FROM t GROUP BY cat`.
- CASE branch: `SELECT cat, CASE WHEN avg(x) > 0 THEN avg(x) OVER () END FROM t GROUP BY cat`.
- Single-argument wrapper: `SELECT cat, abs(avg(x) OVER ()) FROM t GROUP BY cat`.
- Implicit GROUP BY: `SELECT sum(x) + avg(x) OVER () FROM t`.
- Pure window-function name without OVER: `SELECT sum(x) + row_number() FROM t`.
- ORDER BY lifted into bottom-up columns: `SELECT cat, avg(x) FROM t GROUP BY cat ORDER BY avg(x) - avg(x) OVER ()`.
- Mixed agg-wrapped + bare window: `SELECT cat, max(avg(x) OVER ()) - avg(x) OVER () FROM t GROUP BY cat`.
- GROUP BY ordinal resolving to a window-bearing column: `SELECT x - avg(x) OVER () FROM t GROUP BY 1` (caught by pre-existing `validateGroupByExpression`).

New positive tests in the same method:

- Sub-query workaround: `SELECT cat, max(w) FROM (SELECT cat, avg(x) OVER () AS w FROM t) GROUP BY cat`.
- No-GROUP-BY regression: `SELECT (x - avg(x) OVER ()) AS diff FROM t` still works.

New test method `testAggregateOverWindowInExpressionFails` covering the aggregate-over-window guard:

- `max(agg(window)) - min(agg(window))` with GROUP BY and the same shape without GROUP BY (implicit grouping).
- `max(agg(window)) + literal`.
- `max(agg(window)) + agg(col)`.
- Aggregate-wrapped window inside a CASE THEN branch.
- Positive regression guards: bare `max(avg(x) OVER ())`, two separate columns each with `agg(window)`, plain aggregate arithmetic without any window.

`pivot_errors.test` updated:

- `test PIVOT (sum(41) over() FOR j IN ('a', 'b'))` previously returned a structurally-meaningless `NULL NULL` from the empty test table; converted to a `query error` block expecting the new compile-time rejection. The neighbouring `sum(sum(41) over())` case still passes since the walker skips aggregate subtrees.


bluestreak01 (changes requested):
@jovfer — review findings below.

## Moderate

### 1. Code duplication with `checkForChildWindowFunctions` — `SqlOptimiser.java:10876-10892`

The new `hasNestedWindowExpression` is functionally identical to the existing `checkForChildWindowFunctions` at line 10900. Both walk the AST checking `node.windowExpression != null`, with the same `paramCount < 3` branching logic. The only difference is recursive vs. iterative implementation.

Since the call site at line 9083 is in an instance method of `SqlOptimiser`, the instance field `sqlNodeStack` (line 193) is available. The new method should be removed and replaced with a call to the existing one:

```java
if (explicitGroupBy && !qc.isWindowExpression() && checkForChildWindowFunctions(sqlNodeStack, qc.getAst())) {
```

This eliminates duplication and avoids introducing a recursive traversal where the codebase already uses an iterative stack-based approach for the same logic.

### 2. Method placement violates member ordering — `SqlOptimiser.java:10876`

The new method is `private static` but is placed among instance methods (between `wrapWithSelectModel` at line 10837 and `wrapWithSelectWildcard` at line 10894). Per CLAUDE.md, static members should be grouped with other statics and sorted alphabetically. The other `private static` methods live at lines 431–785; `hasNestedWindowExpression` belongs between `extractAndTerms` (431) and `isCompileTimeConstant` (444). Moot if finding 1 is addressed.

## Minor

### 3. `node.args.size()` vs `node.paramCount` — `SqlOptimiser.java:10886`

The new method loops over `node.args.size()`, while the existing `checkForChildWindowFunctions` at line 10926 uses `node.paramCount`. These should match in practice, but `paramCount` is the canonical bound per the ExpressionNode invariants (line 64–68). Moot if finding 1 is addressed.

### 4. Test uses multiple INSERT statements — `WindowFunctionTest.java:8217-8219`

Per coding conventions: "use a single INSERT statement to insert multiple rows." The three separate INSERTs should be consolidated:

```java
execute("INSERT INTO t VALUES (1, 'A', 1000000), (2, 'A', 2000000), (3, 'B', 3000000)");
```

### 5. Error position points at the root expression, not the window function — `SqlOptimiser.java:9084`

`qc.getAst().position` yields position 33 (the `-` operator), whereas the actual offending character is the window function `avg` at position 35. The convention is that the position should point at the specific offending character. The existing top-level check at line 9253 uses hardcoded position 0, so position 33 is still an improvement — this is a nit.

### 6. Limited test coverage — `WindowFunctionTest.java:8214-8229`

Only one query shape is tested. Consider adding:
- Deeper nesting: `SELECT cat, 1 + (2 + avg(x) OVER ()) FROM t GROUP BY cat` — verifies the AST walk recurses correctly
- CASE expression: `SELECT cat, CASE WHEN avg(x) > 0 THEN avg(x) OVER () END FROM t GROUP BY cat` — exercises the `paramCount >= 3` branch
- Positive regression test: a query with nested window functions WITHOUT GROUP BY still works (e.g., `SELECT (avg(x) - avg(x) OVER ()) FROM t`) — guards against the check accidentally rejecting valid queries

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|2|2|100.00%|

questdb-butler:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Fails</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Please update the PR title to match this format:
`type(subType): description`

Where `type` is one of:
`feat`, `fix`, `chore`, `docs`, `style`, `refactor`, `perf`, `test`, `ci`, `revert`

And: `subType` is one of:
`build`, `sql`, `log`, `mig`, `core`, `ilp`, `pgwire`, `http`, `conf`, `ui`, `wal`, `parquet`, `utils`

For Example:

```
perf(sql): improve pattern matching performance for SELECT sub-queries
```
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 94e22a47bc7e26e9859879dabeb6844d15afe751
</p>

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=226687)
**Tested Commit:** `94e22a47bc7e26e9859879dabeb6844d15afe751`

bluestreak01 (approved):
Verified the doc-comment and PR-description fixes from the review. Approved.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 59 / 61 (96.72%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|59|61|96.72%|

PR #6952, opened 2026-04-09
perf(sql): speed up simple single column GROUP BY queries

puzpuzpuz:
WIP (evening experiment)

```sql
-- Rosti 27ms
-- master's Java path 32ms
-- patch's Java path 16ms
SELECT RegionID FROM hits GROUP BY RegionID;

-- Rosti 27ms
-- master's Java path 48ms
-- patch's Java path 18ms
SELECT RegionID, count() FROM hits GROUP BY RegionID;

-- master 317ms
-- patch 276ms (268ms with faster hash function)
SELECT count_distinct(UserID) FROM hits;

-- master 446ms
-- patch 313ms (307ms with faster hash function)
SELECT UserID, COUNT(*) AS c FROM hits GROUP BY UserID ORDER BY c DESC LIMIT 10;

-- master 290ms
-- patch 220ms
SELECT ClientIP, ClientIP - 1, ClientIP - 2, ClientIP - 3, COUNT(*) AS c FROM hits GROUP BY ClientIP, ClientIP - 1, ClientIP - 2, ClientIP - 3 ORDER BY c DESC LIMIT 10;
```

### DIFF
--- a/core/src/main/java/io/questdb/cairo/map/Map.java
+++ b/core/src/main/java/io/questdb/cairo/map/Map.java
@@ -25,6 +25,7 @@
 package io.questdb.cairo.map;
 
 import io.questdb.cairo.Reopenable;
+import io.questdb.griffin.engine.table.GroupByMapFragment;
 import io.questdb.std.Mutable;
 import org.jetbrains.annotations.TestOnly;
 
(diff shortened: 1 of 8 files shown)


puzpuzpuz:
Closed this one since the speed up is modest after #6959

PR #6955, opened 2026-04-10
fix(sql): fix nested window inside aggregate with explicit GROUP BY

### LINKED ISSUE #6954: window function inside aggregate argument failing with explicit GROUP BY
## Summary

Window functions used as arguments to aggregate functions work with implicit grouping but fail with explicit GROUP BY, producing a confusing "Invalid column" error.

## Reproducer

```sql
-- Works (implicit grouping):
SELECT symbol, sum(row_number() OVER ()) FROM trades WHERE timestamp IN '2025-01-01'

-- Fails with "[0] Invalid column: symbol":
SELECT symbol, sum(row_number() OVER ()) FROM trades WHERE timestamp IN '2025-01-01' GROUP BY symbol
```

Both queries are valid SQL. The optimiser correctly extracts the nested window function to a `select-window` layer when there is no GROUP BY clause, but fails when one is present.

## Root cause

`extractAndRegisterNestedWindowFunctions` creates an `innerWindowModel` containing only the extracted window function. This model gets chained between `groupByModel` and `translatingModel`:

```
groupByModel [symbol, sum(row_number)]
  └── innerWindowModel [row_number() OVER (...)]
       └── translatingModel [symbol, timestamp]
```

The `groupByModel` needs `symbol` from its nested model, but `innerWindowModel` has no pass-through column for it. The existing pass-through logic (`SqlOptimiser.java:9512-9525`) propagates columns from `windowModel` (top-level window expressions), which is empty when window functions only appear inside aggregates.

With implicit grouping the `groupByModel` only references `sum(literal_ref)`, and `literal_ref` is already in the `innerWindowModel` from the extraction — so no pass-through is needed and the query works.

## Affected code

- `SqlOptimiser.java:3759` — `extractAndRegisterNestedWindowFunctions`
- `SqlOptimiser.java:9512-9525` — pass-through propagation (misses `groupByModel` columns)
- `SqlOptimiser.java:9565-9572` — inner window model chaining

jovfer:
Fixes #6954

## Summary

- `max(avg(x) OVER (...)) GROUP BY category` and similar queries — a window function used as the argument of an aggregate, combined with explicit `GROUP BY` — previously failed with a confusing `Invalid column: <key>` error.
- With the fix, the same queries return correct results. Implicit-grouping shapes (no `GROUP BY` clause) already worked and are unaffected.

## Approach

The rewrite pipeline puts an inner window model between `groupByModel` and `translatingModel` whenever an aggregate argument contains a nested window function. The existing pass-through logic only propagated columns from the top-level `windowModel`, which is empty in this shape, so the inner window model did not expose the `GROUP BY` keys or the base columns referenced from aggregate arguments. The fix adds a dedicated pass after the inner window models are populated, walking `groupByModel`'s columns and the later inner window models' OVER clauses, and propagating referenced literals through the chain with an owner-index filter so a later inner window model's alias is never added to a deeper model that cannot resolve it.

## Scope

- Simple literal `GROUP BY` keys (`GROUP BY category`), positional references (`GROUP BY 1`), expression `GROUP BY` keys (`GROUP BY upper(cat)`, `GROUP BY a + b`), multiple `GROUP BY` keys, multiple aggregates each wrapping a distinct nested window, `PARTITION BY` on columns outside the `GROUP BY` set, and `NULL` values in source data. Both `STRING` and `SYMBOL` GROUP BY keys are covered.
- Performance impact is limited to compile time: one extra traversal of `groupByModel`'s columns, plus, when more than one inner window model exists, an extra pass over the later models' window expressions. Traversal is bounded by `MAX_WINDOW_FUNCTION_NESTING_DEPTH` and the group-by column count, and uses pooled allocations. No data-path impact.

## Test plan

- New tests reproduce the original failure with both `STRING` and `SYMBOL` GROUP BY keys.
- New tests cover base columns mixed with a window function inside the aggregate, multiple `GROUP BY` keys, a non-aggregate `GROUP BY` expression (`GROUP BY upper(cat)`), multi-aggregate cases with distinct nested windows (two and three inner window models), `PARTITION BY` on a column outside the GROUP BY set, and `NULL` source values.
- Existing implicit-grouping tests (`testWindowFunctionAsArgumentToAggregate`, `testWindowFunctionArithmeticAsArgumentToAggregate`) still pass.
- WindowFunctionTest, SqlOptimiserTest, SqlParserTest, GroupByTest pass with no regressions.


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 119 / 155 (76.77%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/http/processors/SqlValidationProcessorState.java|0|3|00.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|0|4|00.00%|
|:large_blue_circle:|io/questdb/DefaultServerConfiguration.java|0|2|00.00%|
|:large_blue_circle:|io/questdb/cairo/CairoException.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/std/str/Utf8s.java|2|8|25.00%|
|:large_blue_circle:|io/questdb/std/Unsafe.java|3|8|37.50%|
|:large_blue_circle:|io/questdb/cairo/wal/WalWriter.java|44|54|81.48%|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|25|30|83.33%|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/LineWalAppender.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/LineTcpMeasurementScheduler.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/TableUpdateDetails.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpConnectionContext.java|17|17|100.00%|
|:large_blue_circle:|io/questdb/cairo/pool/WalWriterPool.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/arr/BorrowedArray.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/JsonQueryProcessorState.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/std/CharSequenceIntHashMap.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/LineHttpProcessorState.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpServer.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/udp/LineUdpParserImpl.java|2|2|100.00%|

bluestreak01:
@jovfer — here's a detailed review of this PR.

## Moderate

### 1. Multi-aggregate with different nested window functions causes AIOOBE

`SqlOptimiser.java:1026-1055` — `addLiteralPassThroughToInnerWindowModels` adds ALL literals from an aggregate's AST to ALL inner window models indiscriminately. When multiple aggregates contain different nested window functions, window function aliases from one aggregate's inner model get added as pass-through to another aggregate's inner model, which can't resolve them from its nested model.

**Scenario:**
```sql
SELECT cat, max(avg(x) OVER()), min(sum(x) OVER()) FROM tab GROUP BY cat
```

**Trace:**
1. Extraction creates `innerWm_A` (owns `avg_alias`) and `innerWm_B` (owns `sum_alias`)
2. New code at line 9600 traverses `min(sum_alias)`, finds `sum_alias` literal
3. `innerWm_A.getAliasToColumnMap().excludes("sum_alias")` → true → adds `sum_alias` to `innerWm_A`
4. Chain: `groupByModel → innerWm_B → innerWm_A → translatingModel`
5. `SqlCodeGenerator.java:8283` — `generateSelectWindow` resolves `innerWm_A`'s pass-through `sum_alias` via `baseMetadata.getColumnIndexQuiet("sum_alias")` → returns `-1`
6. `SqlCodeGenerator.java:8284` — `baseMetadata.getColumnMetadata(-1)` → `ObjList.getQuick(-1)` → `ArrayIndexOutOfBoundsException`

**Impact:** This case was already broken before the PR (failed with clean "Invalid column" SqlException). After the PR, it fails with an unhandled AIOOBE — a worse failure mode.

**Suggested fix:** In `addLiteralPassThroughToInnerWindowModels`, skip literals whose tokens match window function aliases in any inner window model. Only propagate base-column literals. Alternatively, check each literal against `translatingModel`'s alias map before adding.

### 2. GROUP BY tests lack ORDER BY — non-deterministic result ordering

`WindowFunctionTest.java:13019,13042,13066,13091,13115` — Tests 2-6 all use explicit GROUP BY but no `ORDER BY`. The expected results assume `A` before `B` (or `A/X, A/Y, B/X, B/Y` for multi-key test). Existing GROUP BY tests in the same file (lines 6329, 6375, 7699) consistently include `ORDER BY`. Without ORDER BY, result order is not guaranteed and could change with GROUP BY implementation changes.

**Suggested fix:** Add `ORDER BY category` (or `ORDER BY cat1, cat2`) to each GROUP BY query.

## Minor

### 3. Tree traversal pattern inconsistency

`SqlOptimiser.java:1026-1055` — `addLiteralPassThroughToInnerWindowModels` uses a `while (node != null)` loop with an inline stack-pop check at the bottom. The dominant pattern in this file (15+ locations, e.g., `hasGroupByFunc` at line 307, `addMissingTablePrefixesForGroupByQueries` at line 1070) uses:
```java
while (!sqlNodeStack.isEmpty() || node != null) {
    if (node != null) { /* traverse */ }
    else { node = sqlNodeStack.poll(); }
}
```
The new pattern is functionally equivalent but breaks consistency. Consider refactoring to match.

### 4. Missing test: multiple aggregates with different window functions

No test covers `SELECT cat, max(avg(x) OVER()), min(sum(x) OVER()) FROM tab GROUP BY cat`. This would exercise the multi-inner-window-model pass-through path and would have caught finding #1.

| Original claim | Dismissal reason |
|---|---|
| `sqlNodeStack` reuse could lose state | Method is a leaf; no caller has `sqlNodeStack` in use at the call site (line 9600). Prior use at 9414 completes first. |
| NULL `node.lhs` mishandled | Null lhs causes `node = null`, which falls through to `sqlNodeStack.isEmpty()` check — correct behavior per `ExpressionNode` contract. |
| `ast.type != FUNCTION \|\| !isGroupBy()` misclassifies expressions | Non-aggregate FUNCTION nodes (e.g., `upper`, `+`, `CASE`) correctly return `!isGroupBy() = true` → treated as GROUP BY keys. |
| Alias vs token mismatch | `nextColumn(name)` creates alias=name=token, consistent with how translating model aliases work. |
| Duplicate column not prevented | Both `excludes` guard (line 1032, 9593) and `addBottomUpColumn`'s internal `contains` check prevent duplicates. |
| Resource leak from pool allocation | `ObjectPool` is bulk-cleared on session end. `excludes` guard makes the `addBottomUpColumn` throw path unreachable. |
| Performance regression from nested loops | Compile-time only. Loops bounded by column count × model count (single digits). Pool-based allocation. |

## Summary

**Verdict: Request changes** — The core fix is correct for the stated scenario (single aggregate with nested window function + GROUP BY). Two issues should be addressed: (1) `addLiteralPassThroughToInnerWindowModels` should filter out window function aliases to avoid adding unresolvable columns to inner window models in the multi-aggregate case, and (2) the GROUP BY tests should include ORDER BY for deterministic ordering. Both are straightforward fixes.

12 findings examined, 5 confirmed (2 moderate, 2 minor, 1 test gap), 7 false positives removed.

jovfer:
Thanks @bluestreak01 — addressed in 7ec0b2d487.

**Moderate 1 (spurious pass-through → AIOOBE):** Reproduced locally. In my branch the same input first fails with `Invalid column: x` at `parseFunction` time before it reaches the `getColumnMetadata(-1)` call, but the root cause is what you described.

Fix:
- `addLiteralPassThroughToInnerWindowModels` now takes an `upToExclusive` bound and uses a new `findInnerWindowModelOwnerIdx` helper. For each literal it finds the first inner window model that owns the token and only propagates to models strictly above the owner — so a later model's alias is never added to a deeper model that can't resolve it.
- To keep the multi-aggregate case actually working (B can't see `x` through A unless A forwards it), I added a second pass after the aggregate-arg pass: for each later inner window model (index ≥ 1), walk its own window expression (AST, PARTITION BY, ORDER BY, frame bounds) and propagate referenced literals through the deeper inner window models.

**Moderate 2 (missing ORDER BY):** Added `ORDER BY category` / `ORDER BY cat1, cat2` to the three affected GROUP BY tests.

**Minor 3 (tree traversal pattern):** Rewrote `addLiteralPassThroughToInnerWindowModels` to use the canonical `while (!sqlNodeStack.isEmpty() || node != null)` form.

**Minor 4 (missing test):** Added `testWindowFunctionAsArgumentToAggregateWithMultipleAggregatesAndGroupBy` exercising `max(avg(x) OVER (PARTITION BY category)), min(sum(x) OVER (PARTITION BY category))` with GROUP BY — this reproduces the scenario from finding 1 and would have caught the bug.

WindowFunctionTest, SqlOptimiserTest, SqlParserTest, GroupByTest, WindowJoinTest, ExplainPlanTest all green locally.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 48 / 54 (88.89%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|48|54|88.89%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 90 / 97 (92.78%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|48|54|88.89%|
|:large_blue_circle:|io/questdb/cutlass/parquet/HybridColumnMaterializer.java|26|27|96.30%|
|:large_blue_circle:|io/questdb/cutlass/parquet/SQLSerialParquetExporter.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|14|14|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/ExportQueryProcessor.java|1|1|100.00%|

bluestreak01 (approved):
Approving. Owner-filtered propagation correctly handles 1-, 2-, and 3-inner-window-model shapes. Multi-aggregate AIOOBE and missing-ORDER BY concerns from the prior round are resolved and covered by new tests.

Minor nits to consider (non-blocking):

1. Redundant `if (nInner > 1)` guard at SqlOptimiser.java:9780 — the enclosing `for (int i = 1; i < nInner; i++)` already no-ops when `nInner <= 1`.

2. Comment at SqlOptimiser.java:1030-1032 reads as if only literals with an owner >= 0 propagate, but base-column literals (ownerIdx == -1) correctly fan out to all inner models via `ownerIdx + 1 = 0`. Worth rephrasing so the -1 case is not lost.

3. Missing test for frame bound expressions referencing a column — `wc.getRowsLoExpr()`/`getRowsHiExpr()` traversal paths are not exercised by the existing tests. If the feature isn't supported, drop those two traversal calls.

4. Missing test for GROUP BY key not in SELECT list — e.g. `SELECT sum(avg(x) OVER ()) FROM tab GROUP BY category`. Verified to work (`replaceIfAggregateOrLiteral` at SqlOptimiser.java:6633-6649 always inserts GROUP BY literals into `groupByModel.getBottomUpColumns()`), but a regression test would pin the behavior.

5. Sanity-check `supportsRandomAccess=true, expectSize=true` for the GROUP BY + ORDER BY tests — CI is green, so likely fine, but these flags assert factory properties and a mismatch fails silently in the wrong direction.

PR #6965, opened 2026-04-13
fix(core): fix slow WAL apply for indexed high-cardinality SYMBOL columns on non-last partitions

ideoma:
## Summary

- Fix dramatically slow WAL apply when inserting into non-last partitions of tables with high-cardinality indexed SYMBOL columns. The apply time grew quadratically with each transaction — e.g. from seconds to hours over a series of batch inserts into the same historical partition.
- The root cause was a typo in `MemoryCMARWImpl.of()` (5-arg fd-based overload): it stored the raw page size (e.g. 16,777,216) into `extendSegmentMsb` instead of its most significant bit position (24). Since `extend0()` uses Java bit shifts (`n & 63`), `16,777,216 & 63 = 0`, so the effective shift was `>>> 0`. This caused the bitmap index key file to grow by ~1 byte per write instead of in page-sized chunks, triggering a separate `posix_fallocate` + `mremap` for every `putLong` call on new key entries.
- The 7-arg fd-based overload and all path-based overloads already applied `Numbers.msb()` correctly — only this one overload had the bug.

## Test plan

- `testBitmapIndexExtendOnNonLastPartitionAppendDateOnly` opens a BitmapIndexWriter with init=false via the fd-based API (the exact O3 append code path), adds 10K new keys, and asserts fewer than 100 `allocate()` calls. Before the fix this produced 39,912 calls (+8 bytes each); after the fix it produces 1 call (+16MB page-aligned).
- `testBitmapIndexExtendOnNonLastPartitionWal` is an end-to-end WAL test that inserts multiple batches of rows with unique symbols into a non-last partition via O3, verifying the bitmap index .k file extends efficiently across all batches.


mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/vm/MemoryCMARWImpl.java|2|2|100.00%|

bluestreak01 (approved):

PR #6967, opened 2026-04-13
chore(core): fix flaky HyperLogLogTestUtils

glasstiger:
## Summary
- Widen the cardinality tolerance in `HyperLogLogTestUtils.assertCardinality` from 5 to 6 standard errors when `precision <= 5`, to fix flakiness at the lowest precisions.
- At precision 4–5 only 16–32 registers are used; the estimator is known to be biased high and the error distribution has fatter tails than the 3σ bound assumed by Flajolet et al., so the existing 5σ envelope was occasionally tripped.

## Test plan
- [x] Re-run `HyperLogLog*` tests locally / in CI at the affected precisions


glasstiger:
an example of the test failure fixed in this change:
[775.txt.zip](https://github.com/user-attachments/files/26678224/775.txt.zip)

bluestreak01 (approved):

PR #6971, opened 2026-04-13
fix(core): avoid redundant bitmap index rollback after WAL O3 append

jerrinot:
WAL apply could store the bitmap index max row as the exclusive end when appending to an indexed SYMBOL column in a non-last partition. The next apply then treated the index as one row ahead and rolled it back before indexing new rows.

For users with high-cardinality indexed symbols, this added unnecessary index scan work and could increase WAL apply latency.

Store the bitmap index max row as an inclusive row id, clarify the row-id adjustment in the O3 index update path, and add WAL regression coverage for the inclusive max-row invariant.


jerrinot:
Approve — 1 verified fix, 1 verified regression test, 3 verification agents' concerns dropped as false positives. Only blocking ask: add the missing labels (Bug, Core, WAL, Performance). Optional: link the issue, move MAX_VALUE_OFFSET to a non-private constant.

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/O3CopyJob.java|1|1|100.00%|

ideoma (approved):

PR #6983, opened 2026-04-16
fix(pgwire): fix memory corruption from malformed wire messages

bluestreak01:
## Summary

Several code paths in `io.questdb.cutlass.pgwire` trusted length and dimension fields read from the wire without validating them. In the pre-auth handshake this allowed a network attacker to drive pointer arithmetic to positions outside the receive buffer; in the post-auth extended query protocol it allowed an authenticated client to wrap the flat length of an array to a value smaller than the shape implied.

### Pre-authentication

- `PGCleartextPasswordAuthenticator.processInitMessage` accepted any signed `int` as `msgLen`. Negative values passed the `msgLen > availableToRead` check and flowed into `processStartupMessage`, where `msgLimit = recvBufStart + msgLen` pointed *before* the receive buffer. `recvBufReadPos = msgLimit` then moved the read cursor backwards, and `compactRecvBuf` invoked `Vect.memmove(recvBufStart, recvBufReadPos, recvBufWritePos - recvBufReadPos)` with a source pointer outside the allocated buffer.
- `PGCleartextPasswordAuthenticator.processPasswordMessage` had the same class of issue with its `msgLen`.
- Both handlers now require `msgLen` to sit between the protocol minimum and the capacity of the receive buffer; out-of-range values terminate the connection via `PGMessageProcessingException`.
- `DefaultPGCircuitBreakerRegistry.cancel` compared the index with `<` instead of `<=`, so a `CancelRequest` with `pid == circuitBreakers.size()` indexed one slot past the logical end of the `ObjList`. The method also read `circuitBreakers.getQuick(idx)` and `cb.getSecret()` outside the registry spin lock, while `add`/`remove`/`close` all hold it. The cancel path now runs the whole lookup, secret check and `cb.cancel()` under the same lock. It additionally rejects the `-1` sentinel that `clear()` installs, closing the window between `clear()` and the next `init()` during which the slot held no real secret.

### Post-authentication

- `PGNonNullBinaryArrayView.addDimLen` and `PGNonNullVarcharArrayView.addDimLen` computed `flatViewLength *= dimLen` on signed `int`s without overflow checking. Dimensions such as `[65537, 65537]` wrapped `flatViewLength` to `131073` while `shape` still claimed `~4.3B` elements. `setPtrAndCalculateStrides` then validated the on-wire byte count against the wrapped length and passed. Both methods now use `Math.multiplyExact`.
- `PGPipelineEntry.setBindVariableAsArray` and `setBindVariableAsVarcharArray` did not check that `valueSize` stayed non-negative as it was decremented, and did not reject negative `dimensions` or `dimensionSize`. They now validate `valueSize` before each field read, reject non-positive dimension counts and negative dimension sizes, and translate `ArithmeticException` from `addDimLen` into a `PGMessageProcessingException`.

## Tradeoffs

- Well-formed clients already send lengths within the validated ranges, so there is no observable behavior change for legitimate traffic.
- The pre-auth checks add two comparisons on the startup/password paths; those paths are not performance-sensitive.
- The per-dimension checks in `setBindVariableAsArray` run once per dimension per `Bind`. Arrays typically have few dimensions, and this is not on a hot per-row path.
- Moving the cancel path under the registry spin lock means `cb.cancel()` now runs while the lock is held. `close()` already calls `cb.cancel()` under the same lock, so the contention pattern is unchanged.
- The `-1` sentinel rejection in `cancel` is conservative: a real random secret equal to `-1` (1 chance in 2^32) would be rejected. That is an acceptable cost to eliminate the clear/init window.

## Test plan

- mvn -pl core -Dtest='DoubleArrayParserTest,VarcharArrayParserTest,PGSecurityTest' test
- mvn -pl core -Dtest='PGArraysTest,PGVarcharArrayBindVariablesTest' test
- mvn -pl core -Dtest='PGJobContextTest' test
- mvn -pl core test (full suite)
- Manual: connect via pgjdbc and psql to confirm normal auth and query flow

### DIFF
--- a/core/src/main/java/io/questdb/cutlass/pgwire/PGPipelineEntry.java
+++ b/core/src/main/java/io/questdb/cutlass/pgwire/PGPipelineEntry.java
@@ -2869,11 +2869,14 @@ public class PGPipelineEntry implements QuietCloseable, Mutable {
     }
 
     private void setBindVariableAsArray(int i, long lo, int valueSize, long msgLimit, BindVariableService bindVariableService) throws SqlException, PGMessageProcessingException {
+        if (valueSize < 3 * Integer.BYTES) {
+            throw kaput().put("malformed array header [valueSize=").put(valueSize).put(']');
+        }
         int dimensions = getInt(lo, msgLimit, "malformed array dimensions");
         lo += Integer.BYTES;
         valueSize -= Integer.BYTES;
-        if (dimensions == 0) {
-            throw kaput().put("array dimensions cannot be zero");
+        if (dimensions <= 0) {
+            throw kaput().put("array dimensions must be positive [dimensions=").put(dimensions).put(']');
         }
         if (dimensions > ColumnType.ARRAY_NDIMS_LIMIT) {
             throw kaput().put("array dimensions cannot be greater than maximum array dimensions [dimensions=").put(dimensions).put(", max=").put(ColumnType.ARRAY_NDIMS_LIMIT).put(']');
@@ -2891,8 +2894,18 @@ public class PGPipelineEntry implements QuietCloseable, Mutable {
 
         PGNonNullBinaryArrayView arrayView = arrayViewPool.next();
         for (int j = 0; j < dimensions; j++) {
+            if (valueSize < 2 * Integer.BYTES) {
+                throw kaput().put("malformed array dimension header [dimensionIndex=").put(j).put(']');
+            }
             int dimensionSize = getInt(lo, msgLimit, "malformed array dimension size");
-            arrayView.addDimLen(dimensionSize);
+            if (dimensionSize < 0) {
+                throw kaput().put("array dimension size cannot be negative [dimensionIndex=").put(j).put(", size=").put(dimensionSize).put(']');
+            }
+            try {
+                arrayView.addDimLen(dimensionSize);
+            } catch (ArithmeticException e) {
+                throw kaput().put("array size overflow");
+            }
             lo += Integer.BYTES;
             valueSize -= Integer.BYTES;
 
@@ -3072,6 +3085,9 @@ public class PGPipelineEntry implements QuietCloseable, Mutable {
     }
 
     private void setBindVariableAsVarcharArray(int i, long lo, int valueSize, long msgLimit, BindVariableService bindVariableService) throws SqlException, PGMessageProcessingException {
+        if (valueSize < 5 * Integer.BYTES) {
+            throw kaput().put("malformed varchar array header [valueSize=").put(valueSize).put(']');
+        }
         int dimensions = getInt(lo, msgLimit, "malformed array dimensions");
         lo += Integer.BYTES;
         valueSize -= Integer.BYTES;
@@ -3095,7 +3111,14 @@ public class PGPipelineEntry implements QuietCloseable, Mutable {
 
         // dimensions == 1
         int dimensionSize = getInt(lo, msgLimit, "malformed array dimension size");
-        arrayView.addDimLen(dimensionSize);
+        if (dimensionSize < 0) {
+            throw kaput().put("array dimension size cannot be negative [size=").put(dimensionSize).put(']');
+        }
+        try {
+            arrayView.addDimLen(dimensionSize);
+        } catch (ArithmeticException e) {
+            throw kaput().put("array size overflow");
+        }
         lo += Integer.BYTES;
         valueSize -= Integer.BYTES;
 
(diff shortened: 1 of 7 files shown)


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 39 / 49 (79.59%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGPipelineEntry.java|15|23|65.22%|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGNonNullBinaryArrayView.java|4|6|66.67%|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGCleartextPasswordAuthenticator.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cutlass/pgwire/DefaultPGCircuitBreakerRegistry.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGNonNullVarcharArrayView.java|2|2|100.00%|

core/src/test/java/io/questdb/test/cutlass/pgwire/PGJobContextTest.java
```diff
@@ -1630,6 +1656,36 @@ public void testBadPasswordLength() throws Exception {
...
+        assertHexScript(
+                """
+                        >0000000804d2162f
+                        <4e
+                        >0000007500030000757365720061646d696e006461746162617365006e6162755f61707000636c69656e745f656e636f64696e67005554463800446174655374796c650049534f0054696d655a6f6e65004575726f70652f4c6f6e646f6e0065787472615f666c6f61745f64696769747300320000
+                        <520000000800000003
+                        >7000200000
+                        <!!"""
+        );
+    }
```
RaphDal, line 1687: How are we supposed to review that?
Please abstract the hex generation away to make this readable

core/src/test/java/io/questdb/test/cutlass/pgwire/PGJobContextTest.java
```diff
@@ -2205,6 +2261,50 @@ public void testBatchInsertWithTransaction() throws Exception {
...
+        assertHexScript(
+                """
+                        >0000006b00030000757365720061646d696e006461746162617365006e6162755f61707000636c69656e745f656e636f64696e67005554463800446174655374796c650049534f0054696d655a6f6e6500474d540065787472615f666c6f61745f64696769747300320000
+                        <520000000800000003
+                        >700000000a717565737400
+                        <520000000800000000530000001154696d655a6f6e6500474d5400530000001d6170706c69636174696f6e5f6e616d6500517565737444420053000000187365727665725f76657273696f6e0031312e33005300000019696e74656765725f6461746574696d6573006f6e005300000019636c69656e745f656e636f64696e670055544638004b0000000c0000003fbb8b96505a0000000549
+                        >500000002b0073656c6563742024312066726f6d206c6f6e675f73657175656e6365283129000001000003fe420000002e00000001000100010000001c0000000200000000000002bd000100010000000100010001000000010000450000000900000000005300000004
+                        <450000002b433030303030004d61727261792073697a65206f766572666c6f7700534552524f5200503100005a0000000549"""
+        );
+    }
```
RaphDal, line 2346: Same

core/src/main/java/io/questdb/cutlass/pgwire/PGPipelineEntry.java
```diff
@@ -2891,8 +2894,18 @@ private void setBindVariableAsArray(int i, long lo, int valueSize, long msgLimit
 
         PGNonNullBinaryArrayView arrayView = arrayViewPool.next();
         for (int j = 0; j < dimensions; j++) {
+            if (valueSize < 2 * Integer.BYTES) {
+                throw kaput().put("malformed array dimension header [dimensionIndex=").put(j).put(']');
+            }
```
RaphDal, line 2899: This check could be done before the loop

RaphDal (changes requested):
The fixes are looking good, tests should be made understandable

RaphDal (dismissed):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 42 / 49 (85.71%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGNonNullBinaryArrayView.java|4|6|66.67%|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGPipelineEntry.java|18|23|78.26%|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGCleartextPasswordAuthenticator.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cutlass/pgwire/DefaultPGCircuitBreakerRegistry.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGNonNullVarcharArrayView.java|2|2|100.00%|

Issue #4678, opened 2024-06-13
"Sample by align to calendar time zone" does not aggregate results under certain conditions


oz-brg:
### To reproduce

Under certain conditions, using `SAMPLE BY 1d ALIGN TO CALENDAR TIME ZONE` does not aggregate data during a DST change, resulting in the unaggregated data being shown for that day.

Possibly related to https://github.com/questdb/questdb/issues/4544 but the resulting bug is different.

This appears to happen when there is no data in the day immediately prior to a DST change and when the first data point available again is after the DST change.

To reproduce in the demo environment (DST change was on March 9th 2024 in the US):
```sql
SELECT timestamp, sum(amount)
FROM (
    SELECT timestamp, amount
    FROM trades
    WHERE timestamp IN '2024-03-08' 
    OR timestamp BETWEEN('2024-03-10T08:00:00Z', '2024-03-12')
) SAMPLE BY 1d ALIGN TO CALENDAR TIME ZONE 'America/New_York';
```

This results in:
![image](https://github.com/questdb/questdb/assets/157364047/f4e2c8fc-dcdf-42e1-921d-b1754e4db9bb)

However, if there is data immediately prior to the DST change, then the correct results are shown: (in the example below, the selected time is adjusted back 2 hours)
```sql
SELECT timestamp, sum(amount)
FROM (
    SELECT timestamp, amount
    FROM trades
    WHERE timestamp IN '2024-03-08' 
    OR timestamp BETWEEN('2024-03-10T06:00:00Z', '2024-03-12')
) SAMPLE BY 1d ALIGN TO CALENDAR TIME ZONE 'America/New_York';
```

Result:
![image](https://github.com/questdb/questdb/assets/157364047/79683884-04b4-4625-b311-a3f826d68287)

### QuestDB version:

8.0.0

### OS, in case of Docker specify Docker and the Host OS:

Public demo instance

### File System, in case of Docker specify Host File System:

Public demo instance

### Full Name:

Oliver Zhu

### Affiliation:

Bridgeton

### Have you followed Linux, MacOs kernel configuration steps to increase Maximum open files and Maximum virtual memory areas limit?

- [X] Yes, I have

PR #6987, opened 2026-04-17
fix(sql): fix crash on select-choose with explicit timestamp and column rename

bluestreak01:
## Summary

When a `select-choose` model declared an explicit `timestamp(...)` clause matching the child's designated timestamp, `SqlCodeGenerator.generateSelectChoose` treated it as a redundant (entity) model and returned the child factory as-is. The check did not verify that the projection columns actually matched the child's columns, so a model that also renamed a column (e.g. `timestamp AS ts`) was silently elided — the rename was discarded and the renamed name then failed to resolve upstream.

Two user-visible symptoms depending on the enclosing query shape:
- `SqlException: Invalid column: ts` during SAMPLE BY execution.
- `AssertionError: wtf? ts` in `generateSelectChoose` (line 7406) when an extra wrapper layer pushed the lookup past the sub-factory's metadata.

Reproducer (the CTE double-parens form triggers the assertion trip):

```sql
WITH Test AS ((
  SELECT timestamp AS ts, amount
  FROM (
    SELECT timestamp, amount
    FROM trades
    WHERE timestamp IN '2024-03-08' OR timestamp BETWEEN('2024-03-10T08:00:00Z', '2024-03-12')
  ) timestamp(timestamp))
) SELECT ts, sum(amount) FROM Test
SAMPLE BY 1d ALIGN TO CALENDAR TIME ZONE 'America/New_York'
```

## Fix

Tighten the entity check in `generateSelectChoose`:
- Require the child column count to match the projection count.
- Require each projected column's AST token and alias to match the child's column name at the same position.
- Only then, if the model declares an explicit `timestamp(...)`, verify it aligns with the child's designated timestamp.

Before, the else-branch accepted entity purely on the timestamp alignment, ignoring the projection entirely. The three previously-buggy cases (`timestamp != null` with mismatched column count, or with renamed columns) now correctly produce a real `SelectedRecordCursorFactory` instead of handing back the child's metadata.

## Tradeoffs

- Two code paths that previously elided the select-choose (when `timestamp != null` and columns didn't match, or counts differed) now materialize a projection factory. These paths were incorrect; no existing test depended on the old behavior — the full `SqlParserTest` suite (1060 tests) passes.
- No change for queries without an explicit `timestamp(...)` clause or without renamed columns.

## Test plan

- Added `SqlCodeGeneratorTest.testCteWithDoubleParensAndTimestampClauseSampleBy`: end-to-end query with 9 inserted rows crossing the 2024-03-10 DST boundary, verifying filtered rows are correctly bucketed and summed.
- Added `SqlParserTest.testCteWithDoubleParensAndTimestampClauseSampleBy`: asserts the parsed model shape so future changes to the optimizer surface are detectable.
- Full `SqlParserTest` run locally: 1060 passed, 0 failed, 4 pre-existing skipped.

Note: issue #4678 also mentions that the bucket *timestamp* is wrong on the DST-start day when the previous day has no data. That is a separate bug in the SAMPLE BY cursor's DST state machine (`AbstractNoRecordSampleByCursor`) and will be addressed in a follow-up.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
+++ b/core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
@@ -7348,7 +7348,19 @@ public class SqlCodeGenerator implements Mutable, Closeable {
 
         boolean entity;
         // the model is considered entity when it doesn't add any value to its nested model
-        if (metadata.getColumnCount() == selectColumnCount) {
+        boolean hasRename = false;
+        for (int i = 0; i < selectColumnCount; i++) {
+            QueryColumn qc = columns.getQuick(i);
+            if (qc.getAlias() != null && !Chars.equals(qc.getAlias(), qc.getAst().token)) {
+                // user projection renames a column (e.g. `timestamp AS ts`); the wrapper
+                // carries real semantics and must not be elided.
+                hasRename = true;
+                break;
+            }
+        }
+        if (hasRename) {
+            entity = false;
+        } else if (timestamp == null && metadata.getColumnCount() == selectColumnCount) {
             entity = true;
             for (int i = 0; i < selectColumnCount; i++) {
                 QueryColumn qc = columns.getQuick(i);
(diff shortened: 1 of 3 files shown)


nwoolmer (dismissed):
LGTM in principle

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|9|9|100.00%|

ideoma (approved):

PR #6993, opened 2026-04-19
perf(sql): apply parallel top-K to queries with SELECT projections

### LINKED ISSUE #6528: Complex projection disables top-k optimization
### To reproduce

table:
```sql
create table tab as (select x, x::timestamp as ts, x::timestamp as ts2 from long_sequence(1_000)) timestamp (ts) partition by hour wal;
```

This query uses the top-k algo:
```sql
select * from tab
where ts2 in '2025'
order by ts2 desc
limit 10;
```

This also uses top-k
```sql
select x from tab
where ts2 in '2025'
order by ts2 desc
limit 10
```

This uses top-k too:
```sql
select x, * from tab
order by ts2 desc
limit 10;
```

but this uses Sort light (=no top-k optimization)
```sql
select x, * from tab
where ts2 in '2025'
order by ts2 desc
limit 10;
```

That's suboptimal, as the top-k algorithm can have a huge impact on larger datasets. 

### QuestDB version:

9.2.2

NA

NA

### Full Name:

Jaromir Hamala

### Affiliation:

QuestDB

- [x] Yes, I have

### Additional context

_No response_

DHRUV6029:
Fixes #6528

## Summary

Extend the parallel top-K gate in `SqlCodeGenerator#generateOrderBy` so it also fires when a column-projection wrapper (`SelectedRecordCursorFactory` or `VirtualRecordCursorFactory`) sits between the `ORDER BY ... LIMIT N` and the filtered page-frame scan. Before this change, any non-literal SELECT list combined with a `WHERE` clause forced the query onto the generic `Sort light` path, because the projection wrapper hid the steal-able filter from the gate.

## Problem

Reproduction from the issue:

```sql
CREATE TABLE tab AS (
    SELECT x, x::timestamp AS ts, x::timestamp AS ts2
    FROM long_sequence(1_000)
) TIMESTAMP (ts) PARTITION BY HOUR WAL;
```

| Query | Plan before | Plan after |
| --- | --- | --- |
| `SELECT * FROM tab WHERE ts2 IN '2025' ORDER BY ts2 DESC LIMIT 10` | Async JIT Top K | Async JIT Top K |
| `SELECT x FROM tab WHERE ts2 IN '2025' ORDER BY ts2 DESC LIMIT 10` | Async JIT Top K | Async JIT Top K |
| `SELECT x, * FROM tab WHERE ts2 IN '2025' ORDER BY ts2 DESC LIMIT 10` | Sort light | Async JIT Top K |
| `SELECT x + 1 AS xp, ts2 FROM tab WHERE ts2 IN '2025' ORDER BY ts2 DESC LIMIT 10` | Sort light | Async JIT Top K |

The first two cases already hit the top-K gate because the generated plan is `AsyncJitFilter -> PageFrame` with no wrapper in between. The third and fourth cases generate an extra `SelectedRecord` (duplicate columns) or `VirtualRecord` (computed column) above the filter, and the gate's `supportsFilterStealing()` / `supportsPageFrameCursor()` checks look only at the immediate child, so they return false and fall through to `Sort light`. The result set is correct, but the query materializes and sorts every matching row instead of keeping a bounded heap of N.

## Approach

The codebase already has a projection-peel precedent in `generateAsOfJoin` and `generateLatestBy`: when a projection wrapper hides a useful optimization target, peel the wrapper, apply the optimization to the inner factory, and re-wrap.

This PR collapses the direct page-frame, direct filter-stealing, and projection-peel shapes into a **single unified branch** in `generateOrderBy`, driven by two default methods on `RecordCursorFactory`:

- `translateOrderByColumnToBase(int projectedIndex)` — returns the corresponding column index in the base metadata, or a negative value when the projected column cannot be resolved (e.g., a computed `VirtualRecord` column). The default is identity; `SelectedRecordCursorFactory` and `VirtualRecordCursorFactory` override it to chain through their base, so nested wrappers resolve transparently.
- `rewrapOverTopK(RecordCursorFactory topK, RecordMetadata orderedMetadata)` — re-wraps a freshly-built top-K factory so the output shape is preserved. The default is a pass-through. Projection wrappers override to re-create themselves over the new base.

A cheap structural predicate `canReachPageFrameLeafForTopK` decides whether the gate should attempt the peel at all. When it holds, a shared helper `buildAsyncTopKOverStolenFilter` steals the filter state from the inner filter factory via `halfClose()`, builds `AsyncTopKRecordCursorFactory` on the page-frame leaf, and the call site re-wraps via the projection wrapper's override. Ownership of the half-closed inner base transfers to the new top-K factory, matching the AsOf/LatestBy peel precedent.

The predicate and the call site peel a single projection layer. Two or more stacked projection wrappers above the filter (rare today) fall through to `Sort light` rather than recursing.

### Incidental fix: AsOf join peel guards (keyed and non-keyed)

Widening `isProjection()` to also return `true` on `VirtualRecordCursorFactory` is required by the unified branch above. That widening incidentally made two existing branches in `SqlCodeGenerator#generateJoinAsof` reachable for `VirtualRecord` slaves: the keyed peel at line 4369 and the non-keyed peel at line 4502. Both peel a slave projection into a `FilteredAsOfJoin*FastRecordCursorFactory` and require a non-null column cross index — only `SelectedRecord` provides one. Before this PR, `isProjection()` was effectively equivalent to "is `SelectedRecord`", so the branches relied on that implicit invariant.

The keyed branch trips `assert stolenCrossIndex != null`. The non-keyed branch is worse: with assertions disabled, `FilteredAsOfJoinNoKeyFastRecordCursorFactory` treats a null cross index as "no projection remap" and would silently drop the `VirtualRecord` layer's column translation, producing wrong results.

This PR adds an explicit `getColumnCrossIndex() != null` guard at both call sites so `VirtualRecord` slaves fall through to whichever later branch catches them — the same path they took before the widening. The fix is a single extra condition per branch; the AsOf branches' logic is otherwise untouched and the deferred AsOf refactor remains a follow-up (as noted during review). Two regression tests pin the fix: `AsOfJoinTest.testAsOfJoinWithTopDownFilteredDateaddTimestampProjection` covers the keyed path, and `AsOfJoinTest.testAsOfJoinNoKeyWithTopDownFilteredDateaddTimestampProjection` covers the non-keyed path with row-level assertions, so a regression surfaces both with and without `-ea`.

## Alternatives considered

- Teach `AsyncTopKRecordCursorFactory` to accept a projection wrapper directly. Rejected: it would spread projection concerns into the execution engine and duplicate the column-translation logic that `SelectedRecordCursorFactory` already owns.
- Push the projection below the filter at parse time. Rejected: it changes semantics for any projection whose computed columns reference the filter's bound variables, and it would interact poorly with `isProjection()` consumers elsewhere in the generator.
- Extend the gate to cover `ORDER BY` on computed (non-passthrough) columns. Rejected for this PR: `RecordComparatorCompiler` operates on the base metadata, and lifting computed keys into it is a larger, separate change. The passthrough-only constraint keeps row-level semantics identical to the non-peeled plan.
- Three separate disjuncts (the original attempt before the refactor). Rejected: duplicated filter-stealing code across branches. The unified branch with wrapper-side defaults is smaller, uniformly testable, and lets nested wrappers compose.
- Revert `VirtualRecordCursorFactory.isProjection()` to `false` and keep the AsOf branch untouched. Rejected: it would force the top-K gate back to `instanceof VirtualRecordCursorFactory` checks, diverging from the unified-branch design.

## Changes

- `core/src/main/java/io/questdb/cairo/sql/RecordCursorFactory.java` — two new default methods: `translateOrderByColumnToBase` and `rewrapOverTopK`.
- `core/src/main/java/io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java` — overrides both defaults; `translateOrderByColumnToBase` guards bounds and chains through base; `rewrapOverTopK` recursively rewraps base first, then wraps over the result.
- `core/src/main/java/io/questdb/griffin/engine/table/VirtualRecordCursorFactory.java` — overrides `isProjection()` to return true; overrides both defaults. `translateOrderByColumnToBase` returns a negative value when the projected column is a computed expression (`ColumnFunction` + `PriorityMetadata.getBaseColumnIndex()` check), making the gate fall back to `Sort light` transparently.
- `core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java` — unified top-K branch in `generateOrderBy` with single-pass key translation (collapses the two former passes over `listColumnFilterA`); a `canReachPageFrameLeafForTopK` predicate; a shared `buildAsyncTopKOverStolenFilter` helper with a leak guard around `compileWorkerFiltersConditionally`. Also narrows the slave-projection peel guards in `generateJoinAsof` (lines 4369 and 4502) with an explicit `getColumnCrossIndex() != null` check (see Incidental fix above).
- `core/src/test/java/io/questdb/test/griffin/ExplainPlanTest.java` — plan-shape tests for single-key and multi-key `ORDER BY` through both wrappers, plus negative cases that pin the `Sort light` fallback when guards don't hold.
- `core/src/test/java/io/questdb/test/griffin/LimitTest.java` — row-level correctness tests through the full WAL lifecycle. `ts2` uses a reversed sequence (`(1_001 - x)::timestamp`) so a regression mistranslating `ORDER BY ts2` to the designated timestamp `ts` surfaces as a row-order diff. The virtual-projection test runs five iterations to catch resource-cleanup regressions.
- `core/src/test/java/io/questdb/test/griffin/SqlOptimiserTest.java` — updates the two `testJoinAndUnionQueryWithJoinOnDesignatedTimestampColumnWithLastFunction` plan assertions to the new top-K shape: `SelectedRecord` now sits above `Async Top K`, and the key references the base column (`ts`) rather than the projected alias (`LAST`). Behavior is unchanged; only the plan arrangement reflects the unified-branch design.
- `core/src/test/java/io/questdb/test/griffin/engine/join/AsOfJoinTest.java` — adds `testAsOfJoinNoKeyWithTopDownFilteredDateaddTimestampProjection`, the non-keyed analog of the existing keyed regression test. Asserts row-level results so a regression in the line-4502 guard surfaces both with `-ea` (assertion error) and without (silent wrong results).

## Tradeoffs and limitations

- When the gate fires above a projection wrapper, queries that were already fast on `Sort light` (small result set after filtering) pay the parallel top-K worker-queue setup cost. The break-even point is dataset-dependent.
- `ORDER BY` on a computed column still uses `Sort light`. Lifting computed keys into `RecordComparatorCompiler` is a larger follow-up.
- The predicate and call site peel only a single projection layer. Nested projection wrappers above the filter fall back to `Sort light`.
- Widening `isProjection()` to cover `VirtualRecord` required guards in `generateJoinAsof`'s slave-projection peels at lines 4369 and 4502 (see Incidental fix). No functional divergence for pre-PR inputs: factories that used to enter those branches still enter them; factories that used to skip them still skip them.

## Test plan

- `mvn -Dtest=ExplainPlanTest test` passes, including new plan-shape tests for single-key, multi-key, chained-wrapper, and negative-fallback cases.
- `mvn -Dtest=LimitTest test` passes, including the top-K projection tests that exercise `drainWalQueue()` and assert row ordering with the reversed `ts2` sequence.
- `mvn -Dtest=ParallelTopKFuzzTest test` passes.
- `mvn -Dtest=OrderByAdviceTest,OrderByWithFilterTest,OrderByWithAsyncFilterTest,OrderByExpressionTest,OrderByWithIntervalFilterTest test` passes unchanged.
- `mvn -Dtest=SqlOptimiserTest#testJoinAndUnionQueryWithJoinOnDesignatedTimestampColumnWithLastFunction,SqlOptimiserTest#testQueryPlanForJoinAndUnionQueryWithJoinOnDesignatedTimestampColumnWithLastFunction test` passes with the updated plan assertions.
- `mvn -Dtest=AsOfJoinTest#testAsOfJoinWithTopDownFilteredDateaddTimestampProjection test` passes with the keyed slave-peel guard (line 4369).
- `mvn -Dtest=AsOfJoinTest#testAsOfJoinNoKeyWithTopDownFilteredDateaddTimestampProjection test` passes with the non-keyed slave-peel guard (line 4502); row-level assertions catch both the `-ea` assertion path and the production silent-wrong-result path.
- Manual verification against the issue reproduction: all four queries either keep their prior `Async JIT Top K` plan or move from `Sort light` to `Async JIT Top K`.


DHRUV6029:
@coderabbitai review

DHRUV6029:
@coderabbitai review

DHRUV6029:
@puzpuzpuz @bluestreak01 would you mind taking a look when you get a chance? Thanks!

puzpuzpuz:
There seems to be a slightly simpler approach. Could you take a look?

# PR #6993 — Simpler approach

A proposal to collapse the three near-identical filter-stealing branches in `SqlCodeGenerator.generateOrderBy` into one, by pushing the projection-specific ORDER BY translation onto the wrapper factories.

## Problem with the current PR

The PR adds two new disjuncts (SelectedRecord peel, VirtualRecord peel) next to the existing direct filter-stealing branch. All three do the same high-level work:

1. `halfClose()` a filter-stealable factory.
2. Take ownership of `compiledFilter` / `filter` / `bindVarMemory` /
   `bindVarFunctions`.
3. Build `AsyncTopKRecordCursorFactory` on the page-frame leaf.
4. Optionally re-wrap the result with the projection wrapper.

The only thing that varies between the three branches is how an ORDER BY column index expressed in the **outer** metadata is translated to the column index in the **base** (page-frame) metadata:

| Branch | Translation |
| --- | --- |
| Direct (existing) | identity (projected == base) |
| SelectedRecord peel | `columnCrossIndex.getQuick(projectedIdx)` |
| VirtualRecord peel | `priorityMetadata.getBaseColumnIndex(columnFn.getColumnIndex())` with a passthrough-only bail |

Because that one varying piece is inlined into each branch, each branch carries its own copy of the steal/build/re-wrap scaffold. The author's own tradeoffs section acknowledges this:

> The shared helper takes the already-translated `ListColumnFilter` and
> first-key index as inputs, so the two disjuncts still carry their own
> translation logic inline.

The result: the VirtualRecord rewrap initially passed `virtualFactory.getMetadata()` instead of `orderedMetadata` (caught by CodeRabbit in afcf38e929). The SelectedRecord branch got it right. That kind of drift is exactly what duplication invites.

## Proposed shape

Push the translation onto the wrapper factory, and have one top-K branch consume it. The default on `RecordCursorFactory` is identity; the two wrapper types override it.

### 1. A translation contract on the factory

```java
// RecordCursorFactory.java

/**
 * Translates an ORDER BY column index expressed in this factory's output
 * metadata to the corresponding column index in the base (page-frame) metadata.
 *
 * Returns the input unchanged for factories that do not re-arrange or hide
 * base columns. Returns a negative value if the projected column cannot be
 * resolved to a base column (e.g. a computed VirtualRecord column); callers
 * must fall back to the generic sort path in that case.
 */
default int translateOrderByColumnToBase(int projectedIndex) {
    return projectedIndex;
}
```

### 2. Wrapper overrides

```java
// SelectedRecordCursorFactory.java

@Override
public int translateOrderByColumnToBase(int projectedIndex) {
    return columnCrossIndex.getQuick(projectedIndex);
}
```

```java
// VirtualRecordCursorFactory.java

@Override
public int translateOrderByColumnToBase(int projectedIndex) {
    Function fn = functions.getQuick(projectedIndex);
    if (!(fn instanceof ColumnFunction columnFn)) {
        return -1;
    }
    return priorityMetadata.getBaseColumnIndex(columnFn.getColumnIndex());
}
```

### 3. One top-K branch in `generateOrderBy`

The three current branches collapse into one. Pseudocode (not final — shows the shape, not every edge case already handled in the method):

```java
// Single top-K branch. Handles:
//   - direct page-frame or direct filter-stealing factory (identity translation)
//   - projection wrapper (SelectedRecord or VirtualRecord) around a filter-stealing factory
if (parallelTopKEnabled && canReachPageFrameLeafForTopK(recordCursorFactory)) {
    final RecordCursorFactory projectionWrapper = recordCursorFactory.isProjection()
            || recordCursorFactory instanceof VirtualRecordCursorFactory
            ? recordCursorFactory
            : null;
    final RecordCursorFactory filterFactory = projectionWrapper != null
            ? projectionWrapper.getBaseFactory()
            : recordCursorFactory;
    final RecordCursorFactory pageFrameLeaf = filterFactory.supportsFilterStealing()
            ? filterFactory.getBaseFactory()
            : filterFactory;
    final RecordMetadata baseMetadata = pageFrameLeaf.getMetadata();

    // Translate every ORDER BY key, bail if any cannot be resolved.
    final ListColumnFilter baseOrderByFilter = new ListColumnFilter();
    int baseFirstOrderByIdx = -1;
    boolean allKeysResolved = true;
    for (int i = 0, n = listColumnFilterA.size(); i < n; i++) {
        int signed = listColumnFilterA.getQuick(i);
        int projectedIdx = (signed > 0 ? signed : -signed) - 1;
        int baseIdx = recordCursorFactory.translateOrderByColumnToBase(projectedIdx);
        if (baseIdx < 0) {
            allKeysResolved = false;
            break;
        }
        baseOrderByFilter.add(signed > 0 ? (baseIdx + 1) : -(baseIdx + 1));
        if (i == 0) {
            baseFirstOrderByIdx = baseIdx;
        }
    }

    if (allKeysResolved) {
        IQueryModel.restoreWhereClause(expressionNodePool, model);

        final RecordCursorFactory topK = buildAsyncTopKOverStolenFilter(
                executionContext,
                filterFactory,
                pageFrameLeaf,
                baseMetadata,
                baseOrderByFilter,
                baseFirstOrderByIdx,
                lo
        );

        return projectionWrapper == null
                ? topK
                : projectionWrapper.rewrapOverTopK(topK, orderedMetadata);
    }
}
```

`rewrapOverTopK` is the small amount of rewrap logic that does differ per wrapper type (one re-creates a `SelectedRecordCursorFactory`, the other a `VirtualRecordCursorFactory` reusing `priorityMetadata` + `functions`). It is the only projection-specific piece left; everything else is one path.

`canReachPageFrameLeafForTopK` is a cheap structural predicate: the outer factory itself is a filter-stealing factory or a page-frame factory, or is a projection wrapper over one. It replaces the per-branch guard conditions.

### 4. `buildAsyncTopKOverStolenFilter` stays

The helper introduced by the PR stays — with one signature change so callers pass a `filterFactory` (which may be the outer factory itself in the direct case, or the projection base in the peel case). When `filterFactory.supportsFilterStealing()` is false (pure page-frame case, no filter to steal), the helper short-circuits and calls `AsyncTopKRecordCursorFactory` with null filter state. That already matches what the existing direct branch does at line 6554 when the factory supports page frames but not filter stealing.

## What this fixes

### Duplication goes away

Three copies of "steal filter state and build top-K" become one. The only code that varies per wrapper is the translation (4 lines) and the rewrap (one factory constructor call). Both live on the wrapper type that understands
them.

### The `orderedMetadata` class of bug becomes unrepresentable

The current PR's VirtualRecord rewrap was shipped with `virtualFactory.getMetadata()` and had to be fixed in a follow-up commit after CodeRabbit flagged it. In the proposed shape, `rewrapOverTopK` is called once from the top-K branch with a single `orderedMetadata` argument, so both wrapper implementations receive the same value and cannot disagree.

### Extending to a new wrapper is one method, not one branch

If a future wrapper type wants to participate in top-K stealing, the change is to implement `translateOrderByColumnToBase` and `rewrapOverTopK` on that type. No new disjunct in `generateOrderBy`, no new copy of the steal/build scaffold.

## Migration notes

- The AsOf precedent at lines 4264-4308 / 4396-4430 has the same two-disjunct shape (SelectedRecord peel + direct filter stealing, no VirtualRecord). The same refactor applies there. Not required for this PR — file as a follow-up
  so the blast radius stays contained.
- Keep the existing tests unchanged. They pin plan shapes; the proposed refactor produces the same plans, so the same assertions hold.
- The test gap noted in the review (`SelectedRecord -> VirtualRecord -> filter -> pageframe` at a single model level) should be added regardless of which approach lands. In the proposed shape it is a natural case because `translateOrderByColumnToBase` can be chained through a nested wrapper by delegating to the base factory's own implementation — something the current PR cannot do without a third disjunct.

## Scope

- Touches `RecordCursorFactory` (one default method), `SelectedRecordCursorFactory` and `VirtualRecordCursorFactory` (one override each, plus a small `rewrapOverTopK` helper), and the top-K branch in
  `SqlCodeGenerator.generateOrderBy` (replaces three disjuncts with one).
- Net line count: fewer than the current PR.
- Keeps `buildAsyncTopKOverStolenFilter` introduced by the PR; generalises its caller set from two to one.

DHRUV6029:
@coderabbitai review

DHRUV6029:
@coderabbitai review this PR critically

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -6525,56 +6624,64 @@ private RecordCursorFactory generateOrderBy(
...
+                                                    pageFrameLeaf,
+                                                    baseMetadata,
+                                                    baseOrderByFilter,
+                                                    baseFirstOrderByIdx,
+                                                    lo
+                                            );
+
+                                            return projectionWrapper == null
+                                                    ? topK
+                                                    : projectionWrapper.rewrapOverTopK(topK, orderedMetadata);
```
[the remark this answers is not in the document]
DHRUV6029: Thanks for the catch on this one. I checked it carefully and I think this is a false positive: `GenericRecordMetadata extends AbstractRecordMetadata implements RecordMetadata, Mutable` — it is not `Closeable` / `QuietCloseable` and holds no native resources (only Java `ObjList`/`String` fields). `Misc.free(orderedMetadata)` would be a no-op. I left the direct-return path as `return topK;`. Happy to add an explicit comment at the return site if you think it is worth documenting.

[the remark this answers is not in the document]
DHRUV6029: I think this one is also a false positive. The abandoned `projectionWrapper` matches the AsOf / LatestBy peel precedent explicitly referenced in the `rewrapOverTopK` Javadoc ("after this call the caller must not close the original wrapper; its state has either transferred to the returned factory or been dropped on the floor"). At codegen time the factory holds no open cursor resources — the inner `cursor` field is only wired up inside `getCursor()`. The inner `filterFactory` has already had its filter state stolen and `halfClose()`d, which is designed to leave it safe to drop. Adding an explicit `Misc.free(projectionWrapper)` here would double-free resources that ownership transfer already accounts for. If there is a real leak in this broader pattern, it would need to be addressed system-wide rather than in this PR.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -6525,56 +6630,66 @@ private RecordCursorFactory generateOrderBy(
...
```
DHRUV6029, line 7412: I think this one is also a false positive. The abandoned `projectionWrapper` matches the AsOf / LatestBy peel precedent explicitly referenced in the `rewrapOverTopK` Javadoc ("after this call the caller must not close the original wrapper; its state has either transferred to the returned factory or been dropped on the floor"). At codegen time the factory holds no open cursor resources — the inner `cursor` field is only wired up inside `getCursor()`. The inner `filterFactory` has already had its filter state stolen and `halfClose()`d, which is designed to leave it safe to drop. Adding an explicit `Misc.free(projectionWrapper)` here would double-free resources that ownership transfer already accounts for. If there is a real leak in this broader pattern, it would need to be addressed system-wide rather than in this PR.

core/src/main/java/io/questdb/griffin/engine/table/VirtualRecordCursorFactory.java
```diff
@@ -159,6 +165,18 @@ public boolean recordCursorSupportsRandomAccess() {
         return supportsRandomAccess;
     }
 
+    @Override
+    public RecordCursorFactory rewrapOverTopK(RecordCursorFactory topK, RecordMetadata orderedMetadata) {
+        RecordCursorFactory rewrappedBase = base.rewrapOverTopK(topK, base.getMetadata());
```
DHRUV6029, line 170: Intentional. The default `rewrapOverTopK` implementation ignores its metadata argument and returns `topK` unchanged, so passing `base.getMetadata()` here is a provable no-op today. I considered passing `topK.getMetadata()` instead but `base.getMetadata()` is more consistent with "recurse into base before rewrapping", which is what this recursive call is meant to express. If a future factory overrides `rewrapOverTopK` in a way that depends on the metadata argument, the contract would need clarification regardless. Happy to revisit if you have a concrete scenario in mind.

DHRUV6029:
@puzpuzpuz I refactored my approach based on your suggestions,  please review it 
Thanks

puzpuzpuz:
Could you check this list?

## Critical

### 1. `generateJoinAsof` line 4502: missing `getColumnCrossIndex() != null` guard (same bug as fixed at 4369)

**Status:** CONFIRMED with a reproducer that crashes on the PR branch with `AssertionError`.

The PR widens `VirtualRecordCursorFactory.isProjection()` to return `true`. The author correctly added a guard at `SqlCodeGenerator.java:4369` for the **keyed** AsOf peel:

```java
} else if (slave.isProjection() && slave.getColumnCrossIndex() != null) {
    // peel requires a column cross index; only SelectedRecord provides one
```

But the **non-keyed** AsOf peel at `SqlCodeGenerator.java:4502` has the identical pattern and was left unguarded:

```java
if (slave.isProjection()) {
    RecordCursorFactory projectionBase = slave.getBaseFactory();
    ...
    if (projectionBase.supportsFilterStealing()) {
        ...
        if (filterStealingBase.supportsTimeFrameCursor()) {
            IntList stolenCrossIndex = slave.getColumnCrossIndex();
            assert stolenCrossIndex != null;     // line 4514 — fires for VirtualRecord slaves
```

`VirtualRecordCursorFactory` does not override `getColumnCrossIndex()`, so it returns the `null` default. Before this PR, `isProjection()` was only true for `SelectedRecord`, so the branch was never reached with a `VirtualRecord` slave. After the widening, `VirtualRecord` slaves sitting over a filter-stealing + time-frame-capable base enter this branch and hit `assert stolenCrossIndex != null`.

**Reproducer** (dropped into `core/src/test`, run on checked-out `pr-6993`):

```java
execute("CREATE TABLE m (ts TIMESTAMP, a INT) TIMESTAMP(ts) PARTITION BY DAY");
execute("CREATE TABLE s (ts TIMESTAMP, b DOUBLE) TIMESTAMP(ts) PARTITION BY DAY");
execute("INSERT INTO m VALUES ('2026-01-01T00:00:10.000000Z', 1), ...");
execute("INSERT INTO s VALUES ('2026-01-01T00:00:05.000000Z', 1.0), ...");
assertSql("...",
    "SELECT m.*, s_sub.* FROM m ASOF JOIN ("
    + " SELECT dateadd('s', 30, ts) AS ts, b * 2 AS computed FROM s WHERE b > 0"
    + ") s_sub");

```

Crashes with:

```
java.lang.AssertionError
    at io.questdb.griffin.SqlCodeGenerator.generateJoinAsof(SqlCodeGenerator.java:4514)
```

Impact:

- With `-ea` (QuestDB CI): `AssertionError` kills compilation for a previously-working query class.
- Without `-ea` (prod): `null` crossIndex is passed into `FilteredAsOfJoinNoKeyFastRecordCursorFactory`, which tolerates null as "no remap" — silently dropping the `VirtualRecord` layer's computed functions and producing **wrong results** (the computed slave columns disappear).

**Fix:** Add the same guard at line 4502:

```java
} else if (slave.isProjection() && slave.getColumnCrossIndex() != null) {
```

and add a regression test for the non-keyed path (e.g., `AsOfJoinNoKeyTest`) analogous to `testAsOfJoinWithTopDownFilteredDateaddTimestampProjection` but without the `ON (…)` clause.

## Moderate

### 2. No regression test for the non-keyed AsOf + VirtualRecord slave scenario

The PR acknowledges the `isProjection()` widening and its interaction with the AsOf peel, but the only regression test cited (`testAsOfJoinWithTopDownFilteredDateaddTimestampProjection`) exercises the **keyed** path (line 4369). The parallel non-keyed path at line 4502 has no coverage, which is how the bug above slipped through.

Once the critical fix is in, add a non-keyed variant. Grepping `AsOfJoinNoKeyTest.java` shows no `VirtualRecord`/`dateadd`/`Projection` coverage.

## Minor

### 3. Translation loop runs twice over `listColumnFilterA`

`SqlCodeGenerator.java:6655-6675` iterates `listColumnFilterA` once to validate `allKeysResolved`, then a second time to build `baseOrderByFilter`. Could be a single pass: build into `baseOrderByFilter` as you go and abandon if a key fails to resolve. Purely a micro-refactor; `listColumnFilterA` size is small.

DHRUV6029:
@puzpuzpuz thanks for the careful review. All three items addressed in `e74ba041b0`.

## 1. Critical — `generateJoinAsof` line 4502 guard

Added the same `&& slave.getColumnCrossIndex() != null` guard at line 4502 that already exists at line 4369:

```java
if (slave.isProjection() && slave.getColumnCrossIndex() != null) {
    // peel requires a column cross index; only SelectedRecord provides one
```

This restores the pre-PR narrow meaning at that specific call site. `VirtualRecord` slaves now fall through to the generic `AsOfJoinLightNoKeyRecordCursorFactory` path (the same path they took before `isProjection()` was widened), instead of entering a peel branch that needs a cross index they don't have.

I also called out the production-correctness risk in the commit body and the PR description: with `-ea` off, `FilteredAsOfJoinNoKeyFastRecordCursorFactory` treats a null `slaveColumnCrossIndex` as "no remap" and would have silently dropped the `VirtualRecord` layer's translation. Worth flagging because it would have shipped as wrong results, not a crash.

## 2. Moderate — non-keyed regression test

Added `AsOfJoinTest.testAsOfJoinNoKeyWithTopDownFilteredDateaddTimestampProjection`, the non-keyed analog of the existing keyed test. Same fixture (`fx_trades` / `market_data` with a `dateadd`-style projected timestamp on the slave), same shape, just without the `ON (symbol)` clause.

It uses `assertQueryNoLeakCheck` against expected rows rather than `assertPlan`, so:

- with `-ea` (CI): catches the `AssertionError` regression at line 4514;
- without `-ea` (prod): catches the silent wrong-result regression in `FilteredAsOfJoinNoKeyFastRecordCursorFactory`.

Putting it in `AsOfJoinTest.java` rather than `AsOfJoinNoKeyTest.java` because `AsOfJoinNoKeyTest` is parameterized over `JoinType` (ASOF/LT) and the bug is `ASOF`-only (`generateJoinAsof` is only invoked from `JOIN_ASOF`, line 4822-4823). Keeping it next to the keyed test also reads as a pair.

## 3. Minor — single-pass key translation in `generateOrderBy`

Collapsed the two iterations over `listColumnFilterA` into one. The ListColumnFilter is now built incrementally and the loop bails on the first unresolved key:

```java
final ListColumnFilter baseOrderByFilter = new ListColumnFilter();
int baseFirstOrderByIdx = -1;
boolean allKeysResolved = true;
for (int i = 0, n = listColumnFilterA.size(); i < n; i++) {
    int signed = listColumnFilterA.getQuick(i);
    int projectedIdx = (signed > 0 ? signed : -signed) - 1;
    int baseIdx = recordCursorFactory.translateOrderByColumnToBase(projectedIdx);
    if (baseIdx < 0) {
        allKeysResolved = false;
        break;
    }
    baseOrderByFilter.add(signed > 0 ? (baseIdx + 1) : -(baseIdx + 1));
    if (i == 0) {
        baseFirstOrderByIdx = baseIdx;
    }
}
```

This matches the pseudocode in your earlier "simpler approach" proposal verbatim. The fallback semantics are unchanged: when `allKeysResolved` is false, the partially-built `baseOrderByFilter` is discarded and the query falls through to `Sort light`.

## Verification

Local runs all green:

- `AsOfJoinTest#testAsOfJoinNoKeyWithTopDownFilteredDateaddTimestampProjection` and `testAsOfJoinWithTopDownFilteredDateaddTimestampProjection`: both pass.
- `ExplainPlanTest`, `LimitTest`, `ParallelTopKFuzzTest`: 576 tests, 0 failures.
- `SqlOptimiserTest`: 172 tests, 0 failures.

PR description has been updated to document both peel guards and the test pair.

DHRUV6029:
Hello @puzpuzpuz addressed all the concerns and issues , Please take a look
Thanks

DHRUV6029:
Hello @puzpuzpuz Gentle Reminder to please review the PR.
Thanks

puzpuzpuz:
@DHRUV6029 sorry for the delay. Your PR is on my radar. Going to check the updates today.

puzpuzpuz:
Minor review comments only. Would be great to address item 1 while we're on it. Others are nit.

## Minor

### 1. `VirtualRecordCursorFactory.translateOrderByColumnToBase` does not peel `MemoizerFunction`

**File:** `core/src/main/java/io/questdb/griffin/engine/table/VirtualRecordCursorFactory.java:197`

```java
Function fn = functions.getQuick(projectedIndex);
if (!(fn instanceof ColumnFunction columnFn)) {
    return -1;
}
```

When an alias is referenced more than once, `SqlCodeGenerator.generateSelectVirtual` (around line 8333) wraps the underlying `ColumnFunction` in a `MemoizerFunction`. The codebase already provides `ColumnFunction.unwrap(fn)` precisely to peel these wrappers — and `generateSelectVirtual` uses it at line 8240. The new translation logic does not, so a passthrough column with memoization silently falls back to `Sort light`.

This is consistent with the existing `VirtualFunctionRecordCursor.getLongTopKColumnIndex` (line 95) which has the same gap — so it is a missed-optimization carry-over, not a regression. Worth fixing both at once with `ColumnFunction.unwrap(fn)`.

### 2. Test name and comment for `testSelectWhereOrderByLimit_chainedWrapper` are misleading

**File:** `core/src/test/java/io/questdb/test/griffin/ExplainPlanTest.java`

The test comment claims `translateOrderByColumnToBase` "must chain from `SelectedRecordCursorFactory` through `VirtualRecordCursorFactory` to the filter leaf". But `canReachPageFrameLeafForTopK` (`SqlCodeGenerator.java:1015`) returns `false` for `SelectedRecord(VirtualRecord(JITFilter))`:

- `SelectedRecord.supportsPageFrameCursor()` forwards to `VirtualRecord.supportsPageFrameCursor()` which returns `false` (default).
- `supportsFilterStealing()` is `false` for both.
- The predicate's projection branch only inspects `f.getBaseFactory()` once — the inner `VirtualRecord`, whose `supportsFilterStealing()` is `false`.

The recursive call `base.rewrapOverTopK(topK, base.getMetadata())` in `SelectedRecordCursorFactory.rewrapOverTopK` is also forward-only — no current path constructs a chained input.

In practice the test almost certainly hits the gate at the inner `VirtualRecord` level (single-layer peel), and the outer `SelectedRecord` is added later by the parent `generateSelectChoose` pass. The test still exercises a real path; the comment overstates what it covers. Either drop the comment or adjust the input (e.g. an explicit subquery) so the gate is actually invoked on a stacked wrapper.

### 3. `SelectedRecordCursorFactory.recordCursorSupportsLongTopK` lacks the same bounds guard as `translateOrderByColumnToBase`

**File:** `core/src/main/java/io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java:205`

```java
@Override
public boolean recordCursorSupportsLongTopK(int columnIndex) {
    return base.recordCursorSupportsLongTopK(columnCrossIndex.getQuick(columnIndex));
}
```

The PR (correctly) added a bounds check to `translateOrderByColumnToBase` at line 247. The neighbouring `recordCursorSupportsLongTopK` override does not bound-check `columnIndex`. Inconsistent. Both have the same metadata-derived caller today, so it is not exploitable, but the asymmetry is a footgun for future callers.

### 4. Redundant cleanup in `buildAsyncTopKOverStolenFilter`

**File:** `core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java:1468`

The new try/catch around `compileWorkerFiltersConditionally` frees the stolen state on failure. But `generateOrderBy`'s outer catch (line 6791) already calls `recordCursorFactory.close()`, which cascades into the (still-referenced) filter factory's `_close()` — see `AsyncJitFilteredRecordCursorFactory._close()` at line 409 — which frees `compiledFilter`, `filter`, `bindVarMemory`, and `bindVarFunctions` again.

The duplicate is harmless because all of those `close()` implementations are idempotent (e.g. `CompiledFilter.close()` guards on `fnAddress > 0`), but the pre-PR code did not have this redundancy. If you want to keep the explicit catch as defense in depth, fine; if not, the original pattern is cleaner.

### 5. Comments on the AsOf-join guards reference the SelectedRecord constraint without naming it

**File:** `core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java:4379, 4512`

The single-line comment "peel requires a column cross index; only SelectedRecord provides one" is fine but loses the rationale recorded at length in the PR description (the silent wrong-results path through `FilteredAsOfJoinNoKeyFastRecordCursorFactory` when the cross index is null). A one-liner pointing at the test name would make the intent harder to inadvertently undo.

DHRUV6029:
@puzpuzpuz all 5 minor items addressed in `746678e223` (single linear commit on top of `upstream/master`).

| # | Item | Status |
|---|------|--------|
| 1 | `ColumnFunction.unwrap(fn)` applied in `VirtualRecordCursorFactory.translateOrderByColumnToBase:197` and `VirtualFunctionRecordCursor.getLongTopKColumnIndex:98` so memoized aliases hit top-K instead of `Sort light` | Fixed |
| 2 | Test renamed `testSelectWhereOrderByLimit_chainedWrapper` -> `testSelectWhereOrderByLimit_virtualRecordPeel`; comment rewritten to describe the single-layer peel and note that the outer `SelectedRecord` is added later by `generateSelectChoose` | Fixed |
| 3 | Bounds guard added to `SelectedRecordCursorFactory.recordCursorSupportsLongTopK:205` so it matches the neighbouring `translateOrderByColumnToBase` override | Fixed |
| 4 | Inner try/catch removed from `buildAsyncTopKOverStolenFilter`; ownership cascades through `generateOrderBy`'s outer `recordCursorFactory.close()` -> `filterFactory._close()`, with a one-line comment recording the chain | Fixed |
| 5 | AsOf-guard comments at `SqlCodeGenerator.java:4373` and `:4509` now name `FilteredAsOfJoinNoKeyFastRecordCursorFactory` and the silent-wrong-results path, and point at `AsOfJoinTest#testAsOfJoinWithTopDownFilteredDateaddTimestampProjection` / `#testAsOfJoinNoKeyWithTopDownFilteredDateaddTimestampProjection` | Fixed |

Local regression run after the fixes: `ExplainPlanTest` (531), `AsOfJoinTest` (110), `OrderByLimitTest` (2) all green. Ready for another look when you have a moment.

core/src/main/java/io/questdb/griffin/engine/table/VirtualRecordCursorFactory.java
```diff
@@ -145,6 +146,11 @@ public boolean implementsLimit() {
         return base.implementsLimit();
     }
 
+    @Override
+    public boolean isProjection() {
+        return true;
```
puzpuzpuz: When `isProjection()` returns true, it means the following:
```
Returns true if the factory stands for nothing more but a projection, so that
the above factory (e.g. a parallel GROUP BY one) can steal the projection.

Projection consist of cross-indexes and metadata columns.
```

`VirtualRecordCursorFactory` can return arbitrary SQL functions, not just projected base factory's columns.

puzpuzpuz:
Critical — `ExtraNullColumnCursorFactory` is silently peeled by the new top-K gate

### Reproducer

```sql
CREATE TABLE trades (ts TIMESTAMP, sym SYMBOL, price DOUBLE) TIMESTAMP(ts) PARTITION BY DAY;
CREATE TABLE prices (ts TIMESTAMP, sym SYMBOL, price DOUBLE) TIMESTAMP(ts) PARTITION BY DAY;
INSERT INTO trades VALUES
    ('2023-01-01T09:10:00.000000Z', 'AAA', 100.0),
    ('2023-01-01T09:11:00.000000Z', 'BBB', 200.0),
    ('2023-01-01T09:12:00.000000Z', 'CCC', 300.0),
    ('2023-01-01T09:13:00.000000Z', 'DDD', 400.0);
INSERT INTO prices VALUES ('2023-01-01T09:00:00.000000Z', 'AAA', 1.0);

SELECT t.sym, t.price, t.ts, sum(p.price) AS window_price
FROM trades t
WINDOW JOIN prices p ON (0 = 1)
RANGE BETWEEN 1 MINUTE PRECEDING AND 1 MINUTE FOLLOWING
ORDER BY t.sym DESC LIMIT 3;
```

| | master | this PR |
| --- | --- | --- |
| plan | `Async Top K -> ExtraNullColumnRecord -> PageFrame` | `Async Top K -> PageFrame` (layer dropped) |
| output columns | `sym, price, ts, window_price` (4) | `sym, price, ts` (3) — `window_price` missing |

A second variant — `ORDER BY window_price DESC LIMIT 3` — throws `AssertionError: index out of bounds, 3 >= 3` from `baseMetadata.getColumnType(baseFirstOrderByIdx)` in `buildAsyncTopKOverStolenFilter` (`SqlCodeGenerator.java:1441`); without `-ea` the same path produces an `IndexOutOfBoundsException`.

### Root cause

`ExtraNullColumnCursorFactory.isProjection() == true` (line 146, pre-PR), but it does not override the new `translateOrderByColumnToBase` or `rewrapOverTopK`. The unified gate at `SqlCodeGenerator.java:6643-6697` then:

1. `canReachPageFrameLeafForTopK` returns true via the forwarded `supportsPageFrameCursor()`.
2. `projectionWrapper = ExtraNullColumnCursorFactory`, `pageFrameLeaf = master`.
3. `translateOrderByColumnToBase(idx)` falls through to the default identity. For `idx < columnSplit` it succeeds; for `idx >= columnSplit` it returns an OOB index that hits the assertion above.
4. `projectionWrapper.rewrapOverTopK(topK, orderedMetadata)` returns `topK` unchanged (default impl) — the `ExtraNullColumn` layer is silently dropped.

Pre-PR the gate wrapped `recordCursorFactory` directly as the top-K base, so `ExtraNullColumn.getPageFrameCursor()` continued to splice in the null columns. The PR sets `base = pageFrameLeaf`, which is the bare master.

### Tie-in with [r3180067265](https://github.com/questdb/questdb/pull/6993#discussion_r3180067265)

Same disease. The Javadoc on `isProjection()` says "projection consists of cross-indexes and metadata columns" — `SelectedRecord` and `ExtraNullColumn` (historically). The PR widens the predicate to mean "anything the new top-K gate can peel" and then has to add `getColumnCrossIndex() != null` guards at AsOf lines 4371 and 4507 to keep the *other* `isProjection()` consumers honest. Those guards restore the old meaning at those callsites; the new top-K gate at line 6653 has no such guard, so `ExtraNullColumn` slips through.

### Fix

Three options. The cleanest, IMO, is the first:

1. Revert `VirtualRecordCursorFactory.isProjection()` to `false`, drop the AsOf `getColumnCrossIndex() != null` guards (they become unnecessary), and give the top-K gate its own predicate — either `instanceof SelectedRecordCursorFactory || instanceof VirtualRecordCursorFactory` directly, or a separate default method (`canPeelForTopK()` or similar) that only the two new-overrides factories return `true` from. Net change: the unified gate handles exactly the two factories with the new overrides; `ExtraNullColumn` falls back to Sort light (same as master); the existing `isProjection()` contract is unchanged.

2. Keep `VirtualRecord.isProjection() = true` and add `getColumnCrossIndex() != null` to the new gate at line 6653, mirroring the AsOf guards. `ExtraNullColumn` (no cross-index) falls back. Minimal diff but keeps the overloaded predicate.

3. Override `translateOrderByColumnToBase` and `rewrapOverTopK` on `ExtraNullColumnCursorFactory` to preserve the optimization. Best perf, more code, and still leaves `isProjection()` overloaded.

Either way, please add a regression test using the SQL above (both the "ORDER BY real column" silent-rows variant and the "ORDER BY null column" assertion variant — the second catches the runtime path with `-ea` off).

puzpuzpuz:
Smaller items

### Moderate — missing regression test for `MemoizerFunction` peel

The PR added `ColumnFunction.unwrap` calls in two places (`VirtualRecordCursorFactory.translateOrderByColumnToBase:197` and `VirtualFunctionRecordCursor.getLongTopKColumnIndex:98`) explicitly to handle the memoizer wrapping that `generateSelectVirtual` inserts when an alias is referenced more than once. No test in the PR actually exercises that path — the new `LimitTest` and `ExplainPlanTest` cases all use single-reference projections. Without a test, future refactors can quietly regress the optimization back to `Sort light`. A query along the lines of `SELECT x AS a, a, a FROM ... WHERE ... ORDER BY a DESC LIMIT N` (or whichever shape forces memoization through `generateSelectVirtual`) would pin the new `unwrap` behavior.

### Minor — comment at `SqlCodeGenerator.java:1469` references `_close()` instead of `close()`

```
// generateOrderBy's outer catch closes recordCursorFactory, which cascades into the
// still-referenced filterFactory._close() and frees the stolen filter handles.
```

`_close()` is the protected template method on `AbstractRecordCursorFactory`; the cascading call from the outer catch goes through `close()`. Either drop the underscore or rephrase as "cascades through `close()` into `filterFactory._close()`".

### Nit — `VirtualRecordCursorFactory.rewrapOverTopK` shares `functions` between old and new wrapper

The new factory at `VirtualRecordCursorFactory.java:171-177` is constructed with the same `functions` `ObjList` reference as the orphan it replaces. The Javadoc on `RecordCursorFactory.rewrapOverTopK` already documents this ("its state has either transferred to the returned factory or been dropped on the floor, matching the AsOf/LatestBy peel precedent"), and the orphan is unreachable after the gate returns, so there's no double-free in practice. Worth a one-line code comment on the override pointing at the Javadoc — the shared mutable reference is non-obvious next to `_close()` calling `Misc.freeObjList(functions)`.

puzpuzpuz (commented):
The PR is almost good to go, thanks for handling all of the review feedback. Could you address the following moderate level items? The fixes are pretty straight-forward, so it shouldn't take long.

## Moderate

### M1 — Latent: recursive `rewrapOverTopK` passes pre-ORDER-BY metadata down the chain

**Files:** `core/src/main/java/io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java:218-224`, `core/src/main/java/io/questdb/griffin/engine/table/VirtualRecordCursorFactory.java:169-181` (in-diff)

Both overrides do:

```java
RecordCursorFactory rewrappedBase = base.rewrapOverTopK(topK, base.getMetadata());
```

`base.getMetadata()` is the wrapper's pre-ORDER-BY metadata, with the *original* timestamp index. Today the predicate `canReachPageFrameLeafForTopK` peels at most one wrapper, so the recursive call always lands on a non-projection base whose default `rewrapOverTopK` ignores the metadata argument — the bug is dormant. But the file comment at `SqlCodeGenerator.java:1052-1056` explicitly contemplates widening the predicate. The moment a future change lets the chain run two layers deep (e.g. `SR(VR(filter(pageFrame)))`), the inner wrapper's `getMetadata().getTimestampIndex()` will lie about ordering whenever `AsyncTopK`'s `baseOrderedMetadata` set the timestamp index to `-1` (ORDER BY a non-timestamp column).

**Fix:** either thread the gate's `orderedMetadata` down each layer with the appropriate timestamp index, or `assert !base.isProjection();` at the recursive call site so the trap is loud if the predicate is ever widened.

### M2 — Comment in the keyed AsOf guard describes the wrong factory and wrong failure mode

**File:** `core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java:4420-4424` (in-diff)

The comment in the **keyed** AsOf branch reads:

> Without it, **FilteredAsOfJoinNoKeyFastRecordCursorFactory** silently produces wrong results …

But this branch builds `FilteredAsOfJoinFastRecordCursorFactory` (line 4396), not the NoKey variant. The PR description itself notes that the keyed branch trips `assert stolenCrossIndex != null` — i.e. it asserts, it does not silently produce wrong results. The non-keyed branch at line 4556 correctly references the NoKey class and the silent-wrong-results path.

**Fix:** rewrite the keyed comment to reference `FilteredAsOfJoinFastRecordCursorFactory` and the assertion failure, so future readers don't conflate the two failure modes.

### M3 — `isProjection()` Javadoc is now inconsistent with the widened semantics

**File:** `core/src/main/java/io/questdb/cairo/sql/RecordCursorFactory.java:270-282` (out-of-diff but directly relevant)

The Javadoc still reads:

> Returns true if the factory stands for nothing more but a projection… **Projection consist of cross-indexes and metadata columns.**

`VirtualRecordCursorFactory.isProjection()` now returns `true` even though the wrapper holds arbitrary SQL functions (not pure column references). The PR's defence is the new `&& slave.getColumnCrossIndex() != null` guard at the two AsOf join sites, which correctly excludes `VirtualRecord`. Without a doc fix, future callers of `isProjection()` will keep assuming the old narrower contract (cross-index always non-null).

**Fix:** update the Javadoc to document the widened contract — something like *"the wrapper hides a steal-able base; cross-index/base-column-name are guaranteed only when `getColumnCrossIndex() != null`"*.

### M4 — `ExtraNullColumnCursorFactory` is the third `isProjection()=true` class and silently relies on never reaching the new gate

**Files:** `core/src/main/java/io/questdb/griffin/engine/table/ExtraNullColumnCursorFactory.java:145, 169-171` (out-of-diff)

`ExtraNullColumnCursorFactory.isProjection()` returns `true` (pre-existing), `getColumnCrossIndex()` is the default `null`, and `supportsPageFrameCursor()` forwards to `base`. It does not override `translateOrderByColumnToBase` or `rewrapOverTopK`. Today it only wraps a window-join `master` (which doesn't support page-frame cursors), so `canReachPageFrameLeafForTopK` rejects it and the gate doesn't fire.

The arrangement is fragile: any future change that makes ExtraNull wrap a page-frame-capable base, or any base whose `supportsFilterStealing()` is true, will silently flow into the new gate and produce wrong results because:

1. the default identity `translateOrderByColumnToBase` ignores the null-padding split, and
2. the default `rewrapOverTopK` returns the bare `topK` and drops the null-padding wrapper entirely (output column count would shrink).

**Fix:** either gate the new path on `getColumnCrossIndex() != null` (mirroring the AsOf fix at lines 4420 and 4556) or have ExtraNull explicitly override both methods even with no-op semantics that fall back to `Sort light`.

DHRUV6029:
@puzpuzpuz Thanks for the deep review. Applied **option 1** (your preferred): introduced a dedicated `canPeelForTopK()` default method, reverted `VirtualRecordCursorFactory.isProjection()` to false, and dropped the AsOf `getColumnCrossIndex() != null` guards now that they're redundant. Pushed in `038d422`.

| Issue | Status | Fix |
| --- | --- | --- |
| **Critical** — `ExtraNullColumnCursorFactory` silently peeled by top-K gate | Fixed | New `canPeelForTopK()` predicate; only Selected and Virtual override it. ExtraNullColumn (and any future projection-like wrapper) inherits `false` by default and is correctly excluded. |
| AsOf `getColumnCrossIndex() != null` guards | Removed | Restored to `slave.isProjection()` only — guards are unnecessary now that `isProjection()` is back to its original narrow "cross-indexes and metadata columns" meaning. |
| `VirtualRecord.isProjection()` widening | Reverted | Override removed; falls back to default `false`. |

**Regression tests** (LimitTest):
- `testTopKDoesNotPeelExtraNullColumnSplice` — your WINDOW JOIN repro with `ORDER BY t.sym DESC LIMIT 3`. Confirms the splice layer survives (output keeps all 4 columns including `window_price`).
- `testTopKDoesNotPeelExtraNullColumnSpliceOrderByNullColumn` — your repro with `ORDER BY window_price DESC LIMIT 3`. Confirms no `AssertionError` (with `-ea`) or `IndexOutOfBoundsException` (without `-ea`); top-K runs to completion over the spliced layer.

Both new tests fail on the prior version of this branch and pass after the fix. Full local sweep: `LimitTest` (all green), `ExplainPlanTest` (574 tests, 0 failures, 2 pre-existing skips), `AsOfJoinTest` + `AsOfJoinNoKeyTest` + `LtJoinTest` (120 tests, 0 failures), `WindowJoinTest` (352 tests, 0 failures, 186 parameterization skips).

The earlier smaller items from your 2026-05-04T08:24:13Z comment (memoizer test, `_close()` comment wording, shared-functions comment on `rewrapOverTopK`) were already addressed in the prior round; verified still in place.

Ready for another look.

core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java
```diff
@@ -1040,6 +1040,29 @@ private static void buildMultiHorizonColumnMappings(
...
+    // light rather than recursing; that keeps the predicate O(1) and the
+    // wrapper overrides for translateOrderByColumnToBase / rewrapOverTopK
+    // still chain correctly when they run.
+    private static boolean canReachPageFrameLeafForTopK(RecordCursorFactory f) {
+        if (f.supportsPageFrameCursor() || f.supportsFilterStealing()) {
+            return true;
+        }
+        if (f.canPeelForTopK()) {
+            RecordCursorFactory base = f.getBaseFactory();
+            return base != null && (base.supportsPageFrameCursor() || base.supportsFilterStealing());
```
puzpuzpuz: One final tiny tweak: let's reject stacked wrappers here to avoid bugs later on, if optimizer's behavior changes:
```suggestion
            return base != null && !base.canPeelForTopK() && (base.supportsPageFrameCursor() || base.supportsFilterStealing());
```

puzpuzpuz (approved):

puzpuzpuz:
Thanks for the contribution!

PR #6959, opened 2026-04-11
perf(sql): speed up parallel keyed GROUP BY queries by introducing batch calls

puzpuzpuz:
## Summary

This patch introduces batched aggregate dispatch for the parallel keyed GROUP BY path. Instead of per-row virtual dispatch through `GroupByFunctionsUpdater` → `GroupByFunction.computeFirst/computeNext` → `MapValue.putLong/addLong`, the reducer splits each page frame into sub-batches (2048 rows by default, configurable via `cairo.sql.parallel.groupby.batch.size`) and processes them in two phases:

1. **Probe phase** — find or create map entries and pack row index + entry offset into a scratch buffer. For example, `Unordered4Map.probeBatch()` inlines the hash lookup with direct field access, bypassing the `MapKey`/`MapValue` abstraction. A fast path reads keys directly from column memory when the key is a single fixed-size column.

2. **Update phase** — call `GroupByFunction.computeKeyedBatch()` once per function per sub-batch. The default implementation loops over the batch and delegates to `computeFirst`/`computeNext`. Hot functions (`count`, `sum`, `min`, `max`, `avg`, `bit_and`/`bit_or`/`bit_xor`) override with tight `Unsafe` loops that skip the per-row `MapValue` dispatch and, when the argument is a direct column, read values straight from the page frame.

Each batch entry is a packed long: `[isNew:1][rowIndex:24][offset:39]`.

### Limitations

* Sharded reducer code path stays as is for now as the whole sharding design may change
* Batched aggregation is disabled in case of late materialization

### Other changes

Also fixes a pre-existing data-correctness bug in `count(uuid)`. On master, `CountUuidGroupByFunction` read the high 64 bits of the UUID argument twice instead of reading both halves — so any UUID whose high half equals `Long.MIN_VALUE` (the NULL-hi sentinel) was treated as NULL regardless of its low half, and silently excluded from the count. Fixed by reading the low half via `getLong128Lo`.

## TODO (to productionize)

- [x] Etalon buffer: pre-initialize new map entries with identity-element bytes (`GroupByFunction#setEmpty()`'s result) so `computeKeyedBatch` overrides can skip new/existing branching
- [x] Implement `computeKeyedBatch` overrides for hot functions: sum, min, max, avg, bitwise
- [x] Consider other `computeKeyedBatch` override opportunities for hot functions
- [x] Support filtered path (`aggregateFilteredNonSharded`)
- [x] Apply the same pattern to `Unordered8Map` and `OrderedMap`
[...]

PR #6997, opened 2026-04-21
perf(sql): speed up GROUP BY, hash joins, and count_distinct with a faster hash finalizer

puzpuzpuz:
Depends #6959

## Summary

Consolidates QuestDB's non-Murmur 64-bit hash paths onto a single finalizer derived from xxHash3's 64-bit avalanche, and raises the load factor of the affected hash tables from `0.5` to `0.7`.

Before this change, `Hash.java` exposed three competing 64-bit mixers:

- `fmix64` (Murmur3 finalizer) - 2 multiplies, ideal avalanche.
- `fastHashInt64` / `fastHashLong64` (FxHasher) - 1 multiply, poor avalanche, used in six hot hash tables with a compensating `0.5` load factor.
- (new) `xxh3Avalanche64` - 1 multiply + 2 shift-XORs, ideal avalanche.

`xxh3Avalanche64` is roughly 1.7x faster than `fmix64` in latency mode and matches its quality tier. Once it's available, the `fmix64` / `fastHash*` split no longer earns its keep, so `fastHash*` is removed and every non-Murmur finalizer in `Hash.java` routes through the new mixer. `fmix64` survives as a private helper used only by `murmur3ToLong`, which `ApproxCountDistinct*` (HyperLogLog) depends on bit-for-bit.

The six former `fastHash*` callers had `0.5` load factors chosen to absorb FxHasher's weak avalanche on sequential-ish inputs. With the stronger mixer they have headroom to run denser, so their load factor moves to `0.7` - the project's standard for `Unordered4Map` / `Unordered8Map` / `OrderedMap`.

## Motivation

PR #6959 (batched GROUP BY aggregate dispatch) exposed `Hash` as a first-order cost in the reducer. Profiling showed the finalizer on the critical path of `Unordered4Map` / `Unordered8Map` probes, which motivated revisiting the mixer.

A secondary motivation: removing `fastHash*` eliminates a footgun. FxHasher's `0.5` load factor requirement was easy to forget at new call sites, and its avalanche is catastrophically bad on low-entropy keys (`avgProbes > 250_000` on `i << 20` style inputs at `0.48` load).

### `Hash.java`

- New private `xxh3Avalanche64(long)`:

```java
  private static long xxh3Avalanche64(long h) {
      h ^= h >>> 37;
      h *= 0x165667919E3779F9L; // xxh3 prime
      return h ^ (h >>> 32);
  }
  ```

- `hashInt64`, `hashLong64`, `hashLong128_64`, `hashLong256_64`, `hashMem64`, `hashUtf8(Utf8String)` now finalize via `xxh3Avalanche64`.
- `fastHashInt64` / `fastHashLong64` deleted.
- `fmix64` demoted to `private` with a Javadoc note scoping it to `murmur3ToLong` only.
- `hashMem64`'s polynomial body (`h = h * M2 + block` per 8 bytes) is unchanged - only the finalizer swaps.

### Call-site swaps (6 files)

- `DirectIntIntHashMap`, `DirectIntLongHashMap`, `DirectIntMultiLongHashMap`: `fastHashInt64` → `hashInt64`.
- `DirectLongLongHashMap`: `fastHashLong64` → `hashLong64`.
- `GroupByIntHashSet`: `fastHashInt64` → `hashInt64`.
- `GroupByLongHashSet`: `fastHashLong64` → `hashLong64`.

### Load factor bumps (`0.5` → `0.7`)

- `FastGroupByAllocator.chunks` (`DirectLongLongHashMap`).
- `SortKeyMaterializingRecordCursor.rowIdToOrdinal` (`DirectLongLongHashMap`).
- `EqSymFunctionFactory.Func.lookupCache` (`DirectIntIntHashMap`).
- `AsyncWindowJoinFastAtom.SLAVE_MAP_LOAD_FACTOR` (propagates to three construction sites: `AsyncWindowJoinFastAtom`, `WindowJoinPrevailingCache`, `WindowJoinFastRecordCursorFactory`).
- `DirectLongHashSet.DEFAULT_LOAD_FACTOR` (0.4 -> 0.7, affects `InLongFunctionFactory`). Unrelated to the FxHasher removal - this set already used `Hash.hashLong64` - brought in line with the rest of the direct-map family. The bump also exposed a pre-existing arithmetic bug in this class's `rehash()`, fixed below.
- `cairo.sql.count.distinct.load.factor` default, in both `PropServerConfiguration` and `DefaultCairoConfiguration`. This touches every `count_distinct` / `string_distinct_agg` hash set - `GroupByIntHashSet`, `GroupByLongHashSet`, `GroupByLong128HashSet`, `GroupByLong256HashSet`, `Utf8SequenceHashSet`. Note that the `Long128` / `Long256` / Varchar / String sets never used FxHasher, so the load-factor bump there is independent of the mixer change; the `0.5` default was a blanket conservative value that `0.7` supersedes regardless of which mixer the set uses.

### Incidental fix: `DirectLongHashSet.rehash()` arithmetic

`DirectLongHashSet.rehash()` computed `free = (int)((capacity - size) * loadFactor)`, which expands to `capacity * loadFactor - size * loadFactor` and leaves the set operating roughly `size * (1 - loadFactor)` slots past its configured load factor. Pre-existing bug (introduced with the class in commit `9968b7ce23`), amplified by the `0.4 -> 0.7` bump above: the effective plateau would have climbed from ~0.64 to ~0.80. Replaced with `free = (int)(capacity * loadFactor) - size`, matching the invariant every sibling `Direct*HashMap` class maintains.

Audited the rest of the hash-map family (`Unordered*Map`, `OrderedMap`, `GroupBy*HashSet`, all `AbstractLongHashSet` / `AbstractIntHashSet` descendants, `CharSequence*` / `Utf8Sequence*` subclasses, `CompactUtf8SequenceHashSet`, `CompactIntHashSet`) - `DirectLongHashSet` was the sole outlier.

### Mixer choice

| Candidate | Muls | Avalanche | Outcome |
|---|---:|---|---|
| `fmix64` (baseline) | 2 | ideal | serial dep chain of 2 muls |
| `moremur64` | 2 | ideal | same shape as `fmix64`, no speed win |
| `rrmxmx64` | 2 | ideal | no speed win |
| `xxh3Avalanche64` | 1 | ideal | chosen |
| `wyMix64` (`Math.multiplyHigh`) | ~1 | near-ideal | 2.6x throughput loss on HotSpot JDK 17 - `multiplyHigh` does not fuse with the lo-half multiply |
| `wyMixUnsigned64` | ~1 | near-ideal | 5x throughput loss |
| `fxHash64` | 1 | broken | `avgProbes > 250_000` on `i << 20` keys; disqualified |

### `hashMem64` body

The polynomial loop was kept as-is. Rejected alternatives:

- **`poly4Xxh3`** - four interleaved accumulators, 1.5-1.8x JMH speedup for keys >= 64 bytes. The win disappears into memory-bandwidth noise on end-to-end queries; the added complexity (four lanes + explicit fold) did not pay off.
- **`wyhashMem`** - pathological tail handling; 12-byte keys were 3x slower than baseline. Also hits the `Math.multiplyHigh` throughput ceiling.
- **`xxh64`** - strictly slower than the current polynomial for every key length below 256 bytes.

### Single-bit avalanche (100K random inputs, ideal = 32)

| Mixer | Avalanche |
|---|---:|
| `fmix64` | 31.999 |
| `xxh3Avalanche64` | 32.760 |
| `fastHash` (FxHasher, for reference) | 17.653 |

### 256-bucket shard distribution (chi2, uniform ~= 255)

`xxh3Avalanche64` yields chi2 near zero on structured inputs (sequential, timestamps, shifted-low-bit) because a single good multiply acts as a near-perfect permutation modulo 2^8 - better-than-random, not worse. Max
bucket sizes within ~5% of expected for both mixers across every workload.

### Probe quality at shipped configurations

The PR affects three groups of callers with different before/after setups. All three are measured below on the same workloads (1M keys, 2^20-slot open-addressed linear-probe table, load factor as configured).

**Group A — former `fastHash*` callers (6 edited sites).** Before: FxHasher at load 0.5. After: xxh3 at load 0.7. Affects `DirectIntIntHashMap`, `DirectIntLongHashMap`, `DirectIntMultiLongHashMap`, `DirectLongLongHashMap`, `GroupByIntHashSet`, `GroupByLongHashSet`.

| Workload | FxHasher @ 0.5 avg / max | xxh3 @ 0.7 avg / max |
|---|---:|---:|
| random-long | 1.50 / 45 | 2.17 / 116 |
| int32 zero-extended | 1.50 / 35 | 2.16 / 121 |
| sequential | 1.00 / 1 | 1.92 / 85 |
| stride=997 | 1.00 / 1 | 1.93 / 152 |
| stride=1000 | 2.50 / 4 | 2.20 / 86 |
| timestamp 1ms | 2.50 / 4 | 2.12 / 123 |
| `i << 20` | **262144 / 524288** | 2.88 / 63 |
| `i << 40` | **262144 / 524288** | 2.24 / 168 |

Orders-of-magnitude gain on the two adversarial workloads where FxHasher degrades into a linear scan. Modest probe-count tax on benign inputs (sequential and stride=997 go from effectively perfect to ~1.9 average). Net trade: robustness, ~29% less memory, and a faster finalizer per probe, for slower average probes on keys that are already cheap.

**Group B — `count_distinct` Long128/Long256/Varchar/String sets.** Before: fmix64 at load 0.5. After: xxh3 at load 0.7. Affects `GroupByLong128HashSet`, `GroupByLong256HashSet`, `Utf8SequenceHashSet` via the `cairo.sql.count.distinct.load.factor` default.

| Workload | fmix64 @ 0.5 avg / max | xxh3 @ 0.7 avg / max |
|---|---:|---:|
| random-long | 1.50 / 60 | 2.17 / 116 |
| timestamp 1ms | 1.50 / 48 | 2.12 / 123 |

Probe count rises by ~45% average going from 0.5 to 0.7, which is the expected density cost. The practical impact on count_distinct is small — the HyperLogLog / sparse-set layers above these hash sets spend far more time on their own work than on probe chains.

**Group C — `Unordered4Map` / `Unordered8Map` callers.** Before: fmix64 at load 0.7. After: xxh3 at load 0.7. No load factor change, finalizer swap only. This is the biggest group by call volume (every GROUP BY / fixed-key path).

| Workload | fmix64 @ 0.7 avg / max | xxh3 @ 0.7 avg / max |
|---|---:|---:|
| random-long | 2.16 / 115 | 2.17 / 116 |
| int32 zero-extended | 2.17 / 114 | 2.16 / 121 |
| sequential | 2.17 / 101 | 1.92 / 85 |
| stride=1000 | 2.18 / 114 | 2.20 / 86 |
| timestamp 1ms | 2.17 / 104 | 2.12 / 123 |
| `i << 20` | 2.16 / 126 | 2.88 / 63 |
| `i << 40` | 2.15 / 122 | 2.24 / 168 |

Tied on average probes across realistic workloads; xxh3 worst-case tails run slightly longer on two adversarial bit-patterns (`i << 20`, `i << 40`). Real win for this group comes from the finalizer latency, not probe quality.

### JMH microbenchmark (JDK 17.0.18, Linux x86_64)

| Mixer | Latency (ns/op) | Throughput (ns/op) |
|---|---:|---:|
| `fmix64` | 2.300 | 0.209 |
| `xxh3Avalanche64` | 1.335 | 0.193 |

`xxh3Avalanche64` is 1.72x faster in latency mode and roughly tied in throughput mode.

### `hashMem64` variable-length finalizer swap

The polynomial body is unchanged; only the finalizer differs. The finalizer-only swap saves roughly 1 ns per call - around 7% for 12-byte keys, falling to ~0.5% for 256-byte keys where the polynomial dominates.

### qdb-bench, 100M rows, single-aggregate fixed-size GROUP BY

Seven queries on the same IPv4 (Unordered4Map) key, one aggregate each. Medians across 5 iterations, single warmup discarded.

| Function | `fmix64` (ms) | `xxh3` (ms) | Delta |
|---|---:|---:|---:|
| `sum(v_long)` | 43.53 | 35.29 | -18.9% |
| `min(v_long)` | 40.69 | 33.02 | -18.9% |
| `max(v_long)` | 39.09 | 30.78 | -21.3% |
| `avg(v_double)` | 44.16 | 34.63 | -21.6% |
| `bit_and(v_int)` | 38.69 | 30.34 | -21.6% |
| `bit_or(v_int)` | 35.54 | 30.14 | -15.2% |
| `bit_xor(v_int)` | 38.02 | 31.67 | -16.7% |

Single-column `count()` across key types:

| Query | `fmix64` (ms) | `xxh3` (ms) | Delta |
|---|---:|---:|---:|
| `k_ip50, count()` (Unordered4Map, 50 keys) | 26.63 | 19.80 | -25.7% |
| `k_ip5k, count()` (Unordered4Map, 5K keys) | 35.15 | 31.03 | -11.7% |
| `k_long50, count()` (Unordered8Map, 50 keys) | 19.30 | 19.31 | 0% |
| `k_long5k, count()` (Unordered8Map, 5K keys) | 34.13 | 28.69 | -15.9% |
| `k_tstp50, count()` (Unordered8Map, 50 keys) | 21.28 | 19.01 | -10.7% |
| `k_tstp5k, count()` (Unordered8Map, 5K keys) | 35.92 | 37.51 | +4.4% |

Multi-aggregate queries (6 aggregates sharing one probe phase) sit within run-to-run noise - the per-probe hash cost is a small share of total time, and measured deltas on `k_long5k`, `k_vc5k`, `k_str5k`, and composite keys
range from +6.4% to +10.8% at 5 samples per query. Three of the apparent regressions route through `hashMem64`, whose finalizer swap was not included in that particular sub-comparison, so they are noise rather than real regression. These bench numbers set a floor on what this workload can resolve.

### Long-key `hashMem64` (10M rows, 5000 distinct keys)

| Key | baseline (ms) | patch (ms) | Delta |
|---|---:|---:|---:|
| `vc32`  | 23.32 | 25.28 | +8.4% (outlier skews median; mean is -12.4%) |
| `vc64`  | 30.55 | 29.24 | -4.3% |
| `vc128` | 43.19 | 42.09 | -2.5% |
| `vc256` | 73.18 | 70.12 | -4.2% |
| `str64`  | 49.05 | 50.48 | +2.9% |
| `str128` | 77.95 | 76.43 | -1.9% |
| `str256` | 149.50 | 149.49 | 0% |

Deltas are small because these queries are memory-bandwidth bound once keys exceed a few tens of bytes. A 10M-row scan over 128-byte keys streams ~1.3 GB of column data; at ~20 GB/s that's ~65 ms of pure DRAM reads, which matches the observed timings. Finalizer cycles get absorbed into the bandwidth stall.

### ClickBench (99,997,497 rows, 43 queries, 10 runs + 1 warmup)

Aggregate time summed across all 43 queries, using ClickBench's "cold + 10ms" convention (`min(execute) + 10ms` per query):

| Metric | master | patch | Delta |
|---|---|---|---|
| Total time | 14.68 s | 14.50 s | -1.21% |

Small net improvement. Most queries sit within run-to-run noise. Queries moving more than 5% either direction:

| Q | master | patch | Delta | Query |
|---|---|---|---|---|
| Q8 | 636.93 ms | 544.55 ms | -14.3% | `RegionID count_distinct GROUP BY` |
| Q17 | 781.63 ms | 735.83 ms | -5.8% | `UserID, SearchPhrase GROUP BY LIMIT` |
| Q27 | 127.26 ms | 96.50 ms | -22.4% | `CounterID AVG(length) GROUP BY HAVING` |
| Q42 | 3.82 ms | 3.12 ms | -5.1% | `SAMPLE BY 1m` |

## Tradeoffs and risks

- **Probe-count tradeoff for former `fastHash*` callers.** These sites lose FxHasher's very-fast probe chains on benign inputs (sequential-key avg goes from 1.00 to 1.92 probes per lookup; random-key avg from 1.50 to 2.17) in exchange for eliminating FxHasher's catastrophic failure mode on low-entropy bit patterns (avg 262k probes -> avg 2.9). The finalizer latency win per probe more than compensates in the qdb-bench and ClickBench numbers below.
- **Worst-case probe tails on two adversarial workloads** (`i << 20`, `i << 40`) run slightly longer under xxh3 than under fmix64 at 0.7 load. Average probe counts are unchanged for the fixed-size `Unordered*Map` callers on any realistic workload.
- **Load factor 0.7 reduces free-slot margin** from 50% to 30%. For the six affected hash tables this is still well below the upper bound at which linear probing degrades, and matches the project's standard elsewhere.
- **Memory footprint** for the six former `fastHash*` callers drops by roughly `1 - 0.5/0.7 = 29%` of previously-allocated map capacity at steady state.

## What was left unchanged

- **`murmur3ToLong`** - public contract is "Murmur3 128-bit", consumed by `ApproxCountDistinctIPv4/Int/Long` (HyperLogLog) and asserted bit-for-bit in `HashTest`. Internal `fmix64` calls remain; `fmix64` stays in `Hash.java` as a private helper scoped to this caller.
- **`hashMem64` polynomial body** - `h = h * M2 + block` stays serial. Multi-accumulator variants were prototyped and rejected (see Alternatives).
- **`hashLong128_64` / `hashLong256_64` structure** - only finalizers swap. Paired-product rewrites were considered but the expected savings (2-4 cycles) are below measurement noise for the `count_distinct` UUID / Long256 callers.

## Tests

Shard-distribution and random-distribution regression fences added in `HashTest` for the primitive paths that now route through `xxh3Avalanche64`:

- `testHashInt64RandomDistribution`, `testHashLong64RandomDistribution`, `testHashLong128_64RandomDistribution`, `testHashLong256_64RandomDistribution` - 100k-input unique-hash tests on the 32 LSBs, mirroring the existing `testHashMemRandomCorpus_hashMem64` pattern.
- `testHashLong64SequentialShardDistribution`, `testHashLong64ShiftedLow20ShardDistribution`, `testHashLong64ShiftedLow40ShardDistribution` - 256-bucket shard-balance checks on the inputs that broke FxHasher. The `i << 20` variant collapses all 100k entries into bucket 0 under FxHasher and serves as the real fence against any future mixer regressing on low-zero-bit inputs.
- `testHashMem64SequentialIntKeyShardDistribution`, `testHashMem64SequentialLongIntKeyShardDistribution` - shard balance for the 4-byte tail-only and 12-byte long+int-tail paths of `hashMem64` (the code paths `hashLong64` doesn't exercise).

`DirectLongHashSetTest.testRehashMaintainsLoadFactor` added as a regression fence for the `rehash()` arithmetic fix - constructs a set at `loadFactor=0.5`, fills to the first-rehash threshold, then asserts the second rehash fires at exactly the load-factor budget on the doubled capacity. Would fail under the old formula (which would have let the set run to ~20 elements before re-rehashing instead of exactly 16).


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 25 / 26 (96.15%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/DirectLongLongHashMap.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/std/DirectIntMultiLongHashMap.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/std/DirectIntIntHashMap.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/GroupByLongHashSet.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/Hash.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/std/DirectIntLongHashMap.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/eq/EqSymFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/GroupByIntHashSet.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyMaterializingRecordCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/FastGroupByAllocator.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/DirectLongHashSet.java|1|1|100.00%|

bluestreak01 (approved):

PR #6998, opened 2026-04-21
test(wal): fix WalLockerFuzzTest crash on timeout

RaphDal:
## Summary

`WalLockerFuzzTest.testFuzz` runs 8 worker threads performing 200_000 operations each against a native `WalLock`. When the overall deadline of 30 seconds elapses, the current code asserts the timeout and falls into `@After tearDown()`, which calls `locker.close()` and frees the Rust `Arc<WalLock>` through `Box::from_raw`. The worker threads are never signaled to stop, so any worker still inside a JNI call (e.g. `setWalSegmentMinId0`) continues to operate on the freed pointer. Its next `DashMap` shard-lock unlock reads a recycled atomic state word and fails the `WRITE_LOCKED` sanity check in `dashmap::RawRwLock::unlock_exclusive_slow`. Because that assertion is marked no-unwind, the Rust runtime aborts the JVM with SIGABRT, and Surefire reports:

```
The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
```

This masks the real timeout and also terminates every subsequent test scheduled on the same fork (in the original failure, `TtlTest.testMonthOneMicrosBeyondTtl[WITH_WAL]` was in-flight when the panic landed).

The fix introduces an `AtomicBoolean isStopRequested` that each worker checks in its loop guard, tracks the spawned workers in an array, and, on timeout, sets the flag and joins all workers before asserting. By the time `tearDown` runs, no thread is still holding a JNI handle to the native lock.

## Changes

- `core/src/test/java/io/questdb/test/cairo/wal/WalLockerFuzzTest.java`
  - Add `AtomicBoolean isStopRequested` and include it in the worker loop guard.
  - Retain references to the spawned threads in a `Thread[] workers` array.
  - On timeout, set the stop flag and join every worker before `Assert.assertTrue(...)` fails the test.

No production code is touched.

### DIFF
--- a/core/src/test/java/io/questdb/test/cairo/wal/WalLockerFuzzTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/wal/WalLockerFuzzTest.java
@@ -38,6 +38,7 @@ import org.junit.Test;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
 public class WalLockerFuzzTest {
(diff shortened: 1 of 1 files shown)


questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=228171)
**Tested Commit:** `4cf2a443e960b82cb249b711137da379252af886`

bluestreak01 (approved):

PR #7005, opened 2026-04-22
chore(parquet): populate dictionary page offset in column chunk metadata

RaphDal:
## Summary

When QuestDB writes a Parquet column chunk that contains a dictionary page, `build_column_chunk` was leaving `dictionary_page_offset` as `None` and setting `data_page_offset` to the offset of the first page in the chunk — which, for dict-encoded columns, is the dictionary page itself rather than the first data page. Per the Parquet thrift spec, `dictionary_page_offset` must point at the dictionary page when one exists, and `data_page_offset` must point at the first data page.

This PR replaces the single-offset assignment with a helper, `extract_page_offsets`, that walks the page write specs and reports the two offsets separately. The helper also rejects column chunks that contain more than one dictionary page, which is not valid per the spec.

`ColumnChunk.file_offset` was previously computed as `data_page_offset + total_compressed_size`. Once `data_page_offset` correctly skipped past the dictionary, that formula started overshooting the end of the column chunk's pages by `dict_bytes_written`. The thrift spec recommends new writers point `file_offset` at the inline `ColumnMetaData`, so this PR threads the column-metadata offset through `build_column_chunk` and writes it into `file_offset` directly. The value lands on the byte right after the last page (and bloom filter, when present), which is exactly where the inline `ColumnMetaData` is serialized by `write_column_chunk`.

Compression handling (`extract_compression`) was pulled into its own helper at the same time; behavior is unchanged — it still rejects mixed codecs within a column chunk and defaults to `Uncompressed` for empty spec lists.

The same dict-vs-data offset distinction surfaced a second, dormant bug on the update path. `build_raw_row_group` in `core/rust/qdbr/src/parquet_write/update.rs` was reading only `data_page_offset` to set `RowGroup.file_offset`. With the new offsets, that overshoots the row group's start byte by `dict_bytes_written` whenever the first column is dict-encoded. The fix swaps the open-coded lookup for parquet2's existing `ColumnOffsetsMetadata::calc_row_group_file_offset` helper, so the update path now uses the same dict-or-data fallback as `write_row_group` and stays in lockstep with future changes to that helper.

### Effects

- Parquet files written by QuestDB now declare `dictionary_page_offset` correctly, which readers can use for random-access reads directly to the dictionary page and for validation checks.
- `data_page_offset` now points past the dictionary page for dict-encoded columns, matching the spec. Files written before this change still parse because readers generally read pages sequentially from `data_page_offset`. External tools that previously seeked using the (incorrect) `data_page_offset` will now seek to a different — and correct — byte; tools that relied on the old value to land on the dictionary page will need updating.
- `ColumnChunk.file_offset` now points at the inline `ColumnMetaData` per the thrift spec recommendation, rather than at `data_page_offset + total_compressed_size` (which overshot end-of-pages by `dict_bytes_written` for dict-encoded columns). Internal QuestDB code reads `dictionary_page_offset` / `data_page_offset` and is unaffected. External tools that used the deprecated `file_offset` for sanity checks or footer validation will observe a different value.
- `RowGroup.file_offset` written by the update path now points at the dictionary page when the first column is dict-encoded, matching the row group's actual starting byte. Before, it pointed at the first data page, which overshot the start by the dictionary page's size; the bug was masked while `data_page_offset` itself was wrong.
- No change to the on-disk page layout or to file sizes — only the values written into the column chunk metadata change.
- The write path does one extra pass over the spec list per column chunk. The list is small (one entry per page) and the added work is negligible compared to page encoding and compression.

## Test plan

- New unit tests in `core/rust/parquet2/src/write/column_chunk.rs` cover `extract_page_offsets`, `extract_compression`, and `build_column_chunk`:
  - dictionary page followed by one data page — both offsets reported
  - data pages only — `dictionary_page_offset` is `None`
  - dictionary page followed by several data pages — `data_page_offset` points at the first data page
  - `DATA_PAGE_V2` is recognised as a data page
  - empty spec list returns `(0, None)` without panicking
  - two dictionary pages in one chunk returns an error
  - uniform compression across pages is reported as-is
  - empty spec list defaults to `Uncompressed`
  - mixed compression codecs return an error
  - `build_column_chunk` with dict + data page: `file_offset` equals the column-metadata offset and is not the previously-buggy `data_page_offset + total_compressed_size` (regression guard for the overshoot)
  - `build_column_chunk` with no dict page: `dictionary_page_offset` is `None`, `data_page_offset` matches the data page, and `file_offset` lands on end-of-pages
  - `build_column_chunk` with a bloom filter: `file_offset` sits past end-of-pages by the bloom filter size, matching where the inline `ColumnMetaData` is serialized
- New unit tests in `core/rust/qdbr/src/parquet_write/update.rs` cover `build_raw_row_group`:
  - first column has both dict and data pages: `RowGroup.file_offset` matches the dict offset, not the data offset (regression guard)
  - first column has only data pages: `RowGroup.file_offset` matches the data offset
  - first column has `dictionary_page_offset = Some(0)`: the sentinel is ignored and `RowGroup.file_offset` falls back to the data offset, mirroring `calc_row_group_file_offset`'s `> 0` filter
- Existing Parquet write tests continue to pass.

## Bug Fixes
- Fixed bloom filter position tracking and column chunk offset calculations in Parquet file writes
- Corrected row group footer layout handling for dictionary-encoded columns to prevent offset miscalculation

## Tests
- Added regression tests validating correct Parquet column offsets with and without dictionary pages
- Enhanced file descriptor identity validation in writer tests for more reliable assertions


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 28 / 37 (75.68%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/parquet2/src/write/column_chunk.rs|28|37|75.68%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=231452)
**Tested Commit:** `e1d245764e34b1869bbe509dd3657a68ebf6d9e1`

bluestreak01:
@RaphDal — review notes from a level-2 pass:

## Critical

### `ColumnChunk.file_offset` overshoots end-of-pages by `dict_bytes_written` whenever a dictionary page is present (in-diff)

`core/rust/parquet2/src/write/column_chunk.rs:283`

The line was unchanged textually, but its meaning silently changed because `data_page_offset` no longer points at the first page in the chunk.

- `bytes_written` per page = `header_size + compressed_page_size` (`page.rs:72-88`).
- `total_compressed_size` is summed over **all** pages, including dict (`column_chunk.rs:199-202`), so it equals `sum(bytes_written)`.
- Before PR: `data_page_offset = first_page.offset` (dict if present), so `file_offset = first_page.offset + sum(bytes_written) = end_of_last_page`.
- After PR: `data_page_offset = first_DATA_page.offset = first_page.offset + dict_bytes_written` when dict exists, so `file_offset = end_of_last_page + dict_bytes_written` — i.e., it points past where the chunk's pages actually end.

Internal QuestDB code is unaffected (`ColumnOffsetsMetadata::calc_row_group_file_offset` reads `dictionary_page_offset`/`data_page_offset`, not `file_offset`; no Java JNI consumer reads it). External Parquet readers treat `file_offset` as deprecated per the Thrift spec but it is still emitted on disk; tools that use it for sanity checks or footer validation will now see a value past the actual end of pages for any dict-encoded column.

Fix — compute `file_offset` from the first page rather than from `data_page_offset`:

```rust
file_offset: specs.first().map(|s| s.offset as i64).unwrap_or(0) + total_compressed_size,
```

or the more defensible form:

```rust
file_offset: specs.last().map(|s| (s.offset + s.bytes_written) as i64).unwrap_or(0),
```

## Moderate

### No round-trip test asserts the metadata fix actually produces correct offsets

`core/rust/parquet2/src/write/column_chunk.rs:338-534`

The PR fixes a metadata-correctness bug yet ships no test that writes a parquet file containing a dict-encoded column, reads back the `FileMetaData`, and asserts `dictionary_page_offset == Some(<correct value>)` and `data_page_offset > dictionary_page_offset`.

The Java tests in `ExpParquetExportTest` and `CopyExportTest` only assert byte counts — they would pass for any change of equal byte size and cannot catch a regression that re-introduces `dictionary_page_offset = None`. A `read_metadata`-based assertion would close this gap and would also catch the Critical above.

### Async `write_column_chunk_async` path (`column_chunk.rs:147`) is not exercised by any test

It calls the same `build_column_chunk`, so it inherits the fix and the bug, but no test covers it. Pre-existing gap, but enlarged by this PR's behavior change.

### PR title and description (CLAUDE.md conventions)

- Title is `chore(parquet): populate ...`. The body opens with "`build_column_chunk` was leaving `dictionary_page_offset` as `None`... Per the Parquet Thrift spec, `dictionary_page_offset` must point at the dictionary page" — this is a spec-compliance fix, not a chore. Per CLAUDE.md the title should be `fix(parquet): fix ...` and must repeat the verb (e.g. `fix(parquet): fix dictionary_page_offset emission in column chunk metadata`).
- The Test plan section says new tests live in `core/rust/qdbr/parquet2/src/write/column_chunk.rs`. Actual path is `core/rust/parquet2/src/write/column_chunk.rs`.
- No labels are set. Should have at least `rust`; arguably `Bug` and/or `Compatibility` since external readers may see the changed `file_offset`/`data_page_offset` fields.
- The "Effects" section says "Files written before this change still parse" but does not characterize the converse: external tools that previously seeked using the (incorrect) `data_page_offset` will now seek to a different — correct — byte. Worth one sentence.

## Minor

### Inconsistent panic policy for unknown `PageType` within the same function

`build_column_chunk` at lines 212/227/240 uses `try_into().unwrap()` and panics on any unknown variant. The new `extract_page_offsets` at lines 298-318 uses a permissive `_ =>` arm that silently ignores them. The `_ =>` is unreachable in practice because `CompressedPage` only has `Data` and `Dict` variants, so the writer never produces other types. Two policies for the same input in the same function is a code smell.

### `extract_page_offsets` relies on writer-side ordering without a spec citation

`column_chunk.rs:303` breaks on the first data page assuming dict precedes data. Comment says so but does not cite the Thrift spec ("Column chunk contains optionally a dictionary page followed by data pages").

### Local var naming inconsistency

`column_chunk.rs:297, 307`: `dict_page_offset` (local) vs `dictionary_page_offset` (Thrift field, line 274).

### Test helper duplication

`dict_page_spec` / `data_page_spec` / `data_page_v2_spec` (`column_chunk.rs:342-425`) share ~80% of field construction; a shared `default_spec()` + per-test mutator would shrink the module by ~50 lines.

## Downgraded (considered and dismissed)

- **"`extract_page_offsets` returns `data_page_offset = 0` when no data page exists"** — Behavior identical to pre-PR (`unwrap_or(0)` produced 0 for empty specs). Not a regression. Realistically unreachable since column chunks always carry data pages.
- **"data-page-before-dict-page malformed input is silently dropped"** — Unreachable: `CompressedPage` has only `Data` and `Dict` variants and the writer always emits canonical order. Not a real bug.
- **"`spec.offset as i64` overflow for offsets > i64::MAX"** — Requires an 8 EiB Parquet file.
- **"`extract_compression` performance regression"** — Net improvement: removes a `HashSet<Compression>` allocation and short-circuits on first mismatch.

## Summary

Request changes for the one critical `file_offset` regression. The rest is moderate/minor. All findings are in-diff; no out-of-diff regressions surfaced.

RaphDal:
@bluestreak01 - Thanks for the level-2 pass — addressed everything below. New commit covers the code changes; PR description and labels updated separately.

## Critical

### `file_offset` overshoot

You're right that the formula was broken once `data_page_offset` skipped past the dictionary. I took a slightly different fix from the two you suggested.

Both of your candidates land `file_offset` on end-of-last-page. The thrift spec on `ColumnChunk.file_offset` says new writers should point it at the `ColumnMetaData` instead:

> New implementations should set this to the location of the ColumnMetaData and use the column_meta_data field to indicate the location of column data.

`write_column_chunk` does serialize the inline `ColumnMetaData` immediately after the pages and bloom filter (line 95-102), so the offset that satisfies the spec recommendation is `end_of_pages + bloom_filter_bytes`. I threaded that value through `build_column_chunk` as `column_metadata_offset` and assigned it directly to `file_offset`. With no bloom filter present this collapses to your suggestion (end-of-pages); with a bloom filter it's strictly past end-of-pages, by design.

Net effect for external tools is the same as what you described — they'll see a different value than before — but at least it's a value the spec acknowledges, rather than a third interpretation.

## Moderate

### No round-trip test asserts the metadata fix

Added three unit tests on `build_column_chunk` directly in `core/rust/parquet2/src/write/column_chunk.rs`. They construct realistic page specs and assert on the resulting `ColumnChunk`:

- `test_build_column_chunk_offsets_with_dict_page` — the actual regression guard. Asserts `file_offset == column_metadata_offset`, `dictionary_page_offset == Some(<dict offset>)`, `data_page_offset` past the dict, and explicitly `assert_ne!` against the buggy `data_page_offset + total_compressed_size` formula. Under the old code this test fails with `file_offset=440` vs expected `350`.
- `test_build_column_chunk_offsets_no_dict_page` — sanity case, locks `dictionary_page_offset == None`.
- `test_build_column_chunk_file_offset_includes_bloom_filter` — verifies the bloom-filter advancement gets recorded in `file_offset`. Buggy formula gives `340`, fix gives `314`.

I went with `build_column_chunk`-level tests rather than full file write/read round-trip because constructing real `CompressedPage`s through `DynStreamingIterator` is heavyweight and the bug lives in `build_column_chunk`. Happy to add a full round-trip in a follow-up if you'd prefer the integration coverage too.

### Async path

Inherits the fix verbatim — `write_column_chunk_async` calls the same `build_column_chunk` with the same `offset` arithmetic. Agree it would benefit from coverage, but the gap is pre-existing and out of scope for this PR.

### PR title and description

- Test plan path corrected (`core/rust/qdbr/parquet2/...` -> `core/rust/parquet2/...`).
- Effects section now calls out that external tools previously seeking via the old `data_page_offset` will land on a different byte, and that `file_offset` semantics changed.
- Labels: added `rust`, `Compatibility`.
- Title: kept `chore(parquet)`. I use `fix` only for customer-reported bugs and `chore` for self/colleague/CI finds — this came out of your review, so it stays `chore`. Adjusted the body wording so it doesn't lean on "fix" language and conflict with the title.

## Minor

### Inconsistent panic policy in `extract_page_offsets`

Switched the match to convert via `try_into()?` and match exhaustively on `crate::page::PageType`. The wildcard `_ =>` arm is gone — same enum, same exhaustiveness as `build_column_chunk`. Left `build_column_chunk`'s `unwrap()`s alone to keep the diff scoped.

### Spec citation

Docstring now cites `parquet.thrift`, ColumnMetaData ordering rule.

### Naming

`dict_page_offset` -> `dictionary_page_offset` to match the thrift field.

### Test helper duplication

Extracted `default_spec(offset, compression)` returning a blank spec; `dict_page_spec` / `data_page_spec` / `data_page_v2_spec` now mutate only the fields they care about. Each helper is ~10-14 lines instead of ~25.

## Downgraded

Agree with all five — no action.

bluestreak01 (dismissed):
Reviewed at level 2. Critical `file_offset` overshoot is correctly fixed by threading `column_metadata_offset` through `build_column_chunk`; tests are good regression guards.

One out-of-diff follow-up worth a separate PR: `build_raw_row_group` in `core/rust/qdbr/src/parquet_write/update.rs:1061-1064` reads only `data_page_offset` to set `RowGroup.file_offset`, so for files written with the new (correct) dict offsets it will overshoot the row-group start by `dict_bytes_written` whenever the first column is dict-encoded. Pre-existing bug, dormant before this PR; not blocking.

Minor nits (also not blocking): the `assert_ne!(chunk.file_offset, buggy_file_offset)` regression guard is informational since the equality check on the line above already nails the value down; `extract_compression`'s `Uncompressed` default for empty specs is unreachable from production callers; and the `chore(parquet)` title is slightly at odds with the `Compatibility` label and the user-facing nature of the fix, but defensible.

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=237367)
**Tested Commit:** `6fe93e184669e38f2a6e8685836528b944691823`

bluestreak01 (dismissed):
Approving — re-reviewed at level 2, no blocking issues. The `build_raw_row_group` follow-up I flagged in the earlier (dismissed) pass is now folded into this PR and correctly uses `calc_row_group_file_offset`.

The core fix is correct and backward-compatible by construction. Every reader seeks via `byte_range()` = `dictionary_page_offset.unwrap_or(data_page_offset)`, and the `_pm` cache fingerprints the same way, so `byte_range_start` is invariant under the fix: old files (dict=None, data=dict-offset) and new files (dict=dict-offset, data=past-dict) resolve to the same physical byte. Only the declared metadata values change. `ColumnChunk.file_offset` and `RowGroup.file_offset` have no internal readers, so the changed values affect external Parquet tools only. `cargo clippy --all-targets` is clean on qdbr, and the new unit tests are good regression guards.

Non-blocking follow-ups, author's discretion:

Moderate
- No Rust-level write->read-*data* round-trip for a dict column. The new `mod.rs` test reads back metadata only. The data path is covered by Java integration tests (`CopyExportTest`, `O3ParquetPartitionFuzzTest`), so this is a cheap-guard gap, not a correctness hole. A dict-enabled variant of the `decode_column.rs` round-trip would close it.
- No integration test drives `build_raw_row_group` with a dict-encoded *first* column. All existing O3/update tests put LONG/timestamp first and SYMBOL second, and `RowGroup.file_offset` is first-column-derived. A SYMBOL-first table converted then UPDATE-d would exercise it end to end.
- The bundled `jni.rs` fd-identity test change is unrelated to dictionary-page offsets and is not mentioned in the description. Document it or split it into its own PR. Consider adding the `Bug` label since the body describes spec-compliance fixes.

Minor
- `column_chunk.rs:300`: the doc comment puts quotes around a paraphrase attributed to `ColumnMetaData`; the dictionary-page placement rule is actually on `DictionaryPageHeader`.
- The "Multiple dictionary pages" guard only catches consecutive leading dict pages (it breaks on the first data page), which is narrower than "rejects column chunks that contain more than one dictionary page". Unreachable in practice since the writer always emits a single leading dict page.
- Optional: `build_column_chunk` makes ~5 separable passes over `specs` that could fuse into one. Immaterial given `specs.len()` is 1-2 on the export path, and the change is already a net improvement over the prior HashSet-based code.

bluestreak01 (dismissed):
Approving - independent level-2 re-review, no blocking issues.

The core fix is correct and backward-compatible by construction. Every reader seeks via `byte_range() = dictionary_page_offset.unwrap_or(data_page_offset)`, which resolves to the same physical byte for old files (`dict=None, data=dict_offset`) and new files (`dict=Some, data=past-dict`). Both `ColumnChunk.file_offset` and `RowGroup.file_offset` are write-only - the `file_offset()` accessor is never called and `parquet_read` never references the row-group field - so the changed values affect external Parquet tools only. The `offset` accumulator lands exactly on the inline `ColumnMetaData` byte (per-page `bytes_written = header_size + compressed_page_size`, and the bloom block now advances `offset` past the filter), and the update path mirrors `write_row_group` via `calc_row_group_file_offset`.

Non-blocking follow-ups, author's discretion:

## Moderate

- The Effects section states "No change to the on-disk page layout or to file sizes," but the footer does grow: `dictionary_page_offset` is an optional thrift field flipping `None` -> `Some(i64)`, so each dict-encoded column chunk now serializes it. The PR's own Java test deltas confirm it (+8 bytes at default, scaling with row-group count up to +394 at `row_group_size=201`). The page layout is unchanged, but the sentence about file sizes should be corrected.
- The now-live `update.rs` dict-offset paths lack direct coverage: `adjust_column_chunk_offsets`'s dict-shift branch (update.rs:1173) and an end-to-end UPDATE on a symbol-column table. Both are correct by inspection and reads stay safe via `byte_range` invariance, but no test would catch a future regression in the dict shift. A unit test on `adjust_column_chunk_offsets` with `Some(dict)` is the cheapest guard; an end-to-end "write symbol parquet -> copy_row_group -> read back values" test would also close it. The new `mod.rs` test reads back metadata only, never decoding symbol data.

## Minor

- `extract_page_offsets` doc cites `parquet.thrift` ColumnMetaData for the dict-precedes-data / single-dict rule; that rule is actually documented on `DictionaryPageHeader` (parquet_format.rs:3821-3822).
- The "multiple dictionary pages" guard only catches consecutive leading dict pages (it breaks on the first data page), narrower than the body wording. Unreachable in practice since the writer emits one leading dict page.
- The jni.rs fd-identity change (fcntl -> fstat dev/ino identity) is unrelated to dictionary offsets; consider splitting it into its own PR. Consider adding the `Bug` label given the spec-compliance nature of the fix.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 357 / 364 (98.08%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/parquet_write/mod.rs|19|20|95.00%|
|:large_blue_circle:|parquet2/src/write/column_chunk.rs|283|288|98.26%|
|:large_blue_circle:|qdbr/src/parquet_write/update.rs|55|56|98.21%|

bluestreak01 (dismissed):
Approving after a full (level 3) review: change-surface map, fresh-context adversarial correctness + performance passes, and per-finding source verification. Ran the affected Rust tests — 13/13 parquet2 `column_chunk` and 3/3 qdbr `build_raw_row_group` pass; no new clippy warnings in changed code.

The fix is correct and spec-aligning: `dictionary_page_offset` is now populated, `data_page_offset` correctly skips the dictionary, `ColumnChunk.file_offset` points at the inline `ColumnMetaData`, and the update path's `RowGroup.file_offset` is brought into lockstep with the write path (the dormant bug the old incorrect `data_page_offset` was masking). QuestDB's own read path is unaffected because `byte_range()` already prefers `dictionary_page_offset`, and old files remain readable via the `data_page_offset` fallback. Verified no production code reads `ColumnChunk.file_offset`.

No blocking issues. Non-blocking follow-ups:
- Description says 'No change to ... file sizes', but the test updates bump export sizes upward — the footer now carries the previously-absent `dictionary_page_offset` per dict-encoded chunk (delta scales with row-group count). Please correct this line.
- Optional: `fix(parquet):` would fit better than `chore` (this fixes two bugs); consider a `Bug` label.
- Optional/pre-existing: the raw-copy path (`adjust_column_chunk_offsets`) doesn't shift `ColumnChunk.file_offset`, so copied chunks keep a source-coords value (harmless internally, matters only to external tools) — one-line follow-up.
- Optional: `extract_page_offsets` returns `data_page_offset = 0` for the (currently unreachable) dict-only/empty-chunk case.

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=242669)
**Tested Commit:** `9b0398d95197ddf6c15d38e499deeb806a90f062`

PR #7010, opened 2026-04-22
refactor(sql): require explicit compacted index in filtered record positioning

puzpuzpuz:
## Summary

This PR replaces the two-overload `setRowIndex` API on `PageFrameFilteredMemoryRecord`
with a single explicit `setFilteredRowIndex(rowIndex, compactedRowIndex)` method.
The inherited single-argument `setRowIndex(long)` now throws
`UnsupportedOperationException` on filtered records: callers must supply both the
absolute and compacted indices.

The previous API relied on an internal auto-incrementing counter
(`compactRowIndex`, initialized to `-2`) to track the compacted position during
sequential iteration. A call to `setRowIndex(long)` outside the intended
sequential walk would desync the compacted index from the absolute index, and
non-filter columns would then read from the wrong row without raising an error.

The PR updates callers in `AsyncGroupBy{,NotKeyed}`, `AsyncHorizonJoin{,NotKeyed}`,
and `AsyncMultiHorizonJoin{,NotKeyed}` to pass both indices. Sites that
previously used `record.setRowIndex(0); record.getRowId()` to compute `baseRowId`
now call `Rows.toRowID(frameIndex, 0)` directly, which is equivalent and does
not depend on positioning the record.

The PR also adds a parallel `setFilteredRowIndex(long, long)` to the base
`PageFrameMemoryRecord` (ignoring the compacted index) so the same call site
works polymorphically against filtered and unfiltered records.

## Tradeoffs

- Positive: the API rejects unsupported positioning instead of silently
  desyncing the compacted index. Call sites state the compacted index
  explicitly at every positioning point.
- Negative: every caller of a filtered record must now track the compacted
  index itself. This is a small amount of extra code at each call site
  compared to the implicit auto-increment.
- No runtime behavior change for correctly-written callers; no measurable
  performance impact expected (hot paths still do a single field store).

## Test plan

- Existing `AsyncHorizonJoin*` tests in `ParallelHorizonJoinFuzzTest` cover the
  out-of-order K-way merge positioning path explicitly via the
  `*LateMaterialization*` cases.
- Existing `AsyncGroupBy*` and `AsyncMultiHorizonJoin*` fuzz tests exercise the
  sequential GROUP BY positioning path and the keyed/non-keyed factory variants;
  late-materialization coverage is incidental, via filters that fall under the
  selectivity threshold.
- All call sites that previously used `setRowIndex(long)` or
  `setRowIndex(long, long)` on a filtered record have been migrated;
  `setRowIndex(long)` on a filtered record now throws and would fail any
  remaining caller loudly.


bluestreak01 (approved):
Reviewed the refactor end-to-end. Verified:

- All call sites of the old setRowIndex(long) and setRowIndex(long, long) on filtered records have been migrated to setFilteredRowIndex.
- AsyncGroupByRecordCursorFactory dispatches to the batched probe path only when not late-materialized, so Map.probeBatch* never receives a filtered record.
- baseRowId computation via Rows.toRowID(frameIndex, 0) is equivalent to the previous setRowIndex(0); getRowId() pattern.
- Slot model in AsyncFilterContext preserves single-threaded access per record.
- BoolList reuse in of() is safe across frames with differing column counts.
- Resource cleanup is unchanged.

LGTM.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 54 / 64 (84.38%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameFilteredMemoryRecord.java|31|41|75.61%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryRecord.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/HybridColumnMaterializer.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncMultiHorizonJoinRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncHorizonJoinRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncMultiHorizonJoinNotKeyedRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncGroupByRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncHorizonJoinNotKeyedRecordCursorFactory.java|3|3|100.00%|

PR #7024, opened 2026-04-24
feat(core): support column type conversions with parquet partitions

ideoma:
Tandem https://github.com/questdb/questdb-enterprise/pull/1089

## Summary

`ALTER TABLE ... ALTER COLUMN ... TYPE` now works on tables that contain parquet partitions. Before this change, the operation supported native partitions only — running it against a table with any parquet partition either failed or silently left the parquet data unconverted, so subsequent reads returned values in the old type or NULL after a later O3 merge.

End-user behaviour:

- `ALTER COLUMN TYPE` on a parquet-partitioned table now succeeds and returns the converted data on every read, matching the native partition behaviour exactly. Null sentinels, BYTE/SHORT/CHAR column tops, date/timestamp scaling and UTF-16/UTF-8 transcoding all match.
- Conversion is lazy: parquet partitions are not rewritten by the ALTER. Decoding cost moves to the query path, with a one-time decode per row group buffered for subsequent rows. O3 merges that land on a converted partition rewrite the partition with the new type.
- Conversions involving SYMBOL, and chained conversions where the parquet column type no longer matches current metadata, eagerly convert the affected parquet partitions to native before the ALTER. This keeps semantics consistent at the cost of one rewrite per affected partition.

## Implementation

- `ConvertOperatorImpl` orchestrates the conversion: pre-pass to convert parquet→native where lazy decode is impossible (target/source SYMBOL, or a chained conversion with type mismatch), then per-partition dispatch. Column tops are propagated from the old to the new column index on every parquet partition so a later parquet→native conversion finds data at the correct row offsets.
- Each `ALTER COLUMN TYPE` creates a new column index with a `replacingIndex` link to the previous index. `O3PartitionJob` and `PageFrameMemoryPool` walk this chain to map current writer indexes to columns inside an older parquet file.
- The query path (`PageFrameMemoryRecord`) carries a per-column conversion strategy (`-1` = none, `>= 0` = source-tag for fixed→var, `< -1` = negative-tag for var→fixed) and converts at accessor level using pooled `StringSink`/`Utf8StringSink` to stay zero-GC. Fixed→fixed conversions happen in the Rust decoder via `post_convert()` (boolean expansion, date/timestamp ×/÷1000).
- Parquet `Repetition` for BYTE/SHORT/CHAR is narrowed to Optional so column-top rows can carry NULL via def-level=0. Without this, lazy SHORT→INT would produce 0 in INT space where the native path produces `Integer.MIN_VALUE`, breaking the contract that both paths produce identical results. Encoder dispatch in `parquet_write/encode.rs` is routed accordingly; `core/src/main/java/io/questdb/griffin/CLAUDE.md` documents the rule and the schema/dispatch consistency requirement.
- When O3 lands on a parquet partition with type-converted columns, the partition is rewritten end-to-end and the lazy conversion is materialised permanently.


ideoma:
## Review: PR #7024 — column type conversions with parquet partitions

**Level 2** review (8 parallel agents: Java correctness, Rust value-correctness, Rust panic-safety, concurrency, resources, performance + adversarial-performance, tests, code-quality; cross-context folded into the correctness/concurrency passes). Scope: 91 files, +13,581/−793.

This is a large, unusually careful feature PR. After deep verification — two agents built the Rust crate and ran the parquet test suites, the concurrency agent proved the pool threading model, every correctness claim was traced against the native `ColumnTypeConverter` — **no Critical correctness, concurrency, resource-leak, or JVM-abort/panic bug was confirmed.** The findings below are a Moderate performance issue on the DDL path, real test-coverage gaps, and code-quality items. None are blocking on their own.

### Critical
None confirmed.

### Moderate

**M1 — DDL pre-pass re-mmaps + re-parses each parquet partition footer per partition, then decodes a second time on conversion** _(in-diff)_
`TableWriter.getParquetColumnType` (`TableWriter.java:2401`), called per parquet partition from `ConvertOperatorImpl.convertColumn0` (`ConvertOperatorImpl.java:252`).
Each call does `mapRO` (mmap the whole parquet file) + `parquetFileDecoder.of()` (parse footer) + `Misc.free(parquetFileDecoder)` in the `finally` — and the decoder is a *reusable* field, so it's freed and reallocated on every partition. Partitions that then need conversion get mmap'd + decoded a second time by `convertPartitionParquetToNative`. For `ALTER COLUMN TYPE` on a table with many cold parquet partitions (hundreds–thousands of daily partitions, the common layout), this is an avoidable 2× I/O/parse multiplier plus per-call native-alloc churn.
_Fix:_ don't `Misc.free` the reusable decoder inside the loop (`clear()`/reuse); share one footer decode between the inspection pass and `convertPartitionParquetToNative`. (Reading the footer directly rather than the `_pm` sidecar, for robustness when the sidecar is missing/stale, is fine; the redundant second decode and the free/realloc churn are the real waste.) DDL path, not per-query, so Moderate not Critical.

> **Update — M1 retracted.** This is essentially unavoidable; I overstated it. The inspection loop only runs for symbol / chained / dedup-key conversions (`ConvertOperatorImpl.java:246` — `if (hasPriorConversion || isTargetSymbol || isDedupKey)`), so a plain first-time lazy conversion never enters it and pays none of this cost — the "hundreds–thousands of cold partitions" framing was wrong. `mapRO` is an mmap (only footer pages fault in), not a whole-file read, and you must open the file to learn the stored column type anyway. The only genuinely redundant work is the second decode on partitions that *do* get converted — but those are about to be fully rewritten (read every row group + re-encode), which dwarfs a footer reparse by orders of magnitude. Sharing the decoder is non-trivial plumbing for a rounding-error saving, and not freeing the reusable decoder per iteration just trades peak native memory for a tiny CPU win on a DDL path. Not worth doing.

**M2 — Test-coverage gaps on the riskiest new code** _(in-diff)_
The assertion-rule audit came back **clean** (no `assertSql`/`TestUtils.assertSql`, no `.returnsOnce(...)`, no banned plan helpers, no lone-`assertQuery`-in-`assertMemoryLeak`), and the differential `native-table vs parquet-table` design is sound and correctly supplemented with absolute-oracle tests for the float/decimal boundary cases the native path shares. But concrete holes remain:
- **Rust decoder arms genuinely untested by anything** (CI shows `decode.rs` at 13.67% — most arms are reached only via JNI integration tests the coverage tool can't see, but these are not): the `Plain` and `DeltaByteArray` byte-array decode arms (QuestDB's writer can't emit them, so no writer-driven test reaches them — they fire only for externally produced parquet) and the `decode_double_dispatch` `Array` arm (no `DOUBLE[]` conversion test). Add native-Rust unit tests over hand-built pages for these.
- **`parquet_write` encoder paths** (`primitive.rs`/`numeric.rs` at 25%) have no targeted Rust unit test pinning encoded bytes.
- **Converted column as a JOIN equality key** is untested (all JOIN tests join on the unconverted column and only aggregate/filter on the converted one).
- **var→fixed and var→var eager rewrite** are only tested at single-row-group size; the fixed→var arm has a multi-row-group test (`testIntToStringConvertToNativeMultiRowGroup`, which exists precisely because it caught a running-offset bug) but the symmetric arms don't, so a per-row-group offset bug there wouldn't be caught.
- **DECIMAL→DECIMAL narrowing across mismatched scales** is explicitly skipped (author comment, `ParquetColumnTypeConversionTest.java:741`) — an acknowledged hole in the conversion matrix.

**M3 — Misattributed Javadoc; `private static` method interleaved among instance methods** _(in-diff)_
`PageFrameMemoryRecord.java:893-907`: the Javadoc describing `convertFixedToStr` ("Converts a fixed-type value to a STRING CharSequence… caches the `Fixed2VarConverter`…") sits directly above `private static boolean isNoNullSentinelFixedType(...)`, which was wedged in between the doc and the method it documents. Net effect: the doc is attached to the wrong method and `convertFixedToStr` is left undocumented, and a static helper sits in the instance-method block. _Fix:_ move `isNoNullSentinelFixedType` to the static group and reattach the Javadoc to `convertFixedToStr`.

### Minor

- **Per-row double work on converted columns** (`PageFrameMemoryRecord.getStrLen:1411`, `getVarcharSize`; same in `PageFrameFilteredMemoryRecord`): `getStrLen` runs the full `convertFixedToStr`/`readVarValueForConversion` just for `.length()`, and a following `getStrA` re-runs it; `getVarcharSize` formats into the sink and discards all but `.size()`. Length-then-value consumers (sorting, varchar key building) pay the formatter twice per row. Still zero-GC; a last-cell `(columnIndex,rowIndex)` cache would avoid it.
- **`init()` O(columnCount) snapshot copy of `sourceColumnTypes` per frame navigation** (gated behind `hasTypeCasts`, so zero cost for the common path). Could skip when bind generation / source-array identity is unchanged.
- **`PageFrameFilteredMemoryRecord` scalar getters delegate to `super` on the cast path without the `rowIndex = getRowIndex(...)` save/restore the string/varchar getters perform** (`:171-437` vs `:441-466`). Safe today (scalar getters are only reached on the var→fixed path, which re-derives the row through overridden `getStr0`/`getVarchar`), but fragile — a future fixed→var conversion routed through a scalar getter would silently read the wrong filtered row. Add a comment/assert.
- **`row_groups.rs:1436` and `:1626`** add `(row_group_hi - row_group_lo)` unsigned subtraction (debug panic / release wrap if `hi < lo`). Pre-existing caller contract, not newly exposed, but inconsistent with the sibling planning path in `parquet_meta_decode.rs:104` which uses `saturating_sub`. Use `saturating_sub` for consistency.
- **`CLAUDE.md` rendering bugs** (`core/src/main/java/io/questdb/griffin/CLAUDE.md`): line 8 opens a code fence as `` ```2 `` (stray `2` breaks the block) and line 23 reads `For each PARQUET partition:2` (stray trailing `2`) — accidental edits. Also: a package-level design doc named `CLAUDE.md` may be misread as directory-scoped agent guidance; consider a `package-info` or `docs/` location. Content itself is good and on-topic.
- **Member ordering** (alphabetical-within-kind): `PageFrameMemoryPool.resolveParquetColumn:1088`; six new statics appended at end of `O3PartitionJob.java` (`convertFixedColumnToString:4202`, …, `estimateStringDataSize:4390`, `estimateVarcharDataSize:4413`, `isNoNullSentinelFixedType:4196`); `PageFrameFilteredMemoryRecord.needsLazyConversion:623`.
- **`needsLazyConversion(...)`** should use the `is…`/`has…` boolean-method prefix.
- **Non-ASCII arrows (`→`) in code comments** (`TableWriter.java:289`, `TableUtils.java:89`) — inconsistent with the PR's own em-dash/arrow→ASCII cleanup elsewhere.
- **`isNoNullSentinelFixedType` is duplicated** in `O3PartitionJob` and `PageFrameMemoryRecord` — consider hoisting to a shared util.

### Downgraded (false positives dismissed after verification)
- `-encoded == -1` sentinel collision for var→fixed — **disproved**: STRING=11, VARCHAR=26 ⇒ `-encoded ≤ -11 < -1`; no-sentinel fixed types (BOOLEAN=1…CHAR=4) stored as positive. No collision.
- `columnTops` dangling after buffer free — **safe**: it's a strict peer of `pageAddresses` in `ParquetBuffers` (allocated/freed/cleared/reopened/bind-generation-guarded together).
- Shared `sourceColumnTypes`/`hasTypeCasts` data race across workers — **safe**: every `PageFrameMemoryPool` is single-thread-confined (per-reduce-task / per-worker-slot / per-cursor), and `init()`/`copyOf()` deep-copy the conversion state; cross-thread record sharing is mediated by `boundPool` + `bindGeneration`.
- `resolveParquetColumn` stale-slot fall-through and the non-symbol→symbol `CairoException.critical` — **safe**: `queryToSlot`/`sourceColumnTypes` pre-set to −1; the critical is a defensive assertion the pre-pass makes unreachable in normal operation.
- Rust panic/overflow sites — **none reachable**: `try_reserve`-backed buffers, `checked_*` on all decimal/timestamp scaling; the diff actually *converts* five encoder `assert!`s into recoverable errors (a net panic-safety improvement).
- `getDate`/`getTimestamp` → `getLong` regression — **safe**: master already resolved these to the identical `Record` defaults.

### Summary
**Verdict: approve with minor changes.** No blocking issues. Across the parallel passes, ~30 draft findings were verified and ~10 dropped as false positives (listed above); essentially all confirmed findings are in-diff — the cross-context pass actively walked `getPageAddress` callers, the `Async*` factory chain, and the pool threading model and cleared them.

Tradeoffs the PR makes honestly: conversion cost moves to the query path (lazy decode, one decode per row group), and SYMBOL/chained/dedup-crossing conversions force an eager per-partition rewrite — both stated in the description. The residual *risk* is concentrated in M2: the Rust decode/encode arms CI flags at 13–25% are covered (where covered) only via JNI integration tests, leaving the writer-unreachable byte-array arms, the DOUBLE Array arm, var→fixed/var→var multi-row-group rewrite, and converted-column-as-JOIN-key genuinely untested.

One item to confirm rather than dismiss: the **Enterprise CI failure** (build 243159) was run against commit `d491a67b`, which is **not** the current head — so it may predate later fixes, but that should be confirmed by re-running enterprise CI on the current head rather than assumed stale.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) — `/review-pr` level 2</sub>

ideoma:
## Review: PR #7024 — column type conversions with parquet partitions

**Level 2** review (7 parallel agents: Java correctness, Rust value-correctness, Rust panic-safety, concurrency+resources, tests, performance+adversarial, code-quality+metadata; cross-context folded into the correctness/concurrency passes). Scope: 91 files, +13,581/−793. Every finding verified against source.

The lazy-conversion contract — byte-identical to the native `ColumnTypeConverter`: NULL sentinels, BOOLEAN null_count, DATE/TIMESTAMP single-scaling, decimal half-away rounding + overflow→NULL, UTF transcoding, Required→Optional schema/dispatch consistency — holds under tracing.

### Critical
**None confirmed.**

One **blocking unknown (not a code finding):** Enterprise CI is red on the current head — build **244967**, tested commit `6e129640`. Per the "never dismiss a failure" rule it must be investigated and proven unrelated (or fixed) before merge. Not reproducible from the OSS checkout. This is the single item gating merge.

### Moderate

**M1 — Test-coverage gaps on the new code — RETRACTED.**
Original concern: foreign-parquet byte-array DECIMAL decode arms and the DOUBLE-array arm untested; no empty-partition test; GEO/BINARY/ARRAY only in a skip-on-throw loop; DECIMAL/fixed→fixed untested at multi-row-group; converted-column-as-JOIN-key untested for some types.
Resolution: this PR flips the column-conversion probability `0.0 → 0.1` in `testConvertPartitionToParquet` and `testCreateTableAsParquet` (`WalWriterFuzzTest`), so the fuzzer now walks conversions on parquet across multi-row-group, all encodings, all writer-emittable conversion pairs, dedup, and O3 merges — the matrix is covered by fuzz, and enumerating it explicitly is a combinatorial explosion. Converted-column-as-JOIN-key is not a distinct code path (the cast happens in the accessor before any join sees it). The only residual was the writer-unreachable foreign-parquet byte-array arms — and those (`decode_byte_array_decimal_mode` / `_dict_mode`, decode.rs:10/:1149) exist on master **unchanged**; this PR neither adds nor re-routes them, so they're out of scope. The only genuinely new foreign-parquet arms (`Int32ToDoubleConverter`, the DOUBLE-array arm) already carry constructor-validation tests. **No valid coverage finding against this PR.**

**M2 — `getStrLen`/`getVarcharSize` double-format per row — DOWNGRADED to Minor / won't-fix.**
`PageFrameMemoryRecord.getStrLen` runs the full `convertFixedToStr`/`readVarValueForConversion` just for `.length()`, then a following `getStrA` re-runs the identical conversion; `getVarcharSize` formats into the sink and discards all but `.size()`. So length-then-value consumers pay the formatter 2× per row on a lazily-converted parquet STRING/VARCHAR column (still zero-GC). Resolution: the trigger is narrow (needs both a pending lazy conversion to STRING/VARCHAR *and* a length-then-value consumer on that column; non-converted columns never enter the `hasTypeCasts` branch), and the only fix is per-cell memoization whose validity state has to live on the hot flyweight record that every query carries and resets. Complicating the hot record for all queries to recover a 2× constant factor on a rare path is a bad trade. Real, but Minor and won't-fix.

### Minor (verified)

- **`griffin/CLAUDE.md` markdown bugs** — line 9 opens a fence as ```` ```2 ```` (stray `2`); line 23 reads `For each PARQUET partition:2` (stray trailing `2`). The one concrete thing worth fixing.
- **`needsLazyConversion(int)`** (`PageFrameFilteredMemoryRecord.java:623`) returns boolean — should use the `is…`/`has…` prefix.
- **`getStrLen` var-branch passes the unmasked source tag** to `readVarValueForConversion` (`PageFrameMemoryRecord.java:~700`) where every sibling masks `& 0xFF`. Unreachable today (var→fixed/decimal/nano columns are never read via `getStrLen`), but inconsistent — mask it for future-proofing.
- **`PageFrameFilteredMemoryRecord` fixed→var column-top index domain** — string/varchar getters swap in the compacted `rowIndex` before delegating, so `convertFixedToStr`'s column-top guard compares a compacted index against an absolute column-top for no-sentinel sources under late materialization. Reachability couldn't be fully proven — please confirm a compacted filtered frame can't carry a divergent `columnTops`, and add an assert/comment.
- **`init()` snapshots `sourceColumnTypes` O(columnCount) per frame navigation** (gated by `hasTypeCasts`, zero cost on the common path). Avoidable with a pool generation/identity check. `PageFrameMemoryRecord.java:1526`/`:816`.
- **`row_groups.rs` fixed→fixed numeric conversion is scalar** (`convert_numeric_in_place`, `scale_i64_in_place`) — monomorphized `fn`-pointer loops (no capture/boxing), but auto-vectorization is likely blocked by unaligned access + the null-sentinel branch. A `chunks_exact` SIMD path would speed the O3 merge/rewrite path. Low priority.
- **Naming the package design docs `CLAUDE.md`** (`cairo/`, `griffin/`) risks them being auto-loaded as agent guidance rather than read as human architecture docs. Content is accurate. Maintainer call.
- **Non-ASCII (`→ × ÷ μ ≤` em-dash) in new code comments** — the ASCII rule is scoped to log messages and master already uses these in comments. Mention only.
- **`isNoNullSentinelFixedType` duplication** in the diff snapshot — already hoisted to `ColumnType.isNoNullSentinelFixedType` in the live head. No action.

### Downgraded (false positives dismissed after verification)
- **`navigateTo` cache-hit forces full `openParquet` rebuild on every scan** — disproved: `openParquet(frameIndex)` returns at lines 991-993 when `hasFullProjectionMap` is true; the cache-hit path is a boolean check + `activateDecoder`. No regression on non-converted scans; the unconditional call is the fix for the stale-conversion-metadata bug.
- **Misattributed Javadoc above `isNoNullSentinelFixedType`** — disproved: the `convertFixedToStr` Javadoc sits directly above `convertFixedToStr`.
- **Shared `sourceColumnTypes`/pooled-sink data race across reduce workers** — safe: every `PageFrameMemoryPool` is single-thread-confined; `init()`/`copyOf()` deep-copy the conversion state; cross-thread reuse is mediated by `boundPool` + `bindGeneration`.
- **`TableWriter.getParquetColumnType` reusable-decoder double-free / UAF** — safe: `ParquetFileDecoder.close()` is idempotent and reopenable; free in `finally` alongside munmap is correct on all paths.
- **Rust panics/overflow** — none reachable: `try_reserve`-backed buffers, `checked_*`/`saturating_sub` on all decimal/timestamp/row-group arithmetic; bare `row_group_hi - row_group_lo` remains only in `#[cfg(test)]`. The diff converts eight encoder `assert_eq!` aborts into recoverable `Err` — a net panic-safety improvement. `decode.rs:85 dict.unwrap()` is guarded by a `Some(_)` match arm.
- **DDL pre-pass redundant footer re-parse** — essentially unavoidable: the inspection loop only runs for symbol/chained/dedup conversions, and partitions that re-decode are about to be fully rewritten anyway.

### Summary
**Verdict: approve. The only thing standing between this PR and merge is the red enterprise CI (build 244967, commit `6e129640`).**

Both Moderate findings resolved after discussion: M1 retracted (fuzz now walks the writer-emittable conversion matrix; the residual foreign-parquet byte-array arms predate this PR and are unchanged), M2 downgraded to a won't-fix Minor (rare trigger, costly per-query state). No Critical or Moderate code defect stands across all 7 passes. ~30 draft findings verified, ~8 dropped as false positives; essentially all confirmed findings are in-diff — the cross-context pass walked the `Async*` factory chain, the pool threading model, and `getPageAddress`/`columnTops` ownership and cleared them. The PR honestly states its tradeoffs (conversion cost moves to the query path; SYMBOL/chained/dedup-crossing conversions force eager per-partition rewrite).

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) — `/review-pr` level 2</sub>

ideoma:
Review: OSS PR #7024 — column type conversions with parquet partitions

Level 2, OSS-only (tandem enterprise #1089 out of scope per your choice). 9 parallel agents (Java correctness+cross-context, Rust value-correctness, Rust panic-safety, concurrency, resources, performance, adversarial-performance, tests, code-quality) + independent verification of every Critical-candidate. Scope: 102 files, +13.5k/−793 (Java + Rust + tests). Local checkout is at PR head 192ab623.

This is a large, unusually disciplined feature PR. After deep verification — the Rust crate was built and its parquet suites run, the conversion math was traced against the native ColumnTypeConverter/C++ ground truth, the concurrency model was proven single-thread-confined, and every cross-context fast-path was walked — no Critical correctness, concurrency, resource-leak, or JVM-abort bug was confirmed. Findings are performance (cast-column-scoped), test-quality, and code-quality. None blocks merge.

Critical

None confirmed.

Moderate

M1 — getStrLen/getVarcharSize on a fixed→var converted column run the full formatter per row just to test null (in-diff, perf, LIVE)
PageFrameMemoryRecord.getStrLen:713 / getVarcharSize:803-805 call convertFixedToStr/convertFixedToVarchar (the entire Fixed2VarConverter.convert(...) format) and then discard everything but .length()/.size(). I verified a live consumer the structured perf agent missed: CountStrGroupByFunction.computeKeyedBatch:71 and computeNext:80 (and CountVarcharGroupByFunction:70/79) call arg.getStrLen(record)/getVarcharSize(record) per row on the page-frame record. So count(c) over an e.g. INT→STRING converted parquet column formats every value to a string solely to check non-null — O(format-width)/row where an O(1) null probe suffices. Correct results, just slower; bounded to converted columns. Fix: a null-only fast path (the address==0 + sentinel/column-top checks already in convertFixedToStr run before formatting), or a last-cell (columnIndex,rowIndex) sink cache mirroring the PR's own parseVarToUuid cache.

> Reply: Not a blocker on its own, and the same idea is a worthwhile future perf improvement across the board — every conversion accessor (not just the fixed→var string/varchar getters flagged here) could short-circuit on a null check before materializing the converted value, returning early whenever the caller only needs null-ness (count, IS NULL, length/size probes) instead of running the full converter. Worth tracking as a general follow-up rather than gating this PR.

Minor

- (Minor performance — reclassified from Moderate) Rust cast-decode for Int→Long/Date/Timestamp/Float/Double is two-pass and non-vectorizable (in-diff, cast-decode path). row_groups.rs routes these Int→* pairs to DecodeAs::Source + a second full convert_numeric_in_place pass (correct, because i32::MIN must map to the wider null sentinel), while the sibling Byte/Short→* pairs were flipped to single-pass DecodeAs::Target. Two compounding issues: (a) the second pass is genuinely avoidable — the PR already ships the pattern (FloatToIntRangeCheckConverter, a per-value Converter mapping out-of-range to the sentinel during decode); an analogous Int→Long sentinel-widen converter would make it single-pass. (b) convert_numeric_in_place takes its null-check/convert transforms as fn pointers and is not #[inline], so even within one monomorphization LLVM cannot inline/vectorize the loop — unlike the sibling contract_to_bool one function away, which takes impl Fn. Per-row-group decode loop over converted columns. Fix: change the two params to generic impl Fn, and/or add the Int sentinel-widen converter for single-pass.
- (Minor test improvement — reclassified from Moderate) Cross-context parity tests verify results but don't pin factory selection (in-diff). The ~46 Async* / SampleBy / JOIN cross-context tests (ParquetColumnTypeConversionTest) assert correctness only via assertSqlCursors parity against the native oracle. Only testVectorizedGroupByOverAlteredParquetColumn:3283 asserts the plan. Each of these tests exists to exercise a specific ... || frameMemory.hasColumnTypeCasts() gate (e.g. AsyncJitFilteredRecordCursorFactory.java:332 + 9 siblings). A parity-only test does not prove the gated factory was selected — if the planner ever degrades to a fallback, parity still passes on the fallback and the gate is never executed, so the test protects nothing. Fix: add .assertsPlanContaining(...) / .withPlanContaining(...) to the cross-context helpers to pin factory selection (especially the JIT-filter test, which never asserts JIT engaged).
- SortKeyEncoder.encodeVarcharBatch:1346 lacks the getSourceColumnType(...) != -1 guard its three siblings (encodeFixed8Batch/encodeFixedWideBatch/encodeStringBatch) have. Verified not a live bug — fixed→VARCHAR decodes as a fixed type, so remapColumns leaves aux=0 and the existing auxAddr==0 check falls back to the converting per-row path; var→var keeps sourceColumnTypes==-1. Add the guard for parity/defense-in-depth — current safety rests on an implicit remapColumns invariant.
- init() copies sourceColumnTypes O(columnCount) on every frame switch (gated behind hasTypeCasts, so zero cost on normal tables), with no skip when the source array/generation is unchanged — the common case of a uniformly-converted table re-snapshots identical data each frame. A generation guard makes it O(1).
- O3PartitionJob.convertVarColumnToFixed:4438-4443 copies an already-contiguous UTF-16 STRING value char-by-char into a sink before parsing, purely to obtain a CharSequence; a DirectString flyweight over (charAddr,len) is zero-copy. O3 write path (per row group), not per query.
- PageFrameFilteredMemoryRecord scalar getters delegate to super without the rowIndex save/restore the string getters perform (:171-437 vs :441-466). Verified correct today (scalar var→fixed re-derives the row through the overridden getStr0/getVarchar), but fragile — a future fixed→var routed through a scalar getter would read the wrong filtered row. Add a comment/assert.
- Split bit-layout codec with no shared definition: the encode side (PageFrameMemoryPool.resolveParquetColumn <<8/<<16/<<24) and decode side (PageFrameMemoryRecord &0xFF/>>8/>>16/bit-24) implement the same sourceColumnTypes wire format, coupled only by hand-maintained duplicated comments. Extract shared constants/codec.
- Member ordering (alphabetical-within-kind): PageFrameMemoryPool.resolveParquetColumn:1110 (appended after unbind); O3PartitionJob new statics out of order and the package-private static group wedged inside the private-static block; ConvertOperatorImpl.isParquetStorageCompatible:539 (static after instance methods); ColumnType.isNoNullSentinelFixedType:567; PageFrameFilteredMemoryRecord needsLazyConversion/getRowIndex order.
- Naming: PageFrameFilteredMemoryRecord.needsLazyConversion(int):623 is a boolean predicate without the required is/has prefix (CLAUDE.md).
- CLAUDE.md-named design docs (core/.../cairo/CLAUDE.md, core/.../griffin/CLAUDE.md) are feature design docs (200-380 lines) but, being named CLAUDE.md, get auto-loaded as directory-scoped agent guidance for all work under those huge packages (they were injected into this very session). Content is excellent; consider renaming to PARQUET_COLUMN_CONVERSION.md/README.md. (May be intentional — your call.)
- PR body has no Test Plan section despite +13.5k of tests; QuestDB convention expects one (plain - bullets).
- Lower-priority test gaps: lazy fixed→var read across multiple row groups is asserted only after the eager rewrite (testIntToStringConvertToNativeMultiRowGroup, vs the var→fixed/var→var arms which assert the lazy multi-RG read); HTTP /exp has no converted-column export test (COPY does, shared gate); Rust-only edge cases (malformed pages, try_reserve failure, decimal overflow boundary) lack direct #[test]s (functionally covered end-to-end via Java).
- Rust readability footgun (not a bug): PlainPrimitiveDecoder<phys,out> vs DeltaBinaryPackedDecoder<out,phys> use opposite generic-param order; both widen correctly.

Downgraded (false positives / already-fixed, dismissed after source verification)

- Sync vectorized GROUP BY (VectorAggregateEntry) reads converted columns as raw bytes — REFUTED. SqlCodeGenerator.java:8758 sets canVectorize = !factory.hasParquetConvertedColumns(executionContext) and gates the whole GroupByRecordCursorFactory on it; SampleBy first/last is gated the same way (:8049). The unguarded getPageAddress path is unreachable for converted columns. (I chased this as the prime P0 candidate; the author handled it.)
- -encoded == -1 sentinel collision for var→fixed — impossible: STRING=11/VARCHAR=26 ⇒ -encoded ≤ -11; no-sentinel fixed sources stored positive.
- Data race on shared sourceColumnTypes/hasTypeCasts — safe: every PageFrameMemoryPool is single-thread-confined (per-cursor / per-reduce-task), the record deep-copies sourceColumnTypes on every init(), and the worker→collector handoff is fenced by an AtomicInteger happens-before edge.
- ParquetBuffers.columnTops leak / dangling — safe: freed in close(), the ctor catch, and clearAddresses(); reopened in reopen() — full parity with pageAddresses.
- Reachable Rust panic / JVM abort — none: encode dispatch is consistent by construction (pt.field_info.repetition drives both), the 5 encoder assert!s are now recoverable Errs, saturating_sub replaced the row-group subtraction, and allocations use fallible AcVec/try_reserve. Decimal/timestamp scaling uses checked_*/wrapping_* matching C++.
- isNoNullSentinelFixedType duplicated and detached convertFixedToStr Javadoc (both from prior reviews) — already fixed: the helper is hoisted to ColumnType:567; the Javadoc is correctly attached.
- DDL pre-pass re-mmaps/re-decodes each partition 2× (prior review M1) — retraction stands: the inspection loop only runs for symbol/chained conversions, mmap faults only footer pages, and converted partitions are fully rewritten anyway.
- getStrLen double-work is "latent" (structured perf agent) — corrected to LIVE (M1 above); count() over a converted var column is a real consumer.

Summary

Verdict: approve with minor changes. No blocking issues. ~30 draft findings verified, ~9 dropped as false positives / already-fixed (listed above). Findings are essentially all in-diff; the cross-context pass actively walked the out-of-diff callsites (hasColumnTops consumers — all 18 paired with hasColumnTypeCasts; VectorAggregateEntry/vect GROUP BY — gated; SortKeyEncoder batch encoders; ColumnMapping triple callers) and cleared them, with SortKeyEncoder.encodeVarcharBatch the single out-of-diff Minor (parity, not a live bug).

Tradeoffs the PR makes honestly (all stated in its description): conversion cost moves to the query path (lazy per-row decode), and SYMBOL/chained conversions force an eager per-partition rewrite. The most useful follow-up is the test plan-pinning item (now Minor) — the cross-context suite proves correctness but not path selection, which is the exact property those tests were written to guarantee; worth tightening but not blocking.

Two notes outside this OSS review's scope: (1) the tandem enterprise PR #1089 must wire up any ACL for the new state-mutating behavior and stay consistent — not reviewed here per your OSS-only choice; (2) per the "never dismiss a failure" rule, enterprise/tandem CI must be confirmed green on the current head 192ab623 before merge rather than assumed (a prior review flagged a red enterprise build at an older commit).

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 3695 / 4314 (85.65%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetFileDecoder.java|0|1|00.00%|
|:large_blue_circle:|qdbr/src/parquet_write/encoders/numeric.rs|6|24|25.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/ReusablePageFrameMemory.java|1|2|50.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameFilteredMemoryRecord.java|41|71|57.75%|
|:large_blue_circle:|qdbr/src/parquet_read/decode.rs|479|789|60.71%|
|:large_blue_circle:|qdbr/src/parquet_write/encoders/plain/primitive.rs|83|113|73.45%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinRecordCursorFactory.java|6|8|75.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/TablePageFrameCursor.java|4|5|80.00%|
|:large_blue_circle:|io/questdb/cairo/TableUtils.java|25|30|83.33%|
|:large_blue_circle:|io/questdb/cairo/DecimalColumnTypeConverter.java|11|13|84.62%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|106|125|84.80%|
|:large_blue_circle:|qdbr/src/parquet_read/parquet_meta_decode.rs|60|69|86.96%|
|:large_blue_circle:|qdbr/benches/decode_page.rs|7|8|87.50%|
|:large_blue_circle:|qdbr/benches/encode_page.rs|7|8|87.50%|
|:large_blue_circle:|qdbr/src/parquet_read/row_groups.rs|1187|1315|90.27%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryRecord.java|369|393|93.89%|
|:large_blue_circle:|qdbr/src/parquet_write/update.rs|240|253|94.86%|
|:large_blue_circle:|io/questdb/cairo/O3ParquetMergeContext.java|36|38|94.74%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|164|171|95.91%|
|:large_blue_circle:|qdbr/src/parquet_write/encode.rs|143|150|95.33%|
|:large_blue_circle:|io/questdb/griffin/ConvertOperatorImpl.java|45|46|97.83%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|418|425|98.35%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/RowGroupBuffers.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinFastRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/ReadParquetRecordCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableReaderMetadata.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/rle_dict_varchar_slice.rs|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/encoders/plain/tests.rs|17|17|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriterMetadata.java|8|8|100.00%|
|:large_blue_circle:|qdbr/src/parquet/qdb_metadata.rs|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileReader.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableColumnMetadata.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|4|4|100.00%|
|:large_blue_circle:|qdbr/tests/row_groups.rs|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/buffers.rs|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/ColumnTypeConverter.java|53|53|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ParquetRowGroupFilter.java|5|5|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/delta_binary_packed.rs|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/NanosTimestampDriver.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractRecordMetadata.java|2|2|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/plain.rs|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/RecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/MillisTimestampDriver.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/str/Utf8s.java|5|5|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/rle_dictionary.rs|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/rle.rs|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyEncoder.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/ColumnMapping.java|5|5|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/encoders/delta_binary_packed.rs|25|25|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/delta_length_array.rs|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/ExportQueryProcessor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/ColumnType.java|2|2|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/converters.rs|38|38|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/ParquetExportMode.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/MicrosTimestampDriver.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AbstractPageFrameRecordCursorFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncJitFilteredRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CopyExportFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetPartitionDecoder.java|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/mod.rs|9|9|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/schema.rs|1|1|100.00%|

bluestreak01 (approved):
Approving after a level-3 review (independent pass over the full change surface, verifying candidate findings against source on both sides of the JNI boundary).

No Critical or Moderate correctness, concurrency, resource-leak, JVM-abort, or heap-safety issue survived verification. Confirmed findings are all Minor:

- Dead code: convertVarToStr and the srcTag < -1 branches in getStrA/getStrB/getStrLen are unreachable (STRING/VARCHAR targets never encode sourceColumnTypes < -1). getStrLen's < -1 branch also passes the source tag unmasked where siblings mask & 0xFF - harmless while dead, latent if reachable.
- needsLazyConversion(int) should use the is.../has... boolean prefix.
- getStrLen/getVarcharSize double-format the value per row for length-then-value consumers (zero-GC, narrow trigger).
- hasParquetConvertedColumns is coarse: disables JIT/vectorization/sample-by-first-last for the whole query whenever any column is chain-converted and any parquet partition exists, even if the query does not touch the converted column. Correct (fail-safe) but a broad pessimization.
- dict.unwrap() at decode.rs:827 is guarded but inconsistent with sibling Some(dict_page) arms - bind directly.
- FLOAT string-size estimate (15) is exactly the Ryu worst case; correct but zero margin, and the write precedes the assert. A comment pinning the derivation would harden it.

Verified safe (candidate bugs dismissed): filtered-record column-top guard (Rust row-filter compacts the column top, row_groups.rs:1930-1947); scalar-getter delegation without rowIndex save/restore (var->fixed reads route through the overridden getStr0/getVarchar); raw page-address hot paths (systematically gated); O3 batch conversion buffer sizing (DOUBLE 24 / FLOAT 15 fit 25/15); chained-conversion column resolution (getReplacingChainHead walks to the head with cycle detection); -encoded sentinel collision (STRING=11/VARCHAR=26); Rust panic/overflow sites (test-only or guarded).

Non-code item to confirm, not block: re-run enterprise CI (tandem #1089) on the current head to prove it green rather than assume stale.

Tradeoffs are disclosed honestly: conversion cost moves to the query path, and SYMBOL/chained/dedup-crossing conversions force an eager per-partition rewrite.

PR #7025, opened 2026-04-24
chore(core): remove orphan directory left behind when CREATE TABLE fails

RaphDal:
## Summary

`CairoEngine.createTableOrViewOrMatViewUnsecure` runs two filesystem-visible steps in sequence: `createTableOrViewOrMatViewUnsafe` (or `createTableOrMatViewInVolumeUnsafe`) writes the table directory and `_meta` to disk, then `tableNameRegistry.registerName` hydrates the metadata cache, which re-opens `_meta` via `MetadataCache.hydrateTableStartup` -> `MemoryCMRImpl.smallFile` -> `TableUtils.openRO`. If that second `openRO` fails, the exception propagates out and the existing catch at `CairoEngine.java:2098` only cleans up the sequencer side for WAL tables:

```java
} catch (Throwable th) {
    if (struct.isWalEnabled()) {
        tableSequencerAPI.dropTable(tableToken, true);
    }
    throw th;
}
```

Nothing undoes the on-disk directory. For a non-WAL table, the dirname is just `<tableName>` (or `<tableName>~` with mangle) and is stable across retries, so the next CREATE for the same name hits the `TableUtils.exists` guard at `CairoEngine.java:2004` (or `TableUtils.existsInVolume` for IN VOLUME) and throws `[-1] name is reserved [table=<name>]`. The name is effectively poisoned until an operator removes the directory manually. For a WAL table the dirname is `<tableName>~<tableId>` and the retry generates a fresh `tableId`, so the existence check returns NOT_EXISTS regardless; retry already worked on master, the orphan just accumulated as clutter under `dbRoot`.

The fuzz harness surfaced the non-WAL case on CI in `WalWriterFuzzTest.testWalApplyEjectsMultipleTables`. The CI log shows `createTableOrViewOrMatViewUnsafe` succeeding and unlocking the pool, then an injected IO failure hitting the `openRO(_meta)` inside `registerName`, then `FuzzRunner.applyNonWal`'s retry via `FuzzDropCreateTableOperation.recreateTable` failing with `name is reserved [table=testWalApplyEjectsMultipleTables_0_nonwal]`. The table is non-WAL (note the `_nonwal` suffix) even though the fuzz class is `WalWriterFuzzTest`.

This PR tracks whether the on-disk create succeeded with a local `filesystemCreated` flag and, in the existing outer catch for non-WAL tables, runs a best-effort `FilesFacade.unlinkOrRemove` on `dbRoot/<tableDir>`. For IN VOLUME tables the dbRoot entry is a soft link into the volume, so before unlinking the dbRoot entry the cleanup calls `ff.readLink()` to recover the volume-side target and removes that too — mirroring the pattern `WalPurgeJob` already uses on table drop. Removing only the dbRoot soft link would not satisfy the subsequent `existsInVolume` check, which reads the volume directly. Errors inside the cleanup are logged and swallowed so the original exception still propagates.

WAL tables are intentionally excluded from the cleanup: their retry already works on master (fresh `tableId` -> fresh dirname), so adding an unlink there would only have a cosmetic effect and is out of scope here.

## Changes

- `core/src/main/java/io/questdb/cairo/CairoEngine.java`
  - `createTableOrViewOrMatViewUnsecure`: add `boolean filesystemCreated`, set after the on-disk create call. In the outer `catch (Throwable th)`, when `filesystemCreated && !struct.isWalEnabled()`, best-effort `unlinkOrRemove(dbRoot/<tableDir>)`. If that entry is a soft link (IN VOLUME table), first resolve the target with `ff.readLink()` and then `unlinkOrRemove` it too. Cleanup errors are logged and swallowed; the original exception is rethrown.

- `core/src/test/java/io/questdb/test/cairo/CreateTableTest.java`
  - `testCreateNonWalLeavesOrphanDirWhenRegisterNameHydrateFails`: installs a `TestFilesFacadeImpl` whose `openRO` returns `-1` for the `_meta` file of the target table when a per-test flag is set. CREATE -> DROP -> arm flag -> CREATE (must fail with `could not open`) -> CREATE (must succeed). On master the final CREATE fails with `name is reserved`; after the fix it succeeds.
  - `testCreateInVolumeCleansUpOnRegisterNameHydrateFailure`: same fault injection, but the table is created `IN VOLUME '<alias>'`. Exercises the volume-side unlink — on master the retry fails with `name is reserved` because the volume target dir remains after the dbRoot symlink is gone; after the fix it succeeds.
  - `testCreateCleanupFailureDoesNotMaskOriginalError`: arms both the `openRO(_meta)` fault and an `unlinkOrRemove` fault. The original `CairoException` from the injected `openRO` failure must still propagate — the swallowed cleanup error must not replace or shadow it.

## Effects

Positive:
- Transient IO errors during `registerName`/`hydrateTable` no longer leave a non-WAL table name blocked. Callers (SQL CREATE TABLE, ILP table auto-create, REST API, DDL listeners) can retry under the same name without operator intervention. Covers both regular and IN VOLUME tables.
- The fuzz-observed failure in `WalWriterFuzzTest.testWalApplyEjectsMultipleTables` and any similar fuzz runs that inject IO during hydrate can progress past the poisoned state.

Neutral and negative:
- WAL tables are not touched by this PR. Retry was already working there on master; the dbRoot orphan directory it leaves behind on failure still lingers until an operator removes it. Addressing that cleanliness issue is a separate concern.
- The cleanup is best-effort. If `unlinkOrRemove` itself fails (for example, the filesystem error that caused the original exception is still in effect), the directory will persist and a retry will still see `name is reserved`. The error is logged as `could not clean up partial table after failed create` (and, for the volume arm, `could not clean up partial table in volume`). This is the same observable behavior as master; the fix is a strict improvement for the common case where the failure was on a specific `openRO` rather than blanket-failing.
- The cleanup changes the contents of `dbRoot` and of the volume in the error path. Callers that previously inspected the orphan directory for post-mortem diagnosis will no longer find it. The pre-create `LOG.info` lines in `TableUtils.createTableOrViewOrMatView` plus the new cleanup error logs still document what happened.
- No allocations on either path: the volume-side target is recovered through `ff.readLink()` into a thread-local `Path`, same as `WalPurgeJob` does on drop.

## Test plan

- `mvn -Dtest='CreateTableTest#testCreateNonWalLeavesOrphanDirWhenRegisterNameHydrateFails+testCreateInVolumeCleansUpOnRegisterNameHydrateFailure+testCreateCleanupFailureDoesNotMaskOriginalError' test` — passes with the fix. The first two fail on master with `name is reserved`; the third passes on master too but locks in the cleanup-error-swallowing contract.
- `mvn -Dtest='CreateTableTest,TableNameRegistryTest,CairoEngineTest' test` — 94 tests, no regression.
- `mvn -Dtest='WalWriterFuzzTest#testWalApplyEjectsMultipleTables' test` — happy-path sanity check on the fuzz test that originally surfaced the bug. Note the fuzz is stochastic: this run is not guaranteed to re-exercise the exact CI seed (`3099845320031L, 1776975582561L`), but the deterministic unit tests above cover the failure mechanism.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/CairoEngine.java
+++ b/core/src/main/java/io/questdb/cairo/CairoEngine.java
@@ -2056,6 +2056,7 @@ public class CairoEngine implements Closeable, WriterSource {
             while (!lockTableCreate(tableToken)) {
                 Os.pause();
             }
+            boolean filesystemCreated = false;
             try {
                 String lockedReason = lockAll(tableToken, "createTable", true);
                 boolean locked = true;
@@ -2066,6 +2067,7 @@ public class CairoEngine implements Closeable, WriterSource {
                         } else {
                             createTableOrViewOrMatViewUnsafe(mem, blockFileWriter, path, struct, tableToken);
                         }
+                        filesystemCreated = true;
 
                         if (struct.isWalEnabled()) {
                             tableSequencerAPI.registerTable(tableToken.getTableId(), struct, tableToken);
@@ -2099,6 +2101,21 @@ public class CairoEngine implements Closeable, WriterSource {
                 if (struct.isWalEnabled()) {
                     // tableToken.getLoggingName() === tableName, table cannot be renamed while creation hasn't finished
                     tableSequencerAPI.dropTable(tableToken, true);
+                }
+                // If on-disk files were created before the failure (e.g. registerName threw),
+                // remove them so the table name does not stay blocked by an orphan directory.
+                if (filesystemCreated) {
+                    try {
+                        final FilesFacade ff = configuration.getFilesFacade();
+                        path.of(configuration.getDbRoot()).concat(tableToken).$();
+                        if (!ff.unlinkOrRemove(path, LOG)) {
+                            LOG.error().$("could not clean up partial table after failed create [table=").$(tableToken)
+                                    .$(", errno=").$(ff.errno()).I$();
+                        }
+                    } catch (Throwable cleanupEx) {
+                        LOG.error().$("cleanup after failed create threw [table=").$(tableToken)
+                                .$(", err=").$(cleanupEx).I$();
+                    }
                 }
                 throw th;
             } finally {
--- a/core/src/main/java/io/questdb/cairo/CairoEngine.java
+++ b/core/src/main/java/io/questdb/cairo/CairoEngine.java
@@ -2101,15 +2101,23 @@ public class CairoEngine implements Closeable, WriterSource {
                 if (struct.isWalEnabled()) {
                     // tableToken.getLoggingName() === tableName, table cannot be renamed while creation hasn't finished
                     tableSequencerAPI.dropTable(tableToken, true);
-                }
-                // If on-disk files were created before the failure (e.g. registerName threw),
-                // remove them so the table name does not stay blocked by an orphan directory.
-                if (filesystemCreated) {
+                } else if (filesystemCreated) {
+                    final FilesFacade ff = configuration.getFilesFacade();
                     try {
-                        final FilesFacade ff = configuration.getFilesFacade();
                         path.of(configuration.getDbRoot()).concat(tableToken).$();
+                        Path volumeTarget = null;
+                        if (ff.isSoftLink(path.$())) {
+                            volumeTarget = Path.getThreadLocal2("");
+                            if (!ff.readLink(path, volumeTarget)) {
+                                volumeTarget = null;
+                            }
+                        }
                         if (!ff.unlinkOrRemove(path, LOG)) {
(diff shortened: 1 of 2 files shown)


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 17 / 20 (85.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|17|20|85.00%|

bluestreak01 (approved):
Reviewed. Fix is targeted and correct. Suggested follow-ups in review thread (DDL listener rollback, WAL orphan cleanup, commit scope) are non-blocking.

PR #7026, opened 2026-04-24
chore(parquet): avoid false read failure on mostly-null dictionary-encoded columns

RaphDal:
## Summary

A QuestDB reader could reject a valid Parquet file with `Layout: Unexpected end of rle iterator` when a dictionary-encoded column contained a data page with zero non-null values and a non-empty global dictionary. This was observed on QuestDB-written ClickBench `hits.parquet` for mostly-null SYMBOL columns split across multiple data pages.

Parquet's RLE-dictionary data-page payload starts with a one-byte `bits_per_key` followed by a hybrid-RLE index stream. When a page has zero non-null values but the global dictionary is non-empty, `bits_per_key` is positive and the hybrid-RLE encoder emits a bitpacked-zero-groups header (`0x01`) and nothing else. The overall payload is `[bits_per_key, 0x01]` and is a legal encoding -- a caller that only consumes nulls for this page should never need to pull a run from the stream.

Three `try_new` constructors pulled the first RLE run eagerly, before knowing whether the caller would actually consume any non-null values:

- `RleDictionaryDecoder::try_new` at `core/rust/qdbr/src/parquet_read/decoders/rle_dictionary.rs` -- the primitive / SYMBOL path (the site reproduced in the field).
- `RleDictVarcharSliceDecoder::try_new` at `core/rust/qdbr/src/parquet_read/decoders/rle_dict_varchar_slice.rs` -- the `VarcharSlice` path.
- `RleDictionarySlicer::try_new` at `core/rust/qdbr/src/parquet_read/slicer/rle.rs` -- the legacy `Varchar` slicer path.

On the valid zero-values payload, the first eager `decoder.next()` returns `None` and each constructor turns that into `Unexpected end of rle iterator`, failing the entire column chunk.

The fix drops the eager decode in all three constructors. The decoders are already initialized with an empty inner `RleIterator`, and every consumer method (`push`, `push_slice`, `skip`, `next`, `next_into`) already falls back to `decode()` / `decode_next_run()` when the inner iterator is exhausted. So the eager call is redundant on a normal page and is the sole cause of the false-positive error on a zero-values page. Payloads that really are malformed (for example, a page that claims a positive `bits_per_key` but provides no header byte at all) still surface the same error, just on the first consumer call rather than at construction time.

The `RleDictionarySlicer::try_new` constructor also read the bit-width byte via `buffer[0]`, which would panic on an empty buffer. Since a panic inside a JNI-called Rust function aborts the JVM, this is replaced with `*buffer.first().ok_or_else(...)` returning a layout error consistent with the defensive check already present in the other two constructors.

The `RleBooleanDecoder` path was also audited and already uses lazy decoding via `ensure_run()`; no change there.

## Changes

- `core/rust/qdbr/src/parquet_read/decoders/rle_dictionary.rs` -- `RleDictionaryDecoder::try_new` no longer calls `res.decode()?` before returning.
- `core/rust/qdbr/src/parquet_read/decoders/rle_dict_varchar_slice.rs` -- `RleDictVarcharSliceDecoder::try_new` no longer calls `res.decode_next_run()?` before returning.
- `core/rust/qdbr/src/parquet_read/slicer/rle.rs` -- `RleDictionarySlicer::try_new` no longer calls `res.decode()?` before returning, and now reads the bit-width byte via `buffer.first()` so that an empty buffer surfaces a layout error instead of panicking on `buffer[0]`.
- Test coverage. Each of the three constructors gets a matching set of regression tests so the contract is pinned symmetrically across primitive / `VarcharSlice` / legacy `Varchar` paths:
  - `test_*_zero_values_positive_bit_width` -- constructs the decoder from a `[bits_per_key, 0x01]` payload, asserts `try_new` returns `Ok`, and (where applicable) verifies that pushing only nulls never touches the RLE stream.
  - `test_*_truncated_payload_errors_on_first_consume` -- pins the error-timing shift: a positive `bits_per_key` with no hybrid-RLE body now constructs successfully and surfaces `Unexpected end of rle iterator` on the first consumer call.
  - `test_*_empty_buffer_errors_on_construction` (and the upgraded `test_empty_rle_buffer` for the `VarcharSlice` path) -- pins the `buffer.first()` bounds check with a message-level assertion rather than a bare `is_err()`.

The edits are strictly a removal of the eager call plus a short comment explaining why the construction path must stay lazy, plus the `buffer[0]` bounds check on the legacy slicer. All other behaviour -- including the existing error message text and the error path for genuinely truncated payloads -- is preserved.

## Tradeoffs and caveats

- Error-surfacing timing shifts. An RLE payload that is malformed past the header byte (for example, `[6, 0x01]` immediately followed by a truncated second run, or a positive `bits_per_key` with an empty tail) used to fail inside `try_new`; it now fails on the first `push` / `push_slice` / `skip` that actually needs data. The error type and message are unchanged. Callers that wrap construction and consumption in separate error-handling blocks should confirm both paths surface the error the same way -- in this repo, both paths propagate `ParquetResult` identically.
- No format change, no on-disk change, no new allocations. The fix is read-only and does not affect writer output.
- The fault was not hit by existing unit tests because every prior test for these constructors used a payload with at least one value. The new tests close that gap for the zero-values-positive-bit-width shape, the truncated-payload error timing, and the empty-buffer bounds check, symmetrically across all three decoder paths.

## Test plan

- `cargo test --lib parquet_read` in `core/rust/qdbr` -- full `parquet_read` suite passes (279 tests), including the new regression tests.
- The `*_zero_values_positive_bit_width` tests were confirmed to fail on `master` before the fix with `Layout: Unexpected end of rle iterator`, and to pass after it.
- Pre-existing coverage for the three decoders (skip / push / next / push_slice across RLE, bitpacked, byte-indices, and zero-bit-width paths) still passes; the change does not alter any non-empty-page code path.


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 151 / 167 (90.42%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/parquet_read/slicer/tests.rs|35|41|85.37%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/rle_dictionary.rs|57|62|91.94%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/rle_dict_varchar_slice.rs|55|60|91.67%|
|:large_blue_circle:|qdbr/src/parquet_read/slicer/rle.rs|4|4|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=229092)
**Tested Commit:** `ca781bd9f0446b33392bbcbc1bae9ee3d1b8942d`

PR #7037, opened 2026-04-28
feat(sql): add nth_value support for LONG and TIMESTAMP arguments

jovfer:
Part of https://github.com/questdb/roadmap/issues/120 gap 2.

## Summary

- Add `nth_value(<long>, n)` window function — resolves `nth_value` to a `LONG`-returning factory whose routing mirrors the Double variant: per-partition / whole-partition / ROWS / RANGE bounded and unbounded frames, plus the lock-in fast path for `UNBOUNDED PRECEDING ... K PRECEDING` and the current-row 1-row frame.
- Add `nth_value(<timestamp>, n)` window function — same coverage, returning `TIMESTAMP`. Each inner class overrides `getType()` to return `arg.getType()` so `TIMESTAMP_MICROS` and `TIMESTAMP_NANOS` subtypes propagate intact, matching `FirstValueTimestamp`.
- Both factories use `Numbers.LONG_NULL` as the sentinel for "n exceeds the current frame size" and for genuine NULL inputs, consistent with `FirstValueLong` / `LastValueLong` / `FirstValueTimestamp` / `LastValueTimestamp`.

## Side effect

Adding multiple `nth_value` overloads (`D`, `L`, `N`) shifts arity validation from the per-factory body into the overload resolver. The pre-existing `testNthValueRejectsArity` is updated to expect "there is no matching function `nth_value` with the argument types" instead of "wrong number of arguments for function `nth_value`". This is the same shape of message every other multi-overload window function emits.

## Tradeoffs / limitations

- Each per-type factory duplicates the implementation of the Double variant (same pattern as `FirstValue*`). Compared to a single generic factory this is more code but it preserves zero-GC primitive arithmetic on the hot path and keeps each factory's routing readable in isolation.
- `nth_value` peer-row semantics under `RANGE ... CURRENT ROW` continue to follow the project-wide QuestDB convention (no peer look-ahead). This is unchanged from the Double variant.

## Test plan

- 19 new tests in `WindowFunctionTest.java` covering both factories: basic ordered, partitioned, current-row, ROWS frame (sliding and unbounded preceding), RANGE frame, whole-result-set `over ()`, NULL data propagation, n=1 fast path, and error-case `n=0` for the Long factory; `IGNORE NULLS` rejection for the Timestamp factory.
- Existing `testNthValueRejectsArity` updated to the new dispatcher error message.
- All 90 `nth_value*` tests pass; the wider 149-test ntile / cume_dist / nth_value suite is green.

## Related PRs
- #6925
- https://github.com/questdb/documentation/pull/430


bluestreak01:
@jovfer there are large chunks of code in both functions untested

jovfer:
@bluestreak01 thanks, updated. Bumped to ~95%, the rest is currently dead toTop (hidden by cached 2pass) and `catch` blocks

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 1465 / 1543 (94.94%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/NthValueLongWindowFunctionFactory.java|727|766|94.91%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/NthValueTimestampWindowFunctionFactory.java|738|777|94.98%|

bluestreak01 (approved):

PR #7041, opened 2026-04-29
chore(qwp): add durable-ack opt-in and zone-aware query failover

bluestreak01:
## Summary

This PR has grown from its original scope (making QWP async sender errors visible end-to-end) to bundle three streams of work on the QWP protocol stack:

1. **Async error visibility** — server flush of pending ACKs no longer conditional on a clean frame boundary; client `close()` rethrows latched terminal errors; durable-ack opt-in negotiated through the WebSocket upgrade.
2. **Failover** — per-`Execute` failover loop on the egress query client (`wire-egress.md` §11.9), cursor-engine reconnect loop on the ingress client (`sf-client.md` §13.6), shared host-health / role-filter / error-classification primitives (`failover.md`). Zone-aware preference for same-locality endpoints on both sides.
3. **Documentation** — `docs/qwp/` reorganised into a normative spec tree (wire-ingress, wire-egress, wire-udp, sf-client, failover) with a top-level README that maps audience to required reading.

## Async error visibility

Server (`QwpWebSocketUpgradeProcessor`):
- Always `flushPendingAck` after processing the recv buffer, regardless of whether the buffer ended on a clean frame boundary. The previous COMPLETE-only check starved senders of ACKs whenever a recv landed mid-frame, which under `forceRecvFragmentationChunkSize` (and any small-MTU recv) left the client's `drainOnClose` to time out. Drops the now-unused `FrameProcessResult` enum.
- Stop logging a stack trace when the WebSocket peer disconnects. `PeerDisconnectedException` is caught ahead of the generic `Throwable` handler and logged at info with just the fd; every other `Throwable` still hits the error-with-stack-trace path.
- Echo `X-QWP-Durable-Ack: enabled` in the 101 response when the client sent `X-QWP-Request-Durable-Ack: true` and the server's `DurableAckRegistry` is enabled. The header is omitted on OSS builds, on primaries that have not initialised the registry yet, and when the client did not ask for it.
- Echo `X-QuestDB-Zone: <id>` on 421 role-reject responses so a client whose upgrade fails can still record the rejected host's zone in `WalkTracker`. Without it, role-rejected hosts stay `zone_tier=Unknown` and lose their preference ranking after the first failover.

Client:
- `close()` rethrows latched terminal errors. Any `Throwable` from the flush/drain path is captured, shutdown completes, then the error is rethrown, so a user who only ever calls `close()` cannot silently lose data. Gated on whether the error has already been delivered to a user-registered handler so callers that consume errors out-of-band still get the legacy clean-close behaviour.
- The cursor SF sender's initial WebSocket handshake is async — `connect` no longer blocks the calling thread.
- Defer cursor SF trim to durable-ack when opted in. When `request_durable_ack=on`, OK frames no longer advance the trim watermark on the on-disk SF; only `STATUS_DURABLE_ACK` frames do. Closes the last data-loss window in the QWP ingestion path: with OK-driven trim, a primary that committed to its WAL but failed before uploading to object store could lose rows whose only remaining copy was already trimmed from the client's SF.
- Send keepalive PINGs while waiting on `STATUS_DURABLE_ACK` frames. The OSS server's `flushPendingAck` only runs on inbound recv events (binary message, ping, close), so an opted-in client that finishes publishing and goes idle would otherwise never see the ack frame even after the WAL upload completes server-side. Cadence configurable via `durable_ack_keepalive_interval_millis`.
- `SenderProgressHandler` async dispatcher pairs with the existing `SenderErrorDispatcher`: bounded inbox, lazy-started daemon, volatile handler swap. Watermarks are monotonic, so a full inbox evicts the oldest entry rather than the newest — drops compress, they don't lose information.

## Failover

Both the ingress cursor-engine and the egress query client now follow a documented failover loop with shared primitives in `failover.md` (host-health model, backoff math, role filter, error classification, cross-context defaults cheat sheet).

Ingress (`sf-client.md` §13.6):
- Cursor-engine reconnect loop walks the addr list with backoff and role-filter, classifies errors into terminal / transient / role-reject categories, and short-circuits on durable-ack opt-in mismatch so a `request_durable_ack=on` client landing on a replica fails at the first attempt rather than after the full reconnect budget.

Egress (`wire-egress.md` §11.9):
- Per-`Execute` failover loop with `WalkTracker` / `OnFailoverReset`: mid-stream failover restarts the cursor against the next reachable host, but only after the new host's SERVER_INFO confirms it is a valid target for the chosen `target=` filter.
- `target=primary` / `target=replica` walks past non-matching roles at connect; with only the wrong role available the client throws `QwpRoleMismatchException` rather than silently degrading.
- Address-list parsing: `ws::addr=h1:p1,h2:p2` is equivalent to the repeated-key form (parity with the .NET egress client).

## Zone preference (egress)

- `CAP_ZONE` (0x00000001) added to `QwpEgressMsgKind`. New `QwpServerInfoProvider` interface exposes `getZoneId()` / `getCapabilities()`; the OSS provider returns `null` / `0`, preserving today's STANDALONE wire bytes. `QwpEgressFrameWriter.writeServerInfo` appends a `u16_len+utf8` zone trailer when `CAP_ZONE` is set. With null zoneId and capabilities=0, the writer emits a byte-for-byte v2.0 frame.
- The provider is the single source of truth for the bit/trailer alignment: passing `CAP_ZONE` in capabilities while leaving zoneId null is treated as a no-op, the bit is masked off rather than letting an invalid frame escape.
- Client `QwpHostHealthTracker` ranks hosts on the lexicographic `(state, zone_tier)` tuple; sticky-Healthy preserves only same-zone HEALTHY entries, so a same-zone host that flickers is preferred over a remote-zone host that is currently healthy.
- Ingress silently accepts `zone=` so the same connect string can be shared across ingress and egress clients.

## Documentation: `docs/qwp/`

Reorganises the old top-level `QWP_SPECIFICATION.md` / `QWP_UDP_DESIGN.md` into a normative spec tree:

- `README.md` — index, audience matrix, drift-control checklist for spec-affecting PRs.
- `wire-ingress.md` — QWP1 ingest wire format (WebSocket + UDP shared core); §15.5 points at the cursor-engine reconnect loop.
- `wire-egress.md` — egress wire format; §11.9 owns the per-`Execute` failover loop and `WalkTracker`.
- `wire-udp.md` — UDP-specific deviations from `wire-ingress`.
- `sf-client.md` (new, ~1000 lines) — client-side cursor-engine substrate spec covering Memory mode, Store-and-Forward, FSN, ACKs, errors, connect-string, on-disk slot layout, and the §13.6 ingress reconnect loop.
- `failover.md` (new, ~400 lines) — shared host-health model, backoff math, role filter, error classification, cross-context defaults cheat sheet.
- `design/egress-phase2-backlog.md` — non-normative working notes.

Each spec carries a Reference Implementation footer with a commit hash; the README drift-control section documents the contract that interop-affecting changes update the relevant spec in the same PR.

## Connect-string knobs

Added:
- `request_durable_ack=on` — opt into `STATUS_DURABLE_ACK`-driven trim. Negotiated via the WebSocket upgrade headers; if the server does not echo `X-QWP-Durable-Ack: enabled`, connect fails with a `LineSenderException` rather than letting the SF grow until disk fills. Against a replica or other role-rejecting server, the failure is classified as terminal on the first attempt.
- `durable_ack_keepalive_interval_millis=<millis>` — cadence for the keepalive PING the I/O loop sends while `request_durable_ack=on` AND there are pending durable-ack confirmations. Default 200 ms. `0` or negative disables the keepalive entirely, which is the intended setting for callers that send continuous traffic (organic recv events already prod the server) or that drive their own ping. WebSocket-only — TCP path raises.
- `zone=<id>` (egress) — case-insensitive locality preference. The client compares against the server's advertised zone (SERVER_INFO trailer + `X-QuestDB-Zone` on 421) and ranks same-zone endpoints higher. Ingress accepts the key but ignores it.
- `addr=h1:p1,h2:p2` (egress) — comma-separated address list, equivalent to the repeated-key form.

Semantics changed:
- `close_flush_timeout_millis=0` — now genuinely opts out of `close()` draining AND the pre-drain `checkError()`. Previously `checkError()` ran unconditionally, so a HALT-policy rejection escaped through `close()` even when the user explicitly asked for a fast close. Users opting out observe outcomes via the `SenderProgressHandler` / `SenderErrorHandler` async callbacks instead.

Removed:
- `IN_FLIGHT_WINDOW_SIZE` constant and `inFlightWindowSize(int)` builder method. The in-flight window was already a no-op after the cursor-engine rewrite; the constant and builder method are gone, callers and tests cleaned up.

## Ingest error reporting

- `WalColumnarRowAppender.numericParseError` throws `CairoException.nonCritical()` instead of `QwpParseException` so numeric-coercion failures surface through the same path as other Cairo conversion errors. The now-unused `put(Utf8Sequence)` and `put(char)` overloads on `QwpParseException` are removed.

## Test infrastructure (`AbstractQwpWebSocketTest`)

- New `connectWs` overloads route through `Sender.fromConfig` / `Sender.builder` so refactoring of `QwpWebSocketSender`'s programmatic constructors cannot break tests.
- An `errorHandler` variant lets tests observe `SenderError` dispatches deterministically (`DROP_AND_CONTINUE` never throws from `flush()`, so the async handler is the only signal).
- All overloads default `closeFlushTimeoutMillis` to 60s so fuzz tests that push hundreds of batches against a single-worker server with concurrent ALTERs do not flake on the 5s production default.
- `RestartableQwpServer` (new) — fixture for bouncing the server mid-stream without tearing down the test harness; consumed by the new restart fuzz and failover-loop tests.
- `QwpRow` / `QwpTable` load helpers — typed in-memory representation of expected rows used by the oracle fuzz to compare server state against a pure-Java oracle.

## Test refactors and additions

- `QwpSenderE2ETest`, `QwpWebSocketSenderReceiverTest`: coercion-error tests now register a `SenderErrorHandler` and assert via `CompletableFuture<SenderError>`; the new `testCloseRethrowsLatchedTerminalError` pins the `close()`-rethrow contract end-to-end against a server-side rejection.
- `QwpSenderLowLevelTest`, `QwpWebSocketSenderTlsE2ETest`, `QwpWebSocketBinaryMessageTest`: switch construction sites from `QwpWebSocketSender.connect(...)` to `Sender.fromConfig` / `connectWs` helpers.
- `QwpSenderFuzzTest`: adds a `clientAutoFlushRows` knob and uses small values for `testLoadSmallBuffer`, `testLoadSendSymbolsWithSpace`, and `testCaseVariationReorderingColumnsSendSymbolsWithSpace` so wire frames stay under the server's recv buffer (the bytes-based threshold underestimates wire size due to QWP framing and dict).
- `QwpCursorBoundsCheckFuzzTest`, `QwpEgressFragmentationFuzzTest`: drop frozen-seed regressions back to `TestUtils.generateRandom` now that the corresponding fixes have landed.
- `QwpServerInfoFrameTest`: adds `CAP_ZONE` / zone-trailer round-trip cases — capabilities=0 emits the byte-identical v2.0 frame; `CAP_ZONE` with empty / utf8 / null-with-bit-set is exercised on both write and read paths.
- `QwpQueryClientFailoverLoopTest` (new, ~450 lines): covers the per-`Execute` failover loop end-to-end against a restartable in-process server — initial bind addr-order, mid-stream restart, `target=primary` / `target=replica` walking, zone preference under `(state, zone_tier)` churn.
- `QwpIngressOracleFuzzTest` (new, ~1050 lines): randomised ingest with a pure-Java oracle; asserts the server-side table state matches the oracle row-set after every interleaving of restart, ALTER, flush, and durable-ack windows.
- `QwpIngressServerRestartFuzzTest` (new): one producer thread writes 25-row batches through a single SF sender while a second thread bounces the server 3-5 times. Sender configured with `reconnect_max_duration_millis=120s` and `close_flush_timeout_millis=120s` so the bounces never break user-facing calls. After the bouncer finishes the producer drains on close; the table is asserted to hold exactly the produced row set (count, distinct id, and min/max id all match), which proves no loss and that `DEDUP UPSERT KEYS(ts, id)` collapsed every replay.

## Benchmark

- `QwpEgressReadBenchmarkWide` `DEFAULT_ROW_COUNT` bumped to 1B and the sender now writes through an SF dir under `java.io.tmpdir`, so the bench exercises the cursor SF wire path under sustained load.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added zone ID support for multi-host failover endpoint selection.
  * Introduced durable-ack handshake capability for improved message acknowledgment.
  * Enhanced store-and-forward client durability across server restarts.

* **Documentation**
  * Added comprehensive QWP specification updates including failover primitives, store-and-forward client behavior, and wire protocol enhancements.

* **Tests**
  * Added extensive end-to-end and fuzz tests covering failover scenarios, server restart resilience, and multi-sender durability validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/questdb/questdb/pull/7041)


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 20 / 22 (90.91%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpWebSocketHttpProcessor.java|9|11|81.82%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpWebSocketUpgradeProcessor.java|11|11|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 20 / 22 (90.91%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpWebSocketHttpProcessor.java|9|11|81.82%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpWebSocketUpgradeProcessor.java|11|11|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 41 / 45 (91.11%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|2|4|50.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpWebSocketHttpProcessor.java|9|11|81.82%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressUpgradeProcessor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/DynamicPropServerConfiguration.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpWebSocketUpgradeProcessor.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/codec/QwpServerInfoProvider.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/codec/QwpEgressFrameWriter.java|12|12|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 95 / 159 (59.75%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpWebSocketUpgradeProcessor.java|38|83|45.78%|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|2|4|50.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressUpgradeProcessor.java|15|23|65.22%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpProcessorState.java|15|22|68.18%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpWebSocketHttpProcessor.java|9|11|81.82%|
|:large_blue_circle:|io/questdb/DynamicPropServerConfiguration.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/codec/QwpServerInfoProvider.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/codec/QwpEgressFrameWriter.java|12|12|100.00%|

PR #7044, opened 2026-04-29
fix(sql): between and inside sub-expression

### LINKED ISSUE #6534: assertion error with invalid SQL
### To reproduce

found via fuzzing:
```sql
SELECT fun(col BETWEEN (0 AND 1, col), col;
```

totally non-sensical SQL, but `AssertionError` is still not acceptable:
```
2025-12-14T17:52:35.738916Z C i.q.c.h.p.JsonQueryProcessorState [135291470075] internal error [ex=
java.lang.AssertionError: opStack is empty at ','
	at io.questdb.griffin.ExpressionParser.parseExpr(ExpressionParser.java:373)
	at io.questdb.griffin.SqlParser.expr(SqlParser.java:4292)
	at io.questdb.griffin.SqlParser.expr(SqlParser.java:4303)
	at io.questdb.griffin.SqlParser.parseSelectClause(SqlParser.java:3024)
	at io.questdb.griffin.SqlParser.parseDml0(SqlParser.java:1990)
	at io.questdb.griffin.SqlParser.parseDml(SqlParser.java:1880)
	at io.questdb.griffin.SqlParser.parseSelect(SqlParser.java:2972)
	at io.questdb.griffin.SqlParser.parse(SqlParser.java:4329)
	at io.questdb.griffin.SqlCompilerImpl.compileExecutionModel(SqlCompilerImpl.java:2564)
[...]
```

### QuestDB version:

c2343aad

NA

NA

### Full Name:

Jaromir Hamala

### Affiliation:

QuestDB

- [x] Yes, I have

### Additional context

_No response_

brunocalza:
## Summary

Fixes <https://github.com/questdb/questdb/issues/6534>

The issue here is not the malformed query. The problem is the `AND` inside a sub-expression. Currently, the parser thinks that the `AND` inside the sub-expression refers to `BETWEEN`'s `AND` operator. 

There was a special treatment for `BETWEEN-CASE`, so the problem does not arise in 

```sql
SELECT x BETWEEN 0 AND CASE WHEN 1 AND 2 THEN 5 ELSE 10 END
```

but in other cases, such as

```sql
SELECT 1 BETWEEN (1 and 2) and 3;
SELECT 1 BETWEEN ARRAY[0 AND 1] AND 2 
SELECT 1 BETWEEN (ARRAY[10, 20, 30])[1 AND 1] AND 3
SELECT 1 BETWEEN CAST(0 AND 1 AS INT) AND 3
```

the issue arises. These are syntactically correct statements, and they should throw a type mismatch error.

## The fix

We keep track of the scope depth when we reach `BETWEEN` on a variable called `betweenStartScopeDepth`, and whenever we see an `AND` we use the current scope depth to know whether the `AND` refers to `BETWEEN` or not.

This fix is identical to the treatment given to the `BETWEEN-CASE`, but it's more generic. We had a `betweenStartCaseCount` variable that we removed.

## How this was tested

This was manually tested, and a unit test was added for each one of the scope cases.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/ExpressionParser.java
+++ b/core/src/main/java/io/questdb/griffin/ExpressionParser.java
@@ -1063,7 +1063,7 @@ public class ExpressionParser {
             int betweenAndCount = 0;
             int caseCount = 0;
             int argStackDepth = 0;
-            int betweenStartCaseCount = 0;
+            int betweenStartScopeDepth = 0;
             savedScopeStackBottom = scopeStack.getBottom();
             scopeStack.setBottom(scopeStack.sizeRaw());
             boolean parsedDeclaration = false;
@@ -1528,7 +1528,8 @@ public class ExpressionParser {
                                 processDefaultBranch = true;
                             }
                         } else if (SqlKeywords.isAndKeyword(tok)) {
-                            if (caseCount == betweenStartCaseCount && betweenCount > betweenAndCount) {
+                            if (scopeStack.size() == betweenStartScopeDepth
+                                    && betweenCount > betweenAndCount) {
                                 betweenAndCount++;
                                 thisBranch = BRANCH_BETWEEN_END;
                                 argStackDepth = popAndOpStack(listener, argStackDepth, prevBranch);
@@ -1603,7 +1604,7 @@ public class ExpressionParser {
                                 throw SqlException.$(lastPos, "between statements cannot be nested");
                             }
                             betweenCount++;
-                            betweenStartCaseCount = caseCount;
+                            betweenStartScopeDepth = scopeStack.size();
                         }
                         processDefaultBranch = true;
                         break;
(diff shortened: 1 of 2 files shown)


bluestreak01 (approved):
Fix is correct and minimal. Verified against existing BETWEEN tests in SqlParserTest, ExpressionParserTest (295), WhereClauseParserTest (842), and IntervalFilterTest (26) — all pass. The new `scopeStack.size()` check is a strict superset of the old `caseCount` check, so CASE behavior is preserved while PAREN/BRACKET/ARRAY/CAST scopes are now handled correctly.

Non-blocking suggestions:
- PR title is missing the verb per CLAUDE.md (`fix(sql): fix BETWEEN ...`).
- `Fixes <URL>` won't auto-close issue #6534; use `Fixes #6534`.
- Add `Bug` and `SQL` labels.
- Consider adding an execution-level test that asserts the resulting type-mismatch error, since the original bug surfaced as a runtime AssertionError.

PR #7047, opened 2026-04-30
fix(wal): correct write amplification and dedup counts in tables()

nwoolmer:
## Summary

- `tables().table_write_amp_p50/p90/p99/max` and `tables().wal_dedup_row_count_since_start` could report values much larger than the per-commit log line for the same workload -- sometimes by several orders of magnitude -- when a WAL apply job interleaved data writes with non-data transactions.
- Two `TableWriter` counters (`physicallyWrittenRowsSinceLastCommit`, `dedupRowsRemovedSinceLastCommit`) were only reset by `processWalBlock` (and one of them also by the non-WAL `commit()`), so iterations of `ApplyWal2TableJob.processWalCommit` that took non-resetting branches re-read the previous iteration's value when accumulating into `physicalRowsAdded`, `totalPhysicalRowCount`, and the dedup count passed to `recordWalProcessed`.
- The fix resets both counters at the start of every `processWalCommit` call so each iteration's reads only see that iteration's writes, regardless of branch.

## Affected branches in `processWalCommit`

Each of the following could leave the counter stale for the next iteration's read when preceded by a successful data write:

- `DATA` via `trySkipWalTransactions` (a future `REPLACE_RANGE` or `TRUNCATE` supersedes the current txn)
- `SQL` `UPDATE` that matches no rows (no internal `commit()` triggered)
- `TRUNCATE` (`removeAllPartitions` commits the txn file, not the writer)
- `MAT_VIEW_INVALIDATE`

The dedup mode of the skipped data txn (DEFAULT, UPSERT_NEW, NO_DEDUP) does not change exposure -- only the data branch matters. The `VIEW_DEFINITION` branch also lacks an internal reset, but views never carry `DATA` transactions in their WAL, so the bug pattern cannot form there in normal operation; the unconditional reset still covers it defensively.

## Side effects fixed by the same change

- `wal_apply_physically_written_rows` Prometheus counter no longer over-counts on skip / no-op iterations.
- `sys.telemetry_wal.physicalRowCount` is no longer attributed to skipped transactions.
- The per-job `ampl=` log line now matches what the iteration's writes actually produced.

## Tradeoffs

- The fix adds a single reset call per WAL apply iteration. Both counters are `LongAdder`s, so the reset is constant-time and runs on the WAL apply thread.
- The new public `TableWriter.resetWalApplyCounters()` increases the writer's API surface by one method, used only from `ApplyWal2TableJob`.

## Test plan

New cases in `RecentWriteTrackerIntegrationTest` that each exercise one affected branch:

- testWriteAmpNotInflatedByInterleavedUpdate -- UPDATE matching 0 rows
- testWriteAmpNotInflatedByMatViewInvalidate -- MAT_VIEW_INVALIDATE
- testWriteAmpNotInflatedBySkippedReplaceRange -- trySkipWalTransactions via REPLACE_RANGE
- testWriteAmpNotInflatedBySkippedTruncate -- trySkipWalTransactions via TRUNCATE, cross-drain pooled-writer leakage
- testDedupCountNotInflatedBySkippedReplaceRange -- dedup counter
- testWriteAmpPercentilesMatchPerCommitValues -- baseline contrast

All five regression cases reproduce the inflated value (p50=2.0 or dedup count = 2x) before the fix and resolve to p50=1.0 / correct count after.


nwoolmer:
@coderabbitai review

bluestreak01 (approved):
Approved. Five minor nits flagged in review (SQL string underscores in test literals, doc comment for testDedupCountNotInflatedBySkippedReplaceRange, unasserted count, "iter" abbreviation, optional javadoc on resetWalApplyCounters) — all non-blocking.

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/wal/ApplyWal2TableJob.java|1|1|100.00%|

PR #7048, opened 2026-04-30
fix(sql): keep JIT filter active for ::TYPE casts in WHERE clauses

nwoolmer:
## Summary

- Folds `cast(<inner>, T)` into the inner operand in JIT-compiled
  filters when T matches the column's full type, so the legacy
  `::uuid` workaround from PR #6413 (and any other literal
  `'...'::TYPE`) keeps using JIT instead of falling back to the
  Java filter.
- Three fold paths: literal constants (incl. negated), bind
  variables (when bound type matches), and no-op column-side casts.
- Fixes a latent sign-truncation in `serializeNumber` that emitted
  the negated MIN value of each integer width (BYTE_MIN, SHORT_MIN,
  INT_MIN, LONG_MIN) as the positive sentinel instead of the
  negative MIN.

## What is NOT covered

- Cross-type casts (e.g. `cast(int_col, long) = 100`) still bail
  out — the JIT IR has no cast opcode to perform runtime coercion,
  and folding would change semantics.
- PG's `::float` is rewritten by the parser to DOUBLE, so
  `afloat = 1.5::float` against a FLOAT column is treated as a
  cross-type cast and bails out. The unambiguous
  `cast(1.5 AS float)` form folds correctly.
- TIMESTAMP_MICRO and TIMESTAMP_NANO share the TIMESTAMP type tag
  but use different precision drivers — the guard compares full
  type, not tag, to prevent silent precision loss.
- STRING bind variables with `::symbol` cast on SYMBOL columns
  don't fold (they route through the deferred-symbol backfill
  path which the fold doesn't hook into).

## Test plan

- New \`CompiledFilterTest.testConstantCastPreservesJit\`,
  \`testColumnCastPreservesJit\`, \`testBindVariableCastPreservesJit\`
- New \`CompiledFilterCastFuzzTest\` with three probe shapes
  (matching, column-side, cross-type) across all supported numeric
  types including the negated MIN boundary
- Updated 8 fixtures in \`ExplainPlanTest.testSelectWithNonJittedFilter*\`
  — 6 flipped to \`Async JIT Filter\`, 2 kept as cross-type Java
- Verified across 4 random fuzz seeds and the full
  \`CompiledFilter*\`, \`JIT*\`, \`AsyncFiltered*\`, \`ExplainPlan*\`,
  \`ParallelFilter\` suites (833 tests, all green)

### DIFF
--- a/core/src/main/java/io/questdb/jit/CompiledFilterIRSerializer.java
+++ b/core/src/main/java/io/questdb/jit/CompiledFilterIRSerializer.java
@@ -52,6 +52,7 @@ import io.questdb.std.GenericLexer;
 import io.questdb.std.IntStack;
 import io.questdb.std.LongList;
 import io.questdb.std.LongObjHashMap;
+import io.questdb.std.Misc;
 import io.questdb.std.Mutable;
 import io.questdb.std.Numbers;
 import io.questdb.std.NumericException;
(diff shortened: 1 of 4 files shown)


nwoolmer:
@coderabbitai review

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 60 / 66 (90.91%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/jit/CompiledFilterIRSerializer.java|60|66|90.91%|

PR #7049, opened 2026-05-01
fix(sql): fix wrong results from JIT filter with UUID bind variables

nwoolmer:
## Summary
- A JIT-compiled filter using a UUID bind variable alongside any other bind variable produced wrong results — usually zero rows, sometimes spurious matches — because the JIT addressed bind-variable slots at an 8-byte stride while the Java side wrote 16 bytes for a UUID. With a UUID slot followed by another bind, the second 8 bytes of the UUID shadowed the next bind, which then decoded garbage.
- The bind-variable area now uses a fixed 16-byte stride end-to-end. UUID slots are unchanged; non-UUID slots get an 8-byte zero pad. Filters with a single UUID bind variable were already correctly addressed and continue to work.
- Trade-off: 8 wasted bytes per non-UUID bind variable in a per-query scratchpad. The hot-loop addressing stays a single immediate-offset load on both x86 and aarch64. The aarch64 immediate-form fast path now covers ``idx < 256`` instead of ``idx < 512``; past that the codegen falls back to the existing register-offset path, so behaviour at large bind-variable counts is unchanged in semantics.
- Why this design over a per-bind offset table: the indirection saves a handful of bytes in a memory area that is already small, at the cost of either a memory load per bind read in the JIT hot loop or a second invariant the Java and native sides have to agree on. Fixed stride keeps the agreement to a single number and matches what the codegen already does for column reads.

## Test plan
- Added ``CompiledFilterTest.testUuidBindFollowedByOtherBinds`` exercising UUID + SYMBOL + two TIMESTAMP binds against a UUID-keyed table; previously returned zero rows, now returns the expected row.
- Ran the full ``CompiledFilterTest`` (30 tests) and the broader ``Jit*`` / ``Uuid*`` / ``AsyncFiltered*`` / ``ParquetLateMaterializationFuzzTest`` / ``FilterParser*`` set — 396 tests total — all green with the rebuilt native library.
- Verified end-to-end against the originally reported failure (sqlx 0.8 binding ``Uuid`` to a ``product UUID`` column on a WAL table with DEDUP keys): with the fix, ``.bind(uuid)`` returns the matching row instead of zero rows; the workaround ``.bind(uuid.to_string())`` continues to work as well.
- Native library needs to be rebuilt for every supported platform (linux-x86-64, linux-aarch64, darwin-x86-64, darwin-aarch64, windows-x86-64). I rebuilt linux-x86-64 locally for verification; the in-repo binaries under ``core/src/main/resources/io/questdb/bin/`` are unchanged in this commit and will need regenerating in the release pipeline.


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 17 / 18 (94.44%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncFilterUtils.java|17|18|94.44%|

jerrinot (approved):

PR #6861, opened 2026-03-09
feat(core): add compact, high-performance posting index for symbol columns

bluestreak01:
## Summary

Adds the posting index, a new index format for symbol columns, and covering index support that stores selected column values in sidecar files alongside the posting list.

### SQL syntax

```sql
-- Inline
CREATE TABLE t (
    ts TIMESTAMP,
    sym SYMBOL INDEX TYPE POSTING INCLUDE (price, qty),
    price DOUBLE,
    qty INT
) TIMESTAMP(ts) PARTITION BY DAY;

-- Out-of-line
CREATE TABLE t (sym SYMBOL, ts TIMESTAMP),
    INDEX(sym TYPE POSTING INCLUDE (price))
    TIMESTAMP(ts) PARTITION BY DAY;

-- ALTER TABLE
ALTER TABLE t ALTER COLUMN sym ADD INDEX TYPE POSTING INCLUDE (price, qty);

-- Encoding variants (default is adaptive)
CREATE TABLE t (sym SYMBOL INDEX TYPE POSTING DELTA, ts TIMESTAMP) ...;
CREATE TABLE t (sym SYMBOL INDEX TYPE POSTING EF, ts TIMESTAMP) ...;
```

Configuration properties:
- `cairo.posting.index.row.id.encoding` — `adaptive` (default), `ef`, or `delta`
- `cairo.posting.index.auto.include.timestamp` — `true` (default): the designated timestamp column is auto-appended to the INCLUDE list

### Core index engine

`PostingIndexWriter` / `PostingIndexFwdReader` / `PostingIndexBwdReader` implement the read/write path with concurrent reader support via a double-buffered seqlock protocol in the `.pk` key file.

Row IDs are encoded per-key using delta + Frame-of-Reference (FoR64) bitpacking. At seal time the writer groups keys into strides of 256 and trial-encodes each stride in two modes:

- **Delta mode**: per-key delta encoding, then per-block FoR bitpacking. Wins when values/key >= ~10 — per-key delta distributions produce low bitwidths.
- **Flat mode**: stride-wide FoR with a single base and bitwidth, prefix-count array for per-key slicing. Wins when values/key <= ~8 — eliminates per-key metadata overhead.

The writer picks whichever mode produces fewer bytes per stride. When a flat-mode bitwidth lands on 8, 16, or 32, `bp_bitpack.cpp` decodes via native AVX2 (Lemire-style vpshufb + vpsrlvd extraction). A Java scalar fallback handles all other widths and platforms without AVX2.

[...]

PR #7051, opened 2026-05-02
chore(core): seal POSTING indexes after WAL fast-lag commit

bluestreak01:
## Summary

The fast-lag commit path (`applyLagToLastPartition`) used by WAL apply when transactions land entirely inside the open last partition calls `updateIndexesParallel` to populate the in-memory state of each column indexer, but it does not seal the result. POSTING writers hold their entries in memory until `seal()` flushes them to the `.pv` file; BITMAP writers persist incrementally so they were unaffected. The O3 commit path catches every POSTING column it touches via `sealPostingIndexesForO3Partitions`, but the fast-lag path bypasses `processO3Block` entirely.

The visible effect: partitions whose data arrives exclusively through the fast-lag path - typically the very first partition of an empty WAL table - end up with an empty `.pv` for any non-covering POSTING column. Indexed queries (`WHERE sym = 'X'`) over those partitions return zero rows, while full scans return the data correctly. Tables that fall through to a full O3 commit on every transaction are unaffected.

This PR adds `sealPostingIndexesForLastPartitionFastLag()`, which iterates indexed POSTING columns on the last partition and runs the non-covering seal sequence (`setNextTxnAtSeal`, `seal`, `publishPendingPurges`). It is invoked from `applyLagToLastPartition` immediately after `updateIndexesParallel`. Covering POSTING indexes are intentionally skipped: they require the full reseal in `sealPostingIndexForPartition` (covering-file remap plus `rebuildSidecars`), which the next O3 commit catches; wiring covering reseal into the fast-lag path is left for a follow-up so it can be exercised under its own focused tests.

## Tradeoffs

- Each fast-lag commit on a table with non-covering POSTING indexes now does an extra `seal()` plus `publishPendingPurges()` per indexed column. For tables that mostly batch through fast-lag, this adds I/O proportional to the number of indexed columns per commit. Tables with no POSTING indexes (BITMAP-only or no indexes) take an early-return on the column-type check and pay nothing measurable.
- Covering POSTING indexes still go through the existing O3-driven seal path. If a table has only covering POSTING indexes and stays exclusively on the fast-lag path for many commits, the seal happens later than it does for non-covering columns. Behavior is not worse than today; it is the unchanged baseline.
- The bug exists only on the unreleased `master` introduced by #6861, hence `chore` rather than `fix`.

## Test plan

- `WalWriterFuzzTest.testWalSmallWalFdReuse` with seeds `7012496711609388L,1777753420105L` and `7012496713968699L,1777753420107L` deterministically failed before this change with row mismatches in `assertRandomIndexes`; passes after.
- Full `WalWriterFuzzTest` class (24 tests, all 4 symbol-index flavours: BITMAP, POSTING, POSTING DELTA, POSTING EF) passes with random seeds.
- Wider sweep across `*FuzzTest`, `WalTableSqlTest`, `WalAlterTableSqlTest`: 15,609 tests, 0 failures.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -3681,6 +3681,7 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
                 }
             }
             updateIndexesParallel(initialTransientRowCount, newTransientRowCount);
+            sealPostingIndexesForLastPartitionFastLag();
         }
         // set append position on columns so that the files are truncated to the correct size
         // if the partition is closed after the commit.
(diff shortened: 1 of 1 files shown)


questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=230738)
**Tested Commit:** `02b5fc879e3823445398085cb6286583aa3207c8`

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 16 / 18 (88.89%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|16|18|88.89%|

PR #7052, opened 2026-05-02
fix(core): refresh covering sidecar mappings on posting-index reader reload

bluestreak01:
## Summary

- Long-lived posting-index readers (e.g. those reused by `TableReader` across queries) call `reloadConditionally()` to track writer publishes. The reader resized `.pv` to the new `VALUE_MEM_SIZE` from the picked chain entry but never touched the covering `.pcN` mappings. After a same-`sealTxn` gen flush extended `.pcN` past the reader's pre-allocated extent, covered reads of the new gen dereferenced past the mapping. After a `sealTxn` advance, the reader kept reading the old `sealTxn`-suffixed sidecar file with offsets from the new chain entry. The first case risks `SIGBUS`; the second silently returns wrong data.
- This PR publishes each `.pcN`'s authoritative valid-byte extent in the chain entry itself, in a per-cover-column footer that lives right after the gen-dir region. The writer fills the footer in `extendHead` (in-place, ordered before the `GEN_COUNT` store) and `appendNewEntry` (alongside the rest of the header). The reader picks the footer under the same chain seqlock and sizes its sidecar mappings from the published value rather than `ff.length(fd)`.
- The footer length is derived from `coverCount`, which is fixed for the `.pk` file's posting column instance and lives in the `.pci` sidecar header. Because `coverCount` is constant for every entry in a `.pk`, no `coveringFormat` flip or format-version bump is required. Per-entry overhead is `coverCount * 8` bytes (16 bytes for the typical `coverCount=2` — about 0.4% on a full 4068-byte entry). No change for entries with no cover columns.

### Reader-side flow changes

- `of()` now opens `.pci` before the chain pick so `coverCount` is known when reading the entry's footer.
- `reloadConditionally()` drops sidecar mappings on a `sealTxn` advance (so the next `ensureSidecarOpen` lazy-maps the new `sealTxn` file) and resizes them in place to the new published end-offset on same-`sealTxn` growth.
- `ensureSidecarOpen()` maps to the chain-published extent rather than `ff.length(fd)`, so the mapping is tight to actual data even when the writer pre-allocates pages past it.

### Considerations / tradeoffs

- 8 bytes per cover column added per chain entry. Negligible in absolute terms; chain entries are GCed once no reader is pinned on them.
- The chain reader (`PostingIndexChainPicker.pick`) and the entry codec (`PostingIndexChainEntry.read/writeHeader`) gained an explicit `coverCount` parameter. Backward-compat overloads delegate to `coverCount=0` so existing chain-writer recovery paths and unit-test fixtures compile unchanged.
- The chain writer's `appendNewEntry` / `extendHead` now also take a `LongList coverEndOffsets`; same backward-compat overloads cover the no-cover call sites.

## Test plan

- New regression test `CoveringIndexTest#testCoveringReloadDoesNotRefreshStaleSidecarMmap`. It overrides `cairo.writer.data.index.value.append.page.size` down to the OS page size so two same-`sealTxn` gens force the writer to extend the file beyond gen-0's mmap, then reads every row across both gens through one long-lived reader. Without this fix the post-reload reads dereference past the stale mapping (SIGBUS, or zero-padded data depending on FS / page state). With the fix every covered value round-trips and the post-reload sidecar mmap covers the new extent.
- Full posting-index + covering-index suites: `Tests run: 1479, Failures: 0, Errors: 0, Skipped: 1` (the one skip is pre-existing).

### DIFF
--- a/core/src/main/java/io/questdb/cairo/idx/AbstractPostingIndexReader.java
+++ b/core/src/main/java/io/questdb/cairo/idx/AbstractPostingIndexReader.java
@@ -72,6 +72,11 @@ public abstract class AbstractPostingIndexReader implements IndexReader {
     protected final IntList sidecarColumnIndices = new IntList();
     protected final IntList sidecarColumnTypes = new IntList();
     protected final LongList sidecarCovTs = new LongList();
+    // Per-cover-column file extents picked from the chain entry. Slot c is
+    // the writer's published append offset in .pc{c}.<...>.<sealTxn> at the
+    // moment the picked entry was (re)published. Sized by openSidecarFilesIfPresent
+    // to coverCount; refreshed on each readIndexMetadataFromChain() success.
+    protected final LongList sidecarFileEndOffsets = new LongList();
     protected final ObjList<MemoryMR> sidecarMems = new ObjList<>();
     protected final MemoryMR valueMem = Vm.getCMRInstance();
     protected long columnTop;
@@ -257,6 +262,12 @@ public abstract class AbstractPostingIndexReader implements IndexReader {
                     -1
             );
 
+            // Discover coverCount from .pci before reading the chain entry
+            // — the entry's cover end-offset footer length is sized by
+            // coverCount, and PostingIndexChainPicker.pick() needs it to
+            // populate the snapshot's coverFileEndOffsets list.
+            openSidecarFilesIfPresent(path.trimTo(pLen), columnName, columnNameTxn);
+
             readIndexMetadataFromChain();
 
             if (headEntryOffset == PostingIndexUtils.V2_NO_HEAD || valueMemSize <= 0) {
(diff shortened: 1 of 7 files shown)


questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=230743)
**Tested Commit:** `5b4e3c45273ca57b947a05ee05efaa769c10f0e6`

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=230754)
**Tested Commit:** `fbe72555707c7c1cdfce84b67ca5fa21aaad64cf`

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 81 / 87 (93.10%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/AbstractPostingIndexReader.java|24|29|82.76%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexChainEntry.java|23|24|95.83%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|15|15|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexChainWriter.java|17|17|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexChainPicker.java|2|2|100.00%|

PR #7055, opened 2026-05-05
chore(core): add level switch to review-pr skill

bluestreak01:
## Summary

The `review-pr` skill currently runs the full mission-critical pass on every invocation — 10 parallel review agents, repo-wide callsite inventory, and per-finding verification. The token cost is high regardless of PR scope, so casual reviews end up paying for blast-radius analysis they do not need.

This PR adds a level token to the skill so the caller can pick the depth:

- `--level=0` (default): single-pass inline review of the diff. No agent spawn, no change surface map, no verification step.
- `--level=1`: adds the semantic-delta map plus three light agents — correctness, tests, code quality.
- `--level=2`: full change surface map limited to `public`/`protected` callsites, agents 1-7, batched verification.
- `--level=3`: the previous behavior — all 10 agents, per-finding verification, full callsite inventory.

Levels accept `--level=N`, `-lN`, or a bare digit (e.g., `/review-pr 1234 2`). The skill announces the chosen level at the start of every review and, when the level was defaulted, reminds the user that level 3 exists.

## Tradeoffs

- Level 0 will miss bugs at out-of-diff callsites that level 3 catches by design. The chosen level is announced up front so the depth/coverage tradeoff is explicit at call time, but a level-0 review is materially less thorough than the prior default and the user has to opt into the deeper passes.
- The level gating lives in a single section near the top of the skill rather than as inline annotations on each step. The existing step structure stays unchanged, but the gate logic is centralized and must be cross-referenced with the step it controls.
- The level mapping is heuristic. Cost ratios across levels are not measured; level 2 may still be expensive on PRs that touch widely-used symbols because Step 2.5b grows with the public/protected callsite count.

## Test plan

- Run `/review-pr <PR>` with no level and confirm the skill announces level 0 and skips agent spawn
- Run `/review-pr <PR> --level=2` and confirm agents 1-7 are launched and agents 9 and 10 are skipped
- Run `/review-pr <PR> 3` (bare digit) and confirm full pass behavior is unchanged
- Run `/review-pr <PR> -l1` and confirm only agents 1, 5, 6 are launched

### DIFF
--- a/.claude/skills/review-pr/SKILL.md
+++ b/.claude/skills/review-pr/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: review-pr
 description: Review a GitHub pull request against QuestDB coding standards
-argument-hint: [PR number or URL]
+argument-hint: [PR number or URL] [--level=0..3]
 allowed-tools: Bash(gh *), Read, Grep, Glob, Agent
 ---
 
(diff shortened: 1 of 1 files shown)


.claude/skills/review-pr/SKILL.md
```diff
@@ -28,6 +28,21 @@ You are a senior QuestDB engineer performing a blocking code review. QuestDB is
   that should never occur in a non-corrupt database. Only flag an `assert` if the condition can plausibly be triggered
   by normal (non-corrupt) user operations.
 
+## Review level
+
+Parse `$ARGUMENTS` for a level token: `--level=N`, `-lN`, or a bare single digit `0`-`3`. **If no level is given, default to 0.** Strip the level token before feeding the remainder (PR number or URL) to `gh` commands.
```
puzpuzpuz, line 33: ## Moderate

### 1. Step 1's `gh` examples still use raw `$ARGUMENTS` — contradicts the new "strip the level token" rule

The new section (lines 31–45) says:

> Strip the level token before feeding the remainder (PR number or URL) to `gh` commands.

But Step 1 (lines 35–39, unchanged) still reads:

```bash
gh pr view $ARGUMENTS --json number,title,body,labels,state
gh pr diff $ARGUMENTS
gh pr view $ARGUMENTS --comments
```

`$ARGUMENTS` is substituted by the slash-command runtime *before* the model parses anything. If the user types `/review-pr 7055 --level=2`, Step 1 renders as `gh pr view 7055 --level=2 --json ...` — `gh` rejects `--level=2` as an unknown flag and Step 1 fails on every non-default invocation.

The skill relies on the model reading the new textual rule and silently rewriting the bash block, which is exactly the kind of cross-section coordination that fails in practice. Fix one of two ways:

- Replace `$ARGUMENTS` in lines 36–38 with a placeholder like `<PR>` and add a one-liner: "substitute `<PR>` with the PR identifier obtained after stripping the level token from `$ARGUMENTS`".
- Or, more robust: have the model assign the stripped identifier to a shell variable in a preceding bash call (`PR=...`) and reference `$PR` in the gh commands.

puzpuzpuz (commented):
## Moderate

### 2. Bare-digit form is positionally ambiguous and the rule is underspecified

The PR body shows the bare-digit form as `/review-pr <PR> 3` (level is the second positional arg), but the skill text only says "a bare single digit `0`-`3`" without pinning the position. As written, `/review-pr 3` is ambiguous: PR #3 or level 3 with no PR? `argument-hint` doesn't list the bare-digit form either, so users won't see it advertised. State the rule explicitly: "the bare digit is recognized only as the second whitespace-separated token; `/review-pr 3` is treated as PR identifier `3`, not a level."

## Minor

### 3. `argument-hint` only advertises `--level=0..3`

The `-lN` and bare-digit forms are documented in the body but not in the hint. Either add them or drop them from the body — pick one surface for the syntax.

### 4. Level 0 still references "Step 4" but Step 4's "Downgraded (false positives)" subsection presumes Step 3b ran

Level 0 skips 3b ("verify each finding inline as you write it"), so there are no formally-classified false positives to list. Either drop the Downgraded section for level 0 or tell the model to inline-merge inline-dismissed findings into Downgraded so the output schema stays uniform.

puzpuzpuz:
Please treat all comments as nit except for https://github.com/questdb/questdb/pull/7055#discussion_r3186754861 (this one is probably worth addressing).

ideoma (dismissed):

puzpuzpuz (approved):

PR #7058, opened 2026-05-05
feat(core): auto-seal covering posting index past gen threshold

nwoolmer:
## Summary

`sealPostingIndexesForLastPartitionFastLag` exempts covering POSTING from the per-commit seal because covering reseal is more expensive than non-covering (sidecar rebuild + covering-file remap). The exemption defers to the next O3 commit. For pure in-order WAL ingestion into a single partition (the common forward-only ILP path), no O3 commit ever runs, so covering gens accumulate one per commit until the writer closes.

This adds a threshold check using the existing `cairo.posting.seal.gen.threshold` config so covering reseal kicks in proactively once unsealed gen count crosses the threshold.

## What changed

- `TableWriter.sealPostingIndexesForLastPartitionFastLag` calls `sealIfMultiGen(threshold)` for covering instead of unconditionally skipping. Below the threshold the behaviour is unchanged (still defers to O3); at or above, covering reseals on the fast-lag commit.
- The threshold key is reused — same value, same default — so a deployment already setting it sees consistent behaviour at both the partition-rollover and fast-lag callsites. No new config to document.
- `@TestOnly` accessor `AbstractPostingIndexReader.getGenCount()` so tests can observe gen-count cycling.
- `PostingSealGenThresholdTest` covers three settings: 0 (seal every commit), 4 (cycle, cap at 5), Integer.MAX_VALUE (no auto-seal).

## Numbers

Synthetic harness, 10K-key sweep, in-order commits to one partition, scan throughput vs sealed layout:

| gen count | scan ops/s | scan vs sealed | point ops/s | point vs sealed |
|---:|---:|---:|---:|---:|
| 1 | 504 | 100% | 688 | 100% |
| 8 | 188 | **55%** | 455 | 68% |
| 16 | 170 | 49% | 415 | 60% |
| 32 | 107 | 34% | 276 | 43% |
| 100 | 59 | 17% | 133 | 19% |

The default threshold of 16 keeps scan within 2x of sealed and point reads within ~1.7x. Lower thresholds tighten the read window at write-throughput cost — rough numbers from a 200-commit / 400K-row covering write loop:

| seal frequency | covering commits/s | vs no-auto-seal |
|---|---:|---:|
| every commit | 89 | 0.03x |
| every 8 | 680 | 0.23x |
| every 16 | 1,415 | **0.48x** |
| every 32 | 1,990 | 0.67x |
| every 64 | 2,893 | 0.97x |
| no auto-seal | 2,973 | 1.00x |

So 16 sits at the inflection point where doubling the threshold buys ~2x writes for ~1.2x worse reads, and halving it costs ~2x writes for marginal read gains.

## Tradeoffs

- Pure-O3 workloads see no behavioural change (their covering reseal already runs via `sealPostingIndexesForO3Partitions`).
- Forward-only ILP into a single partition: covering write throughput drops to roughly half of the no-auto-seal baseline, in exchange for read latency staying bounded. The previous behaviour was effectively unbounded gen accumulation until writer close — a state most users would have hit on hot-shard ingest.
- Threshold is configurable; setting it to `Integer.MAX_VALUE` restores the prior behaviour exactly.
- Non-covering POSTING is unchanged. Fast-lag has always sealed it unconditionally on the last partition, and that path is preserved.

## Test plan

- mvn test --batch-mode -pl questdb/core -Dtest.include="%regex[.*PostingSealGenThresholdTest.*]"
- mvn test --batch-mode -pl questdb/core -Dtest.include="%regex[.*CoveringIndexTest.*]" (the three FSST failures observed on master also fail without this patch — pre-existing, unrelated)
- Spot-check `cairo.posting.seal.gen.threshold` set to a low value with a covering POSTING table under in-order WAL ingestion: confirm gen count stays bounded across many commits.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -10916,17 +10916,20 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
                 continue;
             }
             IntList coveringCols = metadata.getColumnMetadata(colIdx).getCoveringColumnIndices();
-            if (coveringCols != null && coveringCols.size() > 0) {
-                // Covering POSTING indexes need the full reseal in
-                // sealPostingIndexForPartition (covering-file remap +
-                // rebuildSidecars). The fast-lag commit path is followed by
-                // a full O3 commit when more transactions accumulate, which
-                // catches the covering reseal then. Non-covering POSTING is
-                // the only path that has no later sweep.
-                continue;
-            }
+            boolean isCovering = coveringCols != null && coveringCols.size() > 0;
             indexer.getWriter().setNextTxnAtSeal(txWriter.getTxn());
-            indexer.seal();
+            if (isCovering) {
+                // Covering POSTING reseal is expensive (sidecar rebuild via
+                // sealPostingIndexForPartition with covering-file remap), so
+                // it normally waits for the next O3 commit. With pure in-order
+                // ingestion (no O3, no partition rollover), gens accumulate
+                // without bound and per-key cursor probes turn into walks
+                // over every gen. The threshold caps unsealed gen count;
+                // sealIfMultiGen is a no-op below the threshold.
+                indexer.getWriter().sealIfMultiGen(configuration.getPostingSealGenThreshold());
+            } else {
+                indexer.seal();
+            }
             indexer.publishPendingPurges(messageBus, tableToken, partitionBy, timestampType, txWriter.getTxn());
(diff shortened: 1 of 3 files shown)


nwoolmer:
Consolidating into #7056 — same change, simpler review path.

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/AbstractPostingIndexReader.java|1|1|100.00%|

PR #7063, opened 2026-05-06
chore(core): fix yet another posting index bug found by fuzz

bluestreak01:
## Summary

When `TableWriter.createIndexFiles` runs against a partition that already
has a `.pk` posting-index key file, it used to wipe and reinitialise the
file. That reset `genCounter` to `-1`, so the writer's next
`appendNewEntry` reused `sealTxn=0` and recreated `.pv.0` at a path that
an outstanding orphan-purge for `sealTxn=0` would then delete out from
under the live writer. The resulting failure mode is a runtime
`could not open, file does not exist: .../sym2.pv.0` from the next
reader, or — when the purge races with the writer rather than killing it
outright — a silent row mismatch between WAL and non-WAL views of the
same data.

This PR makes the POSTING branch of `createIndexFiles` preserve an
existing `.pk` whose seqlock header is in a published, stable state, so
its chain history survives writer reopen. The `.pk` only gets wiped when
its header is missing, truncated, or zeroed — leftovers from an earlier
init that crashed before publishing Page A and that carry no chain
history worth preserving. Validation lives in a new
`PostingIndexUtils.hasInitialisedKeyFileHeader` helper.

## Tradeoffs

- POSTING now ignores the `force` flag in `createIndexFiles`. Callers
  that pass `force=true` to recreate an index from scratch will instead
  reuse the existing `.pk` if it has a valid header. This is by design:
  `genCounter` is a high-water mark for orphan purges and must not
  regress, so wiping a valid `.pk` is unsafe regardless of intent.
- The validity check opens the file read-only and reads four 8-byte
  fields. That's one extra `openRO`/`close` per indexed column per
  partition open compared to the previous bare `ff.exists` check.
  Negligible against partition-open cost.

## Verification

Fuzz reproducers (frozen seeds) run locally against this branch:

- `WalWriterFuzzTest.testWalWriteRollbackTruncateHeavy` — seeds
  `4160224326802423175L, 630096581114372683L` (failure originally
  manifested as `could not open, file does not exist:
  .../sym2.pv.0` from `PostingIndexFwdReader`).
- `com.questdb.cairo.fuzz.SequencerV2ToV1MigrationFuzzTest.testMigrationWithSchemaChanges`
  — seeds `7259504071682373373L, 6177862713324340969L` (failure
  originally manifested as a row-mismatch in
  `FuzzRunner.checkIndexRandomValueScan`). This test lives in
  questdb-enterprise and depends on parquet APIs that have moved on
  this branch; it was not re-run end-to-end locally. The architectural
  invariant the fix relies on — preserve any `.pk` written by a
  successful `initKeyMemory`, wipe only partial/zeroed leftovers — is
  identical to what the original orphan-purge race needed, so the same
  fuzz failure mode remains covered.

Test plan:

- `mvn -pl core -Dtest=FullFwdPartitionFrameCursorTest test` — 165 pass
  (the three `testIndexFailInConstructorByNoneEmpty{1,2,3}k` cases that
  regressed when the early version of this fix preserved any existing
  `.pk` now go through the header-validity path correctly).
- `mvn -pl core -Dtest='PostingSealPurgeTest,PostingIndexChainTest,PostingIndexChainWriterTest,IndexFactoryTest,IndexBuilderTest,IndexTypeTest,CoveringIndexTest' test`
  — 433 pass.
- Full Cairo and Griffin CI on Azure Pipelines.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -4816,7 +4816,21 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
         try {
             keyFileName(indexType, path.trimTo(plen), columnName, columnNameTxn);
 
-            if (!force && ff.exists(path.$())) {
+            if (IndexType.isPosting(indexType)) {
+                // POSTING preserves chain history across the partition's
+                // lifecycle: the genCounter in the .pk header is the
+                // high-water sealTxn that pending orphan purges may target.
+                // Wiping and reinitialising the .pk would reset genCounter to
+                // -1, so the writer's next appendNewEntry would reuse
+                // sealTxn=0 and recreate .pv.0 at a path that an outstanding
+                // orphan purge for sealTxn=0 would then delete out from under
+                // the live writer. Skip the recreate when the .pk already
+                // exists; the writer's recovery walk drops abandoned chain
+                // entries on reopen, which is the right cleanup for POSTING.
+                if (ff.exists(path.$())) {
+                    return;
+                }
+            } else if (!force && ff.exists(path.$())) {
                 return;
             } else {
                 ff.removeQuiet(path.$());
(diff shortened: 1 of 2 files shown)


mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexUtils.java|13|13|100.00%|

PR #7064, opened 2026-05-06
fix(core): fix rare WAL apply live lock when transactions are committed at a very high rate

ideoma:
Fix rare live lock when WAL apply thread spin in:

```
  'wal-apply_1': RUNNABLE                                                                                                                                                                                                                                                        
      at io.questdb.cairo.wal.WalTxnDetails.readObservableTxnMeta(WalTxnDetails.java:530)                                                                                                                                                                                        
      at io.questdb.cairo.TableWriter.readWalTxnDetails(TableWriter.java:2643)                                                                                                                                                                                                   
      at io.questdb.cairo.wal.ApplyWal2TableJob.applyOutstandingWalTransactions(ApplyWal2TableJob.java:370)                                                                                                                                                                      
      at io.questdb.cairo.wal.ApplyWal2TableJob.applyWal(ApplyWal2TableJob.java:935)                                                                                                                                                                                             
      at io.questdb.cairo.wal.ApplyWal2TableJob.doRun(ApplyWal2TableJob.java:989)                                                                                                                                                                                                
                                                                                     
```

This may happen in some edge-case commit pattern where transactions are added very quickly, and reading the transactions metadata loop always finds new entries.


bluestreak01:
@ideoma — review notes from a level-2 pass over this PR. The core fix is correct (the `do…while` always runs at least one iteration, so each `readWalTxnDetails(cursor, timeLimit)` call still loads at least one batch even when the deadline has elapsed; `loadFromSeqTxn = appliedSeqTxn + 1` keeps the assertions in `ApplyWal2TableJob.java:481-482` reachable-safe; `TableTransactionLogV1.TransactionLogCursorImpl.hasNext()` re-maps on new txns, which is the mechanism the original loop got stuck on). Concerns below.

### Moderate

**1. Lookahead now competes with apply work for the same time budget — undocumented behavior change.**
Old: `final long timeLimit = microClock.getTicks() + tableTimeQuotaMicros;` was declared *after* `readWalTxnDetails(cursor)`. The pre-read had unbounded time, and the main apply loop got a fresh `tableTimeQuotaMicros` budget.
New: `timeLimit` is declared *before* the pre-read (`ApplyWal2TableJob.java:365`). Lookahead and apply share one budget. At the default 1 s quota this is invisible; under tight quotas or slow storage it directly reduces how many txns the same apply pass can commit. Worth calling out in the PR body so operators tuning `cairo.wal.apply.table.time.quota` understand the new accounting.

**2. Bounded lookahead can force extra O3 commits — not acknowledged in the PR.**
`WalTxnDetails.readObservableTxnMeta` back-propagates min timestamps across the loaded prefix to set `commit_to_timestamp` (lines 562–589). When the deadline cuts the load short, the loaded txns never see the min timestamps of the unloaded tail. Those tail txns may have lower min timestamps that would have lowered earlier `commit_to_timestamp` values and avoided O3. Intended tradeoff (don't deadlock, accept some O3), but worth surfacing for anyone later debugging unexpected O3 commits.

**3. Missing regression test for the actual live-lock scenario.**
Both new tests verify *mechanics* (the deadline halts the loop; a 0-quota apply still drains). Neither reproduces the failure mode in the PR description: commits arriving fast enough that `cursor.getMaxTxn()` grows on every `hasNext()`. A test that drives that condition (producer thread inserting in a tight loop while the apply thread pulls) would have caught the original bug and would catch a regression if the deadline plumbing is later refactored away. `WalTxnDetailsFuzzTest` is a more natural home than `WalWriterTest`.

### Minor

**4. Test name `testTruncateLookaheadDeadlineApplyMakesProgress` is misleading.** The test body has no TRUNCATE — it inserts 20 rows and drains with `tableTimeQuotaMicros=0`. The \"Truncate\" prefix looks pasted from the alphabetical neighbor `testTruncateFollowedByTwoInserts`. Rename to e.g. `testApplyMakesProgressWithZeroTimeQuota`; this also moves it to its correct alphabetical slot.

**5. `WalTxnDetailsFuzzTest` callsites all use the 1-arg form.** Seven call sites, none exercising the deadline branch. The deadline path is the only new behavior — it should be fuzzed. A coin-flip between `Long.MAX_VALUE` and `microClock.getTicks() + jitter` per call would cover the partial-load path the deterministic frozen-clock test can't reach (deadline elapsing mid-iteration of the do-while).

**6. PR title English.** `\"when transactions are committed in very high rate\"` reads awkwardly — `\"at a very high rate\"` is the correct preposition. This goes into release notes verbatim.

**7. Enterprise CI is failing** (commit `de4da4299470336dfe1b01c8f8e753a435305c72`, per the bot comment above). Worth resolving before merge.

**8. `config.getMicrosecondClock()` is fetched on every `readObservableTxnMeta` call** (`WalTxnDetails.java:539`). Not a hot path, but `WalTxnDetails` already holds `config` as a final field, and `ApplyWal2TableJob` caches `microClock` as a final field at construction. Cosmetic; matches existing pattern.

### Considered and dismissed

- *Re-read inside apply main loop might not load the next `seqTxn` when deadline is expired* — `loadFromSeqTxn = appliedSeqTxn + 1` and the always-executed first iteration loads a batch starting exactly at the needed `seqTxn`.
- *`microClock.getTicks() + tableTimeQuotaMicros` overflow* — worst case ~`1.78e15`, far below `Long.MAX_VALUE`.
- *Single-arg overload's `Long.MAX_VALUE` breaks the comparison* — `clock.getTicks()` is always well below `Long.MAX_VALUE`; legacy behavior preserved.
- *Member ordering in `TableWriter`* — placing the shorter-signature overload first matches the file's existing convention.

Verdict: needs minor discussion — fix is correct, but the shared-budget and extra-O3 tradeoffs deserve a line in the PR body, the live-lock regression test would harden the fix, and enterprise CI needs to be green.

ideoma:
> @ideoma — review notes from a level-2 pass over this PR. The core fix is correct (the `do…while` always runs at least one iteration, so each `readWalTxnDetails(cursor, timeLimit)` call still loads at least one batch even when the deadline has elapsed; `loadFromSeqTxn = appliedSeqTxn + 1` keeps the assertions in `ApplyWal2TableJob.java:481-482` reachable-safe; `TableTransactionLogV1.TransactionLogCursorImpl.hasNext()` re-maps on new txns, which is the mechanism the original loop got stuck on). Concerns below.

hey @bluestreak01, here are my replies. All major points are non-issues, I fixed the minors and merged the latest


### Moderate #1 — shared time budget / lookahead competing with apply

  One do-while iteration is bounded, not the time budget. `WalTxnDetails.readObservableTxnMeta` always executes at least one iteration that loads up to `txnLoadCount = config.getWalApplyLookAheadTransactionCount()` transactions (default 200, via `PropServerConfiguration.java:901`). The deadline check sits *between* iterations, so a single iteration always completes.

  Within one iteration, `loadTransactionDetails` calls `loadTxns` which radix-sorts the batch by `(walId, segmentId)`, and the inner loop opens each segment's `_event` file exactly once per batch (reusing `walEventCursor` while `lastWalId == walId && segmentId == lastSegmentId`). So 100s of segments per batch is fine.

  The real exit condition is `rowsToLoad > 0`, i.e. the row budget `maxLookaheadRows = getWalMaxLagRows()` (`TableWriter.java:6218`). The deadline can only bite when many iterations are needed because each one loads very few rows — i.e. lots of very small segments where `maxLookaheadRows` is never reached. That is a degenerate shape with no practical workload behind it today, which is why the budget is shared.

  ### Moderate #2 — extra O3 commits

  The only direct change is "load fewer transactions when the deadline fires." Back-propagation then runs over a shorter buffer — the unloaded tail's `WAL_TXN_MIN_TIMESTAMP_OFFSET` values never feed into `runningMinTimestamp`, which can leave some `commit_to_timestamp` slots
  higher than a fully-loaded pass would produce. That's the mechanism behind extra O3.

  But this is the lookahead being *less effective*, not a regression: in the only shape where the deadline actually fires (row budget never reached → many tiny segments), the pre-fix behavior was the live-lock. In any workload where `maxLookaheadRows` is hit first (the normal
  case), `clock.getTicks() < deadlineMicros` is still true at exit and back-propagation is identical to before. Will add a sentence to the PR body for completeness.

  ### Moderate #3 — missing regression test for the live-lock

  Reproducing the actual live-lock deterministically would need the producer to outpace the consumer indefinitely on every test machine — that's a timing race we can't pin down. The condition is `getLastSeqTxn() < cursor.getMaxTxn()` staying true forever, which requires the
  producer thread to add at least one new txn into the sequencer log between each `loadTxns` call on the apply side. That ratio depends on CPU count, scheduler, page cache, disk speed, and JIT warmup — on a fast CI box the consumer catches up and the test passes without
  exercising the bug;  and at the end, the new code will not fall into the live lock, trivial to deduce from the code.

  The two added tests cover what is actually testable: the deadline halts the do-while (`testReadWalTxnDetailsBoundedByDeadline`), and the apply pipeline still drains end-to-end when both budgets are immediately exhausted (`testApplyMakesProgressWithZeroTimeQuota`). Together they
   pin the mechanism the fix relies on.

  ### Minors — fixed

  - **#4** Renamed `testTruncateLookaheadDeadlineApplyMakesProgress` → `testApplyMakesProgressWithZeroTimeQuota` and moved into its alphabetical slot just before `testApplyManySmallCommits2Writers`.
  - **#5** Added `readWalTxnDetailsFuzzy(writer, cursor)` helper in `WalTxnDetailsFuzzTest`. Each call coin-flips between the 1-arg form and a deadline of `clock.now() + rnd.nextLong(50)`; the deadline branch loops until all txns are loaded so deterministic assertions still hold.
   All 7 callsites use the helper.
  - **#6** PR title updated to `"...committed at a very high rate"`.
  - **#7** Enterprise CI: the cited failing commit predates the current branch head; the more recent merge should re-trigger.
  - **#8** `WalTxnDetails` now caches `MicrosecondClock` in a final field assigned in the constructor; the inline `config.getMicrosecondClock()` lookup in `readObservableTxnMeta` is gone.

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/wal/WalTxnDetails.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/ApplyWal2TableJob.java|3|3|100.00%|

bluestreak01 (approved):

PR #7065, opened 2026-05-06
chore(sql): allow disabling covering index for refresh queries

nwoolmer:
This is a temporary measure whilst we improve async group by performance for queries leveraging the covering index.

## Summary

- Adds the `cairo.mat.view.covering.index.enabled` configuration option (default `false`). When false, the SQL planner skips the covering-index path for materialized view refresh queries and uses the regular plan instead.
- Covering-index group-by and multi-threaded performance currently lags the regular plan, which can hurt mat view refresh latency for queries that would otherwise pick up the new path. This option lets operators opt out until that gap is closed.
- Tradeoff: setups where covering would have made the refresh faster (small, highly selective filters with INCLUDE columns) lose that speed-up while the option is left at its default. Ad-hoc queries are unaffected — they always use covering when eligible.

The gate is implemented as a `default` method on `SqlExecutionContext` returning `true`. `MatViewRefreshSqlExecutionContext` is the sole override; it caches the config value at construction. `SqlCodeGenerator` AND's the new check into the existing `!SqlHints.hasNoCoveringHint(model)` condition at all five covering-eligible decision points (LATEST ON single-key, LATEST ON multi-key, IN-list / equality single-key, IN-list multi-key, DISTINCT-via-posting symbol enumeration).

## Test plan

- new `MatViewCoveringIndexTest` exercises all five covering-index gate sites in `SqlCodeGenerator` and verifies the refresh context drops covering by default while the regular context still picks it
- `MatViewCoveringIndexTest` also covers the override path (`cairo.mat.view.covering.index.enabled=true` restores covering for refresh)
- `PropServerConfigurationTest` updated for both default-off and override-on assertions
- regression check: existing `CoveringIndexTest` (344) and `MatViewTest` (175) still pass

### DIFF
--- a/core/src/main/java/io/questdb/PropServerConfiguration.java
+++ b/core/src/main/java/io/questdb/PropServerConfiguration.java
@@ -325,6 +325,7 @@ public class PropServerConfiguration implements ServerConfiguration {
     private final DateLocale logTimestampLocale;
     private final String logTimestampTimezone;
     private final TimeZoneRules logTimestampTimezoneRules;
+    private final boolean matViewCoveringIndexEnabled;
     private final boolean matViewEnabled;
     private final long matViewInsertAsSelectBatchSize;
     private final int matViewMaxRefreshIntervals;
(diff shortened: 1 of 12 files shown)


bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlExecutionContext.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshSqlExecutionContext.java|2|2|100.00%|

PR #7066, opened 2026-05-06
fix(pgwire): fix partial query result when re-binding a suspended portal

RaphDal:
## Summary

After an Execute suspends a portal, its cursor is intentionally retained so the next Execute on the same portal can resume from the same row. However, when the client sends a `Sync` between that suspended Execute and a follow-up `Bind` for the same statement, `pipelineCurrentEntry` becomes `null` at Sync's end, so the existing pre-lookup `closeSuspendedCursor()` guard in `msgBind` no longer fires. The subsequent `lookupPipelineEntryForNamedStatement` then re-introduces the suspended entry with its cursor still alive, and `msgExecuteSelect` skips `factory.getCursor()`, returning a partial result set.

The fix adds a post-lookup `closeSuspendedCursor()` guard in `msgBind`: a `Bind` always starts a fresh execution, so any suspended cursor on the entry the lookup lands on must be freed before `setStateBind(true)`.

A small documentation comment on `clearState()` records the invariant that callers reach for `closeSuspendedCursor()` if they intend to discard a suspended cursor (since `clearState()` deliberately leaves `stateSuspended` and `cursor` untouched so that suspended portals can be resumed).

## Description

The PG extended-query state machine retains a suspended portal's cursor across `Sync` so that the next `Execute` for that portal can resume from the same row. The bug surfaces when the client interleaves `Close-S` of an unrelated statement and a follow-up `Bind` for the suspended statement.

A representative pgjdbc sequence (driven by `setMaxRows()` and `preparedStatementCacheQueries=0`):

```
Bind('' -> S_2) + Execute('', 2) + Sync     -- S_2 suspended after row 2
... other commands ...
Close-S(S_1) + Sync + Bind('' -> S_2) + Execute('', 6) + Sync
```

State at the start of `Close-S(S_1)`:
- `pipelineCurrentEntry == S_2` with `stateSuspended = true`, cursor alive.

`msgClose` finds the close target (`S_1`) is different from `pipelineCurrentEntry` (`S_2`). Because `S_2` is not "dirty" by the existing `isDirty()` definition (suspended is intentionally not dirty -- a suspended portal must be resumable), `msgClose` calls `releaseToPoolIfAbandoned(S_2)`, which calls `clearState()` only. `clearState()` does NOT touch `cursor` or `stateSuspended`, so `S_2` keeps its suspended cursor (correct: a Close-S on an unrelated statement must not invalidate it). `pipelineCurrentEntry` is then set to `S_1` with `stateClosed = true`.

The intervening `Sync` drains the pipeline (the `Close-S` reply is sent), and `pipelineCurrentEntry` ends up `null`. `S_2` is still in `namedStatements` with cursor alive, `stateSuspended = true`.

`Bind('' -> S_2)` enters `msgBind`. The pre-lookup `closeSuspendedCursor()` guard is not triggered because `pipelineCurrentEntry` is null. `lookupPipelineEntryForNamedStatement(S_2)` re-introduces S_2 as the current entry with its cursor still alive. Without the post-lookup guard, `setStateBind(true)` runs and the next `Execute` lands in `msgExecuteSelect` with `cursor != null`, so `factory.getCursor()` is skipped and the cursor resumes from row 3 instead of starting fresh.

The post-lookup `closeSuspendedCursor()` guard mirrors the existing pre-lookup check: a `Bind` always starts a fresh execution, so any suspended cursor on the entry the lookup lands on must be freed before `setStateBind(true)`.

## What was tried and rejected

Two earlier iterations of the fix turned out to introduce regressions instead of fixing the bug:

- **Adding `stateSuspended` to `isDirty()`** so that `msgClose` would queue the suspended entry onto `pipeline` rather than calling `releaseToPoolIfAbandoned`. This created a duplicate-reference bug: the entry was reachable from both `pipeline` and `namedPortals` simultaneously. A subsequent `Execute` lookup re-introduced it as `pipelineCurrentEntry`, the next `Sync` queued it again, and `syncPipeline` polled the same instance twice -- the second poll triggered the existing `if (nextEntry != null && isSuspended) closeSuspendedCursor()` branch, freeing the cursor mid-stream. `testCursorFetch` reproduced this when pgjdbc evicts an unrelated cached statement (the `CREATE TABLE`) during the cursor-fetch loop.
- **Closing the suspended cursor in `releaseToPoolIfAbandoned`** when the displaced entry was a suspended named entry. This makes any `Close-S` / `Close-P` / `Describe` / `Execute` on an unrelated entity destroy the suspended portal's cursor, which contradicts PG protocol semantics: portals persist across the transaction and are only invalidated by `Close-P` of that portal, `Close-S` of its parent statement, or a re-`Bind`. The same `testCursorFetch` failure mode resulted: pgjdbc's mid-loop `Close-S(S_1)` displaced the suspended portal and reset its cursor, returning row 1 on the next batch instead of continuing.

The accepted fix is narrower: only `msgBind` itself frees the cursor of the entry being rebound. `Close`, `Describe`, and `Execute` on unrelated entities preserve suspended cursors, matching PG semantics.

## Tradeoffs

- Positive: pgjdbc clients that re-execute a `PreparedStatement` (the `setMaxRows` / `preparedStatementCacheQueries=0` pattern, where pgjdbc emits a deferred `Close-S` followed by `Sync` and then a fresh `Bind` for another cached statement) no longer get rows resumed from the wrong cursor.
- `msgBind` (the post-lookup guard) gains one branch and one method call per `Bind`. The branch evaluates `pipelineCurrentEntry.isSuspended()` against a boolean field; cost is a non-issue.
- The `clearState()` doc comment is documentation only. No behavior change.
- No change to `Close-S`, `Close-P`, `Describe`, or `Execute` paths -- their handling of unrelated suspended portals is unchanged from master.

## Test plan

- `testCloseStatementMustNotLeakSuspendedCursorOnAnotherNamedStatement` (new, JDBC). Drives pgjdbc with `preparedStatementCacheQueries=0` to force `Close-S` flush on the next `executeQuery`, and `prepareThreshold=-1` to force named statements from the first execute. Asserts `show columns from tab` returns all 3 rows after `Close-S` of an unrelated named statement followed by re-execute. Pre-fix returns only 1 row; post-fix returns 3.
- `testCloseStatementBeforeBindMustNotResumeSuspendedCursor` (new, hex). Variant of the above with no intermediate `Sync` between `Close-S` and the next `Bind` -- the bug-trigger sequence is `Close-S(S_1) + Bind('' -> S_2) + Execute('', 10) + Sync`. Verifies the post-lookup guard fires whether or not `pipelineCurrentEntry` is null at Bind time. Asserts all 5 rows / `SELECT 5` and the protocol-correct response order: `CloseComplete + BindComplete + 5 DataRows + CommandComplete + RFQ`. Pre-fix returns 3 rows / `SELECT 3`.
- `testCursorFetch` (existing, JDBC). Regression coverage for the rejected approaches: pgjdbc evicts the `CREATE TABLE` statement mid-fetch and continues fetching from the suspended `SELECT` portal. Pre-fix (with the rejected `isDirty()` or `releaseToPoolIfAbandoned` changes) failed intermittently with cursor restart at the start of a new batch. Post-fix: 30 consecutive local runs pass.
- All existing `PGJobContextTest` tests pass, including the four `testUnnamedPortalCursor*` regression tests added by `8708e2bc4e` for the postgres.js cursor-resume pattern.

### DIFF
--- a/core/src/main/java/io/questdb/cutlass/pgwire/PGConnectionContext.java
+++ b/core/src/main/java/io/questdb/cutlass/pgwire/PGConnectionContext.java
@@ -722,9 +722,11 @@ public class PGConnectionContext extends IOContext<PGConnectionContext> implemen
         }
 
         if (pipelineCurrentEntry.isSuspended()) {
-            // Symmetric to the pre-lookup check above. The lookup may have brought in
-            // a named entry whose cursor was retained from a previous Execute (e.g.,
-            // queued onto pipeline by an earlier Close-S that displaced it). A new
+            // Symmetric to the pre-lookup check above. The lookup may have re-introduced
+            // a named entry whose cursor was retained from a previous suspended Execute
+            // (e.g., a Close-S of an unrelated statement landed in between, intentionally
+            // preserving the suspended cursor; an intervening Sync then nulled
+            // pipelineCurrentEntry, so the pre-lookup guard could not see it). A new
             // Bind always starts a fresh execution, so the prior cursor must be freed.
             pipelineCurrentEntry.closeSuspendedCursor();
         }
@@ -1553,12 +1555,6 @@ public class PGConnectionContext extends IOContext<PGConnectionContext> implemen
             if (pe.isCopy || (!pe.isPreparedStatement() && !pe.isPortal())) {
                 releaseToPool(pe);
             } else {
-                if (pe.isSuspended()) {
-                    // The entry is being displaced (not retained as the current entry).
-                    // Its cursor must be freed here because nothing else will resume it,
-                    // and clearState() does not touch stateSuspended or cursor.
-                    pe.closeSuspendedCursor();
-                }
                 pe.clearState();
             }
         }
--- a/core/src/main/java/io/questdb/cutlass/pgwire/PGConnectionContext.java
+++ b/core/src/main/java/io/questdb/cutlass/pgwire/PGConnectionContext.java
@@ -721,6 +721,14 @@ public class PGConnectionContext extends IOContext<PGConnectionContext> implemen
             throw msgKaput().put("received a Bind message without a matching Parse");
         }
 
+        if (pipelineCurrentEntry.isSuspended()) {
+            // Symmetric to the pre-lookup check above. The lookup may have brought in
+            // a named entry whose cursor was retained from a previous Execute (e.g.,
+            // queued onto pipeline by an earlier Close-S that displaced it). A new
+            // Bind always starts a fresh execution, so the prior cursor must be freed.
+            pipelineCurrentEntry.closeSuspendedCursor();
+        }
+
         pipelineCurrentEntry.setStateBind(true);
 
         // "bind" is asking us to create portal. We take the conservative approach and assume
@@ -1545,6 +1553,12 @@ public class PGConnectionContext extends IOContext<PGConnectionContext> implemen
             if (pe.isCopy || (!pe.isPreparedStatement() && !pe.isPortal())) {
                 releaseToPool(pe);
             } else {
+                if (pe.isSuspended()) {
+                    // The entry is being displaced (not retained as the current entry).
+                    // Its cursor must be freed here because nothing else will resume it,
+                    // and clearState() does not touch stateSuspended or cursor.
+                    pe.closeSuspendedCursor();
+                }
                 pe.clearState();
             }
         }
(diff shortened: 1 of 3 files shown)


mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGConnectionContext.java|2|2|100.00%|

bluestreak01 (approved):

PR #7021, opened 2026-04-23
chore(core): add randomized query fuzzer and fix multiple SQL correctness and resource-leak bugs

puzpuzpuz:
## Summary

Adds `QueryFuzzTest` under `core/src/test/java/io/questdb/test/griffin/fuzz/` - a seeded randomised JUnit test that generates 1-3 WAL tables with a wide range of column types, inserts rows spanning multiple DAY partitions, and then runs a configurable budget of randomly generated queries against them. The oracle is crash-only: the test fails on any exception outside a short allowlist, and materialises every result row so per-column accessors are exercised end to end.

The same branch also ships fixes for every bug class the fuzzer surfaced during development. After the fixes, repeated runs of 5k-10k queries each are clean.

## Behavior changes

Bug 27 below changes the result type of `INT - FLOAT` from DOUBLE to FLOAT. The new `SubFloatFunctionFactory` resolves `-(FF)` directly, mirroring `+(FF)`, `*(FF)`, and `/(FF)` which already returned FLOAT for the same operand pair. Queries that previously relied on the intermediate result of an `INT - FLOAT` subtree being kept at f64 precision will see it truncated to f32. Downstream casts, comparisons near an f32 ulp boundary, or chained arithmetic that further widens to f64 can produce different values than before. To preserve the old behavior, cast one operand to DOUBLE explicitly: `c0::DOUBLE - c5` or `c0 - c5::DOUBLE`.

Bug 41's follow-up changes how STRING / VARCHAR `<=` and `>=` against a constant NULL behave when the table side is also null. Previously the literal-form short-circuit returned FALSE for every row, so `c <= null` produced 0 rows even when `c IS NULL`. The bind-variable form already returned TRUE through the runtime `Chars.lessThan` / `Utf8s.lessThan` path under QuestDB's `NULL = NULL -> true` convention, so the literal and bind forms disagreed on the both-null edge. The fix now honours `negated && argIsNull(rec)`, so `c <= null` and `c >= null` return the matching NULL rows -- aligning the literal form with the bind form, the runtime path, and the in-PR DECIMAL fix for Bug 26. Queries that relied on the previous "always 0 rows" behaviour for `c <= null` / `c >= null` need to filter explicitly with `c IS NOT NULL`, or use `c < null` / `c > null` which still return 0 rows in every case.

[...]

PR #7067, opened 2026-05-06
chore(parquet): store stats in parquet sidecar in their original format

RaphDal:
## Summary

The `_pm` parquet metadata sidecar stores per-row-group min/max for inline-eligible columns (fixed_size <= 8) in a u64 stat slot. The skip code in `can_skip_row_group` re-reads that slot at parquet physical type width and applies parquet-native interpretations (e.g. `i32::from_le_bytes` for INT32, `is_date * MILLIS_PER_DAY` for DATE). The slot is therefore an opaque parquet-native byte buffer from the reader's point of view.

The convert path violated that contract. `convert_stat_to_qdb` reshaped parquet stats to QuestDB-native width and units before they reached the inline slot:

- INT32-backed narrow types (BYTE, SHORT, GeoByte, GeoShort, Char) were truncated to 1 or 2 bytes. The inline write path then zero-padded to 8 bytes, so a SHORT min of `-74` (`i16 0xFFB6`) round-tripped through the skip code's 4-byte `i32::from_le_bytes` read as `0x0000FFB6 = 65462`. Predicates like `c1 <= 0` then tripped `min_val > v` and dropped every row group whose true min was negative. The query fuzzer surfaced this on seed `785607679928725L, 1778086793079L`, query `SELECT ... FROM fuzz_t0 WHERE 0.147451::FLOAT >= c1 LIMIT 32` against a SHORT column with negative values: shadow returned 32 rows, primary returned 0.
- INT32 Date (days since epoch) was widened to i64 millis. The skip code reads 4 bytes for INT32 physical width and applies `MILLIS_PER_DAY` itself, so it would interpret the low 4 bytes of an 8-byte millis value as days. Latent for QuestDB-written parquet (which always uses INT64 for Date) but wrong for external Int32-Date parquets ingested through the inline path.
- INT64 `Timestamp(Millis)` was multiplied by 1000 to micros. The branch was guarded by `(Int64, ColumnTypeTag::Timestamp)`, but `infer_column_type` maps external Int64+Millis to `ColumnTypeTag::Date`, not `Timestamp`, and QuestDB itself never writes Timestamp as Millis. The conversion was unreachable in any QDB-managed or schema-inferred external workflow; only a hand-crafted `QdbMeta` override could trigger it. Removing it just deletes dead code.
- INT96 was converted to i64 epoch nanos. INT96 stats are always 12 bytes, so the convert path inlined the converted nanos into the slot. `find_row_group_by_timestamp` reads `chunk.min_stat as i64` and could observe these stats when the designated timestamp was an INT96 column. QuestDB never writes INT96, and external parquet inferred as INT96 + Timestamp(NS) is rarely the designated timestamp; after this change those calls fall back to `decode_ts` instead of an inline read. Functionally correct, slightly slower for the rare case.

The fix removes `convert_stat_to_qdb` and stores parquet stats verbatim at parquet physical width. The convert path now reads `min_value`/`max_value` directly from `metadata().statistics` (raw thrift bytes) and writes them through to the inline slot or OOL region without any typed deserialize/re-serialize round-trip. Inline vs OOL is gated purely by stat byte width (1..=8 inline, longer OOL), so the QuestDB column type tag no longer constrains placement: short Varchar/Symbol stats can now inline alongside primitive stats. The `extract_stat_values` helper, the type-driven `is_inline` gate, and the `col_type_tag` plumbing through `build_column_chunk_from_thrift`, `build_row_group_block_from_thrift_with_types`, `generate_parquet_metadata`, `update_parquet_metadata`, and the `ParquetMetaColumnInfo` struct are all gone with it.

PR #7021 sign-extended narrow signed types to INT32 width to make the inline slot read correctly. That fix is sound for the BYTE/SHORT/Geo* case, but it leaves the Int32-Date conversion in place. Removing the conversion altogether collapses the issues into one rule: the inline u64 slot holds parquet bytes verbatim.

## Tradeoffs

- The unit-conversion path for external Int32-Date parquets is gone, and the skip path now sees raw days as it expects. QuestDB writes Date as INT64 + `Timestamp(Millis)`, so QuestDB-written parquet is unaffected.
- INT96 inline stats now go OOL. `can_skip_row_group` does not consume INT96 inline stats today (Int96 maps to `phys_size = 0` in skip.rs), so this is observable only in `find_row_group_by_timestamp`, which now decodes from parquet instead of reading the slot. The path is reachable only for external INT96 designated-timestamp partitions.
- The Int64 `Timestamp(Millis)` -> micros conversion is gone, but as covered in the summary, that branch was unreachable for QDB-managed and schema-inferred external workflows. A future external-parquet path that explicitly tags Millis-stored timestamps as QDB Timestamp via `QdbMeta` would need to convert at the read site, using the logical type already on the column descriptor.

## Test plan

- cargo fmt, cargo check --all-targets, cargo clippy --all-targets all clean
- cargo test --lib in core/rust/qdbr: 889 passed, 0 failed (the 7 unit tests for the removed `extract_stat_values` helper are gone, replaced by integration coverage through the regression tests below)
- Regression test `convert_short_negative_min_round_trips_at_int32_width` writes a SHORT column with values `-50..49`, runs the full `convert_from_parquet` path, and asserts the inline `min_stat` reads back as `-50i32` and `max_stat` as `49i32` at parquet physical width — the exact read the skip path performs
- Regression test `convert_int32_date_round_trips_at_int32_width` constructs an INT32 + Date logical-type column via thrift, runs `build_row_group_block_from_thrift_with_types`, and asserts the inline slot holds i32 days at 4-byte physical width with the high 4 bytes zero — proving the convert path no longer widens days to i64 millis
- New Java tests `testMinMaxPruningGeoByte` and `testMinMaxPruningGeoShort` insert geohash values across two partitions, convert to parquet, and verify equality and full-scan reads round-trip correctly through the new 4-byte inline stat encoding
- mvn -pl core -Dtest=ParquetRowGroupPruningTest test: 112 passed
- mvn -pl core -Dtest=ParquetMetaFileReaderTest,ParquetMetaGenerateTest,ParquetMetaStalePmTest,ParquetMetaWriteTest,ParquetMetaCrcCorruptionTest test: 61 passed

### DIFF
--- a/core/rust/qdbr/src/parquet_metadata/convert.rs
+++ b/core/rust/qdbr/src/parquet_metadata/convert.rs
@@ -1430,6 +1430,125 @@ mod tests {
         );
     }
 
+    /// Regression: an external INT32 + Date logical-type parquet must keep its
+    /// stats as i32 days at parquet physical width when ingested through the
+    /// inline path. The earlier convert_stat_to_qdb widened days to i64 millis,
+    /// so the skip path's 4-byte INT32 read interpreted the low 4 bytes of
+    /// millis as days and produced wildly wrong bounds. QuestDB itself writes
+    /// Date as INT64+millis, so this case only surfaces for external Int32-Date
+    /// parquets converted through `build_column_chunk_from_thrift`.
+    #[test]
+    fn convert_int32_date_round_trips_at_int32_width() {
+        use parquet2::metadata::{ColumnDescriptor, Descriptor};
+        use parquet2::schema::types::{
+            FieldInfo, ParquetType, PhysicalType, PrimitiveLogicalType, PrimitiveType,
+        };
+        use parquet2::schema::Repetition;
+        use parquet2::thrift_format::{
+            ColumnChunk, ColumnMetaData, CompressionCodec, Encoding as ThriftEncoding, RowGroup,
+            Statistics as ThriftStatistics, Type,
+        };
+
+        let min_days: i32 = -100; // ~1969-09-23
+        let max_days: i32 = 365; //  1971-01-01
+
+        let primitive_type = PrimitiveType {
+            field_info: FieldInfo {
+                name: "d".to_string(),
+                repetition: Repetition::Required,
+                id: None,
+            },
+            logical_type: Some(PrimitiveLogicalType::Date),
+            converted_type: None,
+            physical_type: PhysicalType::Int32,
+        };
+        let base_type = ParquetType::PrimitiveType(primitive_type.clone());
+        let descriptor = Descriptor { primitive_type, max_def_level: 0, max_rep_level: 0 };
+        let col_desc = ColumnDescriptor::new(descriptor, vec!["d".to_string()], base_type);
+
+        let stats = ThriftStatistics {
+            max: None,
+            min: None,
+            null_count: Some(0),
+            distinct_count: None,
+            max_value: Some(max_days.to_le_bytes().to_vec()),
+            min_value: Some(min_days.to_le_bytes().to_vec()),
+        };
+        let meta = ColumnMetaData {
+            type_: Type::INT32,
+            encodings: vec![ThriftEncoding::PLAIN],
+            path_in_schema: vec!["d".to_string()],
+            codec: CompressionCodec::UNCOMPRESSED,
+            num_values: 100,
+            total_uncompressed_size: 400,
+            total_compressed_size: 400,
+            key_value_metadata: None,
+            data_page_offset: 4,
+            index_page_offset: None,
+            dictionary_page_offset: None,
+            statistics: Some(stats),
+            encoding_stats: None,
+            bloom_filter_offset: None,
+            bloom_filter_length: None,
+        };
+        let column_chunk = ColumnChunk {
+            file_path: None,
+            file_offset: 0,
+            meta_data: Some(meta),
+            offset_index_offset: None,
+            offset_index_length: None,
+            column_index_offset: None,
+            column_index_length: None,
+            crypto_metadata: None,
+            encrypted_column_metadata: None,
+        };
+        let row_group = RowGroup {
+            columns: vec![column_chunk],
+            total_byte_size: 400,
+            num_rows: 100,
+            sorting_columns: None,
+            file_offset: None,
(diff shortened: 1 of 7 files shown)


puzpuzpuz (dismissed):

core/rust/qdbr/src/parquet_metadata/convert.rs
kafka1991: why not inline vs OOL purely by stat byte width (1..=8), independent of ColumnTypeTag::fixed_size (like short string/varchar also can be inlined) ?

_Current Symbol situation: Symbol::fixed_size() = Some(4), but its parquet stats are variable-length UTF-8 strings — so is_inline = true only sets the intent, and a secondary min_val.len() <= 8 check at convert.rs:336/348 is what actually keeps things correct. The type-based gate ends up being a redundant proxy for the width check it already requires._ 

So I suggest drop the type check entirely here

kafka1991: read raw min_value/max_value bytes straight from metadata().statistics — `col_chunk.statistics()` deserializes into a typed Arc<dyn Statistics> and for `extract_stat_values` to re-serialize back to LE bytes via a physical_type match, all avoidable

kafka1991 (changes requested):

core/rust/qdbr/src/parquet_metadata/convert.rs
RaphDal: Fair point, it largely simplifies the code too, fixed in https://github.com/questdb/questdb/pull/7067/commits/1e30c0a60b54729ce2e151592d02210aae8b1e5a

RaphDal: Good catch, fixed in https://github.com/questdb/questdb/pull/7067/commits/1e30c0a60b54729ce2e151592d02210aae8b1e5a

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 193 / 200 (96.50%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/parquet_metadata/convert.rs|193|200|96.50%|

kafka1991 (approved):
LGTM

PR #7068, opened 2026-05-07
chore(core): fix NPE on ORDER BY against an unknown indexed symbol

bluestreak01:
## Summary

- A query that filters a posting-indexed `SYMBOL` column by a literal the symbol map has never seen, and orders results by that column, crashed with `NullPointerException` instead of returning an empty result. The same crash propagated through `CREATE VIEW` validation, which is how the Fuzz Tests pipeline first surfaced it (Azure build [231547](https://dev.azure.com/questdb/questdb/_build/results?buildId=231547), `ViewFuzzTest.testMultipleLevelDependencyViewsWithRandomSelect`).
- The fix keeps the cursor wired up even when no keys resolve, so upstream operators (`ORDER BY`, etc.) can probe `getSymbolTable()` before iteration. A regression test driven by `assertQueryNoLeakCheck` would have caught this directly without waiting for a fuzz hit.

## Cause

When the WHERE literal resolved to `VALUE_NOT_FOUND`, `CoveringIndexRecordCursorFactory.getCursor()` freed the partition frame cursor and called `ofEmpty()`, leaving `frameCursor` and `tableReader` null. `ORDER BY` on a `SYMBOL` column wraps the base cursor in `EncodedSortRecordCursor`, whose `init()` phase probes `baseCursor.getSymbolTable()` before any iteration to build symbol-rank maps. That probe then dereferenced the null `frameCursor`.

```
java.lang.NullPointerException: Cannot invoke "PartitionFrameCursor.getSymbolTable(int)" because "this.frameCursor" is null
  at CoveringIndexRecordCursorFactory$CoveringCursor.getSymbolTable(...:441)
  at SortKeyEncoder.init(...:217)
  at EncodedSortRecordCursor.of(...:148)
  at EncodedSortRecordCursorFactory.getCursor(...:70)
  at LimitRecordCursorFactory.getCursor(...:63)
  at StaleViewCheckFactory.getCursor(...:102)
  at SqlCompilerImpl.executeCreateView(...:4450)
```

## Change

- `getCursor()` and `getPageFrameCursor()` now wire the cursor up via `of()` in all cases. Symbol-table probes always work; iteration short-circuits naturally when the multi-key list is empty.
- Explicit guards in `SingleKeyCoveringCursor.advanceKey()` / `hasNextLatestBy()` and `SingleKeyCoveringPageFrameCursor.nextImpl()` short-circuit when the resolved key is `VALUE_NOT_FOUND`. The guards also stop `VALUE_NOT_FOUND` (-2) from reaching `TableUtils.toIndexKey()`, which would otherwise produce `-1` and force every partition's index reader to be opened just to return an empty cursor.
- Multi-key paths short-circuit on empty key lists, removing a redundant full-partition scan when an `IN` list contains only unknown literals.
- The dead `ofEmpty()` methods are removed.

## Tradeoffs

- One extra integer comparison per `advanceKey()` / `nextImpl()` call. Offset by the removed full-partition scan in the empty multi-key case.
- The previous page-frame variant returned `null` from `getSymbolTable()` on an empty cursor, which was already unsafe for any caller that did not null-check; one existing test (`testPageFrameCursor_OfEmptyNullSymbolTables`) was pinning down that contract. The new contract always returns a usable symbol table.

## Test plan

- New regression test `testEqFilterOnUnknownSymbolWithOrderBy` in `CoveringIndexTest`. Populated table (2,000 rows across 5 known symbols), queries known and unknown symbols with `ORDER BY` in single-key, single-key + `LIMIT`, `IN`-list, mixed `IN`-list (one known + one unknown), bind-variable single-key, bind-variable `IN`-list, and `LATEST ON` (single-key, multi-key all-unknown, multi-key mixed) shapes. Verifies the EXPLAIN plan still picks the CoveringIndex path. Includes a positive `ORDER BY ... DESC LIMIT` baseline on a known symbol so the same factory chain that produced the original NPE stack is exercised on populated data. Verified red against unfixed code with the same NPE; green with the fix.
- `testPageFrameCursor_OfEmptyNullSymbolTables` was documenting the buggy contract; renamed to `testPageFrameCursor_OfEmptySingleKeyKeepsSymbolTables` and the assertions inverted to require non-null.
- All 1,035 covering-index tests pass across the three code-path variants (`CoveringIndexTest`, `CoveringIndexTestEf`, `CoveringIndexTestDelta`).
- All 4 `ViewFuzzTest` cases pass.
- Posting-index suite (`PostingIndexStressTest`, `PostingIndexNativeTest`, `PostingIndexOracleTest`, `PostingIndexCriticalIssuesTest`, `PostingIndexConcurrencyTest`, `PostingSealPurgeTest`, plus `Delta`/`Ef` variants and `IndexTypeTest`, `SymbolNotEqualsValueTest`) clean.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/table/CoveringIndexRecordCursorFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/table/CoveringIndexRecordCursorFactory.java
@@ -184,7 +184,6 @@ public class CoveringIndexRecordCursorFactory implements RecordCursorFactory {
                 }
                 SymbolMapReader smr = frameCursor.getTableReader().getSymbolMapReader(indexColumnIndex);
                 multiKeyCursor.multiKeys.clear();
-                boolean hasAnyKey = false;
                 for (int i = 0, n = resolvedKeys.size(); i < n; i++) {
                     int key = resolvedKeys.getQuick(i);
                     if (key == SymbolTable.VALUE_NOT_FOUND && keyValueFuncs != null) {
@@ -193,14 +192,13 @@ public class CoveringIndexRecordCursorFactory implements RecordCursorFactory {
                     }
                     if (key != SymbolTable.VALUE_NOT_FOUND) {
                         multiKeyCursor.multiKeys.add(key);
-                        hasAnyKey = true;
                     }
                 }
-                if (!hasAnyKey) {
-                    Misc.free(frameCursor);
-                    multiKeyCursor.ofEmpty();
-                    return multiKeyCursor;
-                }
+                // Always wire up the frame cursor and table reader, even when no
+                // keys resolve. Callers wrap us in operators (e.g. ORDER BY on a
+                // SYMBOL column) that probe baseCursor.getSymbolTable() during
+                // init, before any iteration. An empty multiKeys list naturally
+                // makes advanceKey() report no rows.
                 multiKeyCursor.of(frameCursor);
                 multiKeyCursor.latestByFilter = latestByFilter;
                 if (latestByFilter != null) {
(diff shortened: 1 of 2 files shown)


mtopolnik (dismissed):

mtopolnik (approved):

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/CoveringIndexRecordCursorFactory.java|11|11|100.00%|

PR #7069, opened 2026-05-07
fix(core): fix RENAME COLUMN failure on posting-indexed columns

bluestreak01:
## Summary

`ALTER TABLE RENAME COLUMN` on a non-WAL table with a posting-indexed `SYMBOL` column can throw `[errno=2] could not create hard link` and distress the writer mid-DDL. Recently observed on master as a fuzz failure: `WalWriterFuzzTest.testAddDropColumnDropPartition` ([build 231522](https://dev.azure.com/questdb/questdb/_build/results?buildId=231522)).

## Root cause

`TableWriter.linkPostingIndexAuxFiles` enumerates sealed `.pc<N>` cover-data files via `scanSealedFiles` (`iterateDir`) and hardlinks each one to the new column's namespace. Inside `linkFile`, `ff.exists(from)` runs first, then `ff.hardLink(from, to)`. Between those two syscalls, `PostingSealPurgeJob` — running on a worker thread, independent of the writer — can finish processing a queued `PostingSealPurgeTask` for a superseded sealed generation and `removeQuiet` the file. The subsequent `hardLink` returns `ENOENT`, the rename throws, and the writer is distressed.

The CI log captures the exact interleaving: `PostingSealPurgeOperator purged posting sealed version [column=sym2, postingColumnNameTxn=-1, sealTxn=0]` runs 191 microseconds before `TableWriter writer error: could not create hard link from=...sym2.pc0.0.0, to=...new_col_0.pc0.7.0`.

## Marko's new fix

`linkPostingIndexAuxFiles` reads the live `sealTxn` from the source `.pk` — it already does this for the `.pv` link — and uses it as the filter for the `.pc<N>` visitor. Only files whose `sealTxn` matches the chain head are hardlinked to the dst column's namespace. The visitor now mirrors the long-standing `.pv` link logic exactly.

### Why the original fix is no longer needed

- **No race to tolerate.** The visitor never attempts to link a superseded `.pc<N>`, so the TOCTOU between `ff.exists(from)` and `ff.hardLink(from, to)` is closed structurally rather than swallowed. The live `sealTxn` cannot be subject to a pending or in-flight purge task: rename holds the writer lock so no new seal can land, and purge only targets superseded sealTxns.
- **No `tolerateConcurrentDelete` parameter.** `linkFile` returns to a single, strict signature. `ENOENT` is once again a hard error in every linker — there is no path that can silently swallow it. The original fix's narrow tolerance was correct for the cases it covered, but introducing an "ignore missing source" path widens the surface area where a real bug could hide behind a tolerated `ENOENT`.

### Why the new fix is also better

- It fixes a latent issue the original fix did not address. The original visitor enumerated *every* sealed generation on disk, so the dst column accumulated hardlinks of superseded generations. Those orphans were never targeted by any purge task (purge tasks reference the *src* column name) and survived until the column was dropped. The new filter keeps the dst minimal: only files reachable from the chain head arrive there.

### Tradeoffs of the new fix

- The fix relies on `readSealTxnFromKeyFile`. If the seqlock retries are exhausted or the `.pk` is malformed, it returns `-1` and no `.pc<N>` is linked — same behavior the existing `.pv` path has had since day one. In practice this only happens when the column has never been sealed, in which case there are no `.pc<N>` files to link anyway.
- The new test does not reproduce the original CI error directly, because the original race is no longer structurally reachable. The previous fix's regression test has been removed; coverage of the end-to-end CI scenario falls to the fuzz suite (`WalWriterFuzzTest.testAddDropColumnDropPartition`).

### Test plan for the new fix

- New deterministic test: `PostingIndexCriticalIssuesTest#testRenameColumnLinksOnlyLiveSealGeneration`. Plants a "ghost" `.pc<N>` file with `sealTxn=9999`, runs `ALTER TABLE RENAME COLUMN`, scans the dst partition with `PostingIndexUtils.scanSealedFiles`, and asserts the ghost is absent on dst while the live `sealTxn=0` file is present. The data round-trip is also verified.
- All 128 `PostingIndex*Test` tests pass, all 524 `TableWriterTest`/`AlterTableRenameColumnTest`/`CoveringIndexTest` tests pass, all 15,547 `*FuzzTest` tests pass (including the original `WalWriterFuzzTest.testAddDropColumnDropPartition`). 0 failures, 0 regressions.

## Original fix

`linkFile` gains an opt-in `tolerateConcurrentDelete` parameter. After a failed `hardLink`, if `errno` indicates the file does not exist and `ff.exists(from)` is now false, return `false` silently instead of throwing. Only the `.pc<N>` visitor in `linkPostingIndexAuxFiles` opts in — that is the single call site that races with `PostingSealPurgeJob`. All other linkers (the live `.pv` resolved from `.pk`, `.pci`, symbol map files, column data files) operate on files outside the seal-purge contract and remain strict.

The purge contract guarantees no reader is pinned in the visibility window of a purged generation, so the renamed column does not need it either.

## Tradeoffs

- The visitor now silently skips a missing source. The narrow `Files.isErrnoFileDoesNotExist` + `!ff.exists(from)` re-check after a failed `hardLink` keeps the tolerance to the genuine TOCTOU; other failure modes (EACCES, ENOSPC, EXDEV, etc.) still throw.
- The fix is conservative. A deeper refactor — only link the live generation, mirroring how `.pv` already reads `sealTxn` from `.pk` — would also fix a separate latent issue: hardlinks of dead generations under the new column's namespace are not currently targeted by any purge task and survive until the column is dropped. That refactor wants a wider audit (chain walks, recovery, snapshot/restore, partition detach/attach) and is intentionally left for a follow-up.

## Test plan

- New deterministic regression test: `PostingIndexCriticalIssuesTest#testRenameColumnTolerantToConcurrentSealPurge`. A custom `FilesFacade` removes the `.pc<digit>.` source between the visitor's `exists()` and `hardLink()`, simulating the seal purge winning the race. Pre-fix the test reproduces the exact CI error (`[errno=2] could not create hard link from=...sym.pc0.0.0`); post-fix the rename completes and data is queryable.
- 16,169 tests pass across `TableWriterTest`, `CoveringIndexTest`, `PostingIndex*Test`, and `*FuzzTest` (including the original `WalWriterFuzzTest.testAddDropColumnDropPartition`). 0 failures, 0 regressions.


mtopolnik (approved):

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|

PR #7070, opened 2026-05-07
fix(core): fix JVM crash and dropped cover footers on covering posting indexes

bluestreak01:
## Summary

A SIGSEGV reproduced on the `bench/posting-wal-fastlag` benchmark branch (`hs_err_pid19555.log` from `PostingIndexBenchmarkSuite.walFastLagInsertAndQuery` against a `posting_covering` table) traced to two cooperating bugs - a writer-side bug that produced chain entries with no cover footer, and a reader-side bug that over-read past the entry's `LEN` and crossed the .pk mmap boundary.

This PR fixes both.

### Reader fix - bound chain reads against entry LEN

`PostingIndexChainEntry.read()` trusted the reader-supplied `coverCount` unconditionally and stepped past the entry's own `LEN` to read the cover end-offset footer. When the entry's actual cover-footer size did not match the reader's `coverCount` (sourced from the `.pci` sidecar) and the entry sat flush against the end of the `.pk` mmap, the over-read crossed the mapping boundary and SIGSEGV'd the JVM. The picker's `entry.len` bound check ran AFTER `read()` and so could not catch this class of failure.

- `PostingIndexChainEntry.read()` self-bounds the cover-footer loop against the entry's own `LEN`. Slots the entry does not actually contain are zero-filled, which the rest of the reader treats as "no published cover bytes for this slot" - the same fallback the existing code already uses for tombstoned or not-yet-opened sidecars.
- `PostingIndexChainPicker.pick()` peeks `LEN` before invoking `read()` so the full-entry-fits-in-mapping check happens before any cover read. A torn or corrupted `LEN` that overruns the mapping now produces `RESULT_HEADER_UNREADABLE` consistently.

### Writer fix - preserve covering schema across post-seal cleanup

`PostingIndexWriter.clearCovering()` did two things at once: released the read-side state set up for the most recent seal (covered-column RO mmaps and borrowed source-column addresses), and forgot the covering schema (zeroed `coverCount`, dropped `coveredColumnIndices`, etc.). `TableWriter`'s two post-seal `finally` blocks used `clearCovering()` to do the first job, but the second job was a side effect they did not want: between fast-lag seals, a follow-up `commit()` ran `flushAllPending() -> publishToChain() -> chain.extendHead` with `coverEndOffsetsScratch` empty (because `captureCoverEndOffsets` short-circuits when `coverCount<=0`), and `extendHead` rewrote the head entry's `LEN` using `entrySize(genCount, 0)` - dropping the cover footer.

- New `PostingIndexWriter.releaseCoveredColumnReadMappings()` drops the read-side state but keeps `coverCount`, `coveredColumnIndices`, `coveredColumnNames`, `coveredColumnNameTxns`, `coveredColumnTops`, `coveredColumnShifts`, `coveredColumnTypes`, and `sidecarMems` intact. A subsequent `commit()` between seals reads cover end offsets directly from `sidecarMems[c].getAppendOffset()` (the last seal's published extents) and writes a chain entry whose `LEN` includes a correctly populated cover footer. The footer values match the previous seal's extents, which is correct: covers do not advance between seals.
- `IndexWriter`, `ColumnIndexer`, and `SymbolColumnIndexer` gain matching default and forwarding declarations.
- `TableWriter.sealPostingIndexesForO3Partitions` and `sealPostingIndexesForLastPartitionFastLag` finally blocks switch from `indexer.clearCovering()` to `indexer.releaseCoveredColumnReadMappings()`. `clearCovering()` retains its full-teardown semantics for legitimate uses (writer `discardAndClose`, swap to a different cover schema).

## Effect on existing behavior

- For a chain entry whose physical cover-footer size matches the reader's `coverCount`, behavior is unchanged.
- For a chain entry whose cover footer is missing or truncated relative to the reader's `coverCount` (which can still happen via the fd-based O3CopyJob publish window before `sealPostingIndexesForO3Partitions` republishes), the reader no longer SEGVs and falls back to the non-covering path for that entry.
- The new picker pre-check is a single 8-byte read per chain step. No measurable impact on hot paths.
- `clearCovering()`'s contract is unchanged. Production callers that previously misused it for read-mapping cleanup now use the narrower `releaseCoveredColumnReadMappings()`.

## Test plan

- `PostingIndexChainTest` - five new low-level cases covering the read clamp and the picker's pre-bound-check: a coverCount-mismatched entry, a `read()` that must stop at LEN even when more slots are requested, the happy-path round trip, a stomped LEN exceeding the mapping, and a non-positive LEN.
- `PostingIndexCriticalIssuesTest#testWriterEntrysCoverFooterShrinksAfterClearCoveringCommit` - drives `PostingIndexWriter` through `configureCovering -> seal -> clearCovering -> add -> commit`, asserts the head LEN reflects `entrySize(genCount, 0)` (documents `clearCovering()`'s contract for legitimate teardown), plants a sentinel byte pattern at the false-footer offset and verifies the picker returns `0` instead of the sentinel (validates the reader fix).
- `PostingIndexCriticalIssuesTest#testWriterEntrysCoverFooterPersistsAfterReleaseReadMappingsCommit` - symmetric counterpart that drives `configureCovering -> seal -> releaseCoveredColumnReadMappings -> add -> commit` and asserts the head LEN equals `entrySize(genCount, coverCount)` (validates the writer fix).
- `PostingIndexCriticalIssuesTest#testWalFastLagCoveringPostingDoesNotCrashReader` - end-to-end reproducer modeled on `walFastLagInsertAndQuery`: covering POSTING index over `CREATE TABLE ... WAL`, repeated INSERT batch -> `drainWalQueue` -> `SELECT count() WHERE sym = ?` over 64 iterations. On unfixed code the JVM SIGSEGVs inside the chain walk; with the fix in place queries return correct counts.

Verification:

- 49 tests in `PostingIndexChainTest`, `PostingIndexCriticalIssuesTest`, and `PostingIndexChainWriterTest` pass.
- Broader run across `*PostingIndex*`, `*Covering*`, `*Indexer*`, `*WalApply*` test classes (1331 tests) passes.
- The new picker-clamp test fails on unfixed code with the planted sentinel surfacing as a cover end offset (`expected:<0> but was:<-2401053089206453570>` = `0xdeadbeefcafebabe`), confirming the reader-side test catches the regression.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/ColumnIndexer.java
+++ b/core/src/main/java/io/questdb/cairo/ColumnIndexer.java
@@ -124,6 +124,15 @@ public interface ColumnIndexer extends QuietCloseable {
 
     void refreshSourceAndIndex(long loRow, long hiRow);
 
+    /**
+     * Drop the read-side state set up for the most recent seal but keep
+     * the covering schema intact, so a subsequent commit() still publishes
+     * a chain entry with a correct cover footer. See
+     * {@link io.questdb.cairo.idx.PostingIndexWriter#releaseCoveredColumnReadMappings()}.
+     */
+    default void releaseCoveredColumnReadMappings() {
+    }
+
     void releaseIndexWriter();
 
     void resetColumnTop();
--- a/core/src/main/java/io/questdb/cairo/SymbolColumnIndexer.java
+++ b/core/src/main/java/io/questdb/cairo/SymbolColumnIndexer.java
@@ -240,6 +240,11 @@ public class SymbolColumnIndexer implements ColumnIndexer, Mutable {
         index(ff, fd, loRow, hiRow);
     }
 
+    @Override
+    public void releaseCoveredColumnReadMappings() {
+        writer.releaseCoveredColumnReadMappings();
+    }
+
     public void releaseIndexWriter() {
         Misc.free(writer);
     }
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -11186,9 +11186,13 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
                         }
                         // Clear the writer's reference to o3SealAddrs so that
                         // a subsequent close() → seal() cannot dereference the
-                        // unmapped addresses. The seal would return early (gen0
-                        // is already dense), but this is a defensive measure.
-                        indexer.clearCovering();
+                        // unmapped addresses. Keep the covering schema
+                        // (coverCount, sidecarMems) intact so a subsequent
+                        // commit() between seals still publishes a chain
+                        // entry with a correct cover footer; clearCovering()
+                        // would zero coverCount and the next captureCoverEnd
+                        // Offsets would short-circuit, dropping the footer.
+                        indexer.releaseCoveredColumnReadMappings();
                     }
                 } else {
                     indexer.configureFollowerAndWriter(
@@ -11364,7 +11368,11 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
                                 o3SealAuxAddrs.setQuick(c, 0);
                             }
                         }
-                        indexer.clearCovering();
+                        // Drop the borrowed o3SealAddrs pointers from the
+                        // indexer but keep coverCount and sidecarMems set,
+                        // so a follow-up commit() between fast-lag seals
+                        // still publishes with a correct cover footer.
+                        indexer.releaseCoveredColumnReadMappings();
                     }
                     continue;
                 }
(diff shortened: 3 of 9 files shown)


mtopolnik (dismissed):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 58 / 74 (78.38%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/ColumnIndexer.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/idx/IndexWriter.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexChainEntry.java|7|21|33.33%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|42|42|100.00%|
|:large_blue_circle:|io/questdb/cairo/SymbolColumnIndexer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexChainPicker.java|4|4|100.00%|

PR #7062, opened 2026-05-06
chore(core): fix another posting index bug leading to corrupt results

bluestreak01:
found by a fuzz test

PR #7071, opened 2026-05-07
fix(core): stop posting index from returning rows past sealed coverage

bluestreak01:
## Summary

- POSTING-indexed symbol lookups can surface row ids past what the chain entry has formally sealed, because readers were not consulting the entry's `MAX_VALUE` field.
- The change wires `entryScratch.maxValue` from `readIndexMetadataFromChain` into a new `AbstractPostingIndexReader.entryMaxValue` field. The reader-side row-id paths now clamp their effective upper bound to `min(callerMax, entryMaxValue)`: `PostingIndexFwdReader#getCursor`, `PostingIndexBwdReader#getCursor`, and `AbstractPostingIndexReader#collectDistinctKeysInRange` (used by the POSTING distinct-keys factory's per-partition scan).
- Implicit nulls (rows before `columnTop`) are left clamped by `columnTop` only, since they are independent of the index data. `NullCursor.size()` derives the null portion from the unclamped-derived `nullCount` field rather than recomputing from `Cursor.maxValue` (which now holds the clamped bound), so `size()` stays consistent with `hasNext()` even if `entryMaxValue` were ever below `columnTop`.

## Why

The v2 chain entry's `V2_ENTRY_OFFSET_MAX_VALUE` is documented as "Highest row id covered by this entry's index data" - the boundary between rows the writer has formally indexed and any dirty `(key, rowId)` pairs that can sit further along in `.pv` (for example, after an O3 split shrinks a partition before the next reseal evicts them). The writer maintains the field on every seal and sparse-gen append (`PostingIndexChainWriter#extendHead`, `#updateHeadMaxValue`). The reader did not consume it; cursors clamped only to `pageFrame.getPartitionHi() - 1` in production paths and to `Long.MAX_VALUE` from internal callers, so any `.pv` entry whose row id fell inside that wider window leaked through to the SQL layer regardless of whether the chain claimed to cover it.

This change closes that gap on the reader side. It is the conservative half of the fix: the dirty entries that sit past the chain's tracked coverage are now filtered. Dirty entries that sit *within* `[0, MAX_VALUE]` - for example a stale `(key=0, rowId=50)` left over from a prior generation after row 50 was overwritten with a non-null symbol - are not filtered by this change. Addressing those would require either a per-gen valid-range field (the entry header has 4032 reserved bytes in the page header, so format headroom exists) or an eager writer-side re-encode on overwrite, both of which are larger surgery and out of scope here.

## Tradeoffs and limitations

- The clamp narrows the effective upper bound. In production, cursor callers pass `pageFrame.getPartitionHi() - 1`, which should equal the chain entry's `MAX_VALUE` on a correctly maintained writer; the clamp only fires when they diverge. In tests that pass `Long.MAX_VALUE` the clamp does fire heavily and reduces returned row ids to the entry's tracked extent - this is the intended behaviour.
- The fix relies on the writer's `MAX_VALUE` being maintained correctly. If the writer ever sets `MAX_VALUE` lower than the actual `.pv` extent, the reader will under-report rather than over-report; that direction is safer than the pre-fix over-report but not strictly invisible. Investigating whether any writer path can leave `MAX_VALUE` stale-low is the natural follow-up.

## Test plan

- New deterministic regression test: `PostingIndexCriticalIssuesTest#testReaderClampsCursorToChainEntryMaxValue`. Writes 100 rowids across 5 keys to a fresh chain, commits, then calls `PostingIndexWriter#setMaxValue(49)` to lower the head entry's `V2_ENTRY_OFFSET_MAX_VALUE` in place via `updateHeadMaxValue` (which republishes the chain header without touching `valueMem` or the gen-dir, leaving the encoded gen with rowids 0..99). Both `PostingIndexFwdReader` and `PostingIndexBwdReader` are queried for key 0 with caller-supplied `max=Long.MAX_VALUE`. With the fix, `indexMaxValue = min(callerMax, entryMaxValue) = 49` and the cursor stops once the value exceeds 49. Without the fix the cursor leaks 10 rowids past 49.
- The test exercises the cursor clamp only. The symmetric clamps in `collectDistinctKeysInRange` and `NullCursor.size()` do not have dedicated regression tests in this PR; both reuse the same `entryMaxValue` field and follow the same `entryMaxValue >= 0 ? Math.min(...) : ...` pattern, so the regression surface is shared with the cursor clamp the test does cover.
- mvn -pl core -Dtest='PostingIndex*Test,CoveringIndexTest,BitmapIndexTest,SymbolMapTest,PostingIndexCriticalIssuesTest,CheckpointFuzzTest' test - 536 tests, all green.
- Watch the scheduled fuzz pipeline (build definition 25, "Fuzz Tests") for `testCheckpointEjectedWalApply` failure rate over the next several runs - if the bug is the past-MAX_VALUE shape this PR addresses, the rate should drop.

## Related

- Reader-side counterpart to the recent posting-index hardening commits #7062 and #7063.
- Background: `core/src/main/java/io/questdb/cairo/idx/POSTING_INDEX_CHAIN_DESIGN.md`, sections 3 and 4.3.


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 17 / 19 (89.47%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexFwdReader.java|3|4|75.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexBwdReader.java|6|7|85.71%|
|:large_blue_circle:|io/questdb/cairo/idx/AbstractPostingIndexReader.java|8|8|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 21 / 25 (84.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexFwdReader.java|3|5|60.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexBwdReader.java|6|8|75.00%|
|:large_blue_circle:|io/questdb/cairo/idx/AbstractPostingIndexReader.java|12|12|100.00%|

PR #7077, opened 2026-05-08
fix(core): fix incorrect indexed query results after O3 row-id reuse

mtopolnik:
## Summary

A range-replace, dedup-replace, or O3 split commit that reuses
partition row-ids while changing the value at that row-id could
leave stale `(key, rowId)` entries inside `[columnTop, partitionSize)`
in posting-indexed columns. Subsequent queries against those columns
returned phantom matches: e.g. a row with `sym2 != null` matching
`sym2 = null` because the chain still carried the prior null entry.

The seal sweep on each O3 commit now rebuilds the posting chain from
the column data file rather than relying on `rollbackConditionally`,
which only evicts entries past the new partition size. The rebuild
sequence is `discardForRebuild` (drop in-memory state without I/O) ->
re-add every row from the `.d` file -> `commit` -> `seal`. The new
chain entry supersedes any stale entries from prior commits; the old
`.pv` is rotated and queued for the seal-purge job.

## Tradeoff: extra read per indexed column per O3 seal

The fix performs a full read of the symbol column file (`.d`) plus a
per-row `toIndexKey + writer.add` for every indexed column during
each O3 partition seal, even when the prior `rollbackConditionally`
path would have been a no-op. For the covering branch this piggybacks
on a file open the seal sweep was already doing; for the non-covering
branch this is a new file read that did not happen before. Cost is
O(rows in partition x indexed columns) per O3 commit. No benchmark
numbers are included in this PR; users with large partitions and
many indexed columns may see a measurable seal-time increase.

File-open accounting is unchanged: still one new `.pv` per seal, no
extra `.pk` rewrite.

## Outstanding follow-ups

A self-review flagged a few items that did not make it into this PR.
None block correctness, but they would harden the fix:

- No deterministic regression test reproduces the original phantom-
  match scenario at the SQL level. Validation comes from the fuzz
  runs listed below plus the existing `PostingIndex*Test*` suite. A
  small SQL-level test that drives an O3 replace-range with
  overlapping row-ids and asserts no phantom matches would tighten
  the regression net.
- No targeted unit test exercises `discardForRebuild`'s state-
  preservation invariants (`valueMem`, `valueMemSize`, `keyMem`,
  `sealTxn`, chain head all preserved; pending/spill/genCount/
  keyCount all cleared). Future changes to the writer's in-memory
  representation could break those invariants without tripping a
  local failure -- only a flaky fuzz pass would surface it.
- Seal-time cost is acknowledged but not quantified.
- The four-line rebuild block (`discardForRebuild` -> open `.d` ->
  `index` -> close fd -> `commit`) is duplicated across the covering
  and non-covering branches in `sealPostingIndexForPartition`. A
  small private helper would document the operation as a unit and
  prevent the two paths from drifting.

## Test plan

- 30 sequential runs of `WalWriterFuzzTest.testWalWriteRollbackHeavy`
  (the failure being chased): all pass
- Full `WalWriterFuzzTest` suite (24 cases): all pass
- All 266 `PostingIndex*Test*` cases: pass
- `CheckpointFuzzTest`, `BitmapIndexTest`, `SymbolMapTest`: pass
- `PostingIndexCriticalIssuesTest#testReviewFindingC1` (which counts
  seal-time `.pv` opens) confirms file-open accounting is unchanged


bluestreak01 (commented):
@mtopolnik — review at level 2 (mission-critical pass available at level 3 if you want a wider sweep). I think C1 is a blocker; flagging C2/M2 for the same root cause.

## Critical

### C1. `extendHead` overwrites pre-existing chain head slots — protocol violation, breaks MVCC and crash recovery

**Files:**
- `core/src/main/java/io/questdb/cairo/TableWriter.java:11308-11320, 11381-11388` (covering and non-covering branches of `sealPostingIndexForPartition`)
- `core/src/main/java/io/questdb/cairo/idx/PostingIndexWriter.java:543-576` (`discardForRebuild`)
- `core/src/main/java/io/questdb/cairo/idx/PostingIndexWriter.java:2778-2864` (`publishToChain` / `extendHead` callsite)

**The bug.** `discardForRebuild` zeroes `genCount` / `maxValue` / `hasPendingData` in memory but deliberately preserves `sealTxn`, `valueMemSize`, and the chain head on disk. The subsequent `commit()` runs `flushAllPending()`, which sets `genCount=1` and calls `publishToChain(newGenCount=1, overrideGenIndex=0, …)`. Because `sealTxn == chain.getHeadSealTxn()`, `publishToChain` takes the `extendHead` branch (`PostingIndexWriter.java:2794`).

`extendHead` only assumes `newGenCount > oldGenCount` — its protocol writes a new gen-dir slot at index `oldGenCount` (past the visible end), then bumps `GEN_COUNT` last. The PR violates this: it writes gen-dir slot 0 (`PostingIndexWriter.java:2797-2802`) — already occupied by an old gen — and pulls `GEN_COUNT` from K (the previous head's gen count) down to 1 (`PostingIndexChainWriter.java:277`).

When `sealPostingIndexForPartition` runs, the partition's chain head can have `GEN_COUNT > 1`: every fd-based O3 sparse-gen append took `extendHead` and bumped `GEN_COUNT`. So the case where the discard rebuild is most needed is exactly the case where `oldGenCount > 1`.

**Three concrete failure modes.**

**1) Torn read in concurrent reader.** Between `publishToChain` writing slot 0 dir bytes and `extendHead` storing the new `GEN_COUNT`, a reader running `PostingIndexChainEntry.read` (`PostingIndexChainEntry.java:115-116`) can latch the OLD `GEN_COUNT=K`, `loadFence`, then iterate slots `0..K-1`. Slot 0 has been overwritten with NEW (post-O3) gen-dir bytes, slots `1..K-1` still point at OLD gens. The reader sees a mixed-gen state — exactly the phantom-match shape the PR is trying to remove. The PR's doc comment claims "the chain header's seqlock plus extendHead's GEN_COUNT-last ordering rule out torn reads"; that protection works for `newGenCount > oldGenCount` (no slot < `oldGenCount` is touched), but not for `newGenCount < oldGenCount`.

**2) Old-snapshot reader corruption (persists after seal).** After the trailing `seal()` appends a NEW chain entry at a fresh `sealTxn`, the OLD entry becomes a "prev" entry. Its bytes have been mutated in place: `MAX_VALUE` (via `setMaxValue` inside `index()` → `chain.updateHeadMaxValue` at `PostingIndexChainWriter.java:565`), `KEY_COUNT`, `LEN`, `VALUE_MEM_SIZE`, slot 0 dir, and `GEN_COUNT=1`. Slot 0's `FILE_OFFSET` now points at new data appended to the OLD `.pv.{S}` file (at the preserved `valueMemSize`). The OLD entry's `txnAtSeal` is unchanged.

A reader pinned to a transaction `T` with `T_partition_seal ≤ T < new_committed_txn` walks the chain via `PostingIndexChainPicker.read` (`PostingIndexChainPicker.java:146-156`) and stops at the OLD entry (`txnAtSeal ≤ T`). They observe post-O3 partition state through a chain entry that should reflect pre-O3. This is an MVCC snapshot-isolation violation that persists until the OLD `.pv.{S}` is purged by the seal-purge job.

**3) Crash recovery hole.** `recoveryDropAbandoned` (`PostingIndexChainWriter.java:462-528`) only drops chain entries with `txnAtSeal > currentTableTxn`. The OLD entry's `txnAtSeal` is unchanged by `extendHead`, so it survives recovery — but its bytes are mutated. If the writer crashes between `commit()` and `seal()` (or between `seal()` of column N and the txWriter commit of `finishO3Commit`), recovery drops new entries with `txnAtSeal` beyond the committed txn but does not undo the in-place mutation of the OLD head. Post-recovery readers see the post-O3 mutated state under the OLD entry's pre-O3 `txnAtSeal` tag, and the data file (`.d`) at that pinned txn no longer matches.

**4) Cover footer is left scrambled (covering branch).** In the covering branch, `commit()` runs *before* `configureCovering()` (lines 11320 vs 11327). `writer.of()` inside `configureFollowerAndWriter` calls `close()` which zeroes `coverCount`, so during this `commit()` `coverCount==0`. `captureCoverEndOffsets` short-circuits (`PostingIndexWriter.java:1651-1653`), `extendHead` writes `LEN = entrySize(genCount=1, coverCount=0)`. The OLD entry's actual cover footer (at `header + oldGenCount*GEN_DIR_SIZE`) is left in place. A reader latching new `GEN_COUNT=1` resolves the footer offset to `header + 1*GEN_DIR_SIZE` (`PostingIndexChainEntry.java:159-162`) — which is now reading OLD slot 1's gen-dir bytes as cover end-offsets. The self-bound check at `PostingIndexChainEntry.java:139-140` likely zeroes them out (because the new `LEN` shrank), so covering returns empty for that entry. That is "fail safe" rather than crashing, but it silently disables covering reads through the prev entry until purge.

**Why the pre-PR code didn't have this.** The OLD `rollbackConditionally(partitionSize)` either no-op'd (when `getMaxValue() < partitionSize`) or rotated to a fresh `.pv` at `peekNextSealTxn()` via `reencodeAllGenerations`, which makes the next `publishToChain` take the `appendNewEntry` branch (`sealTxn != chain.getHeadSealTxn()`) — a clean new entry, no in-place mutation of an older entry. The PR replaces this with in-place mutation of the existing head.

**Suggested fix.** `discardForRebuild` should rotate `sealTxn` (e.g., `sealTxn = peekNextSealTxn()` plus `closeSidecarMems()` + reopen `valueMem` at the new `sealTxn` like `truncate()` does at `PostingIndexWriter.java:1380-1391`) so the subsequent `commit()` takes the `appendNewEntry` branch. Or, add a special case in `publishToChain` when `newGenCount < oldGenCount` to force `appendNewEntry` rather than `extendHead`. The current "extendHead with newGenCount=1" path silently violates the `extendHead` contract.

The fuzz tests in the PR's test plan don't exercise concurrent readers with pinned older txns walking back through a partition that was just resealed by an O3 commit, which is why they pass despite this hole.

---

### C2. `index()` calls `setMaxValue` against a chain head whose on-disk `MAX_VALUE` is stale

`SymbolColumnIndexer.index` ends with `writer.setMaxValue(hiRow - 1)` (`SymbolColumnIndexer.java:209`). `setMaxValue` not only sets the in-memory field, it also writes `MAX_VALUE` to the chain head and republishes the chain header (`PostingIndexWriter.java:1322` → `PostingIndexChainWriter.updateHeadMaxValue`). After `discardForRebuild` (which preserves the chain head on disk but resets in-memory `genCount` / `keyCount` to zero), this mutates a head whose on-disk `GEN_COUNT`, `KEY_COUNT`, `VALUE_MEM_SIZE`, etc. still describe the OLD state.

A reader picking up the chain header republished by `updateHeadMaxValue` and then walking to the head entry sees: `MAX_VALUE` = NEW (= `partitionSize - 1`), all other fields = OLD. The combination is internally inconsistent (e.g. `MAX_VALUE` exceeds the rowId range any OLD slot covers).

This compounds C1's race and is a second protocol violation introduced by the PR even before `commit()` runs.

---

## Moderate

### M1. Performance regression — full `.d` re-read per indexed column per O3 commit, even when unnecessary

The PR description acknowledges this. To quantify in concrete terms:

- **Old path:** `rollbackConditionally(partitionSize)` was a true no-op when `getMaxValue() < partitionSize` (`PostingIndexWriter.java:1128-1130`). For most O3 commits that didn't reuse rowids, this was free.
- **New path:** unconditional `discardForRebuild` + full read of `<col>.d` + `add()` per row + `commit()` (which also runs `sync()` unless `NOSYNC` per `PostingIndexWriter.java:404-407`). Cost = `O(rows_in_partition × indexed_columns)` per O3 commit, plus the in-memory work of re-encoding gens, plus the keyMem/valueMem `sync` from `commit()`.

For a workload doing O3 inserts into partitions with millions of rows on multiple indexed symbol columns, this turns "O3 seal" from a near-instant chain rotation into a full per-partition reindex. No benchmark numbers are provided. Suggest at minimum gating the rebuild on a "this commit could have reused rowids" signal (replace-range, dedup-replace, O3 split — these are knowable from the sink block).

### M2. `commit()` runs before `configureCovering()` in the covering branch — cover footer drops on the mutated entry

See C1 failure mode (4). Even setting aside the corruption concern, this is a structural inversion: rebuilding sidecars expects the cover schema to be configured first. The OLD code's `rollbackConditionally` was a writer-only operation that didn't republish the chain entry, so the ordering didn't matter. The new code republishes via `extendHead` while covering is in a half-configured state. Move `discardForRebuild` + `index` + `commit` to *after* `configureCovering` + `setCoveredColumnNameTxns`, or rotate `sealTxn` in `discardForRebuild` so the `commit()` produces a fresh entry that gets superseded cleanly.

### M3. Missing regression test

The PR description explicitly lists this as an outstanding follow-up: "No deterministic regression test reproduces the original phantom-match scenario at the SQL level." For a fix that targets a wrong-results bug in O3 row-id reuse, a deterministic SQL-level test that fails on master and passes on this branch is required, not optional. Fuzz coverage has been demonstrated to occasionally miss this exact class of bug (the original ec4b2adb36 fast-lag commit/seal change). A concrete test that replays a replace-range or dedup-replace touching an indexed symbol column and asserts no phantom matches must land with this PR.

### M4. Missing targeted test for `discardForRebuild`'s state-preservation invariants

PR description acknowledges this. The list of fields the method must preserve (`valueMem`, `valueMemSize`, `keyMem`, `sealTxn`, chain head) versus clear (`pending`, `spill`, `genCount`, `keyCount`, `keyCapacity`, `maxValue`, `hasPendingData`, `activeKeyCount`) is encoded only in the method body and the comment. Future refactors of writer in-memory layout will silently break this without a unit test catching it; the only signal would be flaky fuzz runs.

---

## Minor

### m1. Duplicated rebuild block

Acknowledged in PR description. The five-line `discardForRebuild` → `openRO(dFile)` → `index` → `close fd` → `commit` block is identical between the covering and non-covering branches at `TableWriter.java:11308-11320` and `11381-11388`. A private helper avoids drift.

### m2. `discardForRebuild` doc comment overstates safety

The Javadoc on `IndexWriter.discardForRebuild` (`IndexWriter.java:90-119`) says "The subsequent `commit()` / `seal()` writes use the chain's established in-place mutation and rotation protocols and are safe under the same seqlock guarantees as a normal flush." Per C1, the in-place protocol guarantees do not extend to `newGenCount < oldGenCount`. Either fix the protocol (preferred) or update the doc to state precisely which seqlock guarantees apply and which don't.

### m3. `setNextTxnAtSeal` ordering inconsistency between the two branches

Covering branch: `setNextTxnAtSeal(txWriter.getTxn())` is at line 11346 — *after* `commit()`. Non-covering branch: `setNextTxnAtSeal(txWriter.getTxn())` is at line 11378 — *before* `discardForRebuild` and `commit()`. Both end up correct because `pendingTxnAtSeal` is only consumed when `appendNewEntry` runs (i.e., during `seal()`), and `commit()` here takes the `extendHead` branch which doesn't use it. But the asymmetry is misleading and future-fragile; line up the two branches.

### m4. Bare integer literal `4` in `index()` for symbol-column row size

Pre-existing in `SymbolColumnIndexer.java:193-205`, not introduced by this PR, but the new code path now exercises this read on every O3 seal. While here, consider a named constant — `Integer.BYTES` or `SYMBOL_KEY_SIZE`. Minor.

---

## Summary

The fix's intent (rebuild from `.d` rather than relying on `rollbackConditionally`) is correct: `rollbackConditionally` cannot detect stale `(key, rowId)` pairs inside `[columnTop, partitionSize)` because rowIds don't exceed `partitionSize`. But the implementation reuses the `extendHead` chain protocol in a configuration it was never designed for (`newGenCount < oldGenCount`), introducing in-place mutation of an existing chain entry. That mutation is observable to concurrent readers, persists past the trailing seal as a corrupt prev entry, and is not undone by crash recovery. C1 is a regression in MVCC snapshot consistency — strictly worse than the bug being fixed in some scenarios.

Recommended path forward: change `discardForRebuild` to rotate `sealTxn` (mirror the relevant subset of `truncate()`'s behaviour without recreating the `.pv` file content), so the subsequent `commit()` takes the `appendNewEntry` branch and produces a clean new chain entry rather than mutating the head. Then add a deterministic SQL test that replays an O3 row-id-reuse and asserts no phantom matches under concurrent readers, plus a unit test for `discardForRebuild`'s invariants.

bluestreak01 (commented):
@mtopolnik — re-reviewed `7bba017693`. All 2 critical and all 4 moderate findings from the previous review are resolved. Notes inline so the reasoning is on the record.

## C1 — chain-head in-place mutation: fixed

`discardForRebuild` (`PostingIndexWriter.java:596-617`) now mirrors the rotation block from `truncate()`: captures `sealTxn` into the new `pendingDiscardOldSealTxn` field, computes `newSealTxn = max(1, chain.peekNextSealTxn())`, closes `valueMem` and sidecar mappings, opens `valueMem` against `valueFileName(p, indexName, postingColumnNameTxn, newSealTxn)`, and resets `valueMemSize = 0`. The chain head and OLD `.pv` stay on disk for concurrent readers.

The next `commit()` reaches `publishToChain` with `this.sealTxn != chain.getHeadSealTxn()`, so it takes the `appendNewEntry` branch instead of `extendHead`. The protocol-violating `extendHead(newGenCount=1)` against an `oldGenCount > 1` head no longer happens. The OLD `.pv`'s purge is deferred via `pendingDiscardOldSealTxn`, consumed in `flushAllPending` right after `publishToChain` so `recordPostingSealPurge` sees the post-append chain shape `[REBUILD@new, OLD@old]` and computes a correct `[OLD.txnAtSeal, REBUILD.txnAtSeal)` window.

`testReviewFindingC1_DiscardForRebuildMutatesAlreadyPublishedHeadInPlace` is a tight regression: it builds a 3-gen pre-rebuild head, replays `discardForRebuild + add + commit`, and asserts at least one of `entryCount grew / sealTxn rotated / txnAtSeal advanced / genCount preserved`. Under the fix, three of the four hold.

Edge cases checked:
- Empty chain (`!chain.hasHead()`): rotation is skipped; `appendNewEntry` fires naturally because `chain.hasHead()` is false. Safe.
- Fd-based writer (`partitionPath.size() == 0`): rotation is skipped. The only production caller (`sealPostingIndexForPartition`) is path-based, so this path is unreachable from production — acknowledged in the new comment.

## C2 — `setMaxValue` mutates stale chain head: fixed

`setMaxValue` (`PostingIndexWriter.java:1361-1382`) early-returns from `chain.updateHeadMaxValue` when `pendingDiscardOldSealTxn >= 0`. The upcoming `flushAllPending` takes `appendNewEntry` (rotated sealTxn) and stamps `maxValue` into the FRESH entry's header from the in-memory mirror, so skipping the on-disk republish keeps the OLD entry's `MAX_VALUE` consistent with its other preserved fields.

`testReviewFindingC2_SetMaxValueAfterDiscardForRebuildMutatesPrevEntry` confirms post-rebuild that the OLD entry (now prev) retains its original `MAX_VALUE` while all sibling fields match the pre-rebuild snapshot.

## M1 — full `.d` re-read regression: fixed

`sealPostingIndexForPartition` gains a `canSkipRebuild` parameter (line 11241). The caller computes it from the o3 sink block:

```
canSkipRebuildForPartition = !partitionMutates
        && o3SplitPartitionSize == 0
        && newPartitionSize >= oldPartitionSize;
```

Sink offsets match the existing reader at `TableWriter.java:7257-7262`. Pure-append O3 partitions take the cheap `rollbackConditionally(partitionSize)` path, avoiding the per-row `.d` re-read. The `dataPartitionTimestamp` callsite passes `false` unconditionally, which is correct because a split parent shrunk and may have rewritten rowids inside the new size.

## M2 — `commit()` before `configureCovering()` cover-footer drop: fixed (different mechanism than I suggested; the txnAtSeal hoist works)

The covering branch hoists `setNextTxnAtSeal(txWriter.getTxn() + 1L)` before `discardForRebuild` (line 11302). The intermediate REBUILD entry is tagged out of every current reader's visibility window (`T_pin <= getTxn() < getTxn() + 1`); the picker walks past it to OLD (whose footer is intact) until `rebuildSidecars` publishes SEAL with `txnAtSeal = getTxn()` and a real footer. `recoveryDropAbandoned` can also drop REBUILD on a partial-publish reopen.

The non-covering branch keeps `setNextTxnAtSeal(getTxn())` (no `+1L`): it has no covering footer to lose, so a brief-window picker hit returns the posting index alone, which is correct. Comment captures the asymmetry.

`testReviewFindingM2_RebuildCommitTagsEntryWithMeaningfulTxnAtSeal` validates the writer-API contract.

## M3 — missing regression test: partially addressed, acceptable

Three writer-fixture regression tests added: `testReviewFindingC1_…`, `testReviewFindingC2_…`, `testReviewFindingM2_…`. No SQL-level phantom-match test was added.

I think the writer-fixture coverage is more discriminating than a SQL-level test for this specific protocol violation — they fail loudly on any regression of `extendHead`'s `newGenCount > oldGenCount` precondition, the `setMaxValue` republish, or the M2 hoist. Combined with the existing fuzz coverage (`WalWriterFuzzTest.testWalWriteRollbackHeavy` was the original signal that found the bug), I would treat M3 as resolved. A SQL-level smoke for the full O3 → seal → query loop on top is still worthwhile but not a blocker.

## M4 — missing unit test for `discardForRebuild` invariants: addressed

The new C1/C2/M2 tests collectively pin the preserved-vs-cleared invariants: chain head preservation, `sealTxn` rotation, `MAX_VALUE` preservation in prev, `txnAtSeal` propagation. A future refactor that breaks `discardForRebuild`'s contract will trip at least one.

## One thing to update in the PR description

The PR body still says "File-open accounting is unchanged: still one new `.pv` per seal, no extra `.pk` rewrite." With the fix, the rebuild path opens an additional `.pv` per indexed column per O3 commit (one in `discardForRebuild`, one in `seal`/`rebuildSidecars`). The new `canSkipRebuild` fast-path keeps accounting unchanged for pure-append commits, but mutating O3 commits add one `.pv` open per indexed column. Worth tightening that line so it isn't read as a post-fix invariant.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 53 / 55 (96.36%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/IndexWriter.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|31|32|96.88%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|22|22|100.00%|

bluestreak01 (approved):

PR #7075, opened 2026-05-08
chore(core): switch posting index fast-lag from seal to commit

kafka1991:
## Summary

WAL fast-lag commits to posting-indexed columns now publish via `commit()` instead of `seal()`. The previous path rewrote the whole posting index to a new `.pv.<sealTxn>` file per commit; the new path appends a sparse generation to the active `.pv` and defers compaction to `sealIfMultiGen(16)` on partition retirement in `switchPartition`, or to `flushAllPending`'s built-in auto-seal at `MAX_GEN_COUNT` (143).

`PostingIndexWriter.seal()` also now syncs `.pk` after the chain publish, closing a power-loss gap where readers built around a new sealTxn could observe an unflushed chain page. `O3PartitionJob` branches by index type so BITMAP keeps using `commit()` (its only sync point) and POSTING calls `seal()` directly, now fully self-contained for durability.

## Performance

JMH `PostingIndexBenchmarkSuite.walFastLagInsert` (avgt, ms/op):

| batchRows | keyCount | indexType        | before | after  |
|-----------|----------|------------------|--------|--------|
| 100       | 50       | posting          | 5.145  | 0.762  |
| 100       | 100000   | posting          | 11.219 | 4.499  |
| 1000      | 50       | posting          | 8.485  | 1.140  |
| 1000      | 100000   | posting          | 20.969 | 7.072  |
| 10000     | 50       | posting          | 23.245 | 4.324  |
| 10000     | 100000   | posting          | 63.258 | 23.857 |
| 100       | 50       | posting_covering | 5.405  | 0.723  |
| 1000      | 100000   | posting_covering | 20.796 | 7.317  |
| 10000     | 100000   | posting_covering | 62.676 | 23.639 |

Speedup is 2.5x at the largest batch + cardinality and 7.5x at the smallest: seal's fixed per-commit overhead (file allocation + several fsyncs) dominates when each commit's useful work is small. Covering and non-covering branches improve at similar ratios. `writeInsert` (steady-state ingestion, non fast-lag) and `commitProfile` (forced seal per cycle) are within noise of pre-change values.

## Tradeoffs

- The active partition can accumulate up to 142 sparse generations before auto-seal compacts. Reads against the active partition scan more generations on average until then; old partitions still compact at 16 generations on `switchPartition`.
- `seal()` now performs one extra `.pk` fsync per call. The fsync runs unconditionally, matching the existing seal-internal `sealValueMem` / sidecar syncs that already ignore commit mode.

PR #7080, opened 2026-05-09
fix(core): prevent OOM during ALTER ADD INDEX on large partitions

nwoolmer:
## Summary

`ALTER TABLE ... ALTER COLUMN ... ADD INDEX TYPE POSTING [INCLUDE ...]` on a multi-billion-row partition currently OOMs at `PostingIndexWriter.spillKey`'s per-key buffer doubling, with the JVM's `Unsafe.checkAllocLimit` reporting `global RSS memory limit exceeded` and `memoryTag=NATIVE_INDEX_READER`. The same vulnerability applies to `IndexBuilder` (REINDEX), `TableSnapshotRestore`, and -- once #7077 lands -- every O3 partition rewrite on a posting-indexed wide column.

This PR introduces two cooperating mechanisms that bound peak heap usage on those paths, plus four fuzz tests exercising the new code.

### Indexing-phase back-pressure

A self-imposed memory budget (`cairo.posting.index.indexer.spill.bytes.max`, default 256 MiB) tracks bytes held in the per-key spill arena. When `spillKey` would push the running tally past the budget, the writer drains pending+spill into a fresh sparse generation in `valueMem` and frees the spill arena. The pending arrays stay live across the trigger so the in-flight `add` call can complete its post-`spillKey` write without dereferencing a freed pointer, except for the case where `flushAllPending`'s inline-seal-at-`MAX_GEN_COUNT` frees them too -- handled by reallocating before returning.

### Seal-phase streaming fallback

A pre-flight check at the top of `reencodeAllGenerations`'s seal branch picks between three options based on RSS headroom:

- Fast path (existing `reencodeWithStrideDecoding`): peak bounded by the largest single stride's aggregate row count.
- Streaming path (new `reencodeWithPerKeyStreaming`): peak bounded by the largest single key's row count. Decodes one key at a time from every source generation, encodes directly into `sealTarget`, always emits DELTA-mode strides. Output format is byte-compatible with the fast path so readers cannot distinguish the two.
- Hard fail with an actionable diagnostic if even streaming would exceed the limit.

Fixed-size `INCLUDE` columns (DOUBLE, FLOAT, LONG, etc.) are fully supported in the streaming path via a per-key sidecar writer that reuses the same ALP compression and on-disk layout as the fast path. Variable-size `INCLUDE` columns (STRING, VARCHAR, BINARY) refuse with a clear error; lifting that restriction would need a two-pass write to materialise per-stride offset tables and is left for a follow-up.

### Defensive cleanups along the way

- Three doubling sites (`spillKey`, the merge-spill grow inside `flushAllPending`, `growKeyBuffers`) computed `curCap * 2` in int and silently lost the doubling-amortisation invariant once `curCap >= 2^30`. Now in long arithmetic with an explicit `Integer.MAX_VALUE` clamp and a `CairoException` on exceedance.
- Phase 1's `strideTotal` accumulator was int and could wrap negative when compaction merged multiple gens into a single 256-key stride past `Integer.MAX_VALUE` aggregate. Now long with an explicit `posting index seal stride aggregate exceeds 2^31` error at the boundary.
- `strideValCount` in the streaming reencode and streaming sidecar paths was an int sum that could wrap to 0 on heavily-compacted partitions, silently skipping non-empty strides; replaced with a `boolean hasAnyValues` flag since the sum was only used for an empty-stride skip.

## Effect on existing behaviour

- The default 256 MiB budget is large enough that no existing test trips the back-pressure trigger; commit-time WAL ingestion and the per-commit deltas seen in normal operation stay on exactly the same code path as before.
- `PostingIndexStressTest` (159), `PostingIndexCriticalIssuesTest` (13), `PostingIndexConcurrencyTest` (12), `IndexBuilderTest` (23), and `CoveringIndexTest` (1038) all pass without modification, confirming no behaviour change at the default budget.
- `seal()` now performs more allocations during compaction estimation, but those are gated by `Unsafe.getRssMemUsed` reads (cheap) and a small handful of long arithmetic operations. No measurable hot-path impact.

## Tradeoffs

- Streaming compaction is roughly 2-3x slower than the fast path on the same data because it cannot batch decode across keys and forgoes FLAT-mode encoding (which can be denser when the per-stride min/max range is small). It only kicks in when the fast path would not run at all -- a strict win over the alternative of OOM.
- Variable-size `INCLUDE` columns hit the hard-fail branch when the partition's per-stride aggregate exceeds the RSS headroom and streaming is selected. This regression is bounded to a narrow case (large partition + var-size INCLUDE + tight RSS); the message tells the operator to drop the var-size INCLUDE, reduce partition size, or raise `RSS_MEM_LIMIT`. A follow-up will lift this restriction.
- One new configuration knob. Setting it to 0 or negative disables the back-pressure entirely and recovers the legacy "accumulate until seal" behaviour.

## Compatibility with other open posting-index PRs

- **#7070** (cover-footer SEGV fix): adds `releaseCoveredColumnReadMappings` and reorders members in `PostingIndexWriter`. No semantic conflict; member ordering rebases mechanically.
- **#7075** (WAL fast-lag commit-based): switches WAL fast-lag to `commit()` so the active partition can carry up to `MAX_GEN_COUNT = 143` sparse generations. The default 256 MiB budget never trips during a single WAL commit batch; the streaming fallback handles the eventual auto-seal at `MAX_GEN_COUNT` consistently. No interaction.
- **#7077** (O3 row-id reuse fix): every O3 seal now drives `discardForRebuild` + the full row-by-row `indexer.index` loop. Without this PR's back-pressure, every O3 commit on a posting-indexed wide partition hits the same OOM as the user-reported `ALTER ADD INDEX`. With this PR landed first, that vulnerability is closed automatically.

## Test plan

- `PostingIndexOomFallbackTest` (new, 9 cases):
  - `testIndexerPeriodicFlushUnderBudget`: 64 KiB budget over 50 keys * 4096 rows; asserts genCount > 1 before seal, peak `NATIVE_INDEX_READER` within ~10x the budget, and round-trip correctness.
  - `testIndexerSurvivesMaxGenCountInlineSeal`: 64-byte budget forces hundreds of periodic flushes, exercising the inline-seal-frees-pending recovery path.
  - `testSealStreamingFallbackTriggers`: builds the same workload twice (fast path baseline + tight-RSS streaming-forced) and asserts byte-identical row-id read-back.
  - `testSealHardLimitWhenEvenStreamingDoesNotFit`: single hot key + 256 KiB headroom; asserts the explicit "would exceed RSS limit even with streaming compaction" error.
  - `testSealStreamingVarIncludeFails`: BINARY INCLUDE under tight RSS; asserts the var-size guard fires with the explicit remediation message.
  - `testFuzzPeriodicFlushAcrossBudgets`: 16 random workloads with budgets spanning 5 orders of magnitude; asserts both forward and backward cursors round-trip every (key, rowId) pair.
  - `testFuzzStreamingMatchesFastPathBaseline`: 12 random workloads built twice, asserts streaming and fast-path read-backs are byte-identical.
  - `testFuzzMixedBudgetAndRssAtSeal`: 24 random workloads with random budgets * random RSS limits; asserts each run either succeeds with correct round-trip or fails with a known diagnostic. Sanity-asserts that across the seeds we observe both successes and hard-fails.
  - `testFuzzStreamingFixedIncludeMatchesBaseline`: 8 random workloads with a DOUBLE INCLUDE column built twice; asserts covered DOUBLE values read via the covering cursor are byte-identical between fast path and streaming.
- `RssMemoryLimitTest.testAlterAddPostingIndexUnderRssLimit`: end-to-end SQL test, 96 MiB RSS limit, 1 MiB spill budget, 200_000 row WAL table, full `ALTER TABLE ... ALTER COLUMN sym ADD INDEX TYPE POSTING` round-trip.
- Regression sweep across `PostingIndex*`, `IndexBuilder*`, `Covering*`, `RssMem*` (1358 tests): all pass (1 pre-existing skip).

### DIFF
--- a/core/src/main/java/io/questdb/PropServerConfiguration.java
+++ b/core/src/main/java/io/questdb/PropServerConfiguration.java
@@ -406,6 +406,7 @@ public class PropServerConfiguration implements ServerConfiguration {
     private final boolean posthogEnabled;
     private final boolean postingIndexAutoIncludeTimestamp;
     private final byte postingIndexRowIdEncoding;
+    private final long postingIndexerSpillBytesMax;
     private final int postingSealGenThreshold;
     private final int postingSealPurgeOutboxMax;
     private final int preferencesStringPoolCapacity;
@@ -1568,6 +1569,7 @@ public class PropServerConfiguration implements ServerConfiguration {
                 case "delta" -> PostingIndexUtils.ENCODING_DELTA;
                 default -> PostingIndexUtils.ENCODING_ADAPTIVE;
             };
+            this.postingIndexerSpillBytesMax = getLongSize(properties, env, PropertyKey.CAIRO_POSTING_INDEX_INDEXER_SPILL_BYTES_MAX, 256L << 20);
             this.postingSealGenThreshold = getInt(properties, env, PropertyKey.CAIRO_POSTING_SEAL_GEN_THRESHOLD, 16);
             this.postingSealPurgeOutboxMax = getInt(properties, env, PropertyKey.CAIRO_POSTING_SEAL_PURGE_OUTBOX_MAX, 8192);
             this.sqlJoinMetadataPageSize = getIntSize(properties, env, PropertyKey.CAIRO_SQL_JOIN_METADATA_PAGE_SIZE, 16384);
@@ -4168,6 +4170,11 @@ public class PropServerConfiguration implements ServerConfiguration {
         @Override
         public int getPoolSegmentSize() {
             return poolSegmentSize;
+        }
+
+        @Override
+        public long getPostingIndexerSpillBytesMax() {
+            return postingIndexerSpillBytesMax;
         }
 
         @Override
--- a/core/src/main/java/io/questdb/PropertyKey.java
+++ b/core/src/main/java/io/questdb/PropertyKey.java
@@ -101,6 +101,7 @@ public enum PropertyKey implements ConfigPropertyKey {
     CAIRO_WORK_STEAL_TIMEOUT_NANOS("cairo.work.steal.timeout.nanos"),
     CAIRO_PARALLEL_INDEXING_ENABLED("cairo.parallel.indexing.enabled"),
     CAIRO_POSTING_INDEX_AUTO_INCLUDE_TIMESTAMP("cairo.posting.index.auto.include.timestamp"),
+    CAIRO_POSTING_INDEX_INDEXER_SPILL_BYTES_MAX("cairo.posting.index.indexer.spill.bytes.max"),
     CAIRO_POSTING_INDEX_ROW_ID_ENCODING("cairo.posting.index.row.id.encoding"),
     CAIRO_PAGE_FRAME_REDUCE_QUEUE_CAPACITY("cairo.page.frame.reduce.queue.capacity"),
     CAIRO_PAGE_FRAME_ROWID_LIST_CAPACITY("cairo.page.frame.rowid.list.capacity"),
--- a/core/src/main/java/io/questdb/cairo/CairoConfiguration.java
+++ b/core/src/main/java/io/questdb/cairo/CairoConfiguration.java
@@ -501,6 +501,18 @@ public interface CairoConfiguration {
         return 0.0;
     }
 
+    /**
+     * Maximum bytes the posting index writer's per-key spill buffers may hold
+     * before it triggers a mid-stream {@code flushAllPending} + free cycle to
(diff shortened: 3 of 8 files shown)


bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 356 / 431 (82.60%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|351|426|82.39%|

PR #7084, opened 2026-05-10
perf(sql): parallelise low-precision approx_percentile for LONG arg

mcadariu:
Follow-up from https://github.com/questdb/questdb/pull/6502, replaces heap-based `HdrHistogram` with off-heap `GroupByHistogram` in `ApproxPercentileLongGroupByFunction`.


mcadariu:
@coderabbit please review this PR.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/ApproxPercentileLongGroupByFunction.java
```diff
@@ -165,18 +146,43 @@ public boolean isThreadSafe() {
...
+    public void merge(MapValue destValue, MapValue srcValue) {
+        long srcPtr = srcValue.getLong(valueIndex);
+        if (srcPtr == 0) {
+            return;
+        }
+
+        long destPtr = destValue.getLong(valueIndex);
+        if (destPtr == 0) {
+            destValue.putLong(valueIndex, srcPtr);
+            return;
```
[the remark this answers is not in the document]
mcadariu: Done.

mcadariu:
@puzpuzpuz this is ready for review when you have a spare moment.

puzpuzpuz:
> @puzpuzpuz this is ready for review when you have a spare moment.

@mcadariu sure, it's on my radar. I'll review it tomorrow, if that's fine.

core/src/test/java/io/questdb/test/cairo/fuzz/ParallelGroupByFuzzTest.java
```diff
@@ -472,6 +472,108 @@ public void testParallelCountDistinctFuzz() throws Exception {
         });
     }
 
+    @Test
+    public void testParallelApproxPercentileFuzz() throws Exception {
```
puzpuzpuz: nit: this and `testParallelApproxPercentileFuzzWithNulls` tests are not fuzz tests and they're fully deterministic and there is no randomization. Thus, test names are misleading. Maybe it's worth it turning these into actual fuzz tests?

core/src/main/java/io/questdb/griffin/engine/functions/groupby/ApproxPercentileLongGroupByFunction.java
```diff
@@ -165,18 +150,37 @@ public boolean isThreadSafe() {
...
+    }
+
+    @Override
+    public void setAllocator(GroupByAllocator allocator) {
+        histogramA.setAllocator(allocator);
+        histogramB.setAllocator(allocator);
+    }
+
     @Override
     public void setEmpty(MapValue mapValue) {
```
puzpuzpuz: nit: `setEmpty` is now identical to `setNull`, so we could remove it. The default impl calls `setNull`.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/ApproxPercentileLongGroupByFunction.java
```diff
@@ -35,72 +35,64 @@
...
-        histogramIndex = 0;
+        histogramA.clear();
+        histogramB.clear();
     }
 
     @Override
     public void computeFirst(MapValue mapValue, Record record, long rowId) {
-        final Histogram histogram;
-        if (histograms.size() <= histogramIndex) {
-            // We pre-size the histogram for [1, 1000] range to avoid resizes in some basic use cases
```
puzpuzpuz, line 72: It looks like we no longer pre-size the histogram for the [1, 1000] range. Is that on purpose?

core/src/main/java/io/questdb/griffin/engine/functions/groupby/ApproxPercentileLongGroupByFunction.java
```diff
@@ -165,18 +150,37 @@ public boolean isThreadSafe() {
...
     }
 
+    @Override
+    public void merge(MapValue destValue, MapValue srcValue) {
+        long srcPtr = srcValue.getLong(valueIndex);
+        if (srcPtr == 0) {
+            return;
+        }
+
+        histogramA.of(destValue.getLong(valueIndex));
```
puzpuzpuz: What if `destValue.getLong(valueIndex)` is `0`? We could check this case and simply copy src hisrogram into dest value (see how it's done in `ApproxCountDistinctLongGroupByFunction`).

core/src/test/java/io/questdb/test/cairo/fuzz/ParallelGroupByFuzzTest.java
mcadariu: Right! Added some random elements. As well, I removed the `testParallelApproxPercentileFuzzWithNulls` as I found it a redundant after all.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/ApproxPercentileLongGroupByFunction.java
mcadariu: Good one, removed.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/ApproxPercentileLongGroupByFunction.java
mcadariu: Oops, I mistakenly thought it is done in the `GroupByHistogram`, fixed.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/ApproxPercentileLongGroupByFunction.java
mcadariu: Good one, done

core/src/test/java/io/questdb/test/cairo/fuzz/ParallelGroupByFuzzTest.java
```diff
@@ -2156,8 +2156,60 @@ public void testParallelNonKeyedGroupByDecimalFunctions() throws Exception {
     }
 
     @Test
-    public void testParallelNonKeyedGroupByFaultTolerance() throws Exception {
```
puzpuzpuz, line 2574: Could you restore `testParallelNonKeyedGroupByFaultTolerance`? This seems an unintended test deletion.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/ApproxPercentileLongGroupByFunction.java
```diff
@@ -35,72 +35,66 @@
 import io.questdb.griffin.engine.functions.BinaryFunction;
 import io.questdb.griffin.engine.functions.DoubleFunction;
 import io.questdb.griffin.engine.functions.GroupByFunction;
+import io.questdb.griffin.engine.groupby.GroupByAllocator;
+import io.questdb.griffin.engine.groupby.GroupByHistogram;
 import io.questdb.std.Numbers;
-import io.questdb.std.ObjList;
-import io.questdb.std.histogram.org.HdrHistogram.Histogram;
 
 public class ApproxPercentileLongGroupByFunction extends DoubleFunction implements GroupByFunction, BinaryFunction {
```
puzpuzpuz, line 42: We still have single-threaded `ApproxPercentileLongPackedGroupByFunction`, `ApproxPercentileDoubleGroupByFunction`, and `ApproxPercentileDoublePackedGroupByFunction` after this patch, so we should at least change the PR title to something like `perf(sql): parallelise low-precision approx_percentile for LONG arg` since it doesn't migrate all `approx_percentile` functions.

core/src/main/java/io/questdb/griffin/engine/functions/groupby/ApproxPercentileLongGroupByFunction.java
```diff
@@ -35,72 +35,66 @@
...
-        } else {
-            histogram = histograms.getQuick(histogramIndex);
-            histogram.reset();
-        }
-
         final long val = exprFunc.getLong(record);
         if (val != Numbers.LONG_NULL) {
-            histogram.recordValue(val);
+            histogramA.of(0);
+            histogramA.resize(1000);
```
puzpuzpuz: Could you keep the `We pre-size the histogram for [1, 1000] range...` comment we had in the old code?

core/src/main/java/io/questdb/griffin/engine/functions/groupby/ApproxPercentileLongGroupByFunction.java
```diff
@@ -35,72 +35,66 @@
...
     }
 
     @Override
     public void computeNext(MapValue mapValue, Record record, long rowId) {
-        final Histogram histogram = histograms.getQuick(mapValue.getInt(valueIndex));
         final long val = exprFunc.getLong(record);
         if (val != Numbers.LONG_NULL) {
-            histogram.recordValue(val);
+            long ptr = mapValue.getLong(valueIndex);
+            histogramA.of(ptr).recordValue(val);
```
puzpuzpuz, line 84: In case of zero address here we may need to do `histogramA.resize(1000)` to do the same thing as what `computeFirst()` does. Alternatively, we could add default value fields for `countsArrayLength`, `bucketCount`, `highestTrackableValue` in the flyweight and use them to reset the fields to their default state in the `of(0)` call case. This way, we won't need to make `resize()` public and call it additionally.

puzpuzpuz:
I did an ad-hoc benchmark on my 12c/24t box. Here are the results:
```sql
CREATE TABLE approx_pct_bench AS (
    SELECT
        timestamp_sequence('2024-01-01T00:00:00.000000Z', 1L) AS ts,
        -- Narrow latency-like distribution, no NULLs
        (abs(rnd_long()) % 1000) + 1                                  AS latency_us,
        -- Wide range, forces autoResize beyond the pre-sized 1000
        (abs(rnd_long()) % 1_000_000_000) + 1                         AS wide_val,
        -- 10% NULL on a narrow column
        CASE WHEN rnd_int() % 10 = 0 THEN NULL
            ELSE (abs(rnd_long()) % 1000) + 1 END                    AS sparse10,
        -- 50% NULL on a narrow column (exercises of(0) reset on many groups)
        CASE WHEN rnd_int() % 2 = 0 THEN NULL
            ELSE (abs(rnd_long()) % 1000) + 1 END                    AS sparse50,
        -- Group keys at three cardinalities
        abs(rnd_int()) % 10                                           AS g10,
        abs(rnd_int()) % 1_000                                        AS g1k,
        abs(rnd_int()) % 100_000                                      AS g100k
    FROM long_sequence(100_000_000)
) TIMESTAMP(ts) PARTITION BY DAY WAL;

-- (1) Non-keyed baseline — single global histogram, narrow values
-- patch - 65ms
-- master - 697ms
SELECT approx_percentile(latency_us, 0.99, 1) FROM approx_pct_bench;

-- (2) Cardinality sweep — narrow values, precision 1
-- patch - 72ms
-- master - 1.19s
SELECT g10,   approx_percentile(latency_us, 0.99, 1) FROM approx_pct_bench;
-- patch - 115ms
-- master - 1.85s
SELECT g1k,   approx_percentile(latency_us, 0.99, 1) FROM approx_pct_bench;
-- patch - 1.23s
-- master - 18s
SELECT g100k, approx_percentile(latency_us, 0.99, 1) FROM approx_pct_bench;

-- (3) Value-range stress — wide values force autoResize per group
-- patch - 87ms
-- master - 1.47s
SELECT g10,   approx_percentile(wide_val, 0.99, 1) FROM approx_pct_bench;

-- (4) NULL ratio sweep at medium cardinality
-- patch - 114ms
-- master - 2.11s
SELECT g1k, approx_percentile(latency_us, 0.99, 1) FROM approx_pct_bench;

-- (5) Precision boundary — precision 2 (parallelized) vs precision 3 (Packed, NOT parallel)
-- patch - 228ms
-- master - 2.41s
SELECT g1k, approx_percentile(latency_us, 0.99, 2) FROM approx_pct_bench;
```

core/src/main/java/io/questdb/griffin/engine/functions/groupby/ApproxPercentileLongGroupByFunction.java
mcadariu: Yes agreed, that way we preserve good encapsulation.

core/src/test/java/io/questdb/test/cairo/fuzz/ParallelGroupByFuzzTest.java
```diff
@@ -2160,6 +2160,63 @@ public void testParallelNonKeyedGroupByFaultTolerance() throws Exception {
         testParallelGroupByFaultTolerance("select vwap(price, quantity) from tab where npe();");
     }
 
+    @Test
+    public void testParallelApproxPercentileFuzz() throws Exception {
```
puzpuzpuz, line 2635: Could you add a non-keyed non-keyed variant of this test?

core/src/test/java/io/questdb/test/cairo/fuzz/ParallelGroupByFuzzTest.java
```diff
@@ -2160,6 +2160,63 @@ public void testParallelNonKeyedGroupByFaultTolerance() throws Exception {
...
+            TestUtils.execute(
+                    pool,
+                    (engine, compiler, sqlExecutionContext) -> {
+                        sqlExecutionContext.setJitMode(
+                                enableJitCompiler ? SqlJitMode.JIT_MODE_ENABLED : SqlJitMode.JIT_MODE_DISABLED);
+
+                        execute(
+                                compiler,
+                                "CREATE TABLE tab AS ("
+                                        + "SELECT abs(rnd_long()) % 10000000 AS x, rnd_int() % " + groupCount + " AS g "
```
puzpuzpuz: Might be a good idea to change `abs(rnd_long()) % 10000000` to `case when rnd_int() % 10 = 0 then null else abs(rnd_long()) % 10_000_000 end` or something similar, so that we get nulls consistently generated in the test data.

puzpuzpuz (dismissed):
LGTM. Left a couple comments around test gaps - please treat them as nits.

mcadariu:
Thanks! I will address the remaining points ASAP.

puzpuzpuz (approved):

PR #6843, opened 2026-03-03
feat(sql): add per-column parquet encoding/compression config

RaphDal:
Tandem PR: https://github.com/questdb/questdb-enterprise/pull/932, https://github.com/questdb/documentation/pull/384

## Summary

- Adds per-column Parquet encoding and compression configuration via `CREATE TABLE` and `ALTER TABLE` SQL syntax
- Introduces a global `min_compression_ratio` parameter that discards compressed pages when the compression ratio (`uncompressed/compressed`) falls below a threshold, storing them uncompressed instead
- Add RLE dictionary encoding support for all column types except Boolean and Array
- Improve varchar dict encoding performance by switching from `HashMap` to `RapidHashMap` and storing indices directly in a `Vec<u32>` instead of fetching them from the hashmap
- Enable `lto` and strip `debuginfo` from qdbr library (~50% file size reduction).

## Per-column Parquet encoding and compression

Users can now specify Parquet encoding and compression on a per-column basis. This only applies to Parquet partitions and is ignored for native partitions.

#### `CREATE TABLE`

```sql
CREATE TABLE sensors (
    ts TIMESTAMP,
    temperature DOUBLE PARQUET(rle_dictionary, zstd(3)),
    humidity FLOAT PARQUET(rle_dictionary),
    device_id VARCHAR PARQUET(default, lz4_raw),
    status INT
) TIMESTAMP(ts) PARTITION BY DAY;
```

The syntax is `PARQUET(encoding [, compression[(level)]])`. Both encoding and compression are optional — use `default` for the encoding when specifying compression only. When omitted entirely, the column uses the global defaults (type-based encoding and the server-wide compression codec).

#### `ALTER TABLE`

Setting per-column config on existing tables:

```sql
-- Set encoding only
ALTER TABLE sensors ALTER COLUMN temperature SET PARQUET(rle_dictionary);

-- Set compression only (with optional level)
ALTER TABLE sensors ALTER COLUMN temperature SET PARQUET(default, zstd(3));

-- Set both
ALTER TABLE sensors ALTER COLUMN temperature SET PARQUET(rle_dictionary, zstd(3));

-- Reset to defaults
ALTER TABLE sensors ALTER COLUMN temperature DROP PARQUET;
```

#### `SHOW CREATE TABLE`

Per-column Parquet config appears in `SHOW CREATE TABLE` output:

```sql
CREATE TABLE 'sensors' (
  ts TIMESTAMP,
[...]

PR #7085, opened 2026-05-11
chore(core): remove garbled PARQUET clauses in SHOW CREATE TABLE on pre-9.3.4 tables

RaphDal:
## Summary

- On tables whose `_meta` file was last written before 9.3.4, `SHOW CREATE TABLE` could emit invalid DDL of the form `ts TIMESTAMP PARQUET(unknown(133), unknown(202)(90), bloom_filter)`, which the parser rejects. This breaks copy-schema workflows.
- Root cause: PR #6843 (commit `e8387cdd13`, shipped in 9.3.4) added `parquetEncodingConfig` at column-entry offset 20 in the `_meta` file without bumping `META_FORMAT_MINOR_VERSION_LATEST`. The reader paths in `MetadataCache`, `TableReaderMetadata`, and `TableWriterMetadata` read those 4 bytes unconditionally. Pre-9.3.4 `_meta` files can carry non-zero data at that offset from earlier layouts: the migration `Mig608` (meta version 423 -> 424, used on tables created before October 2021) writes a random `Rnd.nextLong()` at offsets 20-27, and `_meta` files written in the Oct 2021 - Nov 2022 column-hash era have the upper half of the old 8-byte column hash sitting at 20-23. Either way, bit 24 (the "explicit" flag) is set in roughly half of all column entries, which trips the parquet-clause emitter in `ShowCreateTableRecordCursorFactory`; the other bytes flow into `ParquetEncoding.getEncodingName` / `ParquetCompression.getCompressionName` and surface as `unknown(<id>)`.
- Fix: introduce `TableUtils.hasParquetEncodingConfig(metaMem)`, gated on a new named constant `META_FORMAT_MINOR_VERSION_PARQUET_ENCODING_CONFIG = 1` (the value of `META_FORMAT_MINOR_VERSION_LATEST` at the time #6843 landed, i.e. the minor version that guarantees offset 20-23 is either zero or a real config and not stale bytes). All three reader sites call the new helper instead of reading the field unconditionally. When the meta checksum is invalid or the saved minor version is below this threshold, the field defaults to `0` (no per-column config).
- A small refactor extracts the checksum + minor-version comparison from `isMetaFormatUpToDate` into a private `isMetaFormatAtLeast(metaMem, minorVersion)` helper, so the parquet gate and the existing "all fields current" check share one implementation while binding to different version constants. This is the future-proof spelling: when the next versioned field is added and `META_FORMAT_MINOR_VERSION_LATEST` is bumped to 2, the parquet gate stays pinned at version 1 and 9.3.4 / 9.3.5 metas keep their per-column configs.

## Tradeoffs

- The fix only suppresses the bogus values during reads. Old `_meta` files on disk still contain the stale bytes at offset 20-23. They are overwritten with a real `parquetEncodingConfig` (zero by default) the next time the meta is rewritten - i.e., on any schema-changing ALTER TABLE. Until then, the on-disk bytes remain incorrect, but the user-facing output is correct.
- Tables created on >= 9.3.4 with an explicit per-column `PARQUET(...)` clause are unaffected: their meta is written with `META_FORMAT_MINOR_VERSION_LATEST = 1` and passes the `hasParquetEncodingConfig` check.
- An alternative would be to bump `META_FORMAT_MINOR_VERSION_LATEST` to 2 and add a one-time meta-rewrite migration that zeros offset 20-23 on legacy entries. That is heavier than necessary: the read-time gate handles correctness at no on-disk cost, and the named-constant gate introduced here preserves the option to do that later if needed.

## Test plan

- New regression test `ShowCreateTableTest#testParquetIgnoredOnLegacyMetaFormat`: creates a non-WAL table, writes a value with bit 24 set at offset 20 of every column entry in `_meta`, sets `META_OFFSET_META_FORMAT_MINOR_VERSION` to 0, clears the metadata cache, and asserts that `SHOW CREATE TABLE` emits no `PARQUET(...)` annotations. Confirmed the test reproduces the bug on the unpatched code (output contains `PARQUET(unknown(133), unknown(201)(89), bloom_filter)` for every column) and passes with the fix applied.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/MetadataCache.java
+++ b/core/src/main/java/io/questdb/cairo/MetadataCache.java
@@ -215,6 +215,7 @@ public class MetadataCache implements QuietCloseable {
 
             TableUtils.buildColumnListFromMetadataFile(metaMem, columnCount, table.columnOrderList);
             boolean isMetaFormatUpToDate = TableUtils.isMetaFormatUpToDate(metaMem);
+            boolean hasParquetEncodingConfig = TableUtils.hasParquetEncodingConfig(metaMem);
             // populate columns
             for (int i = 0, n = table.columnOrderList.size(); i < n; i += 3) {
                 int writerIndex = table.columnOrderList.get(i);
@@ -246,7 +247,9 @@ public class MetadataCache implements QuietCloseable {
                 column.setIndexBlockCapacity(TableUtils.getIndexBlockCapacity(metaMem, writerIndex));
                 column.setSymbolTableStaticFlag(true);
                 column.setDedupKeyFlag(TableUtils.isColumnDedupKey(metaMem, writerIndex));
-                column.setParquetEncodingConfig(isMetaFormatUpToDate ? TableUtils.getParquetEncodingConfig(metaMem, writerIndex) : 0);
+                // Pre-9.3.4 _meta files can carry non-zero bytes at column-entry offset 20 from
+                // older layouts (e.g. Mig608 wrote a random column hash at 20-27); read 0 there.
+                column.setParquetEncodingConfig(hasParquetEncodingConfig ? TableUtils.getParquetEncodingConfig(metaMem, writerIndex) : 0);
                 column.setWriterIndex(writerIndex);
 
                 boolean isDesignated = writerIndex == timestampWriterIndex;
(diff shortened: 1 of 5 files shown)


questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=232513)
**Tested Commit:** `a56aa296f909e5d35339575405513ddc46f295da`

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableReaderMetadata.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriterMetadata.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableUtils.java|9|9|100.00%|

bluestreak01 (approved):

PR #7086, opened 2026-05-11
chore(core): restore posting index count() fast path on clean partitions

bluestreak01:
## Summary

- Restores the `Cursor.size()` fast path in `AbstractPostingIndexReader` for clean partitions. PR #7071 had to disable it across the board to prevent over-counting row ids past the chain entry's tracked coverage.
- Per-gen classification: gens whose encoded range sits entirely at or below `entryMaxValue` contribute their count, gens whose range sits entirely above contribute 0, and gens that straddle the boundary force a `-1` bail so the caller iterates with the same clamp the cursor applies.
- `CoveringIndexRecordCursorFactory#size()`, which `COUNT(*) WHERE sym = 'X'` uses, returns to `O(genCount)` plus a small constant per gen on clean partitions, instead of full `O(totalRowIdsForKey)` iteration.

## Approach

For every gen visited by `size()`, peek at the encoded layout (without decoding values) to derive a min and an upper bound on the max for the requested key:

- Dense FLAT mode: per-key min and max are exact, read by unpacking the slice endpoints relative to `baseValue`.
- Dense DELTA mode (and the same blob inside sparse gens): min comes from `firstValues[0]` exactly; max is a slack upper bound from the last block's packed-max formula. For constant-stride blocks (`bitWidth == 0`) the bound is also exact.
- Elias-Fano: max is exact (`universe - 1`); min decodes the first set bit of the high-bit stream and the corresponding low chunk.

Classification rules:

- `min > entryMaxValue` => ALL_DIRTY, skip the gen.
- `maxUpperBound <= entryMaxValue` => CLEAN, add the per-gen count.
- otherwise MIXED, return `-1` and let the caller iterate.

The empty-chain branch (`entryMaxValue == -1`) keeps the original count fast path verbatim because no clamping applies.

## Tradeoffs

- DELTA-mode upper bound is slack at high bit-widths, so some borderline-CLEAN gens classify as MIXED and force iteration. This only loses the fast path; it never miscounts. The bound becomes exact when the last block uses `bitWidth == 0`.
- Dirty-data workloads (the bug PR #7071 fixed) still take the iteration path. Performance there is unchanged from current `master`.
- The static peek helpers are reader-only; no on-disk format change. The cursor still walks the same clamped path during iteration, so the PR #7071 correctness fix is untouched.

## Test plan

- `mvn -pl core -Dtest='PostingIndexCriticalIssuesTest' test` -- 27 tests pass, including the 5 new size() tests and the 3 PR #7071 regression tests (`testReaderClampsCursorToChainEntryMaxValue`, `testCollectDistinctKeysDoesNotExposeKeysPastEntryMaxValue`, `testReaderSizeDoesNotOutrunEntryMaxValueClamp`).
- `mvn -pl core -Dtest='PostingIndex*Test,CoveringIndexTest,BitmapIndexTest,SymbolMapTest' test` -- 566 tests pass.
- New tests:
  - `testSizeFastPathOnDenseGen` / `testSizeFastPathOnSparseGen`: `size()` returns the iterated count (not `-1`) when no dirty rows exist, covering both sealed-dense and commit-time-sparse layouts.
  - `testSizeBypassesAllDirtyGen`: gen with every encoded row above `MAX_VALUE` returns 0 without bailing.
  - `testSizeBailsOnMixedGen`: single-gen MIXED case forces `size() == -1`.
  - `testSizeMultiGenMixedClassification`: one CLEAN gen + one ALL_DIRTY gen sum to the iterated count without bailing.
- Performance validation is via `PostingIndexBenchmarkSuite.walLargePartitionQuery` (`SELECT count() FROM walbench WHERE sym = '...'` across `partitionSize` ∈ {1M, 10M, 100M} and `indexType` ∈ {`no_index`, `bitmap`, `posting_covering_10cols`}). The benchmark exercises the recovered fast path end-to-end.

## Direct A/B against master

Ad-hoc test that builds a sealed posting index with 1,000,000 row ids for a single key, then times the fast-path call vs the forced-iteration fallback at the cursor level. The same file also runs the end-to-end `SELECT count() FROM t WHERE sym = 'X'` shape against a 1M-row clean partition. Procedure: run on this branch, `cp` the reader file to `/tmp`, `git checkout master -- core/src/main/java/io/questdb/cairo/idx/AbstractPostingIndexReader.java`, rerun, restore the working tree from `/tmp`. The demo file is not committed.

Results, single key, single sealed partition, 1,000,000 row ids:

| measurement                              |  with fix |    master | ratio |
|------------------------------------------|----------:|----------:|------:|
| `cursor.size()` return value             | 1,000,000 |        -1 | n/a   |
| direct `cursor.size()`-or-fallback, p50  |      7 us | 3,293 us  | ~470x |
| `SELECT count() WHERE sym='X'`, p50      |     76 us | 2,388 us  | ~31x  |
| forced iteration (control), p50          |  2,788 us | 3,315 us  | ~1.2x |

The "forced iteration" row serves as a control: it exercises the slow path the fix lets the caller skip, so its near-equal numbers across the two runs confirm the gap on the other rows is not measurement noise. The SQL ratio sits below the raw-cursor ratio because cursor construction, `hasNext`, and `getLong` overhead is common to both runs and dominates the with-fix case. Dirty-data workloads (per Tradeoffs above) still iterate, unchanged from master. This is a best-case microbenchmark: one key, one sealed partition, no dirty rows; real workloads with mixed-classification gens will see smaller ratios or no change.

Test code, `core/src/test/java/io/questdb/test/cairo/PostingIndexFastPathPerfDemo.java`, not committed:

```java
/*
 * AD-HOC PERF DEMO - NOT INTENDED FOR COMMIT.
 *
 * Validates the size() fast-path restoration in
 * core/src/main/java/io/questdb/cairo/idx/AbstractPostingIndexReader.java.
 *
 * Strategy: build a clean partition with N rows for one symbol, time
 *   SELECT count() FROM t WHERE sym = 'X'
 * across many invocations. With the fast path, count() resolves via
 * cursor.size() in O(genCount); without it, the caller falls back to
 * full O(N) iteration.
 *
 * To compare against master, run this test, then:
 *   git checkout master -- core/src/main/java/io/questdb/cairo/idx/AbstractPostingIndexReader.java
 *   mvn -pl core -Dtest=PostingIndexFastPathPerfDemo -P local-client test
 *   git checkout HEAD -- core/src/main/java/io/questdb/cairo/idx/AbstractPostingIndexReader.java
 */
package io.questdb.test.cairo;

import io.questdb.cairo.idx.PostingIndexFwdReader;
import io.questdb.cairo.idx.PostingIndexWriter;
import io.questdb.cairo.sql.RecordCursor;
import io.questdb.cairo.sql.RecordCursorFactory;
import io.questdb.cairo.sql.RowCursor;
import io.questdb.std.str.Path;
import org.junit.Assert;
import org.junit.Test;

import java.util.Arrays;

import static io.questdb.cairo.TableUtils.COLUMN_NAME_TXN_NONE;

public class PostingIndexFastPathPerfDemo extends AbstractCairoTest {

    private static final int ITERATIONS = 200;
    private static final int ROWS = 1_000_000;
    private static final int WARMUP = 20;

    /**
     * Direct cursor measurement: opens a PostingIndexFwdReader on a sealed
     * index with ROWS row ids for key 0, then alternately measures the time
     * to (a) call cursor.size() and (b) drain the cursor.
     *
     * With the fix: size() returns ROWS (a real count), the caller's fast
     * path skips iteration entirely.
     *
     * Without the fix: size() returns -1, the caller has to drain the
     * cursor, paying O(ROWS).
     */
    @Test
    public void demoCursorSizeVsIterate() throws Exception {
        assertMemoryLeak(() -> {
            final String name = "perf_demo_cursor";
            try (Path path = new Path().of(configuration.getDbRoot())) {
                final int plen = path.size();
                try (PostingIndexWriter writer = new PostingIndexWriter(
                        configuration, path.trimTo(plen), name, COLUMN_NAME_TXN_NONE)) {
                    for (long rowId = 0; rowId < ROWS; rowId++) {
                        writer.add(0, rowId);
                    }
                    writer.commit();
                    writer.seal();
                }

                try (PostingIndexFwdReader reader = new PostingIndexFwdReader(
                        configuration, path.trimTo(plen), name,
                        COLUMN_NAME_TXN_NONE, /* partitionTxn */ 0, /* columnTop */ 0)) {
                    long advertised;
                    try (RowCursor cursor = reader.getCursor(0, 0L, Long.MAX_VALUE)) {
                        advertised = cursor.size();
                    }
                    System.out.printf(">>> cursor.size() returned: %d  (fast path: %s)%n",
                            advertised, advertised >= 0 ? "ENGAGED" : "DISABLED (returned -1)");

                    // Warmup
                    for (int i = 0; i < WARMUP; i++) {
                        try (RowCursor cursor = reader.getCursor(0, 0L, Long.MAX_VALUE)) {
                            long s = cursor.size();
                            if (s < 0) {
                                while (cursor.hasNext()) cursor.next();
                            }
                        }
                    }

                    // Measure size() path (whatever it returns).
                    long[] sizeNs = new long[ITERATIONS];
                    for (int i = 0; i < ITERATIONS; i++) {
                        long t0 = System.nanoTime();
                        try (RowCursor cursor = reader.getCursor(0, 0L, Long.MAX_VALUE)) {
                            long s = cursor.size();
                            if (s < 0) {
                                // Caller's fallback: drain.
                                while (cursor.hasNext()) cursor.next();
                            }
                        }
                        sizeNs[i] = System.nanoTime() - t0;
                    }

                    // Measure forced-iteration (the slow path the fast path skips).
                    long[] iterNs = new long[ITERATIONS];
                    for (int i = 0; i < ITERATIONS; i++) {
                        long t0 = System.nanoTime();
                        long iterated = 0;
                        try (RowCursor cursor = reader.getCursor(0, 0L, Long.MAX_VALUE)) {
                            while (cursor.hasNext()) {
                                cursor.next();
                                iterated++;
                            }
                        }
                        iterNs[i] = System.nanoTime() - t0;
                        Assert.assertEquals(ROWS, iterated);
                    }

                    report("cursor.size()-or-fallback", sizeNs);
                    report("forced iteration         ", iterNs);
                }
            }
        });
    }

    /**
     * SQL end-to-end measurement: times SELECT count() FROM t WHERE sym = 'X'
     * against a 1M-row clean partition. The CountRecordCursorFactory routes
     * through baseCursor.size(); on the fixed branch this resolves via the
     * per-gen classification fast path, on master it falls back to
     * calculateSize() which drains every row id.
     */
    @Test
    public void demoSqlCountFastPath() throws Exception {
        assertMemoryLeak(() -> {
            execute("CREATE TABLE perf_t (ts TIMESTAMP, sym SYMBOL INDEX TYPE POSTING) " +
                    "TIMESTAMP(ts) PARTITION BY DAY");
            execute("INSERT INTO perf_t " +
                    "SELECT dateadd('u', x::INT, '2024-01-01T00:00:00.000000Z'::TIMESTAMP), 'X' " +
                    "FROM long_sequence(" + ROWS + ")");

            String sql = "SELECT count() FROM perf_t WHERE sym = 'X'";

            // Compile once, reuse the factory across iterations so we measure
            // cursor cost only, not parse/plan cost.
            try (RecordCursorFactory factory = select(sql)) {
                // Sanity check the answer once and confirm fast path engagement.
                try (RecordCursor c = factory.getCursor(sqlExecutionContext)) {
                    Assert.assertTrue(c.hasNext());
                    long got = c.getRecord().getLong(0);
                    Assert.assertEquals(ROWS, got);
                }

                // Warmup
                for (int i = 0; i < WARMUP; i++) {
                    try (RecordCursor c = factory.getCursor(sqlExecutionContext)) {
                        c.hasNext();
                        c.getRecord().getLong(0);
                    }
                }

                // Measure
                long[] times = new long[ITERATIONS];
                for (int i = 0; i < ITERATIONS; i++) {
                    long t0 = System.nanoTime();
                    try (RecordCursor c = factory.getCursor(sqlExecutionContext)) {
                        c.hasNext();
                        c.getRecord().getLong(0);
                    }
                    times[i] = System.nanoTime() - t0;
                }
                report("SELECT count() WHERE sym='X'", times);
            }
        });
    }

    private static void report(String label, long[] ns) {
        long[] sorted = ns.clone();
        Arrays.sort(sorted);
        long min = sorted[0];
        long p50 = sorted[sorted.length / 2];
        long p90 = sorted[(int) (sorted.length * 0.9)];
        long max = sorted[sorted.length - 1];
        System.out.printf(">>> %s : min=%6d us  p50=%6d us  p90=%6d us  max=%6d us%n",
                label, min / 1000, p50 / 1000, p90 / 1000, max / 1000);
    }
}
```

## Benchmark

```

╔══════════════════════════════════════════════════════════════════════════════════╗
║              POSTING INDEX BENCHMARK SUMMARY  [encoding: EF   ]                  ║
╚══════════════════════════════════════════════════════════════════════════════════╝

── Decode Throughput (ops/s, higher=better) ──────────────────────────────────────
  batch           1-bit        8-bit       12-bit       16-bit       20-bit       32-bit
  64         23,460,512   11,963,580   10,226,865    8,027,155    7,550,493    6,119,904
  256         5,599,120    3,318,368    2,714,740    2,165,403    2,021,724    1,653,614
  1024        1,552,973      810,553      732,845      557,837      517,882      420,462

── Index Comparison: Posting vs Legacy (ops/s, higher=better) ────────────────────
  indexPointRead:
             LEGACY    POSTING    ratio
  S1            568      2,012    3.54x
  S2          1,324      1,200    0.91x ◄
  S3          1,907      1,838    0.96x ◄
  S4            215        262    1.22x
  S5            407        238    0.59x ◄
  S6          1,096      2,022    1.84x
  S7            366      1,734    4.74x
  S8            392        470    1.20x

  indexScanRead:
             LEGACY    POSTING    ratio
  S1             20         45    2.23x
  S2            415        484    1.17x
  S3          1,003        892    0.89x ◄
  S4            205        250    1.22x
  S5            390        244    0.63x ◄
  S6            118         93    0.78x ◄
  S7             13         28    2.09x
  S8            374        469    1.25x

  indexRangeRead:
             LEGACY    POSTING    ratio
  S1          1,844      4,312    2.34x
  S2            781        987    1.26x
  S3          5,798     11,770    2.03x
  S4          1,190      1,809    1.52x
  S5            711        386    0.54x ◄
  S6          2,470      5,810    2.35x
  S7          1,576      3,961    2.51x
  S8          4,410     16,010    3.63x

── Sidecar Read Throughput (ops/s, higher=better) ────────────────────────────────
  type       baseline   covering    ratio
  DOUBLE          285        184    0.64x
  FLOAT           325        205    0.63x
  LONG            326        189    0.58x
  INT             318        192    0.61x
  DECIMAL64        305        192    0.63x
  DECIMAL32        313        203    0.65x
  SHORT           317        202    0.64x

── SQL Queries (ops/s, higher=better) ────────────────────────────────────────────
  Point lookup:
    covering_where                     23,386 ops/s
    non_covering_where                 22,006 ops/s
  Aggregation:
    covering_agg                        7,825 ops/s
    covering_sum                        8,496 ops/s
    covering_count                     31,939 ops/s
  Filter:
    residual_filter                    14,847 ops/s
    non_covering_filter                15,719 ops/s
    no_index_filter                     2,986 ops/s
  Filter IN-list:
    residual_filter_in                  1,487 ops/s
    non_covering_filter_in              1,617 ops/s
  VARCHAR/FSST:
    varchar_fsst                       27,080 ops/s
    varchar_non_covering               21,545 ops/s
    varchar_in_covering                 5,016 ops/s
  O3:
    o3_covering                        20,303 ops/s
    o3_non_covering                    24,634 ops/s
    o3_distinct                        38,730 ops/s
  Misc:
    latest_on                          39,780 ops/s
    in_list                             1,512 ops/s
    wide_table                         25,172 ops/s
  Bulk throughput:
    bulk_covering                       1,026 ops/s
    bulk_non_covering                     936 ops/s

── Write Overhead (ops/s, higher=better) ─────────────────────────────────────────
  no_index                     41.4 ops/s
  bitmap                       37.1 ops/s
  posting                      39.1 ops/s
  posting_covering             37.8 ops/s
  posting_varchar              35.9 ops/s

── Commit Profile (512 keys × 56 commits × 128 v/commit + seal) ──────────────────
  full write+seal cycle        30.7 ops/s  (32.6 ms/cycle)


══════════════════════════════════════════════════════════════════════════════════════════════════
  Covering Index I/O Projection (CPU + predicted page faults)
══════════════════════════════════════════════════════════════════════════════════════════════════
  500 keys × 2,000 v/k = 1,000,000 rows, querying 200 keys

  type             pages   pages    CPU ms    CPU ms    + NVMe   + cloud     + HDD
                    base   cover      base     cover     cover     cover     cover
  ────────────────────────────────────────────────────────────────────────────────────────
  DOUBLE           1,954     348       7.3       7.5      35.4     355.5    1399.5
                                                         163.6                        (baseline + NVMe for comparison)
  FLOAT              977     181       2.2       6.9      21.4     187.9     730.9
                                                          80.4                        (baseline + NVMe for comparison)
  LONG             1,954     248      10.0      15.5      35.3     263.5    1007.5
                                                         166.3                        (baseline + NVMe for comparison)
  INT                977     175       3.8       5.5      19.5     180.5     705.5
                                                          81.9                        (baseline + NVMe for comparison)

══════════════════════════════════════════════════════════════════════════════════════════════════
  Implied Sidecar Compression (file size, raw column vs encoded sidecar)
══════════════════════════════════════════════════════════════════════════════════════════════════
  type              raw column       sidecar         ratio
  ────────────────────────────────────────────────────────
  DOUBLE               7.63 MB       3.39 MB         2.25x
  FLOAT                3.81 MB       1.77 MB         2.16x
  LONG                 7.63 MB       2.42 MB         3.15x
  INT                  3.81 MB       1.70 MB         2.24x
```

## Compared to master

```
⏺ Comparing the two runs (master = baseline you pasted; PR = #7086 description):

  SQL Queries — meaningful wins

  ┌────────────────────┬────────┬────────┬─────────────────────────┐
  │       Query        │ master │   PR   │          delta          │
  ├────────────────────┼────────┼────────┼─────────────────────────┤
  │ covering_count     │ 28,116 │ 31,939 │ +14% ← fast-path target │
  ├────────────────────┼────────┼────────┼─────────────────────────┤
  │ covering_where     │ 20,461 │ 23,386 │                    +14% │
  ├────────────────────┼────────┼────────┼─────────────────────────┤
  │ non_covering_where │ 19,897 │ 22,006 │                    +11% │
  ├────────────────────┼────────┼────────┼─────────────────────────┤
  │ covering_sum       │  7,488 │  8,496 │                    +13% │
  ├────────────────────┼────────┼────────┼─────────────────────────┤
  │ latest_on          │ 34,991 │ 39,780 │                    +14% │
  ├────────────────────┼────────┼────────┼─────────────────────────┤
  │ wide_table         │ 22,978 │ 25,172 │                    +10% │
  ├────────────────────┼────────┼────────┼─────────────────────────┤
  │ varchar_fsst       │ 24,861 │ 27,080 │                     +9% │
  ├────────────────────┼────────┼────────┼─────────────────────────┤
  │ o3_distinct        │ 36,322 │ 38,730 │                     +7% │
  ├────────────────────┼────────┼────────┼─────────────────────────┤
  │ Most others        │      — │      — │    +1–6% (within noise) │
  └────────────────────┴────────┴────────┴─────────────────────────┘

  covering_count is the headline number and tracks the change — the fast path saves a full row-id drain on clean partitions.

  Index micro-benchmarks — flat-to-slightly-noisy

  - indexPointRead: mixed ±2–5%, except S3 +11% (1,661 → 1,838).
  - indexScanRead: uniformly −2 to −7%.
  - indexRangeRead: −2 to −8% (S7, S8 worst at −8%).
  - Decode throughput: within ~5% noise.

  These paths don't touch size(), so the small swings look like run-to-run variance rather than a real effect.

  Write & commit

  - bitmap write: 42.5 → 37.1 ops/s (−13%) — biggest standout, unrelated to this PR's code path; very likely noise.
  - posting_covering: 40.7 → 37.8 (−7%).
  - Commit cycle: 31.5 → 30.7 (−3%).

  Sidecar reads

  −2 to −7% across the board, again consistent with run-to-run variance.

  Bottom line: the SQL queries that route through cursor.size() (count, where, latest_on) gain 10–14%, matching the PR's
  intent. Everything else moves within typical bench noise, with no path getting materially worse. The PR's microbench (~31×
  SQL ratio) is the best-case 1M-rows-1-key scenario; on the broader suite the wins land at +10–14% for affected query shapes.
```

### DIFF
--- a/core/src/main/java/io/questdb/cairo/idx/AbstractPostingIndexReader.java
+++ b/core/src/main/java/io/questdb/cairo/idx/AbstractPostingIndexReader.java
@@ -570,6 +570,117 @@ public abstract class AbstractPostingIndexReader implements IndexReader {
         return false;
     }
 
+    /**
+     * Returns a (possibly slack) upper bound on the largest row id encoded
+     * for one key, given the per-key encoded blob at {@code encodedOffset}
+     * inside the value memory mapping. Returns {@code -1L} when the key
+     * holds no values, and {@code Long.MAX_VALUE} when bit-width arithmetic
+     * would overflow (handled as MIXED by {@code Cursor.size()}).
+     * <p>
+     * Flat-mode strides do not use the delta layout, so this helper covers
+     * only the per-key delta-FoR / Elias-Fano blob format that
+     * {@code deltaKeyHasValueInRange} consumes.
+     */
+    private static long peekDeltaKeyMaxValueUpperBound(long baseAddr, long encodedOffset) {
+        int firstWord = Unsafe.getInt(baseAddr + encodedOffset);
+        if (firstWord == PostingIndexUtils.EF_FORMAT_SENTINEL) {
+            // EF: writer stores universe == lastValue + 1, so max == universe - 1 exactly.
+            long universe = Unsafe.getLong(baseAddr + encodedOffset + 9);
+            return universe - 1;
+        }
+        if (firstWord <= 0) {
+            return -1L;
+        }
+        int blockCount = firstWord;
+        long valueCountsOff = encodedOffset + 4;
+        long firstValuesOff = valueCountsOff + blockCount;
+        long minDeltasOff = firstValuesOff + (long) blockCount * Long.BYTES;
+        long bitWidthsOff = minDeltasOff + (long) blockCount * Long.BYTES;
+        int last = blockCount - 1;
+        long lastFv = Unsafe.getLong(baseAddr + firstValuesOff + (long) last * Long.BYTES);
+        int lastCount = Unsafe.getByte(baseAddr + valueCountsOff + last) & 0xFF;
+        int numDeltas = lastCount - 1;
+        if (numDeltas == 0) {
+            return lastFv;
+        }
+        long lastMinD = Unsafe.getLong(baseAddr + minDeltasOff + (long) last * Long.BYTES);
+        int lastBitWidth = Unsafe.getByte(baseAddr + bitWidthsOff + last) & 0xFF;
+        if (lastBitWidth == 0) {
+            // Constant-stride block: writer guarantees minD > 0 when count > 1, so the
+            // arithmetic progression's last value is exact.
+            return lastFv + (long) numDeltas * lastMinD;
+        }
+        // Slack upper bound: assume every delta in the last block packs to (1<<bw)-1.
+        // Real values are <= this, so a CLEAN classification stays correct; the cost
+        // is a stricter cutoff for MIXED detection at high bitwidth.
+        if (lastBitWidth >= 32) {
+            return Long.MAX_VALUE;
+        }
+        long maxDelta = lastMinD + (1L << lastBitWidth) - 1;
+        if (maxDelta < lastMinD) {
+            return Long.MAX_VALUE;
+        }
+        long span = (long) numDeltas * maxDelta;
+        if (maxDelta != 0 && span / maxDelta != numDeltas) {
+            return Long.MAX_VALUE;
+        }
+        long max = lastFv + span;
+        if (max < lastFv) {
+            return Long.MAX_VALUE;
+        }
+        return max;
+    }
+
+    /**
+     * Exact smallest row id encoded for one key, taken from the encoded blob
+     * at {@code encodedOffset}. Returns {@code -1L} when the key holds no
+     * values. Mirror to {@link #peekDeltaKeyMaxValueUpperBound}.
+     */
+    private static long peekDeltaKeyMinValue(long baseAddr, long encodedOffset) {
+        int firstWord = Unsafe.getInt(baseAddr + encodedOffset);
+        if (firstWord == PostingIndexUtils.EF_FORMAT_SENTINEL) {
+            return peekEFKeyMinValue(baseAddr, encodedOffset);
+        }
+        if (firstWord <= 0) {
+            return -1L;
+        }
+        long firstValuesOff = encodedOffset + 4 + firstWord;
+        return Unsafe.getLong(baseAddr + firstValuesOff);
+    }
+
+    private static long peekEFKeyMinValue(long baseAddr, long encodedOffset) {
+        long pos = encodedOffset + 4; // skip EF_FORMAT_SENTINEL
+        int totalCount = Unsafe.getInt(baseAddr + pos);
+        pos += 4;
+        if (totalCount == 0) {
+            return -1L;
+        }
+        int bitsL = Unsafe.getByte(baseAddr + pos) & 0xFF;
+        pos += 1;
+        long universe = Unsafe.getLong(baseAddr + pos);
+        pos += 8;
+        long lowOffset = pos;
+        long highOffset = pos + PostingIndexUtils.efLowBytesAligned(totalCount, bitsL);
+        int numHighWords = (int) ((totalCount + (universe >>> bitsL) + 63) / 64);
+        for (int i = 0; i < numHighWords; i++) {
+            long word = Unsafe.getLong(baseAddr + highOffset + (long) i * 8);
+            if (word == 0) {
+                continue;
+            }
+            // First set bit's absolute index in the high-bit stream is the high
+            // part of value 0 (outputCount == 0 at the start, so base == i*64).
+            long high = (long) i * 64 + Long.numberOfTrailingZeros(word);
+            long low = 0;
+            if (bitsL > 0) {
+                long lowMask = (bitsL < 64) ? (1L << bitsL) - 1 : -1L;
+                long lowWord = Unsafe.getLong(baseAddr + lowOffset);
+                low = lowWord & lowMask;
+            }
+            return (high << bitsL) | low;
+        }
+        return -1L;
+    }
+
     private void closeSidecarMems() {
         Misc.freeObjListAndKeepObjects(sidecarMems);
         coverCount = 0;
@@ -1227,18 +1338,45 @@ public abstract class AbstractPostingIndexReader implements IndexReader {
             if (requestedKey < 0) {
                 return 0;
             }
-            if (entryMaxValue >= 0) {
-                // Raw per-gen counts do not know whether encoded row ids sit
-                // past the chain entry's tracked coverage. Decline the fast
-                // path so callers count via the same clamped hasNext() path.
-                return -1;
-            }
+            // Per-gen classification when the chain entry advertises a tracked
+            // coverage: CLEAN gens (max <= entryMaxValue) contribute their count,
+            // ALL_DIRTY gens (min > entryMaxValue) contribute zero, MIXED gens
+            // force a -1 bail so the caller falls back to the clamped iteration.
+            // entryMaxValue == -1 means an empty chain entry where no clamping
+            // applies; the original count fast path is taken verbatim.
             long total = 0;
             for (int g = 0; g < cursorGenCount; g++) {
                 int gkc = genLookup.getGenKeyCount(g);
                 if (gkc >= 0) {
+                    if (entryMaxValue >= 0) {
+                        long minV = peekDenseKeyMinValue(g, gkc);
+                        if (minV < 0) {
+                            continue;
+                        }
+                        if (minV > entryMaxValue) {
+                            continue;
+                        }
+                        long maxV = peekDenseKeyMaxValueUpperBound(g, gkc);
+                        if (maxV > entryMaxValue) {
+                            return -1;
+                        }
+                    }
                     total += getDenseGenKeyCount(g, gkc);
                 } else if (!genLookup.notContainKey(valueMem, g, requestedKey)) {
+                    int activeKeyCount = -gkc;
+                    if (entryMaxValue >= 0) {
+                        long minV = peekSparseKeyMinValue(g, activeKeyCount);
+                        if (minV < 0) {
+                            continue;
+                        }
+                        if (minV > entryMaxValue) {
+                            continue;
+                        }
+                        long maxV = peekSparseKeyMaxValueUpperBound(g, activeKeyCount);
+                        if (maxV > entryMaxValue) {
+                            return -1;
+                        }
+                    }
                     total += getSparseGenKeyCount(g);
                 }
             }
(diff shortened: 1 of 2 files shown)


mtopolnik (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 171 / 201 (85.07%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/AbstractPostingIndexReader.java|171|201|85.07%|

PR #7091, opened 2026-05-12
fix(core): fix POSTING-indexed queries dropping rows after partition squash

bluestreak01:
## Summary

- `squashSplitPartitions` grows the squash target's `.d` files via `FrameAlgebra.append` but never reseals the target partition's POSTING indexes. Any indexed column whose `colTop` sat below the post-squash row count -- i.e. the column has `.d` data inside the partition, e.g. a column added between earlier seal points -- had no chain entries for the rows merged in by squash; an indexed `WHERE col = v` quietly dropped those rows while a non-indexed scan returned them.
- Reseal the target's POSTING indexes from `.d` right before `columnVersionWriter.commit()`, then point each POSTING indexer back at `lastOpenPartitionTs` via a new helper `restorePostingIndexersToLastPartition()` so the next append doesn't go to the just-sealed squash target. The same helper replaces the inline re-target block that previously lived at the end of `sealPostingIndexesForO3Partitions`.
- Both commit paths reach `squashSplitPartitions` via `housekeep` (`commitWalInsertTransactions:1505` and the non-WAL `commit:4505`), and the existing `sealPostingIndexesForO3Partitions` runs inside `finishO3Commit` *before* `housekeep`, so it cannot cover partitions the post-commit squash subsequently grows. The fix lands at the shared `squashSplitPartitions` entry point and therefore protects WAL and non-WAL alike. The reproducer was captured by `FuzzRunner.runFuzz`, which applies identical transactions to a non-WAL table, a WAL table, and a WAL-parallel table and asserts the three return identical query results; the failure surfaced as a non-WAL/WAL output divergence with the indexed `WHERE` predicate dropping rows on the non-WAL side. We do not have a direct WAL-only repro, but WAL is not architecturally exempt from the same `colTop`-below-post-squash-size pattern.

## Test plan

- `WalWriterFuzzTest#testConvertPartitionToParquet` with the captured reproducer seed (`7873130690938663L, 1778548286078L`): the cross-validation assertion between the non-WAL and WAL tables passes 10/10 on this branch; on master it fails 13-20% of runs.
- `WalWriterFuzzTest#testConvertPartitionToParquet` random seeds: 20/20.
- `WalWriterFuzzTest` whole class: 24/24 (1 pre-existing skip).
- `O3Test`, `O3FailureTest`, `O3SplitPartitionTest`, `O3MaxLagTest`: 247/247 (1 pre-existing skip).
- POSTING-index suite (`PostingIndexNativeTest`, `PostingIndexOracleTest`, `PostingSealPurgeTest`, `PostingIndexCriticalIssuesTest`, `PostingIndexConcurrencyTest`, `PostingIndexOomFallbackTest`, `PostingIndexChainWriterTest`, `PostingIndexAdaptiveDeltaThresholdTest`, `PostingIndexChainTest`): 130/130 (1 pre-existing skip).
- `O3SquashPartitionTest`, `O3PartitionPurgeTest`, `AlterTableConvertPartitionTest`, `IndexOnSymbolColumnTest`: 112/112 (1 pre-existing skip).

## Tradeoffs

- The reseal walks every POSTING-indexed column and rebuilds its chain from `.d` for every squash; that is on the housekeep path so it is off the user commit's critical path, but it does turn a previously-free squash into work proportional to the partition's row count for POSTING-indexed tables. BITMAP-indexed tables are unaffected.
- The added re-target helper is a refactor of code that already existed inline; the second call site (squash) is the new caller. Behaviour at the original `sealPostingIndexesForO3Partitions` site is unchanged.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -11904,19 +11904,27 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
             openLastPartition();
         }
 
-        // The squash grew the target partition's .d files via FrameAlgebra.append,
-        // which uses its own short-lived IndexWriter to append POSTING entries
-        // during the data copy. That writer leaves a chain head tagged with its
-        // own internal sealTxn rather than the table txn, so for any indexed
-        // column whose columnTop sits below the post-squash row count -- i.e.
-        // the column has .d data covering some prefix of the merged-in rows --
-        // readers still resolve through the pre-squash chain head, which has
-        // no entries for the rows squash merged in. Reseal the target
-        // partition's POSTING indexes from .d so the published chain head
-        // covers the full post-squash row range with a txn tag consistent with
-        // the upcoming txWriter.commit() below. After seal, point the table's
-        // indexers back at the active partition so the next append uses the
-        // active partition's index files rather than the just-sealed target.
+        // The squash grew the target partition's .d files via FrameAlgebra.append.
+        // FrameAlgebra's short-lived IndexWriter calls commit() after the per-row
+        // add() loop, and that commit() does publish entries to the chain: via
+        // extendHead in the non-copy squash (chain already open, head sealTxn
+        // matches) or via appendNewEntry with txnAtSeal=0 in the copy squash
+        // (fresh chain; pendingTxnAtSeal is never set, so the fallback at
+        // PostingIndexWriter#publishToChain fires). The IndexWriter never sees
+        // configureCovering, however, so coverCount=0 when captureCoverEndOffsets
+        // runs and the new gens land with an empty cover footer. For COVERING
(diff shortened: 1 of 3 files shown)


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 293 / 348 (84.20%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|234|287|81.53%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|45|47|95.74%|
|:large_blue_circle:|io/questdb/cairo/O3CopyJob.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/O3OpenColumnJob.java|11|11|100.00%|

Issue #1, opened 2014-08-09
Event Appender


bluestreak01:
An appender to work in Complex Event Processing environment and implement send-and-forget pattern.

As a developer I want:
- Log events with constant and low thread handover time
- Zero-GC logging
- Back-pressure sensitive logging. e.g. I would prefer to lose an event rather that slow down processing.

PR #7095, opened 2026-05-13
chore(core): fix RENAME COLUMN silently dropping POSTING value file under transient read error

bluestreak01:
## Summary

`TableWriter.linkPostingIndexAuxFiles` (column rename) and
`TableWriter.copyOrRebuildColumnIndexes` (parquet conversion) call
`PostingIndexUtils.readSealTxnFromKeyFile` to resolve the live `sealTxn`
that names the `.pv` value file to hardlink. The function returns `-1`
for several distinct conditions: file absent, empty chain (no head
entry yet), format version mismatch, or both header pages failing
seqlock validation across every retry attempt. Both call sites treated
every `-1` uniformly as "nothing to link" and fell through silently.

Under the writer lock the seqlock cannot actually observe a torn read
-- the file is quiescent. A `-1` from the seqlock-exhaustion branch
therefore signals an I/O error: a `pread` returning a negative value,
or a `FilesFacade` wrapper (e.g. `FailureFileFacade` in the fuzz
harness) injecting one. The silent skip then commits the rename with
`.d` + `.pk` on the destination column and no `.pv`. The chain in the
linked `.pk` still references `.pv.<columnNameTxn>.<sealTxn>`, so any
later indexed predicate against the renamed column fails opening a file
that was never linked. The CI reproducer surfaced this as:

```
CairoException: [2] could not open, file does not exist:
.../testConvertPartitionToParquet_nonwal~/2022-02-26.76/new_col_8.pv.101.4
```

## Change

`PostingIndexUtils.readLiveSealTxnFromKeyFileOrThrow` is a strict
variant for the writer-locked call sites. It calls
`readSealTxnFromKeyFile`, and on `-1` with the `.pk` file present,
scans the partition for any `.pv.<columnNameTxn>.*` file. If one
exists, the chain has sealed data and the `-1` must be an I/O error:
the function throws `CairoException.critical` with the `.pk` path. If
no `.pv` is present, the chain is genuinely empty and `-1` is returned
-- the caller's existing skip is correct in that case.

Both writer-locked call sites switch to the strict variant.
`renameColumn` already wraps its work in `throwDistressException`, so
the throw distresses the writer and the pool replaces it instead of
serving the next operation against a corrupted on-disk state.

## Effects

Positive: the broken-column window during `RENAME COLUMN` and parquet
conversion on POSTING-indexed columns closes. Production exposure was
narrow but real -- any `pread` returning a negative value during the
`.pk` header read window would silently corrupt the operation; this
scope now fails loudly instead.

Tradeoffs:
- A previously-tolerated transient error now distresses the writer.
  The application sees a critical exception where before it would see
  a successful rename followed by a later spurious "file does not
  exist". Total error count is the same; the timing and surface
  change, and the writer is replaced from the pool rather than
  silently carrying corruption forward.
- The strict variant pays one extra directory scan on the `-1` path.
  The success path is unchanged. On the failure path the scan cost is
  dwarfed by the I/O error that triggered it.

## Test plan

- `testConvertPartitionToParquetRenameMissingPostingValue` captures
  seeds from the CI failure of `testConvertPartitionToParquet` and
  reproduces the rename failure deterministically. On master with
  these seeds the test fails with the "file does not exist" error
  above; on this branch the harness's `FailureFileFacade` still
  injects the `.pk` read failure, the strict variant now surfaces it
  as a `CairoException`, and `AbstractFuzzTest` records "expected IO
  failure observed".
- `WalWriterFuzzTest` (25 tests) and `PostingIndexCriticalIssuesTest`
  (36 tests) pass under the change locally.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/ColumnPurgeOperator.java
+++ b/core/src/main/java/io/questdb/cairo/ColumnPurgeOperator.java
@@ -56,6 +56,7 @@ public class ColumnPurgeOperator implements Closeable {
     private final int pathRootLen;
     private final TableWriter purgeLogWriter;
     private final ScoreboardUseMode scoreboardUseMode;
+    private final SealedPostingFileScanProbe sealedPostingFileScanProbe = new SealedPostingFileScanProbe();
     private final String updateCompleteColumnName;
     private final int updateCompleteColumnWriterIndex;
     private long longBytes;
@@ -214,20 +215,37 @@ public class ColumnPurgeOperator implements Closeable {
             if (ff.exists(IndexFactory.keyFileName(indexType, path, columnName, columnVersion))) {
                 return true;
             }
-            // For POSTING, the live .pv has its sealTxn recorded in .pk. We need to check that file.
-            // For BITMAP, the second arg (sealTxn) is ignored.
-            long sealTxn = columnVersion;
             if (IndexType.isPosting(indexType)) {
                 path.trimTo(pathTrimToPartition);
                 long fromPk = PostingIndexUtils.readSealTxnFromKeyFile(
                         ff, PostingIndexUtils.keyFileName(path, columnName, columnVersion));
                 if (fromPk >= 0) {
-                    sealTxn = fromPk;
+                    // .pk gave us the live sealTxn; probe the exact .pv path.
+                    path.trimTo(pathTrimToPartition);
+                    if (ff.exists(PostingIndexUtils.valueFileName(path, columnName, columnVersion, fromPk))) {
+                        return true;
+                    }
+                } else {
+                    // .pk is gone or unreadable -- cannot resolve the live
+                    // sealTxn, and the previous fallback to
+                    // sealTxn = columnVersion probed a path that almost
+                    // never matched. Scan the partition for any orphan
+                    // .pv.<columnVersion>.<sealTxn> or
+                    // .pc<N>.<columnVersion>.<C>.<sealTxn>; if one exists
+                    // the purge is not complete and must retry, otherwise
+                    // the caller's "files all gone" shortcut leaks them.
+                    sealedPostingFileScanProbe.of(columnVersion);
+                    PostingIndexUtils.scanSealedFiles(ff, path, pathTrimToPartition, columnName, sealedPostingFileScanProbe);
+                    if (sealedPostingFileScanProbe.hasMatch()) {
+                        return true;
+                    }
                 }
-            }
-            path.trimTo(pathTrimToPartition);
-            if (ff.exists(IndexFactory.valueFileName(indexType, path, columnName, columnVersion, sealTxn))) {
-                return true;
+            } else {
+                // BITMAP: valueFileName ignores the sealTxn arg.
+                path.trimTo(pathTrimToPartition);
+                if (ff.exists(IndexFactory.valueFileName(indexType, path, columnName, columnVersion, columnVersion))) {
+                    return true;
+                }
             }
         }
         // Always check legacy .k/.v
@@ -544,6 +562,42 @@ public class ColumnPurgeOperator implements Closeable {
     private void setUpPartitionPath(int timestampType, int partitionBy, long partitionTimestamp, long partitionTxnName) {
         path.trimTo(pathTableLen);
         TableUtils.setPathForNativePartition(path, timestampType, partitionBy, partitionTimestamp, partitionTxnName);
+    }
+
+    /**
+     * Reusable {@link PostingIndexUtils.SealedFileVisitor} that flags
+     * whether any sealed POSTING file (.pv or .pc&lt;N&gt;) for a target
+     * columnVersion exists in the scanned partition. Stateful but
+     * non-allocating: callers call {@link #of(long)} to reset state for
+     * a new probe, run {@link PostingIndexUtils#scanSealedFiles}, then
+     * read {@link #hasMatch()}.
+     */
+    private static final class SealedPostingFileScanProbe implements PostingIndexUtils.SealedFileVisitor {
+        private long columnVersion;
+        private boolean hasMatch;
+
+        public boolean hasMatch() {
+            return hasMatch;
+        }
+
+        public void of(long columnVersion) {
+            this.columnVersion = columnVersion;
+            this.hasMatch = false;
+        }
+
+        @Override
+        public void onCoverDataFile(int includeIdx, long postingColumnNameTxn, long coveredColumnNameTxn, long sealTxn) {
+            if (postingColumnNameTxn == columnVersion) {
+                hasMatch = true;
+            }
+        }
+
+        @Override
+        public void onValueFile(long postingColumnNameTxn, long sealTxn) {
+            if (postingColumnNameTxn == columnVersion) {
+                hasMatch = true;
+            }
+        }
     }
 
     public enum ScoreboardUseMode {
--- a/core/src/main/java/io/questdb/cairo/O3OpenColumnJob.java
+++ b/core/src/main/java/io/questdb/cairo/O3OpenColumnJob.java
@@ -27,7 +27,6 @@ package io.questdb.cairo;
 import io.questdb.MessageBus;
 import io.questdb.cairo.idx.IndexFactory;
 import io.questdb.cairo.idx.IndexWriter;
-import io.questdb.cairo.idx.PostingIndexUtils;
 import io.questdb.cairo.vm.api.MemoryMA;
 import io.questdb.log.Log;
 import io.questdb.log.LogFactory;
@@ -1870,8 +1869,7 @@ public class O3OpenColumnJob extends AbstractQueueConsumerJob<O3OpenColumnTask> 
                     );
                 } else {
                     dstKFd = openRW(ff, IndexFactory.keyFileName(indexType, pathToNewPartition.trimTo(pNewLen), columnName, columnNameTxn), LOG, tableWriter.getConfiguration().getWriterFileOpenOpts());
-                    long valueTxn = resolvePostingValueFileTxn(ff, dstKFd, indexType, columnNameTxn);
-                    dstVFd = openRW(ff, IndexFactory.valueFileName(indexType, pathToNewPartition.trimTo(pNewLen), columnName, columnNameTxn, valueTxn), LOG, tableWriter.getConfiguration().getWriterFileOpenOpts());
+                    dstVFd = openRW(ff, IndexFactory.valueFileName(indexType, pathToNewPartition.trimTo(pNewLen), columnName, columnNameTxn, columnNameTxn), LOG, tableWriter.getConfiguration().getWriterFileOpenOpts());
                 }
             }
         } catch (Throwable e) {
@@ -2210,8 +2208,7 @@ public class O3OpenColumnJob extends AbstractQueueConsumerJob<O3OpenColumnTask> 
                         );
                     } else {
                         dstKFd = openRW(ff, IndexFactory.keyFileName(indexType, pathToNewPartition.trimTo(pNewLen), columnName, columnNameTxn), LOG, tableWriter.getConfiguration().getWriterFileOpenOpts());
-                        long valueTxn = resolvePostingValueFileTxn(ff, dstKFd, indexType, columnNameTxn);
-                        dstVFd = openRW(ff, IndexFactory.valueFileName(indexType, pathToNewPartition.trimTo(pNewLen), columnName, columnNameTxn, valueTxn), LOG, tableWriter.getConfiguration().getWriterFileOpenOpts());
+                        dstVFd = openRW(ff, IndexFactory.valueFileName(indexType, pathToNewPartition.trimTo(pNewLen), columnName, columnNameTxn, columnNameTxn), LOG, tableWriter.getConfiguration().getWriterFileOpenOpts());
                     }
                 }
             }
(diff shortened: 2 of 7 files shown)


bluestreak01:
# PR #7095 Review (level 2)

`chore(core): fix RENAME COLUMN silently dropping POSTING value file under transient read error`

## Critical

### 1. Spurious throw on V2_NO_HEAD → READ_FAILURE retry path (in-diff)

**File:** `core/src/main/java/io/questdb/cairo/idx/PostingIndexUtils.java:1275-1331`

Walk-through:

1. `firstAttempt = readSealTxnFromKeyFileTriState(...)` returns `V2_NO_HEAD` (`-1`). The chain is genuinely empty — this requires a clean format-version read, a stable seqlock pair, and sane region descriptors, so the answer is authoritative.
2. The `firstAttempt == V2_NO_HEAD && !ff.exists(keyFilePath)` branch at line 1279 fires only when the file has disappeared. Under the writer lock that's impossible — so we fall through.
3. `secondAttempt = readSealTxnFromKeyFileTriState(...)` returns `READ_FAILURE` (FailureFileFacade one-shot, pread EINTR, etc.).
4. `savedErrno`, `ff.exists`, `hasInitialisedKeyFileHeader` all pass (file is genuinely there and well-formed).
5. `scanSealedFiles` finds the pre-seal `.pv.<colTxn>.0` that ADD COLUMN created (see `PostingIndexWriter`; the live unsealed values file sits at `sealTxn = 0` until the first seal). The visitor filters by `filePostingColumnNameTxn == postingColumnNameTxn`, which matches the current column, so `hasPv[0] = true`.
6. **Throw fires**, even though the first successful read already proved the chain has no sealed data to lose.

This reintroduces the bug class the PR is trying to eliminate, just inverted: instead of silently dropping a real `.pv`, the code loudly fails a rename that should have succeeded. The new `testRenamePostingColumnBeforeFirstSeal` test exercises exactly the pre-seal-rename shape; if its FF injection pattern hits the second `.pk` open rather than the first, the test would surface this spurious throw.

**Suggested fix:** short-circuit on V2_NO_HEAD from the first attempt — under the writer lock the chain state cannot change between reads, so a clean V2_NO_HEAD on the first attempt is authoritative.

```java
if (firstAttempt == V2_NO_HEAD) {
    return -1;
}
// only retry on READ_FAILURE from here on
```

Drop the dead `&& !ff.exists(...)` defensive branch at line 1279-1282 in the process — it's unreachable under the writer lock either way.

### 2. Head commit message is `bugfixes` with no body

**File:** git `HEAD` (`0aa0717d4e`)

CLAUDE.md: commit titles must be "short (up to 50 chars) and descriptive in plain English", and "always include a full long-form description in the commit message body". The current head commit is the literal string `bugfixes` — no body, non-descriptive. Squash/rewrite before merge.

## Moderate

### 3. PR title uses `chore(core)` for a Bug-labeled fix

The change is labeled `Bug` and turns silent data loss into a loud failure. Per Conventional Commits this should be `fix(core):`, not `chore(core):`. The description also leaks implementation detail ("transient read error") rather than user-facing impact. Consider e.g. `fix(core): RENAME COLUMN on indexed columns no longer loses data on I/O errors`.

### 4. Zero direct unit tests for the new public API

**File:** `core/src/main/java/io/questdb/cairo/idx/PostingIndexUtils.java:1267`

`readLiveSealTxnFromKeyFileOrThrow` has ~10 decision branches (firstAttempt sealTxn / V2_NO_HEAD / READ_FAILURE × secondAttempt sealTxn / V2_NO_HEAD / READ_FAILURE × file-exists × header-initialised × hasPv). A grep across `core/src/test` finds zero direct callers — every branch is exercised only transitively through two seeded fuzz tests.

The fuzz harness (`FuzzRunner.java:289-321`) classifies any `CairoException` as "expected" once `ff.failureGenerated()` increases, then retries the transaction. Both the bug (silent .pv miss → later "file does not exist") and the fix (loud throw at rename time) leave the table briefly suspended, and both get swallowed by that catch — JUnit cannot distinguish them. The seeded tests verify that the harness completes, not that the new strict path triggered.

A targeted test class staging real `.pk`/`.pv` files in a tmp dir (pattern: `PostingIndexConcurrencyTest.seedConsistentKeyFile` ~line 640) that drives each branch and asserts the specific `CairoException` message/errno would close this gap and catch finding #1.

### 5. Tri-state region-descriptor sanity check is untested

**File:** `core/src/main/java/io/questdb/cairo/idx/PostingIndexUtils.java:1179-1186`

The new check `regionBase < KEY_FILE_RESERVED || regionLimit < regionBase` distinguishes a torn read from V2_NO_HEAD, and the split between `headEntryOffset < 0` (V2_NO_HEAD) and `headEntryOffset < regionBase || >= regionLimit` (READ_FAILURE) is also new. No test verifies any of these branches.

`readSealTxnFromKeyFdTriState` and `readSealTxnFromKeyFileTriState` are package-private, so any test must live in `io.questdb.test.cairo.idx`.

### 6. Seeded fuzz tests are likely flaky, no disclaimer

**Files:** `WalWriterFuzzTest.java:241` and `ReplaceInsertFuzzTest.java:40`

The sibling test `testConvertPartitionToParquetPostingSquashReseal` (same file, ~line 200) explicitly documents "the shared worker pool adds residual non-determinism, so this is not a fully deterministic reproducer; on master it failed roughly 80% of runs with these seeds." Both new tests use the same shared worker pool (`AbstractFuzzTest.java:57`) and identical fuzz-probability shape. The new tests' comments do not include the flakiness disclaimer.

Either pair them with deterministic unit tests (preferable; see #4) or copy the flakiness note and add the `Flaky Test` GitHub label.

### 7. Method ordering violates CLAUDE.md alphabetical/visibility rule

**File:** `core/src/main/java/io/questdb/cairo/idx/PostingIndexUtils.java:1124, 1229, 1267`

- `readLiveSealTxnFromKeyFileOrThrow` (public, line 1267) sorts as `readL...` which is alphabetically before `readSealTxnFromKeyFd` (public, line 1082). It belongs above line 1082.
- `readSealTxnFromKeyFdTriState` (line 1124) and `readSealTxnFromKeyFileTriState` (line 1229) are package-private and currently interleaved with public methods. The file's existing package-private-static block sits separately around lines 1683/1699 (`efLowBytesAligned`, `readBitsWord`); the tri-state helpers should move there to honour the visibility grouping.

## Minor

### 8. Dead defensive branch under writer lock

**File:** `core/src/main/java/io/questdb/cairo/idx/PostingIndexUtils.java:1279-1282`

`if (firstAttempt == V2_NO_HEAD && !ff.exists(keyFilePath))` is unreachable under the writer lock (the file cannot disappear after a successful read). With the fix from #1, this branch goes away entirely.

### 9. `savedErrno` may carry over from an intervening `close()`

**File:** `core/src/main/java/io/questdb/cairo/idx/PostingIndexUtils.java:1299`

`readSealTxnFromKeyFileTriState` returns after running `ff.close(fd)` in a finally block; `savedErrno = ff.errno()` then reads the OS thread-local. On Linux `close()` does not normally set errno on success, but `FilesFacade` wrappers (and `FailureFileFacade` in particular — which trips on a counter, not a real syscall) can leave `errno` carrying an unrelated prior value. Result: the `CairoException.critical(savedErrno)` thrown at line 1328 may report a misleading errno. Cosmetic — does not affect control flow.

### 10. `readSealTxnFromKeyFile` docstring drift

**File:** `core/src/main/java/io/questdb/cairo/idx/PostingIndexUtils.java:1087-1093`

Says "Returns -1 if the file cannot be opened or both metadata pages fail the seq-lock validation." Also returns -1 for V2_NO_HEAD, file-too-short, format-version mismatch, and out-of-region head pointer. Pre-existing but worth a one-line update now that the PR introduces the tri-state companion that callers may want to reach for.

### 11. PR labels incomplete

Currently `Bug`, `Core`. The diff touches WAL apply and on-disk file layout. Add `WAL` and `storage` from the CLAUDE.md label list.

- **`copyOrRebuildColumnIndexes` previously silently corrupted** — on master, the fall-through `linkSealTxn = columnNameTxn` produced `.pv.<colTxn>.<colTxn>` which `linkFile` would fail to link, surfacing a noisy `CairoException.critical(0).put("index files do not exist")` at the existing throw (TableWriter.java:4797). Not silent corruption in that specific call site. The PR's framing is accurate for `linkPostingIndexAuxFiles` (rename path) but overstated for parquet conversion. The new `linkSealTxn = ... : 0L` fallback is still a real functional improvement for pre-seal parquet conversion.
- **`.pc<N>.<col>.0.*` filter at `pvSealTxn=0`** — `PostingIndexWriter.openSidecarFilesForAppend` is called in pre-seal state with `this.sealTxn = 0`, so the filter correctly matches the pre-seal sidecar files. Verified. No bug.
- **Writer-lock vs concurrent `.pk` writers** — `PostingSealPurgeOperator.purge` only touches `.pv` and `.pc<N>` files, never `.pk`. `renameColumn` runs `commit()` first which joins all parallel indexer workers via `indexLatch.await()` before returning. No concurrent `.pk` mutation is possible during the strict read. The PR description's writer-lock claim holds.
- **Orphan `.d`/`.pk` hardlinks on throw** — when `linkPostingIndexAuxFiles` throws inside `renameColumn`, the destination `.d` and `.pk` hardlinks under the new column name are orphaned. Confirmed but pre-existing — any `linkFile` throw in this path leaks the same way. `_todo` + writer-pool replacement on distress restores metadata consistency but does not unlink the orphans. Out of scope for this PR; worth filing as a separate cleanup.
- **`CairoException.put(LPSZ)` aliasing the partition `path`** — `Utf16Sink.put(Utf8Sequence)` decodes-and-copies eagerly into the message sink. Safe.

## Summary

**Request changes.** 23 draft findings verified or dismissed; 11 reported (1 critical, 6 moderate, 4 minor) and 5 downgraded as false positives. 9 in-diff, 2 are PR metadata. The core design — distinguishing READ_FAILURE from V2_NO_HEAD in writer-locked paths — is sound, but the V2_NO_HEAD branch in the strict reader retries instead of short-circuiting, which converts a clean empty-chain answer into a spurious throw the first moment a second-read fault injection hits (exactly the conditions the new fuzz tests are built to surface). Fix #1 first; #2 (commit message) and #3 (title) are merge-blockers under CLAUDE.md conventions; the rest are cleanup that the PR is the right place to land.

---
Reviewed by Claude Code (Opus 4.7, level 2 review).

kafka1991 (changes requested):
Same silent-skip bug remains at four other callsites:

- **`ColumnPurgeOperator.java:222`** — on transient `.pk` read failure, `existsIndexFile` returns false, caller marks the purge task complete (line 322-325), real `.pv.<cv>.<realSealTxn>` and sealed sidecars leak permanently.

- **`O3OpenColumnJob.java:3332`** (`resolvePostingValueFileTxn`) — `-1` folds to `0`, caller opens the wrong `.pv.<cv>.0`. The method's own docstring warns this leads to SIGSEGV during sidecar rebuild.

- **`TableWriter.java:4205`** — detach-time cleanup leaks one `.pv` into the detached partition dir. `.pci`/`.pc<N>` are fine via `scanSealedFiles`.

- **`TableWriter.java:4060`** — attachPartition validation throws "Index value file does not exist" when the real cause is a `.pk` read failure. Misleading error, no corruption.

bluestreak01:
Thanks for the review @kafka1991 — all four sites addressed in 107ab9e869.

- **#1 ColumnPurgeOperator.java:222** — when `readSealTxnFromKeyFile` returns -1, `existsIndexFile` now scans the partition for any `.pv.<cv>.*` or `.pc<N>.<cv>.<C>.*` and returns true if anything matches, so the purge keeps retrying instead of silently marking the row complete. Red test `testPurgeMarksTaskCompleteAndLeaksPostingValueWhenKeyFileMissing` reproduces the leak on master (manual `.d`/`.o`/`.pk` delete leaving a real `.pv.0`, second pass marks complete and leaves the orphan); after the fix it passes. The new `SealedFileVisitor` is a reusable single-instance field, so the hot path stays allocation-free.

- **#2 O3OpenColumnJob.java:3332** — confirmed the POSTING branch is dead code today (all three call sites only invoke it from the non-POSTING else after `isPosting` routes through `setO3PathContext`/`openFromO3Context`). Deleted the function and inlined `columnNameTxn` at the three sites, removed the now-unused `PostingIndexUtils` import. The SIGSEGV trap goes with the dead code.

- **#3 TableWriter.java:4205** — the visitor's `onValueFile` now removes any `.pv` whose `postingColumnNameTxn` matches, so the explicit conditional removal is no longer load-bearing and the `.pk` read is best-effort. No more leaked `.pv` when `readSealTxnFromKeyFile` returns -1. Visitor is also a reusable single-instance field on `TableWriter`.

- **#4 TableWriter.java:4060** — swapped in `readLiveSealTxnFromKeyFileOrThrow` (the strict reader from earlier in this branch). Persistent `.pk` read failure with a real `.pv` present now throws "could not read live sealTxn from posting key file" with the `.pk` path and errno, instead of the misleading "Index value file does not exist". V2_NO_HEAD falls back to `sealTxn=0`, which is the on-disk pre-seal convention.

No new test for #3 — the existing fuzz harness exercises the detach/attach paths and the bug class is identical to #1.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 75 / 113 (66.37%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexUtils.java|28|53|52.83%|
|:large_blue_circle:|io/questdb/cairo/ColumnPurgeOperator.java|16|22|72.73%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|28|35|80.00%|
|:large_blue_circle:|io/questdb/cairo/O3OpenColumnJob.java|3|3|100.00%|

kafka1991 (approved):

PR #7096, opened 2026-05-14
chore(core): fix POSTING-indexed queries returning wrong rows after a failed commit

kafka1991:
## Summary

- POSTING-indexed `WHERE col = v` against a WAL table occasionally returned a row whose `.d` value did not match `v`. Surfaced as a non-WAL vs WAL divergence in `WalWriterFuzzTest#testWalWriteRollbackHeavy` once writer distress (fuzz IO injection, very short `cairo.wal.apply.table.time.quota`) caused multiple commits to share a single chain entry via the canSkipRebuild fast-path.
- Root cause: PR #7087's `extendHead` mutates the chain entry header in place but never updates `TXN_AT_SEAL`. Recovery's `entry.TXN_AT_SEAL > committedTxn` predicate cannot identify partial-publish failures because the field stays pinned at the first commit's value while later tail gens carry uncommitted data. The reader picks the entry, walks the in-flight tail, and surfaces stale (key, rowId) pairs.

## Change

- Move `TXN_AT_SEAL` from entry header to per-slot. Each gen-dir slot tags itself with the anticipated table `_txn` at write time. Recovery does per-gen tail trim against `slot.TXN_AT_SEAL > committedTxn`; reader cursors skip in-flight tail gens for snapshot isolation. Per-slot `MAX_VALUE` lets recovery and per-gen read recompute the visible cumulative max in O(1).
- `applyHeadTrim` uses GEN_COUNT-first store order (inverse of `extendHead`'s GEN_COUNT-last). Trim is a shrink, so the latch field is written first to prevent a concurrent reader from seeing an old large GEN_COUNT with a new small VALUE_MEM_SIZE.
- `sealPostingIndexForPartition`'s `setNextTxnAtSeal` unified to `getTxn() + 1L` (matching `O3CopyJob`). The earlier `getTxn()` trade-off was tied to entry-header `TXN_AT_SEAL`; with the field moved to slot[0] and `publishPendingPurges` already clamping `toTableTxn` to `getTxn()`, the scoreboard concern no longer applies. Slot `TXN_AT_SEAL` is now monotonic within an entry.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/O3CopyJob.java
+++ b/core/src/main/java/io/questdb/cairo/O3CopyJob.java
@@ -820,8 +820,8 @@ public class O3CopyJob extends AbstractQueueConsumerJob<O3CopyTask> {
                         indexWriter.of(tableWriter.getConfiguration(), dstKFd, dstVFd, row == 0, indexBlockCapacity);
                     }
                 }
-                updateIndex(dstFixAddr, Math.abs(dstFixSize), indexWriter, dstIndexOffset / Integer.BYTES, dstIndexAdjust);
                 indexWriter.setNextTxnAtSeal(tableWriter.getTxn() + 1L);
+                updateIndex(dstFixAddr, Math.abs(dstFixSize), indexWriter, dstIndexOffset / Integer.BYTES, dstIndexAdjust);
                 indexWriter.commit();
             } finally {
                 if (closed) {
--- a/core/src/main/java/io/questdb/cairo/O3CopyJob.java
+++ b/core/src/main/java/io/questdb/cairo/O3CopyJob.java
@@ -821,6 +821,7 @@ public class O3CopyJob extends AbstractQueueConsumerJob<O3CopyTask> {
                     }
                 }
                 updateIndex(dstFixAddr, Math.abs(dstFixSize), indexWriter, dstIndexOffset / Integer.BYTES, dstIndexAdjust);
+                indexWriter.setNextTxnAtSeal(tableWriter.getTxn() + 1L);
                 indexWriter.commit();
             } finally {
                 if (closed) {
--- a/core/src/main/java/io/questdb/cairo/O3PartitionJob.java
+++ b/core/src/main/java/io/questdb/cairo/O3PartitionJob.java
@@ -3377,6 +3377,14 @@ public class O3PartitionJob extends AbstractQueueConsumerJob<O3PartitionTask> {
                             indexWriter.setMaxValue(newPartitionSize - 1);
                         }
 
+                        // Tag the chain entry with the upcoming txn so that if
+                        // the surrounding O3 / parquet-rewrite path crashes
+                        // before txWriter.commit lands, recoveryDropAbandoned
+                        // on the next writer-open drops this entry. Without
+                        // the tag, slot[0].TXN_AT_SEAL defaults to 0 and the
+                        // entry would be undroppable, surfacing stale rows.
+                        // No-op on BitmapIndexWriter.
+                        indexWriter.setNextTxnAtSeal(tableWriter.getTxn() + 1L);
                         if (IndexType.isPosting(indexType)) {
                             indexWriter.commitDense();
                         } else {
--- a/core/src/main/java/io/questdb/cairo/TableReader.java
+++ b/core/src/main/java/io/questdb/cairo/TableReader.java
@@ -371,6 +371,10 @@ public class TableReader implements Closeable, SymbolTableSource {
         final long partitionTxn = txFile.getPartitionNameTxn(partitionIndex);
         IndexReader indexReader = getIndexReaderIfExists(partitionIndex, columnIndex, direction);
         if (indexReader != null) {
+            // Single choke point for refreshing the scoreboard pin on cached
+            // readers. TableReader.txn advances through several paths
+            // (goActive / reload / ...); setting it here covers all of them.
+            indexReader.setPinnedTableTxn(txn);
             if (
                     !indexReader.isOpen()
                             || indexReader.getColumnTxn() != columnNameTxn
@@ -1055,7 +1059,8 @@ public class TableReader implements Closeable, SymbolTableSource {
                         getColumnTop(columnBase, columnIndex),
                         metadata,
                         columnVersionReader,
-                        partitionTimestamp
+                        partitionTimestamp,
+                        txn
                 );
                 if (direction == IndexReader.DIR_BACKWARD) {
                     indexes.setQuick(globalIndex, reader);
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -7640,7 +7640,7 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
                             setStateForTimestamp(other, newSplitPartitionTimestamp);
                             ff.mkdir(other.$(), configuration.getMkDirMode());
                             try (Frame targetFrame = frameFactory.createRW(other, newSplitPartitionTimestamp, metadata, columnVersionWriter, 0)) {
-                                FrameAlgebra.append(targetFrame, sourceFrame, newPrevPartitionSize, prevPartitionSize, configuration.getCommitMode());
+                                FrameAlgebra.append(targetFrame, sourceFrame, newPrevPartitionSize, prevPartitionSize, txWriter.getTxn() + 1L, configuration.getCommitMode());
                             }
                         }
                         addPhysicallyWrittenRows(prevPartitionSize - newPrevPartitionSize);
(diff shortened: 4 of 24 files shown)


questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=233212)
**Tested Commit:** `a6d2a68318fd27f3f14064283979256f75aca426`

glasstiger:
Claude review:
```
⏺ Code Review: PR #7096 — chore(core): fix POSTING-indexed queries returning wrong rows after a failed commit

  Repo: questdb/questdb (OSS). Enterprise-only review was requested but the URL points at the OSS repo, so this reviews the OSS PR. 798 additions / 198 deletions across 18 files.
  The change moves TXN_AT_SEAL from the chain-entry header into per-gen-dir slots, adds per-gen MAX_VALUE, and adds a recovery-time "trim in-flight tail gens" path.

  I verified every draft finding against the actual source in the checked-out posting_add_txn_commit branch.

  ---
  Critical
  
  1. applyHeadTrim corrupts the live gen-dir region for concurrent readers — covering indexes
  core/src/main/java/io/questdb/cairo/idx/PostingIndexChainWriter.java:647-657

  Step 1 of applyHeadTrim copies the cover footer to newFooterOffset = entryOffset + V2_ENTRY_HEADER_SIZE + keepGenCount * GEN_DIR_ENTRY_SIZE. Because keepGenCount < oldGenCount,
  that offset is the start of slot[keepGenCount] — inside the OLD gen-dir region.

  Trace of the race (covering posting index, distress/recovery path):
  - recoveryDropAbandoned republishes the chain-header seqlock only after the whole walk (PostingIndexChainWriter.java:541). During applyHeadTrim the seqlock is unchanged.
  - A concurrent reader in AbstractPostingIndexReader.readIndexMetadataFromChain (:855) picks the head, reads entryScratch.genCount (can still be OLD — step 2 hasn't run), then
  calls genLookup.snapshotMetadata(keyMem, entryScratch.genCount, ...) which reads all OLD slots. Slots [keepGenCount..] now hold footer longs reinterpreted as FILE_OFFSET/SIZE/etc.
  - The reader's stillStable re-check (:902) passes because the writer hasn't republished the header yet → it commits the garbage snapshot.
  - With the default pinnedTableTxn = Long.MAX_VALUE, computeVisibleGenCount returns the full OLD genCount (nothing is > MAX_VALUE), so the reader keeps the corrupted tail slots and
   later dereferences .pv at garbage offsets → wrong rows or SIGSEGV.
   
  The javadoc claim "readers using OLD count don't read the new-offset bytes" (:609) is false — OLD-count readers iterate exactly those slots. This trades the PR's target bug
  ("stale rows") for a crash window. Non-covering indexes are unaffected (step 1 is if (coverCount > 0)), and the window is narrow, but it is real and not covered by any test (all
  applyHeadTrim tests are single-threaded).

  Suggested fix: drive applyHeadTrim under the chain-header seqlock — bump it to an in-progress/odd state before mutating and republish after, so straddling readers retry — or stage
   the trimmed entry copy-on-write at a fresh region offset and publish it as the new head instead of mutating in place.

  ---
  Moderate
  
  2. On-disk layout changed incompatibly without a format-version bump
  core/src/main/java/io/questdb/cairo/idx/PostingIndexUtils.java:178,228 (GEN_DIR_ENTRY_SIZE 28→44, V2_ENTRY_HEADER_SIZE 64→56, every gen-dir/header offset shifted)

  V2_FORMAT_VERSION stays 2. PostingIndexChainWriter.openExisting (:412) accepts any file tagged version 2 and then reads it with the new offsets, so a posting index .pk created by
  an earlier build is silently misparsed → corruption/crash on upgrade. If POSTING indexes are unreleased/experimental this is likely acceptable, but it should be a conscious,
  stated decision in the PR description — confirm no on-disk v2 posting index can survive the upgrade, or bump the version / add a guard.
  
  ---
  Minor
  
  3. MAX_GEN_COUNT = 91 "fits in a 4 KB page" rationale is questionable
  PostingIndexUtils.java:200-204, PostingIndexWriter.java:2724-2729

  entrySize(91, coverCount) = 4060 + coverCount*8 exceeds 4096 once coverCount > 4, and flushAllPending seals on genCount >= MAX_GEN_COUNT only — it never considers coverCount.
  coverCount is coveredColumnNames.size() from the INCLUDE (...) clause and SqlParser enforces no upper bound. This is not a regression (the old 64 + 143*28 math was tighter,
  overflowing at coverCount > 3), and the design doc describes entries as variable-size and packed rather than page-bounded — so either the 4 KB-per-entry constraint isn't actually
  load-bearing (then the comments are misleading) or it is (then it's a pre-existing latent bug). Worth resolving the comment either way.

  4. publishToChain "assert" referenced in a comment does not exist
  PostingIndexWriter.java:3508 silently falls back to slotTxnAtSeal = 0 when pendingTxnAtSeal < 0, but the comment at :3723 claims "The publishToChain assert catches the never-set
  case". There is no assert. If an extendHead ever lands with pendingTxnAtSeal == -1 after slot[0] was stamped with a positive txnAtSeal, slot TXN_AT_SEAL goes non-monotonic and
  trips assert txnAtSeal >= prevTxnAtSeal in PostingGenLookup.snapshotMetadata (:261). Add the real assert the comment claims, or fix the comment.

  5. PR metadata — Title is chore(core): for a correctness fix; per QuestDB conventions this should be fix(core): unless chore is deliberate because POSTING is unreleased. The PR
  also has no labels (convention requires labels matching scope, e.g. Core, Bug, storage).

  6. Test gap — no test exercises a concurrent reader against an entry mid-applyHeadTrim (finding #1's window), and the covering-footer-relocation path in particular is only checked
   single-threaded.

  ---
  Downgraded (false positives)
  
  - "Reader-side computeVisibleGenCount per-gen filtering is broken — never trims." Not a bug: with the default pinnedTableTxn = Long.MAX_VALUE it intentionally returns the full
  genCount; the PR explicitly defers scoreboard-pin plumbing and relies on the writer-side recovery walk. Dormant by design.
  - "PostingIndexChainEntry.read returns txnAtSeal = 0 for genCount == 0, which recoveryDropAbandoned would treat as visible / wrongly drop." Production writers always pass genCount
   >= 1 (appendNewEntry/extendHead call sites confirm), and the genCount == 0 case is documented; not reachable in practice.
  - "publishToChain writes MAX_VALUE then storeFence then TXN_AT_SEAL, but snapshotMetadata reads both with no loadFence between." snapshotMetadata's torn reads are explicitly
  harmless — it fills a staging snapshot that is validated by the header-seqlock stillStable re-check before commitSnapshot on the normal publish path (where the writer does
  republish the header). The storeFence/loadFence pair specifically guards trimInFlightTailGens, which runs single-threaded w.r.t. other writers.

  ---
  Summary
  
  Request changes. One Critical concurrency bug: applyHeadTrim's in-place cover-footer copy overwrites live gen-dir slots that concurrent OLD-GEN_COUNT readers iterate, with no
  seqlock protection during the mutation — a SIGSEGV/wrong-row window for covering posting indexes on the recovery path. The core design (per-slot TXN_AT_SEAL, recovery trim) is
  sound and well-tested single-threaded, but the concurrent-reader safety argument in the applyHeadTrim javadoc is wrong. Also confirm the format-change-without-version-bump is
  intentional. 6 findings verified, 3 draft findings dropped as false positives
```

glasstiger (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 158 / 166 (95.18%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|4|5|80.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|8|9|88.89%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexChainWriter.java|81|86|94.19%|
|:large_blue_circle:|io/questdb/cairo/idx/AbstractPostingIndexReader.java|20|21|95.24%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexChainEntry.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/IndexFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/O3CopyJob.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexBwdReader.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/frm/FrameAlgebra.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/IndexReader.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/frm/file/ContiguousFileIndexedFrameColumn.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingGenLookup.java|13|13|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexFwdReader.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/frm/FrameColumn.java|1|1|100.00%|

PR #7099, opened 2026-05-14
chore(core): fix incorrect query results after writing non-ascii varchars to parquet

ideoma:
## Summary

- Queries against a parquet partition that was written entirely ASCII for a VARCHAR column and later received non-ASCII bytes returned wrong results: `Utf8Sequence.isAscii()` reported `true` for values containing non-ASCII bytes, so UTF-8 fast paths (length, UTF-8 to UTF-16 conversion, `Utf8Sink::put`) treated multi-byte sequences as single-byte ASCII and produced truncated or garbled output. Affects tables with parquet partitions created via `ALTER TABLE ... CONVERT PARTITION TO PARQUET` or a `FORMAT PARQUET` default.
- Cause: `ParquetUpdater::end` reused the source file's column-level `ascii` flag verbatim. A partition born all-ASCII kept `ascii: Some(true)` across an O3 rewrite that merged in non-ASCII data, and the decoder applies that flag to every aux entry on read.
- Fix: track per VARCHAR column whether the column is still all-ASCII, seeded from the old flag and flipped to `false` on the first non-ASCII aux entry during a write. The scan short-circuits once a column is `false`, so columns that already contain non-ASCII bytes pay no scan cost. The read-side fast path stays active for columns that genuinely stay all-ASCII across rewrites.

## Test plan

- New `ParquetTest#testVarcharAsciiFlagAfterRewriteWithNonAsciiAppend` reproducer: passes on this branch, fails on master.
- `WalWriterFuzzTest#testCreateTableAsParquet` and `#testConvertPartitionToParquet`: failing seed plus 3 random-seed runs each.
- `ParquetTest` whole class (98 tests) and Rust `parquet_write` unit tests (505 tests).


bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 59 / 61 (96.72%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/parquet_write/update.rs|59|61|96.72%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=233412)
**Tested Commit:** `283bc325232310bbe033dd02c564e068001bb8cf`

PR #7100, opened 2026-05-15
chore(qwp): expand fuzz tests with full wire-type coverage

bluestreak01:
## Summary

- `QwpIngressOracleFuzzTest` now drives BYTE, SHORT, INT, FLOAT, CHAR, UUID, LONG256, and TIMESTAMP_NANO through the oracle. Base columns are pre-declared in `createTargetTable` (pre-existing-column path); `injectExtra` adds five auto-create cases for the new types. The smallest `sf_max_bytes` pool entry is raised from 64 KiB to 256 KiB so a max-chunk frame fits the wider row.
- `QwpSenderFuzzTest` adds one column per new type to `colTypes` / `colNameBases` / `colValueBases`. New typed columns are always set on every row and excluded from the auto-create-extras pool; only the legacy STRING/DOUBLE slice is eligible for skipping or auto-create injection. Default `recvBufferSize` is raised from 8 KiB to 32 KiB so the wider per-flush frame fits.

## Type coverage added

| Wire type | Client setter used |
|-----------|--------------------|
| BYTE | `byteColumn` |
| SHORT | `shortColumn` |
| INT | `intColumn` |
| FLOAT | `floatColumn` |
| CHAR | `charColumn` |
| UUID | `uuidColumn` |
| LONG256 | `long256Column` |
| TIMESTAMP_NS | `timestampColumn(.., NANOS)` |

The non-public setters route through a `Sender` -> `QwpWebSocketSender` cast inside `QwpRow.publishTo`. The WS connection (`ws::`) always returns that impl, so the cast is safe.

Wire types still uncovered after this change: `DATE` and `GEOHASH`. No client setter exists at any layer for either.

## Notable design points

- `QwpRow.assertCell` handles `TIMESTAMP_NANO` ahead of the tag switch because `ColumnType.tagOf(TIMESTAMP_NANO) == TIMESTAMP`. BYTE/SHORT/CHAR mirror BOOLEAN's no-NULL semantics: the oracle requires producers to always set them.
- `QwpSenderFuzzTest.LEGACY_COLUMN_COUNT = 6` partitions the column-name array. Typed columns sit beyond that index, never skipped, never used as extras. This keeps the `LineData`/`TableData` single-default-per-column model intact when the alter-table thread converts a column across types with different NULL semantics (BYTE 0 renders as `0`; INT NULL renders as `null`).
- `addColumnValue`'s integer cases cap values to byte range. Narrowing conversions (INT -> BYTE) then preserve every stored value losslessly, and the oracle's yielded string still matches the cursor.
- `TableData.getDefaultValue` gains BYTE/SHORT -> `"0"` and INT/LONG -> `"null"` so that when the alter-table thread morphs a legacy column into an integer-family member, the oracle's default cell rendering matches the cursor's NULL/zero rendering.

## Tradeoffs

- Integer-family value bases stay <= 119 so narrowing conversions do not overflow. The numeric span the typed columns exercise is narrower than it could be in exchange for end-to-end test reliability.
- New typed columns in `QwpSenderFuzzTest` are always set on every row. Their wire emission, reordering, duplicate, and case-variation paths are exercised; their auto-create-only path is not.
- 32 KiB is more permissive than the old 8 KiB `recvBufferSize` default. Only `testLoadSmallBuffer` still pins a tight buffer to exercise the small-frame path explicitly.
- The new 256 KiB minimum `sf_max_bytes` is larger than the previous 64 KiB. Segment-rotation coverage is preserved by the remaining `{256 KiB, 1 MiB, 4 MiB}` entries.

## Test plan

- Run `mvn -pl core -Dtest=QwpIngressOracleFuzzTest test` across multiple seeds; 4/4 pass each time
- Run `mvn -pl core -Dtest=QwpSenderFuzzTest test` across multiple seeds; 28/28 pass each time
- Combined `mvn -pl core -Dtest='QwpIngressOracleFuzzTest,QwpSenderFuzzTest' test`; 32/32 pass

### DIFF
--- a/core/src/test/java/io/questdb/test/cutlass/line/tcp/load/TableData.java
+++ b/core/src/test/java/io/questdb/test/cutlass/line/tcp/load/TableData.java
@@ -33,8 +33,12 @@ import io.questdb.std.Rnd;
 
 import java.util.concurrent.atomic.AtomicLong;
 
+import static io.questdb.cairo.ColumnType.BYTE;
 import static io.questdb.cairo.ColumnType.DOUBLE;
 import static io.questdb.cairo.ColumnType.FLOAT;
+import static io.questdb.cairo.ColumnType.INT;
+import static io.questdb.cairo.ColumnType.LONG;
+import static io.questdb.cairo.ColumnType.SHORT;
 
 public class TableData {
     private final IntLongSortedList index = new IntLongSortedList();
@@ -137,8 +141,14 @@ public class TableData {
     }
 
     private String getDefaultValue(short colType) {
+        // NULL renderings must match CursorPrinter.printColumn for each type:
+        // - DOUBLE/FLOAT NULL: "null"
+        // - INT/LONG NULL: Numbers.append() prints "null"
+        // - BYTE/SHORT: no NULL sentinel; cursor renders the stored 0 as "0"
+        // - CHAR/UUID/LONG256/TIMESTAMP and string-like types: empty
         return switch (colType) {
-            case DOUBLE, FLOAT -> "null";
+            case DOUBLE, FLOAT, INT, LONG -> "null";
+            case BYTE, SHORT -> "0";
             default -> "";
         };
     }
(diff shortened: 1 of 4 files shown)


questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=233455)
**Tested Commit:** `a0583722c4c444add101b02c6099a07056d5801e`

PR #7101, opened 2026-05-15
chore(qwp): fix geohash precision loss and sentinel null handling on ingest

kafka1991:
## Summary

- Auto-creating a GEOHASH column from a QWP ingest batch now carries the correct bit precision. Previously the column type was inferred without per-column precision metadata from the wire cursor, producing an incorrect type.
- Fixed-width columns (INT, LONG, DATE, TIMESTAMP, TIMESTAMP_NS, FLOAT, DOUBLE) carrying sentinel NULL values (`INT_NULL`, `LONG_NULL`, `NaN`) without a null bitmap are now recognised as NULL. Without the fix, row-by-row conversion paths could throw on out-of-range coercions (e.g. INT to BYTE with `INT_NULL`) or write the raw sentinel value into storage instead of NULL.

## Test plan

- QwpFixedWidthDecoderTest covers sentinel-NULL detection across all supported types, plus the inverse case (bitmap-marked non-null preserves an explicit sentinel value).
- QwpUdpAllTypesTest covers GEOHASH auto-create end-to-end across all four storage widths (GEOBYTE, GEOSHORT, GEOINT, GEOLONG).

### DIFF
--- a/core/src/main/java/io/questdb/cutlass/line/tcp/QwpWalAppender.java
+++ b/core/src/main/java/io/questdb/cutlass/line/tcp/QwpWalAppender.java
@@ -519,93 +519,100 @@ public class QwpWalAppender implements QuietCloseable {
                          ColumnType.DATE,
                          ColumnType.UUID, ColumnType.LONG256,
                          ColumnType.GEOBYTE, ColumnType.GEOSHORT, ColumnType.GEOINT, ColumnType.GEOLONG -> {
-                        if (cursor instanceof QwpGeoHashColumnCursor geoHashCursor) {
-                            if (ColumnType.isGeoHash(columnType)) {
-                                int wireBits = geoHashCursor.getPrecision();
-                                if (wireBits != ColumnType.getGeoHashBits(columnType)) {
-                                    throw geoHashPrecisionMismatchException(columnType, wireBits, tableBlock, col);
+                        switch (cursor) {
+                            case QwpGeoHashColumnCursor geoHashCursor -> {
+                                if (ColumnType.isGeoHash(columnType)) {
+                                    int wireBits = geoHashCursor.getPrecision();
+                                    if (wireBits != ColumnType.getGeoHashBits(columnType)) {
+                                        throw geoHashPrecisionMismatchException(columnType, wireBits, tableBlock, col);
+                                    }
+                                }
+                                appender.putGeoHashColumn(columnIndex, geoHashCursor, rowCount, columnType);
+                            }
+                            case QwpFixedWidthColumnCursor fixedCursor -> {
+                                if (!isFixedTypeCoercionAllowed(qwpType, columnType)) {
+                                    throw coercionNotSupportedException(qwpType, columnType, tableBlock, col);
+                                }
+                                int wireSize = fixedCursor.getValueSize();
+                                int columnSize = ColumnType.sizeOf(columnType);
+                                int colTag = ColumnType.tagOf(columnType);
+                                boolean isIntegerWire = isIntegerWireType(qwpType);
+                                boolean isFloatWire = isFloatWireType(qwpType);
+                                boolean isFloatTarget = colTag == ColumnType.FLOAT || colTag == ColumnType.DOUBLE;
+                                boolean isIntegerTarget = colTag == ColumnType.BYTE || colTag == ColumnType.SHORT
+                                        || colTag == ColumnType.INT || colTag == ColumnType.LONG;
+                                if (isIntegerWire && (isFloatTarget || wireSize != columnSize)) {
+                                    // Integer → float/double, integer widening (INT → LONG), or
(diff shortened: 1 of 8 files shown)


jerrinot:
codex review:

- [P1] No-bitmap GEOHASH null sentinels are still decoded as real values. The ingress spec says null_flag = 0 uses in-place sentinels, and GEOHASH null is all-ones truncated to the geohash byte width: docs/qwp/wire-ingress.md:355, docs/qwp/wire-ingress.md:446, docs/qwp/wire-ingress.md:656. But core/src/
  main/java/io/questdb/cutlass/qwp/protocol/QwpGeoHashColumnCursor.java:60 only honors bitmap nulls; in no-bitmap mode it sets currentIsNull = false, reads the packed value, and returns non-null. The Java buffer can emit exactly this sentinel with TYPE_GEOHASH + useNullBitmap=false: java-questdb-client/
  core/src/main/java/io/questdb/client/cutlass/qwp/protocol/QwpTableBuffer.java:1040, packed by java-questdb-client/core/src/main/java/io/questdb/client/cutlass/qwp/client/QwpColumnWriter.java:194. The WAL path then writes null only if cursor.isNull(): core/src/main/java/io/questdb/cairo/wal/
  WalColumnarRowAppender.java:1011. Result: sentinel geohash nulls become stored geohash values.

- [P1] Existing GEOHASH columns do not validate incoming wire precision, so later batches can silently truncate or reinterpret values. This branch creates new geohash columns with the first block’s precision: core/src/main/java/io/questdb/cutlass/line/tcp/QwpWalAppender.java:290. But appending to an
  existing geohash column accepts any QwpGeoHashColumnCursor and passes it straight to the WAL appender: core/src/main/java/io/questdb/cutlass/line/tcp/QwpWalAppender.java:509. The WAL appender only casts to the storage width, with no comparison to ColumnType.getGeoHashBits(columnType): core/src/main/java/
  io/questdb/cairo/wal/WalColumnarRowAppender.java:1007, core/src/main/java/io/questdb/cairo/ColumnType.java:355. Since the client buffer only enforces same precision within the current buffer and resets precision after flush: java-questdb-client/core/src/main/java/io/questdb/client/cutlass/qwp/protocol/
  QwpTableBuffer.java:888, java-questdb-client/core/src/main/java/io/questdb/client/cutlass/qwp/protocol/QwpTableBuffer.java:1419, the same sender can create a 5-bit column in one batch and send 20-bit values in the next batch. That should be rejected or explicitly converted, not silently cast.

bluestreak01 (dismissed):
LGTM. Both fixes look correct end-to-end:

- `QwpFixedWidthColumnCursor.advanceRow` / `QwpGeoHashColumnCursor.advanceRow` now recognize the sentinel-mode null encoding, so downstream coercion paths (e.g. INT→BYTE with `INT_NULL`) no longer throw and `putGeoHashColumn` no longer persists a junk value.
- `appendToWalColumnar` validates wire-vs-column precision for existing GEOHASH columns; mismatches now surface a clear `GeoHash precision mismatch` error instead of silently truncating.
- `mapQwpTypeToQuestDB(int, cursor, colIndex)` carries per-column metadata into auto-created column types, so the auto-created GEOHASH width matches the wire precision.

Verified that:
- bitmap-mode plus an explicit `INT_NULL` payload still round-trips as a real value (covered by `testDecodeIntColumnExplicitSentinelWithBitmap`);
- the 60-bit GEOLONG round-trip is unaffected (sentinel is `0xFFFFFFFFFFFFFFFF`, max legitimate value is `0x0FFFFFFFFFFFFFFF`);
- UDP path safely drops mismatched batches via the existing `Throwable` handler in `QwpUdpReceiver`.

Minor nits, non-blocking:
- PR title is `chore(qwp):` but this is a bug fix (label = Bug). Per CLAUDE.md the prefix should be `fix(qwp):` so it lands in release notes.
- Consider adding the `ILP` label alongside `Bug`.
- `QwpWalAppenderTest.java:208` still asserts `mapQwpTypeToQuestDB(TYPE_GEOHASH) == GEOLONG` against the no-cursor overload; that fallback is effectively dead on the production path now. Either drop the fallback or note in Javadoc that the no-cursor variant must not be used for `TYPE_GEOHASH`.

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 181 / 183 (98.91%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/QwpWalAppender.java|169|171|98.83%|
|:large_blue_circle:|io/questdb/cutlass/qwp/protocol/QwpFixedWidthColumnCursor.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/protocol/QwpGeoHashColumnCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpTudCache.java|1|1|100.00%|

PR #7104, opened 2026-05-15
feat(sql): add regr_r2 coefficient of determination aggregate

jovfer:
## Summary

Adds `regr_r2(y, x)`, the standard SQL coefficient-of-determination aggregate. Reports how well a linear regression of Y on X fits the data, on a 0–1 scale. Useful for separating real time-series trends from noise — for example, finding robots whose motor temperature is trending up with high confidence versus those drifting randomly.

The change is shaped to also serve as the foundation for the remaining standard regression aggregates, which can be added later as small subclasses without touching the shared math.

Addresses gap 3 of [questdb/roadmap#120](https://github.com/questdb/roadmap/issues/120). Closes one of seven missing standard regression aggregates; the other six are out of scope here.

Documentation: [questdb/documentation#449](https://github.com/questdb/documentation/pull/449).

## What changed

- New `AbstractRegressionGroupByFunction` — a 6-slot Welford state shared by all regression aggregates: `meanY, Syy, meanX, Sxx, Sxy, count`. Provides the per-row update and Chan parallel merge in one place.
- `regr_slope` and `regr_intercept` migrated onto the shared base. Their inner function classes shrink from ~220 to ~25 lines each; they now override only `getDouble()` and `getName()`.
- New `regr_r2(y, x)` implementing SQL:2003 §10.9 semantics:
  - `Sxx = 0` returns `NULL` (covers `count = 1` and constant-X cases).
  - `Syy = 0` with `Sxx ≠ 0` returns `1.0` (a horizontal line is a perfect fit when Y is constant).
  - Otherwise returns `Sxy² / (Sxx · Syy)`.

## Tradeoffs and limitations

- **Slope and intercept gain an unused state slot** (8 bytes per group key for the `Syy` value they don't read). The slot is still written on every row, but the work is one extra subtract+multiply+add in the hot loop. Per-row overhead is negligible compared to the existing Welford update; per-group memory overhead is negligible compared to the rest of the group-by map machinery.
- **Welford reordering is bit-identical for the shared slots**, but only by inspection: `Sxy = (y - oldMeanY) * (x - newMeanX)` and `Sxx = (x - newMeanX) * (x - oldMeanX)` produce the same IEEE-754 result regardless of whether `Syy` is computed between the two `meanY` and `meanX` updates. All existing `regr_slope` and `regr_intercept` tests pass without modification, which validates the trace empirically.
- **Semantics diverge from `corr()` at one edge**: when Y is constant and X varies (`Syy = 0`, `Sxx ≠ 0`), `regr_r2` returns `1.0` per SQL:2003, while `corr()` returns `NULL`. Users who expect `regr_r2 = corr²` everywhere will see this one boundary disagree. Documented in code.
- **Does not migrate `corr()` onto the new base**, even though its state layout is identical. The migration is mechanical but touches a different test surface; deferred. Note that the empty-partial merge guards introduced here fix a latent parallel-GROUP-BY bug that `corr()` still has — tracked separately as #7160.

## Test plan

- Run the new `RegressionR2FunctionFactoryTest` (15 cases): all-null inputs, no rows, single row (`Sxx = 0 → NULL`), perfect line on DOUBLE/FLOAT/INT (`r² = 1.0`), constant-X (`NULL` per spec), constant-Y (`1.0` per spec), all-same values (`NULL`, `Sxx` precedence), large repeated values (no overflow), some-null pairs mixed with real data, a hand-computed reference `r² = 0.6` for `X={1,2,3,4,5}`, `Y={2,4,5,4,5}`, equality with `corr(y,x)²` on a non-degenerate dataset within `1e-12`, and a multi-group `GROUP BY` to exercise the parallel merge path.
- Run the existing `RegressionSlopeFunctionFactoryTest` and `RegressionInterceptFunctionFactoryTest` unmodified to confirm the bit-identical refactor.
- Run the adjacent `CorrGroupByFunctionFactoryTest`, `CovarSampleGroupByFunctionFactoryTest`, `CovarPopGroupByFunctionFactoryTest` to confirm no unintended fallout.
- Cross-check the `r² = 0.6` reference value against PostgreSQL for the same dataset.

## Follow-ups

This PR ships one of seven SQL-standard regression aggregates, and only in the GROUP BY context. Three natural extensions are out of scope here:

### Apply the same merge-guard fix to `corr()` (#7160)

The empty-partial guards in `AbstractRegressionGroupByFunction.merge()` close a latent NaN-poisoning bug in the parallel-merge path: when several worker partials see only non-finite `(y, x)` pairs for a key, the unguarded Chan formula computes `0 / 0 = NaN` and the resulting NaN then poisons subsequent merges with real-data partials, producing `NULL` instead of the correct result. `corr()` uses an identical 6-slot layout and still has the unguarded merge, so the same bug remains reachable there under sparse-NULL parallel GROUP BY. The fix is mechanical (either migrate `corr()` onto the new base — only `getDouble()` and `getName()` change — or duplicate the two guards into the existing `merge()`); the work is deferred to keep this PR's blast radius narrow and is tracked as #7160. A natural bundle for it is the six-aggregate follow-up below.

### Remaining regression aggregates (GROUP BY)

The SQL standard defines six more regression aggregates that QuestDB still lacks: `regr_count`, `regr_avgx`, `regr_avgy`, `regr_sxx`, `regr_sxy`, `regr_syy`. With the new `AbstractRegressionGroupByFunction` base in place, each becomes a roughly 30-line subclass that overrides only `getDouble()` and `getName()` — no new state, no new merge logic. These map directly to fields the base already tracks (`count`, `meanX`, `meanY`, `Sxx`, `Sxy`, `Syy`). A single follow-up PR can add all six.

### Window-function variants for regression

None of the regression aggregates currently work inside `OVER (...)`. `regr_slope OVER (...)`, `regr_intercept OVER (...)`, `regr_r2 OVER (...)` will not compile today. This is the same gap that `stddev`/`variance`/`covar_samp`/`covar_pop` had before the recent window additions, and it matters for time-series workloads — rolling regression (financial beta, sensor degradation slope, drift detection) is a common pattern that today requires self-joins with time-bucket GROUP BY.

Adding window variants is structurally separate work. QuestDB keeps GBF and window aggregates as independent class trees by design: GBF uses Welford's online algorithm (compact state, not removable on sliding frames), while window uses naive sum-of-products (decrementable, less stable). The window-side abstract class `AbstractBivariateStatWindowFunctionFactory` already tracks the full state regression needs (`sumX, sumXX, sumY, sumYY, sumXY, count`) for `corr`/`covar_samp`/`covar_pop`, parametrized by two booleans (`isCorrelation`, `isSample`).

Extending the window family to cover regression would require:
- Refactoring the boolean parametrization in `AbstractBivariateStatWindowFunctionFactory` into a strategy or enum so it can dispatch to slope/intercept/r2 projections in addition to corr/covar.
- Adding `computeSlope`, `computeIntercept`, `computeR2` static helpers alongside the existing `computeCorr`/`computeCovar` for both Welford (unbounded frames) and naive (sliding frames) code paths.
- Adding small shim factories per function (~48 lines each, modeled on `CorrDoubleWindowFunctionFactory`).
- Tests covering partition framing, RANGE and ROWS frames, and sliding-frame edge cases.


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 135 / 138 (97.83%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/AbstractRegressionGroupByFunction.java|97|100|97.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/RegressionInterceptFunctionFactory.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/RegressionSlopeFunctionFactory.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/RegressionR2FunctionFactory.java|18|18|100.00%|

bluestreak01 (approved):

Issue #1016, opened 2021-05-13
alter table zeros out hysteresis values


bluestreak01:
```sql
CREATE TABLE my_table (timestamp TIMESTAMP, x long, s symbol) timestamp(timestamp)
PARTITION BY DAY WITH o3MaxUncommittedRows=250000, o3CommitHysteresis=240s
```

then

```sql
ALTER TABLE my_table ALTER COLUMN s ADD INDEX
```

then

```sql
tables()
```

shows 0 hysteresis

bluestreak01:
fixed by #1014

PR #7107, opened 2026-05-15
feat(core): support parquet format table level

ideoma:
## Summary

Adds table-level Parquet partition format. Tables can now declare Parquet as the default storage format for new partitions, either at creation time or via `ALTER TABLE`. Native remains the default.

Until now, Parquet partitions could only be produced after the fact, partition by partition, via `ALTER TABLE ... CONVERT PARTITION ... TO PARQUET`. With this change, a table can be marked `FORMAT PARQUET` and every brand-new partition is written directly as Parquet, with no intermediate native partition.

Tandem https://github.com/questdb/questdb-enterprise/pull/1016

## New SQL

`CREATE TABLE` accepts an optional `FORMAT { PARQUET | NATIVE }` clause inside the partitioning section:

```sql
CREATE TABLE trades (ts TIMESTAMP, price DOUBLE, sym SYMBOL)
TIMESTAMP(ts) PARTITION BY DAY FORMAT PARQUET WAL;
```

`ALTER TABLE` switches the default format for new partitions of an existing table:

```sql
ALTER TABLE trades SET FORMAT PARQUET;
ALTER TABLE trades SET FORMAT NATIVE;
```

`SHOW CREATE TABLE` echoes `FORMAT PARQUET` back when the table was created or altered to use it. `FORMAT NATIVE` is the default and is omitted.

## Behaviour and implications

- `FORMAT PARQUET` only applies to **partitioned WAL tables**. Non-WAL tables, non-partitioned tables, and materialized views reject the clause with a clear error at compile time.
- Switching format with `ALTER TABLE ... SET FORMAT` is a metadata-only change. **Existing partitions are not rewritten** — they keep whatever format they were written in. The setting only governs partitions created after the ALTER.
- On a `FORMAT PARQUET` table, every new partition that gets written by the WAL apply path lands directly as a Parquet file. There is no native intermediate, no later conversion step, and `table_partitions()` reports `isParquet = true` from inception.
- Backdated inserts that land into an existing native partition continue to write into that native partition. Backdated inserts that create a brand-new partition on a `FORMAT PARQUET` table create that partition as Parquet.
- Parquet partitions are immutable, so writes to a `FORMAT PARQUET` table always go through the O3 rewrite path, even when the data is strictly in order. Workloads that previously benefited from the in-order append fast path will see additional write amplification on Parquet-format tables — pick the format based on workload.
- `ALTER TABLE ... SET FORMAT` requires write permission on the table. Enterprise builds gate it behind a dedicated permission.

## Compatibility

- Existing tables and `CREATE TABLE` statements without a `FORMAT` clause are unaffected and continue to default to native partitions.
- The `_meta` table-metadata file bumps to minor version 2 to record the default partition format. Older databases open without rewrites and read as `FORMAT NATIVE`.


questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=233797)
**Tested Commit:** `08b265fa34e42fea4db182882d2428920cc03235`

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 244 / 257 (94.94%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/wal/seq/MetadataServiceStub.java|0|2|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/AlterOperation.java|7|10|70.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|22|25|88.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|23|24|95.83%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|130|134|97.01%|
|:large_blue_circle:|io/questdb/cairo/TableReaderMetadata.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CreateTableOperationBuilderImpl.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlParser.java|13|13|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriterMetadata.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoTable.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowCreateTableRecordCursorFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/AlterOperationBuilder.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableUtils.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CreateTableOperationImpl.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableStructure.java|1|1|100.00%|

ideoma:
Reviewed at level 2

  Critical

  None.

  Moderate

  1. Enterprise CI is RED — investigate before merge, don't wave through. (process / out-of-diff)
  questdb-butler reports the enterprise suite failed (Azure buildId 237050, commit 930bd8cb). This PR is tandem with questdb-enterprise#1016, which wires the new authorizeAlterTableSetFormat permission. Because SecurityContext.authorizeAlterTableSetFormat is a non-default abstract method (SecurityContext.java:90),
  every enterprise implementor (EntSecurityContextBase, AdminSecurityContext, AbstractReplicaSecurityContext, replica deniedOnReplica, test mocks) plus Permission.java registration must be updated or enterprise won't compile. Per CLAUDE.md, a red CI is a hypothesis to disprove — confirm the failure is the expected
  tandem interlock and not a real regression before merging.

  2. writeFreshParquetFromO3 omits the partition-directory fsync the analogous CONVERT path performs. (in-diff durability gap)
  writeFreshParquetFromO3 fsyncs _pm (O3PartitionJob.java:3638-3639) but never fsyncs the brand-new partition directory. The established convertPartitionNativeToParquet path does both — it fsyncs _pm and the partition directory on non-Windows (TableUtils.java:1985-1995) with the explicit rationale that _pm is "a
  brand-new file in a brand-new partition directory." The fresh path creates a brand-new directory in exactly the same way. On a power loss right after committing _txn (non-NOSYNC commit mode), the directory entry for the new _pm/parquet file may not be durable even though the _pm contents were fsynced — _txn could
  reference a partition whose directory entry is lost. Mirror the CONVERT path's dir-fsync (drop to the parent and fsyncAndClose the dir fd on non-Windows).

  3. No test for the _meta v1→v2 migration read path. (in-diff test gap)
  TableUtils.getTableFormat() returns TABLE_FORMAT_NATIVE only when isMetaFormatAtLeast(..., META_FORMAT_MINOR_VERSION_TABLE_FORMAT) is false (TableUtils.java:2784-2788). Every TableFormatTest table is created fresh at minor-version 2, so the false branch — the one backward-compat path that matters — has zero
  coverage. EngineMigrationTest never touches getTableFormat. The failure mode (an un-migrated v1 _meta mis-reading bytes at META_OFFSET_TABLE_FORMAT) is exactly what a fixture catches. Add a test that writes/loads a minor-version-1 _meta and asserts getTableFormat() == TABLE_FORMAT_NATIVE (ideally a zip fixture in
  EngineMigrationTest). Note: the PR also correctly froze the TTL and parquet-encoding gates to explicit version constants (META_FORMAT_MINOR_VERSION_TTL=1, ..._PARQUET_ENCODING_CONFIG=1) rather than isMetaFormatUpToDate/LATEST — that pinning is a necessary correctness fix (otherwise the bump to LATEST=2 would have
  silently zeroed TTL on v1 tables), and forward-compat is safe because the checksum hashes only metadataVersion+columnCount, not the minor version.

  4. Rust JNI-safety error arms are largely untested. (in-diff test gap)
  A panic across JNI aborts the JVM, so these guard arms must be covered, not assumed unreachable (per the Rust guidelines). Uncovered:
  - The entire Delta strided encoder has no Rust unit test — neither happy-path nor error arms for encode_designated_timestamp_strided / write_delta_i64_page in delta_binary_packed.rs (only end-to-end via testFreshParquetTimestampDeltaBinaryPackedEncoding). This is the weakest spot.
  - TimestampValues::Contiguous(_) => Err(InvalidLayout) in all three encoders.
  - repetition != Required => Err guards in the Plain/Delta page writers.
  - slice.is_empty() => Ok(vec![]) in the rle strided path.
  These are cheap to test directly (contiguous-flagged column → Err, Optional PrimitiveType → Err, empty &[[i64;2]] → Ok(empty)). The Plain strided happy path is well covered by encode_designated_timestamp_strided_matches_contiguous (byte-identical vs contiguous).

  Minor

  5. Latent footgun — multi-partition strided not guarded. file.rs:462 applies strided_row_group_slice only when columns.len() == 1, but the encoder dispatch (encode.rs) gates only on strided_timestamp_16. Multi-partition strided is unreachable today (the strided flag is set only on the single-partition
  encodePartition path; the streaming/export multi-partition path never sets it), so it's safe — but a future multi-partition strided caller would silently re-emit the full merge index per row group. Add debug_assert!(!columns[0].strided_timestamp_16) to the multi-partition branch, or make it a hard error.

  6. Latent footgun — after-DEDUP FORMAT block doesn't set formatSeen. SqlParser.java sets formatSeen in the before-WAL and after-WAL blocks but not the after-DEDUP block (~line 2316). Harmless today (it's the last FORMAT check in a linear parse), but it'll silently allow a duplicate if a fourth FORMAT position is
  ever added after it. Set the flag for symmetry.

  7. Redundant per-column metadata lookups. populateO3DescriptorColumns (O3PartitionJob.java:1668-1675) calls metadata.getColumnMetadata(i) up to four times per column (type, name, writerIndex, encodingConfig). Hoist to one TableColumnMetadata m = metadata.getColumnMetadata(i) local. It's getQuick(i) (bounds-free
  array read), so this is micro — but the duplicate calls on consecutive lines 1674/1675 are an obvious hoist. Same pattern at the two sibling loops.

  8. Deterministic column-type coverage on the fresh path. TableFormatTest directly covers only timestamp/long/string/varchar/binary/symbol. DOUBLE[]/LONG256/UUID/GEOHASH/IPv4/BOOLEAN/indexed-SYMBOL and all-NULL columns are exercised only stochastically via WalWriterFuzzTest (createWalAsParquet + the full type
  matrix). Given how much rides on the direct-slice assumption, one deterministic multi-type fresh-parquet round-trip test would harden it.

  9. Test robustness. testShowCreateTableEmitsFormatParquet asserts the emitted DDL string but never re-executes it. (Verified it would re-parse fine — the emitted DDL omits WAL, but partitioned tables default to WAL, so FORMAT PARQUET is accepted on replay. Still worth a round-trip assertion.)

  10. Release-mode hardening (optional). Column::from_raw_data / timestamp_values rely on a debug_assert in transmute_slice for the strided length-multiple and alignment. Safe given the single Java caller passes rowCount * 16 from malloc'd memory, but a release build would silently truncate a malformed length rather
  than error. A release-mode length check for the strided case would make the JNI contract self-defending.

  Summary

  Verdict: approve pending (a) resolution of the red Enterprise CI and (b) the moderate test/durability items. No critical issues.

ideoma:
# Code Review: PR #7107 — feat(core): support parquet format table level

**Level-3 review** (9-agent fan-out + per-finding source verification), consolidated and re-evaluated after discussion. This single comment supersedes my earlier follow-ups (C1/M2/M4/M5/M1), folded in below with their final dispositions.

**Bottom line (re-evaluated): Approve.** No Critical or Moderate issues remain. The correctness-critical machinery was traced to source and verified sound:
- **Fresh-parquet O3 write** — the dropped `mergeCopy` is a verified-redundant identity copy (`flatten_index` rewrites the merge-index rowId to sequential; `cthO3SortColumn`/`swapO3ColumnsExcept` physically sort non-timestamp columns before this path; the timestamp uses the strided index). No data-corruption path for out-of-order fresh partitions.
- **Strided-timestamp JNI encoding**, **`_meta` v1→v2 minor-version handling**, **coalesce-boundary-tie dedup** (+ `decode_row_group_range`), and **resource ownership** (non-owning `freshPartitionDescriptor`, fd lifecycle, latch accounting) — all verified correct.

Six originally-raised findings were withdrawn/downgraded after the author's pushback (consistent root cause: I had measured against edge configs / outlier siblings rather than QuestDB's dominant conventions); M3 was fixed by the author.

---

## Previously-raised findings — final disposition

**C1 (was Critical) — `writeFreshParquetFromO3` partition-dir fsync → DOWNGRADED to a minor consistency note.** The fresh path fsyncs only `_pm`, not the partition directory — but that is the product-wide norm: native O3 commits (`O3CopyJob.syncColumns`) fsync only column file FDs + mapped pages, and the existing `processParquetPartition` rewrite syncs only `_pm` (`update.rs:1062-1067`). `convertPartitionNativeToParquet` (`TableUtils.java:1991`) is the lone outlier that also fsyncs the dir (added speculatively with the `_pm` sidecar in #6913). `writeFreshParquetFromO3` is consistent with the dominant convention; the underlying durability gap is real but product-wide, and closing it would be a broad change unrelated to this PR.

**M1 (was Moderate) — SHOW CREATE TABLE omits `WAL` → WITHDRAWN.** SHOW CREATE TABLE intentionally omits `WAL` for WAL tables and relies on `cairo.wal.enabled.default` (true in production), so the emitted DDL round-trips on any normal deployment. The same long-standing convention applies to `DEDUP` (also WAL-only, also emitted without `WAL` — `ShowCreateTableTest.testDedup`). Only legacy `wal.enabled.default=false` instances see a replay failure; not worth correcting.

**M2 (was Moderate) — removal of #7146 captured-seed reproducers → RETRACTED.** This PR changes `FuzzTransactionGenerator.generateSet`'s random-draw sequence (an unconditional new `rnd.nextDouble()` for the SET FORMAT decision, plus two added terms in the probability denominator). A fixed seed therefore generates a different transaction sequence than the one captured, so the reproducers no longer exercise their target bugs — keeping them would be false coverage. Removal is correct and inevitable; the randomized fuzz (with native oracle) preserves coverage of those index paths.

**M3 (was Moderate) — committed `/tmp/qdb_dump` debug tests → FIXED** (commit `710ac447ec`, "remove dump debug only tests"). The two no-assertion `testDump...ForInspection` methods and the `/tmp/qdb_dump` writes are gone. Minor leftover: the now-unused private helper `dumpParquetPartition` (`ParquetWriteTest.java:3488`) is orphaned — worth deleting for cleanliness.

**M4 (was Moderate) — missing `_meta` v1→v2 migration-read test → RETRACTED.** A minor-version bump is not a migration: no `Mig*` step, no `_meta` rewrite; old files are read transparently via the checksum-gated `isMetaFormatAtLeast`, and `EngineMigrationTest` is `@Ignore`d. The gate is already tested (`TableMetadataTest`, `ShowCreateTableTest`); `getTableFormat` is a one-liner over it with the field in the always-mapped 128-byte header. (The TTL / parquet-encoding version-constant pinning in the PR remains a genuine correctness fix.)

**M5 (was Moderate) — missing deterministic indexed-SYMBOL fresh-parquet test → RETRACTED.** Covered by the fuzz: `FuzzRunner.createInitialTable` creates `index(sym2 ...)` + `FORMAT PARQUET WAL` under `createWalAsParquet` (`FuzzRunner.java:988/995`), so every fresh partition runs `writeFreshParquetFromO3` → `updateParquetIndexes`; `assertRandomIndexes` queries the indexed column against the native oracle, and a missing-index SIGABRT would crash the run. Broader than a single deterministic case.

---

## Minor / optional (non-blocking)

- **m1** — Harden `assert decoded == rowGroupSize` (mergeRowGroup range path) into a thrown `CairoException`; assertions are disabled in release and a mismatch reads past the decoded buffer on a data-critical path.
- **m2** — Derive the coalesce gate (`hasCoalescableTie`) and the `coalesceBoundaryTies` flag from one computed boolean; today they cannot diverge dangerously only because the timestamp is always present in Parquet (asserted), which is fragile.
- **m3** — Eliminate the redundant double-scan of row-group timestamp stats on dedup commits (`hasCoalescableBoundaryTie` re-derives what the `rowGroupBounds` build + `computeMergeActions` already compute).
- **m4** — Set `formatSeen = true` in the after-DEDUP FORMAT block (`SqlParser`) for symmetry; harmless today (last FORMAT position).
- **m5** — Add Rust tests for `decode_row_group_range` and the Delta/RLE strided encoders' error arms (`Contiguous => Err`, non-`Required` repetition, empty-slice), per the repo's "cover or prove unreachable" rule.
- **m6** — Hoist `metadata.getColumnMetadata(i)` in `populateO3DescriptorColumns` (called up to 4x/column); negligible but a clean hoist.
- **m7** — Guard or comment `decode_row_group_range` against VarcharSlice column types (latent landmine; safe today because the merge passes materialized types only).
- **m8** — `ALTER ... SET FORMAT NATIVE` on an ineligible table forces a no-op `commit()` + metadata bump; consider an early return when the value is unchanged.
- **m9** — `CREATE TABLE (LIKE src)` inherits FORMAT PARQUET (copied in `updateFromLikeTableMetadata`) but is untested.
- **m10** — No test for the encode-failure / `updateParquetIndexes`-failure cleanup path (the existing failure tests inject at `_pm` open, before the encoder, so the close-`_pm`-before-rmdir branch is never exercised).
- **m11** — Rename the misleading local `isMetaFormatUpToDate` (`MetadataCache.java:218`), now meaning "at least PARQUET_ENCODING_CONFIG".
- **m12** — Optional Rust hardening: add `columns.len() == 1 &&` to the three strided dispatch guards in `encode.rs` so a future multi-partition-strided caller fails loudly instead of silently dropping partitions.
- **m13 (new)** — Remove the orphaned `dumpParquetPartition` helper (`ParquetWriteTest.java:3488`) left behind by the M3 fix.

---

## Verified correct (not issues)

- **Dropped `mergeCopy`** — redundant identity copy after `flatten_index`; new direct-buffer handoff is equivalent across WAL ordered/unordered and block-apply paths.
- **Negative `partitionIndexRaw`** in `o3ConsumePartitionUpdateSink` — `updateAttachedPartitionSizeByRawIndex` handles `index < 0` and genuinely inserts.
- **parquet2 `file_offset` change** — spec-correct fix; no reader regression: `ColumnChunkMetaData::file_offset()` has zero callers, readers use `byte_range()` via `dictionary_page_offset`/`data_page_offset`. Affects all parquet writers but is reader-safe.
- **No resource leak / double-free / latch hang** on the fresh-parquet path.

---

## Verdict

**Approve.** No Critical or Moderate issues; the correctness-critical paths are verified sound. Only minor/optional polish remains (incl. the orphaned `dumpParquetPartition` helper).

Process: the enterprise tandem (#1016 — `authorizeAlterTableSetFormat`, `Permission.SET_TABLE_FORMAT`, replica `deniedOnReplica`) must land together; the red enterprise CI is the expected interlock (OSS compiles; tandem present in the working tree) — confirm before merge per CLAUDE.md.

<sub>🤖 Consolidated level-3 review via [Claude Code](https://claude.com/claude-code)</sub>

ideoma:
# Code Review: PR #7107 — feat(core): support parquet format table level

**Level-3 delta review.** This PR already carries a consolidated level-3 review that reached **Approve**, with C1/M1/M2/M4/M5 withdrawn or downgraded after discussion and M3 fixed. Rather than re-litigate those, I ran a fresh 5-agent adversarial fan-out (one fully unanchored, no checklist) briefed to skip the settled findings and hunt only for *new* issues. All five — including the unanchored agent — independently re-derived the prior conclusions and found **no new correctness bug, leak, panic, or corruption.** Below is only what's new or still-open.

## Critical

None. The correctness-critical machinery (fresh-parquet O3 write, strided-timestamp JNI encoding, parquet2 `column_chunk` offset rewrite, multi-group `decode_row_group_range`, dedup boundary-tie coalescing, fd/latch/descriptor ownership) was re-traced to source and confirmed sound. The parquet2 `file_offset` change is a genuine *fix* (old code emitted `dictionary_page_offset = None` and overshot `file_offset` by the dict-page size); `ColumnChunkMetaData::file_offset()` has zero callers, so no reader regresses.

## Moderate

None.

## Minor

**1. (new) Latent footgun: strided-timestamp identification relies on the `writerIndex == positional-index` invariant.** `populateO3DescriptorColumns` (`O3PartitionJob.java:2611/2639`) sets each column's `col_id` to `getWriterIndex()`, while the descriptor's `timestamp_index` is `metadata.getTimestampIndex()` (positional, `:1705`). On the Rust side, `designated_timestamp = (col_id == timestamp_index)` (`jni.rs:825`), and the strided encoder is selected only when `is_designated_timestamp && strided_timestamp_16` (`encode.rs:461`). If `col_id != timestamp_index` for the timestamp column, `designated_timestamp` goes false, the strided branch is skipped, and the 16-byte `(ts,rowId)` merge index gets decoded as a contiguous `i64` array → silent timestamp corruption. This is **safe today** because `TableWriterMetadata` never compacts on DROP (tombstones kept in place, `writerIndex == i`, verified at `TableWriterMetadata.java:185`), and `populateO3DescriptorColumns` is only ever called with `TableWriterMetadata`. But the coupling is implicit and load-bearing. Suggest a one-line `assert metadata.getColumnMetadata(timestampIndex).getWriterIndex() == timestampIndex;` in `populateO3DescriptorColumns`, or a `debug_assert!` on the Rust side that a `strided_timestamp_16` column is always `designated_timestamp`, so a future reuse with compacted metadata fails loudly instead of corrupting data.

**2. (new) `delta_binary_packed` and `rle_dictionary` strided encoders have no direct Rust unit test.** Only the Plain encoder has the byte-equivalence test (`encode_designated_timestamp_strided_matches_contiguous`, `plain/tests.rs:159`). The two other new strided functions are exercised only end-to-end via the Java round-trips (`testFreshParquetTimestampDeltaBinaryPackedEncoding` / `...RleDictionaryEncoding`). This is distinct from the already-noted error-arm gap (m5): the *happy path* of these two encoders has no unit test mirroring Plain's. Low severity (Java round-trip covers correctness), but the `matches_contiguous` pattern should be extended to all three.

**3. (new, optional) `writeFreshParquetFromO3` trusts `partitionRowCount = srcOooHi - srcOooLo + 1 >= 1` with no guard.** Consistent with the sibling `processParquetPartition` invariant (the O3 dispatcher never creates empty-partition tasks), so safe today — but a future dispatch-contract change would silently feed a 0/negative row count to the encoder. A cheap `assert partitionRowCount > 0` would defend it.

**4. (still open) m4 — after-DEDUP `FORMAT` block doesn't set `formatSeen`** (`SqlParser.java:1871`). Confirmed still present. Harmless (it's the last FORMAT position in the linear parse), but set the flag for symmetry so a future fourth position can't silently accept a duplicate.

**5. (still open) m13 — orphaned `dumpParquetPartition` helper** (`ParquetWriteTest.java:3488`). Confirmed still present after the M3 cleanup commit. Unused private method; delete it.

## Downgraded (re-confirmed against source, not new findings)

These were verified *again* this pass and remain correct / non-issues — listed so the prior dispositions are visible:

- **C1 — fresh path doesn't fsync the partition dir** → product-wide norm (native O3 commit also only fsyncs file FDs; CONVERT is the lone outlier). Not introduced here.
- **Dropped `mergeCopy`** → verified-redundant identity copy after `flatten_index`; direct-buffer handoff is equivalent across var-size, symbol, and NULL cases.
- **parquet2 `file_offset` change** → spec-correct fix, reader-safe (no `file_offset()` callers; `RowGroup.file_offset` derives from dict/data page offsets).
- **Negative `partitionIndexRaw`** → correctly inserts via `updateAttachedPartitionSizeByRawIndex`.
- **Dedup coalesce gate vs coalesce loop** → byte-identical boundary detection keeps them in lockstep; inter-group gaps are provably empty.
- **`VarcharSlice` cross-group dangling pointer in `decode_row_group_range`** → unreachable; the merge path only passes persisted storage types, never the transient `VARCHAR_SLICE`. The `refresh_ptrs` change is a real correctness fix for reallocation across appended groups.
- **M1/M2/M4/M5** (SHOW CREATE WAL omission, fuzz-seed removal, "no migration test", indexed-SYMBOL fuzz coverage) → all re-confirmed as correctly settled.
- One **pre-existing** item (not this PR): `SlicePageReader::from_parts` lacks a `col_start + col_len` overflow guard (added in #6913, inputs come from trusted `_pm`). Flagged only for awareness.

## Summary

**Verdict: Approve** — consistent with the existing review. This delta pass surfaced **3 new minor/latent items** (the strided-timestamp identification footgun is the only one with a data-corruption failure mode, and it is unreachable under the current `writerIndex == position` invariant) plus **2 still-open known minors** (m4, m13). No Critical or Moderate issues; nothing here blocks merge.

- 5 review agents run (4 briefed on settled findings + 1 unanchored fresh-context); **0 new confirmed bugs**, 3 new latent/minor items, 2 known-open minors re-confirmed present.
- All new items are **in-diff** except the latent footgun, whose corruption surface only opens at an out-of-diff reuse (`populateO3DescriptorColumns` called with compacted metadata) that doesn't exist today.
- Process unchanged from the prior review: the enterprise tandem (#1016 — `authorizeAlterTableSetFormat`, replica `deniedOnReplica`) must land together; the red enterprise CI is the expected interlock (OSS compiles; the OSS-side abstract hook and all OSS `SecurityContext` impls are consistent and `ReadOnlySecurityContext` denies).

The single highest-value action is a one-line defensive `assert` on the strided-timestamp `writerIndex == timestampIndex` coupling (Minor #1) — cheap insurance against a silent-corruption footgun that today is held safe only by an unrelated invariant.

<sub>🤖 Delta level-3 review via [Claude Code](https://claude.com/claude-code)</sub>

bluestreak01:
@ideoma — deep review of this PR (multi-agent pass with per-finding source verification; ~27 draft findings checked, 14 dismissed as false positives). Summary at the bottom.

---

## Critical

### C1. `ALTER TABLE SET TYPE BYPASS WAL` escapes the "FORMAT PARQUET is WAL-only" invariant → corruption-capable state — **out-of-diff**
`TableConverter.convertTables` rewrites only the WAL bool (`TableConverter.java:136`) and preserves the new tableFormat int, and `alterTableSetType` (`SqlCompilerImpl.java:1749-1773`) has no format check — while CREATE (`SqlParser.java:1874`), ALTER SET FORMAT (`SqlCompilerImpl.java:1676`), and `TableWriter.setMetaTableFormat` (`TableWriter.java:3162-3170`) all enforce WAL-only. After `CREATE ... FORMAT PARQUET WAL` → `SET TYPE BYPASS WAL` → restart, you get a non-WAL parquet-format table. Concrete breakage: once its last partition is parquet, the non-WAL append path has no guard — `openLastPartitionAndSetAppendPosition` early-returns for parquet (`TableWriter.java:8643-8646`), yet `newRow` (`TableWriter.java:2576-2597`) falls through to append for in-order timestamps, writing into never-opened column memories. WAL tables never reach `newRow`, so only this escaped state hits it. **Fix:** reject the conversion (mirror `SqlCompilerImpl.java:1676-1678` in `alterTableSetType`) or reset format to NATIVE in `TableConverter`.

### C2. Configured bloom filters silently dropped on every fresh parquet partition — permanently — **in-diff**
`writeFreshParquetFromO3` passes `0, 0` for `bloomFilterColumnIndexesPtr/Count` (`O3PartitionJob.java:3753-3754`). The CONVERT path derives the list from per-column `PARQUET_CONFIG_BLOOM_FILTER_FLAG` (bit 25) metadata (`TableUtils.java:1938-1959`). The `parquetEncodingConfig` passed per column does not help: Rust reads bloom columns solely from the explicit ptr/count (`jni.rs:521-525`), never from bit 25. Knock-on: `PartitionUpdater` detects bloom columns from the existing file's row group 0 (`update.rs:234-249`), so a partition born bloom-less can never acquire filters through later merges. A user who configures bloom filters and creates a FORMAT PARQUET table gets files that silently ignore the config forever. **Fix:** extract the `TableUtils.java:1941-1951` loop into a shared helper and pass the flagged columns' dense descriptor indexes from `writeFreshParquetFromO3`.

### C3. Parquet data file is never fsynced; fresh path also skips the parent-dir fsync — **in-diff**
Under `CommitMode != NOSYNC`: the native O3 path msync+fsyncs column data (`O3CopyJob.java:654-655, 736-748`); the CONVERT path fsyncs `_pm` and the partition dir with an explicit rationale comment (`TableUtils.java:1980-1995`). `writeFreshParquetFromO3` fsyncs only the `_pm` fd (`O3PartitionJob.java:3763-3765`) — no dir fsync, and the parquet **data** file is never synced anywhere: Rust does `File::create` + write + drop (`jni.rs:527, 565-567`); the only `sync_data` in `parquet_write` is the `_pm` in `update.rs:1065`. Power loss after the `_txn` commit can lose the data file's contents or the dir entry while `_txn` references the partition → unopenable partition, data loss, on the feature's primary write path. Nuance: the data-file gap also exists in the pre-existing CONVERT path, but this PR makes it the steady-state write path for parquet tables. **Fix:** fsync the parquet data fd (Rust-side `sync_data` before close, or Java-side) and fsync the parent dir like `TableUtils.java:1987-1995`.

---

## Moderate

**M1. Replace-range commits × FORMAT PARQUET (in-diff).** With new-partition `isParquet` now true on parquet-format tables (`TableWriter.java:9236-9238`), a replace-range commit targeting a not-yet-existing partition throws `CairoException.critical("commit replace mode is not supported for Parquet partitions")` and suspends the table (`TableWriter.java:9422-9431`). Mat views (the main replace users) can't be parquet, but `WalWriter.commitWithParams` (`WalWriter.java:345`) is a public embedded API. Additionally, `writeFreshParquetFromO3` stamps the passed `o3TimestampMin` (the replace-range lo, possibly earlier than the first actual row) into sink slot 1, whereas the native brand-new-partition path recalculates it from the data (`O3PartitionJob.java:840-846`) — with `isFirstPartitionReplaced`, `txWriter.minTimestamp` can undershoot real data. Document the limitation, add a rejection test, and recalc `o3TimestampMin` in the fresh path.

**M2. Unchecked int casts guarded only by disabled-in-prod asserts (in-diff).** The coalesced-run row count is summed into a long and cast to int behind `assert rgRowCount <= Integer.MAX_VALUE` (`O3PartitionJob.java:2545-2550`), and `assert decoded == rowGroupSize` after `decodeRowGroupRange` (`O3PartitionJob.java:2877`). Production runs `-da` for customers; a malformed/crafted parquet file can violate both, and the recent direction (#7212) is hardening exactly this surface. Make both real checks throwing `CairoException`.

**M3. Coalesced-run memory spike is an undocumented tradeoff (in-diff).** A dedup commit on a partition where one timestamp straddles many row-group boundaries (mass-duplicate data: same ts across N row groups) coalesces the whole chain into ONE `decodeRowGroupRange` — the entire run is decoded into `RowGroupBuffers` at once, where the previous merge decoded one row group at a time. Worst case is the whole partition in memory, and the thread-local context retains that high-water capacity. The dedup-correctness rationale is sound, but the PR description doesn't mention the memory profile; at minimum document it.

**M4. Row-group timestamp stats scanned three times per dedup commit (in-diff, perf).** `hasCoalescableBoundaryTie` walks all adjacent pairs (`O3PartitionJob.java:2842-2853`); the same stats are re-read building `rowGroupBounds`; `computeMergeActions` re-runs the same `o3ContainsTimestamp` binary searches per tied boundary. The gate also runs before the cheaper rewrite triggers can short-circuit it. Build `rowGroupBounds` once before the rewrite decision and derive the tie scan from it — one O(R) pass. Also: the comment near `O3PartitionJob.java:387` claiming the fallback "decodes actual data" is stale; `rowGroupMin/MaxTimestamp` are O(1) `_pm` reads.

**M5. Strided timestamp encoders recompute min/max per row on input that is sorted by construction (in-diff, perf).** The merge index is ascending after `flattenIndex`, and `create_row_group_from_partitions` forces statistics for the designated timestamp — so stats always run. `write_required_i64_page` does per-row `MaxMin::update` (vs branchless SIMD on the contiguous path); `write_delta_i64_page` makes a whole second O(N) pass just for stats. Optimal: `min = slice[0][0]`, `max = slice[last][0]`, keeping the loop only when a bloom set is present. This is the always-on path for every fresh-partition write and COPY_O3 row group.

**M6. `decodeRowGroupRange` + Varchar→VarcharSlice conversion is a latent use-after-free (in-diff, defensive).** The range decoder copies the conversion overrides (`parquet_meta_decode.rs`) and appends with `reset_bufs=false`, but `fixup_varchar_slice_spill_pointers` (`column_sink/var.rs:552-587`) bakes absolute pointers into the current `data_vec` allocation per chunk; the next appended chunk can reallocate `data_vec`, dangling the prior chunk's pointers, and cleared spill markers prevent re-fixup. Not triggerable today (the only caller passes plain table types), but the API actively invites it — reject VarcharSlice in range mode or defer fixup to end-of-run.

**M7. `writeFreshParquetFromO3` never calls `addPhysicallyWrittenRows` (in-diff).** Native append (`TableWriter.java:9413`) and the parquet MERGE/COPY_O3 arms (`O3PartitionJob.java:558` etc.) account physically written rows; the fresh path writes `partitionRowCount` rows with zero accounting — write-amplification metrics under-report on exactly the tables the PR warns have the highest amplification.

**M8. Test coverage gaps.** Covered well: WAL-pipeline ALTER, boundary-tie dedup end-to-end (real `decodeRowGroupRange`), multi-row-group stride correctness, all three ts encodings, failure+RESUME recovery, TRUNCATE. Missing:
- v1 `_meta` migration read (TTL preserved, format defaults NATIVE) — the exact regression the version pinning prevents has no test
- `CREATE TABLE LIKE` a FORMAT PARQUET table (code copies format, `CreateTableOperationImpl.java:615`, zero tests)
- deterministic `AS SELECT` born-parquet assertion (fuzz-only today)
- deterministic indexed-symbol `WHERE sym='x'` query against a born-parquet partition (fresh path builds .k/.pk — only fuzz coverage)
- the rmdir-after-encode-failure branch (current injection fires before the encoder)
- mat view refresh over a FORMAT PARQUET base table
- replace-range rejection test

**M9. PR metadata.** Labels are empty — add `New feature`, `Core`, `SQL`, `rust`, `storage`, `WAL`. Title `feat(core): support parquet format table level` reads awkwardly for release notes — suggest `feat(core): support table-level Parquet partition format`. Description tone and tradeoff coverage are otherwise good (add the M3 memory note).

---

## Minor

- `SqlParser.java:1867-1872` (parse point 3): missing `formatSeen = true` is a dead store today, but set it anyway; a second FORMAT currently dies with generic "unexpected token" instead of "duplicate FORMAT clause".
- Boolean naming (is/has rule): `formatSeen` → `hasSeenFormat` (`SqlParser.java:1702`), `noLag` (`TableWriter.java:9654`), `private final boolean parquet` → `isParquet` (`FuzzSetTableFormatOperation.java:42`), `createWalAsParquet` (`FuzzRunner.java:102`).
- `OwnedMemoryPartitionDescriptor` is now entirely unused — delete the dead class.
- `dumpParquetPartition` (`ParquetWriteTest.java:3515`) has zero callers — leftover debug helper, delete.
- Promote the strided encoders' `debug_assert_eq!(columns.len(), 1)` to a returned error (`rle_dictionary/mod.rs:51`, `plain/primitive.rs`, `delta_binary_packed.rs`) — unreachable today, silent single-partition encode in release if ever reached.
- `writeFreshParquetFromO3` catch block relies on an undocumented prefix invariant after `updateParquetIndexes` clobbers the thread-local Path via `PostingIndexWriter`; a defensive `path.of(pathToTable)` in the catch (matching `processParquetPartition`) would be sturdier.
- `computeMergeActions` convenience overload is used only by tests — move to a test helper or mark it.
- `updateParquetIndexes` is invoked unconditionally (`O3PartitionJob.java:3782`), paying two mmap/munmap pairs + decoder init even with zero indexed symbols; an O(columnCount) pre-check skips it (same parity issue pre-exists at the other call site).
- Test ordering: `TableFormatTest` and `O3ParquetMergeStrategyTest` new methods violate alphabetical ordering (e.g. `testCoalesce*` after `testCustom*`; `testNewPartitionLandsAsParquet*` trio reversed); `assertParquetTimestampRoundTrip` helper sits among public tests.
- `WalWriterFuzzTest.java:233` comment says `0.01 // disabled` (it isn't); `:206` garbled sentence; `:267-269` stray blank line. `ExpParquetExportTest` modified 6-digit literals lack underscores; `ParquetWriteTest` new tests use lowercase SQL.
- `AlterOperationFlagsTest.testSetTableFormatRoundTrip` asserts byte-equality only — also assert the deserialized format value via `apply`.
- `testO3InsertCreatesParquetPartition` asserts the partition list but not row data.
- Pre-existing (not this PR, worth a follow-up issue): `PartitionUpdater.of(..., timestampIndex)` feeds a writer index into `SortingColumn::new` (`jni.rs:201-202`), which is a parquet schema ordinal — wrong sorting-columns metadata when a column before the timestamp was dropped.

---

1. *"ALTER SET FORMAT on WAL tables crashes via MetadataServiceStub"* — non-structural ALTERs travel as SQL text through WAL events and execute against `TableWriter.setMetaTableFormat` (`OperationExecutor.java:111`); the stub only ever sees structural ops, and `testAlterTableSetFormatParquet` passes the full pipeline.
2. *"timestampIndex passed in the wrong index space"* — Rust marks designated-ness by `col_id == timestamp_index` (`jni.rs:825`); both new call sites pass writer-index-space consistently, and for `TableWriterMetadata` position == writerIndex by construction (dropped columns keep their slots).
3. *"partitionNameTxn diverges between producer and consumer"* — the task carries `getTxn()` at publish; txn only increments at commit boundaries, so `txn - 1 == txWriter.getTxn() - 1` within the cycle; negative raw index is the documented insertion-point convention (`TxWriter.java:563-568`).
4. *"Sink offset 6 left uninitialized"* — pre-written per partition by `processO3Block` (`TableWriter.java:9309`) with exactly the right value; no task path writes it.
5. *"v1 `_meta` tables lose TTL/parquet-encoding config after LATEST bump"* — gates are pinned to per-feature constants (`hasParquetEncodingConfig` ≥1, TTL ≥1), and the diff converted the one risky caller; `isMetaFormatUpToDate` has zero production callers.
6. *"Duplicate FORMAT clause accepted after DEDUP"* — second FORMAT dies in `parseCreateTableExt` with "unexpected token"; the missing flag write is a dead store (kept as Minor).
7. *"CREATE ... FORMAT PARQUET AS SELECT writes native partitions"* — WAL-table AS SELECT copies via `WalWriter` (`SqlCompilerImpl.java:4088-4102`); data lands parquet through WAL apply (kept as a test gap only).
8. *"Multi-partition strided encode silently loses data in release"* — the strided bit flows only into single-partition encode entries (`file.rs:366-380`, `update.rs:412/480`); unreachable (hardening kept as Minor).
9. *"RLE-dict strided emits one unbounded page, unlike non-strided"* — the entire dict family emits 1 dict + 1 data page per chunk (`rle_dictionary/primitive.rs:17-18`); page size is row-group-bounded (default 100k rows); exact parity.
10. *"transmute_slice misalignment for `[i64;2]`"* — `align_of::<[i64;2]>()` is 8, identical to the pre-existing contiguous i64 transmute; `base + o3Lo*16` preserves it.
11. *"Empty-placeholder noLag gate too loose"* — forcing full commit for `rowCount==0` parquet tables is the safe side; no LAG data exists to lose.
12. *"ILP/CSV TableStructure adapters reset format to NATIVE"* — those adapters only create new tables, where NATIVE is the correct default; metadata rewrites preserve the stored format.
13. *"Encoder failure leaves task latches counted but no sink queued"* — the finally counts down, the consumer handles errors via `o3BumpErrorCount`; identical to `processParquetPartition`.
14. *"Thread-local Path clobber breaks the catch path"* — `trimTo(pathSize)` reconstructs correctly because the clobbered path shares the `pathToTable` prefix (fragility kept as Minor).

Also verified clean: the `file_offset`/`dictionary_page_offset` spec-compliance fixes are safe for all readers in both crates (readers use `dictionary_page_offset.unwrap_or(data_page_offset)`, never file_offset arithmetic); `copy_row_group` adjusts offsets on rewrite; append-mode aux offsets are absolute; `refresh_ptrs` always-recompute is required and safe; gap actions inside a coalesced run are provably empty; non-dedup commits correctly avoid the forced rewrite; resource/fd handling in `writeFreshParquetFromO3` is leak-free on all paths; member-ordering moves in `AlterOperation`/`TableWriterMetadata` actually restore alphabetical order.

---

## Summary

**Verdict: request changes.** The core design is sound and unusually well-verified in places (the strided encoder equivalence test, the boundary-tie end-to-end tests, the failure/RESUME tests), but three Critical items stand: the `SET TYPE BYPASS WAL` invariant escape (corruption-capable, out-of-diff), silently dropped bloom-filter config (permanent, unfixable post-hoc per file), and the missing data-file/dir fsync on the feature's primary write path (durability violation under sync commit modes). Tradeoffs to note in the description: forced O3 path for in-order writes (already noted) and the new whole-run decode memory profile under dedup (not noted).

Stats: ~27 draft findings verified, 14 dismissed as false positives. Of the 12 confirmed findings (3 Critical, 9 Moderate), 11 are in-diff and 1 is out-of-diff (`TableConverter`).

bluestreak01:
# Code Review: PR #7107 — feat(core): support parquet format table level

**Level-3 review** (11-agent fan-out + per-finding source verification). This PR has a conflicting review history — one approve track, one request-changes track with 3 Criticals. I re-adjudicated from source. The decisive finding: the working tree (HEAD `23eeed95ed`) **already contains fixes for two of the three disputed Criticals**, so that earlier review predates them.

## Critical

**None.** All three disputed Criticals resolve:

- **C1 (SET TYPE BYPASS WAL escapes the WAL-only invariant) — FIXED on branch.** Commit `769d097c11` added the guard in `alterTableSetType` (`SqlCompilerImpl.java:1767,1772` — rejects on both the table FORMAT flag and a parquet last partition) and an independent refusal in `TableConverter.convertTables` (`TableConverter.java:120-126`, `hasParquetFormatOrLastPartition`). The corruption-capable state (non-WAL `newRow` appending into never-opened parquet column memory) is now unreachable through supported SQL. The write-path hazard itself is real — `openLastPartitionAndSetAppendPosition` early-returns for a parquet last partition without opening column memories — which is why the upstream guards are load-bearing.
- **C2 (configured bloom filters silently dropped on fresh parquet partitions) — FIXED on branch.** Commit `78d1c58982` made `writeFreshParquetFromO3` derive the flagged columns via `TableUtils.deriveBloomFilterColumnIndexes` and pass a real ptr/count (`O3PartitionJob.java:3748-3766`), symmetric with the CONVERT path, with a regression test (`testFreshParquetHonorsBloomFilterColumns`, `TableFormatTest.java:431`).
- **C3 (durability / fsync) — confirmed real but reframed to Moderate** (M1). Not a regression introduced by this PR's code.

## Moderate

**M1 — Parquet data file is never fsynced; under sync commit mode this is a durability asymmetry vs native tables (pre-existing, product-wide).** Across the whole Rust crate the only `sync_data` in `parquet_write` is for `_pm` (`update.rs:1065`). The parquet *data* file is written via `File::create` + buffered writes + drop, never synced; `writeFreshParquetFromO3` fsyncs only `parquetMetaFd` (`O3PartitionJob.java:3775-3777`). The native O3 path msync+fsyncs column data under `commitMode != NOSYNC` (`O3CopyJob.java:736-747`). So after a successful sync-mode commit, a native partition's data survives power loss but a parquet partition's may not (`_txn` references a data file still in the page cache).

Honest framing (corrects both prior tracks): this gap is **pre-existing and product-wide for parquet** — CONVERT *and* the in-PR rewrite path (`processParquetPartition`, `syncParquetMeta()` at `O3PartitionJob.java:580`) also sync only `_pm`. The PR follows the existing parquet convention; what it changes is making fresh-parquet the **steady-state ingestion path**, so the gap now applies to normal writes, not just explicit conversions. Not a hard blocker (consistent with all existing parquet paths), but it warrants a tracked follow-up for parquet durability under sync commit mode; if native parity is wanted, the fresh path is the natural place to add a data-file `sync_data` (Rust-side, gated on commit mode). The dir-fsync sub-claim is **minor**: the fresh path matches its sibling rewrite path; CONVERT is the lone outlier that also fsyncs the dir, so the "product-wide norm" downgrade holds for the directory.

**M2 — Corruption-class `assert`s on the parquet decode/merge worker path should be real `CairoException` checks.** `O3PartitionJob.java:2152` (`assert decoded == rowGroupSize`) guards merge-index sizing after `decodeRowGroupRange`; `:458`/`:505` (`assert rgRowCount/copyRgRowCount <= Integer.MAX_VALUE`) guard `long→int` casts of coalesced-run row counts; `:3534` similar. These run over data decoded from a parquet file. A malformed/foreign file whose metadata row counts disagree with the decoded payload would, with assertions disabled, read past the decoded buffer or wrap a negative int — on the WAL-apply worker thread. This aligns with the team's recent hardening direction (#7212, already on this branch).

**M3 — Perf: strided timestamp encoders compute min/max with a scalar per-row loop (delta makes a second full O(N) pass) on input that is ascending by construction.** The merge index is fully sorted before partition processing (QuestDB itself reads `o3TimestampMin = index[0]`, `o3TimestampMax = index[last]`), and the strided encoder receives a contiguous sub-slice. So for the designated timestamp `min = slice[0]`, `max = slice[last]` is **O(1)**, replacing the O(N) `MaxMin::update` loop (`plain/primitive.rs:140-148`) and eliminating the dedicated stats pass in the delta path (`delta_binary_packed.rs:155-179`, which documents "the iterator is consumed twice"). The contiguous CONVERT path already computes these branchlessly via SIMD fused into the def-level scan; the new strided path is scalar/branchy on more-structured input. This is the always-on write path for every fresh partition and every COPY_O3 row group.

**M4 — Test coverage gaps** (ranked by risk):
- **Post-encoder failure cleanup branch is untested.** Both injection tests fault at `_pm` open, before the encoder. The path where the encoder succeeds and `updateParquetIndexes` throws — rmdir of a partition dir already holding a written `data.parquet` (`O3PartitionJob.java:3824-3840`) — has no coverage, and depends on a fragile thread-local `Path`-prefix invariant (clobbered by `PostingIndexWriter` inside `updateParquetIndexes`).
- **Rust strided `delta_binary_packed` and `rle_dictionary` encoders have no unit test.** Only Plain got `encode_designated_timestamp_strided_matches_contiguous`. None of the three `Contiguous => Err(InvalidLayout)` arms, the delta `repetition != Required` arm, nor the rle `empty => Ok(vec![])` arm have a Rust unit test.
- **v1→v2 `_meta` read path untested** — no test asserts an old minor-version-1 `_meta` reads back `getTableFormat() == NATIVE` with TTL preserved (the exact regression the version-constant pinning prevents).
- **`CREATE TABLE (LIKE src)` inheriting FORMAT PARQUET** (`CreateTableOperationImpl.updateFromLikeTableMetadata`) and **mat-view refresh over a FORMAT PARQUET base table** are both unexercised.

## Minor

- **`writeFreshParquetFromO3` never calls `addPhysicallyWrittenRows`** (`O3PartitionJob.java:3650`), unlike the native fresh path (`:3199`) and every merge arm (`:500/533/558`). Write-amplification metrics under-report on exactly the tables the PR warns have the highest amplification. (Metrics-only.)
- **Dedup boundary-tie detection has a duplicated binary-search pass and an eager gate.** `hasCoalescableBoundaryTie` and the `computeMergeActions` coalesce loop run the identical `o3ContainsTimestamp` search; the gate is a pre-computed variable, so `||` short-circuiting doesn't stop it running when the cheaper O(1) unused-bytes triggers already force a rewrite. (Per-commit; passes 1-2 read O(1) footer stats, not decoded data.)
- **`updateParquetIndexes` does two `mmap`s + decoder spin-up before checking whether any column is an indexed SYMBOL** (`O3PartitionJob.java:3441-3479`), paid per fresh partition even with zero indexed columns.
- **`populateO3DescriptorColumns` calls `metadata.getColumnMetadata(i)` 4x/column** (`:2587-2593`) — hoist to one local.
- **Coalesced-run decode loads the whole tied run into one buffer** (vs previous per-row-group), and the thread-local `RowGroupBuffers` retains that high-water capacity for the worker's life. Correctness-required, but the memory profile isn't in the PR description — document it.
- **Orphaned `dumpParquetPartition` helper** (`ParquetWriteTest.java:3515`) — zero callers; delete.
- **`formatSeen` not set in the after-DEDUP FORMAT parse block** (`SqlParser.java:1868` checks it but the block doesn't set it) — dead store today; set for symmetry.
- **Strided encoders use `debug_assert_eq!(columns.len(), 1)`** (`plain/primitive.rs:82`, `delta_binary_packed.rs:99`, `rle_dictionary/mod.rs:51`) — release-unsafe but each is backed by a real `Err` return; promote to a hard error. Same for the `transmute_slice` 16-byte-multiple/alignment guard (load-bearing, debug-only, unreachable because the Java caller always passes `rowCount * 16`).
- **Boolean naming**: `formatSeen` → `hasSeenFormat`, plus `noLag`, `createWalAsParquet`, `private final boolean parquet` per the `is/has` rule.
- **`testO3InsertCreatesParquetPartition` asserts the partition list but not row data**; **`AlterOperationFlagsTest.testSetTableFormatRoundTrip` asserts byte-equality only** — also read the decoded format back from the deserialized op.
- **PR labels are empty** — add `New feature`, `Core`, `SQL`, `rust`, `storage`, `WAL`.

## Downgraded (false positives / verified correct)

- **C1, C2** — real when written, **fixed on branch** (`769d097c11`, `78d1c58982`) with guards + regression test.
- **C3 "fresh path is worse than CONVERT on dir-fsync"** — the fresh path matches its sibling rewrite path; only CONVERT also fsyncs the dir. Reframed to the data-file gap (M1).
- **`_pm` records `Required` repetition for no-null SYMBOL columns** — pre-existing (`parquet_write/jni.rs` is not in the diff), and harmless: decode keys off `max_def_level` (correctly 1), `copy_row_group` copies raw bytes against the target schema. Separate follow-up, not a finding against this PR.
- **Rust panics across JNI** — none reachable. The strided transmute is guarded by the Java `rowCount * 16` sizing; the remaining `column_chunk.rs` `.unwrap()`s are safe because `write_page` sets `type_` and the matching header `Some(..)` in lockstep; `decode_row_group_range` bounds-checks indices and returns `Err`; offset arithmetic can't overflow at QuestDB scale.
- **Resource leaks / latch hangs / format-decision races** — none. The `_pm` fd is closed on every path with no double-close; latch counted down exactly once in `finally`; `isParquet` is snapshotted on the writer thread and `setMetaTableFormat` calls `commit()` to drain in-flight O3 before mutating the format. Thread-locals are per-thread, reclaimed at pool shutdown.
- **Test assertion-API compliance** — clean. No new query-result `assertSql`; the lone `.returnsOnce` (`ParquetWriteTest.java:615`) is pre-existing and justified.
- **parquet2 `file_offset`/`dictionary_page_offset` rewrite** — spec-correct fix (old formula overshot end-of-pages by the dict-page size); reader-safe (`file_offset()` has zero callers).

## Summary

**Verdict: Approve** — converging with the approve track, because the two blocking Criticals from the request-changes track were subsequently fixed on the branch. No Critical issues remain; the correctness-critical machinery (fresh O3 write, strided encoding, dedup coalescing, fd/latch ownership, `_meta` v1→v2 read, Rust panic safety) verified sound.

The one genuine durability observation (M1: parquet data file never fsynced) is real but pre-existing and product-wide for parquet — defensible to land here and track separately, since the new fresh path is consistent with the existing rewrite/CONVERT convention. Worth doing before merge: the M2 assert→exception hardening on the decode path, the M3 O(1) min/max win on the always-on write path, and the M4 test gaps (post-encoder failure branch + the two untested Rust strided encoders).

- **Tradeoffs to keep in the description**: the forced-O3 path for in-order writes (already noted) and the new whole-run decode memory profile under dedup (not noted).
- **Verification stats**: ~24 draft findings verified against source; 2 headline Criticals dropped as fixed-on-branch, ~6 dropped as false-positive/pre-existing. Surviving: 4 Moderate, ~12 Minor — all in-diff.
- **Process**: the enterprise tandem (#1016 — `authorizeAlterTableSetFormat`, replica `deniedOnReplica`) must land together; a red enterprise CI is the expected interlock (OSS compiles; `ReadOnlySecurityContext` denies, all OSS `SecurityContext` impls consistent).

<sub>🤖 Level-3 review via [Claude Code](https://claude.com/claude-code)</sub>

ideoma:
@bluestreak01 — flagging one change since your last review (which had C1 resolved via `769d097`): the latest commit, **`04db252` ("address C1")**, takes a different approach to C1 than the guard you approved.

`769d097` *blocked* `SET TYPE BYPASS WAL` on a parquet table (reject in `alterTableSetType` + a refusal in `TableConverter`). `04db252` removes those guards and instead **allows** the conversion. Why:

- Converting a WAL table to non-WAL is a deliberate operational escape hatch — the standard way to clear a poison-pill WAL transaction (one that keeps failing to apply and suspends the table) is to go non-WAL, which drops the WAL and its sequencer, then convert back to WAL. Rejecting the conversion (or silently resetting `FORMAT` to `NATIVE`) takes that away.
- The escape isn't introduced by this PR: `master` has no parquet guard in `alterTableSetType`, and the corruption path you described (non-WAL + parquet last partition + in-order insert) is already reachable on `master` via `CONVERT PARTITION TO PARQUET` on a plain WAL table's last partition, then `SET TYPE BYPASS WAL`. This PR only adds the table-level `FORMAT` flag; it doesn't touch the conversion — so guarding it here would reach beyond the PR's scope and disable a capability that works today.
- The intended flow never writes while non-WAL; it converts straight back to WAL. The `FORMAT` flag is preserved across the round-trip.

`TableFormatTest` updated to match: `testSetTypeBypassWalConvertsFormatParquetTable`, `testSetTypeBypassWalConvertsParquetLastPartition`, and `testSetTypeBypassWalRoundTripPreservesFormat` (format + data preserved across WAL → non-WAL → WAL).

bluestreak01 (dismissed):
Approve (level-3 review). No Critical issues found — the correctness-critical machinery all verified sound against the current branch:

- Fresh O3→parquet write (`writeFreshParquetFromO3`): fd/latch ownership balanced (exactly one countdown in `finally`; error path closes fd, rmdirs the dir, bumps error count, throws; `processPartition` returns immediately after — no double-decrement).
- New-parquet branch in `o3ConsumePartitionUpdateSink`: `updateAttachedPartitionSizeByRawIndex(<0,…)` correctly decodes the binary-search insertion point, so the partition inserts at the right position.
- Var-size aux-offset slicing: sorted O3 buffers have monotonic absolute offsets; the Rust varlen encoder reads absolute offsets (no `aux[0]==0` assumption); `dataExtent` bounds are exact.
- Strided timestamp: writer-index == dense-index for the designated timestamp; multi-row-group splitting narrows per group in `file.rs`.
- Dedup coalescing: rewrite gate (`hasCoalescableBoundaryTie`) and `computeMergeActions(coalesce=isCommitDedup)` use the identical `hasTimestampStats` predicate against the same `meta` — a coalesced multi-group MERGE can only be emitted under dedup + an actual tie, which also forces rewrite. Lockstep confirmed.
- `_meta` v1→v2: checksum hashes only metadataVersion+columnCount; both former `isMetaFormatUpToDate` callers (TTL, parquet-encoding-config) re-pinned to explicit version constants, so the LATEST bump cannot regress v1 reads.
- parquet2 offset rewrite: spec-correct; read paths resolve chunk start via `byte_range()`/dictionary offset, old and new files yield identical starts, `file_offset()` has no readers.

Cross-context: all six widely-used changed symbols SAFE at every out-of-diff callsite. All OSS `SecurityContext` implementors compile.

Non-blocking follow-ups (do not block merge):
- M1 (perf): `commit00` now calls `txWriter.hasParquetPartitions()` (O(numPartitions) scan) unconditionally on every commit of every table. Guard with `metadata.getTableFormat()==TABLE_FORMAT_PARQUET || hasNotifiedParquetPartitions`.
- M2 (durability): `writeFreshParquetFromO3` fsyncs `_pm` but not the brand-new partition directory; the CONVERT path fsyncs both for the same brand-new-dir scenario. Mirror its dir-fsync (non-Windows).
- M3/M4 (test gaps): `_meta` v1→v2 read path; var-size multi-row-group fresh commit; indexed-SYMBOL fresh partition; all-null non-key column in a coalesced dedup run.
- Minor: dead `getMetaTableFormat()` / `OwnedMemoryPartitionDescriptor`; orphaned `dumpParquetPartition` test helper; bloom skip predicate `<=0` vs `<0`; `formatSeen`→`hasSeenFormat`; empty PR labels.

Note: enterprise tandem #1016 must land together (a red enterprise CI is the expected interlock — OSS compiles).

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 1219 / 1333 (91.45%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/wal/seq/MetadataServiceStub.java|0|2|00.00%|
|:large_blue_circle:|qdbr/src/parquet_read/parquet_meta_decode.rs|85|116|73.28%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/AlterOperation.java|11|14|78.57%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/partition_decoder.rs|68|84|80.95%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CreateTableOperationBuilderImpl.java|10|12|83.33%|
|:large_blue_circle:|qdbr/src/parquet_write/encoders/delta_binary_packed.rs|62|74|83.78%|
|:large_blue_circle:|qdbr/src/parquet_write/encoders/plain/primitive.rs|61|72|84.72%|
|:large_blue_circle:|qdbr/src/parquet_write/encoders/rle_dictionary/mod.rs|53|61|86.89%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|33|36|91.67%|
|:large_blue_circle:|qdbr/src/parquet_write/schema.rs|50|54|92.59%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|171|182|93.96%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|23|24|95.83%|
|:large_blue_circle:|qdbr/src/parquet_write/mod.rs|19|20|95.00%|
|:large_blue_circle:|qdbr/src/parquet_write/encoders/plain/tests.rs|66|68|97.06%|
|:large_blue_circle:|io/questdb/cairo/O3ParquetMergeStrategy.java|34|35|97.14%|
|:large_blue_circle:|qdbr/src/parquet_write/update.rs|55|56|98.21%|
|:large_blue_circle:|parquet2/src/write/column_chunk.rs|283|288|98.26%|
|:large_blue_circle:|io/questdb/cairo/TableReaderMetadata.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/O3ParquetMergeContext.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlParser.java|25|25|100.00%|
|:large_blue_circle:|qdbr/src/parquet_metadata/convert.rs|19|19|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/file.rs|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriterMetadata.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoTable.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/MetadataCache.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowCreateTableRecordCursorFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/AlterOperationBuilder.java|6|6|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/encode.rs|10|10|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableUtils.java|18|18|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decode_column.rs|9|9|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CreateTableOperationImpl.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableStructure.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/ParquetPartitionDecoder.java|7|7|100.00%|

Issue #26, opened 2014-12-13
Visualisation and / or Grafana Support for Stored Data


sirinath:
One of main applications of stored TS data would be visualise it. Perhaps adding this functionality out of the box would be an advantage.

May be through: http://grafana.org/

PR #7108, opened 2026-05-15
feat(qwp): expand ingest wire-type coverage and advertise batch cap

bluestreak01:
Tandem: https://github.com/questdb/questdb-enterprise/pull/1010

## Summary

This PR bundles several QWP ingest improvements that landed on the same branch:

### 1. Server advertises max batch size on handshake

- QWP ingest senders previously had no way to discover the server's payload cap. They learned it by sending a batch larger than 16 MB and reading `STATUS_PARSE_ERROR`. The 101 upgrade response on `/write/v4` now carries an `X-QWP-Max-Batch-Size` header with the server's effective payload cap (clamped to the smaller of the protocol ceiling and the HTTP recv buffer minus WebSocket framing overhead).
- The header is additive — older clients ignore it. The paired client-side parsing and budget clamping land in questdb/java-questdb-client#26.

### 2. IPv4 ingest support

- QWP egress had advertised `TYPE_IPV4` (0x18) since v1 cut, but ingress rejected it: `QwpColumnDef.validate()` range-checked `0x01..0x16 (CHAR)`, so any sender pushing an IPv4 column failed with `INVALID_COLUMN_TYPE` before the batch reached the table writer.
- `validate()` switches from a range check to an explicit allowlist that admits `TYPE_IPV4`. `BINARY` (0x17) stays rejected — no ingest decoder exists yet, but the explicit list now surfaces that gap rather than hiding it behind a range.
- `QwpConstants.getFixedTypeSize` / `isFixedWidthType`, `QwpFixedWidthColumnCursor`, and `QwpTableBlockCursor` treat IPv4 as a 4-byte fixed-width type, identical to INT on the wire.
- `QwpWalAppender` gains `TYPE_IPV4 -> ColumnType.IPv4` in `mapQwpTypeToQuestDB` (enables auto-create) and a dedicated IPv4 dispatch arm in `appendToWalColumnar` that accepts `TYPE_IPV4` or `TYPE_INT` wire input. Other coercions are rejected — IPv4 stays distinct from INT on read.
- `WalColumnarRowAppender.putFixedColumn` previously threw `unsupported column type for direct copy: IPv4` whenever any row in a batch was null (the fast path is a raw memcpy and worked without a switch). The null-aware switch now writes `Numbers.IPv4_NULL` (0) for null-bitmap positions and copies the int otherwise — matching QuestDB's IPv4 NULL convention.

The paired client-side IPv4 sender API lands in questdb/java-questdb-client#26.

### 3. GEOHASH end-to-end and fuzz coverage

The companion client adds the public `geoHashColumn(...)` API; this PR adds the end-to-end test coverage that proves GEOHASH round-trips through the server across all four storage classes, plus the bit-precise (non-multiple-of-5) wire path.

- `QwpSenderE2ETest` gains three GEOHASH tests:
  - `testAutoCreateGeoHashColumns` drives the new sender API in both overloads across GEOHASH(5b) / (15b) / (20b) / (35b), exercising the four storage classes (GEOBYTE / SHORT / INT / LONG) and the implicit-NULL bitmap path.
  - `testGeoHashColumnValidation` covers precision out of range, null / empty / over-12-char / invalid-base32 / cross-row precision mismatch — client-side `LineSenderException` paths that never reach the wire.
  - `testGeoHashColumnPrecisionPersistsAcrossBatches` pins the between-batch precision lock: a batch with no values for a previously-flushed GEOHASH column must still encode the schema-locked precision varint. Without the matching client fix, the server rejects the batch as `wireBits=1` vs the auto-created `GEOHASH(Nc)`.
- `QwpIngressOracleFuzzTest` adds five base GEOHASH columns at precisions 5 / 7 / 15 / 20 / 35 bits (covering the four storage classes plus one non-multiple-of-5 to exercise the bit-precise wire path), and four auto-create GEOHASH extras in `injectExtra` — `ex_g20` routes through the base32-string overload so the client-side decoder gets coverage on the auto-create path too.
- `QwpSenderFuzzTest` adds four typed GEOHASH columns to the always-written typed block, using `ColumnType.GEOBYTE / GEOSHORT / GEOINT / GEOLONG` as switch discriminants with fixed precision per case (multiples of 5 only so CursorPrinter's base32 render matches the yielded oracle string verbatim).
- `QwpRow` (the oracle row model) gains `ValueType.GEOHASH`, `setGeoHash(name, bits, precision)` plus a base32-string overload, dispatch through `qwp.geoHashColumn(...)`, and four storage-class-specific `assertCell` arms. `TypedValue`'s leftover IPv4 `i` field is renamed to `ipv4` to live cleanly next to `i32`.
- `QwpCursorBoundsCheckTest` had a gap that surfaced on CI: random-byte corruption that flipped a column-type byte to `TYPE_IPV4` (0x18) made the test's value consumer assert `unsupported fuzz column type: 24`. `TYPE_IPV4` is now in the consumer's `TYPE_INT` group, and an `ipv4` column is populated in `populateFuzzTable` so the non-corruption path exercises it too.

`questdb.client.version` default bumped to `1.2.2-SNAPSHOT` so e2e tests exercise the new client APIs. Pin back with `-Dquestdb.client.version=1.2.1` if you need to reproduce against the released client.

### 4. Cursor sentinel-null for IPv4 / UUID / LONG256

`QwpFixedWidthColumnCursor.readCurrentValue` handles `TYPE_IPV4`, `TYPE_UUID`, and `TYPE_LONG256`, but the matching `isCurrentValueSentinelNull` switch only handled `TYPE_INT` / `TYPE_LONG` / `TYPE_FLOAT` / `TYPE_DOUBLE`. The wire protocol omits the null bitmap whenever `nullCount == 0` (see `QwpColumnWriter.writeNullHeader`), so the sentinel switch is the protocol's intended null detector for those columns. The gap meant any consumer reading these types through `cursor.isNull()` saw the NULL bit-pattern as a real value.

- The bulk WAL ingest path (`WalColumnarRowAppender.putFixedColumn`) bypasses the cursor with a `nullBitmapAddress == 0 → bulk memcpy` fast path, so existing same-type ingest is unaffected.
- The type-conversion paths `putFixedOtherToStringColumn` / `putFixedOtherToVarcharColumn` dispatch UUID and LONG256 through `formatFixedOtherValue` and call `cursor.isNull()` per row, so the gap was live for those two types. After the fix, the NULL bit-pattern is correctly written as NULL into the target STRING / VARCHAR column rather than the literal `"00000000-0000-0000-0000-000000000000"` or `"0x000...000"`.
- IPv4 was latent (no `formatFixedOtherValue` arm dispatches to it today), but the asymmetry is closed for consistency with `readCurrentValue` and to harden against the next consumer that hits the cursor.

Three regression tests added to `QwpFixedWidthDecoderTest` mirror the existing `testDecodeIntColumnSentinelNullNoBitmap` pattern: `testDecodeIPv4ColumnSentinelNullNoBitmap`, `testDecodeUuidColumnSentinelNullNoBitmap`, `testDecodeLong256ColumnSentinelNullNoBitmap`.

### 5. End-to-end batch-cap regression tests

`QwpSenderOversizeRowInBatchTest` stands up a single QwpSidecar with a tight 128 KB recv buffer (advertised cap ~131 KB) and drives two scenarios that pin the producer-side guards added in the paired client PR:

- `testOversizeRowInsideBatchTripsProducerGuard`: a small row followed by a 600 KB row to the same table. Asserts the per-row guard throws `"row too large for server batch cap"` on row 2's `at()`, not just on row 1.
- `testOversizeBatchFlushTripsProducerGuard`: five 30 KB rows with byte-based auto-flush disabled, then a manual `flush()`. Asserts the flush-time guard throws `"batch too large for server batch cap"` rather than letting the oversize WS frame reach the wire (where the server would close the connection with `ws-close[1009]` asynchronously).

Both tests fail with "expected exception not thrown" against an unfixed client and pass against the matching client commit — true regression tests, not smoke tests.

Plus a minor tidy of `QwpSenderFailoverBatchSizeTest`: `countOf` signature simplified, sidecar made `static`, and a `drainWalQueue() + engine.releaseInactive()` block added to the teardown to satisfy `assertMemoryLeak`.

### 6. JMH micro-bench for the `pendingBytes` hot path

`QwpTotalBufferedBytesBenchmark` in the `benchmarks` module compares three implementations of the per-row `pendingBytes` update:

- `walkAllTables` — the pre-fix behaviour (walk every table per row).
- `currentTableOnly` — what the new delta-based code does (look at the active table only).
- `shortCircuit` — what a no-op would cost (lower bound for reference).

Parametrized on `numTables ∈ {1, 4, 16}` and `numColumns ∈ {1, 8}`. Confirms the `O(numTables)` shape of the old code (12.8 ns/row at 1 table, 264 ns/row at 16 tables on the dev machine) and the flat per-table cost of the new approach (~9 ns/row regardless of table count). Not a CI gate; ship-and-keep as documentation for the hot path.

### 7. Cross-type ingest hardening and test pins

Seven follow-ups that close asymmetric type-routing and test-coverage gaps surfaced by the new `TYPE_IPV4` / `TYPE_BINARY` allowlist additions and the cursor sentinel-null fix.

- **INT_NULL → IPv4_NULL translation.** The IPv4 arm of `appendToWalColumnar` accepts `qwpType == TYPE_INT` as a legacy-client migration path, but INT and IPv4 disagree on the NULL bit pattern (`INT_NULL = Integer.MIN_VALUE` vs `IPv4_NULL = 0`). Without translation, the fast-path memcpy in `WalColumnarRowAppender.putFixedColumn` would land `MIN_VALUE` verbatim and the value would read back as `128.0.0.0` instead of NULL. New `putIntToIPv4Column` method (added to both the `ColumnarRowAppender` interface and the WAL impl) uses the sentinel-aware `cursor.isNull()` to translate per row. Pinned by `testIntColumnIntoIPv4TranslatesNullSentinel`.
- **BINARY → VARCHAR coercion guard.** The VARCHAR arm of `appendToWalColumnar` pattern-matched the cursor by class only, so `TYPE_BINARY` wire data (newly valid after the allowlist refactor) routed through `putVarcharColumn` unchecked — raw possibly-non-UTF-8 bytes landing in a column QuestDB treats as UTF-8. A guard now rejects `qwpType == TYPE_BINARY` into a VARCHAR target with `coercionNotSupportedException`. Symmetric to the existing guard in the BINARY arm. Pinned by `testBinarySourceRejectedByVarcharTarget`.
- **UUID / LONG256 sentinel-null E2E pin.** The cursor sentinel-null arms for `TYPE_UUID` and `TYPE_LONG256` (section 4) flow through `cursor.isNull()` consumed by `putFixedOtherToStringColumn` / `putFixedOtherToVarcharColumn`, but existing `testCoercionToString` / `testCoercionToVarchar` only exercised random non-sentinel values. New `testCoercionToStringPreservesUuidAndLong256NullSentinels` sends `(LONG_NULL, LONG_NULL)` UUIDs and all-`LONG_NULL` LONG256 values into STRING and VARCHAR targets and asserts SQL NULL on read.
- **IPv4 → STRING / VARCHAR formatter arm.** `formatFixedOtherValue` had no `TYPE_IPV4` case, so a client sending `TYPE_IPV4` into a pre-existing STRING / VARCHAR column threw `"unsupported wire type for string conversion: 24"` mid-row and aborted the whole batch. Added a `TYPE_IPV4` arm that calls `Numbers.intToIPv4Sink` for dotted-quad rendering; the upstream `cursor.isNull()` filter already short-circuits the IPv4 NULL sentinel so this arm only fires for real addresses. Pinned by `testCoercionToStringAndVarcharFromIPv4`.
- **`setBinary` precondition tightened** (`QwpRow.java`, fuzz oracle helper). The Javadoc said "pass `new byte[0]` for empty", but `MemoryPARWImpl.putBin` writes `NULL_LEN` when `len == 0` so the column reads back as SQL NULL while `assertCell`'s BINARY arm asserts non-null + length match. Fuzz tests never emit length-0 today; the contract is tightened so future scenarios fail at the `setBinary` call site rather than at read-back assertion time.
- **Handshake cap test pin.** `testOnHeadersReadyAdvertisesEffectiveBatchSize` re-derived the "expected" value with the same `Math.min(Math.max(0, recvBufferSize - 14), DEFAULT_MAX_BATCH_SIZE)` formula the production code uses, so a drift in `MAX_WS_FRAME_HEADER_BYTES` (currently 14) would silently track on both sides. The test now pins the literal `131058` for a fixed `recvBufferSize` of 128 KiB — a future drift on either side fails this assertion.
- **`isFixedTypeCoercionAllowed` documentation.** The reverse direction `TYPE_IPV4` wire into a `BYTE` / `SHORT` / `INT` / `LONG` column is intentionally rejected (a valid IPv4 with bit pattern `Integer.MIN_VALUE` would land as `INT_NULL` and surface as SQL NULL on read — a worse surprise than rejecting). Javadoc on `isFixedTypeCoercionAllowed` and a cross-reference comment in the IPv4 arm now make this asymmetry and its NULL-sentinel reasoning explicit. No behaviour change.

### 8. Cache QWP handshake response header bytes (zero-GC)

Three response header values (`X-QWP-Version`, `X-QWP-Content-Encoding`,
`X-QWP-Max-Batch-Size`) were being built via
`Integer.toString().getBytes(US_ASCII)` and `String` concatenation on
every handshake. None of those values comes from client input — they
come from closed sets the server itself defines.

`QwpEgressCompressionNegotiator` now returns precomputed `byte[]`
instances from a per-zstd-level static table;
`QwpWebSocketHttpProcessor.VERSION_BYTES` is a static lookup by
version; `QwpWebSocketUpgradeProcessor` caches the max-batch-size
bytes at construction since the effective cap is derived from
`recvBufferSize` (per-instance config-fixed). Per-handshake
allocations for these three header values drop to zero. Wire format
byte-identical.

### 9. Default zstd compression level to 1 + operator force-level

Two related changes that give operators bounded CPU control over the
QWP egress compression path.

**Default**. The negotiator now picks level 1 (was 3) when a client
sends `zstd` without a level, or with an empty / unparseable level
value. Level 1 is the cheapest CPU; ratio gain at higher levels is
small for typical columnar payloads. Wire format unchanged.

**Force-level config**. New `qwp.egress.compression.force.level`
property (`PropertyKey.QWP_EGRESS_COMPRESSION_FORCE_LEVEL`).
- Value `0` (default) honors the client-negotiated level.
- Value in `[1, 9]` forces every ZSTD-negotiated connection on this
  server to that level regardless of what the client asked. Operators
  capping a misbehaving client at level 1 must beat clients that ask
  for level 9.

Reloadable: the property belongs to
`DynamicPropServerConfiguration.dynamicProps`, and
`QwpEgressUpgradeProcessor` reads
`engine.getConfiguration().getQwpEgressForcedZstdLevel()` on every
handshake (no caching at construction). Already-established
connections keep their negotiated level — mutating an in-flight ZSTD
context is not safe; the next new connection picks up the new value.

Out-of-range property values are rejected at startup; the override
site clamps as defense-in-depth.

New `QwpEgressCompressionNegotiatorTest` pins 11 cases between the
default-level (bare zstd, empty level, unparseable level), the
[1, 9] wire clamp at the negotiation site, and the operator override
semantics (no-op when off, no-op when codec != ZSTD, overrides client
choice in both directions, out-of-range clamp at the override site).

### 10. End-to-end test for force-level + hot reload

`QwpEgressCompressionTest.testForceLevelEnforcedOnWireAndHotReloadable`
writes `force.level=3` to `server.conf`, opens a client requesting
`compression_level=1`, asserts `client.getNegotiatedZstdLevel() == 3`
— proves the server's forced level beats the client request on the
wire. Then it rewrites the file with `force.level=9`, calls
`engine.getConfigReloader().reload()` synchronously, opens a fresh
client, and asserts `getNegotiatedZstdLevel() == 9` — proves hot
reload picks up on the next new connection without restart.

Uses the direct reload API (vs. file-watcher +
`assertReloadConfigEventually`) for determinism, matching
`DynamicPropServerConfigurationTest`'s convention.

Depends on the paired client-side accessor for
`X-QWP-Content-Encoding` (java-questdb-client #26 section 13).
Without that the test would have no observability for the wire-level
outcome.

## Tradeoffs

- Handshake header: 101 response body is ~30 bytes longer per upgrade. Negligible.
- The advertised batch cap is the protocol constant / recv-buffer-derived value — there is no per-server config knob to override it. Adding one is a separate change.
- Bumping `questdb.client.version` to `1.2.2-SNAPSHOT` means contributors on this branch must either `cd java-questdb-client && mvn clean install -DskipTests` once, or pass `-P local-client` to every build, until 1.2.2 ships. Without that, a stale local cache surfaces as a Maven resolution error.
- IPv4 ingest only accepts `TYPE_IPV4` and `TYPE_INT` wire data into IPv4 columns. Senders that previously crammed IPv4 addresses into INT columns and expected the server to interpret them as IPv4 still need an explicit IPv4 column type.
- The `QwpColumnDef.validate()` refactor swaps a single comparison for a switch. Cost is one extra branch per column-def parse — irrelevant in the noise.
- The GEOHASH coverage additions widen the oracle fuzz row by five columns and the non-oracle fuzz by four columns. Per-row wire payload grows by roughly 30-50 bytes; existing per-frame size budgets accommodate this without auto-flush behaviour changes.
- Forward-compatible: a 1.2.1 client connecting to this server ignores the unknown handshake header and behaves identically to before. Older clients that don't know about IPv4 / GEOHASH ingest simply never send those wire types.
- The cursor sentinel-null fix is a visible behaviour change for any consumer that today reads the NULL bit-pattern as a non-null value. UUID and LONG256 conversion-to-string paths are the live ones affected; their stored values now correctly read as SQL NULL rather than the literal all-zeros string. Bulk WAL ingest is unaffected (cursor bypassed).
- The new e2e tests in `QwpSenderOversizeRowInBatchTest` rely on the matching client-side guards landing. Without them the assertions fail with "expected exception not thrown".
- `QwpTotalBufferedBytesBenchmark` lives in the `benchmarks` module and is not run by CI. The numbers it reports are dev-machine-specific.
- The `INT_NULL → IPv4_NULL` translation changes observable behaviour for any caller today sending `Integer.MIN_VALUE` via `intColumn` into an IPv4 column. Before: stored as `128.0.0.0` silently. After: surfaces as SQL NULL. Strictly an improvement but a behaviour change for code relying on the old quirk.
- The `BINARY → VARCHAR` rejection breaks any caller today writing `TYPE_BINARY` into a pre-existing VARCHAR column. The only "working" use case was "VARCHAR happens to be UTF-8-safe random bytes" which is by definition fragile; a clear coercion error is the right surface.
- The `IPv4 → STRING/VARCHAR` formatter is purely additive: a code path that throws mid-row today succeeds and produces dotted-quad output. No caller relied on the mid-row throw.

## Cross-repo

- Client side: questdb/java-questdb-client#26 — parses `X-QWP-Max-Batch-Size`, clamps batches accordingly, exposes `sender.ipv4Column(...)` and `sender.geoHashColumn(...)` on the public `Sender` interface, ships the matching client-side GEOHASH precision-persistence fix, adds the producer-side batch-cap guards pinned by the e2e tests in this PR, and fixes the closed-sender check precedence across every column setter that null-short-circuits before the close check.
- Build pin: questdb/questdb-enterprise#1010 — bumps the enterprise client default to 1.2.2-SNAPSHOT.

## Test plan

- mvn test -pl core -Dtest="WebSocketHandshakeTest" — covers the header emission (positive value emits, zero omits).
- mvn test -pl core -Dtest="QwpSenderE2ETest" — end-to-end tests pass against the SNAPSHOT client, including IPv4 auto-create + the three new GEOHASH tests (auto-create across storage classes, validation paths, between-batch precision lock).
- mvn test -pl core -Dtest="QwpUdpAllTypesTest" — adds `testIPv4Direct` (three concrete addresses + explicit-bitmap-null + explicit 0.0.0.0 -> both NULL on read) and grows `testAllTypesInOneRow` with an IPv4 column.
- mvn test -pl core -Dtest="QwpIngressOracleFuzzTest" — oracle fuzz now generates `ip` + five base GEOHASH columns and per-id IPv4 / GEOHASH extras in `injectExtra`, exercising SF replay and the server-side auto-add-column path under randomized load.
- mvn test -pl core -Dtest="QwpSenderFuzzTest" — non-oracle fuzz writes IPv4 and four GEOHASH columns per row across 28 test scenarios.
- mvn test -pl core -Dtest="QwpCursorBoundsCheckTest" — payload-corruption fuzz no longer trips on `TYPE_IPV4` and exercises IPv4 in the clean path.
- mvn test -pl core -Dtest="QwpColumnDefTest" — `TYPE_IPV4` is in the valid-types fixture; 0x17 (BINARY) and 0x00 stay rejected.
- mvn test -pl core -Dtest="QwpFixedWidthDecoderTest" — 34 tests including the three new sentinel-null cases (`testDecodeIPv4ColumnSentinelNullNoBitmap`, `testDecodeUuidColumnSentinelNullNoBitmap`, `testDecodeLong256ColumnSentinelNullNoBitmap`).
- mvn test -pl core -Dtest="QwpSenderOversizeRowInBatchTest" — 2 tests, both pass against the paired client commit.
- mvn test -pl core -Dtest="QwpSenderE2ETest" — full E2E pass, including the four new tests added in section 7: `testIntColumnIntoIPv4TranslatesNullSentinel`, `testBinarySourceRejectedByVarcharTarget`, `testCoercionToStringPreservesUuidAndLong256NullSentinels`, `testCoercionToStringAndVarcharFromIPv4`.
- mvn test -pl core -Dtest="QwpWebSocketUpgradeProcessorOnHeadersReadyTest" — 15 tests, including the now-pinned `testOnHeadersReadyAdvertisesEffectiveBatchSize`.
- mvn test -pl core -Dtest.include="%regex[.*[Qq]wp.*]" — full core QWP regression: 1285 tests pass.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -91,6 +91,22 @@ offending character, not the start of the expression.
 
 ## Git & PR Conventions
 
+- **PRs are squash-merged. Commit history on a PR branch is throwaway** — only
+  the squashed commit message that lands on `master` is preserved. Do not
+  spend effort tidying the branch's history: no soft resets to "commit all at
+  once", no rewording prior commits, no force pushes to clean up. Adding a
+  fix-up commit on top is always fine. The squash flow folds the lot at merge
+  time anyway.
+
+## Investigating failures
+
+- **Never dismiss a failure as "pre-existing", "flaky", "unrelated", or "a
+  known issue" without actually proving it.** That label is a hypothesis,
+  not a conclusion. Treat any red test, red CI job, or surprising log line
+  as a live bug to investigate until the evidence — git log, reproduction
+  on master, a real timing constraint, an upstream report — forces a
+  different conclusion. Only after that proof can the issue be set aside,
+  and the proof itself should be reported back so it can be verified.
 - **`java-questdb-client/` is a separate git repo** (a git submodule). Always
   `cd` into it and commit there independently. Never commit it from the parent
   repo as a submodule pointer update without also committing inside it first.
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -117,7 +117,7 @@
           -->
         <uberjar.name>benchmarks</uberjar.name>
 
-        <questdb.client.version>1.2.1</questdb.client.version>
+        <questdb.client.version>1.2.2</questdb.client.version>
     </properties>
 
     <build>
--- a/benchmarks/src/main/java/org/questdb/QwpTotalBufferedBytesBenchmark.java
+++ b/benchmarks/src/main/java/org/questdb/QwpTotalBufferedBytesBenchmark.java
@@ -1,0 +1,194 @@
+/*+*****************************************************************************
+ *     ___                  _   ____  ____
+ *    / _ \ _   _  ___  ___| |_|  _ \| __ )
+ *   | | | | | | |/ _ \/ __| __| | | |  _ \
+ *   | |_| | |_| |  __/\__ \ |_| |_| | |_) |
+ *    \__\_\\__,_|\___||___/\__|____/|____/
+ *
+ *  Copyright (c) 2014-2019 Appsicle
+ *  Copyright (c) 2019-2026 QuestDB
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ ******************************************************************************/
+
+package org.questdb;
+
+import io.questdb.client.cutlass.qwp.protocol.QwpConstants;
+import io.questdb.client.cutlass.qwp.protocol.QwpTableBuffer;
+import io.questdb.client.std.CharSequenceObjHashMap;
+import io.questdb.client.std.Misc;
+import io.questdb.client.std.ObjList;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Microbench for the per-row bytes-tracking step inside
+ * {@code QwpWebSocketSender.sendRow()}.
+ * <p>
+ * Today {@code sendRow()} sets {@code pendingBytes = totalBufferedBytes()} on
+ * every call. {@code totalBufferedBytes()} walks {@code tableBuffers.keys()},
+ * does a hashmap {@code get()} for each table, and sums per-column
+ * {@code getBufferedBytes()} values. We want a number on what that walk
+ * actually costs, since the two consumers of {@code pendingBytes} are guarded:
+ * <ul>
+ *   <li>the byte-based auto-flush check, gated by
+ *       {@code effectiveAutoFlushBytes > 0};</li>
+ *   <li>the server-cap pre-flight, only checked when
+ *       {@code pendingRowCount == 1 && serverMaxBatchSize > 0}.</li>
+ * </ul>
+ * When neither gate is active the walk is dead work.
+ * <p>
+ * Three variants:
+ * <ul>
+ *   <li>{@link #walkAllTables} -- current behaviour.</li>
+ *   <li>{@link #currentTableOnly} -- alternative: track the active table only
+ *       (requires upstream delta bookkeeping to be correct).</li>
+ *   <li>{@link #shortCircuit} -- proposed fix's cheap path: skip the walk when
+ *       no gate is active.</li>
+ * </ul>
+ * Params sweep {@code numTables} and {@code numColumns} so the O(N{*}K)
+ * scaling is visible.
+ * <p>
+ * Run from the benchmarks module:
+ * <pre>
+ *   mvn -pl benchmarks -P local-client -DskipTests package
+ *   java -jar benchmarks/target/benchmarks.jar QwpTotalBufferedBytesBenchmark
+ * </pre>
+ * The {@code local-client} profile rebuilds the client from source so the
+ * bench measures the {@code QwpTableBuffer} version sitting in this tree
+ * rather than the Maven-cached release.
+ */
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Warmup(iterations = 5, time = 1)
+@Measurement(iterations = 5, time = 1)
+@Fork(1)
+public class QwpTotalBufferedBytesBenchmark {
+
+    @Param({"1", "4", "16"})
+    private int numTables;
+
+    @Param({"1", "8"})
+    private int numColumns;
+
+    @Param({"100"})
+    private int rowsPrePopulated;
+
+    private QwpTableBuffer currentTableBuffer;
+    private CharSequenceObjHashMap<QwpTableBuffer> tableBuffers;
+
+    public static void main(String[] args) throws RunnerException {
+        Options opt = new OptionsBuilder()
+                .include(QwpTotalBufferedBytesBenchmark.class.getSimpleName())
+                .build();
+        new Runner(opt).run();
+    }
+
+    /**
+     * Alternative B: only the currently-active table's bytes are summed.
+     * The sender would need to maintain {@code pendingBytes} as a running
+     * delta updated by column setters and {@code nextRow()} for this to be
+     * correct across multi-table interleaving; the bench just measures the
+     * single-table {@code getBufferedBytes()} cost in isolation.
+     */
+    @Benchmark
+    public long currentTableOnly() {
+        return currentTableBuffer.getBufferedBytes();
+    }
+
+    @Setup(Level.Trial)
+    public void setUp() {
+        tableBuffers = new CharSequenceObjHashMap<>();
+        for (int t = 0; t < numTables; t++) {
+            String name = "tbl_" + t;
+            QwpTableBuffer tb = new QwpTableBuffer(name);
+            for (int r = 0; r < rowsPrePopulated; r++) {
+                for (int c = 0; c < numColumns; c++) {
+                    QwpTableBuffer.ColumnBuffer col =
+                            tb.getOrCreateColumn("c" + c, QwpConstants.TYPE_LONG, true);
+                    col.addLong((long) r * 31L + c);
+                }
+                tb.nextRow();
+            }
+            tableBuffers.put(name, tb);
+        }
+        currentTableBuffer = tableBuffers.get("tbl_0");
+    }
+
+    /**
+     * Proposed cheap path: when neither byte-based auto-flush nor server cap
+     * are active, {@code pendingBytes} is never read, so skip the walk.
+     * Returns a constant; JMH still has to call the method.
+     */
+    @Benchmark
+    public long shortCircuit() {
+        return 0L;
+    }
+
+    @TearDown(Level.Trial)
+    public void tearDown() {
+        if (tableBuffers != null) {
+            ObjList<CharSequence> keys = tableBuffers.keys();
+            for (int i = 0, n = keys.size(); i < n; i++) {
+                CharSequence k = keys.getQuick(i);
+                if (k != null) {
+                    Misc.free(tableBuffers.get(k));
+                }
+            }
+            tableBuffers.clear();
+            tableBuffers = null;
+        }
+        currentTableBuffer = null;
+    }
+
+    /**
+     * Baseline: replicates {@code QwpWebSocketSender.totalBufferedBytes()}
+     * exactly so the bench measures the same map walk + per-table column
+     * loop the sender pays on every row.
+     */
+    @Benchmark
+    public long walkAllTables() {
+        long total = 0;
+        ObjList<CharSequence> keys = tableBuffers.keys();
+        for (int i = 0, n = keys.size(); i < n; i++) {
+            CharSequence name = keys.getQuick(i);
+            if (name == null) {
+                continue;
+            }
+            QwpTableBuffer tb = tableBuffers.get(name);
+            if (tb != null) {
+                total += tb.getBufferedBytes();
+            }
+        }
+        return total;
+    }
+}
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -46,7 +46,7 @@
         -->
         <qdbr.rustflags>-D warnings</qdbr.rustflags>
 
-        <questdb.client.version>1.2.1</questdb.client.version>
+        <questdb.client.version>1.2.2-SNAPSHOT</questdb.client.version>
     </properties>
 
     <version>9.3.6-SNAPSHOT</version>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -45,7 +45,7 @@
         -->
         <qdbr.rustflags>-D warnings</qdbr.rustflags>
 
-        <questdb.client.version>1.2.2-SNAPSHOT</questdb.client.version>
+        <questdb.client.version>1.2.2</questdb.client.version>
     </properties>
 
     <version>9.4.1-SNAPSHOT</version>
--- a/core/src/main/java/io/questdb/DynamicPropServerConfiguration.java
+++ b/core/src/main/java/io/questdb/DynamicPropServerConfiguration.java
@@ -99,7 +99,8 @@ public class DynamicPropServerConfiguration implements ServerConfiguration, Conf
             PropertyKey.CAIRO_SQL_JIT_MAX_IN_LIST_SIZE_THRESHOLD,
             PropertyKey.CAIRO_PARQUET_EXPORT_COPY_REPORT_FREQUENCY_LINES,
             PropertyKey.CAIRO_SQL_COPY_EXPORT_ROOT,
-            PropertyKey.CAIRO_SQL_COPIER_CHUNKED
+            PropertyKey.CAIRO_SQL_COPIER_CHUNKED,
+            PropertyKey.QWP_EGRESS_COMPRESSION_FORCE_LEVEL
     ));
     private static final Function<String, ? extends ConfigPropertyKey> keyResolver = (k) -> {
         Optional<PropertyKey> prop = PropertyKey.getByString(k);
--- a/core/src/main/java/io/questdb/PropServerConfiguration.java
+++ b/core/src/main/java/io/questdb/PropServerConfiguration.java
@@ -415,6 +415,7 @@ public class PropServerConfiguration implements ServerConfiguration {
     private final PublicPassthroughConfiguration publicPassthroughConfiguration = new PropPublicPassthroughConfiguration();
     private final int queryCacheEventQueueCapacity;
     private final boolean queryWithinLatestByOptimisationEnabled;
+    private final int qwpEgressForcedZstdLevel;
     private final int qwpMaxRowsPerTable;
     private final int qwpMaxSchemasPerConnection;
     private final int qwpMaxTablesPerConnection;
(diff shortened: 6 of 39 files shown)


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 76 / 81 (93.83%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/QwpWalAppender.java|13|15|86.67%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpWebSocketHttpProcessor.java|21|24|87.50%|
|:large_blue_circle:|io/questdb/cutlass/qwp/protocol/QwpFixedWidthColumnCursor.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/protocol/QwpConstants.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpWebSocketUpgradeProcessor.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/protocol/QwpColumnDef.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalColumnarRowAppender.java|29|29|100.00%|

kafka1991:
## Code Review

Reviewed at full depth (correctness, concurrency, resources, performance, tests, standards, adversarial). Every changed symbol was traced to its callsites repo-wide. **No correctness bugs, resource leaks, concurrency hazards, or hot-path allocation regressions found in the production code.** Findings below are one metadata defect, one CI hazard, and minor style items.

### Critical
None.

### Moderate

**1. PR description section 2 contradicts the shipped code.** Section 2 says *"`BINARY` (0x17) stays rejected — no ingest decoder exists yet"*, but BINARY ingest is fully wired: `QwpColumnDef.validate()` admits `TYPE_BINARY`, `mapQwpTypeToQuestDB` maps it to `ColumnType.BINARY` (auto-create), `QwpTableBlockCursor` decodes it via the string cursor, and `WalColumnarRowAppender.putBinaryColumn` is a complete decoder. Section 7 even documents BINARY coercion guards. Release notes derive from the description — section 2 must be rewritten to state BINARY ingest is supported.

**2. Several committed tests are red against any client other than the paired `1.2.2-SNAPSHOT`.** `QwpSenderOversizeRowInBatchTest`, `QwpSenderFailoverBatchSizeTest.testReconnectToTighterCapRefreshesServerMaxBatchSize` (Javadoc: "red today"), `QwpEgressCompressionTest.testForceLevelEnforcedOnWireAndHotReloadable`, and many `QwpSenderE2ETest` IPv4/GEOHASH/BINARY tests hard-fail (no `Assume` guard) without the paired client. Acceptable only because the submodule pointer moves in the same PR — but any CI lane resolving the client from the Maven cache rather than building the submodule from source will fail. Please confirm CI builds the submodule from source.

**3. Non-null zero-length BINARY ingests as SQL NULL.** `putBinaryColumn` -> `MemoryPARWImpl.putBin(from, len)` writes `NULL_LEN` whenever `len <= 0`. A client sending `binaryColumn("b", new byte[0])` stores SQL NULL. This is consistent with QuestDB's existing BINARY semantics (the row-based `WalWriter.putBin` does the identical call), so it is not a defect introduced here and is not blocking — but it is new behavior for the QWP wire path and is currently only noted inside a test-oracle Javadoc. Worth a line in `docs/qwp/` or an explicit ingest-time rejection.

### Minor

1. **Member ordering** — `putIntToIPv4Column` is placed between `putFixedToVarcharColumn` and `putFloatToDecimalColumn` in both `WalColumnarRowAppender.java` and the `ColumnarRowAppender` interface. It should sort after all `putFloat*`/`putGeo*` methods, immediately before `putIntegerToNumericColumn`. (`putBinaryColumn` is correctly placed.)
2. **Banner comment** — `QwpEgressCompressionNegotiatorTest.java` adds `// --- resolveEffectiveZstdLevel (operator force-level override) ---`. CLAUDE.md forbids `// ---` banner headings, including in test code. Remove it.
3. **Numeric literal** — `QwpWebSocketUpgradeProcessorOnHeadersReadyTest` uses bare `131058`; should be `131_058` per CLAUDE.md.
4. **Scope creep** — the PR modifies `CLAUDE.md` (squash-merge bullet, "Investigating failures" section) unrelated to QWP; should be a separate PR.
5. **`testBinarySourceRejectedBy{Varchar,Char,String,Symbol}Target`** swallow the flush exception and rely only on `SELECT count() = 0`. Asserting the rejection message would make them true regression tests.

### Checked and dismissed (false positives)

- IPv4 value `0` with an explicit bitmap rendering `"0.0.0.0"` via `formatFixedOtherValue` — a row marked non-null in the wire bitmap is genuinely non-null; rendering the value is correct. IPv4-column round-trip losing `0.0.0.0` to NULL is QuestDB's inherent `IPv4_NULL == 0` limitation, not a defect.
- `VERSION_BYTES` / `ZSTD_LEVEL_BYTES` array indexing — version and zstd level are clamped before either table is indexed; the null slot 0 is never reached.
- `parseLevel` integer overflow — an overflowed negative value is caught by `negotiate()`'s `< MIN` clamp.
- Resource leak in `putBinaryColumn` / `putIntToIPv4Column` — `MemoryMA` handles are writer-owned; error paths roll back via `cancelColumnarWrite()`.
- `getQwpEgressForcedZstdLevel()` reload race — config swaps wholesale behind `CairoConfigurationWrapper`'s `AtomicReference`; each handshake sees a coherent value.

### Verdict

**Request changes — minor only.** No correctness, concurrency, or resource defects. The cross-context audit confirmed all `responseSize`/`writeResponse` callers updated to the `byte[]` signature and all `CairoConfiguration` implementors inherit the new `default` method. The zero-GC handshake-caching and config-reload-safety claims both verified accurate.

Blocking before merge: fix the PR description (Moderate 1), confirm CI builds the client submodule from source (Moderate 2). Everything else is cosmetic.

One tradeoff to weigh: this PR bundles 10 loosely-related changes. It is well-tested and internally consistent, but the bundling makes the description hard to keep accurate (Moderate 1 is direct evidence) and couples the QWP feature work to an unreleased client.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 110 / 116 (94.83%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/PropServerConfiguration.java|3|5|60.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpWebSocketUpgradeProcessor.java|6|7|85.71%|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/QwpWalAppender.java|19|21|90.48%|
|:large_blue_circle:|io/questdb/cutlass/qwp/protocol/QwpFixedWidthColumnCursor.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressUpgradeProcessor.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpWebSocketHttpProcessor.java|23|23|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/protocol/QwpConstants.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressCompressionNegotiator.java|15|15|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/protocol/QwpColumnDef.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalColumnarRowAppender.java|29|29|100.00%|

PR #7112, opened 2026-05-18
perf(core): skip unchanged buckets on O3 incremental refresh

nwoolmer:
## Summary

- When an out-of-order historical write into a mat view's base table lands
  far behind the current commit position, the incremental refresh on
  master scans every non-empty bucket between the O3 timestamp and "now",
  including pre-existing ones that no WAL transaction touched. The cursor
  filter is set to the full `[minTs, maxTs]` envelope and the iterator's
  step-group spans the gap. For a 512-symbol mat view with a 12-hour O3
  lag the refresh takes 160 ms instead of the ~2 ms the actual work
  requires.

- This PR adds a cost-aware clustering pass in `MatViewRefreshJob`. Two
  rolling EMAs on `MatViewState` (avg commit latency in ns, avg scan
  latency per timestamp-unit) drive a gap-width threshold below which two
  adjacent cached intervals get merged. `computePerClusterSteps` then
  produces one iterator step per cluster (`min(naturalStep,
  clusterWidthBuckets)`) so a step-group never straddles two clusters --
  the existing gap-skip excises the gap cheaply, and each cluster gets
  the step that best fits its own width.

- The iterator (`SampleByIntervalIterator` + `FixedOffsetIntervalIterator`
  + `TimeZoneIntervalIterator`) carries a parallel `LongList
  stepPerInterval` and swaps the active step on cluster transition. A new
  `snapToCluster()` hook aligns the iterator's hi to the next cluster's
  bucket-aligned lo, so a small per-cluster step doesn't walk the gap
  one bucket at a time.

- `materialized_views()` gains three columns (`refresh_avg_commit_nanos`,
  `refresh_avg_scan_nanos_per_ts_unit`, `refresh_gap_threshold_ts_units`)
  for operator visibility into the cost model. `REFRESH MATERIALIZED VIEW
  <name> STATS` resets the EMAs to cold-start when the workload shape
  changes.

- One new configuration key: `cairo.mat.view.refresh.max.clusters`
  (default 32) caps the number of clusters per refresh to prevent
  pathological many-disjoint-intervals workloads from emitting hundreds
  of tiny commits.

## Benchmark numbers

`MatViewO3RefreshBenchmark`, 512 symbols, 1440-minute seed.

**Original sweep** (single-tick recent writes, `wideBuckets=1`):

| rowsPerQuery | o3LagMin | Before | After | Delta |
|---|---|---|---|---|
| 1,000,000 | 0 | 0.45 ms | 0.40 ms | baseline |
| 1,000,000 | 30 | 6.4 ms | 2.2 ms | 2.9x faster |
| 1,000,000 | 120 | 22 ms | 1.8 ms | 12x faster |
| 1,000,000 | 720 | 160 ms | 1.7 ms | 94x faster |
| 1,000 | 720 | 2.7 ms | 1.6 ms | similar |

**Mixed-width sweep** (`rowsPerQuery=1M, o3LagMin=720`, varying the
number of consecutive recent ticks per iteration). This exposes the
scenario where a narrow far-back O3 cluster co-exists with a wide
recent cluster:

| wideBuckets | Master | Global step-cap | Per-cluster step |
|---|---|---|---|
| 1 | 160 ms | 1.79 ms | 1.73 ms |
| 10 | 160 ms | 1.76 ms | 1.58 ms |
| 100 | 160 ms | 1.03 ms | 0.36 ms |
| 1000 | 160 ms | 5.33 ms | 0.76 ms |

The global step-cap variant pins the iterator step to the narrowest
cluster's width, so a 1000-bucket wide cluster paired with a point
narrow cluster pays 1000 `factory.getCursor()` calls. The per-cluster
step recovers that gap at K=1000 (-86%) and K=100 (-65%) while leaving
the K=1 case unchanged.

## Tradeoffs

- The previously warm fast path (no O3, default config) moves from
  0.45 ms to 0.40 ms; the difference is at the level of bench noise.
- Two-cluster commit overhead is unchanged (one REPLACE_RANGE per
  cluster); the K=1 mixed case is ~1.7 ms regardless of step strategy
  because it is two point-interval commits.
- `RefreshContext` gains four fields (`naturalStep`, `stepPerInterval`,
  `refreshIntervals`, `approxBucketSize`) so the OOM / too-many-rows
  retry paths can recompute the per-cluster step list without re-running
  `findRefreshIntervals`.

## Test plan

- 404 mat view tests pass (390 existing + 14 new).
- Unit tests for the cost-model helpers:
  - `MatViewRefreshJobClusterTest` -- 23 tests covering empty/single
    intervals, adjacent merges, distant splits, maxClusters cap,
    overflow handling at `Long.MAX_VALUE`, 200-point-interval cap
    pressure, point intervals, malformed inputs, and the new
    `computePerClusterSteps` per-cluster cap semantics
  - `MatViewStateEmaTest` -- 16 tests covering cold-start, EMA
    convergence, zero/negative samples, outlier capping, pathological
    cold-start scan rates, `STATS` reset, overflow at `Long.MAX_VALUE`
- Iterator transition tests:
  - `FixedOffsetIntervalIteratorTest.testPerClusterStepSnapsAcrossGap` --
    verifies the snap on cluster transition produces a fresh step-group
    anchored at the cluster boundary
  - `FixedOffsetIntervalIteratorTest.testPerClusterStepThreeClusters` --
    multi-cluster transition with mixed step widths
- Integration tests in `MatViewTest`:
  - `testRefreshIntervalsO3SplitsWideEnvelope` -- regression test for
    the original wasted-recompute scenario
  - `testRefreshIntervalsO3MergesNarrowGap` -- auto-tune merge branch
  - `testRefreshIntervalsO3SingleIntervalUnaffected` -- guards against
    step-cap over-shrinking on a single dirty interval
  - `testRefreshIntervalsO3PeriodMatView` -- period mat view with
    clustering active
  - `testRefreshMaterializedViewStatsResetsEma` -- SQL-level smoke test
    for the new `STATS` form
- All existing `MatViewTest`, `MatViewStateTest`, `MatViewTelemetryTest`,
  `MatViewReloadOnRestartTest`, `CreateMatViewTest`, `MatViewFuzzTest`
  tests pass unchanged.


questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=234337)
**Tested Commit:** `43c1f5d1e9e31d6f5250d57e27e2c3f4e5a83b01`

core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
```diff
@@ -147,6 +152,127 @@ private static long approxStepDuration(long step, long approxBucketSize) {
...
+     * the given intervals list. Returns {@code step} unchanged if the list is
+     * null, has fewer than one entry, or the bucket size is non-positive.
+     * <p>
+     * Goes together with {@link #clusterIntervals}: each list entry is one
+     * cluster after auto-tuning, and we want a step-group to fit inside one
+     * cluster so the iterator's gap-skip can excise the gaps between them
+     * without dragging unchanged buckets into the cursor's scan range.
+     */
+    // kept public for testing
+    public static long capStepByNarrowestInterval(@Nullable LongList intervals, long approxBucketSize, long step) {
```
kafka1991: Does `capStepByNarrowestInterval` penalize mixed-width workloads by capping the iterator step globally?

Consider the scenario: a narrow far-back O3 interval plus a wide recent interval, kept as two clusters. Doesn't the step then get pinned to the narrow cluster's width (possibly one bucket), so the wide cluster is walked in many tiny step-groups, one `factory.getCursor()` call plus setup per step-group? Wouldn't any single narrow interval (in the extreme, a point interval) pin the step for the entire refresh?

nwoolmer: It can, let me assess impact and see if fix is justified

nwoolmer: You're right -- that's a real penalty in mixed-width workloads. I extended `MatViewO3RefreshBenchmark` with a `WIDE_RECENT_BUCKETS_SWEEP` dimension to measure the case you describe (narrow far-back O3 + wide recent cluster):

| rowsPerQuery=1M, o3LagMin=720 | wideBuckets=1 | 10 | 100 | **1000** |
|--|--|--|--|--|
| Before (global cap) | 1.79ms | 1.76ms | 1.03ms | **5.33ms** |

That `5.33ms` at K=1000 is exactly the failure mode you described -- the 1000-bucket wide cluster paid 1000 `factory.getCursor()` calls because the narrowest cluster's width (1 bucket) capped the iterator step globally.

Pushed a fix replacing `capStepByNarrowestInterval` with `computePerClusterSteps`. The iterator's base class now holds a parallel `LongList stepPerInterval` and swaps `step` on cluster transition in `next()`. A new `snapToCluster()` hook aligns the iterator's hi to the cluster boundary on transition so a small per-cluster step doesn't walk the gap one bucket at a time.

After-numbers on the same matrix:

| rowsPerQuery=1M, o3LagMin=720 | wideBuckets=1 | 10 | 100 | **1000** |
|--|--|--|--|--|
| After (per-cluster) | 1.73ms | 1.58ms | 0.36ms | **0.76ms** |
| Delta | flat | -10% | -65% | **-86%** |

K=1 is unchanged (no cap to remove there). Wide-only baselines at `o3Lag=0` are unchanged within noise across all K, so the change is targeted at the mixed-width path. The benchmark extension is in the same commit so the matrix is reproducible.

nwoolmer: ^ robot

bluestreak01 (dismissed):

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 268 / 275 (97.45%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewState.java|51|57|89.47%|
|:large_blue_circle:|io/questdb/cairo/mv/SampleByIntervalIterator.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/FixedOffsetIntervalIterator.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|13|13|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlKeywords.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/TimeZoneIntervalIterator.java|47|47|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/MatViewsFunctionFactory.java|16|16|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|110|110|100.00%|

PR #7115, opened 2026-05-19
fix(sql): prevent internal errors in symbol hash joins

jerrinot:
Queries joining on SYMBOL plus UUID or DECIMAL keys could fail with an UnsupportedOperationException instead of returning results. Delegate the missing record accessors so these joins work correctly.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/sql/DelegatingRecord.java
+++ b/core/src/main/java/io/questdb/cairo/sql/DelegatingRecord.java
@@ -26,6 +26,8 @@ package io.questdb.cairo.sql;
 
 import io.questdb.cairo.arr.ArrayView;
 import io.questdb.std.BinarySequence;
+import io.questdb.std.Decimal128;
+import io.questdb.std.Decimal256;
 import io.questdb.std.Interval;
 import io.questdb.std.Long256;
 import io.questdb.std.str.CharSink;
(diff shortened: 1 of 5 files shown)


puzpuzpuz (dismissed):
A few method ordering violations, but I don't want to list them (worth running reformat in IDEA though). Other than that, all minor comments (treat as nits):

m1. Title scope is narrower than the fix. Title: `fix(sql): prevent internal errors in symbol hash joins`, but the same bug class affects ASOF joins (the PR itself adds ASOF tests), LT joins (LtJoinLightRecordCursorFactory), and HORIZON joins (HorizonJoinRecordCursorFactory, MultiHorizonJoinRecordCursorFactory, AsyncHorizonJoinRecordCursorFactory, AsyncMultiHorizonJoinRecordCursorFactory — all wrap their key record in SymbolTranslatingRecord). Consider `fix(sql): fix internal errors in keyed joins with UUID or DECIMAL columns` or similar.

m2. No NULL-key tests. The added tests use only non-NULL UUID/DECIMAL values. The interesting SymbolTranslatingRecord.translate() branch is VALUE_IS_NULL, and RecordSink paths must handle NULL Long128/Decimal* keys distinctly from real values. Add at least one row per type with a NULL key on each side, with the expected JOIN/LEFT JOIN behavior (NULL never matches NULL by default).

m3. getArrayDouble1d2d, getLongIPv4, getVarchar(int, Utf16Sink) are still not delegated. Each has a default implementation that ends up dispatching through getArray / getIPv4 / getVarcharA — those are delegated, so behavior is functionally correct. But this silently bypasses optimized overrides on PageFrameMemoryRecord (it overrides getArrayDouble1d2d to skip the ArrayView allocation, and overrides getLongIPv4). Whenever SymbolTranslatingRecord wraps a PageFrameMemoryRecord, the optimized fast paths are lost. Not blocking for the correctness fix this PR is making, but worth a follow-up.

jerrinot:
@puzpuzpuz: thanks! everything should be addressed now.

puzpuzpuz (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 10 / 16 (62.50%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/sql/DelegatingRecord.java|10|16|62.50%|

PR #7121, opened 2026-05-20
chore(core): storage policy fixes

glasstiger:
Storage policy fixes, required by https://github.com/questdb/questdb-enterprise/pull/1019.

## Changes

- **Drain the writer's async command queue after each WAL apply batch** (`ApplyWal2TableJob`): the WAL apply loop holds the writer across a batch of transactions and never ticks its command queue, so async commands published while the writer is busy (e.g. storage policy parquet-commit / drop-local / squash on a continuously-ingested table) could sit unprocessed until the writer is returned to the pool. Once a batch is applied and its sequencer txn finalized, the apply job now drains the queue via `writer.tick(false, drainDeadline)`. The drain shares the per-table apply time quota: it gets at most half of whatever quota is left, and it stops after processing a command once the deadline has passed (so it always makes forward progress), leaving the rest queued for the next tick. That keeps a backlog of expensive commands from monopolizing an apply worker shared with other WAL tables. Draining is a side activity — the WAL transactions are already durably committed, so a drain failure does not fail the apply or suspend the table: each command's own failure is reported on its correlation channel, and only a rare infrastructure error escaping the queue loop marks the writer distressed (forcing the pool to recreate it). `commit()` itself does **not** drain; non-WAL writers continue to be drained by their ingestion `tick()` and on pool return.
- **Reject structural async commands on a WAL table's queue** (`TableWriter.publishAsyncWriterCommand`): a WAL table's structural ALTERs route through the sequencer, so only non-structural commands (e.g. storage policy) may legitimately reach its async command queue. Now that the apply job drains that queue, a structural command slipping in would mutate the writer's metadata out of band with the sequencer and corrupt subsequent WAL transactions. Such a command is now rejected at publish time, with a redundant consumer-side assertion as a backstop. `startAsync()` moves from `WriterPool` into `publishAsyncWriterCommand` so it runs only after this invariant has been checked.
- **Bound the command-queue drain by a deadline** (`TableWriter.tick` / `processCommandQueue`): new `tick(boolean, long deadlineMicros)` and `processCommandQueue(boolean, long)` overloads check the wall-clock deadline after each command and stop the drain when it is reached. Passing `Long.MAX_VALUE` drains the whole queue, so existing callers' behavior is preserved.
- **Increase writer async command queue capacity** (`PropServerConfiguration`): bump the `cairo.writer.command.queue.capacity` default from 32 to 256 so the writer's async queue can hold more messages during bursts of storage policy tasks. The queue is allocated per writer (`capacity * slotSize`), so this raises each open writer's async-queue reservation accordingly.
- **Reject negative `cairo.wal.max.lag.size`** (`PropServerConfiguration`): add a `getLongSize(..., minValue)` overload and use it for `walMaxLagSize` with a minimum of 0, so a negative configured value fails fast at startup with a `ServerConfigurationException` instead of producing a nonsensical lag bound. The default (75 MB) is unchanged.
- **SHOW PARTITIONS reports `hasParquetGenerated` for parquet partitions** (`ShowPartitionsRecordCursorFactory`): report `hasParquetGenerated = true` whenever a partition is in parquet format (`isParquet`), even if the raw parquet-generated bit is cleared. This covers pre-storage-policy parquet tables and divergent states where a partition is parquet format but its generated bit was reset. A parquet partition implies a parquet file was generated for it. The column is display-only; internal logic still reads the raw `txWriter` bit directly.
- **Clarify the TTL granularity error message** (`PartitionBy.validateTtlGranularity`): reword to `TTL value must be an integer multiple of the partition size (its time interval)`.
- **Replace a hand-rolled clamp with `Math.clamp`** (`TableWriter.getWalMaxLagRows`): rewrite `Math.min(Math.max(0, ...), getWalMaxLagSize())` as `Math.clamp(..., 0, getWalMaxLagSize())`. Behavior is preserved (`getWalMaxLagSize()` is non-negative, so the clamp bounds are always valid).
- **Remove a tautological assertion** (`TableWriter.produceNativeFromParquet`): drop `assert parquetMetadata.getRowGroupSize(rowGroupIndex) <= Integer.MAX_VALUE`. `getRowGroupSize` returns an `int`, so the assertion was always true and never had any effect.

## Tests

- Async command-queue drain (`TableWriterTest`): `testNonWalCommitDoesNotDrainAsyncCommandQueue` (a plain commit must not drain — a queued command stays unapplied until an explicit `tick()`), `testTickStopsAtDeadlineAndLeavesRemainingCommandsQueued` (the deadline bound applies exactly one command then leaves the rest queued), `testWalApplyTickDrainsAsyncCommandQueue` (hand-replicated commit-then-tick on a held WAL writer drains a queued command), and `testWalApplyTickIsolatesFailingAsyncCommand` (a failing drained command does not abort the WAL commit or distress the writer).
- Structural-command rejection: `testWalStructuralAsyncCommandRejectedAtPublish` asserts a structural ALTER published to a WAL table's async queue is rejected at publish time.
- End-to-end drain via the real `ApplyWal2TableJob` (`WalApplyDrainTest`, `WalApplyDrainFailureTest`): driven through `drainWalQueue()`, these distinguish the apply-job drain (`tick(false, <finite deadline>)`) from the pool-return tick (`tick(true, MAX_VALUE)`) so the apply-job call-site is genuinely exercised, and verify that a drain failure marks the writer distressed without suspending the table.
- Config validation: `testInvalidWalMaxLagSizeNegative` rejects a negative `cairo.wal.max.lag.size` at startup; `testWalMaxLagSizeZeroAccepted` confirms 0 is accepted and round-trips (guards against a `>` vs `>=` regression).
- `testShowPartitionsParquetFormatWithoutGeneratedFlag` exercises a partition in parquet format with the generated bit cleared, asserting `hasParquetGenerated` is still reported as true.
- Updated `testAllDefaults` and `testShowParameters` for the new `cairo.writer.command.queue.capacity` default (256), and updated the TTL granularity message expectations across `TtlTest`, `CreateMatViewTest`, and `MatViewTest`.


ideoma (dismissed):

glasstiger:
@coderabbitai review

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=235684)
**Tested Commit:** `1f1730c5c08309ef5668bb93392e774fb7b232e7`

bluestreak01:
@glasstiger — flagging two issues from a code review of this PR.

### WAL+structural-command invariant is enforced only by `assert` (silent corruption risk in production builds)

`core/src/main/java/io/questdb/cairo/TableWriter.java:8168`

```java
assert !tableToken.isWal() || !asyncWriterCommand.isStructural()
        : "structural command must not reach a WAL table's async command queue";
```

Production JVMs run without `-ea`. If any caller publishes a structural `AlterOperation` (e.g. `ADD_COLUMN`, `RENAME_COLUMN`, `CHANGE_COLUMN_TYPE`, `SET_DEDUP_ENABLE/DISABLE`) to a WAL writer's async queue, the new `writer.tick()` drain in `ApplyWal2TableJob` will dispatch it via `processAsyncWriterCommand` → `AlterOperation.apply` and mutate the writer's metadata out-of-band with the sequencer. `AlterOperation.doApply` does not gate `ADD_COLUMN`/`ADD_INDEX`/`DROP_INDEX`/`SET_DEDUP_*` on `contextAllowsAnyStructureChanges`, so the writer's metadata diverges from the sequencer's; the next WAL txn reads stale metadata. The comment immediately above the assert ("would otherwise be applied out of band") acknowledges the exact failure mode.

No current call path violates the invariant — but a single future regression (e.g. a new caller of `engine.getWriterOrPublishCommand(walToken, structuralCmd)`, or the existing `OperationDispatcher` fallback when `walWriterPool` throws `EntryUnavailableException(POOL_SIZE)`) corrupts production silently. The assert documents an invariant the surrounding code does not actually enforce.

**Suggested fix:** convert into a production-visible guard:

```java
if (tableToken.isWal() && asyncWriterCommand.isStructural()) {
    LOG.critical().$("structural command on WAL table's async queue [type=").$(cmdType)
        .$(", table=").$(tableToken).I$();
    errorCode = STRUCTURE_CHANGE_NOT_ALLOWED;
    errorMsg = "structural ALTER on WAL table must route through the sequencer";
    // skip apply, fall through to publishTableWriterEvent
} else {
    affectedRowsCount = asyncWriterCommand.apply(this, contextAllowsAnyStructureChanges);
}
```

Either that, or reject at publish time in `WriterPool.addCommandToWriterQueue`.

bluestreak01:
@glasstiger — second finding from the same review.

### `writer.tick()` drain runs outside `tableTimeQuotaMicros` and is unbounded

`core/src/main/java/io/questdb/cairo/wal/ApplyWal2TableJob.java:531-536`

```java
try {
    writer.tick();
} catch (Throwable th) {
    LOG.error().$("error draining async command queue after WAL apply [table=")
            .$(writer.getTableToken()).$(", error=").$(th).I$();
}
```

Two concerns about this drain:

**1. No time bound, no fairness budget.** `applyOutstandingWalTransactions` budgets per-table work via `tableTimeQuotaMicros` (default 1000ms) for fairness across tables sharing an apply worker. The new tick runs after `commitSeqTxn`, outside the quota. Combined with this PR's storage policy default changes on the enterprise side (worker count 2 → 4, check interval 15 min → 5 min) and the writer queue capacity bump (32 → 256), the queue can accumulate many commands. Per-command cost is not bounded:

- `StoragePolicyWriterCommand` DROP_LOCAL → `removePartition` (mmap I/O on partition files);
- `TASK_SQUASH` → `preparePartitionForParquetConversion` → `squashPartitionForce` (copies/merges partition data, easily multiple seconds on wide tables);
- `TASK_PARQUET_COMMIT` → `switchNativePartitionWithParquet` (file ops).

Worst case: a wide table with hundreds of historical partitions, queue near its 256 cap, every command a squash — one tick occupies the apply worker for tens of seconds. Other WAL tables sharing that worker wait that whole time. The PR description's framing as "a single lock-free sequence read" is true only when the queue is empty.

**Suggested fix:** add an inner deadline (e.g. half of the remaining `tableTimeQuotaMicros`, or a small fixed budget) and break out of `processCommandQueue` when it elapses. Remaining commands stay queued for the next tick.

**2. `Throwable` from tick() is swallowed without marking the writer distressed.** The comment claims "Genuine writer corruption still surfaces via the distressed flag," but that's a property of each individual command's `apply` path, not of `tick()` itself. Many failure paths inside, e.g., `StoragePolicyWriterCommand.apply0` after a partial `removePartition` mutation do not call `markDistressed`. The outer for-loop then continues using a writer in inconsistent state until something else trips the distressed check or the writer is returned to the pool.

**Suggested fix:** mark the writer distressed in the catch block so the pool recreates it on next acquisition. Also narrow `catch (Throwable)` to `catch (Exception)` so genuine `Error`s (OOM, StackOverflow, LinkageError) propagate to the apply-loop's existing mem-pressure handling instead of being downgraded to a log line.

```java
try {
    writer.tick();
} catch (Exception ex) {
    LOG.error().$("error draining async command queue after WAL apply [table=")
            .$(writer.getTableToken()).$(", error=").$(ex).I$();
    writer.markDistressed();
}
```

Also: the catch block is the cold path here, and the PR coverage bot shows `ApplyWal2TableJob.java` at 40% (2/5 covered) — please add a test that forces an exception out of `tick()` via the apply job (not via a direct `writer.tick()` call) and asserts the table is not suspended.

glasstiger:
> ### WAL+structural-command invariant is enforced only by `assert` (silent corruption risk in production builds)
> `core/src/main/java/io/questdb/cairo/TableWriter.java:8168`
> 
> ```java
> assert !tableToken.isWal() || !asyncWriterCommand.isStructural()
>         : "structural command must not reach a WAL table's async command queue";
> ```
> 
> Production JVMs run without `-ea`. If any caller publishes a structural `AlterOperation` (e.g. `ADD_COLUMN`, `RENAME_COLUMN`, `CHANGE_COLUMN_TYPE`, `SET_DEDUP_ENABLE/DISABLE`) to a WAL writer's async queue, the new `writer.tick()` drain in `ApplyWal2TableJob` will dispatch it via `processAsyncWriterCommand` → `AlterOperation.apply` and mutate the writer's metadata out-of-band with the sequencer. `AlterOperation.doApply` does not gate `ADD_COLUMN`/`ADD_INDEX`/`DROP_INDEX`/`SET_DEDUP_*` on `contextAllowsAnyStructureChanges`, so the writer's metadata diverges from the sequencer's; the next WAL txn reads stale metadata. The comment immediately above the assert ("would otherwise be applied out of band") acknowledges the exact failure mode.
> 
> No current call path violates the invariant — but a single future regression (e.g. a new caller of `engine.getWriterOrPublishCommand(walToken, structuralCmd)`, or the existing `OperationDispatcher` fallback when `walWriterPool` throws `EntryUnavailableException(POOL_SIZE)`) corrupts production silently. The assert documents an invariant the surrounding code does not actually enforce.
> 
> **Suggested fix:** convert into a production-visible guard:
> 
> ```java
> if (tableToken.isWal() && asyncWriterCommand.isStructural()) {
>     LOG.critical().$("structural command on WAL table's async queue [type=").$(cmdType)
>         .$(", table=").$(tableToken).I$();
>     errorCode = STRUCTURE_CHANGE_NOT_ALLOWED;
>     errorMsg = "structural ALTER on WAL table must route through the sequencer";
>     // skip apply, fall through to publishTableWriterEvent
> } else {
>     affectedRowsCount = asyncWriterCommand.apply(this, contextAllowsAnyStructureChanges);
> }
> ```
> 
> Either that, or reject at publish time in `WriterPool.addCommandToWriterQueue`.

Claude is being stupid here.
I explained that nothing goes on the async queue for WAL tables, apart from the new storage policy tasks, so eventually we agreed on this assert.
Now it wants to harden it further.

If we want to enforce this, then I would rather throw an exception when the command is put on the async queue of the WAL table, instead of logging an error when it gets processed.

Addressed in https://github.com/questdb/questdb/pull/7121/commits/93417433a07fc93fed1c91f339c49a8466c0c531

glasstiger:
> ### `writer.tick()` drain runs outside `tableTimeQuotaMicros` and is unbounded
> `core/src/main/java/io/questdb/cairo/wal/ApplyWal2TableJob.java:531-536`
> 
> ```java
> try {
>     writer.tick();
> } catch (Throwable th) {
>     LOG.error().$("error draining async command queue after WAL apply [table=")
>             .$(writer.getTableToken()).$(", error=").$(th).I$();
> }
> ```
> 
> Two concerns about this drain:
> 
> **1. No time bound, no fairness budget.** `applyOutstandingWalTransactions` budgets per-table work via `tableTimeQuotaMicros` (default 1000ms) for fairness across tables sharing an apply worker. The new tick runs after `commitSeqTxn`, outside the quota. Combined with this PR's storage policy default changes on the enterprise side (worker count 2 → 4, check interval 15 min → 5 min) and the writer queue capacity bump (32 → 256), the queue can accumulate many commands. Per-command cost is not bounded:
> 
> * `StoragePolicyWriterCommand` DROP_LOCAL → `removePartition` (mmap I/O on partition files);
> * `TASK_SQUASH` → `preparePartitionForParquetConversion` → `squashPartitionForce` (copies/merges partition data, easily multiple seconds on wide tables);
> * `TASK_PARQUET_COMMIT` → `switchNativePartitionWithParquet` (file ops).
> 
> Worst case: a wide table with hundreds of historical partitions, queue near its 256 cap, every command a squash — one tick occupies the apply worker for tens of seconds. Other WAL tables sharing that worker wait that whole time. The PR description's framing as "a single lock-free sequence read" is true only when the queue is empty.
> 
> **Suggested fix:** add an inner deadline (e.g. half of the remaining `tableTimeQuotaMicros`, or a small fixed budget) and break out of `processCommandQueue` when it elapses. Remaining commands stay queued for the next tick.
> 
> **2. `Throwable` from tick() is swallowed without marking the writer distressed.** The comment claims "Genuine writer corruption still surfaces via the distressed flag," but that's a property of each individual command's `apply` path, not of `tick()` itself. Many failure paths inside, e.g., `StoragePolicyWriterCommand.apply0` after a partial `removePartition` mutation do not call `markDistressed`. The outer for-loop then continues using a writer in inconsistent state until something else trips the distressed check or the writer is returned to the pool.
> 
> **Suggested fix:** mark the writer distressed in the catch block so the pool recreates it on next acquisition. Also narrow `catch (Throwable)` to `catch (Exception)` so genuine `Error`s (OOM, StackOverflow, LinkageError) propagate to the apply-loop's existing mem-pressure handling instead of being downgraded to a log line.
> 
> ```java
> try {
>     writer.tick();
> } catch (Exception ex) {
>     LOG.error().$("error draining async command queue after WAL apply [table=")
>             .$(writer.getTableToken()).$(", error=").$(ex).I$();
>     writer.markDistressed();
> }
> ```
> 
> Also: the catch block is the cold path here, and the PR coverage bot shows `ApplyWal2TableJob.java` at 40% (2/5 covered) — please add a test that forces an exception out of `tick()` via the apply job (not via a direct `writer.tick()` call) and asserts the table is not suspended.

This makes sense, we should limit the time spent with processing storage policy tasks.

Addressed in https://github.com/questdb/questdb/pull/7121/changes/768306bce9e67f91c151f7a854d2dc7bfc979249

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 31 / 31 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/PartitionBy.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowPartitionsRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|15|15|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/ApplyWal2TableJob.java|8|8|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=235796)
**Tested Commit:** `768306bce9e67f91c151f7a854d2dc7bfc979249`

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 31 / 31 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/PartitionBy.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowPartitionsRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|15|15|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/ApplyWal2TableJob.java|8|8|100.00%|

bluestreak01 (approved):

PR #7124, opened 2026-05-20
fix(core): do not execute statements during SQL validation

glasstiger:
## Summary

The SQL validation endpoint (`/api/v1/sql/validate`) puts the execution context into validation-only mode and compiles a statement purely to check its syntax — it must not enforce authorization, execute any side effects, or pollute the server log. Previously validation-only compilation still enforced permissions, ran some statements inline during compile, and emitted a query-progress log line (plus an error metric) for every validated statement.

This PR makes validation-only compilation side-effect free:

- **Inline execution is skipped.** The previous coarse short-circuits at the top of `compileAlter`, `compileTruncate`, `compileVacuum`, and `compileReindex` (which returned early during validation) are replaced with fine-grained `isValidationOnly()` guards around the individual side-effecting calls. Validation now runs the full compile path — resolving the target object and checking semantics — while skipping the mutation itself. The statements whose inline side effect is now guarded: REINDEX, TRUNCATE, VACUUM, RESUME/SUSPEND WAL, ALTER ... SET TYPE (convert file), ALTER VIEW / CREATE OR REPLACE VIEW (replace view definition), and COMPILE VIEW (enqueue compile, including the error path). Removing the `compileAlter` short-circuit also lets the Enterprise storage-policy guards in the paired PR take effect on the same full compile path.
- **Authorization is bypassed.** A new `SecurityContext.asValidationContext()` returns a no-op (allow-all) view of the security context. While `validationOnly` is set, `SqlExecutionContextImpl.getSecurityContext()` returns this view, so every `authorize*` call during compilation is a no-op. The default returns `AllowAllSecurityContext` (sufficient for open-source builds); enforcing builds override it. `SqlCompilerImpl` routes all authorization through `executionContext.getSecurityContext()` rather than a captured `securityContext` field, so the validation view actually takes effect. `REFRESH MATERIALIZED VIEW` now resolves authorization through that view as well — the `authorizeMatViewRefresh` call moved out of the execution guard. Because authorization now flows through that single accessor, the previously added explicit `isValidationOnly()` guard around `authorizeColumnAccess` in `SqlOptimiser` is redundant and was removed.
- **Parser callback sees the execution context.** `SqlParserCallback.parseCreateTableExt` / `parseCreateMatViewExt` / `parseCreateViewExt` now receive the full `SqlExecutionContext` instead of a bare `SecurityContext`. This lets enforcing builds resolve identity and authorization (e.g. `OWNED BY`) through the validation-aware `getSecurityContext()` during CREATE TABLE / MATERIALIZED VIEW / VIEW parsing, instead of capturing the real security context at parse time.
- **Logging and metrics are suppressed.** `shouldLogSql()` returns `false` in validation-only mode, suppressing the `exe`/`fin` query-progress lines. `QueryProgress.logError()` returns early for validation-only failures so they are not logged as server-side `err` lines and do not increment the query-error metric — the error is already returned to the caller over HTTP. Query tracing is likewise skipped. Reader leaks are still logged and counted in both paths, since a leak indicates a real bug independent of validation mode.
- `SqlValidationProcessor` frees the compiled record cursor factory for every non-SELECT statement and resets `validationOnly` in a `finally` block, so the no-op security-context view can never outlive the request.

## Effects and tradeoffs

- Validation requests no longer enforce permissions, execute side effects, or appear in the server log, regardless of success or failure.
- Removing the coarse short-circuits means validation now resolves the target object for ALTER, REINDEX, TRUNCATE, and VACUUM. Validating such a statement against a non-existent table or view now reports "table does not exist" instead of passing as syntactically valid. This catches bad object references earlier but is a behavior change for callers that previously got an OK for these statements.
- Validation failures no longer increment the query-error metric. Operators who watched the log or that metric to spot malformed validation requests lose that signal; the error is still delivered to the client in the HTTP response.
- Normal (non-validation) execution, WAL apply, and telemetry contexts are unaffected.

## Test plan

- New `ValidationOnlyAlterTest`, `ValidationOnlyViewTest`, and `QueryProgressValidationTest` cover the alter/view/logging paths. `ValidationOnlyAlterTest` also pins the reject-on-missing-object behavior (`testValidateAlterMissingTableFails`, `testValidateReindexMissingTableFails`, `testValidateTruncateMissingTableFails`, `testValidateVacuumMissingTableFails`).
- `SqlValidationTest.testDoesNotMessUpTable` updated for the resolve-on-validate behavior, including the VACUUM-on-missing-table case.
- Added validate-then-execute coverage for the remaining inline-guarded paths: `CheckpointTest.testValidateCheckpointCreateDoesNotCreate` / `testValidateCheckpointReleaseDoesNotRelease`, `CancelQueryFunctionFactoryTest.testValidateCancelQueryDoesNotCancel`, and `MatViewTest.testValidateRefreshDoesNotRefresh`.
- Existing `SqlCompilerImplTest` passes.
- Manual: send valid and invalid SQL to `/api/v1/sql/validate` and confirm no `exe`/`fin`/`err` lines are produced and no state changes, while normal `/exec` queries still log and execute as before.

Required by https://github.com/questdb/questdb-enterprise/pull/1020


mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/QueryProgress.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlExecutionContextImpl.java|1|1|100.00%|

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/QueryProgress.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlExecutionContextImpl.java|1|1|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 79 / 87 (90.80%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|58|66|87.88%|
|:large_blue_circle:|io/questdb/cairo/SecurityContext.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/SqlValidationProcessor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlParser.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/QueryProgress.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlExecutionContextImpl.java|6|6|100.00%|

bluestreak01 (approved):
Approving. Verified the focused question: no statements execute during validation. The coarse short-circuits at the top of `compileAlter`, `compileReindex`, `compileTruncate`, `compileVacuum` are correctly replaced with fine-grained `isValidationOnly()` guards around each mutation site, so the matching Enterprise guards in https://github.com/questdb/questdb-enterprise/pull/1020 also take effect. All 18 guards line up with their mutation sites.

Minor follow-ups, none blocking:
- `SqlValidationProcessor.compileAndValidate` lines 275-285: factory acquired by `compile()` at 275, then three setters run before `Misc.free()` at 284. If any setter throws, factory leaks. Low risk, but moving `Misc.free` to a `finally` or immediately after `compile()` would tighten it.
- `compileTruncate` (lines 3689, 3718): `queryRegistry.register`/`unregister` still runs during validation because they're outside the new `!isValidationOnly()` guard at 3647. The for-loop they bracket never iterates (empty `tableWriters`), so this is a transient and self-cleaning side effect — but the queryId is briefly observable to `SHOW QUERIES`. Move the pair inside the guard.
- Default `SecurityContext.asValidationContext()` returns `AllowAllSecurityContext.INSTANCE` whose `getPrincipal()` is the constant `"admin"`. The new leaked-reader log in `QueryProgress.logEnd` (the only path that still reads principal during validation) will attribute leaks to `"admin"` instead of the real user on OSS. Either delegate `getPrincipal()` in the default, or read principal before swapping.
- VACUUM has no positive validation test (only the missing-table case). Legacy CHECKPOINT PREPARE/COMPLETE guards (3331, 3337) have no test. Add coverage.

PR #7126, opened 2026-05-20
fix(sql): fix wrong twap() and sparkline() results under concurrent query load

### LINKED ISSUE #7123: twap() / array_agg() / sparkline() can return wrong results under concurrent query load
## Summary

Three parallel-GROUP-BY aggregate functions — `twap()`, `array_agg()`, and `sparkline()` — share a flawed precondition in their `merge()` implementations. Under concurrent query load with cross-query work-stealing, the precondition is silently violated and the functions return wrong results. No exception, no log, no test catches it.

## Affected functions

- `io.questdb.griffin.engine.functions.groupby.TwapGroupByFunction`
- `io.questdb.griffin.engine.functions.groupby.SparklineGroupByFunction`
- `io.questdb.griffin.engine.functions.groupby.AbstractArrayAggDoubleGroupByFunction` (both `array_agg(D)` and `array_agg(D[])`)

All three carry javadoc claiming the same parallelism strategy:

> Within a single worker, page frames arrive in rowId order, so per-worker buffers are sorted runs. During the merge phase two sorted runs are combined with a merge-sort step ...

## Root cause

The "per-worker sorted runs" assumption is incorrect. Buffers are bound to **slots**, not worker threads:

- `AsyncGroupByAtom` creates `workerCount` `GroupByAllocator` instances; `setAllocator()` is called once at construction, wiring each per-slot `GroupByFunction` instance to its slot's allocator.
- For each page frame, `atom.maybeAcquire(workerId, owner)` performs linear-probe slot acquisition via `PerWorkerLocks`. Under contention a worker can land on a slot other than its primary.
- The cursor thread (`workerId = -1`) work-steals tasks from the global queue via `stealWork()`. For foreign tasks, `owner = false` and `acquireSlot(-1, ...)` randomizes the starting slot — the foreign cursor can land on any slot of the foreign atom.

As a result, a single slot N can accumulate page frames from multiple threads in arbitrary order:

```
t=0: worker A acquires slot N, processes frame for rowIds 500-599, releases
t=1: foreign cursor acquires slot N, processes frame for rowIds 200-299, releases
```

Slot N's buffer is now `[rowIds 500..599, rowIds 200..299]` — not a sorted run.

`computeNext` in all three functions appends unconditionally at the tail (no rowId/ts comparison). The subsequent two-pointer merge in `merge()` assumes both inputs are sorted and produces wrong output when they are not. `AbstractArrayAggDoubleGroupByFunction.tryMergeDisjointRuns` makes it worse — its fast path inspects only first/last entries and can blind-memcpy an unsorted buffer.

## Failure modes

| Function | Symptom |
|---|---|
| `twap()` | Wrong numeric result. Step-function deltas can go negative; the function silently returns a meaningless TWAP, sometimes falling back to plain arithmetic mean. |
| `array_agg()` | Visibly wrong element order in the returned array. |
| `sparkline()` | Wrong glyph shape — values plotted in slot-acquisition order rather than rowId order. |

## Reproduction

End-to-end repro tests forthcoming as a comment on this issue. They use:

- 2-worker pool, 50-row page frames, aggressive work-stealing config
- 8 concurrent threads firing the same query 30 times
- Deterministic input (`val = rowId`) so the correct result is exactly known

Observed under the repro:

- `twap()` returns ~1400 instead of the exact correct 2500.0 — wrong on ~80% of runs.
- `array_agg()` produces ~14k element-order inversions over 240 runs.
- `first()` / `last()` are unaffected (regression guard included), confirming the bug is specifically the sorted-runs assumption and not a general parallel-GROUP-BY issue.

The pathology is single-query-test-invisible because slot affinity holds under low contention.

mtopolnik:
Fixes #7123.

## Summary

Under concurrent query load with cross-query work-stealing, `twap()` and `sparkline()` can silently return wrong results during parallel GROUP BY. Both functions maintain per-slot native buffers of `(key, value)` entries and run a 2-pointer merge at slot-combine time. Their javadocs claim "Within a single worker, page frames arrive in rowId order, so per-worker buffers are sorted runs" - but the buffers are bound to slots, not workers. The cursor thread work-steals tasks across queries, and `PerWorkerLocks` lets foreign work land on any slot, so a single slot can accumulate frames in non-monotonic order. The 2-pointer merge then runs on unsorted input and silently produces wrong values for `twap()` and a wrong-order glyph for `sparkline()`.

This PR introduces `SortedRunsMerge`, a stable bottom-up pairwise mergesort over the natural sorted runs that a single scan detects in the buffer. The only invariant it relies on is one we can prove from the executor: within one `computeNext` loop, entries land in non-decreasing key order. Runs across frames can share keys, can interleave in key range, and are not required to be disjoint - a K-way merge of sorted runs handles all of that.

`TwapGroupByFunction.merge()` and `SparklineGroupByFunction.merge()` now delegate to `SortedRunsMerge.compactInto`. The read paths (`TwapGroupByFunction.getDouble`, `SparklineGroupByFunction.renderForPtr`) call `SortedRunsMerge.compactInPlace` to handle the case where no merge phase ran (single-frame query) but work-stealing still left the buffer in multi-run state.

An earlier attempt assumed runs were pairwise key-disjoint and tried to sort by permuting whole runs by their first key. That broke when two frames' key ranges happened to ascend across a gap (e.g. frame `[0..49]` then frame `[100..149]`, with frame `[50..99]` appended later): the run-discovery pass fuses them into one "fake" sorted run whose internal gap hides a missing range, and a concat-after-sort then produces wrong output. The pairwise K-way merge handles that case correctly because it does not require disjointness.

The `array_agg()` variant of the same bug is fixed on a separate PR branch and will land separately.

## Tradeoffs

- Adds a small allocation-free merge pass on the slot-combine path. For queries that never trip the multi-run condition, this is one extra linear scan of the buffer over what the previous 2-pointer merge did.
- The read paths now also call `compactInPlace` for safety; for fully single-run buffers this is a single-scan no-op but it is still extra work compared to the previous (incorrect) implementation that skipped the check.

## Test plan

- `TwapUnsortedRunReproTest` flipped from asserting the bug fires (mismatches > 0) to asserting correctness (mismatches == 0). Before fix: 182 wrong out of 240 runs. After fix: 0.
- `FirstLastUnderContentionTest` added as a regression guard for `first()` / `last()`, which are immune by construction (they compare rowIds explicitly) but should fail loudly if anyone ever refactors them to rely on insertion order.
- Existing `TwapGroupByFunctionFactoryTest` (20 tests) and `SparklineGroupByFunctionFactoryTest` (41 tests) - all green.

### DIFF
--- a/core/src/test/java/io/questdb/test/griffin/engine/functions/groupby/TwapUnsortedRunReproTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/engine/functions/groupby/TwapUnsortedRunReproTest.java
@@ -1,0 +1,168 @@
+/*+*****************************************************************************
+ *     ___                  _   ____  ____
+ *    / _ \ _   _  ___  ___| |_|  _ \| __ )
+ *   | | | | | | |/ _ \/ __| __| | | |  _ \
+ *   | |_| | |_| |  __/\__ \ |_| |_| | |_) |
+ *    \__\_\\__,_|\___||___/\__|____/|____/
+ *
+ *  Copyright (c) 2014-2019 Appsicle
+ *  Copyright (c) 2019-2026 QuestDB
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ ******************************************************************************/
+
+package io.questdb.test.griffin.engine.functions.groupby;
+
+import io.questdb.PropertyKey;
+import io.questdb.cairo.CairoEngine;
+import io.questdb.cairo.sql.Record;
+import io.questdb.cairo.sql.RecordCursor;
+import io.questdb.cairo.sql.RecordCursorFactory;
+import io.questdb.griffin.SqlExecutionContext;
+import io.questdb.mp.SOCountDownLatch;
+import io.questdb.mp.WorkerPool;
+import io.questdb.test.AbstractCairoTest;
+import io.questdb.test.tools.TestUtils;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * End-to-end repro that proves a per-slot buffer in
+ * {@link io.questdb.griffin.engine.functions.groupby.TwapGroupByFunction}
+ * can become unsorted by timestamp under realistic parallel-GROUP-BY execution
+ * with concurrent queries.
+ *
+ * <p>The dataset is constructed so the correct TWAP is exactly 2500.0 with no
+ * floating-point error:
+ * <ul>
+ *   <li>5000 rows, ts = rowIdx * 1000, price = rowIdx + 1</li>
+ *   <li>Step-function weightedSum = sum(i * 1000) for i = 1..4999
+ *       = 1000 * 4999 * 5000 / 2 = 12 497 500 000 (exact in double)</li>
+ *   <li>totalDuration = 4999 * 1000 = 4 999 000</li>
+ *   <li>TWAP = 12 497 500 000 / 4 999 000 = 2500.0 exactly</li>
+ * </ul>
+ * All intermediate sums stay well within 2^53, so integer addition in double
+ * is exact regardless of summation order - a correct (sorted) buffer always
+ * produces exactly 2500.0. Any deviation is direct evidence that merge()
+ * received an unsorted-by-ts input, which can only happen if a per-slot
+ * buffer accumulated frames in non-rowId order.
+ */
+public class TwapUnsortedRunReproTest extends AbstractCairoTest {
+
+    private static final int NUM_ITERATIONS = 30;
+    private static final int NUM_THREADS = 8;
+    private static final int ROWS = 5_000;
+    private static final double EXPECTED_TWAP = 2500.0;
+
+    @Test
+    public void testParallelTwapCanProduceWrongResult() throws Exception {
+        setProperty(PropertyKey.CAIRO_SQL_PAGE_FRAME_MAX_ROWS, 50);
+        setProperty(PropertyKey.CAIRO_SQL_PARALLEL_GROUPBY_ENABLED, "true");
+        setProperty(PropertyKey.CAIRO_SQL_PARALLEL_WORK_STEALING_THRESHOLD, 1);
+        setProperty(PropertyKey.CAIRO_SQL_PARALLEL_GROUPBY_SHARDING_THRESHOLD, 1_000_000);
+        setProperty(PropertyKey.CAIRO_SQL_PARALLEL_GROUPBY_BATCH_SIZE, 8);
+
+        assertMemoryLeak(() -> {
+            try (WorkerPool pool = new WorkerPool(() -> 2)) {
+                TestUtils.execute(pool, (engine, compiler, sqlExecutionContext) -> {
+                    engine.execute(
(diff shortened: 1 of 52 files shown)


nwoolmer (commented):
```
Re-tracing the merge path more carefully:                                                                                                                         
  
  - Old merge held dest + src + merged = peak ~2×(d+s) entries.                                                                                                                                                             
  - New merge holds dest + src + merged + aux = peak ~3×(d+s) entries during the sort.
  - Plus a count-of-scans regression in the happy path: when both inputs are themselves single sorted runs (the normal case), old code did one 2-pointer pass writing into merged. New code does: concat (2 memcpys) +      
  run-discovery scan + 2-way merge round + copy-back from aux = roughly 5 linear passes. Run-discovery may return 1 run and exit early only if the concatenated buffer happens to be globally sorted (A's max ≤ B's min) —  
  under work-stealing that's the exception, not the rule.                                                                                                                                                                   
                                                                                                                                                                                                                            
  The PR's "Tradeoffs" section calls this "one extra linear scan" — that undersells it. Not a blocker (these buffers are typically small, entries are 16 bytes), but the tradeoff section should be honest about the cost in
   both extra memory and extra passes so reviewers and future readers don't underestimate it.
                                                                                                                                                                                                                            
  There's also an easy win available for the common 2-input-single-run case: detect destCount > 0 && srcCount > 0 and that each input is a single run (cheaply, from the merge caller's perspective), then do the old       
  2-pointer merge from inputs directly into mergedPtr without the concat-then-aux dance. Falls back to the general path only when needed.
```

Looks good in principle, would be good to verify the extremes - high cardinality, many groups - to assess performance change. This is a correctness fix but we should still be sure we know the CPU/RAM scaling.

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 99 / 100 (99.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SortedRunsMerge.java|88|89|98.88%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/TwapGroupByFunction.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/SparklineGroupByFunction.java|4|4|100.00%|

core/src/test/java/io/questdb/test/griffin/engine/functions/groupby/TwapUnsortedRunReproTest.java
```diff
@@ -0,0 +1,283 @@
...
+                    }
+                    engine.execute(sb.toString(), sqlExecutionContext);
+
+                    final CyclicBarrier barrier = new CyclicBarrier(NUM_THREADS);
+                    final SOCountDownLatch latch = new SOCountDownLatch(NUM_THREADS);
+                    final Map<Integer, Throwable> errors = new ConcurrentHashMap<>();
+                    final AtomicInteger mismatches = new AtomicInteger();
+                    final AtomicInteger threadsThatSawMismatches = new AtomicInteger();
+                    // Track one observed wrong value so the failure print is concrete.
+                    final java.util.concurrent.atomic.AtomicReference<Double> sampleWrongValue =
```
puzpuzpuz, line 111: Fully qualified class names should be avoided. Claude often writes them when the context gets large.

core/src/test/java/io/questdb/test/griffin/engine/functions/groupby/TwapUnsortedRunReproTest.java
```diff
@@ -0,0 +1,283 @@
...
+ * </ul>
+ * All intermediate sums stay well within 2^53, so integer addition in double
+ * is exact regardless of summation order - a correct (sorted) buffer always
+ * produces exactly 2500.0. Any deviation is direct evidence that the
+ * compaction step in
+ * {@link io.questdb.griffin.engine.groupby.SortedRunsMerge} either was not
+ * called or failed to restore a sorted buffer before the integration walk in
+ * {@code getDouble}.
+ */
+public class TwapUnsortedRunReproTest extends AbstractCairoTest {
```
puzpuzpuz, line 74: While this test is cool, we should improve `ParallelGroupByFuzzTest`. The idea is to let the number of slots in async group by atoms to be overridable in tests, so that we configure less slots than the number of worker threads. In that case, order violations should reproduce easily in `ParallelGroupByFuzzTest` and it'll be enough to add tests with twap or any other problematic function there.

bluestreak01: too late, followup PR @mtopolnik

PR #7127, opened 2026-05-20
chore(core): fix flaky LineHttpSender and QWP failover tests

glasstiger:
## Summary

Two test-only fixes for intermittent failures in the ingestion test harness.
Neither touches production code or changes ingestion/rename behaviour.

- `LineHttpSenderTest.testConcurrentRenameWhileProcessingLargeBatch`
- `QwpSenderFailoverBatchSizeTest.testReconnectToTighterCapRefreshesServerMaxBatchSize`

Note: the core rename race in the first test (claiming the `batchPausedOnce`
slot before counting down `walSegmentRolled`, plus excluding the test thread
from the `openRW` interceptor) already landed on master via #7131, which this
branch merged in. The remaining net-new change to that test here is only the
timeout surfacing described below.

## testConcurrentRenameWhileProcessingLargeBatch

The parked `openRW` interceptor previously swallowed a timed-out
`renameRegistered.await(60, SECONDS)`. If the rename ever failed to register
within 60s, the batch committed unparked, no 503 was produced, and the test
failed with the misleading `AssertionError: Expected LineSenderException due to
table rename` rather than pointing at the real cause.

Change:

- Record the `await` timeout in `renameRegisteredTimedOut` instead of
  discarding it, and assert on it on the main thread after the sender joins.
  A rename that fails to register within 60s now fails the test with an
  accurate message.

## testReconnectToTighterCapRefreshesServerMaxBatchSize

Failed intermittently inside `assertMemoryLeak()` with a leaked file descriptor
count (e.g. 5 extra OS fds), not on the assertion the test was written to check.

Each sidecar's QWP processor acquires a pooled WAL writer to ingest the
warm-up/failover rows. The cleanup called `engine.releaseInactive()` while
serverB's sidecar worker pool was still running. `releaseInactive()` only frees
pool entries that are not currently checked out, so it raced serverB's worker
thread returning its WAL writer to the pool. When the test thread won the race,
the still-checked-out writer was skipped and its file descriptors leaked; when
serverB returned the writer first, the test passed. serverA was unaffected
because it is stopped earlier in the test body.

Change:

- Stop serverB before draining the WAL queue and calling `releaseInactive()`,
  so its worker pool halts and returns the WAL writer to the pool first.
  `QwpSidecar.stop()` is idempotent, so the existing `finally` block remains
  safe.

## Test plan

- Ran `testConcurrentRenameWhileProcessingLargeBatch` repeatedly (6 consecutive
  runs); all pass in ~2s.
- Ran `testReconnectToTighterCapRefreshesServerMaxBatchSize` repeatedly to
  confirm the fd leak no longer reproduces.

### DIFF
--- a/core/src/test/java/io/questdb/test/cutlass/http/line/LineHttpSenderTest.java
+++ b/core/src/test/java/io/questdb/test/cutlass/http/line/LineHttpSenderTest.java
@@ -861,6 +861,7 @@ public class LineHttpSenderTest extends AbstractBootstrapTest {
             CountDownLatch renameRegistered = new CountDownLatch(1);
             AtomicBoolean trackOpens = new AtomicBoolean(false);
             AtomicBoolean batchPausedOnce = new AtomicBoolean(false);
+            AtomicBoolean renameRegisteredTimedOut = new AtomicBoolean(false);
 
             FilesFacade ff = new TestFilesFacadeImpl() {
                 @Override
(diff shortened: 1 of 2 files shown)


bluestreak01:
duplicate

bluestreak01 (approved):

PR #7128, opened 2026-05-20
fix(qwp): fix WebSocket ingest disconnects during fragmented 101 handshake

bluestreak01:
## Summary

The QWP/WebSocket ingress upgrade processor sent the 101 Switching
Protocols response synchronously from `onHeadersReady` and caught
`PeerIsSlowToReadException` as a fatal "handshake blocked" error.
`HttpResponseSink.sendBuffer` deliberately throws PISR after a chunk
when `forceSendFragmentationChunkSize` is smaller than the response,
and also throws on the `n == 0` branch when a real slow peer's kernel
send window cannot accept the whole 101 in one shot. In both cases
the processor converted the signal into an `HttpException` and
disconnected mid-handshake. A client connecting via QWP/WebSocket
ingest then saw exactly `chunkSize` bytes followed by EOF before the
`\r\n\r\n` header terminator arrived.

In production with the default `forceSendFragmentationChunkSize =
Integer.MAX_VALUE`, the fragmentation branch never trips. The
`n == 0` branch can still fire when the kernel send buffer is small
or a proxy / TCP-window quirk fragments the 101 response, so the bug
is rare in default deployments but possible in practice. The
egress upgrade processor on the same code path already handles this
correctly; this PR brings the ingress processor in line.

The fix mirrors `QwpEgressUpgradeProcessor`: `onHeadersReady` writes
the response into the send buffer and records the byte count on the
per-connection state; `onRequestComplete` performs `rawSocket.send`
and finalises the protocol switch; `resumeSend` drains any parked
bytes via `context.resumeResponseSend()` before finalising. PISR
propagates into the framework's park-on-write path and the dispatcher
resumes the flush until the buffer is empty. The protocol switch
happens only after the 101 response is fully on the wire, so the
connection never starts parsing WebSocket frames while the handshake
is still in flight.

## Tradeoffs and notes

- `QwpProcessorState` gains two fields (`handshakeFlushPending`,
  `pendingHandshakeBytes`) and getters/setters, paralleling
  `QwpEgressProcessorState`. `onDisconnected()` resets them.
- `onRequestComplete`'s declared exceptions widen to match the
  `HttpRequestProcessor` contract (`PeerDisconnectedException`,
  `PeerIsSlowToReadException`, `ServerDisconnectException`) since
  it now performs network I/O. Callers in the framework already
  expect these.
- The protocol switch shifts from `onHeadersReady` to a new private
  `finalizeHandshake` helper invoked once the send completes. The
  framework invokes `onHeadersReady` and `onRequestComplete`
  back-to-back inside `handleClientRecv`, so the synchronous fast
  path for callers that never park is observationally identical to
  before.
- The buffer-too-small failure paths (400 / 426 / 421) keep their
  existing synchronous send. Those responses fit in one chunk and
  the failure semantics there are unchanged.
- Three white-box unit tests that asserted the old in-`onHeadersReady`
  send semantics now also drive `onRequestComplete`; no assertion
  intent changed. `testHandshakeSendSlowToRead` was flipped from
  asserting an `HttpException` to asserting PISR propagation, since
  PISR is now the correct outward signal of a parked handshake send.

## Test plan

- Added `QwpWebSocketProtocolTest#testHandshakeSurvivesSmallSendChunkSize`,
  which boots the server with `forceSendFragmentationChunkSize = 7`
  and performs a WebSocket upgrade. On master this fails with
  `Unexpected end of stream during handshake` after exactly 7 bytes;
  with this change it passes.
- Ran the full `Qwp*Test, *WebSocket*` test set (1391 tests): all
  pass with no skips beyond the two pre-existing skipped tests.
- Ran adjacent suites (`LineHttp*Test`, `*HttpServer*Test`,
  `*HttpProcessor*Test`, 164 tests): all pass.
- Updated white-box unit tests
  (`QwpWebSocketUpgradeProcessorOnHeadersReadyTest`,
  `QwpWebSocketUpgradeProcessorResumeRecvTest`) to drive
  `onRequestComplete` where they previously relied on
  `onHeadersReady` having already sent the response. The c-questdb-client
  Rust fuzz test `binary_qwp_ws_random_fuzz` referenced in the
  original bug report exercises the same path and should start
  passing without further change.

### DIFF
--- a/core/src/main/java/io/questdb/cutlass/qwp/server/QwpProcessorState.java
+++ b/core/src/main/java/io/questdb/cutlass/qwp/server/QwpProcessorState.java
@@ -93,10 +93,12 @@ public class QwpProcessorState implements QuietCloseable, ConnectionAware {
     private byte deferredErrorStatus;
     private boolean durableAckEnabled;
     private long fd = -1;
+    private boolean handshakeFlushPending;
     private long highestProcessedSequence = -1;
     private long lastAckedSequence = -1;
     private long messageSequence;
     private byte negotiatedVersion = QwpConstants.VERSION_1;
+    private int pendingHandshakeBytes;
     private int recvBufferLen;
     private long resumeAckSequence = -1;
     private SecurityContext securityContext;
@@ -275,6 +277,10 @@ public class QwpProcessorState implements QuietCloseable, ConnectionAware {
         return pendingAckSeqTxns;
     }
 
+    public int getPendingHandshakeBytes() {
+        return pendingHandshakeBytes;
+    }
+
     public int getRecvBufferLen() {
         return recvBufferLen;
     }
@@ -297,6 +303,10 @@ public class QwpProcessorState implements QuietCloseable, ConnectionAware {
 
     public boolean isDurableAckEnabled() {
         return durableAckEnabled;
+    }
+
+    public boolean isHandshakeFlushPending() {
+        return handshakeFlushPending;
     }
 
     public boolean isOk() {
@@ -415,6 +425,8 @@ public class QwpProcessorState implements QuietCloseable, ConnectionAware {
         sendState = SEND_STATE_READY;
         clearDeferredError();
         clearDeferredClose();
+        handshakeFlushPending = false;
+        pendingHandshakeBytes = 0;
         wsHandshakeSent = false;
 
         // Drop any durable-ack state; the connection is going away, so even if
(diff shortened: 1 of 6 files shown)


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 26 / 27 (96.30%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpWebSocketUpgradeProcessor.java|18|19|94.74%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpProcessorState.java|8|8|100.00%|

bluestreak01:
duplicate

PR #7130, opened 2026-05-21
fix(qwp): allow query results larger than the HTTP send buffer

kafka1991:
QWP egress was failing any query whose `RESULT_BATCH` frame did not fit in the HTTP send buffer (default 2 MiB). The most common trigger is a high-cardinality SYMBOL column: the connection-scoped dict delta grows linearly with first-seen values and can exceed the buffer in a single batch, causing the server to return `STATUS_INTERNAL_ERROR` with "egress: batch too large for send buffer". The same failure mode applied to any sufficiently wide table block (large VARCHAR, multi-column rows, ARRAY).

`QwpEgressUpgradeProcessor.streamResults` now splits oversized results across multiple batches instead of failing. Two overflow sources are handled separately because dict delta and table block have different properties:

- **Dict delta** ships as one unbreakable unit on the wire. The fill loop monitors `connSymbolDict.heapBytes()` and stops appending rows when the new delta would exceed 60% of the send buffer. The partially-filled batch ships; the remainder rolls into the next iteration.
- **Table block** can be sliced row-by-row. New `QwpResultBatchBuffer.findLargestEmittablePrefix` binary-searches the largest K such that the table block for K rows fits the remaining budget, and ships that prefix via a new `emitTableBlockPrefix`. The un-shipped suffix stays in the scratch (tracked via a new `startRow` field) for the next iteration.

When a single row exceeds the send buffer the server now throws a `QwpRowExceedsBufferException` mapped to `STATUS_LIMIT_EXCEEDED` with a message naming the actual constraint, rather than the generic `STATUS_INTERNAL_ERROR`.

### Implementation notes

- `QwpResultBatchBuffer` gains a `startRow` field so partial emits can leave a suffix in place without a memmove. `advanceStartRow` triggers an in-place position reset when the buffer drains (the common case), matching today's per-batch reset behaviour on the happy path.
- A single `emitTableBlockImpl(..., dryRun)` underpins both `computeTableBlockSize` (used by the prefix search) and the real emit, so emit-time bytes and pre-computed bytes share one code path and cannot drift.
- ARRAY column slicing required adding a per-row `arrayOffsetsAddr` table to `QwpColumnScratch`, mirror of the existing `stringOffsetsAddr`.


bluestreak01:
@kafka1991 — review at level 2 (level 3 exists for a full mission-critical pass).

## Summary

This PR adds partial-emit support to the QWP egress streaming path: a binary-search-based prefix selector splits oversized RESULT_BATCH frames row-by-row, with a separate 60% dict-budget cap for the unbreakable SYMBOL dict delta. The slicing logic for null bitmaps, dense values, strings/binary, arrays, symbols, and Gorilla timestamps is **mostly correct end-to-end**, and the resource cleanup paths are clean. However, the budget arithmetic has two off-by-margin bugs at small `bufSize` (the regime this PR is supposed to support), the new public surface has no direct unit tests, one test name is misleading, and the PR metadata violates project conventions.

**Verdict: request changes.** Two correctness bugs reachable at small `bufSize` (the targeted regime), plus diagnostic and test-coverage gaps. No critical safety issues.

## Critical

None. (No memory safety, no leak, no concurrency, no data-corruption bug found in the partial-emit slicing.)

## Moderate

### M1 — `emitDeltaSection` per-entry guard uses `MAX_VARINT_BYTES`, can spuriously fail at the budget boundary
**File:** `core/src/main/java/io/questdb/cutlass/qwp/codec/QwpResultBatchBuffer.java:432` (and the header preflight at line 420).

`streamResults` (`QwpEgressUpgradeProcessor.java:1846-1852`) computes the table-block budget by subtracting `computeDeltaSize()` from `bufSize`, where `computeDeltaSize()` returns the **exact** wire bytes via `QwpVarint.encodedLength(...)`. `findLargestEmittablePrefix` then picks the largest `K` with `tableBlockSize <= budget`. So the total bytes for delta + table block are guaranteed to fit.

But `emitDeltaSection`'s per-entry guard reserves the pessimistic `MAX_VARINT_BYTES = 10` instead of `QwpVarint.encodedLength(entryLen)`. At the last entry, when `varint(entryLen)` is 1 byte but the table-block budget consumed almost all of its allowance, the guard requires `9 - varint(lastEntryLen)` extra bytes that aren't there. The check trips and `sendResultBatch:1457` throws `HttpException.instance("egress: delta section overflows send buffer")` — which is mapped to `STATUS_INTERNAL_ERROR`, not a clean limit error.

Concrete repro: `bufSize=54` (minimum allowed), 1 symbol entry of length 12, no rows: `deltaBytes=15`, `budget=8`, `tableBlockSize(1)=7` fits. At per-entry check: `p + 10 + 12 = 22 > wireLimit - preludeEnd = 21`. Fails. The bug also exists in the **header preflight** at line 420 (`p + 2 * MAX_VARINT_BYTES > wireLimit`), which spuriously fails when the actual two header varints are short and budget is tight.

**Fix:** in the per-entry guard, use `QwpVarint.encodedLength(entryLen)` instead of `MAX_VARINT_BYTES`. In the header preflight, compute exact sizes upfront via `encodedLength(deltaStart) + encodedLength(deltaCount)`.

### M2 — `dictBudget` measures heap bytes, but the wire delta includes per-entry varint overhead
**File:** `core/src/main/java/io/questdb/cutlass/qwp/server/egress/QwpEgressUpgradeProcessor.java:1772-1802`.

`dictBudget = (bufSize * 6) / 10` is compared against `connSymbolDict.heapBytes() - dictBytesAtStart`, i.e. raw concatenated UTF-8 bytes. The actual wire delta is `2 + sum(varint(entryLen_i) + entryLen_i)` — heap plus 2-byte header plus 1-10 bytes of varint per entry. For many short symbols, wire-bytes can be ~2x heap. At small `bufSize` with short symbol values this can push wire delta past `bufSize - constants` even though heap stayed under the 60% cap.

In that case `budget = bufSize - constants - deltaBytes` goes negative, `findLargestEmittablePrefix` returns -1, and the user sees `QwpRowExceedsBufferException → STATUS_LIMIT_EXCEEDED` with message "row prefix does not fit send buffer" — but the real failure is the **dict delta**, not the row. Concrete reachable case at minimum `bufSize=54` with 32 single-character symbols.

**Fix:** budget the dict against `computeDeltaSize()` (wire bytes) rather than `heapBytes()`. Either change the cap check to incrementally track wire delta, or recompute `computeDeltaSize()` after each batch of new entries and break when wire delta approaches a wire-byte ceiling.

### M3 — Misleading `QwpRowExceedsBufferException` message for non-row overflows
**File:** `core/src/main/java/io/questdb/cutlass/qwp/server/egress/QwpEgressUpgradeProcessor.java:1860-1864`.

`findLargestEmittablePrefix` returns `-1` for "even the empty header overflows the budget" — a pathological send buffer config (very small `bufSize` with a wide full schema, or M2 above). The call site unconditionally throws `QwpRowExceedsBufferException.instance(colCount, bufSize, rowsBuffered)` whose message says `"egress: row prefix does not fit send buffer"`. When `rowsBuffered == 0` or when the actual blocker is the schema header, the diagnostic points at the wrong thing.

**Fix:** distinguish `k == 0` (single-row overflow) from `k == -1` (header/schema/delta overflow) and emit a separate marker exception or message text for the latter.

### M4 — `computeDeltaSize` is a hand-rolled mirror of `emitDeltaSection`, not actually a shared code path
**File:** `core/src/main/java/io/questdb/cutlass/qwp/codec/QwpResultBatchBuffer.java:374-389`.

The comment claims "mirrors the emit byte layout step-by-step ... tests guarantee the two stay in sync" but no direct parity unit test exists in this PR. `computeTableBlockSize` correctly delegates to `emitTableBlockImpl(dryRun=true)` (single source of truth) — `computeDeltaSize` does NOT. A future change to either side that doesn't update the other will silently produce off-by-N wire frames.

**Fix:** add `computeDeltaSection(wireBuf, wireLimit, boolean dryRun)` that drives both paths, mirroring the `emitTableBlockImpl(dryRun)` pattern. Or add a unit test that asserts `computeDeltaSize() == emitDeltaSection(buf, MAX)` across varint-boundary entry lengths (1, 127, 128, 16383, 16384, large).

### M5 — Direct unit tests are missing for the new public methods
The PR adds six new public methods on `QwpResultBatchBuffer` — `advanceDeltaStart`, `advanceStartRow`, `computeDeltaSize`, `computeTableBlockSize`, `emitTableBlockPrefix`, `findLargestEmittablePrefix` — and none has a direct unit test. Coverage is entirely end-to-end at 256K `bufSize`, and the dictBudget vs dictCapHit branches are not isolated. Specifically missing:
- `findLargestEmittablePrefix` returning `-1` (header alone too big) and `0` (header fits, no row).
- `computeDeltaSize` vs `emitDeltaSection` byte parity across varint-boundary entry lengths.
- `advanceStartRow` reset-to-zero path when partial-emit drains the buffer exactly.
- Wire-byte equivalence between N partial-emit frames and one full-emit frame for the same logical batch (the PR's "byte-identical" claim is untested).
- Mid-stream `QwpRowExceedsBufferException` after some rows already shipped (leak test).
- `PeerIsSlowToReadException` parked-send resume during partial-emit carryover.
- Isolated dictCapHit-only trigger separate from table-block overflow.

### M6 — `testOverflowPreservesBooleanWithNulls` is misnamed and does not actually test wire NULL preservation
**File:** `core/src/test/java/io/questdb/test/cutlass/qwp/QwpEgressDeltaSymbolDictTest.java:323-379`.

QuestDB's BOOLEAN has no NULL representation: `CASE ... ELSE NULL` is coerced to `false` at INSERT time before egress ever sees it. The test asserts `nullCount==0` and `falseCount==167+167`, which is consistent with QuestDB storage but the test name promises something the code doesn't do. The wire path is never asked to transport a BOOLEAN null. Rename to reflect that this exercises BOOLEAN bit-packing across partial-emit boundaries, OR use a nullable column type (UUID/INT/VARCHAR with explicit NULLs) to actually test wire-null preservation across the slice boundary.

### M7 — Binary-search dry-run pays O(sliceNonNull) per TIMESTAMP probe
**File:** `core/src/main/java/io/questdb/cutlass/qwp/codec/QwpResultBatchBuffer.java:1129-1173`.

`findLargestEmittablePrefix` does `~log2(N)` probes of `computeTableBlockSize`, each materializing column sizes. Non-TS columns use the prefix-sum (O(1)). TS columns still run `QwpGorillaEncoder.calculateEncodedSizeIfSupported` per probe (O(sliceNonNull)). The `cachedGorilla*` cache is keyed by `(nonNullsBefore, sliceNonNull, nonNullCount)`; binary-search probes have distinct `sliceNonNull` values so every probe misses the cache. For 100-column tables with 4096 rows that always overflow, that's ~50K TS-value comparisons per iteration *before* the real emit.

Acceptable for the fix but worth a follow-up: precompute a per-batch Gorilla running-bits prefix-sum and look up `gorilla(start, K) = bits[start+K] - bits[start]` in O(1) per probe, OR use the raw-bytes upper bound during search and only run the real Gorilla on the chosen `K`.

### M8 — PR metadata violates project conventions
- **Title:** `fix(qwp): allow query results larger than the HTTP send buffer`. The description part should repeat the verb (the squashed release-note line). Should be `fix(qwp): fix query results being rejected when larger than HTTP send buffer` or similar.
- **Labels:** empty. Should have at least `Bug` and `REST API` (QWP is the egress HTTP protocol).
- **No `### Test plan`** section. Add plain `-` bullets covering: small-buffer + high-cardinality SYMBOL, oversized VARCHAR row, oversized ARRAY row, `STATUS_LIMIT_EXCEEDED` single-row case, happy-path regression.
- **Tradeoffs not presented.** Project conventions require presenting negative effects "with equal weight". Missing: binary-search dry-run cost, extra wire frames, 60% dict ceiling wasting headroom.

## Minor

- **Member ordering in `QwpColumnScratch.java`:** `ensureArrayHeapCapacity` (line 939), `markNonNullAndAdvanceRow` (946), `resetPositions` (959) are misplaced — sit after `updatePeakUsage` (924) instead of alphabetised among the other package-private methods. `recordArrayOffset` (264) and `recordSymbolBytes` (275) sit in the private-helper region between `recordStringOffset` and `setNullBit`.
- **`dictCapHit` (QwpEgressUpgradeProcessor.java:1775) should be `isDictCapHit`** per the boolean-naming convention.
- **`int physicalRowCount` / `int startRow` long-tail overflow:** in sustained partial-emit steady state these never reset. At 10K batches/sec × 4096 rows ≈ 58 hours before INT_MAX. Realistic only on permanently-overflowing streams at tiny `bufSize`. Consider `long`, or add an assert/guard.
- **`sendResultBatch:1507` comment cross-references `QwpEgressUpgradeProcessor.java:1492-1497`** — that line range no longer points at the referenced explanation after this PR's refactor.

## Downgraded (false positives removed)

- **BOOLEAN/BYTE/SHORT/CHAR column-top divergence between columnar and per-row paths:** pre-existing behavior, not introduced by this PR. The columnar `fillNulls` path was added in earlier work; the PR only changes how the bitmap is sliced.
- **`QwpRowExceedsBufferException` ThreadLocal cross-talk:** identical pattern to `HttpException`. The HTTP worker is single-threaded per fd and the exception is consumed inline within the catch frame before any other connection can reuse the worker. Not a new issue.
- **Concurrency:** `Counter` is `LongAdder`-backed (safe). Memory visibility across `PeerIsSlowToReadException` park/resume runs on the same worker via the dispatcher's per-fd affinity. No new races.
- **Resource leaks:** both new native buffers (`arrayOffsetsAddr`, `symbolBytesCumAddr`) are paired with capacity counters and freed in `close()`. `Unsafe.realloc` semantics keep state consistent on OOM. Connection close reaches `QwpColumnScratch.close()` via `Misc.free(batchBuffer) → freeObjListIfCloseable(scratches)`.
- **`rollbackCurrentBatch` re-entrancy / Gorilla cache reset gap:** in practice every rollback is followed by `endStreaming` → next query → `beginStreaming` → `resetForNewQuery` → next `beginBatch` resets `cachedGorillaNonNullCount = -1`. Reachable only on contrived call sequences.
- **`isCursorExhausted = false` when `rowsToAdd == 0`:** correctness-preserving (just adds an extra empty round trip on the cursor). Not a bug.
- **Hot-path allocations introduced by the PR:** none found. No `java.util.*`, no autoboxing, no capturing lambdas, no String concat on the data path. `wireTypeFixedSize`'s `default` throws via a chained `CairoException.put(...)` but only on unreachable wire codes.

## Verification summary

10 confirmed findings (2 in-diff correctness, 1 diagnostic, 1 cross-file maintainability, 3 test-coverage, 1 perf cost, 1 minor ordering, 1 PR metadata) plus 4 style/minor. ~7 draft findings downgraded as false positives during verification. All findings are in-diff. No out-of-diff callsites broken: the only callers of `getRowCount()` outside the file are unrelated classes (`QwpUdpReceiver.TableBlock`, `QwpTudCache.cursor`); all in-package callers were updated.

kafka1991:
Pushed fixes for the level-2 review:

**Correctness**
- **M1** `emitDeltaSection` per-entry guard and header preflight now use exact `QwpVarint.encodedLength(...)` instead of `MAX_VARINT_BYTES`. No more spurious failures at tight `bufSize`.
- **M2** dict cap compares against wire bytes (not heap) via a new O(1) incremental tracker: `QwpEgressConnSymbolDict.totalWireBytes` maintained in `addEntry` / `rollbackTo` / `clear`; `QwpResultBatchBuffer.currentBatchDeltaWireBytes()` returns `varint(deltaStart) + varint(deltaCount) + (totalWireBytes - snapshot)`. Replaces the row-fill loop's prior `computeDeltaSize()` call, eliminating the worst-case O(rows²) when every row introduces a new symbol. `computeDeltaSize()` (the O(N) walker) stays for the outer-iteration budget computation as byte-layout SSOT.
- **M3** `QwpRowExceedsBufferException.instance(..., boolean isHeaderOverflow)` branches the message: `k == -1` -> "table block header does not fit send buffer"; `k == 0` -> "single row exceeds send buffer".

**Maintainability**
- **M4** extracted `emitDeltaSectionImpl(wireBuf, wireLimit, dryRun)`. `emitDeltaSection` and `computeDeltaSize` drive through it. Drift is structurally impossible.

**Tests**
- **M5** added in `QwpEgressReviewFindingsTest`:
  - `testComputeDeltaSizeMatchesEmitDeltaSection` (byte parity across varint boundaries)
  - `testCurrentBatchDeltaWireBytesMatchesComputeDeltaSize` (O(1) tracker vs walker after 7 varint-boundary inserts + partial rollback)
  - `testFindLargestEmittablePrefixReturnsMinusOneOnHeaderOverflow`
  - `testFindLargestEmittablePrefixReturnsZeroWhenNoRowFits`
- **M6** renamed `testOverflowPreservesBooleanWithNulls` -> `testOverflowPreservesBooleanBitPacking` with a comment noting QuestDB BOOLEAN has no NULL representation; the test exercises bit-packing across partial-emit boundaries.

**Minor**
- `physicalRowCount` / `startRow` int-overflow asserts on the three increment paths (`-ea`-only).
- Stale comment reference at `sendResultBatch:1507` rewritten self-contained.

**Rejected after cost analysis (not deferred)**
- **M7** Gorilla binary-search prefix-bits optimisation. Would add per-row work to every TIMESTAMP append (and the bulk path is shared with LONG/DATE/DECIMAL64, so it needs hot-path branching or method duplication) in order to speed up the rare overflow path. Inverted tradeoff: it taxes the common case to optimise the rare case. The existing per-emit `cachedGorilla*` cache already collapses the happy-path 2× scan to 1×, which was the only real regression. Won't pursue absent profile evidence.

**Tradeoffs** (per project conventions, with equal weight):
- Wire-bytes tracker adds 16 bytes of state (one `long` on dict + snapshot on batchBuffer). Tiny.
- 60% dict ceiling intentionally wastes headroom for predictable splits; adaptive sizing is a follow-up if `qwp_egress_batch_overflow_splits` shows pressure.
- Binary-search dry-run still re-scans TIMESTAMP columns per probe (M7 rejected above).

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 421 / 477 (88.26%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpRowExceedsBufferException.java|14|17|82.35%|
|:large_blue_circle:|io/questdb/cutlass/qwp/codec/QwpResultBatchBuffer.java|236|278|84.89%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressUpgradeProcessor.java|58|65|89.23%|
|:large_blue_circle:|io/questdb/cutlass/qwp/codec/QwpColumnScratch.java|91|95|95.79%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressMetrics.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/codec/QwpEgressSchemaWriter.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/codec/QwpEgressConnSymbolDict.java|10|10|100.00%|

bluestreak01 (approved):

PR #7136, opened 2026-05-22
perf(sql): stop materialized views emitting no-op WAL transactions under base table backlog

puzpuzpuz:
Fixes primary instance-side of the bug previously fixed in #7064. With this patch, primary enterprise instances no longer generate bursts of no-op replace-range transactions.

When an incremental, full, or range refresh finds no intervals to query, `insertAsSelect()` still recorded the refresh by committing a no-rows `resetMatViewState` WAL transaction. For non-period views an earlier guard skips this when the base txn is unchanged, but period views bypass that guard, so they committed the transaction even when neither the base txn watermark nor the period hi advanced.

Under base table apply backlog the base table's readable txn stays static while the dependent-view refresh task re-enqueues in a tight loop. Each iteration produced one such no-op commit, so a single view could emit millions of replace-range transactions in minutes. The flood bloats the view's WAL and, on replicas, stalls WAL apply.

This PR makes two related changes to that path.

First, `insertAsSelect()` now commits the no-rows watermark only when it actually advances: `commitBaseTxn > lastRefreshBaseTxn` or `commitPeriodHi > lastPeriodHi`. Otherwise it writes no transaction and returns false. Legitimate advances still commit - a new base txn consumed with no matching rows, a completed period (the interval iterator is non-null then), and a full refresh of an empty base table (`lastRefreshBaseTxn` is -1 after `resetInvalidState`, so it advances). This removes the harmful WAL flood and is what protects replicas.

Second, `refreshDependentViewsIncremental()` acknowledges the base txn the refresh caught up to more precisely. The base table refresh notification uses a clean/dirty txn handshake: the apply side stamps the latest readable base txn, and the refresh job acknowledges the txn it reached, re-enqueueing itself until the two meet. The method previously sampled the base writer txn once, before refreshing the dependent views, so when the readable txn advanced between that sample and a view opening its reader - as it does under apply backlog - the acknowledged txn trailed what the views examined and the refresh re-enqueued with no new work. It now acknowledges the minimum base txn examined across the refreshed views. That minimum never exceeds what any view examined, so no refresh is dropped and no view goes stale, and it is always at least the previous sample, so the handshake converges sooner.

The second change reduces, but does not fully eliminate, the redundant re-enqueues. The window between a view opening its reader and the acknowledgement remains, so under sustained apply backlog the primary can still re-run the refresh without new work - now without emitting WAL transactions, because of the first change. Full convergence of the re-enqueue handshake is left for a follow-up.

The PR also removes the `@TestOnly` annotations from `SeqTxnTracker.getSeqTxn()` and `getWriterTxn()`; several production call sites already use both methods.

## Test plan

- Adds `MatViewTest.testNoOpIncrementalPeriodRefreshDoesNotCommitWal`: refreshes a period view once, then runs a no-op incremental refresh and asserts the view's `wal_transactions` listing is unchanged.
- `MatViewTest` (176) and `MatViewFuzzTest` (22) pass; the fuzz suite exercises concurrent refresh and confirms views do not go stale with the new acknowledgement value.
- No dedicated test for the acknowledgement change: the sustained re-enqueue spin does not reproduce in a single-node test - the handshake converges there with or without the change - so such a test would pass regardless of the fix.


puzpuzpuz:
@jerrinot fixed stale metadata in f8d1b2f4154d6ff8359dff663e6ffb7370b53550. Thanks for flagging this!

core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
```diff
@@ -81,6 +81,10 @@ public class MatViewRefreshJob implements Job, QuietCloseable {
     private final FixedOffsetIntervalIterator fixedOffsetIterator = new FixedOffsetIntervalIterator();
     private final MatViewGraph graph;
     private final LongList intervals = new LongList();
+    // The base table txn the most recent refreshIncremental0() examined (its reader's seqTxn), or
+    // Numbers.LONG_NULL if it did not open a reader. Read by refreshDependentViewsIncremental() to
+    // acknowledge the base txn that all refreshed views actually caught up to.
+    private long lastExaminedBaseTxn = Numbers.LONG_NULL;
```
ideoma: Nit: `lastExaminedBaseTxn` is a hidden out-parameter. Returning a value via a mutable instance field reset between calls is fragile. The current dance (reset → call → conditionally read) is correct because every caller path resets before reading, but a future caller that forgets the reset will silently read a stale value. Consider refactoring `refreshIncremental0` to return a `long`, or accepting a `LongHolder` out-param.

core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
```diff
@@ -1298,6 +1326,9 @@ private boolean refreshIncremental0(
         try (TableReader baseTableReader = engine.getReader(baseTableToken)) {
             final long fromBaseTxn = viewState.getLastRefreshBaseTxn();
             final long toBaseTxn = baseTableReader.getSeqTxn();
+            // Record the base txn this refresh examines, even on the early-out paths below, so the
+            // caller can acknowledge only what was actually examined.
+            lastExaminedBaseTxn = toBaseTxn;
```
ideoma: Nit: the assignment happens before the `fromBaseTxn > toBaseTxn` throw a few lines below. The catch in `refreshDependentViewsIncremental` (around line 1182) skips the field read at line 1177, so it's safe today. But the comment above ("even on the early-out paths below") is misleading — the throw is also "below" but the value is never consumed there. Move the assignment after the sanity check, or qualify the comment.

ideoma (dismissed):

core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
puzpuzpuz: Addressed in 00e1f192c39edfc4776b2d1ba2d67f1aedccc5ae by bit-packing

core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java
puzpuzpuz: Addressed in 00e1f192c39edfc4776b2d1ba2d67f1aedccc5ae

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 20 / 20 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|20|20|100.00%|

PR #7140, opened 2026-05-22
perf(sql): speed up parallel twap() and sparkline()

mtopolnik:
Relates to #7123

Accompanying docs change: https://github.com/questdb/documentation/pull/452

## Benchmarks master vs. this PR

The common parallel and concurrent paths (10K, 16, 16×8) speed up 2-3×. The high-cardinality memory-stress scenario (1M groups, ~10 obs/group) regresses by ~15% — the per-group state widening without enough observations per group to amortize the new merge step.

```
master

scenario                                              mean          p50          p90  NATIVE_FAST_MAP   NATIVE_GROUP_BY_FN
---------------------------------------------------------------------------------------------------------------------------
1M groups, ~10 obs/group (memory-stress)            788 ms       774 ms       853 ms              0 B              3.97 GB
10K groups, ~1K obs/group                           305 ms       301 ms       405 ms              0 B              1.92 GB
16 groups, ~625K obs/group (merge-heavy)            254 ms       250 ms       274 ms              0 B              1.13 GB
16 groups x 8 concurrent (cross-query stealing)   1,358 ms     1,329 ms     1,771 ms              0 B              8.34 GB


twap-optimize

scenario                                              mean          p50          p90  NATIVE_FAST_MAP   NATIVE_GROUP_BY_FN
---------------------------------------------------------------------------------------------------------------------------
1M groups, ~10 obs/group (memory-stress)            886 ms       885 ms       902 ms         704.5 MB              3.59 GB
10K groups, ~1K obs/group                           143 ms       143 ms       147 ms          10.0 MB              1.19 GB
16 groups, ~625K obs/group (merge-heavy)            144 ms       139 ms       181 ms           2.5 MB              1.12 GB
16 groups x 8 concurrent (cross-query stealing)     457 ms       439 ms       530 ms          16.5 MB              6.26 GB

```

## What changed

This branch reworks parallel and serial execution of the `twap()` and `sparkline()` aggregates and makes one of their preconditions explicit.

**Per-frame batch bounds (3b8130ddfb).** The parallel aggregates buffer per-group observations and sort them at merge and read time. The sort previously recovered run boundaries by scanning for a key decrease, which conflated page frames whose key ranges ascended across a gap and forced an element-wise merge. `computeNext` now records the exact per-frame batch boundary (the page frame is read from `rowId >>> 44`) in a lazily allocated descriptor buffer, so the sort permutes whole batches with one bulk copy each and leaves an already-ordered buffer untouched.

**Consecutive-frame batches (03062356f5).** A descriptor was recorded on every page-frame change, even when frames arrived consecutively and in order. Such a buffer is already one sorted run, so the descriptors were pure overhead. `computeNext` now extends the current batch when the new frame is the immediate successor of the previous one, recording a boundary only at a gap or an out-of-order frame. A group whose frames arrive as one consecutive run keeps its descriptor buffer unallocated.

**Hot-path tuning (c2f38fb838).** A JFR profile of the 1M-groups scenario flagged three small wins in the merge step that the descriptor approach amplified per group: `sortByFirstKey` runs an insertion sort for `n <= 16` (the common case during the reduce phase, where each group has a handful of batches) instead of the bottom-up mergesort's per-level overhead; `gatherBatches` pre-grows the scratch list once via `setPos`/`setQuick` and carries the previous descriptor's `end` forward as the next batch's `start`; `merge()` folds the entry buffer and (optional) descriptor buffer into a single `allocator.malloc` call.

**Designated-timestamp requirement (6be6e8a161).** The batch model relies on each page frame's observations already being sorted by the timestamp argument, which holds only for the table's designated timestamp on a forward scan. `twap()` now validates this at compile time.

## Impact

- On the common parallel and concurrent paths, the sort permutes whole batches with one bulk copy each instead of merging element-wise, and a slot that sees frames in order allocates no descriptor buffer at all. The merge step also skips the per-call auxiliary buffer the old algorithm allocated.
- Per-group state widens from 3 longs to 7 (24 → 56 bytes in `NATIVE_FAST_MAP`) to carry the descriptor pointer, count, capacity, and last frame index. On a high-cardinality query where each group has too few observations to amortize the per-batch setup, this shows up as a slowdown (1M groups × ~10 obs/group: 886 ms vs. master's 788 ms). `NATIVE_GROUP_BY_FUNCTION` drops on the same scenario (3.59 GB vs. 3.97 GB) because the new merge skips the auxiliary buffer master allocated per call.
- `twap()` no longer accepts a non-designated timestamp argument. This is a behavior change: a query that previously ran (and could return an incorrect result) now fails at compile time with a clear error.
- `sparkline()` keys on row id, so the timestamp restriction does not apply to it.

## Test plan

- `mvn -pl core -Dtest='Twap*,Sparkline*,SortedRunsMergeTest' test` passes: the new rejection, the parallel keyed/sharded reduce paths, the existing designated-timestamp cases, and the `sparkline()` suite.
- `TwapGroupByFunctionTest` drives `computeNext` directly, asserting the descriptor layout for consecutive, gapped, and out-of-order frames.
- `SortedRunsMergeTest` exercises the batch-descriptor sort helper, including the insertion-sort small-`n` path.
- `TwapUnsortedRunReproTest` runs the keyed and non-keyed reduce paths under heavy cross-query work stealing (8 threads × 30 iterations × 5000 rows), asserting `twap() == 2500.0` exactly — any deviation would mean the compaction step failed to restore key order. Locks in the fix for #7123.


bluestreak01:
@mtopolnik — review at level 2 (`/review-pr 7140 --level 2`). The 8 parallel review agents I dispatched all returned HTTP 529 (API overload), so this was completed inline against the diff and source. Some breadth was sacrificed; a level-3 rerun may surface more.

## Critical

None confirmed against source.

## Moderate

### M1. `validateTimestampArg` rejects legal wrappers around the designated timestamp
`TwapGroupByFunction.java:413-416` requires `tsFunc instanceof ColumnFunction cf`. `ColumnFunction.unwrap(Function)` (in `ColumnFunction.java:43-48`) already exists to peel `MemoizerFunction` wrappers the code generator inserts when an alias is referenced more than once. Memoization is currently applied to top-level projection functions only (`SqlCodeGenerator.java:9189`), so aggregate-arg leaves aren't wrapped today — but the rule is brittle: any future change that wraps an aggregate's arg (cast, memoizer, type-conversion) will silently turn a previously-valid `twap(price, ts)` into a compile error.

Suggested fix: replace `tsFunc instanceof ColumnFunction cf` with `ColumnFunction.unwrap(tsFunc)` and check `cf != null && cf.getColumnIndex() == designatedTimestampIndex`. The helper exists for exactly this purpose.

### M2. Imports out of alphabetical order
`TwapGroupByFunction.java:37-38` and `SparklineGroupByFunction.java:40-41` both have:
```java
import io.questdb.griffin.engine.groupby.SortedRunsMerge;
import io.questdb.griffin.engine.groupby.GroupByAllocator;
```
`G` sorts before `S`, so `GroupByAllocator` must come first.

### M3. `SortedRunsMerge` private static methods out of alphabetical order
`SortedRunsMerge.java:419` places `insertionSortByFirstKey` at the end, after `sortByFirstKey:387`. Alphabetically `insertionSortByFirstKey` (`in…`) must come before `isAscending` (`is…`) at line 324. Current order: `copyRecord, emit, gatherBatches, isAscending, mergeRuns, sortByFirstKey, insertionSortByFirstKey`. Move `insertionSortByFirstKey` between `gatherBatches` and `isAscending`.

### M4. Stale `runScratch` field name after semantic rename
`TwapGroupByFunction.java:103` and `SparklineGroupByFunction.java:97` declare `private final LongList runScratch`. The PR rewrote `SortedRunsMerge` to operate on "batch records" — its public API now uses `LongList scratch` and `RUN_DESC_LONGS` was renamed to `RECORD_LONGS`. The field name still says `run`. Rename to `batchScratch` (or `scratch`) to match.

### M5. Dead `@Ignore`d test
`TwapGroupByFunctionFactoryTest.java:101-122` adds `@Ignore("non-designated timestamp not currently supported for TWAP")` to `testTwapNullTimestamp`. The PR description states this is a deliberate permanent behavior change. An `@Ignore`d test that will never be unignored is dead code (plus the unused `org.junit.Ignore` import). Delete the test; `testTwapRejectsNonDesignatedTimestamp` already covers the rejected case.

### M6. Test coverage gap: mergesort path never exercised
`SortedRunsMergeTest.java` covers `n` up to 4 batches. `SortedRunsMerge.sortByFirstKey:391` switches to bottom-up mergesort only when `n > INSERTION_SORT_THRESHOLD (16)`. Every test hits the insertion-sort branch; `mergeRuns` and the ping-pong have **zero unit-test coverage**. Add tests with `n = 17` and `n = 32` batches, including out-of-order arrival and (firstKey, lastKey) ties.

### M7. Test coverage gap: descriptor-buffer growth via `compactInPlace`
`testAppendBatchStartLazyAllocationAndGrowth` covers descriptor growth from `appendBatchStart`, but no `compactInPlace` test exercises a buffer with more than 8 batches (`DESC_INITIAL_CAPACITY`). Add a `compactInPlace` test where descriptors have grown past the initial capacity (e.g., 12 batches).

### M8. Test coverage gap: NULL-prefix recovery in `computeNext`
`TwapGroupByFunctionTest.java` has no test for the `count <= 0` recovery branch (`TwapGroupByFunction.java:159-172`), which fires when `computeFirst` saw a NULL and a later `computeNext` brings the first valid value. The branch re-initialises `lastFrameId` from a frame index different from `setNull`'s starting value of 0. Non-obvious and untested.

### M9. `merge()` sets descriptor capacity equal to count
`TwapGroupByFunction.java:379-380` and `SparklineGroupByFunction.java:397-398`:
```java
destValue.putLong(valueIndex + 4, mergedDescPtr != 0 ? mergedBatches : 0);  // descCount
destValue.putLong(valueIndex + 5, mergedDescPtr != 0 ? mergedBatches : 0);  // descCapacity
```
After merge, `descCapacity == descCount`. If any future code path calls `appendBatchStart` on a post-merge MapValue, the very first append immediately triggers a `realloc`. Today the parallel reduce path doesn't add rows after merge, so this never fires — latent contract that will silently bite a future change. Either grow capacity proactively, or `assert` that `appendBatchStart` is never called on a merged value.

### M10. PR body lacks top-level issue cross-reference
The test-plan section says "Locks in the fix for #7123" but the PR body has no top-of-body reference. This PR isn't the fix itself, so strict `Fixes #` isn't required — but a top-level `Relates to #7123` makes the relationship discoverable from the issue.

## Minor

- **m1.** `lastFrameId = 0` set by `setNull` plus the `count <= 0` re-init branch — correct but a one-line comment in `setNull` would save the next reader a trace.
- **m2.** The "consecutive frame" heuristic comment (`TwapGroupByFunction.java:173-178`) should cite `validateTimestampArg`/`validateScanDirection` as the compile-time guard the runtime invariant depends on, so a future reader doesn't try to relax the validation.
- **m3.** `compactInPlace`'s aux malloc deliberately lives below the `isAscending` early-return (`SortedRunsMerge.java:191-196`); a code comment ("aux deliberately allocated after the early-return check") would prevent a future refactor from leaking it.
- **m4.** `SortedRunsMerge.appendBatchStart` hardcodes the MapValue layout (descIndex+1 = count, descIndex+2 = capacity). Style-only: consider duplicating it as a private helper in the two callers, or taking three explicit slot offsets.
- **m5.** `INSERTION_SORT_THRESHOLD = 16` is documented as JFR-tuned but has no benchmark guard — a comment naming the JFR profile would help future maintainers retune.

## Downgraded (false positives explored during verification)

- **Sort-factory false-positive validation.** Initially suspected `factory.getScanDirection() == FORWARD` could be true for sort factories sorting on a non-timestamp column. Dismissed: `SqlCodeGenerator.java:7350-7356` builds `orderedMetadata` via `copyOfSansTimestamp` when the first sort column isn't a TIMESTAMP, so `validateTimestampArg`'s `designatedTimestampIndex < 0` check rejects the query before the FORWARD-reporting sort factory can mislead anyone.

- **MemoizerFunction wrapping the ts arg.** Dismissed (with M1's caveat): `SqlCodeGenerator.java:9189`'s memoization loop targets top-level projection functions only, not aggregate-argument leaves. M1 still recommended for future-proofing.

- **`rowId >>> 44` collision between parallel and serial paths.** Dismissed: in the serial path `frameId` is always 0, no descriptors recorded, `compactInPlace` early-returns on `descPtr == 0`. Degenerates to a no-op.

- **Allocator OOM leaving half-state.** Dismissed: `allocator.malloc` throws `CairoException`, propagates out, cursor closes, allocator resets.

- **Descriptor pointer invalidation on `realloc` against an embedded slice.** Dismissed: `appendBatchStart` is never invoked on merged MapValues (parallel reduce is terminal). Latent risk noted as M9.

## Summary

**Verdict:** request changes. None of the moderate findings are blocking individually, but M1 (validation strictness), M3 (member ordering), and M5 (dead code) violate CLAUDE.md conventions and should land before merge. M6/M7/M8 plug real test-coverage gaps.

Tradeoffs the PR honestly discloses (15% high-cardinality regression; twap() breaking change rejecting non-designated-timestamp args) are real and documented.

**Stats:** ~10 verified findings, 5 false positives dropped during verification. All findings are in-diff; the cross-context callsite walk (intended Agent 9 work) confirmed the public-API blast radius is narrow — only `assembleGroupByFunctions`'s 6 call sites in `SqlCodeGenerator.java`, all updated in-diff.

If you want the deeper level-3 pass (per-finding verification agents, full cross-context walk, fresh-context adversarial agent), retry once the API overload clears.

— *Generated with [Claude Code](https://claude.com/claude-code)*

bluestreak01:
@mtopolnik — level-2 review (8 parallel agents: correctness, concurrency, performance, resources, tests, code-quality, metadata, adversarial-performance; single batched verification pass). This re-ran the review that earlier hit API overload, so it supersedes the inline-only run above.

## :rotating_light: Blocking — M1: `sparkline()` has no parallel-contention test

**The headline issue, and it's a process one: every parallel-capable aggregate must ship a test that forces the parallel path. `sparkline()` does not.**

`sparkline()` received the *identical* descriptor-sort rework as `twap()` — same `appendBatchStart` (`SparklineGroupByFunction.java:224`), same `compactInPlace` (`:517`), same `compactInto` (`:387`), same 7-slot `MapValue`, same `rowId >>> 44` batch logic. But its test coverage of that rework is **zero at the integration level**:

- `SparklineGroupByFunctionFactoryTest` contains **no** `CAIRO_SQL_PARALLEL_WORK_STEALING_THRESHOLD`, no `CAIRO_SQL_PAGE_FRAME_MAX_ROWS`, no `CyclicBarrier`.
- Its only two parallel tests (`testParallelCorrectness`, `testParallelChunky`) assert via `assertSqlCursors(sql, sql)` — the query compared against *itself*. That cannot catch a systematically-wrong-but-stable render.
- With the default 1M-row page frames, a slot never sees out-of-order frames, so the reordering path that this whole PR introduces is never reached.

Contrast `twap()`, which is tested properly: `TwapUnsortedRunReproTest` sets `PAGE_FRAME_MAX_ROWS=50`, `WORK_STEALING_THRESHOLD=1`, runs 8 threads through a `CyclicBarrier`, and asserts an **exact, independently-known** value (`twap() == 2500.0`). A descriptor-sort regression in sparkline's `getVarcharA → renderForPtr → compactInPlace` wiring would render a deterministically-wrong sparkline that **every current test passes**. The shared `SortedRunsMergeTest` covers the primitive, but not sparkline's wiring to it.

**Please add a `SparklineUnsortedRunReproTest`** mirroring the twap repro: `PAGE_FRAME_MAX_ROWS=50`, `WORK_STEALING_THRESHOLD=1`, multi-thread + `CyclicBarrier`, on a dataset whose correct sparkline is known exactly (e.g. monotonically increasing values ⇒ output must be `▁▂▃▄▅▆▇█`), asserting the exact rendered string on every iteration — plus a keyed (`GROUP BY` + low `SHARDING_THRESHOLD`) variant and a duplicate-rowId-across-frame-boundary variant to hit the `(firstKey, lastKey)` tie path. The principle: if a function declares `supportsParallelism()`, it ships a contention test with an exact oracle, not a self-comparison.

---

The rest of the review. This is a follow-up **optimization** of code whose correctness bug (#7123) already landed in #7126 — "Relates to #7123" is the correct verb (this PR doesn't touch `array_agg`, so it isn't the closer). The load-bearing assumption checks out end-to-end: `rowId >>> 44` is the page-frame sequence index (`PageFrameMemoryRecord.getRowId()` = `Rows.toRowID(frameIndex, rowIndex)`; both reduce paths pass `baseRowId + r`), monotonic in scan order, and `SortedRunsMerge` is reachable only from `twap()`/`sparkline()`. The algorithm is sound.

### Critical

**None.** No correctness, concurrency, or resource-safety bug survived verification. The highest-risk out-of-diff investigations came back clean:

- **Validation proxy is safe.** `getScanDirection()` defaults to `FORWARD`, but the parallel descriptor path runs only over page-frame factories (raw partition storage, always designated-ts-ordered). Derived/reordering factories (LATEST ON, joins, sorts) fall to the serial path, which uses a monotonic `rowId++` ⇒ single batch ⇒ compaction is a no-op. `twap()`'s extra guard (`timestampIndex >= 0 && unwrap(tsFunc).getColumnIndex() == designatedTimestamp`) closes the rest.
- **No post-merge heap corruption.** `merge()` embeds the descriptor buffer inside the entry allocation and stamps `lastFrameId = -1`; `computeNext` is never called on a merged value (reduce → merge → read are barrier-ordered); the assert backstops it under `-ea`.
- **No race.** `isThreadSafe()==false` ⇒ per-slot instances; `scratch`, render caches, and in-place compaction are slot-confined under `PerWorkerLocks`, with merge/read separated by worker-pool barriers.
- **No leak / double-free.** Every `malloc` is balanced or reclaimed at cursor close; the embedded descriptor slice is never `free`/`realloc`'d; validation-failure cleanup closes the function via the `catch` in `assembleGroupByFunctions`.

### Moderate

**M2 — The documented ~15% high-cardinality regression is per-group-state width, and it is partially reducible.** Per-group `MapValue` widened 3→7 longs (24→56 B). `entryCap` (+2) is read only in `computeNext` growth; `descCap` (+5) only by `appendBatchStart`; for the dominant high-card group (~10 obs, in-order frames) `descPtr` stays 0 so those slots carry constant zero yet still widen every probe-loop value. Packing `descCount`+`descCap` into one long (two ints), and/or deriving `lastFrameId` from the last entry for sparkline (where the key *is* the rowId: `Unsafe.getLong(ptr+(count-1)*16) >>> 44`), would claw back the regression the PR itself reports. `entryCap` is genuinely needed during growth, so this is an opportunity, not a defect.

**M3 — `compactInto` re-sorts the full concatenation every merge, ignoring that the accumulator is already sorted.** The merge driver (`GroupByShardingContext.mergeOwnerMap`/`mergeShard`) accumulates sequentially: `destMap.merge(srcMap_i)` repeatedly. After the first merge, `dest`'s batch list is already `(firstKey,lastKey)`-sorted (since `emit` writes batches+descriptors in sorted order), yet `compactInto` (`SortedRunsMerge.java:234-239`) unconditionally calls `sortByFirstKey` over all batches and — unlike `compactInPlace` — has **no `isAscending` fast path**. A guarded 2-way merge (`mergeRuns` already exists) of the two sorted runs is `O(batchesA+batchesB)` vs `O(n log n)`. Biggest win on the low-cardinality/merge-heavy and concurrent paths. Correct today, just not optimal.

**M4 — PR body states the regression as two different numbers.** Benchmark table: `1M groups ... 895 ms`. Impact prose: "1M groups × ~10 obs/group: **904 ms** vs. master's 788 ms." Reconcile 895 vs 904 — matters precisely because the honest disclosure is otherwise a strength.

**M5 — Enterprise CI is red and must be investigated before merge.** `questdb-butler` reports the enterprise suite failed for tip `e12164cdca` (Azure build 236961). Plausible blast radius: the `twap()` non-designated-timestamp rejection is a compile-time breaking change an enterprise test may rely on. Please confirm the cause or prove it unrelated with evidence.

### Minor

- **Import ordering** — `TwapGroupByFunction.java:37-38` and `SparklineGroupByFunction.java:40-41` both list `SortedRunsMerge` before `GroupByAllocator`; `G` precedes `S`. Swap.
- **Dead import** — `TwapGroupByFunctionFactoryTest.java:30` imports `org.junit.Ignore` with no usage (orphaned when `testTwapNullTimestamp` was deleted). Remove.
- **TWAP duplicate-timestamp tie-break non-determinism** — if multiple *entire* page frames each hold a single identical-microsecond timestamp K and arrive out of scan order, the stable `(firstKey,lastKey)` sort keeps them in arrival (not scan) order, so the price carried into the next distinct-ts interval can differ from the serial result. Mechanism real; reachability essentially nil; TWAP over duplicate timestamps is inherently ambiguous. Worth a one-line code comment, not a fix.
- **Recovery test asserts the field, not the consequence** — `testNullPrefixRecoveryReinitializesLastFrameId` checks `lastFrameId==5` but doesn't drive a follow-on same-frame `computeNext` to prove no phantom batch boundary is recorded.
- **Dead-from-SQL branch now untested** — removing `testTwapNullTimestamp` leaves the `ts == Numbers.LONG_NULL` skip unreachable from SQL and uncovered. Note it as defensive, or keep a white-box unit test driving `computeNext` with `LONG_NULL`.
- **Allocation asymmetry** — `Twap.computeNext` grows via `malloc`+`Vect.memcpy` while `Sparkline.computeNext` uses `realloc`. Not a leak; `realloc` for symmetry would lower peak footprint on the regressing path.
- **`getDouble` computes `priceSum` unconditionally** but uses it only in the `totalDuration==0` fallback — one wasted FP add per observation; compute it lazily.

### Downgraded (false positives removed during verification)

- **`SortedRunsMerge` method ordering** (flagged in the earlier inline run) — REFUTED. `copyRecord, emit, gatherBatches, insertionSortByFirstKey, isAscending, mergeRuns, sortByFirstKey` is correctly alphabetical (`insertion…` < `is…`, 'n' < 's').
- **`validateTimestampArg` rejects wrapped ts args** — already fixed: current code uses `ColumnFunction.unwrap(tsFunc)`.
- **Validation proxy admits reordered factories; post-merge `computeNext` corruption; `scratch`/render-cache race; aux leak on throw** — all cleared (see Critical).

## Summary

**Approve with changes.** Blocking: add the sparkline parallel-contention test (**M1**), reconcile the benchmark number (M4), resolve the enterprise CI red (M5), and fix the import/dead-import nits. M2/M3 are real performance opportunities against the disclosed regression but are enhancements, not blockers. The algorithm is correct and the common-path 2-3× win is real.

**Tradeoffs (honestly disclosed):** ~15% slowdown on the 1M-group/~10-obs path from the wider per-group state, and `twap()` now rejects non-designated-timestamp arguments at compile time (a breaking change that previously could return wrong results).

~7 findings verified, 5 false positives dropped. All findings are in-diff; the out-of-diff blast radius (the 6 `assembleGroupByFunctions` callsites, the reduce/merge lifecycle, the two `SortedRunsMerge` callers) was walked and cleared.

— *Generated with [Claude Code](https://claude.com/claude-code)*

mtopolnik:
Key cross-context checks came back clean: `SortedRunsMerge`'s signature changes touch only twap/sparkline (even on master — `array_agg` never used it), the cast passthrough follows established precedent with no double-close, and the `assembleGroupByFunctions` signature change is handled at all 6 callsites. Let me write the final report.

---

# Review: PR #7140 — `perf(sql): speed up parallel twap() and sparkline()`

**Reviewing at level 2.** (Level 3 exists for a full 11-agent, per-finding pass — reserve it for on-disk-format / replication / JNI changes.) Ran 8 parallel agents (correctness, concurrency, performance, resources, tests, code-quality, metadata, adversarial-performance), then verified each finding against source. **24 draft findings → 16 verified, 8 dropped as false positives.** In-diff: 16. **Out-of-diff confirmed bugs: 0** — and that zero is a verified conclusion, not an underrun (cross-context callsites checked below).

This is a high-quality PR. The core algorithm is provably correct, leak-free, and race-free under the actual parallel execution model. No blocking issues. The findings are one corroborated performance opportunity, one test/documentation gap on a concurrency-sensitive path, and minor nits.

## Critical
None.

## Moderate

**M1 — `merge()`/`compactInto` has no ordered-concatenation fast path; this is the likely driver of the documented ~15% regression. (in-diff, performance)**
`SortedRunsMerge.compactInto` (`SortedRunsMerge.java:226`) always runs `gatherBatches(A) + gatherBatches(B) + sortByFirstKey + emit`, and `merge()` (`TwapGroupByFunction.java:361`, `SparklineGroupByFunction.java:381`) always allocates a descriptor buffer (`mergedBatches = destBatches + srcBatches ≥ 2`) — even when both inputs are single sorted runs that concatenate cleanly. Two independent agents (perf + adversarial-perf) converged on this. Because page frames cover disjoint timestamp ranges, two slots' partials for the same group are *always* cleanly orderable (in one direction or the other), so this is precisely the 1M-groups scenario. Worse, the 2-batch result is marked multi-batch *permanently*, so batch count grows with merge depth and the cost compounds up the fold tree.

Fix: in `compactInto` (or `merge`), when both inputs are single-batch (`descPtrA==0 && descPtrB==0`) and orderable (`A.lastKey <= B.firstKey` or the reverse), emit two bulk `memcpy`s and set `mergedDescPtr = 0` — no descriptor allocation, no sort, result stays single-batch. Verified safe: `[A][B]` with `A.lastKey <= B.firstKey` is a single sorted run. This directly attacks the one regressing scenario the PR documents.

**M2 — The `sharedDependents` path is correct only under an undocumented single-thread-read invariant, and has no end-to-end parallel test. (in-diff, concurrency/tests)**
`setAllocator` shares the owner's *non-thread-safe* `FastGroupByAllocator` across all shared dependents (`TwapGroupByFunction.java:411`, `SparklineGroupByFunction.java:426`), and a sparkline dependent even `malloc`s render output from it. The concurrency agent verified this is safe *today* — shared cursors (`AsyncGroupBySharedCursor`) are consumed on the single query-driving thread — but:
- The in-code justification ("aux buffer is freed before return") argues buffer *lifetime*, not *concurrency*, and would mislead a maintainer into thinking concurrent reads are safe. They are not: two concurrent `malloc`s on this allocator corrupt the heap.
- Both shared-dependent tests (`TwapUnsortedRunReproTest.java:424`, `SparklineGroupByFunctionFactoryTest.java:699`) call `setParallelGroupByEnabled(false)` and assert only the cache-reset field via reflection. **No test runs twap/sparkline as a shared dependent through the async/parallel shared-cursor path** — the exact path this mechanism (added to fix a real NPE) exists to support.

Recommend: (a) a comment on `setAllocator`/`initSharedFrom` stating the hard same-thread-read requirement; (b) an end-to-end test — a multiply-referenced keyed group-by subquery with parallel group-by enabled and a low sharding threshold — asserting exact known twap/sparkline values.

**M3 — Per-group state could shrink 56B→40B to claw back the high-cardinality regression. (in-diff, performance/enhancement)**
The MapValue widened 3→7 longs. Two slots are dead weight outside `computeNext`: `descCapacity` (+5, used only by `appendBatchStart`) and `lastFrameId` (+6, overwritten with the `-1` sentinel at merge). `descCapacity` is safely derivable (`Numbers.ceilPow2(descCount)`); `lastFrameId`/`descCount` are small ints that pack two-to-a-long. At 1M groups the value width dominates (the PR's own `NATIVE_FAST_MAP` jumps 0→704MB), so this is the principled fix for the regression the PR documents. Packing adds shift/mask to the per-row boundary check, so measure — but dropping `descCapacity` is unconditionally safe (cold path only).

## Minor

- **Test gaps** (none blocking, all in-diff): parallel/merge path never exercises sparkline `renderSubsampled` (`width < size`); no twap test for a NULL *timestamp* (only NULL price is tested, though `ts == Numbers.LONG_NULL` is a distinct guard at `TwapGroupByFunction.java:140,162`); twap `merge()` lacks the direct white-box assertions sparkline has (empty-dest-copies-src, counts/descriptors); no sparkline-over-`LatestByLight` rejection test (twap has one); the `compactInto` mergesort branch (n>16) is untested (insertion-sort branch only). Eight twap parallel tests self-compare `assertSqlCursors(sql, sql)`, which can't catch a stable-but-wrong result — mitigated by the exact-value `TwapUnsortedRunReproTest`.
- **`TwapGroupByFunction.java:266`** — comment uses a Unicode em-dash (`—`); the rest of the PR uses ASCII `--`. Make it consistent.
- **`benchmarks/.../TwapBenchmark.java:45`** — unused import `java.util.concurrent.TimeUnit`.
- **twap `computeNext` growth** uses `malloc + memcpy + abandon` while sparkline uses `realloc` (`TwapGroupByFunction.java:200`). **Pre-existing** (master had the same asymmetry — not introduced here), but `FastGroupByAllocator.realloc` has a copy-free bump-pointer fast path, so switching twap to `realloc` is never worse and reduces abandoned memory. Free win.
- **`SortedRunsMerge` reads `lastKey` per batch even for sparkline**, whose keys are unique rowIds that can never tie — so the `lastKey` tie-break is dead work there. A sparkline-specific 3-long record would cut gather cache-line touches and scratch traffic ~25%. Micro-opt.
- **Removed runtime out-of-order backstop** (commit `45a74be606`): production now relies entirely on the compile-time guards. This is a deliberate, documented decision and the commit states an audit of every `RecordCursorFactory` found no other base that reports `FORWARD` while delivering out-of-order rows (the `LatestByLight` fix patches the one known liar). Residual risk: `getScanDirection()` defaults to `FORWARD`, so any *future* reordering factory that keeps the timestamp index is a silent-wrong-result vector. Worth a one-line note at the `isBaseTimestampAscending` predicate flagging this contract for future factory authors.

## Downgraded (false positives, verified against source)
- **Resource leaks / double-frees** — none. `compactInPlace`'s `aux` is allocated only after both early-returns; `merge()`'s co-allocated block is exact-fit and never freed (best-effort allocator contract, reclaimed at close); the `GroupByUtils` error-path `outGroupByFunctions.clear()` correctly prevents the JOIN callsites' `Misc.freeObjList` from double-closing. (Agent 4, full trace.)
- **Concurrency races** — none. Per-slot instances, `PerWorkerLocks` mutual exclusion, and the merge barrier isolate all mutable state; reads are single-threaded post-barrier. (Agent 2.)
- **`CastTimestampToTimestamp` returning `arg` → double-close** — false. `args.clear()` on the success path doesn't free; the type constant is a no-op-close singleton; identical ownership to master; matches the precedent of every other `AbstractCastToXxx` identity cast.
- **Duplicate-timestamp-across-frame-boundary → wrong sort** — false. The `(firstKey, lastKey)` tie-break + disjoint-frame-interval invariant provably yields a non-decreasing concatenation.
- **`computeNext` after `merge()` corrupts the co-allocated descriptor buffer** — false. Structurally impossible (two-phase aggregation); guarded by the `-1` sentinel + assert.
- **`SortedRunsMerge` signature change breaks `array_agg`** — false. Only twap/sparkline use it, on this branch *and* master.
- **`maxValues` cap inconsistency (serial throws at accumulation, parallel at render)** — real but **pre-existing**, not introduced here.

## Summary
**Verdict: approve with suggestions.** No correctness, concurrency, or resource defects. The algorithm — recording exact per-frame batch boundaries so the sort permutes whole key-disjoint batches — is sound and well-documented, and the change is correctly gated behind a compile-time designated-timestamp-ascending precondition (including the `LatestByLight` `getScanDirection` fix that closes the one factory that misreported its order).

The PR is honest about its tradeoffs: it documents the ~15% high-cardinality regression prominently rather than hiding it. The two highest-value follow-ups both target that regression and the one undertested path: **M1** (ordered-concatenation fast path — likely recovers most of the regression, safe, cheap) and **M2** (end-to-end parallel shared-dependent test + a thread-affinity comment). Neither blocks merge. Metadata is compliant: Conventional-Commits title, end-user-facing description, level-headed tone presenting the regression with equal weight, and correct SQL/Performance/Compatibility labels (`Compatibility` for the new compile-time rejection of non-designated-timestamp `twap()`).

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=237505)
**Tested Commit:** `09a485dfe9c12981a08e4e9115ea8fc2f24aaf09`

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 336 / 338 (99.41%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/cast/CastTimestampToTimestampFunctionFactory.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SortedRunsMerge.java|152|153|99.35%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/TwapGroupByFunction.java|78|78|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/SparklineGroupByFunction.java|71|71|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByLightRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/GroupByUtils.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|19|19|100.00%|

mtopolnik:
# Review: PR #7140 — `perf(sql): speed up parallel twap() and sparkline()`

**Level 2** (full change-surface map; Agents 1–7 + 11; batched verification). 

This branch has matured well past the two prior review rounds visible on the PR — the dead `@Ignore` test, import/method ordering, the `runScratch` rename, the missing sparkline contention test, the mergesort-branch coverage, and the `Relates to #7123` reference are all already addressed. The four logical changes (per-frame batch bounds → whole-batch permutation sort; consecutive-frame batching; hot-path tuning; compile-time designated-timestamp gate) are sound. Three deep agents (correctness, concurrency, resources) cleared the load-bearing invariants after extensive tracing.

## Critical
**None.** No correctness, concurrency, or resource-safety defect survived verification.

The highest-risk investigations came back clean, with evidence:
- **(firstKey, lastKey) sort + concatenation invariant** is correct. A batch only spans *consecutive* frames (a gap/reorder forces a boundary), and the global ascending-timestamp scan makes batches key-disjoint except at a shared boundary key — so the lastKey tie-break orders the narrow all-firstKey batch ahead of the wider one, matching scan order. `isAscending`'s `firstKey >= prevLast` check never wrongly accepts an unsorted buffer.
- **Validation gate** closes the silent-reorder hole: a reorder factory either keeps the designated ts (and FORWARD then genuinely means ts-ascending) or drops it via `copyOfSansTimestamp` so `getTimestampIndex()` no longer matches. The `LatestByLightRecordCursorFactory` ts-strip is the correct fix for that one gotcha.
- **Same-thread-read invariant** holds: `dispatchAndAwait`'s done-latch barriers reduce from merge; `PerWorkerLocks` slot CAS binds each `merge()`/`scratch` to one thread; joins/projections consume primary and shared cursors serially on the result thread.
- **Double-close fix** (`GroupByUtils.java:370` `outGroupByFunctions.clear()`) is correct on every error path — functions are freed through `outerProjectionFunctions` in all 6 callsites, so the JOIN callers' `Misc.freeObjList` iterates an emptied list (no double-free), and non-JOIN callers don't leak.

## Moderate

**1. `compactInto` re-sorts the already-sorted accumulator on every merge (in-diff, `SortedRunsMerge.java:226-240`).**
The merge driver (`GroupByShardingContext.mergeOwnerMap`/`mergeShard`) folds sequentially: `destMap.merge(srcMap_i)` repeatedly into one `destValue`. After the first merge, `emit` has written `dest`'s batches in `(firstKey,lastKey)` sorted order, so `dest` is an already-sorted run. Yet `compactInto` unconditionally calls `sortByFirstKey` over *all* `batchesA + batchesB` records — unlike `compactInPlace`, it has no `isAscending` fast path and no 2-way merge of the two sorted runs. The existing `mergeRuns` already implements exactly the stable 2-way merge needed; sorting only `src`'s batches (guarded per-side by `isAscending`) then `mergeRuns` drops the sort term from `O(m log m)` to `O(batchesA + batchesB)`.

*Honest caveat (this is why it is Moderate, not Critical, and arguably Minor):* the sort operates on the small batch-*record* array, while `emit`'s entry memcpy (`O(countA+countB)`) dominates each merge — and that copy is unavoidable because `merge()` mallocs a fresh destination. The deeper cost is the `O(W²·entries)` re-copy inherent to the pairwise-merge contract, which this change does not touch. So the realizable speedup is small and should be profiled before investing; the value is mainly that the code stops ignoring structure it already has. Reuses existing code at low risk.

**2. PR body benchmark inconsistency: 886 ms vs 904 ms (release-facing).**
The benchmark table reports the 1M-groups branch mean as **886 ms** (p50 885, p90 902); the "Impact" prose says "**904 ms** vs. master's 788 ms". The "~15%" regression claim matches 904 (788→904 = +14.7%), not 886 (+12.4%). Per CLAUDE.md's accuracy/level-headed-tone rule, reconcile to one figure (other numbers — 3.59/3.97 GB, 24→56 bytes — check out). This is a doc fix, not a code issue.

## Minor

- **Per-group state width — partial claw-back of the disclosed ~15% regression.** `descCount` (+4) and `descCapacity` (+5) are both batch counts well under 2³¹ and could pack into one long (two ints via `Numbers.encodeLowHighInts`), narrowing the MapValue 56→48 bytes. `appendBatchStart` is the only writer and reads both together. `lastFrameId` (+6) is *not* reducible for twap (its key is the timestamp, not the rowId), so leave it. Low-risk, modest gain on the high-cardinality path the PR regresses.

- **`computeNext` realloc ordering defeats the in-place grow on the coincident-row case** (`TwapGroupByFunction.java:185-202`, same in Sparkline). When a row both starts a new batch (`appendBatchStart` mallocs/reallocs the descriptor buffer) *and* grows the entry buffer, the descriptor alloc runs first and moves the allocator tail, so the entry `realloc` misses `FastGroupByAllocator`'s in-place fast path and falls to malloc+memcpy. Growing entries *before* `appendBatchStart` is a free, safe reorder (the batch-start offset `count` is unchanged). Narrow applicability (mostly single-group/non-keyed in-order growth), so marginal.

- **Em-dash in a comment, `TwapGroupByFunction.java:263`** (`// All observations at the same timestamp — simple average`). The only non-ASCII in changed production code (excluding the license banner and `SPARK_CHARS` glyph data). CLAUDE.md's strict-ASCII rule scopes to log messages, so this is not a hard violation — but every other comment in the PR uses ASCII hyphens, so it's an inconsistent outlier. Replace with ` - `.

- **Test gaps (both low-risk, same code path covered elsewhere):**
  - sparkline's `validateScanDirection` is never exercised through the serial **SAMPLE BY FILL** callsite (`SqlCodeGenerator.java:7851`); twap has `testTwapRejectsSampleByFillOverNonTimestampOrderedBase`, sparkline has no counterpart.
  - `compactInto` with **empty src** (`countB==0`, non-empty A) is not unit-tested in `SortedRunsMergeTest` (only empty-dest). The `gatherBatches` `entryCount<=0` return is reached only from the A side at the unit level.

- **PR body nits:** "speed up 2-3×" overstates the 16-groups case (254→144 ms is 1.8×); the test plan's "Locks in the fix for #7123" sits in slight tension with the correct "Relates to #7123" (the fix landed in #7126); sparkline's compile-time backward-scan rejection is a behavior change not surfaced in the body (only "Compatibility" label covers it).

## Downgraded (false positives / non-issues dismissed after verification)

- **`supportsParallelism()` override in Twap (`:451`) is dead code** — FALSE. `TwapGroupByFunction` inherits conflicting defaults from `GroupByFunction` (false) and `BinaryFunction` (`getLeft()&&getRight()`); without the explicit override the call is ambiguous and won't compile. Same for Sparkline (`:469`).
- **`compactInto` re-sort is a major perf problem** — DOWNGRADED to the caveated Moderate above; the unavoidable entry copy dominates the sort, and the deeper `O(W²·entries)` re-copy is inherent to the pairwise merge contract.
- **CastTimestampToTimestamp identity-cast (return `arg` unwrapped) has a dangerous global blast radius** — FALSE. The branch returns `arg` only when `arg.getType() == target` exactly (type-preserving); "return arg for a no-op cast" is the pervasive idiom across cast factories; no double-close (the parser folds arg as one node). Only visible deltas: EXPLAIN renders the bare column, and `findColumnKeyIndex` now recognizes `cast(ts AS timestamp)` as a column key — both intended/more correct. All cast tests pass.
- **`merge()` post-state corrupts the allocator (descCapacity==descCount + embedded descriptor)** — FALSE. The `-1` sentinel in slot +6 plus the `computeNext` assert forbid post-merge appends, and the reduce→merge→read barrier ordering means no path calls `computeNext` after `merge`.
- **sparkline accepts a base with no designated timestamp** (`isBaseTimestampAscending(factory, -1)` returns true since `-1==-1`) — NOT A BUG. sparkline keys on rowId, so rowId-order rendering is deterministic regardless; matches master (which had no scan gate). Correctness-safe.
- **Concurrency races on `scratch`/render caches**, **compactInPlace aux-buffer leak**, **realloc abandons buffers unboundedly** — all FALSE per Agents 2/4: per-slot instances under `PerWorkerLocks`, no throwing call between aux malloc/free, and abandoned buffers are bounded (`O(workerCount × groupSize)`) and reclaimed at the bump allocator's wholesale close.
- **`renderSubsampled` divide-by-zero on an empty bucket** — FALSE. Reached only when `width < size` ⇒ `bucketSize > 1` strictly ⇒ `to > from` for every bucket.

## Summary

**Verdict: approve.** No blocking issues. This is a mature, well-tested optimization: the common parallel/concurrent paths speed up 2–3×, the high-cardinality regression (~12–15%) and the twap() breaking change are honestly disclosed, and the correctness/concurrency/resource invariants hold under adversarial tracing. The remaining items are non-blocking — one caveated algorithmic opportunity (`compactInto` 2-way merge), two small perf claw-backs, an em-dash, two minor test gaps, and a 886/904 benchmark-number fix in the PR body.

**Stats:** ~14 draft findings verified, ~8 dismissed as false positives. **All confirmed findings are in-diff** — the cross-context walk confirmed a narrow blast radius (`assembleGroupByFunctions`'s 6 callsites all updated in-diff; `SortedRunsMerge` reachable only from twap/sparkline; `CastTimestampToTimestamp` is global but the change is type-preserving and safe; `LatestByLightRecordCursorFactory` has one constructor caller). The one genuinely cross-cutting change (the cast factory) was verified safe rather than left unexamined.

**Tradeoffs the PR carries:** the disclosed high-cardinality slowdown from the 24→56-byte per-group state (partially reducible per Minor #1), and twap()/sparkline() now rejecting at compile time some queries that previously ran (twap with a non-designated timestamp could return wrong results; this is the intended fix).

bluestreak01 (approved):
Approving. The core batch-descriptor rewrite in `SortedRunsMerge` is correct (verified from source: `rowId >>> 44` is the page-frame index, frame indices are globally sequential in scan order, each frame is processed atomically by one slot, so batches are contiguous frame-index intervals and the (firstKey, lastKey) sort yields a globally sorted buffer). No correctness, concurrency, or resource issues found; the `outGroupByFunctions.clear()` double-free guard and the shared-allocator same-thread-read contract both hold. Headline 2-3x speedups are real.

Optional perf follow-ups (non-blocking, would offset the disclosed ~15% high-cardinality regression):
- Pack descCount/descCapacity/lastFrameId into one long (all fit in 63 bits), shrinking the per-group MapValue from 7 to 5 longs.
- Steal src's entry/descriptor pointers on the empty-destination first merge per group instead of malloc+memcpy (safe: buffers live in the GroupByAllocator and outlive srcMap).

Minor: the dead `ts == LONG_NULL` sub-clause in twap computeNext/computeFirst is now unreachable (designated timestamp is non-nullable) and could be dropped or annotated; the new LatestByTest comment referencing SCAN_DIRECTION_OTHER is stale (the impl strips the timestamp instead). The LatestByLight timestamp-strip is a correct, narrowly-scoped behavior change (silent-wrong ASOF/SAMPLE BY over a light subquery now errors) - the Compatibility label covers it.

PR #7141, opened 2026-05-22
fix(core): fix crash adding a covering index when the latest partition is Parquet

nwoolmer:
## What

`ALTER TABLE ... ALTER COLUMN <sym> ADD INDEX TYPE POSTING INCLUDE (...)`
threw a `NullPointerException` and suspended the table when the table's
most recent partition was stored as Parquet. Adding the same covering
index when only historic partitions were Parquet worked fine.

## Why

`TableWriter.writeIndex()` indexes historic partitions through
`indexHistoricPartitions()`, which checks `isPartitionParquet()` and routes
Parquet partitions to the Parquet-aware `indexParquetPartition()`. The last
partition went through `indexLastPartition()`, which had no such branch — it
indexed a Parquet last partition as if it had native column files, so the
covering `seal()` dereferenced a `FilesFacade` that was never set for the
Parquet path.

`indexLastPartition()` now detects a Parquet last partition and delegates to
`indexParquetPartition()`, the routine `indexHistoricPartitions()` already
uses. The change adds one branch; the native path is unchanged.

Non-WAL tables never reached this bug — `CONVERT PARTITION TO PARQUET` skips
the active partition, so a non-WAL table's last partition is always native.
The non-WAL test below pins that boundary.

## Test plan

- New `CoveringIndexTest.testAddPostingCoveringIndexWhenLastPartitionIsParquetWal` — reproduces the bug (table suspended) before the change, passes after.
- New `CoveringIndexTest.testAddPostingCoveringIndexNonWalActivePartitionCannotBeParquet` — asserts a non-WAL table's active partition stays native, so the covering index builds on the native path.
- `CoveringIndexTest`, `PostingIndexCriticalIssuesTest`, `AlterTableConvertPartitionTest`, `MatViewCoveringIndexTest` — 432 tests, all green.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -6491,6 +6491,24 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
     private void indexLastPartition(SymbolColumnIndexer indexer, CharSequence columnName, long columnNameTxn, int columnIndex, int indexValueBlockSize, byte indexType) {
         final int plen = path.size();
 
+        // A Parquet last partition must be indexed the same way historic Parquet
+        // partitions are (see indexHistoricPartitions): indexParquetPartition()
+        // reads column data through the Parquet decoder and wires up the covering
+        // sidecars. The native path below assumes native column files; for a
+        // Parquet partition the covering seal would dereference a null FilesFacade.
+        // Non-WAL tables cannot have a Parquet active partition (see
+        // convertPartitionNativeToParquet), so this only fires for WAL tables.
+        final int lastPartitionIndex = txWriter.getPartitionCount() - 1;
+        if (lastPartitionIndex >= 0 && txWriter.isPartitionParquet(lastPartitionIndex)) {
+            try {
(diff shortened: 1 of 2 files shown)


mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|6|6|100.00%|

bluestreak01 (approved):
Reviewed at level 2 — approve with nits.

Fix is correct: `indexLastPartition` was the only `writeIndex` branch missing the Parquet dispatch that `indexHistoricPartitions` already had. The new branch delegates to the same `indexParquetPartition` and wraps it in the same `try/finally indexer.releaseIndexWriter()` pattern. Path state, resource cleanup, rollback, and throw-safety all check out.

Non-blocking suggestions:
- Add the `SQL` label (change is triggered by ALTER TABLE ... ADD INDEX).
- Widen test coverage: a multi-partition layout (last-only Parquet, and mixed native+Parquet historic with last Parquet) and an ADD INDEX TYPE POSTING variant without INCLUDE would exercise the new branch beyond the minimal single-partition reproducer.
- Optional: a non-DOUBLE INCLUDE type (e.g. LONG or TIMESTAMP) to cover the per-type compression branches in PostingIndexWriter.

PR #7146, opened 2026-05-24
fix(core): fix indexed NULL lookups against Parquet partitions

bluestreak01:
## Summary

Indexed `WHERE col = null` could return zero rows from a Parquet partition whose SYMBOL column was entirely NULL in a row group. Three call sites that rebuild a covering index from a decoded Parquet chunk shared the same defect:

- `O3PartitionJob.updateParquetIndexes` runs after an O3 insert rewrites a Parquet partition.
- `TableWriter.indexParquetPartition` runs when `ALTER TABLE ALTER COLUMN ... ADD INDEX` covers a historic Parquet partition.
- `TableSnapshotRestore.rebuildTableFiles` runs during checkpoint restore for any Parquet partition that needs its bitmap index rebuilt.

The `_pm`-backed Rust decoder has a documented optimisation (see the `RowGroupBuffers` javadoc): when a column chunk's stats report `null_count == num_values` it skips materialising the buffer and returns `size = 0`. In all three sites the walk iterated zero bytes and wrote no index entries; in the O3 path, the subsequent `zeroColumnTopsAfterParquetRewrite` also set `columnTop = 0`, disabling `PostingIndexFwdReader`'s implicit-null synthesis. The net effect was an index that recognised no NULL rows for that column in that partition.

The fix detects the documented `size == 0` convention and emits explicit null index entries for every row the row group covers. The null-fill loop is extracted into `BitmapIndexUtils.addNullEntries(IndexWriter, startRowId, endRowIdExclusive)` so the convention is named in one place and the next caller that touches this code path cannot forget it. No Rust-side change.

## Tradeoffs

- Adds one extra branch in the per-row-group loop at each call site. Each check is a single comparison and only takes the new path on chunks the decoder fast-pathed away, so runtime cost is negligible.
- Other consumers of `RowGroupBuffers` outside these three sites have not been audited.

## Tests

- `WalWriterFuzzTest.testConvertPartitionToParquetAllNullChunkIndex` was the original CI-seeded fuzz reproducer. It depends on the shared worker pool schedule to land an all-null SYMBOL chunk in a rewritten row group, and on `FuzzRunner.checkIndexRandomValueScan` happening to pick a NULL row from that partition for its random assertion. On master with these seeds it fails reliably; a future schedule shift or different random pick could mask a re-regression. Kept for incidental coverage.
- `ParquetTest.testIndexAllNullSymbolChunkOnParquetPartitionAfterO3Rewrite` constructs the all-null chunk explicitly (insert NULLs with the column already indexed, convert to Parquet, O3-insert another NULL row to force the rewrite) and asserts on `WHERE id = null` directly. The indexed query plan is pinned to `Index forward scan on: id  filter: id=0` so a silently-degraded full scan cannot mask a broken index by returning the right answer for the wrong reason. Confirmed to fail without the `O3PartitionJob` fix.
- `ParquetTest.testIndexAllNullSymbolChunkOnParquetPartition` covers the `ALTER TABLE ADD INDEX` path with the same explicit-chunk + pinned-plan pattern. Confirmed to fail without the `TableWriter.indexParquetPartition` fix.
- `CheckpointTest.testCheckpointRestoreRebuildsBitmapIndexesOnParquetAllNullChunk` invokes `TableSnapshotRestore.rebuildTableFiles` directly against a Parquet partition whose only SYMBOL chunk is all-NULL, and asserts `count() WHERE sym = null` returns the inserted rows. Confirmed to fail without the `TableSnapshotRestore` fix.

## Test plan

- mvn -pl core -Dtest='ParquetTest#testIndexAllNullSymbolChunkOnParquetPartition*' test
- mvn -pl core -Dtest='CheckpointTest#testCheckpointRestoreRebuildsBitmapIndexesOnParquetAllNullChunk' test
- mvn -pl core -Dtest='WalWriterFuzzTest#testConvertPartitionToParquet*' test

### DIFF
--- a/core/src/main/java/io/questdb/cairo/O3PartitionJob.java
+++ b/core/src/main/java/io/questdb/cairo/O3PartitionJob.java
@@ -25,9 +25,9 @@
 package io.questdb.cairo;
 
 import io.questdb.MessageBus;
+import io.questdb.cairo.idx.BitmapIndexUtils;
 import io.questdb.cairo.idx.IndexWriter;
 import io.questdb.cairo.sql.RecordMetadata;
-import io.questdb.cairo.sql.SymbolTable;
 import io.questdb.cairo.sql.TableRecordMetadata;
 import io.questdb.cairo.vm.api.MemoryCR;
 import io.questdb.cairo.vm.api.MemoryMA;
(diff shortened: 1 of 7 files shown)


mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/BitmapIndexUtils.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableSnapshotRestore.java|4|4|100.00%|

PR #7147, opened 2026-05-24
fix(core): fix covering index returning NULL values on Parquet partitions

bluestreak01:
## Summary

- Adding a covering index (`ALTER TABLE ... ADD INDEX TYPE POSTING INCLUDE (...)`) to a Parquet partition wrote NULL values for all covered columns because `seal()` tried to read source data from native `.d` files which do not exist in Parquet partitions
- `indexParquetPartition()` now decodes covered columns directly from the Parquet file and passes their memory addresses to the address-based `configureCovering()` overload, handling both fixed-size (DOUBLE, INT, TIMESTAMP) and variable-size (VARCHAR, STRING) column types

## Root cause

`indexParquetPartition()` called `configureCoveringIfNeeded()` which wires the `PostingIndexWriter` with column file paths. During `seal()` → `writeSidecarsPerColumn()` → `mapCoveredColumn()`, the writer calls `TableUtils.dFile()` to locate the source column data. In a Parquet partition, no such file exists — `ff.openRO()` returns -1, the address stays 0, and `writeSidecarFixedStrideForColumn` writes NULL sentinels for every row.

The posting index itself (row IDs) was correct (decoded from Parquet), so `count(*)` queries worked — only the covered column values were wrong.

## Test plan

- New test `testAddPostingCoveringIndexOnParquetPartitionFixedSize` verifies DOUBLE, INT, and TIMESTAMP covered values from a Parquet partition
- New test `testAddPostingCoveringIndexOnParquetPartitionVarSize` verifies VARCHAR + DOUBLE covered values from a Parquet partition
- All 353 existing CoveringIndexTest tests pass
- PostingIndexCriticalIssuesTest, PostingIndexStressTest, IndexBuilderTest, AlterTableConvertPartitionTest all pass

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -10,6 +10,14 @@ with native C/C++ libraries for performance-critical operations. It features
 column-oriented storage, SIMD-accelerated vector execution, and specialized
 time-series SQL extensions.
 
+## Code analysis rules
+
+When analyzing code (validating claims, tracing bugs, assessing null safety,
+etc.), list every code path you haven't checked before making any claim. No
+conclusions until all paths are traced. Read every assignment, every branch,
+every caller — then answer. If you start talking before you've finished reading,
+you're pattern-matching, not analyzing.
+
 ## Coding guidelines
 
 Java class members are grouped by kind (static vs. instance) and visibility, and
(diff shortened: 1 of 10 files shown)


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 127 / 156 (81.41%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|127|156|81.41%|

PR #7148, opened 2026-05-25
chore(parquet): include bloom filters in parquet metadata generated by migration

RaphDal:
## Summary

- `_pm` sidecar files produced by `Mig940`, `TableSnapshotRestore`, and the attach-existing-parquet code path in `TableWriter` now inline parquet bloom-filter bitsets, matching the layout the write path already produces. Queries with equality predicates on bloom-indexed columns no longer have to read the bitset from `data.parquet` at plan time on partitions that arrived through one of these paths.
- `MIGRATION_VERSION` bumps from 428 to 429 so installs that already ran the 428 migration overwrite their existing `_pm` files on the next startup.

## Background

`_pm` is the per-partition parquet metadata sidecar introduced with migration version 428. The Rust side has two `_pm`-generation paths:

- `parquet_write` (used when the engine writes a parquet partition) captures bloom-filter bitsets in memory and inlines them into the resulting `_pm`.
- `convert_from_parquet` (used by `Mig940`, `TableSnapshotRestore`, and the attach-parquet path in `TableWriter`) recorded each chunk's `bloom_filter_offset` but did not read the bitset. The `_pm` it produced therefore had no bloom-filter section, and readers had to re-read the bitset from the parquet file on every query touching that column.

The two paths drifted because the write path can capture bitsets while writing each row group, whereas the conversion path has to read them back out of the finished parquet file. This change closes that gap.

## Changes

- `convert_from_parquet` takes an optional `&[u8]` view of the parquet file's bytes. When present, the converter reads each chunk's bitset via `parquet2::bloom_filter::read_from_slice_at_offset` and inlines it through `RowGroupBlockBuilder::add_bloom_filter`. When absent, it behaves as before.
- The `ParquetMetadataWriter.generate` JNI bridge always mmaps the parquet file (previously the mmap was conditional on the partition having a designated timestamp) and passes that slice. All three Java call sites (`Mig940`, `TableSnapshotRestore`, `TableWriter`'s attach-partition path) produce `_pm` files identical to the write path's output.
- `MIGRATION_VERSION` 428 → 429. `Mig940`'s registration moves from 428 to 429 and gains a `migrateForceRegenerate` entry that bypasses the staleness check, so any bloom-less `_pm` produced by the 428 release in 9.4.0 is rewritten on the next startup. Removing the 428 slot also avoids running `Mig940` twice on a table that has not yet upgraded — it now hits the migration exactly once at the 429 slot regardless of starting version.
- In-tree callers that do not need bloom inlining (the `pm_generate` binary, the `parquet_meta_decode` round-trip helper, unit tests in `convert.rs` and `decode_pm_e2e.rs`, the `pm_inspect_e2e` integration test) pass `None` for the new argument and behave as before.

## Trade-offs

- Databases already on migration 428 pay a one-off `_pm` regeneration cost on next startup. `Mig940` re-reads every parquet partition's footer and bloom-filter section and rewrites the sidecar; the cost scales with total parquet partition count and bloom-filter size, and is comparable to the original 428 migration run. Databases coming from below 428 pay a single `Mig940` walk.
- `_pm` file size grows by the size of any inlined bitsets. A bitset is absent when a partition has no bloom-indexed columns, so tables without bloom configuration see no size change. With bloom configuration the growth is roughly the parquet `bloom_filter_length` summed across affected chunks.
- `ParquetMetadataWriter.generate` always mmaps the parquet file now, even on partitions without a designated timestamp and without bloom filters. The mmap itself is cheap; the converter only faults pages it reads, so the page-cache footprint is bounded by what bloom inlining and timestamp backfill would already touch.

## Test plan

- `mvn test -Dtest.include='%regex[.*Mig940Test.*]' -DfailIfNoTests=false`
- `mvn test -Dtest.include='%regex[.*ParquetMetaFileReaderTest.*|.*ReadParquetFunctionTest.*]' -DfailIfNoTests=false`
- `cargo test --lib` in `core/rust/qdbr`
- `cargo clippy --all-targets -- -D warnings` in `core/rust/qdbr`
- Manual: start the engine on a database already at migration 428 that contains parquet partitions with bloom-filtered columns, confirm `_pm` files are regenerated to include the bitsets, and verify equality predicates on those columns do not read `data.parquet` during plan.


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 67 / 77 (87.01%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/bin/pm_generate.rs|0|1|00.00%|
|:large_blue_circle:|qdbr/src/parquet_metadata/convert.rs|26|31|83.87%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/converter.rs|22|26|84.62%|
|:large_blue_circle:|qdbr/tests/decode_pm_e2e.rs|11|11|100.00%|
|:large_blue_circle:|io/questdb/cairo/mig/EngineMigration.java|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/parquet_meta_decode.rs|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/mig/Mig940.java|5|5|100.00%|
|:large_blue_circle:|qdbr/tests/pm_inspect_e2e.rs|1|1|100.00%|

ideoma (commented):
I see the change re-runs `_pm` generation on every restart locally 

```
2026-05-26T10:44:40.302121Z I i.q.c.m.EngineMigration upgrading [path=/Users/alpel/questdb-root/db/idx_test_1h~222, fromVersion=426, toVersion=429]
2026-05-26T10:44:40.302226Z I i.q.c.m.EngineMigration upgrading table [path=/Users/alpel/questdb-root/db/idx_test_1h~222, toVersion=429]
2026-05-26T10:44:40.303875Z I i.q.c.m.EngineMigration generating parquet metadata files [path=/Users/alpel/questdb-root/db/idx_test_1h~222/_txn]
```

This runs on every restart; it should only run once.

RaphDal:
@ideoma
I addressed your comment in https://github.com/questdb/questdb/pull/7148/commits/cb75826ab16fcba82af0a2ea24a12f574d0ea903

ideoma (commented):
## Review (level 3)

14 draft findings across 10 parallel review agents. 7 verified findings retained, 7 false positives removed. All findings are in-diff — cross-context caller analysis verified all callsites (Mig941, TableSnapshotRestore, TableWriter attach-parquet, pm_generate CLI) as SAFE. No correctness bugs, no resource leaks, no concurrency issues, no Rust panic sites, no query-path performance regressions found.

### Moderate

**M1. `testMigrateSkipsHealthyPm` tests behavior that no longer exists**

`Mig941Test.java:1050` — The PR removed the `isParquetMetadataStale()` check from `Mig941.migrate()`. The migration now unconditionally regenerates every `_pm`. But `testMigrateSkipsHealthyPm` (line 1050) asserts `"_pm should not be rewritten"` by comparing file sizes (line 1083). This assertion passes accidentally because a regenerated `_pm` for the same input produces the same size. The test name, the comment at line 1076 ("should skip healthy _pm"), and the assertion message are all wrong — the file IS being rewritten, it just happens to be the same size.

Suggested fix: rename to `testMigrateIdempotentPm` with updated comments, or verify actual skip behavior if that was the intent (it clearly wasn't since the staleness check was removed).

**M2. Bloom filter read error during migration crashes the JVM with no diagnostic trail**

`convert.rs:273` — `convert_from_parquet()` line 273 reads the bloom filter bitset from the parquet mmap via `read_from_slice_at_offset()`. On error, the `?` operator propagates the error up through `generate_parquet_meta()` → JNI → `Mig941.generateParquetMetaForPartition()` → re-thrown at line 229 → `EngineMigration.upgradeTables()` line 211 re-throws → startup fails.

A single parquet partition with a corrupt or non-standard bloom filter header blocks the entire database from starting. Worse, the crash does not flush QuestDB's async logging, so there is no clear way to identify which table or partition caused the failure. The operator sees a startup crash with no actionable diagnostic. A missing bloom filter in `_pm` is a performance degradation (readers fall back to reading from `data.parquet`), not a correctness issue, so the severity of the failure is disproportionate to the impact.

Suggested fix: degrade gracefully — log a warning (which identifies the table, partition, and column) and skip the bloom filter for that column chunk rather than aborting:

```rust
match parquet2::bloom_filter::read_from_slice_at_offset(offset, file_data) {
    Ok(bitset) if !bitset.is_empty() => {
        rg_builder.add_bloom_filter(col_idx, bitset)?;
    }
    Err(err) => {
        // Log warning with table/partition/column context, continue without inlined bloom filter.
    }
    _ => {}
}
```

**M3. Test coverage gaps for bloom filter inlining**

`convert.rs` tests — All new bloom filter tests use a single row group with a single bloom-filtered column:
- `convert_from_parquet_inlines_bloom_filter_from_slice` — 1 RG, 1 bloom column
- `testMigrateInlinesBloomFiltersInPm` — 3 rows (1 RG), 1 bloom column

Missing coverage:
- Multiple row groups with bloom filters — a bug where the same bitset is copied to all RGs, or only the last RG gets a bitset, would go undetected.
- Multiple bloom-filtered columns — OOL region collisions between per-column bitsets are untested.
- Migration doesn't re-run on restart — the `repeatMigrationFromVersion` default change from 426 to -1 is a behavioral fix (commit cb75826ab1 says "to prevent data corruption") with no direct test.

**M4. Stale test comments reference removed `isParquetMetadataStale` method**

`ParquetMetaCrcCorruptionTest.java:82-83` — "Mig940 must treat the _pm as stale (its isParquetMetadataStale helper swallows CairoException as 'stale')" — `isParquetMetadataStale` was removed in this PR. Mig941 unconditionally regenerates, so the comment is factually wrong. The method name `testCrcRefusedLetsMig940RegenerateAfterFlip` (line 81) also references the old class name.

### Minor

**N1. Stale "Mig940" references in production Rust comments**

`converter.rs:125` and `:189`:
- Line 125: `"concurrently because Mig940 and TableSnapshotRestore open it read-only."` → should say Mig941
- Line 189: `"so it is callable from Mig940 before any _pm exists."` → should say Mig941

**N2. PR title type should be `fix`, not `chore`**

The PR fixes a functional gap: `_pm` files from the migration/snapshot-restore/attach paths lacked inlined bloom filter bitsets, causing readers to fall back to reading from `data.parquet` on every query. This is a bug fix (`fix(parquet): ...`), not housekeeping (`chore`). The description should also repeat the verb: e.g., `fix(parquet): fix missing bloom filters in metadata sidecar produced by migration`.

**N3. Helper method and test names still reference "Mig940"**

- `Mig941Test.java:1240` — `runMig940()` method name (calls `Mig941.migrate()`)
- `ParquetMetaCrcCorruptionTest.java:81` — `testCrcRefusedLetsMig940RegenerateAfterFlip`
- Numerous comments throughout `Mig941Test.java` still say "Mig940"

RaphDal:
Thanks for the thorough pass. All seven findings are addressed below. The
net code change is test-only plus comments and a config default already
committed in `cb75826ab1`; the bloom-read control flow in `convert.rs` is
unchanged, so no native rebuild is required.

## M1 — `testMigrateSkipsHealthyPm` tested behavior that no longer exists

Fixed. Renamed to `testMigrateRegeneratesEquivalentPm` and rewrote it to test
what the migration actually does. `Mig941` carries no staleness check, so it
regenerates every `_pm` unconditionally; the test now asserts the regeneration
is idempotent — the regenerated `_pm` matches the write-path `_pm` size, and a
second run leaves the file unchanged. The misleading "should skip healthy _pm"
comment and the "_pm should not be rewritten" assertion message are gone.

## M2 — bloom read error during migration

Addressed, but I landed on **fail-fast** rather than graceful degradation,
after the suggested "log a warning and skip" turned out not to be achievable
in a useful form:

- `log::warn!` in `qdbr` emits nothing. The crate depends on the `log` facade
  but registers no backend (no `JNI_OnLoad`, no `set_logger`), so log macros
  are silent no-ops at runtime.
- The only Rust-to-log path that actually reaches the QuestDB log today is
  `eprintln!` to stderr (the JVM forwards it; see `qwp_zstd.rs`), and that
  carries no table or partition.
- `convert_from_parquet` only has the parquet fd and footer metadata. The
  table and partition directory live in the Java callers (`Mig941`,
  `TableSnapshotRestore`, `TableWriter`), so a Rust-side log line cannot name
  the affected partition without a JNI signature change.

A log that names neither the table nor the partition is unactionable, and
silently dropping the bitset is worse than failing. So the conversion aborts
on an unreadable bitset, and the failure surfaces through the existing path:
`Mig941.generateParquetMetaForPartition` removes the partial `_pm` and
re-throws, and `EngineMigration.upgradeTables` logs
`failed to upgrade table [path=...]` together with the
`could not read parquet bloom filter at offset ...` detail. The bloom-read
control flow is therefore unchanged from the merged code; I added a comment
documenting the fail-fast intent and kept the test asserting propagation.

Consistency note: the parquet-write path (`convert.rs`, `build_row_group_from_thrift`)
swallows the same error with `if let Ok(bitset)`. That predates this PR and is
out of scope here; happy to align the two paths in a follow-up if preferred.

## M3 — test coverage gaps

Added:

- `convert_from_parquet_inlines_bloom_filters_across_row_groups_and_columns`
  (Rust): two row groups x two bloom-filtered columns, asserting every
  `(row group, column)` pair inlines a distinct, non-empty bitset at a distinct
  `_pm` offset. Catches a bitset copied to all row groups, only the last row
  group keeping a bitset, and per-column out-of-line collisions.
- `testMigrateDoesNotRerunOnRestart` (`Mig941Test`): drives the dispatcher
  twice with `force=false`, deletes the `_pm` between runs, and asserts it is
  not regenerated on the second pass. Direct guard for the
  `repeatMigrationFromVersion` default change — the migration must run once,
  not on every startup.
- `testRepeatMigrationFromVersionDefault` (`PropServerConfigurationTest`):
  asserts the production default is `-1` and never equals `ColumnType.VERSION`.
  That equality is the exact collision that reset the recorded migration
  version every boot and caused the every-startup `_pm` regeneration.

## M4 — stale comments referencing removed `isParquetMetadataStale`

Fixed. `ParquetMetaCrcCorruptionTest` no longer cites the removed
`isParquetMetadataStale` helper; the comment now states that `Mig941`
regenerates the `_pm` unconditionally, which is why a CRC-flipped `_pm` is
overwritten. Renamed `testCrcRefusedLetsMig940RegenerateAfterFlip` to
`...Mig941...` and corrected the "After Mig940" comment.

## N1 — stale `Mig940` references in Rust comments

Fixed. `jni/converter.rs` (two sites), `convert.rs` (test doc referencing
`Mig940Test`), and `parquet_read/decode_column.rs` now say `Mig941`.

## N2 — PR title type

Disagree; keeping `chore(parquet)`. There is no incorrect behavior here:
queries against partitions whose `_pm` lacked an inlined bitset always returned
correct results — the reader simply fell back to reading the bitset from
`data.parquet` at plan time. This change brings the migration / snapshot-restore
/ attach paths in line with the write path so that fallback is no longer needed,
which is a consistency and query-plan-performance improvement, not a bug fix.
`chore` is the right type.

## N3 — `Mig940` references in helper and test names

Fixed. `runMig940` is now `runMig941`, and every remaining `Mig940` reference
across `Mig941Test`, `TableWriterTest`, and `ParquetMetaCrcCorruptionTest` is
updated. No `Mig940` references remain in the tree.

## Validation

- `cargo fmt`, `cargo clippy --all-targets` clean; `cargo test --lib parquet_metadata`
  passes (62 tests).
- `Mig941Test`, `ParquetMetaCrcCorruptionTest`, `PropServerConfigurationTest`,
  and `ServerMainTest#testShowParameters` pass.
- No native rebuild required: the net Rust change is test and comment only.

bluestreak01 (dismissed):
Reviewed at level 2. No correctness bugs, no resource leaks, no Rust panic risks, no query-path performance regressions. The bloom-filter inlining code is solid.

Three minor items worth addressing at your discretion:

- **Stale comment** at `Mig941Test.java:879-880` still references the removed `isParquetMetadataStale()` method.
- **Five `mig940` references survive** (fixture path, Rust test function name, and comments) despite the rename sweep.
- **`testMigrateInlinesBloomFiltersInPm`** only checks the `BLOOM_FILTERS` feature-flag bit, not the actual bitset content — the Rust unit test covers content, but the end-to-end Java path does not.

None are blocking.

ideoma (dismissed):

RaphDal:
Thanks for the review @bluestreak01
I addressed the stale comment and the `Mig940` references in https://github.com/questdb/questdb/pull/7148/commits/0323886995d5af003a5c89d105e4e4076a04d985.
I didn't modify `testMigrateInlinesBloomFiltersInPm` because as the comment state, the content is already checked by a unit-test.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 303 / 321 (94.39%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/bin/pm_generate.rs|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/mig/Mig941.java|0|1|00.00%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/converter.rs|22|26|84.62%|
|:large_blue_circle:|qdbr/src/parquet_metadata/convert.rs|266|278|95.68%|
|:large_blue_circle:|qdbr/tests/decode_pm_e2e.rs|11|11|100.00%|
|:large_blue_circle:|io/questdb/cairo/mig/EngineMigration.java|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/parquet_meta_decode.rs|1|1|100.00%|
|:large_blue_circle:|qdbr/tests/pm_inspect_e2e.rs|1|1|100.00%|

bluestreak01 (approved):
Reviewed at level 2. No correctness bugs, no resource leaks, no Rust panic risks, no query-path performance regressions. The bloom filter inlining code is solid — error propagation is complete, lifetimes are sound, mmap is properly scoped, and all production callsites are correct.

Three minor items (non-blocking):

- **PR body**: 6 stale "Mig940" references remain in the description (source code is clean). Worth updating before squash-merge since the body feeds the commit message.
- **`pm_generate` CLI**: passes `None` for `parquet_file_data`, so its output diverges from production `_pm` when bloom filters are present. The file already opens the parquet — adding an mmap for parity would be straightforward.
- **PR title type**: `chore` vs `fix` is defensible either way given the author's reasoning that queries always returned correct results.

Issue #3, opened 2014-08-09
Abstract configuration builder


bluestreak01:
As a developer I want to be able to configure Journal factory from sources other that db.xml.

Other implementations could be database-based, code based or may be even some Sprint based configuration.

PR #7150, opened 2026-05-25
fix(sql): fix wrong results and memory leaks in posting-index, join and group-by queries

puzpuzpuz:
## Summary

This PR extends the query fuzzer with new query shapes and index variants, then ships the engine bug fixes that the expanded coverage surfaced. The fuzzer changes are test-only; the engine fixes each carry a targeted regression test independent of the fuzzer.

## Query fuzzer changes

- **Fuzz posting indexes with covering columns** - SYMBOL columns now draw uniformly across all four index kinds (bitmap, plain posting, DELTA-encoded posting, EF-encoded posting), with optional `INCLUDE` covering columns on posting variants. The shadow sibling re-draws independently, so storage diffs compare posting/covering against bitmap or no index over identical data.
- **Generate posting-index read query shapes** - `PostingClause` emits the two shapes the planner needs to pick the posting read operators: bare `SELECT DISTINCT <postingSym>` (PostingIndex) and `SELECT <subset> WHERE <postingSym> = ... | IN (...)` (CoveringIndex). `SymbolType` switches to a small known domain so literal predicates actually match rows.
- **Generate richer temporal join ON conditions** - ASOF/LT/SPLICE joins now get empty, one-key, or two-key ON clauses in both shorthand and explicit `a.x = b.y` forms. Pairing follows QuestDB's actual join-key rule (exact type match or SYMBOL/STRING/VARCHAR family) to avoid spurious "join column type mismatch" skips.
- **Fuzz generic SELECT DISTINCT queries** - `SimpleClause` emits SELECT DISTINCT over arbitrary projections and source shapes, exercising the DISTINCT-to-GROUP-BY rewrite and its async variants separately from the narrow single-column DISTINCT that drives the posting index.
- **Tolerate splice-join random-access asymmetry** - SPLICE JOIN rejects master inputs without random access; when the master's SYMBOL column carries a posting/bitmap index the planner picks an index-driven scan that lacks random access, while the unindexed shadow full-scans and compiles. This is a planner choice driven by index configuration, so the runner now treats it as a planner-sensitivity asymmetry.
- **Tolerate ASOF full-fat var-size slave column asymmetry** - ASOF JOIN falls back to `createFullFatJoin` when the slave factory does not support random access (e.g. shadow's slave becomes a `CoveringIndexRecordCursorFactory` once a `sym` predicate propagates onto an indexed slave). The full-fat path materialises non-key slave columns into a map value and rejects variable-length slave columns (STRING, VARCHAR, BINARY, arrays). The primary, with no slave index, keeps a page-frame scan that supports random access and takes the light path, which has no such type restriction. Same planner-sensitivity shape as the SPLICE skip above; the runner now allowlists the matching SqlException.
- **Skip bind-vs-literal int-overflow asymmetry** - `(53256 * abs(58539)) < c0` (c0 SHORT, no LONG anywhere) diverges 0 rows vs 116 rows: the literal path constant-folds through `FunctionParser.functionToConstant0` to a `LongConstant`, so the comparison runs at long width; the bind path keeps the subtree opaque, picks `*(II)`, wraps to a negative INT, and `<(II)` reads via `getInt` so the SHORT widens and every row passes. The production fix (read both operands via `.getLong()` in `LtIntFunctionFactory` / `EqIntFunctionFactory`) ripples to `>`/`>=`/`<=`/`!=` via the negating/swapping wrappers and is a noticeable behavior change, so detect in the runner instead: walk balanced parens of the literal SQL, compile each subexpression through `testParseExpression` + `parseFunction`, treat a `LongConstant` whose long value does not fit INT as the overflow signal. A `NON_INT_TYPE_HINT` pre-filter excludes candidates with LONG/FLOAT/DOUBLE/DECIMAL/TIMESTAMP/DATE literals or casts (type promotion, not overflow).

## Engine bug fixes surfaced by the fuzzer

While this branch was open, upstream independently landed equivalent fixes for two of the items below: the multi-key CoveringIndex ordering (#7182) and the `toTop()` row-cursor leak (#7187). After merging `master`, items 5 and 6 describe only what remains on top of those upstream fixes. The numbering is left unchanged so the `follow-up to (N)` cross-references still hold.

1. **Off-by-one when stepping past null array map key** - `ArrayTypeDriver.getPlainValueSize(long)` returned `8 + (-1) = 7` for null array slots, causing `OrderedMapVarSizeRecord` to land one byte inside the prior slot when stepping past a null array key, tripping the "typeTag of encodedType is not ARRAY" assert.
2. **ASOF Light path for String/Varchar to Symbol joins** - `SymbolJoinKeyMapping.of(RecordCursor)` had a default that threw `UnsupportedOperationException`; only the symbol-to-symbol variant overrode it. STRING/VARCHAR master keys routed through the Light factory (when the slave does not support `TimeFrameCursor`, e.g. parquet + posting index) crashed.
3. **TablePageFrameCursor on covering page frames** - parallel keyed GROUP BY over a `CoveringIndex` factory wrapped by `SelectedRecord` crashed with `ClassCastException` because `CoveringPageFrameCursor` implemented `PageFrameCursor` directly while `SelectedRecordCursorFactory.getPageFrameCursor()` casts to `TablePageFrameCursor`.
4. **Free row cursor factory in parent factory _close()** - when `HeapRowCursor.of()` threw mid-iteration (e.g. a DECIMAL scale overflow in the post-filter), the singleton row cursor was never assigned to `PageFrameRecordCursorImpl.rowCursor`, leaving per-symbol `PostingIndexFwdReader.Cursor` instances outside the reader's freeCursors pool and leaking their block buffers (NATIVE_INDEX_READER). `HeapRowCursorFactory` and `SequentialRowCursorFactory` gain a `close()` that frees the inner singleton, and each parent factory that wraps them (`FilterOnExcludedValues`, `FilterOnValues`, `FilterOnSubQuery`) frees its `RowCursorFactory` from `_close()`, matching the ownership pattern `PageFrameRecordCursorFactory` already used.
5. **Free row cursor on PageFrame cursor toTop()** *(landed upstream in #7187; no longer in this PR's diff after merging master)* - `PageFrameRecordCursorImpl.toTop()` assigned `rowCursor = null` instead of closing it. Keyed `SAMPLE BY ... FILL(PREV)` drives the base cursor twice (initializeMap + buildMap) and called `toTop()` between passes, stranding the `PostingIndexFwdReader.Cursor` and leaking its native block buffer (NATIVE_INDEX_READER). Upstream's #7187 made the same `toTop()` fix, so after merging master the branch and master agree on this file and it drops out of the diff.
6. **Honest scan direction and duplicate-key dedup for multi-key CoveringIndex** - the underlying ordering bug (a multi-key IN drained each key's posting cursor sequentially, emitting `(partition, key, row-id)` order while advertising `SCAN_DIRECTION_FORWARD`) was independently fixed upstream in #7182, which reworked the multi-key cursor to merge the per-key posting cursors by row-id so each partition emits in ts-ascending order. After merging master this PR keeps the two pieces #7182 does not cover. (a) `getScanDirection()` is overridden honestly: `SCAN_DIRECTION_OTHER` only for multi-key LATEST BY (one row per key in key order, not ts order), and `SCAN_DIRECTION_FORWARD` for single-key LATEST BY (a single row, trivially ts-ordered) and for all non-latestBy scans (the row-id merge is ts-ascending). SAMPLE BY's entry gate, ORDER BY ts elision in `SqlCodeGenerator`, and window/join scan-direction forwarding all trust this advertisement, so without the override a multi-key LATEST BY still hands a downstream the FORWARD claim it does not satisfy. (b) IN-list elements that resolve to the same symbol id (via bind variables or runtime constants) are deduplicated before the merge, so a key is not opened twice and its row-id stream merged twice (duplicate rows / inflated aggregates). The ordering bug was latent since posting indexes landed in #6861; it surfaces only when at least two keys are in the IN list, their row-ids interleave in time, and a downstream exploits the FORWARD claim.
7. **Per-worker array key memory leak in GROUP BY** - `extractWorkerFunctionsConditionally` re-checked only the VIRTUAL function-key slots when deciding whether to keep per-worker copies, so a thread-safe `ArrayConstant` key alongside a non-thread-safe partner (SYMBOL column, `count_distinct`) had its per-worker copies orphaned, leaking the backing `DirectArray` (NATIVE_ND_ARRAY) once per compile per worker.
8. **Use group-by alias for outer column on duplicate refs** - in `SqlOptimiser.createSelectColumn`, the duplicate-literal branch re-used the inner column object as the outer reference; the outer-virtual / distinct models live under the group-by model and expose `innerAlias`, not `translatedColumnName`. `SELECT x AS e1, x AS e2 ... DISTINCT` hit `wtf? x` in `generateSelectChoose`.
9. **Gate trivial-group-by LIMIT push on ORDER BY column presence** - `rewriteTrivialGroupByExpressions` pushed LIMIT from the wrapping VIRTUAL down to a nested GROUP BY unconditionally. When the VIRTUAL exposed columns the GROUP BY did not (e.g. constant aliases after the DISTINCT-to-GROUP-BY rewrite), `rewriteOrderBy` attached `ORDER BY <constant-alias>` to the limited GROUP BY and code generation crashed with `ArrayIndexOutOfBoundsException`. The push is now gated on every ORDER BY token resolving in the GROUP BY's alias map, walking the nested chain from the VIRTUAL downward since after `rewriteSelectClause` the ORDER BY sits on the deepest base model. Narrowing the gate to ORDER BY tokens keeps the bounded-cursor benefit for shapes like ClickBench Q35.
10. **Name posting-index distinct output by alias** - the posting-index DISTINCT optimization in `generateSelectGroupBy` named its output column after the source token instead of the projection alias, so outer models could not resolve the alias (`AssertionError: wtf? k` or `Invalid column: k`).
11. **Widen narrow JIT operands when LONG only appears as literal** - the JIT widen pre-pass observed only FLOAT/DOUBLE constants in the CONSTANT case, so a predicate whose only LONG source was a literal (e.g. `-432577L`) never reached `NarrowI64WidenDetector.typesObserver`. `needsNarrowI64Widening` stayed false, no `SX_I64` was emitted, and a narrow LHS computed at int32 (wrapping mod 2^32) before the i32->i64 widening at the comparison, diverging from the Java path which computed at long width throughout. The detector now recognises an L/l suffix or a magnitude outside the int range and emits the widening so JIT and Java agree.
12. **Full-fat AsOf/Lt join projection of cross-type SYMBOL key** - the full-fat join's `SymbolWrapOverJoinRecord` wraps slave-key reads over the master cursor because slave symbol IDs are lost when the key is copied to the join map as a string. With a STRING master key and SYMBOL slave key, master has no symbol table, so projecting the slave key trips `symbolTableStatic && getStaticSymbolTable() != null` in `SymbolColumn.init` before any row is read. The fuzzer hit this with a parquet-only master joined on STRING/SYMBOL to a slave forced onto the no-random-access covering-index path. `createFullFatJoin` now exposes such cross-type slave SYMBOL keys in the JOIN metadata as the master's type, so the planner picks a `StrColumn` that reads via `getStrA()` from the map record where the key is already stored as a string.
13. **JIT widen pre-pass misclassifies null / true / false tokens** - follow-up to (11): `longConstantTypeCode` and `floatConstantTypeCode` classified any token ending in `l` / `L` / `e` (most notably the `null`, `true` and `false` literals) as I8 or F8 before parsing. In a predicate combining narrow-int arithmetic with such a literal the spurious observation flipped `NarrowI64WidenDetector.shouldWiden` and emitted `SX_I64` around the column read, so the JIT computed at long width while the scalar Java filter wrapped at int. Both helpers now reject the reserved literal keywords up front.
14. **Full-fat join slaveTypes lags metadata rewrite for cross-type SYMBOL key** - follow-up to (12): `createFullFatJoin` rewrites the slave key column metadata to the master's type for SYMBOL-to-non-SYMBOL keys, but `slaveTypes` still carried SYMBOL. `NullRecordFactory` then built a `SymbolConstant.NULL` slot for that column, and the outer-join no-match path crashed in `SymbolFunction.getVarcharSize` when the master was VARCHAR. `slaveTypes` is now overridden alongside the metadata.
15. **Resolve qualified column in expression under DISTINCT** - `SqlOptimiser.addColumnToTranslatingModel` strips the table prefix in the single-base case, so `t0.c2 AS e1` registered the translating-model entry under the unqualified key `c2`. `doReplaceLiteral0` only mirrored that strip in its `joinCount > 1` branch; in the single-table case it looked the literal up qualified, missed the entry, treated the column as new, allocated a duplicate alias, and overwrote the original `columnNameToAlias` mapping. The translating model was later elided as redundant and the parent virtual model resolved the duplicated alias against GROUP BY metadata that no longer exposed the original column, raising `Invalid column: c2` for `SELECT DISTINCT ..., t0.c2 AS e1, ..., (k * t0.c2) AS e2 FROM t t0`. `doReplaceLiteral0` now strips the prefix and retries the lookup when the literal arrives qualified against a single base model, the symmetric counterpart to the existing multi-join branch.
16. **Return NULL from constant symbol valueOf on VALUE_IS_NULL key** - `SymbolConstant.valueOf(int)` and `valueBOf(int)` ignored the key entirely and returned the constant value, breaking the `SymbolTable` contract that `VALUE_IS_NULL` must map to null. Group-by aggregates such as `FirstSymbolGroupByFunction` (parent of `last` / `last_not_null`) and `ModeSymbolGroupByFunction` store `VALUE_IS_NULL` in their state on `setEmpty` / `setNull` and read it back via `arg.valueOf(getInt(rec))`; a `SymbolConstant` arg resurrected the constant instead of yielding null. `SELECT last(constant_symbol) FROM t WHERE <folds_to_false>` therefore returned the constant for one row instead of NULL: the WHERE-folded `EmptyTableRecordCursorFactory` drove the not-keyed group-by into `updateEmpty` and the read-back went through the buggy `SymbolConstant.valueOf`. The bind variant of the same query went through `AbstractCastToSymbolFunction.valueOf`, which honours `VALUE_IS_NULL` via `TableUtils.toIndexKey`, hence the literal-vs-bind divergence the fuzzer reported. Both `valueOf` and `valueBOf` now return null when the key is `VALUE_IS_NULL`.
17. **Strip qualified column ref in `createSelectColumn`** - follow-up to (15): `doReplaceLiteral0` got the strip-and-retry for the single-base-model case, but `createSelectColumn` makes the same `translatingModel.getColumnNameToAliasMap().get(columnAst.token)` lookup and was never updated. `SELECT DISTINCT :b0::CHAR, (t0.x)::CHAR, t0.x FROM long_sequence(N) t0` from the fuzzer (seeds 1096908972480L / 1779975290744L) hits the same shape: the bind cast steals the early `cast` / `cast1` aliases, so the bare `t0.x` projection keeps user alias `x`. `(t0.x)::CHAR`'s emit already registered the translating entry under stripped key `x`, then `createSelectColumn` looked the bare `t0.x` up qualified, missed, and synthesised alias `x1`. `rewriteDistinct` had captured the original `x` for the wrapper, so the outer pass raised `Invalid column: x`. The literal form happens to work only because case-insensitive collision between an earlier `'X'` constant and `t0.x` pre-renames the bare projection's auto-alias to `x1`, accidentally matching the inner rename; swapping the constant for a bind cast frees the alias and exposes the divergence. `createSelectColumn` now strips the prefix and retries the lookup when the literal arrives qualified against a single base model.
18. **Add IN function factory for IPv4** - QuestDB had no `in(Xv)` overload for IPv4, so `c IN ('1.2.3.4'::ipv4, ...)` against an IPv4 column had no exact dispatch. The planner fell back to `InStrFunctionFactory`, which casts the LHS to STRING and then validates each list element against the STRING-compatible set; an IPv4 constant in that list failed at compile time with "cannot compare STRING with type IPv4". The fuzzer surfaced this on seeds 6329374986193L / 1779980523210L via `WHERE (sym = ...::varchar AND sym IS NULL) AND NOT (c3 IN ('18.32.198.182'::ipv4))`: on the primary table the POSTING-indexed `sym` contradiction folded during the indexed-scan path so the residual IPv4 IN filter never had to compile; the unindexed shadow had no such short-circuit, reached the InStr factory, and threw. The new `InIPv4FunctionFactory` matches `in(Xv)` directly, accepting NULL, STRING, VARCHAR, SYMBOL, IPv4, and bind variables (UNDEFINED). String-like elements run through `Function.getIPv4`, which delegates to `SqlUtil.implicitCastStrAsIPv4` -- the same parse path `EqIPv4StrFunctionFactory` uses, with `ImplicitCastException` re-thrown as `SqlException` at the list element's position. When the LHS is also constant the factory folds to `BooleanConstant`. Storage uses `LongHashSet` rather than `IntHashSet` because IPv4 covers the full 32-bit int range (`IPv4_NULL = 0` and broadcast `255.255.255.255 = -1` are both valid), so no `IntHashSet` sentinel is safe; widening to long puts every sign-extended int outside the `Long.MIN_VALUE` sentinel.

## Minor cleanup

- Drop a dead `lhi != rhi` guard in `SqlOptimiser.analyseEquals` (the preceding early return makes the two indices distinct by the time control reaches the guard).

## Test plan

- All new engine fixes carry targeted regression tests (`ArrayAsMapKeyTest`, `AsOfJoinTest`, `CompiledFilterTest`, `CoveringIndexTest`, `DistinctTest`, `InIPv4Test`, `FirstSymbolGroupByFunctionFactoryTest`, `LastSymbolGroupByFunctionFactoryTest`, `ModeSymbolGroupByFunctionFactoryTest`, `ParallelGroupByFuzzTest`, `ParallelHorizonJoinFuzzTest`)
- Merged latest `master`: upstream's #7182 (multi-key CoveringIndex row-id merge), #7187 (fluent `QueryAssertion` builder) and #7157 (ORDER BY byte caps) reconciled with this branch's covering-index and test changes; affected suites re-run green (`CoveringIndexTest`, `ExplainPlanTest`, `AsOfJoinTest`, `CompiledFilterTest`, `LimitTest` and the migrated group-by / distinct / IPv4 / array tests)
- Query fuzzer runs at `-Dquestdb.fuzz.queries=900` with `diffJit` and `diffShadow` enabled, across 1-, 2- and 3-table runs, pass with zero failures and zero "join column type mismatch" skips
- `mvn -pl core test` green


puzpuzpuz:
@nwoolmer could you check d5dd0298e72e09266bd1531333ec57b115679a7c - it's a noticeable covering index bugfix.

nwoolmer:
> @nwoolmer could you check [d5dd029](https://github.com/questdb/questdb/commit/d5dd0298e72e09266bd1531333ec57b115679a7c) - it's a noticeable covering index bugfix.

Looks good to me. I have a follow up piece to speed up group by on covering index (currently, the decode step is serialized) so this will help to prevent regressions.

puzpuzpuz:
Verification complete. Here is the final review.

  ---
  Code Review — PR #7150 (Level 3)
  
  fix(sql): fix wrong results and memory leaks in posting-index, join and group-by queries — 46 files, +4212/−1096, 18 engine bug fixes + query-fuzzer test changes. No .rs files.

  Method: 5 change-surface mappers + 10 parallel review agents (correctness, concurrency, perf, resource, tests, quality, meta, cross-context, fresh-adversarial, adversarial-perf), then 4
  deep adversarial re-probes (per-worker ownership, optimiser strip-retry, JIT widening, cross-context blast radius), plus my own source-level adjudication of 9 high-risk hypotheses. The
  structured pass surfaced 0 correctness/concurrency/resource/cross-context findings; I independently re-ran the riskiest surfaces to confirm that is real, not an underrun.

  Critical

  None. No correctness, resource-leak, concurrency, double-free, or undefined-behavior issue survived verification.

  Moderate

  M1 — 18 independent fixes bundled into one squash-merged commit (out-of-diff / process)
  The PR carries 18 numbered, technically independent fixes spanning four subsystems (cairo storage, jit, griffin SQL, griffin join/group-by). Per CLAUDE.md, PRs are squash-merged — all 18
  collapse into one atomic commit on master, defeating per-fix bisect/revert. The change-surface maps confirm no hard technical coupling (the only intra-PR dependencies are
  iterate-on-same-fix pairs that squash harmlessly: #13→#11, #14→#12, #17→#15). This is your call as author; if kept as one PR, a line in the body stating the bundle is intentional would
  help future bisectors. Splitting candidates: (a) test-only fuzzer expansion, (b) storage/JIT correctness (ArrayTypeDriver + the two CompiledFilterIRSerializer fixes), (c) SqlOptimiser
  model-rewrite cluster, (d) row-cursor lifecycle leaks, (e) join/IPv4/SymbolConstant.

  Minor

  m1 — IN-key dedup is O(n²) — CoveringIndexRecordCursorFactory.java:212 (getCursor) and :283 (getPageFrameCursor) dedup via !multiKeys.contains(key) in a loop where multiKeys is an IntList
  (contains is a linear scan), making dedup O(n²) in IN-list cardinality. Init-path (once per cursor acquisition, not per row), so impact is bounded and only material for pathologically
  large IN (...) lists. Fix is a clean drop-in: declare multiKeys as IntHashSet (O(1) dedup via add's return, stable positional get(slot) confirmed at IntHashSet.java:104) and drop the
  contains guard — precedent at InSymbolCursorFunctionFactory.java:218. Valid on algorithmic-optimality grounds.

  m2 — Multi-key k-way merge uses a sorted-array list (O(k)/row) not a binary heap (O(log k)) — MultiKeyCoveringCursor.hasNext() (:1869) and
  MultiKeyCoveringPageFrameCursor.fillFramePartition() (:2123) merge per-key posting cursors through IntLongSortedList, whose pollAndReplace/pollValue do an O(k) array shift per emitted row
  → O(N·k) vs O(N·log k) for a binary heap. The adversarial-perf agent rated this moderate; I temper to minor for two verifiable reasons (the algorithmic point itself stands): (1) k =
  IN-keys with rows in the current partition, realistically small/bounded; (2) IntLongSortedList is the codebase's universal merge primitive — HeapRowCursor (the very pattern this fix
  deliberately mirrors) uses it identically, and no binary-heap primitive exists in std/. A real heap would be a codebase-wide change, not a PR-local one.

  m3 — longConstantTypeCode recomputes floatConstantTypeCode the caller already evaluated — CompiledFilterIRSerializer.java:1936 calls floatConstantTypeCode(token), then :1938 calls
  longConstantTypeCode(token), which re-invokes floatConstantTypeCode(token) internally (:515). Compile-time, per-literal, tiny tokens — negligible, but a free tidy-up (pass the precomputed
  float code down, or call longConstantTypeCode directly since it already short-circuits floats).

  m4 — maxRowsPerFrame field out of alphabetical order — CoveringIndexRecordCursorFactory.java:748: when widened private→protected, it was appended after tableReader instead of its
  alphabetical slot between isExhausted (:737) and pendingCoveringCursor (:743). Violates the CLAUDE.md member-ordering rule.

  m5 — New test SQL uses 5+ digit integer literals without underscores — e.g. DistinctTest (278444 * t0.x) (pr.diff:4504), CompiledFilterTest INSERT ... (127, 20000000, ...) (pr.diff:3081,
  whose adjacent comment correctly writes 20_000_000), plus widespread 10000, 348512, etc. CLAUDE.md mandates underscores for 5+ digit numbers. Test-only/cosmetic; apply only to genuine
  integer SQL literals (leave fraction strings, fuzzer seeds, magnitudes).

  m6 — Labels under-represent scope (Core missing) — Labels are Bug, SQL, regression, but fix #1 (ArrayTypeDriver, cairo storage) and fixes #11/#13 (CompiledFilterIRSerializer, JIT) are
  Core-engine, not SQL-surface. Precedent #7176 (posting-index storage) = Bug,Core,storage. Add Core (consider storage). Correctly not Performance — the JIT changes are correctness fixes.

  m7 — Coverage completeness: first_not_null/last_not_null(symbol-constant) over empty table — The SymbolConstant NULL-sentinel fix is regression-tested for first/last/mode (all
  non-tautological), but the *_not_null aliases lack a dedicated assertion. They extend FirstSymbolGroupByFunction and inherit the exact setNull→valueOf(VALUE_IS_NULL) read-back, so there is
  no untested behavioral path — optional to add.

  m8 — Coverage: setScanProfile covering-reader delegation untested — The TablePageFrameCursor promotion flips setScanProfile from no-op to getTableReader().setScanProfile(). Benign and
  covered by the shared TablePageFrameCursor contract (16 sibling impls), but no test asserts the profile reaches the covering reader. Optional.

  (There is no Fixes #NNN; an issue search found no matching tracking issue, so this is defensible — these are internal fuzzer discoveries. Noted, not flagged.)

  Downgraded (investigated and cleared)

  No agent finding was a false positive — but I probed 9 high-risk hypotheses + the full cross-context blast radius and cleared every one against source. Documenting so the reasoning is
  auditable:

  - ArrayTypeDriver.getPlainValueSize off-by-one — correct: the totalSize <= 0 → Long.BYTES branch exactly mirrors getPlainValue's null check (:387); real arrays always have totalSize ≥ 8.
  - toTop() + factory close() double-free — non-issue: RowCursor.close() is contractually idempotent, HeapRowCursor.close() clears its list → inner POSTING cursors freed exactly once; super
  _close() doesn't free the row cursor factory; each FilterOn* factory owns a distinct instance.
  - toStaticSymbolTable now throws AssertionError — gated unreachable by SqlCodeGenerator:2413 isSymbolTableStatic; strictly safer than the old null-return-then-NPE.
  - InIPv4 dropped-constant ownership — matches InLong/InChar convention (constants are GC-managed; only native sets need freeing).
  - Multi-key merge correctness — RowCursor.next() idempotency holds per contract; IntLongSortedList poll/peek semantics correct; row-ids are unique per key within a partition (no tie
  ambiguity).
  - Per-worker GROUP BY fn ownership (H7) — clean graph: COLUMN→null, VIRTUAL→factory, GROUP_BY→factory or freeWorkerFunctionsByFlag; no leak/double-free/NPE; null COLUMN slots never escape
  the temp list.
  - SqlOptimiser qualified strip-retry (H8) — survived 25+ adversarial queries; fixes proven real/non-tautological; multi-join branch byte-identical to master.
  - JIT long-literal widening (H9) — no over-widening divergence constructible (an I8 in a comparison forces FunctionParser's non-INT resolution, so the Java filter widens the subtree too);
  token misclassification unreachable via the ExpressionParser CONSTANT-node gate.
  - getScanDirection FORWARD claim — honest; fixes two latent master unsoundness bugs (latestBy and the old multi-key concatenation both falsely advertised FORWARD). Likewise the
  TablePageFrameCursor widening fixes a latent ClassCastException at the pre-existing SelectedRecordCursorFactory:153 cast — that is fix #3's purpose, now independently confirmed reachable.

  Summary

  Verdict: Approve. After an exhaustive level-3 pass — 15-agent workflow, 4 deep adversarial re-probes, and independent source adjudication of every high-risk surface — I found no 
  correctness, resource-leak, concurrency, or security issue. Each of the 18 fixes was confirmed to fix a real bug without breaking sibling paths, and three of them additionally close latent
  unsoundness in unchanged code (the TablePageFrameCursor CCE and the two false-FORWARD scan-direction bugs).

  - 11 draft findings → 10 unique (the two floatConstantTypeCode-double-compute findings merged); all 10 verified, 0 false positives. Plus ~13 high-risk hypotheses probed and cleared.
  - In-diff vs out-of-diff: 7 in-diff (m1, m2, m3, m4, m5, m7, m8), 3 out-of-diff (M1, m6, and the no-issue-ref note) — all out-of-diff items are metadata/process, not code. The
  cross-context pass did not underrun: I independently re-ran SAFE/BROKEN verdicts on every high-blast-radius symbol (TablePageFrameCursor casts, SymbolJoinKeyMapping.of, getScanDirection,
  SymbolConstant.valueOf, compilePerWorkerInnerProjectionFunctions, the three FilterOn* _close()s) and found zero broken callsites.
  - Tradeoffs to weigh: the only judgment call is M1 (bundling — your decision as author); everything else is optional polish. The two perf items (m1, m2) are valid optimality observations
  with low practical impact; neither blocks merge.

  The working tree is clean and the core module compiles; the new ParallelGroupByFuzzTest/ParallelHorizonJoinFuzzTest exercise the leak-prone paths under assertMemoryLeak.

puzpuzpuz:
Code Review — PR #7150 (Level 3)

  fix(sql): fix wrong results and memory leaks in posting-index, join and group-by queries — 47 files, +3863/−1054, 18 engine bug fixes + query-fuzzer test expansion. No .rs files (Agent 8
  skipped).

  Method: 10 parallel agents (5 per-cluster correctness, resource+concurrency, performance, tests, fresh-adversarial, quality+meta), each verifying against source — several reverted
  individual fixes and watched the regression tests fail to prove the fixes are load-bearing. I then independently adjudicated the highest-risk surfaces (FIX 9 pass ordering,
  getScanDirection under O3, SymbolConstant blast radius, per-worker double-free) against source myself. This is a fully independent pass; where it differs from the review already posted on
  the PR, I say so.

  Critical

  None. No correctness, resource-leak, concurrency, double-free, or undefined-behavior issue survived verification. The most sensitive surfaces were cleared with proof (see Downgraded).
  Three of the 18 fixes additionally close latent unsoundness in unchanged code (the TablePageFrameCursor ClassCastException, and two false-FORWARD scan-direction advertisements).

  Moderate

  M1 — FIX 9's LIMIT-push gate silently disables the bounded cursor for ORDER BY <position> and ORDER BY <qualified.column> (in-diff, NEW — not in the posted review).
  SqlOptimiser.java:10565 gates the LIMIT push-down on allOrderByColumnsPresentIn(model, nestedModel) (helper at :438), which tests
  nestedModel.getAliasToColumnMap().contains(orderBy.getQuick(i).token) against the raw, un-normalized ORDER BY token. But rewriteTrivialGroupByExpressions runs at optimise():11822, before
  rewriteOrderByPosition (:11830, position→alias) and rewriteOrderBy (:11832, qualifier strip). So:
  - ... ORDER BY 5 DESC LIMIT 10 → token is "5" → not in the alias map → gate fails → push skipped.
  - ... ORDER BY hits.ClientIP LIMIT 10 → token is "hits.ClientIP" → not in the unqualified alias map → push skipped.
  - ... ORDER BY c DESC LIMIT 10 (bare alias) → token "c" → in map → push happens.

  Master had no gate and always pushed. So for the ClickBench-Q35 class this rewrite exists to optimize, two semantically identical queries (ORDER BY 5 vs ORDER BY c, where position 5 is c)
  now get different plans: the positional/qualified form leaves the GROUP BY unbounded and runs the outer VIRTUAL projection over all groups instead of the top-N. Results stay correct — this
  is a silent perf regression, ironic given the PR carries the regression label.

  Suggested direction: normalize the token before the membership test — resolve a numeric position against the outer projection to its target column, and strip a table qualifier — so the
  gate reflects what rewriteOrderBy will actually bind. Caution: do not blanket-treat positions as "present." A position that points at a virtual-only/constant alias must still block the
  push, or the exact ArrayIndexOutOfBoundsException in generateOrderBy that this fix prevents returns. The current behavior is safe (conservative); it's just over-conservative for the common
  positional case.

  M2 — 18 independent fixes bundled into one squash-merged PR (process). CLAUDE.md: PRs are squash-merged, so all 18 collapse into one master commit, defeating per-fix git bisect/revert. The
  change-surface maps show no hard technical coupling (the only intra-PR couplings are iterate-on-same-fix pairs that squash harmlessly: #13→#11, #14→#12, #17→#15). Several fixes are
  behavior changes with blast radius beyond the fuzzer (#16 SymbolConstant NULL contract, #11/#13 JIT widening). This is the author's call; if kept as one PR, a one-line note in the body
  that the bundle is intentional would help future bisectors.

  M3 — the horizon-join half of FIX 7 has probabilistic test coverage only (in-diff, test gap). The GROUP BY half (the extractWorkerFunctionsByFlag rewrite) is covered deterministically by
  ArrayAsMapKeyTest (fixed 4-worker pools, assertMemoryLeak, empirically proven to fail without the fix). But the horizon-join half — freeWorkerFunctionsByFlag(GROUP_BY) in
  generateHorizonJoinFactory/generateMultiHorizonJoinFactory — is exercised only by ParallelHorizonJoinFuzzTest.testParallelHorizonJoinConstArrayKeyDoesNotLeak, gated by
  Assume.assumeTrue(enableParallelHorizonJoin) where enableParallelHorizonJoin = rnd.nextBoolean() with a per-run random seed (ParallelHorizonJoinFuzzTest.java:52,103). It is skipped ~50% of
  runs (the count_distinct group-by variant ~20%, nextInt(5) != 0). This is the codebase's standard fuzz convention (random seed, logged, CI runs frequently), so the path is exercised over
  repeated runs — but a single CI run or a developer's local run may not touch it. Consider a deterministic horizon-join NATIVE_ND_ARRAY leak test mirroring the group-by ones.

  Minor

  m1 — IN-key dedup is O(n²) (in-diff, perf). CoveringIndexRecordCursorFactory.java:211 (getCursor) and :295 (getPageFrameCursor) dedup via !multiKeys.contains(key) where multiKeys is an
  IntList (contains = linear scan) → O(n²) in IN-list cardinality. Per-cursor-acquisition, not per-row, and literal duplicates are already deduped by WhereClauseParser, so n (distinct
  resolved keys, e.g. bind vars folding to the same symbol) is normally tiny. Still a valid optimality finding: precedent InSymbolCursorFunctionFactory.java:218 uses an IntHashSet. The merge
  indexes multiKeys positionally, so the fix is to dedup into the list using a scratch IntHashSet for the membership test, not to replace the list.

  m2 — longConstantTypeCode recomputes floatConstantTypeCode the caller already computed (in-diff, perf, compile path). CompiledFilterIRSerializer visit() calls floatConstantTypeCode(token),
  then on UNDEFINED calls longConstantTypeCode(token), which re-invokes floatConstantTypeCode(token) internally (and re-runs isReservedConstantKeyword). Compile-time, per-literal, tiny
  tokens — negligible, but a free tidy-up (call longConstantTypeCode first, or thread the precomputed float code through).

  m3 — Core/storage label missing (out-of-diff, meta). Labels are Bug, SQL, regression. Fix #1 (ArrayTypeDriver, cairo storage) and #11/#13 (CompiledFilterIRSerializer, JIT) are Core-engine,
  not SQL-surface. Add Core (consider storage). Performance is correctly absent — the JIT changes are correctness fixes, not optimizations.

  m4 — 5+ digit integer SQL literals without underscores in new tests (in-diff, style). CLAUDE.md mandates underscores for 5+ digit numbers. Examples: DistinctTest (278444 * t0.x);
  CoveringIndexTest 20000000, 86400000000; CompiledFilterTest 78941, 5000000000 (this one also overflows int — 5_000_000_000 would read clearest), 939722, 403251, etc. Cosmetic; apply only
  to genuine integer literals (leave timestamps, decimals, fuzzer seeds, and result-assertion rows that mirror raw DB output).

  m5 — missing cross-context execution tests (in-diff, coverage). (a) getScanDirection (FIX 6) is consumed by window and join factories, but no test runs a window function (row_number() OVER
  (ORDER BY ts)) or a join over a multi-key covering IN that relies on the advertised FORWARD direction — SAMPLE BY, the hardest consumer, is covered. (b) The
  qualified-column-under-DISTINCT change (FIX 15/17) visibly alters window-model generation (SqlParserTest/PgAttributeFunctionFactoryTest goldens shift), but only parser-model assertions
  changed — no execution-level test confirms a window function over a single-table qualified-column query returns correct rows. (c) SequentialRowCursorFactory.close() (the non-heap branch of
  FIX 4) has no dedicated leak test (the three regression tests all drive the heap branch).

  m6 — assorted style nits (in-diff). CoveringIndexTest:~1978 uses cast(NULL as SYMBOL)/cast('A' as SYMBOL) where CLAUDE.md prefers NULL::SYMBOL/'A'::SYMBOL (the surrounding lines use ::TYPE
  correctly). allOrderByColumnsPresentIn is a boolean method without the is.../has... prefix (matches the neighboring columnNotExistsInJoinModels, so local precedent). One non-ASCII em-dash
  in an ArrayAsMapKeyTest comment (outside the LOG-only ASCII rule, but the repo prefers plain ASCII). extractWorkerFunctionsByFlag introduces a naming split from its
  compileWorker*Conditionally siblings (the rename is justified — it no longer extracts conditionally — just flagging the local inconsistency).

  Downgraded (investigated and cleared)

  No agent finding was a false positive on the code, but I probed the high-risk surfaces and cleared each against source — documenting so the reasoning is auditable:

  - ArrayTypeDriver.getPlainValueSize off-by-one — correct. Null arrays write exactly 8 bytes with NULL_LEN = -1; the totalSize <= 0 → Long.BYTES branch exactly mirrors getPlainValue:387; a
  real (incl. empty/0-cardinality) array always has totalSize ≥ 8. The sole consumer OrderedMapVarSizeRecord:603 now lands on the correct slot boundary. Reverting it makes ArrayAsMapKeyTest
  fail with the exact typeTag ... is not ARRAY assert.
  - getScanDirection FORWARD claim under O3 — honest. The multi-key non-latest path is a row-id k-way merge; row-id order == designated-timestamp order within a partition (O3 physically
  sorts partitions at commit), partitions iterate ASC → globally ts-ascending. Same invariant + precedent FilterOnValuesRecordCursorFactory.getScanDirection() already relies on in
  production. The override is strictly more honest than master (it downgrades multi-key LATEST BY to OTHER and stops the old false-FORWARD concatenation claim).
  - SymbolConstant.valueOf/valueBOf contract change — blast radius contained. VALUE_IS_NULL = Integer.MIN_VALUE; a non-null SymbolConstant only ever produces key 0, so valueOf(VALUE_IS_NULL)
  is reachable only via the aggregate empty/null path (the fix target) or SymbolConstant.NULL (value already null). All callers scanned; non-aggregate callers short-circuit before passing
  the sentinel. Reverting it fails all five First/Last/Mode/FirstNotNull/LastNotNull tests.
  - SymbolJoinKeyMapping.of(RecordCursor) now abstract — all 3 implementors override it; the only caller (ASOF Light) passes null for the non-single-symbol case (never invoked), and
  toStaticSymbolTable's AssertionError is gated unreachable by the isSymbolTableStatic contract. Strictly safer than the old default-throws.
  - Per-worker GROUP BY function ownership (FIX 7) — clean graph: COLUMN→null, VIRTUAL→factory/atom, GROUP_BY→factory or freeWorkerFunctionsByFlag; disjoint flags, no leak/double-free/NPE.
  The one apparent double-free (single-horizon catch + atom close) is neutralized by Misc.freeObjList element-nulling + idempotent DirectArray.close(), and actually closes a pre-existing
  leak. The COLUMN thread-safety skip is safe (column keys are read natively by the RecordSink, never invoked as functions). Reverting the rewrite fails the const-array leak tests with
  NATIVE_ND_ARRAY, difference: 32.
  - createFullFatJoin cross-type SYMBOL key (FIX 12/14) — correct i-th master/slave key pairing; only STRING/VARCHAR masters reach this branch; the slave SYMBOL key reads correctly via
  getStrA()/getVarcharA() from the map; slaveTypes override makes NullRecordFactory use Str/VarcharConstant.NULL, killing the real getVarcharSize crash with no new one.
  - FilterOn* _close() + Heap/SequentialRowCursorFactory.close() (FIX 4) — fixes a real pre-existing leak (reverting yields NATIVE_INDEX_READER, difference: 72); each FilterOn* owns a
  distinct factory; double-close is idempotent via freeObjListAndClear; super._close() doesn't touch the row cursor factory.
  - TablePageFrameCursor promotion (FIX 3) — fixes a reachable ClassCastException at SelectedRecordCursorFactory:153; the throwing 2-arg of() override is never invoked (the cursor is built
  via its own package-private 3-arg of); getTableReader/setScanProfile delegation is correct.
  - InIPv4FunctionFactory (FIX 18) — NULL semantics consistent with InLong/InDouble; the LongHashSet(LONG_NULL) sentinel is necessary (broadcast -1 collides with IntHashSet's default) and
  never false-collides; const/runtime branching and init() rebuild correct; no native leak (on-heap LongHashSet, constants GC-managed); in(Xv) dispatch wins for IPv4 LHS by exact-match
  priority.
  - JIT long-literal widening (FIX 11/13) — no over- or under-widening divergence is constructible (an INT-arith subtree compared against a LONG literal forces the Java (LL) overload, so the
  scalar filter widens too); only CONSTANT-node tokens reach the branch (column refs are LITERAL nodes); isReservedConstantKeyword is load-bearing (reverting fails the null-token test).
  - analyseEquals cleanup — safe: the lhi == rhi case returns above (:1652), so the unconditional linkDependencies(min,max) always has min != max.
  - SqlOptimiser qualified strip-retry (FIX 8/15/17) — survived ~30 adversarial queries; single-base guard + bare-projection restriction are identity-preserving; the multi-join branch is
  byte-identical to master.
  - Multi-key k-way merge "should be a heap" — cleared as not actionable here. It's O(N·k), but it is pre-existing (landed with #6861/#7182, untouched by this PR), and io.questdb.std has no
  binary-heap primitive — the cited precedent HeapRowCursor is itself backed by IntLongSortedList (a sorted array, also O(k)). A real heap would be a separate codebase-wide change. (One
  in-file comment calling HeapRowCursor "O(log N)" is factually wrong — O(k) — worth correcting if touched.)
  - maxRowsPerFrame field out of alphabetical order — false positive (this was in the posted review). The PR did not widen any field's visibility;
  maxRowsPerFrame/isExhausted/pendingCoveringCursor/tableReader were already protected (from #7145) and appear only as unchanged context lines. No ordering violation.

  Summary

  Verdict: Approve. After an exhaustive level-3 pass I found no correctness, resource-leak, concurrency, or security issue; every one of the 18 fixes was confirmed to fix a real bug without
  breaking sibling paths, and three close latent unsoundness in unchanged code.

  - 0 critical, 3 moderate (M1–M3), 6 minor (m1–m6).
  - Verified vs dropped: ~13 agent findings + ~13 high-risk hypotheses adjudicated; 0 confirmed false positives in the code, and I overturned one false positive from the previously-posted
  review (the maxRowsPerFrame ordering claim). The one genuinely new finding beyond that review is M1 (the FIX 9 positional/qualified ORDER BY regression) — a structural perf regression I
  confirmed from the pass ordering and the raw-token check, not just an agent claim.
  - In-diff vs out-of-diff: of the 9 findings, 6 are in-diff (M1, M3, m1, m2, m4, m5/m6), 3 are out-of-diff/process (M2 bundling, m3 labels, and the no-issue-ref note). The cross-context
  pass did not underrun: I re-ran SAFE/BROKEN verdicts on every high-blast-radius symbol (SymbolConstant.valueOf, SymbolJoinKeyMapping.of, getScanDirection, the TablePageFrameCursor casts,
  the three FilterOn*._close(), the per-worker compile paths) and found zero broken callsites.
  - Tradeoffs to weigh: M1 is the only finding I'd genuinely consider acting on before merge (it's a silent regression in the optimization's own target query class, though perf-only); M2 is
  your call as author; everything else is optional polish.

  The working tree is clean and matches the committed PR (HEAD 9c9ec00bcf) — the scratch files and experimental reverts that the parallel agents mention in their logs were confined to their
  isolated sandboxes and never touched the canonical tree.

bluestreak01 (dismissed):
Approve. Reviewed against `origin/master` (which already carries #7182), so the real covering-index delta here is small (~+33/-5: IN-key dedup, `getScanDirection()`, and the `TablePageFrameCursor` promotion) -- the row-id merge and descending scan are upstream and out of scope.

No correctness, NULL, resource-leak, concurrency, double-free, or security issue survived verification. Each of the 18 fixes was confirmed to fix a real bug without breaking sibling callsites, and three additionally close latent unsoundness in unchanged code (the `SelectedRecordCursorFactory` ClassCastException, the `FilterOn*`/row-cursor-factory leak, and the false-FORWARD scan-direction advertisement for multi-key LATEST BY). Specifically cleared with source proof: the per-worker function ownership (no double-free -- the inner projection list is never bulk-freed and each flag reaches exactly one owner), and the JIT narrow-i64 widening (per-comparison predicate scoping means a LONG literal can never over-widen narrow arithmetic in a different comparison).

Non-blocking follow-ups, in priority order:

- Test coverage for three of the fixes' own behavior changes is incomplete (the code is correct; the verification is thin):
  - The horizon-join half of the per-worker array-key leak fix is exercised only by `ParallelHorizonJoinFuzzTest.testParallelHorizonJoinConstArrayKeyDoesNotLeak`, gated behind `Assume.assumeTrue(enableParallelHorizonJoin)` with `rnd.nextBoolean()`, so it runs ~50% of CI runs and has no deterministic backstop (the group-by half does, via `ArrayAsMapKeyTest`). A non-fuzz `assertMemoryLeak` horizon-join test with the property forced on would close it.
  - `SequentialRowCursorFactory.close()` (the Cursor-order branch of the FilterOn* leak fix) is untested -- the three regression tests all take the Heap branch. A `count()` variant of the throwing-filter tests reaches it.
  - The `getScanDirection()` change (multi-key LATEST BY -> `SCAN_DIRECTION_OTHER`) has no test asserting the sort is retained for `... IN (a, b) LATEST ON ts PARTITION BY sym ORDER BY ts`, which the old false-FORWARD would have wrongly elided.

- Bundling: 18 technically-independent fixes land as one squash commit, which defeats per-fix bisect/revert. Your call as author; a note in the body that the bundle is intentional would help future bisectors.

- Minor: the IN-key dedup is O(n^2) (`!multiKeys.contains(key)` over an `IntList`) -- a scratch `IntHashSet` for membership, keeping the `IntList` for positional indexing, makes it O(n); `InIPv4FunctionFactory` always probes a `LongHashSet` even for 1-2 element lists where `InLongFunctionFactory` specializes to register compares; `perWorkerKeyFunctions` isn't nulled after transfer (benign idempotent double-free on a factory-ctor throw); add the `Core` label.

The `WindowFunctionTest` reorder (+12k/-12k, `@Test` count unchanged) is pure churn and inflates the diff but loses no coverage.

bluestreak01:
Heads up since the diff stats look alarming after I merged master in — **no `WindowFunctionTest` cases were removed.** Verified:

| | `@Test` methods | file size |
|---|---|---|
| branch before merge | 602 | 21,064 lines |
| master | 602 | 20,980 lines |
| branch after merge | 602 | 20,986 lines |

Same 602 methods (identical method-name sets), near-identical file size, and the full `WindowFunctionTest` suite passes locally.

The PR's line count is measured against the merge-base, and merging master moved that base forward:

- Before, the merge-base (`8a53c15898`) predated this window-function test suite, so its ~12k lines all counted as additions in the PR.
- After merging, the base **is** master, which already contains the same tests — so they cancel out. Shared work, not deleted.

Concretely, `WindowFunctionTest` went from ~24.7k changed lines to ~6. The branch's copy differed from master's only in assertion-API style — e.g. `timestamp("ts###desc")` -> `timestampDesc("ts")`, and `explain ... returnsOnce("QUERY PLAN...")` -> `assertsPlan(...)`. Same queries, same expected output. I took master's version and re-added the 6 `.noRandomAccess()` lines the optimiser changes on this branch require.

tl;dr: the window tests are all still here and green; they just live in the base now rather than showing up in the diff.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 208 / 223 (93.27%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/constants/SymbolConstant.java|2|3|66.67%|
|:large_blue_circle:|io/questdb/griffin/engine/table/CoveringIndexRecordCursorFactory.java|4|5|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SymbolJoinKeyMapping.java|6|7|85.71%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/bool/InIPv4FunctionFactory.java|66|72|91.67%|
|:large_blue_circle:|io/questdb/jit/CompiledFilterIRSerializer.java|26|28|92.86%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|47|49|95.92%|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|41|43|95.35%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SymbolToSymbolJoinKeyMapping.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FilterOnExcludedValuesRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FilterOnSubQueryRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/StringToSymbolJoinKeyMapping.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/HeapRowCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SequentialRowCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/VarcharToSymbolJoinKeyMapping.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/arr/ArrayTypeDriver.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FilterOnValuesRecordCursorFactory.java|1|1|100.00%|

bluestreak01 (approved):
Approve. Independent level-3 pass — 12 adversarial agents with source access plus manual verification of the highest-risk surfaces (scan-direction honesty, the `in(Xv)` signature grammar). No correctness, NULL, resource-leak, concurrency, double-free, or security issue survived verification. Each of the 18 fixes was confirmed to fix a real bug without breaking sibling callsites, and three additionally close latent unsoundness in unchanged code: the `SelectedRecordCursorFactory` ClassCastException (FIX 3), the `FilterOn*`/row-cursor-factory native leak (FIX 4), and the false-FORWARD scan-direction advertisement for multi-key LATEST BY (FIX 6).

Specifically cleared with source proof:
- `getScanDirection()` is honest — `multiKeyCursor`/`latestBy` are final ctor-set fields (not stale), the default was already FORWARD so the override only downgrades multi-key LATEST BY to OTHER, and the non-latestBy row-id merge is genuinely ts-ascending.
- The per-worker GROUP BY function ownership has no double-free — `perWorkerInnerProjectionFunctions` is block-local and never bulk-freed, each flag reaches exactly one owner, and `Misc.freeObjList` element-nulling makes the error-path catch idempotent.
- The JIT narrow-i64 widening cannot diverge from the scalar filter — the widen observer is strictly per-comparison, and an INT-arith subtree vs a LONG literal resolves uniquely to the `(LL)` overload so Java widens the subtree too.
- `SymbolConstant.valueOf`/`valueBOf` blast radius is contained — a non-null constant's `getInt` only ever yields key 0, so `VALUE_IS_NULL` reaches it only via the aggregate empty/null path, which wants null.

Non-blocking follow-ups, in priority order:
- Two of the fixes' own behavior changes have only thin verification (the code is correct):
  - The horizon-join half of the per-worker array-key leak fix runs only under `ParallelHorizonJoinFuzzTest` gated behind `Assume.assumeTrue(rnd.nextBoolean())` (~50% of CI runs), with no deterministic backstop — the GROUP BY half has one via `ArrayAsMapKeyTest`. A non-fuzz `assertMemoryLeak` horizon-join test with the property forced on would close it.
  - `SequentialRowCursorFactory.close()` (the Cursor-order branch of the FilterOn* leak fix) is untested — the three throwing-filter tests all take the Heap branch. A `... ORDER BY sym` variant reaches it.
- Bundling: 18 technically-independent fixes land as one squash commit, which defeats per-fix bisect/revert. Your call as author; a note in the body that the bundle is intentional would help future bisectors.
- Minor: the multi-key IN dedup is O(n^2) (`!multiKeys.contains(key)` over an `IntList`) — a scratch `IntHashSet` for membership, keeping the `IntList` for positional access, makes it O(n); `longConstantTypeCode` recomputes `floatConstantTypeCode` the caller already evaluated; add the `Core` label (FIX 1 is cairo storage, FIX 11/13 are JIT, not SQL-surface); a few 5+ digit integer literals in the new tests lack underscore separators.

The current HEAD already contains the normalized `allOrderByColumnsPresentIn` (positional + qualified handling), so the earlier positional-ORDER-BY LIMIT-push regression is resolved — `ORDER BY 5 DESC LIMIT 10` pushes correctly again.

PR #7156, opened 2026-05-26
perf(sql): speed up cached window queries and lower memory footprint

### LINKED ISSUE #7154: Sub-optimal plan used for window functions with ORDER BY ts DESC

```sql
qdb=> EXPLAIN
select
  symbol, timestamp,
  lag(timestamp, 1) OVER (PARTITION BY symbol ORDER BY timestamp DESC)
from test_table
limit 100;
                                              QUERY PLAN
------------------------------------------------------------------------------------------------------
 Limit value: 100 skip-rows-max: 0 take-rows-max: 100
     CachedWindow
       orderedFunctions: [[timestamp desc] => [lag(timestamp, 1, NULL) over (partition by [symbol])]]
         PageFrame
             Row forward scan
             Frame forward scan on: test_table
```

Here we sort the rows which is slow and OOM-prone on large tables. In case of `ORDER BY timestamp ASC` we skip the sort:
```sql
qdb=> EXPLAIN
select
  symbol, timestamp,
  lag(timestamp, 1) OVER (PARTITION BY symbol ORDER BY timestamp ASC)
from test_table
limit 100;
                               QUERY PLAN
-------------------------------------------------------------------------
 Limit value: 100 skip-rows: 0 take-rows: 100
     Window
       functions: [lag(timestamp, 1, NULL) over (partition by [symbol])]
         PageFrame
             Row forward scan
             Frame forward scan on: test_table
```

_No response_

_No response_

### Full Name:

Andrei Pechkurov

### Affiliation:

QuestDB

### Additional context

_No response_

kafka1991:
Fixes #7154.

Speeds up window queries with ORDER BY and lowers their per-row memory footprint, with the largest gains on wide tables read from random-access bases.

## Summary

- **Light cached window factory.** New `CachedWindowLightRecordCursorFactory` skips materialising base columns into the chain when the base cursor supports random access. It stores only the deferred window output columns in the narrow chain plus a parallel `DirectLongList` of base rowIds keyed by dense row index, fetching base columns on demand via `baseCursor.recordAt`. Window functions see a `WindowLightRecord` that dispatches per-column reads to the live base record or the narrow chain via a single sign-encoded `IntList` source map. `SqlCodeGenerator` routes to it when `base.recordCursorSupportsRandomAccess()` is true, every ordered group is encoded-sort-eligible, and the new `cairo.sql.window.cached.light.enabled` flag is on (default `true`). The tree fallback is reserved for the non-light factory to avoid O(N log N) random base reads on cold bases.

- **Encoded sort buffer for cached window queries.** `CachedWindowRecordCursorFactory` selects between two sort strategies per ordered window group via a new `WindowSortBuffer` interface. `EncodedWindowSortBuffer` packs sort keys plus rowIds into a flat `DirectLongList` and finalises with `Vect.sortEncodedEntries` (parallelised above a configurable threshold). When the cursor knows the base row count, the buffer pre-sizes its backing memory up front to avoid geometric reallocations. `TreeWindowSortBuffer` retains the original `LongTreeChain` + `RecordComparator` path for groups whose sort keys do not fit the encoded layout. `SqlCodeGenerator` skips comparator bytecode generation for groups that will go down the encoded path.

- **`lead(ts, 0)` / `lead(m, 0)` return TIMESTAMP/DATE.** `LeadTimestampFunctionFactory` and `LeadDateFunctionFactory` no longer route `offset == 0` through `LagLongFunctionFactory.LeadLagValueCurrentRow`; they use their TIMESTAMP/DATE counterpart so `lead(ts, 0)` and `lead(m, 0)` agree with `lag(ts, 0)` / `lag(m, 0)` on result type instead of returning LONG.

## Tradeoffs

- The light path only fires when the base cursor supports random access AND every ordered group fits the encoded sort buffer. Mixed-key window queries (e.g. one group on VARCHAR ORDER BY) fall back to the non-light factory.

- The light path replaces full chain materialisation with one `baseCursor.recordAt(...)` per row during pass1/pass2 ordered iteration and external iteration. Whether this is a net win depends on the base cursor's random-access cost. For `PageFrameRecordCursor` bases it is cheap; for filter-wrapped async chains the gain shrinks. The `cairo.sql.window.cached.light.enabled` flag (default `true`) allows opt-out.

- The encoded sort buffer pre-allocates `SqlWindowTreeKeyPageSize` (4 KB default) per ordered group, vs. the red-black tree's 4 KB key heap + 1 KB rowid heap. Comparable footprint for small queries; the encoded buffer grows by doubling when the row count is unknown, or sizes up front when the base cursor reports a row count.

- `lead(ts, 0)` / `lead(m, 0)` now return TIMESTAMP/DATE instead of LONG, a behavioural change for callers depending on the old return type.

## Test plan

- Run `WindowFunctionTest`, `MaxLongWindowFunctionTest`, `PercentRankFunctionTest`, `EmaWindowFunctionTest`, `VwemaWindowFunctionTest`, `WindowDecimalFunctionTest`, and `NamedWindowFuzzTest` with `cairo.sql.window.cached.light.enabled` toggled both ways; outputs must match between the light and non-light paths.
- Verify EXPLAIN on `SELECT *, LAG(ts, 1) OVER (PARTITION BY symbol ORDER BY ts DESC) FROM t LIMIT 100` shows the `CachedWindowLight` plan label.
- Verify a window query with `OVER (... ORDER BY <varchar>)` routes through the non-light factory (encoded sort not eligible).
- `testLagOverDescDuplicateTimestampRegression` checks that `LAG OVER (ORDER BY ts DESC)` over duplicate designated-timestamp values matches a genuine DESC sort (ascending-rowId tie-break), the same result master produces.


nwoolmer (dismissed):
```
LEAD remains ONE_PASS, so a swapped LAG OVER ts DESC still uses CachedWindow rather than streaming Window. The sort is removed but the chain is still materialised.
```

The fix is wrong imo, if anything we should be rewriting LEAD -> LAG in these cases, which supports non-cached (streaming) window.

Please can you share benchmarks to demonstrate what performance gains you expect from these changes?

kafka1991:
Bench on 100M rows, `LAG(ts, 1) OVER (PARTITION BY symbol ORDER BY ts DESC)`:

```
master:  43.73s
patch:    2.56s     (~17x)
```

The rewrite swaps `LAG over DESC` to its per-row equivalent `LEAD over ASC`. The O(N log N) `LongTreeChain` sort goes away; nothing else changes.

The patch handles `LEAD over DESC` -> `LAG over ASC` too, which hits streaming `Window` (LAG is `ZERO_PASS`). LEAD-only would leave the LAG case (the 17x) unfixed.

nwoolmer:
The user's scenario was actually ~2 billion rows, and wide ones. What is the behaviour on LIMIT, and how much memory is still allocated by the CachedWindow? Even without the sort, I suspect this is still liable to OOM?:

kafka1991:
This PR adds `CachedWindowLightRecordCursorFactory`, used whenever `base.recordCursorSupportsRandomAccess()` is true. It only materialises the deferred window output columns plus the base rowId (embedded as column 0 of the narrow chain) and fetches base columns on demand. The routing applies to every cached window query (RANK, ROW_NUMBER, AVG OVER, NTH_VALUE, Window with order by etc.), not just LAG.

For a `LAG(ts, 1)` output, per-row memory drops:
- 5-column table: ~40 B (full chain) -> 16 B
- 10-column table: ~80 B -> 16 B

For the ~2 B-row scenario: 2 B x 16 B is still ~32 GB of narrow chain. Note that light pushes the ceiling out a few times but does not eliminate the OOM risk at that scale.

nwoolmer (commented):
I think this will be good in conjunction with this PR: https://github.com/questdb/questdb/pull/7163

7156 reduces the painfulness of CachedWindow, and 7163 reduces the number of cases where CachedWindow is required for this sort of query.

bluestreak01:
@kafka1991 pls use assertQuery() in all window function tests

kafka1991:
> @kafka1991 pls use assertQuery() in all window function tests

Done. Replaced all existing `assertSql` with `assertQuery` in Window Function Tests.

bluestreak01:
@kafka1991 — code review against PR head `619239f0d3...`. Listing the issues I'd want addressed before merge; the rest are nits.

## Critical (blocks merge)

### 1. Native memory leak window in `CachedWindowRecordCursorFactory` ctor

`core/src/main/java/io/questdb/griffin/engine/window/CachedWindowRecordCursorFactory.java:115`

```java
} catch (Throwable t) {
    Misc.freeObjList(sortBuffers);
    Misc.free(recordChain);
    throw t;
}

this.cursor = new CachedWindowRecordCursor(columnIndexes, recordChain, sortBuffers); // <-- outside try
```

If `new CachedWindowRecordCursor(...)` throws (OOM), `this.cursor` is never assigned, `sortBuffers` and `recordChain` go out of scope, and their native pages leak. The outer catch calls `close()` → `Misc.free(cursor)` which is null-safe and frees nothing.

The new LIGHT factory at `CachedWindowLightRecordCursorFactory.java:107-117` already has the correct pattern (cursor construction inside the inner try). This is a regression when the non-LIGHT factory was refactored to share the `WindowSortBuffer` interface.

**Fix:** move line 115 inside the inner try.

### 2. Enterprise CI failed, no follow-up

`questdb-butler` posted Enterprise CI Failed for `619239f0d3...` (Azure build 236810). No comment from the author or reviewer states the failure is investigated or resolved.

## Moderate

### 3. `TreeWindowSortBuffer` LIGHT path → O(N log N) random base reads

`CachedWindowLightRecordCursorFactory.java:100-104`, `TreeWindowSortBuffer.java:94`, `LongTreeChain.java:120`

When the LIGHT factory picks `TreeWindowSortBuffer` (sort key not byte-encodable: `VARCHAR`, `STRING`, `BIN`, `UUID`, `INTERVAL`, `ARRAY`, `LONG256`, or total > 32 bytes), each tree compare calls `sourceCursor.recordAt(rightRecord, rowId(ref))` → `positionRecordB` → `baseCursor.recordAt(...)`. That's O(N log N) random base lookups during tree insertion.

The non-LIGHT factory does O(N log N) reads against a contiguous chain memory block. For a 10M-row query with `OVER (ORDER BY varchar_col)` on a partitioned table over cold storage, this could regress 10-100x. And there is no test that would catch it (see #11).

**Fix options:** (a) restrict the LIGHT gate to encoded-supported keys, falling back to non-LIGHT when any ordered group cannot use `EncodedWindowSortBuffer`; or (b) project the tree-sort key columns into `narrowChainTypes` so the comparator reads from local memory.

### 4. `SortKeyEncoder` ctor mid-loop OOM leaks `DirectIntList`s

`core/src/main/java/io/questdb/griffin/engine/orderby/SortKeyEncoder.java:82-96`

The constructor loop does `rankMaps.add(new DirectIntList(1024, NATIVE_DEFAULT, true))` per symbol column with no try/catch. If iteration *i* throws, allocations from `0..i-1` leak. Mirror the `LongTreeChain` pattern: `try { ... } catch (Throwable th) { Misc.freeObjList(rankMaps); throw th; }`.

### 5. `TreeWindowSortBuffer` ctor leaks `rankMaps` on `LongTreeChain` OOM

`TreeWindowSortBuffer.java:46-59`, callers `CachedWindowLightRecordCursorFactory.java:100-104` and `CachedWindowRecordCursorFactory.java:102-107`

Caller passes `SortKeyEncoder.createRankMaps(...)` inline. If `createRankMaps` succeeds (allocating native rankMaps) and `new TreeWindowSortBuffer(...)` then fails inside `new LongTreeChain(...)`, the half-built buffer is unreachable — `rankMaps` was never added to `sortBuffers` so the factory's catch cannot free it. Wrap the buffer ctor body in try/catch that frees `rankMaps`.

### 6. Forward unordered Pass1 should be a sequential scan

`CachedWindowLightRecordCursorFactory.java:459-468`

The fused forward loop does `positionRecordA(rIdx)` for `rIdx = 0..size-1`. Every iteration walks the narrow chain to recover `baseRowId` and calls `baseCursor.recordAt(...)`. Since `rIdx` is dense forward, the N-th lookup yields exactly the row the N-th `baseCursor.hasNext()` would naturally produce. Replace with `baseCursor.toTop(); while (baseCursor.hasNext()) {...}` to turn 100M random accesses into a sequential scan. Symmetric for the backward path.

### 7. Encoded entry stores dense rowIndex — should store baseRowId

`CachedWindowLightRecordCursorFactory.java:406`, `EncodedWindowSortBuffer.java:113-117`, `CachedWindowLightRecordCursorFactory.java:430-438`

The sort buffer stores the dense rowIndex. Step 2 retrieves it, then `positionRecordA(rIdx)` walks the narrow chain just to recover the `baseRowId` that the buffer could have stored directly. Save one auxMem read + one chain-record load + one `getLong(0)` per row per ordered group in Pass1 and Pass2. The tree path still needs the dense index, so the encoded path needs a different rowId encoding.

### 8. Per-row per-function direction branch not hoisted

`CachedWindowLightRecordCursorFactory.java:462-467, 474-479`

After splitting the row-outer loop into forward/backward halves, the inner `if (f.getPass1ScanDirection() == FORWARD)` still runs per (row × function). `getPass1ScanDirection()` is a final per-function value — partition `unorderedFunctions` into `forwardFunctions` / `backwardFunctions` at construction time and remove the inner branch entirely.

### 9. Narrow chain page-size config diverges from non-LIGHT factory

`CachedWindowLightRecordCursorFactory.java:87-88`

LIGHT uses `getSqlWindowTreeKeyPageSize`/`MaxPages` (default 512 KB). Non-LIGHT uses `getSqlWindowStorePageSize`/`MaxPages` (default 1 MB). The narrow chain is row storage, not tree-key memory. A user tuning `cairo.sql.window.tree.max.pages` down to cap sort memory will inadvertently throttle the narrow chain. Use the store config or add a dedicated property.

### 10. `lead(ts,0)` / `lead(m,0)` return-type change is backwards-incompatible

`LeadTimestampFunctionFactory.java:74`, `LeadDateFunctionFactory.java:73`

Before this PR, `lead(ts,0)` returned LONG. Now it returns TIMESTAMP/DATE. Necessary for the LAG↔LEAD swap. Tradeoffs section flags it; please confirm release notes call it out and check no MatView/cached factory schemas persist the old LONG type.

### 11. Coverage gap: TreeWindowSortBuffer not exercised through LIGHT path

No new test in `WindowFunctionTest.java` / `WindowDecimalFunctionTest.java` uses `OVER (... ORDER BY <varchar|string|bin|long256|interval> ...)`. The tree fallback on the LIGHT factory is unverified end-to-end. Combined with #3, the regression risk is real *and* unsurfaced by tests.

### 12. Coverage gap: parallel sort threshold not crossed

Default `cairo.sql.sort.encoded.parallel.threshold = 1_024_000`. Largest window-test dataset is 1,000,000 rows (`WindowFunctionTest.java:2430`). The parallel branch inside `Vect.sortEncodedEntries` is not covered. Either insert > 1M rows in a dedicated test or set the threshold to a small value.

### 13. PR body has no Test plan section

CLAUDE.md PR template expects `## Test plan` with `-` bullets. Body has `## Summary` and `## Tradeoffs` only.

## Minor

- **`RecordChain` 2-arg constructor is dead code** (`RecordChain.java:72-78`). Coverage 0/2. Every callsite uses the 4-arg form. Remove.
- **WindowLightRecord coverage gap** for GEOHASH / LONG256 / LONG128 / INTERVAL / BIN / ARRAY accessors (53/96 lines). Add a CachedWindowLight test over a table with these column types.
- **WindowLightRecord SYMBOL/VARCHAR NPE forward-compat hazard**: narrow chain has no `setSymbolTableResolver`. Currently unreachable (no window function returns SYMBOL/VARCHAR/STRING) but the original non-LIGHT factory wires up the resolver; the LIGHT path silently regresses if a SYMBOL-returning window function is ever added. Add an `assert` at `SqlCodeGenerator.java:9815` that `m.getColumnType()` is fixed-width and not SYMBOL.
- **`EncodedWindowSortBuffer.finishPut` design fragility** (line 89-105): unconditionally frees encoder; reopen relies on `DirectIntList.setCapacity` being safe on a closed list. Functional but fragile — move encoder cleanup to `close()`.
- **`circuitBreaker.statefulThrowExceptionIfTripped` inside ordered-group inner loop** (`CachedWindowLightRecordCursorFactory.java:404-407, 410-413`). Once-per-row is sufficient. Pre-existing in non-LIGHT too.
- **`CachedWindowRecordCursor` field ordering** (`CachedWindowRecordCursorFactory.java:283-285`): `sortBuffers` should be alphabetically after `recordChain`.
- **`instanceof Reopenable`** at `CachedWindowLightRecordCursorFactory.java:576-578` and `CachedWindowRecordCursorFactory.java:501-503` can use a Java 17 pattern variable.
- **18 `assertSql("plan...", "explain query")` callsites** in `WindowFunctionTest.java`, `PercentRankFunctionTest.java`, `EmaWindowFunctionTest.java` should be `assertPlanNoLeakCheck(query, plan)` per @bluestreak01's review comment — the new test blocks correctly use `assertPlanNoLeakCheck` but old patterns were not migrated.
- **`WindowFunctionTest.java:7292`** uses `assertQuery(...)` inside `assertMemoryLeak` (nested). Use `assertQueryNoLeakCheck`.
- **`hasNonLagLeadSiblingInSortGroup`** logic is correct but load-bearing-by-coincidence: a sibling at a different direction is filtered by the `direction` check earlier — please add a comment that the loop tolerates in-place mutation because of this.
- **Pre-existing**: narrow-chain mid-Step-1 exception leaves stale rows (`CachedWindowLightRecordCursorFactory.java:401-413`). Same pattern in non-LIGHT factory.
- **Pre-existing**: `RecordArray.hasPrev` / `hasNext` non-defensive about position invariant.
- **Title grammar**: "reduce the memory use" reads off — consider "reduce memory usage" or "lower memory footprint".
- **PR body** opens with implementation-detail-heavy section names; a one-line user-facing opener would help.

## Summary

Verdict: request changes. The 17x speedup claim is plausible and the architecture is sound, but four issues should block merge: the leak window (#1), the LIGHT-path tree-sort regression risk for var-width keys (#3) combined with no test that would catch it (#11), and the unaddressed Enterprise CI failure (#2). Findings #4-5 are also worth fixing.

Findings #6-8 are missed optimizations on top of the claimed gain — easy wins.

---

*Generated by a Claude Code level-2 review (8 specialist agents + verification pass). 25 verified findings; 4 false positives dropped; all in-diff. Happy to drill into any item.*

kafka1991:
Thanks @bluestreak01. Addressed:

**Fixed:**
- Moved `CachedWindowRecordCursor` construction inside the inner try in `CachedWindowRecordCursorFactory` so an OOM during cursor allocation no longer leaks `sortBuffers` / `recordChain`.
- Restricted the LIGHT dispatcher to queries where every ordered group is encoded-sort-eligible. VARCHAR / STRING / BIN / >32 B sort keys fall back to the non-light factory, eliminating the `O(N log N)` random base reads risk during tree.put. Removed the now-unreachable tree branch and `comparators` param from the light factory.
- Wrapped the `SortKeyEncoder` ctor loop in try/catch so a mid-loop OOM frees the partially-allocated `DirectIntList` rank maps.
- Wrapped `TreeWindowSortBuffer`'s `LongTreeChain` construction so the caller-provided `rankMaps` are freed if tree allocation fails.
- Switched the narrow chain in the light factory from `SqlWindowTreeKeyPageSize` / `MaxPages` to `SqlWindowStorePageSize` / `MaxPages` (narrow chain is row storage, not tree-key memory).
- Added `## Test plan` to the PR body.
- Title grammar fixed.
- Dead 2-arg `RecordChain` ctor removed (was 0/2 coverage).
- Field ordering in `CachedWindowRecordCursor`: `recordChain` before `sortBuffers`.

**New test coverage:**
- `WindowFunctionTest.testCachedWindowLightVarcharOrderByFallsBackToNonLight` — pins the light flag on, asserts EXPLAIN routes a `ORDER BY varchar_col` query to plain `CachedWindow`. Locks down the new gate behavior.
- `WindowFunctionTest.testCachedWindowLightCumeDistTwoPassUnderLightFactory` — pins the light flag on and verifies `cume_dist() OVER (ORDER BY v)` results, deterministically exercising the two-pass branch under the light factory.
- `WindowFunctionTest.testCachedWindowLightLagSwapBlockedByMixedSortGroup` — EXPLAIN locks down `hasNonLagLeadSiblingInSortGroup` blocking the LAG -> LEAD swap when LAG shares a sort group with `SUM(...) OVER (...)`.
- `RecordArrayTest.testFixedWidthSchemaSkipsAuxMem` — round-trips 1k rows on an all-fixed-width chain (`varOffset == 0`, `auxMem == null`), validating the auxMem-skip path via both `recordAtRowIndex` and `hasNext`.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 579 / 646 (89.63%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/window/WindowLightRecord.java|55|96|57.29%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyEncoder.java|12|15|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/window/TreeWindowSortBuffer.java|29|33|87.88%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|77|81|95.06%|
|:large_blue_circle:|io/questdb/griffin/engine/window/CachedWindowLightRecordCursorFactory.java|263|275|95.64%|
|:large_blue_circle:|io/questdb/griffin/engine/window/EncodedWindowSortBuffer.java|55|57|96.49%|
|:large_blue_circle:|io/questdb/griffin/engine/window/CachedWindowRecordCursorFactory.java|28|29|96.55%|
|:large_blue_circle:|io/questdb/griffin/engine/window/LightWindowSPI.java|14|14|100.00%|
|:large_blue_circle:|io/questdb/PropServerConfiguration.java|21|21|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LongTreeChain.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/RecordArray.java|14|14|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoConfiguration.java|3|3|100.00%|

core/src/main/java/io/questdb/griffin/engine/window/CachedWindowLightRecordCursorFactory.java
```diff
@@ -0,0 +1,597 @@
...
+        ObjList<WindowSortBuffer> sortBuffers = null;
+        DirectLongList baseRowIds = null;
+        try {
+            this.base = base;
+            this.orderedGroupCount = sortKeys.size();
+            assert orderedGroupCount == orderedFunctions.size();
+            this.orderedFunctions = orderedFunctions;
+            narrowChain = new RecordArray(
+                    narrowChainTypes,
+                    null,
```
bluestreak01, line 91: passing null into @Notnull field, annotation mismatch

kafka1991:
## Final Review Report — `perf(sql): speed up cached window queries and lower memory footprint`

Level-3 review (parallel specialist agents + per-finding source verification), diffed against `master`.

**Verdict: Approve, pending green CI.** No blocking issues remain.

### What the PR does
Adds a "light" cached-window execution path (`CachedWindowLightRecordCursorFactory`) that skips materializing base columns when the base cursor supports random access. It stores only the deferred window-output columns in a narrow `RecordArray` plus a parallel `DirectLongList` of base rowIds, fetching base columns on demand via `recordAt`. Sorting for each ordered group goes through a flat encoded-key buffer (`EncodedWindowSortBuffer` → `Vect.sortEncodedEntries`, parallel above a threshold) instead of a red-black tree, with `TreeWindowSortBuffer` retained for the non-light factory's non-encodable keys. `SqlCodeGenerator` routes to the light path only when the base supports random access, every ordered group is encoded-sort-eligible, and `cairo.sql.window.cached.light.enabled` (default `true`) is on. Separately, `lead(ts,0)` / `lead(m,0)` now return TIMESTAMP/DATE instead of LONG.

### Correctness & safety — verified
- **Light vs non-light result equivalence.** The ordered pass1 loop positions only the base record (`positionRecordABaseOnly`); correct because no cached window function reads a window-output column through the record — outputs are read via `spi.getAddress(rowIndex,…)` / `getRecordAt`, both rowId-addressed. Verified across RANK, ROW_NUMBER, AVG/SUM-OVER, NTH_VALUE, FIRST/LAST_VALUE, cume_dist, percent_rank, and the cached LAG/LEAD variants.
- **Encoded-sort ordering.** The native comparator tie-breaks equal keys by ascending stored rowId; the light path stores dense indices (ascending base-scan order), matching the red-black tree's insertion-order tie-break. Duplicate-key `ORDER BY … DESC` is correct (regression-tested). The encoder/native sort are pre-existing and shared with ORDER BY.
- **Resource lifecycle.** All constructor error paths free every native allocation exactly once (no leak, no double-free); early-free of single-pass sort buffers is idempotent and safe within the enforced getCursor/close contract; `RecordArray`'s nullable-`auxMem` and reordered `beginRecord` paths are correct for both fixed-width and var-width callers.
- **Parquet decode budget.** The light cursor propagates `ParquetDecodeHint.SCATTERED` to the base, so Parquet-backed bases use the full decode budget for the out-of-order pass1/pass2 reads rather than the MONOTONIC default (¼ budget / 4 buffers). The non-light factory correctly relies on the MONOTONIC default — it drains the base in a single sequential pass.
- **`lead(ts,0)` / `lead(m,0)` type change.** Intentional and contained: fixes a real asymmetry with `lag(ts,0)`, MatViews reject window functions at parse time (no persisted-schema break), and tests assert the new type and value.
- **Cross-context.** The only out-of-diff caller of the changed `RecordArray` contract (`MarkoutHorizonRecordCursorFactory`) uses absolute-offset access and is unaffected.

### Remaining findings (Minor, non-blocking)
1. **Redundant narrow-chain positioning (perf).** The backward pass1 and the two pass2 loops in `CachedWindowLightRecordCursorFactory` use `positionRecordA` (base + narrow) where `positionRecordABaseOnly` suffices — those passes read only base columns. Small per-row saving; also removes an inconsistency with the ordered pass1 loop.
2. **Early-free `isOpen` invariant asymmetry (robustness).** `computeWindow()` frees single-pass sort buffers while the cursor stays open — a latent fragility the non-light factory lacks. Currently unreachable (cursor close/reopen contract) and non-crashing; worth hardening by restoring `isOpen` unconditionally in `reopen()`/`of()`.
3. **Test coverage gaps.** No positive light-path test for 3–4 column composite keys (FIXED_24/FIXED_32 padding), none with NULLs *in the sort key*, and `RecordArrayTest.testFixedWidthSchemaSkipsAuxMem` does not exercise `hasPrev` on the auxMem-skip path (covered indirectly by `testSkipAndRefill`). The differential light-vs-non-light coverage (randomized flag flip across all window suites + explicit fallback/plan tests) is otherwise solid, and there are no `assertSql`/`returnsOnce` style violations.
4. **Diagnostics nit.** `EncodedWindowSortBuffer`'s memory-overflow message names no actionable config property.

### Tradeoffs (as documented in the PR)
- The light path swaps chain materialization for one `baseCursor.recordAt` per row: a large win on cheap random-access bases (~17× on the reported `LAG OVER ts DESC` case at 100M rows), smaller or negative on filter-wrapped async bases — hence the opt-out flag. It lowers per-row memory but does not eliminate OOM risk at multi-billion-row scale.
- `lead(ts,0)` / `lead(m,0)` returning TIMESTAMP/DATE is a deliberate compatibility change (carries the `Compatibility` label and is documented).

### Summary
The engineering is sound and the high-risk areas — light/non-light equivalence, encoded-sort tie-breaking, native-memory lifecycle, Parquet decode hinting, and the `lead` type change — are verified correct. The four remaining items are all Minor and non-blocking. Recommend merge once CI is green; addressing finding #1 (and optionally #2–#3) would be reasonable follow-ups but need not block.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 619 / 690 (89.71%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/window/WindowLightRecord.java|55|96|57.29%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyEncoder.java|12|15|80.00%|
|:large_blue_circle:|io/questdb/cairo/RecordArray.java|24|30|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/window/TreeWindowSortBuffer.java|31|35|88.57%|
|:large_blue_circle:|io/questdb/griffin/engine/window/EncodedWindowSortBuffer.java|64|68|94.12%|
|:large_blue_circle:|io/questdb/griffin/engine/window/CachedWindowLightRecordCursorFactory.java|282|293|96.25%|
|:large_blue_circle:|io/questdb/griffin/engine/window/CachedWindowRecordCursorFactory.java|74|76|97.37%|
|:large_blue_circle:|io/questdb/griffin/engine/window/LightWindowSPI.java|18|18|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|22|22|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyType.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/PropServerConfiguration.java|25|25|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LongTreeChain.java|5|5|100.00%|

bluestreak01:
@kafka1991 — independent **level-3 review** (parallel specialist agents + per-finding source verification) against PR head `3857fc0d37`. The earlier fixes from @bluestreak01's review all landed and verified. The core engineering is sound — light/non-light result equivalence, the sign-encoded column dispatch, the `RecordArray` memory pre-extension, the encoded-sort tie-break (checked against `ooo.cpp`), the native-memory lifecycle, and the `lead` type fix are all correct. Findings below are one CI blocker, one operational behavior change, hardening, and test gaps.

## Critical (blocks merge)

### C1. Enterprise CI is failing on the current head — unresolved, unproven-unrelated
`gh pr checks 7156` shows `enterprise-ci  fail  buildId=241798  "Enterprise tests failed"`, while the full OSS pipeline (`241797`) is green. This was flagged in the prior review (#2) and is **still red** on `3857fc0d37`; the resolution was never posted. A green OSS run does not clear an enterprise-only failure (enterprise has its own window/ACL/SecurityContext tests). Please root-cause and fix, or prove it's unrelated with evidence — it shouldn't be dismissed silently.

## Moderate

### M1. Encoded window sort buffer no longer honors the window-specific memory caps
`EncodedWindowSortBuffer.java:71-73` sizes its budget from `cairo.sql.sort.key.max.bytes` + `cairo.sql.sort.light.value.max.bytes`, **not** `cairo.sql.window.tree.max.bytes` / `cairo.sql.window.rowid.max.bytes` that the tree path used. This applies to **both** the light and the refactored non-light factory for every encoded-eligible ordered group — i.e. the common case (TIMESTAMP/LONG/INT/DATE/DECIMAL keys). A user who explicitly lowered `cairo.sql.window.tree.max.bytes` to bound window sort memory — the exact multi-billion-row OOM-mitigation scenario raised on this PR — silently loses that bound (encoded sort is then capped only by the default-unbounded, ~34 GB-clamped global sort caps). Defaults are unaffected. The fact that `CachedWindowMemoryCapTest` was rewritten to VARCHAR keys to keep the cap tests firing confirms this consequence was known but it's undocumented. Recommend honoring the window caps in the encoded buffer, or documenting it in Tradeoffs and making the overflow message name the now-relevant keys (`SortKeyEncoder.throwSortHeapOverflow` already does this).

### M2. `lead(ts,0)`/`lead(m,0)` return-type change is in the body but not the release-note-bound title
The LONG→TIMESTAMP/DATE change is well documented in Summary/Tradeoffs, but CLAUDE.md says the title description is what reaches release notes. As squash-merged, an upgrading user wouldn't learn that `lead(ts,0)` now returns TIMESTAMP, that `CREATE TABLE AS SELECT lead(ts,0) OVER (...)` now persists a TIMESTAMP column instead of LONG, or that PG-wire/BI clients keying off the old LONG type see a different OID. (No MatView break — window funcs are rejected at MatView parse time, verified.) Add an explicit breaking-change line that reaches release notes, or split the `lead` change into its own titled commit.

### M3. LIGHT gate has no fixed-width guard — a future var-width window output would segfault (latent)
The light narrow chain assumes every window-output column is fixed-width (`RecordArray.beginRecord` never initializes var-pointer slots; `LightWindowSPI.getAddress` only resolves fixed slots). This holds **today** — no window function outputs a var-width type (`count(varchar)`/`count(symbol)` return LONG; there's no `first_value`/`lag`/`lead`/`nth_value`/`max`/`min` over STRING/VARCHAR). But the gate in `SqlCodeGenerator.java:9843-9888` doesn't assert `narrowChainTypes` is all-fixed-width. If a var-width-returning window function is ever added, `varOffset>0` allocates `auxMem`, the uninitialized var pointers are read as offsets → `Unsafe` OOB → JVM abort (not a recoverable error). Add a `ColumnType.isVarSize` check that routes var-width outputs to the non-light factory (or at minimum an `assert`).

## Minor

- **Wasted per-row narrow positioning in both pass2 loops** — `CachedWindowLightRecordCursorFactory.java:521` and `:533` use `positionRecordA` (base **and** narrow) where `positionRecordABaseOnly` suffices. Every two-pass function reads only base columns through the record and reads/writes its own output via `spi.getAddress(rowIndex,…)` (position-independent). You already made this realization for `pass1` (`:470`, `:485`); `pass2` was missed. Correct as-is, just wasted work over millions of rows.
- **`RecordArray.beginRecord` double `jumpTo` per row** (`:89-90`) — for the LIGHT chain (`recordSink==null`, written only via random `getAddressAtRowIndex` *after* build), the rewind `jumpTo` is dead and the per-row extend is hoistable to one `jumpTo(size*fixOffset)` after the build loop. (Both jumpTos are justified for the non-light `put()` path.)
- **Member ordering** — `SortKeyType.java:33-34`: `MAX_HEAP_BYTES` ('H') is before `MAX_ENTRY_LONGS` ('E'); both `public static final`, mutually independent — trivial swap.
- **Coverage gaps** (CI green, paths unverified): (a) `testCachedWindowLightWindowLightRecordReadsExoticColumnTypes` (~line 839) asserts only the EXPLAIN plan + `count()` — it never reads GEOHASH/LONG256/LONG128/BIN/ARRAY/INTERVAL **values** through `WindowLightRecord` (matches the 53% coverage report); (b) no value-level test exercises `TreeWindowSortBuffer` end-to-end (only cap-failure + empty-table plan); (c) the parallel `Vect.sortEncodedEntries` branch is only crossed via a lowered threshold over ≤1 000 rows — the real 1 024 000-row default path is uncovered; (d) FIXED_24/FIXED_32 composite-key **success** path not value-asserted (only the >32-byte fallback).
- **Out-of-scope churn in `PropServerConfiguration.java`** — ~70 lines of unrelated reordering (`validate*` helpers moved, `getSqlJitMode` moved, `getPostingIndex*` getters swapped). The resulting order is correct but it bloats the diff and invites merge conflicts on a hot file; consider dropping or splitting into a `chore` commit.
- **Test boolean naming** — `cacheLightWindowEnabled` (no `is`/`has` prefix) in 6 window test classes.
- **Diagnostics** — `EncodedWindowSortBuffer`'s overflow message names no actionable config key; the `assert maxEntries > 0 : "put() called before of()"` can also trip on a legitimately tiny configured cap with a misleading message (production still throws the correct `LimitOverflowException`).
- **Latent reopen fragility** — if `of()` were re-entered without an intervening `close()` (a cursor-contract violation), `reopenSortBuffers()` is skipped and the resurrected `entryMem` of an early-freed single-pass group leaks. Non-crashing; doesn't occur in normal flows; the non-light path lacks this.

## Downgraded (dismissed after source verification)

- "Encoded buffer stores dense rowIndex then re-walks the narrow chain to recover baseRowId" — FALSE: recovery is a single O(1) `baseRowIds.get(rIdx)` load; dense rowIndex is required anyway for narrow addressing + `spi.getAddress`.
- "`finishPut` frees/reopens the encoder per group" — FALSE: encoder freed only in `close()`; `finishPut` just sorts.
- Parallel-sort concurrency hazard — FALSE: `Vect.sortEncodedEntries` joins all threads before returning; per-group private buffers; the window cursor is unreachable from parallel-reduce/top-K/AsyncWindowJoin (`supportsPageFrameCursor()`/`canPeelForTopK()` both `false`).
- Encoded-sort tie-break differs from tree — FALSE: `ooo.cpp` `encoded_Nx::operator<` tie-breaks equal keys by ascending `rowId`, matching tree insertion order (ASC & DESC); the duplicate-timestamp DESC regression test pins it.
- `lead(ts,0)` breaks a persisted MatView schema — FALSE: MatViews reject window functions at parse time.
- prior review #1 (cursor-ctor leak window), #3 (light tree-sort regression), #4/#5 (rankMaps leaks), #9 (narrow-chain page-size config) — all VERIFIED FIXED on the current head.

## Summary

**Verdict: request changes**, gated on the failing Enterprise CI (C1) — the only hard blocker; correctness and architecture are otherwise sound. The base-column elision + flat encoded sort is a real algorithmic win and the ~17× `LAG OVER ts DESC` claim is plausible. Two items I'd most want a response on alongside CI: **M1** (the `window.tree.max.bytes` bound silently no longer applies to the encoded-key case — the exact OOM scenario this PR targets) and **M3** (the missing fixed-width guard). All retained findings are in-diff; the cross-context pass walked the real out-of-diff callers (`MarkoutHorizonRecordCursorFactory`, the non-light encoded path, `lead` callers) and confirmed each SAFE — the contract changes are additive.

<sub>Generated by an independent level-3 review (10 specialist agents + per-finding verification). 13 findings verified, 6 draft findings dropped as false positives / confirmed-fixed.</sub>

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 598 / 673 (88.86%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/window/WindowLightRecord.java|51|96|53.12%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyEncoder.java|12|15|80.00%|
|:large_blue_circle:|io/questdb/cairo/RecordArray.java|24|30|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/window/TreeWindowSortBuffer.java|31|35|88.57%|
|:large_blue_circle:|io/questdb/griffin/engine/window/EncodedWindowSortBuffer.java|68|72|94.44%|
|:large_blue_circle:|io/questdb/griffin/engine/window/CachedWindowLightRecordCursorFactory.java|282|293|96.25%|
|:large_blue_circle:|io/questdb/griffin/engine/window/CachedWindowRecordCursorFactory.java|74|76|97.37%|
|:large_blue_circle:|io/questdb/griffin/engine/window/LightWindowSPI.java|18|18|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|24|24|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyType.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LongTreeChain.java|5|5|100.00%|

kafka1991:
Independent **level-3 review** (parallel specialist agents for correctness/resources + native-comparator verification + inline checks of concurrency, tests, performance, metadata) against PR head `5fb62c8814`. The earlier fixes all landed and verified. **no blocking issues remain** — the core engineering is sound.

## Critical (blocks merge)

None.

## Moderate

### M2. `lead(ts,0)` / `lead(m,0)` LONG -> TIMESTAMP/DATE is a breaking change absent from the release-note-bound title
`LeadTimestampFunctionFactory.java:74`, `LeadDateFunctionFactory.java:73`.

The change is correct and intentional (makes `lead(ts,0)` agree with `lag(ts,0)`, which already returns TIMESTAMP via `LagTimestampFunctionFactory.LeadLagValueCurrentRow`), is documented in the body, and carries the `Compatibility` label. But the title description is what reaches release notes on squash-merge. An upgrading user wouldn't learn from release notes that `lead(ts,0)` now returns TIMESTAMP / `lead(m,0)` returns DATE, that `CREATE TABLE AS SELECT lead(ts,0) OVER (...)` now persists a TIMESTAMP column instead of LONG, or that PG-wire/BI clients keying off the old LONG type OID see a different type. No MatView break (window functions are rejected at MatView parse time - verified). Recommend a release-note-reaching breaking-change line, or splitting the `lead` change into its own titled commit.

## Minor

- **Out-of-scope churn in `PropServerConfiguration.java`** - ~70 lines of pure member-reordering on a hot, conflict-prone file (`validate*` helpers relocated, `getSqlJitMode` moved, two posting-index getters swapped). The order is alphabetically correct but unrelated to the window/sort work; consider dropping or splitting into a `chore` commit.
- **Encoded buffer pre-sizes to `expectedRows` up front** (`EncodedWindowSortBuffer.of():136-141`) - `min(expectedRows, maxEntries) * longsPerEntry` allocated in one shot. A win for completing queries (avoids realloc copies, equals real footprint), but on a multi-billion-row base a query cancelled immediately commits the full (tens of GB) allocation before the first per-row circuit-breaker check. Documented tradeoff, not a bug.
- **Test coverage gaps** (CI green, paths thin):
  - `testCachedWindowLightWindowLightRecordReadsExoticColumnTypes` asserts only the plan + `count()`; it exercises the GEO*/LONG256/LONG128/BIN/ARRAY/INTERVAL read-*dispatch* in `WindowLightRecord` but never equality-checks the returned *values* (matches the 57% line coverage). Add a `.returns(...)` over those columns.
  - The parallel branch of `Vect.sortEncodedEntries` (`cairo.sql.sort.encoded.parallel.threshold`, default `1_024_000`) is only crossed via a lowered threshold over small datasets, never at the real default.
  - `TreeWindowSortBuffer`'s value-level path is exercised by cap-failure/plan tests, not value-asserted end-to-end.
- **`circuitBreaker.statefulThrowExceptionIfTripped()` called per-row** in the ordered-group inner loops (`CachedWindowLightRecordCursorFactory.java:468`, `:519`). Once-per-row is more than necessary, but it mirrors the non-light factory (pre-existing), so not a new regression.

## Verified correct / confirmed-fixed (not findings)

**Verified correct this review:**
- **Light vs non-light equivalence** - `WindowLightRecord` sign-encoded dispatch uniform across all ~30 accessors (the `+1` narrow offset removes the column-0 ambiguity); `LightWindowSPI.getRecordAt` shares `getRecordB()` while the ordered loops position `recordA` via `getRecord()` -> no aliasing, and `getRecordAt` is only live during `computeWindow`.
- **Encoded-sort tie-break matches the tree** - checked against `ooo.cpp` `encoded_2x/3x/4x/5x::operator<` (lines 1259-1309): equal keys break by ascending unique `rowId` -> total deterministic order reproducing the tree's insertion-order tie-break for ASC and DESC (incl. the duplicate-timestamp DESC regression test).
- **Native-memory lifecycle** - both factory ctors null the locals after cursor construction and free-once on every throw path; ctor-catches and `close()` are mutually exclusive (no double-free); the `computeWindow()` early-free of single-pass groups is idempotent and the "reopen fragility" is unreachable under the getCursor/close contract (self-heals via `realloc(0,...)` even if violated); `RecordArray.auxMem` is freed via the virtual `clear()` dispatched from `RecordChain.close()`.
- **`RecordArray` all-fixed-width path** - `beginRecord()`'s double `jumpTo` extends pages to the row end before rewinding, so random writes via `getAddressAtRowIndex` land on allocated memory; `rowToOffset = rowIndex * fixOffset` when `auxMem == null` is consistent with the append sequence.
- **Concurrency** - no window factory overrides `supportsPageFrameCursor` (default false); cursor is single-threaded; `Vect.sortEncodedEntries` joins all workers before returning, per-group private buffers, parallel phase touches only pre-encoded bytes.
- **Tests** - migrated to the builder API: 23 new `assertQuery(...)` with `.assertsPlan(...)`/`.withPlan(...)`, zero banned `assertSql`/`returnsOnce`/`assertPlanNoLeakCheck`/`getPlan`; the `assertMemoryLeak` wrappers hold DDL+inserts+assertion (legitimate); boolean is `isCacheLightWindowEnabled`; `## Test plan` present.

**Confirmed fixed since prior reviews:** M1 (encoded buffer now honors `getSqlWindowTreeKeyMaxBytes()` + `getSqlWindowRowIdMaxBytes()`, with `maxHeapBytes` clamping each operand before summing to avoid `Long.MAX_VALUE` overflow); M3 (`isAllWindowOutputFixedWidth` added to the light gate so a future var-width window output routes to the non-light factory); prior #1 (cursor-ctor leak window), #3 (light tree-sort regression - gated out by `isAllGroupsEncodedEligible`), #4/#5 (rankMaps leaks), #6-#8 (forward/backward partition + base-only positioning), #9 (narrow-chain page-size now uses `getSqlWindowStorePageSize`), and the pass2 `positionRecordABaseOnly` perf nit (lines 524, 537).

## Summary

**Verdict: approve.** With the stale-branch CI failure excluded, the engineering is verified sound - base-column elision + flat encoded sort is a real algorithmic win, the encoded-sort tie-break provably matches the old tree (checked against `ooo.cpp`), the native-memory lifecycle is leak/double-free-free on every traced path, and light/non-light equivalence holds. Only **M2** (a release-note disclosure for the `lead` compatibility change) is worth addressing before merge; the four Minors are optional follow-ups. All retained findings are in-diff; the cross-context pass confirmed the out-of-diff callers (`MarkoutHorizonRecordCursorFactory`, the non-light encoded path, `lead` callers, MatView parse-time rejection) are SAFE - the contract changes are additive (nullable `recordSink`, nullable `auxMem`, an extra `LongTreeChain.put` overload).

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 598 / 673 (88.86%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/window/WindowLightRecord.java|51|96|53.12%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyEncoder.java|12|15|80.00%|
|:large_blue_circle:|io/questdb/cairo/RecordArray.java|24|30|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/window/TreeWindowSortBuffer.java|31|35|88.57%|
|:large_blue_circle:|io/questdb/griffin/engine/window/EncodedWindowSortBuffer.java|68|72|94.44%|
|:large_blue_circle:|io/questdb/griffin/engine/window/CachedWindowLightRecordCursorFactory.java|282|293|96.25%|
|:large_blue_circle:|io/questdb/griffin/engine/window/CachedWindowRecordCursorFactory.java|74|76|97.37%|
|:large_blue_circle:|io/questdb/griffin/engine/window/LightWindowSPI.java|18|18|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|24|24|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyType.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LongTreeChain.java|5|5|100.00%|

bluestreak01 (approved):

PR #7157, opened 2026-05-26
feat(sql): add byte caps for CachedWindow and ORDER BY native memory

puzpuzpuz:
## Summary

- Adds six byte-denominated per-operator caps for native memory growth in the CachedWindow record store, the `LongTreeChain` key and value heaps (used by both window functions and ORDER BY), and the ORDER BY sort heaps. The caps are unset (uncapped) by default; setting any of them bounds the operator and raises `LimitOverflowException` with a `(raise X)` hint naming the key that needs to change:
```
cairo.sql.window.cache.max.bytes
cairo.sql.window.rowid.max.bytes
cairo.sql.window.tree.max.bytes
cairo.sql.sort.key.max.bytes
cairo.sql.sort.light.value.max.bytes
cairo.sql.sort.value.max.bytes
```

- Marks `cairo.sql.window.tree.max.pages`, `cairo.sql.window.rowid.max.pages`, `cairo.sql.sort.key.max.pages`, `cairo.sql.sort.light.value.max.pages`, and `cairo.sql.sort.value.max.pages` deprecated. They continue to be parsed: if a user has one set, the derived byte default for the corresponding new key becomes `pageSize * maxPages`. An explicit new `*.max.bytes` value wins when both are set.

- For `cairo.sql.window.cache.max.bytes`, precedence is resolved against the legacy `cairo.sql.window.store.max.pages` (and its `cairo.sql.analytic.store.max.pages` alias), which still drive the per-partition `MemoryCARW` window function buffers. The new bytes key wins when set explicitly; the legacy pages key wins when only it is explicit; otherwise the new bytes default wins. The resolved key path is plumbed into the runtime so the error message names the binding constraint instead of sending the user down the wrong knob. One nuance worth flagging: when only the legacy `cairo.sql.analytic.store.max.pages` alias is explicit, the runtime hint names `cairo.sql.window.store.max.pages` (the modern key that supersedes it) rather than the alias the user has in their config — raising the modern key still takes effect via the same resolution path.

- When sort-key materialization is engaged, `cairo.sql.sort.key.max.bytes` is split across the materialized column buffers in proportion to each column's fixed-size width (via `ColumnType.sizeOf`). A wider column (e.g. `DECIMAL256`, 32 B per row) receives a proportionally larger share than a narrow one (e.g. `BOOLEAN`, 1 B per row), so each buffer's row capacity stays roughly balanced and the total reachable memory tracks the operator-wide budget rather than the wide column's footprint times the column count. Each buffer is still floored at one page so the cursor can initialise; the only overshoot is for sub-page budgets, where the floor exceeds the requested cap by at most `bufferCount * PAGE_SIZE`.

- Page-size config keys (`cairo.sql.sort.{key,light.value,value}.page.size` and the three `cairo.sql.window.*.page.size` keys) clamp to a minimum of 1 byte at read time, so a misconfigured `*.page.size=0` fails over to a tiny page rather than propagating a 0 into downstream divisions and surfacing as an `ArithmeticException` from the factory constructor.

- Tradeoff: with the caps unset by default, this PR provides opt-in protection rather than preventing OOM out of the box. Operators that previously survived a `lag(...)` over a multi-billion-row table on unbounded growth keep that behavior; deployments that want a guard must set one of the new keys. The previous iteration of this PR defaulted each cap to 4 GiB, which would have been a breaking change for any operator already exceeding that ceiling in production; this version is not.

- Without any cap set, the only ceilings remain the internal compressed-offset limits in `AbstractRedBlackTree` (~32 GiB key heap) and `LongTreeChain` (~16 GiB value heap), plus the JVM's native memory budget.

## Test plan

- `CachedWindowMemoryCapTest` exercises each of the cache, rowid, and tree caps firing, the happy path with uncapped defaults, the same workload succeeding after raising the cap, and per-cursor reset across repeated executions.
- `OrderBySortKeyMaterializationTest` covers the operator-wide ceiling for single-column and equal-width multi-column sort-key materialization, plus the size-weighted split via a mixed `BYTE + LONG` schema where the even split would fire and the weighted split succeeds.
- `PropServerConfigurationTest` covers deprecation precedence (including the case where both the modern deprecated key and its analytic alias are set with conflicting values), the resolved CachedWindow cap key, explicit-`Integer.MAX_VALUE`-on-the-deprecated-key behavior, and the page-size clamp under misconfigured `*.page.size=0`.
- `ServerMainTest` "show parameters" snapshot updated to include the six new keys at `Long.MAX_VALUE`.
- `SampleByFillTest.testSortedRecordCursorFactoryHandlesKeyHeapOverflow` switches from the `max.pages=-1` cleanup probe to a 64-byte `sort.key.max.bytes` budget that triggers the same `LimitOverflowException` cleanup path, and asserts the `(raise cairo.sql.sort.key.max.bytes)` hint so a rename to the property key would fail the test instead of silently breaking the user-facing remediation guidance.


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 183 / 198 (92.42%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LimitedSizeLongTreeChain.java|3|6|50.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LimitedSizeSortedLightRecordCursorFactory.java|4|8|50.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortRecordCursor.java|15|20|75.00%|
|:large_blue_circle:|io/questdb/std/MemoryPages.java|7|9|77.78%|
|:large_blue_circle:|io/questdb/PropServerConfiguration.java|45|46|97.83%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortLightRecordCursor.java|17|17|100.00%|
|:large_blue_circle:|io/questdb/cairo/vm/Vm.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortedLightRecordCursorFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/RecordChain.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/AbstractRedBlackTree.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/DefaultCairoConfiguration.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/PropertyKey.java|13|13|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKAtom.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyMaterializingRecordCursor.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/vm/MemoryCARWImpl.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoConfigurationWrapper.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LongTreeChain.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyMaterializingRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/RecordArray.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortedRecordCursorFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/RecordTreeChain.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/window/CachedWindowRecordCursorFactory.java|6|6|100.00%|

puzpuzpuz:
Confirmed fresh-28 independently: RecordArray line 59 allocates auxMem with page pageSize >> 4 (= 4 bytes at the new 64-byte store minimum), and beginRecord() line 64 does
  auxMem.putLong(...) (8 bytes) — so with a cache cap resolving to 1 page the aux side throws before storing any record, and the justification comment at line 153-154 ("store >> 4, so it
  needs >= 64") is arithmetically wrong (64>>4 = 4 < 8).

  I now have everything needed. Here is the consolidated, source-verified review.

  ---
  Review: PR #7157 — feat(sql): add byte caps for CachedWindow and ORDER BY native memory (level 3)
  
  29 raw findings → 26 verified, 3 dropped as false positives. Split: all confirmed code findings are in-diff; the only out-of-diff items are PR metadata/description. The cross-context sweep
  of all ~30 unmigrated window-function callsites of getSqlWindowStoreMaxPages/getSqlWindowStorePageSize came back clean (no compile break, no overflow, no silently-dead overrides). No
  resource-leak finding survived verification, and the adversarial-performance pass found nothing — the arithmetic is allocation-free and overflow-safe at Long.MAX_VALUE.

  Critical

  None. No confirmed correctness, safety, concurrency, resource-leak, or UB defect. Results stay correct on every path; the overflow clamps prevent negative/overflowed budgets; per-worker
  tree chains are single-thread-owned under PerWorkerLocks then merged behind the dispatchAndAwait barrier.

  Moderate

  M1 — On the default ORDER BY path, sort.key.max.bytes alone does not bound memory, and the error message misdirects

  (merges correctness-0, concurrency-1, resource-7, tests-10 — in-diff)

  EncodedSortLightRecordCursor.java:75-77 (the default ORDER BY path, selected at SqlCodeGenerator.java:7449 whenever the sort key is fixed-width-encodable):

  keyCap   = min(getSqlSortKeyMaxBytes(),   LIMIT);   // LIMIT = (2^32-2)<<3 ≈ 32 GiB
  valueCap = min(getSqlSortLightValueMaxBytes(), LIMIT);
  maxEntryMemBytes = min(keyCap + valueCap, LIMIT);

  Both caps default to Long.MAX_VALUE, so leaving one unset saturates its operand to LIMIT; the sum is then always ≥ LIMIT and the configured cap is fully masked. A user who sets only
  cairo.sql.sort.key.max.bytes to bound ORDER BY gets no effective bound on this path (both keys must be lowered). throwLimitOverflow() (EncodedSortLightRecordCursor.java:229-237) compounds
  this by saying "raise X or Y", implying either knob is the binding constraint. This is not a regression (master summed page budgets the same way) and not a correctness bug — but for a PR
  whose stated purpose is opt-in memory bounding, the headline knob is ineffective in isolation on the most common path, and the remediation guidance is wrong. The sibling paths behave
  correctly: EncodedSortRecordCursor (key cap drives entryMem, value cap a separate RecordChain) and LongTreeChain (two independent region caps) each bind on a single knob. No test exercises
  a single-knob cap on the EncodedSortLight path.

  Fix: either make each cap bind independently on this path (treat an unset cap as 0-contribution when the other is explicit), or document the combined-budget semantics in server.conf +
  reword the error to "both X and Y bound this operator", and add an OrderByMemoryCapTest case that engages the encoded-light path with only the key cap set.

  M2 — The new (raise ...) remediation hint on the EncodedSort paths is asserted by no test

  (tests-9 — in-diff)

  throwLimitOverflow() in both EncodedSortLightRecordCursor.java:229-237 and EncodedSortRecordCursor.java:243-249 appends the new (raise cairo.sql.sort.key.max.bytes[ or ...]) hint, but the
  only tests hitting these paths (SecurityTest.java:560,809,1092,1156) assert only the pre-existing substring "memory exceeded in EncodedSort". The new user-facing hint is unverified and a
  property rename would silently break it — exactly the failure mode the PR deliberately guards against elsewhere (SampleByFillTest, OrderByMemoryCapTest, CachedWindowMemoryCapTest all pin
  their (raise ...) text). Fix: tighten one SecurityTest assertion (or add a dedicated test) to assert the full hint for both encoded paths.

  M3 — PR description misstates page-size handling (says "clamp to 1" where the code rejects at startup)

  (metadata-17 — PR body)

  The body bullet claims all six page-size keys "clamp to a minimum of 1 byte at read time, so a misconfigured *.page.size=0 fails over to a tiny page." In reality, only
  cairo.sql.sort.value.page.size clamps (Math.max(1, …)); the other five (sort.key≥64, sort.light.value≥12, window.store≥64, window.rowid≥16, window.tree≥32) are rejected at startup via
  validatePageSizeAtLeast (PropServerConfiguration.java:2711) — the server won't boot. The PR's own server.conf text ("rejected at startup if smaller") and testPageSizeBelowBlockSizeRejected
  contradict the body. Fix: rewrite the bullet to state heap-backed keys are rejected (naming the key) and only sort.value.page.size clamps; fix the matching test-plan line.

  Minor

  - m1 — RecordArray aux index page can't hold one 8-byte slot at the new store minimum (fresh-28, in-diff). RecordArray.java:59 allocates auxMem with page pageSize >> 4; at the new
  MIN_WINDOW_STORE_PAGE_SIZE=64 that's 4 bytes, but beginRecord() writes an 8-byte putLong. With window.store.page.size in [64,127] and a cache cap resolving to ~1 page, the aux side throws
  "breached in VirtualMemory" before any record is stored (and the message names the cap key, not the page-size key). The justification comment at PropServerConfiguration.java:153-154
  ("store >> 4, so it needs >= 64") is arithmetically wrong — 64>>4=4<8. **Fix:** raise MIN_WINDOW_STORE_PAGE_SIZE to 128, or allocate auxMem with Math.max(pageSize >> 4, 8); correct the
  comment.
  - m2 — Migrated sort.key.max.pages gives materialization buffers a ~16× looser ceiling (perf-4 + resource-8 + crosscontext-23, in-diff). The derived byte cap is sort.key.page.size(128k) × 
  maxPages, but the materializer's internal unit is PAGE_SIZE=8192, so per-buffer caps scale by 128k/8k = 16× vs pre-PR for a single column (looser for <16 materialized columns, tighter for
  >16). Intentional/documented operator-wide semantics and only affects deployments that set the deprecated key; a relaxation that can't break a passing query. Fix: add an explicit migration
  note ("raise sort.key.max.bytes if you relied on the old per-buffer ceiling").
  - m3 — DefaultCairoConfiguration sort caps changed from finite (1024 pages) to uncapped (crosscontext-22 + fresh-24, in-diff). Removes the implicit ~128 MB–16 GB test-time safety ceiling
  for every AbstractCairoTest. Production (PropServerConfiguration, already Integer.MAX_VALUE) is unchanged and no current test depends on the old default. Risk is only a future runaway-sort
  test silently OOMing instead of surfacing a clean LimitOverflowException. Fix (optional): return 1024L * getSqlSortKeyPageSize() to preserve the finite test default, or note the
  intentional removal.
  - m4 — Parallel ORDER BY ... LIMIT cap is per-worker; aggregate is (workers+1) × cap, undocumented (concurrency-2, in-diff). AsyncTopKAtom.java:106-132 gives the owner chain and every
  per-worker chain the full byte budget. Concurrency-safe; just a documentation gap in server.conf.
  - m5 — warnIfMaxBytesBelowPageSize compares raw page size; consumers floor by ceilPow2(pageSize) (fresh-26, in-diff). For non-power-of-two sort.value.page.size (and sort.key via
  RecordTreeChain.derivePageBudget), the advisory can disagree with the real flooring. Advisory-log accuracy only; default page sizes are powers of two. Note the suggested fix's mention of 
  window.store is a no-op (those are already ceilPow2'd at config time) and must not be applied to sort.light.value (consumed raw).
  - m6 — New boolean locals miss the is/has prefix (quality-14, in-diff). PropServerConfiguration.java:1790-1792 cacheMaxBytesExplicit/storeMaxPagesExplicit violate the CLAUDE.md naming rule
  (the sibling hasAlias in the same PR follows it). Rename to isCacheMaxBytesExplicit/isStoreMaxPagesExplicit.
  - m7 — Resolved cache-cap error names the canonical key even when only the analytic.* alias was set (quality-15, in-diff). PropServerConfiguration.java:1795 hard-codes
  window.store.max.pages though storeMaxPagesExplicit is true for either key. Cosmetic (keys are deprecated-equivalent; raising the named key works).
  - m8 — Sub-minimum legacy alias page.size is rejected naming the new key, not the alias the user set (quality-16, in-diff). e.g. analytic.tree.page.size=16 is rejected as
  cairo.sql.window.tree.page.size. Cosmetic; the value is invalid regardless.
  - m9 — EncodedSortRecordCursor (full_encoded) value-chain byte cap is untested (tests-11, in-diff). The only sort.value overflow test uses full_recordchain → RecordTreeChain. The byte cap,
  page-rounding, message, and config-key hint are shared with the tested path, so the gap is the class wiring only.
  - m10 — testParallelTopKPerWorkerCapFires can't distinguish per-worker from owner-chain overflow (tests-12, in-diff). Both chains throw the identical message, so the assertion can't prove
  the per-worker path fired as the comment claims. Soften the comment or add a per-worker sentinel.
  - m11 — Negative legacy-cap floor is asserted only at the config layer (tests-13, in-diff). The old end-to-end max.pages=-1 probe was rewritten to a positive value; no query-level test now
  drives a negative cap through AbstractRedBlackTree/SortKeyMaterializingRecordCursor. The floor logic is verified correct by inspection; the gap is coverage only.
  - m12 — Performance label not justified (metadata-18, out-of-diff). feat(sql) memory-guard PR with no benchmark/perf claim; the only benchmark touched is cleanup from the API rename.
  Remove the label.
  - m13 — LIMIT 40000 missing thousands underscore (metadata-19, in-diff). OrderByMemoryCapTest.java:59,93 — use 40_000 to match the rest of the file and CLAUDE.md.
  - m14 — AMI server.conf omits # Deprecated: notes on analytic.{rowid,tree}.max.pages (metadata-20, in-diff). The main server.conf carries them; the AMI copy doesn't (though it did add them
  for the sort keys). analytic.store.max.pages is correctly left undeprecated in both.

  Downgraded (false positives)

  - fresh-25 — divide-by-zero in DefaultCairoConfiguration.getSqlWindowCacheMaxPagesResolved(). Dropped: no subclass anywhere overrides getSqlWindowStorePageSize() to 0; the cited
  WindowFunctionTest =0 value flows through PropServerConfiguration, which ceilPow2s and rejects it (MIN=64), never reaching the divisor.
  - fresh-27 — window.cache.max.bytes doesn't bound the per-partition window buffers. Dropped: factually true but the in-diff server.conf already scopes the key to "the CachedWindow
  materialised record store" and explicitly ties per-partition buffers to window.store.max.pages. The finding's own suggested fix is already implemented; no contradiction with stated intent.
  - crosscontext-21 — "all callsites SAFE". Not a defect — a clean cross-context confirmation (verified: zero references to the 5 removed int getters, mvn -pl core test-compile green,
  RankFunctionFactory pageSize*maxPages/2 casts to long first so no overflow, SecurityTest translated 1:1).

  Summary

  Verdict: approve with changes recommended. This is a clean, well-tested PR with no blocking issues — no critical findings, no surviving resource-leak or correctness defects, and a
  verified-safe blast radius across the ~30 unmigrated window-function callsites. The arithmetic (overflow-safe divide-then-multiply, clamps before sums, negative-cap flooring) is sound.

  Address before merge: M1 (single-knob sort.key.max.bytes is a no-op on the default ORDER BY path + misleading error message — the most important, since it undercuts the feature's headline
  use case) and M3 (PR description contradicts the shipped reject-at-startup behavior). M2 wants a migration callout. M2/M3/M4/m3 are the tradeoffs worth stating plainly in the description:
  the cap is per-worker for parallel top-K, migrated max.pages shifts the materialization ceiling, and the test-time default is now uncapped. Everything else is minor (one concrete
  edge-config wart in RecordArray/MIN_WINDOW_STORE_PAGE_SIZE, plus naming/test/doc polish).

  Counts: 29 draft findings → 26 verified, 3 false positives removed. In-diff/out-of-diff: all confirmed code findings in-diff; 1 finding (label) out-of-diff PR metadata, 1 (description) in
  the PR body. The non-trivial-but-zero-code-bug out-of-diff result reflects a genuinely clean caller sweep, not an underrun — the cross-context agent enumerated and cleared every unmigrated
  callsite with grep evidence and a successful compile.

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 224 / 235 (95.32%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LimitedSizeSortedLightRecordCursorFactory.java|4|8|50.00%|
|:large_blue_circle:|io/questdb/std/MemoryPages.java|7|9|77.78%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortLightRecordCursor.java|10|12|83.33%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortRecordCursor.java|15|17|88.24%|
|:large_blue_circle:|io/questdb/PropServerConfiguration.java|70|71|98.59%|
|:large_blue_circle:|io/questdb/cairo/vm/Vm.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortedLightRecordCursorFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/RecordChain.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/AbstractRedBlackTree.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/cairo/DefaultCairoConfiguration.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/PropertyKey.java|13|13|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKAtom.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyMaterializingRecordCursor.java|16|16|100.00%|
|:large_blue_circle:|io/questdb/cairo/vm/MemoryCARWImpl.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoConfigurationWrapper.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LimitedSizeLongTreeChain.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LongTreeChain.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyMaterializingRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/RecordArray.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortedRecordCursorFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/RecordTreeChain.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/window/CachedWindowRecordCursorFactory.java|6|6|100.00%|

PR #7162, opened 2026-05-26
feat(core): breaking change 💥 - preserve and backfill materialized view data that is beyond REFRESH LIMIT

nwoolmer:
## Breaking change

`FULL` refresh will preserve rows beyond REFRESH LIMIT, fully clearing the view requires a re-create.

## Summary
- Materialized views with `REFRESH LIMIT` now treat rows older than the boundary as a frozen zone: refresh no longer touches them and users can backfill historical data directly via INSERT / COPY / ILP / QWP.
- Plain `REFRESH MATERIALIZED VIEW <name> FULL` preserves the frozen zone instead of wiping the whole view. Views with no `REFRESH LIMIT` keep the previous wipe-and-reinsert behaviour.
- New `materialized_views().backfill_max_ts` column surfaces the current cutoff so users can see what timestamps are safe to backfill.

## Why
Before this change the `REFRESH LIMIT` boundary used `wallClock - LIMIT`, and `FULL` refresh ran an unconditional `truncateSoft()` before re-inserting only the within-limit data. Anything older than the limit was silently lost on every FULL refresh, which made `REFRESH LIMIT` impractical for views that wanted to keep aggregate history past the base table's TTL.

## Behaviour changes
- Boundary formula changes to `min(max(base_ts), wallClock) - LIMIT`, mirroring the TTL formula in `TableUtils.getMaxTimestamp`. A stale-ingestion gap no longer pushes the boundary forward.
- `INSERT` / `COPY` / ILP HTTP/TCP / QWP into a mat view succeed when `REFRESH LIMIT` is set; rejected otherwise (existing "cannot modify materialized view" error).
- Backfill rows are validated at WAL commit: a whole transaction is rejected if its max row's containing sample-by bucket extends past the boundary's bucket floor. Refresh-job writes are exempt.
- Plain `FULL` skips `truncateSoft()` when `REFRESH LIMIT` is set; per-interval `REPLACE_RANGE` commits inside `insertAsSelect` already rewrite managed-zone buckets atomically. Users who want a full wipe should `DROP MATERIALIZED VIEW; CREATE MATERIALIZED VIEW ...;`.
- `ALTER MATERIALIZED VIEW ... SET REFRESH LIMIT <new>` logs an advisory when `<new>` shrinks or removes the existing limit; previously-managed buckets become frozen and stop being refreshed, which can surprise users expecting continued recomputation.
- New `materialized_views().backfill_max_ts` column (TIMESTAMP_MICRO) surfaces the boundary's bucket floor. Rows with `ts < backfill_max_ts` are safe to backfill; NULL when `REFRESH LIMIT` is unset.
- New `cairo.mat.view.refresh.limit.wall.clock.enabled` config (default false). When true, the entire frozen-zone feature reverts: backfill is rejected at the entry point, plain `FULL` truncates as before, and the boundary returns to `wallClock - LIMIT`.

## Tradeoffs
- A managed-zone bucket whose base data has gone missing (manual `DROP PARTITION` on base, aggressive TTL) is no longer wiped by plain `FULL` since the iterator never emits it as a refresh interval. Stale-bucket recovery requires `DROP MATERIALIZED VIEW; CREATE ...;`.
- Computing `backfill_max_ts` opens one base table reader per row in the `materialized_views()` cursor. Acceptable for interactive metadata queries; tools that poll this rapidly pay the cost per poll.
- The bucket-whole validator allocates a small `TimestampSampler` on its first commit per writer; cached for the writer's lifetime thereafter.

## Test plan
- New `MatViewTest` scenarios: boundary formula (4 cases), backfill accept/reject paths, FULL preserves frozen zone, escape-hatch full revert, bucket-whole rejection, `materialized_views().backfill_max_ts` snapshot.
- Updated `MatViewStateTest` expected schema for the new column.
- Backward-clock regressions verify that accepted backfill and previously frozen orphan buckets survive FULL refresh.
- LIMIT extend/shrink regressions verify that intentional boundary movement still works.
- Checkpoint regressions verify restoration of the never-refreshed bootstrap frontier and the persisted frozen floor/LIMIT pair.
- Focused materialized-view/checkpoint tests and the parameterized `MatViewStateTest` suite pass locally.
- Enterprise companion PR (questdb/questdb-enterprise#1028) adds the `INSERT INTO MATERIALIZED VIEW` permission and 3 replication scenarios (backfill, FULL-with-base-partition-drop, primary->replica failover, backfill replicates).

## Summary by CodeRabbit

* **New Features**
  * Materialized views can now accept backfill writes when a refresh limit is configured, while still blocking unsupported cases.
  * Added a new view metric showing the backfill cutoff timestamp in `materialized_views()`.

* **Bug Fixes**
  * Improved refresh behavior so backfill and refresh boundaries stay consistent across restarts and WAL replay.
  * Fixed full refresh handling to better preserve frozen-zone data when appropriate.
  * Updated ingestion paths to recognize the new backfill rules for HTTP, TCP, SQL, text, and QWP inputs.


nwoolmer:
## Second-pass review of `e5cdfb5e54` ("Address frozen-zone review: coordinate validator with refresh")

Level 2 review (no tandem). Scope: the review-fix commit and master merge `3f6e1a5ae0`. The original PR was covered in the first review.

### Critical

#### C1' — Validator's `min(own, published)` clamp does not account for `ALTER … SET REFRESH LIMIT` between refresh publish and user commit; user backfill silently overwritten on SHRINK

- `core/src/main/java/io/questdb/cairo/mv/MatViewBackfillValidator.java:257-284`
- `core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java:592-625`

The fix publishes the raw `boundaryAnchor` and re-derives `rawBoundary = anchor − LIMIT_now` at validator time. Refresh's actual REPLACE_RANGE.lo was locked in at publish time under `LIMIT_at_publish`. If `ALTER … SET REFRESH LIMIT` shrinks the limit between publish and the user commit, the validator's snapped floor sits *above* refresh's REPLACE_RANGE.lo, and the validator accepts buckets that fall inside the still-pending wipe.

Reproducer (1h sampling):
1. Refresh tick T0 under `LIMIT=2h`. Anchor = 12:00. `viewState.setLastRefreshFrozenBoundaryAnchor(12:00)`. Iterator's `minTs = 10:00`. Refresh does NOT yet emit its terminal REPLACE_RANGE commit.
2. `ALTER … SET REFRESH LIMIT 1h` applies; new LIMIT = 1h.
3. User backfill INSERT at ts=10:30. Validator: `publishedAnchor = 12:00`, `ownAnchor = 12:00`, `effective_anchor = 12:00`. `LIMIT_now = 1h`. `rawBoundary = 11:00`. `sampler.round(11:00) = 11:00`. Row's bucket `[10:00, 11:00)`, `bucketEnd = 11:00`. NOT `> 11:00`. **ACCEPTED.**
4. Refresh's terminal `commitMatView(replacementTimestampLo=10:00, …)` enqueues REPLACE_RANGE `[10:00, ∞)`. Apply wipes the user's row at 10:30.

EXTEND is safe (validator becomes more restrictive). Only SHRINK is unsafe — same direction as the original C1, just rearmed by `ALTER` timing.

**Fix:** publish the snapped REPLACE_RANGE floor refresh actually used (in driver units), not the raw anchor. Validator then does not need to re-apply `LIMIT_now`.

```java
// MatViewRefreshJob.findRefreshIntervals, after the minTs computation:
final long publishedFloor = viewDefinition.getTimestampSampler().round(minTs);
viewState.setLastRefreshFrozenBoundaryFloor(publishedFloor);
```

```java
// MatViewBackfillValidator.computeBoundaryBucketFloor:
final long ownFloor = sampler.round(ownRawBoundary);
final long publishedFloor = state != null ? state.getLastRefreshFrozenBoundaryFloor() : Numbers.LONG_NULL;
return publishedFloor != Numbers.LONG_NULL ? Math.min(ownFloor, publishedFloor) : ownFloor;
```

Walk-through with fix: at T2, `publishedFloor = 10:00`. `min(ownFloor=11:00, publishedFloor=10:00) = 10:00`. Row's bucket end 11:00 > 10:00 → REJECTED. Safe.

#### C2' — `materialized_views()` sampler cache miskeyed on `(interval, unit)` without timestamp driver; wrong `backfill_max_ts` for mixed-driver fleets

- `core/src/main/java/io/questdb/griffin/engine/functions/catalogue/MatViewsFunctionFactory.java:262-274`

Cache key omits the timestamp driver. `MatViewBackfillValidator.createSampler(def)` builds a driver-specific sampler. `SimpleTimestampSampler.round` divides by `bucket` in native driver units (3_600_000_000 micros vs 3_600_000_000_000 nanos for `1h`).

Reproducer: view A on `TIMESTAMP_MICRO` base with `SAMPLE BY 1h`. View B on `TIMESTAMP_NANO` base with `SAMPLE BY 1h`. Cursor iterates A first, builds MICROS sampler. On B, cache hit on `(1, 'h')`, reuses MICROS sampler. `sampler.round(boundaryAnchor /* NANOS */)` returns garbage. `backfill_max_ts` for view B is silently wrong (off by ~1000×).

**Fix:** include the driver in the cache key. Drivers are singletons; `==` works.

```java
private TimestampDriver cachedSamplerDriver;
…
if (cachedSampler == null
        || cachedSamplerDriver != viewDefinition.getBaseTableTimestampDriver()
        || cachedSamplerInterval != viewDefinition.getSamplingInterval()
        || cachedSamplerUnit != viewDefinition.getSamplingIntervalUnit()) {
    cachedSampler = MatViewBackfillValidator.createSampler(viewDefinition);
    cachedSamplerDriver = viewDefinition.getBaseTableTimestampDriver();
    …
}
```

### Moderate

**M1'** — `MatViewBackfillValidator.java:62-66` class-level Javadoc says "When `REFRESH LIMIT == 0` the validator rejects user DATA writes outright" but lines 202-204 return (pass through). Stale doc from the M4 reversion. Rewrite to mirror the in-method comment.

**M2'** — `MatViewState.java:648-650` `setLastRefreshFrozenBoundaryAnchor` does not assert `latch.get()`. Every other `set*` in the class does. Add `assert latch.get();` to match the file's convention and prevent a future non-locked write site from regressing the anchor.

**M3'** — `WalPreCommitValidator.java:27-37` Javadoc says implementations *MUST* throw `CairoException`, but `WalWriter.commit0:927-941` catches `Throwable` and distresses the writer on non-`CairoException`. Either tighten the signature to `throws CairoException` or replace the "MUST" wording with explicit treatment of non-`CairoException` as an internal validator fault.

**M4'** — `MatViewBackfillValidator.java:324` cache check `cachedMaxBaseTs != Long.MIN_VALUE` conflates "miss" with "empty base table" (`TxReader.getMaxTimestamp()` returns `Long.MIN_VALUE` on empty). Defeats the M2 optimisation precisely in the new-mat-view / dev / test case. Use a separate `boolean cachedValid` flag.

### Minor

- **Min1'** `MatViewBackfillValidator.java:76-83` field ordering — alphabetical demands `cachedBaseToken, cachedBaseTxn, cachedMaxBaseTs, cachedSampler`. Currently the sampler is first.
- **Min2'** `MatViewBackfillValidator.java:77-80` group-header comment above three fields. CLAUDE.md bans this — alphabetical auto-sort breaks the comment-to-field link.
- **Min3'** `MatViewBackfillValidator.java:300-304` and `:334-338` emit identical advisory chains. Extract a helper.
- **Min4'** commit title is 61 chars (CLAUDE.md asks ≤50).
- **Min5'** seven body lines in `e5cdfb5e54` exceed the 72-char wrap by 1-2 chars.

### Missing tests

- **P0** SHRINK-LIMIT-after-refresh case (matches C1'). `testMatViewBackfillRefreshDoesNotOverwriteAcceptedRow` holds LIMIT constant.
- **P0** Multi-driver `materialized_views()` sampler cache (matches C2'). One mat view on MICROS base, one on NANOS base, matching sampling; assert `backfill_max_ts` correct on both rows.
- **P1** Non-`CairoException` validator fault marks writer distressed. Install a test validator that throws `RuntimeException`; verify `walWriter.isDistressed()`.
- **P1** Base table drop+recreate invalidates validator's `cachedBaseToken`.

### Verdict

Request changes — two real P0 correctness bugs introduced by the review-fix commit, both rooted in the assumption that the published anchor is sufficient state for the validator to derive its boundary.

The highest-leverage change is publishing the snapped REPLACE_RANGE floor (instead of the raw anchor); that single change fixes C1' and removes a class of LIMIT-mismatch races. C2' is a one-line cache-key addition.

7 agents launched in this review; 13 in-diff findings verified; 4 false positives dropped. Out-of-diff was 0; if the C1' fix is non-trivial, re-running at level 3 (with the cross-context agent) before final approval is recommended.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 173 / 222 (77.93%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewBackfillValidator.java|82|117|70.09%|
|:large_blue_circle:|io/questdb/cairo/wal/WalWriter.java|10|13|76.92%|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|7|9|77.78%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|26|33|78.79%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/MatViewsFunctionFactory.java|17|18|94.44%|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/LineTcpMeasurementScheduler.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/LineHttpTudCache.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewState.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/pool/WalWriterPool.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cutlass/text/CairoTextWriter.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|13|13|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpTudCache.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/udp/LineUdpParserImpl.java|1|1|100.00%|

nwoolmer:
### Follow-up: closed the pre-first-refresh backfill window + hardened edges (964a5d7)

Addressing review findings on the frozen-zone backfill path:

**1. Pre-first-refresh data-loss window — `backfillFrontier`.**
A backfill accepted into a view that hasn't been refreshed yet can't be protected by `max(view_ts)` (the backfilled row *is* the view's max, and it's old). If `max(base_ts)` then retreats (base partition drop / re-ingestion) before the first refresh, the boundary would pull back over the backfill and a FULL refresh would wipe it.

Fix: the commit-time validator records, on accept, the boundary anchor it validated against (`MatViewState.advanceBackfillFrontier`, lock-free CAS max). The refresh job folds it into its boundary anchor → `max(max(base_ts), max(view_ts), backfillFrontier)`. This extends the existing `max(view_ts)` frontier idea to cover backfill accepted before the first refresh. Bootstrap backfill keeps working, and `min(own, published)` + overwrite-publish stay intact so SHRINK (existing shrink-race test) and EXTEND behave as before.

**2. `REFRESH LIMIT` overflow guard.** Replaced the `assert`-only guard (refresh + validator) with saturating arithmetic (`boundaryFromAnchor` → `Long.MIN_VALUE` = whole view frozen). Removes both the release-build silent wrap and the AssertionError path that would have permanently distressed a mat-view WAL writer on a user commit.

**3. `materialized_views()`** now memoizes `max(base_ts)` per `(base token, writer txn)` across cursor rows — one base-reader open per distinct base table per pass instead of per row.

**4.** Corrected a misleading `WalWriter` pre-commit-validator comment.

#### ⚠️ Known limitation regarding the view (worth a doc note)
The frozen boundary is **not persisted through the WAL**; it's re-derived from the view's own data. After a restart/failover, protection for a view that was refreshed before the restart is recovered from `max(view_ts)` (on-disk). The one residual edge: a view that is **backfilled before its first-ever refresh**, then **restarted**, then has a **base partition dropped — all before any refresh runs** — loses both `backfillFrontier` (in-memory) and has no `max(view_ts)` frontier yet, so that pre-refresh backfill could still be overwritten. This is extremely narrow (requires no refresh to have ever run). Full WAL-event persistence of the floor would close it but is a much larger format change; flagging for the docs / a possible follow-up.

Tests: new `MatViewTest` cases (`...FrontierSurvivesBaseRetreatBeforeFirstRefresh` — verified as a real guard via negative control — and `...ExtendRefreshLimitOverwritesBackfillInNewlyManagedRange`) plus a `boundaryFromAnchor` saturation unit test. `MatViewTest` (214), `MatViewBackfillValidatorTest`, `MatViewStateTest`, `ServerMainMatViewBackfillTest`, `QwpIngressProcessorStateTest` and the ILP HTTP gate test all pass.

nwoolmer:
## Level-3 review — frozen-zone backfill (5 adversarial passes)

I took the materialized-view frozen-zone backfill through five adversarial level-3 passes, each one re-targeting the **freshest** code — since on this change every fix tended to become the next pass's review surface. Every finding below was reproduced with a failing test first (RED → GREEN) and verified against the existing mat-view / WAL-apply suites.

### Data-loss bugs found and fixed

| # | Bug | Fix |
|---|-----|-----|
| 1 | Pre-first-refresh backfill wiped after a restart — the `backfillFrontier` was in-memory only, so a retreating `max(base_ts)` slid the boundary back over the row | persist the frontier to `_mv.s` (new block type) + reconstruct it on WAL replay |
| 2a | Crash between the data commit and the post-commit `_mv.s` write loses the frontier (under `commit.mode=sync/async`) — the applied txn is never re-reconstructed on restart | persist the frontier **before** the data commit (durable no later than the data it guards) — `5b37e75` |
| 2b | The pre-commit validator advanced the frontier (a no-rollback CAS-max) before the txn sealed, so a failed commit permanently over-pinned the boundary | advance only on the apply path; store the minimal `bucketEnd + LIMIT` — `5b37e75` |
| 3a | Months `REFRESH LIMIT`: `addMonths(bucketEnd, +L)` then `−L` clamps day-of-month, so the frontier round-trips **below** `bucketEnd` at a month-end bucket → wipe after a base retreat | day-bump the months anchor until its inverse no longer underflows — `2a2e83d` |
| 3b | Timezone-aligned views bucketed the validator / published floor / catalogue **tz-blind**, while the refresh `REPLACE_RANGE` buckets **tz-aware** → accept-then-wipe (e.g. `Asia/Kolkata` +5:30 with `SAMPLE BY 1h`) | one shared tz-aware `createBucketSampler` matching the refresh grid — `adbb91c` |
| 4 | Super-day stride + DST zone: the validator used the SAMPLE BY key sampler, which omits the DST shift-interval **collapse** the refresh's `TimeZoneIntervalIterator` performs → accept-then-wipe for ~17 IANA zones whose transition is at local midnight (Cairo, Havana, Santiago, Beirut, …) and any zone with `WITH OFFSET` | bucket the super-day-DST case through the refresh's own iterator — `4d7fdcd` |

### Confirmed by-design / not bugs (verified against the code's own tests + comments)

- `ALTER … SET REFRESH LIMIT` **extend** reclaiming/overwriting backfill in the newly-managed range — there is an explicit guard test and the monotonic floor clamp was deliberately omitted.
- `SET REFRESH LIMIT 0` validator pass-through — documented; transient rows are reset by the next FULL refresh's `truncateSoft`.
- `materialized_views()` sampler-cache reuse; period-view + `REFRESH LIMIT` range refresh (its lower bound stays at `lastPeriodHi`, never the frozen zone — regression test added).
- A couple of low/informational items: transient `_mv.s`-read failure during restore (loss-of-protection, but unavoidable corruption handling); a narrow in-memory frontier-vs-refresh TOCTOU whose restart variant is closed by the persistence fix.

### Round 5 — clean

Targeted the round-4 grid sampler with a harness built against the compiled classes: **~1.16 billion** scenarios (DST zones × strides {1d, 2d, 7d, 1w, 1M} × offsets {0, 30m, 60m, 120m} around every 2019–2027 transition; spring-forward gaps; `WITH OFFSET`; and end-to-end modeling of the exact validator decision vs the exact `REPLACE_RANGE` coverage) → **zero data-loss findings, zero wrong-rejects**. One benign single-point quirk was documented rather than changed (`9b91b269`).

---

**Five passes, five real data-loss bugs fixed, round 5 clean.** To the depth five independent passes can establish, the frozen-zone path is free of data-loss defects. Any further change to this code would warrant one more pass on the new diff.

nwoolmer:
## CodeRabbit review — validation

CodeRabbit reviewed `2a2e83d` (the months-limit fix), before the timezone fixes landed. I validated all four comments against the current tree; addressed in `f1660cf612`.

| # | Comment | Verdict |
|---|---------|---------|
| 1 | Published frozen boundary doesn't use the tz-aware bucket floor (`MatViewRefreshJob`) | **Valid — already fixed.** The published floor is now derived via `MatViewBackfillValidator.createBucketSampler(...)`, the same tz-aware grid the refresh iterator and the validator use (`adbb91c`, `4d7fdcd`). This is the same tz-grid mismatch caught in the level-3 review above — good corroboration. |
| 2 | Month-anchor adjustment may return an under-protective anchor / overflow (`ApplyWal2TableJob`) | **Applied.** The day-bump loop now fails safe: it saturates to `Long.MAX_VALUE` (whole view frozen — conservative) rather than ever returning an anchor whose boundary still underflows `bucketEnd`, with an overflow guard on `+= oneDay`. In practice the day-of-month clamp deficit resolves in a single step and the anchor is nowhere near overflow, so this is defence-in-depth that makes the safety invariant explicit. |
| 3 | QWP test should poll `isBackfillableMatView` (`QwpIngressProcessorStateTest`) | **Not applicable.** `LineHttpSenderTest` polls because it runs a full async `TestServerMain` (WAL applied on a worker pool). The QWP test uses the synchronous `drainWalQueue()` helper, so the backfill gate is visible as soon as the drain returns — it is not timing-sensitive. |
| 4 | Fuzz loop catches `Throwable`, masking fatal failures (`ServerMainMatViewBackfillTest`) | **Applied.** Narrowed to `catch (Exception)` so an `AssertionError` / `OutOfMemoryError` fails the test fast instead of being miscounted as a backfill rejection. |

Two applied, two explained. Full mat-view suite green (581/581).

nwoolmer:
This one is green and rebased on master — @bluestreak01 could you take a look when you have a moment? Thanks.

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=259135)
**Tested Commit:** `43f1b5a8a3caa53ff5d4bff7867190e74ae857a0`

ideoma:
#### [PR Coverage check]

:heart_eyes: **pass** : 432 / 512 (84.38%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/line/udp/LineUdpParserImpl.java|8|12|66.67%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewBackfillValidator.java|113|159|71.07%|
|:large_blue_circle:|io/questdb/cairo/wal/WalWriter.java|11|13|84.62%|
|:large_blue_circle:|io/questdb/cairo/wal/ApplyWal2TableJob.java|94|110|85.45%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/MatViewsFunctionFactory.java|40|46|86.96%|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|8|9|88.89%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|30|33|90.91%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewStateReader.java|37|38|97.37%|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/LineTcpMeasurementScheduler.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/LineHttpTudCache.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewState.java|32|32|100.00%|
|:large_blue_circle:|io/questdb/cairo/pool/WalWriterPool.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/DatabaseCheckpointAgent.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cutlass/text/CairoTextWriter.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalUtils.java|13|13|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|30|30|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpTudCache.java|1|1|100.00%|

PR #7165, opened 2026-05-26
fix(sql): fix statistical aggregates returning NULL on sparse-NULL data

### LINKED ISSUE #7160: corr() returns NULL on parallel GROUP BY with sparse-NULL data
## Summary

`corr(y, x)` can return NULL instead of the correct correlation coefficient under parallel GROUP BY when the data contains many NULL pairs distributed across worker partials. The bug surfaces only under the parallel-merge path (`cairo.sql.parallel.groupby.enabled=true`, default), not the single-threaded path.

## Reproduction

```sql
CREATE TABLE tbl AS (
    SELECT
        CASE WHEN x BETWEEN 1500 AND 1509 THEN cast(x AS double) ELSE cast(null AS double) END x,
        CASE WHEN x BETWEEN 1500 AND 1509 THEN cast(2 * x + 5 AS double) ELSE cast(null AS double) END y
    FROM long_sequence(4_000)
);

SELECT corr(y, x) FROM tbl;
-- Expected: 1.0 (perfect linear y = 2x + 5)
-- Observed under parallel GROUP BY: NULL
```

## Root cause

`CorrGroupByFunction.merge()` in `core/src/main/java/io/questdb/griffin/engine/functions/groupby/CorrGroupByFunctionFactory.java:159-195` lacks the empty-partial guards.

The 6-slot Welford state layout is `(meanY, Syy, meanX, Sxx, Sxy, count)`. When a worker only sees non-finite `(y, x)` pairs for a key, `computeFirst()` initializes the slot to `(0, 0, 0, 0, 0, 0)` with `count = 0`. The sharded merger then chains:

1. Wholesale-copies worker A's `(0, 0, 0, 0, 0, 0)` into the destination map.
2. `merge(dest=zeros_count0, src=zeros_count0)`: `mergedCount = 0`, `weighting = 0/0 = NaN`, `mergedMeanY = 0/0 = NaN`. Destination is now `(NaN, NaN, NaN, NaN, NaN, 0)`.
3. `merge(dest=NaN_count0, src=real_countN)`: `mergedMeanY = (N · real + 0 · NaN) / N = NaN` (because `0 · NaN = NaN`). Destination stays poisoned with `count > 0`.
4. `getDouble()` reads `sumY = NaN` and `sumX = NaN`, returns `NaN` -> SQL NULL.

## Fix

#7104 fixed the identical bug in the regression family (`regr_slope`, `regr_intercept`, `regr_r2`) by introducing `AbstractRegressionGroupByFunction.merge()` with two early-return guards at lines 138-163:

```java
long srcCount = srcValue.getLong(valueIndex + 5);
if (srcCount == 0) {
    return;
}
long destCount = destValue.getLong(valueIndex + 5);
if (destCount == 0) {
    destValue.putDouble(valueIndex, srcValue.getDouble(valueIndex));
    // ... copy all slots from src
    return;
}
```

`CorrGroupByFunction` uses an identical 6-slot layout, so it can extend `AbstractRegressionGroupByFunction` directly — only `getDouble()` (corr = `sumXY / sqrt(Sxx · Syy)`) and `getName()` ("corr") would need to be implemented. Alternatively, duplicate the guards into the existing `CorrGroupByFunction.merge()` without refactoring.

## Test coverage

`#7104` adds `AbstractRegressionParallelGroupByTest` with `testParallelSparseNulls`, `testParallelAllNull`, and `testParallelKeyedSparseNulls` to exercise the parallel-merge guards. A `CorrParallelGroupByTest` modeled on this harness would lock in the fix and regression-test future merge changes.

## Scope

`#7104` explicitly defers this:

> Does not migrate corr() onto the new base, even though its state layout is identical. The migration is mechanical but touches a different test surface; deferred.

The same follow-up could also cover the remaining six SQL-standard regression aggregates (`regr_count`, `regr_avgx`, `regr_avgy`, `regr_sxx`, `regr_sxy`, `regr_syy`) as listed in #7104's follow-up section, since they would all share the same base.

brunocalza:
## Summary

Fixes #7160 

`corr`, `stddev_pop`, `stddev_samp`, `var_pop`, `var_samp`, `covar_pop`, and `covar_samp` were returning `NULL` on sparse-NULL data on GROUP BY parallel.

The merge algorithm that is called when parallel workers need to merge partial results of the aggregate was not considering empty partials. When partials were empty, a NaN value was being propagated and was surfaced as an SQL `NULL`. 

## Changes

- Applied the same merge fix guard from #7104 to all those aggregates
- Tests that confirmed the bug were added, and after the fix, they became green

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/functions/groupby/AbstractStdDevGroupByFunction.java
+++ b/core/src/main/java/io/questdb/griffin/engine/functions/groupby/AbstractStdDevGroupByFunction.java
@@ -100,13 +100,23 @@ public abstract class AbstractStdDevGroupByFunction extends DoubleFunction imple
     // Chan et al. [CGL82; CGL83]
     @Override
     public void merge(MapValue destValue, MapValue srcValue) {
+        long srcCount = srcValue.getLong(valueIndex + 2);
+        if (srcCount == 0) {
+            return;
+        }
+        long destCount = destValue.getLong(valueIndex + 2);
+        if (destCount == 0) {
+            destValue.putDouble(valueIndex, srcValue.getDouble(valueIndex));
+            destValue.putDouble(valueIndex + 1, srcValue.getDouble(valueIndex + 1));
+            destValue.putLong(valueIndex + 2, srcCount);
+            return;
+        }
+
         double srcMean = srcValue.getDouble(valueIndex);
         double srcSum = srcValue.getDouble(valueIndex + 1);
-        long srcCount = srcValue.getLong(valueIndex + 2);
 
         double destMean = destValue.getDouble(valueIndex);
         double destSum = destValue.getDouble(valueIndex + 1);
-        long destCount = destValue.getLong(valueIndex + 2);
 
         long mergedCount = srcCount + destCount;
         double delta = destMean - srcMean;
--- a/core/src/main/java/io/questdb/griffin/engine/functions/groupby/CorrGroupByFunctionFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/functions/groupby/CorrGroupByFunctionFactory.java
@@ -157,19 +157,32 @@ public class CorrGroupByFunctionFactory implements FunctionFactory {
         // Chan et al. [CGL82; CGL83]
         @Override
         public void merge(MapValue destValue, MapValue srcValue) {
+            long srcCount = srcValue.getLong(valueIndex + 5);
+            if (srcCount == 0) {
+                return;
+            }
+            long destCount = destValue.getLong(valueIndex + 5);
+            if (destCount == 0) {
+                destValue.putDouble(valueIndex, srcValue.getDouble(valueIndex));
+                destValue.putDouble(valueIndex + 1, srcValue.getDouble(valueIndex + 1));
+                destValue.putDouble(valueIndex + 2, srcValue.getDouble(valueIndex + 2));
+                destValue.putDouble(valueIndex + 3, srcValue.getDouble(valueIndex + 3));
+                destValue.putDouble(valueIndex + 4, srcValue.getDouble(valueIndex + 4));
+                destValue.putLong(valueIndex + 5, srcCount);
+                return;
+            }
+
             double srcMeanY = srcValue.getDouble(valueIndex);
             double srcSumY = srcValue.getDouble(valueIndex + 1);
             double srcMeanX = srcValue.getDouble(valueIndex + 2);
             double srcSumX = srcValue.getDouble(valueIndex + 3);
             double srcSumXY = srcValue.getDouble(valueIndex + 4);
-            long srcCount = srcValue.getLong(valueIndex + 5);
 
             double destMeanY = destValue.getDouble(valueIndex);
             double destSumY = destValue.getDouble(valueIndex + 1);
             double destMeanX = destValue.getDouble(valueIndex + 2);
             double destSumX = destValue.getDouble(valueIndex + 3);
             double destSumXY = destValue.getDouble(valueIndex + 4);
-            long destCount = destValue.getLong(valueIndex + 5);
 
             long mergedCount = srcCount + destCount;
             double deltaY = destMeanY - srcMeanY;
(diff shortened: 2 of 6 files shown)


core/src/main/java/io/questdb/griffin/engine/functions/groupby/AbstractStdDevGroupByFunction.java
```diff
@@ -100,13 +100,23 @@ public boolean isConstant() {
     // Chan et al. [CGL82; CGL83]
     @Override
     public void merge(MapValue destValue, MapValue srcValue) {
+        long srcCount = srcValue.getLong(valueIndex + 2);
```
brunocalza, line 103: `AbstractStdDevGroupByFunction.merge` fixes `stddev_pop`, `stddev_samp`, `var_pop`, `var_samp` and the aliases `stddev` and `variance`

core/src/main/java/io/questdb/griffin/engine/functions/groupby/CorrGroupByFunctionFactory.java
```diff
@@ -157,19 +157,32 @@ public boolean isConstant() {
         // Chan et al. [CGL82; CGL83]
         @Override
         public void merge(MapValue destValue, MapValue srcValue) {
+            long srcCount = srcValue.getLong(valueIndex + 5);
```
brunocalza, line 160: `CorrGroupByFunctionFactory.merge` fixes `corr`

core/src/main/java/io/questdb/griffin/engine/functions/groupby/CovarSampleGroupByFunctionFactory.java
```diff
@@ -82,15 +82,26 @@ public String getName() {
         // Chan et al. [CGL82; CGL83]
         @Override
         public void merge(MapValue destValue, MapValue srcValue) {
+            long srcCount = srcValue.getLong(valueIndex + 3);
```
brunocalza, line 85: `CovarSampleGroupByFunctionFactory.merge` fixes `covar_pop`, and `covar_samp`

bluestreak01 (approved):
Production code fix is correct, complete, and matches the established pattern from PR #7104. The two early-return guards in `merge()` prevent NaN poisoning on all Welford/Chan merge paths — verified across `AbstractStdDevGroupByFunction`, `CorrGroupByFunction`, and `CovarSampleGroupByFunction`. All subclasses inherit the fix correctly; no unguarded merge methods remain.

A few test-hardening suggestions (non-blocking):

- **Run sparse-null queries 8 times** (as in `AbstractRegressionParallelGroupByTest`) to reliably exercise the `destCount == 0` guard, which fires nondeterministically.
- **Wrap assertions in `round(..., 10)`** in the covar and stddev tests to absorb floating-point drift from nondeterministic merge order.
- **Add `testParallelAllNull`** to `CovarParallelGroupByTest` and `StdDevParallelGroupByTest` (already present in `CorrParallelGroupByTest`).
- Minor: add `Bug` + `SQL` labels; replace em dash (U+2014) with ASCII dash in `CorrParallelGroupByTest.java:36`.

bluestreak01 (approved):
Approve.

The production fix is correct, complete, and root-caused. The two early-return guards in each `merge()` prevent NaN poisoning on every Welford/Chan parallel-merge path, verified across `AbstractStdDevGroupByFunction`, `CorrGroupByFunction`, and `CovarSampleGroupByFunction`. All subclasses (`var_pop`/`var_samp`, `stddev_pop`/`stddev_samp`, `covar_pop`/`covar_samp`, `corr`) inherit the fix; no merge() in the family is missed; `count` (long) is a robust emptiness sentinel for both the `computeFirst` (zeroed) and `setNull` (NaN) init paths. No correctness, concurrency, or resource issues.

Non-blocking follow-ups (test hardening + metadata; can land separately):

- Sparse-null tests run once rather than 8x. The `destCount == 0` copy branch — the exact path this PR fixes — fires roughly 75% of the time per run, so a single run is a weaker regression guard than the #7104 precedent (`AbstractRegressionParallelGroupByTest` loops 8x for this reason).
- `testParallelAllNull` is present for corr but missing for covar and stddev, leaving the all-empty terminal case (result must be NULL) unasserted for those families.
- ASCII-only rule: em dash (U+2014) in `CorrParallelGroupByTest.java:36`.
- Add `Bug` and `SQL` labels.
- Pre-existing (out of scope): `AbstractStdDevGroupByFunction:130` and `CorrGroupByFunctionFactory:197-200` compute `(double)(srcCount * destCount)` (multiply-then-cast); `CovarSampleGroupByFunctionFactory:117` and the #7104 reference cast first. Aligning would complete the parity.

Note: the non-rounded exact-value assertions are safe here — the 10 non-null rows fall in a single page frame, so only one non-empty partial exists and merge() never runs the two-partial Chan arithmetic that would introduce float drift.

PR #7170, opened 2026-05-27
chore(core): storage policy TTL interval forward shift

glasstiger:
Adds `TableUtils.checkStoragePolicyTtl`, a storage-policy variant of the existing `checkTtl` partition-age check. Both methods now share a private `isOlderThanTtl` helper.

The two checks differ only in the boundary they measure from:

- `checkTtl` (table TTL) measures a partition's age from its **ceiling** — the start of the next logical partition. A partition is old enough only once even its newest possible record exceeds the TTL.
- `checkStoragePolicyTtl` measures age from the partition's **floor** — its own start. This shifts every threshold forward by one partition width: a partition qualifies as soon as the next (active) partition begins, as long as the interval does not exceed the partition width.

Table TTL behaviour is unchanged. `checkTtl` keeps the ceiling-based calculation and is only refactored to call the shared helper.

The new method exists for questdb/questdb-enterprise#1032 to call; there is no OSS caller yet, so on its own this PR only adds the helper and its unit tests.

Tradeoffs:

- The two checks now look almost identical and differ only in floor vs. ceiling, so they could be misread as duplicates or accidentally collapsed back together. `testCheckStoragePolicyTtlShiftsOnePartitionVsTtl` pins the distinction.

Test plan:

- `TableUtilsTest` gains unit tests covering the floor-based hours boundary, the months boundary, the zero-TTL assertion guard, and the one-partition shift relative to `checkTtl`.
- The existing `checkTtl` tests are unchanged and continue to pass.

Required by questdb/questdb-enterprise#1032.


mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableUtils.java|7|7|100.00%|

bluestreak01 (approved):

PR #7171, opened 2026-05-27
perf(sql): single-pass parquet decode for covering index build

bluestreak01:
## Summary

- Merge the two separate row-group decode loops in `indexParquetPartition` into a single pass that decodes the SYMBOL column and all covered columns together in one `decodeRowGroup` call per row group
- Replace the `write()`/re-`mmap` round-trip for covered column data with direct mmap-backed `MemoryMARW` writes: decoded data flows into the mmap and the same addresses feed `PostingIndexWriter`'s addr-based `configureCovering` path during seal
- Eliminate the separate `materialiseCoveredColumnsFromParquet` call on this code path

### Before

1. `materialiseCoveredColumnsFromParquet` decoded all covered columns from every row group, writing to temp files via `write()` syscalls
2. A second row-group loop decoded just the SYMBOL column and fed it to `indexWriter.add()`
3. `seal()` opened the temp files via `mmap` to read covered column data for sidecar writing

The parquet file was fully decoded twice and the covered column data made a disk round-trip through `write()` then `mmap` read.

### After

1. A single row-group loop decodes SYMBOL + all covered columns together
2. Covered column data accumulates into `MemoryMARW` instances (file-backed mmap) via `putBlockOfBytes`
3. `seal()` reads from the same mmap addresses — no re-open, no second decode pass

## Test plan

- CoveringIndexTest (364 tests) passes, including all parquet-specific covering index tests
- O3ParquetPartitionFuzzTest passes
- PostingIndexOomFallbackTest passes
- Updated fault-injection test to match new file-open path (`openRW` instead of `openAppend`)

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -825,17 +825,22 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
         columnMetadata.setCoveringColumnIndices(coveringColumnIndices);
 
         final SymbolColumnIndexer indexer = new SymbolColumnIndexer(configuration, indexType);
-        writeIndex(columnName, indexValueBlockSize, indexType, columnIndex, indexer);
+        try {
+            writeIndex(columnName, indexValueBlockSize, indexType, columnIndex, indexer);
 
-        columnMetadata.setIndexType(indexType);
-        columnMetadata.setIndexValueBlockCapacity(indexValueBlockSize);
+            columnMetadata.setIndexType(indexType);
+            columnMetadata.setIndexValueBlockCapacity(indexValueBlockSize);
 
-        // set the index flag in metadata and create new _meta.swp
-        rewriteAndSwapMetadata(metadata);
-        clearTodoAndCommitMeta();
+            // set the index flag in metadata and create new _meta.swp
+            rewriteAndSwapMetadata(metadata);
+            clearTodoAndCommitMeta();
 
-        indexers.extendAndSet(columnIndex, indexer);
-        populateDenseIndexerList();
+            indexers.extendAndSet(columnIndex, indexer);
+            populateDenseIndexerList();
+        } catch (Throwable th) {
+            Misc.free(indexer);
+            throw th;
+        }
 
         try (MetadataCacheWriter metadataRW = engine.getMetadataCache().writeLock()) {
             metadataRW.hydrateTable(metadata);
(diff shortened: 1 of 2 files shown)


bluestreak01:
## Code review — single-pass parquet decode for covering index build

Level-3 review (parallel multi-agent + per-finding source verification). Scope: `TableWriter.java` + `CoveringIndexTest.java`. 20 draft findings → 12 verified, 8 dropped as false positives. No Critical findings.

The core data path is sound: the aux/data row-group stitching is a faithful byte-for-byte port of the removed `materialiseCoveredColumnsFromParquet`; the addr-based seal reads are provably in-bounds under the enforced `columnTop ∈ {0, partitionSize}` invariant; the mmap ownership model has no double-free (`coveredColReadSizes=0` → release no-ops, `TableWriter` solely owns `covMmaps`); and the path is single-threaded per writer. The headline perf claim (two decode passes → one, `write()`+re-mmap round-trip eliminated) holds.

### Moderate

**M1. Undocumented removal of a corruption tripwire in the native seal path** — `TableWriter.java:7284` (`mapCoveringColumnsForSeal`). The throw for an empty fixed-size covered file was replaced with a silent skip (`addr=0` → all-null). This method is **not on the parquet path** — its callers are the native O3 seal (`11841`) and WAL fast-lag seal (`10530`).

- Correct for the legitimate all-null case (`colTop == partitionSize`): the reader synthesizes nulls via the `rowId < colTop` fast path, so the old throw was a false failure.
- But for a genuinely corrupt 0-byte `.d` on a column that should have data (`colTop < partitionSize`), the old code failed loud; the new code silently returns wrong (all-null) results.

This change is orthogonal to "single-pass parquet decode," is not mentioned in the PR description, and has no test. Suggest gating the tolerance (`throw` only when `colTop < partitionSize`) or adding a regression test for the legitimate case it fixes.

**M2. Peak-memory tradeoff for fixed-size covered columns** — `TableWriter.java:8537-8544` (`prepareCoveredColumnMmaps`). Fixed-size data scratch is presized to `ceilPageSize(sizeOf × partitionSize)` and opened via `of(ff, file, dataPreSize, 0L, …)`. With `size=0`, `MemoryCMARWImpl.map()` immediately allocates + maps the **entire** partition-sized file, for all K covered columns simultaneously, held through decode + seal — e.g. 10 `LONG` covered columns over a 50M-row partition ≈ 4 GB mapped RW before the first byte is decoded. The old path streamed via `write()` during decode and mapped covered files RO only at seal, so its decode-phase peak was ~0 covered mappings. This is a real `O(K×N)` peak mapped/resident regression.

Mitigations (equal weight): file-backed and evictable (not anonymous), on-disk usage unchanged, `ADD INDEX` is one-shot DDL. Fix: presize fixed-size data to `getDataAppendPageSize()` like var-size already does, letting `extend0`/`mremap` grow it — caps peak at `O(K × pageSize)`. The description lists only benefits and omits this tradeoff.

**M3. Test gap: var-size data-mmap extend (mremap) is never exercised** — the riskiest new path: var-size data written via `putBlockOfBytes` triggers `mremap` (relocating the base) once the page size is exceeded; `configureCoveringFromMmaps` then reads the post-extend `addressOf(0)`. Correct by design, but no test forces an extend — every parquet var-size test uses a few KB, and all large var-size covering tests are native (different memory class). Add a parquet test with a VARCHAR/STRING/BINARY covered column whose decoded data exceeds the test `dataAppendPageSize` (2 MB) over ≥2 row groups, asserted via a `WHERE sym = …` covering query (a bare `ORDER BY ts` does not select the covering factory).

### Minor

- **m1.** `addIndex` double-frees the indexer via double-close (`TableWriter.java:840-843`): `writeIndex` already frees it and rethrows (`12921-12926`), then the new outer catch frees it again. Benign only because `close()` is idempotent. Also, if `populateDenseIndexerList()` throws after `indexers.extendAndSet` (OOM only), the catch frees an indexer still referenced in `indexers`. Scope the catch with a `built` flag (or install into `indexers` last). The intended post-build-failure case is real and tested.
- **m2.** `getColumnTop` (prepare, `8506`, returns -1 for absent) vs `getColumnTopQuick` (configure, `4880`, returns 0) inconsistency. Benign today (skipped slots have `addr=0` → null at seal regardless), but use the same accessor.
- **m3.** All-null chunk does `malloc → fill → memcpy → free` per row group (`3638`, `3652`); the presized mmap exposes `appendAddressFor`, so the fixed/STRING-data branches can fill the null pattern directly into the mmap in one pass. Fires only on null-region row groups (low impact).
- **m4.** Loop-invariant `ColumnType.getDriver`/`isVarSize` recomputed per slot per row group in `accumulateCoveredColumnsFromRowGroup` (`3729-3746`); cache them in `prepareCoveredColumnMmaps`.
- **m5.** ~70 lines duplicated: `accumulateAllNull{Fixed,VarSize}Chunk` (mmap) vs the still-live `appendAllNull{Fixed,VarSize}Chunk` (fd). Justified by different sinks, but a divergence risk.
- **m6.** Stale naming: `cleanupMaterialisedCoveredColumnTempFiles` and `testMaterialiseFromParquetCleansUpTempFilesOnOpenAppendFailure` retain "Materialise"/"OpenAppend" though that path now mmaps via `openRW`.
- **m7.** Misleading docstring in `prepareCoveredColumnMmaps` (`8472-8482`): claims the all-null skip fires via "columnTop >= partitionSize", but for columns added after the partition `getColumnTop` returns -1, so the actual skip is the `findParquetColumnIndex < 0` branch.
- **m8.** `belowColumnTop` is defensive dead code — never true while the loop runs given the asserted invariant + `partitionSize > columnTop` guard. Either drop it or note it guards a future ATTACH-PARQUET path. Related: the covered-column colTop invariant is relied on but not asserted (only the symbol's is), and the addr-based seal has no bounds check — an invariant violation reads OOB rather than failing cleanly. Consider asserting covered colTop too.
- **m9.** Trivia: em-dash in the comment at `6897` (prefer ASCII `--`); the comment in `testAlterAddIndexParquetCoveredMmapLeakOnAuxOpenFailure` says `dataMem` is added "only after both opens succeed" but the code adds it before the aux open (code is safer than the comment).
- **m10.** PR metadata: title doesn't repeat a verb (e.g. `perf(sql): speed up …`); description is implementation-focused and one-sided (omits the M2 tradeoff and the M1 behavior change); consider a `storage`/`Core` label.

### Dropped as false positives (after source verification)

- **Double-free crashes the JVM** — `close()` is idempotent end-to-end (every native free addr-guarded + zeroed). Reclassified to m1.
- **Stale freed addresses linger in `coveredColumnAddrs` → use-after-free on next commit** — `releaseCoveredColumnReadMappings()` clears the borrowed-address copy *before* `covMmaps` is freed, and the next `configureCovering` re-clears both lists.
- **Concurrency race on `coveringAddrs`/shared fields** — `ADD INDEX` indexes partitions sequentially on the writer thread; addresses are copied into the indexer before `seal()`, all in one call stack.
- **`covSlotMeta` leaks (no `resetCapacity`)** — it's the pooled `tempDirectMemList`, freed at writer close.
- **Correctness bug in aux/data stitching** — byte-for-byte port; totals equal `getAuxVectorSize(partitionSize)` exactly (STRING N+1 and VARCHAR 16B/row verified).
- **`prepareCoveredColumnMmaps` leaks the `.d` mmap if `.i` open fails** — `dataMem` is added to `covMmaps` before the aux open, so the finally reclaims it.
- **Fixed presize causes repeated remaps** — it's a single up-front mapping (that's the basis of M2, not a remap problem).
- **Vacuous `assertSqlCursors` in `AcrossManyParquetRowGroupsAllVarSize`** — out of scope (pre-existing test from #7145, not this PR), though it's a real pre-existing test-quality issue: the full-scan compare doesn't select the covering factory.

### Verdict

Needs discussion, not blocking. No correctness, concurrency, or resource-safety defects survived verification. Address M1–M3 before merge (disclose/fix the memory tradeoff, gate or test the corruption-tripwire removal, cover the var-size mmap extend); the minors can ride along or follow up.

bluestreak01:
**M1 and M3 follow-up** (commit e82ddff496).

**M3 (var-size scratch mmap extend) — addressed.** Added `testAddPostingCoveringIndexParquetVarSizeScratchMmapExtendsWal`: a parquet partition with a STRING covered column (~3.2 MB of decoded `.d` data over four 2000-row groups) whose running total crosses the 2 MB test `dataAppendPageSize`, forcing the var-size data scratch to extend mid-decode. Instrumented confirmation shows the `mremap` relocates the mapping base on the third row group; the remaining groups append past the relocation and `configureCoveringFromMmaps` reads the post-extend `addressOf(0)`. The covering scan matches the `no_covering` fallback row-for-row, so the post-extend base resolves correctly. This is the first test that forces the extend.

**M1 (removed tripwire) — investigated; narrower than the finding implied.** The throw is in `mapCoveringColumnsForSeal`, the commit-seal path (O3 reseal / WAL fast-lag), not the `ADD INDEX` build. An all-null fixed covered column does not reach that branch with a 0-byte `.d`:
- O3 materializes an all-null fixed column (`colTop`->0, explicit null sentinels), so its `.d` is non-empty at reseal.
- The build path uses `PostingIndexWriter.mapColumnFile`, which already skips empty files and never threw.

Across the full `CoveringIndexTest` suite plus a direct O3 reproduction attempt, the only 0-byte covered file observed at commit-seal was an inlined VARCHAR (var-size, always skipped). So the removed fixed-size throw guarded a state normal SQL does not produce: the removal is low-risk, but the exact line is not reproducible without injecting file corruption. Added two tests pinning the native all-null indexing path (the analog of the parquet all-null skip), covering the fixed-size case the finding flagged: `testAddPostingCoveringIndexIncludesFixedSizeColumnAddedToFullyPopulatedNativeWal` and `testAddPostingCoveringIndexAllNullFixedAndVarSizeNativeWal`.

M2 (peak-memory tradeoff) is not addressed here.

bluestreak01:
**M2 follow-up: re-examined, the peak-memory framing is overstated.**

Tracing the old vs new profile against the code, the "O(K x N) peak mapped/resident regression" does not hold up - it is a memory-*profile* change, not a higher peak.

The old `materialiseCoveredColumnsFromParquet` path did stream covered data via `write()` during decode (decode-phase covered footprint ~0), but the subsequent `seal()` mapped every covered column file RO at once: `PostingIndexWriter.ensureCoveredColumnReadMaps` loops `c = 0..coverCount` and full-file mmaps each via `mapColumnFile`, and the old `indexParquetPartition` held those K mappings alongside the parquet decoder mmap through sidecar writing. So the old *overall* peak was already O(K x N) mapped, at seal. The new path peaks at the same O(K x N); it reaches that peak during decode and holds it through seal rather than only at seal.

Two further points:
- `mmap` does not pin resident memory. The "~4 GB mapped RW before the first byte is decoded" is virtual address space plus a file-length reservation (`MemoryCMARWImpl.map()` with `size=0` maps `minMappedMemorySize`); RSS only grows as covered data is decoded and written, which happens in both paths because the data has to be materialized regardless. The pages are file-backed and reclaimable either way.
- The suggested fix (presize fixed-size to the data-append page and grow via `mremap`) would not cap the peak. The scratch must hold all decoded covered data until `seal()` reads it, so a grow-on-demand mapping reaches the same O(N) per column by end of decode (the var-size path already does this). Grow-on-demand only trims the up-front virtual reservation, at the cost of ~N/pageSize extra `mremap` calls per column, which works against the single-map intent. Because fixed-size knows its exact length, the one-shot map is both syscall-minimal and size-exact.

What is genuinely new and worth disclosing: for fixed-size covered columns the scratch is mapped RW at the exact partition size up front and stays live for the whole build (decode + seal), versus the old RO mapping that existed only at seal. That is a longer-lived, writable, pre-allocated mapping - a behavioral change, not a higher peak, and it is the cost of collapsing two decode passes and the write/reopen round-trip into one.

Net: downgrading M2 from "address before merge" to a one-line behavioral note in the PR description; no code change.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 184 / 205 (89.76%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|184|205|89.76%|

Issue #4620, opened 2024-05-29
domain specific functions for finance and aggregation


nwoolmer:
### Is your feature request related to a problem?

Many users of QuestDB are in the finance industry. Within this domain, there are common operations that can be complicated to represent in standard SQL.

We should add functions to make these operations easier for our users, and start with the most common operations.

Such functions should support `ARRAY` types in future, where useful for performance improvements.

This is a tracking issue, and we welcome suggestions from users!

### Describe the solution you'd like.

More domain-specific finance functions around trading and order books.

- [x] **l2price**: level two order-book price
    - PR: https://github.com/questdb/questdb/pull/4462
    - Issue: https://github.com/questdb/questdb/issues/4454
- [x] **vwap**: volume-weighted average price
    - PR: https://github.com/questdb/questdb/pull/3885
- [ ] **twap**: time-weighted average price
    - PR: https://github.com/questdb/questdb/pull/4670
    - PR2: https://github.com/questdb/questdb/pull/5085
- [x] **avg**: avg window function, can be used to calculate simple moving average
    - docs: https://questdb.io/docs/reference/function/aggregation/#avg 
- [ ] **vwma**: volume-weighted moving average (window)
    - Issue: https://github.com/questdb/questdb/issues/4727
    - **not looking for contributions, some internal work needed**
- [x] **mid(bid,ask)**: `average(bid, ask) = bid+ask / 2`
    - Issue: https://github.com/questdb/questdb/issues/4730 
    - PR: https://github.com/questdb/questdb/pull/4734
- [x] **spread(bid, ask)**: `ask-bid`
    - Issue: https://github.com/questdb/questdb/issues/4731
    - PR: https://github.com/questdb/questdb/pull/4744
- [x] **wmid(bid_size, bid, ask, ask_size)**: weighted mid point: 
    - `wmid = (imbalance * ask_price) + (1 - imbalance) * bid_price` where `imbalance = bid_size / (bid_size + ask_size)`
    - Issue: https://github.com/questdb/questdb/issues/4800
    - PR: https://github.com/questdb/questdb/pull/4801
- [x] **spread_bps(bid, ask)**:  ` spread(bid,ask)/mid(bid,ask) * 10,000`
    - PR: https://github.com/questdb/questdb/pull/4939 
- [ ] **format_price(decimal_form, tick_size)**: treasury price format, decimal to fraction, e.g. `format_price(100.5, 32)= 100-16`
[...]

PR #4960, opened 2024-09-16
feat(sql): add format_price function for finance

thalesreisbr:
Implements format_price to #4620

PR #7175, opened 2026-05-27
feat(sql): add format_price function for finance

anish3333:
Implements the `format_price(decimal_price, tick_size)` finance function proposed in #4620.

This PR adds a new finance-oriented SQL function:

```sql id="p7m1k5"
format_price(decimal_price, tick_size)
```

The function converts decimal prices into fixed fractional tick notation commonly used in fixed-income and treasury-style pricing.

Examples:

```sql id="m4v8k2"
select format_price(100.5, 32);
-- 100-16

select format_price(99.375, 32);
-- 99-12

select format_price(100.5, 128);
-- 100-064
```

This implementation revisits and extends the previous unmerged PR #4960.

Changes and improvements compared to the previous implementation include:

* uses `VarcharFunction` + `Utf8StringSink` to avoid UTF16-to-UTF8 re-encoding on the wire
* avoids `String.format()` and intermediate string allocations on the hot path
* uses reusable UTF8 sinks for zero-allocation formatting
* supports dynamic tick widths (e.g. 32 -> 2 digits, 128 -> 3 digits)
* supports negative prices
* supports constant tick-size optimization through cached width calculation
* adds floating-point overflow normalization safety
* follows QuestDB naming/style conventions (`camelCase`, sink-based formatting, etc.)

The implementation currently formats prices using integer tick fractions only.

For example:

```sql id="x2m7v4"
format_price(100.5, 32) -> 100-16
```

This PR does not yet implement treasury-specific sub-tick notation such as:

* `+`
* quarter 32nds
* eighth-of-32nd encoding

which can be added later as a separate extension if desired.

Tests added include:

* multiple tick sizes
* negative values
* null handling
* invalid tick sizes
* multi-row result sets
* ordering behavior
* constant and non-constant tick paths
* large whole-number values
* rounding behavior
### Implementation Notes

A few implementation details that may be useful during review:

* constant tick sizes cache formatting width during `init()` to avoid repeated per-row width calculation
* formatting width is derived dynamically from the tick size (e.g. `32 -> 2 digits`, `128 -> 3 digits`)
* floating-point overflow normalization is included to safely handle edge rounding cases such as `31.999999999 -> 32`


bluestreak01 (changes requested):
@anish3333 Thanks for the contribution. I did a deep review (verified end-to-end: the function compiles, is auto-discovered by `FunctionFactoryScanner`, and all 56 tests pass). There's one blocking correctness issue around NULL handling that the test suite currently can't catch, plus a performance miss and some standards items. Details below.

## Critical

### 1. `format_price` returns an empty string `''` instead of SQL NULL for null/invalid inputs
`FormatPriceFunctionFactory.java:144-146`
```java
if (Numbers.isNull(decimalPrice) || tick <= 0) {
    return sink;   // sink was just cleared -> empty Utf8 sequence, NOT null
}
```
Every other varchar function returns Java `null` to signal SQL NULL — e.g. `MD5VarcharFunctionFactory` (`if (sequence == null) return null;`), `SplitPartVarcharFunctionFactory`, `TrimVarcharFunctionFactory`. This function returns a length-0 sink instead, conflating NULL with the empty string.

Proven empirically against the engine:
- `format_price(NULL, 32) IS NULL` -> **false**
- `format_price(1.0, NULL) IS NULL` -> **false**
- `format_price(100.5, 0) IS NULL` -> **false**
- `length(format_price(NULL, 32))` -> **0**

This breaks `IS NULL` / `IS NOT NULL`, `DISTINCT` / `GROUP BY`, joins, and what pgwire/HTTP clients receive (they get `''`, not NULL). For a finance function fed market data with gaps, NULL-vs-empty is a meaningful correctness distinction.

**Fix:** `return null;` in the guard instead of `return sink;`.

### 2. The NULL test suite cannot detect issue #1 — false confidence
`testNullTickSize`, `testNullPrice`, `testBothNull`, `testNegativeTickSize`, `testZeroTickSize`, `testMultiRowNullsAndValidMixed`, etc. all assert `"format_price\n\n"`. `CursorPrinter` renders a `null` varchar and an empty sink identically (both emit nothing), so **every one of these tests passes whether the function returns `null` (correct) or an empty sink (current, wrong)** — the NULL suite pins nothing.

**Fix:** assert the actual semantics, e.g. `assertQuery("is_null\ntrue\n", "select format_price(NULL, 32) is null is_null")`, and add them together with the fix for #1 so they would fail without it.

## Moderate

### 3. Width is recomputed per row for parameterized (`:tick`) queries (performance)
`init()` caches `constantWidth` only when `tickSize.isConstant()` (compile-time constant). For the canonical `format_price(price_col, :tick)`, the bind variable is **runtime**-constant: `isConstant()` is false but `isRuntimeConstant()` is true, so `calculateWidth(tick)` — an integer-division loop — runs on every row even though the tick is fixed for the whole cursor.

**Fix:** cache when `tickSize.isConstant() || tickSize.isRuntimeConstant()`, mirroring `SplitPartVarcharFunctionFactory.init()`. `init()` runs per cursor after bind values are set, so `tickSize.getInt(null)` is valid there.

### 4. Banner comments violate an explicit project rule (test file)
The repo standard forbids `// ===` / `// ---` banner section headings in any Java file (production or test). The test file has ~9 of them. Please remove.

### 5. Tests use the single-pass helper, never exercising `getVarcharB` / random access
`AbstractFunctionFactoryTest.assertQuery(expected, sql)` runs a single cursor pass, so `getVarcharB(...)` (which has its own `sinkB`) and the random-access path are untested. Prefer the builder `assertQuery(sql).returns(expected)` — its second pass + `calculateSize()` + random-access checks cover those paths.

## Minor
- **Boolean naming:** `constantTick` field -> `isConstantTick`; local `negative` -> `isNegative` (project requires `is.../has...` prefixes).
- **Member ordering:** instance fields and public methods aren't sorted alphabetically (e.g. `init()` sits between `getRight()` and `getVarcharA()`).
- **Malformed output for `|price| >= 2^63`** (edge): `(long)` saturation makes `format_price(1e19, 32)` -> `"9223372036854775807-0-1"`. Unrealistic for prices but unguarded; consider returning null above the representable range.
- **`"-0-00"` for tiny negatives that round to zero** (edge): the sign is taken before rounding, so `format_price(-0.001, 32)` -> `"-0-00"`.
- **`calculateWidth(Numbers.INT_NULL)` underflows in `init()`** when the tick is a constant NULL (`temp = MIN_VALUE - 1`). Harmless (result unused) but worth guarding.
- Missing trailing newline in both files. PR has no labels (add `SQL` / `New feature`); consider `Closes #4620` in the body.

## Verified as NOT issues (in case they come up)
- **Registration:** not needed in `function_list.txt` — `FunctionFactoryScanner` reflectively discovers everything under `io.questdb.griffin.engine.functions.*`; that file only orders conflicting signatures.
- **Constant folding:** `BinaryFunction.isConstant()` already defaults to `left && right`, so `format_price(const, const)` is folded — no override needed.
- **No leak:** `Utf8StringSink` is heap-backed (`byte[]`), so the two sinks need no `close()`.
- **No data race:** `Function.isThreadSafe()` defaults to false, so the framework clones the function per worker; same pattern as `RPadVarcharFunctionFactory`.

## Summary
The numeric core (round-half-up on the absolute value, the `fractionalTicks == tick` carry, zero-pad width) is correct for the realistic price range and the test arithmetic checks out. The blocker is the **NULL semantics** (#1) combined with a test suite that is structurally blind to it (#2). Please address #1-#2, ideally #3-#5, and the minor items. Happy to re-review after.

anish3333:
@bluestreak01 Thank you for the detailed review, I will work on all the changes thoroughly.

anish3333:
@bluestreak01 I have gone through the changes and updated the PR:
* Fixed the `NULL` return bug and added strict `IS NULL` tests to catch it.
* Swapped all tests to use the `.returns()` builder to cover those random-access memory paths.
* Fixed the performance issue by caching the width for runtime variables.
* Handled the math edge cases (like the `-0-00` sign bug) and cleaned up the banner comments.

Everything is passing locally now. Let me know how it looks!

PR #7176, opened 2026-05-28
fix(core): fix storage policy for tables with posting indexes

RaphDal:
## Summary

Tables with a posting-indexed symbol column (`TYPE POSTING`, `TYPE POSTING DELTA`, `TYPE POSTING EF`) could not be used in storage policy after the switch to parquet.

## Root cause

`TableWriter.linkPartitionIndexFiles`, called from `switchNativePartitionWithParquet`, only hard-linked the legacy bitmap layout (`.k` / `.v`). A posting-indexed column has no `.v`: its value data lives in `.pv.<colTxn>.<sealTxn>` plus the `.pci` sidecar and per-column `.pc<N>.{colTxn}.{sealTxn}` data files. The link against the non-existent `.v` returned ENOENT and the commit aborted before `txWriter.commit`, leaving the partition stuck partway through the switch on every retry.

## Changes

A new private helper `linkColumnIndexFiles(srcDirLen, dstDirLen, columnName, columnNameTxn, indexType)` owns the per-column hard-link step:

- Resolve `sealTxn` for posting columns from the live chain head in `.pk` via `PostingIndexUtils.readLiveSealTxnFromKeyFileOrThrow`, falling back to `0` for the pre-seal empty-chain case where the live data still sits in `.pv.<colTxn>.0`.
- Route the key/value-file paths through `IndexFactory.keyFileName/valueFileName(indexType, ...)`, which resolve to `.k`/`.v` for bitmap (the `sealTxn` argument is ignored) and `.pk`/`.pv.<colTxn>.<sealTxn>` for posting.
- For posting columns, also call `linkPostingIndexAuxFiles` to hard-link the `.pci` sidecar and the per-column `.pc<N>.{colTxn}.{sealTxn}` data files for the live generation.

Both pre-existing callers now delegate to the helper, so the bitmap/posting dispatch lives in exactly one place:

- `copyOrRebuildColumnIndexes`'s no-column-top branch collapses to a single call. It previously inlined the same dispatch.
- `linkPartitionIndexFiles` collapses to a `colTop == -1` skip plus a single call. The inner `rmdir` rollback also goes away — `switchNativePartitionWithParquet`'s outer `catch (Throwable)` at the call site already removes the new partition dir on any throw.

## Test plan

- New regression test `TableWriterTest#testSwitchNativePartitionWithParquetLinksPostingIndexFiles` drives `switchNativePartitionWithParquet` on a partition with a `TYPE POSTING` symbol column and asserts the call returns `SWITCH_OK` and `TxWriter.isPartitionParquet` flips. Reverting the production change reproduces the original failure (`[2] could not hard link index value file [table=posting_switch, column=sym]`); with the fix the test passes.
- `mvnd -pl core -Dtest='TableWriterTest#testSwitchNativePartitionWithParquet*' test` — 4/4 pass, including the three pre-existing tests covering the non-indexed and active-partition paths.
- `mvnd -pl core -Dtest='AlterTableConvertPartitionTest' test` — 38/38 pass, covering the other consumer (`copyOrRebuildColumnIndexes`) that now also routes through the shared helper.
- `mvnd -pl core -Dtest='PostingIndex*Test,ParquetTest,TimeFrameCursorTest' test` — 334/334 pass (2 pre-existing skips), broader sanity over posting + parquet paths.


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 23 / 28 (82.14%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|23|28|82.14%|

bluestreak01 (approved):

PR #7179, opened 2026-05-28
chore(core): fix wrong symbol values on replicas after a column type conversion - fuzz test fix

glasstiger:
## What

`ApplyWal2TableJob.calculateSkipTransactionCount()` decides which WAL transactions can be skipped because a later `REPLACE_RANGE` commit fully covers their time range. The forward scan already treated a `SQL` transaction as a barrier (an `UPDATE` may read existing data), but it walked past structural metadata changes. As a result an `INSERT` could be skipped even when a column type conversion sat between it and the covering replace.

A `STRING`/`VARCHAR` to `SYMBOL` conversion builds the column's symbol map from the rows present when it runs. If the insert before it is skipped, the conversion sees fewer rows and assigns symbol keys differently. Within a single instance this stays self-consistent, so query results are unaffected. But the skip decision depends on which later transactions are already visible when a transaction is applied, so two instances can legitimately skip different sets - and under WAL replication a replica that skips the insert (while the primary applied it before the replace existed) builds a divergent symbol map, after which rows resolve to the wrong symbol value.

## Change

Two parts to the skip look-ahead:

1. **Any non-data transaction is now a barrier.** `SQL` was already one; now a structural change such as a column type conversion (`walId < 1`) and every other non-data WAL transaction (`MAT_VIEW_INVALIDATE`, `VIEW_DEFINITION`, ...) is one too. The scan stops at the first such barrier, so transactions before it are no longer skipped and every instance converts the same set of rows.

2. **Materialized views are exempt from that barrier.** A column type conversion cannot run on a materialized view, so the cross-instance divergence the barrier guards against cannot arise there; meanwhile a full mat view refresh truncates, and materialising data only to immediately overwrite it is wasted work - and mat view refresh is the main production user of `REPLACE_RANGE`. So the table token is now threaded into the look-ahead, and for a mat view the scan keeps the original behaviour: it records the barrier and scans on, so a later `TRUNCATE` or covering `REPLACE_RANGE` can still skip the data before it. An `SQL` transaction stays a hard barrier even for a mat view, as it may read existing data. Regular tables are unaffected by part 2.

The `firstNonSkippableTxn` bookkeeping is kept, but its role narrows. For a regular table the scan now breaks at the first barrier before it can ever reach a later `TRUNCATE`, so `firstNonSkippableTxn` stays at its initial `MAX` and the truncate path skips straight to the truncate. It still does real work on the mat-view path, where the scan walks past barriers: the `TRUNCATE` return clamps `min(firstNonSkippableTxn, truncateTxn)` so a recorded barrier is never skipped past.

Supporting refactor in `WalTxnDetails`: a new `getWalTxnTypeAndFlags()` accessor plus static `walTxnTypeOf()` / `dedupModeOf()` decoders let the look-ahead read the packed type+flags slot once per future transaction and decode both the txn type and the dedup mode from it, instead of fetching the same slot twice. `getWalTxnType()` and `getDedupMode()` now delegate to the same decoders - no behaviour change.

## Trade-off

For a regular table the optimization is now disabled across every non-data transaction, not only column conversions. When a `REPLACE_RANGE` fully covers an insert that is separated from it by a structural change (for example an `ADD COLUMN`), the insert is now materialized and then overwritten rather than skipped, which is extra apply work in that case. The common case the optimization targets - a run of data inserts followed by a covering replace with no structural change in between - is unchanged, and materialized views (the main `REPLACE_RANGE` user) keep the optimization in full. Applying instead of skipping is always correct, so this is a performance trade-off, not a change to query results.

A narrower fix limited to conversions-to-`SYMBOL` is possible, but it requires threading the structural-change details into the look-ahead (the scan still has only the coarse transaction type plus, now, the table token) and would silently re-regress if another data-dependent structural operation were added later.

## Test plan

Added twelve tests to `WalWriterReplaceRangeTest`. The ten apply-path tests each commit the insert + the intervening transaction(s) + the covering operation *before* applying, so the apply job sees the later operation while deciding whether to skip the insert (and the skip is decided with an empty writer LAG, which is what `trySkipWalTransactions()` requires); the remaining two run no apply path and pin the transaction-type classification the look-ahead depends on:

**Strict regression guards** (regular table; single-instance observable via the append-only symbol map, so they actually fail when the fix is reverted):
- `testReplaceRangeDoesNotSkipInsertBeforeSymbolConversion` - insert + `STRING`->`SYMBOL` conversion + covering `REPLACE_RANGE`, then asserts via the symbol map that the conversion processed the inserted rows. Fails with `expected:<20> but was:<10>` when the fix is reverted, passes with it.
- `testReplaceRangeDoesNotSkipInsertBeforeVarcharToSymbolConversion` - the same strict guard for a `VARCHAR`->`SYMBOL` conversion (same `20` vs `10` symbol-count proxy).

**Barrier path coverage** (regular table; a single instance is self-consistent whether or not the insert is skipped - the covering operation overwrites the rows either way - so these pass with and without the fix and only pin that the barrier path applies cleanly and returns correct data):
- `testReplaceRangeDoesNotSkipInsertBeforeColumnAdd` - an `ADD COLUMN` (structural, `walId < 1`) barrier.
- `testReplaceRangeDoesNotSkipInsertBeforeMatViewInvalidate` - a `MAT_VIEW_INVALIDATE` barrier (non-data, `walId > 0`): the `!isDataType` arm whose behaviour the fix changed (`VIEW_DEFINITION` shares this arm). Driven via `WalWriter.resetMatViewState`, which emits a synthetic invalidate without needing a real mat view.
- `testReplaceRangeDoesNotSkipInsertBeforeTruncateAfterBarrier` - a barrier between the insert and a `TRUNCATE`, the interaction the `firstNonSkippableTxn` clamp governs (the regular-table scan breaks at the barrier before reaching the truncate).

**Materialized-view exemption** (real `MATERIALIZED VIEW`s; the covering operation overwrites/wipes the rows so the rendered data is identical either way - but each test except the first pins the skip *decision* directly via a physically-written-rows delta, so it fails if the exemption regresses):
- `testReplaceRangeSkipsInsertBeforeBarrierOnMatView` - the insert *is* skipped across a `MAT_VIEW_INVALIDATE` soft barrier via the covering `REPLACE_RANGE` route (the opposite of the regular-table guards). Pinned via a physically-written-rows delta of `10` (only the replace's rows reach disk); the delta would be `20` if the `isMatView()` exemption regressed and applied the insert.
- `testReplaceRangeSkipsInsertBeforeTruncateAfterBarrierOnMatView` - the same exemption via the `TRUNCATE` skip-clamp route (`INSERT -> barrier -> TRUNCATE`, returning `min(barrier, truncate) = barrier`, so the insert is skipped but the invalidate itself is still applied). Also pinned via physically-written rows.
- `testReplaceRangeSkipsInsertBeforeMultipleBarriersAndTruncateOnMatView` - the stacked-barrier variant (`INSERT -> MAT_VIEW_INVALIDATE -> MAT_VIEW_INVALIDATE -> TRUNCATE`), which exercises the `firstNonSkippableTxn = Math.min(...)` accumulation across more than one barrier: the skip is clamped to the *first* barrier, so neither invalidate is skipped past. Again pinned via physically-written rows.
- `testReplaceRangeDoesNotSkipInsertBeforeSqlBarrierOnMatView` - the complementary case: an `SQL` transaction stays a *hard* barrier even on a mat view (the `futureType != SQL` carve-out), so the insert is *not* skipped. Uses `ADD INDEX` - the realistic mat-view `SQL` barrier: non-data and non-structural, so it commits as a `walId > 0` `SQL` transaction and builds row-derived index state. Pins that the arm applies cleanly, returns correct data, and leaves a usable index.
- `testReplaceRangeDoesNotSkipInsertBeforeSqlBarrierAndTruncateOnMatView` - the truncate-clamp counterpart of the previous test: an `SQL` barrier (`ADD INDEX`) followed by a `TRUNCATE`, confirming `SQL` stays a hard barrier even when a `TRUNCATE` follows it (the scan breaks at the `SQL` barrier and never reaches the truncate clamp). Pinned strictly via a physically-written-rows delta of `11` (the insert's 10 rows are materialised then truncated, plus 1 surviving row); the delta would be `1` if the carve-out regressed and skipped the insert.

**Classification guards** (no apply path; assert the transaction-type invariants the barrier and the mat-view exemption depend on):
- `testMatViewPermittedColumnAltersStayNonStructural` - the exemption is safe only because no row-order-dependent structural change can reach a mat view. This pins that the column alters a mat view is permitted to run (`ADD INDEX`, `DROP INDEX`, `SYMBOL CAPACITY`) are non-structural - so they commit as `walId > 0` `SQL` hard barriers, not exempted soft barriers - while a column type conversion is structural (and is separately rejected on a mat view). If one of those alters were ever made structural, it would silently become an exempted soft barrier on a mat view and could reopen the cross-instance divergence; this test fails the moment that assumption breaks. The `calculateSkipTransactionCount()` comment references it by name.
- `testNonDataWalTxnTypesStopTheSkipScan` - pins the `WalTxnType.isDataType()` classification the `!isDataType(futureType)` barrier rests on: only `DATA` and `MAT_VIEW_DATA` are data types; `SQL`, `TRUNCATE`, `MAT_VIEW_INVALIDATE`, `VIEW_DEFINITION` and `NONE` are all barriers. This is the only coverage for `VIEW_DEFINITION` (type 5), which has no `WalWriter` entry point to commit in a test. If a new type were added and misclassified as data, an insert before it could be skipped and reopen the divergence.

Ran the WAL apply, replace-range and materialized-view suites (`WalWriterReplaceRangeTest`, `MatViewTest`, `CreateMatViewTest`, `MatViewIdenticalReplaceTest`, `WalTxnDetailsFuzzTest`, `WalAlterTableSqlTest`, `ReplaceInsertFuzzTest`, `DedupInsertFuzzTest`) - all green. Materialized-view refresh is the main production user of `REPLACE_RANGE` and shows no regression.

Required by https://github.com/questdb/questdb-enterprise/pull/1034


bluestreak01:
## Code review (mission-critical pass)

`origin/master` already contains the basic fix (the skip scan stops at any non-data transaction, so an `INSERT` covered by a later `REPLACE_RANGE` is no longer skipped across a `STRING/VARCHAR→SYMBOL` conversion). This PR is the follow-up that (1) **exempts materialized views** from that barrier — for a mat view the inner scan records `firstNonSkippableTxn` and continues past non-data barriers (except `SQL`, kept hard), with the `TRUNCATE` early-return clamped to `min(firstNonSkippableTxn, futureSeqTxn)` so a barrier is never skipped past; (2) **refactors `WalTxnDetails`** to read the packed type+flags slot once; (3) adds 12 tests.

### Critical
**None.** No correctness, concurrency, or resource blocker survived verification.

The core safety argument checks out: the mat-view exemption is safe because the only apply-window-dependent persistent state on a mat view is the `STRING/VARCHAR→SYMBOL` conversion, which is **doubly blocked** — `compileAlterTable` rejects every `ALTER TABLE <matview>` via `checkMatViewModification` (`SqlCompilerImpl.java:2293`), and the `ALTER MATERIALIZED VIEW` grammar admits only `SYMBOL CAPACITY`/`ADD INDEX`/`DROP INDEX`, all non-structural (`AlterOperation.isStructural()` → `walId>0` `SQL` = hard barrier). The refactor is bit-for-bit behavior-preserving, and regular-table behavior is provably identical to master (`firstNonSkippableTxn` stays `MAX`, so `min(MAX, futureSeqTxn) == futureSeqTxn`).

### Moderate
1. **PR type should be `fix`, not `chore`.** The body and diff implement a data-correctness defect ("rows resolve to the wrong symbol value" on replicas). The description is copied to release notes, and `chore` commits are typically excluded from the changelog — a user-facing replication-correctness fix would silently vanish. The title also self-contradicts: "fix wrong symbol values" cannot honestly be a `chore`.
2. **Labels insufficient.** Only `tandem` is set. Missing `WAL`, `Bug`, `Materialized View`, `Core`. Confirm whether `regression` also applies.
3. **No test asserts the safety hinge directly.** The whole exemption is safe only because a column type conversion cannot reach a mat view, yet no test asserts that rejection. `testMatViewPermittedColumnAltersStayNonStructural` asserts conversion-*is-structural* on a **regular** table — it never exercises a mat view rejecting the conversion. A one-liner closes the gap:
   ```java
   assertException("alter materialized view mv alter column s type symbol", /*pos*/, "cannot modify materialized view");
   ```

### Minor
1. **Inaccurate comment** at `ApplyWal2TableJob.java:140-141`: "NONE for structural (walId < 1)". `walTxnTypeOf(0L)` decodes to `DATA` (0), not `NONE` (-1). Harmless (the `walId < 1 ||` short-circuit means it's never evaluated), but the wording is wrong — it was borrowed from the inner loop where `: NONE` is explicitly assigned.
2. **Redundant `getWalId` read in the hottest loop** (perf, not a blocker). Structural txns store `NONE` in the type slot (`WalTxnDetails.java:764`), so the inner-loop `futureWalId < 1 || !isDataType(futureType)` is equivalent to `!isDataType(futureType)` if the type slot is read unconditionally — letting the `getWalId(futureSeqTxn)` load (line 164, a different cache line) be dropped, halving array loads over a lookahead window of ~200+ txns. Reasonable to defer.
3. **Mat-view exemption raises the inner-scan worst case to O(K·N)** (was bounded by distance-to-first-barrier). Bounded by the in-memory lookahead window (all array reads, no I/O) and trades cheap reads for avoiding physical writes the truncate/replace discards — favorable for the targeted refresh pattern; worth a one-line note in the comment.
4. **Test-comment over-crediting.** The regular-table strict guards (`...BeforeSymbolConversion`, `...BeforeVarcharToSymbolConversion`, `...BeforeMatViewInvalidate`) actually guard the pre-existing master barrier + refactor preservation, not this PR's net delta. Only the mat-view trio (`testReplaceRangeSkipsInsertBefore{Barrier,TruncateAfterBarrier,MultipleBarriersAndTruncate}OnMatView`) fails on reverting *this* PR.
5. **Disclosed coverage gaps**: `VIEW_DEFINITION` soft-barrier-on-mat-view is pinned only by classification, never driven e2e; a multi-soft-barrier-then-`REPLACE_RANGE` (not `TRUNCATE`) mat-view case is untested (the `REPLACE_RANGE` skip path doesn't consult `firstNonSkippableTxn`); an `SQL` barrier on a regular table isn't directly exercised. All low-risk/path-equivalent and honestly disclosed.
6. **Style nits**: the two `"s\nw0\n...\nw9\n"` expected results should use text blocks; `getWalTxnTypeAndFlags` is `public` though only used in-package.
7. **Global-metric proxy** `getPhysicallyWrittenRows()` is deterministic here (single-table, synchronous `drainWalQueue`, base never inserted so no refresh writes) but latently fragile if copied into a multi-table context — a one-line note on that assumption would help.

### Summary
**Approve** after the metadata fixes and ideally the one-line safety-hinge test. 10 findings verified, 0 false positives; all in-diff (cross-context caller inventory confirmed every refactored accessor byte-for-byte equivalent, and the only new behavior is gated behind `isMatView` with a single caller). Tradeoffs are fairly stated in the description. The mat-view exemption's safety holds against current code but rests on a forward-looking invariant (no future row-dependent op may be made structural *and* allowed on a mat view) — documented in-code and tripwired by `testMatViewPermittedColumnAltersStayNonStructural`.

<sub>🤖 mission-critical review pass</sub>

bluestreak01 (dismissed):

bluestreak01 (dismissed):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 22 / 22 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/wal/WalTxnDetails.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/ApplyWal2TableJob.java|17|17|100.00%|

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 27 / 30 (90.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/ConcurrentLongHashMap.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/std/ConcurrentIntHashMap.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/std/ConcurrentHashMap.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalTxnDetails.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/ApplyWal2TableJob.java|17|17|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlException.java|5|5|100.00%|

PR #7182, opened 2026-05-29
fix(sql): return the last N rows for LIMIT -N over a covering index

### LINKED ISSUE #7181: LIMIT -N over a posting/covering index with a residual filter returns the first N rows instead of the last N
## Summary

`SELECT ... LIMIT -N` over a covering (posting) index with a residual filter returns the **first N rows instead of the last N**. The result is silently wrong — no error, no warning. The covering factory ignores the requested scan order when serving a page-frame cursor, so the parallel negative-limit machinery collects the lowest-timestamp rows while believing they are the highest.

## Reproducer

```sql
CREATE TABLE t_m2 (ts TIMESTAMP, sym SYMBOL, price DOUBLE) TIMESTAMP(ts) PARTITION BY DAY BYPASS WAL;

INSERT INTO t_m2
SELECT dateadd('d', x::INT, '2024-01-01T00:00:00Z'::TIMESTAMP), 'A', x::DOUBLE
FROM long_sequence(10);   -- one row per DAY => 10 partitions => 10 page frames at default frame size

ALTER TABLE t_m2 ALTER COLUMN sym ADD INDEX TYPE POSTING INCLUDE (price);

SELECT price FROM t_m2 WHERE sym = 'A' AND price > 0 LIMIT -3;
```

- Expected (last 3 by ts): `8.0`, `9.0`, `10.0`
- Actual (observed): `3.0`, `2.0`, `1.0`  — the first/lowest-ts rows

Captured plan:

```
SelectedRecord
    Async Filter (limit: 3, filter: 0<price)
        CoveringIndex on: sym with: price filter: sym='A'
```

The residual `price > 0` is what routes the query through the parallel filter wrapper. The limit is absorbed into the `Async Filter` node (no separate `Limit` node), which is the heart of the bug.

## Affected query shape

All of the following must hold:

- a covering (posting) index that can answer the query from the index (`... ADD INDEX TYPE POSTING INCLUDE (...)`),
- a residual filter on top of the index predicate (so the query goes through `wrapCoveringWithFilter`),
- parallel filtering enabled (`isParallelFilterEnabled()` — the default),
- a **negative** single-arg limit (`LIMIT -N`), or a **bind-variable** limit whose sign is unknown at compile time.

`LIMIT -N` is the only broken consumer. `ORDER BY ts DESC` over a covering index is **not** affected — the covering factory advertises `SCAN_DIRECTION_FORWARD`, so DESC is never elided onto it (it gets a real `Sort` over the genuinely ascending cursor). Vectorized GROUP BY and SAMPLE BY FIRST/LAST over a covering index always request `ORDER_ASC`, so they are unaffected.

## Root cause

`CoveringIndexRecordCursorFactory.getPageFrameCursor(SqlExecutionContext, int order)` takes an `order` parameter but **ignores it**, always opening `ORDER_ASC` partition frames:

```java
// CoveringIndexRecordCursorFactory.java
public PageFrameCursor getPageFrameCursor(SqlExecutionContext executionContext, int order) throws SqlException {
    ...
    PartitionFrameCursor frameCursor = dfcFactory.getCursor(
            executionContext,
            columnIndexes,
            PartitionFrameCursorFactory.ORDER_ASC   // <-- hardcoded; `order` unused
    );
    ...
}
```

while `getScanDirection()` advertises `SCAN_DIRECTION_FORWARD`.

The dispatch that exploits this (all in `master`, none introduced recently):

1. `SqlCodeGenerator.wrapCoveringWithFilter` wraps the covering factory in `AsyncFilteredRecordCursorFactory` with the limit function.
2. `AsyncFilteredRecordCursorFactory.implementsLimit()` returns `limitLoFunction != null` (true), so `generateLimit` does not add an outer `LimitRecordCursorFactory` — the async factory's internal negative-limit handling is final.
3. For `LIMIT -N`, `AsyncFilteredRecordCursorFactory.getCursor` computes `order = reverse(baseOrder) = ORDER_DESC` and routes through `AsyncFilteredNegativeLimitRecordCursor`.
4. `PageFrameSequence` passes that `ORDER_DESC` to `base.getPageFrameCursor(ctx, ORDER_DESC)` — which the covering factory ignores, returning ascending frames.
5. `AsyncFilteredNegativeLimitRecordCursor` was built with `hasDescendingOrder = (base.getScanDirection() == BACKWARD) = false`, so it takes the first-N-encountered rows expecting the highest timestamps. Fed ascending frames, it returns the first N instead of the last N.

## Pre-existing

This reproduces on `master`: `getPageFrameCursor` hardcodes `ORDER_ASC` there too, the covering factory already returns `true` from `supportsPageFrameCursor()`, and it does not override `getScanDirection()` (inheriting the `FORWARD` default). The negative-limit machinery (`AsyncFilteredRecordCursorFactory`, `AsyncFilteredNegativeLimitRecordCursor`, `PageFrameSequence`) is long-standing. The covering/posting index feature landed in #6861.

## Possible fixes

Two correct approaches (details worked out during the PR #7150 review):

- **Approach A (proper):** honor `order` — give the covering factory a real backward page-frame scan: iterate partitions DESC, emit the high row-range sub-frame first within a partition (mirroring `BwdTableReaderPageFrameCursor.computeNativeFrame`), and drain the multi-key path via a max-heap. `getScanDirection()` stays `FORWARD` (the per-call `order` controls emission). Preserves parallel filtering for negative-limit covering queries; moderate effort and real correctness risk in the sub-frame/row/partition interplay.

  > The naive "pass `order` to `dfcFactory.getCursor`" one-liner is **wrong**: it reverses partition order but leaves sub-frames ascending within a partition, returning wrong results for any partition with more than `maxRowsPerFrame` (default 1,000,000) matching rows.

- **Approach B (codegen guard, lower risk):** in `wrapCoveringWithFilter`, when the limit may be negative (`!isConstant()` — bind var, sign unknown — or `getLong(null) < 0`), skip the parallel path and fall back to `FilteredRecordCursorFactory`. Its `implementsLimit()` is the default `false`, so `generateLimit` wraps it in a `LimitRecordCursorFactory`, which computes last-N via size + skip (no random access needed). Forfeits parallel filtering for the narrow negative/bind-var-limit + covering + residual-filter shape; no new cursor code.

Either way, do **not** change `getScanDirection()` to `BACKWARD` — that would wrongly elide `ORDER BY ts DESC` onto the ascending cursor and create a new wrong-results bug.

## Suggested regression test

The reproducer above, plus a multi-row-per-partition variant (`setMaxRowsPerFrameForTesting(2)` or several rows per day) so a partition-DESC-only (still-wrong) fix would also fail, not just the single-frame-per-partition case.

bluestreak01:
Fixes #7181

## Summary

`SELECT ... LIMIT -N` over a covering (posting) index with a residual filter returned the **first N rows instead of the last N**. The result was silently wrong — no error, no warning. This affected single-key queries (`WHERE sym = 'A' AND price > 0 LIMIT -3`) and the same shape with a multi-key (`sym IN (...)`) predicate or a bind-variable limit.

The covering factory ignored the requested scan order when serving a page-frame cursor: `getPageFrameCursor(executionContext, order)` always opened ascending partition frames, while `getScanDirection()` advertised `FORWARD`. The parallel negative-limit machinery then collected the lowest-timestamp rows believing they were the highest.

## Fix

The two query shapes need different handling because they have different ordering guarantees.

**Single-key** covering queries are globally timestamp-ordered (partitions ascending, rows ascending within a partition), so a backward scan yields the true last N rows. `getPageFrameCursor` now honors `ORDER_DESC`:

- iterate partitions latest-to-earliest, and
- within each partition, split the row range into sub-frames at most `maxRowsPerFrame` wide and emit the highest row-range sub-frame first, each read forward so rows stay ascending *within* a frame.

This mirrors `BwdTableReaderPageFrameCursor` and stays correct when a partition holds more than `maxRowsPerFrame` matching rows — a case where a partition-reverse-only fix would still return wrong results. Parallel filtering is preserved for this shape.

**Multi-key** covering is only per-key-per-partition ordered, never globally timestamp-ordered, so a correct last-N-by-timestamp cannot be produced from it without a k-way merge. When the limit may be negative (a constant below zero, or a bind variable whose sign is unknown at compile time), codegen routes multi-key queries to the serial `FilteredRecordCursorFactory`, where `LimitRecordCursorFactory` computes last-N via size plus skip. A defensive guard in `getPageFrameCursor` rejects a backward scan on the multi-key path.

`getScanDirection()` stays `FORWARD` so `ORDER BY ts DESC` is not wrongly elided onto the ascending cursor.

## Tradeoffs

- Single-key negative-limit covering queries keep parallel filtering and now read fewer rows for small `N` (early termination from the high end), instead of scanning ascending and discarding.
- Multi-key (and bind-variable) negative-limit covering queries forfeit parallel filtering for this narrow shape: they fall back to the serial filter plus a `Limit` node. The bind-variable case takes the serial path even when the value turns out positive at runtime, because the sign is unknown at compile time. This is a correctness-for-throughput trade on an otherwise silently-wrong path; positive-limit and no-limit multi-key queries are unaffected and still run in parallel.

## Test plan

- New regression tests in `CoveringIndexTest`:
  - the issue reproducer (10 single-row partitions, `LIMIT -3` returns `8,9,10`), including a filter that eliminates the low rows and a larger limit spanning all partitions
  - a multi-frame-per-partition variant with `setMaxRowsPerFrameForTesting(2)`, which a partition-reverse-only fix would still fail, across `-3/-4/-5` to cross sub-frame boundaries
  - a bind-variable limit on a single-key query, asserting last-N for a negative value and head-N for a positive value on the same compiled statement
  - a multi-key query asserting the plan does not use the parallel `Async Filter` path and returns the correct tail, while a positive multi-key limit still does
- Confirmed the single-key tests fail on the unpatched code and pass with the fix.
- Full `CoveringIndexTest` (378), `PostingIndexCriticalIssuesTest` (40), and `CoveringCompressorTest` (17) pass.

## New Features
- Added support for negative LIMIT values on covering index queries, enabling efficient retrieval of the last N rows

## Bug Fixes
- Fixed row ordering correctness for multi-key covering index queries to maintain timestamp order across partitions
- Improved LIMIT pushdown behavior on covering indexes to prevent incorrect results when ORDER BY clauses are present

## Tests
- Added comprehensive test suite verifying covering index query correctness across multiple scenarios including filters, projections, and LIMIT variations


bluestreak01:
## Code review

Reviewed the full change surface, not just the diff. The core fix is correct and well-engineered: the descending scan replicates `BwdTableReaderPageFrameCursor.computeNativeFrame` exactly (the `adjustedLo = Math.max(lo, hi - maxRows)` split), the negative-limit contract holds (`AsyncFilteredNegativeLimitRecordCursor` with `hasDescendingOrder=false` for the FORWARD-direction covering factory), all 19 new tests + 378 existing `CoveringIndexTest` pass, and the risky shapes cross-check against the `no_covering` oracle. No correctness bug, no leak in new code, no concurrency hazard. The items below are real but none block.

**Verdict: Approve with minor changes.**

### Critical
None.

### Moderate

**M1 — PR description overstates the performance benefit.** The Tradeoffs bullet says single-key negative-limit queries "now read fewer rows for small `N` (early termination from the high end)." This is not delivered. The parallel negative-limit path (`AsyncFilteredNegativeLimitRecordCursor.fetchAllFrames` → `prepareForDispatch()` → `PageFrameSequence.buildAddressCache()`) drains and buffers **every frame of every partition** before the `rowCount >= rowLimit` cancel can fire; `cancel()` only short-circuits the residual-filter reduce stage, not frame production or buffer allocation. This is a pre-existing, universal property of the parallel negative-limit machinery (the cursor serves all filtered bases, not just covering), so it's not a regression — but the claimed benefit doesn't exist. The descending scan delivers correctness (and saves some residual-filter CPU because high-timestamp frames reduce first), not fewer rows read. Suggest rewording the bullet.

**M2 — Multi-key k-way merge uses an O(N)/row linear min-scan; the "mirrors HeapRowCursorFactory" comment is inaccurate.** `MultiKeyCoveringCursor.hasNext` and `MultiKeyCoveringPageFrameCursor.fillMergedFrame` pick the smallest head with an unconditional linear scan over all N keys per emitted row. The comment claims this mirrors the non-covering `HeapRowCursor`, but that path uses `IntLongSortedList` — O(N) worst case but **adaptive** (`pollAndReplace` shifts only `p` elements, ≈O(1) when one key dominates), whereas the linear scan is always O(N). A true binary min-heap would be O(log N). For realistic IN-lists (2–20 values) the linear scan is fine and likely wins on cache locality; for large generated lists (hundreds) over dense data it's a real per-row cost. Not a regression (the merge is *required* for correctness — the old key-grouped order was wrong) and not blocking. Either switch to `IntLongSortedList` to match the established pattern, or correct the comment to state the actual tradeoff.

**M3 — Test coverage gaps on the new paths.** Two worth closing:
- *Variable-width covered columns + negative limit / multi-key merge.* Every negative-limit test uses `price DOUBLE` only; the VARCHAR/STRING/BINARY/ARRAY IN-list tests have no limit and no residual filter. `finalizeFrame`/`writeCoveredRow` serve all types and the write code is shared with the tested forward path (so risk is moderate), but the descending sub-frame split + `growFrameBuffers` relocating var-data buffers is an untested interaction. Add a covered-VARCHAR (or ARRAY) `LIMIT -N` test and a multi-key var-width cross-check with a filter.
- *Sub-frame cap × filter-drop × negative limit.* No test combines `setMaxRowsPerFrameForTesting(2)` + a residual filter that drops rows + a negative limit. That's exactly where an off-by-one in the descending accumulation would hide (e.g. cap=2, single YEAR partition, `price > 5 LIMIT -3` so the top sub-frame survives and a lower one is partly dropped).

**M4 — `mayBeNegativeLimit` is misplaced in alphabetical member order.** `SqlCodeGenerator.java:11337` wedges it between `validateSubQueryColumnAndGetGetter` and `validateTimestampNotInJoinKeys`. The surrounding private instance methods are strictly alphabetical, so it should move between `lookupColumnIndexesUsingVanillaNames` (11088) and `prepareLatestByColumnIndexes` (11099).

### Minor

**m1 — `maxRowsPerFrame == 0` hangs the descending scan.** In `nextImplDescending`, `subLo = Math.max(descPartitionLo, descSubHi - maxRowsPerFrame)`; with `maxRowsPerFrame==0`, `subLo == descSubHi`, so `descSubHi` never decreases and the loop spins forever. Production-unreachable (`PropServerConfiguration.validatePageFrameRows` rejects `<1`), but the test-only `setMaxRowsPerFrameForTesting(0)` hook would hang CI rather than fail. Cheap guard: `long step = Math.max(1, maxRowsPerFrame);`.

**m2 — Pre-existing error-path leak in `wrapCoveringWithFilter`.** If `compileWorkerFiltersConditionally(...)` (compiles JIT filters, can throw) or `deepClone(...)` throws before the `AsyncFilteredRecordCursorFactory` constructor takes ownership, `coveringFactory` + `filter` (+ limit function) leak — the only catch (`~10548`) frees only `dfcFactory`. Pre-existing, not a regression. The new `mayBeNegativeLimit → init()` adds one more throw point but it's essentially unreachable (init runs only for constant-folded limits, a no-op). Since this PR already carries an "error path leak fix" commit, consider wrapping the two callsites (10297, 10397) in a `try/catch` that frees `coveringFactory` + `filter` on throw.

**m3 — Stale comment.** `CoveringIndexRecordCursorFactory.java:~213` says an empty `multiKeys` list "makes `advanceKey()` report no rows" — but `advanceKey()` now throws `UnsupportedOperationException`; the empty-list short-circuit lives in `hasNext()` → `openNextPartitionCursors()`. Reword.

**m4 — Boolean field naming.** `protected boolean descending` should be `isDescending` per the `is…`/`has…` rule — it sits directly above `isExhausted`, which uses the prefix.

**m5 — Duplicated merge logic.** The min-head selection loop and the park-before-probe partition-open loop are near-identical in `MultiKeyCoveringCursor` and `MultiKeyCoveringPageFrameCursor` (only `openForwardCoveringCursor` is shared). A future fix to the merge invariant must touch both. Consider a shared helper.

**m6 — `fillMergedFrame` double-pass (micro).** The `anyHead` pre-scan over N keys duplicates the first iteration of the per-row min-scan. Defer `allocFrameBuffers()` until after the first `best >= 0` selection to fuse the two passes and drop the "allocate-then-discard" rationale.

**m7 — Test convention.** The explicit-value tests (`testSinglePartitionInterleavedExplicit`, `testSingleKeyOrderByTsDescLimitExplicit`, `testReferenceOracleIsTimestampOrdered`) use a custom `assertResult`/`printSql` instead of `assertQueryNoLeakCheck`, skipping the factory-property assertions (`expectSize`/`expectedTimestamp`) that matter for a LIMIT fix. The oracle cross-check method itself is correctly custom (it compares two queries and asserts the plan is genuinely covering).

**m8 — Labels.** Consider adding **Performance** (k-way merge + buffer-alloc commit) and **regression** (9.4.0 shipped `getPageFrameCursor` hardcoding `ORDER_ASC` — a user-facing correctness regression vs the plain-index path).

### Verified safe (checked and cleared)
- **Concurrency:** the reused `frame` object is safe — `buildAddressCache`/`PageFrameAddressCache.add` snapshots all frame addresses into off-heap lists on the single producer thread before any worker runs; covering frames are always `NATIVE` so no lazy re-read. Producer-only merge state.
- **Multi-key DESC `CairoException.critical` guard:** unreachable defensive guard — `getScanDirection()==FORWARD` blocks `ORDER BY ts DESC` elision and `recordCursorSupportsRandomAccess()==false` blocks top-K; multi-key negative limits route to serial.
- **k-way merge duplicate/tie on shared row ids:** cannot happen — `WhereClauseParser` dedups IN-list strings and the symbol map is a bijection, so distinct keys never share a row id (verified with `sym IN ('A','A')`).
- **Multi-key serial fallback last-N correctness, deferred-symbol resolution, off-by-one bounds, latestBy interaction, generateLimit double/dropped-limit:** all traced and correct.

### Summary
Merge after addressing the description accuracy (M1), member ordering (M4), and ideally the var-width / filter-drop negative-limit tests (M3). The performance items (M2) and minors are non-blocking. The multi-key output order now matches the plain-index path (timestamp-ordered), which is why 7 existing assertions were updated — a correct, user-visible consistency improvement.

puzpuzpuz:
> **Multi-key** covering is only per-key-per-partition ordered, never globally timestamp-ordered, so a correct last-N-by-timestamp cannot be produced from it without a k-way merge.

Heads-up: k-way merge is introduced in #7150 (there is a separate bug which requires k-way merge similar to FilterOnValues cursor's one).

puzpuzpuz (dismissed):

puzpuzpuz:
Two moderate items I'd like to highlight. They're not blockers and are more or less essential for the chosen fix, but that's something we could improve in the future patches.

###  M1 — Negative-limit descending path over-reads/over-materializes [in-diff, perf]

SingleKeyCoveringPageFrameCursor.nextImplDescending (CoveringIndexRecordCursorFactory.java:2366) sizes each backward sub-frame to a fixed maxRowsPerFrame (default 1,000,000) window of row-ids and calls fillFrameForKey(..., Integer.MAX_VALUE), which materializes every matching row in that window into native buffers (writeCoveredRow) before the consumer reads anything. AsyncFilteredNegativeLimitRecordCursor.fetchAllFrames (:215) cancels only after the first frame is fully built. So ... LIMIT -5 over a hot key in a large latest partition copies up to ~1M rows to return 5.

Unlike the standard backward cursor, this path ignores pageFrameMinRows/sharedQueryWorkerCount — BwdTableReaderPageFrameCursor:480 sizes frames via calculatePageFrameRowLimit(...). The per-row copy is inherent to covering (no mmap'd column to point at, unlike the zero-copy BwdTableReaderPageFrameCursor), but the fixed 1M window is an avoidable parameter choice. Note: enabling covering can therefore be slower than the plain index for large LIMIT -N. Confirmed independently by Agents 3, 10, 11. Mitigation: thread the limit N down to bound the first sub-frame, and/or adopt calculatePageFrameRowLimit for consistency.

###  M2 — Multi-key negative-limit serial fallback scans ~2–3× [in-diff, perf tradeoff]

  The PR routes multi-key negative/bind-var limits to FilteredRecordCursorFactory + LimitRecordCursorFactory. Because MultiKeyCoveringCursor.size() returns -1 (:1924), LimitRecordCursorFactory.ensureBoundsResolved (LimitRecordCursorFactory.java:279) runs a full calculateSize counting pass, then toTop + skip + take — reconstructing the k-way merge ~2–3 times. The PR explicitly accepts this throughput tradeoff, and it's correct. But SingleKeyCoveringCursor.size() (:2202) already shows the pattern: a cheap MultiKeyCoveringCursor.size()
  summing each key's posting size() (which has an O(genCount) fast count path) would eliminate the redundant counting pass. Worth doing since this path is the deliberate fallback for a common shape. (Agents 3, 11)

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 243 / 256 (94.92%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|40|45|88.89%|
|:large_blue_circle:|io/questdb/griffin/engine/table/CoveringIndexRecordCursorFactory.java|203|211|96.21%|

PR #7183, opened 2026-05-29
chore(core): keep parquet sort metadata consistent across out-of-order merges

RaphDal:
## Summary

When a partition that has been converted to parquet receives out-of-order (O3) data, the merge rewrites the partition by copying the unchanged row groups and appending the new ones. The copied row groups dropped their per-row-group sorting-column metadata while the freshly written ones kept it, so the resulting file declared different sorting columns on different row groups.

QuestDB's query readers tolerate that inconsistency, so it normally goes unnoticed on a running instance. A replica restoring from a backup does not: engine bootstrap runs the `Mig941` migration to generate `_pm` parquet-metadata sidecars, and that path validates per-row-group consistency strictly. It aborted with `sorting columns differ between row groups: rg 0 has 0 but rg 1 has 1`, which crashed the replica during `createReplica()`. The failure surfaced through the enterprise `ReplicationFuzzTest.testFuzzWithBackup` (replica restore from backup is an enterprise feature, so this test is not part of the OSS suite).

This change makes the migration accept the mixed metadata that already exists on disk -- trusting `qdb_meta` over the footer -- and stops the merge from producing it going forward. It also corrects the index that the merge declares for the sort column: that index came from the table's raw timestamp slot, which becomes stale once a column before the timestamp is dropped, so the declared sort column could point at the wrong parquet column (or out of range).

## Root cause

Two write paths disagreed on sorting columns:

- `copy_row_group` / `copy_row_group_with_null_columns` rebuild the copied row groups through `build_raw_row_group`, which hard-coded `sorting_columns: None`. The raw-copy writer pushed that verbatim, so copied groups declared zero sorting columns.
- Freshly written O3 row groups went through the normal write path, which stamps the file's sorting columns, so they kept the designated-timestamp sort column.

A partition that was converted to parquet and then O3-merged therefore ended up with rg 0 (copied) declaring zero sorting columns and rg 1 (fresh) declaring one.

Only replicas tripped over it because the two `_pm`-generation paths differ in strictness. The primary generates `_pm` inline at O3 time from the file-level target sorting columns and never inspects the footer. The replica restored a backup whose parquet files have no `_pm`, so bootstrap ran `Mig941`, which reads the on-disk footer through the strict validator.

A second, independent defect rode along on the same metadata. The sort-column index that the merge stamps came from the table's raw timestamp writer slot (`META_OFFSET_TIMESTAMP_INDEX`). `DROP COLUMN` only marks a column deleted -- it leaves a tombstone in the raw layout and never re-indexes the timestamp slot or decrements the column count -- so once a column before the timestamp is dropped, that slot no longer matches the file's dense column layout. `SortingColumn.column_idx` is a dense positional index, so the merge then declared the timestamp sort column at the wrong position, or out of range when the dense file held fewer columns than the slot. This is independent of the consistency bug: it corrupts the index value itself, so even once every row group declares the same sorting column, that column can still be the wrong one.

## Changes

Fixes 1 and 2 are in the `qdbr` native crate. Fix 3 spans `qdbr`, the vendored `parquet2` crate, and a small read-only addition to the Java `ParquetMetaFileReader`.

Fix 1 -- tolerate the mismatch (`parquet_metadata/convert.rs`). `resolve_sorting_columns` resolves the sort column from the designated timestamp in `qdb_meta` -- which the read path already treats as authoritative -- and consults the footer only for files that have no designated timestamp (e.g. external parquet). For a QuestDB file the footer is not read at all, so it no longer matters what shape the buggy merge left it in: copied groups (no sort column) mixed with fresh ones, a stale index shared by every group, or declaring groups that genuinely conflict (an update-mode merge across this upgrade can leave a cached group's stale index beside a freshly appended corrected one). The migration takes the dense designated position regardless and never aborts -- and, because that position comes from `qdb_meta`, the regenerated `_pm` records the correct dense index even for a legacy file whose footer carries a stale or out-of-range one. This is the load-bearing change: files already written by the buggy merge, on primaries and inside backups, carry the inconsistency on disk and cannot be rewritten retroactively, so the migration has to read them. The footer reader (`extract_sorting_columns`) still skips row groups that declare no sorting columns and still rejects a genuine cross-row-group conflict, but only on the no-`qdb_meta` path: a conflicting footer on a file without a designated timestamp is still rejected.

Fix 2 -- stop producing the mismatch (`parquet_write/update.rs`). `build_raw_row_group` takes the sorting columns as a parameter, and both copy sites pass the updater's file-level target sorting columns instead of `None`. Using one file-level value rather than each copied footer's own value is necessary because ADD/DROP COLUMN can shift the timestamp column index; this matches what the inline `_pm` path already does. Every row group in newly written files -- copied and fresh -- now declares the same sorting columns.

Fix 3 -- declare the correct sort-column index (`parquet_write/update.rs`, plus a setter on the vendored `parquet2` crate). The file-level sorting columns that Fix 2 stamps were themselves derived from the raw timestamp slot, so Fix 2 made the row groups agree on a stale index after a pre-timestamp DROP COLUMN. `ParquetUpdater` now derives the sort column from the dense designated-timestamp position recorded in the file's QuestDB metadata: at construction for the no-schema-change rewrite, and re-stamped in `set_target_schema` from the dense target schema for the ADD/DROP COLUMN rewrite. This mirrors the fresh-encode path and the designated-timestamp position that `end()` already computed densely, so the copy paths now agree with the encoder. The write path reads the sort direction from the designated timestamp's order flag rather than assuming ascending, so it derives the same `SortingColumn` the migration does. A new `ParquetFile::set_sorting_columns` setter lets `set_target_schema` re-stamp the value after it rebuilds the schema. The raw `timestamp_index` threaded elsewhere -- the designated-timestamp flag and the raw-slot merge loops that test `i == timestampIndex` -- is left unchanged, since it is correct as a writer-slot index there; only the derivation of `sorting_columns` changes.

Fix 1 alone resolves the crash. Fix 2 keeps new files internally consistent, and Fix 3 makes that consistent value the correct dense index. Fix 1 cannot be dropped once Fix 2 and Fix 3 ship, because it is what keeps already-written files -- including those carrying the stale index -- readable.

## Tradeoffs and limitations

- Fix 1 leans the migration on `qdb_meta` rather than the footer for QuestDB files. The footer's per-row-group sorting columns are advisory, so the migration adopts `qdb_meta`'s designated timestamp and only falls back to (or, for files without one, rejects via) the footer. This deliberately accepts mixed or conflicting footers that already exist on disk and in backups; it trusts `qdb_meta` over the footer, which is exactly what the read path already does.
- Fix 2 affects newly written files only. Files written by the buggy path before this change stay inconsistent on disk; their readability depends entirely on Fix 1.
- Copied row groups now claim the file-level target sort order without re-checking the copied page bytes against it. This relies on the existing invariant that QuestDB parquet partitions are sorted by the designated timestamp. The inline `_pm` path already assumes the same thing, so this aligns the two paths rather than adding a new assumption -- but it remains an assumption about the data rather than a verified property of the bytes.
- Fix 3 is latent on a running QuestDB instance. The query read path resolves the designated timestamp from the QuestDB metadata flag, never from the footer `sorting_columns`, so the stale index produced no wrong results and no ordering, pruning, or stability impact. The visible surface is external tooling that reads the `.parquet` directly: a metadata inspector would show the wrong sorted column, and a strict footer validator could reject the out-of-range case.
- Fix 3 changes newly written files only. Files written with the stale index before this change keep it on disk; every row group shares that index, so they are internally consistent and read correctly (the index is advisory). An update-mode merge after upgrading can append a corrected index beside the stale one, leaving a mixed footer -- the migration tolerates that via `qdb_meta` (Fix 1), so still no data migration is needed.
- Fix 3 adds two read-only getters to `ParquetMetaFileReader` (`getSortingColumnCount`, `getSortingColumnIndex`) so a test can read the declared sort index back from the `_pm` sidecar. They mirror the existing `qdb-parquet-meta` reader and add no on-disk format change.
- No on-disk format change and no migration of existing parquet data. Existing `_pm` files are untouched.

## Test plan

- Added `extract_sorting_columns_tolerates_groups_without_sorting` (convert.rs): builds a `FileMetaData` shaped like an O3-merged partition (rg 0 with no sorting columns, rg 1 with the timestamp sort column) and asserts the validator accepts it; also asserts it still rejects a different sort column and a different sort-column count.
- Added `extract_sorting_columns_handles_many_row_groups` (convert.rs): the same validator with more than two row groups and non-declaring groups interleaved among declaring ones, so the reference is adopted from a non-first group and a conflict on a later, non-adjacent group is still rejected.
- Added `convert_from_parquet_tolerates_conflicting_sort_indices_via_qdb_meta` (convert.rs): a footer whose declaring groups disagree (stale 1 vs corrected 0) is rejected by the footer reader but accepted by the full conversion via `qdb_meta`, recording the dense position 0; without `qdb_meta` it still errors.
- Added `update_mode_append_tolerated_by_migration_via_qdb_meta` (update.rs): drives a real in-place update-mode append onto a legacy stale-index file, producing the conflicting footer, and asserts the migration tolerates it via `qdb_meta`.
- Added `extract_sorting_columns_handles_zero_row_groups` and `extract_sorting_columns_skips_empty_present_sorting_columns` (convert.rs): footer-reader edge cases (no row groups, present-but-empty sorting vector).
- Added `testSortingColumnIndexDtsAsc` and `testSortingColumnIndexExplicitArray` (ParquetMetaFileReaderTest): cover both branches of the new `getSortingColumnIndex` getter, including the explicit on-disk array.
- Added `testCorruptedSortingColumnCountValidatedBeforeArrayRead` (ParquetMetaFileReaderTest): corrupts the `_pm` header's sorting-column count to an oversized value and asserts `resolveFooter` rejects it with a clean `CairoException` rather than reading past the on-disk sorting array. This brings the new sorting-count bound check to parity with the existing `columnCount` / `rowGroupCount` corruption tests.
- Added `convert_from_parquet_tolerates_mixed_sorting_columns` (convert.rs): drives the full Mig941 conversion (resolve_sorting_columns -> detect_designated_timestamp -> `_pm` writer -> `ParquetMetaReader`) over a mixed footer and asserts the designated timestamp is still recorded.
- Updated `convert_sorting_columns_propagated` (convert.rs, pre-existing): now that `qdb_meta` is authoritative, a designated-timestamp file records the dense sort column even when its footer declares none, so the test asserts the recorded index directly instead of branching on whether the footer happened to carry sorting columns.
- Added `copy_row_group_preserves_sorting_columns` (update.rs): reproduces the rewrite-mode O3 flow (copy an existing row group, append a fresh one) with a designated timestamp, and asserts both row groups declare identical sorting columns and that the index is the dense position 0. The consistency assertion fails before Fix 2 and passes after; the dense timestamp already sits at 0 here, so the Fix-3 shift is isolated by the two tests below.
- Added `copy_row_group_with_null_columns_preserves_sorting_columns` (update.rs): the ADD/DROP COLUMN variant. The old schema places the timestamp at index 1, the target schema drops the leading column so it lands at index 0, and the updater is constructed with a deliberately bogus sort index (7). It asserts the stamped index is the dense target position 0, proving `set_target_schema` recomputes the index rather than threading the constructed value through. Fails before Fix 3, passes after.
- Added `rewrite_ignores_stale_raw_timestamp_slot` (update.rs): an already-dense `[ts, val]` file (sort column 0) with the updater fed the stale raw slot (1) and no schema change, so the construction-time derivation is the only guard. Asserts the footer declares the dense index 0, not the stale slot. Fails before Fix 3, passes after.
- Added `testO3MergeAfterDropColumnStampsDenseSortColumn` and `testO3MergeNoSchemaChangePreservesDenseSortColumn` (Java, `O3ParquetStaleReaderTest`): drive the real O3 path -- convert a partition to parquet then DROP a leading column and O3-merge (schema change), and DROP then convert then O3-merge a single-row-group partition (no schema change) -- and assert the declared sort index in the `_pm` sidecar is the dense timestamp position 0, read back through the new `ParquetMetaFileReader.getSortingColumnIndex`. Both fail before Fix 3 (`expected:<0> but was:<1>`) and pass after.
- Added `copy_row_group_preserves_absent_sorting_columns` (update.rs): the no-designated-timestamp case, where the target declares no sort order. Both the copied and the fresh group must stay without sorting columns and the validator must return an empty set. This guards the inverse regression -- the fix must not stamp sort columns onto a no-timestamp file -- and so passes both before and after Fix 2.
- `cargo fmt`, `cargo check --all-targets`, `cargo clippy --all-targets` (zero warnings) and `cargo test --lib` pass; `cargo llvm-cov` confirms the changed branches are exercised. The Java `ParquetMetaFileReaderTest` (48 tests) passes with the new getters and the sorting-count corruption test.
- The end-to-end reproduction for the crash is the enterprise `ReplicationFuzzTest.testFuzzWithBackup`, which drives the full Java replica-bootstrap path (a native rebuild plus a Maven run) and runs in enterprise CI, not OSS CI. No OSS test reproduces the mixed-footer (consistency) shape through a real O3 merge -- the Rust unit tests above construct that footer directly -- but the two `O3ParquetStaleReaderTest` cases do reproduce the stale-index defect (Fix 3) through a real O3 merge.

### DIFF
--- a/THIRD_PARTY_LICENSES.txt
+++ b/THIRD_PARTY_LICENSES.txt
@@ -1,7 +1,7 @@
 Third-party licenses for Rust dependencies of QuestDB
 ======================================================
 
-Generated on 2026-05-25 by cargo-deny
+Generated on 2026-05-29 by cargo-deny
 
 0BSD (1): adler2@2.0.1
 Apache-2.0 (75): adler2@2.0.1, alloc-checked@0.1.3, autocfg@1.5.0, bitflags@2.11.1, bytemuck@1.25.0, cc@1.2.60, cesu8@1.1.0, cfg-if@1.0.4, crc32fast@1.5.0, crossbeam-deque@0.8.6, crossbeam-epoch@0.9.18, crossbeam-utils@0.8.21, either@1.15.0, fallible-streaming-iterator@0.1.9, find-msvc-tools@0.1.9, flate2@1.1.9, getrandom@0.3.4, hashbrown@0.14.5, itoa@1.0.18, jni@0.21.1, jni-sys@0.3.1, jni-sys@0.4.1, jni-sys-macros@0.4.1, jobserver@0.1.34, libc@0.2.185, lock_api@0.4.14, log@0.4.29, memmap2@0.9.10, miniz_oxide@0.8.9, nonmax@0.5.5, num-traits@0.2.19, once_cell@1.21.4, parking_lot_core@0.9.12, parquet-format-safe@0.2.4, parquet2@0.17.2, pastey@0.1.1, pkg-config@0.3.33, proc-macro2@1.0.106, qdb-core@0.1.0, qdb-parquet-meta@0.1.0, qdbr@0.1.0, quote@1.0.45, r-efi@5.3.0, rapidhash@4.4.1, rayon@1.12.0, rayon-core@1.13.0, rustversion@1.0.22, scopeguard@1.2.0, seq-macro@0.3.6, serde@1.0.228, serde_core@1.0.228, serde_derive@1.0.228, serde_json@1.0.149, shlex@1.3.0, smallvec@1.15.1, streaming-decompression@0.1.2, syn@2.0.117, thiserror@1.0.69, thiserror-impl@1.0.69, unicode-ident@1.0.24, wasip2@1.0.2+wasi-0.2.9, windows-link@0.2.1, windows-sys@0.45.0, windows-sys@0.61.2, windows-targets@0.42.2, windows_aarch64_gnullvm@0.42.2, windows_aarch64_msvc@0.42.2, windows_i686_gnu@0.42.2, windows_i686_msvc@0.42.2, windows_x86_64_gnu@0.42.2, windows_x86_64_gnullvm@0.42.2, windows_x86_64_msvc@0.42.2, wit-bindgen@0.51.0, zstd-safe@7.2.4, zstd-sys@2.0.16+zstd.1.5.7
--- a/THIRD_PARTY_LICENSES.txt
+++ b/THIRD_PARTY_LICENSES.txt
@@ -1,7 +1,7 @@
 Third-party licenses for Rust dependencies of QuestDB
 ======================================================
 
-Generated on 2026-05-29 by cargo-deny
+Generated on 2026-05-30 by cargo-deny
 
 0BSD (1): adler2@2.0.1
 Apache-2.0 (75): adler2@2.0.1, alloc-checked@0.1.3, autocfg@1.5.0, bitflags@2.11.1, bytemuck@1.25.0, cc@1.2.60, cesu8@1.1.0, cfg-if@1.0.4, crc32fast@1.5.0, crossbeam-deque@0.8.6, crossbeam-epoch@0.9.18, crossbeam-utils@0.8.21, either@1.15.0, fallible-streaming-iterator@0.1.9, find-msvc-tools@0.1.9, flate2@1.1.9, getrandom@0.3.4, hashbrown@0.14.5, itoa@1.0.18, jni@0.21.1, jni-sys@0.3.1, jni-sys@0.4.1, jni-sys-macros@0.4.1, jobserver@0.1.34, libc@0.2.185, lock_api@0.4.14, log@0.4.29, memmap2@0.9.10, miniz_oxide@0.8.9, nonmax@0.5.5, num-traits@0.2.19, once_cell@1.21.4, parking_lot_core@0.9.12, parquet-format-safe@0.2.4, parquet2@0.17.2, pastey@0.1.1, pkg-config@0.3.33, proc-macro2@1.0.106, qdb-core@0.1.0, qdb-parquet-meta@0.1.0, qdbr@0.1.0, quote@1.0.45, r-efi@5.3.0, rapidhash@4.4.1, rayon@1.12.0, rayon-core@1.13.0, rustversion@1.0.22, scopeguard@1.2.0, seq-macro@0.3.6, serde@1.0.228, serde_core@1.0.228, serde_derive@1.0.228, serde_json@1.0.149, shlex@1.3.0, smallvec@1.15.1, streaming-decompression@0.1.2, syn@2.0.117, thiserror@1.0.69, thiserror-impl@1.0.69, unicode-ident@1.0.24, wasip2@1.0.2+wasi-0.2.9, windows-link@0.2.1, windows-sys@0.45.0, windows-sys@0.61.2, windows-targets@0.42.2, windows_aarch64_gnullvm@0.42.2, windows_aarch64_msvc@0.42.2, windows_i686_gnu@0.42.2, windows_i686_msvc@0.42.2, windows_x86_64_gnu@0.42.2, windows_x86_64_gnullvm@0.42.2, windows_x86_64_msvc@0.42.2, wit-bindgen@0.51.0, zstd-safe@7.2.4, zstd-sys@2.0.16+zstd.1.5.7
--- a/THIRD_PARTY_LICENSES.txt
+++ b/THIRD_PARTY_LICENSES.txt
@@ -1,7 +1,7 @@
 Third-party licenses for Rust dependencies of QuestDB
 ======================================================
 
-Generated on 2026-05-31 by cargo-deny
+Generated on 2026-06-01 by cargo-deny
 
 0BSD (1): adler2@2.0.1
 Apache-2.0 (75): adler2@2.0.1, alloc-checked@0.1.3, autocfg@1.5.0, bitflags@2.11.1, bytemuck@1.25.0, cc@1.2.60, cesu8@1.1.0, cfg-if@1.0.4, crc32fast@1.5.0, crossbeam-deque@0.8.6, crossbeam-epoch@0.9.18, crossbeam-utils@0.8.21, either@1.15.0, fallible-streaming-iterator@0.1.9, find-msvc-tools@0.1.9, flate2@1.1.9, getrandom@0.3.4, hashbrown@0.14.5, itoa@1.0.18, jni@0.21.1, jni-sys@0.3.1, jni-sys@0.4.1, jni-sys-macros@0.4.1, jobserver@0.1.34, libc@0.2.185, lock_api@0.4.14, log@0.4.29, memmap2@0.9.10, miniz_oxide@0.8.9, nonmax@0.5.5, num-traits@0.2.19, once_cell@1.21.4, parking_lot_core@0.9.12, parquet-format-safe@0.2.4, parquet2@0.17.2, pastey@0.1.1, pkg-config@0.3.33, proc-macro2@1.0.106, qdb-core@0.1.0, qdb-parquet-meta@0.1.0, qdbr@0.1.0, quote@1.0.45, r-efi@5.3.0, rapidhash@4.4.1, rayon@1.12.0, rayon-core@1.13.0, rustversion@1.0.22, scopeguard@1.2.0, seq-macro@0.3.6, serde@1.0.228, serde_core@1.0.228, serde_derive@1.0.228, serde_json@1.0.149, shlex@1.3.0, smallvec@1.15.1, streaming-decompression@0.1.2, syn@2.0.117, thiserror@1.0.69, thiserror-impl@1.0.69, unicode-ident@1.0.24, wasip2@1.0.2+wasi-0.2.9, windows-link@0.2.1, windows-sys@0.45.0, windows-sys@0.61.2, windows-targets@0.42.2, windows_aarch64_gnullvm@0.42.2, windows_aarch64_msvc@0.42.2, windows_i686_gnu@0.42.2, windows_i686_msvc@0.42.2, windows_x86_64_gnu@0.42.2, windows_x86_64_gnullvm@0.42.2, windows_x86_64_msvc@0.42.2, wit-bindgen@0.51.0, zstd-safe@7.2.4, zstd-sys@2.0.16+zstd.1.5.7
--- a/THIRD_PARTY_LICENSES.txt
+++ b/THIRD_PARTY_LICENSES.txt
@@ -1,7 +1,7 @@
 Third-party licenses for Rust dependencies of QuestDB
 ======================================================
 
-Generated on 2026-05-30 by cargo-deny
+Generated on 2026-05-31 by cargo-deny
 
 0BSD (1): adler2@2.0.1
 Apache-2.0 (75): adler2@2.0.1, alloc-checked@0.1.3, autocfg@1.5.0, bitflags@2.11.1, bytemuck@1.25.0, cc@1.2.60, cesu8@1.1.0, cfg-if@1.0.4, crc32fast@1.5.0, crossbeam-deque@0.8.6, crossbeam-epoch@0.9.18, crossbeam-utils@0.8.21, either@1.15.0, fallible-streaming-iterator@0.1.9, find-msvc-tools@0.1.9, flate2@1.1.9, getrandom@0.3.4, hashbrown@0.14.5, itoa@1.0.18, jni@0.21.1, jni-sys@0.3.1, jni-sys@0.4.1, jni-sys-macros@0.4.1, jobserver@0.1.34, libc@0.2.185, lock_api@0.4.14, log@0.4.29, memmap2@0.9.10, miniz_oxide@0.8.9, nonmax@0.5.5, num-traits@0.2.19, once_cell@1.21.4, parking_lot_core@0.9.12, parquet-format-safe@0.2.4, parquet2@0.17.2, pastey@0.1.1, pkg-config@0.3.33, proc-macro2@1.0.106, qdb-core@0.1.0, qdb-parquet-meta@0.1.0, qdbr@0.1.0, quote@1.0.45, r-efi@5.3.0, rapidhash@4.4.1, rayon@1.12.0, rayon-core@1.13.0, rustversion@1.0.22, scopeguard@1.2.0, seq-macro@0.3.6, serde@1.0.228, serde_core@1.0.228, serde_derive@1.0.228, serde_json@1.0.149, shlex@1.3.0, smallvec@1.15.1, streaming-decompression@0.1.2, syn@2.0.117, thiserror@1.0.69, thiserror-impl@1.0.69, unicode-ident@1.0.24, wasip2@1.0.2+wasi-0.2.9, windows-link@0.2.1, windows-sys@0.45.0, windows-sys@0.61.2, windows-targets@0.42.2, windows_aarch64_gnullvm@0.42.2, windows_aarch64_msvc@0.42.2, windows_i686_gnu@0.42.2, windows_i686_msvc@0.42.2, windows_x86_64_gnu@0.42.2, windows_x86_64_gnullvm@0.42.2, windows_x86_64_msvc@0.42.2, wit-bindgen@0.51.0, zstd-safe@7.2.4, zstd-sys@2.0.16+zstd.1.5.7
--- a/core/rust/parquet2/src/write/file.rs
+++ b/core/rust/parquet2/src/write/file.rs
@@ -626,6 +626,10 @@ impl<W: Write> ParquetFile<W> {
         self.schema = schema;
     }
 
+    pub fn set_sorting_columns(&mut self, sorting_columns: Option<Vec<SortingColumn>>) {
+        self.sorting_columns = sorting_columns;
+    }
+
     pub fn sorting_columns(&self) -> Option<&[SortingColumn]> {
         self.sorting_columns.as_deref()
     }
--- a/core/rust/qdbr/src/parquet_metadata/convert.rs
+++ b/core/rust/qdbr/src/parquet_metadata/convert.rs
@@ -34,6 +34,7 @@ use crate::parquet_metadata::types::{
 };
 use crate::parquet_metadata::writer::ParquetMetaWriter;
 use parquet2::metadata::FileMetaData;
+use parquet2::metadata::SortingColumn;
 use parquet2::schema::types::{PhysicalType, PrimitiveLogicalType};
 use qdb_core::col_type::ColumnTypeTag;
 
@@ -765,22 +766,20 @@ pub(crate) fn extract_sorting_columns(
     // copied groups (no sorting columns) with freshly written ones (the
     // timestamp sort column). Genuinely conflicting sort orders -- two groups
     // sorted on different columns -- are still rejected.
-    let mut reference: Option<(usize, Vec<SortingCol>)> = None;
+    // The reference holds the first declaring group's sorting columns borrowed
+    // straight from the footer (with its rg index, kept only for error
+    // messages). Later declaring groups are compared against it in place, so no
+    // per-row-group Vec is allocated; the result is materialized once at the end.
+    let mut reference: Option<(usize, &[SortingColumn])> = None;
 
     for (rg_idx, rg) in file_metadata.row_groups.iter().enumerate() {
-        let current: Vec<SortingCol> = match rg.sorting_columns() {
-            Some(cols) if !cols.is_empty() => cols
-                .iter()
-                .map(|sc| SortingCol {
-                    column_idx: sc.column_idx,
-                    descending: sc.descending,
-                })
-                .collect(),
+        let current: &[SortingColumn] = match rg.sorting_columns() {
+            Some(cols) if !cols.is_empty() => cols.as_slice(),
             _ => continue, // no sorting columns -> ignore this row group
         };
 
         let (ref_idx, prev) = match reference {
-            Some(ref r) => (r.0, &r.1),
+            Some(r) => r,
             None => {
                 reference = Some((rg_idx, current));
                 continue;
@@ -810,7 +809,16 @@ pub(crate) fn extract_sorting_columns(
         }
     }
 
-    Ok(reference.map(|(_, cols)| cols).unwrap_or_default())
+    Ok(reference
+        .map(|(_, cols)| {
+            cols.iter()
+                .map(|sc| SortingCol {
+                    column_idx: sc.column_idx,
+                    descending: sc.descending,
+                })
+                .collect()
+        })
+        .unwrap_or_default())
 }
 
 pub(crate) fn detect_designated_timestamp(
@@ -1089,6 +1097,19 @@ mod tests {
         assert_eq!(cols[0].column_idx, 0);
         assert!(!cols[0].descending);
 
+        // Reverse order: rg0 declares the sort column, rg1 declares none -> still tolerated.
+        metadata.row_groups = vec![
+            // fresh: ts sort
+            RowGroupMetaData::with_sorting_columns(vec![], 10, Some(ts_sort.clone()), 0),
+            // copied: no sort cols
+            RowGroupMetaData::with_sorting_columns(vec![], 10, None, 0),
+        ];
+        let cols = extract_sorting_columns(&metadata)
+            .expect("a mix of empty and non-empty sorting columns must be tolerated");
+        assert_eq!(cols.len(), 1);
+        assert_eq!(cols[0].column_idx, 0);
+        assert!(!cols[0].descending);
+
         // Two row groups sorted on DIFFERENT columns is a real conflict -> still rejected.
         metadata.row_groups = vec![
             RowGroupMetaData::with_sorting_columns(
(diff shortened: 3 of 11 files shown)


bluestreak01:
@RaphDal I took a close look at how the designated-timestamp index flows into the Parquet footer sort metadata on the O3 rewrite path, since that is exactly what this PR makes consistent. There is an edge case worth flagging: the value this PR makes consistent is wrong after a particular DROP COLUMN. It does not affect QuestDB and it does not cause the bootstrap crash this PR targets, but it is worth a tracking issue, and one of the obvious fixes would reintroduce it.

## The index is the writer-space slot, not the dense footer position

`O3PartitionJob.java:187/294` passes `tableWriterMetadata.getTimestampIndex()` into `PartitionUpdater.of(...)`. That is the raw persisted slot, not the dense column position:

- `TableWriterMetadata` reads it straight from `META_OFFSET_TIMESTAMP_INDEX` (`:155`), and `removeColumn` only `markDeleted()` (`:280-287`) - no compaction, no re-index. `TableReaderMetadata` remaps it to the dense `outIndex` (`:485-486`); the writer deliberately does not (the comment at `TableWriter.java:11886-11891` spells this out).
- `jni.rs:201-202` builds `SortingColumn::new(timestamp_index, ...)` from that slot. `set_target_schema` (`update.rs:592-594`) swaps in the dense target schema but never re-stamps `sorting_columns`, so the slot value survives against a dense schema.
- The fresh-encode path is correct: `jni.rs:532-540` derives the index via `find_map` over the dense `partition.columns`. So an O3 rewrite degrades a previously-correct index rather than introducing a brand-new one.

Reachable trace: `CREATE TABLE t (a INT, ts TIMESTAMP) TIMESTAMP(ts)` then `ALTER TABLE t DROP COLUMN a`. `a` becomes a tombstone at slot 0 and `timestampIndex` stays 1 (`rewriteMetadata` persists it unchanged at `:11574`). A later out-of-order write rewrites the partition to the dense schema `[ts]` (one column at index 0), but the footer declares sort column 1, which is out of range for a one-column file. With more columns it lands on the wrong column instead of out of range.

## Symptom: none inside QuestDB, narrow for external readers

QuestDB takes the designated timestamp from the dense QdbMeta flag (`detect_designated_timestamp`, `convert.rs:830-836`) and `_pm.designated_timestamp` (`getDesignatedTimestampColumnIndex()`), and nothing in the read path consults the footer sort-column list. `extract_sorting_columns` (`convert.rs:758-810`) checks only cross-row-group agreement, not range, so the consistent-but-wrong value passes - which is also why it does not trip the bootstrap crash this PR fixes. No consumer indexes an array with the value, so there is no panic.

Result: no crash, no wrong results, no ordering or pruning impact inside QuestDB. The only surface is external tools reading the partition `.parquet` directly after the drop-column + rewrite sequence:

- Most engines (pyarrow, DuckDB, Spark, Polars) treat `sorting_columns` as advisory and ignore it on plain reads, so no visible effect.
- Metadata inspectors report the wrong or out-of-range sorted column (the most likely observable effect).
- A strict footer validator may reject the out-of-range case.
- An engine that trusts sort metadata for an optimization could in principle mis-order or skip rows; this is not default behavior in mainstream tools.

So it is a latent metadata-fidelity issue on the export/interop surface, not a correctness or availability problem. Its practical cost is that it undercuts the "truthful per-row-group metadata" goal for this one case.

## If a follow-up addresses it

`tableToParquetIdx.getQuick(timestampIndex)` (already computed at `O3PartitionJob.java:351`) is not a correct fix. That map is keyed to the existing on-disk file (`:160-172`), which still contains the dropped column, so in the schema-change case it returns the source position (1 for `[a, ts]`), not the dense target position (0) the rewritten footer needs. It is correct only when the source file is already dense. The robust fix mirrors the encoder: `find_map` over the dense target `partition.columns`, or equivalently count non-deleted columns before `timestampIndex`.

One note on coverage: `copy_row_group_with_null_columns_preserves_sorting_columns` hand-feeds the correct dense value `SortingColumn::new(0)` into `ParquetUpdater::new`, so it proves the updater stamps consistently but does not exercise the value production actually supplies via `getTimestampIndex()`. A test driving the Java `of(...)` path for a dropped-leading-column partition would close that gap.

RaphDal:
@bluestreak01 
Great catch, 3dc83b1ec6f13a0ef86f8961f0d0603f2b0a646a should solves this.
From PR body: Fix 3 -- declare the correct sort-column index (parquet_write/update.rs, plus a setter on the vendored parquet2 crate). The file-level sorting columns that Fix 2 stamps were themselves derived from the raw timestamp slot, so Fix 2 made the row groups agree on a stale index after a pre-timestamp DROP COLUMN. `ParquetUpdater` now derives the sort column from the dense designated-timestamp position recorded in the file's QuestDB metadata: at construction for the no-schema-change rewrite, and re-stamped in set_target_schema from the dense target schema for the ADD/DROP COLUMN rewrite. This mirrors the fresh-encode path and the designated-timestamp position that end() already computed densely, so the copy paths now agree with the encoder. A new `ParquetFile::set_sorting_columns` setter lets `set_target_schema` re-stamp the value after it rebuilds the schema. The raw timestamp_index threaded elsewhere -- the designated-timestamp flag and the raw-slot merge loops that test i == timestampIndex -- is left unchanged, since it is correct as a writer-slot index there; only the derivation of sorting_columns changes.

bluestreak01:
@RaphDal — thorough review pass on this PR. The three fixes are individually well-reasoned and most of the change is clean: the `extract_sorting_columns` rewrite is correct and a genuine improvement (K allocations → 1), the `_pm`/footer offset math matches byte-for-byte across the Rust↔Java boundary, there are no reachable Rust panics, no resource leaks, and no concurrency hazards (every changed path is single-threaded). But verification surfaced one **Critical** gap: the fix is incomplete for the update/append merge mode, where it can re-create the exact replica-bootstrap crash this PR exists to eliminate — and that mode has no test coverage.

---

## Critical

### C1 — Update/append-mode O3 merge re-introduces the cross-row-group sorting conflict crash

**Status: confirmed** — reproduced independently three times (identical output `rg[0]=[1], rg[1]=[1], rg[2]=[0] → Err: "sorting column 0 differs between row groups 0 and 2"`), and the mechanism is verified directly in source.

**Classification:** in-diff root cause (`update.rs:234`, plus the validator's refusal in `convert.rs:789-808`), manifesting through the **out-of-diff** update-mode path (`parquet2` `Mode::Update::end()` + `O3PartitionJob`).

**Mechanism (each link source-verified):**
1. `ParquetUpdater::new` sets `self.sorting_columns = sorting_columns.and(designated_sorting_columns(&source_qdb_meta))` (`update.rs:234`). This runs in **both** rewrite and update mode — it precedes the `is_rewrite` split.
2. In update mode, `parquet2` `add_row_group` stamps only the **freshly inserted** group with `self.sorting_columns` (`parquet2/src/write/file.rs:702`), while `Mode::Update::end()` keeps every **existing** row group's on-disk `sorting_columns` verbatim (`file.rs:887-917`, `RowGroupSource::Cached`) — it never re-stamps them.
3. `extract_sorting_columns` hard-errors when two *declaring* groups disagree on `column_idx` (`convert.rs:799-808`). Fix 1 only tolerates `None`/empty groups (`convert.rs:777` `continue`); it does **not** tolerate two differing non-empty indices.

**Reachable trace (cross-version upgrade):**
1. Pre-PR binary. `CREATE TABLE x (a INT, ts TIMESTAMP) TIMESTAMP(ts) PARTITION BY DAY WAL`; `CONVERT PARTITION TO PARQUET`.
2. `ALTER TABLE x DROP COLUMN a` → the raw timestamp writer slot stays `1` (DROP leaves a tombstone) while the dense file has `ts` at `0`.
3. Pre-PR O3 merge → rewrite (forced by `hasSchemaChange`) → dense multi-row-group file whose declaring groups carry the stale index `[1]`, copied groups carry `None`. Footer = `[None, [1], …]`. Dead-space reset to 0. **Fix 1 tolerates this static file on restore — the intended fix works here.**
4. Upgrade to this PR.
5. O3 merge → **update mode** (`!hasSchemaChange` now that the file is dense; `rowGroupCount > 1`; dead-space ≈ 0 → not a rewrite, `O3PartitionJob.java:210-213`). Existing groups keep `[None, [1]]`; the appended group is stamped `[0]`. Footer = `[None, [1], [0]]`.
6. A backup is taken and restored to a replica (backups carry no `_pm`, so `Mig941`/`TableSnapshotRestore` reads the footer) → `extract_sorting_columns` → `SchemaMismatch` → bootstrap aborts. Same crash class this PR fixes.

**Why this contradicts the PR's own guarantee.** The PR body states: *"Files written with the stale index before this change keep it on disk; because every row group shares that same index they are internally consistent and read correctly ... so no migration is needed."* That breaks the moment such a file is update-merged post-upgrade: Fix 3 corrects only the appended group, turning a Fix-1-tolerable `[None,[1]]` into an intolerable `[None,[1],[0]]`. For this specific update-mode case the new code is strictly worse than master (master appended a matching `[1]`, producing no intra-footer conflict among declaring groups).

**Reachability is narrow and conjunctive** (pre-PR binary + pre-timestamp `DROP COLUMN` + multi-row-group partition + version upgrade + an update-mode merge + backup taken *after* that merge + replica restore). But every link is realistic, and the outcome is a replica bootstrap crash in precisely the enterprise restore path this PR targets.

**Remediation (strongest first):**
- **(a)** Make the migration crash-proof for QuestDB files: when declaring groups disagree on a non-empty index, resolve via `qdb_meta`'s designated timestamp instead of hard-erroring. The read path already treats `qdb_meta` as authoritative and `detect_designated_timestamp` already prefers it, so `extract_sorting_columns`/`convert_from_parquet` need never abort on footer sort inconsistency. This also subsumes the original bug.
- **(b)** In update mode, derive the appended group's sorting from the *existing footer's* declared columns (match on-disk) rather than the recomputed dense index, keeping the file internally consistent.
- **(c)** In `O3PartitionJob`, force a rewrite when the existing footer's per-group sort index diverges from the dense designated position (now detectable via the new `getSortingColumnIndex` vs `getDesignatedTimestampColumnIndex`), normalizing all groups in one pass.

Whichever path is chosen, add the update-mode test coverage in M1.

---

## Moderate

### M1 — The update/append-mode O3 sorting path has zero test coverage (in-diff)
All four new Rust `ParquetUpdater` tests construct in rewrite mode (`write_file_size == 0`: `update.rs:2261/2356/2462/2574`). Both Java tests force rewrite (schema-change, or single-row-group). The one pre-existing update-mode multi-row-group test (`testStaleReaderSurvivesO3MergeWithColumnTops`) asserts nothing about sorting and has no `DROP COLUMN`, so the stale-slot condition can't manifest. C1 lives precisely in this untested mode. Add a Rust update-mode test (`write_file_size != 0`, multi-RG source whose cached groups carry `[1]`, append a group, assert footer consistency or documented rejection) and a Java end-to-end update-mode case that round-trips footer→`_pm` through `ParquetMetadataWriter.generate`.

### M2 — Comment on the load-bearing validator contradicts this PR's own change (in-diff)
`convert.rs:763-766` states an O3 merge "copies unchanged row groups through `copy_row_group` **without re-stamping** the partition's designated-timestamp sort column." This PR's Fix 2 makes `copy_row_group` re-stamp. The tolerance now exists purely for backward-compatibility with files written *before* this fix. As written, the comment will mislead a future reader into believing the current write path still leaves copied groups unstamped — contradicting the sibling hunk in `update.rs`. Reword to frame it as legacy-file compatibility.

### M3 — Java tests assert less than they appear; one assertion is tautological (in-diff)
`getSortingColumnIndex` returns `getDesignatedTimestampColumnIndex()` whenever the `SORTING_IS_DTS_ASC` flag is set — the always-true case for O3/converted files — reading the *same* header offset. So `O3ParquetStaleReaderTest.java:550-552` (`getSortingColumnIndex(0) == getDesignatedTimestampColumnIndex()`) can never fail; it is dead weight. Only assertion 2 (`== expectedSortColumnIndex`) actually catches the bug. Separately, these tests read the regenerated `_pm` (correct in both modes), not the parquet footer, so they structurally cannot detect C1. Drop/rework assertion 3, and add a case that forces the explicit-array branch (a sort column that is not the ascending designated timestamp) plus a footer-round-trip case.

---

## Minor

- **m1** — `getSortingColumnIndex` (`ParquetMetaFileReader.java:518-525`) lacks the bounds check its Rust counterpart enforces (`qdb-parquet-meta/src/header.rs:397-417` returns an error on out-of-range; the explicit-array branch here does an unchecked `Unsafe.getInt`, and the DTS-ASC branch only `assert`s, disabled in prod). Test-only caller today, so unreachable — but asymmetric with Rust and with the class's other assert-guarded accessors. Add `assert sortingColumnIndex >= 0 && sortingColumnIndex < getSortingColumnCount()` and validate the array region in `resolveFooter` before any production caller lands. (in-diff)
- **m2** — `.and()` comment overstates (`update.rs:227-234`): "keep the caller's presence decision" — `.and()` can flip the caller's `Some` to `None` when the source `qdb_meta` has no designated column. Tighten the wording. (in-diff)
- **m3** — `designated_sorting_columns` hardcodes `descending=false` (`update.rs:1208-1214`), losing direction for a descending designated timestamp on the rewrite/update footer. Pre-existing convention (`jni.rs:202/540` also hardcode `false`) and unreachable via table partitions (only `COPY … TO PARQUET` export sets descending, and those files never enter O3 merge); `_pm` recovers direction from column flags. Informational. (in-diff)
- **m4** — The "bogus index 7" test (`copy_row_group_with_null_columns_preserves_sorting_columns`) proves less than its comment claims: `7` is discarded by `.and(designated_sorting_columns(source))` at `update.rs:234` before `set_target_schema` runs, so it can never reach the assertion. The test legitimately distinguishes the target recompute (→0) from the source recompute (→1), but the comment's "not the bogus 7" is unobservable. Fix the comment. (in-diff)
- **m5** — Missing small edge tests: zero-row-group `extract_sorting_columns` (returns `Ok(empty)`, untested) and a `Some(vec![])` empty-but-present group mixed with `Some([0])` (the `!cols.is_empty()` guard at `convert.rs:777` is uncovered). Trivial to add. (in-diff)

---

## PR metadata & conventions

- **Title type `chore(core):` undersells a crash fix.** This prevents a replica bootstrap crash and changes both the parquet write path and the migration validator; `fix(core):` is the accurate type, and the **`Bug` label is missing** (current labels: Core, rust, storage). The description text is copied to release notes, where a crash fix should surface.
- **Title is implementation-focused.** It could convey end-user impact, e.g. "prevent replica bootstrap failure when restoring backups of parquet partitions that received out-of-order writes."
- No `Fixes #NNN` — acceptable if there is no public issue (the reproduction is the enterprise `ReplicationFuzzTest`).
- The "Tradeoffs and limitations" section is thorough and level-headed; the one gap is C1, which it does not anticipate.

---

## Downgraded (considered and dismissed after source verification)

- **"PR added a 3rd `QdbMeta` deserialize in `ParquetUpdater::new`"** — false. `git show master` confirms master's `new()` already had 3 deserialize calls; HEAD also has 3. The PR *reused* the existing validation deserialize (bound it to `source_qdb_meta`). The 3× redundancy is pre-existing and out of scope.
- **"`.and(designated_sorting_columns(source))` drops the footer sort for files lacking the designated flag"** — not reachable: the encode path always marks the designated column in `qdb_meta` (`jni.rs:704-710, 825-839`) and the O3 caller always passes `Some`; the inline `_pm` `designated_timestamp` is computed independently from `qdb_meta`, not from `sorting_columns`.
- **"`set_target_schema`'s unconditional re-stamp diverges from `new()`'s presence-preserving `.and()`"** — safe: `DROP`/`ADD COLUMN` can neither drop the designated timestamp nor add a second; only its position shifts.
- **Concurrency / torn reads / mmap aliasing** — none: every changed path is single-threaded; `extract_sorting_columns` borrows heap `FileMetaData` structs, not the mmap.
- **Resource leaks on error paths** — none: `new()`'s `File` values are owned and dropped on every early return; the added logic is infallible/resource-free; the migration mmap is RAII-dropped on both old-`Err` and new-`Ok` paths.
- **Rust panic via JNI** — none reachable: `is_designated()→tag().expect()` can't fire because `QdbMeta::deserialize` routes every `column_type` through a validating `TryFrom<i32>` that rejects corrupt footers at the `?` before `tag()` runs; `pos as i32` needs >2³¹ columns.

---

## Summary

**Verdict: request changes.** One Critical finding (C1) plus its enabling test gap (M1).

- C1 is an incompleteness, not a fresh regression against master's common case: the PR genuinely fixes the originally-reported crash (a static legacy file is now tolerated), but Fix 3 applied only to the appended group in update mode can convert a tolerable footer into an intolerable one, re-opening the same crash on a narrower cross-version-upgrade path. Because it lands in the exact replica-restore domain this PR targets and is confirmed by three independent reproductions, it should be resolved (or consciously deferred with a tracking issue + the robustness fix). Remediation (a) also hardens the migration against the whole class.
- Real tradeoffs worth keeping visible: Fix 1 deliberately relaxes the validator (it no longer flags a row group genuinely missing its sort column), and Fix 3 is latent inside QuestDB (the read path ignores footer sorting). Both are sound given the read path trusts `qdb_meta`.

*Review verification: ~17 draft findings examined, 10 verified (1 Critical, 3 Moderate, 5 Minor, plus metadata), 7 dropped as false positives. All verified findings are in-diff; C1's crash manifests at out-of-diff callsites (`parquet2` update-mode `end()` and `O3PartitionJob` mode selection), which is why a diff-only reading misses it.*

RaphDal:
@bluestreak01 - Thanks for the thorough pass — the C1 analysis in particular was worth the read. I independently reproduced the mechanism from source before acting on it, and it holds end to end:

- update mode stamps only the appended group (`add_row_group` → `write_row_group(&self.sorting_columns)`) and copies cached groups verbatim (`RowGroupSource::Cached`);
- the appended group carries the *corrected* dense index (`new()`'s `.and(designated_sorting_columns(source))`) while a pre-fix legacy file's cached group still carries the stale one, so a cross-version update-merge yields `[None, [1], [0]]`;
- both migration call sites (`converter.rs` and inside `convert_from_parquet`) ran `extract_sorting_columns(...)?`, which hard-errors on that footer.

Confirmed and fixed. Details below, by your IDs.

## C1 — fixed via remediation (a)

The migration now resolves the sort column from `qdb_meta`'s designated timestamp — which the read path already treats as authoritative — and consults the footer only as a fallback. I applied it at **both** call sites (the `converter.rs` one crashes first), via a shared `resolve_sorting_columns(file_metadata, qdb_meta)`.

This subsumes the original bug, so it's the migration-side hardening you outlined as the strongest option.

## M1 — update-mode coverage added

- `update_mode_append_tolerated_by_migration_via_qdb_meta` (update.rs): a real in-place update-mode append (`write_file_size != 0`) onto a legacy stale-index file, producing the `[1]` vs `[0]` conflicting footer, asserting the footer reader rejects it and the migration tolerates it via `qdb_meta`.
- `convert_from_parquet_tolerates_conflicting_sort_indices_via_qdb_meta` (convert.rs): the same footer shape end to end through the conversion, including the negative (no `qdb_meta` → still errors).

## M2 — comment reworded

`extract_sorting_columns`'s doc now frames the no-sorting-columns skip as legacy-file compatibility for the footer fallback, not a claim about the current write path (which Fix 2 does re-stamp).

## M3 — already partly addressed; coverage added

The tautological `getSortingColumnIndex(0) == getDesignatedTimestampColumnIndex()` assertion was already removed in the earlier "address feedback" commit — the current `assertParquetSortColumn` asserts each getter against `expectedSortColumnIndex` independently. The real remaining gap was the **explicit-array branch**, which had no Java coverage. Added `testSortingColumnIndexExplicitArray` (sort column != designated ts → reads the on-disk array) and `testSortingColumnIndexDtsAsc` to `ParquetMetaFileReaderTest`.

## Minors

- **m1** — the bounds-check assert is already present (also from the earlier commit); trimmed the doc.
- **m2 / m4** — reworded. For m4 specifically: agreed the "bogus 7" framing was unobservable since `.and(designated_sorting_columns(source))` reduces it to the source's `1` at construction; the comment and assertion now describe the real distinction (target `0` vs source `1`).
- **m3** — left as-is on the write side (it's the existing convention and unreachable for table partitions). Note the new migration-side helper *does* honor the order bit, so that path is correct for a descending designated timestamp.
- **m5** — added `extract_sorting_columns_handles_zero_row_groups` and `extract_sorting_columns_skips_empty_present_sorting_columns`.

## PR metadata

Keeping `chore(core)` and not adding `Bug`. Our convention reserves `fix` for customer-reported bugs; this surfaced through `ReplicationFuzzTest`, not a customer report, so `chore` is the consistent type here. Happy to revisit if we'd rather treat fuzz-discovered crashes as `fix`.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 573 / 589 (97.28%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/parquet_write/update.rs|338|352|96.02%|
|:large_blue_circle:|qdbr/src/parquet_metadata/convert.rs|216|218|99.08%|
|:large_blue_circle:|qdbr/src/parquet_metadata/jni/converter.rs|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileReader.java|15|15|100.00%|
|:large_blue_circle:|parquet2/src/write/file.rs|3|3|100.00%|

bluestreak01 (approved):
Approved. The core fix is correct and the previously-found Critical (update-mode cross-version conflict re-introducing the replica crash) is genuinely closed via the qdb_meta-authoritative resolution at both migration call sites. Independent level-2 review confirms: no reachable Rust panics via JNI, no resource leaks, no concurrency hazards, and the _pm offset math matches the Rust writer byte-for-byte.

Non-blocking follow-ups worth tracking:
- The extract_sorting_columns relaxation also loosens the no-qdb_meta (external parquet) path reachable via ATTACH PARTITION: a footer mixing declaring/non-declaring row groups that master rejected is now accepted. Either scope the skip-empty tolerance to the qdb_meta-present path or document the more permissive external inference.
- The descending designated-timestamp branch of designated_sorting_col / designated_ts_sorting_columns is untested (all tests hardcode ascending). Add one descending case to back the 'honors the order bit' claim.
- designated_sorting_col (convert.rs) and designated_ts_sorting_columns (update.rs) are the same computation in the same crate and must stay in sync; worth consolidating.
- Explicit sort-array entries are written into _pm unchecked against columnCount on the no-qdb_meta path (asymmetric with the dts-index guard); test-only consumer today.

PR #7185, opened 2026-05-29
fix(core): read a column added after Parquet conversion as NULL

bluestreak01:
## What

A query that projects **only** column(s) absent from a Parquet partition — i.e. columns added with `ALTER TABLE ADD COLUMN` after the partition was converted to Parquet — could crash the server (`SIGSEGV`) or return stale data instead of `NULL`.

It triggers whenever the projected column set contains no column that exists in the Parquet file, for example:

```sql
ALTER TABLE t CONVERT PARTITION TO PARQUET WHERE ...;
ALTER TABLE t ADD COLUMN s SYMBOL;
SELECT s FROM t;   -- reads only 's', which the Parquet file does not contain
```

`SELECT *`, and any projection that also reads a column present in the Parquet file, are unaffected. Native partitions are unaffected.

## Root cause

`PageFrameMemoryPool.ParquetBuffers.decode()` sizes and zeroes the per-column page-address lists only inside the `parquetColumns.size() > 0` branch, through `remapColumns()`. When every projected column is absent from the Parquet file, `parquetColumns` is empty, `remapColumns()` is skipped, and the lists retain stale or uninitialized native memory (`clearAddresses()` only resets the list position, and `DirectLongList.get()` does no bounds check).

`PageFrameMemoryRecord` then reads a non-zero, invalid page address for the absent column and dereferences it:

- a freshly allocated buffer holds uninitialized `malloc` memory, so the address is a wild pointer and `getSymA()`/`getInt()`/... `SIGSEGV`;
- a reused buffer (the parquet frame cache evicts and reuses a previously decoded buffer) holds a leftover pointer from another frame, so the read returns a wrong, non-NULL value.

This is a regression from the introduction of column add/drop over Parquet partitions (#6875): before that, every projected column was guaranteed to exist in the Parquet file.

## Fix

Two coupled changes in `PageFrameMemoryPool`:

1. Hoist `remapColumns()` out of the `parquetColumns.size() > 0` guard so the address lists are always sized and zeroed; only `decodeRowGroup()` stays conditional. When no columns are decoded, the remap loop does no work, so every column resolves to address 0 and the absent column reads as `NULL`, matching native-partition behaviour.

2. Add an early `if (columnCount == 0) return;` guard at the top of `remapColumns()`. The hoist makes `remapColumns()` run unconditionally, including for zero-column reads such as `count(*)`. Without the guard, `ensureCapacityAndZero(list, 0)` calls `DirectLongList.setCapacity(0)` and trips its `assert capacity > 0`. The guard leaves the already-cleared lists empty, which is the correct state for a zero-column read.

A follow-up also early-returns from `remapColumns()` when `parquetColumns` is empty, after the four lists have been zeroed. In that case every loop iteration would resolve to `parquetIdx < 0` and `continue`, so the loop is provably dead. This skips it. It is a micro-optimization, not a correctness change.

## Impact and tradeoffs

- Correctness/stability fix: absent Parquet columns now read as `NULL` instead of crashing or returning stale data.
- Overhead is negligible: in the empty-projection case the change adds one `remapColumns()` call that zeroes four small `DirectLongList`s (and now returns before the dead loop).
- No behaviour change for any query that reads at least one column present in the Parquet file.

## Test plan

- `ParquetTest.testSelectOnlyAddedColumnAbsentFromParquetPartition` reproduces the bug deterministically. It lays out enough column-present Parquet partitions to drain the parquet frame cache, forcing the column-absent frame to reuse a decoded buffer, so the stale read surfaces as a wrong value rather than depending on a `SIGSEGV`. It is parameterized over the added column type: SYMBOL exercises the data-vector read path (`getSymA` gates on `pageAddresses`), and a VARCHAR variant (`testSelectOnlyAddedVarcharColumnAbsentFromParquetPartition`) exercises the aux-vector path (`getVarcharSize`/`getVarchar` gate on `auxPageAddresses`). The test pins `CAIRO_SQL_PARQUET_FRAME_CACHE_CAPACITY` and ties the partition count to it, so a future change to the default cache capacity cannot silently defeat the buffer-reuse trigger. Both fail on `master` (the absent column returns its prior value instead of `NULL`) and pass with this change.
- `ParquetTest.testCountStarOverParquetPartition` covers the zero-column (`count(*)`) read path that the `columnCount == 0` guard protects. Note this test passes on `master` — there `remapColumns()` is never called for a zero-column projection, so it guards an `AssertionError` introduced by the hoist, not a master crash. It pins the plan with `assertPlanNoLeakCheck` so an optimizer change cannot silently route the query to the metadata-only `Count` factory and skip the frame decode.
- The result-set tests use `assertQueryNoLeakCheck`, which re-creates the cursor and re-reads, additionally exercising parquet decode re-entry.
- `ParquetTest.testAbsentColumnLateMaterializationOverParquet` covers the filtered / late-materialization read of an absent column through `decodeRemainingColumns` and `PageFrameFilteredMemoryRecord` (a separate path from the record-scan tests). The query filters on a present column and groups by the absent VARCHAR column over a parquet partition, comparing against a native sibling. It passes on `master` — the present filter column makes the filter pass size and zero the address lists — so it is a guard for the aux-vector zeroing on that route rather than a reproduction.
- Full `io.questdb.test.griffin.ParquetTest` (and the sqllogictest sibling) passes: 118 run, 0 failures, 1 skipped (pre-existing `@Ignore`).

### DIFF
--- a/core/src/main/java/io/questdb/cairo/sql/PageFrameMemoryPool.java
+++ b/core/src/main/java/io/questdb/cairo/sql/PageFrameMemoryPool.java
@@ -28,8 +28,8 @@ import io.questdb.cairo.CairoException;
 import io.questdb.cairo.ColumnType;
 import io.questdb.cairo.Reopenable;
 import io.questdb.griffin.engine.table.parquet.ParquetDecoder;
-import io.questdb.griffin.engine.table.parquet.ParquetPartitionDecoder;
 import io.questdb.griffin.engine.table.parquet.ParquetFileDecoder;
+import io.questdb.griffin.engine.table.parquet.ParquetPartitionDecoder;
 import io.questdb.griffin.engine.table.parquet.RowGroupBuffers;
 import io.questdb.std.DirectIntList;
 import io.questdb.std.DirectLongList;
@@ -273,6 +273,35 @@ public class PageFrameMemoryPool implements RecordRandomAccess, QuietCloseable, 
         frameMemory.clear();
     }
 
+    private void activateDecoder(int frameIndex) {
+        final ParquetDecoder frameDecoder = addressCache.getParquetDecoder(frameIndex);
+        if (frameDecoder instanceof ParquetPartitionDecoder parquetMetaFrame) {
+            if (parquetMetaDecoder.getFileAddr() != parquetMetaFrame.getFileAddr() || parquetMetaDecoder.getFileSize() != parquetMetaFrame.getFileSize()) {
+                parquetMetaDecoder.of(parquetMetaFrame);
+                buildColumnIdMap(parquetMetaDecoder);
+            }
+            activeDecoder = parquetMetaDecoder;
+        } else {
+            ParquetFileDecoder legacyFrame = (ParquetFileDecoder) frameDecoder;
+            if (legacyDecoder.getFileAddr() != legacyFrame.getFileAddr() || legacyDecoder.getFileSize() != legacyFrame.getFileSize()) {
+                legacyDecoder.of(legacyFrame);
+                buildColumnIdMap(legacyDecoder);
+            }
+            activeDecoder = legacyDecoder;
+        }
+    }
+
+    private void buildColumnIdMap(ParquetDecoder decoder) {
+        final int parquetColumnCount = decoder.getColumnCount();
+        columnIdToParquetIdx.clear();
+        for (int i = 0; i < parquetColumnCount; i++) {
+            final int id = decoder.getColumnId(i);
+            // External parquet files may not have field IDs (all -1).
+            // Fall back to positional index so the lookup in openParquet() works.
+            columnIdToParquetIdx.put(id < 0 ? i : id, i);
+        }
+    }
+
     // We don't use additional data structures to speed up the lookups
     // such as <frame_index, buffers> hash table. That's because we don't
     // expect the cache size to be large.
@@ -326,35 +355,6 @@ public class PageFrameMemoryPool implements RecordRandomAccess, QuietCloseable, 
                 .put("insufficient memory pool size [size=").put(parquetCacheSize)
                 .put(", usageBit=").put(usageBit)
                 .put(']');
-    }
-
-    private void buildColumnIdMap(ParquetDecoder decoder) {
-        final int parquetColumnCount = decoder.getColumnCount();
-        columnIdToParquetIdx.clear();
-        for (int i = 0; i < parquetColumnCount; i++) {
-            final int id = decoder.getColumnId(i);
-            // External parquet files may not have field IDs (all -1).
-            // Fall back to positional index so the lookup in openParquet() works.
-            columnIdToParquetIdx.put(id < 0 ? i : id, i);
-        }
-    }
-
-    private void activateDecoder(int frameIndex) {
-        final ParquetDecoder frameDecoder = addressCache.getParquetDecoder(frameIndex);
-        if (frameDecoder instanceof ParquetPartitionDecoder parquetMetaFrame) {
-            if (parquetMetaDecoder.getFileAddr() != parquetMetaFrame.getFileAddr() || parquetMetaDecoder.getFileSize() != parquetMetaFrame.getFileSize()) {
-                parquetMetaDecoder.of(parquetMetaFrame);
-                buildColumnIdMap(parquetMetaDecoder);
-            }
-            activeDecoder = parquetMetaDecoder;
-        } else {
-            ParquetFileDecoder legacyFrame = (ParquetFileDecoder) frameDecoder;
-            if (legacyDecoder.getFileAddr() != legacyFrame.getFileAddr() || legacyDecoder.getFileSize() != legacyFrame.getFileSize()) {
-                legacyDecoder.of(legacyFrame);
-                buildColumnIdMap(legacyDecoder);
-            }
-            activeDecoder = legacyDecoder;
-        }
     }
 
     private void openParquet(int frameIndex) {
@@ -637,6 +637,17 @@ public class PageFrameMemoryPool implements RecordRandomAccess, QuietCloseable, 
             ensureCapacityAndZero(pageSizes, columnCount);
             ensureCapacityAndZero(auxPageAddresses, columnCount);
             ensureCapacityAndZero(auxPageSizes, columnCount);
+
+            if (parquetColumns.size() == 0) {
+                // No parquet column was decoded (every projected column was added
+                // after this partition became parquet). openParquet() only adds a
+                // column to parquetColumns when columnIdToParquetIdx maps it to a
+                // present parquet column, so an empty parquetColumns means every
+                // column below would resolve to parquetIdx < 0 and continue. The
+                // zeroing above already left them all at address 0 (NULL), so skip
+                // the dead remap loop.
+                return;
+            }
 
             final ColumnMapping columnMapping = addressCache.getColumnMapping();
             final int readParquetColumnCount = columnMapping.getColumnCount();
--- a/core/src/main/java/io/questdb/cairo/sql/PageFrameMemoryPool.java
+++ b/core/src/main/java/io/questdb/cairo/sql/PageFrameMemoryPool.java
@@ -626,6 +626,13 @@ public class PageFrameMemoryPool implements RecordRandomAccess, QuietCloseable, 
         // address pair into their respective query slots.
         private void remapColumns() {
             final int columnCount = addressCache.getColumnCount();
+            if (columnCount == 0) {
+                // The query reads no columns (e.g. count(*)). clearAddresses() already
+                // left the lists empty, which is the correct state; there is nothing to
+                // remap. Sizing them to 0 would trip DirectLongList.setCapacity()'s
+                // assert capacity > 0.
+                return;
+            }
             ensureCapacityAndZero(pageAddresses, columnCount);
             ensureCapacityAndZero(pageSizes, columnCount);
             ensureCapacityAndZero(auxPageAddresses, columnCount);
--- a/core/src/main/java/io/questdb/cairo/sql/PageFrameMemoryPool.java
+++ b/core/src/main/java/io/questdb/cairo/sql/PageFrameMemoryPool.java
@@ -566,8 +566,15 @@ public class PageFrameMemoryPool implements RecordRandomAccess, QuietCloseable, 
             clearAddresses();
             if (parquetColumns.size() > 0) {
                 decoder.decodeRowGroup(rowGroupBuffers, parquetColumns, rowGroup, rowLo, rowHi);
-                remapColumns();
             }
+            // Always size and zero the page-address lists, even when there are no
+            // parquet columns to decode (every projected column was added after this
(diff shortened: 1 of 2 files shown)


RaphDal:
@bluestreak01 review from Claude

Critical

None.

Moderate

M1 — The regression test silently rots if the parquet-frame-cache default ever rises above 8 (in-diff,
ParquetTest.java testSelectOnlyAddedColumnAbsentFromParquetPartition).
The test's determinism depends on the cache holding exactly 8 buffers so the 9th (column-absent) frame is
forced to reuse an already-decoded buffer and read its stale slot-0 pointer. That 8 lives only in a comment
(lines ~1925, ~1954). The test never pins it — ParquetTest doesn't set CAIRO_SQL_PARQUET_FRAME_CACHE_CAPACITY,
so it inherits DefaultCairoConfiguration's 8. If that default is raised above 8, the 06-09 frame draws a
fresh (never-decoded) buffer instead; fresh OS pages are typically zero, so on master the absent column would
read NULL and the test would silently pass on master — a false-green regression test that no longer guards
anything. (The config floor is Math.max(..., 8), so the risk is one-directional: lowering keeps it red,
raising defeats it.)
Fix: pin it in the test, e.g. node1.setProperty(PropertyKey.CAIRO_SQL_PARQUET_FRAME_CACHE_CAPACITY, "8") (or
set it to a smaller value like 2 and build that many present partitions), so a future default change can't
quietly neuter the test.

M2 — No test exercises the absent column through the late-materialization / aux-vector paths (missing
coverage).
The shipped test covers only a SYMBOL column read via the data-vector (pageAddresses) on the record-scan path.
The fix also zeroes the aux lists (auxPageAddresses/auxPageSizes) and is relied upon by the
late-materialization path (populateRemainingColumns → decodeRemainingColumns, whose unchanged if
(parquetColumns.size() > 0) guard is only safe because decode() now always sizes the lists). Agent 1 proved
both are correct by construction, but neither is tested:
- A regression that zeroed only the data lists and not the aux lists would pass the current test yet crash on
an absent VARCHAR/STRING column.
- The filtered/async late-mat path (WHERE on a present column, project an absent column) is untested.

M3 — PR conventions: fix vs chore, and "crash" in the title (matches your saved preferences).
The body states the bug "originally surfaced by BackupFuzzTest.testMatViewRefreshAfterBackup" — a fuzz/CI
find, not a customer report. Per your own convention (fix vs chore), this should be chore(core):, not
fix(core):. And per PR title tone, avoid "crash." Both point to a calmer title, e.g.:
chore(core): return NULL for a column added after Parquet conversion

M4 — The PR description omits half the production change.
The "Fix" section describes only hoisting remapColumns(). It never mentions the second production change — the
new if (columnCount == 0) return; guard inside remapColumns() — which is required to keep the hoist safe for
zero-column reads (count(*)): without it the always-called remapColumns() would hit
ensureCapacityAndZero(list, 0) → setCapacity(0) → trip assert capacity > 0. The test plan also lists only
testSelectOnly... and omits testCountStarOverParquetPartition. A reviewer reading the description wouldn't
know why the guard or the count(*) test exist. Add a sentence on the guard and list both tests.

Minor

m1 — Wasted no-op loop on the (rare) empty-projection path (in-diff, PageFrameMemoryPool.java ~line 643). When
parquetColumns.size() == 0, the remap loop runs columnCount iterations that every time hit parquetIdx < 0 →
continue, doing columnCount hash probes for nothing. The mandatory work is the four ensureCapacityAndZero
zeroings; the loop is provably dead here. A one-line if (parquetColumns.size() == 0) return; after the four
zeroings removes it. Flagged independently by Agents 3 and 11; both call it non-blocking (decode runs once per
row-group per cache miss, amortized over all rows). Optional — the "always run remapColumns, the loop is a
harmless no-op" simplification is defensible.

m2 — testCountStarOverParquetPartition passes on master; its comment overstates it (in-diff). The comment
opens "Regression for a parquet read crash." It is not — on master remapColumns() is never called for a
zero-column projection, so master never trips the assert and this test is green there. It actually guards an
AssertionError introduced by this PR's own hoist. Still valuable, but reword the opening line to say so (the
mechanism explanation that follows is accurate).

m3 — "Red regression" phrasing (in-diff, testSelectOnly... comment ~line 1911). Jargon-y; "Regression test
(red on master)" reads clearer. Trivial.

m4 — assertSql vs assertQueryNoLeakCheck. Both tests use assertSql inside assertMemoryLeak. Per CLAUDE.md,
griffin query tests should prefer assertQueryNoLeakCheck (asserts factory properties). This matches the
dominant ParquetTest house style (≈96/98 use assertSql) and these are storage-read-behavior tests, so it's
defensible — but testSelectOnly... asserts a real result set and is the better candidate to convert if you
want to honor the guideline.

m5 — bugfix commit title. The second commit (59a767e2cb) is titled just "bugfix" with no body. Squash-merge
makes branch history throwaway (CLAUDE.md), so this is cosmetic, but the squashed message should carry the
full rationale (hoist + the columnCount == 0 guard).

Downgraded (false positives / out of scope)

- "PageFrameMemoryPool.of() doesn't clear cachedParquetBuffers" (Agent 1). Real but pre-existing and not
exploitable: each pool is factory-scoped with invariant columnCount, and reduce-task pools call
releaseParquetBuffers() per frame. Not introduced or worsened by this PR.
- "zero() memsets full capacity, not size; per-frame recompute of query-invariant q→slot mapping" (Agent 11).
Pre-existing micro-inefficiencies; the diff only makes remapColumns() run on more frames. Out of scope for
this fix; would be a separate optimization PR.
- Concurrency (Agent 2) and native-memory/leak (Agent 4) findings: none — both passes returned clean.
PageFrameAddressCache is frozen and safely published before parallel reads; the count(*) guard correctly
avoids the realloc-to-0 / assert path on all close/reopen sequences.

Summary

Verdict: the fix is correct — approve it; address the test-robustness (M1) and metadata (M3/M4) items before
merge, and consider the coverage additions (M2). Both production changes (hoist remapColumns() + the
columnCount == 0 guard) are necessary, coupled, and verified: absent parquet columns now read NULL instead of
a wild/stale pointer, and zero-column reads no longer trip setCapacity(0). No regressions or tradeoffs beyond
a per-frame O(columnCount) zeroing that is mandatory for correctness and accurately described in the PR.

- ~16 draft findings; ~10 reported, ~6 downgraded (1 stale-file false positive, 3 pre-existing/out-of-scope,
plus clean concurrency/resource passes).
- In-diff vs out-of-diff: all reported findings are in-diff or about the PR (tests/description/metadata); 0
out-of-diff production findings. This is justified, not an under-run: both changed methods are private and
reachable only within PageFrameMemoryPool (the 3 navigateTo overloads → decode() → remapColumns()), and Agent
1 explicitly walked every consumer (PageFrameMemoryRecord.getSymA, hasColumnTops, populateJitAddresses, the
late-mat path, varchar/fixed-size accessors) and found each safe under the new contract.

bluestreak01:
Thanks for the thorough review. Addressed across two pushes (`98a9c9b894` test hardening, `4e9eaac161` comment rewording) plus a title/description update.

**Moderate**

- **M1 (test rots if the cache default rises above 8):** Fixed. `testSelectOnlyAddedColumnAbsentFromParquetPartition` now pins `node1.setProperty(CAIRO_SQL_PARQUET_FRAME_CACHE_CAPACITY, cacheCap)` and generates exactly `cacheCap` column-present partitions, tying the partition count to the same constant. A future default change can no longer quietly neuter it.

- **M2 (aux-vector + late-mat coverage):** Partially addressed. Added a VARCHAR variant (`testSelectOnlyAddedVarcharColumnAbsentFromParquetPartition`) exercising the aux-vector path (`getVarcharSize`/`getVarchar` gate on `auxPageAddresses`), so a regression that zeroed only the data lists would now fail. The filtered/late-materialization path (`decodeRemainingColumns`) projecting an absent column is still uncovered — I checked, and the existing `LateMaterializationOverParquet` tests all filter-and-late-materialize a column that is present in the Parquet file. Flagged as a known gap in the PR description. I can add a "filter on a present column, project an absent column" variant in this PR if you'd prefer it not be a follow-up.

- **M3 (title: fix vs chore, "crash"):** Dropped "crash" — now `fix(core): read a column added after Parquet conversion as NULL`. Kept it a `fix`: a real query (add a column, then read it over a converted partition) does crash, so it is a user-facing correctness fix rather than internal cleanup.

- **M4 (description omits half the change):** Fixed. The Fix section now documents both production changes — the `remapColumns()` hoist and the `if (columnCount == 0) return;` guard (with the `setCapacity(0)` assert rationale) — and the test plan lists `testCountStarOverParquetPartition`.

**Minor**

- **m1 (dead no-op loop):** Applied. Added `if (parquetColumns.size() == 0) return;` after the four `ensureCapacityAndZero` zeroings. When `parquetColumns` is empty, `openParquet()` guarantees every column resolves to `parquetIdx < 0`, so the loop was provably dead.

- **m2 (count test comment overstates "crash on master"):** Fixed. Reworded the opening to say it guards an `AssertionError` introduced by this PR's hoist and that it passes on master.

- **m3 ("Red regression" phrasing):** Fixed. Now "Regression test (red on master)".

- **m4 (assertSql vs assertQueryNoLeakCheck):** Converted both new tests. Beyond the factory-property assertions, `assertQueryNoLeakCheck` re-creates the cursor and re-reads, which also exercises parquet decode re-entry — worthwhile on this path.

- **m5 (bugfix commit title):** Branch history is throwaway under squash-merge; the squashed message will carry the full rationale (hoist + the `columnCount == 0` guard).

Open item: the M2 late-mat coverage — let me know if you want it in this PR or tracked separately.

bluestreak01:
Follow-up on M2: added `testAbsentColumnLateMaterializationOverParquet` (commit 02f414cef8), closing the late-mat coverage gap.

It filters on a present column and groups by a column added after conversion (absent from the parquet file), so the absent VARCHAR is read through `PageFrameFilteredMemoryRecord` + `decodeRemainingColumns` — confirmed via the plan (`Async JIT Group By`, `filter: null!=c1`, `keys: [s]`) over a parquet `PageFrame` (verified `isParquet=true` for the partition after the ADD COLUMN). The parquet result is compared against a native sibling table.

One honest caveat: this test passes on `master`. With the filter column present in parquet, the filter pass's `decode()` already calls `remapColumns()` and sizes/zeroes the address lists, so the absent column reads NULL there too. So it is a guard for the aux-vector zeroing on the late-mat route, not a reproduction of the original bug. The record-scan tests remain the red-on-master reproductions. I considered a red-on-master late-mat variant (filtering on the absent column itself, to leave the filter pass's list unsized) but it relies on the optimizer not folding the always-NULL predicate, which is brittle; the guard test is the more robust choice.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 23 / 23 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|23|23|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=237514)
**Tested Commit:** `02f414cef82f6ff849092d716295d6cd0cbcf207`

PR #7186, opened 2026-05-29
chore(core): return NULL for columns added after parquet conversion

RaphDal:
# Background

When `ALTER TABLE ADD COLUMN` adds a column after a partition has already been converted to parquet, that column is not part of the partition's parquet file. Reading it should return NULL for every row of that partition, the same as a native column top.

A query that projects only such absent column(s) decodes no parquet columns for that partition. `PageFrameMemoryPool` reuses a small pool of per-frame page-address buffers across frames. The decode pass reset the buffer positions but only re-sized and zeroed the address slots when at least one parquet column was decoded. With nothing to decode, the buffers kept the page pointers written by the previously decoded frame. The record accessors then read those stale pointers and returned a neighbouring partition's value instead of NULL; when a retained pointer referenced memory that had since been freed, the read faulted.

The path is reachable from a plain `SELECT` on a primary instance. It surfaced in the parquet late-materialization fuzz test on CI rather than from a user report.

## Change

`ParquetBuffers.decode()` now calls `remapColumns()` unconditionally. `remapColumns()` sizes and zeroes the four address lists and then maps each query column to its decoded buffer. When no parquet columns were decoded, every column lacks a parquet index and maps to address 0 (NULL) without touching the row-group buffers, which is the intended result for a column absent from the partition.

## Effects

- Projecting only added-after-conversion columns returns NULL, matching both the native read path and the result when the same column is projected alongside a present column.
- `remapColumns()` now runs on every decode, including the empty case. It walks the column mapping and zeroes four `DirectLongList`s sized to the column count. This adds a small constant per-frame cost that the previous code skipped when nothing was decoded. The cost is negligible next to a row-group decode, but the empty pass is no longer free.
- Only the record-scan cursor exhibited the fault in practice. It reuses an eight-buffer pool across frames, so a populated buffer can be reused for a later empty pass. The parallel aggregation and late-materialization paths use a single-buffer pool that they release per task, so their empty passes always saw a fresh, zeroed buffer and their behavior does not change.
- The sibling late-materialization path `decodeRemainingColumns()` needs no change. It relies on `decode()` having sized and zeroed the buffer first, which now always happens.

## Test plan

- Added `testProjectOnlyAddedColumnsAcrossParquetPartitions` to `ParquetLateMaterializationFuzzTest`. It builds older partitions whose parquet files carry the added columns and newer partitions whose parquet files omit them, captures the correct result while the older partitions are still native, converts them to parquet, then asserts the parquet read matches. It covers LONG, SYMBOL and VARCHAR projections plus a multi-column projection.
- Confirmed this test fails before the change (newer rows read stale values from older partitions) and passes after.
- Added `testAggregateOnlyAddedColumnsAcrossParquetPartitions` covering a non-keyed aggregation over only the absent columns. It passes both before and after the change, since the aggregation path was not affected, and guards NULL handling on that path.
- `ParquetLateMaterializationFuzzTest` passes in full (48 tests).
- `AlterTableConvertPartitionTest` passes in full (38 tests).

### DIFF
--- a/core/src/main/java/io/questdb/cairo/sql/PageFrameMemoryPool.java
+++ b/core/src/main/java/io/questdb/cairo/sql/PageFrameMemoryPool.java
@@ -566,8 +566,14 @@ public class PageFrameMemoryPool implements RecordRandomAccess, QuietCloseable, 
             clearAddresses();
             if (parquetColumns.size() > 0) {
                 decoder.decodeRowGroup(rowGroupBuffers, parquetColumns, rowGroup, rowLo, rowHi);
-                remapColumns();
             }
+            // Always size and zero the address slots, even when nothing is decoded.
+            // When parquetColumns is empty (every projected column was added after the
+            // partition was converted to parquet), remapColumns() maps every column to
+            // address 0 (NULL) without touching rowGroupBuffers. Skipping this left the
+            // slots holding stale pointers from a previously decoded frame whose buffers
+            // this frame reuses, which the accessors then dereferenced.
+            remapColumns();
         }
 
         public void decodeRemainingColumns(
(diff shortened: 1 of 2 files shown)


RaphDal:
Duplicate of https://github.com/questdb/questdb/pull/7185

PR #7187, opened 2026-05-30
test(sql): add fluent QueryAssertion builder for query tests, fix 9 discovered bugs

bluestreak01:
Absorbs #7193 because `master` is compile-broken without it.

## What

Introduces a standalone fluent query-assertion builder, `io.questdb.test.QueryAssertion`,
reached through `assertQuery(CharSequence query)` on the test base classes. The expected
output moves to a terminal step, so a chain reads as query -> options -> expectation:

```java
assertQuery("select * from x")
        .timestamp("ts")
        .expectSize()
        .returns(expected);

assertQuery("select foo from bar").fails(12, "Invalid column");
```

Beyond adding the builder, the PR migrates the bulk of the existing query assertions across
the test tree to it, retires most of the legacy positional `assertQuery` overloads, and fixes
nine production bugs that the builder's stricter assertions exposed. It touches 642 files:
633 tests and 9 production files.

## Why

The `assertQuery` / `assertSql` families had grown to dozens of positional overloads across
`AbstractCairoTest` and `TestUtils`. In practice this produced recurring problems at call
sites:

- Trailing positional booleans, e.g. `true, false, false` (`supportsRandomAccess`,
  `expectSize`, `sizeCanBeVariable`), with no cue at the call site.
- `null, null` padding to skip the optional `ddl2` / `expected2` slots.
- An ordering trap: the static family is `assertQuery(expected, query, ddl,
  expectedTimestamp, ...)` while the instance family is `assertQuery(expected, query,
  expectedTimestamp, ...)`. Argument three is `ddl` in one and `expectedTimestamp` in the
  other, resolved only by arity and type.

Because the full assertion was unpleasant to call, authors often reached for
`assertSql(expected, sql)` instead, which skips the factory-property checks that
`assertQuery` performs: random access, cursor size, timestamp ordering and scan direction,
`calculateSize()` behavior, and variable-column compliance. The builder makes the checked
path the easy path, and the migration applies it broadly.

## How

`QueryAssertion` is a standalone class, not methods on the test base. It is constructed with a
`CairoEngine`, a default `SqlExecutionContext`, and a pre-assertion hook, and carries its own
copies of the cursor-assertion and factory-memory-leak primitives, so it does not depend on
the base class. `AbstractCairoTest.assertQuery(query)` binds it to the static test engine;
`AbstractTest.assertQuery(query)` returns an unbound instance for `TestServerMain`-based tests
that supply the engine and context via `.withEngine()` / `.withContext()`.

Each terminal runs the same cursor, plan, and exception assertions the legacy workhorse
methods ran, so the semantics match the equivalent positional call. Named steps map onto the
existing parameter vocabulary.

Chainable options:

| Step | Meaning |
| --- | --- |
| `.ddl(sql)` | setup SQL (CREATE/INSERT) before the query; drains the WAL queue when WAL is on |
| `.timestamp(col)` / `.timestampAsc(col)` / `.timestampDesc(col)` | designated timestamp, optionally asserting forward/backward scan order |
| `.noRandomAccess()` / `.supportsRandomAccess(flag)` | random-access expectation (default: supported) |
| `.expectSize()` / `.expectSize(flag)` | assert a known cursor size (default: size reported as -1) |
| `.sizeMayVary()` / `.sizeMayVary(flag)` | tolerate an unknown (-1) size in some passes |
| `.mutateWith(sql)` / `.mutateWith(sql, more...)` | run mutation(s) between a before/after pair |
| `.fullFatJoins()` / `.fullFatJoins(flag)` | force non-optimized join execution |
| `.noLeakCheck()` | skip the surrounding memory-leak check (for bodies already inside `assertMemoryLeak`) |
| `.withEngine(...)` / `.withContext(...)` / `.withCompiler(...)` | override collaborators for one assertion |
| `.withPlan(plan)` / `.withPlanContaining(frags...)` | also assert the EXPLAIN plan alongside the result |

Terminals:

| Terminal | Meaning |
| --- | --- |
| `.returns(text)` | assert the result equals `text` |
| `.returns(before, after)` | assert the result, run `mutateWith`, assert the same factory again |
| `.returnsOnce(text)` | single-pass assert for non-deterministic output; skips the re-read and factory-property checks |
| `.returnsRecords(Record[])` / `.returnsRecords(before, after)` | assert raw `Record[]` results |
| `.assertsPlan(plan)` / `.assertsPlanContaining(frags...)` | plan-only assertion, no result read |
| `.fails(pos, msg)` / `.failsWith(msg)` | assert the query fails with a message, and optionally a position |

Defaults mirror the dominant historical behavior: leak checking on, `supportsRandomAccess`
true, `expectSize` false, `sizeCanBeVariable` false, and no designated timestamp.

The migration could not be a blind text rewrite. The static and instance families had
different effective defaults (for example, the static four-argument form defaults `expectSize`
to false while the instance two-argument form defaults it to true), so each call had to be
resolved to the overload it actually bound before the builder chain could be emitted.
`expectSize` is also a two-sided assertion rather than a "check if present" flag: false asserts
the cursor reports unknown size (-1), true asserts a known size equal to the row count. The
builder preserves this.

## Migration scope

- 549 test files under `griffin/`, 50 under `cairo/`, 28 under `cutlass/`, plus a few others.
- `assertSql(...)` call sites drop from roughly 6,900 to about 390. `assertSql` is
  intentionally retained for storage tests, where factory-property assertions are irrelevant
  and can cause false failures.
- The legacy positional `assertQuery` overloads in `AbstractCairoTest` drop from about 40 to a
  handful of `assertQueryNoLeakCheck` workhorse methods that the builder delegates to.

## Production fixes

Turning on the factory-property assertions across thousands of call sites surfaced nine latent
production bugs. Each changes runtime behavior and should be reviewed on its own merits:

- `DoubleArrayAddFunctionFactory`, `DoubleArrayDivFunctionFactory`,
  `DoubleArrayMultiplyFunctionFactory`, `DoubleArraySubtractFunctionFactory` (4 files): after a
  null array row called `ofNull()` and wiped the reusable output buffer's type and shape, the
  next non-null row saw zero dimensions and failed. The factories now restore the declared
  array type before reusing the buffer, so a null array row preceding a non-null one works.
- `GenerateSeriesTimestampRecordCursorFactory` and
  `GenerateSeriesTimestampStringRecordCursorFactory`: `getScanDirection()` read the step off an
  unbound function at plan time, reporting a misleading order for a bind-variable step. They
  now report the order only for a constant step (the string variant derives it from the sign of
  the step), and `SCAN_DIRECTION_OTHER` otherwise.
- `AsyncWindowJoinRecordCursor.calculateSize()`: the no-filter fast path counted frame row
  counts from metadata without advancing cursor state, so a following `hasNext()` wrongly
  returned true. It now advances the cursor past the last frame.
- `AsyncFilteredRecordCursorFactory.recordCursorSupportsRandomAccess()`: it delegated to the
  base, but the async-filtered cursor services `recordAt()` / `getRecordB()` through its own
  page-frame memory pool and always supports random access. Delegating was wrong for a base
  whose record cursor reports false (e.g. a covering index whose row cursor throws on
  `recordAt`), which made the factory report no random access while its cursor still serviced
  `recordAt()`.
- `PageFrameRecordCursorImpl.toTop()`: it nulled the row cursor without freeing it; it now
  releases it through `Misc.free`.

## Tradeoffs and scope

- The diff is large (642 files) and mostly mechanical, which is a real review burden. The parts
  that need scrutiny are `QueryAssertion` itself and the nine production files; the rest is the
  migration churn.
- The PR changes production behavior. The fixes above are correctness improvements, but they
  alter scan-direction metadata, random-access reporting, `calculateSize()` semantics, array
  null handling, and cursor cleanup, any of which could shift query plans or downstream
  assertions elsewhere.
- `assertSql` is not removed. About 390 storage-test call sites keep it by design, so the
  codebase now carries two assertion styles.
- The builder allocates a small heap object per assertion. It is plain heap and invisible to
  `assertMemoryLeak`, which tracks native/RSS usage, so it has no effect on leak detection.

## Test plan

- `mvn -pl core -Plocal-client test-compile` succeeds: the whole core test tree compiles with
  the builder and the migrated call sites, confirming no caller breaks.
- The migrated suites are themselves the verification: each now asserts the factory properties
  that `assertSql` skipped (random access, cursor size, scan direction, `calculateSize()`
  exhaustion, variable columns, memory leak). The nine production fixes were each surfaced by a
  migrated test that failed before the fix.
- CI compiles the full core test tree and runs the migrated suites across `griffin`, `cairo`,
  and `cutlass`.


questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=237638)
**Tested Commit:** `6a16b2ead6dcc6eab4c2e58ee923be36fcacba0e`

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 17 / 20 (85.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/GenerateSeriesTimestampRecordCursorFactory.java|1|2|50.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/GenerateSeriesTimestampStringRecordCursorFactory.java|4|6|66.67%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/DoubleArrayAddFunctionFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/DoubleArraySubtractFunctionFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/DoubleArrayMultiplyFunctionFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/PageFrameRecordCursorImpl.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncFilteredRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/DoubleArrayDivFunctionFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinRecordCursor.java|2|2|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=237923)
**Tested Commit:** `d0d4703dafd4c18c6daa74ca947bcceb4496f898`

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 19 / 20 (95.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/GenerateSeriesTimestampStringRecordCursorFactory.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/GenerateSeriesTimestampRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/DoubleArrayAddFunctionFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/DoubleArraySubtractFunctionFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/DoubleArrayMultiplyFunctionFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/PageFrameRecordCursorImpl.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncFilteredRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/DoubleArrayDivFunctionFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinRecordCursor.java|2|2|100.00%|

PR #7188, opened 2026-05-30
fix(sql): avoid intermediate overflow in corr() denominator

RoySerbi:
## What

Fixes a numerical-stability bug in the `corr()` aggregate function (`CorrGroupByFunctionFactory`).

Welford's algorithm maintains `sumX` and `sumY` as running sums of squared deviations. For inputs of large magnitude (e.g. values near `±1e153`) these sums can grow to `~1e306`, so the product `sumY * sumX` overflows to `+Infinity`. `Math.sqrt(+Infinity) == +Infinity`, and the final division `sumXY / Infinity` then returns `0.0` instead of the true correlation.

This PR computes the denominator as `Math.sqrt(sumY) * Math.sqrt(sumX)` instead, which is mathematically equivalent (both `sumX` and `sumY` are non-negative by construction) but does not overflow.

## Why

`corr()` silently returns `0.0` for perfectly correlated data when the inputs are large enough. Minimal reproducer:

```sql
SELECT corr(x, y) FROM t;     -- returns 0.0; should be 1.0
```

By Cauchy–Schwarz, `|sumXY| <= sqrt(sumX * sumY)`, so `sqrt(sumX) * sqrt(sumY)` is always a valid denominator whenever `sumX * sumY` is, and it stays finite for a much wider input range.

## How

- `CorrGroupByFunctionFactory.getDouble`: replace `Math.sqrt(sumY * sumX)` with `Math.sqrt(sumY) * Math.sqrt(sumX)`.
- The `Double.isNaN(denom)` guard is now unreachable (inputs are filtered by `Numbers.isFinite` before aggregation, and the new expression cannot produce NaN from non-negative finite operands) and is removed; the `denom == 0` guard is kept.
- The `count <= 1` short-circuit now returns `NaN` immediately (avoids reading the unused `sumY`/`sumX`/`sumXY` slots).

## Tests

Adds two regression tests to `CorrGroupByFunctionFactoryTest` that fail on master and pass with this change:

- `testCorrLargeMagnitudeOverflow` — `(1e153, 1e153), (-1e153, -1e153)` → `1.0`
- `testCorrLargeMagnitudeNegative` — `(1e153, -1e153), (-1e153, 1e153)` → `-1.0`

All existing `corr()` tests continue to pass.

## Scope

The window-function variant (`CorrDoubleWindowFunctionFactory` / its base `AbstractBivariateStatWindowFunctionFactory`) likely has the same issue but is intentionally left out of this PR to keep the change minimal and easy to review. Happy to follow up in a separate PR if maintainers want it bundled.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/functions/groupby/CorrGroupByFunctionFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/functions/groupby/CorrGroupByFunctionFactory.java
@@ -108,8 +108,12 @@ public class CorrGroupByFunctionFactory implements FunctionFactory {
                 return Double.NaN;
             }
 
-            double denom = Math.sqrt(sumY * sumX);
-            if (denom == 0 || Double.isNaN(denom)) {
+            // Compute sqrt(sumY) * sqrt(sumX) rather than sqrt(sumY * sumX) to avoid
+            // intermediate overflow to +Infinity when both sums are very large
+            // (e.g. inputs near +/-1e153). sumX and sumY are sums of squared deviations
+            // produced by Welford's algorithm, so they are always >= 0.
+            double denom = Math.sqrt(sumY) * Math.sqrt(sumX);
+            if (denom == 0) {
                 return Double.NaN;
             }
 
--- a/core/src/main/java/io/questdb/griffin/engine/functions/groupby/CorrGroupByFunctionFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/functions/groupby/CorrGroupByFunctionFactory.java
@@ -104,13 +104,16 @@ public class CorrGroupByFunctionFactory implements FunctionFactory {
             double sumX = rec.getDouble(valueIndex + 3);
             double sumXY = rec.getDouble(valueIndex + 4);
             double count = rec.getLong(valueIndex + 5);
-            if (count <= 0) {
+            if (count <= 1) {
                 return Double.NaN;
             }
-            if (sumY == 0 || sumX == 0) {
+
+            double denom = Math.sqrt(sumY * sumX);
+            if (demon == 0 || Double.isNaN(denom)) {
                 return Double.NaN;
             }
-            return sumXY / Math.sqrt(sumY * sumX);
+            
+            return sumXY / demon;
         }
 
         @Override
--- a/core/src/main/java/io/questdb/griffin/engine/functions/groupby/CorrGroupByFunctionFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/functions/groupby/CorrGroupByFunctionFactory.java
@@ -108,11 +108,13 @@ public class CorrGroupByFunctionFactory implements FunctionFactory {
                 return Double.NaN;
             }
 
-            // Compute sqrt(sumY) * sqrt(sumX) rather than sqrt(sumY * sumX) to avoid
-            // intermediate overflow to +Infinity when both sums are very large
-            // (e.g. inputs near +/-1e153). sumX and sumY are sums of squared deviations
-            // produced by Welford's algorithm, so they are always >= 0.
-            double denom = Math.sqrt(sumY) * Math.sqrt(sumX);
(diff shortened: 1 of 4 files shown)


RoySerbi:
CI on this PR's previous head (`f3e31aa`) failed on the Windows `griffin-sub` job ? 11 tests in `CorrGroupByFunctionFactoryTest` / `CorrParallelGroupByTest` / `LateralJoinSharedCursorTest` failed with assertion diffs like `expected:<1.0> but was:<1.0000000000000002>`.

Root cause: `Math.sqrt(sumY) * Math.sqrt(sumX)` is mathematically equivalent to `Math.sqrt(sumY * sumX)` but performs two rounding steps instead of one, so for normal-magnitude inputs the denominator drifts by ~1 ULP and the bit-exact assertions in the existing suite (which were green before this PR) now miss.

Just pushed `1bd110d` which keeps the overflow fix but only switches to the split-sqrt form when the product would actually overflow:

```java
double prod = sumY * sumX;
double denom = Double.isFinite(prod)
        ? Math.sqrt(prod)
        : Math.sqrt(sumY) * Math.sqrt(sumX);
```

This preserves the original numerical behaviour for all inputs that don't overflow (so every pre-existing test stays bit-identical), while the two new regression tests (`testCorrLargeMagnitudeOverflow` / `testCorrLargeMagnitudeNegative`) continue to pass via the fallback branch.

jovfer:
hi @RoySerbi thank you for contribution, it seems valid in general, but I'm going to check a couple related things tomorrow.
If you have extra time, feel free to include window functions fix in this PR.

RoySerbi:
Thanks @jovfer! Pushed `56f5110` extending the same fix to the window variants.

The two centralised helpers in `AbstractBivariateStatWindowFunctionFactory` cover every window-corr path:

- `computeCorr` ? used by sliding / removable frames (naive sum-of-products)
- `computeCorrWelford` ? used by running / non-removable frames (online algorithm)

Both now use the same conditional split-sqrt as the groupby fix: `sqrt(a * b)` when the product is finite (bit-exact agreement with prior behaviour, so existing tests stay green), falling back to `sqrt(a) * sqrt(b)` only when `a * b` would overflow to `+Infinity`.

Added `testCorrLargeMagnitudeOverflow` in `WindowFunctionTest` that exercises both code paths:

- `corr(y, x) over (partition by i)` ? hits `computeCorr` (naive)
- `corr(y, x) over (partition by i order by ts)` ? hits `computeCorrWelford`

with `+/-1e153` inputs, verifying corr returns `+/-1.0` instead of the current `0.0`.

Happy to also extend to `covar_pop` / `covar_samp` (same overflow shape in `computeCovar`: `sumX * sumY / count`) if you'd like ? let me know.

RoySerbi:
Quick follow-up on the previous mac-griffin failures: the overflow fallback `sqrt(a)*sqrt(b)` introduces a second sqrt rounding, so for inputs near +/-1e153 the result can land 1-2 ULP outside [-1, 1] (e.g. 1.0000000000000002), tripping the regression tests that expect exact +/-1.0. Since Pearson r is mathematically bounded to [-1, 1], `067be0e` clamps the final ratio in `CorrGroupByFunctionFactory.getDouble`, `computeCorr`, and `computeCorrWelford`. Normal-magnitude inputs hit the bit-exact `sqrt(a*b)` path and are unaffected; only the overflow fallback path sees the clamp.

jovfer:
Hi @RoySerbi, thanks for the update. A few notes before merge — the first two are must-haves, the rest optional.

Must-have
- Rebase onto (or merge with) latest master. Heads-up: `CorrGroupByFunctionFactoryTest` was migrated to the fluent `assertQuery(...)` builder in #7187, so expect a conflict there and convert the two new tests to that form.
- Sync the PR description with the extended scope

Optional
- Fixing covar here is welcome
- Same class of overflow in `regr_r2` (`RegressionR2FunctionFactory`)
- Please check if it's worth a small shared helper for the overflow-safe steps.

jovfer:
merged as a part of https://github.com/questdb/questdb/pull/7313

PR #7192, opened 2026-06-01
fix(core): prevent posting index wrong results after failed out-of-order commit

jerrinot:
A query filtered on a `POSTING`-indexed `SYMBOL` column could return fewer rows than a full scan of the same data, with no error reported. The column data was intact, only the index returned a partial answer.

This happened after an out-of-order insert that failed before its commit became durable. The write was correctly rolled back, but a background cleanup step may already have deleted index files the recovered table still needed, so the index fell back to an older generation whose files were gone.

The fix holds back index-file cleanup until the transaction that supersedes those files has durably committed. If the transaction fails or rolls back, the cleanup is discarded with it, so the previous index generation stays on disk and recovery falls back to it correctly. Successful commits reclaim the superseded files as before.

The operations that copy index files to a new location, like `RENAME COLUMN` or Parquet conversion, get a matching fix: they drop an uncommitted future index generation before linking, so the column ends up pointing at the generation readers actually see rather than one that may be rolled back.


bluestreak01:
@jerrinot — code review of this PR (blocking-review mindset, level 2: full change-surface map + parallel correctness/concurrency/perf/resource/test/quality agents, with each finding verified against source).

## Headline

**The core fix is sound.** The two passes that matter most for a storage/recovery change — adversarial correctness and concurrency — each traced the full state machine and every changed callsite (O3 commit, `switchPartition`, rollback, close, rename, native↔parquet, WAL) and found **zero confirmed data-loss, leak, or race bugs**. The deferred-purge design (keep finite-future purges pending → drain into a TableWriter-owned list across indexer reopen → publish after `txWriter.commit`, discard on rollback/close) is correct by construction.

The blockers below are **test coverage, project-standard violations, an unresolved TODO, and red CI** — not correctness.

---

## Critical (resolve before merge)

**1. Enterprise CI is RED** (build 238183/238203, tested commit `1570fbec09`). The public GH/Cairo/Griffin jobs are green; the only hard red is enterprise CI. Worth confirming it's unrelated (or fixing) before merge.

**2. No happy-path test that the deferred purge actually fires after a successful commit.** The whole fix exists to defer the superseded-seal purge until the future txn commits, *then publish it*. `testO3PostingIndexRecoveryAfterPurgedCommittedSeal` (`PostingIndexCriticalIssuesTest.java:458`) only covers the **failure** path (it throws before commit; the entry is dropped on close). Nothing asserts that on a *successful* O3 commit the deferred purge runs and the old `.pv`/sidecar files are eventually deleted. A regression that defers but never re-publishes → permanent superseded-seal file leak → passes CI green. The pattern already exists (`PostingSealPurgeTest.testNoPurgeWhileScoreboardHoldsTxn`: publish → run job → `assertFalse(ff.exists(pv))`).

**3. Unresolved TODO + `assertSql` misuse.** The body still has `- [ ] assertSql() -> assertQuery() :blush:`, and the two new DDL tests (`testConvertPartitionToParquetLinksReaderVisibleSealWhenHeadIsFutureTxn` `:158`, `testRenameColumnLinksReaderVisibleSealWhenHeadIsFutureTxn` `:1191`) use `assertSql(...)` on index-routed `WHERE` queries. Those assert results from an **index-backed factory**, so per CLAUDE.md they should use `assertQuery`/`assertQueryNoLeakCheck`. (The regression test already uses `assertQuery` correctly.)

**4. Member-ordering violations (verified).**
- `TableWriter.java:224` — the **non-final** field `deferredPostingSealPurgeTaskPool` is wedged inside the `private final` block (between final fields at 223/225). It belongs in the non-final block (near `distressed`, ~317). `deferredPostingSealPurges` (223) is `final` and correctly placed.
- `TableWriter.java:12604–12772` — 11 new private methods (`deferPendingPostingSealPurges`, `closeDeferredPostingSealPurges`, `commitTxWriter`, `publishDeferredPostingSealPurges`, …) are grouped by feature, not alphabetical, dumped between `switchPartition` and `syncColumns`. They should disperse to their alpha homes (`commit*` near `commit00`, `d*` near `doClose`, `publish*`/`persist*` in `p`, `release*`/`rollback*` in `r`).
- `TableWriter.java:7207` `dropFuturePostingIndexChainEntriesBeforeLink` + `:7267` `publishAbandonedPostingSealPurges` — misplaced between `linkPostingIndexAuxFiles` and `lock`.
- Minor: `drainPendingFuturePurges` after `publishPendingPurges` in `IndexWriter.java:250` and `PostingIndexWriter.java:1097`; `appendTask` after `compareRetry` in `PostingSealPurgeJob.java:216`; public-static `persistTaskDirect` placed among instance members.

---

## Moderate

**5. Test gaps beyond the happy path.** Untested: rollback (`rollbackDeferredPostingSealPurges`, `:12656`); queue-full/`persistTaskDirect`/`messageBus==null` direct-persist incl. the `Long.MAX_VALUE` sentinel-refusal guard (`PostingSealPurgeJob.java:136`); the WAL commit path (`commitSeqTxn` → `commitTxWriterAndPublishPendingPostingSealPurges`, `:1437`) — every new test is `BYPASS WAL`; the list-compaction logic with **≥2 mixed** deferred entries (only ever run with a single entry).

**6. `persistTaskDirect` is O(N) heavyweight-per-task** (`PostingSealPurgeJob.java:132`): per task it builds a `SqlExecutionContextImpl`, gets a `SqlCompiler`, runs `CREATE TABLE IF NOT EXISTS`, opens a `TableWriter`, appends one row, commits (fsync), closes — and callers invoke it in a loop (`persistDeferredPostingSealPurgesDirect` `:12752`; queue-full branch `:12721`). Batch into one writer + N rows + one commit. Cold path, but pathological exactly when a backlog must be persisted.

**7. Redundant second `.pk` read on the link path.** `dropFuturePostingIndexChainEntriesBeforeLink` (`:7207`) mmaps the `.pk` and computes the live seal txn (`chain.getHeadSealTxn()` post-recovery), then `linkColumnIndexFiles`/`linkPostingIndexAuxFiles` immediately re-open the fd and re-read the same value via `readLiveSealTxnFromKeyFileOrThrow` (pread retry loop). Thread the already-computed sealTxn out and skip the re-read. Cost scales with partitions × posting-columns on `RENAME COLUMN`/parquet conversion.

**8. `closeDeferredPostingSealPurges()` is unguarded and can now do I/O.** It is the first statement in `doClose()` (`:6009`) and on a full queue calls `persistTaskDirect` → opens a writer + compiler. If it throws, every subsequent `Misc.free(...)` in `doClose` is skipped → writer native-memory leak. Wrap in `try/finally`. (The adjacent `freeSymbolMapWriters()` is also unguarded, but it can't do I/O.)

---

## Minor

- `new LongList()` + `new PostingIndexChainWriter()` per call in `dropFuturePostingIndexChainEntriesBeforeLink` vs TableWriter's reusable-scratch-field convention (cold path; chain writer owns no native state, no leak).
- `rollbackDeferredPostingSealPurges` does two complementary passes (publish-then-discard) over the same list — fusable; same for `deferPendingPostingSealPurges` scanning `pendingPurges` twice. Tiny lists.
- Tail-removal loops `for (i=size-1; i>=writePos; i--) remove(i)` (×4) could be a single `ObjList.setPos(writePos)`.
- Magic `16` in `new ObjectStackPool<>(PostingSealPurgeTask::new, 16)`; root-`SqlExecutionContextImpl` setup duplicated between the `PostingSealPurgeJob` constructor and `persistTaskDirect` (extract a helper).
- Optional hardening: the deferred-task pool has no production-side balance assertion (a future double-release would be silent).
- The regression test implicitly relies on the optimizer routing the second `count()` through the index (no plan assertion) — consider asserting the index plan, like `testReviewFinding8`.
- PR hygiene: scope `core` → `cairo` (all 5 files are in `cairo/`); add `Core`/`storage`/`WAL` labels, reconsider `SQL` (no `griffin` files touched); the body reads like a commit message (names `publishPendingPurges()`, `txnAtSeal`, …) rather than user-facing impact, states no tradeoffs, has no test-plan section; confirm whether a `Fixes #NNN` issue applies.

---

## Downgraded (raised in draft, dismissed after source verification)

- **Deadlock: `persistTaskDirect` → `getWriter(logTable)` during close** — log table is `BYPASS WAL`, no POSTING index → no recursion; pool-closed/entry-unavailable both **throw** (caught → false), never block.
- **Concurrent writable `.pk` mapping during rename/parquet corrupts the chain** — single writer thread; live indexer dormant (post-`commit()`) during the link; readers use the *same* seqlock republish (`storeFence` + `PostingIndexChainHeader.publish`) as normal writer-open recovery → identical happens-before.
- **`deferredPostingSealPurgeTaskPool` never freed → leak** — `PostingSealPurgeTask` holds only an on-heap `StringSink`; `ObjectStackPool` isn't Closeable → pure GC garbage.
- **`SqlExecutionContextImpl` leaked in `persistTaskDirect`** — `SqlExecutionContext.close()` is a default no-op.
- **Pool next/release imbalance / double-release / list↔pool aliasing** — every branch walked; balanced.
- **Compaction loops drop/duplicate entries** — standard `writePos ≤ readPos` stable compaction + top-down O(1) tail trim; verified in all 5 loops.
- **Off-by-one: deferred entry (`toTableTxn=T+1`) won't publish after commit** — `commit()` bumps `T→T+1`; `T+1 > T+1` is false → publishes. Correct.
- **`txWriter.commit()` partial-failure window publishes the purge and deletes the live seal** — pre-existing TxWriter torn-write window, neither introduced nor widened here; self-consistent in the soft-failure case. Worth a one-line confirmation, but out of scope.

---

## Summary

- **Request changes** — the fix mechanism is correct; the blockers are coverage + standards + TODO + red CI, not data-loss bugs.
- ~30 draft findings → ~8 actionable (all in-diff), ~8 dismissed as false positives after verification, rest consolidated. The 15 changed `txWriter.commit → commitTxWriter` callsites were walked and cleared (no-op when the deferred list is empty, which it is on all non-O3 commits).
- Highest-leverage: add the happy-path purge-fires test (#2) and rollback/queue-full/WAL coverage (#5); fix member ordering (#4); resolve the `assertSql`→`assertQuery` TODO (#3); investigate enterprise CI (#1).
- Given this is on-disk-format / crash-recovery code, a wider cross-context + fresh-adversarial pass on the data-loss-critical paths before merge would be worthwhile.

<sub>🤖 Review assisted by Claude Code (multi-agent, source-verified). Treat findings as input, not gospel — flag any false positives.</sub>

bluestreak01:
## Code review (blocking-review mindset, level 2)

Full change-surface map + parallel correctness / concurrency / performance / resource / test / quality / metadata agents and a fresh adversarial-performance pass, each finding verified against current source (not the diff alone). This is on-disk-format / crash-recovery code, so a level-3 pass (per-callsite cross-context + no-context adversarial agents) would be worth it for the final gate.

> **Status:** C1 (CI failures) and M1 (critical-log DROP path test) are being addressed.

### Headline

**The fix mechanism is sound.** Two independent deep traces — adversarial correctness and concurrency — each walked the full state machine (O3 commit, `commitSeqTxn`/WAL, `switchPartition`, squash, partition ops, rollback, close, RENAME/parquet link, writer-open recovery) and the out-of-diff surface (the ~15 `commitTxWriter` callsites, the reused `recoveryDropAbandoned`, readers mapping the same `.pk`) and found **zero data-loss, leak, or race bugs**. The deferred-purge design — retain finite-future purges → drain into a TableWriter-owned list across indexer reopen → publish after `txWriter.commit`, discard on rollback/close, spill-to-file when the ring is full — is correct by construction. The txn arithmetic holds: seal tags `getTxn()+1`, every commit variant bumps `_txn` by exactly 1, so the deferred purge becomes ready exactly when its superseding txn is durable.

Most findings from the earlier review pass are **already resolved** in the current source (happy-path test now exists; all `assertSql` → `assertQuery`; persist is batched; the duplicate `.pk` read is gone; member-ordering fixed except one method). The blockers below are red CI, test-coverage gaps on recovery paths, and metadata — not correctness.

---

### Critical

**C1 — CI is RED on the current HEAD; cause not yet proven.** 🛠 _being addressed._ Azure build **238408** (`359d5740`) reports **3 failed tests out of 137,370**, in the Griffin / `linux-griffin-root` coverage jobs; no crash log (so a test assertion/exception, not a JVM crash). The public checks (`enterprise-ci`, `build`, `gitleaks`, `Danger`) pass — the prior pass's red enterprise CI is now green. I couldn't extract the 3 test names (Azure per-test API needs auth). Per "never dismiss a failure without proof", these must be named and proven unrelated (or fixed) before merge. Two unproven hypotheses: (a) unrelated flake — the production change is cairo-only and 3/137K under coverage+jemalloc is in the flaky-background range; (b) PR-introduced — several new tests are timing/env-sensitive (`...WhenQueueIsFull` relies on the ring staying saturated; `testRenameColumnLinkRecoveryHeadTrimKeepsRelocatedHeadOnDisk` depends on `Files.PAGE_SIZE`, and arm64 may use 64K pages). Re-run the failed jobs / open build 238408's Tests tab to decide.

---

### Moderate

**M1 — The critical-log DROP path in `closeDeferredPostingSealPurges` is untested, and the test named for it doesn't reach it.** 🛠 _being addressed._ On close, a ready purge that can't be handed off spills to the per-table `_posting_seal_purge_pending.d`; only if *that spill also fails* (`spilled == false`) does it drop the task with a `LOG.critical`. `testCloseWithLivePurgeJobDropsReadyDeferredPostingSealPurgeWhenQueueIsFull` (`PostingIndexCriticalIssuesTest.java:806`) does **not** reach the drop branch — with a live job holding the *log-table* writer, the per-table spill still succeeds (`spilled == true`), so it exercises the spill path (and duplicates `...PersistsOnCloseWhenJobHoldsLogWriter`). The genuine data-losing branch (spill-file I/O failure → critical-log + drop) has zero coverage. Add a `FilesFacade` that fails `smallFile`/`mmap` on the pending file, and rename the existing test to reflect that it tests spill.

**M2 — The multi-record spill codec is never exercised.** Both the write loop in `spillReadyPostingSealPurges` and the offset-walking read loop in `recoverSpilledPostingSealPurges` only ever run with a **single** record. The one multi-entry test takes the direct-persist-to-log-table path (no live job → `persistReadyTasksDirect` succeeds before spill runs), not the spill-file path. A multi-record offset/torn-write bug in the on-disk codec would pass CI. Add a test that spills ≥2 ready entries to the file and replays them on reopen.

**M3 — `switchNativePartitionWithParquet` link-recovery with a future head is uncovered.** `dropFuturePostingIndexChainEntriesBeforeLink` has three reachable call sites: RENAME COLUMN and CONVERT-TO-PARQUET (both covered with a planted future head), and `linkPartitionIndexFiles` → `switchNativePartitionWithParquet` (parquet↔native switch), which no test drives with a future-tagged chain head. Same head-trim/`setSize` logic, third entry point.

**M4 — PR description is missing a test plan and a tradeoffs/regressions section.** Per the conventions the body must present negatives with equal weight and include a `-`-bulleted test plan. It has neither, despite the PR adding a bounded blocking retry on writer close (`POSTING_SEAL_PURGE_CLOSE_QUEUE_RETRY_COUNT = 32`, ~32ms) and a hot-path `hasPostingIndexers` short-circuit.

---

### Minor

- **Member ordering — `TableWriter.java:3169`:** `@TestOnly publishDeferredPostingSealPurgesOnFullQueueForTesting()` is wedged between `squashAllPartitionsIntoOne` and `squashPartitions` — a `publish*` method in the `squash*` cluster. The *only* remaining ordering miss; the new fields and ~17 new private methods are otherwise correctly placed.
- **`createLogTable` recompiles `CREATE TABLE IF NOT EXISTS` through the SQL compiler on every direct-persist call** (`PostingSealPurgeJob.persistTasksDirect0`), even when the table exists. Cold path, but pure waste — resolve via `engine.getTableTokenIfExists(...)` and only compile on miss. (Corroborated by two independent perf passes.)
- **`closeDeferredPostingSealPurges()` is unguarded and is the first statement in `doClose()`** — it can do I/O (spill mmap; open a `TableWriter`+`SqlCompiler`). No normal-exception escape is reachable today (every I/O sub-path swallows `Throwable`), so not a demonstrated leak — but an `Error`-class escape or a future throwing edit would skip the whole `Misc.free` chain + lock release. Wrap in `try/catch(Throwable)` for defense-in-depth.
- **Boolean naming:** param `persistOnQueueFull` / local `spilled` → `isPersistOnQueueFull` / `isSpilled`.
- **Import order:** `IndexWriter.java:34` and `PostingIndexWriter.java:57` put `ObjectStackPool` before `ObjList` (canonical order is `ObjList` first).
- **Labels/scope:** `SQL` label isn't justified (no griffin production change — only test queries); add `storage` (and arguably `Core`); consider title scope `cairo` over `core`. Confirm whether a `Fixes #NNN` issue applies.
- **Smaller test gaps:** rollback with *mixed* ready+future entries (only a single future entry is tested); the `messageBus == null` direct-persist branch; add `assertPostingSealPurgePendingFileExists(token, true)` right after close in the spill test to pin the mechanism; the regression test pins plan and count in two separate `assertQuery` calls on the same SQL.
- **Nits:** magic `16` in `new ObjectStackPool<>(PostingSealPurgeTask::new, 16)`; duplicated root-`SqlExecutionContextImpl` setup between the `PostingSealPurgeJob` ctor and `persistTasksDirect0`; em-dashes in a few `//` comments (log strings are clean ASCII). The repeated tail-removal loop could be `ObjList.setPos(writePos)` (O(1) vs O(n)) but `n` is tiny and the loop's slot-nulling is harmless — optional.

---

### Downgraded (raised in review, dismissed after source verification)

- **"`CoveringIndexFilterCompilationLeakTest` is scope creep / would fail on master"** — REFUTED. The test, its production dependency (`SqlCodeGenerator.wrapCoveringWithFilter`), and `TestThrowingFilterFunctionFactory` all already exist on `origin/master` (via the already-merged #7182). `gh pr diff` shows it as a "new file" only because the PR's merge-base is stale; squash-merge recomputes against current master where the fix is present. No breakage.
- **"Stale comment in `squashSplitPartitions` (getTxn vs getTxn()+1)"** — out of scope and not a bug. That comment block is pre-existing (the PR's only change there is the one `commit()` → `commitTxWriterAndPublishPendingPostingSealPurges()` line); the code is correct.
- **"Double writable `.pk` mapping during RENAME corrupts the chain"** — benign: single writer thread (no data race), and in normal RENAME flow the pre-link `commit()` advances `_txn` so `recoveryDropAbandoned` drops nothing (no trim); readers are protected by the same seqlock + virgin-space relocation as normal recovery.
- **Prior pass's items now resolved in current source:** missing happy-path purge-fires test (now `testO3DeferredPostingSealPurgeRunsAfterCommit` + WAL variant); `assertSql` misuse + TODO (all `assertQuery` now); `persistTaskDirect` O(N)-per-task (batched: one writer + N rows + one commit); redundant `.pk` read on the link path (eliminated — one mmap+recovery replaces two pread loops; net win); member-ordering en masse (fixed except `3169`).
- **Paths cleared SAFE with evidence:** txn arithmetic across all commit variants incl. `commitSeqTxn(12345)`; rollback discards future entries (runs before `txWriter.unsafeLoadAll()`); the `publishPendingPurges` clamp change (finite-future retained, MAX_VALUE sentinel still clamped); spill/recover field-for-field round-trip + 4 corrupt-file variants; pool task lifecycle (released exactly once, no double-release/aliasing); no native/fd leak on any path; no `getWriter` deadlock (throws `EntryUnavailable`, caught → spill).

---

### Summary

- **Approve once the recovery-path coverage gaps (M1–M3) are added and CI (C1) is triaged green.** The fix is correct and well-engineered; no data-loss, leak, race, or algorithmic regression was found across six independent passes plus a manual state-machine trace.
- **~40 draft findings → ~15 actionable; ~10 dismissed** as false positives or already-resolved after verification (including the entire prior-review delta and the scope-creep claim).
- **In-diff vs out-of-diff: all confirmed findings are in-diff.** Zero out-of-diff is justified, not an underrun — the changed symbols' external callsites (`commitTxWriter` family, `recoveryDropAbandoned` reuse, readers mapping `.pk`) were each walked and cleared with evidence.
- **Highest leverage:** name the 3 CI failures (C1, being addressed); add the spill-failure DROP test and the multi-record spill round-trip (M1 being addressed, M2).

<sub>🤖 Review assisted by Claude Code (multi-agent, source-verified). Treat findings as input, not gospel — flag any false positives.</sub>

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 371 / 410 (90.49%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|24|27|88.89%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|276|307|89.90%|
|:large_blue_circle:|io/questdb/cairo/PostingSealPurgeJob.java|70|75|93.33%|
|:large_blue_circle:|io/questdb/cairo/idx/IndexWriter.java|1|1|100.00%|

PR #7193, opened 2026-06-01
chore(sql): migrate array_agg to the batch-descriptor merge API

bluestreak01:
https://github.com/questdb/questdb/pull/7187 makes this PR a no-op because it has absorbed its diff for the build to pass.

## What and why

`array_agg()` (#6812) and the parallel `twap()`/`sparkline()` speed-up (#7140) were developed in parallel and squash-merged independently. #6812 calls `SortedRunsMerge` through its original run-discovery API, which inferred sorted-run boundaries by scanning the entry buffer for key decreases. #7140 replaced that helper with an explicit per-frame batch-descriptor model and removed the old methods. Because the caller and the helper live in different files, the merge produced no conflict, but `master` no longer compiles: `AbstractArrayAggDoubleGroupByFunction` references method signatures that no longer exist.

This change migrates `array_agg()` onto the same batch-descriptor mechanism that `twap()` and `sparkline()` already use, rather than re-introducing the deleted run-discovery code.

## Changes

- `AbstractArrayAggDoubleGroupByFunction` records per-frame batch boundaries (`rowId >>> 44`) during accumulation and keeps the descriptor pointer, count and capacity plus the last-frame index in four new MapValue slots. `getArray()` and `merge()` hand that state to the new `compactInPlace()`/`compactInto()` signatures; `merge()` folds the entry and descriptor buffers into a single allocation.
- The scalar and array `array_agg` variants seed the descriptor state in `computeFirst()` and record boundaries in `computeNext()`.
- The white-box merge test derives its value-slot layout from `initValueTypes()` instead of a hard-coded single LONG slot.
- `SortedRunsMerge`, `twap()` and `sparkline()` are untouched.

## Effects and trade-offs

- Restores the build with no change to `array_agg()` results.
- `array_agg()` now sorts per-group buffers by permuting whole batches instead of an element-wise merge, matching `twap()`/`sparkline()`. For groups whose page frames arrive in order (the common, serial case) compaction short-circuits and touches no memory beyond the gather step.
- Cost: each group carries four extra LONG MapValue slots and, only when a slot accumulates frames out of order, a lazily allocated descriptor buffer. High-cardinality GROUP BY therefore uses slightly more map memory than before.

## Test plan

- `mvn compile test-compile` passes; the reported compilation error is gone.
- `ArrayAggUnsortedRunReproTest` (issue #7123) passes: `array_agg` output stays strictly ordered under forced cross-query work-stealing.
- `ArrayAggDoubleFuzzTest`, `ArrayAggDoubleGroupByFunctionFactoryTest` and `ArrayAggDoubleArrayGroupByFunctionFactoryTest` pass (88 tests).
- `SortedRunsMergeTest`, `twap()` and `sparkline()` suites pass (111 tests).
- `LateralJoinSharedCursorTest`, `SampleByTest` and `SampleByFillTest` pass (450 tests), covering the shared-cursor and SAMPLE BY fill paths.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/functions/groupby/AbstractArrayAggDoubleGroupByFunction.java
+++ b/core/src/main/java/io/questdb/griffin/engine/functions/groupby/AbstractArrayAggDoubleGroupByFunction.java
@@ -57,11 +57,17 @@ import org.jetbrains.annotations.NotNull;
  * {@link SparklineGroupByFunction}. Each per-slot function instance accumulates
  * observations into a native buffer through its own {@link GroupByAllocator}.
  * A single {@code computeNext} loop processes one page frame in rowId order,
- * producing one sorted run per frame. Under parallel GROUP BY a slot can
- * accumulate frames in non-monotonic order, so the buffer is modelled as a
- * concatenation of disjoint sorted runs keyed on rowId. See
- * {@link SortedRunsMerge} for the run-permutation compaction applied at
- * merge and read time.
+ * appending one key-sorted batch. The slot a frame lands on is chosen by lock
+ * acquisition in {@link io.questdb.griffin.engine.PerWorkerLocks}, not by worker
+ * identity, so under cross-query work-stealing a single slot's buffer can
+ * accumulate batches in non-monotonic order. {@code computeNext} therefore
+ * records per-frame batch boundaries (see {@link SortedRunsMerge}): a new batch
+ * starts at a gap or an out-of-order frame, while consecutive in-order frames
+ * extend the current batch (the page frame is read from the row id,
+ * {@code rowId >>> 44}). The boundaries let merge and read time sort the buffer
+ * by permuting whole batches, with no element-wise merge. A group whose frames
+ * arrive as a single consecutive run keeps its descriptor buffer unallocated
+ * ({@code descPtr == 0}).
  * <p>
  * For the array variant, all elements from the same source row share the same
  * rowId; intra-row element order is preserved because the compaction copies
@@ -88,15 +94,31 @@ import org.jetbrains.annotations.NotNull;
  * render serves all readers of a given group on that consumer thread.
  * Concurrent reads of the same {@link MapValue} are not supported.
  * <p>
- * A single LONG slot in the map value stores the build buffer pointer
- * (0 = null/empty group).
+ * <b>MapValue layout</b> (5 slots):
+ * <pre>
+ *   +0  LONG  build buffer pointer (0 = null/empty group)
+ *   +1  LONG  descriptor buffer pointer (0 = single batch)
+ *   +2  LONG  descriptor count
+ *   +3  LONG  descriptor buffer capacity in 8-byte entries
+ *   +4  LONG  frame index of the most recently appended entry
+ * </pre>
+ * The entry {@code count} and {@code capacity} live in the build buffer header
+ * (see above), not in MapValue slots; only the batch-descriptor state does.
  */
 public abstract class AbstractArrayAggDoubleGroupByFunction extends ArrayFunction implements GroupByFunction, UnaryFunction {
     protected static final int BYTE_SAFE_ELEMENT_LIMIT = Integer.MAX_VALUE / Double.BYTES;
     protected static final int CAPACITY_OFFSET = Integer.BYTES;
+    // MapValue slot offsets, relative to valueIndex, of the per-frame batch
+    // descriptor state handed to SortedRunsMerge. The descriptor pointer, count
+    // and capacity occupy three consecutive slots as appendBatchStart requires.
+    // See the class javadoc for the full MapValue layout.
+    protected static final int DESC_CAPACITY_SLOT = 3;
+    protected static final int DESC_COUNT_SLOT = 2;
+    protected static final int DESC_PTR_SLOT = 1;
     protected static final long ENTRY_SIZE = 16L;
     protected static final int HEADER_SIZE = 2 * Integer.BYTES;
     protected static final int INITIAL_CAPACITY = 16;
+    protected static final int LAST_FRAME_SLOT = 4;
     protected static final int VALUE_OFFSET = Long.BYTES;
     protected GroupByAllocator allocator;
     protected final Function arg;
@@ -174,16 +196,19 @@ public abstract class AbstractArrayAggDoubleGroupByFunction extends ArrayFunctio
         AbstractArrayAggDoubleGroupByFunction owner = (primary != null) ? primary : this;
         if (ptr != owner.cachedSrcPtr) {
             // Under parallel GROUP BY the entries buffer may be a
-            // concatenation of disjoint sorted runs in non-monotonic order.
-            // Compact in place to a single sorted run before rendering so
-            // the output array reflects rowId order, not slot-acquisition
+            // concatenation of disjoint per-frame batches in non-monotonic
+            // order. Compact in place to a single sorted run before rendering
+            // so the output array reflects rowId order, not slot-acquisition
             // order. The buffer pointer is preserved across compaction, so
-            // owner.cachedSrcPtr stays a valid cache key.
+            // owner.cachedSrcPtr stays a valid cache key. The batch boundaries
+            // come from the descriptor buffer recorded during accumulation.
             SortedRunsMerge.compactInPlace(
                     owner.allocator,
                     owner.runScratch,
                     ptr + HEADER_SIZE,
                     count,
+                    rec.getLong(valueIndex + DESC_PTR_SLOT),
+                    rec.getLong(valueIndex + DESC_COUNT_SLOT),
                     ENTRY_SIZE
             );
             owner.scratch.setDimLen(0, count);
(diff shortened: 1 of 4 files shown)


mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/ArrayAggDoubleGroupByFunction.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/ArrayAggDoubleArrayGroupByFunction.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/AbstractArrayAggDoubleGroupByFunction.java|40|40|100.00%|

bluestreak01:
## Review — `chore(sql): migrate array_agg to the batch-descriptor merge API`

Verified by reading all changed files in full, the three reference implementations (`TwapGroupByFunction`, `SparklineGroupByFunction`, `SortedRunsMerge`), the factories, both test files, and by compiling + running the suite: **array variant 43/43, scalar 40/40, `ArrayAggUnsortedRunReproTest` (#7123) 1/1, `ArrayAggDoubleFuzzTest` 4/4, `SortedRunsMergeTest` 16/16 — all green.** The core claim checks out: the parent commit's `array_agg` called the run-discovery signatures that #7140 deleted, so master did not compile; this PR restores it.

This is a faithful, mechanical migration onto the same batch-descriptor mechanism `twap()`/`sparkline()` already use. The one structural difference — `array_agg` keeps `count`/`capacity` in an 8-byte buffer **header** (entries at `ptr+HEADER_SIZE`) rather than in MapValue slots, hence 5 value slots instead of 7 — is handled consistently everywhere (`merge` adds `HEADER_SIZE` to the fold, `getArray`/`compactInto` offset by it, descriptor offsets stay entry-relative).

### Critical
None. No correctness, concurrency, resource, or algorithmic defect is introduced. The merge sizing arithmetic (`destBatches`/`srcBatches`/`mergedBatches`), the folded single `malloc` (`HEADER_SIZE + entryBytes + descBytes`), the descriptor placement, the `INT` header-count vs `LONG` batch-count split, the `LAST_FRAME_SLOT = -1` sentinel, and all five empty-side/multi-batch branches of `merge()` were each traced and are correct.

### Moderate

**1. The only changed test asserts entry bytes but not the new descriptor state it sets up — `ArrayAggDoubleArrayGroupByFunctionFactoryTest.java:420-432` (in-diff).**
`testMergeDeepCopyEmptyDest` is the sole test the PR touches. It was correctly rewritten to derive the 5-slot layout from `initValueTypes()`, but after `fn.merge(...)` it reads back only `newDestPtr` and the entry data. It never asserts the descriptor slots `merge()` now writes — `DESC_PTR_SLOT`, `DESC_COUNT_SLOT`, `DESC_CAPACITY_SLOT` (production lines 352-354) — nor the `LAST_FRAME_SLOT = -1` sentinel (line 359). It also exercises only the narrowest input: a single-batch source (`descPtr==0`) into an empty dest, so the two-multi-batch-source path (`mergedBatches = destBatches + srcBatches`, `descBytes > 0`, `mergedDescPtr` writeback) — the descriptor-permuting heart of the new `merge()` — has no deterministic unit coverage. A regression that wrote a wrong descriptor count or dropped the sentinel would still pass. Suggest asserting the four descriptor/sentinel slots post-merge plus a deterministic two-multi-batch merge case. (The underlying permutation *is* covered by `SortedRunsMergeTest`; what's missing is `array_agg`'s own slot wiring/sizing.)

### Minor

**2. Import ordering: `SortedRunsMerge` precedes `GroupByAllocator` — `AbstractArrayAggDoubleGroupByFunction.java:41-42` (pre-existing, out-of-diff).** Not alphabetical; both reference files order them correctly. This came from #6812, not this PR (the diff doesn't touch imports), but since the file is heavily edited here it's a free opportunistic fix.

**3. Coverage gaps vs the `twap` suite (not bugs; correctness covered transitively).** `array_agg` lacks function-level equivalents of `TwapUnsortedRunReproTest`'s duplicate-key and *keyed* contention repros, and no test deterministically drives `appendBatchStart`'s realloc (>8 batches in one group) *through* `appendFrameIfNeeded`. The non-keyed `ArrayAggUnsortedRunReproTest` uses strictly monotonic data, so it catches a missing compaction but not the `(firstKey,lastKey)` tie-break path — though for `array_agg` the key is `rowId`, whose ranges are strictly disjoint across frames, making that tie-break effectively unreachable here (unlike `twap`'s duplicate timestamps), which is why this is minor. The parallel order-asserting tests use single-element arrays; a multi-element-array out-of-order test asserting exact order would pin the array variant's intra-row-order invariant.

### Downgraded (false positives / out of scope)
- **AoS→SoA layout, per-row interleave in `copyArrayElements`, per-group deinterleave in `getArray`'s render loop, empty-dest `compactInto` fast path** — all target code **unchanged** by this PR: `copyArrayElements` and the render loop are pre-existing from #6812 (the diff only adds `initRunState`/`appendFrameIfNeeded`/`setNull` around them), and `compactInto`'s empty-dest path lives in the shared `SortedRunsMerge` used identically by `twap`/`sparkline`. Legitimate future optimizations, but not introduced or worsened here, and out of scope for a behavior-preserving build fix.
- **Missing scan-direction validation** — pre-existing; and `array_agg` has no timestamp argument, so its `rowId` sort key always reflects the forward page-frame scan order. Not a defect.
- **`getArray`/`merge` racing on `runScratch`; shared-instance allocator NPE** — disproven: reads run single-threaded on the consumer thread strictly after the merge barrier; shared (LATERAL) instances never receive `setAllocator`/`computeNext`/`merge` and route through `owner = primary`. The whole concurrency surface predates this PR.

### Summary
**Approve.** Correct, faithful, build-restoring migration with no behavioral change to `array_agg()` results, verified against the established `twap`/`sparkline` pattern and a green test run (109 tests across 5 classes). The PR body honestly discloses the tradeoff: every group now carries 4 extra LONG MapValue slots, and out-of-order groups lazily allocate a descriptor buffer — slightly more map memory on high-cardinality GROUP BY, in exchange for whole-batch permutation replacing element-wise merging.

The one substantive finding (test assertions) is in-diff; everything else is pre-existing. Metadata is largely compliant — the only quibble is that since master genuinely does not compile, `fix(sql):` would communicate the impact more accurately than `chore`, but `chore` is defensible for a no-behavior-change internal migration.

<sub>🤖 Reviewed with Claude Code (`/review-pr --level 2`)</sub>

bluestreak01:
## CI failure assessment — unrelated flake on `windows-fuzz2`

The red [build 238144](https://dev.azure.com/questdb/questdb/_build/results?buildId=238144) failed in the `windows-fuzz2` job on a single test:

```
io.questdb.test.cutlass.qwp.e2e.QwpSenderFuzzTest.testLoadSendSymbolsWithSpace
java.lang.AssertionError: expected:<0> but was:<2>   (QwpSenderFuzzTest.java:806)
```

This is in QWP/ILP ingestion, a different subsystem from this PR (SQL `array_agg`). Root cause below; it is environmental and not caused by this change.

### What actually failed
Line 806 is `Assert.assertEquals(0, failureCounter.get())`, so **2 of the 4 sender threads failed** — and it's a *connect* failure, not a data mismatch:

```
Data sending failed [e=LineSenderException: Failed to connect: all 1 endpoint(s) unreachable; last=localhost:53348
  Caused by: HttpClientException: timed out [errno=0]
    at WebSocketClient.readUpgradeResponse(WebSocketClient.java:1021)
    at WebSocketClient.upgrade(WebSocketClient.java:645)
    at QwpWebSocketSender.buildAndConnect(QwpWebSocketSender.java:2429)
```

The TCP socket connected (`errno=0` = no OS error, the client's own timer fired). It timed out waiting for the server's `101 Switching Protocols` **WebSocket upgrade** response. Budget = `authTimeoutMs` = `DEFAULT_AUTH_TIMEOUT_MS` = **15 s**.

### Server-side smoking gun
Same window, `QwpIngressUpgradeProcessor` accepted all 4 TCP connections but only ran the handshake for 2 of them:

| client fd | TCP accept | upgrade handshake | outcome |
|---|---|---|---|
| `…5151` | connCount=1 | **"WebSocket handshake sent"** | ingests, closes `code=1000` |
| `…0144` | connCount=2 | **"WebSocket handshake sent"** | ingests 703 symbols, closes `code=1000` |
| `…3233` | connCount=3 | **never processed** | `disconnect read=-2` (client gave up after 15 s) — failure 1 |
| `…2895` | connCount=4 | **never processed** | `disconnect read=-2` (client gave up after 15 s) — failure 2 |

The agent is `windows-x86-64 [2 processors]`, 8 GiB RAM; the server HTTP pool is `workers=2, pool=shared-network`. The test fires **4 concurrent senders** plus a `columnConvertProb>0` ALTER thread (continuous `ALTER COLUMN … TYPE`) — **24 WAL "rolling … with type conversion" segment rolls in that same 15 s window**. The 2 HTTP workers got pinned serving the 2 connections that upgraded first (sustained ingest + WAL conversions on the shared write pool); on a 2-core box they never freed up to run the handshake for the remaining 2 connections before the 15 s client budget expired.

Timeline matches to the millisecond: test starts 15:32:37.30 → senders connect ~15:32:38 → both failures at 15:32:53.01/.02 = +15 s.

### Why frozen seeds don't reproduce it
The seeds (`1673215709700L, 1780327957300L`) control only the generated rows/columns/symbols. The failure is a worker-availability race — whether 2 HTTP workers, pinned by 2 active ingest streams + WAL conversions on 2 physical cores, reach a 3rd/4th connection's handshake within 15 s. On a normal multi-core box the handshake completes in milliseconds regardless of seed, so it never reproduces. It needs the 2-core `windows-fuzz2` agent under this specific concurrent ALTER/WAL load.

### Conclusion
Pre-existing flake in `QwpSenderFuzzTest`, independent of this PR. The signature (`Failed to connect` / `readUpgradeResponse` / `timed out [errno=0]`) appears exactly twice in the entire ~280k-line log — both in this one test. A re-run should go green.

Possible de-flake (separate from this PR): cap sender threads to the server's HTTP worker count on the low-core agents — `initLoadParameters(100, …, 4, …)` hard-codes 4 senders against a 2-worker pool, while iterations already special-case `Os.isLinux() ? 5 : 3`. Bumping the client timeout would not reliably help, since both workers stay pinned for the run.

<sub>🤖 Investigated with Claude Code from the Azure DevOps build logs.</sub>

PR #7195, opened 2026-06-01
fix(sql): fix a parquet query crash and wrong results in latest-by, ASOF joins, and counts

bluestreak01:
## What

Migrates the execution-plan assertions in 15 test classes from the legacy
`assertPlanNoLeakCheck(query, plan)` / `assertPlan(...)` helpers to the
`QueryAssertion` builder's plan-only terminal:

```java
assertQuery(query).noLeakCheck().assertsPlan(plan);
```

This continues the move toward a single fluent way of asserting query results and
plans. `.assertsPlan()` runs the same `EXPLAIN <query>` and compares the cursor
against the expected plan with the same Async-JIT/Async fallback as the legacy
helper. Most conversions are direct plan-only swaps that move the query and plan
text verbatim and are byte-for-byte equivalent. A minority of sites (about 34)
needed real edits rather than a verbatim move: explain-as-data assertions had their
`explain `/`QUERY PLAN` headers stripped, and the `assertSqlAndPlanNoLeakCheck`
conversions were strengthened (added `.timestamp*()` / `.sizeMayVary()`, upgraded
`returnsOnce` -> `returns`). Those edits strengthen the assertions; none weaken them.

`ExplainPlanTest` additionally drops its three now-unused local plan helpers and
routes `assertBindVarPlan` / `assertSqlAndPlanNoLeakCheck` through the builder. The
full-fat compiler-form plan assertions migrate as well, via
`.withCompiler(compiler).assertsPlan(...)` -- the builder does accept a compiler. The
shared `assertPlanNoLeakCheck` / `assertSqlAndPlanNoLeakCheck` helpers are then
removed and all their callers migrated.

About 1,150 plan assertions across these files were converted:

- ExplainPlanTest, GroupByTest
- SampleByTest, SampleByNanoTimestampTest, SampleByFillNullValueTest, SampleByFillPrevTest, GroupByFunctionCaseTest
- WindowFunctionTest, WindowDecimalFunctionTest, VwemaWindowFunctionTest
- IntervalFunctionTest, TimestampFloorFromOffsetFunctionFactoryTest, TimestampAddWithTimezoneFunctionFactoryTest
- ArrayTest, ArrayOrderBookTest

The branch also strengthens result assertions by replacing single-pass
`returnsOnce()` calls with the full `returns()` battery where the query is
deterministic, and by migrating result tests off `assertSql()`. One such
conversion, `AsOfJoinTest.assertAlgoAndResult`, additionally folds its separate
`EXPLAIN` plan check into the same assertion through `.withPlanContaining(...)`, so
plan and result are now checked in a single pass. That stronger checking surfaced
five production bugs, fixed here (see Bug fixes).

## Bug fixes

These bugs stayed hidden under weaker assertions -- `returnsOnce()`, which reads
the cursor once, or `assertSql()`, which never exercises random access. The full
`returns()` battery re-reads the cursor, calls `skipRows()` / `calculateSize()`,
checks timestamp metadata, and walks the cursor with `recordAt()` random access.
That is what exposed them.

**Cross join `skipRows()` was not re-entrant.**
`CrossJoinRecordCursor.skipRows()` produced the correct result only when called
once from a master-row boundary. A second call while the cursor sits partway
through the current master row's slave scan -- for example the call
`LimitRecordCursor.calculateSize()` issues against an unknown-size base -- re-skips
the already-consumed master cursor and drops the remaining slave rows of that
master row. Effect: `LIMIT lo, hi` and result-size calculation over a cross join
whose input reports an unknown size (such as an indexed or filtered scan) could
count or skip too few rows. The fix drains the current master row's remaining
slave rows before applying the per-master-row skip arithmetic. New regression
test: `JoinTest.testCrossJoinSkipRowsIsReentrant` (fails without the fix, passes
with it).

**Multi-value indexed latest-by did not order by the designated timestamp.**
`LatestByValuesIndexedRecordCursor` (the `<indexed symbol> in (...) latest on ts`
path) emitted its result rows in index/partition discovery order without sorting,
unlike its sibling `LatestByValuesIndexedFilteredRecordCursor`, which already
sorts. Across multiple partitions the output was therefore not ordered by the
designated timestamp, yet the factory advertises that timestamp and a forward
scan direction -- so the optimizer elided `ORDER BY ts` and returned unsorted
rows. Effect: `... in (...) latest on ts partition by ... ORDER BY ts` over a
multi-partition table returned rows not sorted by ts (and the un-ordered form
returned an arbitrary order). The fix sorts the result set into ascending
designated-timestamp order -- inverting the page-frame index, then
`rows.sortAsUnsigned()` -- mirroring the filtered sibling. Covered by
`SqlCodeGeneratorTest.testLatestByAllIndexedListMultiplePartitions`, now asserted
through `.timestamp("ts").expectSize().returns(...)`.

**Parquet random access could read freed memory and crash the JVM.**
`PageFrameMemoryPool.navigateTo(int, record)` early-returned whenever the record's
frame index already matched the requested one, assuming the record still pointed
at the pool's live buffers. But an `AsyncFilteredRecordCursor` record is also bound
to a reduce task's parquet buffers during sequential iteration (via
`record.init(task.getFrameMemory())`), and `PageFrameReduceTask.collected()` frees
those parquet buffers eagerly on every collect (`DirectLongList.address=0`,
`RowGroupBuffers.ptr=0`). A later `recordAt()` to the same frame index took the
early-return, skipped rebinding to the pool's own live buffers, and the next column
read dereferenced freed native memory -- a JVM SIGSEGV. Effect: random access over
a parquet partition (the `recordAt()` path a parent operator uses, for example
under `LIMIT`) could crash the process. The pool's flyweight `navigateTo(int)` was
unaffected because `releaseParquetBuffers()` resets its frame index; only a foreign
record dangles. The fix keeps the early-return for native frames -- whose page
addresses come from the address cache and stay valid for the whole query -- and
always re-navigates parquet records; `nextFreeBuffers()` returns the cached buffer
without re-decoding, so a repeated visit to the same frame stays cheap. Covered by
`ParquetLateMaterializationFuzzTest.testLateMaterializationWithLimit`, whose
migration from `assertSql()` to `returns()` first triggered the crash.

**Dense ASOF join produced stale results when its cursor was re-read.**
`AsOfJoinDenseRecordCursorBase.toTop()` reset every forward/backward scan field
except `backwardScanExhausted`, and `of()` (cursor re-acquisition from the same
factory) reset none of the five scan fields -- only the two scan hash maps. Once a
first pass exhausts the backward scan and sets `backwardScanExhausted = true`, a
later pass keeps that stale flag and short-circuits before scanning backward, so
any match the backward scan would have found is reported as no-match. Effect: a
Dense ASOF join read more than once -- by a window function, a cached factory
re-running the query, or a parent operator calling `toTop()` -- dropped matches on
every pass after the first; a single read was correct, and the Fast and Light
algos were unaffected because they reset correctly. The fix extracts a
`resetScanState()` helper that clears all five scan fields, including
`backwardScanExhausted`, and calls it from both `of()` and `toTop()`. New
regression test: `AsOfJoinTest.testAsOfJoinDenseReReadProducesStableResults`
(re-reads via `toTop()` and re-`getCursor()`; fails without the fix).

**`count(*)` over a full scan threw on an empty partition missing from disk.**
`FullFwdPartitionFrameCursor.calculateSize()` and its backward sibling opened every
partition unconditionally to sum its row count, unlike `next()`, which already skips
a partition whose metadata row count is zero. A size-0 partition with no directory
on disk -- for example one a backup left out of the restore because it held no rows
-- has nothing to open, so `openPartition()` threw "Partition ... does not exist in
table ... directory" instead of contributing zero rows. Effect: `SELECT count(*)`
and full-scan size calculation over such a table failed outright, while
`next()`-based iteration was already correct. The fix guards `openPartition()` with
`getPartitionRowCountFromMetadata(partitionIndex) > 0` in both the forward and
backward cursors, bringing `calculateSize()` to parity with `next()` and the
interval cursors. Like the others it was surfaced by the stronger `returns()`
battery -- here its `calculateSize()` cross-check, which `assertSql()` never ran.

## Intentionally left out

- A handful of explain-as-data assertions
  (`assertQuery("explain ...").returnsOnce("QUERY PLAN\n...")`) remain -- about 6
  sites across a few files. Most such assertions in the migrated classes were
  converted (stripping the `explain ` prefix and `QUERY PLAN` header, as noted
  above); this tail is left for follow-ups.
- The `returnsOnce` -> `returns` strengthening is applied only where the query is
  deterministic. Genuinely non-deterministic assertions -- unseeded `rnd_*`,
  `now()` / `sysdate()` and the like -- intentionally keep `returnsOnce`.

## Tradeoffs

- The plan-assertion conversion is a large, mechanical diff. Depending on the
  original formatting some files grow (single-line calls expand into the chain) and
  some shrink (multi-line calls compact), so the line delta is not a meaningful
  signal of risk. Correctness of the plan-only conversions rests on the text being
  moved unchanged; the roughly 34 header-stripped or strengthened sites (see What)
  are individually reviewable.
- The five bug fixes change production code (`CrossJoinRecordCursorFactory`,
  `LatestByValuesIndexedRecordCursor`, `PageFrameMemoryPool`,
  `AsOfJoinDenseRecordCursorFactoryBase`, `FullFwd`/`FullBwdPartitionFrameCursor`).
  Four change query results: the latest-by fix alters the row order of affected
  `... in (...) latest on ts` queries, the Dense ASOF fix restores dropped matches
  when such a join is read more than once, and the empty-partition fix makes
  `count(*)` and full-scan size return the correct count -- instead of throwing --
  on a table whose empty partition is absent from disk. All are visible behavior
  changes, in the direction of correctness. A single read of a Dense ASOF join is
  unaffected. The parquet fix changes no
  results; it prevents a use-after-free crash and adds a small amount of repeated
  work on parquet random access (re-navigating to the cached buffer instead of
  early-returning). Each fix is narrow and covered by a test that fails -- or
  crashes -- without it.
- One unrelated change rides along: `WarningsProcessor` swaps `current.length() == 0`
  for `current.isEmpty()` (equivalent; no behavior change). It is incidental to this
  migration and could be dropped without affecting anything else.

## Test plan

- `mvn -pl core test-compile` succeeds.
- All 15 converted plan-assertion classes pass, run individually: 2,660 tests, 0
  failures, 0 errors. `ExplainPlanTest` reports 2 skips, which are its pre-existing
  `@Ignore` tests.
- Cross-join fix: `JoinTest` passes including the new
  `testCrossJoinSkipRowsIsReentrant`; the seed-dependent
  `SqlOptimiserTest.testOrderByAdviceWorksWithCrossJoin1a` passes 25/25 runs.
- Latest-by fix: about 3,300 tests across LatestByTest, ParallelLatestByTest,
  ExplainPlanTest, SqlCodeGeneratorTest, SqlParserTest, SampleByTest,
  SampleByNanoTimestampTest, GroupByTest and others pass with no regressions; only
  the one previously-incorrect test changed.
- Parquet fix: `ParquetLateMaterializationFuzzTest` (46 tests) passes, including
  the previously-crashing `testLateMaterializationWithLimit`, and 5 runs with fresh
  random seeds are green. About 370 further parquet and async-filter tests
  (ParquetTest, ParquetRowGroupPruningTest, ReadParquetFunctionTest,
  AsyncFilteredRecordCursorFactoryTest) pass with no regressions.
- Dense ASOF fix: `AsOfJoinTest` (113 tests) passes including the new
  `testAsOfJoinDenseReReadProducesStableResults`; `AsOfJoinFuzzTest` (6),
  `SqlHintsTest` (6) and `SqlOptimiserTest` (174) pass with no regressions.


questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=238812)
**Tested Commit:** `af679df8f9ef07ed2f96c31e5cda4d86f79717ab`

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=239006)
**Tested Commit:** `53272267d6f5da5d0d1bf4b8cbb552492da704d5`

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=239528)
**Tested Commit:** `422d266ebf80fd90993d0b3aee79c2b47be08b72`

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 32 / 33 (96.97%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/join/CrossJoinRecordCursorFactory.java|7|8|87.50%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinDenseRecordCursorFactoryBase.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByValuesIndexedRecordCursor.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullBwdPartitionFrameCursor.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/WarningsProcessor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullFwdPartitionFrameCursor.java|4|4|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 43 / 44 (97.73%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/join/CrossJoinRecordCursorFactory.java|7|8|87.50%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinDenseRecordCursorFactoryBase.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryRecord.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByValuesIndexedRecordCursor.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/ReusablePageFrameMemory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullBwdPartitionFrameCursor.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/WarningsProcessor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullFwdPartitionFrameCursor.java|4|4|100.00%|

bluestreak01:
## Code review (level 2)

The 5 production bug fixes are all **correct** — I verified each by tracing callers, siblings, and bit-layout, corroborated by parallel correctness/concurrency/performance/resource passes. The findings below are about test infrastructure and coverage, not the fixes themselves.

### Critical

**1. Enterprise CI is red, and the likely cause is this PR removing shared `protected`/`public` test helpers — out-of-diff (separate repo)**

This PR deletes, from the **shared** test base classes that the enterprise repo extends:
- `AbstractCairoTest.assertSql(CharSequence expected, CharSequence sql)` (protected, query-result)
- `AbstractCairoTest.assertSql(CharSequence sql, ObjList<BindVariableTestTuple> tuples)` (protected, bind-var)
- `AbstractCairoTest.assertPlanNoLeakCheck(query, plan)` + the `SqlCompiler` overload (protected)
- `TestUtils.assertSql(...)` — **public static**, both overloads
- the `BindVariableTestTuple` class (deleted outright)

No compatibility shim remains (no surviving `assertSql` in the base classes). Within core tests alone these were called **218 + 1,340 times**. `questdb-butler` reports the enterprise suite failing at commit `e1dc10cc3c` — which is a descendant of the removal commits (`ce929032bc remove assertSql`, `3e5369ba1e`, `8621f598e1`), so the enterprise build ran with these gone. The enterprise repo (`EntGriffinServerMain` and its test tree) depends on questdb-core's test artifacts and, at that call density, almost certainly references the removed surface — which would be a hard compile failure.

I can't read the Azure log to confirm it's a compile failure versus an unrelated test failure, so this isn't certainty — but it's the highest-probability cause and it's **in scope of this PR**. **Action:** open the [Azure build](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=239943); if it's the removed helpers, the enterprise test tree must be migrated to the builder in lockstep (or land a coordinated enterprise PR) before this merges. A green enterprise CI is a merge gate.

### Moderate

**2. The empty-partition `count(*)` fix has no dedicated regression test — in-diff**

Four of the five fixes ship a named, deterministic regression test (`testCrossJoinSkipRowsIsReentrant`, `testAsOfJoinDenseReReadProducesStableResults`, `testLatestByAllIndexedListMultiplePartitions`, `testLateMaterializationWithLimit`). The fifth — `FullFwd/FullBwdPartitionFrameCursor.calculateSize()` skipping a size-0 partition absent from disk — does not. The PR says it was "surfaced by the stronger `returns()` battery" of some migrated test, but names none, and I couldn't identify it.

Crucially, the coverage bot's "4/4 lines covered" does **not** prove the fix: the four covered lines (`getPartitionRowCountFromMetadata`, `openPartition`, `if (hi > 0)`, `counter.add`) are all hit by a *normal non-empty* `count(*)`. The actual bug branch — `getPartitionRowCountFromMetadata(...) > 0` being **false**, skipping `openPartition()` — adds no line, so its coverage is invisible to the metric and unconfirmed. For a fix that converts a hard `count(*)` crash ("partition does not exist") into correct behavior, this branch should be locked in. **Action:** add a test that builds a table with a size-0 partition whose directory is absent (the "backup left it out of the restore" scenario) and asserts `SELECT count(*)` returns the right count through `.expectSize().returns(...)` — or at minimum name the migrated test that exercises it.

### Minor

**3. A few new test literals omit thousands separators — in-diff.** e.g. `timestamp_sequence(0, 60000000)` should be `60_000_000` (ParquetLateMaterializationFuzzTest), and assorted `100000L`. Per the repo convention, numbers ≥ 5 digits use `_`. Cosmetic.

**4. Unrelated `WarningsProcessor` change rides along — in-diff.** `current.length() == 0` → `current.isEmpty()` (equivalent; `CharSequence.isEmpty()` is a Java 15+ default, fine on the Java 17 core). The description already flags it as incidental. Harmless; doesn't belong to this migration.

### Downgraded (false positives ruled out during verification)

- **"Latest-by now emits in descending order (sort direction inverted wrong)."** False — the latest-by factory builds an `ORDER_DESC` (backward) page-frame cursor, so `frameCount` grows as time goes *back*; `MAX_SAFE_PARTITION_INDEX - frameIndex` + `sortAsUnsigned()` ascending therefore yields ascending designated-timestamp order. It is byte-identical to the shipped `LatestByValuesIndexedFilteredRecordCursor` sibling, and the shared factory advertises the default `SCAN_DIRECTION_FORWARD` — which is exactly why the old unsorted emission was the bug.
- **"Moving `openParquet()` into the cache-miss branch leaves `parquetColumns`/`parquetIdxToDecodeSlot` stale on cache hits."** False — both are read only by `decode()`/`remapColumns()`, which run *only* in the `!cacheHit` branch; the cache-hit path re-points the record at the `ParquetBuffers`' own already-remapped address lists and never touches them.
- **"`boundPool` is written on a worker thread and read on the collector thread without synchronization (data race)."** False — `boundPool` is thread-confined per `(pool, record)` pair; the worker binds it on a *per-worker* record it never random-accesses, the collector binds/reads on the cursor's own record, and the handoff publishes through the lock-free ring queue (happens-before).
- **"The `boundPool == this` early-return can still hit a same-pool use-after-free via `releaseParquetBuffers()`."** False — the fix only makes the early-return *stricter*; `releaseParquetBuffers()` is called only on reduce-side pools, never on the random-access pool mid-scan, and the record's usage bit pins its buffer against eviction.
- **"`resetScanState()` misses a scan field."** False — the dense ASOF cursor has exactly five per-scan fields; all five are reset, and it's now called from both `of()` (previously reset none) and `toTop()` (previously omitted `backwardScanExhausted`).

### Summary

**Verdict: request changes** — not for the production code (all 5 fixes are correct and well-targeted), but to resolve the red enterprise CI (#1) before merge and to add the one missing regression test (#2).

- **Production correctness:** clean. The cross-join `skipRows` re-entrancy arithmetic, the dense-ASOF 5-field reset, the parquet `boundPool` UAF guard + `openParquet` relocation, the latest-by invert-and-sort, and the empty-partition guard are each verified against callers, siblings, and the `Rows` bit-layout. No correctness, concurrency, resource, or algorithmic findings.
- **Tradeoff worth restating:** four of these fixes change query results (latest-by row order, dense-ASOF restored matches on re-read, `count(*)` on absent empty partitions) — all toward correctness; the parquet fix changes no results and adds bounded O(cache=8) repeat work on parquet random access. The PR describes these accurately.
- **The plan/result migration audit is clean:** no surviving query-result `assertSql`, the 18 newly-added `.returnsOnce(...)` are all on genuinely non-deterministic queries (concurrent-rename races, blocking `sleep()` queries, unseeded `rnd_*`), assertion counts are conserved across files (e.g. WindowFunctionTest 942 → 942), and the EXPLAIN-folds strengthen rather than weaken.

nwoolmer (approved):
L G T M

PR #7197, opened 2026-06-02
feat(sql): cache decoded Parquet row groups on disk for cold-tier partitions

kafka1991:
## Summary

This PR adds an optional cold-tier disk-spill path to the Parquet decoded-frame cache. When a SCATTERED-access cursor is about to evict a decoded row group that lives on a cold-tier (remote) partition, the pool copies the decoded bytes to a local-disk scratch file and reads them back on a later miss, instead of re-decoding from the (remote) Parquet file.

**Depends on #7230**, which carries the in-memory byte-cap cache and the access-pattern hint this path builds on. Until #7230 merges, the diff here still includes the #7230 changes; once it merges, this branch will be rebased onto master and the diff will narrow to the disk-spill path alone.

Disk spill stays dormant under default configuration. It is gated on three conditions (below), and the cold-tier flag is currently a stub that always returns false in production, so the path does not engage until remote-storage metadata is wired in.

## New configuration

- `cairo.sql.parquet.cache.disk.size` — per-cursor disk quota for the cold-tier spill path. Default 0 (spill disabled).
- `cairo.sql.parquet.cache.disk.dir` — scratch directory for spill files. Default `${cairo.root}/tmp`. The engine wipes stale spill files (matching the `qdb-parquet-spill-` prefix) from this directory on startup.

## How it engages

All three gates must hold for a buffer to spill on eviction:

- the factory access pattern is SCATTERED,
- the partition is marked cold-tier (`PageFrameAddressCache.isColdParquetPartition`), and
- `cairo.sql.parquet.cache.disk.size > 0`.

The cold-tier flag is a stub returning false in production, so the spill path stays dormant until future remote-storage work supplies real cold-tier metadata. The in-memory byte budget (#7230) governs the cache on its own meanwhile.

## Behaviour and footprint

- For SCATTERED cursors over cold-tier partitions with a positive disk quota, eviction writes decoded chunk data to a scratch file rather than discarding it; a later miss on the same frame reads it back instead of re-decoding. Quota exhaustion silently degrades to discard (re-decode on the next miss).
- Spill files are tagged with a per-cursor ID and deleted on cursor close; stale files left by a crashed prior process are wiped on engine startup.
- Worst-case disk use attributable to spill equals `cairo.sql.parquet.cache.disk.size x concurrent SCATTERED cursors over cold-tier partitions`.
- Corrupted, short, or truncated spill files degrade to re-decode rather than failing the query; write/open/allocate failures are counted and degrade to discard.

## Internal changes

- `PageFrameMemoryPool.SpillManager`: chunk-data write on eviction, read back on miss, per-cursor quota, file lifecycle bound to cursor close and engine startup wipe. The `ParquetBuffers` restore-state machinery (restored slot metadata, restored scratch, `setupRestored`) backs the read-back path.
- `ParquetDecodeMetrics` (`parquet_decode_buffer_spills`, `parquet_decode_buffer_spill_failures`, `parquet_decode_buffer_spills_skipped_quota`, `parquet_decode_buffer_restores`) registered with the standard Metrics registry. All stay at zero in production until cold-tier spill engages.
- VARCHAR round-trip uses a materialize step. A VARCHAR_SLICE chunk's aux entries can reference multiple Rust-owned page or dict buffers that are not adjacent in memory (RLE-dict entries point into the dict page; DeltaByteArray spill entries point into the data vector after fixup), so a naive contiguous copy is unsafe. Spill walks aux, copies each non-null entry's bytes into a contiguous blob, and rewrites the entry's pointer field with the offset within the blob. Restore reads the blob back and rebases the offsets to absolute pointers.
- Cold-tier flag plumbing: `PageFrame.isColdParquetPartition` (stub default false), `PageFrameAddressCache` per-frame cold bits, and forwarders in `SelectedRecordCursorFactory` / `ExtraNullColumnCursorFactory`.
- `Bootstrap` verifies the spill directory; `CairoEngine` wipes stale spill files on startup.

## Test plan

- ParquetTest: spill round-trip tests across sort, hash join, hash outer join, cross join, nested loop join, latest by, cached window, async TopK, async window join, window join, many-partition, and VARCHAR workloads, each with the cold flag forced on, asserting the spill and restore counters move above zero. Negative paths: corrupted spill files (bad magic, wrong version, bad or below-current slot counts, truncated/oversized/short header or body, schema mismatch) degrade to re-decode; quota exhaustion degrades to discard; allocate/openRW/write failures bump the failure counter and degrade; hot-partition and MONOTONIC scans do not spill; spill files deleted after cursor close; stale-file cleanup preserves foreign-PID files and subdirectories.
- ParquetLateMaterializationFuzzTest: disk-spill-enabled fuzz over cold partitions, including a VARCHAR late-materialization case.
- MetricsTest, ServerMainTest, PropServerConfigurationTest: the new counters and disk config keys exposed.

The in-memory cache, access-hint, and byte-budget benchmarks are covered in #7230.


bluestreak01:
@kafka1991 — code review below (level 2: full change-surface map, parallel correctness/concurrency/perf/resource/test/quality agents, findings verified against source).

This is a large (89 files, +2008/-226), carefully-engineered PR. The highest-risk areas were independently verified in source. **No Critical correctness, memory-safety, or resource-leak bug was found** — the VARCHAR_SLICE pointer rebasing, the pin/eviction invariant, the spill/restore error paths, and `cachedBytes` accounting are all correct. The new data structures (`byFrameIndex` O(1), intrusive LRU, skip-`openParquet`-on-hit) are genuine improvements over the old triple-O(n)-scan code. Findings are concentrated in performance regressions, test coverage, and docs.

## Critical
None in the code.

**Merge gate (external):** Enterprise CI failed on the tested commit `f18f5b6d0c` (Azure build 238639). A fresh run is currently pending — please confirm it goes green or investigate the root cause before merge.

## Moderate

**M1 — Per-frame `ParquetBuffers` allocation on the hot async-parquet path (regression vs master).** `releaseParquetBuffers` (`PageFrameMemoryPool.java:355`) is called **per frame** by `PageFrameReduceTask.collected():342` and the `Async*RecordCursorFactory` finally-blocks (e.g. `AsyncTopKRecordCursorFactory.java:256`). The old version moved buffers to a persistent `freeParquetBuffers` list via `Misc.freeObjListAndKeepObjects` — freeing native memory but **retaining the Java objects** for `reopen()` reuse. The new version `close()`s and drops every `ParquetBuffers`, so the next frame's `acquireBuffer` (line 410) allocates a fresh `new ParquetBuffers()` = 1 shell + 5 `DirectLongList` + 1 `RowGroupBuffers` per frame (more when `recordB` is live). Native alloc/free churn is unchanged; the regression is purely the Java-object allocation the old free-list eliminated. Magnitude is bounded (young-gen, dwarfed by decode), but it's an avoidable per-frame allocation on the most common parquet query path. Suggest restoring a free-list (or retaining the small bookkeeping lists across release), or justifying the change in the PR.

**M2 — `SpillManager.findSpilled` is an O(n) linear scan.** `PageFrameMemoryPool.java:788`. Called once per `spill()`, per `restore()` miss, and per `dropSpilledFrame`. `n` is bounded only by `diskBudget / avgFrameBytes` (hundreds–thousands), so a scattered sweep is O(n) per access → O(n^2) overall. Off by default today (disk.size=0, cold flag is a prod stub), but it's on the explicitly-documented future path, and the fix is the same `IntIntHashMap frameIndex->index` pattern already used for `byFrameIndex` with O(1) swap-remove. Worth fixing while the code is being written.

**M3 — Spill error/integrity paths and most SCATTERED contexts are untested; `reDecodeInitial` is unreachable-and-untested.** Spill tests exercise only `ORDER BY` (light-sort). Untested: (a) the 7 integrity-rejection branches in `restore()` (bad `slotCount`, negative sizes, `bodyBytes`/`fileLen` mismatch, prefix mismatch) — file-content-derived native code; (b) degrade-to-discard on spill `write`/`allocate`/`openRW` failure (only `mkdirs` failure is fault-injected); (c) SCATTERED join / latest-by / cached-window / full-sort / top-K spill paths (a dropped SCATTERED hint there silently disables spill with no test to catch it); (d) two-record (A+B) simultaneous-pin eviction. Separately, `reDecodeInitial` + the `isRestored` branch in `populateRemainingColumns` (lines 601, 1211) appear **effectively unreachable**: late-mat only targets the freshly-decoded `FRAME_MEMORY`-pinned buffer, while restores occur only for `recordB` comparison navigations that never late-materialize — so restore-then-late-mat never coincide. `testWithDiskSpillEnabledVarcharLateMaterialization` does not reach it (measured re-decode count = 0). Either prove it unreachable and guard/document it, or add a test that forces restore-then-late-mat.

**M4 — Shipped config template not updated.** `core/src/main/resources/io/questdb/site/conf/server.conf:526` still advertises the now-inert `#cairo.sql.parquet.frame.cache.capacity=8` and omits the 3 new keys (`cairo.sql.parquet.cache.memory.size`, `.disk.size`, `.disk.dir`). Operators reading the template will tune a dead key and never discover the new controls. Update the template and ship/reference the public-docs update.

## Minor

- **`PageFrameMemoryPool.java:1313-1314, 1355-1356`** — `slotParquetIndexes`/`slotColumnTypes` are populated on every decode but read only by `spill()`. When `spillManager == null` (default) this is dead work on the decode path; guard the two appends with `if (spillManager != null)`. (The `decodedBytes` byte-sum in the same loop must stay.)
- **`RecordCursor.java:140`** — `default void setParquetDecodeHint` is inserted between `close()` and `expectLimitedIteration()`, breaking alphabetical instance-method order. It belongs between `preComputedStateSize()` and `size()` (where the overriding wrappers correctly place it).
- **`PageFrameMemoryPool.java:450`** — local `boolean restored;` violates the `is`/`has` prefix rule; rename to `isRestored`.
- **`PropertyKey.java:139`** — `CAIRO_SQL_PARQUET_FRAME_CACHE_CAPACITY` lacks the `// deprecated, use ...` comment used for other deprecated keys, though the body/`CairoConfiguration` mark it `@Deprecated`.
- **`PropServerConfiguration.java:2216`** — `Math.max(getInt(...), 8)` still clamps the deprecated capacity that nothing consumes; dead work.
- **`PageFrameAddressCache.java:55,123`** — `public static volatile FORCE_COLD_PARQUET_PARTITION_FOR_TEST` is a test-only flag in production code, read on every `add()`. Acceptable for now; debt to remove once real cold-tier metadata lands.
- **Spill-dir robustness** (`PageFrameMemoryPool.java:171, :680`): startup `cleanStaleDiskSpillFiles` deletes all `qdb-parquet-spill-*` files ignoring `cursorId`, so two instances explicitly sharing the dir would wipe each other's live files (default dir is per-`dbRoot`, so default config is safe); and the `!exists && mkdirs!=0` ctor check can spuriously throw under concurrent first-query init on a fresh custom dir (treat post-mkdirs `exists` as success). Both misconfiguration/rare-race only.
- **`PageFrameMemoryPool.clear():182`** doesn't call `spillManager.close()`/`deleteAllSpilledFiles()` (unlike `close()`). Harmless today (only no-spill `0L` pools call `clear()`), but a latent leak if a spill-enabled pool ever uses `clear()` as terminal cleanup — document `clear()` as reset-only or mirror the cleanup.
- **Memory footprint tradeoff**: the default 256MB budget is **per cursor** (SCATTERED -> 256MB, MONOTONIC -> 64MB); worst case is `budget x concurrent cursors` vs master's fixed 8-slot cap. The body acknowledges it; please state the product explicitly in the docs so operators size against cursor concurrency. (Most parallel paths pass `0L`, so fan-out doesn't multiply.)
- **Tests**: new integration tests use lowercase SQL keywords (CLAUDE.md prefers UPPERCASE; sibling tests in the file use UPPERCASE) and `assertSqlCursors` rather than `assertQueryNoLeakCheck` (so factory `supportsRandomAccess`/`expectSize` aren't asserted on the spill path).
- **Labels**: add `storage` (the bulk of the diff is `cairo/` storage-engine code + a disk-spill subsystem).

## Verified and dismissed (false positives)
- *VARCHAR_SLICE spill dereferences inlined strings as pointers* — the slice aux format never inlines; `VarcharTypeDriver.getSliceValue:246` always reads `hdr>>>4` length + `entry+8` pointer, which the spill code matches exactly.
- *Removing the `keepClosed=true` ctor flag changes memory behavior* — `keepClosed=false` allocates eagerly, consistent with the removed `reopen()` lifecycle.
- *Startup `verifyFileSystem` fails on a missing spill dir* — with `failOnNfs=false` it only logs "NOT FOUND".
- *`cachedBytes` drift / double-free on error paths* — every error path subtracts exactly once via `evictHalfInitialized` and adds only after success; `restore`'s `bodyPtr` ownership handoff (`isOk` flag) is correct.
- *`Path.getThreadLocal` clobbering; use-after-free from pin clearing* — verified safe (path fully consumed before nested rebuild; victims always have `usageFlags==0` so bound buffers are never evicted).

## Verdict
Request changes — but the changes are perf/coverage/docs, not correctness. Blocking before merge: confirm enterprise CI (gate), M2 (cheap O(1) fix), and M4 (config docs). M1, M3, and the minors are strongly recommended.

<sub>Reviewed with Claude Code.</sub>

kafka1991:
Reviewing PR #7197 at level 2.

35 draft findings → **0 Critical, 0 Moderate, 27 Minor** after verification against source (5 false positives dropped, 2 deduped).

## Critical
None.

## Moderate
None.

## Minor

**Performance**
- `openParquet` rebuilds `parquetColumns` + `parquetIdxToDecodeSlot` on every cache miss (`PageFrameMemoryPool.java:580-606`). Output is a pure function of `columnMapping` (cursor-stable) and `columnIdToParquetIdx` (file-stable); cache once per file switch.
- `remapColumns` does 2 IntIntHashMap lookups per query column per miss (`PageFrameMemoryPool.java:1560-1577`). A flat `int[] qToSlot` built in `openParquet` collapses both to one array read.
- `decodeAndAccount` calls `spillManager.restore()` and `decode()` populates per-slot metadata even when the cursor is MONOTONIC (`PageFrameMemoryPool.java:488-507, 1386-1393`). MONOTONIC cursors can never spill (gated at line 420 by `accessPattern == SCATTERED`); cache `spillPossible = (spillManager != null && accessPattern == SCATTERED)` once and gate on it.
- `SpillManager.spill()` emits `1 + 2*slotCount` syscalls (`PageFrameMemoryPool.java:932,955,964`). Disabled by default; when enabled a single-staging-buffer write or `writev` collapses this to one syscall.
- `sumVarcharStringBytes` + `materializeVarchar` make two passes over the varchar aux array (`PageFrameMemoryPool.java:784-798, 848-869`). Could track high-water bytes during `decode()` so `spill()` skips the sizing pass.
- `PageFrameAddressCache.add()` does `4 * columnCount` individual `DirectLongList.add(0)` calls per parquet frame to zero-fill (`PageFrameAddressCache.java:110-115`). `ensureCapacity` + `setPos` suffices since underlying memory is zero-initialised.

**Tests**
- No fault-injection coverage for `restore()` negative-slot-size (line 1111), `diskBudget`-exceeded `bodyBytes` (line 1120), or body-read short-read (line 1137). Existing corruption tests cover magic/version/slotCount/file-size/schema-prefix but not these three.
- VARCHAR spill round-trip exercises only long-form (14-char) strings. Add a case mixing NULL / inlined (≤9 bytes) / long / empty in a single column to cover every branch of `rebaseVarcharAux` + the inlined `setRestoredSlot` aux path.
- Spill is exercised end-to-end through HashJoinLight + SortedLight + LimitedSizeSortedLight only. No spill round-trip for AsOf/Lt/Splice, LATEST BY, CachedWindow, scalar subquery, DistinctTimeSeries, AsyncTopK with SCATTERED — a dropped or mis-wired hint on those factories first surfaces as production OOM.
- `testParquetCacheWrapperChainPreservesScatteredHint` infers hint forwarding from `metrics.spills() > 0`; a future override that swallows the hint in one wrapper of a chain still passes if another wrapper keeps SCATTERED. Direct unit tests on each wrapper would catch that.
- No test for budget-overshoot when a single decoded row group exceeds `maxCacheBytes > 0`. JavaDoc promises temporary overshoot; nothing pins the behaviour.
- No test for shell-pool reuse + cap. `SHELL_POOL_CAP=8` LIFO is exercised only incidentally.

**Resource / Robustness**
- `PageFrameMemoryPool.clear()` calls `spillManager.close()` (line 202-212), which frees scratches and clears tracking. Pool reuses via `of()` and relies on lazy realloc. Works today only because both `clear()` callers (`PageFrameReduceTask`, `LatestByTask`) construct with `maxCacheBytes=0` → `spillManager==null`. Latent if either caller ever enables spill. Add a `SpillManager.clear()` that resets without freeing scratch, or document the contract on `close()`.
- `releaseParquetBuffers` ignores `usageFlags` pin bits (line 383-405). JavaDoc warns about it; add `assert boundForRecordA == null && boundForRecordB == null && boundForFrameMemory == null` at the head so misuse fails at the call site instead of later as SIGSEGV.
- `ParquetBuffers` constructor (line 1346-1353) and `reopen()` (1473-1480) allocate 6 native handles without try/catch. OOM mid-sequence leaks the already-allocated handles and (in the `reopen` case) the shell is already removed from the free pool. Latent; native OOM is rare. Wrap in try/catch → `close()` → rethrow.
- `cleanStaleDiskSpillFiles` only deletes current-PID files (line 739-766). Stale files from prior JVM crashes (different PID) accumulate forever; operator must wipe manually. Intentional behaviour (avoids stomping on concurrent QuestDB instances sharing a spill dir) and tested by `testParquetCacheCleanStaleSpillFilesPreservesForeignPidFiles`, but worth documenting that a single-instance deployment doesn't benefit from cross-restart cleanup.
- `SpillManager.nameStartsWith` has a `pid == 0 ? 1 : 0` branch (line 747) that is dead code on POSIX/Windows. Either drop it or `assert PID > 0` once at static init.
- `SpillManager.spill()` catches I/O errors and `return`s without rethrowing (lines 920, 928, 934, 957, 966), but `acquireBuffer` wraps the call in `try/catch (Throwable)` (lines 422-428) that logs a separate "discarding buffer" message — that outer catch is effectively dead for I/O failures. Pick one model: rethrow from `spill()`, or drop the outer try/catch.

**Code quality**
- `SpillManager` static method ordering: private statics `nameStartsWith` (739), `rebaseVarcharAux` (768), `sumVarcharStringBytes` (784) and package-private `deleteSpillFilesInDir` (989) sit between instance methods. CLAUDE.md requires static-vs-instance grouping.
- `freeParquetBufferShells = new ObjList<>(8)` (line 117) uses the literal `8` instead of `SHELL_POOL_CAP`. The two can drift independently.
- `IS_COLD_PARQUET_PARTITION_FORCED_FOR_TEST` is `public static volatile` (line 57 of `PageFrameAddressCache.java`). Production injection point enforced only by comment. Worth moving behind a `@VisibleForTesting`-style setter or a `CairoConfiguration` flag when cold-tier metadata lands properly.

## Verified and dismissed (false positives)
- *Restore-path memory leak for `count(*)` queries* — `count(*)` doesn't go through `navigateTo` (entity cursor bypasses parquet decode); when there is a filter, the base subquery's metadata has its real column count. Unreachable.
- *Sibling cursor restores spill with different projection* — `of()` calls `spillManager.deleteAllSpillFiles()` at every cursor re-attach; the next cursor starts with empty tracking. Prefix mismatch never occurs.
- *`populateRemainingColumns` failure leaves Rust state half-mutated* — `evictHalfInitialized` calls `buffers.close()` which destroys the Rust `ColumnChunkBuffers` regardless of state. No leak.
- *`populateRemainingColumns` on a restored buffer would segfault* — every async caller releases between frames, wiping the cache before any restore can race; sync callers never invoke `populateRemainingColumns`. `assert !isRestored` is defensive and reachable only via deliberate refactor; production runs with `-ea`.
- *`spill()` skip-on-duplicate leaves on-disk copy stale after late-mat extension* — depends on the same restore + `populateRemainingColumns` callgraph that does not exist.

## Verdict

**Approve.** No correctness, memory-safety, or resource-leak bug found. The 27 Minor items are polish; none gate the merge.

<sub>Reviewed with Claude Code.</sub>

core/src/main/java/io/questdb/cairo/CairoConfiguration.java
```diff
@@ -774,6 +774,17 @@ default RostiAllocFacade getRostiAllocFacade() {
...
     int getSqlParallelWorkStealingThreshold();
 
+    CharSequence getSqlParquetCacheDiskDir();
+
+    long getSqlParquetCacheDiskSize();
+
+    long getSqlParquetCacheMemorySize();
+
+    /**
+     * @deprecated Replaced by {@link #getSqlParquetCacheMemorySize()}. The value is
```
bluestreak01: why not remove this?

bluestreak01: there is a mechanism to support old config key and new at the same time & keep the code clean of warnings

kafka1991: fixed

bluestreak01:
@kafka1991 this is performance related PR, can you add benchmarks before and after?

kafka1991:
> @kafka1991 this is performance related PR, can you add benchmarks before and after?

hey @bluestreak01 I had ran my own benchmark on my local(no regression) will paste the data later

bluestreak01:
@kafka1991 — level-3 code review below (full change-surface map, parallel correctness/concurrency/perf/resource/test/quality + cross-context and adversarial passes, every finding verified against source).

Reviewed at current HEAD `7604db32a0` — 8 commits past the commit the earlier review comments used (`f18f5b6d0c`). Diff: 96 files, +3352/-318, no Rust changes.

**Prior-review items re-verified as FIXED at HEAD:** ctor native-OOM try/catch (`PageFrameMemoryPool` 158-175 + `ParquetBuffers` 1363-1397), `findSpilled` now O(1) via `spilledFrameIndexToSlot` (853-855), per-frame shell-pool reuse, all three `server.conf` files, the `isTrackingForSpill` decode guard (1430), `PropertyKey` deprecation comment, the dead `Math.max(...,8)` clamp removal, and the cold-flag snapshot moved out of `add()`. Those are genuinely resolved.

## Critical
**None.** No correctness bug, memory-safety violation, leak, double-free, use-after-free, or concurrency hazard was found in any production-reachable path. The VARCHAR_SLICE spill/restore round-trip, `cachedBytes` accounting symmetry, the pin/eviction invariant, and the restore ownership-handoff were each traced clean and confirmed against source.

## Moderate

### M1 — The byte cap silently undercounts VARCHAR/STRING memory; actual RSS can be several× the configured budget *(in-diff)*
`PageFrameMemoryPool.java:1432` — `decodedBytes = Σ (getChunkDataSize(s) + getChunkAuxSize(s))`.

`openParquet` rewrites every VARCHAR column to `VARCHAR_SLICE` (609-611). Verified against the Rust side (`parquet_read/mod.rs`, `decode.rs:58-62`, `column_sink/var.rs`, the dict/delta decoders): `getChunkDataSize` returns `data_vec.len()`, but for the **common** encodings (Plain, DeltaLengthByteArray, RLE/Plain dictionary) the decoded string bytes live in a separate, retained Rust buffer `page_buffers: Vec<Vec<u8>>` that the aux entries point into — `data_vec` stays empty, so `data_size == 0`. Only the rare DeltaByteArray encoding copies into `data_vec`. The codebase corroborates this itself: the spill path uses `sumVarcharStringBytes()` (sums `hdr>>>4`) instead of `getChunkDataSize` for VARCHAR_SLICE, with a comment that the bytes aren't contiguous (901-903, 880-881).

Consequence: each cached VARCHAR frame counts only its 16-byte-per-value aux toward the budget while holding `16 + stringLen` bytes natively. For 100-byte strings that's a ~7× undercount; for 256-byte strings ~17×. The pool evicts on `cachedBytes >= effectiveBudgetBytes` (414), so with a 256 MB SCATTERED budget it retains ~160 VARCHAR frames (~1.8 GB resident) before evicting, and ~40 frames (~460 MB) at the 64 MB MONOTONIC budget. This is **new behavior** — master capped the cache by frame count (default 8), so for VARCHAR-heavy workloads the branch retains materially **more** memory than master and **exceeds the operator's configured ceiling several-fold**, which is exactly the guarantee a memory-constrained deployment relies on to bound RSS.

Not a correctness issue and the memory is bounded per-cursor and released on close, so not Critical — but it defeats the PR's central promise (a decoded-bytes ceiling) for the most memory-heavy column type. This is the most important item to address.

Fix options: include the real string bytes in `decodedBytes` for VARCHAR_SLICE slots — reuse the existing `sumVarcharStringBytes(getChunkAuxPtr(s), getChunkAuxSize(s))` at decode time (adds a per-decode aux scan for varchar columns — a hot-path cost), or expose the per-chunk `page_buffers` byte total from Rust and add it (cleaner/exact, needs a Rust change). At minimum, document that the budget excludes VARCHAR string bytes so operators can size accordingly.

### M2 — Async window-join slave decode cache left at MONOTONIC, while this PR upgraded the *sync* window-join to SCATTERED *(out-of-diff)*
This PR added `slaveCursor.setParquetDecodeHint(SCATTERED)` to the sync window joins (`WindowJoinRecordCursorFactory.java:187`, `WindowJoinFastRecordCursorFactory.java:240`). The **async** equivalent, `AsyncWindowJoinAtom` (verified: **not touched by this PR**), builds its owner + per-worker slave `TimeFrameCursor`s in `initTimeFrameCursors` (590-607) and never calls `setParquetDecodeHint`, so each slave pool stays MONOTONIC (quarter budget, no LRU-survival reordering). The slave is random-accessed in the same scattered cross-frame-lookback pattern as the sync path (`WindowJoinTimeFrameHelper.recordAt`). Net: window joins that take the parallel path — the large-data case that matters most — run the slave Parquet cache at ¼ the intended budget and re-decode more. Perf only (spill is prod-dormant), but it's a clear inconsistency that undercuts the PR's own optimization on the higher-value path.

Fix: in `AsyncWindowJoinAtom.initTimeFrameCursors`, after each `*.of(...)`, call `setParquetDecodeHint(ParquetDecodeHint.SCATTERED)` on `ownerSlaveTimeFrameCursor` and every per-worker slave cursor, mirroring the sync factories.

### M3 — `openParquet` + `remapColumns`/`remapRemainingColumns` recompute a file-invariant mapping on every cache miss *(in-diff)*
Independently flagged by both performance passes. `openParquet` (589-648) clears and rebuilds `parquetColumns` + `parquetIdxToDecodeSlot` on **every miss**, and `remapColumns` (1582-1627) does two hash probes per query column per miss (`columnIdToParquetIdx.get` → `parquetIdxToDecodeSlot.get`). Both outputs are a pure function of the active decoder file + the cursor-stable `columnMapping`/`columnTypes` — nothing per-frame feeds them. `activateDecoder` already detects the file-identity change (the same guard it uses for `buildColumnIdMap`); gate the rebuild on that and precompute a flat `int[] qToSlot` so each miss does one array read per column instead of clearing two hashmaps and doing 2·Q probes. The MONOTONIC ¼-budget cut raises the miss rate, amplifying this. Miss-path (not per-row), so not Critical, but it's the dominant avoidable per-miss Java cost. Note the late-mat path's `openParquet(...,false)` writes a different column set into the same structures, so any caching must re-mark dirty after late materialization. (This is also present on master, so it's a missed optimization rather than a regression — see note in the summary.)

## Minor

- **Zero-`decodedBytes` frames never evict** *(in-diff)*. `acquireBuffer` only evicts when `cachedBytes >= effectiveBudgetBytes` (414); a frame whose projected columns are all-null fixed-size decodes to `decodedBytes == 0` (1432, 1440-1441), so `cachedBytes` stays 0 and every such frame allocates a fresh `ParquetBuffers` that is never reused. Reachable via `SELECT added_col FROM t` over partitions predating `added_col`. Footprint is small (empty buffers) and released on `toTop`/`close`, so low severity — but unbounded within a long scan over many such partitions. A `> 0` floor per entry, or a count backstop, would close it. (Same root-cause family as M1.)
- **`@TestOnly` removed from `Counter.reset()`** *(in-diff)*. Forced by `ParquetDecodeMetrics.clear()` (production) calling `reset()` via `Metrics.clear()`. `Metrics.clear()` has no production callers, so this is cosmetic today — but it weakens a guard. Cleaner: don't reset monotonic counters in `ParquetDecodeMetrics.clear()` (Prometheus counters are monotonic), or keep the annotation and reset only in test setup. (Don't simply re-add `@TestOnly` without removing the production call — that would break the inspection.)
- **`IS_COLD_PARQUET_PARTITION_FORCED_FOR_TEST` lacks `@TestOnly`** *(in-diff, `PageFrameAddressCache.java:57`)*. A `public static volatile` test escape-hatch in production code; comparable fields carry `@TestOnly`. Production never writes it (verified) and it's snapshotted at `of()`, so it's safe — add the annotation for consistency.
- **Member-ordering violations** *(in-diff, `PageFrameMemoryPool.java`)*. In `SpillManager`, the `private static` methods `nameStartsWith` (748), `rebaseVarcharAux` (777), `sumVarcharStringBytes` (793) sit before the private instance methods instead of in alphabetical position among them; package-private `deleteSpillFilesInDir` (998) precedes `deleteAllSpillFiles` (1022).
- **Dead branch** *(in-diff, `PageFrameMemoryPool.java:757`)*. `int digits = pid == 0 ? 1 : 0;` — `ProcessHandle.current().pid()` is never 0; simplify to `int digits = 0;`.
- **Literal `8` instead of the constant** *(in-diff, `PageFrameMemoryPool.java:117`)*. `new ObjList<>(8)` should be `new ObjList<>(SHELL_POOL_CAP)`; `releaseParquetBuffers` already uses the constant, so the two can drift.
- **Shell-pool reuse discards the native capacity reuse the rewrite is built around** *(in-diff)*. `releaseParquetBuffers` (383-405) `close()`s native memory before pooling the shell; `acquireBuffer` then `reopen()`s it (442-459), re-allocating from zero — so the shell path saves only the Java wrapper objects, not the native Vec capacity the in-place LRU-victim path (415-440) preserves. Likely intentional for the async per-frame release path (it wants to free decoded bytes between frames), but worth a one-line comment so the asymmetry isn't read as an oversight.
- **`assert !isRestored` invariant is sound but untested** *(in-diff, 1299/1461)*. Restore and late-materialization never land on the same buffer (the only path that could collide is AsyncTopK over a cold partition with a non-filter projection), and with `-da` a regression degrades to a catchable `CairoException` via the Rust null-guard rather than a JVM abort. But no test forces restore-then-late-mat. Add that one differential test — it also closes the untested A+B simultaneous-pin eviction path.
- **Async-cursor ctor-catch `close()` is `isOpen`-gated and frees nothing** *(in-diff)*. `AsyncFilteredRecordCursor`/`AsyncFilteredNegativeLimitRecordCursor` wrap their ctor in `try/catch → close()`, but `close()` is guarded by `if (isOpen)` (141) and `isOpen` is false during construction. The window is unreachable today (the only post-pool statement is a non-throwing getter, and a fresh pool/record holds no native memory), so it's not a live leak — but the pattern is dead and inconsistent with `AbstractPageFrameRecordCursor`'s un-gated ctor cleanup.
- **Test coverage gaps**. Strong where present (every spill test pairs correctness with a `spills()/restores()` counter assertion; the disk-file `created==removed` balance test and the corrupted-spill suite are high quality). Missing: (a) no AsyncTopK-over-cold-partition spill+late-mat test (closes the `isRestored` and A+B-pin gaps — the only place spill and late-mat can meet); (b) no SCATTERED spill round-trip for light-sort or through a projection/scalar-subquery wrapper (a dropped hint there is a silent perf cliff); (c) VARCHAR spill mixes NULL/empty/inline/long only in the fuzz test, which asserts correctness but not `restores()>0`; (d) the restore 12-byte header-prefix short-read (1059) and the data/aux (vs header) write-failure branches (964/973) aren't independently hit; (e) no test would catch M1 — add an assertion that actual native memory for a VARCHAR scan stays near `cache.memory.size`.
- **PR description inaccuracies**. It says "15 factories declare it; the 16th (SortKeyMaterializing) does not call recordAt" — there are 17 declaring files, and `SortKeyMaterializingRecordCursor.recordAt` *does* call `baseCursor.recordAt` (204) but correctly forwards the hint, so the conclusion (safe) holds while the rationale is wrong.
- **Scope creep** *(in-diff)*. `Bootstrap.java` (`_` unnamed lambda params, `sb.repeat`) and the `CairoConfiguration.java` reordering of the three posting-index methods are unrelated to the parquet cache. All valid on Java 25, but they don't belong here; the posting-index reorder also flips the A/R/e ordering — reverting it removes both the noise and the ordering debate.
- **AMI template** (`pkg/ami/marketplace/assets/server.conf`) drops the obsolete `frame.cache.capacity` key but doesn't add the three new keys the shipped template documents.
- **Horizon/Markout join master cursor un-hinted** *(out-of-diff, low confidence)*. `HorizonJoin*`/`MultiHorizonJoin*`/`MarkoutHorizon*` set the slave MONOTONIC but leave the master un-hinted, while the master is revisited across offsets via a min-heap (potentially scattered). Lower confidence than M2 (the master is the time-ordered driver and revisitation may be locally bounded) — worth your judgment, not a clear defect.

## Downgraded (false positives / dismissed after source verification)
- *`tryHit`'s `lruMoveToTail` does 9 field-writes per hot row* — FALSE POSITIVE. `lruMoveToTail` early-returns when `b == lruTail` (567); a repeated same-frame hit does zero writes. Reordering occurs only on genuine interleaved A/B access (correct LRU behavior).
- *`spillPossible` guard would speed up MONOTONIC misses* — latent-only; no shipped MONOTONIC pool has a `spillManager` and takes the restore path, and for SCATTERED spill *is* possible, so the guard saves nothing on any current query.
- *VARCHAR_SLICE spill dereferences inlined strings / corrupts NULL/empty* — VARCHAR_SLICE never inlines; `materializeVarchar`/`rebaseVarcharAux` correctly skip NULL and handle empty/long; verified against the Rust sink layout.
- *`restore()`/`spill()` leak/double-free `bodyPtr`/fd, or `populateRemainingColumns` error path double-frees* — ownership flags (`isOwnershipTransferred`/`isRestoreSucceeded`), `evictHalfInitialized`, and idempotent `close()` make every path free exactly once; fd closed in the outer finally.
- *`PageFrameAddressCache.add()` per-element zero-fill is a PR regression* — pre-existing (identical on master); not introduced here.
- *`clear()` calling `spillManager.close()` leaks shells on a spill-enabled pool* — both `clear()` callers (`PageFrameReduceTask`, `LatestByTask`) construct with budget `0L` → `spillManager == null`.

## Performance regressions vs master (since this is a perf PR, calling these out explicitly)
- **Memory (high confidence):** M1 — VARCHAR-heavy workloads retain several× the configured budget and more than master's 8-frame cap (OOM risk on constrained hardware). The clearest regression.
- **Latency (edge case):** for wide projections / large row groups, the ¼ MONOTONIC budget (and to a lesser degree the SCATTERED budget) can cache fewer frames than master's fixed 8, increasing re-decodes — mainly for multi-frame-working-set operators (Horizon/Markout, async window join per M2). The PR body acknowledges wide schemas need tuning, but the *default* serves them worse than master's flat cap.
- **Not regressions:** hit/miss bookkeeping is O(1) vs master's O(n) scan (faster); the spill machinery is fully skipped in default config (`disk.size=0` → `spillManager==null`); buffer lifecycle ≈ master; M3's per-miss recompute exists on master too (missed optimization, not a regression).

## Verdict
**Request changes** — the changes are memory-accounting, perf-consistency, and coverage, not correctness or safety. No Critical issue; the hard parts (VARCHAR pointer rebasing, pin/eviction, spill error paths, ownership handoff) are correct, and all prior-review blockers are fixed at HEAD.

The one item that genuinely matters is **M1**: the byte budget doesn't bound VARCHAR memory (the cached string bytes in Rust `page_buffers` are uncounted), so actual RSS can run several× the configured cap and exceeds master for VARCHAR-heavy workloads — fix or clearly document before merge. **M2** (async window-join hint) and **M3** (per-miss recompute) are perf follow-ups; the rest are minors.

In-diff vs out-of-diff: M1/M3 + most minors are in-diff; **M2 and the Horizon/Markout item are out-of-diff** (the changed `setParquetDecodeHint` contract is incomplete at untouched callsites).

<sub>Reviewed with Claude Code (level 3).</sub>

bluestreak01:
### Why: the decode-cache byte cap undercounted varchar frames

`openParquet` decodes varchar columns as `VARCHAR_SLICE`. For the common encodings (Plain, DeltaLengthByteArray, RLE/Plain dictionary) the decoded string bytes stay in retained native page buffers that the aux entries point into, while `data_vec` stays empty. So `getChunkDataSize()` returns 0, and the budget (`data_size + aux_size`) counted only the 16-byte-per-row aux. Only the rarer DeltaByteArray path copies into `data_vec` and was counted.

The result was that a varchar-heavy query recorded ~16 bytes/row per frame while actually holding 16 + stringLen bytes resident, so the pool retained well beyond the configured cap before evicting. The spill path already sized itself with `sumVarcharStringBytes` instead of `getChunkDataSize`, which is what surfaced the gap.

### The fix

`ColumnChunkBuffers` now tracks `page_buffers_size` — the summed length of the retained page/dict buffers — computed once per column chunk in `refresh_ptrs`, exposed to Java through a new offset accessor. `PageFrameMemoryPool` adds it to `decodedBytes` at the two budget sites (`decode` and `sumChunkBytes`). The page-address remap that feeds the query engine is left untouched, so only the budget changes.

### Why compute it in Rust rather than scan the aux in Java

- **Accuracy.** The field measures the actual retained heap, not logical bytes. Compressed pages decompressed onto the heap are counted; uncompressed pages borrowed from the mmap contribute 0 (the decode allocates no heap for them); a low-cardinality dictionary contributes only its small shared dict page. A Java aux scan would sum logical per-row lengths and would over-count the dictionary and uncompressed cases.
- **Cost.** The sum is O(pages), folded into the decode that already runs, so the Java budget path stays a field read. A JMH micro-benchmark put the alternative per-row aux scan at ~0.6 ns/row (~63 us per 100k-row varchar column, ~10% of an LZ4-compressed single-column frame decode); the field read measures ~0.001 us.

### Tradeoff

The budget now reflects real RSS, so fewer frames fit under the same cap for varchar-heavy workloads. That is the intended effect — it is what bounds RSS — but those queries can re-decode more often than before this change. Uncompressed partitions see no change, since their string bytes are file-backed in the mmap rather than on the heap.

### Validation

- Rust: `fmt`/`clippy` clean; 315 `parquet_read` lib tests including a new unit test for the field.
- Java: 282 parquet tests green (decoder/encoder/read plus griffin `ParquetTest`).
- The benchmark setup confirms the accounting: 100k rows x 128-byte varchars under LZ4 report `page_buffers_size` ~28 MB (previously recorded as 0), while uncompressed reports 0.

Touches Rust, so the committed `libquestdbr` libraries need a refresh via the "Build and Push Release Rust Libraries" workflow on this branch before CI runs against them.

bluestreak01:
@kafka1991 — level-2 code review below (full change-surface map, parallel correctness/concurrency/perf/resource/test/quality + Rust-safety + adversarial-perf agents, every finding verified against source).

Reviewed at HEAD `8887361002`. This is a re-review; the newest commit (the M1 VARCHAR-undercount fix) is the least-reviewed code, so I scrutinized it most: I verified it end-to-end myself, then ran 6 parallel agents and verified every finding against source.

**31 draft findings → 1 Moderate (in-diff), 1 Moderate (out-of-diff), ~15 Minor, 6 test gaps; ~10 false positives dropped.**

## Critical
**None.** No correctness bug, memory-safety violation, leak, double-free, use-after-free, or concurrency hazard in any production-reachable path. The M1 VARCHAR accounting, the VARCHAR_SLICE spill/restore pointer rebasing, the pin/eviction invariant, `cachedBytes` symmetry, and the restore ownership handoff were each traced clean. The Rust side is clean (`cargo check` + `clippy` green, all 5 shipped dylibs 64-bit, `#[repr(C)]` so `offset_of!` is sound, zero production logic changed in the decoders).

## Moderate

### Finding 1 — The byte cap does not bound zero-`decodedBytes` frames; native buffers accumulate unbounded within a scan *(in-diff, regression vs master)*
**Where:** `PageFrameMemoryPool.acquireBuffer():414` interacting with `decode():1446-1448`.

**Path:** When a projection resolves to **zero** parquet columns (`parquetColumns.size() == 0`), `decode()` sets `decodedBytes = 0`, so `decodeAndAccount` adds 0 to `cachedBytes` (line 526). The eviction/reuse loop is gated solely by `if (cachedBytes >= effectiveBudgetBytes)` (line 414). For any positive budget that gate never opens, so each miss takes the allocation path (line 458) and links a fresh `ParquetBuffers` — eagerly allocating 5× `DirectLongList(16)` + a `RowGroupBuffers` (~1 KB native each, ctor at 1363) — into `byFrameIndex` + the LRU, **never reused**. The hot cursors (`AbstractPageFrameRecordCursor`, `TimeFrameCursorImpl`) don't release per frame, so these accumulate across the whole scan, freed only at `toTop`/`close`.

**Reachability:** plain `SELECT col FROM t` where `col` was added (ALTER TABLE ADD COLUMN) *after* the scanned partitions became parquet — the column is absent, reads all-NULL, `parquetColumns` is empty. One buffer held-until-close per row group.

**Why it's a regression:** master's `nextFreeBuffers` held at most `parquetCacheSize` (default/min **8**) buffers and **threw "insufficient memory pool size" rather than allocate beyond the cap** (verified at the merge-base). HEAD removed that hard count ceiling, and zero-payload frames exert no byte pressure to replace it — so this defeats the PR's central promise (a decoded-bytes ceiling bounding RSS) for this query shape. Related sub-gap: the per-buffer bookkeeping overhead (~1 KB) is *never* charged to the budget, so even nonzero-tiny frames let actual RSS overshoot the configured cap.

**Not a leak** (freed at close) and bounded by row-group count — hence Moderate not Critical — but a genuine behavioral regression. The symptom is already visible in your own test comment at `ParquetTest.java:4088-4092`: *"Without this the default 256 MiB budget swallows all tiny partitions, no eviction happens."* (Three independent agents flagged it; the earlier level-3 review rated it Minor — I'm elevating it.)

**Fix:** charge a per-buffer minimum (e.g. the shell overhead) toward `decodedBytes`/`cachedBytes`, or add a buffer-count ceiling alongside the byte budget. Either makes zero-payload and tiny frames exert eviction pressure again.

### M2 — Async window-join slave pools left at MONOTONIC while the sync path and async TopK both use SCATTERED *(out-of-diff, perf-only)*
**Where:** `AsyncWindowJoinAtom.initTimeFrameCursors():590-603` sets up the owner + per-worker slave `TimeFrameCursor`s via `.of(...)` but **never** calls `setParquetDecodeHint`. By contrast `WindowJoinRecordCursorFactory:187` and `WindowJoinFastRecordCursorFactory:240` set the slave SCATTERED, and the analogous async path `AsyncTopKRecordCursor:136` calls `initMemoryPools(..., ParquetDecodeHint.SCATTERED)`. So the async window-join — the parallel, large-data path that matters most — is the *only* async random-access path left at the ¼ (MONOTONIC) budget, re-decoding more than the sync path it mirrors. Correctness is unaffected (a lost hint only resizes the budget); spill is prod-dormant.

**Fix:** in `initTimeFrameCursors`, after each `.of(...)`, call `setParquetDecodeHint(ParquetDecodeHint.SCATTERED)` on `ownerSlaveTimeFrameCursor` and each per-worker slave cursor. (Previously raised in the level-3 review; still open.)

## Minor

**Performance / tradeoffs**
- **MONOTONIC quarter-budget default (`ParquetDecodeHint`, `>>>2` = 64 MB).** For wide-schema / large-row-group workloads where 8 frames fit comfortably on master's flat-8 count cap, the 64 MB byte cap can cache *fewer* than 8 frames → more re-decodes. A genuine workload-dependent tradeoff, not a bug — the description leans on the favorable narrow-frame case; state it plainly and justify quarter (`>>>2`) vs half (`>>>1`) with a wide-schema AsOf-join benchmark. (No degenerate zero-frame case exists — the gate reuses in place rather than rejecting, so the pool always holds ≥1 frame.)
- **N1 — redundant `chunksPtr` reloads** in the new budget loops (`decode:1437-1439`, `sumChunkBytes:1666-1671`): each of `getChunkDataSize`/`getChunkAuxSize`/`getChunkPageBuffersSize` re-reads the base `chunksPtr` (`RowGroupBuffers.java:82/93/107`) → ~6 native loads/slot where 4 suffice. Miss-path only, negligible, but pure waste; hoist `chunksPtr` once per slot. New in this PR.
- **M3 — per-miss mapping recompute** (`openParquet` ~589-616; `remapColumns` ~1616-1633 does 2 hash probes/column). Both are pure functions of file identity + the cursor-stable `ColumnMapping`; `activateDecoder` already detects file change. **Pre-existing on master** (missed optimization, not a regression), but cheap to fix now that the class is being reworked. The hit path itself is a genuine O(1) improvement over master's double-O(n) scan.

**Code quality / standards** (re-verified still open at HEAD)
- `PageFrameMemoryPool.SpillManager` member ordering: `private static` `nameStartsWith`(748)/`rebaseVarcharAux`(777)/`sumVarcharStringBytes`(793) interleave with instance methods; package-private `deleteSpillFilesInDir`(998) follows instance `spill()`. Statics must group before instance methods, alphabetized.
- `PageFrameMemoryPool.java:117` — `new ObjList<>(8)` should be `new ObjList<>(SHELL_POOL_CAP)` (the constant is used at 388; can drift).
- `PageFrameMemoryPool.java:757` — dead branch `int digits = pid == 0 ? 1 : 0;` (`ProcessHandle.current().pid()` is never 0).
- `PageFrameAddressCache.java:57` — `public static volatile IS_COLD_PARQUET_PARTITION_FORCED_FOR_TEST` lacks `@TestOnly`.
- `Counter.reset()` lost its `@TestOnly` to let `ParquetDecodeMetrics.clear()` call it "in production" — but `Metrics.clear()` has **no production caller** (every caller is test teardown), so the annotation could have stayed; the premise of the removal is wrong.
- **Scope creep:** `Bootstrap.java` (`_` unnamed lambda params, `sb.repeat`) and the `CairoConfiguration.java` posting-index method reorder are unrelated to the parquet cache.
- `pkg/ami/marketplace/assets/server.conf` drops the obsolete `frame.cache.capacity` key but doesn't add the 3 new keys the main template documents.
- **PR description inaccuracy:** "15 factories declare it; the 16th (SortKeyMaterializing) does not call `recordAt`" — there are **17** declaring factories, and `SortKeyMaterializingRecordCursor.recordAt:204` **does** call `baseCursor.recordAt` (it forwards the hint, so the conclusion holds but the rationale is false).
- Commit titles `7604db32a0`/`3e2a6e4df0`/`c77dd1a69b` are bare with no body (low impact on a squash branch, but the squashed message must not inherit the vagueness).

**Test coverage gaps**
- **G1** — Finding 1 (zero-byte accumulation) is unasserted: `testSelectOnlyAddedColumnAbsentFromParquetPartition` checks only the NULL result, not buffer-count bounding.
- **G2** — the M1 varchar-bytes budget integration has **no Java assertion** (only the Rust unit test + a non-CI benchmark). `cachedBytes` has no getter and there's no byte metric, so a regression dropping the `getChunkPageBuffersSize` term — undercounting the heaviest column type, the fix's whole point — would pass CI. Add a `cachedBytes()` hook + an assertion that a varchar scan reports materially more than data+aux alone.
- **G3** — AsyncTopK SCATTERED + the restore-then-late-mat path (`assert !isRestored`) is entirely untested; single-worker tests route `ORDER BY ... LIMIT` to `LimitedSizeSortedLight`, and `ParallelTopKFuzzTest` enables no spill.
- **G4** — `SortedLightRecordCursorFactory:97`, `SortedRecordCursorFactory:102`, `LongTopKRecordCursorFactory:73` SCATTERED callsites are not actually exercised (no-limit `ORDER BY` routes to the EncodedSort variants).
- **G5** — no spill round-trip for `HashOuterJoin*`, `NestedLoop{Left,Right,Full}`, or sync `WindowJoin*` SCATTERED factories.
- **G6** — the varchar NULL/empty/inline spill round-trip (`ParquetLateMaterializationFuzzTest`) asserts correctness but not `restores() > 0`, so it doesn't prove `rebaseVarcharAux` was exercised.

## Downgraded (false positives, verified against source)
- *M1 fix double-counts / reads a stale `page_buffers_size`* — false; `refresh_ptrs()` re-sums from scratch as the last decode step, `reset()` zeroes it, every drain is immediately followed by `reset()`, and it's added only at the two budget sites (not the page-address remap).
- *`assert !isRestored` is a reachable crash* — false; restore (recordB chain hops) and late-mat (fresh decode) never land on the same buffer in any prod path, and under `-da` it degrades to a catchable `CairoException`, not a JVM abort.
- *VARCHAR_SLICE spill dereferences inlined strings / corrupts NULL/empty* — false; the slice aux is always pointer-based; the helpers skip NULL and handle empty/long correctly.
- *`cachedBytes` drift / double-free; spill fd/file/`bodyPtr` leak* — false; every add has a matching subtract, `evictHalfInitialized` + idempotent `close()` free exactly once, and fd/file/`bodyPtr` are released on all error branches.
- *Eviction victim scan is O(cache size)* — false; ≤3 pins, so it skips ≤3 nodes (effectively O(1)).
- *`clear()` calling `spillManager.close()` leaks shells on a spill-enabled pool* — false; both `clear()` callers construct with budget `0L` → `spillManager == null`.
- *Rust `.sum()` overflow / `usize`→`long` sign issue* — false at any realistic scale; all shipped targets 64-bit.
- *Eager-restore / per-slot metadata work on MONOTONIC cursors* — gated by `spillManager != null`, null in default config; no wasted work on the default path.

## Verdict
**Request changes** — the changes are memory-accounting, perf-consistency, and coverage, **not correctness or safety**. The hard parts are correct, and the newest commit (the M1 Rust `page_buffers_size` fix) is sound end-to-end.

Two items genuinely matter: **Finding 1** (the byte cap doesn't bound zero-payload frames — a regression from master's hard count cap; fix by charging a per-buffer minimum, which also covers the uncounted bookkeeping overhead) and **M2** (wire SCATTERED into `AsyncWindowJoinAtom.initTimeFrameCursors`). The rest are docs, the known minors, and coverage — most importantly an assertion that the byte budget actually bounds VARCHAR RSS (G2), since that's the entire point of the latest commit and nothing in CI currently proves it.

<sub>Reviewed with Claude Code (level 2).</sub>

kafka1991:
> @kafka1991 this is performance related PR, can you add benchmarks before and after?

Hey @bluestreak01  Benchmark Added — see [Benchmarks](https://github.com/questdb/questdb/pull/7197#benchmarks) section in the PR description. SCATTERED-hinted sort paths show 12-31% wall-time reductions on a 502M-row, 240-partition parquet workload; sequential aggregations sit within noise as expected.

kafka1991:
## Review report — Level 3 mission-critical pass

**Methodology:** 9 parallel review agents (correctness, concurrency, perf×2, resources, tests, code quality+metadata, Rust safety, cross-context, fresh-context adversarial) produced 22 candidate findings. Self-verification against source dropped 6 as false positives and downgraded the severity of 3 more.

---

### Critical
None. No correctness regression, no resource leak, no race, no panic site. Safe to merge.

---

### Moderate

#### M1. `coldParquetPartitions` not refreshed by `PageFrameAddressCache.updateAddresses` (latent)

- **Location:** `core/src/main/java/io/questdb/cairo/sql/PageFrameAddressCache.java:271-285`
- **Trace:** `TimeFrameCursorImpl.addUninitializedFrame` / `ConcurrentTimeFrameState.addUninitializedFrame` pre-populate the cache with placeholder NATIVE entries — `PageFrameAddressCache.add` line 114 only sets `isCold` on the PARQUET branch. When `ensurePartitionOpened` later patches via `updateAddresses(idx, realFrame)`, the method refreshes column addresses and parquet decoder but does not refresh `coldParquetPartitions`. Parquet partitions opened through the lazy TimeFrameCursor path stay `isCold=false` forever.
- **Impact today:** none — `PageFrame.isColdParquetPartition()` is a documented stub returning `false` everywhere, so the gate is unreachable. The two AsyncWindowJoin spill tests pass because they take the eager-build branch.
- **Impact tomorrow:** once cold-tier metadata is wired, AsyncWindowJoin / HorizonJoin / AsOf-style slaves over cold parquet partitions on the lazy path will silently fail to engage disk spill.
- **Fix:** in the `updateAddresses` PARQUET branch, add `coldParquetPartitions.set(frameIndex, frame.isColdParquetPartition() || isColdParquetForcedSnapshot)`. One line; worth doing now while the lazy-path contract is fresh.

---

### Minor

#### m1. Subclass `close()` virtual dispatch in `AbstractPageFrameRecordCursor` constructor can NPE on uninitialized subclass field

- **Location:** `core/src/main/java/io/questdb/griffin/engine/table/AbstractPageFrameRecordCursor.java:55-63` × `core/src/main/java/io/questdb/griffin/engine/table/LatestByValueListRecordCursor.java:85-94`
- The PR's new try/catch in the base ctor calls virtual `close()` on throw. `LatestByValueListRecordCursor.close()` accesses `rowIds.getCapacity()` without null-check; `rowIds` is `final`, assigned in the subclass ctor body, still `null` if the base ctor throws (e.g. native OOM in a new `DirectLongList`, or `mkdirs` failure on the spill dir).
- **Net:** the PR is an improvement over master (master leaked the partially-constructed fields with no cleanup at all). The remaining issue is only that the NPE masks the original exception for diagnostics.
- **Fix:** `if (rowIds != null && rowIds.getCapacity() > shrinkToCapacity)` in the subclass close.

#### m2. Static field ordering in `PageFrameMemoryPool`

`core/src/main/java/io/questdb/cairo/sql/PageFrameMemoryPool.java:101-108` — `PID` precedes `OWN_PID_PREFIX_BYTES` (alphabetical order is the opposite). Load-bearing because the prefix initializer reads PID. Fix: inline `ProcessHandle.current().pid()` into the prefix initializer, or move into a static block, then reorder per CLAUDE.md.

#### m3. Missing `rust` label on the PR

9 Rust files modified plus 5 rebuilt native libraries.

#### m4. 10-digit literal without underscores in tests

`core/src/test/java/io/questdb/test/griffin/ParquetTest.java:2090, 2092, 3630, 3632` — `3600000000` → `3_600_000_000` per CLAUDE.md.

#### m5. PR title narrowed

The title understates the work — the PR also adds access-pattern hints, true LRU, disk-spill scaffolding, four metrics, and a Rust-side VARCHAR_SLICE accounting fix. Consider broadening for release notes (e.g. `feat(sql): cap Parquet decode cache by bytes and add access-pattern hints`).

#### m6. Over-conservative SCATTERED hint at two sequential-walk callsites

`core/src/main/java/io/questdb/griffin/engine/window/CachedWindowRecordCursorFactory.java:205` and `core/src/main/java/io/questdb/griffin/engine/orderby/EncodedSortRecordCursorFactory.java:71` tag their base cursor SCATTERED, but `CachedWindowRecordCursor.buildRecordChain` and `EncodedSortRecordCursor` only iterate the base via `hasNext()` (no `recordAt`). The hint over-allocates the parquet decode-cache byte budget that the sequential build phase never exercises. MONOTONIC matches the actual access pattern. Verify intent.

#### m7. `byFrameIndex` always sized for SCATTERED (256 slots) on MONOTONIC pools

`PageFrameMemoryPool.java:163` — MONOTONIC pools (4-entry working set) carry the 256-slot hash table for their lifetime. ~4 KB ref array per pool, multiplied by concurrent cursors. Not load-bearing, just noted.

#### m8. Test coverage gaps

- **All-pinned eviction overflow** — no targeted test exercises the fall-through `new ParquetBuffers()` allocation when every cached buffer is pinned.
- **`populateRemainingColumns` failure path** — the `cachedBytes -= prevDecodedBytes` / `evictHalfInitialized` rollback on throw has no direct test injecting a decoder failure during remaining-columns decode.
- **SCATTERED-hint factories without spill round-trip tests:** `NestedLoopFullJoinRecordCursorFactory`, `NestedLoopRightJoinRecordCursorFactory`, `WindowJoinFastRecordCursorFactory`, `LongTopKRecordCursorFactory`, `LimitedSizeSortedLightRecordCursorFactory`, `HashOuterJoinFilteredLightRecordCursorFactory` all call `setParquetDecodeHint(SCATTERED)` but no spill round-trip test selects their plan.
- **VARCHAR NULL spill:** `testParquetCacheVarcharRoundTripsThroughSpill` uses only non-null strings (NULL-skip branch of `materializeVarcharAppend` is covered indirectly by `ParquetLateMaterializationFuzzTest` but not in the focused varchar spill test).

---

### Downgraded (false positives, dropped after source verification)

- **`releaseParquetBuffers` wipes spill files mid-reduce → wasted I/O.** Withdrawn. The PR description says explicitly: "restore runs only on recordB chain-comparison hops to already-reduced frames". The spill→restore cycle is intra-frame by design — within a single reduce, evicted frames spill and are restored by chain comparator hops. End-of-frame `releaseParquetBuffers` clearing files is correct, otherwise disk usage would be unbounded. Confirmed by `testParquetCacheAsyncTopKSpillRoundTrip` asserting both `spills > 0` AND `restores > 0` in a single reduce pass.
- **Hoist per-decode slot-meta copy to `openParquet`.** Withdrawn. `parquetColumns` is a pool-level mutable list that grows when `populateRemainingColumns` appends late-materialization columns. Each buffer must freeze its own decode-time column set because spill (which can happen long after decode) needs the snapshot at decode time.
- **`maxCacheBytes == 0` pool grows unbounded.** Withdrawn. `tryHit` clears the prior usage-bit pin before `acquireBuffer` runs. With three usage bits, at most three frames are pinned; eviction always finds an unpinned victim. Cache stays at ≤ 3 entries.
- **`nameMatchesOwnPid` may OOB-read short filenames.** Withdrawn. The kernel dirent `d_name` buffer is fixed at `NAME_MAX+1` (256 bytes); reading 25 bytes is always within the struct. Early-return on first mismatching byte handles the short-name case.
- **`spill()` irreversibly mutates VARCHAR_SLICE aux entries before file write.** Withdrawn. `acquireBuffer` always immediately re-decodes the victim after eviction; the Rust `ColumnChunkBuffers::reset` clears `data_vec`/`aux_vec` on the next decode.
- **Cache key is `frameIndex` only — full-set vs subset decode mismatch.** Withdrawn. No current caller interleaves full-set and subset `navigateTo` against the same pool.
- **`releaseParquetBuffers` shell-push leaves records dangling.** Withdrawn. All callers are in finally blocks where records are explicitly cleared first.

---

### Strengths

- **Resource ownership protocol on the spill restore path** is exemplary: `isOwnershipTransferred` / `isRestoreSucceeded` double-flag guards ensure exactly the right side frees `bodyPtr` on every error path.
- **Corrupted-spill-file test matrix** covers 10 distinct guards (magic, version, slotCount-too-large, slotCount-below-current, schema-prefix mismatch, negative slot size, oversized body, short header read, short body read, truncated file) — each verified to reject and degrade to re-decode cleanly.
- **PID-prefix isolation** for spill file cleanup correctly protects sibling QuestDB processes sharing a host.
- **Hot/cold × MONOTONIC/SCATTERED quadrant** fully covered by negative-assertion tests.
- **PR description tone** is level-headed: the +3% ASOF JOIN regression and 0L-budget aggregation neutrality are reported with the same prominence as the -31% sort speedup. The memory-footprint formula (`effectiveBudget × concurrent cursors`) is published so operators can size correctly.
- **Cross-context audit** of the 42 `RecordCursor` impls, 7 `PageFrame` impls, and 3 `TimeFrameCursor` impls found zero forwarding gaps.

---

### Summary

- **22 draft findings → 8 confirmed (1 moderate, 7 minor) + 7 false positives + 1 internal severity downgrade.**
- **Verdict: approve and merge.** Land M1's one-line `coldParquetPartitions` refresh before the cold-tier wiring lands; everything else is optional polish.
- Zero blocking issues. Zero correctness regression, zero memory leak, zero race, zero panic site.

puzpuzpuz:
@RaphDal could you check the disk spill changes from this PR? To me, cold storage's own caching is doing pretty much the same thing sans parquet-to-native decoding.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 905 / 1011 (89.52%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/sql/TimeFrameCursor.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/QueryProgress.java|0|2|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/DistinctTimeSeriesRecordCursorFactory.java|0|3|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ExtraNullColumnCursorFactory.java|1|5|20.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncFilteredNegativeLimitRecordCursor.java|5|10|50.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AbstractPageFrameRecordCursor.java|7|10|70.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncFilteredRecordCursor.java|8|11|72.73%|
|:large_blue_circle:|io/questdb/Bootstrap.java|3|4|75.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/RowGroupBuffers.java|10|13|76.92%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|655|735|89.12%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameAddressCache.java|11|12|91.67%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinLightNoKeyRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByValueListRecordCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinDenseRecordCursorFactoryBase.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncFilterContext.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/WindowJoinRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortLightRecordCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/NestedLoopLeftJoinRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FilteredRecordCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/MultiHorizonJoinNotKeyedRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/union/AbstractSetRecordCursor.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/AbstractVirtualFunctionRecordCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/NestedLoopRightJoinRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ConcurrentTimeFrameCursorImpl.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/ParquetDecodeMetrics.java|23|23|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/rle_dict_varchar_slice.rs|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortedLightRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/HashOuterJoinLightRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LimitedSizeSortedLightRecordCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/DefaultCairoConfiguration.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ExtraNullColumnRecordCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinAtom.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LimitedSizePartiallySortedLightRecordCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFirstLastRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/PropertyKey.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKAtom.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/HorizonJoinNotKeyedRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SpliceJoinLightRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/vect/GroupByRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyMaterializingRecordCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/ParquetDecodeHint.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFillRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/jni/buffers.rs|6|6|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/delta_binary_packed.rs|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/MultiHorizonJoinRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/plain.rs|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrame.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/NestedLoopFullJoinRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/LtJoinLightRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FilterOnSubQueryRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/WindowJoinFastRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/rle_dictionary.rs|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/rle.rs|1|1|100.00%|
|:large_blue_circle:|io/questdb/PropServerConfiguration.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKRecordCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestBySubQueryRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/Metrics.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortLightRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LongTopKRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/delta_length_array.rs|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decode.rs|21|21|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/RecordAsAFieldRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LimitedSizeSortedLightRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/CrossJoinRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortedLightRecordCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByLightRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/VirtualFunctionRecordCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressProcessorState.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/tasks/LatestByTask.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/LtJoinNoKeyRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinFastRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/async/PageFrameReduceTask.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/TimeFrameCursorImpl.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortedRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/BaseAsyncHorizonJoinAtom.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoConfiguration.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/LimitRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/window/CachedWindowRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/HorizonJoinRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/HashOuterJoinFilteredLightRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LongTopKRecordCursor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/HashJoinLightRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/RecordCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinLightRecordCursorFactory.java|1|1|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=240516)
**Tested Commit:** `1201cd888e5667e157ecfc20bb18bbe5211a3cd9`

kafka1991:
The memory budget part has been split out into a separate PR: https://github.com/questdb/questdb/pull/7230. The disk spill part won't be merged into OSS.

PR #7200, opened 2026-06-03
chore(qwp): simplify the wire protocol to one version with inline schemas

mtopolnik:
Tandem with questdb/java-questdb-client#39. Merge that PR first, then re-point the `java-questdb-client` submodule to its main branch before merging this PR.

Tandem with https://github.com/questdb/questdb-enterprise/pull/1041. Merge this PR first, then update `questdb` module, then merge Enterprise PR.

## What

Removes the QWP schema-reference mechanism and collapses the protocol to a single version (drops the v2 designation) on the server. Column schemas now always travel inline on the wire.

This is the server (Phase 1) half of a tandem change; the Java client half is questdb/java-questdb-client#39. `java-questdb-client` submodule pointer is set at that PR, so end-to-end tests see it.

## Why

The schema-reference optimisation let a client register a column schema once and reference it by id on later frames. Store-and-forward retry made the server's registry and the client's view drift apart -- frame N+1 can land while frame N is still being resent -- so clients already stopped using it. The server still maintained the registry and the wire still carried a mode byte + schema id on every table block.

## Wire format

Ingress table block: drops the mode byte and `schema_id`; columns are always inline (`name_len, name, type_code` per column).

Egress `RESULT_BATCH`: the schema rides only `batch_seq == 0`; continuation batches carry rows only. The empty-result case still emits one `batch_seq == 0` with `row_count == 0`, which carries the schema.

The per-frame version byte stays at offset 4, written as `1` and validated `== 1`. The `X-QWP-Max-Version` handshake negotiation is retained for a future bump.

## Removed

- `QwpSchemaRegistry` (ingress) and `QwpEgressSchemaWriter` (egress fingerprint dedup + reference emission).
- The egress schema-fingerprint cache and its 4096 soft cap; the `RESET_MASK_SCHEMAS` half of `CACHE_RESET` (the dict half stays).
- Protocol-v2 constants (`VERSION_2`, `MAX_SUPPORTED_*`) collapse to `VERSION = 1`.
- `qwp.max.schemas.per.connection` server property and its plumbing.
- The in-repo `docs/qwp/*.md` design docs (the authoritative spec moved to the documentation repo); code references repoint to https://questdb.com/docs/.

## Trade-offs

- Re-sent schemas: every result re-ships its column descriptors in batch 0, even for a repeated query against the same table -- a few dozen wire bytes plus a client-side re-parse and per-column String alloc per batch 0.
- The empty-result invariant becomes load-bearing: every cursor path must emit at least one `RESULT_BATCH` so the client receives the schema. Covered by tests, including the reused-schema empty case.
- Wide schema + small send buffer: the full inline schema always rides batch 0, so a query whose schema alone exceeds the egress send-buffer budget now fails with a controlled error where a second query against an already-registered schema could previously fit a tiny reference. Guarded (no corruption); remedy is a larger send buffer.

## Test plan

- `mvn -Dtest='Qwp*' -pl core -Plocal-client test` (QWP unit + e2e suites)
- Schema-registry / schema-id-exhaustion / schema-ref tests removed; bounds-check, cursor, and hardening coverage retained.
- Empty-result and reused-schema empty-result invariants covered in `QwpEgressBootstrapTest`.
- Wire-level interop under `cutlass-test-suite` greens only after the submodule pointer reaches the client branch (already bumped on this branch).


mtopolnik:
# Review: PR #7200 — `refactor(qwp): simplify the wire protocol to one version with inline schemas`

**Reviewing at level 3** (full mission-critical pass: change-surface map, all applicable agents, per-finding verification).

This is a large diff (~9,200 lines) but the production change is small and cohesive: it deletes the QWP schema-reference optimization so column schemas always travel inline, and collapses two protocol versions to one. The bulk of the diff is deleted docs and mechanical test updates. I read the core egress/ingress logic myself, dispatched 6 review agents (correctness+cross-context, concurrency+resource, structured perf, adversarial perf, tests, code-quality+metadata), and one fresh-context adversarial agent that independently `clean test-compile`d core and ran ~340 QWP tests — **all green**.

## Critical
**None.** The load-bearing change — `isFirstBatch = state.getStreamingBatchSeq() == 0`, replacing the old `writeFullSchema`/`streamingFullSchemaSent` flag — is correct across every path I traced and re-verified: multi-batch (schema rides seq 0 only), **partial emit on batch 0** (prefix ships as seq 0 with schema, suffix carries to seq 1 without, `beginBatch` not re-run on the suffix), **PISR park+resume** (`onStreamingBatchSent` increments and rows advance before the throw; bytes are committed, so resume never re-emits the schema or double-counts the seq), **empty result** (the `getStreamingBatchSeq() > 0` guard keeps the first iteration from short-circuiting, so one RESULT_BATCH with row_count 0 + schema is always sent), and **wide-schema/tiny-buffer** (`findLargestEmittablePrefix` returns −1 → controlled `QwpRowExceedsBufferException`, no corruption). Wire size↔emit byte-identity holds (`assert p - startP == schemaBytes`; `columnCount == columns.size()` set together in `beginBatch`). Ingress bounds checks are intact. No dangling references to any removed symbol in `core`. No leaks (the removed `Utf8StringSink`/`Utf8SequenceIntHashMap` were on-heap; the deleted `SCHEMA_ID_EXHAUSTED` free-path is covered by the generic `catch (Throwable)` that frees factory/cursor when `!streamingHandedOff`).

## Moderate
None blocking. The one item with any engineering weight is the efficiency nit below; I've placed it in Minor because of its narrow trigger.

## Minor

**1. (Efficiency, in-diff) `inlineColumnsSize()` is recomputed across the partial-emit binary search though it's invariant.**
`core/src/main/java/io/questdb/cutlass/qwp/codec/QwpResultBatchBuffer.java:1116-1118`. When batch 0 overflows the send buffer, `findLargestEmittablePrefix` calls `computeTableBlockSize(K, true)` ≈ `3 + ⌈log2 R⌉` times, and each call recomputes `colsVarLen + inlineColumnsSize()` — an O(columns) walk — even though the schema byte count depends only on `columns` (fixed for the batch), not on `K`. Both perf agents derived this independently; I confirmed there is no cached field. Severity is genuinely low: it fires only on a first batch that overflows the buffer, the per-column walk is cheap (`getNameUtf8()` returns a pre-encoded `byte[]`, `QwpVarint.encodedLength` is branchless), and the *dominant* per-probe cost is actually the pre-existing per-column null-bitmap popcount in `emitColumnSlice` (out of this PR's scope). But the dedup removal collapsed the old two-branch schema sizing into a single source, so the optimal fix is now a clean one-liner: compute `colsVarLen + inlineColumnsSize()` once in `beginBatch` into an `int` field and read it in `emitTableBlockImpl`. Worth doing while this code is open.

**2. (Doc consistency, in-diff) Five assertion messages still cite a deleted doc.**
`core/src/test/java/io/questdb/test/cutlass/qwp/QwpEgressColumnTopNonNullableTypesTest.java:110,149,165,181,197` still say `"... (wire-egress.md sec 11.5)"`. The PR repointed this file's class Javadoc to the questdb.com URL but missed these inline strings — directly contradicting the PR's stated "repoint code references to https://questdb.com/docs/". Cosmetic; replace the substring with the same URL/phrasing used in the header.

**3. (Doc consistency, out-of-diff) `failover.md` reference left dangling.**
`core/src/test/java/io/questdb/test/cutlass/qwp/QwpQueryClientFailoverLoopTest.java:49,274` reference `failover.md §4.3` and `§6`, which this PR deletes (`docs/qwp/failover.md`). Repoint to `https://questdb.com/docs/high-availability/client-failover/concepts/` (the URL the production `QwpEgressMsgKind` Javadoc now uses).

**4. (Test gaps) Small, verified holes — none blocking:**
- **`columnCount == 0` parse is untested.** The refactor dropped the `if (length < 1)` guard from `QwpSchema.parse`; `QwpSchemaTest` covers counts 1/2/3/10 but never 0, so the new "empty schema, 0 bytes consumed" behavior has no direct assertion. (`testColumnNameEmpty` tests an empty *name* at count 1, not count 0.)
- **Egress-side version cap untested.** `QwpEgressServerInfoTest.testNegotiatesVersion` uses a client that only advertises v1, so `QwpEgressUpgradeProcessor.negotiateQwpVersion`'s `Math.min(clientMax, VERSION)` is never exercised with a `>1` header. Only the *ingress* negotiator is tested (`testVersionNegotiationReturnsServerMax`). The two are separate methods.
- **Partial-emit-on-batch-0 has no direct, intention-revealing assertion.** It is functionally covered end-to-end (`QwpEgressWideSchemaBatchCapTest.testWideSchemaQueryDoesNotOverflowBatch` would fail decode if the suffix re-emitted the schema), but no test decodes the seq-1 frame to assert the schema is absent. Given it's the riskiest path, a targeted assertion is advisable.
- **`testContinuationBatchOmitsSchema`** uses column names ≤100 chars (all 1-byte `name_len` varints). Adding a ≥128-char name would cover the 2-byte varint boundary in `inlineColumnsSize()`.
- **`QwpMessageHeaderTest.testUnsupportedVersion`** rejects 0 and 255 but not 2 — the exact boundary this PR changes (v2 is correctly rejected via `version != VERSION`).

## Downgraded (verified non-issues)
- **"Always-send SERVER_INFO + continuation batches drop col_count breaks v1 egress clients."** By design — this is the whole point of the single-version collapse, it's a tandem change with `java-questdb-client#39` (the fresh-context agent confirmed the submodule decoder matches the new layout), and the PR is labeled `Compatibility` and documents the merge order. QWP is Phase-1/experimental. Not a defect.
- **`columnCount == 0` consumes 0 schema bytes.** Pre-existing downstream behavior (old code only consumed the mode byte + schema-id varint); not a regression introduced here.
- **Per-probe popcount/Gorilla re-scan in the binary search.** Real algorithmic cost (O(columns × rows × log rows) on the overflow path) but entirely pre-existing and untouched by this PR — out of scope. Flagging only as context for the Minor #1 fix.
- **Suggested `ILP` label.** Declined — QWP is QuestDB's own wire protocol, distinct from InfluxDB Line Protocol; `ILP` would mislabel it. Existing `Compatibility` + `Core` + `tandem` are appropriate.

## Summary
**Verdict: approve** — clean, careful, well-tested refactor; no correctness, concurrency, or resource defects. **9 findings verified, ~5 draft findings dropped** as by-design/pre-existing/out-of-scope. **In-diff vs out-of-diff: 8 in-diff (the perf nit, 1 doc-ref, the test gaps), 1 out-of-diff (the `failover.md` reference in `QwpQueryClientFailoverLoopTest`).** Everything verified is cosmetic or a test-coverage nit — none blocks merge on code grounds.

The one real gate is **operational, not code**: the `java-questdb-client` submodule pointer currently sits on the feature branch (`a3ec95d` / `mt_remove-schema-ref`) and must move to the client's main branch after `java-questdb-client#39` merges, per the PR body's documented merge order. The wire-interop tests under `cutlass-test-suite` only validate against the client main branch.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 63 / 64 (98.44%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressUpgradeProcessor.java|12|13|92.31%|
|:large_blue_circle:|io/questdb/cutlass/qwp/codec/QwpResultBatchBuffer.java|31|31|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/protocol/QwpTableBlockCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpIngressUpgradeProcessor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpStreamingDecoder.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpIngressHttpProcessor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpUdpReceiver.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/protocol/QwpMessageHeader.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressProcessorState.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/protocol/QwpSchema.java|5|5|100.00%|

bluestreak01 (approved):

PR #7202, opened 2026-06-03
fix(sql): fix count() over a GROUP BY subquery reporting phantom duplicates

### LINKED ISSUE #7201: Wrong result: count() over a GROUP BY subquery reports duplicates that do not exist (WAL dedup table)
### Description

On a WAL table with `DEDUP UPSERT KEYS(ts, s)` holding 12 **distinct** `(ts, s)` rows (all sharing one timestamp), a `count()` over a `GROUP BY (ts, s)` subquery filtered to `c > 1` returns a non-zero value — i.e. it reports duplicate groups that do not exist. The materialized form of the same subquery is correct, so the two query shapes disagree on identical data.

### Steps to reproduce

```sql
CREATE TABLE tab (ts TIMESTAMP, s SYMBOL INDEX)
  TIMESTAMP(ts) PARTITION BY DAY WAL
  DEDUP UPSERT KEYS(ts, s);

INSERT INTO tab
  SELECT '2020-02-27T12:00:00.000000Z'::TIMESTAMP, ('sym' || (x - 1))
  FROM long_sequence(12);

-- 12 (correct)
SELECT count() FROM tab;

-- BUG: expected 0 (all 12 (ts,s) pairs are unique) — returns 1
SELECT count() dups FROM (SELECT ts, s, count() c FROM tab) WHERE c > 1;

-- Both of these confirm there is no real duplicate:
SELECT ts, s, count() c FROM tab ORDER BY c DESC, s;                 -- 12 rows, every c = 1
SELECT ts, s, c FROM (SELECT ts, s, count() c FROM tab) WHERE c > 1; -- empty
```

### Expected

`SELECT count() dups FROM (SELECT ts, s, count() c FROM tab) WHERE c > 1` returns `0`.

### Actual

Returns `1`. A phantom group with `c > 1` is counted, even though the materialized group-by (`SELECT ts, s, count() c ...`) shows all 12 symbols with `c = 1` and the row-projection over the same subquery is empty. The discrepancy is isolated to the non-materializing `count()`/size path over the keyed `GROUP BY` subquery.

### Notes

The fault is layout/timing sensitive — it reproduces reliably under memory pressure (e.g. within a larger test run) but on a freshly started instance running only this snippet it may take several executions to trip. The core defect is the disagreement between the `count(subquery)` result and the materialized subquery on identical data.

### Version

QuestDB 9.4.1-SNAPSHOT

ideoma:
Fixes #7201

## What was wrong

`count()` over a keyed `GROUP BY` subquery, filtered on the aggregate alias (a HAVING-style predicate), reported duplicate groups that do not exist. For example, on a table holding 12 distinct `(ts, s)` rows:

```sql
-- expected 0 (all 12 pairs are unique), returned 1
SELECT count() dups FROM (SELECT ts, s, count() c FROM tab) WHERE c > 1;
```

The materialized projection over the same subquery (`SELECT ts, s, c FROM (...) WHERE c > 1`) was empty and the bare group by (`SELECT ts, s, count() c FROM tab`) showed all 12 symbols with `c = 1`, so the two query shapes disagreed on identical data.

## Root cause

The query plan showed the keyed group by had been replaced by a scalar `count()`:

```
Count
    Filter filter: 1<c
        Count            <- subquery collapsed to count() over the whole table = 12
            PageFrame
```

The `WHERE c > 1` is pushed into the subquery as a filter on the group by model. `propagateTopDownColumns0()` has a guard that re-adds a group by sub-query's grouping keys to its top-down columns so column pruning cannot drop them, but it ran before the model's own WHERE/HAVING and ORDER BY literals had been emitted into the top-down list. For the `count()` shape the outer query contributes no key literals, so the HAVING alias `c` was the sole top-down contributor and was added only after the guard had already observed an empty list and skipped. The keys `ts` and `s` were left unprotected, pruning collapsed the keyed group by into a scalar aggregate, and `12 > 1` falsely counted one group. The materialized projection and the bare group by both contribute key literals, which is why only the non-materializing `count()`/size path was affected.

## Fix

`retainGroupByKeysAsTopDownColumns()` now runs twice: once at the original early position, which preserves key ordering for the common case where the parent already contributed top-down columns, and once more after the WHERE/HAVING and ORDER BY literals have been emitted, which covers the HAVING-only contributor case. `addTopDownColumn()` dedupes by alias, so the second pass is idempotent and adds keys only where the early pass missed them.

This is purely additive relative to the previous behavior: keys can only be added where the original pass missed them, never reordered or removed. There is no measurable performance change — it only affects query compilation, adding one idempotent pass over a model's columns.

## Test plan

- Added regression test `CountTest#testCountOverKeyedGroupBySubqueryWithHavingFilter`, which asserts the plan keeps the keyed group by, that the buggy query returns `0`, that a genuine duplicate is still reported (`dups=3`, `sum=12`), and that the unaffected shapes (`count()=12`, empty projection, single-key) stay correct.
- Ran the optimizer-adjacent suites with zero failures: CountTest, GroupByTest, SqlOptimiserTest, ExplainPlanTest, SampleByTest, UnionTest, DistinctTest, SelectTest, OrderByTest, SubQueryTest, SqlParserTest, WindowFunctionTest, JoinTest, HorizonJoinTest, LimitTest, LatestByTest, NestedSetOperationTest, CoveringIndexTest, MatViewTest.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/SqlOptimiser.java
+++ b/core/src/main/java/io/questdb/griffin/SqlOptimiser.java
@@ -6215,17 +6215,14 @@ public class SqlOptimiser implements Mutable {
             propagateTopDownColumns0(jm, false, model, true);
         }
 
-        // If this is group by model we need to add all non-selected keys, only if this is sub-query
-        // For top level models top-down column list will be empty
-        if (model.getSelectModelType() == IQueryModel.SELECT_MODEL_GROUP_BY && model.getTopDownColumns().size() > 0) {
-            final ObjList<QueryColumn> bottomUpColumns = model.getBottomUpColumns();
-            for (int i = 0, n = bottomUpColumns.size(); i < n; i++) {
-                QueryColumn qc = bottomUpColumns.getQuick(i);
-                if (qc.getAst().type != FUNCTION || !functionParser.getFunctionFactoryCache().isGroupBy(qc.getAst().token)) {
-                    model.addTopDownColumn(qc, qc.getAlias());
-                }
-            }
-        }
+        // If this is a group by sub-query whose projection will be pruned to its top-down columns,
+        // retain all non-selected grouping keys so pruning cannot collapse the keyed group by into a
+        // scalar aggregate. This early pass preserves the key ordering for the common case where the
+        // parent already contributed top-down columns; it is repeated once more below, after the
+        // model's own WHERE/HAVING and ORDER BY literals have been emitted, to also cover a HAVING-style
+        // filter that references only an aggregate alias (its literals are the sole top-down contributor,
+        // so an empty list here would skip retention). For top level models the list is empty -> no-op.
+        retainGroupByKeysAsTopDownColumns(model);
 
         // latest on
         if (model.getLatestBy().size() > 0) {
@@ -6258,6 +6255,14 @@ public class SqlOptimiser implements Mutable {
         if (!topLevel) {
             emitLiteralsTopDown(model.getOrderBy(), model);
         }
+
+        // Repeat group by key retention now that the model's own WHERE/HAVING and ORDER BY literals
+        // have been emitted into top-down columns. This covers HAVING-style filters on a group by
+        // sub-query whose only top-down contribution is an aggregate alias (e.g. count() > 1); the
+        // early pass above would have observed an empty list and skipped, letting pruning collapse the
+        // keyed group by into a scalar aggregate. addTopDownColumn() dedupes by alias, so re-running is
+        // idempotent and leaves the key ordering from the early pass untouched in the common case.
+        retainGroupByKeysAsTopDownColumns(model);
 
         if (nestedIsFlex && nestedAllowsColumnChange) {
             emitColumnLiteralsTopDown(model.getColumns(), nested);
@@ -7248,6 +7253,23 @@ public class SqlOptimiser implements Mutable {
                 }
             }
             node = sqlNodeStack.poll();
+        }
+    }
+
+    // Adds every non-aggregate grouping key of a GROUP BY model to its top-down column list, so that
+    // top-down column pruning cannot drop the keys that define the grouping. Runs only when the model
+    // already has top-down columns, i.e. when it is a sub-query whose projection will be pruned; for a
+    // top level model the top-down list is empty and the bottom-up projection is used verbatim, so there
+    // is nothing to protect. addTopDownColumn() dedupes by alias, making repeated calls idempotent.
+    private void retainGroupByKeysAsTopDownColumns(IQueryModel model) {
+        if (model.getSelectModelType() == IQueryModel.SELECT_MODEL_GROUP_BY && model.getTopDownColumns().size() > 0) {
+            final ObjList<QueryColumn> bottomUpColumns = model.getBottomUpColumns();
+            for (int i = 0, n = bottomUpColumns.size(); i < n; i++) {
+                QueryColumn qc = bottomUpColumns.getQuick(i);
+                if (qc.getAst().type != FUNCTION || !functionParser.getFunctionFactoryCache().isGroupBy(qc.getAst().token)) {
+                    model.addTopDownColumn(qc, qc.getAlias());
+                }
+            }
         }
     }
 
--- a/core/src/test/java/io/questdb/test/griffin/engine/groupby/CountTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/engine/groupby/CountTest.java
@@ -154,6 +154,57 @@ public class CountTest extends AbstractCairoTest {
     }
 
     @Test
+    public void testCountOverKeyedGroupBySubqueryWithHavingFilter() throws Exception {
+        // https://github.com/questdb/questdb/issues/7201
+        // count()/size over a keyed GROUP BY subquery, filtered on the aggregate alias
+        // (a HAVING-style predicate), must keep the grouping keys. The top-down column
+        // pruning used to drop the unreferenced keys, collapsing the keyed group by into
+        // a scalar count() and producing phantom "duplicate" groups.
+        assertMemoryLeak(() -> {
+            execute("CREATE TABLE tab (ts TIMESTAMP, s SYMBOL INDEX) " +
+                    "TIMESTAMP(ts) PARTITION BY DAY WAL DEDUP UPSERT KEYS(ts, s)");
+            execute("INSERT INTO tab " +
+                    "SELECT '2020-02-27T12:00:00.000000Z'::TIMESTAMP, ('sym' || (x - 1)) " +
+                    "FROM long_sequence(12)");
+            drainWalQueue();
+
+            // the keyed group by must survive in the plan (not collapse into a bare Count)
+            assertPlanNoLeakCheck(
+                    "SELECT count() dups FROM (SELECT ts, s, count() c FROM tab) WHERE c > 1",
+                    """
+                            Count
+                                Filter filter: 1<c
+                                    Async Group By workers: 1
+                                      keys: [ts,s]
+                                      values: [count(*)]
+                                      filter: null
+                                        PageFrame
+                                            Row forward scan
+                                            Frame forward scan on: tab
+                            """
+            );
+
+            // all 12 (ts,s) pairs are unique, so no group has count > 1
+            assertSql("dups\n0\n", "SELECT count() dups FROM (SELECT ts, s, count() c FROM tab) WHERE c > 1");
+            // the materialized projection over the same subquery agrees: it is empty
+            assertSql("ts\ts\tc\n", "SELECT ts, s, c FROM (SELECT ts, s, count() c FROM tab) WHERE c > 1");
+            // without the predicate, count() over the subquery counts the 12 groups
+            assertSql("count\n12\n", "SELECT count() FROM (SELECT ts, s, count() c FROM tab)");
+            // a single-key subquery behaves the same way
+            assertSql("count\n0\n", "SELECT count() FROM (SELECT s, count() c FROM tab) WHERE c > 1");
+
+            // genuine duplicates are still reported: 3 symbols, each repeated 4 times
+            execute("CREATE TABLE dup (ts TIMESTAMP, s SYMBOL) TIMESTAMP(ts) PARTITION BY DAY");
+            execute("INSERT INTO dup " +
+                    "SELECT ('2020-02-27T12:00:00.000000Z'::TIMESTAMP + x), ('sym' || (x % 3)) " +
+                    "FROM long_sequence(12)");
+            assertSql("dups\n3\n", "SELECT count() dups FROM (SELECT s, count() c FROM dup) WHERE c > 1");
+            // a non-count outer aggregate over the filtered subquery is also correct: 3 groups of 4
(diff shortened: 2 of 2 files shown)


mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|9|9|100.00%|

ideoma:
# Code review (level 2: multi-agent, empirically verified)

## Verified context

- **Bug confirmed real on master**: the new test fails there at the plan assertion — the keyed group by collapses to a bare `Count` and the query returns 1 instead of 0. The test is a genuine regression guard.
- **Fix verified purely additive**, not just by reading the diff: `addTopDownColumn()` dedupes by alias (`QueryModel.java:335`), per-model bottom-up aliases are unique (`addBottomUpColumn` throws on duplicates), set-op branches are unaffected (`copyBottomToTopColumns()` pre-populates everything, so the retain pass is a no-op), top-level projections are unaffected (HAVING-style filters live on a wrapper model, so the GROUP BY is never the output-shaping top model), and positional consumers (`SELECT *`, `INSERT ... SELECT`, `UNION` by position) resolve by alias through projection layers — each of these was probed with concrete queries, not assumed.
- **Compile-time only**: `retainGroupByKeysAsTopDownColumns` runs once per query-model node per compilation, O(columns) with O(1) hash primitives (`isGroupBy` is a `LowerCaseCharSequenceHashSet` lookup), no allocation. No data-path exposure.
- Suites pass: CountTest, SqlOptimiserTest, GroupByTest, ExplainPlanTest (534 exact-plan assertions — the strongest guard against plan reordering), SampleByTest, UnionTest, DistinctTest, InsertAsSelectTest.

## Critical

None.

## Moderate

None outstanding. The six `assertSql` calls in the new test were converted to `assertQuery(...).noLeakCheck()...returns(...)` builder chains (per project convention for griffin tests), which adds factory-property coverage: the five scalar-aggregate shapes pin `noRandomAccess` + known size on the `Count`/`GroupByNotKeyed` factories, and the materialized projection pins `supportsRandomAccess=true` on the filter-over-group-by cursor (calibrated empirically — an initial `noRandomAccess` guess failed, so these assertions are not trivially passing). Fix-up commit to follow on this branch.

## Minor

1. **No plan assertion on the genuine-duplicate path** (`dup` table, `dups=3`). Results and factory properties are asserted, but a plan check there would prove the keyed group by also survives when groups actually contain duplicates, on a plain (non-WAL) table. The builder's `.withPlan(...)` could do both in one chain.
2. **The WAL + SYMBOL INDEX + DEDUP apparatus is incidental to the repro** — verified the bug reproduces identically on a plain `(ts TIMESTAMP, s SYMBOL) PARTITION BY DAY` table. Coverage is lopsided: multi-key only on the WAL/indexed table, plain table only single-key. One plain-table multi-key case (plan + result) would decouple the regression guard from WAL machinery. Non-symbol key types, compound predicates (`c > 1 AND c < 5`), and a NULL-key group are untested nice-to-haves.
3. **Comment triplication**: the same ~5-line rationale appears at both call sites (`SqlOptimiser.java:6218-6224`, `6259-6264`) and in the method comment (`7259-7263`) — ~18 comment lines for two one-line calls. Suggest keeping the full contract in the method doc (as Javadoc `/** */`, matching `resolveWindowJoinBoundColumns` and other neighbors) and trimming each call site to a one-liner stating only what is unique to that position.
4. **The early retain call is not demonstrably load-bearing**. Verified in isolation: with only the late call, every probed shape (plain 2/3-column tables, WAL/indexed table, single-key, multi-key, ORDER-BY-only) produced byte-identical plans and correct results, and the committed test passes; only removing the *late* call reproduces the collapse. The ORDER-BY-only shape is protected independently of either call (confirmed on master), so the WHERE/HAVING alias is the only contributor that genuinely needs the late timing. Keeping the early call for ordering stability is defensible and near-free (idempotent O(columns) hash probes at compile time) — but the "preserves key ordering for the common case" justification is currently unproven. If simplifying to a single late call, run ExplainPlanTest to confirm no exact-plan reorder.
5. **Near-duplicate inline loop left un-unified**: `SqlOptimiser.java:6282-6286` (UNION branch handling) contains the same retain-keys filter (`type != FUNCTION || !isGroupBy(token)`) inline. Not a drop-in replacement (parallel-indexes into `nestedBu`, targets `nested`), but same intent — worth a cross-referencing comment so a future fix to one doesn't miss the other.

## Downgraded (false positives dismissed during verification)

- *"The early pass is load-bearing on plain tables; removing it ships an untested bug"* — refuted empirically; removing the early call changes nothing on any table shape.
- *"The second pass could switch a top-level query's projection to top-down columns"* — refuted; HAVING filters are modeled on a wrapper, and key-only-in-WHERE top-level group bys get a `VirtualRecord` projection.
- *"Top-down order `[c, ts, s]` could break positional consumers"* — refuted; `SELECT *`, positional `INSERT`, and `UNION` probes all resolve by alias.
- *"An alias collision could make the dedup drop a key"* — refuted; per-model aliases are unique.
- *"Set-op branches could be affected"* — refuted; `copyBottomToTopColumns()` makes the pass a no-op there.
- *"Missing ORDER-BY-contributor regression test"* — refuted; that shape doesn't regress even on master, so there is no bug path to guard.

## Summary

**Approve.** The fix is correct and verified safe across set-ops, positional consumers, and top-level shapes; the regression test is genuine and now also asserts factory properties. Remaining items are minor polish (plan assertion on the duplicate path, comment trimming). Tradeoff worth noting: the belt-and-suspenders early retain call buys unproven ordering stability at the cost of a redundant but negligible compile-time scan.

13 draft findings → 7 verified (1 since resolved), 6 dropped as false positives after source/empirical verification; 6 in-diff, 1 out-of-diff-adjacent (pre-existing duplicate loop). Both changed symbols are private with all callsites inside `SqlOptimiser`, so the cross-context surface is the compilation entry point itself (`optimise()` → `propagateTopDownColumns`, `SqlOptimiser.java:11800`), covered by the suite runs and probes above.

ideoma:
# Code review (level 2: multi-agent, source-verified — second pass)

**Verified context:** the regression test fails without the fix (plan assertion: keyed `Async Group By` vs collapsed `Count`); no pre-existing test covers this shape. Correctness, concurrency, and resource agents came back clean with code-path evidence: all emissions into `model`'s own top-down list happen at or before line 6256, so the late call at `SqlOptimiser.java:6265` observes the model's final contributor state; `SqlOptimiser` is thread-confined via the exclusive-borrow `SqlCompilerPool`; the bottomUp/topDown `QueryColumn` aliasing is pool-safe and pre-existing. SAMPLE BY fill alignment (`SqlCodeGenerator.java:7805-7826`) is immune to retained keys since it only matches group-by functions. Suites pass: CountTest, SqlOptimiserTest (174), GroupByTest (112), SampleByTest (312), UnionTest (47).

## Critical

None.

## Moderate

**1. The late retention call redundantly re-runs the full loop when the early call already retained everything** — in-diff, `SqlOptimiser.java:6265`, corroborated independently by two performance agents.

In the common case (parent already contributed top-down columns), the early call at 6225 adds every non-aggregate key; the late call then re-walks all `bottomUpColumns` paying 2k hash probes (`isGroupBy` + `addTopDownColumn` dedup) per GROUP BY frame for zero effect. Verification proved a cheap gate is correct: the loop's output depends only on `bottomUpColumns`, which nothing between lines 6225 and 6265 mutates (LATEST BY/timestamp/WHERE/ORDER BY emissions all write `topDownColumns` only) — so if the early guard fired, the late call is a provable no-op:

```java
boolean retainedEarly = retainGroupByKeysAsTopDownColumns(model); // return whether guard fired
...
if (!retainedEarly) {
    retainGroupByKeysAsTopDownColumns(model);
}
```

This preserves the ordering the early call exists for (top-down order is load-bearing: it drives the page-frame column read order in `SqlCodeGenerator.generateTableQuery`, `SqlCodeGenerator.java:1577-1599`). Nuance, stated with equal weight: this is compile-time only, O(k) with small k, and the verification agent judged the unconditional idempotent double-call the simpler, safer design. Redundant work with a one-boolean fix is still a finding; author's call whether the branch is worth the state.

## Minor

2. **Comment triplication** — `SqlOptimiser.java:6218-6224`, `6259-6264`, `7259-7263`: ~18 comment lines for two one-line calls; "dedupes/idempotent" and "empty list would skip" each appear twice. Keep the full contract on the method (optionally as `/** */` to match `propagateTopDownColumns0`), trim each call site to the one line unique to its position.
3. **Naming** — `retainGroupByKeysAsTopDownColumns` *adds*; "retain" reads as remove-others. `addGroupByKeysToTopDownColumns` would be literal. Optional.
4. **No plan assertion on the genuine-duplicate path** — the `dup` table asserts results (`dups=3`, `sum=12`) but not that the keyed group by survives in the plan there.
5. **Cheap extra test shapes** — mixed predicate referencing both key and aggregate alias (`WHERE c > 1 AND s != 'x'`), non-SYMBOL key types, a NULL group key, an inner `sum(x)` instead of `count()` as the filtered aggregate, and the group-by subquery inside a UNION branch. All low-risk (the mechanism is contributor-agnostic and UnionTest passes), but each is one assertion.
6. **Cross-reference the near-duplicate loop** — `SqlOptimiser.java:6282-6286` (positional UNION key promotion) embeds the same `type != FUNCTION || !isGroupBy(token)` filter inline; a one-line comment pointing at the new method would keep future fixes in sync.
7. **Commit nits** — first commit title is 71 chars (limit 50); fixup commit `use assertQuery` has no body. Throwaway under squash-merge, noting for completeness.

## Downgraded (false positives dismissed during verification)

- *"ORDER BY-only contributor untested — HIGH gap"* — moot: `optimiseOrderBy` (`SqlOptimiser.java:11795`) runs before `propagateTopDownColumns` (11800) and clears a subquery's ORDER BY when the parent needs no ordering (`model.getOrderBy().clear()`, lines 5743-5747); an aggregate outer resolves to `ORDER_BY_INVARIANT`, so the ORDER BY literals never become top-down contributors. No live bug path; matches the earlier review's empirical result.
- *"WAL DEDUP + SYMBOL INDEX table is incidental complexity"* — issue #7201's repro DDL is exactly that table; the test mirrors the reporter's case verbatim. Deliberate, not incidental. (A plain-table variant would be an addition, folded into Minor 5.)
- *"GROUP BY model with LATEST BY as sole contributor newly retains keys"* — structurally impossible: LATEST BY attaches to the table-owning nested model (`SqlParser.java:3771/3794`), never the synthetic `SELECT_MODEL_GROUP_BY` projection layer (`SqlOptimiser.java:9308`) that the guard targets.
- *"`getFunctionFactoryCache()` re-fetched per column"* — trivial field getter, pattern unchanged from master and repeated at line 6284; not a regression.

## Summary

**Approve.** The fix is correct, purely additive, and verified safe across recursion order, set-ops, SAMPLE BY fill, union positional mapping, and pooling; the regression test is a genuine guard that fails on master. The one substantive tradeoff: the unconditional double pass buys ordering stability with redundant compile-time work that a one-boolean gate would eliminate (Moderate 1). ~15 draft findings: 7 verified, 4 dropped as false positives after source verification, rest were explicit clean-checks. All verified findings are in-diff — expected here, since both changed symbols are private to `SqlOptimiser` and the out-of-diff surface (codegen consumers of `topDownColumns`, union index mapping, SAMPLE BY) was walked and cleared rather than skipped.

bluestreak01 (approved):

Issue #1042, opened 2021-05-21
When insterting via ILP and there is an issue on database side the error is not propagated back to client


postol:
**Describe the bug**
When insterting via ILP and there is an issue on database side the error is not propagated back to client.

**To Reproduce**
1. Start inserting some invalid data, for example send field with value with space
2. QuestDB will start throwing errors that the format of ILP is wrong
3. Client continuous to insert data without knowing that the writes are actually failing

**Expected behavior**
Errors should be propagated back to client so the application that is inserting data can react.

**Environment (please complete the following information):**
 - version 6.0.1
 - 
**Additional context**
The same is happening for any error I believe as we saw the same behavior when the writes were failing on out of disk space

patrickSpaceSurfer:
ILP as a protocol, does not have any mechanism to propagate errors back to the client. If ILP is the ingestion choice then some other mecahnism needs to be used to monitor the status of questdb

postol:
ok, makes sense ... feel tree to close if not relevant

guywilsonjr:
where do you go to view the errors? the logs just say connected to[ip]. disconnected from [ip]

puzpuzpuz:
Note: 6.2.1 added `line.tcp.disconnect.on.error` config option to forcefully close ILP TCP connection on any error.

PR #7203, opened 2026-06-03
fix(core): fix posting index crash and missing rows after partition squash

nwoolmer:
tandem: https://github.com/questdb/questdb-enterprise/pull/1042

This fixes a set of correctness, resource and concurrency bugs in the POSTING /
covering index that surface across partition squash, the plain O3 commit path,
parquet partition reseal, and the WAL fast-lag apply path. It also corrects the
default covering behaviour of a bare `INDEX TYPE POSTING` declaration. Each
change is described below with its symptom and fix.

## 1. JVM crash and missing rows after partition squash

A table with a `POSTING` index can crash the JVM (covering index) or return
short/incorrect indexed counts (non-covering index) after its partitions are
squashed. The reported failure is a `SIGSEGV` in
`PostingIndexWriter.decodeDenseGenStride`, reached from a `wal-apply` thread:

```
squashSplitPartitions -> sealPostingIndexForPartition -> rebuildSidecars
  -> rebuildSidecarsByCopy -> writeSidecarForColumn -> decodeDenseGenStride  (SIGSEGV)
```

The squash and O3 reseal path runs `discardForRebuild()` -> `index()` ->
`commitDense()`. `commitDense()` assumed it writes the first and only generation
at `.pv` offset 0. When `index()` rebuilds a large or squashed partition that
holds a hot key, the `add()` loop trips the spill budget
(`cairo.posting.index.indexer.spill.bytes.max`, default 256 MiB) and
`compactIfOverBudget()` runs `flushAllPending()`, persisting one or more
**sparse** generations to the `.pv` mid-stream. `flushAllPendingDense()` then
appends the dense gen-0 at a non-zero file offset and `extendHead(newGenCount=1)`
overwrites gen-dir slot 0 to point at it, orphaning the sparse generations:

- Covering: `rebuildSidecars`' by-copy path decodes gen-0 from
  `valueMem.addressOf(0)` (the orphaned generation) while
  `accumulateDenseGen0Counts` reads the real file offset. The stride-index
  mismatch makes `decodeDenseGenStride` over-read native memory and crash.
- Non-covering: no sidecar rebuild, so no crash, but the orphaned rows are
  silently dropped and indexed predicates return short counts.

Fix: `commitDense()` consolidates through `seal()` when generations already
exist (`genCount > 0`). `seal()` drains the final pending batch and re-encodes
every generation into a single dense gen-0 at offset 0 -- the shape both the
by-copy rebuild and the dense fast path expect. `coverCount` is 0 at
`commitDense()` time (covering config is applied afterward), so `seal()` stays
sidecar-free and `rebuildSidecars()` writes the sidecars as before. The common
path (no spill, `genCount == 0`) still takes `flushAllPendingDense()` unchanged.

This also fires on the plain O3 commit path (an in-range O3 insert that mutates a
partition, no squash) and the WAL apply auto-squash path -- both reproduced
below.

## 2. Covered reads return NULL after an O3/squash reseal of a parquet partition

A parquet partition that carries a **covering** posting index returned NULL for
its covered columns after an O3 write rewrote it (e.g. `sum(price)` over the
covering index silently dropped the whole partition's rows). Two distinct causes:

- The O3 worker that rewrites a parquet partition builds only the non-covering
  `.pv`. `sealPostingIndexForPartition` previously skipped parquet partitions
  entirely, so the new parquet version had no `.pci`/`.pc` covering sidecars: a
  reader walked the chain (count correct) but found `coverCount == 0` and
  resolved covered values as NULL. `sealPostingIndexForPartition` now rebuilds
  the covering sidecars for parquet partitions from the rewritten parquet
  (`resealParquetCoveringForPartition` -> `indexParquetColumn`), pre-commit, in
  `finishO3Commit` / squash.
- A same-`sealTxn` chain-head republish (`extendHead`) overwrote the head
  entry's cover-end footer with an empty footer whenever the writer's live
  `coverCount` was transiently 0 (between `clearCovering` and the next
  `configureCovering`) or its sidecar handles were closed. A concurrent covered
  read of the republished entry then saw `sidecarFileEndOffsets == 0`, failed to
  map the existing `.pc`, and returned NULL for the whole partition.
  `PostingIndexChainWriter.readHeadCoverEndOffsets` snapshots the on-disk footer
  into `coverEndOffsetsCache` before the gen-dir overwrite, so `captureCoverEndOffsets`
  republishes the real extents instead of an empty footer.

## 3. WAL fast-lag covered sidecar incomplete after a mid-stream spill

On the WAL fast-lag apply path, a mid-stream spill flush during
`updateIndexesParallel` could write an incomplete `.pc` covered sidecar:
`publishPostingIndexesForLastPartitionFastLag`'s addr-based `configureCovering`
clears the writer-owned cover schema set at `openPartition`, so a spill flush
during indexing produced a sidecar missing its covered data, and a later covered
read over-read the `.pc`. `applyLagToLastPartition` now calls
`configureCoveringForLastPartitionFastLag()` to restore the covering
configuration on each posting writer **before** indexing, so spill-flushed
generations carry their covered data. The seal paths also forward the covered
columns' mapped byte lengths (`setCoveredColumnAddrSizes`), which bound the
addr-based covered reads under assertions.

## 4. Covering lost on a symbol-capacity change

`ALTER TABLE ... ALTER COLUMN ... SYMBOL CAPACITY` rebuilds the column metadata.
`updateColumnSymbolCapacity` did not carry over the covering-column indices, so
the next `rewriteAndSwapMetadata()` persisted a `_meta` with no covering
flag/section and every reader thereafter saw the posting index as non-covering.
`updateColumnSymbolCapacity` now preserves `getCoveringColumnIndices()`.

## 5. DDL: a bare `INDEX TYPE POSTING` is now non-covering (intentional, technically breaking)

A bare `INDEX TYPE POSTING` with no `INCLUDE` clause used to auto-append the
designated timestamp to the covering set, silently making the index covering.
That was an unintended default and should not have shipped in that state: a plain
posting index without an explicit `INCLUDE` is now **non-covering**. The
timestamp auto-include (`cairo.posting.index.auto.include.timestamp`, still
default true) now only rounds out an **explicit** `INCLUDE` set; it never turns a
non-covering index into a covering one on its own.

This is a technically breaking change to the default, surfaced consistently
across `CREATE TABLE`, CTAS, `ALTER TABLE ... ADD INDEX` and `SHOW CREATE TABLE`.
It is not expected to affect existing deployments: the covering flag is persisted
per-column in `_meta`, so existing tables created with a bare posting index keep
their covering flag on disk and do not change on read; only newly created or
re-declared indexes are affected. No known usage relies on the old implicit
covering. `SHOW CREATE TABLE tab` for `s SYMBOL INDEX TYPE POSTING` now renders
`INDEX TYPE POSTING` (no `INCLUDE (ts)`), and `SELECT * WHERE s = ...` takes the
plain posting-filter cursor (random access, lazy size) instead of the covering
cursor.

## 6. Resource: parquet reseal value-file leak

When the parquet index rebuild (`O3PartitionJob.updateParquetIndexes`) spills,
the post-fix `commitDense` seals and rotates the `.pv`, recording a purge for the
superseded file. The pooled parquet writer is freed immediately and its
`close()` discards the purge outbox without draining it through a `TableWriter`
(as the native reseal does), so the superseded intermediate `.pv` leaked
permanently on disk -- no directory sweep or recovery walk reclaims it.

`deferParquetPostingSealPurges` hands the rebuild's seal-purges to the
`TableWriter`'s deferred queue, published after the commit; the scoreboard-gated
`PostingSealPurgeJob` deletes the `.pv` only once no reader is pinned in its
`[from, to)` txn window. The hand-off runs in a `finally` around `commitDense`
(both the parquet-worker path and the writer-thread covering reseal), so an I/O
fault inside the seal's `sync` -- which can throw after the purge is recorded --
still hands the entry off instead of dropping it (idempotent no-op on an empty
outbox).

## 7. Concurrency hardening: a single lock for the seal-purge state

Parquet index rebuilds run on parallel O3 workers and stash seal-purges
concurrently, while the writer-thread paths touch the same `deferredPostingSealPurges`
list and task pool only after the O3 workers join. The writer side previously
held no lock and leaned on that join ordering alone. Every list/pool access --
writer and worker -- now takes `parquetSealPurgeLock` (a leaf lock, uncontended
on the post-join writer paths), so the non-thread-safe `ObjList`/`ObjectStackPool`
are safe by construction, not by timing. The post-join invariant itself is
asserted (`o3PartitionUpdRemaining.get() == 0`) on every seal/defer/publish path,
so a future change that moved a mutation into the O3 window can no longer corrupt
them silently. The per-commit publish path short-circuits on an empty list before
taking the lock.

## Performance

The `commitDense` change is a single branch gated on `genCount > 0`, so the
common path is untouched; only a reseal that already spilled past the budget
takes the new `seal()` route. The `seal()` route re-encodes the spilled
generations into a single dense gen-0 -- the irreducible cost of consolidating
them correctly; the old `flushAllPendingDense` was cheaper only because it was
wrong (orphaned the spilled gens).

`resealParquetCoveringForPartition` decodes a partition's parquet once and feeds
every covering posting column from the shared decoder, instead of re-opening and
re-mapping the whole parquet file per column. `configureCoveringForLastPartitionFastLag`
and the per-commit seal-purge publish short-circuit before any work for tables
with no posting / no covering posting index.

`PostingIndexBenchmarkSuite.walLargePartitionO3SpillReseal` (added here) measures
an O3 reseal of a ~1M-row covering-posting partition with the spill budget
parameterised. Indicative single-shot numbers (n=2, noisy):

| `spillBytesMax` | reseal path | ms/op |
| --- | --- | --- |
| 256 MiB | `flushAllPendingDense` (common) | ~129 |
| 2 MiB | `seal()` consolidation | ~143 |
| 256 KiB | `seal()` consolidation | ~139 |

- The 256 MiB row is the no-regression baseline: that path runs the same
  `flushAllPendingDense` with or without this change.
- The 2 MiB / 256 KiB rows cost roughly 10% more on this reseal, but they only
  exist because of the fix -- before it they crashed (covering) or dropped rows
  (non-covering), so there is no faster correct baseline.
- No standard benchmark sets a sub-256-MiB spill budget, so headline numbers are
  not expected to move. Run it with `-Dquestdb.suite.bench=wal_o3_spill`.

## Test plan

Squash path (the reported trigger), each verified to fail with the fix reverted:

- `testSquashCoveringPostingWithMidStreamSpillFlush` -- covering index,
  `squashPartitions()` -> `squashSplitPartitions` reseal; SIGSEGVs without the
  fix (confirmed via hs_err).
- `testSquashSplitPartitionsCoveringPostingMultiSplitSpill` -- merges several
  split sub-partitions in one `squashSplitPartitions` call (the reported
  overload), asserting correctness after the squash.
- `testSquashNonCoveringPostingWithMidStreamSpillFlush` -- non-covering index
  (auto-include-timestamp disabled); returns short counts without the fix,
  correct after.
- `testWalApplyAutoSquashCoveringPostingWithMidStreamSpillFlush` -- drives the
  crash through WAL apply housekeep auto-squash, the reported entry point.
- `testO3CommitCoveringPostingResealWithMidStreamSpillFlush` -- drives the reseal
  through the plain O3 commit path (`sealPostingIndexesForO3Partitions`, no
  squash) via an in-range O3 insert that mutates the partition.

Parquet covering reseal (cause 2 above):

- `testO3CoveringPostingParquetResealKeepsCoveredValues` -- deterministic: an O3
  merge into an already-parquet covering-posting partition; asserts exact covered
  `sum(price)` / `first(tag)` through the covering cursor (`expectSize()` +
  `noRandomAccess()` confirm the covering cursor, not a forward-scan fallback).
  Reads NULL covered values without the parquet-covering reseal.
- `testConvertToParquetPostingResealSpillDoesNotLeakValueFile` -- reproduces the
  parquet reseal `.pv` leak on disk and asserts a single value file remains.

WAL fast-lag, cover-footer preservation, symbol-capacity covering, and the
non-covering DDL default each have targeted regression tests; the concurrent
covered-read race and the lock/temporal-separation invariants are covered by
`PostingIndexO3ConcurrencyFuzzTest` (native + parquet, requires `-ea`).

Regression: `PostingIndexCriticalIssuesTest`, all covering-package tests,
`PostingIndexOracleTest`/`Ef`/`Delta`, `PostingFSSTIndexTest`,
`PostingIndexStressTest`/`Ef`/`Delta` and `O3SquashPartitionTest` all pass.

### DIFF
--- a/benchmarks/src/main/java/org/questdb/PostingIndexBenchmarkSuite.java
+++ b/benchmarks/src/main/java/org/questdb/PostingIndexBenchmarkSuite.java
@@ -107,6 +107,9 @@ public class PostingIndexBenchmarkSuite {
         //   "wal":             walFastLag* and walLargePartition*
         //   "wal_o3":          walLargePartitionO3* — O3 commit paths
         //                      (commitDense + rebuildSidecarsByCopy)
+        //   "wal_o3_spill":    walLargePartitionO3SpillReseal — reseal cost as
+        //                      the spill budget forces mid-stream flushes
+        //                      (commitDense's flushAllPendingDense vs seal())
         //   "io":              page-fault analysis only (skip JMH benchmarks)
         //   anything else:     literal regex body, expanded to
         //                      "PostingIndexBenchmarkSuite\.(<token>)$"
@@ -119,6 +122,7 @@ public class PostingIndexBenchmarkSuite {
             case "wal" -> suiteName + "\\.(walFastLag|walLargePartition).*";
             case "wal_o3" -> suiteName + "\\.walLargePartitionO3.*";
             case "wal_o3_append" -> suiteName + "\\.walLargePartitionO3AppendInsert.*";
+            case "wal_o3_spill" -> suiteName + "\\.walLargePartitionO3SpillReseal.*";
             case "io" -> null;
             default -> suiteName + "\\.(" + filter + ")$";
         };
@@ -128,7 +132,7 @@ public class PostingIndexBenchmarkSuite {
             org.openjdk.jmh.runner.options.ChainedOptionsBuilder builder = new OptionsBuilder()
                     .include(includePattern)
                     .resultFormat(ResultFormatType.TEXT);
-            if ("wal_o3".equals(filter) || "wal_o3_append".equals(filter)) {
+            if ("wal_o3".equals(filter) || "wal_o3_append".equals(filter) || "wal_o3_spill".equals(filter)) {
                 builder.forks(1)
                         .warmupIterations(1)
                         .measurementIterations(2);
@@ -484,6 +488,38 @@ public class PostingIndexBenchmarkSuite {
                 ", 0), '2024-01-01T00:00:00.000000Z'::TIMESTAMP), " +
                 "rnd_symbol(" + s.keyCount + ", 4, 8, 0), " + s.extraValues + " " +
                 "FROM long_sequence(" + WalLargePartitionO3State.BATCH_ROWS + ")";
+        s.engine.execute(sql, s.ctx);
+        s.applyJob.drain(0);
+        s.checkJob.run(0);
+        s.applyJob.drain(0);
+    }
+
+    @Benchmark
+    @BenchmarkMode(Mode.SingleShotTime)
+    @OutputTimeUnit(TimeUnit.MILLISECONDS)
+    @Fork(1)
+    @Warmup(iterations = 1)
+    @Measurement(iterations = 3)
+    public void walLargePartitionO3SpillReseal(WalLargePartitionO3SpillState s) throws Exception {
+        // Same O3-mutating insert as walLargePartitionO3Insert (forces
+        // sealPostingIndexForPartition's canSkipRebuild=false reseal:
+        // discardForRebuild + index + commitDense + rebuildSidecars), but with
+        // cairo.posting.index.indexer.spill.bytes.max parameterised so the
+        // reseal's index() loop spills a controlled number of times:
+        //   - spillBytesMax=256MiB: no mid-stream flush. commitDense stays on
+        //     flushAllPendingDense (the unchanged fast path) -- a no-op for the
+        //     fix, so this column is the no-regression baseline.
+        //   - spillBytesMax=2MiB / 256KiB: re-indexing the ~1M-row partition
+        //     trips compactIfOverBudget, so commitDense routes through seal()
+        //     (flushAllPending + sealFull consolidation). This is the path the
+        //     fix added; before the fix it crashed (covering) or dropped rows
+        //     (non-covering), so there is no pre-fix number to compare against.
+        // Single shot on a freshly preloaded partition (Level.Iteration setup).
+        String sql = "INSERT INTO walbench(ts, sym, " + s.extraColumns + ") " +
+                "SELECT dateadd('u', rnd_int(1, " + WalLargePartitionO3SpillState.PRELOAD_TS_LIMIT_US +
+                ", 0), '2024-01-01T00:00:00.000000Z'::TIMESTAMP), " +
+                "rnd_symbol(" + s.keyCount + ", 4, 8, 0), " + s.extraValues + " " +
+                "FROM long_sequence(" + WalLargePartitionO3SpillState.BATCH_ROWS + ")";
         s.engine.execute(sql, s.ctx);
         s.applyJob.drain(0);
         s.checkJob.run(0);
@@ -2248,6 +2284,105 @@ public class PostingIndexBenchmarkSuite {
             // The benchmark body then interleaves new rows across this
             // entire range, forcing O3 mutation on every commit.
             int batches = partitionSize / BATCH_ROWS;
+            for (int i = 0; i < batches; i++) {
+                int batchOffsetUs = i * BATCH_ROWS + 1;
+                String batchSql = "INSERT INTO walbench(ts, sym, " + extraColumns + ") " +
+                        "SELECT dateadd('u', x::INT + " + batchOffsetUs + ", '2024-01-01T00:00:00.000000Z'::TIMESTAMP), " +
+                        "rnd_symbol(" + keyCount + ", 4, 8, 0), " + extraValues + " " +
+                        "FROM long_sequence(" + BATCH_ROWS + ")";
+                engine.execute(batchSql, ctx);
+                applyJob.drain(0);
+                checkJob.run(0);
+                applyJob.drain(0);
+            }
+        }
+
+        @TearDown(Level.Iteration)
+        public void tearDown() {
+            Misc.free(applyJob);
+            Misc.free(compiler);
+            Misc.free(engine);
+            deleteDirRecursive(tmpDir.toFile());
+        }
+    }
+
+    /**
+     * State for {@code walLargePartitionO3SpillReseal}: preload a single ~1M-row
+     * DAY partition, then let the benchmark O3-mutate it so the posting index
+     * reseals from the column file. {@code keyCount} is small so each symbol is
+     * hot and re-indexing the partition spills enough rowids to cross the small
+     * {@link #spillBytesMax} budgets. {@code spillBytesMax} controls how often
+     * the reseal's index() loop flushes mid-stream, selecting commitDense's
+     * fast path (no flush) vs its seal() consolidation path (>=1 flush).
+     */
+    @State(Scope.Benchmark)
+    @BenchmarkMode(Mode.SingleShotTime)
+    @OutputTimeUnit(TimeUnit.MILLISECONDS)
+    public static class WalLargePartitionO3SpillState {
+        static final int BATCH_ROWS = 100_000;
+        static final int PARTITION_SIZE = 1_000_000;
+        // Matches the preload span (1us spacing over PARTITION_SIZE rows) so the
+        // benchmark's O3 rows land inside the existing data and force mutation
+        // (canSkipRebuild=false), not a pure append.
+        static final int PRELOAD_TS_LIMIT_US = PARTITION_SIZE;
+
+        ApplyWal2TableJob applyJob;
+        CheckWalTransactionsJob checkJob;
+        SqlCompilerImpl compiler;
+        SqlExecutionContextImpl ctx;
+        CairoEngine engine;
+        String extraColumns;
+        String extraValues;
+        @Param({"posting_covering", "posting"})
+        String indexType;
+        int keyCount = 100;
+        // cairo.posting.index.indexer.spill.bytes.max. 256MiB never flushes
+        // mid-stream (flushAllPendingDense fast path = no-regression baseline);
+        // 2MiB and 256KiB trip compactIfOverBudget so commitDense routes through
+        // seal() (the path the fix added).
+        @Param({"268435456", "2097152", "262144"})
+        long spillBytesMax;
+        java.nio.file.Path tmpDir;
+
+        @Setup(Level.Iteration)
+        public void setup() throws Exception {
+            tmpDir = Files.createTempDirectory("suite-largepart-o3-spill");
+            CairoConfiguration config = new DefaultCairoConfiguration(tmpDir.toString()) {
+                @Override
+                public long getPostingIndexerSpillBytesMax() {
+                    return spillBytesMax;
+                }
+
+                @Override
+                public byte getPostingIndexRowIdEncoding() {
+                    return IS_DELTA ? PostingIndexUtils.ENCODING_DELTA : PostingIndexUtils.ENCODING_ADAPTIVE;
+                }
+
+                @Override
+                public int getRndFunctionMemoryMaxPages() {
+                    return 4096;
+                }
+            };
+            engine = new CairoEngine(config);
+            ctx = new SqlExecutionContextImpl(engine, 1)
+                    .with(config.getFactoryProvider().getSecurityContextFactory().getRootContext(),
+                            null, null, -1, null);
+            compiler = new SqlCompilerImpl(engine);
+
+            String ddl = switch (indexType) {
+                case "posting" -> "CREATE TABLE walbench (ts TIMESTAMP, sym SYMBOL INDEX TYPE " + POSTING_SQL +
+                        ", price DOUBLE, name VARCHAR) TIMESTAMP(ts) PARTITION BY DAY WAL";
+                case "posting_covering" -> "CREATE TABLE walbench (ts TIMESTAMP, sym SYMBOL INDEX TYPE " + POSTING_SQL +
+                        " INCLUDE (price), price DOUBLE, name VARCHAR) TIMESTAMP(ts) PARTITION BY DAY WAL";
+                default -> throw new IllegalArgumentException(indexType);
+            };
+            extraColumns = "price, name";
+            extraValues = "rnd_double() * 1000, rnd_varchar(10, 20, 0)";
+            engine.execute(ddl, ctx);
+            applyJob = new ApplyWal2TableJob(engine, 0);
+            checkJob = new CheckWalTransactionsJob(engine);
+
+            int batches = PARTITION_SIZE / BATCH_ROWS;
             for (int i = 0; i < batches; i++) {
                 int batchOffsetUs = i * BATCH_ROWS + 1;
                 String batchSql = "INSERT INTO walbench(ts, sym, " + extraColumns + ") " +
--- a/benchmarks/src/main/java/org/questdb/PostingIndexBenchmarkSuite.java
+++ b/benchmarks/src/main/java/org/questdb/PostingIndexBenchmarkSuite.java
@@ -499,7 +499,7 @@ public class PostingIndexBenchmarkSuite {
     @OutputTimeUnit(TimeUnit.MILLISECONDS)
     @Fork(1)
     @Warmup(iterations = 1)
-    @Measurement(iterations = 3)
+    @Measurement(iterations = 2)
     public void walLargePartitionO3SpillReseal(WalLargePartitionO3SpillState s) throws Exception {
         // Same O3-mutating insert as walLargePartitionO3Insert (forces
         // sealPostingIndexForPartition's canSkipRebuild=false reseal:
@@ -2361,6 +2361,17 @@ public class PostingIndexBenchmarkSuite {
                 @Override
                 public int getRndFunctionMemoryMaxPages() {
                     return 4096;
+                }
+
+                @Override
+                public boolean isPostingIndexAutoIncludeTimestamp() {
+                    // Without this, POSTING auto-covers the designated timestamp,
+                    // so the "posting" variant would also be covering and the two
+                    // @Param values would measure the same path. Disabling it keeps
+                    // "posting" on the non-covering reseal (commitDense -> seal,
+                    // no rebuildSidecars) and "posting_covering" on the covering
+                    // reseal via the explicit INCLUDE (price) below.
+                    return false;
                 }
             };
             engine = new CairoEngine(config);
--- a/core/src/main/java/io/questdb/MessageBusImpl.java
+++ b/core/src/main/java/io/questdb/MessageBusImpl.java
@@ -265,6 +265,8 @@ public class MessageBusImpl implements MessageBus {
     @TestOnly
     public void clear() {
         columnPurgeSubSeq.clear();
+        postingSealPurgeSubSeq.clear();
+        groupByLongTopKSubSeq.clear();
         groupByMergeShardSubSeq.clear();
         indexerSubSeq.clear();
         latestBySubSeq.clear();
@@ -279,6 +281,7 @@ public class MessageBusImpl implements MessageBus {
         copyExportRequestSubSeq.clear();
         vectorAggregateSubSeq.clear();
         walTxnNotificationSubSequence.clear();
+        queryCacheEventSubSeq.clear();
         unorderedPageFrameReduceSubSeq.clear();
         for (int i = 0, n = pageFrameReduceSubSeq.length; i < n; i++) {
             pageFrameReduceSubSeq[i].clear();
--- a/core/src/main/java/io/questdb/cairo/ColumnIndexer.java
+++ b/core/src/main/java/io/questdb/cairo/ColumnIndexer.java
@@ -142,6 +142,9 @@ public interface ColumnIndexer extends QuietCloseable {
     default void seal() {
     }
 
+    default void setCoveredColumnAddrSizes(LongList dataSizes, LongList auxSizes) {
+    }
+
     default void setCoveredColumnNameTxns(LongList txns) {
     }
 
--- a/core/src/main/java/io/questdb/cairo/O3PartitionJob.java
+++ b/core/src/main/java/io/questdb/cairo/O3PartitionJob.java
@@ -3396,10 +3396,21 @@ public class O3PartitionJob extends AbstractQueueConsumerJob<O3PartitionTask> {
                             // rotating the .pv and recording a purge for the
                             // superseded file. The pooled writer is freed below
                             // without draining that outbox through a TableWriter,
-                            // so reclaim the superseded value file inline. Safe
-                            // here: this is a fresh, uncommitted parquet index
-                            // whose old gens no committed reader can pick.
-                            indexWriter.unlinkPendingSealPurgesDirect();
+                            // so the superseded value file would leak. Reclaim it
+                            // inline -- but ONLY in rewrite mode, where the index
+                            // is built into a fresh txn-named directory that is
+                            // invisible to readers until the commit bumps _txn
+                            // (the old dir stays at srcNameTxn): no committed
+                            // reader can pick the superseded prev gen, whose entry
+                            // is tagged with the pre-seal txn. In update-in-place
+                            // mode the directory is the live committed one, where
+                            // a reader pinned at the pre-commit txn could still
+                            // walk to and map the superseded .pv, so it is not
+                            // safe to unlink here (that far-rarer case still leaks
+                            // the intermediate, like the pre-fix behaviour).
+                            if (isRewrite) {
+                                indexWriter.unlinkPendingSealPurgesDirect();
+                            }
                         } else {
                             indexWriter.commit();
                         }
--- a/core/src/main/java/io/questdb/cairo/O3PartitionJob.java
+++ b/core/src/main/java/io/questdb/cairo/O3PartitionJob.java
@@ -3405,9 +3405,15 @@ public class O3PartitionJob extends AbstractQueueConsumerJob<O3PartitionTask> {
                             // is tagged with the pre-seal txn. In update-in-place
                             // mode the directory is the live committed one, where
                             // a reader pinned at the pre-commit txn could still
-                            // walk to and map the superseded .pv, so it is not
-                            // safe to unlink here (that far-rarer case still leaks
-                            // the intermediate, like the pre-fix behaviour).
+                            // walk to and map the superseded .pv, so it is NOT
+                            // safe to unlink here. That far-rarer case keeps one
+                            // superseded .pv on disk -- bounded, not growing: the
+                            // next in-place rebuild's of(isInit) truncates and
+                            // reuses it, so the on-disk count stays at two
+                            // (.pv.0 + active). That is a known gap (full
+                            // reclamation needs a scoreboard-gated deferred purge
+                            // routed to the writer thread), still strictly better
+                            // than the pre-fix data loss it replaces.
                             if (isRewrite) {
                                 indexWriter.unlinkPendingSealPurgesDirect();
                             }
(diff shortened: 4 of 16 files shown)


nwoolmer:
## Self-review (multi-agent, level 3) + current state

I ran a thorough adversarial review of this branch (correctness, cross-context callers, resource/purge accounting, concurrency, performance, tests). Two reviewers independently reproduced the bug by reverting the fix (covering test SIGSEGVs; non-covering test returns `count 1` instead of `401`). Posting this for transparency since I authored it.

### Verdict: ready, with the tradeoff noted below.

### Root cause / fix recap
`commitDense()` assumed it writes the first and only generation at `.pv` offset 0. On a posting-index reseal over a large/squashed partition with a hot key, the rebuild's `index()` loop trips the spill budget (`cairo.posting.index.indexer.spill.bytes.max`, default 256 MiB), so `compactIfOverBudget()` persists sparse generations mid-stream. The old `flushAllPendingDense()` then appended the dense gen-0 at a non-zero offset and `extendHead(newGenCount=1)` orphaned those sparse gens — a SIGSEGV in `decodeDenseGenStride` for covering indexes, silent short counts for non-covering. The fix routes `commitDense()` through `seal()` when `genCount > 0`, consolidating every gen into a single dense gen-0 at offset 0. The common path (`genCount == 0`) is unchanged.

### Review findings (all addressed)
- **Comment overclaim** — the fix comment said "all sparse here, so never the incremental path"; an inline `MAX_GEN_COUNT` seal during `index()` can densify gen-0, so `seal()` may take the incremental re-encode. Outcome is still correct (both land gen-0 at offset 0). Reworded.
- **Benchmark variant mislabeled** — `isPostingIndexAutoIncludeTimestamp()` defaults true, so the benchmark's `posting` param was silently covering. Disabled it in the state so the two params now measure distinct paths (covering ~269 ms vs non-covering ~223 ms on a ~1M-row reseal at the 256 KiB budget).
- **Coverage gap** — the plain O3-commit reseal (no squash, the most common production trigger) had no assertion-backed test. Added `testO3CommitCoveringPostingResealWithMidStreamSpillFlush` (verified it SIGSEGVs without the fix).

### Cleared during review (no change needed)
Double-seal `.pv` purge accounting (each superseded file purge-recorded once, empty `[T+1,T+1)` window); concurrency (`appendNewEntry` is no less safe than the old in-place `extendHead`; new entries are `txnAtSeal=getTxn()+1`, invisible to T-pinned readers until commit); the `seal()` early-return can't fire on this path (sparse gen-0); all three `commitDense` callers (covering reseal, non-covering reseal, parquet O3 build) verified safe.

### Current state
Commits:
- `c819d8a` Fix posting reseal gen loss on mid-stream spill
- `4a23a45` Add posting reseal spill benchmark and squash test
- `42587d1` Refine commitDense comment and benchmark variants
- `b051098` Add O3 reseal test; hardcode spill assertions

Tests (each verified to fail with the fix reverted; all assert hardcoded expected aggregates):
- `testSquashCoveringPostingWithMidStreamSpillFlush` — covering, `squashPartitions()`
- `testSquashSplitPartitionsCoveringPostingMultiSplitSpill` — covering, multi-split squash
- `testSquashNonCoveringPostingWithMidStreamSpillFlush` — non-covering (data loss, not crash)
- `testWalApplyAutoSquashCoveringPostingWithMidStreamSpillFlush` — WAL housekeep auto-squash
- `testO3CommitCoveringPostingResealWithMidStreamSpillFlush` — plain O3 reseal, no squash

Regression: `PostingIndexCriticalIssuesTest` (65), all covering-package tests (1177), posting oracle/FSST/native/oom (65), posting stress + `O3SquashPartitionTest` (180) all pass.

Benchmark: `PostingIndexBenchmarkSuite.walLargePartitionO3SpillReseal` (`-Dquestdb.suite.bench=wal_o3_spill`) parameterises the spill budget so the 256 MiB column is the no-regression baseline (unchanged `flushAllPendingDense`) and the small budgets exercise the new `seal()` consolidation.

### Remaining tradeoff (not addressed — pre-existing, out of scope)
On the covering `genCount > 0` path, `sealFull` writes a clean dense gen-0 and then `rebuildSidecarsByCopy` immediately `copyData`s that whole `.pv` again into another file before attaching sidecars — a redundant full-file pass. It predates this PR (the by-copy already did this) and only fires on a >256 MiB mid-stream spill, where the alternative was a crash. A follow-up could let `rebuildSidecars` consume `sealFull`'s output in place rather than re-copying it.

bluestreak01:
Independent corroboration from a production crash.

I root-caused a production `hs_err` SIGSEGV separately and it lands on exactly this bug and call path, so this isn't only a synthetic/fuzzer find -- it reproduces in the field.

**Faulting frame and path**
- `J <n> jvmci io.questdb.std.Unsafe.getByte(J)B`, on a `wal-apply` thread, in an `-ea` build.
- `ensureCoveredColumnReadMaps` deopts immediately before the fault, the covering sidecar files (`.pc0`/`.pc1`/`.pci`) are mapped rw for writing, and partition-split directories are present -- i.e. the squash/O3 covering reseal, decoding at `writeSidecarForColumn -> decodeDenseGenStride(valueMem.addressOf(0), keyCount, ...)`.

**The address arithmetic matches the "orphaned sparse gen at offset 0" mechanism**
- `.pv` base `B` (a sub-GiB file); faulting address `= B + 0x2_0000_0031`, i.e. `genBase + siSize + strideOff` with `strideOff ~= 0x2_0000_0031` (~8.6 GiB) -- a read ~8 GiB past the value file, hence `SEGV_MAPERR`.
- `0x2_0000_0031` is two little-endian int32 halves, `0x31` (49) and `0x02`. That is sparse-gen bytes read through `getLong` as if they were a dense stride-offset-table entry -- exactly what happens when `addressOf(0)` lands on the orphaned sparse generation instead of the dense gen-0. `decodeDenseGenStride`'s only guard (`nextStrideOff == strideOff`) doesn't fire because adjacent sparse bytes differ.

Routing `commitDense()` through `seal()` when `genCount > 0` is the right fix, and `testWalApplyAutoSquashCoveringPostingWithMidStreamSpillFlush` covers the `wal-apply -> squashSplitPartitions -> sealPostingIndexForPartition` path the crash actually took.

**One non-blocking suggestion**
`decodeDenseGenStride` / `decodeDenseGenSingleKey` dereference `strideOff` with no range check, so a future layout mismatch degrades straight to a JVM-killing SIGSEGV rather than a catchable `AssertionError` (the test comment notes "no AssertionError"). A cheap bounds check on `strideOff` against the gen's value-region size at those two sites would make the next such bug fail cleanly under `-ea`. Worth considering here or as a follow-up.

**CI**: GitHub `build` is green; the Azure self-hosted suites are still pending and `enterprise-ci` is currently red -- worth confirming that red is pre-existing/infra and not from this change before merge.

bluestreak01:
## Code review — tandem with enterprise #1042 (level 2)

Independent multi-agent adversarial pass (correctness, concurrency, performance, resources, tests). Two reviewers reproduced the bug by reverting `commitDense`: the covering test SIGSEGVs and the non-covering test returns a short count; the fix restores full results. Confirming the analysis below.

### Verdict: approve

**Correctness — confirmed.** Routing `commitDense` through `seal()` when `genCount > 0` is the minimal correct fix. After a mid-`index()` spill, gen-0 is always sparse (`flushAllPending` publishes `overrideKeyCount = -activeKeyCount`), so `seal()`'s `isIncrementalCandidate` is false and it falls to `sealFull()`, re-encoding every gen into one dense gen-0 at offset 0 — exactly what `rebuildSidecarsByCopy` (`addressOf(0)`) and the fast path expect. The `genCount==1 && gen0KeyCount>=0` early-return cannot fire with a dense gen-0 at a non-zero offset (the only producer of that layout is `flushAllPendingDense`, which is terminal), so it never short-circuits the consolidation onto a broken layout.

**Concurrency / double-seal — confirmed safe.** On the covering path the fix introduces an extra intermediate `.pv` (commitDense's `seal()` → `.pv.B`, then `rebuildSidecars` → `.pv.C`). Both chain entries carry the identical `txnAtSeal = txWriter.getTxn()+1`, so `PostingIndexChainPicker.pick()` (head-first walk) permanently shadows `B` with `C`; no pinned `_txn` can resolve to `B`. `recordPostingSealPurge(B)` therefore gets an empty `[X,X)` window, the purge is deferred until after `txWriter.commit`, and `isRangeAvailable(X,X)` is vacuously true — the intermediate file is never visible and never unlinked under a live reader.

**Performance — no regression.** The common `genCount==0` path is byte-for-byte unchanged (one added `int>0` branch). `sealFull` is single-pass `O(total_values)` — the optimal lower bound for consolidating sparse gens, and the only correct option since the live rows sit in already-drained sparse gens. The benchmark's 256 MiB row confirms the no-spill baseline is untouched.

**Resources — no new leak.** With `coverCount==0` at `commitDense`, `seal()`'s sidecar-snapshot allocations are skipped; `sealFull`'s native buffers are all freed in `finally`, and `valueMem.swapState` closes the old mapping exactly once.

### Findings

**Moderate — test gap (parquet build).** The `commitDense` `genCount>0 → seal()` branch reached from `O3PartitionJob` (the parquet O3 build) is not exercised with a forced spill — `testParquetIndexWriteUsesCommitDense` sets no spill budget and asserts the `flushAllPendingDense()` (`genCount==1`) branch. *Tempered:* `O3PartitionJob` is unchanged by this PR and `seal()` was the historical parquet behavior, but the non-covering dropped-rows scenario on a parquet build is unverified. Suggest: convert a partition to parquet over a hot key (≥400 rows of one symbol), set `CAIRO_POSTING_INDEX_INDEXER_SPILL_BYTES_MAX=256`, O3-insert into it, and assert full `count()/sum()`.

**Low — test hardening.**
- The 5 spill tests don't pin the query plan (`assertPlan`). `WHERE sym='A'` is assumed index-served; a future optimizer fallback to a base-table scan would silently mask a re-introduced bug (sym.d stays intact; only the index is corrupted). One `assertPlan` on a covering/posting scan node would harden them.
- The `MAX_GEN_COUNT` inline-seal-mid-`index()` edge (the early-return at `seal()`) and multi-hot-key / exact-budget-boundary spills are untested degenerate cases.

**Minor — defensive.** The double-seal's safety rests on the two `setNextTxnAtSeal(getTxn()+1)` calls (`TableWriter` covering + non-covering reseal) staying identical. A one-line assertion or test guarding `B.txnAtSeal == C.txnAtSeal` would lock the invariant against a future divergence.

**Label nit:** consider adding `WAL` — the reported crash is on a `wal-apply` thread and one test drives it through WAL-apply auto-squash. `Performance` is correctly omitted (the PR states headline numbers don't move).

<sub>🤖 Multi-agent review via Claude Code (level 2, tandem). Findings verified against source; treat as input, not gospel.</sub>

bluestreak01 (dismissed):

nwoolmer:
## Follow-up: parquet reseal `.pv` leak (found in review, now fixed)

A deeper pass on the third `commitDense` caller — the parquet index rebuild (`O3PartitionJob.updateParquetIndexes`) — found that the `commitDense → seal()` change introduces a permanent on-disk `.pv` leak there, which my earlier review missed (it only traced the two `TableWriter` reseal callers, which drain correctly via `deferPendingPostingSealPurges`).

**Mechanism (verified against source + demonstrated on disk):** when the parquet rebuild's `index()` loop spills, `commitDense` seals, rotating the `.pv` and recording a purge for the superseded file. The pooled parquet writer is then freed, and `close() → releasePendingPurges()` discards the outbox without ever publishing it (no `TableWriter` drain on this path). Nothing reclaims the orphan — `PostingSealPurgeOperator.onValueFile` is a no-op (no directory sweep) and the recovery walk is abandonment-only. A reproduction (`CONVERT TO PARQUET` + O3 rewrite + spill budget 256) leaves two `sym.pv.*` files in the partition dir surviving the purge job:

```
[sym.pv.1, sym.pk, _pm, sym.pv.0, data.parquet]   # sym.pv.0 leaked
```

**Fix (`34fcfa9`):** `IndexWriter.unlinkPendingSealPurgesDirect()` (overridden by `PostingIndexWriter`) unlinks the recorded value files inline. The direct unlink — bypassing the scoreboard-gated queue the native path needs — is safe here because the parquet rebuild is a *fresh, uncommitted* index: the chain head is tagged with the upcoming txn so no committed reader can pick the superseded gens, and a rewrite failure removes the whole partition directory. The reproduction test now asserts exactly one value file remains, and the index returns correct results (`count = 2012`).

The originally proposed one-line `publishPendingPurges(...)` fix would **not** have worked — I confirmed the purge window is `toTxn = getTxn()+1`, which `publishPendingPurges` retains (`toTableTxn > currentTableTxn → continue`), so `close()` would still drop it. Deferring past the conversion commit (the native mechanism) is also unsafe to drive from `updateParquetIndexes`, which runs on parallel O3 workers; the inline unlink sidesteps both.

Regression: full `PostingIndexCriticalIssuesTest` (66), `CoveringIndexTest` (379), `AlterTableConvertPartitionTest` (38), `BitmapIndexTest` (42), O3-parquet merge/stale-reader/fuzz, posting stress/oracle — all green.

nwoolmer:
## Re-review caught a hole in my own leak fix (now closed)

Reviewing the inline value-file reclaim adversarially against the chain picker, I found it was **not unconditionally safe** — exactly the "don't introduce a worse bug" risk, since the reclaim *deletes* a file.

**The hole:** `PostingIndexChainPicker.pick` walks from the head and returns the **first** entry with `txnAtSeal <= pinnedTableTxn` (`PostingIndexChainPicker.java:74`). The mid-stream spill gens that produced the superseded `.pv` are tagged `txnAtSeal=0` (they run before `setNextTxnAtSeal`), so a reader pinned at the **pre-commit** table txn would skip the `SEAL@getTxn()+1` head and **select the `prev` entry, mapping the superseded `.pv`**. My "no reader can pick it" claim only holds if that reader can't resolve the rebuild's directory at all.

**Where it holds vs. breaks:**
- `isRewrite=true` (rewrite mode): the index is built into a **fresh `txn`-named directory**; the old dir stays at `srcNameTxn` until the commit bumps `_txn` (`O3PartitionJob.java:240-243`). So no pre-commit reader opens the new dir, and post-commit readers pin at the seal txn and pick the head. **Safe.**
- `isRewrite=false` (update-in-place): `txnName = srcNameTxn`, the live committed dir — a reader pinned at the pre-commit txn could map the superseded `.pv`. **Unsafe to unlink.**

**Resolution (`0165b2c`):** gate the inline unlink on `isRewrite`. The rarer in-place case keeps leaking the intermediate (the pre-fix behaviour) rather than risking a reader fault. The existing `O3ParquetStaleReaderTest` confirms the in-place merge is the one exercised with concurrent readers, so leaving it untouched is the conservative choice.

**Fuzz coverage (`testParquetPostingSpillConcurrentReadFuzz`):** 3 background threads continuously `count() WHERE sym='A'` on the parquet partition while the main thread O3-rewrites it under a 256-byte spill budget across random batches. A reclaim that unlinked a live `.pv` would surface as a concurrent query failing to open it; the count must stay monotonic and exact with no reader crash. Stable across 5 runs with random seeds. The in-place-merge fuzz (`O3ParquetMergeStrategyFuzzTest`) and full posting suite remain green.

Net: the picker walk is the kind of thing static reasoning alone can miss — thanks for pushing on the re-review.

nwoolmer:
## Latest changes: parquet covering reseal + auto-include correction (`b04725729a`)

Two correctness fixes for the covering posting index, both surfaced while auditing the index against O3 jobs and the table writer.

### 1. Covering reads return NULL after O3 rewrites a parquet partition

`sealPostingIndexForPartition` skipped parquet partitions outright. When an O3 (or squash) commit rewrote a covering partition to a new `sealTxn`-versioned `.pv`, the writer never materialized that version's `.pci`/`.pc` covering sidecars. A concurrent reader that opened the rewritten version found the `.pci` absent, set `coverCount = 0`, and returned NaN/NULL for every covered column — e.g. `sum(price)` silently dropped whole partitions while `count()` stayed correct (`sum` skips NaN, `count` does not).

**Fix:** `sealPostingIndexForPartition` now calls `resealParquetCoveringForPartition` for parquet partitions, which re-runs `indexParquetPartition` per covering column (rebuilding `.pv` + `.pci`/`.pc` for the new version) and defers the superseded files for scoreboard-gated purge. `indexParquetPartition` also calls `discardForRebuild` after `setNextTxnAtSeal`, so a re-index supersedes the old value file instead of double-counting rows. Non-covering columns are skipped, preserving the `commitDense` fast path.

### 2. Behavior correction: a bare `INDEX TYPE POSTING` (no INCLUDE) is no longer covering

The designated timestamp was being auto-appended to the covering set of **every** posting index, so a plain `INDEX TYPE POSTING` with no `INCLUDE` clause silently became a covering index. That is wrong: with no `INCLUDE` the user asked for a plain posting index, and making it covering rotates value files at seal time (tripping the `commitDense` path) and changes query plans and cursor shapes.

The timestamp auto-include now only rounds out an **explicit** `INCLUDE` set: a posting index is covering iff the user gave `INCLUDE` with at least one column, and only then is the designated timestamp appended. Gated at all three resolution sites — ALTER ADD INDEX (`SqlCompilerImpl.alterTableColumnAddIndex`), inline CREATE TABLE and CTAS (`CreateTableOperationImpl.resolveCoveringColumnIndices` / `resolveCoveringFromAugmented`) — plus the matching `hasCovering` meta flag. This reverts the production logic of an earlier "auto-include on bare POSTING" change while keeping the auto-include benefit for indexes that do list `INCLUDE` columns.

### Tests

- Multi-threaded concurrent-reader regression for the parquet covered-read race (un-ignored, green).
- Tests that had baked in the bare-POSTING-is-covering behavior now assert the non-covering result; tests that genuinely exercise covering machinery were given an explicit `INCLUDE` so they keep their coverage.
- Added `testAutoIncludeTimestampCoversWhereFilter` and `testAutoIncludeTimestampCoversLatestOn` to lock in that the auto-included timestamp alone keeps the WHERE-filter and `LATEST ON` special paths covering.
- Green: `CoveringIndexTest` (×3 encodings), `CreateTableTest`, `PostingIndexCriticalIssuesTest`, and the enterprise replication/backup posting tests.

nwoolmer:
## Code review — level 2 (re-review after fixes)

Re-reviewed the branch at level 2 after pushing the review follow-up fixes (`6df71a38d7`, `ad722c8a4d`). A focused 5-agent pass (correctness, concurrency, resource management, fresh-context adversarial, adversarial performance) on the new delta, since the prior level-2 pass already cleared the base.

**Verdict: no Critical findings, no regressions.**

### What the follow-up fixes addressed
- **Resource:** `deferParquetPostingSealPurges` / `deferPendingPostingSealPurges` now drain in a `finally` around the rebuild's `commitDense`/`seal`, so an I/O fault that throws *after* the seal records its purge still hands the superseded `.pv`/`.pc` to the deferred queue instead of leaking the value file (both the O3-worker parquet path and the writer-thread covering reseal).
- **Performance:** `resealParquetCoveringForPartition` opens one parquet decoder per partition and feeds every covering posting column from the shared mmap, instead of re-opening/re-mapping the file per column (K opens -> 1). It also bails before the `stat(2)` when no covering posting index exists, and the per-commit seal-purge publish skips the lock on an empty queue.
- **Correctness:** the parquet covering reseal passes `allowDestructiveRecovery = false` to `createIndexFiles`, so a live committed `.pk` is never destructively reset; only a fresh `ADD INDEX` build passes `true`.
- **Test:** added `testO3CoveringPostingParquetResealKeepsCoveredValues`, a deterministic regression test asserting exact covered values through the covering cursor after an O3 merge rewrites an already-parquet covering-posting partition.

### Re-review confidence
- **Concurrency:** the new unlocked `size() == 0` fast-path on `publishDeferredPostingSealPurges` is safe. Every caller runs post-join (`o3PartitionUpdRemaining == 0`), and the worker->writer join barrier (`SOUnboundedCountDownLatch`, `volatile count`) supplies the happens-before edge, so a stale-zero read (which would leak a purge) cannot occur on any JMM-conformant platform.
- **Resource:** every parquet teardown path is leak-free and idempotent (`ParquetPartitionDecoder.destroy()` / `ParquetMetaFileReader.clear()` guard their native pointers); the shared-decoder `finally` is byte-identical to the existing `indexParquetPartition` teardown, and the metadata mapping is released even if `mapRO` throws after the metadata open.
- **Correctness:** the extracted `indexParquetColumn` is behaviorally identical to the old inline body apart from the threaded `allowDestructiveRecovery`; the `hasCoveringPostingIndex` / `hasPostingIndexers` guards are strict supersets of their loop bodies, so they never skip needed work.

### Minor fixes applied during re-review (`ad722c8a4d`)
- Restored the per-reseal `LOG.info` (the extraction had silently dropped the reseal-path log line).
- Corrected a comment that named the wrong early-return guard on the non-destructive `createIndexFiles` path.

### Validation
`PostingIndexCriticalIssuesTest` 86/86, `CoveringIndexTest` 1143/1143, and `PostingIndexO3ConcurrencyFuzzTest` (native + parquet) all pass.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 249 / 267 (93.26%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/ColumnIndexer.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/idx/IndexWriter.java|1|2|50.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|44|48|91.67%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexChainWriter.java|11|12|91.67%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|180|191|94.24%|
|:large_blue_circle:|io/questdb/cairo/SymbolColumnIndexer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriterMetadata.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CreateTableOperationImpl.java|4|4|100.00%|

bluestreak01 (dismissed):

nwoolmer:
Flaky test is fixed here: https://github.com/questdb/questdb/pull/7207

bluestreak01 (approved):

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=239879)
**Tested Commit:** `8a9a3b775b20e3d65d4582dfbb2c476c5782a3e1`

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 254 / 274 (92.70%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/ColumnIndexer.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/idx/IndexWriter.java|1|2|50.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|4|6|66.67%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|44|48|91.67%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexChainWriter.java|11|12|91.67%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|180|191|94.24%|
|:large_blue_circle:|io/questdb/cairo/SymbolColumnIndexer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriterMetadata.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/MessageBusImpl.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CreateTableOperationImpl.java|4|4|100.00%|

PR #6759, opened 2026-02-09
perf(parquet): improve decoding performance for parquet

RaphDal:
This PR brings several improvements to parquet support, improving decoding performance.

## Changes
- Adding a benchmark to measure decoding throughput of different column type/encoding combinations.
- Remove unnecessary allocation/zeroing in decompression path.
- Skip definition decoding (null bitmap) when `null_count=0`.
- Improved `Plain`/`DeltaBinaryPacked`/`Rle`/`RleDictionary` decoders implementation for better performance.
- Adding lookup Table for boolean unpack.
- Specialized UUID byte swapping.
- Adding batch decoder for `DeltaBinaryPacked`.
- Batching nullable bitmap.
- Bulk copying of `PlainPrimitiveDecoder`.
- Comprehensive testing of decoders.
- Fixing bug with incorrect null sentinel values.

## Benchmarks

Ran on Apple M5, there is a lot of jitter.

### Decoding

Iteration: decode 100k values (default page size).

- Median change: -87.10%

| Type | Encoding | Cardinality | Null % | Master (ns/iter) | Patched (ns/iter) | Patched (M elements/s) | Change (±) |
|---|---|---:|---:|---:|---:|---:|---:|
| `array` | `delta_len` |  | 0 | 1,142,737 | 959,031 | 104.3 | -16.08% ± 1.70% |
| `array` | `delta_len` |  | 20 | 916,331 | 786,131 | 127.2 | -14.21% ± 1.29% |
| `array` | `plain` |  | 0 | 577,010 | 386,182 | 258.9 | -33.07% ± 0.85% |
| `array` | `plain` |  | 20 | 493,360 | 339,179 | 294.8 | -31.25% ± 1.22% |
| `binary` | `delta_len` |  | 0 | 1,056,638 | 916,466 | 109.1 | -13.27% ± 1.29% |
| `binary` | `delta_len` |  | 20 | 936,500 | 757,334 | 132.0 | -19.13% ± 2.10% |
| `binary` | `plain` |  | 0 | 536,423 | 403,817 | 247.6 | -24.72% ± 1.36% |
| `binary` | `plain` |  | 20 | 485,270 | 352,759 | 283.5 | -27.31% ± 0.69% |
| `boolean` | `plain` |  | 0 | 188,961 | 4,454 | 22,452.4 | -97.64% ± 0.27% |
| `boolean` | `rle` |  | 0 | 193,054 | 2,042 | 48,980.6 | -98.94% ± 0.03% |
[...]

PR #7206, opened 2026-06-07
perf(parquet): speed up partition-to-parquet conversion and reduce file size

RaphDal:
## Summary

Two independent improvements (separate commits) to the parquet write path shared by `ALTER TABLE ... CONVERT PARTITION TO PARQUET`, `COPY ... TO` parquet export, and the `/exp` HTTP parquet endpoint:

1. Compute symbol/dictionary statistics once per distinct dictionary key per page, not once per row.
2. Default the designated (monotonic) timestamp to `delta_binary_packed` when no explicit encoding is given.

On a 20M-row HOUR partition (release build), conversion is single-threaded and spends ~34% of CPU in symbol min/max statistics; change 1 removes most of that, change 2 shrinks the timestamp column.

## Change 1: symbol statistics per distinct dictionary key

`build_symbol_page_stats` compared every row's referenced dictionary value to update a page's min/max. That min/max depends only on the distinct keys a page references, so the repeated comparisons are redundant. The page now records visited keys in a reused per-column-chunk bitmap (one bit per key) and updates statistics at most once per distinct key (for a 1000-symbol Zipf column with 100k-row pages, ~100k comparisons per page become ~1000).

The bitmap is sized by the dictionary and cleared once per page, so it helps only while the dictionary is small. The encoder uses it for dictionaries up to 65,536 keys (an 8 KiB, L1-resident bitmap) and otherwise compares every row, as before: a high-cardinality symbol has little repetition to dedup within a page, while clearing a key-sized bitmap each page would add an O(num_keys) pass that the dedup cannot repay. Both paths produce identical statistics; the cap only bounds the work to compute them.

Serialized statistics are byte-for-byte identical, so row-group pruning and predicate pushdown are unaffected. The same per-row pattern in VARCHAR/STRING dictionaries (`rle_dictionary/varlen.rs`) is left for a follow-up.

## Change 2: default designated timestamp to delta_binary_packed

`to_encodings` previously defaulted an unencoded `TIMESTAMP` to `plain`. The designated timestamp is monotonic, so it now defaults to `delta_binary_packed`, scoped to the designated timestamp only; other columns keep their defaults, explicit `PARQUET(...)` encodings still win, and compression/bloom settings are untouched. The encoding already had validation, a write path, and reader support, so no new format work is needed.

## Benchmarks

One machine, 32 physical cores, Linux, release build. Table and data, 8 HOUR partitions of 20M rows each:

```sql
CREATE TABLE trades (
  symbol SYMBOL CAPACITY 2048 CACHE,  -- Zipf over 1000 values (exponent 2.0)
  side   SYMBOL CAPACITY 4 CACHE,     -- 'buy' / 'sell'
  price  DOUBLE, amount DOUBLE,       -- uniform in [10, 30)
  timestamp TIMESTAMP
) TIMESTAMP(timestamp) PARTITION BY HOUR BYPASS WAL;

INSERT INTO trades SELECT
  rnd_symbol_zipf(1_000, 2.0), rnd_symbol('buy', 'sell'),
  rnd_double() * 20 + 10, rnd_double() * 20 + 10, generate_series
FROM generate_series('2025-01-01T00:00:00', '2025-01-01T08:00:00', '180u');
```

The 180us step yields 20M rows per partition; `BYPASS WAL` makes `CONVERT` run synchronously on the calling connection so timing is exact. Data is generated once and reused across encodings via `INSERT ... SELECT`, so every row below converts identical bytes. Each time is the mean of 8 per-partition conversions after JIT warm-up (run-to-run spread under 1%).

| designated timestamp encoding | time before | time after | parquet size (vs 640 MB native) |
| --- | --- | --- | --- |
| plain | 1.96 s | 1.32 s | 432 MB (-32.5%) |
| delta_binary_packed | 1.72 s | 1.05 s | 338 MB (-47.2%) |

The within-row time drop is change 1: a CPU profile shows `BinaryMaxMinStats::update` falling from 34.6% to ~0.1% of conversion CPU, threading unchanged. Change 2 moves an unencoded-timestamp table from the first row to the second (1.96 s -> 1.05 s, 432 MB -> 338 MB) with no DDL change; the ~94 MB size gap is the timestamp column.

## Tradeoffs

- Change 2 changes the designated timestamp's on-disk encoding for unencoded tables (now delta, was plain). QuestDB reads both, so existing files stay readable; only newly written partitions differ. For partitions of practical size delta is materially smaller on the ordered timestamp (see the benchmark above). For very small partitions it is reliably larger, not smaller: delta carries fixed per-block framing overhead that dominates when a page holds only a handful of rows. This shows up on ordered data in the refreshed test expectations, not just as a theoretical irregular-gap case -- the 3-row and 6-row `/exp` exports in `ExpParquetExportTest` grow 607 -> 758 B and 860 -> 1160 B, and several small `CopyExportTest` partitions tick up a few bytes (e.g. 610 -> 614 B). The net file-size win holds at scale and reverses for tiny partitions.
- Change 1 adds a small scratch bitmap (one bit per key, e.g. 128 bytes for 1000 keys) and a per-row bit test in place of the per-row update, for dictionaries up to 65,536 keys; larger dictionaries keep the original per-row update with no bitmap and no per-page clear. No measurable wall-clock cost in either regime.
- Numbers are from one machine and one synthetic Zipf data set; the direction should hold broadly, magnitudes vary with hardware, distribution, and codec.

## Test plan

- `cargo test --lib` (qdbr): 910 passed, 0 failed, 1 ignored; `cargo clippy --all-targets` and `cargo fmt --check` clean.
- Change 1 equivalence: identical 2M-row data converted with and without the change; `parquet_metadata` `stats_min_value` / `stats_max_value` / `stats_null_count` matched on all 100 column x row-group entries.
- Change 1 unit tests: `symbol_multi_word_bitmap_stats_reset_across_pages` drives dictionary keys spanning several bitmap words (70, 130) and re-references them in a later page, proving the per-page `seen` bitmap clears on every word, not just word 0; `symbol_all_null_page_has_absent_stats` covers a page with no referenced keys (no min/max, all rows null); `symbol_page_stats_dedup_and_per_row_agree` runs one page through both the bitmap and per-row paths and asserts identical serialized statistics; `symbol_large_dictionary_uses_per_row_stats` forces a dictionary past the 65,536-key cap and checks the per-row path still computes correct min/max/null_count.
- Change 2 unit tests: `test_write_parquet_with_designated_timestamp_ascending` and `_descending` both assert an unencoded designated timestamp resolves to `delta_binary_packed`; `designated_timestamp_explicit_encoding_overrides_delta_default` asserts an explicit `PARQUET(...)` encoding still wins over the default.
- Java test expectations affected by the new sizes/encoding are refreshed in `ExpParquetExportTest` and `CopyExportTest`; any other Java test hardcoding converted-partition sizes or the old default timestamp encoding will surface in CI. `ParquetEncodingTest` is unaffected (delta was already valid for `TIMESTAMP`).


RaphDal:
**Reviewed at level 2** (Agents 1–8 + adversarial performance, batched verification). Level 3 exists for a full mission-critical pass with per-finding verification.

Two independent changes to the Rust parquet write path (shared by `CONVERT PARTITION TO PARQUET`, `COPY ... TO` parquet, and `/exp`):

1. **symbol.rs** — dedup per-page symbol min/max stats via a reused `seen` bitmap (≤65536-key dicts), reuse the validity buffer across pages, and derive null/non-null counts instead of rescanning.
2. **schema.rs** — default an unencoded *designated* timestamp to `delta_binary_packed` instead of `plain`.

The core logic was verified inline and by five parallel review agents (correctness+safety, performance, Java-test completeness, concurrency+resources, quality+metadata). **No correctness, concurrency, resource, or panic bug survived verification.** The change is sound and well-tested. All findings below are optimization or quality nits.

## Critical

None.

## Moderate

### M1 — Two separate O(rows) passes over the same slice; the sibling encoder fuses them

*in-diff, `symbol.rs:152` + `:167`*

Per page, the code traverses `page_values` (a contiguous `&[i32]`) once to fill the validity bitmap (`:280`) and again to compute deduped stats (`:377`). These have no data dependency and could be a single loop. The sibling `fixed.rs` encoder already does exactly this — *"Single pass: build validity bitmap, stats, bloom, and encoded values together"* (`fixed.rs:121`). The PR already fused the cheaper `null_count`/`non_null_len` derivations, so leaving the two main passes separate is an inconsistency on the very path it set out to optimize.

Honest magnitude: small. After this PR, stats fell to ~0.1% of CPU, so the win is one fewer streaming read of the key slice (~512 KiB/page at default page size) — well under 1% of conversion time. Worth doing for consistency with `fixed.rs`, but not blocking, and the separated form is arguably clearer.

## Minor

### m1 — Per-page `seen.fill(0)` has a narrow regression regime the cap doesn't guard

*in-diff, `symbol.rs:394`*

The cap bounds the bitmap by *dictionary size*, but the per-page clear cost is `O(span/64)` regardless of how few rows a page holds. With a near-cap dictionary (~65536 keys → 8 KiB clear/page), **small** pages, and low repetition, the new code does an 8 KiB memset plus a per-row bit-test per page while saving ~zero `stats.update` calls — strictly more work than master's plain per-row loop.

This is unreachable under the default `data_page_size` of 1 MiB (`page_sizing.rs:9`; symbols ≈ 131072 rows/page, making the 8 KiB clear ~0.006% overhead). It only appears if an admin sets `cairo.partition.encoder.parquet.data.page.size` small. Low severity, but a one-line guard (e.g. only dedup when `rows_per_page` is large relative to `span`) would close it. The PR's "no measurable wall-clock cost in either regime" claim holds only for default config.

### m2 — A generation/epoch counter would remove the cap and the per-page clear, at a memory cost

*design note, `symbol.rs:53`, `:392`*

A `Vec<u32>` of "last-seen page index" per key needs no per-page clear (the monotonic page index is the discriminator), which would eliminate `MAX_SYMBOL_STATS_BITMAP_KEYS`, the dual `Some`/`None` path, and m1 in one move. It is **not strictly better**, though: it reintroduces 2–4 bytes/key of scratch for high-cardinality columns (e.g. ~40 MB for a 10M-distinct-value symbol, across parallel column threads), which the current cap deliberately bounds to *zero* extra allocation via the per-row fallback. The current design is a defensible, well-documented tradeoff; flagging the alternative only so the choice is explicit.

### m3 — Stale comment names a non-existent function

*in-diff, test, `symbol.rs:603`*

`// ... the bloom.insert path inside build_symbol_dict_page` — no such function exists; the bloom insert lives in `build_dict_buffer` (via `prepare_symbol_dictionary`). Rename in the comment.

### m4 — Dead `_not_null_hint` param on legacy `symbol_to_pages`

*in-diff, `symbol.rs:420`*

Now ignored, retained only so the two benchmark callers (`benches/encode_page.rs`, `benches/decode_page.rs`) compile. Acceptable as-is (the doc comment says "kept for benchmarks"); optionally drop the param and update both bench calls.

- **i64::MIN/LONG_NULL through delta → decoder overflow panic** — Designated timestamps are `Required` and validated non-null at write time (`MicrosTimestampDriver`/`NanosTimestampDriver.validateBounds0`), so the NULL sentinel never reaches delta encoding. The decoder's non-wrapping `min_delta + delta` is pre-existing vendored parquet2 code unchanged by this PR, and the release profile leaves `overflow-checks` off (`Cargo.toml:42` — wraps, never panics). The pre-PR `plain` path also already used the `encode_int_notnull` (no-null) assumption.
- **Trailing-bit leak from reusing the validity buffer across pages** — `encode_bitmap_def_levels_payload` masks the last partial byte (`util.rs:312`) and reads only `bits[..used_bytes]`; full bytes are all explicitly written by `push_present`/`push_null`. Output is identical to a fresh buffer.
- **`null_count` accumulating across the reused validity** — `reset()` zeros `null_count` and `write_idx` each page (`validity.rs:31`).
- **Dropping `column_top` from `build_symbol_validity` loses leading nulls** — column tops are pre-materialized as `-1` keys in `merged_keys` (`symbol.rs:110`); the only caller always passed `0`.
- **Bitmap OOB indexing** — every referenced key ≤ `max_key`, `value_ranges.len() == max_key+1`, so `key/64 < seen.len()`; empty/all-null pages never enter the `key >= 0` branch.
- **Other Java tests with hardcoded parquet sizes breaking** — exhaustive search of `core/src/test/java/`: only Change 2 alters bytes, and only when `meta.getTimestampIndex() >= 0` (a `TIMESTAMP(ts)` table or `ORDER BY ts` query). Every such `assertGetParquet`/`diskSizeHuman` assertion is in the two files the PR updated; all other size assertions are timestamp-free (plain, unchanged). `ShowPartitionsTest`/`AlterTableConvertPartitionTest` use `parquetFileSize > 0` or placeholders post-convert. `CopyExportTest` ran 5/5 pass locally. (The `/exp` `ExpParquetExportTest` couldn't run in-sandbox — socket bind blocked — but is covered by the same mechanism; CI will confirm the exact bytes.)
- **Reused scratch buffers shared across threads** — `seen_in_page`/`validity` are per-`encode`-call locals; rayon parallelism fans out across distinct column indices (`file.rs:477`), so no sharing. Bloom lock scope unchanged. No native memory; clean `Drop` on all `?` paths.

## Summary

**Verdict: approve.** Both changes are correct, the equivalence of the stats fast-path is proven by tests (`symbol_page_stats_dedup_and_per_row_agree`, `symbol_large_dictionary_uses_per_row_stats`), and the test updates are complete. The PR description is accurate and unusually well-balanced — it states the small-partition size *regression* from delta framing overhead with equal weight to the wins.

- Draft findings: ~9 raised across agents → **2 reported (1 Moderate, 4 Minor), 7 dismissed** as false positives after source verification.
- **All findings in-diff.** Out-of-diff surface (Java test callsites, `to_encodings`/`symbol::encode` callers, the parallel write path) was exhaustively checked and is genuinely clean — the diff is small and self-contained, so zero out-of-diff bugs is a verified conclusion, not an under-run.
- Tradeoffs worth the author's eye: the multi-pass left unfused (M1) on the optimized path, and the per-page-clear soft spot under non-default small-page config (m1). Neither blocks merge.

The one residual external step is letting CI confirm the exact `ExpParquetExportTest` byte values, since that suite can't bind a socket in this environment.

kafka1991:
If use delta encoding for _designed timestamp column_, will that hurt `ParquetTimestampFinder.findTimestamp` performance? (timestamp interval queries lean pretty heavily on this.)

RaphDal:
@kafka1991 
Decoding may or may not take a hit.
From https://github.com/questdb/questdb/pull/6759:
| Type | Encoding | Cardinality | Null % | Master (ns/iter) | Patched (ns/iter) | Patched (M elements/s) | Change (±) |
|---|---|---:|---:|---:|---:|---:|---:|
| `timestamp` | `delta_bp` |  | 0 | 441,311 | 37,593 | 2,660.1 | -91.48% ± 1.05% |
| `timestamp` | `delta_bp` |  | 20 | 494,362 | 69,732 | 1,434.1 | -85.89% ± 0.38% |
| `timestamp` | `plain` |  | 0 | 9,910 | 10,835 | 9,229.8 | +9.33% ± 9.33% |
| `timestamp` | `plain` |  | 20 | 107,857 | 15,945 | 6,271.4 | -85.22% ± 0.31% |

Note that in this benchmark, the dataset is able to live in L2/L3 without contention (100k values * 8 bytes -> 800kB).

In a real-world scenario where we're more memory bound, using `delta_bp` might reduce the amount of data to be transferred from RAM to the cores, making it worth it for decoding too.

kafka1991:
> @kafka1991 Decoding may or may not take a hit. From #6759:
> 
> Type	Encoding	Cardinality	Null %	Master (ns/iter)	Patched (ns/iter)	Patched (M elements/s)	Change (±)
> `timestamp`	`delta_bp`		0	441,311	37,593	2,660.1	-91.48% ± 1.05%
> `timestamp`	`delta_bp`		20	494,362	69,732	1,434.1	-85.89% ± 0.38%
> `timestamp`	`plain`		0	9,910	10,835	9,229.8	+9.33% ± 9.33%
> `timestamp`	`plain`		20	107,857	15,945	6,271.4	-85.22% ± 0.31%
> Note that in this benchmark, the dataset is able to live in L2/L3 without contention (100k values * 8 bytes -> 800kB).


From the performance difference between _delta_ and _plain_,the changes actually trading CPU/read performance for storage space/ write performance.

 >In a real-world scenario where we're more memory bound

Not sure. did we have performance profile to support this conclusion?

Just from intuitively, I think this to matter quite a bit for `ParquetTimestampFinder.findTimestamp` (fetching the value at a given point). With plain encoding we just locate the point in the array directly; but with delta we have to decode all the values from the start of the page just to reconstruct the one we're after.

RaphDal:
@kafka1991
> Just from intuitively, I think this to matter quite a bit for ParquetTimestampFinder.findTimestamp (fetching the value at a given point). With plain encoding we just locate the point in the array directly; but with delta we have to decode all the values from the start of the page just to reconstruct the one we're after.

Indeed, but these plain pages still need to be fetched.

> > In a real-world scenario where we're more memory bound

> Not sure. did we have performance profile to support this conclusion?

This is hard to answer to this one as it depends on the scenario.
In the benchmark I shared earlier, 9B elems/s is ~72GB/s on one core. That is fine from cache on one core, but if the same path becomes an out-of-cache scan and scales across cores, it can very quickly become constrained by memory/storage bandwidth rather than decode throughput.
In cold reads, the smaller delta page also means fewer bytes fetched from disk/object storage; in hot reads, it means fewer bytes moved from RAM to the core. That is the scenario where `delta_bp` can still win despite extra decode work.

If you have any idea of workloads we could try out please share, I'll gladly run them on my workstation.

bluestreak01:
@RaphDal — code review of this PR (level 2: Agents 1–8 + adversarial performance, batched verification).

Two changes to the Rust parquet write path: (1) dedup per-page symbol min/max stats via a reused `seen` bitmap + reuse the validity buffer + derive null counts; (2) default an unencoded **designated timestamp** to `delta_binary_packed` instead of `plain`.

The core write-path logic is **correct and safe** — no correctness, concurrency, resource, or panic bug survived verification (a check round-tripped the reused-validity buffer through Arrow and confirmed null positions). The headline issue is a **read-path performance regression** — exactly the `@kafka1991` concern that the thread left unresolved.

## Moderate

### M1 — Defaulting the designated timestamp to `delta_binary_packed` regresses single-value timestamp decode from O(1) to O(row-group-size)
*in-diff cause (`schema.rs:501-502`), out-of-diff impact (read path)*

Traced end to end:

- **`ParquetTimestampFinder.timestampAt(rowIndex)`** (`ParquetTimestampFinder.java:179-196`) decodes a single row: `decodeRowGroup(..., rowLo, rowLo + 1)`.
- A sub-page range is served by `sink.skip(row_lo); sink.push_slice(row_hi - row_lo)` (`decode.rs:1566-1567`). This branch is taken because the designated timestamp is written `Required`, so there's no null bitmap.
- `skip` cost is **encoding-dependent**: plain is O(1) — `self.values_offset += count` (`plain.rs:252`); delta is **O(count)** — it walks the delta chain `wrapping_add` over every skipped value, block by block (`delta_binary_packed.rs:556-573`). There is no random access into a delta block.
- Under default config the timestamp chunk is a **single page**: page size 1 MiB / 8 bytes = 131,072 rows > 100,000 row-group size (`page_sizing.rs:9,14-15`, `file.rs:26`), and `_pm` carries no intra-chunk page index — so the skip spans the **whole row group**.
- **Hotness:** `IntervalBwdPartitionFrameCursor.java:191` calls `timestampAt(limitHi)` **unconditionally, once per non-skipped partition** on every backward interval scan, with `limitHi = rowCount - 1` for the working partition (`:159`). So each such partition now decodes its entire timestamp row group (~100k delta values, ~tens of µs) just to read the last value — master did this in O(1) with `plain`. `findTimestamp` (`:104-110`) additionally pays a `skip(rowGroupRowLo)` prefix; for the common interval case where the range spans most of the group this is just per-element cost (and delta is competitive there), but it is O(row-group-size) when the matched span sits deep in the group.

**Scoping:** the stats-backed fast paths (`find_row_group_by_timestamp`, `minTimestampExact`/`maxTimestampExact`) are **not** affected — they read inline min/max stats and only decode when stats are absent, which doesn't happen for QDB-written partitions. Read-path **correctness** is tested (`TimestampFinderTest` converts then verifies `timestampAt`/`findTimestamp` against an oracle). So this is a bounded, per-partition (not per-row), correctness-safe regression — but a genuine O(1)→O(n) one on a query path, introduced by a *performance* PR and never measured.

This needs a decision, not necessarily a code change: it's a real read-vs-write tradeoff. At minimum, **measure it** (a `timestampAt`/`findTimestamp` microbench over a delta vs plain converted partition) so the tradeoff is informed. See M2.

### M2 — PR description omits the read-path regression (equal-weight disclosure)

The description discloses the write-side small-partition file-size regression in concrete numbers (607→758 B, 860→1160 B, 610→614 B) and hedges benchmark generality — genuinely balanced **on the size axis**. But it says nothing about read/decode cost; the only occurrence of "read" is "QuestDB reads both, so existing files stay readable" (a compatibility statement). The project convention is to present both positive and negative effects with equal weight and point out regressions as prominently as improvements. The read regression in M1 is the more consequential one (it hits interval-scan queries, not just bytes-on-disk) and was explicitly raised in review. Suggest adding a Tradeoffs bullet describing it, ideally with the M1 measurement, plus a `regression` label.

### M3 — Validity and stats are two separate passes over the page key slice; the sibling encoder fuses them
*in-diff, `symbol.rs:152` + `:167`*

Per page, `build_symbol_validity` (`:282`) streams the contiguous `&[i32]` once to fill the validity bitmap, and `build_symbol_page_stats` (`:395`/`:408`) streams it again for min/max. These have no data dependency and could be one loop — the RLE-encode pass (`:254`) legitimately must follow validity (def-levels are written before data to avoid a buffer memmove), but validity+stats need not be separate. `fixed.rs:121` documents doing exactly the fused version: *"Single pass: build validity bitmap, stats, bloom, and encoded values together."* Honest magnitude: **small** — post-PR stats are ~0.1% of conversion CPU, so this saves one streaming read of the key slice. Flagged for consistency on the very path the PR optimizes, not as a hot-path defect.

## Minor

### m1 — `seen.fill(0)` is O(dict/64) per page regardless of row count; the "no measurable cost in either regime" claim is default-config-only
*in-diff, `symbol.rs:394`*

The per-page clear scales with dictionary size, not rows. The `MAX_SYMBOL_STATS_BITMAP_KEYS` cap bounds it to an 8 KiB clear/page at the cap. At the default 1 MiB page (~131k rows/page) that's ~0.008 clears/row — free. But under a small non-default `cairo.partition.encoder.parquet.data.page.size` (e.g. 512 → 64 rows/page) with a near-cap dictionary and low repetition, the code clears 8 KiB/page while saving ~zero `stats.update` calls — up to ~16× more clearing than rows. The benchmark used a 16-key dictionary, so this regime was never measured. Clearing only touched words (track min/max touched word index) would close it while keeping the 8 KiB bound.

### m2 — Stale comment names a non-existent function
*in-diff, test, `symbol.rs:603`*

`// ... the bloom.insert path inside build_symbol_dict_page` — no such function exists (the string appears only in this comment). The bloom insert is in `build_dict_buffer` (`:365`), via `prepare_symbol_dictionary`. Rename in the comment.

### m3 — Missing targeted tests
*in-diff*

- **Cap boundary:** `symbol_large_dictionary_uses_per_row_stats` uses dict size 65,537 (`> cap` → per-row). No test sits at exactly 65,536 (the last bitmap-path value) to pin the `<=` boundary against an off-by-one.
- **Reused-validity shrink:** the new unit tests assert header/stats counts derived from the validity `null_count` *counter*; none decode the def-level *payload* back. The buffer reuse makes the trailing-byte mask (`util.rs:312-313`) load-bearing for the first time on the symbol path (previously fresh buffers were zeroed). A symbol round-trip with a larger-then-smaller page sequence and interspersed nulls would catch a regression if that mask were ever broken. The code is correct today (verified), so this is defensive coverage.

### m4 — Dead `_not_null_hint` param
*in-diff, `symbol.rs:427`*

Now ignored, kept only so two bench callers compile. Acceptable per its doc comment; optionally drop it and update the 4 callsites (2 benches + 2 tests).

## Design note

A generation/epoch counter (`Vec<u32>` of last-seen page index per key) would remove *both* the per-page clear and the `MAX_SYMBOL_STATS_BITMAP_KEYS` cap, eliminating m1 and the cap's dense-span false-negative. It is **not strictly better**: it costs 2–4 bytes/key of uncapped scratch (≈4 MiB for a 1M-key column, ×parallel column threads) where the current bitmap is capped at 8 KiB and otherwise allocates nothing. The current bitmap+cap is a defensible, well-documented memory tradeoff — noted only so the choice is explicit.

- **Stale-bit leak from reusing the validity buffer** — `encode_bitmap_def_levels_payload` copies only the `full_bytes = len/8` fully-written bytes and masks the trailing partial byte (`util.rs:305-314`); reads are bounded to `bits[..used_bytes]`. Output is identical to a fresh buffer. Round-trip verified.
- **`seen` bitmap OOB / dedup not equivalent to per-row** — every page key ≤ `max_key`, and both `value_ranges` and `seen` are sized from that same `max_key`, so `key/64 < seen.len()` always; min/max is order-independent so dedup ≡ per-row (proven + tested by `symbol_page_stats_dedup_and_per_row_agree`).
- **`non_null_len = len - total_null_count` underflow** — `null_count` is a per-element count of the same slice, so ≤ len. Cannot underflow.
- **Designated-timestamp NULL/`i64::MIN` → delta decode overflow panic** — designated timestamps are `Required`/non-null at write time; release builds leave `overflow-checks` off (wrap, no panic) and parquet2's delta enc/dec are `wrapping_sub`/`wrapping_add` inverses; the pre-PR `plain` path already used the no-null assumption.
- **Dropping `column_top` from `build_symbol_validity` loses leading nulls** — column tops are pre-materialized as `-1` keys in `merged_keys` (`symbol.rs:110`); the only caller passed `0`.
- **Read-path correctness untested after the delta default** — covered: `TimestampFinderTest` converts and verifies `timestampAt`/`findTimestamp` vs an oracle over delta timestamps.
- **Reused scratch buffers shared across threads / resource leak** — `seen_in_page` and `validity` are per-`encode()`-call stack locals; rayon fans out across distinct column indices (`file.rs:477`); RAII drops both on every `?` path; no native memory, no `Drop` impl. Bloom lock scope unchanged.
- **Rust panics on valid or corrupt input** — none reachable in the changed code; all indexing provably in-bounds, all new asserts are side-effect-free debug-only.

## Summary

**Verdict: needs discussion (the code is correct; the tradeoff and disclosure are not settled).** The write-path stats dedup and the file-size win are solid, correct, and well-tested. But the delta-default change introduces a real, reachable, **unmeasured and undisclosed** read-path regression (M1) — single-value timestamp decode goes O(1)→O(row-group-size) on interval scans over converted partitions — exactly the `@kafka1991` concern. Before merge: disclose it with equal weight (M2), ideally measure it, and weigh read-heavy vs write/storage-heavy workloads. M3 is a consistency nit; the rest are minor.

- ~19 draft findings across 9 agents → **9 reported** (3 Moderate + 4 Minor + 1 design note), **9 dismissed** as false positives after source verification.
- All *code* findings are in-diff; M1's cause is a one-line in-diff change in `schema.rs` whose **impact is out-of-diff** (`ParquetTimestampFinder`, the interval cursors, the Rust decoder). The full out-of-diff surface (callsites of `to_encodings`, `symbol::encode`, the validity helpers, the whole timestamp read path) was walked — no broken-correctness out-of-diff callsite; the only out-of-diff consequence is the M1 regression.
- Not completed in-sandbox: exact `ExpParquetExportTest` byte values (socket bind blocked) and a clean Rust rebuild (`cargo check` hit a stale-toolchain `E0514` on cached deps, unrelated to this PR) — CI will confirm both.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

RaphDal:
@bluestreak01 @kafka1991 - I asked Claude to make some benchmarks to verify whether `delta_bp` is a good default for the designated timestamp or not.

TL;DR: decoding even a single row still decompresses the whole page it sits in. With `delta_bp`, `LZ4` often can't compress the (already bit-packed) data, so that "decompression" is just a copy, much faster than really decompressing the genuinely-compressed plain page.

Built the `timestampAt` microbench and pushed it in https://github.com/questdb/questdb/pull/7206/commits/a94b7be6e763829e85fde05533958d2217bcc455. Correction up front: plain isn't O(1) under the default LZ4 codec. Its `skip()` is O(1), but it runs after `decodeRowGroup`

decompresses the whole page, so plain is O(page) (flat) while delta is O(offset).
At the `IntervalBwd:191` call (`timestampAt(rowCount-1)` = last row), delta is faster.

`timestampAt`, µs/op:
| row offset in its row group (1 page) | PLAIN | DELTA |
| --- | --- | --- |
| first row | 800.7 | 3.4 |
| ~50k | 800.5 | 57.6 |
| last row (the `IntervalBwd` call) | 805.8 | 108.6 |

Repro:
- `CREATE TABLE t (price DOUBLE, timestamp TIMESTAMP) TIMESTAMP(timestamp) PARTITION BY HOUR BYPASS WAL`
- 2,000,000 rows, irregular timestamps: `(1704067200000000 + (x-1)*2000 + rnd_long(0,1999,0))::timestamp FROM long_sequence(2000000) x` (2000µs step + uniform jitter → ~13-bit deltas; ~1.8M rows land in partition 0)
- PLAIN arm: `ALTER TABLE t ALTER COLUMN timestamp SET PARQUET(PLAIN)` before convert; DELTA arm: default

- `ALTER TABLE t CONVERT PARTITION TO PARQUET WHERE timestamp >= 0`
- codec `LZ4_RAW` (default); default 1 MiB page → 100k-row row group = 1 page/group (matches your trace)
- measure `ParquetTimestampFinder.timestampAt()` on partition 0 (non-active); JMH AverageTime; offsets = row 0 / ~50k / row-group end / partition last row

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 200 / 207 (96.62%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/parquet_write/encoders/symbol.rs|172|179|96.09%|
|:large_blue_circle:|qdbr/src/parquet_write/encoders/helpers/validity.rs|3|3|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/schema.rs|2|2|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/mod.rs|23|23|100.00%|

PR #7207, opened 2026-06-07
test(core): fix flaky easing-up log assertion in RssMemoryLimitTest

glasstiger:
`RssMemoryLimitTest.testLargeTxEventuallySucceeds` intermittently times out,
mostly on the slower Windows and Mac CI jobs, at:

```
capture.waitForRegex("table writing memory pressure is easing up \\[table=");
```

## Root cause

The assertion waits for an INFO line emitted by `ApplyWal2TableJob`, but that
logger is non-guaranteed. QuestDB's async logger discards a record when the ring
buffer is full: `Logger.next()` receives a negative cursor from `seq.next()` and
returns `NullLogRecord`, dropping the message. A guaranteed logger instead calls
`seq.nextBully()`, which blocks until a slot frees.

During the test, each WAL apply merges a 500k-row transaction into ~366 daily
partitions and emits a burst of per-partition `TableWriter` records into the
shared log buffer. The apply job logs the "easing up" line immediately behind
that burst; while the consumer is still draining the burst the buffer stays
full, so the line is dropped and the wait never completes. The "job finished"
line emitted just before it disappears for the same reason. The memory-pressure
recovery itself behaves correctly in the failing runs — the data is applied, the
count matches, and the table is never suspended — only the log line the test
waits on is lost.

The test already marks `QueryProgress` as guaranteed, which is why the
count-query assertions never flake, but it did not mark the apply job.

## Change

- `RssMemoryLimitTest` adds `ApplyWal2TableJob.class` to the
  `enableGuaranteedLogging` / `disableGuaranteedLogging` set, so the apply job's
  producer blocks for a buffer slot instead of dropping the record.
- `ApplyWal2TableJob.LOG` becomes non-final. `LogFactory.setLogger` installs the
  guaranteed logger by setting the static `LOG` field via reflection, which is
  not reliable on a final field. `QueryProgress.LOG` is non-final for the same
  reason.

## Tradeoffs

- Production behavior is unchanged: the `LOG` field is only swapped by test
  setup, never at runtime.
- The fix covers only the line the test asserts on. The high-volume
  `TableWriter` records stay non-guaranteed and can still be dropped under buffer
  pressure; this PR does not change that.
- Guaranteed logging lets the apply job's producer block on a full buffer. The
  number of `ApplyWal2TableJob` records is small, so the added latency stays
  within the test.
- `ApplyWal2TableJob.LOG` is now a mutable static field rather than a constant.

## Test plan

- Run `RssMemoryLimitTest` on the Linux, Windows, and Mac CI jobs.
- Confirm `testLargeTxEventuallySucceeds` passes, including repeated runs on the
  Windows and Mac jobs where it previously flaked.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/wal/ApplyWal2TableJob.java
+++ b/core/src/main/java/io/questdb/cairo/wal/ApplyWal2TableJob.java
@@ -82,7 +82,9 @@ import static io.questdb.tasks.TableWriterTask.CMD_ALTER_TABLE;
 import static io.questdb.tasks.TableWriterTask.CMD_UPDATE_TABLE;
 
 public class ApplyWal2TableJob extends AbstractQueueConsumerJob<WalTxnNotificationTask> implements Closeable {
-    private static final Log LOG = LogFactory.getLog(ApplyWal2TableJob.class);
+    // this field is modified via reflection from tests, via LogFactory.enableGuaranteedLogging
+    @SuppressWarnings("FieldMayBeFinal")
+    private static Log LOG = LogFactory.getLog(ApplyWal2TableJob.class);
     private final BlockFileWriter blockFileWriter;
     private final CairoConfiguration config;
     private final CairoEngine engine;
--- a/core/src/test/java/io/questdb/test/cairo/RssMemoryLimitTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/RssMemoryLimitTest.java
@@ -27,6 +27,7 @@ package io.questdb.test.cairo;
 import io.questdb.PropertyKey;
 import io.questdb.cairo.CairoException;
 import io.questdb.cairo.TableToken;
+import io.questdb.cairo.wal.ApplyWal2TableJob;
 import io.questdb.griffin.engine.QueryProgress;
 import io.questdb.log.LogFactory;
 import io.questdb.std.MemoryTag;
@@ -47,7 +48,7 @@ public class RssMemoryLimitTest extends AbstractCairoTest {
 
     @Override
     public void setUp() {
-        LogFactory.enableGuaranteedLogging(QueryProgress.class);
+        LogFactory.enableGuaranteedLogging(QueryProgress.class, ApplyWal2TableJob.class);
         super.setUp();
         capture.start();
     }
@@ -56,7 +57,7 @@ public class RssMemoryLimitTest extends AbstractCairoTest {
     public void tearDown() throws Exception {
         capture.stop();
         super.tearDown();
-        LogFactory.disableGuaranteedLogging(QueryProgress.class);
+        LogFactory.disableGuaranteedLogging(QueryProgress.class, ApplyWal2TableJob.class);
     }
 
     @Test


mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/wal/ApplyWal2TableJob.java|1|1|100.00%|

glasstiger:
## Code review (level 3) — verdict: approve

Correct, minimal flaky-test fix that follows the existing `QueryProgress` / `LineHttpProcessorState` precedents exactly.

### Why the production change is necessary (not cosmetic)
The `final` → non-`final` change on `ApplyWal2TableJob.LOG` is required for the fix to work: `LogFactory.setLogger` installs the guaranteed logger via `Field.set(null, logger)`, and the JVM forbids reflective writes to a `static final` field even with `setAccessible(true)` (it throws `IllegalAccessException`). Without dropping `final`, `enableGuaranteedLogging(ApplyWal2TableJob.class)` would throw `RuntimeException("Could not set logger", ...)` and `setUp()` would fail. The added comment + `@SuppressWarnings("FieldMayBeFinal")` are byte-identical to the two existing precedents.

### Completeness — confirmed
- `testLargeTxEventuallySucceeds` waits on exactly one log line (`"table writing memory pressure is easing up [table="`, `ApplyWal2TableJob:971`), which is now guaranteed.
- The other tests in the class wait only on `QueryProgress` lines (already guaranteed) or assert via API state.
- That line is gated by `onEnoughMemory()` (`TableWriterPressureControlImpl:87`), which returns `true` only after `onOutOfMemory()` has fired at least once — so the assertion meaningfully proves a pressure episode occurred. Because the RSS budget is QuestDB-tracked (`Unsafe.setRssMemLimit`, not OS RSS), each 500k-row apply crosses it deterministically on every platform, so the line is reliably emitted; the Mac/Windows correlation is explained by the slower log consumer leaving the ring full → the non-guaranteed line dropped. The root-cause analysis in the description checks out.

### Minor (all optional, non-blocking)
1. **`tearDown` exception-safety (pre-existing pattern).** `disableGuaranteedLogging` is the last statement, after `super.tearDown()`. If `super.tearDown()` throws, the disable is skipped and `ApplyWal2TableJob.LOG` leaks as *guaranteed* into later test classes in the same JVM fork. The current ordering is deliberately correct — the field must only be reset after the apply-job worker threads stop, to avoid a non-final-static visibility race — so a fully-safe version would wrap it as `try { super.tearDown(); } finally { disable...; }`. Impact is benign (drop-vs-block only, never log content/correctness), and this shape is shared by ~6 other classes that toggle `QueryProgress`, so it's better as a follow-up across all of them than a blocker here.
2. **Comment wording (`RssMemoryLimitTest:121-122`).** "otherwise the test passes trivially" — if pressure were not triggered the test would actually *fail* (the `waitForRegex` times out), not pass trivially. The mechanism is right; only the phrasing is slightly off.
3. **Labels.** Consider adding `WAL` (the touched production file is the WAL apply job); `CI` + `Flaky Test` are otherwise appropriate.

### Checked and clear
- No production concurrency/visibility issue — the field is assigned only in `<clinit>` and never reassigned at runtime (safe publication via class init).
- No hot-path cost — every `LOG.*` site in the class is a per-transaction / per-error control-plane event, never per-row; same tradeoff already accepted for `QueryProgress.LOG` (which sits on the query hot path).
- No class-init clobber — `Field.set` forces `<clinit>` to complete first, then overwrites with the guaranteed logger.
- The diff is a clean, symmetric two-line change to `enableGuaranteedLogging`/`disableGuaranteedLogging` (no whitespace churn, no base asymmetry).

<sub>🤖 Automated review via [Claude Code](https://claude.com/claude-code) (level 3).</sub>

ideoma (approved):

PR #7208, opened 2026-06-08
fix(sql): reject views and materialized views in SHOW CREATE TABLE

nwoolmer:
## What

`SHOW CREATE TABLE <name>` resolved any object kind and rendered a `CREATE TABLE` statement for it. Run against a view or materialized view, it produced misleading DDL: the output looked like a table definition, and executing it would have created a plain table rather than the view. The dedicated `SHOW CREATE VIEW` / `SHOW CREATE MATERIALIZED VIEW` statements already rejected the wrong object kind; the plain-table path did not.

This change makes `SHOW CREATE TABLE` reject views and materialized views.

## Changes

- **`SqlParserCallback.getTableToken`** now throws `table name expected, got view or materialized view name` (pointing at the name) when the resolved token is a view or materialized view. Both the OSS and Enterprise compile paths call this shared helper, so both are covered. The existing "table does not exist" check still runs first, so an unknown name continues to report that rather than the new message.
- **Sibling guards tightened for symmetry.** Previously `SHOW CREATE VIEW <matview>` and `SHOW CREATE MATERIALIZED VIEW <view>` both reported the generic "got table name". They now report "got materialized view name" and "got view name" respectively, so the error names the actual object kind.
- **Defensive assertion** in `ShowCreateTableRecordCursorFactory.ShowCreateTableCursor.of()`: if a view/matview token ever reaches the factory through a future caller that bypasses the parser guard, it fails fast instead of rendering bogus DDL. This is unreachable today (grep confirms the only two construction sites both go through `getTableToken`).

## Behavior change / compatibility

This is a user-visible behavior change: `SHOW CREATE TABLE <view>` previously returned a (misleading) result and now errors. Anyone who relied on the old output should use `SHOW CREATE VIEW` / `SHOW CREATE MATERIALIZED VIEW`. No existing test depended on the old behavior.

## Test plan

- New tests in `ShowCreateTableTest`: matview and view rejection (plain, quoted, and mixed-case identifiers), an explicit check that a non-existent name still reports "table does not exist", and two regression guards confirming the views remain reachable via their dedicated statements.
- New `ViewsFunctionTest.testShowCreateViewFailMatView` and `CreateMatViewTest.testShowCreateMatViewFailView` cover the tightened sibling messages.
- Full `ShowCreateTableTest` passes (45 tests), as does the Enterprise subclass which inherits these tests against the Enterprise compiler path (51 tests).

### DIFF
--- a/core/src/main/java/io/questdb/griffin/SqlParserCallback.java
+++ b/core/src/main/java/io/questdb/griffin/SqlParserCallback.java
@@ -52,15 +52,21 @@ public interface SqlParserCallback {
                 SqlException.matViewDoesNotExist(tableNameExpr.position, tableNameExpr.token)
         );
         if (!viewToken.isMatView()) {
-            throw SqlException.$(tableNameExpr.position, "materialized view name expected, got table name");
+            throw SqlException.$(tableNameExpr.position, viewToken.isView()
+                    ? "materialized view name expected, got view name"
+                    : "materialized view name expected, got table name");
         }
         return viewToken;
     }
 
-    static TableToken getTableToken(ExpressionNode tableNameExpr, SqlExecutionContext executionContext, Path path) throws SqlException {
-        return getTableToken(tableNameExpr, executionContext, path,
+    static @NotNull TableToken getTableToken(ExpressionNode tableNameExpr, SqlExecutionContext executionContext, Path path) throws SqlException {
+        final TableToken tableToken = getTableToken(tableNameExpr, executionContext, path,
                 SqlException.tableDoesNotExist(tableNameExpr.position, tableNameExpr.token)
         );
+        if (tableToken.isMatView() || tableToken.isView()) {
+            throw SqlException.$(tableNameExpr.position, "table name expected, got view or materialized view name");
+        }
+        return tableToken;
     }
 
     static @NotNull TableToken getViewToken(ExpressionNode tableNameExpr, SqlExecutionContext executionContext, Path path) throws SqlException {
@@ -68,7 +74,9 @@ public interface SqlParserCallback {
                 SqlException.viewDoesNotExist(tableNameExpr.position, tableNameExpr.token)
         );
         if (!viewToken.isView()) {
-            throw SqlException.$(tableNameExpr.position, "view name expected, got table name");
+            throw SqlException.$(tableNameExpr.position, viewToken.isMatView()
+                    ? "view name expected, got materialized view name"
+                    : "view name expected, got table name");
         }
         return viewToken;
     }
--- a/core/src/main/java/io/questdb/griffin/engine/table/ShowCreateTableRecordCursorFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/table/ShowCreateTableRecordCursorFactory.java
@@ -186,6 +186,9 @@ public class ShowCreateTableRecordCursorFactory extends AbstractRecordCursorFact
                     throw TableReferenceOutOfDateException.of(this.tableToken);
                 }
             }
+            // SHOW CREATE TABLE rejects views and materialized views during parsing
+            // (see SqlParserCallback.getTableToken); guard against any future caller that bypasses it.
+            assert !tableToken.isView() && !tableToken.isMatView();
 
             toTop();
             return this;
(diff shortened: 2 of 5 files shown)


nwoolmer:
## Code review (level 2)

Reviewed via a level-2 multi-agent pass (correctness/cross-context, test coverage, standards + fresh-context adversarial) with per-finding source verification. Original scope: the `getTableToken` guard only — the sibling-message and assertion changes below were added *in response* to this review (see the follow-up comment).

### Change surface
- **Symbol:** `SqlParserCallback.getTableToken(ExpressionNode, SqlExecutionContext, Path)` — `public static`.
- **Before:** resolves token; throws "table does not exist" if absent; returns the token for *any* object kind.
- **After:** same, plus throws `table name expected, got view or materialized view name` at the name position when `isMatView() || isView()`.
- **Callsite inventory (verified by grep, both repos):** exactly two callers — `SqlParserCallback.generateShowCreateTableFactory` (OSS) and `EntSqlCompilerImpl.generateShowCreateTableFactory` (Enterprise). Both are SHOW CREATE TABLE paths. No existing test exercises SHOW CREATE TABLE on a view/matview, so no regression risk.

### Findings

**Critical / Moderate:** none.

**Minor (both addressed in the follow-up):**
- Defense-in-depth: `ShowCreateTableCursor.of()` rendered DDL without re-checking the token kind. Only reachable if a future caller bypassed the guard, but cheap to harden.
- Pre-existing asymmetry: `getViewToken`/`getMatViewToken` reported the generic "got table name" even when handed the *other* view kind.

### Downgraded (false positives, verified against source)
- **NPE on null token + `@NotNull`** — disproven. `CairoEngine.getTableStatus(path, null)` maps null → `TABLE_DOES_NOT_EXIST` (CairoEngine.java:1100), so the method throws before dereferencing. The `@NotNull` is truthful.
- **Incomplete guard (a view-like kind slipping through as false/false)** — disproven. `TableToken.Type` is the exhaustive enum `{TABLE, VIEW, MAT_VIEW}` set to exactly one value; `isMatView() || isView()` covers both non-table kinds.

### Verified correct
- Error position (= name position) matches the project convention and the sibling guards.
- Existence check runs **before** the type check — unknown name still reports "table does not exist".
- Error wording is verbatim-reused from the established `SqlCompilerImpl` DROP-TABLE guard.
- Exception propagates cleanly through `SqlOptimiser` dispatch to the user.
- Standards: member ordering preserved, `@NotNull` already imported, ASCII-only message, no LOG chains.

**Verdict: approve.** 9 draft findings → 2 false positives dropped, 0 blocking. All findings in-diff; both out-of-diff callers confirmed safe.

nwoolmer:
## Fixes applied in response to the review

Both Minor findings from the review above were addressed in this PR (they are part of the diff, not deferred):

**1. Defense-in-depth assertion** — `ShowCreateTableRecordCursorFactory.ShowCreateTableCursor.of()` now asserts the resolved token is neither a view nor a materialized view:

```java
// SHOW CREATE TABLE rejects views and materialized views during parsing
// (see SqlParserCallback.getTableToken); guard against any future caller that bypasses it.
assert !tableToken.isView() && !tableToken.isMatView();
```

This is unreachable through today's call graph (the parser guard catches it first); it exists so a future caller that constructs the factory directly fails fast under `-ea` rather than silently rendering bogus DDL.

**2. Symmetric sibling-guard messages** — `getViewToken` and `getMatViewToken` previously collapsed every wrong-kind input to "got table name". They now name the actual object kind:

- `SHOW CREATE VIEW <matview>` → `view name expected, got materialized view name`
- `SHOW CREATE MATERIALIZED VIEW <view>` → `materialized view name expected, got view name`
- the plain-table cases are unchanged (`got table name`), so existing tests still pass.

### Tests added for the fixes
- `ViewsFunctionTest.testShowCreateViewFailMatView`
- `CreateMatViewTest.testShowCreateMatViewFailView`
- `ShowCreateTableTest.testMissingNameReportsDoesNotExist` (pins the existence-before-type ordering flagged during the test-coverage pass)

### Not changed
- The `@NotNull` and incomplete-guard concerns were verified as false positives (see the review comment) — no code change warranted.

All affected suites pass with assertions enabled: full `ShowCreateTableTest` (45), `EntShowCreateTableTest` (51), plus the new `ViewsFunctionTest` / `CreateMatViewTest` cases.

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlParserCallback.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowCreateTableRecordCursorFactory.java|1|1|100.00%|

PR #7209, opened 2026-06-08
fix(sql): fix internal error when an IN subquery contains a join

puzpuzpuz:
## What

A query that nests a join inside an `IN` sub-query — for example:

```sql
SELECT * FROM trades
WHERE symbol IN (
    SELECT t.symbol FROM fx_trades t
    HORIZON JOIN market_data m ON (symbol)
    RANGE FROM -30s TO 30s STEP 5s AS h
);
```

failed at compile/validation time with an internal `NullPointerException` (`Cannot read field "type" because "col" is null` in `WhereClauseParser.analyzeIn`). The same shape reached users through the web console's query validation.

## Why

`SqlParser` parses a join's `ON` clause with its shared `ExpressionTreeBuilder` and drains the parsed operands with `poll()`, but — unlike every other expression parse, which goes through `expr()` and brackets `parseExpr` with `pushModel`/`popModel` — it never raised an arg-stack floor for the `ON` clause.

Expression parsing is reentrant. When the join sits inside a lambda sub-query that is an operand of an outer operator (here, the `IN`), the outer operand (`symbol`) is still on the shared stack. The `ON`-clause drain loop runs down to the outer floor and consumes that operand as a bogus join column. The `in` node is then built with a null left operand, which the code generator dereferences and crashes on.

The symptom varied with the `ON` syntax, but the cause was identical:

- shorthand `ON (col)` → `NullPointerException`
- explicit `ON a = b` → `Column name expected`

## How

`ExpressionTreeBuilder` gains `pushArgStackBottom()` / `popArgStackBottom()`, and `parseJoin` brackets the `ON`-clause parse with them. For the duration of that parse they do two things, leaving the current model in place (unlike `pushModel`):

- raise the arg-stack floor (via the existing `argStackBottomStack`) so the drain can only consume the join's own operands;
- block sub-queries, so an `ON`-clause sub-query rejects with `query is not allowed here`.

The sub-query block is needed because a join nested in a lambda sub-query parses its `ON` clause with the outer query model in scope. Without it, the validation that rejects `ON`-clause sub-queries at top level would not fire when nested, and the sub-query would attach to the wrong scope — compiling `ON x IN (SELECT ...)` with cross-join semantics, or crashing on a bare `ON (SELECT ...)` with a second `NullPointerException`.

## Effects and tradeoffs

- Queries with a join inside an `IN` (or other lambda) sub-query now compile and run instead of crashing.
- The explicit `ON a = b` form inside such a sub-query, which previously returned `Column name expected`, now compiles. A caller that relied on that error message will see the query succeed instead.
- A sub-query written inside the nested `ON` clause itself — `ON x IN (SELECT ...)` or a bare `ON (SELECT ...)` — now rejects with `query is not allowed here`, the same as at top level. Previously the first form compiled with wrong cross-join semantics and the second crashed with a `NullPointerException`. Both are unsupported syntax that errored or crashed before, so no correct result changes.
- Top-level joins are unaffected: with no pending outer operand the floor equals the stack size, and the sub-query block is redundant with the existing top-level rejection, so the drain behaves exactly as before. The parser/join/where-clause suites confirm no change.

## Test plan

- Regression test `JoinTest#testInSubQueryWithJoinOnClause` covers HORIZON and ASOF joins inside an `IN` sub-query (shorthand `ON (col)` and explicit `ON a = b`), a `NOT IN` variant, and two negative cases asserting that a sub-query inside the nested `ON` clause rejects with `query is not allowed here`.
- Ran SqlParserTest, JoinTest, WhereClauseParserTest, HorizonJoinTest, WindowJoinTest, AsOfJoinTest, HashJoinTest, LatestByTest, InTest, ExpressionParserTest, SqlOptimiserTest, ExpressionParserFuzzTest — all green.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/ExpressionTreeBuilder.java
+++ b/core/src/main/java/io/questdb/griffin/ExpressionTreeBuilder.java
@@ -78,9 +78,20 @@ public final class ExpressionTreeBuilder implements ExpressionParserListener {
         return argStack.size() > argStackBottom ? argStack.poll() : null;
     }
 
+    void popArgStackBottom() {
+        argStackBottom = argStackBottomStack.notEmpty() ? argStackBottomStack.pop() : 0;
+    }
+
     void popModel() {
         this.model = modelStack.poll();
         this.argStackBottom = argStackBottomStack.notEmpty() ? argStackBottomStack.pop() : 0;
+    }
+
+    // Raises the arg-stack floor so a reentrant parse cannot poll operands of the
+    // enclosing expression. Unlike pushModel(), it leaves the current model in place.
+    void pushArgStackBottom() {
+        argStackBottomStack.push(argStackBottom);
+        argStackBottom = argStack.size();
     }
 
     void pushModel(IQueryModel model) {
--- a/core/src/main/java/io/questdb/griffin/SqlParser.java
+++ b/core/src/main/java/io/questdb/griffin/SqlParser.java
@@ -3500,6 +3500,10 @@ public class SqlParser {
             case IQueryModel.JOIN_FULL_OUTER:
                 expectTok(lexer, tok, "on");
                 onClauseObserved = true;
+                // A join nested in a lambda sub-query (e.g. "x IN (SELECT ... JOIN ... ON ...)")
+                // leaves the outer operand on the shared arg stack; raise the floor so draining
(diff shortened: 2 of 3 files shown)


puzpuzpuz:
Closing in favor of #7211

PR #7211, opened 2026-06-08
fix(sql): fix internal error when an IN subquery contains a join, and handle declared variables in join ON clauses

puzpuzpuz:
## What

A query that nests a join inside an `IN` sub-query — for example:

```sql
SELECT * FROM trades
WHERE symbol IN (
    SELECT t.symbol FROM fx_trades t
    HORIZON JOIN market_data m ON (symbol)
    RANGE FROM -30s TO 30s STEP 5s AS h
);
```

failed at compile/validation time with an internal `NullPointerException` (`Cannot read field "type" because "col" is null` in `WhereClauseParser.analyzeIn`). The same shape reached users through the web console's query validation.

This PR also makes the join `ON` clause handle declared variables (`DECLARE @v := ...`) consistently, in the same parser path.

## Why

`SqlParser` parses a join's `ON` clause with its shared `ExpressionTreeBuilder` and drains the parsed operands with `poll()`, but — unlike every other expression parse, which goes through `expr()` and brackets `parseExpr` with `pushModel`/`popModel` — it never raised an arg-stack floor for the `ON` clause.

Expression parsing is reentrant. When the join sits inside a lambda sub-query that is an operand of an outer operator (here, the `IN`), the outer operand (`symbol`) is still on the shared stack. The `ON`-clause drain loop runs down to the outer floor and consumes that operand as a bogus join column. The `in` node is then built with a null left operand, which the code generator dereferences and crashes on.

The symptom varied with the `ON` syntax, but the cause was identical:

- shorthand `ON (col)` → `NullPointerException`
- explicit `ON a = b` → `Column name expected`

## How

`ExpressionTreeBuilder` gains `pushArgStackBottom()` / `popArgStackBottom()`, and `parseJoin` brackets the `ON`-clause parse with them. For the duration of that parse they do two things, leaving the current model in place (unlike `pushModel`):

- raise the arg-stack floor (via the existing `argStackBottomStack`) so the drain can only consume the join's own operands;
- block sub-queries, so an `ON`-clause sub-query rejects with `query is not allowed here`.

The sub-query block is needed because a join nested in a lambda sub-query parses its `ON` clause with the outer query model in scope. Without it (floor raised but sub-queries not blocked), the validation that rejects `ON`-clause sub-queries at top level would not fire when nested, and the sub-query would attach to the wrong scope — compiling `ON x IN (SELECT ...)` with cross-join semantics, or crashing on a bare `ON (SELECT ...)` with a second `NullPointerException`. This PR ships the floor raise and the block together, so that intermediate state never reaches master.

A second, pre-existing defect surfaced on the same path. When a parse error unwinds, the error handler clears the shared parse stacks, but the `finally` blocks that restore each stack's saved bottom did so unconditionally; restoring a saved bottom over the emptied stack threw an internal `IllegalStateException` ("Tried to set bottom beyond the top of the stack") and masked the real positioned error. Two methods were affected: `ExpressionParser.parseExpr` (the scope stack, reachable when a parse error occurred two or more lambda levels deep) and `ExpressionParser.parseWindowExpr` (the operator, param-count, and arg-stack-depth stacks, reachable when a window frame bound had a parse error inside a lambda sub-query). Both now clamp each restored bottom to the current stack size, so the original positioned error survives. Neither case requires a join.

## Declared variables in the ON clause

A declared variable (`DECLARE @v := ...`) is a literal at parse time and only expands to its definition later, in `rewriteKnownStatements`. `ON`-clause sub-queries are unsupported, but a variable bound to a sub-query — `@q := (SELECT ...)` used as `ON x IN @q` — slipped past the parse-time block and compiled to the same surprising cross-join semantics this PR closes for the inline form. The reject was first added for the *criteria* forms; it ran in `parseJoin`'s single-node non-literal branch, after the variable was expanded. The shorthand join-column branches never expanded the variable, so a declared variable used as a bare/shorthand join column slipped through.

`parseJoin` now expands declared variables up front in both shorthand branches, before the literal/criteria dispatch, so the declared form matches its inline expansion at every nesting depth and in every `ON`-clause position:

- a variable bound to a sub-query is rejected with `query is not allowed here` — in the operator forms (`IN`/`=`/`>`), the single-column shorthand `ON (@q)` / `ON @q`, the multi-column shorthand `ON (@q, ts)`, and nested in a lambda sub-query alike. Previously only the operator forms rejected; the bare-column forms leaked the unexpanded literal as `Invalid column: s.@q`.
- a variable bound to a bare column expands to a real shorthand join column: `ON (@c)` with `@c := symbol` behaves exactly like `ON (symbol)` → `src.symbol = ref.symbol`, instead of erroring with `Invalid column: s.@c`. This is a deliberate, additive feature: a declared variable equals its inline expansion in every `ON`-clause position.

Plain `ON (a)`, `ON (a, b)`, and `ON a = b` are unchanged — `rewriteKnownStatements` is a no-op on a bare column literal.

## Effects and tradeoffs

- Queries with a join inside an `IN` (or other lambda) sub-query now compile and run instead of crashing. This covers `IN`/`NOT IN` and scalar operands (`=`, `>`, etc.) alike, since the floor raise protects any lambda operand.
- The explicit `ON a = b` form inside such a sub-query, which previously returned `Column name expected`, now compiles. A caller that relied on that error message will see the query succeed instead.
- A sub-query written inside the nested `ON` clause itself — `ON x IN (SELECT ...)` or a bare `ON (SELECT ...)` — now rejects with `query is not allowed here`, the same as at top level. On master both forms returned a misleading `Column name expected` (the drain reached a non-literal node after consuming the enclosing operand). Both are unsupported syntax that errored before, so no correct result changes.
- A declared variable bound to a sub-query, used anywhere in an `ON` clause, now rejects with `query is not allowed here`, consistently across operator and shorthand forms. Previously the shorthand forms (`ON (@q)`, `ON @q`, `ON (@q, ts)`) returned a misleading `Invalid column: s.@q`. Unsupported syntax that errored before, so no correct result changes.
- A declared variable bound to a column may now be used as a shorthand join column (`ON (@c)`), matching the inline column it expands to. This previously errored, so it is strictly additive.
- A parse error that unwinds through a raised parse-stack bottom now surfaces its positioned `SqlException` instead of an internal `IllegalStateException`. This covers a general expression error nested two or more lambda levels deep and a window frame-bound error nested in a lambda sub-query. Both were latent bugs on master, reachable without any join.
- Top-level joins are unaffected: with no pending outer operand the floor equals the stack size, and the sub-query block is redundant with the existing top-level rejection, so the drain behaves exactly as before. The parser/join/where-clause suites confirm no change.

## Test plan

- Regression test `JoinTest#testInSubQueryWithJoinOnClause` covers HORIZON, ASOF, and INNER joins inside an `IN` sub-query (shorthand `ON (col)`, explicit `ON a = b`, and multi-column `ON (a, b)`), scalar `=` and `>` operands whose sub-query contains a join, a `NOT IN` variant, four negative cases asserting that a sub-query inside the nested `ON` clause rejects with `query is not allowed here` (at one and two lambda levels deep), and a pooled-compiler reuse case proving the shared parser state is clean after a rejection.
- `JoinTest#testJoinOnClauseRejectsDeclaredSubQuery` covers the declared-variable sub-query reject across the operator forms and now also the single-column `ON (@q)`, no-parentheses `ON @q`, multi-column `ON (@q, ts)`, and nested shorthand forms, plus a pooled-compiler reuse case.
- `JoinTest#testJoinOnClauseDeclaredColumnShorthand` proves a column-valued declared variable joins correctly as a shorthand join column (single-column, no-parentheses, and multi-column) and matches the equivalent inline `ON (symbol)` form.
- Regression test `SqlParserTest#testNestedSubQueryParseErrorIsPositioned` covers the scope-stack clamp: a parse error two lambda levels deep must surface its positioned `SqlException` rather than a masking internal `IllegalStateException`.
- Regression test `SqlParserTest#testNestedWindowFrameParseErrorIsPositioned` covers the `parseWindowExpr` clamp: a window frame-bound parse error inside an `IN` sub-query must surface its positioned `SqlException` rather than the masking internal `IllegalStateException`.
- Ran SqlParserTest, JoinTest, DeclareTest, WhereClauseParserTest, HorizonJoinTest, WindowJoinTest, AsOfJoinTest, HashJoinTest, LatestByTest, InTest, ExpressionParserTest, ExpressionParserFuzzTest, SqlOptimiserTest, PercentRankFunctionTest, EmaWindowFunctionTest — all green.


puzpuzpuz:
Review: PR #7211 — fix(sql): fix internal error when an IN subquery contains a join
  
  Level 3 (full mission-critical pass: change-surface map, 10 parallel agents — Agent 8/Rust skipped, no .rs files — plus per-finding source verification and an independent reproducer
  build).

  The core fix is correct, well-tested, thread-safe, perf-neutral, and regresses nothing. The arg-stack floor (pushArgStackBottom/popArgStackBottom), the subQueryBlockDepth ON-clause
  sub-query block, and the scopeStack clamp all hold up under adversarial scrutiny. I verified the central regression risk — that blocking ON-clause sub-queries could break a
  previously-working query — and it does not: top-level ON-clause sub-queries were always rejected (model==null at ExpressionParser.java:948), and the nested form was always buggy (NPE /
  "Column name expected"), so no correct result changes.

  One substantive finding: the PR's secondary fix (the masking-IllegalStateException cleanup) is incomplete and the PR body overstates its reach.

  Critical

  None.

  Moderate

  M1 — The masking IllegalStateException fix is incomplete; parseWindowExpr has the identical unclamped pattern (out-of-diff, pre-existing)

  ExpressionParser.parseExpr's finally now clamps the scope-stack restore (line 2200). But the identical save-raise-parse-restore-in-finally pattern in parseWindowExpr
  (ExpressionParser.java:565-567) was left unclamped:

  } finally {
      // Restore stack bottoms
      opStack.setBottom(savedOpStackBottom);                       // unclamped
      paramCountStack.setBottom(savedParamCountStackBottom);       // unclamped
      argStackDepthStack.setBottom(savedArgStackDepthStackBottom); // unclamped
  }

  On the error path, the inner parseExpr catch (lines 2191-2194) clears all three stacks to sizeRaw()==0; the finally then restores a saved bottom >0 → IllegalStateException: Tried to set 
  bottom beyond the top of the stack, masking the real positioned error.

  I reproduced this against clean HEAD (no agent involvement):

  This is pre-existing — parseWindowExpr is byte-identical on master, and the PR touches only line 2200 — so it is not a regression and does not block merge on correctness grounds. But:
  - It falls squarely within the bug class the PR's "Effects and tradeoffs" bullet claims to have eliminated: "A parse error nested two or more lambda levels deep now surfaces its positioned
  SqlException instead of an internal IllegalStateException." That statement is false for this reachable input.
  - The completion is the identical 3-line pattern and is safe by the same argument (no-op on the happy path where stacks rebalance to ≥ saved; clamps to 0 on the cleared error path):

  opStack.setBottom(Math.min(savedOpStackBottom, opStack.sizeRaw()));
  paramCountStack.setBottom(Math.min(savedParamCountStackBottom, paramCountStack.sizeRaw()));
  argStackDepthStack.setBottom(Math.min(savedArgStackDepthStackBottom, argStackDepthStack.sizeRaw()));

  Recommendation: either complete the fix here (+ a regression test for the window-frame-bound-in-lambda shape) so the PR's claim holds, or narrow the "Effects" bullet to the scope-stack
  path and file a follow-up. I'd take the former — it's three lines of an already-proven pattern. I can apply it if you want.

  Minor

  - m1 — Test comment mislabels the retained memory (JoinTest.java, the noMemoryUsageCheck() comments ~lines 2179/2186/2214). The comment calls it ">64 KiB of legitimate map RSS". Verified
  by memory-tag probe: the retained ~128 KiB is overwhelmingly the hash join's LongChain slaveChain (NATIVE_DEFAULT), eagerly allocated at the 128 KiB default page size and freed at factory
  close; NATIVE_UNORDERED_MAP is only ~1.5 KiB and NATIVE_JOIN_MAP is excluded from the check. The substance is right (legitimate, owned, freed at factory close — not a leak; the outer
  assertMemoryLeak remains the authoritative zero-tolerance gate), and noMemoryUsageCheck() is used correctly. Just reword "map RSS" → "hash-join slave-chain RSS".
  - m2 — Join-type coverage is partial but redundant (JoinTest.testInSubQueryWithJoinOnClause). The test exercises HORIZON, ASOF, INNER; LEFT/RIGHT/FULL OUTER, LT, SPLICE, WINDOW, and
  LATERAL all fall through into the same shared ON-drain block in parseJoin, so they traverse byte-identical code — a true gap only in appearance. Adding a LEFT JOIN case (the most common
  after INNER) would cheaply broaden confidence. No NULL-symbol or empty-src boundary case, but those are immaterial here — the fix is a parse-time arg-stack/scope-stack correction that runs
  before any row is read.
  - m3 — PR title under-states scope (optional). "fix internal error when an IN subquery contains a join" is accurate for the headline symptom, but the fix also covers scalar operands (=, >)
  and the (partial) deep-parse-error path. Acceptable as a release-note line; if you want it broader, "...when a sub-query operand contains a join" covers the scalar cases too. Labels (Bug,
  SQL) are correct — do not add Core (parser-only). Fixes #NNN correctly absent (no public issue exists).

  Downgraded (false positives / cleared concerns)

  - "Test failure: JoinTest IllegalState at line 2246" — FALSE. My first test run hit a dirty tree: Agents 5/9/10 had concurrently reverted the line-2200 clamp to prove it was load-bearing.
  Re-ran on a clean tree: both PR tests pass 1/1, 0 failures.
  - Scope-stack clamp masks real corruption / changes happy-path behavior — refuted. Proven no-op on every non-error path (sizeRaw() ≥ savedScopeStackBottom); only diverges after the catch's
  clear(), where the result is discarded.
  - subQueryBlockDepth gets stuck >0 / goes negative / wrongly rejects a later query — refuted. Push/pop balanced on all paths (incl. non-SqlException throwables, since push is outside the
  try and pop in finally); guarded decrement survives the reset-then-pop ordering; SqlCompilerImpl.compile()→clear()→reset() zeroes it before every statement.
  - Floor raise breaks the top-level join ON parse — refuted. At top level argStack is empty, so the floor is 0 (no-op).
  - Concurrency / thread-safety of the new field — refuted. SqlParser/ExpressionTreeBuilder are single-thread-confined via the compiler pool's per-slot CAS; the field is a plain int on a
  confined object.
  - noMemoryUsageCheck() masks a real native leak — refuted. It suppresses only the per-cursor 64 KiB RSS-delta bound; the outer assertMemoryLeak enforces exact-equality native-leak checking
  over the whole test body and passes.
  - Performance regression — refuted. Every added op is O(1) on the compile path; not reachable from any per-row/page-frame/partition path.

  Summary

  Verdict: approve — the core fix is correct and the implementation is clean; M1 is a pre-existing sibling bug, not a regression, so it does not block merge, but I recommend completing it in
  this PR since it's the same proven 3-line pattern and the PR body currently claims it's already fixed.

  - Draft findings: ~14 raised across the agents → 5 verified, ~9 dropped as false positives/cleared concerns (listed above).
  - Verified findings: 1 Moderate, 3 Minor.
  - In-diff vs out-of-diff: 3 in-diff (m1 test comment, m2 test coverage, m3 title) / 1 out-of-diff (M1 parseWindowExpr, pre-existing).
  - Regressions/tradeoffs: none introduced. The one honest behavior change the PR already discloses — ON a = b inside such a sub-query now compiles instead of erroring with "Column name
  expected" — is genuinely new-capability, not a correctness regression.

bluestreak01 (changes requested):
@puzpuzpuz

**Critical**
C1. Declared subqueries bypass the new JOIN `ON` subquery block. In-diff/cross-callsite.

`parseJoin` rejects literal `SELECT` nodes during `parseExpr` via `ExpressionTreeBuilder.java:48`, but join criteria are then passed through `rewriteKnownStatements(...)` at `SqlParser.java:3521`. That later expands declared variables at `SqlParser.java:5474`, after `subQueryBlockDepth` is gone.

Verified repro on this PR branch: literal nested `ON ... IN (SELECT ...)` rejects with `query is not allowed here`, but the declared equivalent compiles:

```sql
DECLARE @q := (SELECT symbol FROM trades)
SELECT * FROM trades
WHERE symbol IN (
  SELECT s.symbol FROM src s JOIN ref r ON s.symbol IN @q
);
```

That violates the PR's stated contract that ON-clause subqueries remain unsupported at every nesting depth. Fix by rejecting `ExpressionNode.QUERY` after declaration rewrite for join criteria, or by making declaration rewrite context-aware for JOIN `ON`.

**Moderate**
M1. Unrelated repo-local Codex review tooling is included in a SQL bugfix PR. In-diff.

`.codex/skills/review-pr/SKILL.md` and `.codex/skills/review-pr/agents/openai.yaml` add a new `.codex` tree; `origin/master` has none, and the PR title/body/labels describe only the SQL parser bug. If QuestDB wants repo-owned AI review policy, this should be split into a tooling/process PR. Otherwise remove these files from this branch.

**Minor**
m1. Commit metadata violates the review convention: `1b8e1a5` uses a Conventional Commits prefix (`chore: add Codex review PR skill`), and the merge commit subject is over 50 chars.

**Downgraded**
Dropped parser-state and resource concerns after verification: `pushArgStackBottom/popArgStackBottom` are balanced on `SqlException` paths, the `parseWindowExpr` clamp is present with a regression test, LEFT/RIGHT/FULL joins share the same ON-drain arm covered by INNER, and the new operations are compile-time O(1). The `.noLeakCheck()` usage is inside a broader `assertMemoryLeak()` scope.

**Verification**
Ran targeted tests successfully:

```bash
mvn -Dtest=JoinTest#testInSubQueryWithJoinOnClause,SqlParserTest#testNestedSubQueryParseErrorIsPositioned+testNestedWindowFrameParseErrorIsPositioned -pl core test
```

Also ran a temporary `/tmp` JUnit reproducer for C1; it confirmed the declared-subquery bypass. Worktree remained clean.

**Summary**
Verdict: request changes.

Verified findings: 1 Critical, 1 Moderate, 1 Minor. Draft findings verified vs dropped: 3 verified, 6 dropped. Split: 2 in-diff findings plus 1 cross-callsite parser finding involving in-diff behavior and out-of-diff declaration rewrite.

puzpuzpuz:
Thanks for the careful review. Follow-up on each item.

## C1 (Critical) -- confirmed valid, fixed in 19a7479

I reproduced it and the mechanism is exactly as described: `@q` is a `LITERAL` at parse time, so it passes the parse-time ON-clause sub-query block in `ExpressionTreeBuilder.onNode`; the sub-query only materialises later when `rewriteKnownStatements` -> `rewriteDeclaredVariables` swaps in the declaration's `QUERY` node, after `subQueryBlockDepth` is gone.

Critically assessing severity, I ran the compiled query rather than stopping at "it compiles". The declared form degrades to a Cartesian product filtered by the sub-query -- the exact "cross-join semantics" footgun the literal rejection exists to prevent:

```
declare @q := (select symbol from trades)
select s.symbol, r.symbol from src s join ref r on s.symbol in @q

plan:   Cross Join
            filter: symbol in cursor
result: (A,A) (A,B) (B,A) (B,B)   -- not the equi-join (A,A)(B,B)
```

One nuance worth recording: this bypass is **not new to the nesting fix** -- the *top-level* declared form (`... join ref r on s.symbol in @q`, no enclosing `IN`) already compiled to the same cross join on master. What this PR changed is the *nested* case: on master it errored with the misleading "Column name expected" (the original arg-stack drain consumed the operand), and the floor raise turned that accidental rejection into a clean compile. So the fix needs to close both the nested case and the pre-existing top-level one to make the contract actually hold.

Fix: `parseJoin` now walks the rewritten ON criteria with `traversalAlgo.traverse` and rejects any `ExpressionNode.QUERY` with `query is not allowed here` -- the first of your two suggested options. The declared form now errors identically to the literal one at every nesting depth (nested, top-level, and scalar `=` operand). Column-valued and scalar declared variables in `ON` still compile unchanged; the reject fires only on sub-query nodes.

Regression coverage in `JoinTest#testJoinOnClauseRejectsDeclaredSubQuery`: nested `IN @q`, top-level `IN @q`, a scalar `= @q`, a valid `@x = @y` column-valued declared join that must still run, and a pooled-compiler reuse case proving shared parser state stays clean after a rejection. SqlParserTest (1081), DeclareTest (77), JoinTest (185), HorizonJoinTest (165), WindowJoinTest (368), WhereClauseParserTest (846), and ExpressionParserTest (295) are all green.

## M1 (Moderate) -- agree

The `.codex/skills/review-pr/` tree is unrelated to this parser fix, but let's keep it.

## m1 (Minor) -- noted

The branch history is throwaway under our squash-merge flow (only the final squashed subject lands on master), so the Conventional-Commits prefix on `1b8e1a5` and the merge-commit subject length won't survive the merge.

bluestreak01 (approved):
Review: PR #7211 — fix(sql): fix internal error when an IN subquery contains a join

Level 3 (full pass: change-surface map, 4 fresh-context parallel agents — correctness+cross-context, fresh-adversarial, adversarial-performance, tests; Rust agent skipped, no `.rs` files — plus per-finding source verification). Reviewed against PR head `b2806ff66`.

## Verdict: approve

The core fix is correct, complete, thread-safe, perf-neutral on every data path, and regresses no previously-correct query. I independently verified the central regression risk — that blocking ON-clause sub-queries could break a working query — and it holds: top-level ON-clause sub-queries were always rejected (`model == null`), and the nested form was always broken (NPE / "Column name expected"), so no correct result changes. Both prior reviews' substantive findings are **resolved** in the current branch:

- **C1** (declared `@q := (SELECT...)` bypassing the ON-clause block) — fixed via up-front `rewriteKnownStatements` + a deep `traversalAlgo.traverse(rejectJoinSubQueryRef)` in the criteria branch + a per-element QUERY check in the multi-column branch. I traced every sink (`addJoinColumn`, `setJoinCriteria`) and confirmed no QUERY node can reach compilation at any nesting depth, including chained/OVERRIDABLE/UNION-propagated declarations (declarations are stored pre-expanded, so a single rewrite pass fully resolves them).
- **M1** (`parseWindowExpr` unclamped) — both `parseExpr` (line 2203) and `parseWindowExpr` (lines 568-570) clamps are present, each with a dedicated positioned-error regression test.

## Critical
None.

## Moderate

**M1 — Unrelated review tooling bundled into a SQL bugfix PR (in-diff).** `.pi/skills/review-pr/SKILL.md` (485 lines = 54% of the +890 diff) is added by this PR and is byte-identical to the harness's active review skill; `origin/master` already carries the equivalent `.codex/skills/review-pr/` tree. The PR title/body/labels describe only the parser bug. This belongs in a separate tooling/process PR. Already raised earlier; the author chose to keep it. Flagged for the record — scope/process concern, not a correctness issue.

## Minor

**m1 — Multi-column shorthand rejects nested sub-queries with the wrong message; the code comment overstates consistency (in-diff).** `SqlParser.java:3539-3545` (the `default` / `ON (a, b, ...)` arm) only checks the **root** node type:

```java
expr = rewriteKnownStatements(expr, decls, null);
if (expr.type == ExpressionNode.QUERY) { throw ... "query is not allowed here"; }
if (expr.type != ExpressionNode.LITERAL) { throw ... "Column name expected"; }
```

A declared variable bound to an expression whose sub-query is **not at the root** — e.g. `@q := symbol + (SELECT count(*) FROM trades)` used as `ON (@q, ts)` — expands to an `OPERATION(+, symbol, QUERY)`, so the root check misses it and it falls through to `"Column name expected"` rather than `"query is not allowed here"`. No sub-query slips through (a LITERAL is atomic and every non-LITERAL is rejected), so this is message-quality only. But the comment on `rejectJoinSubQuery` / the `default` arm — and the PR body — claim the declared form "errors the same as the literal one ... in every ON-clause position -- criteria, single-column shorthand, and multi-column lists alike," which is inaccurate. Found independently by two agents. Fix: either run `traversalAlgo.traverse(expr, rejectJoinSubQueryRef)` in the `default` arm too, or soften the comment/PR wording.

**m2 — `rejectJoinSubQueryRef` does not walk `windowExpression` sub-trees; currently masked, defense-in-depth (in-diff + pre-existing).** `PostOrderTreeTraversalAlgo.traverse` descends only `lhs`/`rhs`/`args`, not `node.windowExpression`, whereas `recursiveReplace` (used by `rewriteDeclaredVariables`) *does* descend into partition-by/order-by/frame bounds. So a declared sub-query nested inside `OVER (PARTITION BY @q ...)` in an ON criterion expands to a QUERY the reject never visits. I confirmed this is **unreachable with harm**: any such criterion contains a window function, and `SqlOptimiser.validateNoWindowFunctionsInWhereClauses` (line 11277) rejects it — `checkForWindowFunctionsOrNames` returns `true` the moment any `node.windowExpression != null`, so "window function is not allowed in JOIN ON clause" fires before the QUERY can affect results. Also a pre-existing asymmetry shared by every `rewriteKnownStatements` visitor. Optional hardening only.

**m3 — Redundant compile-time tree passes (in-diff, low severity).** The PR makes shorthand `ON (col)` / `ON (a, b, c)` run the full 6-pass `rewriteKnownStatements` (5 of 6 are no-ops on a column LITERAL) where master ran **zero** passes — and the multi-column arm runs it once per column. The criteria branch also adds an unconditional `traversalAlgo.traverse(rejectJoinSubQueryRef)` pass that is pure waste whenever no `DECLARE` is in scope, because directly-written ON sub-queries are already blocked at parse time by `subQueryBlockDepth`. All compile-time, on single-digit-node trees → negligible wall-clock, but genuinely redundant. Cheap wins: guard the reject traverse with `decls != null && decls.size() > 0`, and expand declared vars first then dispatch so the no-`DECLARE` shorthand path stays at 0 passes.

**m4 — Declared sub-query rejects report the DECLARE-site position, not the ON-clause use site (in-diff).** The QUERY node carries its parse position from `DECLARE`, so `ON s.symbol IN @q` errors at the `(SELECT...)` inside the declaration (e.g. position 15/53), not at `@q`. The tests pin this. It points at the actual offending sub-query and is consistent with other declared-variable diagnostics, but per the "point at the offending character" convention it's slightly surprising. Very minor.

**m5 — Join-type test breadth (in-diff).** Tests cover INNER, ASOF, HORIZON. LEFT/RIGHT/FULL OUTER, LT, SPLICE, WINDOW, LATERAL all share the same ON-drain block in `parseJoin`, so they traverse byte-identical code — not a true gap. A single LEFT OUTER case would cheaply broaden confidence. Optional.

## Downgraded (false positives / cleared after verification)
- **C1 declared-subquery bypass** — RESOLVED; the deep traverse + per-element check + up-front rewrite close it at every depth. Only the unreachable windowExpression asymmetry (m2) remains.
- **parseWindowExpr unclamped** — RESOLVED; both clamps present with regression tests.
- **Floor raise breaks top-level join ON** — refuted: at top level `argStack` is empty so the floor is 0 (no-op).
- **`subQueryBlockDepth` stuck >0 / negative / leaks across compiles** — refuted: balanced on happy, SqlException (reset()-then-guarded-pop), and non-SqlException paths; zeroed by `reset()`/`clear()` before every compile; the two pooled-compiler-reuse tests prove it.
- **Shared `argStackBottomStack` interleaves with pushModel/popModel** — refuted: the ON-clause parse blocks sub-queries, so no `pushModel` can intervene between push/popArgStackBottom; strict LIFO preserved.
- **Clamp masks real corruption on the happy path** — refuted: proven no-op when `sizeRaw() >= saved`; only diverges after the catch's `clear()`, where the stack is discarded.
- **`traversalAlgo` corruption when the reject visitor throws** — refuted: `traverse` wraps the walk in `backup()`/`restore()` with try/finally.
- **Concurrency / thread-safety of the new field** — refuted: `SqlParser`/`ExpressionTreeBuilder` are single-thread-confined per pooled compiler; `subQueryBlockDepth` is a plain `int` on a confined object.
- **`noMemoryUsageCheck()`/`noLeakCheck()` masks a native leak** — refuted: those suppress only the per-cursor 64 KiB RSS bound (hash-join slave chain held until factory close); the outer `assertMemoryLeak` remains the authoritative gate. An agent ran all 5 tests — green.
- **Performance regression on a data path** — refuted: every added op is O(1) and strictly compile-time.

## Summary
- **Verdict: approve.** Correct, complete, regression-free core fix; clean implementation.
- **Findings:** 0 Critical, 1 Moderate (scope/tooling, maintainer-accepted), 5 Minor.
- **Draft → verified:** ~18 draft findings/concerns across 4 agents → 6 reported (1 Moderate + 5 Minor), ~12 dropped as false positives/cleared.
- **In-diff vs out-of-diff:** all 6 reported findings touch in-diff code (m2 also reflects a pre-existing `traverse`/`recursiveReplace` asymmetry). No confirmed out-of-diff bug — the cross-context pass walked every callsite of the changed symbols (`pushArgStackBottom`/`popArgStackBottom` are package-private, called only from `parseJoin`; `onNode`'s new arm is inert outside ON clauses) and found none broken.
- **Regressions/tradeoffs:** none that change a correct result.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 26 / 26 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlParser.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/ExpressionTreeBuilder.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/griffin/ExpressionParser.java|4|4|100.00%|

PR #7212, opened 2026-06-08
fix(core): fix all-null Parquet column reads and crashes on malformed files

glasstiger:
## What and why

Reading a Parquet partition could fail when an integer-family column had a `DELTA_BINARY_PACKED` data page that contained only nulls — a run of nulls long enough to fill a whole page. The read failed with:

```
could not decode page for column "<name>" in row group 0:
delta binary packed block size must be greater than zero
```

The failure shows up on an *eager*, full-row-group read — converting the partition back to native, or WAL apply re-running that conversion under replication — which decodes every page up front. A lazy `SELECT` over the same partition does not hit it: it fills nulls from the page's definition levels and never constructs the value decoder for an all-null page. Under replication the eager read runs during WAL apply, suspending the table on both primary and replica. The trigger is data-dependent — only a page with zero non-null values hits it — so it surfaced as an intermittent fuzz failure rather than a consistent one.

Note on what does *not* trigger it: a column added by `ALTER TABLE ADD COLUMN` is a column top in older partitions (no data file there). A column top is written all-null straight from its definition levels, without going through the delta value encoder, so it produces a valid page and is unaffected. The bug needs a real, delta-encoded data page whose values are all null.

The bug is specific to the integer-family delta writer. Varchar, string, and binary columns also delta-encode their lengths, but their writers always emit a self-describing `value_count = 0` header for an all-null page (never an empty buffer), so those pages already decoded correctly. This was verified by reverting both fixes: the all-null `long`/`int` round-trips fail with the original error, while the all-null `varchar`/`string`/`binary` round-trips pass unchanged.

## Root cause

For a page with zero non-null values the integer writer (`SimdEncodable::encode_data`) short-circuited and emitted an empty values buffer with no `DELTA_BINARY_PACKED` header. The reader (`MiniblockIterator::try_new`) then decoded `block_size = 0` from that empty buffer and rejected it. PLAIN-encoded all-null pages were unaffected, because the plain decoder reads lazily and tolerates an empty buffer. The varlen writers never produced an empty buffer, so they were never affected.

## Fix

- Reader (`MiniblockIterator::try_new`): treat an empty values buffer as a zero-value page and return an iterator that yields no values, instead of erroring. Definition levels drive null filling, so the value decoder is never asked to produce a value. This keeps Parquet files already written by the affected versions readable. The varchar length decoder reuses `MiniblockIterator`, so it inherits this handling defensively, though no varlen writer in this codebase produces the empty shape.
- Writer (`SimdEncodable::encode_data`): emit a self-describing DELTA header (`value_count = 0`) for all-null integer pages instead of nothing, so newly written files are valid Parquet — matching what the varlen writers already did. PLAIN keeps its empty-buffer fast path.

## Tradeoffs and risks

- The reader now accepts an empty DELTA values buffer (a zero-value page) where it previously errored. A genuinely truncated or corrupt page that nonetheless claims non-null values via its definition levels is still rejected: a zero-value page (an empty buffer or a `value_count = 0` header) fails on the first value request with "not enough values to iterate" rather than silently decoding a phantom 0. All-null pages are unaffected — their definition levels drive only null filling, never a value request.
- The integer writer's byte output for all-null delta pages changes — it now carries a header. This is a forward change only; readers in this codebase could not read the previous empty output anyway, and the new output is standard Parquet.
- No format or version bump: both the old (empty) and new (header) shapes of an all-null delta page are now readable.

## Tests

- Reader unit tests: an empty buffer yields zero values, an all-null page decodes as nulls, and a value request on a zero-value page (empty buffer or `value_count = 0` header) fails on the first request with "not enough values to iterate" (`delta_binary_packed.rs`).
- Varchar reader unit tests: both an empty buffer and a real `value_count = 0` header decode as nulls (the two distinct shapes), and requesting a value errors (`delta_length_array.rs`).
- Writer unit test: an all-null delta page carries a header, while PLAIN stays empty (`numeric.rs`).
- Integration round-trips through QuestDB's own writer and reader for all-null `long`/`int` (delta) and `long` (plain) columns (`encode_primitives.rs`), and for all-null `varchar`/`string`/`binary` columns (`encode_varchar.rs`, `encode_strings.rs`, `encode_binary.rs`). A VarcharSlice all-null delta decode (`decode_varchar_slice.rs`) covers the `DeltaLAVarcharSliceDecoder` read path, which previously had all-null coverage only for Plain and RleDictionary. The existing suite did not cover any of these: its decode tests write with the reference `parquet` crate, and its null patterns never reached 100%.
- Verified both ways: with both fixes reverted the integer integration tests fail with the original error and the varlen ones still pass; with the fixes applied the full `qdbr` suite passes, and `cargo fmt`/`clippy` are clean.

The deterministic Rust round-trips are the regression guard for the codec itself. The failure originally surfaced one layer up, at replication / WAL apply (`ReplicationFuzzTest` "table is suspended"); a companion deterministic guard at that layer — converting a partition with an all-null delta page to Parquet and back to native, under replication, and asserting the table is not suspended — is added separately in the QuestDB Enterprise repository.

---

## Additional hardening: panics on malformed or foreign Parquet

Beyond the all-null fix above, this branch closes a set of reachable panics in the same Parquet read path. They are reachable from the public `read_parquet()` SQL function and from partition conversion on malformed or foreign-produced pages. A Rust panic crosses the JNI boundary as a JVM abort with no recovery, so each was a crash on untrusted input rather than a recoverable error. None are regressions — they predate this branch — but they sit in the decode path this PR already touches.

The decoders and the vendored `parquet2` now reject these inputs with a clean error (surfaced as a `CairoException`):

- DELTA_LENGTH_BYTE_ARRAY varchar pages passed a per-miniblock bit width straight to the 32-bit unpacker, which hit `unreachable!` for any width above 32. The varchar decoder now rejects widths above 32, matching the guard the integer decoder already had, and replaces an unchecked pointer-slice with a bounds-checked one.
- DELTA_BINARY_PACKED headers: `parquet2`'s decoder asserted that the block size is a multiple of 128 and the values-per-miniblock a multiple of 8, and divided by the miniblock count without checking it for zero. A foreign header could trip any of these — an assertion failure or a divide-by-zero. They now return errors. The same change covers the DELTA_BYTE_ARRAY path, including the second (suffix) decoder reached when the suffix section is empty.
- DELTA_BINARY_PACKED miniblocks: `parquet2`'s `Block::advance_miniblock` read each per-miniblock bit width from the page (range 0..=255) and passed it to the 64-bit unpacker, which hit `unreachable!` for any width above 64. It also sized the miniblock with an unchecked `values_per_mini_block * num_bits` multiply, which a 2^63 block size — still a multiple of 128, so it clears the header guard — overflows: in release (overflow checks off) it wraps to a short miniblock and then panics on an `unwrap()`, in debug it panics on the multiply. `advance_miniblock` now rejects bit widths above 64, uses a checked multiply, and propagates the bitpacked decoder error with `?`. The byte-array slicers reach this `Block` for foreign String, Varchar, VarcharSlice, Binary and Array DELTA pages; the earlier varchar bit-width guard covered only the `MiniblockIterator` path, not this vendored decoder.
- ULEB128 varint decoding panicked on a value wider than 64 bits; it now returns an error. `u64::MAX` still decodes.

`parquet2` is vendored in-tree as a QuestDB-maintained fork, so these fixes sit at the source: every caller — the slicers, which already propagate with `?` — gets the clean error without changes. The all-null empty-buffer guards in the slicers stay, since they carry separate semantics.

### Tradeoffs

- The vendored `parquet2` diverges a little further from upstream: in the delta decoder a `panic!` and several `assert!`/`unwrap()` calls become `Result` errors, two unchecked multiplies become checked, and a per-miniblock bit-width bound is added. The behaviour is unchanged for valid input — only previously-panicking malformed input is affected, and only to turn a JVM abort into a returned error.
- No format or version change. Valid files decode exactly as before.

### Tests

- `parquet2` unit tests for each malformed header (block size, miniblock count, values-per-miniblock) and for the ULEB128 overflow.
- Decoder and slicer tests that assert an error rather than a panic for each case, across the `push`, `push_slice` and `skip` paths, plus a bit-width-of-32 boundary case that must still decode.
- Coverage for two pre-existing but untested bounds guards in the delta-length slicer (`skip` and `next_into`).
- `parquet2` and slicer tests for a per-miniblock bit width above 64 and an oversized (2^63) block size, a bit-width-of-64 boundary that must still decode, a bad bit width on the second miniblock (the iteration path), and a bitpacked `length * num_bits` overflow. The slicer tests drive both malformed shapes through `DeltaLengthArraySlicer` and `DeltaBytesArraySlicer`, including the latter's suffix decoder (a valid prefix with a malformed suffix).
- Each guard was checked both ways: reverting the guard makes the test panic; restoring it makes the test return an error.

## Minor cleanups

- The all-null DELTA arm of the integer writer (`encode_data`) no longer scans the whole page through a non-null filter when there are no non-null values; it emits the same `value_count = 0` header from an empty iterator. The output is byte-identical and the all-null write path drops from O(n) to O(1). PLAIN already had this guard. This is a write-time, per-page cost off the read hot path.
- Removed `DeltaBinaryPackedSlicer`, dead code with no construction site anywhere in the crate. It predates this work (introduced with `read_parquet`) and never received the empty-buffer/all-null handling its sibling slicers have, so it was a latent hazard if wired up later. The surviving slicers still use the same `parquet2` decoder and the `DataPageSlicer` trait.

## Test infrastructure

- Deflaked the three file-descriptor ownership tests in `parquet_write::jni`. They probed `fd_is_open(number)` after closing a descriptor, but descriptor numbers are process-global and the suite runs in parallel, so a concurrent test could reopen that number between the close and the probe. That produced intermittent failures in the Rust CI job, unrelated to the codec logic. The tests now detect closure through a pipe's EOF, which the kernel tracks by reference count rather than by descriptor number, so descriptor reuse cannot fool the check. Under the descriptor churn that made the old approach fail, the new one saw no spurious failures over 50,000 iterations.

## Tandem

required by https://github.com/questdb/questdb-enterprise/pull/1050


bluestreak01:
@glasstiger — code review of this PR (level-3 pass: correctness, concurrency, performance, resource safety, Rust panic/UB, tests, cross-context callers, plus adversarial passes; every finding verified against source).

**Verdict: approve with minor changes.** The core fix is correct, well-scoped, and memory-safe. Valid data round-trips in every direction I traced — no data loss, no UB, no leaks, no panic-via-JNI on valid input, no concurrency hazard, and no cross-version/replication hazard. The items below are a corruption-detection nuance, two misleading comments, an inaccurate description claim, and some test-coverage gaps.

Mechanics I verified independently: parquet2's `encode` writes a 5-byte header `[0x80,0x01,0x01,0x00,0x00]` for empty input (writer half produces a valid page); `validate_encoding` keeps floats out of the DELTA arm; `ExactSizedIter` can't underflow on the empty filter; and `split_buffer` yields an empty `values_buffer` for a real all-null V2 page (reader half is genuinely reachable).

## Critical
None.

## Moderate

### M1 — Empty-buffer tolerance silently decodes a corrupt page's lone non-null value as `0`; the "failure moves, it is not lost" claim is inaccurate
In-diff code (`delta_binary_packed.rs:47-67`), manifesting at the unchanged `push`/`push_slice` (same file, 409-419 / 479-487), driven by `decode.rs:1557`.

The degenerate iterator has `current_value = first_value = 0` and `consumed_initial_value = false`. On a page whose def-levels claim a non-null value, `decode_page0` dispatches `push_slice(1)`/`push()`. The **first** call writes the phantom `current_value` (=`0`) and returns `Ok` *without consulting the iterator*; only the **second** value reaches `unpack_next → next_miniblock → None → Err("not enough values to iterate")`.

So a corrupt/truncated page claiming **exactly one** non-null value silently decodes it as `0` (two or more still error). For LONG/INT/TIMESTAMP/DATE, `0` is a plausible value, not a sentinel.

This is reachable **only on a corrupt/foreign malformed page** — a legitimate all-null page has zero non-null def-levels and only ever drives `push_nulls`, so the masking cannot occur on valid data. It is a narrow reduction in corruption-detection robustness, not a data-integrity regression. But the "Tradeoffs and risks" claim *"it fails when a value is actually requested … the failure moves, it is not lost"* is not accurate — the first requested value does not fail; it returns `0`.

Suggested fix: when `try_new` took the empty-buffer path, set a flag so `push`/`push_slice`/`skip` return `"not enough values to iterate"` before consuming the phantom first value (all-null pages are unaffected — they only call `push_nulls`). At minimum, correct the description and in-code comment to say the first value decodes as `0`.

### M2 — Self-contradictory comment on the empty-buffer branch
`delta_binary_packed.rs:48-50` says the empty buffer is emitted by *"QuestDB's own writer for column-top integer columns."* This same PR's writer change (`numeric.rs:414-428`) makes the current writer emit a 5-byte header instead — so the current writer does **not** produce an empty buffer. The branch is reachable only for pre-fix QuestDB files and foreign encoders. Reword to attribute the empty buffer to pre-fix/foreign writers and note the current writer emits a header.

### M3 — Neither fix half is independently guarded end-to-end; the reader's backward-compat path never runs through `decode.rs`
Revert experiments confirm the three `encode_primitives.rs` integration tests pass with the **reader** fix reverted (new header read via the normal path) *and* with the **writer** fix reverted (empty buffer read via the new reader branch); they fail only when both are reverted. So the suite is an OR of the two halves.

The reader's empty-buffer branch — the backward-compat guarantee for reading old/foreign files — is exercised only by iterator-level unit tests, never through the real `decode.rs` decode path (every integration test goes through the new header-emitting writer). Add a decode-path test that feeds an empty DELTA `values_buffer` through `decode_page0` and asserts all-null output.

### M4 — Missing all-null round-trip coverage for distinct-sentinel types and the varchar writer→reader path
Integration round-trips cover only INT/LONG. The reader dispatches DELTA pages for many more types with **distinct null sentinels / widths**, none round-tripped: Decimal64 (`nulls::DECIMAL64`), GeoByte/Short/Int/Long (`nulls::GEOHASH_*`), IPv4, Byte, Short/Char. (Agent verification confirmed the sentinels passed at each `decode.rs` callsite are *correct* — this is a coverage gap, not a live bug.)

Most importantly, there is **no end-to-end test writing an all-null VARCHAR/STRING column through QuestDB's writer and reading it back via QuestDB's reader.** VARCHAR/STRING default to delta-length, so all-null column-top varchar is a common path; the varchar change is unit-only with a hand-built `&[]` the writer never produces.

There is also no **Java end-to-end** test of the user scenario: `ALTER TABLE ADD COLUMN` → entirely-null old partition → `CONVERT PARTITION TO PARQUET` with DELTA → query. Existing tests use PLAIN (never broken) or 50%-null data.

## Minor
- **m1** — `get_end_pointer` comment (`delta_binary_packed.rs:140`) names the wrong trigger: the normal `value_count<=1` path sets `miniblock_index = miniblocks_per_block` (>0) and does **not** hit the early-return guard; only the empty-buffer synthetic state does.
- **m2** — Description overstates scope: *"affects … varchar columns."* The varchar/string/binary delta-length writers (`varlen.rs:466/487/629/913`) call `delta_bitpacked::encode` unconditionally and always write a header — they never produced empty buffers in any version. The bug was integer-only (the SIMD `encode_data` short-circuit was the sole writer that emitted empty DELTA buffers). Varchar reader coverage is incidental robustness, not a QuestDB-triggerable path.
- **m3** — `round_trip_all_null_long` names a LONG column `"str_top"` (`encode_primitives.rs:451`); the INT sibling uses `"int_top"`. Rename to `"long_top"`.
- **m4** — `encode_data_delta_all_null_writes_header` (`numeric.rs:1005`) asserts only `!buf.is_empty()`; assert the exact 5 header bytes (`[128,1,1,0,0]`) so it guards header content.
- **m5** — Wasted O(n) slice scan for a *materialized* all-null page (`numeric.rs:454/486` via parquet2 `encoder.rs:44`): the filter iterator scans the whole slice to discover nothing even though `non_null_count==0` is known. Low impact — the real column-top case passes an empty slice and scans nothing; only a physically-materialized all-null page pays it. Optional: early-return the constant header.
- **m6** — `Fixes #NNN` absent. If a tracking issue exists, add it; if found purely by fuzzing, this is fine.

## Downgraded (verified false positives / out of scope)
- "All-null f32/f64 + DELTA now errors" — unreachable; `validate_encoding` downgrades floats to Plain.
- "Unsupported-encoding arm now errors for all-null" — unreachable; static dispatch only routes Plain/DELTA to `encode_data`.
- Cross-version/replication hazard — none; OSS replication ships row data + SQL, not Parquet bytes (each node re-encodes locally), and new header pages are readable by old readers via the existing `value_count<=1` path.
- Concurrency — none; per-thread owned state only (owned `Vec<u8>`; exclusive `&mut ColumnChunkBuffers`), no statics, Java callers each own their `RowGroupBuffers`.
- Resource leak / OOB write / `data_len` underflow / panic-via-JNI — all SAFE (`reserve` precedes `push_nulls` and refreshes the pointer; `get_end_pointer` provably stays within `data`; `unpack_next` errors before any raw read; corroborated under Miri).
- parquet2 native delta decoder div-by-zero on a truly empty String/Binary/Array buffer — unreachable from QuestDB files (those writers always emit a header); pre-existing, separate hardening if untrusted Parquet ingestion is a concern.

## Summary
Approve with minor changes. Both halves are justified — the writer for spec-compliance/forward-compat, the reader for backward-compat with pre-fix integer files. No regressions on valid data; M1 affects only already-corrupt/foreign pages. Most actionable: fix the two misleading comments (M2, m1) and the inaccurate description claims (M1's "failure moves" wording, m2's varchar scope); optionally harden the degenerate decoder (M1); and close the test gaps (M3, M4) — especially an end-to-end all-null **varchar** round-trip and a Java reproduction, since the suite currently guards neither fix half independently.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) — automated review, please verify findings before acting.</sub>

bluestreak01 (dismissed):

bluestreak01:
Code review of this PR — **level-3 pass** (correctness, concurrency, performance, resource safety, Rust panic/UB across JNI, tests, and cross-context callers, plus fresh-context adversarial and panic-audit passes; every finding verified against source, and the headline finding empirically reproduced).

**Verdict: approve.** The in-scope change is correct, well-tested, and a real safety improvement — no correctness regression, leak, UB, or new panic on valid data, and the DELTA-path malformed-input hardening is complete and well-defended. There is one high-value pre-existing item (M1) of the *same class* on a sibling path that the "crashes on malformed files" theme implies but does not close, plus some test-coverage gaps.

Mechanics I verified independently: the `advance_miniblock` `bitwidths.next().unwrap()` is provably non-None for any foreign header (call count ≤ `num_mini_blocks` via the `length = min(length, M*V)` clamp); the writer's `encode(empty)` fast path is byte-identical to the filtered-empty iterator for both i64 and i32; `validate_encoding` keeps floats out of the DELTA arm so the float→`Err` path is unreachable; the empty-buffer/`value_count=0` reader branch is genuinely reachable; and the new boundary tests pin the guards at `> N` (not `>= N`).

## Critical
None. The PR introduces no correctness regression, leak, UB, or new panic-via-JNI on valid data.

## Moderate

### M1 (out-of-diff, pre-existing) — the "crashes on malformed files" hardening is incomplete: an equally-reachable sibling JVM-abort panic remains on the RLE_DICTIONARY index path

Not a regression — it predates this branch — but it is the **same bug class** this PR's hardening section closes (an unguarded, page-controlled bit width reaching `unpack`'s `unreachable!()`), on a sibling path the diff leaves untouched. Confirmed by source trace and reproduced at runtime: `internal error: entered unreachable code: invalid num_bits 40` at `parquet2/src/encoding/bitpacked/unpack.rs:96`.

Code path:
- A foreign `RLE_DICTIONARY` / `PlainDictionary` data page's **first byte is the index bit width** (range `0..=255`, attacker-controlled). It is read with only a `> 0` check:
  - `RleDictionaryDecoder::try_new` — `core/rust/qdbr/src/parquet_read/decoders/rle_dictionary.rs:281`
  - `RleDictionarySlicer` (`SlicerInner`) — `core/rust/qdbr/src/parquet_read/slicer/rle.rs:142`
  - `RleDictVarcharSliceDecoder::try_new` — `core/rust/qdbr/src/parquet_read/decoders/rle_dict_varchar_slice.rs:58`
- It then flows **unchecked** into the u32 unpacker (only `num_bits == 8` is fast-pathed):
  - `rle_dictionary.rs:46` and `slicer/rle.rs:33`: `bitpacked::Decoder::<u32>::try_new(values, num_bits, count)?`
- `bitpacked::Decoder::try_new` guards only `num_bits == 0` (and now overflow), **not** `num_bits <= 32`, then eagerly `decode_pack` → `unpack32(.., num_bits)` → `unreachable!()` for `num_bits` in `33..=255` → **panic → JVM abort** with no recovery.
- Reachable from public `read_parquet()` via ~15 `RleDictionaryDecoder::try_new` dispatch sites in `decode.rs`. `RLE_DICTIONARY` is the default encoding Arrow/pandas/Spark emit for low-cardinality columns — a maximally common, trivially-crafted foreign page. (Definition/repetition levels are safe: their width comes from `get_bit_width(i16) <= 15`, schema-bounded.)

Why raise it here: the hardening section states it "closes a set of reachable panics in the same Parquet read path … reachable from the public `read_parquet()` SQL function … on malformed or foreign-produced pages." This one satisfies that exact description but is not closed. Given it's a small fix consistent with the PR's own pattern, folding it in seems better than deferring.

Suggested fix (mirrors the delta-path pattern, same vendored fork) — a single guard inside `bitpacked::Decoder::<T>::try_new`:

```rust
if num_bits > size_of::<T>() * 8 {
    return Err(Error::oos(format!("bitpacking num_bits {num_bits} exceeds target width")));
}
```

This covers the u32 dictionary path, defensively re-covers the u64 delta path, and protects every future caller from one point. (Alternatively, explicit `num_bits > 32` guards at the three RLE callsites.) Add a slicer-level test feeding a dictionary page whose leading width byte is e.g. `40`.

## Minor

- **m1** — `skip()` on a zero-value DELTA page is correct but untested. `DeltaBinaryPackedDecoder::skip` (`delta_binary_packed.rs:587`) and `DeltaLAVarcharSliceDecoder::skip` both correctly route a phantom value to "not enough values to iterate" via `consumed_initial_value=true`, but only `push`/`push_slice` are tested. Since `skip` writes no output, a regression that silently advanced past the phantom would go undetected. Add a `skip(1)`-errors test for both.
- **m2** — the integer empty-page error tests assert the message but not that the output buffer stayed untouched. `empty_page_value_request_errors` / `value_count_zero_header_value_request_errors` don't check `data_vec` is still empty, unlike the varlen `test_empty_page_value_request_errors` push() case (which asserts `read_aux_entries().len() == 0`). Add the symmetric check.
- **m3** — `encode_data_delta_all_null_writes_header` (`numeric.rs:1004`) asserts only `!is_empty()`. It does catch the regression on revert, but a prior request for the exact 5 header bytes `[0x80,0x01,0x01,0x00,0x00]` is still unmet at the unit level (round-trips cover it indirectly). The i32 fast path is also unit-untested (only via the `int` integration round-trip).
- **m4** — the new `.get(offset..offset+pack_size)` bound that replaced `from_raw_parts` in `DeltaLAVarcharSliceDecoder::unpack_next` ("DELTA_LENGTH_BYTE_ARRAY pack exceeds miniblock size") is unreachable-by-invariant for `num_bits <= 32` (acceptable), but it's a new branch with no test and no `debug_assert!`/`expect()` marking it unreachable.
- **m5** (informational, pre-existing, not a crash) — `DeltaBytesArraySlicer::next` `last_value.truncate(prefix_len)` silently no-ops when a corrupt foreign `prefix_len` exceeds the previous value's length, yielding a wrong-but-in-bounds string. Matches upstream parquet2; no panic/UB.
- **m6** — `Fixes #NNN` absent. If a tracking issue exists (the description mentions `ReplicationFuzzTest`), add it; if found purely by fuzzing, this is fine.

## What I verified correct (cleared / downgraded)

The in-scope PR is sound and complete for the DELTA path:

- **All-null fix / phantom-0:** `has_values` + `consumed_initial_value` correctly make a zero-value page (empty buffer or `value_count=0` header) error on the first value request rather than decode a phantom `0`, across push/push_slice/skip, in both the integer and varchar decoders. All-null pages (push_nulls only) are unaffected.
- **DELTA-path panic hardening, all present and off-by-one-correct:** `advance_miniblock` `num_bits > 64` + `checked_mul` (both overflow sites) + `.unwrap()`→`?`; `Decoder::try_new` three asserts/division → `Err`; `uleb128` panic → `Err`; qdbr `unpack_next` `num_bits > 32` (varchar) / `> max_num_bits` (integer); `get_end_pointer` fully checked. Boundary tests defend `> N` vs `>= N`.
- **Slicers:** empty-buffer early returns + bounds-checked length/slice access in `DeltaLengthArraySlicer`/`DeltaBytesArraySlicer`/`PlainVarSlicer` (the last fixing a real pre-existing out-of-bounds read/UB on truncated prefixes); `checked_len` covers every delta-decoded length before any `as usize`. Dead `DeltaBinaryPackedSlicer` removal is safe (zero references; no trait-object dispatch).
- **Writer:** moving the `non_null_count==0` early-return into the Plain arm and always emitting a DELTA header is correct. The `other =>` arm and the float→`Err` path are unreachable (`validate_encoding` permits DELTA only for the 12 integer-family tags, all mapping to i32/i64). The i64/i32 `encode(empty)` fast path is byte-identical to the filtered-empty iterator.
- **Tests:** every malformed-header/bit-width test has teeth — decoding the byte sequences by hand confirms each reaches its guarded line. The new Java `testConvertAllNullDeltaDataPageRoundTrips` is a genuine regression guard: 64-byte page size → 8 LONG rows → all-null first page (`column_top=0`, not a column-top shortcut), read eagerly via `CONVERT TO NATIVE` (the path that crashed pre-fix); `.noLeakCheck()` usage is correct and the page-size property is reset via `overrides.reset()` on teardown.
- **Concurrency / resources:** per-thread owned state, raw-pointer writes into a caller-owned `data_vec` with reserve-before-write; no leak/double-free/use-after-free.
- **fd-test deflake (`jni.rs`):** pipe-EOF detection instead of probing process-global fd numbers — a legitimate fix for the intermittent Rust-CI failure, unrelated to codec logic.

Cleared false positives: all-null f32/f64+DELTA error (unreachable — floats route to Plain); unsupported-encoding arm behavior change (dead code via `validate_encoding`); cross-version/replication hazard (OSS replication ships rows+SQL, re-encodes locally); `DeltaBinaryPackedSlicer` removal breaking a caller (zero callers).

## Summary
Approve. The substantive item is **M1**: a pre-existing, equally-reachable JVM-abort panic of the same class on the `RLE_DICTIONARY` index path, which this PR's "crashes on malformed files" hardening implies but does not close — a small, one-point fix consistent with the PR's own pattern. The remaining items are test-coverage gaps. 7 findings verified (1 Moderate out-of-diff + 6 Minor), 6 draft false-positives dropped; the one substantive finding is out-of-diff but in the same read path the diff hardens — the cross-context pass was the highest-value angle for a small diff to a widely-reachable subsystem.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) — automated review, please verify findings before acting.</sub>

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 1039 / 1075 (96.65%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/parquet_write/jni.rs|42|47|89.36%|
|:large_blue_circle:|parquet2/src/encoding/hybrid_rle/decoder.rs|13|14|92.86%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/delta_length_array.rs|118|128|92.19%|
|:large_blue_circle:|parquet2/src/encoding/delta_bitpacked/decoder.rs|69|75|92.00%|
|:large_blue_circle:|qdbr/tests/encode_primitives.rs|48|50|96.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/delta_binary_packed.rs|78|81|96.30%|
|:large_blue_circle:|qdbr/tests/encode_varchar.rs|29|30|96.67%|
|:large_blue_circle:|qdbr/src/parquet_read/decode.rs|217|222|97.75%|
|:large_blue_circle:|qdbr/src/parquet_read/slicer/tests.rs|159|162|98.15%|
|:large_blue_circle:|parquet2/src/encoding/uleb128.rs|5|5|100.00%|
|:large_blue_circle:|qdbr/tests/encode_binary.rs|27|27|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/slicer/rle.rs|4|4|100.00%|
|:large_blue_circle:|qdbr/tests/encode_strings.rs|27|27|100.00%|
|:large_blue_circle:|qdbr/tests/decode_varchar_slice.rs|24|24|100.00%|
|:large_blue_circle:|parquet2/src/encoding/bitpacked/decode.rs|9|9|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/slicer/mod.rs|112|112|100.00%|
|:large_blue_circle:|qdbr/src/parquet_write/encoders/numeric.rs|25|25|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decode/decimal.rs|33|33|100.00%|

bluestreak01 (approved):

PR #7214, opened 2026-06-09
feat(sql): add `is_end_of_month` SQL function

Dracylfrr:
Adds a SQL helper function, is_end_of_month(timestamp), which returns true when the provided timestamp falls on the final calendar day of its month.

This is useful for monthly reports, billing-period logic, finance workflows, and calendar-based time-series analysis.

Includes tests for leap-year February, non-leap-year February, 30-day months, 31-day months, and timestamp_ns inputs.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/functions/date/IsEndOfMonthFunctionFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/functions/date/IsEndOfMonthFunctionFactory.java
@@ -34,6 +34,7 @@ import io.questdb.griffin.SqlExecutionContext;
 import io.questdb.griffin.engine.functions.BooleanFunction;
 import io.questdb.griffin.engine.functions.UnaryFunction;
 import io.questdb.std.IntList;
+import io.questdb.std.Numbers;
 import io.questdb.std.ObjList;
 
 public class IsEndOfMonthFunctionFactory implements FunctionFactory {
@@ -44,7 +45,13 @@ public class IsEndOfMonthFunctionFactory implements FunctionFactory {
     }
 
     @Override
-    public Function newInstance(int position, ObjList<Function> args, IntList argPositions, CairoConfiguration configuration, SqlExecutionContext sqlExecutionContext) {
+    public Function newInstance(
+            int position,
+            ObjList<Function> args,
+            IntList argPositions,
+            CairoConfiguration configuration,
+            SqlExecutionContext sqlExecutionContext
+    ) {
         final Function arg = args.getQuick(0);
         return new Func(arg, ColumnType.getTimestampDriver(ColumnType.getTimestampType(arg.getType())));
     }
@@ -67,6 +74,9 @@ public class IsEndOfMonthFunctionFactory implements FunctionFactory {
         @Override
         public boolean getBool(Record rec) {
             final long value = arg.getTimestamp(rec);
+            if (value == Numbers.LONG_NULL) {
+                return false;
+            }
             return driver.getDayOfMonth(value) == driver.getDaysPerMonth(value);
         }
 
--- a/core/src/main/java/io/questdb/griffin/engine/functions/date/IsEndOfMonthFunctionFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/functions/date/IsEndOfMonthFunctionFactory.java
@@ -1,0 +1,78 @@
+/*+*****************************************************************************
+ *     ___                  _   ____  ____
+ *    / _ \ _   _  ___  ___| |_|  _ \| __ )
+ *   | | | | | | |/ _ \/ __| __| | | |  _ \
+ *   | |_| | |_| |  __/\__ \ |_| |_| | |_) |
+ *    \__\_\\__,_|\___||___/\__|____/|____/
+ *
+ *  Copyright (c) 2014-2019 Appsicle
+ *  Copyright (c) 2019-2026 QuestDB
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


bluestreak01:
Hi @Dracylfrr, thanks for the contribution! I reviewed this and there's a blocking correctness bug. I built the branch and ran `IsEndOfMonthFunctionFactoryTest` locally — `testNull` fails, so this needs a fix before it can merge.

## Critical — `is_end_of_month(NULL)` returns `true` instead of `false`

`IsEndOfMonthFunctionFactory.java:67-71`:
```java
public boolean getBool(Record rec) {
    final long value = arg.getTimestamp(rec);
    return driver.getDayOfMonth(value) == driver.getDaysPerMonth(value);
}
```

There's no `Numbers.LONG_NULL` guard. For a NULL timestamp both driver methods return `Numbers.INT_NULL`:
- `MicrosTimestampDriver.getDayOfMonth(LONG_NULL)` → `INT_NULL` (`MicrosTimestampDriver.java:401-403`)
- `MicrosTimestampDriver.getDaysPerMonth(LONG_NULL)` → `INT_NULL` (`MicrosTimestampDriver.java:428-430`)
- (same in `NanosTimestampDriver` at lines 399-401 / 426-428)

So the function computes `INT_NULL == INT_NULL` → **`true`** for NULL. This is not just the literal-`null` case — **any NULL row in a real timestamp column** returns a spurious `true`, so `WHERE is_end_of_month(ts)` would wrongly select NULL rows.

Reproduced by running your own test:
```
[ERROR] IsEndOfMonthFunctionFactoryTest.testNull:36 expected:<is_end_of_month
false
> but was:<is_end_of_month
true
>
[ERROR] Tests run: 3, Failures: 1, Errors: 0, Skipped: 0
```

The sibling function `IsLeapYearFunctionFactory.java:67-74` handles this correctly. Suggested fix (matches that precedent and short-circuits the NULL decomposition):
```java
import io.questdb.std.Numbers;
...
@Override
public boolean getBool(Record rec) {
    final long value = arg.getTimestamp(rec);
    if (value == Numbers.LONG_NULL) {
        return false;
    }
    return driver.getDayOfMonth(value) == driver.getDaysPerMonth(value);
}
```

The non-NULL logic (`dayOfMonth == daysPerMonth`) is correct — `testTimestamp` and `testTimestampNs` both pass.

## Moderate

**1. Redundant date decomposition (performance).** `getDayOfMonth(value)` and `getDaysPerMonth(value)` are each self-contained: both independently call `getYear` → `isLeapYear` → `getMonthOfYear` before their final step (`MicrosTimestampDriver.java:404-407` and `431-434`). So `getBool` computes the year/month/leap decomposition **twice per row**, with `getYear` (the dominant cost) running twice. On a large `WHERE` scan this doubles the per-row date-math cost. A clean fix would add a driver primitive like `boolean isLastDayOfMonth(long timestamp)` that decomposes once (keeping the micro/nano abstraction intact). Not a correctness blocker.

**2. Test coverage gaps.** All three tests pass constant literals, which the optimizer constant-folds at compile time — so the actual per-row `getBool(Record)` path against a real column is never exercised. Please add a case that reads a timestamp column from a table with month-end, mid-month, and NULL rows. Also missing: last day of a 31-day month at the year boundary (e.g. `2025-12-31`), a NULL `timestamp_ns` case, and usage inside a `WHERE` clause.

## Minor

- **Not added to `core/src/main/resources/function_list.txt`.** Functionally harmless — the scanner auto-discovers factories by classpath reflection and the list only feeds an overload-ordering map — but every other date function is listed there, so add it for consistency.
- **No trailing newline** at EOF in both new files.
- **Redundant `.ddl(null)`** on every test assertion — `ddl` already defaults to null, so the call is a no-op. Drop it.
- **PR has no labels** — add `SQL` and `New feature`.

Once the NULL guard is added the test will pass; happy to take another look after.

Dracylfrr:
Hi @bluestreak01 

Thanks for the detailed review. I fixed the blocking NULL correctness issue by adding a Numbers.LONG_NULL guard in IsEndOfMonthFunctionFactory#getBool, so is_end_of_month(NULL) now returns false instead of comparing two null sentinel values.

I also expanded the tests to cover literal NULL, timestamp, timestamp_ns, real timestamp columns with NULL rows, usage inside a WHERE clause, and a year-boundary month-end case.

I removed the redundant .ddl(null) calls and added the factory to function_list.txt for consistency.

Verified locally:

mvn -pl core "-Dtest=IsEndOfMonthFunctionFactoryTest" "-Denforcer.skip=true" "-Dmaven.compiler.release=25" test

Tests run: 6, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS

Please let me know if this addresses the issue fully or if there is anything else I should adjust.

bluestreak01 (approved):
Approved. Reviewed at full depth (built the branch, ran the test suite — 6/6 pass — plus adversarial probes for DATE input, nanos columns per-row, pre-epoch negatives, and parallel-safe WHERE filtering; all correct).

The earlier blocking NULL bug is fixed by the `Numbers.LONG_NULL` guard, and the driver-based precision selection is actually more correct than the existing `is_leap_year` (which is latently wrong for nanos). No correctness blockers remain.

Non-blocking follow-ups (fine to address in this PR or later):

**Moderate (performance):** `getBool` calls `driver.getDayOfMonth(value)` and `driver.getDaysPerMonth(value)` separately, and each independently decomposes year/month/leap (`getYear` -> `isLeapYear` -> `getMonthOfYear`). So that decomposition runs twice per row on a hot WHERE predicate. Consider a single `TimestampDriver.isLastDayOfMonth(long)` primitive that decomposes once.

**Minor:**
- Add a trailing newline at EOF in both new files.
- Add `SQL` and `New feature` labels to the PR.
- Optional: add coverage for the STRING/VARCHAR/DATE coercion paths (verified working empirically).

Nice contribution — thanks!

PR #7215, opened 2026-06-09
perf(sql): skip parquet decoding of rows that LIMIT discards

kafka1991:
## Summary

`SELECT ... FROM t LIMIT 10` on a parquet partition currently decodes the entire first row group before LIMIT discards all but 10 rows; `LIMIT m, n` decodes the skip-landing row group in full before jumping to the skip position. This PR pushes a `maxRowsAfterSkip` hint from `LimitRecordCursor` down to the parquet decoder so the decode window covers only the rows that LIMIT will actually return — for `LIMIT m, n` the window starts at the skip landing row, so the skipped prefix is not decoded either.

## Mechanism

`RecordCursor.skipRows` gains a second parameter:

```java
void skipRows(Counter rowCount, long maxRowsAfterSkip);
```

The hint is advisory: iteration must remain correct if it is ignored. `RecordCursor.UNBOUNDED_ROW_COUNT` is the sentinel for "no upper bound known".

- `LimitRecordCursor` passes the hint at three sites: `calculateSize` passes `0`, `toTop` passes `baseRowsToTake`, its own `skipRows` passes `remaining - rowsToSkip`. The `if (counter.get() > 0)` guard in `toTop` is removed so the hint reaches the base even when there is no offset.
- `PageFrameRecordCursorImpl` honours the hint when the row cursor is a forward 1:1 scan (`filter == null && rowCursorFactory.isEntity() && rowCursorFactory.isForwardScan()`), clamping the parquet decode range via a new `PageFrameMemoryPool.navigateTo(int, int, int)` overload that takes a `[inFrameRowLo, inFrameRowHi)` window. The landing frame decodes `[skipTarget, skipTarget + take)` instead of `[0, skipTarget + take)`.
- Decoded buffers stay frame-origin-addressable: `remapColumns` shifts the published column vectors back by `inFrameRowLo` entries (data vector for fixed-size columns, aux vector for var-size columns, whose entries hold compacted-data offsets or absolute pointers), so records and row cursors keep using absolute frame-relative row indexes with no read-side changes.
- `ParquetBuffers` tracks the decoded window on both ends (`decodedRowLo`/`decodedRowHi`); a cache hit only stands when the cached window covers the request, so `toTop()` followed by a full rescan transparently re-decodes from row 0. The record-bound and column-subset `navigateTo` paths re-decode a partially decoded buffer to the full frame window, keeping `recordAt`/`getRecordB` random access safe.
- `ReadParquetRecordCursor` clamps post-skip row groups to the remaining row budget.
- Transparent wrappers forward the hint to their base. `UnionAllRecordCursor` passes the same cap to both legs.
- `AsyncFilteredRecordCursor` and `CrossJoinRecordCursor` accept the parameter but pass `UNBOUNDED_ROW_COUNT` downstream.

## What gets faster

- `SELECT ... FROM parquet_table LIMIT N` - the first row group decodes `min(rowGroupSize, N)` rows.
- `SELECT ... FROM parquet_table LIMIT m, n` - the landing row group decodes only the `n` taken rows; the skipped prefix and trailing rows are not decoded. Cost no longer depends on how deep the offset lands in the row group.
- Negative limits (`LIMIT -m, -n`) - the size pass runs on metadata and the landing row group decodes only the taken rows.
- `read_parquet('file') LIMIT N` - the post-skip clamp applies to standalone `read_parquet`.
- LIMIT over UNION ALL, projections, virtual functions - the hint passes through.

## Benchmarks

Table with 17 columns (2 SYMBOL, 3 LONG, VARCHAR, TIMESTAMP_NS, 10 `DOUBLE[][]` of 2x10, designated TIMESTAMP), 200_000 rows in one DAY partition converted to parquet (two row groups of 100_000), default config, local run:

| Query | master | this PR |
|---|---|---|
| `SELECT * FROM t LIMIT 99990, 100010` | 53.04 ms | 8.91 ms |
| `SELECT * FROM t LIMIT -10, -20` | 26 ms | 6 ms |

The first query lands the skip at row 99_990 of the first row group and crosses into the second: master decodes both groups in full (200k rows x 17 columns), the PR decodes 20 rows. The second lands near the end of the last row group: master decodes that group in full, the PR decodes 10 rows. The residual time is per-query fixed cost (file open, metadata, result serialization), so the gap widens with column count and row group size.

## What does not change

- `WHERE ... LIMIT N` - the async filter can drop rows, so an output-row hint cannot be converted into a per-frame decode clamp without knowing filter selectivity. Deferred to a follow-up.
- `CROSS JOIN ... LIMIT N` - master/slave cap split is non-trivial; uncommon in time-series workloads.
- Native (non-parquet) partitions - page frames are mmapped, nothing to clamp.
- `ReadParquetRecordCursor` skip path - row groups iterated through during the skip are still decoded fully; only post-skip row groups benefit. Whole-row-group skip via metadata is a separate change.

## Tradeoffs

- The default-method change touches ~20 cursor implementations and ~30 call sites (mostly tests). Most are one-line changes to forward the new argument.
- `LimitRecordCursor.toTop()` now always calls `base.skipRows`, even when there is no offset. For `LIMIT N`, this brings forward one frame fetch + decode that used to happen in the first `hasNext`. Total work is unchanged.
- For `UNION ALL`, the second leg receives the full cap rather than the cap minus first-leg output. It can over-decode by up to (rows produced by first leg) - still less than `UNBOUNDED_ROW_COUNT`.
- A query that random-accesses a clamped frame (`getRecordB`/`recordAt`) re-decodes that frame to the full window, giving up the clamp for that frame in exchange for unconditional safety. Plain sequential `LIMIT` reads keep the full benefit.

## Bug fixes

The same code path also fixes a stale-pointer bug exposed by the new test. `FullFwdPartitionFrameCursor.next(skipTarget)` and `FullBwdPartitionFrameCursor.next(skipTarget)` returned a skip-only frame skeleton that carried over a `parquetMetaDecoder` reference from a prior call. If the reader was evicted between cursor uses, the stale pointer reached `PageFrameAddressCache`, producing a use-after-free that could crash or return wrong rows for LIMIT queries against parquet-converted partitions. Both cursors now reset `frame.format = NATIVE` and `frame.parquetMetaDecoder = null` on the skip-only fast path. `FwdTableReaderPageFrameCursor.computeNativeFrame` applies the same reset. `ParquetTest.testLimitOffsetOverParquetAfterReaderEviction` covers this regression.

Review of the clamp gate found a second bug: `SortedSymbolIndexRowCursorFactory` overrides neither `isUsingIndex()` nor `isForwardScan()`, so the original `!isUsingIndex()` gate let its scattered symbol-index scan (`WHERE <single-partition interval> ORDER BY <indexed symbol> LIMIT n`) through the clamp, reading undecoded rows. The gate now requires `isEntity()`, which only the plain forward scan satisfies; `ParquetTest.testLimitOverParquetIndexedSymbolOrderBy` pins the index plan and would return wrong rows under the old gate.

## Test plan

- `ParquetTest.testLimitOffsetOverParquetAfterReaderEviction` covers the reader-eviction crash path; `testLimitOverParquetDescendingAfterReaderEviction` covers the backward cursor equivalent.
- `ParquetTest.testLimitOffsetSkipsLandingFramePrefixDecode` compares parquet against an identical native table across all column kinds (fixed-size, SYMBOL, STRING, VARCHAR, UUID, LONG256, `DOUBLE[][]`) for mid-row-group offsets, take spanning row groups, interval scans starting mid row group, and descending scans.
- `ParquetTest.testLimitOffsetThenToTopWidensDecodeWindow` verifies a partial decode window re-decodes from row 0 after `toTop()`.
- `ParquetTest.testLimitOverParquetIndexedSymbolOrderBy` covers the clamp-gate regression; `testLimitOverParquetForwardNoOffset`, `testLimitOverParquetSpansMultipleRowGroups`, `testLimitOffsetLandsMidParquetFrame`, `testLimitOffsetLandsMidParquetFrameTail`, `testLimitOverParquetDescendingScan`, `testLimitOverParquetUnionAll` cover the clamp shapes.
- Existing parquet LIMIT tests, `OrderByAscRowSkippingTest`, `OrderByDescRowSkippingTest`, `WindowJoinTest`, `JoinTest`, `ReadParquetFunctionTest`, `CoveringIndexTest`, `O3ParquetStaleReaderTest`, and `ParallelTopKFuzzTest` verify behaviour preservation.


puzpuzpuz:
Review: PR #7215 — perf(sql): skip parquet decoding of rows that LIMIT discards
  
  Reviewed at level 3 (full mission-critical pass: change-surface map, 11 parallel agents, per-finding verification against source at PR head a97bfdf6).

  This is a high-blast-radius PR: it changes the RecordCursor.skipRows contract across ~20 cursors, rebases raw column pointers for the parquet decode path, touches the Rust/JNI decode
  boundary, and bundles a use-after-free fix. I verified the core invariants myself (aux-offset linearity, frame↔row-group coordinate mapping, the LIMIT clamp math, the stale-pointer fix
  asymmetry) before and after agent dispatch.

  Critical

  None. No SQL-reachable correctness bug, resource leak, UAF, double-free, Rust panic, or data race was found. The pointer-rebasing math is algebraically exact (all var-size drivers'
  getAuxVectorOffset(n) are pure width·n with zero constant term, so addr+size is preserved), the clamp gate (filter==null && isEntity() && isForwardScan()) correctly excludes
  scattered/index/descending scans, the slicer drain-loop fix is correct and actually repairs a pre-existing latent bug, and the stale-parquetMetaDecoder fix is sound.

  Moderate

  M1 — DELTA-encoded text columns: the clamp does not reduce the O(row-group) length-block decode (in-diff, Rust). core/rust/qdbr/src/parquet_read/slicer/mod.rs:284-287,411-426. After
  .take(row_hi), both slicers drain all remaining length values to locate the data block (for r in decoder.by_ref() { r.map_err(...)?; }). QuestDB's own encoder emits DELTA_LENGTH_BYTE_ARRAY
  by default for STRING/VARCHAR/BINARY (parquet_write/schema.rs:693), so for SELECT <text_col> ... LIMIT n over a converted partition the landing page still walks every length — O(page
  rows), not O(n). This is not a regression vs master (master decoded all lengths anyway) — the data-byte copy is clamped — but the PR's "What gets faster" implies uniform savings. Two asks:
  (a) disclose this in the Tradeoffs section; (b) optionally, an O(miniblocks) skip_to_end on the vendored delta_bitpacked::Decoder (walking miniblock byte-spans via bitwidth headers
  without unpacking) would close the gap (~32× fewer ops at default mini-block size).

  M2 — Random access to a clamped frame does strictly more work than master for that frame (in-diff). PageFrameMemoryPool.java:169-172. getRecordB/recordAt re-decode the full row group, so a
  clamped frame that is then random-accessed costs clamped decode + full decode > master's single full decode. Correctness is fine (verified by Agents 1, 9). The PR discloses "gives up the
  clamp" but frames it as neutral; it should state it can exceed master for that frame. Reachable when a LIMIT result feeds a random-accessing parent (some joins/window funcs); top-level
  streaming LIMIT is unaffected. Note (Agent 3): QueryAssertion's second pass triggers this in every .returns() test, so the extra decode is exercised but invisible (only values are
  asserted).

  M3 — Latent window-coverage gap in the record-bound navigateTo fast path (in-diff; NOT currently SQL-reachable). PageFrameMemoryPool.java:158 early-returns on
  record.getFrameIndex()==frameIndex && record.getBoundPool()==this without the decodedRowLo<=… && decodedRowHi>=… coverage check that the 3-arg navigateTo has (line 237-242). If skipRows is
  re-entered mid-frame with a larger maxRowsAfterSkip than the cached narrow window, a later hasNext reads past decodedRowHi into undecoded memory. I proved this cannot happen through the
  SQL engine — LimitRecordCursor is the sole originator of a finite cap, its first op establishes the window at baseRowsToTake, and remaining decreases monotonically so every later baseMax ≤
  baseRowsToTake (verified LimitRecordCursorFactory.java:249,297-303). But it's a fragile, undocumented invariant that the PR's own direct-cursor test style (skipRows(counter, 2)) could
  trip, and the in-diff AsyncFilteredRecordCursor comment explicitly anticipates a future caller pushing this hint. Cheap fix: add the coverage check to the fast-path early-return, or
  document the invariant on the navigateTo(int,int,int) overload.

  M4 — Test coverage gaps.
  - Column-top / NULL through the clamped shift (highest value): testLimitOffsetSkipsLandingFramePrefixDecode covers 8 column kinds but the parquet partition has no NULLs/column-tops, so the
  dataAddr!=0/auxAddr!=0 guards in remapColumns (PageFrameMemoryPool.java:741,748) are never exercised with frameRowLo>0. Add an ALTER TABLE … ADD COLUMN (column-top → address 0) then LIMIT
  <offset>,<take> landing mid-frame, compared against a native twin.
  - Clamped leg B in UNION ALL: testLimitOverParquetUnionAll uses LIMIT 7 (skip=0), so leg B is never clamped. Add LIMIT m,n where the offset crosses into B (sets B's cap).
  - Rust slicer row_lo>0: the new unit tests call try_new(&encoded, row_count, row_count) — only the row_hi==sliced_row_count (start-at-0) case. Add a case with sliced_row_count < row_count
  to lock the drain + non-zero-window arithmetic at the unit level (currently only covered indirectly at the Java level, and only if those columns actually use DELTA).
  
  Minor

  - UnionAll comment is inverted (in-diff). UnionAllRecordCursor.java:81-86. The "passing the full cap to B can over-decode by up to (rows consumed from A)" scenario does not occur: when the
  skip lands in A, B's skipRows is never called (B runs UNBOUNDED); when the skip consumes A, A yields 0 post-skip rows so B's cap is exact. Reword to match actual behavior.
  - remapColumns comment wording (in-diff). PageFrameMemoryPool.java:695-697: "shift only the aux vector, since aux entries hold compacted-data offsets … that need no adjustment" reads as
  self-contradictory (the code does shift the aux base pointer). Clarify that the data vector is unshifted and the offset values inside aux entries are unchanged.
  - Missing labels (metadata). The PR changes .rs files and on-disk decode behavior but lacks rust and storage labels (it has Bug/SQL/Performance).
  - Title hides correctness fixes (judgment). perf(sql): bundles a use-after-free crash fix and two wrong-results fixes (index-scan gate, DELTA partial-decode). Defensible since they share
  the exposing test path, but consider whether the UAF deserves a separate backportable fix(core): PR, or at least surface it in the squash body.
  - Benchmark presents only the best-case shape (deep offset, 17 wide columns). Add a neutral case (no-offset LIMIT N) and acknowledge the toTop brought-forward decode, per the
  equal-weight-tradeoffs convention.
  - Doc nits (in-diff). PageFrameMemoryPool.java:202 (2-arg navigateTo overload) has no javadoc; line 254 dropped the // parquet buffers use 0 offset comment that its sibling at 291 still
  carries.
  - Verify shipped binaries (process). The PR commits rebuilt .so/.dll/.dylib; these cannot be verified against source from here. Confirm CI rebuilds them from a97bfdf6 — an intermediate
  build would be buggy for DELTA_BYTE_ARRAY partial windows.

  Downgraded (false positives — verified against source)

  - "Missing parquetMetaDecoder reset in BwdTableReaderPageFrameCursor.computeNativeFrame." False positive. The Bwd cursor's getParquetDecoder() returns the cursor-level
  reenterParquetDecoder (:531), which nextSlow clears on the native branch (:479) and the reenter native path gates on (:163). Only the Fwd cursor uses a frame-level parquetMetaDecoder
  field, so only it needed the reset. The asymmetry is correct.
  - "decode() throwing mid-decode leaves a corrupt cached buffer (stale frameIndex marked cacheHit)." False positive. The decodedRowLo/Hi = -1 sentinel forces a re-decode on the next visit
  (-1 < rowGroupHi is always true), so the post-throw state self-heals. This actually closes a pre-existing hole in the old bare !cacheHit guard.
  - "New mutable state (maxRowsAfterSkip, decodedRowLo/Hi, isInSkipRows) needs volatile/synchronization." False positive. Only PageFrameRecordCursorImpl (single-threaded) ever produces a
  clamped buffer; all parallel/reduce-worker pools call only the full-window navigateTo overloads. No clamped state crosses a thread boundary. (One LOW future-proofing note: nothing enforces
  this, and the anticipated async-dispatch change would need to preserve it — overlaps with M3.)

  Summary

  Verdict: approve. No blocking issues. The mechanism is carefully engineered and the bundled UAF/wrong-results fixes are correct and well-tested. Address the Moderate items before merge if
  convenient — none blocks: M1/M2 are honesty-of-disclosure + optional optimization, M3 is a cheap defensive hardening of a non-reachable gap, M4 are worthwhile test additions (M4's
  column-top case is the most valuable as it directly exercises the new addr==0 shift guards).

  - Draft findings: ~14 verified, 3 dropped as false positives after source verification.
  - In-diff vs out-of-diff: all confirmed findings are in-diff. The cross-context pass ran in full (Agent 9 walked every skipRows caller/override, every navigateTo callsite, and all 10
  RowCursorFactory implementations; I independently verified the LIMIT lifecycle and the Bwd-cursor asymmetry) and legitimately found zero out-of-diff breakage — because the new skipRows
  contract is upheld at every callsite (sole finite-cap originator is LimitRecordCursor with monotonic remaining), not because the pass underran.
  - Key tradeoffs to weigh: the headline speedup is real for forward LIMIT/LIMIT m,n over fixed-size and long-string columns, but is materially smaller for short DELTA-encoded text columns
  (M1) and inverts to a small net loss for random-accessed clamped frames (M2) — both correctness-safe, both under-disclosed in the PR description.

kafka1991:
## Code review — `perf(sql): skip parquet decoding of rows that LIMIT discards`

Full mission-critical pass (correctness, concurrency, resources, Rust safety, performance, tests, cross-context callers). Every finding was verified against source.

**Verdict: Approve with minor follow-ups.** No correctness, concurrency, resource-leak, Rust-panic, or memory-safety bug found. The pointer-rebase math, the clamp gate (`isEntity() && isForwardScan()`), the Rust slicer drain fix, the stale-`parquetMetaDecoder` fix, and the `skipRows` signature migration (~20 overrides, ~17 callers, all SAFE) are sound.

### Critical
None.

### Moderate

**1. Var-len LIMIT decode is O(page_rows) to find the data offset, when an O(page_rows/miniblock) primitive already exists in the codebase.**
`core/rust/qdbr/src/parquet_read/slicer/mod.rs`: `DeltaLengthArraySlicer::try_new` (drain at 284-287) and `DeltaBytesArraySlicer::try_new` (drains at 411-413 and 423-425).

The drain loops walk **every remaining length value** one at a time (`for r in decoder.by_ref()`) only to advance `consumed_bytes()` to the concatenated-data start. Combined with `take(row_hi)`, `try_new` bit-unpacks **all N length values of the page regardless of how few rows the LIMIT slices** — and `DeltaBytesArraySlicer` does it twice (prefix + suffix). These chunks are typically one page spanning the whole row group, so `SELECT s,v FROM t LIMIT 2` over a 1000-row group still decodes ~1000 (string) / ~2000 (delta-byte) lengths to serve 2 rows. This is the exact hot path the PR exists to speed up, so the var-len win is largely given back.

The cheaper primitive is already in this repo and already used by the sibling path: `MiniblockIterator::get_end_pointer()` (`decoders/delta_binary_packed.rs:152`) sums per-miniblock bitwidth bytes — O(N/miniblock_size), zero value unpacking. `DeltaLAVarcharSliceDecoder::try_new` (`decoders/delta_length_array.rs:48-52`) already computes `data_offset` with it. Both symbols are `pub(crate)` under `pub mod decoders`, reachable from `slicer/mod.rs`.

**Do not just delete the drain** — for `DeltaLengthArraySlicer` it is a genuine correctness fix (master took `consumed_bytes()` mid-block and served garbage on multi-block partial decodes, which the new tests pin). Replace the drain with `get_end_pointer()`: keeps correctness, restores the speedup. Not a regression versus master (master walked all lengths too), so this is an enhancement, not a blocker — but a perf PR leaving its headline win on the table for STRING/VARCHAR/BINARY columns.

### Minor

**2. `PageFrameRecordCursorImpl.of()` does not reset the two new clamp fields** (`PageFrameRecordCursorImpl.java:155-168`). `of()` resets `isExhausted`/`rowCursor`/`areCursorsPrepared` but not `maxRowsAfterSkip` / `rowsProducedSinceSkip` (only `toTop()`/`skipRows()` do). Not reachable today (a finite cap is only ever set under a `LimitRecordCursor`, which always routes through `toTop()`/`skipRows()` before the first `hasNext()`), but a stale finite `maxRowsAfterSkip` on a reused cursor would make `hasNext()` (line 111 guard) silently return zero rows. A one-line defensive reset in `of()` removes the fragility.

**3. `navigateTo(int, IntHashSet)` early-returns without the window-coverage check** (`PageFrameMemoryPool.java:275-278`), unlike the new 3-arg overload which verifies `decodedRowLo/decodedRowHi` cover the request. Safe today because its only caller (`PageFrameReduceTask`) runs on a separate non-clamping pool. Worth extending the documented invariant (or adding the coverage check) so a future caller can't get a partial buffer for a full request.

**4. `ReadParquetRecordCursor.calculateSize` can undercount after a clamp** (`ReadParquetRecordCursor.java:212`): `remaining = rowGroupRowCount - currentRowInRowGroup - 1`, but `rowGroupRowCount` is now the clamped decode count. Not reachable today (`read_parquet` `size()` is always known, so `LimitRecordCursor` never descends into this `calculateSize` after a clamped skip). A guard (skip the clamp inside `calculateSize`, or re-decode full) would harden it.

**5. Test coverage gaps:**
- `testLimitOffsetThenToTopWidensDecodeWindow` reads the 2 in-window rows then jumps to `toTop()` — it never asserts `hasNext() == false` at the clamp boundary, so it doesn't prove the `rowsProducedSinceSkip >= maxRowsAfterSkip` guard stops before reading undecoded memory (frame row 7, value 8). Add the `assertFalse(hasNext())`.
- No test exercises `calculateSize` → `skipRows(counter, 0)` over an **unknown-base-size** parquet scan (every new `.expectSize()` test has a known base size and takes the other branch).
- No end-to-end test of the hint through a **VirtualFunction** projection (`SELECT a+1 FROM parquet_tbl LIMIT m,n`) or a **CrossJoin/AsOf-join** parquet leg, both of which the PR threads the parameter through.
- Rust: the DELTA_BYTE_ARRAY **prefix-block** truncation error path has no dedicated test (only suffix-block truncation is covered).
- The `-m,-n` (negative offset + negative limit) case is only covered via `read_parquet()`, not the CONVERT-PARTITION-TO-PARQUET storage path that is the PR's main target.

**6. Metadata:** title/tone/tradeoffs/benchmarks are compliant. Consider adding a **`rust`** label since the PR makes a non-trivial Rust change. Cosmetic: the two slicers drain with different idioms (`decoder.by_ref()` vs `&mut decoder`) — identical behavior, unify for readability.

### Investigated and dismissed (false positives)
- **Record bound to a partial window reading outside it via the fast path** — the decoded window of a `ParquetBuffers` is monotonically non-shrinking (a coverage miss only widens), `LimitRecordCursor`'s window only shrinks across a scan, and `getRecordB`/`recordAt` force a full re-decode that keeps both records' absolute-index addressing consistent.
- **`remapColumns` rebase reads before buffer / breaks var-size** — `getAuxVectorOffset(row)` is linear (`WIDTH*row`, no header) for Varchar/String/Array, so the aux-base shift is the exact inverse; fixed-size data-base shift is `frameRowLo<<pow2`; `addr != 0` guards keep NULL/column-top columns at 0. Verified across all column kinds against native baselines.
- **Descending / indexed-symbol scans clamp and read undecoded rows** — `isForwardScan()` (false for DESC) and `isEntity()` (false for all 10 index/scattered factories) close the gate; only `PageFrameRowCursorFactory`'s forward 1:1 entity scan passes.
- **UnionAll over-decodes / drops rows by passing the same cap to both legs** — when the skip lands in A, B runs unclamped later; when it crosses into B, A yields zero post-skip rows so B's cap is exact.
- **Rust panic/overflow on corrupt pages** — `&data[data_offset..]` can't overflow (`consumed_bytes() <= len` invariant), drained lengths never reach `checked_len`, parquet2's decoder returns `Err` (not panic) on malformed blocks, and errors propagate to a Java `CairoException`. `cargo check/clippy/test` clean (963 lib tests pass).

### Summary
The mechanism is correct and carefully reasoned; the two bundled bug fixes (stale `parquetMetaDecoder` UAF on skip-only skeletons; partial-decode garbage for delta var-len columns) are legitimate and well-tested. Tradeoffs to weigh: `LIMIT N` no-offset brings one frame fetch/decode forward into `toTop()` (reordered, not added work); every serial-scan `hasNext()` gains two unconditional long ops even without a LIMIT (negligible); the var-len win is partly forfeited by Finding 1.

The single item worth acting on before merge is **Finding 1** (replace the slicer drain with `get_end_pointer()`) so the var-len LIMIT path matches the headline win — an enhancement, not a correctness gate, so it could also land as a fast follow-up.

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=241521)
**Tested Commit:** `1bdc295cebb58fd82c8e88bd93ae0839a96811f3`

bluestreak01 (changes requested):
@kafka1991

Review at full depth (change-surface map, parallel review passes, per-finding source verification at PR head `48158653df`). I reproduced the one Critical finding with a locally compiled test, and verified the highest-stakes claims (Rust return semantics, master diffs, the clamp-gate audit) directly against source. One finding from an earlier review on this PR is overturned with empirical evidence (see Downgraded).

## Critical

### C1. Nested LIMIT over `generate_series` returns wrong rows — regression, reproduced

```sql
SELECT * FROM (SELECT * FROM generate_series(1, 100, 1) LIMIT 5, 55) LIMIT 3
-- PR head returns: 4, 5, 6      correct: 6, 7, 8
```

Ran as a JUnit test against the PR head; it fails exactly as shown (`expected <6 7 8> but was <4 5 6>`).

**Code path.** `LimitRecordCursor.toTop()` now calls `base.skipRows(counter, baseRowsToTake)` unconditionally (`LimitRecordCursorFactory.java:288-291`; master guarded it with `if (counter.get() > 0)`). For the nested query, the outer cursor's `toTop` first triggers the inner LIMIT's `toTop`, which positions `generate_series` at row 5; the outer's second call then reaches the inner's `skipRows(0, 3)`, which forwards `base.skipRows(0, ...)` to the already-positioned series cursor. `GenerateSeriesLongRecordCursor.skipRows` (`GenerateSeriesLongRecordCursorFactory.java:121-129`) computes `currentRowId = recordA.getRowId() - 2` — a correction only valid at the top sentinel, where `curr = start - step` makes `getRowId()`'s `abs()` alias the position to rowId 2. Mid-state, `getRowId()` returns the true rowId `p`, so the code computes `p - 2` and `recordAt` moves the cursor back two rows; the skipped rows 4 and 5 are re-produced.

The same math is in `GenerateSeriesTimestampRecordCursorFactory.java:136-144` and `GenerateSeriesTimestampStringRecordCursorFactory.java:203-216` (random-access units only; `'M'`/`'y'` fall back to the safe default loop).

**Regression scope.** On master the offset-0 outer LIMIT never invoked `skipRows`, so this query was correct; an outer LIMIT *with* offset over an inner offset-LIMIT was already wrong on master via the same math (pre-existing, narrower). The PR widens the bug to the common shape. The QueryAssertion battery cannot see it: its skip passes verify counts, not values, and the count arithmetic is independent of the (wrong) position.

**Fix.** Compute the position from the signed offset in all three cursors — `long currentRowId = (recordA.curr - start) / step + 1` yields 0 at the sentinel and `p` mid-state, making skip-of-0 idempotent and fixing the pre-existing master bug too. Add a regression test for the nested shape. (Guarding the `toTop` call would only mask the broken contract — `skipRows(0)` must be a positional no-op for every implementation.)

## Moderate

### M1. `navigateTo(int)` lost its same-frame early-return — per-row regression on filtered `LIMIT -N`

Master's `navigateTo(int)` began with `if (frameMemory.frameIndex == frameIndex) return frameMemory;`. The PR delegates to the 3-arg overload (`PageFrameMemoryPool.java:290-292`), which re-runs the full body — `getFrameFormat`, and for native frames `unbind` + 6 address-cache getters + 8 stores; for parquet, two row-group lookups, `tryHit`, window compares — on every call. `AsyncFilteredNegativeLimitRecordCursor.hasNext()` (`AsyncFilteredNegativeLimitRecordCursor.java:159`, not in the diff) calls this once per output row with long same-frame runs, so `WHERE <filter> ... LIMIT -N` pays roughly an order of magnitude more per row than master's single compare, for native and parquet tables alike.

The removal was load-bearing only for the one pool that mixes windows; the sibling `navigateTo(int, IntHashSet)` overload kept its early-return (line 382-384). Fix: restore a guarded early-return — return early when the frame matches and the binding is native or the bound buffer's window covers the request (an `isFullWindow` flag set in `decode()` makes this one compare).

### M2. `ReadParquetRecordCursor`: clamped decode stores the full row count with no read guard

Verified empirically with instrumented Rust tests: `decodeRowGroup(..., 0, decodeHi)` returns the **full chunk row count**, not the decoded window size — the Rust page loop accumulates `row_count` for every page while decoding only the overlap (`row_groups.rs:805-861`). So after the new clamped decode (`ReadParquetRecordCursor.java:386-397`), `rowGroupRowCount == rowGroupSize` while the buffers hold only `decodeHi` rows, and `hasNext()` (`:249-253`) has no cap check — it will position the record past the decoded window. Not reachable today because `LimitRecordCursor` reads at most its `remaining`; but `PageFrameRecordCursorImpl` hard-stops for the identical situation (`:111-114`) while this class would serve garbage — and the `isInSkipRows` comment ("relies on rowGroupRowCount == full size") suggests the return semantics were misread during development.

Fix: mirror the page-frame hard stop — return `false` from `hasNext()` once `rowsProducedSinceSkip >= maxRowsAfterSkip`. Do **not** clamp `rowGroupRowCount` instead: `calculateSize` (`:212`) and the skip arithmetic (`:346`) correctly rely on it being the full size.

### M3. Failed in-place widen evicts a buffer that sibling accessors still alias

The two new widen branches (`PageFrameMemoryPool.java:256-265` record path, `:358-366` frame path) are the first code paths that close a pinned, possibly record-aliased buffer on failure: `decodeWindowAndAccount` -> `evictHalfInitialized` frees all five native buffers, but the catch clears only the navigating accessor (`record.clear()` / `frameMemory.clear()`). A recordA bound earlier via `init(frameMemory)` keeps `frameIndex == f && boundPool == this`, so the fast path (`:232`) would early-return into freed `DirectLongList`s on the next access. Master's decode-throw paths only ever closed freshly acquired, never-aliased buffers. Bounded today — the exception terminates the query in every in-tree consumer — but it converts a transient decode error into a latent UAF for any future catch-and-continue caller. Fix: clear all three bindings in `evictHalfInitialized` (it already inspects `boundForFrameMemory`), or state on the 3-arg `navigateTo` that a throwing call poisons all bound records.

### M4. Empty decode windows: fabricated pointers published, plus a `LIMIT 0` regression

`LimitRecordCursor.calculateSize` -> `base.skipRows(counter, 0)` -> `navigateTo(frameIndex, k, k)` issues an empty `[k, k)` decode. Verified empirically: the Rust side decodes nothing but `refresh_ptrs()` (`decode.rs:61-70`) republishes a non-null pointer — the previous decode's stale allocation, or `NonNull::dangling()` = `0x1` on fresh buffers. `remapColumns(frameRowLo > 0)` (`PageFrameMemoryPool.java:1320-1338`) guards only `addr != 0`, so it shifts the stale/dangling pointer backwards and publishes a fabricated "readable" range into the record. Nothing dereferences it today (the 0 cap stops all reads; any later request fails window coverage and re-decodes), but the cursor holds wild address ranges indistinguishable from valid ones. Separately, `LIMIT 0`/`LIMIT n,n` now open the first partition and run this empty JNI decode where master's guarded `toTop` touched nothing — and the comment at `LimitRecordCursorFactory.java:182-184` ("max=0 lets the base skip the landing-frame decode altogether") overstates what happens.

One fix covers all of it: short-circuit `decodeLo == decodeHi` in `navigateTo(int, int, int)` — publish zeroed addresses, skip `openParquet`/JNI entirely.

### M5. `ParquetLimitProbeTest` is a committed review-scratch file

Missing the Apache license header (the only file in `test/griffin/` without one), no class javadoc, non-alphabetical methods, a "Probe" name describing review activity, and comments narrating the PR ("Not covered by the PR's tests..."). The five tests themselves are valuable — `testRandomAccessDuringClampedScan` is the only direct pin of the record-path widen — so keep them: fold into `ParquetTest` at alphabetical positions with the header and constraint-stating comments.

### M6. Coverage gaps worth closing in this PR

- A regression test for C1's nested-LIMIT shape (with the fix).
- `testLimitOffsetThenToTopWidensDecodeWindow`: add `assertFalse(cursor.hasNext())` at the cap before `toTop()` — today nothing pins that the guard stops before reading undecoded frame row 7.
- UNION ALL with the cap smaller than leg A's remaining rows (e.g. add `LIMIT 1, 3` to `testLimitOverParquetUnionAll`) — the in-leg-A hard-stop shape is untested.
- A virtual projection over the clamp (`SELECT i + 1 ... LIMIT 18, 24` over parquet) — `VirtualFunctionRecordCursor` forwards the cap but no test reads through rebased addresses via functions.
- ASOF JOIN with a parquet master and `LIMIT m, n` — the cap forwarding in `AbstractAsOfJoinFastRecordCursor` is the only join-side behavior change and the existing asof-parquet test has no LIMIT.
- Rust units: DELTA_BYTE_ARRAY prefix-block truncation (only the suffix block is covered) and the empty `try_new(x, 0)` shape.

## Minor

- `PageFrameRecordCursorImpl.of()` doesn't reset `maxRowsAfterSkip`/`rowsProducedSinceSkip` (`:155-168`) — a stale finite cap survives cursor reuse; masked only because `LimitRecordCursor` always `toTop()`s first. Two assignments close it.
- `navigateTo(int, IntHashSet)` keeps an unguarded same-frame early-return with no window check (`PageFrameMemoryPool.java:381-384`) — safe purely by caller segregation; add the coverage check or a comment naming the invariant.
- `RecordCursor.skipRows` javadoc calls the cap "advisory" and only binds implementations; it never states the caller-side obligation that finite-cap implementations may hard-stop, so reading past the cap is silent truncation. Same wording tension in the `navigateTo(int,int,int)` javadoc.
- `isForwardScan()` defaults to `true` with no javadoc (`RowCursorFactory.java:63-65`) — the unsafe polarity for a clamp gate (a future entity factory that forgets to override reads undecoded memory; `isUsingIndex()` in the same file defaults safe). At minimum document it. The predicate is also duplicated between `PageFrameRowCursorFactory.getCursor` and `isForwardScan` — share it.
- Missing `@Override` on `AbstractAsOfJoinFastRecordCursor.skipRows:188` (the only override in the diff without it).
- Parameter shadows the sanitized field in `PageFrameRecordCursorImpl.skipRows` (`:181,188,234`) — line 234 must say `this.maxRowsAfterSkip`; rename the parameter.
- Member ordering: `decodeWindowAndAccount` sits between `decodeAndAccount` and `decodeRowFilteredAndAccount` (should follow both); four alphabetical inversions among the 13 new `ParquetTest` methods.
- Unchecked `(int)` casts in the clamp arithmetic (`PageFrameRecordCursorImpl.java:129,238-239`) overflow for hypothetical >2^31-row native frames — harmless (window ignored for NATIVE) but cheap to clamp.
- `TimeFrameCursorTest.assertScatteredCacheInvariants` drive-by: drops the `budget` param and hardcodes `2 * 131072L` while the callsite still defines `budget` — restore the parameter.
- `AsyncFilteredRecordCursor.skipRows` comment narrates a future PR; keep the constraint, drop the process narration.
- The 2-arg `navigateTo` overload has exactly one caller — inline or keep deliberately.
- Performance follow-ups (disclosed or acceptable, listed for tracking): `read_parquet` skip-spans still decode every crossed row group in full (O(skip) where metadata bookkeeping suffices); backward scans (`ORDER BY ts DESC LIMIT N`, the canonical time-series query) get no clamp though the window plumbing supports it; ~2 extra ops per row of cap bookkeeping on every scan (predictable branch, benchmark-watch); the record-path miss fetches `rowGroupLo/Hi` it never uses.
- Metadata: title describes mechanism, not user impact (cf. #7222's "speed up ORDER BY ... LIMIT N queries"); the body's test plan predates the final state (no mention of the Rust `tests.rs` additions, `ParquetLimitProbeTest`, the new `ReadParquetFunctionTest` tests, or the rebuilt DLL); consider the `storage` label; benchmarks lack run methodology. The Windows DLL diff was investigated and is fine — it's the GitHub Actions rebuild workflow's output with PE non-determinism, not a stale local build.

## Downgraded (claims checked and dismissed during verification)

1. "`limit -2, -1999` test case is vacuous (both bounds resolve empty)" — `of()` swaps same-sign out-of-order bounds (`lo > hi && sameSign` -> swap), so the case skips 3001 and takes 1997 rows of 5000. Real coverage.
2. "`calculateSize` undercounts because `rowGroupRowCount` holds the clamped count" (raised in an earlier review on this PR) — the Rust decode empirically returns the **full** chunk count, so `calculateSize` is correct as-is. The actual hazard is the inverse (M2).
3. "`BwdTableReaderPageFrameCursor.computeNativeFrame` is missing the `frame.parquetMetaDecoder = null` fix" — the Bwd frame class has no decoder field; its `getParquetDecoder()` returns the cursor-level `reenterParquetDecoder`, which `nextSlow`'s native branch nulls (`:479`) before `computeNativeFrame` runs. The Fwd/Bwd asymmetry in the fix mirrors an asymmetry in the design.
4. "Bwd interval-scan skip skeletons leave the frame decoder unassigned -> assert fires / stale decoder cached" — the skeleton branch assigns `reenterParquetDecoder` from the current, live partition frame (`:185-189`), which is exactly what the frame's getter returns; NATIVE skeletons null it.
5. "New mutable state (`decodedRowLo/Hi`, cap fields) needs volatile/synchronization" — full thread-attribution trace shows all windowed-decode state confined to the query-owner thread; parallel pools receive only full-window calls.
6. "Empty-window decode can panic in Rust" — empirically clean: the page loop decodes nothing for `[x, x)`, both delta slicers handle `try_new(x, 0)`, errors propagate as `CairoException`, never a panic.

## Summary

**Request changes.** C1 is a reproduced wrong-results regression reachable from plain SQL and must be fixed before merge (fix the GenerateSeries skip math in all three variants + regression test). M1 is a real per-row performance regression on filtered negative-limit queries with a cheap fix. M2-M4 are latent-but-armed hazards where the new machinery's safety rests on undocumented caller discipline — each has a small hardening fix. The core windowed-decode design is otherwise sound: the pointer-rebase algebra, the clamp gate, the cache-coverage protocol, and the bundled UAF fix all survived adversarial verification.

Tradeoffs to weigh alongside the headline speedup: universal per-row cap bookkeeping, the `LIMIT 0`/sizing-path empty decodes (M4), random-access re-decode cost (disclosed in the PR), and the still-unclamped `read_parquet` skip spans and DESC scans.

Verification stats: ~33 deduplicated draft findings -> 1 Critical, 6 Moderate, ~16 Minor confirmed; 7 dropped as false positives after source/empirical verification. The cross-context pass walked all 20 `skipRows` implementations, 17 callers, 10 `RowCursorFactory` implementations, ~50 `navigateTo` sites, and 13 `next(skipTarget)` implementations with explicit verdicts; C1's root cause lives in pre-existing lines of signature-only-touched files with an in-diff trigger, and M1's hot caller is outside the diff.

kafka1991:
## Code review — `perf(sql): skip parquet decoding of rows that LIMIT discards`

Reviewed at full depth (change-surface map, 11 parallel review passes covering correctness / concurrency / performance / resources / Rust safety / tests / cross-context callers / fresh-adversarial / metadata, with per-finding source verification at PR head `9db6e9ccd1`). This PR has been through three prior rounds; the current head incorporates fixes for every blocking finding raised, so this pass focused on verifying those fixes are correct and complete, and on regressions they may have introduced.

### Verdict: approve. No blocking issues.

## Critical
None. No SQL-reachable correctness bug, resource leak, use-after-free, double-free, Rust panic, or data race survived verification.

Load-bearing invariants verified directly against source:
- **Pointer-rebase is algebraically exact.** `getAuxVectorOffset(n)` is `WIDTH*n` with zero constant term for Varchar (`VarcharTypeDriver:469`), String (`StringTypeDriver:129`), Array (`ArrayTypeDriver:334`), Binary (extends String), so `auxAddr -= getAuxVectorOffset(frameRowLo)` exactly inverts the dropped prefix and preserves `addr+size == bufferEnd`. The `addr != 0` guards keep column-tops / all-NULL columns at address 0.
- **C1 wrong-results regression fixed in all three GenerateSeries cursors.** `currentRowId = (recordA.curr - start) / step + 1` is exact because `curr` is always on-grid (exact integer division, incl. negative step and after arg-swap), making skip-of-0 a true positional no-op.
- **Clamp gate `filter == null && isEntity() && isForwardScan()` is sound.** `PageFrameRowCursorFactory` is the only factory with `isEntity()==true` and correctly returns `isForwardScan()==false` for DESC; every index/scattered factory returns `isEntity()==false`, so the prior `SortedSymbolIndexRowCursorFactory` bug class cannot recur.
- **`bindGeneration` closes the failed-widen UAF.** `evictHalfInitialized` bumps the generation, nulls the bound records, and clears `frameMemory`; the record fast path early-returns only on `boundGeneration == bindGeneration`.
- **Rust partial-window `data_offset` is correct and O(miniblocks).** `delta_stream_byte_len`/`get_end_pointer()` walks miniblock byte-spans via bitwidth headers without unpacking values — fixes the prior garbage-strings bug and does not forfeit the var-len win. Both delta slicers guard the empty-buffer 0/0 panic; all accesses are bounds-checked; malformed pages return `Err`, never panic. `cargo check/clippy/test` clean (1014 lib tests).
- **No clamped state crosses a thread boundary.** Windowed `navigateTo` overloads are called only from single-threaded `PageFrameRecordCursorImpl`; parallel/reduce pools use full-window overloads on distinct instances, so the absence of `volatile` is correct.

## Moderate
None. The items below are presentation/hardening only.

## Minor
- **`GenerateSeries*RecordCursor.skipRows` has an undocumented precondition** (`GenerateSeriesLongRecordCursorFactory.java:121` + the two siblings): `currentRowId = (curr-start)/step + 1` is correct only because the sole caller (`LimitRecordCursor.toTop`) always calls `base.toTop()` immediately before `skipRows`, guaranteeing `curr == start-step`. Called mid-iteration it would mis-position. Add a one-line comment stating the precondition.
- **`read_parquet(...) LIMIT m,n` still decodes every row group crossed during the skip in full** (`ReadParquetRecordCursor`, `isInSkipRows` defeats the clamp). Disclosed deferral, not a regression (master did too), but the most valuable unrealized win — a metadata-only whole-row-group skip is feasible (`getRowGroupSize` is already read at the decision point). Worth a follow-up.
- **Redundant lookup on the record-path cache miss** (`PageFrameMemoryPool.java:252-253`): `rowGroupLo/rowGroupHi` are fetched unconditionally but unused on the `parquetBuffers == null` branch (`decodeAndAccount` re-fetches). Two O(1) reads; hoist into the `else if` branch for tidiness.
- **CrossJoin intentionally passes `UNBOUNDED_ROW_COUNT` to both legs** (correct — cartesian output isn't 1:1 with either leg), so there is no clamped CrossJoin path and no test. Confirm deliberate.
- **Metadata:** add the `storage` label (changes on-disk parquet decode-window behavior); `rust` is borderline (the PR's only Rust delta is test code). The `perf(sql):` title undersells the bundled use-after-free crash fix and two wrong-results fixes — covered by the `Bug` label + "Bug fixes" body, but a `perf(sql):` release-note line won't signal the crash fix.
- **Benchmark presents only the two most favorable shapes** (deep offset, 17 wide columns). Per the equal-weight-tradeoffs convention, add a neutral `LIMIT N` no-offset break-even row and note the brought-forward `toTop()` decode (reordered, not added work).
- **Committed rebuilt native binaries** can't be verified from source here; CI must rebuild all five artifacts from `9db6e9ccd1`.

## Downgraded (draft findings dismissed after verification)
- "New clamp state needs `volatile`/synchronization" — false positive; thread-confined, parallel pools only get full-window calls.
- "A decode throwing mid-widen leaves a corrupt cached buffer" — false positive; the `decodedRowLo/Hi = -1` sentinel forces re-decode and `bindGeneration` invalidates aliasing records; self-heals.
- "`navigateTo(int)` lost its early-return" — resolved in current code; `isFrameMemoryCovering()` restores an O(1) early-return.
- Prior rounds' blocking findings — all verified resolved at this head: C1 (GenerateSeries skip math), var-len O(page_rows) offset walk (now O(miniblocks)), `ReadParquetRecordCursor` missing read-guard (now hard-stops on `rowsProducedSinceSkip >= maxRowsAfterSkip`), empty-window fabricated pointers (now short-circuits `decodeLo==decodeHi` before any `openParquet`/JNI), `of()` not resetting clamp fields (now resets), missing `@Override` on the ASOF skipRows, and the `ParquetLimitProbeTest` scratch file (folded into `ParquetTest`).

## Summary
The mechanism is carefully engineered and every previously-flagged blocking defect is fixed and well-tested. Remaining items are presentation (labels, title, benchmark balance) and one cheap doc hardening (the GenerateSeries skip precondition).

- ~5 confirmed findings (all Minor), plus ~7 prior-round findings verified resolved and 3 draft findings dropped as false positives.
- All confirmed findings are in-diff. The cross-context pass ran in full — 18 `skipRows` overrides (all 2-arg, no missed migration), every `navigateTo` callsite, and all `RowCursorFactory` implementations walked with explicit SAFE verdicts — and legitimately found zero out-of-diff breakage, because the new `skipRows` contract is upheld at every callsite (sole finite-cap originator is `LimitRecordCursor`, whose `remaining` is monotonic).
- Tradeoffs: the headline speedup is real for forward `LIMIT`/`LIMIT m,n` over fixed-size and var-len text columns; it inverts to a small net loss for a clamped frame that is later random-accessed (re-decodes full — disclosed); `read_parquet` skip-spans and DESC scans get no clamp yet. All correctness-safe.

<sub>Generated with Claude Code (Opus 4.8, 1M context)</sub>

bluestreak01 (approved):
Approve.

Reviewed at level 3 (change-surface map, 11 agent roles across 6 parallel fresh-context reviewers, per-finding source verification at branch head; merge-base 9a6b561c).

No Critical issues. Independently verified the three load-bearing invariants:
- Pointer-rebasing math is algebraically exact: all var-size drivers have getAuxVectorOffset(row)=c*row with zero constant (Varchar/Array=16*row, String=8*row), so f(r)-f(lo)=f(r-lo) holds and absolute frame-relative indexing lands on the right decoded row; addr+size preserved on both fixed and var-size paths.
- Clamp gate is airtight: of all 12 RowCursorFactory impls, only PageFrameRowCursorFactory returns isEntity()==true, and it overrides isForwardScan() for descending order. The change from !isUsingIndex() to isEntity() is a real fix for the SortedSymbolIndexRowCursorFactory hole.
- skipRows contract propagation is complete (compile-enforced); cross-context pass found zero out-of-diff breakage.
- bindGeneration UAF guard is conservatively safety-additive; clamped buffers are strictly single-threaded (parallel reduce pools use only full-window navigateTo).

Non-blocking follow-ups (address before merge if convenient, none blocks):
- Moderate: PR description over-claims uniform O(n) — DELTA_LENGTH text columns retain an O(page_values/32) length-stream walk on the boundary page (still a large win, optimal for the layout); random-access of a clamped frame can be ~2x master's decode for that frame (disclose it can exceed, not just 'gives up clamp'); consider surfacing the bundled use-after-free fix in the squash body / a Fixes #NNN since perf(sql) hides a crash fix on squash-merge.
- Minor: recordA.setRowIndex(0) after a clamped landing points before the buffer (never dereferenced under the cursor contract, but a latent footgun — prefer setRowIndex(skipTarget)); add a coverage assert to the navigateTo record fast-path; add Windows (and arguably storage) labels; the lone rebuilt Windows DLL has no production Rust source change vs merge-base — confirm CI rebuilds from HEAD.

The mechanism is disciplined and the bundled crash/wrong-results fixes are correct and well-tested.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 309 / 325 (95.08%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/ExtraNullColumnRecordCursor.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/sql/RowCursorFactory.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/ReadParquetRecordCursor.java|34|41|82.93%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|93|100|93.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/GenerateSeriesTimestampRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FwdTableReaderPageFrameCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryRecord.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/QueryProgress.java|1|1|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/slicer/tests.rs|122|122|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/PageFrameRecordCursorImpl.java|24|24|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/PageFrameRowCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FilterOnSubQueryRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/GenerateSeriesTimestampStringRecordCursorFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestBySubQueryRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AbstractAsOfJoinFastRecordCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/CrossJoinRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/VirtualFunctionRecordCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/GenerateSeriesLongRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/LimitRecordCursorFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullBwdPartitionFrameCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/union/UnionAllRecordCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/FullFwdPartitionFrameCursor.java|2|2|100.00%|

PR #7218, opened 2026-06-09
fix(sql): fix wrong results in splice, right and full outer joins with a left-table filter

puzpuzpuz:
## Summary

SPLICE, FULL OUTER and RIGHT OUTER joins NULL-extend the master (left) side: SPLICE emits slave-only timestamps paired with a NULL master, while FULL and RIGHT OUTER emit unmatched right rows with a NULL master. The optimizer pushed a WHERE predicate that referenced only the master table down into that table's
sub-query, because it gated the push solely on the referenced table's own join type (the base table is an inner join) and ignored any downstream NULL-extending join.

The pushed predicate then left those NULL-master rows unfiltered, and for SPLICE it also changed which master row prevailed at each slave timestamp. As a result any of these three joins could return wrong results when a filter referenced only the master table.

The query fuzzer surfaced this as a literal-vs-bind divergence: with a literal constant the optimiser could also propagate it to the slave through the join key, while the bind-variable form could not, so the two forms disagreed on the row count. The literal form looked correct only when the master happened to contain the key.

## Fix

`masterNullingJoinIndex()` reports the outermost downstream SPLICE, FULL OUTER or RIGHT OUTER join that NULL-extends a given table. `analyseEquals()` and`assignFilters()` consult it and keep such a predicate as a post-join filter (`addPostJoinWhereClause`) instead of pushing it into the table's sub-query. LEFT OUTER, ASOF and LT joins keep every master row, so their master-side predicates still push down unchanged.

## Tradeoff

Because the predicate now runs after the join rather than inside the master sub-query, a master-side filter on these three join types can no longer use an index or interval scan on the master table; the query runs as a full scan plus a post-join filter. This is the cost of correctness for these joins - the previous index/interval-scan plan produced wrong results.

## Test plan

- Adds `JoinTest#testSpliceJoinMasterFilterStaysPostJoin` - splice join with a master-side filter; asserts the results and that the filter stays post-join (`Filter filter: ...`), plus literal-vs-bind parity.
- Adds `JoinTest#testOuterJoinMasterFilterStaysPostJoin` - right/full outer join where the master lacks the key; asserts 0 rows (the previously leaked NULL-master rows are gone) for both the literal and bind forms.
- Updates `SqlParserTest` splice and outer-join model assertions to the post-join model, and adds a two-argument `assertQueryWithOuterJoinType` overload so LEFT keeps the pushed-down filter while RIGHT/FULL expect the post-join filter.
- Updates `ExplainPlanTest#testLeftJoinWithPostJoinFilter` and `SqlOptimiserTest#testOrderByAdviceWithMultipleJoinsAndFilters` for the new plan.
- Corrects `EmptyTableRecordCursorFactoryTest#testFoldedFalseFilterAllowsSpliceJoin`: a splice whose master WHERE folds to FALSE must produce 0 rows, not 3.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/SqlOptimiser.java
+++ b/core/src/main/java/io/questdb/griffin/SqlOptimiser.java
@@ -582,6 +582,32 @@ public class SqlOptimiser implements Mutable {
         model.getJoinModels().getQuick(parent).addDependency(child);
     }
 
+    /**
+     * Returns the join-model index of the outermost SPLICE, FULL OUTER or RIGHT OUTER join
+     * that NULL-extends the master (left) side table at {@code tableIndex}, or -1 when no
+     * such join follows it. Each of these joins emits rows in which the master table's
+     * columns are all NULL: slave-only timestamps for SPLICE, unmatched right rows for FULL
+     * and RIGHT OUTER. A WHERE predicate that references only such a table must therefore be
+     * applied as a post-join filter rather than pushed into the table's own sub-query.
+     * Pushing it leaves those NULL-master rows unfiltered (e.g. {@code a RIGHT JOIN b WHERE
+     * a.k = 'v'} would keep unmatched b rows whose a.k is NULL); for SPLICE it also changes
+     * which master row prevails at each slave timestamp. LEFT OUTER, ASOF and LT keep every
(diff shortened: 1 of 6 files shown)


puzpuzpuz:
Closing in favor of #7225

PR #7219, opened 2026-06-09
ci(build): build async-profiler for Linux aarch64 alongside x86-64

bluestreak01:
## What

- Extends the existing `rebuild_async_profiler.yml` workflow to build the async-profiler artifacts (`libasyncProfiler.so`, `asprof`, `jfrconv`) for **both** Linux x86-64 and Linux aarch64, committing the results for both architectures (`core/src/main/bin/linux-x86-64/`, `core/src/main/bin/linux-aarch64/`) in a single commit. There is no separate aarch64 workflow file; one workflow covers both architectures.
- Bumps the async-profiler submodule from v4.2.1 to v4.4 (released 2026-04-20), the latest stable release.

The branch also carries rebuilt x86-64 binaries from a test dispatch of the pre-fold workflow on this branch ([run 27247119369](https://github.com/questdb/questdb/actions/runs/27247119369)). Those binaries were built from v4.2.1; dispatching the updated workflow rebuilds both architectures at v4.4.

## Why

`core/src/main/bin/linux-aarch64/` currently contains only `libjemalloc.so` — there is no async-profiler counterpart, because the rebuild workflow exists for x86-64 only. As a result, aarch64 Linux distributions are assembled without the profiler, so `questdb.sh -p` (and the `asprof` subcommand) fail there even though the loader itself is architecture-agnostic.

## How it works

- A two-leg build matrix compiles async-profiler natively on `ubuntu-latest` (x86-64) and `ubuntu-24.04-arm` (aarch64; GitHub-hosted, free for public repositories).
- Each leg builds inside async-profiler's own musl builder image (`asprof-builder-x86` / `asprof-builder-arm`) with the same `make CC=/usr/local/musl/bin/musl-gcc release` invocation async-profiler uses for its official releases — verified unchanged at the v4.4 tag, as is the release tarball's `lib/` + `bin/` layout the extraction step expects. Both images are pinned by digest and current: `asprof-builder-x86:latest` resolves to the digest already pinned, and the arm digest was resolved from `:latest` on 2026-06-10.
- The build legs upload the three binaries as run artifacts; a final `commit` job downloads both sets and pushes one commit covering both architectures. The single commit removes the push race two independent commit steps would have, and makes it impossible for the two architectures to land on different async-profiler versions.
- The artifact actions stay on v4: like `checkout@v4` they run on node20, which the Debian buster builder containers (glibc 2.28) support; node24-based action versions would not start there.
- The version step fetches tags before `git describe`: the submodule is cloned shallow (`.gitmodules`: `shallow = true`), so the clone carries no tags and the audit commit message previously showed a bare commit sha ("Built from async-profiler 7f60b4d") instead of the release tag.
- The musl-static result has no glibc dependency, so it runs on any aarch64 Linux host, including AWS Graviton.

No loader changes are required: `questdb.sh` resolves `libasyncProfiler.so`/`asprof` from the distribution `lib/` directory by relative path, and the profiler is only engaged on demand (`-p` / `asprof`), so its absence never affects normal startup. Nothing else in the repository references the async-profiler version.

## How the aarch64 binaries get produced

Because `rebuild_async_profiler.yml` already exists on `master`, the updated workflow is dispatchable on this branch directly (Actions tab, select this branch). Dispatching it here rebuilds both architectures at v4.4 and commits all six binaries to this branch, so the PR can ship workflow, version bump, and binaries together. Alternatively, merge first and dispatch on `master`.

## Tradeoffs / notes

- Every dispatch rebuilds and commits both architectures together; an x86-64-only refresh is no longer possible. That is deliberate (the architectures cannot drift apart), but it does remove per-arch flexibility.
- If either architecture's build fails, nothing is committed for either: the matrix is fail-fast and the commit job depends on both legs.
- The commit job reads the async-profiler version from the build legs' job outputs. With a matrix, the last leg to finish wins; both legs check out the same commit and therefore report the same submodule version, so the value is deterministic.
- The v4.2.1 → v4.4 bump takes effect only when the workflow is next dispatched; merging the PR alone does not change shipped artifacts beyond the x86-64 rebuild already on the branch.
- Scope is Linux aarch64 only. macOS arm64 is not covered — QuestDB ships no async-profiler for macOS on any architecture today, and a macOS build would require a separate Mach-O `.dylib`.

## Test plan

- The workflow parses as valid YAML; the build steps are unchanged from the current workflow on `master`, now parameterized by a two-entry matrix (runner, builder image, output directory).
- The aarch64 leg's runner, builder image, and `make ... release` command match async-profiler's own `linux-arm64` release workflow at the v4.4 tag.
- The pre-fold x86-64 workflow was dispatched on this branch end-to-end (run linked above) and committed rebuilt binaries, exercising the shared build/commit steps.
- End-to-end verification of the fold: dispatch the workflow on this branch and confirm it pushes a single commit updating all six binaries, with the commit message reporting v4.4; the resulting `libasyncProfiler.so` files should report no glibc dependency and load via `questdb.sh -p` on matching hosts.


mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/GroupByHistogram.java|46|46|100.00%|

nwoolmer (approved):

PR #7221, opened 2026-06-09
feat(sql): add kurtosis and skewness aggregates

brunocalza:
## Summary

Adds `kurtosis` and `skewness` methods addressing Gap 4 of https://github.com/questdb/roadmap/issues/120. To be precise, the following functions are added

  | Function | Description |
  | --- | --- |
  | `kurtosis` / `kurtosis_samp` | Sample excess kurtosis |
  | `kurtosis_pop` | Population excess kurtosis |
  | `skewness` / `skewness_samp` | Sample skewness |
  | `skewness_pop` | Population skewness |

**Example**
```sql
SELECT skewness(price), kurtosis(price)
FROM trades
WHERE symbol = 'BTC-USD'
SAMPLE BY 1h;
```
Documentation: https://github.com/questdb/documentation/pull/463

## Implementation details
- The state for each group keeps a running mean and the central moments M2, M3 (and M4 for kurtosis) plus a count, updated per row with [Pébay](https://www.osti.gov/biblio/1028931)'s one-pass online algorithm (an extension of Welford's)
- Partial aggregates merge with the matching pairwise-combine formulas, so the functions support parallel GROUP BY execution
- The streaming approach avoids a second pass over the data and is numerically more stable than the naive sum-of-powers formula.

**Conventions**
- Sample skewness returns `null` for fewer than 3 values; sample kurtosis for fewer than 4
- The sample variants apply the standard bias correction (Fisher's definition)
- Population variants return `null` for an empty group
- All variants return `null` when every observation is equal (zero variance), since the
  statistic is undefined

## Testing
- Unit tests using DuckDB as a reference
- A script outside this PR was used to compare the values we compute with this implementation and `scipy`

### DIFF
--- a/core/src/test/java/io/questdb/test/cairo/fuzz/ParallelGroupByFuzzTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/fuzz/ParallelGroupByFuzzTest.java
@@ -311,6 +311,63 @@ public class ParallelGroupByFuzzTest extends AbstractCairoTest {
     }
 
     @Test
+    public void testParallelApproxPercentileFuzz() throws Exception {
+        Assume.assumeTrue(enableParallelGroupBy);
+        assertMemoryLeak(() -> {
+            final Rnd rnd = TestUtils.generateRandom(LOG);
+            final double percentile = rnd.nextDouble();
+            final int precision = rnd.nextInt(6);
+            final int rowCount = rnd.nextInt(ROW_COUNT) + 1;
+            final int groupCount = rnd.nextInt(10) + 1;
+            final WorkerPool pool = new WorkerPool(() -> 4);
+            TestUtils.execute(
+                    pool,
+                    (engine, compiler, sqlExecutionContext) -> {
+                        sqlExecutionContext.setJitMode(
+                                enableJitCompiler ? SqlJitMode.JIT_MODE_ENABLED : SqlJitMode.JIT_MODE_DISABLED);
+
+                        execute(
+                                compiler,
+                                "CREATE TABLE tab AS ("
+                                        + "SELECT CASE WHEN rnd_int() % 10 = 0 THEN NULL ELSE abs(rnd_long()) % 10_000_000 END AS x, rnd_int() % " + groupCount + " AS g "
+                                        + "FROM long_sequence(" + rowCount + "))",
+                                sqlExecutionContext);
+
+                        final String query = "SELECT g, approx_percentile(x, " + percentile + ", " + precision + ") FROM tab GROUP BY g ORDER BY g";
+
+                        sqlExecutionContext.setParallelGroupByEnabled(false);
+                        try {
+                            TestUtils.printSql(
+                                    engine,
+                                    sqlExecutionContext,
+                                    query,
+                                    sink);
+                        } finally {
+                            sqlExecutionContext.setParallelGroupByEnabled(
+                                    engine.getConfiguration().isSqlParallelGroupByEnabled());
+                        }
+
+                        sqlExecutionContext.setParallelGroupByEnabled(true);
+                        final StringSink parallelSink = new StringSink();
+                        try {
+                            TestUtils.printSql(
+                                    engine,
+                                    sqlExecutionContext,
+                                    query,
+                                    parallelSink);
+                        } finally {
+                            sqlExecutionContext.setParallelGroupByEnabled(
+                                    engine.getConfiguration().isSqlParallelGroupByEnabled());
+                        }
+
+                        TestUtils.assertEquals(sink, parallelSink);
+                    },
+                    configuration,
+                    LOG);
+        });
+    }
+
+    @Test
     public void testParallelAvgDecimal128RescaleOverflowFactoryReuse() throws Exception {
         // Regression: AvgDecimal128Rescale256GroupByFunction.merge's "both
         // shards overflowed into 256 bits" branch must add the running
@@ -1773,6 +1830,32 @@ public class ParallelGroupByFuzzTest extends AbstractCairoTest {
     }
 
     @Test
+    public void testParallelGroupByKurtosis() throws Exception {
+        testParallelGroupByAllTypes(
+                "SELECT round(kurtosis_samp(adouble), 12) FROM tab", """
+                        round
+                        -1.173939788697
+                        """,
+                "SELECT round(kurtosis(adouble), 12) FROM tab", """
+                        round
+                        -1.173939788697
+                        """,
+                "SELECT round(kurtosis_pop(adouble), 12) FROM tab", """
+                        round
+                        -1.173979070703
+                        """,
+                "SELECT key, round(kurtosis_pop(adouble), 12) kp, round(kurtosis_samp(adouble), 12) ks FROM tab ORDER BY key", """
+                        key\tkp\tks
+                        k0\t-1.177368704151\t-1.177195920397
+                        k1\t-1.213464478728\t-1.2135590444339999
+                        k2\t-1.156658901168\t-1.156322164928
+                        k3\t-1.1266060077039999\t-1.126049103395
+                        k4\t-1.172914311098\t-1.172699075325
+                        """
+        );
+    }
+
+    @Test
     public void testParallelGroupByRegrIntercept() throws Exception {
         Assume.assumeTrue(enableParallelGroupBy);
         testParallelGroupByAllTypes("SELECT round(regr_intercept(adouble, along), 14) FROM tab", """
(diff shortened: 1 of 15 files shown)


bluestreak01 (changes requested):
Thanks for this — the math is sound. I re-derived the Pébay online update, the pairwise combine, and all four final statistics by hand, and cross-checked the `merge()` combine against a single-pass reference over 2000 randomized 3-way splits (max abs diff ~1e-13, i.e. pure float roundoff). I found **no correctness, concurrency, or resource bug**. The implementation faithfully mirrors the existing `AbstractStdDevGroupByFunction` family. The requests below are about **test coverage** (one of them important) plus two quick PR-hygiene fixes.

## Must address

### 1. The `merge()` Pébay combine — the most complex new code — has zero test coverage
This is the main one. In both `KurtosisParallelGroupByTest` and `SkewnessParallelGroupByTest` the only non-null values are at `x BETWEEN 1500 AND 1509`. With `CAIRO_SQL_PAGE_FRAME_MAX_ROWS = 64`, those 10 rows (0-indexed 1499–1508) all land inside a **single page frame** (frame 23, rows 1472–1535). So exactly one worker partial is ever non-empty, and every `merge()` call hits only the `nB == 0` early-return or the `nA == 0` straight-copy fast path. The actual combine branch — both `nA > 0` and `nB > 0`, i.e. the `delta3`/`delta4` M3/M4 formulas in `AbstractKurtosisGroupByFunction.merge()` / `AbstractSkewnessGroupByFunction.merge()` — **never executes in any test**, serial or parallel.

The combine math is correct (I verified it numerically), but it's the riskiest code in the PR and currently nothing guards it against a future regression.

**To exercise it, the non-null values for a given group must span multiple page frames** so that more than one worker partial accumulates real data for that group and the partials actually get combined. Please look at how the other group-by functions test parallel execution:
- `core/src/test/java/io/questdb/test/griffin/engine/functions/groupby/StdDevParallelGroupByTest.java` — `testParallelKeyedSparseNulls` shows the keyed `GROUP BY grp` shape (note: spread the non-null rows across frames, not 10 consecutive rows).
- `core/src/test/java/io/questdb/test/cairo/fuzz/ParallelGroupByFuzzTest.java` — the canonical pattern: run the aggregate on **random** data over many groups/frames and assert the parallel result equals the same query with `CAIRO_SQL_PARALLEL_GROUPBY_ENABLED = false`. That parallel-vs-serial cross-check is the robust way to validate the merge.

### 2. Bare `kurtosis()` / `skewness()` aliases are untested
`KurtosisGroupByFunctionFactory` / `SkewnessGroupByFunctionFactory` are separate classpath-scanned classes that only override `getSignature()` and inherit behavior from the `_samp` factory. The StdDev family ships `StdDevGroupByFunctionFactoryTest` for exactly this; this PR drops the equivalent, so if an alias were accidentally pointed at the `_pop` factory nothing would catch it. A small test asserting `kurtosis(x)` == `kurtosis_samp(x)` (and likewise skewness) on the same data covers it.

### 3. No keyed `GROUP BY` test anywhere (serial or parallel)
Every query is a whole-table single-group aggregate; the `*MultiColumn` tests are multi-column / single-group. The per-key `valueIndex`-offset isolation and multi-entry map path are never exercised. A keyed test producing more than one output row (ideally folded into #1) covers it.

### 4. Description claims a null-vs-NaN distinction that QuestDB can't surface
The body says "Population variants return `null` only for an empty group" and "When all observations are equal … the result is NaN". In QuestDB, `Double.NaN` *is* the DOUBLE NULL sentinel — both branches return `Double.NaN` and both render as `null`, so there's no observable difference. Please reword.

### 5. Missing labels
`feat(sql)` adding aggregates should carry at least `SQL` and `New feature`.

## Nice to have (non-blocking)
- **SAMPLE BY fill is untested.** The `_samp` variants deliberately set `getSampleByFlags() = SAMPLE_BY_FILL_ALL` (routing through `setDouble`/`setNull`), but no `SAMPLE BY … FILL(…)` test exercises it. (Heads-up: like the StdDev family, `FILL(LINEAR)` will throw at cursor time because no interpolation method is implemented — pre-existing, just be aware.)
- **No numerical-stability / large-n test.** 4th-order moments have a much larger dynamic range than variance — the whole point of Pébay is stability — yet the largest computed sample is `long_sequence(4)`. Compare `StdDevSampleGroupByFunctionFactoryTest.testStddevSampHugeValues` (1M rows).
- **Test anti-pattern (×4):** the `testXxxAllNull` cases (line 34 of each factory test) are a lone single-statement `assertQuery(...)` wrapped in `assertMemoryLeak(() -> …)` + `.noLeakCheck()`. The builder leak-checks itself — drop the wrapper and `.noLeakCheck()`. (It mirrors the StdDev convention, hence non-blocking.)
- **Perf opportunity (consistent with StdDev, not a regression):** no `computeKeyedBatch` override, so the parallel keyed path pays per-row interface dispatch that the `AvgDoubleGroupByFunction` pattern avoids. The per-row `aggregate()` itself is optimal (one division/row; count via `get`+`put` is actually leaner than StdDev's `get`+`addLong`).
- Stray blank line at `KurtosisGroupByFunctionFactory.java:32` (absent in the Skewness sibling). Infinity inputs untested; the two abstract bases differ only by the M4 slot and could share a base.

Net: solid implementation, correct numerics — the blocking ask is real test coverage for the parallel `merge()` combine (plus the alias/keyed gaps), then the description/label tidy-up.

core/src/test/java/io/questdb/test/cairo/fuzz/ParallelGroupByFuzzTest.java
```diff
@@ -310,6 +310,63 @@ public void testParallelApproxCountDistinctOverCastToSymbol() throws Exception {
         });
     }
 
+    @Test
+    public void testParallelApproxPercentileFuzz() throws Exception {
```
brunocalza, line 314: this is just formatting reordering

brunocalza:
Thanks for the review, @bluestreak01 . I have addressed the blocking items and most of the nice-to-haves.

**Must address**
1. `merge()` combine coverage. Added [`testParallelGroupByKurtosis`](https://github.com/questdb/questdb/pull/7221/changes/27302d59f85a95e6f94eb40be1db5d5b1fc2c2c2#diff-78f440f0d75060030b891712b275f56b11215744488ae0af86b785e8d8ef4f9aR1833) and [`testParallelGroupBySkewness`](https://github.com/questdb/questdb/pull/7221/changes/27302d59f85a95e6f94eb40be1db5d5b1fc2c2c2#diff-78f440f0d75060030b891712b275f56b11215744488ae0af86b785e8d8ef4f9aR1883) to `ParallelGroupByFuzzTest`

    I left the `Assume.assumeTrue(enableParallelGroupBy)` gate off from these tests, so the query can be executed serially and in parallel  and divergences can be found. 

2. Bare aliases. Both fuzz tests assert kurtosis(x)/skewness(x) against the _samp variant on the same data, so an alias accidentally wired to _pop would fail.
3. Keyed `GROUP BY`. Each fuzz test adds a SELECT key, ... `GROUP BY key ORDER BY key query over 5 keys, exercising the per-key map merge that the whole-table queries don't.
4. PR description of null vs. NaN. Reworded the Conventions section.
5. I don't have permissions to add those labels

**Nice to have** 
- Large-n/numerical stability. Added [testKurtosisSampHugeValues](https://github.com/questdb/questdb/pull/7221/changes/27302d59f85a95e6f94eb40be1db5d5b1fc2c2c2#diff-3a6e7d47b6a623dbd7f803c22d0567055505d5e1248b1a1ecadc402f85dfc3b0R92) and [testSkewnessSampHugeValues](https://github.com/questdb/questdb/pull/7221/changes/27302d59f85a95e6f94eb40be1db5d5b1fc2c2c2#diff-5e35a1c405ab8f92632dab62c0068936839fcf0b8df2a0336ab7941441e0ead2R92) 
- Test anti-pattern (×4). Dropped the redundant `assertMemoryLeak(...)` wrapper and `.noLeakCheck()` from the four AllNull tests; they now use the bare self-leak-checking `assertQuery(...).returns(...)`.
- Stray blank line. Removed at `KurtosisGroupByFunctionFactory`.

The other suggested nice-to-haves I have left consistent with the stddev/var family (happy to change if you'd prefer).

bluestreak01 (approved):
Approving after a full independent review pass.

**Correctness (verified independently):**
- Online Pébay/Welford update and the pairwise M2/M3/M4 combine match the reference formulas exactly (signs included). Hand-traced n=1, n=2, n=4 cases against the asserted test values.
- All four final statistics are correct: skewness_pop, kurtosis_pop, adjusted Fisher-Pearson skewness_samp, and kurtosis_samp.
- Division-by-zero guards are complete: n<3 (skew_samp), n<4 (kurt_samp), and m2==0.0 for all four. M2 is provably >=0 and stays exactly 0 across merges of all-equal data, so the zero-variance branch fires correctly even in parallel.

**NULL handling:** the int/long->double cast maps the sentinel NULL to NaN and isFinite() filters it; exercised by the INT *MultiColumn tests. Empty/all-null groups return null.

**Concurrency/resources:** no per-function mutable state; accumulators live in MapValue; the nA==0/nB==0 fast paths avoid the empty-partial NaN-poisoning class of bug (cf. #7160). No native allocation, no leak surface.

**Tests:** the parallel merge-combine path is genuinely exercised by the new fuzz tests (~36 frames x 5 keys, serial-vs-parallel cross-check).

Non-blocking nits left to discretion: missing SQL / New feature labels; the two dedicated *ParallelGroupByTest classes only hit the merge fast paths (combine is covered by the fuzz tests); FILL(LINEAR) on the _samp variants throws at cursor time (pre-existing, consistent with the StdDev family); the two abstract bases could share code.

PR #7223, opened 2026-06-09
fix(core): fix Parquet crashes on malformed files and data corruption on partial reads

glasstiger:
Several Parquet inputs could either abort the whole database process or silently
return corrupted data while being read through `read_parquet()` or a
parquet-backed table. This follows up on #7212, which hardened the same readers
but did not cover these paths.

The process aborts come from malformed or foreign-generated files: each reaches
the Rust parquet code over JNI — the metadata parser or the page decoders — where
a panic or an infallible allocation aborts the JVM with no recovery. They are of
two kinds: a release-mode panic, where a bitpacked bit-width read from the page
falls through to an `unreachable!()`; and infallible allocations sized by
attacker-controlled counts — the file's Thrift metadata list counts (read before
any page) and the page-header counts (`num_values`, `uncompressed_page_size`,
per-element definition levels, and DELTA length-stream counts) — which abort the
JVM when the allocator refuses them. Each is converted to a recoverable error so
the server stays up, and the specific guard message now reaches the caller on the
non-parallel read path, which previously masked it behind a generic "corrupted"
error.

The corruption is separate and affects *well-formed* files, including QuestDB's
own: a partial range read of a multi-block DELTA-length-encoded STRING/VARCHAR/
BINARY column computed the value-bytes offset wrong and returned shifted bytes. It
was found while hardening the same slicer and is fixed here too.

A third, unrelated correctness bug surfaced while testing that fix. The backward
(`ORDER BY ts DESC`) variant of the new partial-range test tripped a pre-existing
off-by-one in `IntervalBwdPartitionFrameCursor.calculateSize()`, which under-counts
the rows an unbounded-low interval (`ts < X`) matches over a partial range. It is
not Parquet-specific, but the partial-range Parquet test is what exposed it, so it
is fixed here as well.

The sections below cover each in turn.

## Thrift metadata list counts size an infallible allocation

Before any page is decoded, the file's Thrift metadata is parsed, and
`TInputProtocol::read_list` / `TInputStreamProtocol::read_list` (in
`parquet-format-safe`) reserved their result `Vec` with an infallible
`Vec::with_capacity(list_ident.size)`. `list_ident.size` is the file-controlled
element count of a Thrift list — the row groups, the column chunks per row group,
and so on. The surrounding protocol byte budget charges only `size_of::<usize>()`
per element while each `P` (e.g. `RowGroup`, `ColumnChunk`) is far larger, so the
budget does not bound the reservation: a small metadata blob can declare a list
large enough to drive a multi-gigabyte allocation, and the failure aborts the
process on the metadata-open path — before any of the page-level guards below can
run.

Both `read_list` variants now reserve fallibly with `try_reserve` (mirroring
`read_bytes`/`read_string`, which were already fallible). The loop pushes exactly
`list_ident.size` elements, so the single reservation is never exceeded and there
is no reallocation. A metadata blob with an oversized list count now surfaces a
recoverable error instead of aborting.

## RLE_DICTIONARY bit width over the target width

`bitpacked::Decoder::try_new` validated `num_bits == 0` and a length overflow,
but not an upper bound. The bitpacked unpackers only handle `num_bits` up to the
target width (32 for `u32`, 64 for `u64`); a wider value fell through to an
`unreachable!()`, which panics in release builds and aborts the process.
RLE_DICTIONARY is the default encoding Arrow, pandas and Spark emit for
low-cardinality columns, and its index bit-width byte is file-controlled, so a
crafted file reaches this through the public read path.

`try_new` now rejects `num_bits > size_of::<T>() * 8` at the single point through
which every bitpacked decode passes (hybrid-RLE, the RLE-dictionary decoders, and
the delta miniblocks), mirroring the existing guards on the delta paths.

## Dictionary page num_values over the buffer size

`BaseVarDictDecoder::try_new` called `Vec::with_capacity(num_values)` before
validating `num_values`. That count comes from the dictionary page header and is
bounded only against the page buffer for size, not in its own right, so it can be
as large as roughly 2.1 billion. A small file could therefore request a
multi-gigabyte allocation up front, and the allocation failure aborts the
process.

`try_new` now rejects a header that claims more values than the buffer can hold
(each value needs at least a 4-byte length prefix), which bounds the count to the
actual input size. The fixed-width dictionary decoders already validated the
count against the buffer for the same reason.

## Dictionary reservations still amplify the (decompressed) buffer

Bounding `num_values` against the buffer caps the count but not the bytes
reserved: each decoder reserves `num_values` entries of its *target* type, which
can be far wider than a value's footprint in the buffer. The var-width slice and
aux entries are 16 bytes each (~4× a 4-byte-minimum buffer entry), and the
fixed-length-byte-array decimal dictionaries amplify up to 32× — a `Decimal256`
target over a 1-byte FLBA value. For a compressed page the buffer is the
*decompressed* one, sized only by `uncompressed_page_size` (up to `i32::MAX` and,
unlike `compressed_page_size`, not bounded by `max_page_size`), so a multi-gigabyte
decompressed dictionary could still drive a tens-of-gigabytes infallible
reservation and abort the process even with the count guard in place.

A new `try_reserve_dict_values` helper reserves with `try_reserve_exact`, and all
the dictionary reservations route through it: the var-width values
(`BaseVarDictDecoder`), the varchar-slice aux entries (`RleDictVarcharSliceDecoder`),
and the six FLBA decimal dictionaries (`PreconvertedFlbaDecimalDict<Decimal8..256>`).
An oversized dictionary now surfaces a recoverable `OutOfMemory` error instead of
aborting, while any dictionary that genuinely fits in memory still decodes as
before. The failure is classified `OutOfMemory`, not `Layout`: the layout is
already validated, so on the write path (a parquet merge under
`ApplyWal2TableJob`) it backs off and retries rather than suspending the table.

## Uncompressed page size drives an unbounded allocation

The dictionary, data-page (V1 and V2) and varchar-dictionary decompressors each
sized their output buffer with `Vec::resize(uncompressed_page_size, 0)` before
decompressing. That size comes from the page header and, unlike
`compressed_page_size`, is not bounded by `max_page_size` on the read path:
`max_page_size` caps the compressed column-chunk, and a highly compressible page
can legitimately decompress to more than that, so the size cannot be
range-checked up front. A header claiming up to `i32::MAX` therefore drove a
multi-gigabyte infallible reservation, and the allocation failure aborts the
process.

A new `resize_decompress_buffer` helper reserves the buffer with `try_reserve`,
and all four decompression sites route through it. An oversized header now becomes
a recoverable `OutOfMemory` error instead of an abort, while any page that
genuinely fits in memory still decodes as before. Two related adjustments ride
along:

- The V2 `is_compressed=false` arm returns the page body as-is and never reads
  the decompress buffer, so it no longer sizes one. It now compares the header's
  `uncompressed_size` against the body length directly, instead of sizing an
  `i32::MAX`-capable buffer only to read its length back.
- `resize_decompress_buffer` is grow-only and deliberately does not `clear()`
  (callers thread one buffer through every page of a file, and clearing would
  memset the whole page on every call only for the decompressor to overwrite it).
  `decompress_varchar_slice_dict` reuses a pooled buffer whose aux entries hold
  pointers into the dict for the whole column-chunk decode, so it now `clear()`s
  before resizing — otherwise a malformed page that under-fills the buffer
  (Snappy/LZ4 do not check the fill) could expose stale tail bytes from an earlier
  page.

## Foreign array files drive def_scratch unbounded

The array decoders accumulate the current row's definition levels in a
`def_scratch` buffer, one entry per element, using an infallible `Vec<u32>`. A
foreign `LIST<double>` file whose elements are present-but-null at the def level
grows `def_scratch` to one entry per element, and the infallible `push`/`resize`
aborts the JVM on OOM. QuestDB's own writer never produces this — it encodes a
missing double as `NaN`, never as a def-level null — so it is a foreign-input
path.

`def_scratch` is now an allocator-checked `AcVec<u32>` whose `push`, `reserve` and
`resize` are fallible, threaded from the column's allocator. An oversized foreign
array now returns a recoverable `OutOfMemory` error instead of aborting.

## DELTA length streams sized by an attacker-controlled count

The DELTA_LENGTH_BYTE_ARRAY and DELTA_BYTE_ARRAY slicers decode their length (and,
for DELTA_BYTE_ARRAY, prefix/suffix) streams up front into a `Vec<i32>`, reserving
by the page's row count. That count is `num_values` from the page header,
attacker-controlled up to `i32::MAX`, and a bit-width-0 delta miniblock expands to
its full declared value count while consuming almost no buffer bytes — so a tiny
page can declare billions of lengths and drive a multi-gigabyte infallible
reservation that aborts the process. The DELTA_BYTE_ARRAY suffix stream was worse:
it carried no row-count bound at all and was collected in full, sized by its own
declared `total_count`.

A new `collect_checked_lengths` helper reserves with `try_reserve_exact` and caps
the iteration at the page's row count (`take(limit)`); all three streams (length,
prefix, suffix) route through it. An oversized foreign page now surfaces a
recoverable `OutOfMemory` error instead of aborting, and the suffix stream is now
bounded by the row count like the others. As with the other reservations the
failure is classified `OutOfMemory` so a parquet merge backs off and retries.

## Partial range reads of multi-block DELTA columns returned shifted bytes

This one is a correctness bug, not a crash, and it affects well-formed files —
including those QuestDB writes itself, which encodes STRING, VARCHAR and BINARY
columns as DELTA_LENGTH_BYTE_ARRAY. Such a page lays out `[length stream][value
bytes]`, and to reach the value bytes the slicer must skip the entire length
stream. It derived that offset from `delta_bitpacked::Decoder::consumed_bytes()`
read *after* a `take(row_count)` over the lengths — correct for a full read, but
wrong for a partial range read where `row_count` is below the page's value count
(an interval predicate that ends inside a row group). `take` stops before entering
the later delta blocks, so `consumed_bytes()` omits their bytes and the value-bytes
slice starts *inside* the length stream: every value then shifts (e.g. `"v1"`
decodes as a string with leading NULs), or a valid page trips a spurious "length
exceeds values buffer" error. It only bites when the length stream spans more than
one block (>128 values) and the read stops short of the page end, which is why
existing STRING tests — a handful of rows, or filtering on the value column (a
different, immune decode path) — never caught it.

The offset now comes from `consumed_bytes()` when the length decoder was exhausted
(a full read, O(1) and self-consistent with the lengths it just produced) and from a
structural walk over the block/miniblock layout (`delta_stream_byte_len` /
`MiniblockIterator::get_end_pointer`) only when `take` stopped early, so it no longer
depends on how many values the caller later reads. The walk is bounded by the buffer
size (every block consumes at least its header bytes), so a foreign page cannot
amplify it into an unbounded loop. Both the length stream (DELTA_LENGTH_BYTE_ARRAY)
and the prefix/suffix streams (DELTA_BYTE_ARRAY — foreign-only today, but carrying
the same latent offset bug) are located this way.

The walk counts only the miniblocks that actually hold values
(`ceil(values_in_block / miniblock_size)` per block), not every miniblock. The
format writes a bit-width byte for every miniblock but a body only for the populated
ones, and leaves the trailing miniblocks' bit-width bytes arbitrary; a page with more
than one miniblock per block (Arrow and Spark emit four, QuestDB emits one) would
otherwise have the walk add phantom bodies for the unused trailing miniblocks and
overshoot the length stream — shifting the values the other way, or tripping a
spurious "exceeds page size" error, on a partial read. `get_end_pointer` derives the
last block's populated count from the header, so the bound holds whether it is called
fresh or mid-decode; this also corrects the VARCHAR-slice decode path, which already
used `get_end_pointer` and carried the same over-count.

`MiniblockIterator::try_new` now also rejects a block size that is not a multiple of
128, matching the vendored parquet2 length decoder and the parquet spec so the two
parsers — parquet2 reading the lengths, `get_end_pointer` locating their end — agree
on which headers are valid instead of diverging on a foreign page.

## Backward interval scans miscount rows over a partial range

The backward-order test above (`ORDER BY ts DESC`) drives
`IntervalBwdPartitionFrameCursor`, whose `calculateSize()` — the row count behind
`count()`, certain `LIMIT` optimisations, and the test framework's size
cross-check — under-counted an unbounded-low interval (a `ts < X` / `ts <= X`
predicate, where `intervalLo` is `Long.MIN_VALUE`). It accounts a frame as
`(lo, hi]` with `lo` the exclusive lower index, so the size is `hi - lo` and `lo`
must be `-1` when the interval reaches below row 0. The ternary returned `0`
instead, which drops row 0; and because `0` also routes to "skip interval" rather
than "skip partition", the walk stops after the first matching partition, so a
multi-partition scan can miss whole partitions. `next()` — the iteration path — is
unaffected, so the size disagreed with the rows the cursor actually returns.

The fix sets `lo` to `-1` for the unbounded-low case, matching both the
`findTimestamp()` value the bounded branch already produces and the forward
cursor's accounting. The bug is general, not Parquet-specific: any
`... WHERE ts < X ORDER BY ts DESC` could mis-`count()` through this cursor.

## Surfacing the underlying decode error

The crash-class fixes above turn an abort into a `CairoException` carrying a
specific message — the bit-width guard's `... exceeds ...`, the dictionary guard's
`too short to hold ...`, and so on — but the non-parallel read path was throwing
that message away. The single-threaded `ReadParquetRecordCursor` caught the native
decode error and rethrew a generic "Error reading. Parquet file is likely
corrupted", discarding the cause, so the very guard messages this PR adds never
reached the SQL layer or the user. (The parallel reader already propagated the
cause.)

Both masking sites — `hasNext()` (via `switchToNextRowGroup()`) and `skipRows()` —
now preserve it. `ex` shares the thread-local `CairoException` flyweight that
`nonCritical()` resets, so the underlying message is copied to a `String` first and
then appended: "Error reading. Parquet file is likely corrupted: <cause>". The
server still stays up; the difference is the user now sees which guard fired instead
of a bare "corrupted" string.

## Effects and tradeoffs

- Malformed or foreign inputs that previously aborted the process now return a
  recoverable error and leave the server running.
- The recoverable error now carries the specific guard message end-to-end on both
  read paths: the non-parallel `ReadParquetRecordCursor` previously masked the
  native cause with a generic "Parquet file is likely corrupted" string and now
  appends it, so a user sees which check failed.
- A partial range read of a multi-block DELTA-length-encoded STRING/VARCHAR/BINARY
  column now returns the correct bytes; previously it could return shifted/corrupted
  values or trip a spurious "length exceeds values buffer" error. This corrects
  output for QuestDB's own well-formed files, not only foreign ones.
- `count()` and `LIMIT` over a backward (`ORDER BY ts DESC`) interval scan with an
  unbounded-low predicate (`ts < X`) now report the correct row count; previously
  `IntervalBwdPartitionFrameCursor.calculateSize()` under-counted and, across
  partitions, could drop whole partitions. This fix is general, independent of the
  Parquet changes.
- The bit-width and num_values guards each add a single comparison per page; the
  reservation changes swap infallible reservations for fallible ones; the DELTA
  offset reads the length decoder's own byte count on a full read and walks the
  block structure (bounded by the buffer) only on a partial range read. The cost on
  the valid-file path is negligible in all cases.
- Allocation failures on these paths are classified `OutOfMemory`, not `Layout`,
  so the write path (parquet merges under `ApplyWal2TableJob`) backs off and
  retries on transient memory pressure rather than suspending the table.
- Behavioral change: a subset of already-malformed dictionary pages now fail with
  "too short to hold N values" instead of the previous per-element "too short to
  read value" message. Both are errors; only the message and the point of failure
  differ.
- Behavioral change: a foreign DELTA page whose header declares a block size that is
  not a multiple of 128 (which the parquet spec disallows and QuestDB never writes)
  now fails with a clean "must be a multiple of 128" error instead of being decoded
  on a best-effort basis.
- The fallible reservations change the *failure mode*, not the allocation size: a
  valid file with a genuinely large dictionary or a highly compressible page still
  reserves in proportion to its real contents, because neither `num_values` (past
  the buffer guard) nor `uncompressed_page_size` can be range-checked up front.
  Only a process abort becomes a recoverable error.
- No change to decoded values for well-formed files other than the DELTA
  partial-read fix above; the backward-scan count fix is the only other behavioral
  change, and it corrects row counts rather than column values.

## Test plan

- Thrift metadata `read_list`: no dedicated fixture. The change is a structural
  mirror of the already-fallible `read_bytes`/`read_string` reservations, and the
  happy path is exercised by every parquet read in the suite, since row-group and
  column-chunk lists are parsed through `read_list` on every file open. The
  failure path converts an allocation refusal into a propagated thrift read error
  (`try_reserve`'s `?`) instead of an `abort`.
- Bit-width guard: a unit test on `bitpacked::Decoder::try_new` rejecting widths
  33/40/255 (`u32`) and 65 (`u64`) and accepting the 32/64 boundaries, plus
  width-40 regression tests driven through every consumer of the guard —
  hybrid-RLE, both RLE-dictionary decoders, the RLE-dictionary slicer, and
  end-to-end through `decode_page`.
- num_values guard: tests that a dictionary header with `num_values = i32::MAX`
  over a tiny buffer is rejected with the distinct "too short to hold" message,
  that `num_values == buffer.len() / 4` is accepted, and that one-over is rejected
  (pinning the bound against an off-by-one), plus the same rejection end-to-end
  through `decode_page`.
- Fallible dictionary reservation: a `try_reserve_dict_values` test driving
  `usize::MAX` so `try_reserve_exact` fails deterministically without a real
  allocation, asserting a clean error classified `OutOfMemory`; and a
  `Decimal256`-over-1-byte-FLBA test (plus an all-widths variant) exercising the
  32× amplification call site through a valid dictionary.
- Decompression buffer: a `resize_decompress_buffer` test driving `usize::MAX` for
  a deterministic `OutOfMemory`, a grow-in-place test pinning that it does not
  `clear()` (preserves existing bytes, zeroes only the grown tail), a test that
  the V2 `is_compressed=false` arm allocates no buffer yet still rejects a lying
  size, V1/V2/dict round-trip and lying-size tests, and a
  `decompress_varchar_slice_dict` test pinning the load-bearing `buf.clear()` (an
  under-filling page must read back zeros, not stale pool bytes).
- DELTA length-stream allocation: a `collect_checked_lengths` test driving
  `usize::MAX` for a deterministic `OutOfMemory`, and a test pinning that the count
  is bounded at `limit` (the DELTA_BYTE_ARRAY suffix stream's previously-unbounded
  collect).
- DELTA partial-range offset: Rust slicer tests reading a prefix of a multi-block
  DELTA_LENGTH_BYTE_ARRAY page (constant- and varying-length values) and a
  DELTA_BYTE_ARRAY page, asserting the prefix matches a full decode of just those
  values; the same end-to-end through `decode_page` for STRING and VARCHAR,
  including a multi-block page with nulls in the cut range that pins the row count
  bounding the decode against the smaller non-null length count; a hand-built
  foreign page with four miniblocks per block and non-zero unused trailing bit
  widths driven through `DeltaLengthArraySlicer` on a partial read, plus a
  `get_end_pointer` unit test asserting its offset equals the parquet2 decoder's
  `consumed_bytes()` exactly for such pages (the two pre-existing `get_end_pointer`
  tests strengthened from loose pointer bounds to that exact check); and two Java
  `ParquetRowGroupPruningTest` tests (`testIntervalScanStringMultiBlockPage` and its
  `ORDER BY ts DESC` backward-cursor counterpart) reading a 300-row STRING column
  converted to parquet over an interval that cuts the row group, asserting the
  un-shifted values.
- DELTA block-size validation: a unit test that `MiniblockIterator::try_new` rejects
  a block size that is not a multiple of 128.
- Backward-scan size: the `ORDER BY ts DESC` test above also surfaced the
  `calculateSize()` off-by-one. A new parameterized (native + parquet)
  `IntervalBwdPartitionFrameCursorTest.testCalculateSizeUnboundedLowIntervalAcrossPartitions`
  asserts `calculateSize()` equals the iterated row count for an unbounded-low
  interval that cuts one partition and wholly includes another. Verified both ways:
  reverting the one-line fix makes it report 4 where 29 rows match. No prior test
  drove `calculateSize()` on this cursor, which is why the bug survived.
- Array `def_scratch`: a foreign 1D/2D array of present-but-null elements decodes
  cleanly (unfiltered and filtered), and the same under a memory limit returns a
  clean `OutOfMemory` error rather than aborting, with the error path asserted
  leak-free.
- End-to-end fixtures: two committed broken parquets read through `read_parquet()`
  in `ReadParquetFunctionTest`, each asserted on both the parallel and non-parallel
  readers (parameterized) to surface the guard's specific message rather than abort
  the JVM. `rle_dict_index_bitwidth_over_32.parquet` (a valid dictionary-encoded
  INT32 column with only its index bit-width byte patched to 40) must report
  "exceeds"; `dict_num_values_over_buffer.parquet` (a dictionary page whose
  `num_values` overruns its buffer) must report "too short to hold". The
  non-parallel assertions also pin the error-surfacing fix — reverting the message
  preservation fails both tests' `parallel=false` runs. Ignored Rust generators
  (`generate_rle_dict_index_bitwidth_fixture`,
  `generate_dict_num_values_over_buffer_fixture`) regenerate and re-verify the
  fixtures.
- Verified both ways where a revert reproduces the failure: reverting the
  bit-width guard makes its tests hit `unreachable!()` and abort, and they pass
  with the guard in place; reverting the DELTA offset fix makes the partial-range
  tests return shifted bytes, and reverting the populated-miniblock bound makes the
  multi-miniblock offset and slicer tests fail by over-counting the unused trailing
  miniblocks. The allocation-failure tests are driven with
  `usize::MAX` or a memory limit so they fail deterministically without a real
  multi-gigabyte allocation; they prove abort-freedom of the fallible path (for
  `def_scratch`, a revert shifts the failure to a sibling fallible reservation
  rather than aborting, so that test asserts abort-freedom rather than
  fail-on-revert). All error-path tests assert no memory leak across JNI.
- `cargo test --lib` for the affected modules and `cargo clippy --lib` pass with
  no new warnings.


bluestreak01:
@glasstiger — code review (mission-critical pass, level 2). Net: the three in-diff guards are correct and well-tested; I verified the bitpacked guard makes `decode_pack`'s `copy_from_slice` provably in-bounds for every admissible `num_bits`, the dict guard rejects no valid page, and `try_reserve`-before-`resize` cannot abort. The both-ways revert check reproduces the original panics. What needs attention: one performance regression the PR introduces, and two abort paths of the **identical class** the PR targets that remain reachable from the same `read_parquet()` entry — the cheaper needing only a sub-kilobyte file.

## Critical (out-of-diff, pre-existing, same class as this PR's target)

This PR does not introduce these, but it explicitly continues #7212's malformed-file hardening, and both are JVM-abort-on-malformed-input bugs reachable from the read/merge paths.

### C1 — `array.rs` `def_scratch` is an infallible `std::Vec`, abortable from a ~1 KB file

`core/rust/qdbr/src/parquet_read/decode/array.rs:121` (also `:284`, `:334`): `let mut def_scratch = Vec::new();` is a plain `std::Vec<u32>`, not the fallible `AcVec` used for every other buffer there. It is grown by infallible `reserve`/`resize`/`push` at `array.rs:416-420` (1D) and `:522-526` (2D). A `std::Vec` allocation failure calls `handle_alloc_error` → process abort → JVM down across JNI.

Reachability (traced):
- `infer_column_type` (`meta.rs:97`) routes any `Double` leaf under a `LIST` group to a QuestDB Array; `array_column_type` (`meta.rs:303-318`) keys only off the parquet schema's `List` type, **independent of QuestDB metadata** (which is discarded for foreign files, `meta.rs:157`). A foreign `LIST<double>` file reaches `decode_array_page` → `read_and_append_one_row_1d`.
- `element_count` grows once per rep/def level until `rep == 0`. The level stream is hybrid-RLE from attacker page bytes, bounded only by the header's `num_values` (i32, up to ~2³¹). A single RLE run of `rep=1` × 2³¹ is ~6 bytes; a first `def < max_def_level` flips `has_nulls`, so every level does `def_scratch.push(def)`.
- One fabricated array row drives `def_scratch` to ~2³¹ × 4 B ≈ 8 GiB of infallible allocation from a tiny file — the cheapest trigger of the whole bug class, cheaper than either bug this PR fixes.

Fix: make `def_scratch` an `AcVec`, or `try_reserve`-guard the growth and bound `element_count` against the level-buffer length. Suggest folding into this PR.

### C2 — `PreconvertedFlbaDecimalDict::try_new_*` infallible `Vec::with_capacity`, 32× amplified

`core/rust/qdbr/src/parquet_read/decode/decimal.rs:920,934,948,962,976,992`: each `try_new_decimal{8..256}` calls `Vec::with_capacity(dict_page.num_values)` (infallible). `validate_flba_dict` (`decimal.rs:1011`) enforces only `src_len * num_values == buffer.len()`, and `decode_fixed_decimal_dict_mode` (`decimal.rs:107`) validates `src_len` only as `0 < src_len ≤ 32` — **never against target precision**. So `FixedLenByteArray(1)` + a Decimal logical type of precision ≥ 39 (→ Decimal256, 32 B/elem) gives `num_values == buffer.len()` and a `buffer.len() × 32` reservation. A dict page decompressing to ~2 GiB (now succeeding through the new fallible `resize_decompress_buffer`) → ~64 GiB infallible reservation → abort.

Lower than C1 because it requires a genuine multi-GiB decompressed dict buffer first. Fix: `try_reserve`, or a `num_values`-vs-buffer guard mirroring the one this PR adds to `BaseVarDictDecoder`.

## Moderate

### M1 — `resize_decompress_buffer` adds a full-page memset to every compressed page (in-diff regression)

`core/rust/qdbr/src/parquet_read/decode.rs:2131-2142`. The old `decode.rs` sites were `buffer.resize(size, 0)` **with no `clear()`** — for a buffer reused across pages (`decompress_buffer`/`dict_decompress_buffer` live in `DecodeContext`, created once per file and threaded through every page), `resize` zeroed only the growth delta, amortizing to ~one page-fill per file. The new helper `clear()`s first, so it memsets the entire `uncompressed_page_size` on every compressed page — which the decompressor immediately overwrites. At defaults (1 MiB pages, LZ4_RAW) that's ~25-60 µs of dead writes per page against a ~200-300 µs decompress (~+10-25% on the per-page decompress stage), paid in full on the filtered path even when few rows are read. (`row_groups.rs:131` is not regressed — it already did `clear(); resize`.)

The `clear()` isn't needed for fallibility. Grow-only fix keeps every safety property and the old write count:

```rust
pub(super) fn resize_decompress_buffer(buffer: &mut Vec<u8>, size: usize) -> ParquetResult<()> {
    if size > buffer.len() {
        buffer.try_reserve(size - buffer.len()).map_err(|_| {
            fmt_err!(Layout, "cannot allocate {} bytes for a decompressed page", size)
        })?;
    }
    buffer.resize(size, 0);
    Ok(())
}
```

When growing, `try_reserve(size - len)` guarantees capacity ≥ size so `resize` can't reallocate or abort; when shrinking it truncates. The `usize::MAX` rejection test still passes. The only behavioral delta is that a malformed *under-filling* Snappy/LZ4 page (those codecs don't validate full fill — `parquet2/src/compression.rs:165-187`) leaves stale prior-page bytes in the tail instead of zeros — the pre-existing, same-file-scoped behavior. If deterministic tails are wanted, validate decompressed fill in vendored `parquet2::compression::decompress` rather than memsetting every good page.

### M2 — V2 `is_compressed=false` sizes + memsets a buffer only to compare its length (in-diff, worsened)

`decode.rs:2178` runs `resize_decompress_buffer(...)` before `can_decompress` is read at `:2181`. In the `is_compressed=false` branch (`:2196-2204`) the buffer is never read — only `decompress_buffer.len() != page.buffer.len()` is checked, then `page.buffer` is returned. So it allocates+memsets up to `uncompressed_size` (attacker-controlled to i32::MAX, unbounded by `max_page_size`) purely to compare two integers, with no decompression to amortize against. Writers like arrow-rs emit `is_compressed=false` per page when a page doesn't compress. Fix: hoist sizing into the branches — in the uncompressed arm compare `page.uncompressed_size != page.buffer.len()` directly; size the buffer only in the V1 and V2-compressed arms.

### M3 — allocation failure misclassified as `Layout` instead of `OutOfMemory` (in-diff)

`decode.rs:2133-2139` maps `TryReserveError` to `fmt_err!(Layout, ...)`, bypassing `impl From<TryReserveError> for ParquetError` (`error.rs:208`) which produces `OutOfMemory`. It's the only manual `TryReserveError` mapping in the tree; the other 14 sites use `?`. `into_cairo_exception()` then sets `out_of_memory(false)`. Inert on the pure read path, but real on the write path — parquet merges run under `ApplyWal2TableJob` via `O3ParquetMergeContext`/`ParquetPartitionDecoder`, and `ApplyWal2TableJob:613` branches: `isOutOfMemory()` true → `onOutOfMemory()` backoff; false → `suspendTable()` (`:642`). So a transient host-OOM during a parquet merge now suspends the table where an OOM-classified error would back off and retry. Still better than the pre-PR abort, but the fix is one line: `ParquetError::with_descr(ParquetErrorReason::OutOfMemory(None), format!("cannot allocate {size} bytes for a decompressed page"))` (and amend the "recoverable `Layout` error" doc wording). Every reachable `size` here is ≤ i32::MAX via `slice_reader.rs:140`, so the failure is always allocation, never malformed layout — `OutOfMemory` is also semantically correct.

### M4 — wrong reachability claim in a safety test's comment (in-diff)

`core/rust/parquet2/src/encoding/hybrid_rle/mod.rs:307-310`: the new comment says "A foreign file can declare a level bit width wider than 32." False — level widths are computed, never file-declared: `get_bit_width(max_level: i16) = 16 - leading_zeros` (`parquet2/src/read/levels.rs:3`) caps at 16, and the one caller passing a file byte pre-checks `bit_width > 32` (`deserialize/utils.rs:21`). The genuinely file-controlled >32 path is the RLE_DICTIONARY index-width byte (covered by the qdbr tests). The test is fine as a defense-in-depth check; the comment misattributes the vector and should be reworded.

### M5 — missing test coverage for the changed decompression paths (in-diff)

All 10 new tests pass (158 parquet2 + 958 qdbr); clippy/fmt clean on qdbr. Gaps:
- **No Java end-to-end malformed-parquet test** for either fixed bug. Existing malformed coverage (`sqllogictest/.../broken/`) is structural and never reaches page decode. The abort-vs-clean-error contract is proven only at the Rust unit level. A ~1 KB crafted fixture through `ReadParquetFunctionTest` would close it.
- `decompress_sliced_data`/`decompress_sliced_dict` have **zero** coverage — no test constructs a `SlicedDataPage`; the V2 arm and a compressed page with a lying `uncompressed_size` are untested. The new helper is only unit-tested with `usize::MAX`.
- `resize_decompress_buffer` happy-path growth/shrink/reuse-with-prior-contents is untested — exactly the semantics M1 turns on.

## Minor

- **`BaseVarDictDecoder` still uses infallible `Vec::with_capacity(num_values)`** (`dictionary.rs:67`). The guard caps it at 4×buffer (closing the "i32::MAX from a tiny file" route) but on a multi-GiB decompressed dict it can still abort infallibly, as can `dict_aux` at `rle_dict_varchar_slice.rs:40`. Same class as C2, gated behind a large buffer.
- **`block_size` multiplication precedes the guard** (`bitpacked/decode.rs:31` vs `:45`). `size_of::<T>() * num_bits` runs before the bound check; absurd `num_bits` overflows (debug panic / release wrap). Unreachable today (callers cap at one byte ≤255), but hoisting the guard above line 31 makes it self-contained.
- **"with_capacity is lazy" comment is inaccurate** (`dictionary.rs:294`, `decode.rs:4240`). `with_capacity` allocates eagerly; the non-determinism comes from allocator overcommit (which the comment's own tail states correctly). Reword.
- **Pre-existing stale comment** at `rle_dict_varchar_slice.rs:1661` (`test_dict_value_exceeds_28bit_header`): "BaseVarDictDecoder catches the short buffer" is wrong — the test errors on the 28-bit length check. Since the PR touches this file, worth a 2-line fix.
- **New tests can't detect leaks**: `TestAllocatorState` is a passive counter (no `Drop`/panic-on-leak) and no new test asserts `rss_mem_used() == 0`. Leak-freedom on the new error paths is real (verified by ownership tracing — no leaks/dangling pointers), but the tests don't enforce it. Add `drop(bufs); assert_eq!(tas.rss_mem_used(), 0);` to the two `decode_page` tests.
- **PR metadata**: zero labels — add `Bug, Core, rust, storage` (the set #7212 carried). The body documents only 2 of 3 functional changes — the `resize_decompress_buffer`/`try_reserve` change (third crash vector) is absent, and since PRs squash-merge the body is the permanent record. The intro ("Two malformed... files could abort") and the "Each guard is a single comparison per page" bullet are now stale.

## Verdict

Request changes. The in-diff guards are sound for the three vectors they scope. Before merge: fix the **M1** memset regression (free, one-line, preserves all safety) and coupled **M2**; fix **M3** classification (one line, real WAL-apply consequence); and decide on **C1** — an abort of the identical class, reachable from `read_parquet()` on a sub-kilobyte foreign file, cheaper than either bug fixed here. C1 (ideally C2 too) belongs in this PR or an immediate fast-follow; shipping a "fix crashes on malformed Parquet" PR that leaves the cheapest crash open undercuts its own goal. M4-M5 and the Minors (labels, body, comments, Java e2e test) should also land.

14 findings verified, 0 false positives. In-diff: 10 (M1-M5 + 5 Minors). Out-of-diff: 2 Critical (C1, C2) — the highest-severity bugs are in files the diff never touches but that share the JNI read entry it hardens.

glasstiger:
## Code review (mission-critical pass, level 3)

Reviewing the current state of the branch (the post-feedback revision). Net: **the three classes of fix are correct and well-tested**, and I could not break them. I verified the silent-corruption fix is byte-equal to parquet2's own offset for full reads (differential test, n=1..1000) and reproduced the original corruption by reverting it; the bitpacked guard makes every bitpacked decode in-bounds; and every infallible allocation this PR touches is now recoverable and classified consistently. All findings below are **Minor** — optional polish, none block merge.

This revision already incorporates the earlier level-2 review's C1/C2/M1–M5 + Minors (def_scratch→AcVec, FLBA/var-dict→`try_reserve_dict_values`, grow-only `resize_decompress_buffer`, V2-uncompressed hoist, `OutOfMemory` classification, reworded reachability comment, added round-trip/e2e tests, labels). I re-verified each against the current source.

### What I verified as correct (so it need not be re-checked)

- **DELTA partial-read offset (the only non-crash bug; affects QuestDB's own STRING/VARCHAR/BINARY files).** `delta_stream_byte_len` walks the block structure via `MiniblockIterator::get_end_pointer` instead of `consumed_bytes()`-after-`take`. Proven byte-equal to parquet2's `Decoder::consumed_bytes()` for full reads across n=1..1000 (all block counts, zero/non-zero bit widths); structurally bounds-safe (`end` is always within `[data, data+len]`, so `end - data.as_ptr()` ∈ `[0, len]` — no slice panic, no underflow); `DeltaBytesArraySlicer` prefix/suffix boundaries and the now-`row_count`-bounded suffix collect are correct. Reverting the fix reproduces the exact shifted-bytes corruption (`"v1"` → leading-NUL string) at both the slicer and `decode_page` layers.
- **Bitpacked `num_bits > size_of::<T>()*8` guard.** The `>` boundary admits exactly 32/64 (the `seq!(0..=$bits)` arms exist); it is the single choke point for `parquet2::bitpacked::Decoder`. qdbr's *separate* `unpack` (used by the delta decoders) has its own pre-existing `> 32`/`> 64` guards, so no `unreachable!()` path remains. The `size_of::<T>() * num_bits` at `decode.rs:31` runs before the guard but provably cannot overflow on 64-bit (callers bound `num_bits` ≤ u32::MAX).
- **Fallible reservations** (`read_list` try_reserve, `try_reserve_dict_values`, `collect_checked_lengths`, `resize_decompress_buffer`, `def_scratch`→`AcVec`, `BaseVarDictDecoder` num_values guard): correct integer math, push-not-index, and a consistent split — allocation shortfalls on potentially-valid-large inputs are `OutOfMemory` (so `ApplyWal2TableJob` retries rather than suspends), structural/corrupt inputs are `Layout`. M3 traced through to `ApplyWal2TableJob.handleWalApplyFailure`. `AcVec` resize/reserve/push/clear are semantically identical to std `Vec` plus fallibility; `clear()` retains capacity (no per-row hot-loop regression).
- **`resize_decompress_buffer` grow-only / no-clear** introduces no new stale-tail exposure — byte-identical to the old bare `resize(size, 0)` at the three decode.rs sites; the one pointer-retaining path (`decompress_varchar_slice_dict`) correctly keeps its `buf.clear()`. The V2 `is_compressed=false` check (`page.uncompressed_size != page.buffer.len()`) is exactly equivalent to the old post-resize length compare, minus the wasted i32::MAX-capable allocation.

### Minor

1. **Inaccurate comment "Snappy/LZ4 do not check the fill"** (`decode.rs` doc comment on `resize_decompress_buffer`, and the `row_groups.rs` test comment). Verified against the codec sources: only **Snappy** silently under-fills; **LZ4_RAW, Zstd, Gzip all error** on under-fill. The `clear()` reasoning is still correct (only Snappy needs it), but the comment overstates the set and could mislead later. Reword to name only Snappy.
2. **Missing positive boundary test for u64 widths in (32, 64].** `try_new_num_bits_over_target_width_is_error` pins u32-rejects-33/40/255, u64-rejects-65, and 32/64 accepted, but never asserts `Decoder::<u64>::try_new(.., 40, ..).is_ok()`. One line.
3. **Two error-path tests omit the leak assertion their siblings use.** `rle_dict_varchar_slice.rs::test_bitwidth_over_32_errors` and `rle_dictionary.rs::test_rle_dict_decoder_bitwidth_over_32_errors` allocate tracked `ColumnChunkBuffers` then take the error path but don't `drop(buffers); assert_eq!(tas.rss_mem_used(), 0)`. Same guard's leak-freedom is covered e2e elsewhere, so incremental.
4. **Weak assertion in the non-parallel arm of `testRleDictionaryIndexBitWidthOver32`.** The non-parallel reader replaces the original `CairoException` with a generic "…likely corrupted" string, so `assertContains(message, "corrupted")` proves abort-freedom but not that the bit-width guard specifically fired (any decode error matches). The parallel arm pins `"exceeds"` and the Rust fixture test pins the exact message, so acceptable — just weaker than it reads.
5. **Java multi-block tests depend on *omitting* a property (fragile).** `testIntervalScanStringMultiBlockPage[Backward]` rely on NOT setting `CAIRO_PARTITION_ENCODER_PARQUET_ROW_GROUP_SIZE` so 300 rows form one multi-block page; sibling tests set it per-method. If a global row-group-size ever lands in `setUp`, these silently stop testing the multi-block path without failing. An explicit `setProperty(..., <large>)` would make the intent durable.
6. **Thrift `read_list` has no test** (body admits it). The only fully-untested production change; a structural mirror of the already-fallible `read_bytes`/`read_string`. A single malformed-metadata fixture (footer declaring a `RowGroup`/`ColumnChunk` list count near `i32::MAX`) through `ParquetDecoder::read` would give it an actual regression guard for the metadata-layer abort vector.
7. **`try_reserve_dict_values` failure path only happy-tested at call sites** (acceptable). The reservation-failure path rests on the standalone helper test; the 6 FLBA-decimal / dict_aux sites are happy-path only — defensible because `validate_flba_dict` pins `num_values` to buffer length, so a test can't overflow the reservation there.

### Out-of-diff observation (pre-existing, not introduced here)

- **`get_end_pointer` over-count risk for foreign `num_mini_blocks > 1` pages.** The walk steps over all `miniblocks_per_block` miniblocks; a foreign final block carrying non-zero bit-width bytes for trailing *unneeded* miniblocks could add phantom body bytes parquet2's value decoder (stops at `remaining == 0`) does not. **Foreign-only** (QuestDB writes `mini_blocks = 1` — production files immune, confirmed), **pre-existing** (#7212's primitive), **non-crashing** (bounded wrong-offset). This PR widens the primitive's use to the length/byte-array slicers but does not introduce the edge case. Flagging for awareness.

### Downgraded (false positives, dismissed after verification)

- **Division-by-zero at `hybrid_rle/mod.rs:46` (`packed.len()*8/num_bits`).** Unreachable: `Decoder::next()` returns `None` when `num_bits == 0` (`decoder.rs:30-32`) *before* emitting any run, so `read_next`'s `Bitpacked` arm (which holds the division) is never reached with `num_bits == 0`. The one read-path caller that can pass 0 (`decode.rs:2270`, `get_bit_width(max_rep_level)` for a foreign LIST array) is doubly safe — a `max_rep_level == 0` column has an empty rep-levels buffer, also yielding `None`.
- **"`read_list` metadata failure misclassified (non-OOM) could suspend the table on merge."** For a valid file the row-group/column-chunk lists are tiny — `try_reserve` there cannot fail under transient pressure; it only fails for a genuinely malformed huge count, where suspend-not-retry is correct. The `OutOfMemory` asymmetry the PR applies elsewhere is justified precisely because those reservations can be legitimately large; `read_list` has no "valid-but-large" case.
- **"TestAllocatorState is a passive counter, leak checks meaningless"** (carried from the prior review). The explicit `assert_eq!(rss_mem_used(), 0)` after `drop(...)` is a real error-path leak check — every `track_allocate`/`track_deallocate` updates the atomic.

### Verdict

**Approve.** No correctness, concurrency, resource-leak, or reachable-panic issue in the changed code; the silent-corruption fix (highest severity, affects well-formed files) has the strongest, genuinely fail-without-fix coverage. 9 findings verified (all Minor — 6 in-diff test/comment polish, 1 acceptable-as-is, 1 pre-existing out-of-diff observation, 1 metadata note), 3 false positives removed. In-diff: 6, out-of-diff: 1.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) — level 3 review</sub>

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 1010 / 1120 (90.18%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/parquet_read/decode.rs|425|516|82.36%|
|:large_blue_circle:|qdbr/src/parquet_read/row_groups.rs|13|15|86.67%|
|:large_blue_circle:|parquet2/src/encoding/hybrid_rle/mod.rs|8|9|88.89%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/mod.rs|18|20|90.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/rle_dict_varchar_slice.rs|26|28|92.86%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/rle_dictionary.rs|17|18|94.44%|
|:large_blue_circle:|qdbr/src/parquet_read/slicer/tests.rs|101|105|96.19%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/delta_binary_packed.rs|82|85|96.47%|
|:large_blue_circle:|qdbr/src/parquet_read/decode/array.rs|165|169|97.63%|
|:large_blue_circle:|io/questdb/cairo/IntervalBwdPartitionFrameCursor.java|1|1|100.00%|
|:large_blue_circle:|parquet-format-safe/src/thrift/protocol/mod.rs|2|2|100.00%|
|:large_blue_circle:|parquet2/src/encoding/bitpacked/decode.rs|20|20|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decode/decimal.rs|56|56|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/dictionary.rs|37|37|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/delta_length_array.rs|2|2|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/slicer/mod.rs|37|37|100.00%|

glasstiger:
## Code review (level 3, mission-critical pass)

Reviewed the full change surface across correctness, Rust panic/overflow safety, native-memory leaks across JNI, cross-context callers, concurrency, performance, test coverage, and code quality, plus an independent fresh-context adversarial pass. The `get_end_pointer` core was additionally fuzzed against parquet2's `consumed_bytes()` (135 cases: `miniblocks_per_block ∈ {1,2,4,8}`, 0-width populated and garbage-width trailing miniblocks) — exact match throughout.

**Verdict: approve.** No Critical or Moderate findings. The crash-class hardening, the partial-read corruption fix, and the `calculateSize` off-by-one are all correct, leak-free, and tested. Both behavioral fixes are proven both ways (reverting each reproduces the failure). Only minor polish remains.

### Critical
None.

### Moderate
None.

### Minor

1. **`collect_checked_lengths` over-reserves on null-bearing DELTA columns** — `core/rust/qdbr/src/parquet_read/slicer/mod.rs`. `try_reserve_exact(limit)` reserves `limit = row_hi` (total **rows** in range), but a DELTA_LENGTH / DELTA_BYTE_ARRAY length stream holds one entry per **non-null** value, so the loop pushes only `min(total_count, row_hi)`. The pre-fix `.take(row_count).collect()` sized from the iterator `size_hint` (= `total_count`, exact). Net: a transient per-page over-reservation of `null_count_in_range * 4` bytes (freed when the slicer drops; no per-row cost; zero on a no-null full read). Consider `try_reserve_exact(min(limit, iter.size_hint().0))`.

2. **`ReadParquetRecordCursor.skipRows()` message-preservation is untested** — `ReadParquetRecordCursor.java:314-321`. Reverting the `hasNext()` catch fails 2 fixture tests; reverting the `skipRows()` catch fails none — the fixtures drain via `hasNext`, never the row-skip path. The block is a near-identical copy of the proven one, so risk is low, but the second site has no failing-without-fix guard. Optional: a fixture test with an `OFFSET`/row-skip that forces the `skipRows()` decode.

3. **Duplicated catch block** — `ReadParquetRecordCursor.java`, `hasNext()` and `skipRows()` carry the identical 4-line `getFlyweightMessage().toString()` + wrap. The flyweight-copy idiom is correct (snapshots before `nonCritical()` clears the thread-local). Optional: extract a `private CairoException wrapParquetDecodeError(CairoException)` so the flyweight hazard lives in one place.

4. **Dead `block_size_values == 0` guard** — `decoders/delta_binary_packed.rs:206-208`. `miniblock_size` is a validated non-zero multiple of 32 (and the empty-buffer iterator short-circuits at :175), so `checked_mul` of two non-zero `usize` can never yield `Some(0)` — the branch is unreachable. Per the repo's Rust guideline (uncovered lines must be tested or marked), replace with `debug_assert!(block_size_values > 0)` or drop it (the preceding `checked_mul` already handles overflow).

5. **Misleading early-return comment** — `delta_binary_packed.rs:174-177`. The comment attributes the triple-zero early-return to "`value_count <= 1`", but for `value_count <= 1` on a non-empty buffer `try_new` sets `miniblock_index = miniblocks_per_block` (≠ 0), so the condition can only fire for the `page_data.is_empty()` iterator. Reword to "empty-buffer iterator".

### Verified, not issues (for the record)

- **`block_size % 128` guard is not a regression.** parquet2's encoder hardcodes `block_size=128, 1 miniblock/block`, and parquet2's decoder already rejected non-128 block sizes on master (#7212). QuestDB's own files always comply; the guard newly gates only foreign, spec-violating int/long DELTA pages (`block_size ∈ {32,64,96}`), and the old `32/48/100` tests were hand-built vectors (correctly updated to 128).
- **The 6 failing `parquet2` tests are environmental** — `File::open(...).unwrap()` on the uninitialized `parquet-testing` fixture submodule. The PR touches nothing in `parquet2/src/{read,write}/`; they fail identically on master.
- **No native-memory leak on any new error path** — the decode buffers are borrowed (`&mut`), so an `Err` mid-decode leaves them attached to their Java owners (freed in `close()`); the new scratch (`def_scratch`, dict reservations, length `Vec`s) is dropped on unwind and, in every case, fails before any JNI-persistent buffer is written. The `rss_mem_used()==0` tests confirm the allocator-tracked paths.
- **`get_end_pointer` change is a no-op for QuestDB's 1-miniblock-per-block shape** — `last_block_miniblocks` reduces to `miniblocks_per_block` when `miniblocks_per_block == 1`; only foreign >1-miniblock pages differ (the bug being fixed).
- **`calculateSize()` and `next()` now agree** — the `lo: 0 → -1` fix matches `next()`'s iterated count (29 vs the pre-fix 4) and is correctly scoped to the backward cursor's exclusive-lo accounting; the forward cursor uses `+1` and never had this bug. `COUNT`/`LIMIT`/the framework size cross-check all received the undercount before and get the correct value now.

### Out of scope (not blocking)

- **Pre-existing `skip_till_depth` Map panic** — `parquet-format-safe/.../thrift/protocol/mod.rs:196,199` `.expect("non-zero sized map should contain key type")`. Same malformed-file-abort family this PR addresses, but not touched here, and reachability is unverified (depends on whether the compact `read_map_begin` can return `size > 0` with `key_type = None`). Worth a defensive follow-up.

### Notes

- Coverage is strong and assertions pin specific error messages; the committed `libquestdbr.*` was confirmed rebuilt with this Rust code (the new guard strings are present in the binary), so the Java fixture/pruning tests exercise the real native guards rather than stale binaries.
- Every out-of-diff caller was explicitly cleared (consumers of `MiniblockIterator`/`get_end_pointer`, `decompress_*`, `calculateSize` via `COUNT`/`LIMIT`/the size cross-check, and the parallel page-frame reader).

The Minor items are all optional polish; none should block merge.

---

glasstiger:
## Code review (level 3, mission-critical pass) — re-review at the post-merge head

Re-reviewed the full change surface at `ca9c50955d` across correctness, Rust panic/overflow safety, native-memory leaks across JNI, cross-context callers, concurrency, performance, and test coverage, plus an independent fresh-context adversarial pass. The crash-class hardening, the DELTA partial-read corruption fix, and the `calculateSize` off-by-one are all correct, leak-free, and tested. The headline of this round is an artifact problem the earlier passes could not have seen, now fixed.

### Critical (was blocking — now RESOLVED)

**The committed native libraries were stale — the fix did not ship.** At the merge head `a16c964738`, all five committed `libquestdbr` binaries were byte-identical to the master base and contained none of this PR's Rust changes. The branch did rebuild them (`12c4029086`), but the master-merge (parents `12c40290` + `4fe18837` #7230) resolved the binary conflict in favor of master, silently reverting the rebuild. Verified three ways:

- `strings` on the committed `darwin-aarch64/libquestdbr.dylib`: pre-existing guards present, but `must be a multiple of 128`, `too short to hold`, `for the target type`, `cannot allocate ... decompressed page` all **absent**.
- git blob hashes: HEAD == base across all 5 platforms, != the branch's own rebuild.
- Consequence: a default `mvn test`/`package` links the stale lib, so `ParquetRowGroupPruningTest.testIntervalScanStringMultiBlockPage{,Backward}` run against the unfixed decoder (shifted bytes), and `testRleDictionaryIndexBitWidthOver32` / `testDictionaryNumValuesOverBuffer` would **abort the JVM** — the very crash this PR fixes — rather than throw.

**Fix confirmed in `ca9c50955d "Rebuild Rust libraries"`:** touches only the 5 binaries + `THIRD_PARTY_LICENSES.txt` (no source change), all 5 are freshly built (hashes now differ from both base and the reverted rebuild), and the new guard strings are present in the committed `darwin-aarch64`, `linux-x86-64`, and `windows-x86-64` libraries. The fix now ships.

### Moderate

None.

### Minor (optional polish)

1. **`ReadParquetRecordCursor.skipRows()` message preservation is untested.** `hasNext()` is pinned (reverting it fails 2 fixture tests); the near-identical `skipRows()` catch has no coverage — the fixtures drain via `hasNext()`, never a `LIMIT offset,n` skip. Reverting line ~314 fails nothing. A `SELECT ... FROM read_parquet('<broken fixture>') LIMIT 5,10` test (both `parallel` values) would pin it.
2. **`collect_checked_lengths` over-reserves on null-bearing DELTA pages, and its doc comment is inaccurate** (`slicer/mod.rs`). `try_reserve_exact(limit)` reserves `limit = row_count` (a row count), but the length/prefix/suffix stream yields only non-null values; the old `.collect()` sized from `Take::size_hint()` = the exact non-null count. The comment's "a plain `collect` reserves the same capacity" is wrong for null-heavy pages. Transient over-reservation of `(row_count - non_null) * 4` bytes, freed at page end, cold path. Consider `try_reserve_exact(limit.min(iter.size_hint().0))`, or fix the comment.
3. **Thrift `read_list` abort-fix has no test (sync + async).** The async `stream.rs` variant is entirely uncovered. Low risk (a structural mirror of the already-fallible `read_bytes`/`read_string`), but it's a JVM-abort-class fix worth a malformed-list-count → clean-error test.
4. **The four `def_scratch` alloc-failure tests don't fail-on-revert** (`decode/array.rs`). As their own comments note, reverting the `AcVec` change shifts the failure to the sibling `data_vec` reservation, so they still pass. They prove abort-freedom + no-leak (the contract that matters); flagging only that the fail-on-revert property is absent here.
5. **Dead `block_size_values == 0` guard** (`delta_binary_packed.rs:206`). `miniblock_size` is a validated non-zero multiple of 32, so the `checked_mul` of two non-zero `usize` can't yield `Some(0)`. Replace with `debug_assert!` or drop it (the `checked_mul` already handles overflow).
6. **Misleading early-return comment** (`delta_binary_packed.rs:174`). The "value_count <= 1" example actually falls through the main body (for `value_count <= 1` on a non-empty buffer, `miniblock_index == miniblocks_per_block`, not 0), which still returns correctly. The early-return only fires for the empty-buffer iterator. Reword.

### Verified correct (for the record)

- **DELTA partial-read offset fix.** The `decoder.size_hint().0 == 0 ? consumed_bytes() : delta_stream_byte_len()` dispatch is sound because `size_hint().0` is an exact counter (`values_remaining`, decremented once per value), so it reads 0 iff the decoder is exhausted (full read) and > 0 iff `take(row_count)` stopped early (partial read). `get_end_pointer`'s `last_block_miniblocks` bound mirrors parquet2's populated-miniblock counting exactly (the parquet2 `Block` stops advancing once its values exhaust); brute-forced to byte-exact equality with `consumed_bytes()` across `miniblocks_per_block ∈ {1,2,4,8,12}` × value counts 0..=600 × varied bit-widths. For QuestDB's own 1-miniblock files `last_block_miniblocks` reduces to the old behavior, so the pre-existing VARCHAR-slice path is not regressed.
- **`calculateSize` off-by-one.** `lo: 0 → -1` for the unbounded-low (`ts < X`) backward interval gives `size += hi+1` and routes to "skip partition", matching `next()`'s iterated count (29 vs the pre-fix 4). Bounded-low branch and the forward cursor are correctly untouched.
- **No reachable Rust panic.** All new arithmetic uses `checked_*`/`saturating_*`/guarded divisors; `get_end_pointer`'s loops are bounded by the buffer (`.get()`-None on overrun); the bitpacked `num_bits > size_of::<T>()*8` guard is the sole choke point and boundary-correct (`>`, not `>=`).
- **No native-memory leak on any new error path.** `AcVec`'s `Drop` frees to the tracked allocator, so a partially-grown `def_scratch` dropped on unwind is freed and accounted; the `rss_mem_used()==0` tests genuinely drive the paths they assert.
- **No data race.** All new state is per-iterator/per-decode-context/per-cursor; the allocator's accounting is atomic. The `ReadParquetRecordCursor` flyweight copy snapshots `ex.getFlyweightMessage().toString()` to a `String` before `nonCritical()` resets the thread-local — correct order, thread-local-safe.
- **No hot-path perf regression.** `resize_decompress_buffer` is grow-only (no per-page memset; the earlier `clear()`+resize regression is gone), the full-read DELTA path stays O(1), the structural walk is partial-read-only and bounded, and every new guard/reservation is per-page/per-chunk, not per-value.
- Prior level-2 findings (C1 `def_scratch`, C2 FLBA decimal, M1 memset, M2 V2-uncompressed sizing, M3 `OutOfMemory` classification) all confirmed addressed.

### Out of scope (not blocking)

- `parquet2 cargo test --lib` shows 6 failures — uninitialized `parquet-testing` submodule (`File::open` NotFound); the PR touches nothing in `parquet2/src/{read,write}`. Pre-existing and unrelated.

### Verdict

**Approve.** The one blocker — stale committed binaries — is fixed: the libraries are now freshly rebuilt with this PR's Rust changes across all five platforms, the source is unchanged, and the rebuild commit is clean. The six Minors are optional polish; the `skipRows` and `read_list` test gaps (1, 3) are the most worth doing, but none block merge.

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 1050 / 1271 (82.61%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/ReadParquetRecordCursor.java|4|8|50.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decode.rs|425|622|68.33%|
|:large_blue_circle:|qdbr/src/parquet_read/row_groups.rs|13|15|86.67%|
|:large_blue_circle:|parquet2/src/encoding/hybrid_rle/mod.rs|8|9|88.89%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/mod.rs|18|20|90.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/rle_dict_varchar_slice.rs|26|28|92.86%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/rle_dictionary.rs|17|18|94.44%|
|:large_blue_circle:|qdbr/src/parquet_read/slicer/tests.rs|101|105|96.19%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/delta_binary_packed.rs|118|122|96.72%|
|:large_blue_circle:|qdbr/src/parquet_read/decode/array.rs|165|169|97.63%|
|:large_blue_circle:|io/questdb/cairo/IntervalBwdPartitionFrameCursor.java|1|1|100.00%|
|:large_blue_circle:|parquet-format-safe/src/thrift/protocol/mod.rs|2|2|100.00%|
|:large_blue_circle:|parquet2/src/encoding/bitpacked/decode.rs|20|20|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decode/decimal.rs|56|56|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/dictionary.rs|37|37|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/decoders/delta_length_array.rs|2|2|100.00%|
|:large_blue_circle:|qdbr/src/parquet_read/slicer/mod.rs|37|37|100.00%|

PR #7224, opened 2026-06-09
fix(sql): fix wrong first/last results in parallel GROUP BY

ideoma:
## What this fixes

In a parallel GROUP BY, the order-sensitive aggregates `first()`, `last()`, `first_not_null()` and `last_not_null()` could return a value from the wrong row. There are two distinct defects:

1. **Per-row ordering (`computeNext`)** over `DECIMAL`, `GEOHASH` and `IPv4` columns: the returned value came from an arbitrary row in the group rather than the row with the smallest (first) or largest (last) position.
2. **Shard merge (`last_not_null` `merge`)** over every value type: when a group's value was non-null in one worker's map but null in another, `last_not_null` could silently drop the real value and return null.

`sum`/`avg`/`min`/`max`/`count` are order-independent and were never affected.

Both defects are data- and timing-dependent: they only surface when workers reduce page frames into their aggregation slots out of row order, which is more likely under the HTTP/PGWire worker pools or work stealing, and rare otherwise. Many queries returned correct results most of the time, which is why this went unnoticed.

## Root cause

Parallel keyed GROUP BY reduces page frames into per-worker "slot" maps, then merges those maps pairwise into shard maps. `PerWorkerLocks.acquireSlot` grants mutual exclusion only, not stable worker-to-slot affinity (steal threads start at a random slot), so a single slot can accumulate a higher-position frame's rows before a lower-position frame's rows, and any worker map can be merged in either direction. The inherited `computeFirst` stores the first row of a key unconditionally, including a NULL argument, with a real row id - so a map whose only rows for a key are NULL ends up holding `(realRowId, NULL)`.

1. The order-sensitive aggregates must decide the per-row winner by comparing the row id in `computeNext`, never by relying on processing order. A prior change (`b06bd2a94e`) added that guard to most types but missed the Decimal family, the standalone GeoHash `first()` functions, the `last()` GeoHash factory, and the not-null GeoHash/IPv4/Decimal factories.
2. Every `last_not_null` `merge` compared row ids alone (`srcRowId > destRowId`). When the all-null `(realRowId, NULL)` slot carried a higher row id than another map's single non-null value, the merge kept the NULL and dropped the real value. The `first_not_null` merges already guarded against this with a dest-value-null check; the `last_not_null` merges did not.

## What changed

**`computeNext` ordering** in the missed functions, brought in line with the existing correct references (`First`/`Last` `Long`, `First`/`Last` `NotNull` `Uuid`):

- `first` keeps the smaller row id; `last` keeps the larger.
- `first_not_null` / `last_not_null` consider the argument only when it is non-null, and overwrite the stored value only when that value is null or the row id wins.

**`last_not_null` `merge`** for every value type now also accepts the src when the dest slot still holds a NULL value, mirroring the `first_not_null` merges. This covers Char, Date, Decimal (all widths), Double, Float, GeoHash (all widths), IPv4, Int, Long, Str, Symbol, Timestamp, Uuid and Varchar. The Array merge already carried the guard and is unchanged. `first_not_null` already guards every type and was not affected (verified, including a teeth check that removing the guard reproduces the identical bug).

Correcting the earlier revision of this description: the `merge` path did **not** already compare row ids correctly for `last_not_null` - that gap is fixed here. The non-keyed `computeBatch` backward-scan was reviewed and left unchanged: both call sites build the slot with `computeBatch` only after `setEmpty`, which never produces a `(realRowId, NULL)` state, so its existing `existingRowId == LONG_NULL` guard is sufficient.

## Effects and tradeoffs

- Positive: affected queries now return the correct `first`/`last`/`*_not_null` value deterministically, independent of worker scheduling.
- Behavior change: queries that previously returned an incorrect value will now return a different (correct) value. This is the intended correction, but downstream snapshots/expected outputs that captured the old wrong value will change.
- Performance: `computeNext` gains one row-id comparison (and, for `*_not_null`, one stored-null check) per row, and each `last_not_null` `merge` gains one stored-null check. This matches what every other type already does; no measurable regression is expected, and there is no change to the batch hot path.

## Test plan

- Added `FirstLastNotNullParallelMergeTest`: an end-to-end test over the real parallel sharded GROUP BY path for both `first_not_null` and `last_not_null` across all 14 value types. Each key gets a single non-null value placed so an all-null slot sorts ahead of it (first occurrence for `last`, last occurrence for `first`); the count of non-null results must equal the key count.
- Confirmed the end-to-end test fails on the pre-fix code (14/14 `last_not_null` types) and passes with the fix.
- Verified `first_not_null` has teeth by temporarily removing one dest-null guard: the corresponding test failed with the identical signature, confirming `first_not_null` is correct only because its guard is present.
- Added `FirstLastParallelOrderingTest`: a deterministic unit test driving `computeFirst`/`computeNext` and `merge` for the Decimal, GeoHash and IPv4 shapes across all four aggregate shapes; fails pre-fix, passes with the fix.
- Ran the existing parallel GROUP BY and first/last not-null regression suites with the fix in place; all pass.

Note on reliability: the end-to-end tests are statistically reliable (1000 keys x 10 iterations), not a single-key hard guarantee - per-key merge direction is scheduling-dependent. Across that volume the buggy code fails on every observed run.


ideoma:
## Code review — level 2 (with resolutions)

**Verdict: approve.** The production fix is correct, complete, and thread-safe — I verified the root-cause premise, every changed `computeNext`/`merge`, the concurrency model, and that no type/path was missed. No blocking findings. The items below were either fixed on the branch or are non-blocking; resolution status is inline.

### Root cause — confirmed

Parallel keyed GROUP BY reduces page frames into per-worker slot maps. `PerWorkerLocks.acquireSlot` grants no stable worker→slot affinity (steal threads start at `rnd.nextInt(workerCount)`), and `baseRowId` encodes the frame index, so a slot map genuinely receives frames out of rowId order and `computeNext` sees non-monotonic rowIds. Comparing rowId in `computeNext`, and accepting a non-null src over a null dest in `merge`, is the right, internally-consistent approach. The default `GroupByFunction.computeKeyedBatch` delegates to `computeFirst`/`computeNext`, so the keyed-batch path inherits the fix.

### Findings

**M1 — the 14 `first_not_null` end-to-end cases don't exercise this PR's `computeNext` ordering fix.** ✅ Resolved (documentation + already-complete deterministic coverage).

Empirically: reverting `FirstNotNullIPv4GroupByFunctionFactory.computeNext` to its pre-fix body leaves `testFirstNotNullIPv4` passing, whereas the matching `last_not_null` revert fails. The reason it can't be made to fail by tuning the fixture: the `first_not_null` `merge` already resolves min-rowId correctly (`srcRowId < destRowId || destRowId == LONG_NULL || destVal == NULL`), so it *erases* a single slot's `computeNext` ordering error during shard merge. The wrong answer survives end-to-end only when two non-null rows of the same key reduce into the *same* slot in reverse rowId order — a scheduling window SQL cannot force, so a count-based single-non-null fixture cannot give this fix teeth.

Correction to my own finding: these cases are not entirely toothless — they *are* a regression guard for the merge dest-null invariant (deleting that guard makes them fail with the same signature). They only miss the `computeNext` change, which the correct merge masks.

Action: kept the 14 cases; rewrote the class Javadoc on `FirstLastNotNullParallelMergeTest` to state the real scope and to point at `FirstLastParallelOrderingTest`, which drives `computeFirst`/`computeNext` directly with descending rowIds and covers the ordering defect deterministically for every type/width. No probabilistic fixture added.

**M2 — GeoHash width coverage was asymmetric (3 of 4 widths untested on several paths).** ✅ Fixed.

Added 12 deterministic methods to `FirstLastParallelOrderingTest` so all four widths mirror the Decimal family — plain `last` (+Byte/Long/Short), `first_not_null` computeNext (+Int/Long/Short), `last_not_null` merge (+Byte/Int/Short), `last_not_null` computeNext (+Byte/Int/Short). Added the missing `setNull()` to `GeoIntArg`/`GeoShortArg`. The class is now 64 tests (was 52), all green.

**M3 — `GroupByFunction` Javadoc still asserted the guarantee this bug disproves.** ✅ Fixed, on both entry points.

The false claim ("the value is guaranteed to be growing between subsequent calls … in parallel GROUP BY …") sat on **both** `computeFirst` and `computeNext`, not only `computeNext`. Both now state the real contract: a larger rowId denotes a later row, but rowIds are *not* guaranteed to grow between calls for a group under parallel reduce / work stealing, so order-sensitive aggregates must compare the provided rowId against the stored one and never rely on call order. This closes the trap that produced this bug (and the partial predecessor fix `b06bd2a94e`).

**M4 — `last_not_null` Decimal128/256 `computeNext` did an avoidable per-row wide copy.** ✅ Fixed.

Reordered to test the cheap rowId first (`boolean isWinner = rowId > getLong(valueIndex)`) and read the 16/32-byte `storedValue` only when the rowId doesn't already win (to probe null). The wide copy is now skipped on the common in-order path. Inline widths (8/16/32/64) and the symmetric `first_not_null` copy (genuinely unavoidable) were left alone.

### Minor

- **Two empty test helpers** (`GeoIntArg`/`GeoShortArg` lacked `setNull()`) — ✅ fixed as part of M2.
- **Redundant shadow `decimal128`/`decimal256` buffers** in `LastNotNullDecimalGroupByFunctionFactory` — ↪ deferred. Pre-existing (commit `020ec80464f`), out of scope for a correctness fix; harmless (one extra heap object per instance, no per-row cost).
- **Comment-presence divergence** (the "no `destRowId == LONG_NULL` term needed" note is on the factory merges but not the single-class ones) — ↪ deferred, cosmetic.
- **PR-description leftover** — the "Correcting the earlier revision of this description…" paragraph is meta-commentary that would land in squash history; recommend trimming before merge. ◻ open (non-code).

### Downgraded (investigated, not real)

- **Data race on the new `storedValue` / reused `decimal128` buffers** — false. `FirstLastDecimalFunc.isThreadSafe()` returns `false`, so each worker slot gets a distinct function instance (own buffers); slots are held under exclusive `PerWorkerLocks`; and reduce completes before merge via the `dispatchAndAwait` barrier. No concurrent access.
- **Plain `last` merges compare rowId alone without a null guard** — false; correct for plain `last` (NULL is a legitimate value, the stored rowId is always real). Those sites are all plain `last`, never `last_not_null`.
- **`first_not_null` merges left unfixed** — false; they already carried the dest-null guard pre-PR.
- **`computeKeyedBatch` overrides bypass the fix** — false; the default delegates to `computeFirst`/`computeNext`, and explicit overrides (e.g. `LastNotNullLong`) already inline the correct logic.
- **Non-keyed `computeBatch` mishandles a `(realRowId, NULL)` slot** — false; those scans only write on a non-null value and run after `setEmpty` (rowId `LONG_NULL`), so the state never arises; Decimal has no `computeBatch`.

### Conclusion

The fix was already correct, complete, and thread-safe. With M2/M3/M4 fixed and M1 resolved, the deterministic suite now fully covers both the `computeNext` ordering defect (every type and GeoHash width) and the `last_not_null` merge defect; the regression-inviting Javadoc is corrected on both entry points; and the optional Decimal128/256 hot-path copy is gone. `FirstLastParallelOrderingTest` (64/64) and `FirstLastNotNullParallelMergeTest` (28/28) pass with the fixes in place.

What remains is strictly non-blocking and non-code: the PR-body paragraph trim, plus two optional cleanups (shadow buffers, comment parity) better handled outside a correctness PR.

One note for future work: do not pursue an end-to-end test for the `first_not_null` ordering path — it is structurally unreliable (the correct merge masks it; frame→slot scheduling is not controllable from SQL). The deterministic unit test is the correct and sufficient vehicle.

ideoma:
## Code review — level 2, deeper pass (8 agents) with resolutions

A second, more detailed pass (correctness, concurrency, performance ×2, resource, tests, code-quality, metadata + verification), complementary to the review above. **Verdict: approve.** The fix is correct, the scoping is complete, and the concurrency/resource concerns are cleared — no Critical or blocking issues. This pass was originally written against an earlier branch state; resolution status is inline, and a few items it raised were already closed by later branch commits.

### Critical

None. Every changed `computeNext` uses the correct operator (`<` first, `>` last), every `last_not_null` merge uses the correct per-type null sentinel, the Decimal128/256 buffer aliasing is sound, and no type or path that needed the fix was missed.

### Moderate

**M1 — direct unit coverage for the per-type `last_not_null` merge guard.** ✅ Resolved.
All 22 `*MergeKeepsNonNullOverNullDest` unit tests now exist in `FirstLastParallelOrderingTest`: the 11 simple types (Char, Date, Double, Float, Int, Long, Str, Symbol, Timestamp, Uuid, Varchar) plus Decimal ×6, GeoHash ×4, IPv4. The distinct sentinel paths each now have direct teeth — Str/Varchar `valueIndex+2` bool flag, `Uuid.isNull(lo,hi)`, Symbol `VALUE_IS_NULL`, Char `CharConstant.ZERO`. (The simple-type cases were added on the branch after this pass was written; I added the 3 GeoHash widths.) The count-only `FirstLastNotNullParallelMergeTest` is no longer the sole guard for these.

### Minor

**m1 — `last_not_null` Decimal128/256 *merge* does 3 native decimal reads where 2 suffice.** ◻ Open (non-blocking).
`LastNotNullDecimalGroupByFunctionFactory` merge reads src→`decimal128`, then dest→`decimal128` (clobbering src), then re-reads src→`decimal128` before storing. The class already owns a `storedValue` buffer; reading dest into it instead keeps src intact and drops the 3rd 16/32-byte read. Free and in-scope. (Distinct from m3 below — this is the *merge*, not `computeNext`.)

**m2 — `last_not_null` GeoHash (4 widths) + IPv4 `computeNext` evaluate the arg twice.** ◻ Open (non-blocking).
They null-check `arg.getGeoX(record)` / `arg.getIPv4(record)`, then call `computeFirst(...)`, which re-reads the same arg — and the arg can be an expression (`last_not_null(a+b)`), so the second read re-runs the sub-expression. This PR's own `first_not_null` GeoHash/IPv4 and the Decimal scalars already read once into a local and `putX` directly, so the two not-null families are now inconsistent. Pre-existing pattern, but the PR touched these exact lines; aligning them is a cheap consistency/perf cleanup.

**m3 — `last_not_null` Decimal128/256 `computeNext` read the stored value (16/32 B) unconditionally.** ✅ Fixed.
Reordered to test the cheap 8-byte rowId compare first (`boolean isWinner = rowId > getLong(valueIndex)`); the wide `storedValue` read now happens only off the in-order path (to probe null). `first_not_null` correctly left unchanged — there the value read is load-bearing.

**m4 — misleading rationale in the repeated merge comment.** ◻ Open (re-verified **valid**).
The comment *"Unlike first_not_null, no destRowId == LONG_NULL term is needed: a real srcRowId always exceeds an empty dest's LONG_NULL"* has the right conclusion but a wrong reason. I re-read `OrderedMap.mergeFixedSizeKey`: `mergeFunc.merge(...)` is called only when the src key **already matches** a dest key (line 544); src-only keys are bulk-copied (lines 554-560) without merging. So `merge()` never sees an empty `LONG_NULL` dest — the comment justifies the omission via a case that cannot occur. Suggested reword: *"merge only runs on dest keys that already hold a real row, so destRowId is never LONG_NULL here; the dest-value-null check covers the all-null `(realRowId, NULL)` slot."* Corollary (confirmed): the `|| destRowId == LONG_NULL` term still carried by the sibling `first_not_null` merges is unreachable defensive code.

**m5 — PR description process meta-commentary.** ◻ Open (non-code).
The "Correcting the earlier revision of this description…" sentence addresses a prior draft and would land in the squash-merged master message. Delete it; "What changed" already states the merge fact cleanly.

**m6 — trivia.** ⚪ Partial.
The stray blank line in `GeoIntArg`/`GeoShortArg` is gone — both now have a `setNull()` body (added so the new GeoInt/GeoShort not-null tests can null them). The redundant `func.initValueIndex(0)` after `initValueTypes(types)` is unchanged (harmless).

### Other coverage gaps (optional)

- **Changed GeoHash widths with no direct unit test** (FirstNotNullGeoHash Int/Long/Short, LastNotNullGeoHash Byte/Int/Short, plain LastGeoHash Byte/Short/Long) — ✅ resolved (+12 deterministic tests; `FirstLastParallelOrderingTest` is now 64 tests, was 52, all green).
- **Plain-GeoHash `computeNext` no-overwrite branch never exercised** (descending `ROW_IDS` make the guard true/false on every call depending on direction) — ◻ open, optional; one ascending-rowId case per direction would close it. Low value (the winner assertion already proves the taken branch is correct).
- **No e2e test for plain `first()`/`last()` over Decimal/GeoHash** (the e2e is all `_not_null`) — ◻ open, optional; the deterministic driver already covers plain first/last for all widths.

### Hardening (not a defect) — ◻ open, optional

The thread-safety of the new `storedValue`/`decimal128`/`decimal256` scratch fields rests entirely on `isThreadSafe()` returning `false` (which forces per-worker function clones). Nothing pins that — flipping it to `true` for a decimal first/last func would silently turn these into a data race. A one-line test asserting `isThreadSafe() == false` for the decimal funcs would prevent regression.

### Downgraded (false positives removed)

- **"Shared mutable scratch fields are a data race"** → cleared. Reduce uses `perWorkerGroupByFunctions` (one instance list per worker slot); merge acquires an exclusive `slotId` via `PerWorkerLocks.acquireSlot` and uses that slot's updater; reduce/merge are phase-separated by `dispatchAndAwait()`. No instance is touched by two threads concurrently.
- **"Str/Varchar merge copies a src off-heap pointer into dest → use-after-free"** → cleared. Pre-existing and identical to `first_not_null`; worker `GroupByAllocator`s share the `AsyncGroupByAtom` lifetime and outlive every per-shard `srcMap.close()`.
- **"A `last_not_null` type was missed"** → cleared. All 14 merges carry the dest-null guard (Array already had it); all simple types' `computeNext` were already rowId-guarded; `first_not_null` merges already had the guard and are correctly untouched.
- **Perf agent's claim that the m3 reorder doesn't help** → it conflated `last_not_null` with `first_not_null`; the reorder is valid for `last_not_null`.

### Re-evaluated summary

- Resolved: **M1** (the only Moderate — full per-type merge unit coverage), **m3** (computeNext wide-read reorder), the **GeoHash width** coverage gap, **m6** blank line.
- Open, non-blocking: **m1** (merge 3-read), **m2** (computeNext arg double-eval), **m4** (comment reword — re-verified valid), **m5** (PR-body trim), plus optional coverage/hardening niceties.
- Verification: `FirstLastParallelOrderingTest` 64/64 and `FirstLastNotNullParallelMergeTest` 28/28 pass with the fixes in place.

**Verdict: approve.** The fix is correct and complete; the one Moderate and the headline perf nit are resolved. Everything still open is Minor and non-blocking — three cheap code/comment cleanups (m1/m2/m4), a one-line PR-body trim (m5), and a few optional coverage/hardening niceties. None changes the data-correctness guarantee. Recommend landing m1/m2/m4/m5 as quick follow-ups.

bluestreak01:
@ideoma I completed a level 3 review. Findings below.

**Critical**

- [out-of-diff callsite] `FirstNotNullIPv4GroupByFunctionFactory.Func` and `LastNotNullIPv4GroupByFunctionFactory.Func` still inherit the plain IPv4 `computeKeyedBatch()` implementations. This PR overrides `computeBatch()` and `computeNext()`, but the async keyed GROUP BY non-sharded batched path calls `computeKeyedBatch()` directly from `AsyncGroupByRecordCursorFactory.aggregateFilteredNonShardedBatched()` and `aggregateNonShardedBatched()`. The inherited implementations in `FirstIPv4GroupByFunction` and `LastIPv4GroupByFunction` write `Numbers.IPv4_NULL` as a valid value, so `first_not_null(ipv4_col)` can keep an earlier NULL and ignore a later non-null, while `last_not_null(ipv4_col)` can overwrite a non-null with a later NULL.

  Code path: `AsyncGroupByRecordCursorFactory.java:334` / `:419` -> `GroupByFunction.computeKeyedBatch()` -> inherited `FirstIPv4GroupByFunction.java:71` or `LastIPv4GroupByFunction.java:55`. The not-null factories inherit because `FirstNotNullIPv4GroupByFunctionFactory.java:62` extends `FirstIPv4GroupByFunction` and `LastNotNullIPv4GroupByFunctionFactory.java:55` extends `LastIPv4GroupByFunction` without overriding `computeKeyedBatch()`.

  Suggested fix: override `computeKeyedBatch()` for both IPv4 not-null functions, mirroring the int not-null implementations, or disable batch computation for these two functions if the optimized path is not worth duplicating.

**Moderate**

- The new tests do not cover that keyed batch path and under-assert payload correctness. `FirstLastNotNullParallelMergeTest.java:88` forces sharding, which bypasses the non-sharded keyed batch reducer above; `FirstLastParallelOrderingTest.java:993` checks only the committed row id; and `FirstLastNotNullParallelMergeTest.java:244` checks only non-null count. These can pass while the aggregate stores the wrong payload. Please add targeted value assertions for the winning row, plus a deterministic async SQL test with IPv4 direct columns and sharding threshold high/default.

**Minor**

- `LastNotNullDecimalGroupByFunctionFactory.java:107` and `:193` read the src decimal, overwrite the same scratch with dest, then re-read src before storing. Keeping separate src/dest decimal buffers, or checking row id before reading dest, would avoid the extra 16/32-byte read in the merge path.

- The repeated merge comment at `LastNotNullDecimalGroupByFunctionFactory.java:112`, `LastNotNullGeoHashGroupByFunctionFactory.java:115`, and `LastNotNullIPv4GroupByFunctionFactory.java:103` is misleading. The merge functions are called for matching destination keys; source-only keys are copied separately. I would reword it around existing-dest-key merge semantics instead of an empty-dest `LONG_NULL` case.

**Checks**

- Ran `mvn -pl core -Dtest=FirstLastParallelOrderingTest,FirstLastNotNullParallelMergeTest test`: passed, 92 tests.
- Ran `git diff --check origin/master...HEAD`: clean.
- Verified and dropped draft concerns around row-id direction, null sentinels, decimal thread safety, and str/varchar ownership.

Verdict: request changes because the IPv4 not-null keyed batch path can still produce wrong query results.

bluestreak01 (approved):
Approve. The production fix is correct and complete: verified the root cause, every changed `computeNext` and `merge` across all first/last group-by functions, the null sentinel for each of the 22 types, the concurrency model (per-worker clones + exclusive slot locks + reduce/merge phase barrier), resource lifetimes, and the test teeth (both new suites fail pre-fix / pass post-fix; an adversarial serial-vs-parallel differential probe could not break it).

No Critical or Moderate findings. Out-of-diff paths (computeBatch, computeKeyedBatch, OrderedMap.merge, AsyncGroupBy*) cleared.

Minor, non-blocking follow-ups (optional):
- last_not_null Decimal128/256 merge does 3 wide reads where 2 suffice (read dest into the existing storedValue buffer).
- last_not_null GeoHash/IPv4 computeNext evaluate arg twice via computeFirst; mirror the single-read first_not_null shape rewritten in this PR.
- The 'Unlike first_not_null, no destRowId == LONG_NULL term needed' comment names the rowId compare as the reason, but the dest-value-null term is what catches the all-null (realRowId, NULL) slot; reword and apply uniformly (the 11 single-class merges omit it).
- Merge unit helper asserts only the committed rowId, not the surviving value.
- e2e covers only Decimal64 and 5-bit GeoByte; a wider literal would exercise the Decimal128/256 two-buffer merge end-to-end.
- Trim the 'Correcting the earlier revision of this description:' meta-line from the body before squash-merge.

Tradeoff to keep visible (PR body already does): queries that previously returned a wrong value now return the corrected one, so captured snapshots of old output will change.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 163 / 163 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullUuidGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullSymbolGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/FirstNotNullIPv4GroupByFunctionFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/FirstGeoHashGroupByFunctionLong.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullDecimalGroupByFunctionFactory.java|38|38|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullDoubleGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullLongGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullDateGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/FirstGeoHashGroupByFunctionShort.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/FirstNotNullDecimalGroupByFunctionFactory.java|34|34|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/FirstDecimalGroupByFunctionFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullStrGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/FirstNotNullGeoHashGroupByFunctionFactory.java|20|20|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullTimestampGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullIntGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastGeoHashGroupByFunctionFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullIPv4GroupByFunctionFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastDecimalGroupByFunctionFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullVarcharGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullGeoHashGroupByFunctionFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/FirstGeoHashGroupByFunctionByte.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullCharGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullFloatGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/FirstGeoHashGroupByFunctionInt.java|2|2|100.00%|

PR #7225, opened 2026-06-09
fix(sql): fix wrong results in splice, right and full outer joins with a left-table filter

puzpuzpuz:
Supersedes #7218.

## Summary

SPLICE, FULL OUTER and RIGHT OUTER joins NULL-extend the master (left) side: SPLICE emits slave-only timestamps paired with a NULL master, while FULL and RIGHT OUTER emit unmatched right rows with a NULL master. The optimiser pushed a WHERE predicate that referenced only the master table down into that table's sub-query, because it gated the push solely on the referenced table's own join type (the base table is an inner join) and ignored any downstream NULL-extending join.

The pushed predicate then left those NULL-master rows unfiltered, and for SPLICE it also changed which master row prevailed at each slave timestamp. As a result any of these three joins could return wrong results when a filter referenced only the master table. A closely related case — an equality between two master tables, such as `WHERE t0.a = t1.b` — folds into the inner join's keys instead of pushing into a sub-query, but leaks the same NULL-master rows; it is addressed here too.

The query fuzzer surfaced this as a literal-vs-bind divergence: with a literal constant the optimiser could also propagate it to the slave through the join key, while the bind-variable form could not, so the two forms disagreed on the row count. The literal form looked correct only when the master happened to contain the key.

## Fix

`masterNullingJoinIndex()` reports the outermost downstream join that NULL-extends a given table. Its membership covers SPLICE, FULL OUTER and RIGHT OUTER, plus the `JOIN_CROSS_RIGHT` / `JOIN_CROSS_FULL` joins that `homogenizeCrossJoins` produces for a non-equi RIGHT/FULL OUTER ON clause (for example `ON a.x > b.y`); those CROSS variants still build a NULL master record, so a master-side WHERE over them stays post-join too.

`analyseEquals()`, `assignFilters()` and `moveWhereInsideSubQueries()` consult it and keep such a predicate as a post-join filter (`addPostJoinWhereClause`) instead of pushing it into the table's sub-query. `moveWhereInsideSubQueries()` is the path the predicate takes when the join is wrapped in a sub-query and the master filter sits on the outer model; all three of its push-down branches — single-table, multi-table, and barrier-joined-table — apply the guard, so a wrapped query is held back exactly as the unwrapped one is. LEFT OUTER, ASOF and LT joins keep every master row, so their master-side predicates still push down unchanged.

A master-side equality that spans two master tables (`WHERE t0.a = t1.b`) reaches the same outcome by a different route. `analyseEquals()` would otherwise fold it into the inner join's keys, where it runs before any downstream NULL-extending join; it is now deferred to the post-join filter whenever such a join NULL-extends either referenced table, and `assignFilters()` anchors it in execution order alongside the single-table predicates. An inner-join ON conjunct still folds into the join keys as before. When the two tables share no other join condition, holding the equality back turns their inner join into a cross join followed by the post-join filter — but only when a downstream NULL-extending join is present.

Two further axes matter:

- Origin. Only a WHERE predicate is held back. An inner-join ON conjunct gates the inner join, which runs before any downstream NULL-extending join, so it still pushes down.
- Execution order. `assignFilters()` anchors the post-join filter in `getOrderedJoinModels()` (execution) space, not model-index space, because table reordering can move a homogenized CROSS join to the end of the execution order. A model-index anchor could place the filter below a later nulling join (NULL rows leak) or before the referenced table joins (the filter fails to compile). `moveWhereInsideSubQueries()` anchors its multi-table and barrier-joined branches the same way, for the same reason.

A master-side `column = non-null-literal` predicate still stays post-join, but the constant is propagated to the slave through the join key, so the slave sub-query is pre-filtered. That propagation is result-neutral — matched rows have `slave.key = master.key`, and the post-join filter already drops the NULL-master rows — and matches the existing behaviour of regex (`~`) predicates. A bind variable is excluded from this propagation: it can be NULL at runtime, and because `null = null` is TRUE in QuestDB, propagating it would change which rows survive.

## Secondary fixes

Three smaller changes accompany the optimiser fix. Each corrects a contract in a zero-row situation that the post-join SPLICE/RIGHT/FULL path can now reach, and none of them changes the result of an existing query:

- `EmptySymbolMapReader.keyOf(null)` returns `VALUE_IS_NULL` instead of `VALUE_NOT_FOUND`, so a null symbol round-trips `getInt` -> `keyOf` exactly as `SymbolMapReaderImpl` does.
- `EmptyTableRandomRecordCursor.newSymbolTable()` returns the shared immutable `EmptySymbolMapReader.INSTANCE` instead of inheriting the default that throws `UnsupportedOperationException`; a parallel consumer projecting a SYMBOL from an empty (or folded-to-empty) SPLICE master would otherwise hit that exception.
- `QueryModel.parseWhereClause()` preserves each conjunct's `innerPredicate` origin instead of hardcoding `false`, so the optimiser can still tell an inner-join ON conjunct from a real WHERE predicate after the clause is re-parsed.

## Splice join as the master of a following join

This PR also fixes a separate, pre-existing crash in join code generation that does not depend on any WHERE clause. `generateJoins()` clears the master alias after building every join type except SPLICE. Once a join is built, its result metadata carries fully qualified column names (such as `m.k`), and the next join in the chain copies them with a null alias. The leftover root alias from a leading SPLICE join leaked into the following join's `createJoinMetadata()` call, and `JoinRecordMetadata.addAlias` tripped its `assert tableAlias == null` guard on the already-dotted name. A query such as `m SPLICE JOIN s ON ... RIGHT JOIN r ON ...` threw an `AssertionError`, and any join type following a leading SPLICE join behaved the same way.

The exposure is limited: the crash fires only when assertions are enabled (`-ea`), as in QuestDB's own tests and CI and in a server started with assertions. Without assertions the redundant alias is ignored and the query already returns the correct result, so there is no wrong-results path. The SPLICE case now clears the master alias to match every other join type.

## Tradeoff

Because the predicate now runs after the join rather than inside the master sub-query, a master-side filter on these three join types can no longer use an index or interval scan on the master table; the query runs as a full master scan plus a post-join filter. That post-join filter runs single-threaded and interpreted — the join factories do not support page-frame cursors, so neither the parallel filter nor the JIT compiler applies to it.

For a `column = non-null-literal` predicate the slave sub-query is still pre-filtered through the join key, so only the master side pays the full-scan cost. For inequality, `= null`, or bind-variable predicates the whole filter stays post-join with no slave pre-filter.

When a two-table master-side equality is the only condition linking the two tables, deferring it past a downstream NULL-extending join replaces the equi-join with a cross join plus a post-join filter, which can be substantially more expensive on large inputs. This applies only when such a downstream join is present; without one the equality still folds into the join key as before.

This is the cost of correctness for these joins — the previous index/interval-scan plan produced wrong results. QuestDB has no outer-join strength-reduction pass that could recover the cheaper plan where a null-rejecting predicate makes RIGHT equivalent to INNER (or FULL to LEFT); that remains a possible follow-up.

## Test plan

- `JoinTest` gains a battery of master-filter tests across SPLICE, RIGHT OUTER and FULL OUTER, each asserting the result, that the filter stays post-join (`Filter filter: ...`), and literal-vs-bind parity: `testSpliceJoinMasterFilterStaysPostJoin`, `testOuterJoinMasterFilterStaysPostJoin`, `testOuterJoinMasterFilterKeepsMatchedRow`, `testNonEquiOuterJoinMasterFilterStaysPostJoin`, `testOperatorMasterFilterStaysPostJoin`, `testColumnEqColumnMasterFilterStaysPostJoin`.
- Multi-table, barrier-joined and stacked-join coverage: `testMultiTableMasterFilterStaysPostJoin`, `testMultiTableEqualityMasterFilterStaysPostJoin` (two-table equality across master tables, RIGHT + FULL), `testBarrierJoinedMasterFilterStaysPostJoin`, `testStackedNullingJoinsMasterFilterStaysPostJoin`, `testMasterFilterAnchorsAtLastNullingJoinInOrder`.
- Wrapped-in-sub-query coverage, where the master predicate routes through `moveWhereInsideSubQueries` instead of `assignFilters`: `testWrappedSubQueryMasterFilterStaysPostJoin` (single-table), `testWrappedMultiTableMasterFilterStaysPostJoin` (multi-table, RIGHT + FULL), `testWrappedBarrierSlaveMasterFilterStaysPostJoin` (barrier-joined slave NULL-extended by a later join).
- Cases that must still push down or fold: `testInnerJoinOnConjunctPushesPastNullingJoin` (ON conjunct gates the inner join), `testCrossJoinedMasterFilterPushesDownWhenNotNulled` (master not nulled), `testOuterConstFilterDoesNotLeakIntoNestedNullingJoin` (no foreign constant in a nested sub-query), `testSpliceJoinFoldedFalseMasterFilterProducesNoRows`.
- `testMasterFilterPushesTransitiveSlaveConstForLiteral` pins the literal-constant slave propagation and its absence for bind variables.
- `SqlParserTest` outer-join model assertions move to the post-join model; a two-argument `assertQueryWithOuterJoinType` overload lets LEFT keep the pushed-down filter while RIGHT/FULL expect the post-join filter plus the transitive slave constant. `testJoinClauseAlignmentBug` likewise moves its RIGHT/FULL two-table equality (`c.relnamespace = n.oid`) to a post-join filter while leaving the LEFT model unchanged.
- `EmptyTableRecordCursorFactoryTest#testEmptyMasterFeedsSpliceJoin` covers an empty/folded-false master feeding a SPLICE join; `EmptySymbolMapReaderTest` covers `keyOf(null)` returning `VALUE_IS_NULL`.
- `ExplainPlanTest#testLeftJoinWithPostJoinFilter` and `SqlOptimiserTest#testOrderByAdviceWithMultipleJoinsAndFilters` updated for the new plan.
- `JoinTest#testSpliceJoinAsMasterOfSecondJoin` covers a SPLICE join feeding a following RIGHT OUTER, FULL OUTER and INNER join, each asserting the result.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected predicate handling in outer join queries to ensure WHERE filters on NULL-extended tables remain as post-join filters instead of being incorrectly pushed down, improving query correctness in RIGHT OUTER, FULL OUTER, and SPLICE join scenarios.

* **Tests**
  * Added comprehensive regression tests for join filter correctness with various join types and complex stacked join scenarios.


bluestreak01:
@puzpuzpuz Level-3 review of this PR. **Verdict: request changes.** The core idea — keeping master-side WHERE predicates post-join when a downstream SPLICE/RIGHT/FULL OUTER join NULL-extends the master — is correct, and the nulling-join membership list is exactly right (verified against every join factory: SPLICE/RIGHT/FULL/CROSS_RIGHT/CROSS_FULL null-extend the master via `NullRecordFactory.getInstance(masterMetadata)`; LEFT/CROSS_LEFT/ASOF/LT/WINDOW/HORIZON/UNNEST/LATERAL never do). However, the guard fires on the wrong axis in three places, and probe queries executed on this branch prove **three in-diff regressions**: queries that returned correct results on master now return wrong results or fail to compile. Every Critical below was confirmed by running the failing query on the PR branch, not just by reading the diff.

## Critical

### C1 (in-diff, wrong-results regression): inner-join ON-clause conjuncts are silently converted from ON semantics to WHERE semantics

`SqlOptimiser.java:1671`, `:1752`, `:1869-1882`. The guard keys only on the referenced table and downstream joins — it ignores the predicate's **origin**. `processJoinConditions` (`SqlOptimiser.java:5780`) feeds every inner join's ON conjuncts through the same path as WHERE conjuncts (`innerPredicate=true`), and nothing downstream can tell them apart: `ExpressionNode.innerPredicate` is set (`QueryModel.java:315`) but has **zero readers** in the codebase.

Reproduced on the branch:

- `SELECT ... FROM m JOIN x ON x.k = m.k AND m.c = 1 RIGHT JOIN s ON s.k = x.k` → **0 rows**. Correct (pre-PR, and PostgreSQL) result: 2 NULL-extended rows — the ON conjunct gates the inner join, so the RIGHT join's unmatched `s` rows must survive with a NULL master. The plan shows `Filter filter: m.c=1` *above* the right join.
- The channel is wider than equality: `ON x.k = m.k AND m.c > 1 RIGHT JOIN s ...` → 1 row instead of 2 (non-equality conjuncts reach `assignFilters` via `processJoinConditions`'s default case and the new fork reroutes them on its own). Function-bearing (`abs(m.c)=1`) and regex conjuncts regress the same way.

A LEFT-join control confirms the damage is specific to the new guard (the LEFT variant still pushes down). No test anywhere in the tree covers this shape, which is why CI is green.

**Suggested fix:** make the rerouting origin-aware. In `analyseEquals`, apply the `masterNullingJoinIndex` guard only for WHERE-origin calls (`joinIndex < 0`); in `assignFilters:1871`, reroute only when `!node.innerPredicate` and keep `addWhereNode` for ON-origin nodes (the flag is already populated and currently unread). Add regression tests for equality/non-equality/function ON conjuncts x RIGHT/FULL/SPLICE.

### C2 (in-diff, regression): the anchor is chosen in model-index space but executes in `orderedJoinModels` space — compile failures and a surviving leak

`masterNullingJoinIndex` (`SqlOptimiser.java:604-618`) picks the last nulling join **by model index**; `doReorderTables` appends context-less, dependency-less tables to the **end** of the execution order, and `generateJoins` compiles each model's post-join filter at its execution position (`SqlCodeGenerator.java:6272`). Two confirmed manifestations:

- **Compile error on previously-working queries:** `SELECT * FROM t0 CROSS JOIN t1 RIGHT JOIN t2 ON t2.k = t1.k WHERE t0.c = 1` → `SqlException: [74] Invalid column: t0.c`. Execution order is `[t1, t2, t0]`; the filter anchors at the RIGHT-join model, which executes before `t0` is joined, so the filter compiles against `(t1, t2)` metadata. Pre-PR the predicate pushed into `t0`'s scan and the query ran. A middle-position variant (`v1 CROSS JOIN v2 RIGHT JOIN v3 ON v3.b = v1.b WHERE v2.c = 1`) fails identically.
- **The leak this PR fixes survives reordering:** `SELECT * FROM wm RIGHT JOIN ws1 ON wm.x > ws1.y RIGHT JOIN ws2 ON ws2.q = wm.q WHERE wm.c = 1` — the non-equi join homogenizes to a context-less `CROSS_RIGHT` and executes **last**, above the filter anchored at model-index 2. Confirmed output: one leaked all-NULL row; correct result is 0 rows.

**Suggested fix:** decide placement in execution-order space — `assignFilters` already holds `parent.getOrderedJoinModels()` (line 1850); anchor at the nulling join that executes **last among those executing after the referenced table** (the pattern `mergeConstIntoPostJoinWhereClause:4995-4997` already uses). The first repro also shows the guard is over-broad there: in the actual plan `t0` joins after the RIGHT join and is never null-extended, so pushdown was both safe and optimal.

### C3 (in-diff, wrong-results regression): skipping constant registration resurrects stale `constNameTo*` entries — a *foreign* model's constant gets pushed into the nested slave

The guarded branches no longer write `constNameToIndex/Node/Token` (`SqlOptimiser.java:1681-1683`, `:1763-1765`). Those maps are instance state keyed by **bare column name**, cleared once per query (`clear():358`), and `optimiseJoins` recurses into nested models after processing the outer one — so an outer model's entry survives where pre-PR the nested model's constant overwrote it, and `addTransitiveFilters:1534-1556` matches it by (name, small-int model index), which collides trivially.

Reproduced: `SELECT * FROM tc JOIN (SELECT a.k k1, b.v FROM ta a RIGHT JOIN tb b ON a.k = b.k WHERE a.k = 1) x ON tc.q = x.k1 WHERE tc.k = 2` → **0 rows** (correct: 1). The plan shows the nested slave `tb` filtered by `k=2` — the **outer** query's constant. Pre-PR this query returned the correct row. The mechanism pre-existed via rarer skip paths (`= null`), but the diff widens exposure to the common `col = const`-on-nulling-master class.

**Suggested fix:** on guard rejection, remove/invalidate the bare name in all three maps (or scope the maps per `optimiseJoins` level).

### C4 (in-diff, test API): bind-variable arms are a hand-rolled single-pass `assertSql`

`JoinTest.java:5826-5829, 5917-5920, 5948-5951, 6545-6549, 6627-6631` — five new `printSql(bind); TestUtils.assertEquals(expected, sink)` arms. That is the body of the removed `assertSql` (one cursor pass; skips the second pass, `calculateSize()` cross-check, variable-column and factory-property checks) on fully deterministic queries — and the bind-vs-literal divergence is this PR's entire subject, so the bind arm deserves the stronger battery most. Verified: the builder executes against the same `sqlExecutionContext`/`bindVariableService` and `returns()` never clears binds, so `bindVariableService.setStr(...); assertQuery(bind).noLeakCheck().returns(expected)` works as-is; the purpose-built `.assertBinds(ObjList<BindVarTuple>)` terminal also exists (`QueryAssertion.java:172`, used by SampleByTest/LatestByTest/CoveringIndexTest).

### C5 (out-of-diff, confirmed by execution): four same-class holes remain — the stated fix is incomplete, and one hole trivially resurrects the fixed bug

All four violate the invariant the diff establishes ("master-only predicates must not be evaluated before a downstream nulling join"); all are pre-existing; all reproduced on the branch:

1. **`moveWhereInsideSubQueries` bypasses the fix for wrapped queries.** It runs *after* `optimiseJoins` (`optimise():11926` vs `:11938`) and pushes a wrapper's WHERE into the master with no nulling check. Wrapping this PR's own regression-test query — `SELECT * FROM (SELECT a.sym e0, a.c1 e1 FROM fm a RIGHT JOIN fs b ON a.sym = b.sym) WHERE e0 = 's2'` — returns the 2 leaked NULL rows the PR exists to eliminate (the plan shows `filter: sym='s2'` back inside `fm`'s scan). Confirmed for RIGHT and SPLICE, subquery and plain-table masters alike.
2. **Same-table equality:** the `analyseEquals` `lhi == rhi` path (`:1693-1702`) has no guard — `m RIGHT JOIN s ON m.k = s.k WHERE m.a = m.b` still pushes into the master scan; confirmed 2 rows vs the correct 1 (QuestDB `null = null` is true, so post-join placement keeps exactly one NULL row).
3. **Multi-table master filters (`rs >= 2`):** the untouched `assignFilters` else-branch (`:1883-1897`) anchors at the refs-complete join, below later nulling joins — `t0 JOIN t1 ON ... RIGHT JOIN t2 ON ... WHERE t0.a < t1.b` confirmed leaking `null,null,2` (correct: 1 row). The updated `testJoinClauseAlignmentBug` expectation now bakes this leaking placement into the suite for the right/full instantiations.
4. **Filters on barrier-joined tables:** `u0 LEFT JOIN u1 ON ... RIGHT JOIN u2 ON ... WHERE u1.x = 1` confirmed leaking (2 rows vs 1) via the same else-branch.

These can be follow-ups, but then the PR body should scope the claim explicitly and issues should be filed; items 2 and 3 look cheap to fix here (the guard belongs on the `lhi==rhi` branch, and the else-branch anchor can take `max(refsCompleteIndex, lastNullingExecutingAfter)`).

## Moderate

1. **The performance cliff is bigger than the body states, and a strictly better correct design exists for the dominant case.** Every intercepted predicate of the `col = non-null-literal` form is null-rejecting, which makes RIGHT OUTER ≡ INNER and FULL OUTER ≡ LEFT OUTER — after which the old pushdown (interval pruning, symbol index, parallel JIT, transitivity) is fully legal. QuestDB has no outer-join strength-reduction pass. As merged, such predicates get: a full master scan (the `testOrderByAdviceWithMultipleJoinsAndFilters` plan goes from interval-reducible to `Frame forward scan on: WorkflowEvent` — partition pruning on the designated timestamp lost), an always single-threaded interpreted `FilteredRecordCursorFactory` (join factories never support page frames, `SqlCodeGenerator.java:6272-6310`), and a full-table hash build. SPLICE is genuinely irreducible (prevailing-row counterexample verified), and `!=` / `= null` / bind variables must stay post-join under QuestDB NULL semantics — so the post-join machinery is needed regardless; strength reduction is the right follow-up, not a rewrite of this PR. The Tradeoff section should at least mention the parallel/JIT loss.
2. **Transitive slave constants are given up beyond what correctness requires.** With the master filter retained post-join, pushing the derived `slave.key = const` into the slave remains sound for RIGHT, FULL, and keyed SPLICE for non-null literals (derivation checked; must NOT re-enable for bind variables, which can be NULL at runtime). The codebase is now inconsistent: `analyseRegex:1785-1798` still registers and propagates `m.k ~ '...'` transitively into the slave of a RIGHT join while the strictly analogous equality no longer does.
3. **The new constant-false post-join filter shape isn't folded.** `testSpliceJoinFoldedFalseMasterFilterProducesNoRows` pins `Filter filter: false` over a full Splice Join — the engine executes the entire O(N+M) splice to emit nothing. The constant-fold-to-`EmptyTableRecordCursorFactory` pattern already exists in the const-where path (`SqlCodeGenerator.java:6327-6350`); the post-join path (`:6272-6310`) could do the same.
4. **Test coverage gaps in the fixed area:** (a) no data-level test of a non-equality master predicate that actually passes rows (the `assignFilters` fork carries most real predicates; its only data test folds to FALSE); (b) the timestamp/interval tradeoff has zero data tests, and the lone plan test is degenerate — the `to_timestamp(...)` bounds in `testOrderByAdviceWithMultipleJoinsAndFilters` don't match their format string and parse to NULL (that's why the plan renders `el.CreateDate>=)` with an empty operand); (c) the EmptyTable-master + splice-join compile interaction that the deleted test existed for is no longer exercised anywhere (reachable via `(SELECT * FROM t1 WHERE 1=0) a SPLICE JOIN t0 b`); (d) the stacked-joins test has no bind arm and covers only RIGHT+RIGHT.
5. **PR body gaps:** the Fix section describes only the first commit (the `JOIN_CROSS_RIGHT`/`JOIN_CROSS_FULL` non-equi extension — a production behavior change — is unmentioned); the test-plan bullet still says the EmptyTable test was "corrected" when the final diff moves and renames it; the three tests from the third commit are unlisted; no "Supersedes #7218" pointer to the prior review.
6. **Compile-path shape of `masterNullingJoinIndex`:** it wants the last match but scans forward with no early exit, and at `:1748-1752` the scan runs before `canMovePredicate` even though both outcomes converge for barrier-ON conjuncts (the `aSize==0` branch at `:1662-1665` orders the checks correctly). A backward scan returning on first hit — or one cached "last nulling index" per model, since the answer is `M > tableIndex ? M : -1` — makes each query O(1). Negligible in absolute terms; flagged per compile-path policy.

<details>
<summary><b>Minor</b> (8 items)</summary>

1. The three new inline comments (`:1670`, `:1751`, `:1877-1879`) say "SPLICE/FULL/RIGHT OUTER" but the code also matches `CROSS_RIGHT`/`CROSS_FULL` — exactly what `testNonEquiOuterJoinMasterFilterStaysPostJoin` exercises.
2. Javadoc at `:600` names "NestedLoopRight/FullJoin"; the actual classes are `NestedLoop{Right,Full}JoinRecordCursorFactory`, and the null record is built by `SqlCodeGenerator` (`:5529`, `:5538`) and passed in.
3. New JoinTest code mixes 3-line string concatenation (`:6530-6541`, expected strings at `:5817`, `:5903`) with the file's established text-block style (used at `:6614` in the same PR).
4. The 3-arg `assertQueryWithOuterJoinType` (`SqlParserTest:15004`) now duplicates the 4-arg body; it could delegate with `expected, expected`.
5. `testSpliceJoinMasterFilterStaysPostJoin` omits `bindVariableService.clear()` before its literal arm, unlike its siblings.
6. `"left".equals(outerJoinType)` dispatch couples silently to the literal contents of `outerJoinTypes`; a future entry would silently fall into the right/full arm.
7. First commit title uses a Conventional Commits prefix and is 59 chars; repo rules say commit titles take no prefix, <=50 chars (squash-merge makes this cosmetic). A few passive-voice slips in commit bodies.
8. Drive-by available: `EmptyTableRecordCursorFactoryTest` line 1 has a pre-existing `/*+***` license-header typo.

</details>

<details>
<summary><b>Downgraded</b> (false positives removed during verification, 6 items)</summary>

1. "No `assertBinds`-style helper exists on this branch" — disproven: `QueryAssertion.java:172` defines `assertBinds(ObjList<BindVarTuple>)`, with existing call sites in SampleByTest/LatestByTest/CoveringIndexTest.
2. "Plain-table masters escape the `moveWhereInsideSubQueries` bypass" — disproven by execution: the plain-table wrapped variant leaks identically (C5.1 covers all variants).
3. "`assignFilters`' new call could run O(models x filters x models)" — disproven: the `rs==1` branch consumes each node on first encounter (`postFilterRemoved`), so it's one scan per filter node.
4. "Two nulling joins could swap execution order in general" — for context-bearing joins the topo-sort emits ascending (safe); only the context-less CROSS variant reorders (that real case is folded into C2).
5. "Pooled `JoinContext` orphaned at `:1689` when the guard rejects" — benign: arena-style pool, reclaimed wholesale per compilation, `next()` clears state on reuse.
6. CodeRabbit's two pre-merge claims were both real (they are C5.2 and C5.4 here, now with executions proving them), but its suggested unconditional rerouting diff is not endorsed as-is — without origin awareness it would widen C1.

</details>

## Summary

The WHERE-clause core of the fix is sound and verified (membership list exactly right factory-by-factory; no later pass relocates post-join filters for unwrapped queries; UPDATE/materialized-view/PG-wire contexts safe; literal-vs-bind divergence closed for the tested shapes; concurrency and resource management all-clear). But three in-diff regressions are confirmed by execution — C1 (ON-clause conjuncts WHERE-ized → rows silently vanish), C2 (compile errors plus a surviving leak from model-index vs execution-order anchoring), C3 (stale constant map injects a foreign filter into nested queries) — plus the C4 test-API violation; and the four confirmed C5 bypasses mean the title's claim is broader than what the patch delivers, most visibly because wrapping a fixed query in `SELECT * FROM (...)` un-fixes it.

Process accounting: ~31 consolidated draft findings → 25 verified (4 Critical in-diff mechanisms, 4 Critical out-of-diff bypasses, 6 Moderate, 8 Minor), 6 downgraded as false positives. Every Critical was reproduced with live queries on the PR branch.

---

@puzpuzpuz one process ask: please run a self-review on the branch before requesting human review (e.g. `/review-pr <PR> --level 3`, or `/code-review ultra` for the cloud multi-agent pass). All three in-diff regressions above reproduce with 3-table probe queries that a self-review pass surfaces — catching them pre-publication is much cheaper for everyone.

<sub>🤖 Level-3 multi-agent review generated with [Claude Code](https://claude.com/claude-code); all Critical findings verified by executing probe queries on the PR branch.</sub>

puzpuzpuz:
@bluestreak01 thanks for the review, I'll address it. The PR is still in self-review mode, but my bad is that I forgot to put a "do not merge label" on it. I'll post success review text when it passes.

puzpuzpuz:
Code Review: PR #7225 — level 3
  
  fix(sql): fix wrong results in splice, right and full outer joins with a left-table filter

  This is the 7th fresh level-3 pass on this PR (it has been through ~15 review/fix sessions). The diff vs the true merge-base (4fe188379e, advanced by today's upstream merge) is clean: 5 
  production files (SqlOptimiser.java +429, SqlCodeGenerator.java +3, EmptySymbolMapReader.java, EmptyTableRandomRecordCursor.java, QueryModel.java) and 7 test files. I ran all mandated
  agents (correctness, cross-context, concurrency, resource, performance, adversarial-perf, fresh-adversarial, tests, quality, metadata; Rust N/A) plus my own empirical repro with plan
  inspection, and verified every draft finding against source.

  Critical

  None.

  Moderate

  None.

  Minor

  1. Perf — precomputeNullingJoinAnchors runs an O(J) sweep for inner-only multi-table levels (SqlOptimiser.java:6281). A query with no NULL-extending join still pays one linear anchor sweep
  per multi-table level that produces all--1 results. Bounded by join count and dominated by reorderTables' O(J²); a cheap early-out is possible but saves at most O(J). Consciously declined
  in prior sessions as negligible — restating for completeness, not blocking.
  2. Tests — bare "Filter filter:" plan fragments in SqlOptimiserTest (~lines 158, 183). A position-insensitive substring on "Filter filter:" alone can't distinguish a held post-join filter
  from a pushed one. Mitigated here by the paired "Frame forward scan on: WorkflowEvent\n" fragment (proves no interval pruning) and the corrected-row-count .returns() battery, which is the
  real guard. Low value to tighten.
  3. Quality — minor comment style. A handful of new comments in SqlOptimiser.java use mild passive voice ("the post-join anchor is chosen", "the const is registered"), and a few inline
  blocks (~1682–1687, 1727–1734, 1776–1782) run 6–8 lines against the project's 1–2-line preference. They document subtle leak-scenario invariants central to this PR, so the length is
  defensible. Cosmetic.
  4. Perf — criteriaHasCrossTableEquality double-traverses RIGHT/FULL ON = nodes (once in precomputeHasNonEquiNullingJoin, once in analyseEquals). A 2× constant factor on a cold compile
  path, bounded by ON-clause size, paid only when a RIGHT/FULL OUTER join is present. The natural cache (the join context) doesn't exist at precompute time. Acceptable.

  Downgraded (false positives removed after verification)

  - != / NOT IN master predicate "leaks" a null-master row past a RIGHT/FULL join (my own empirical probe — DISPROVEN). The plan dump is decisive: Filter filter: m.c!=99 sits above Hash 
  Right Outer Join Light — held post-join, identical to the correctly-held < 100. The kept row is correct QuestDB semantics: NULL != const is TRUE (EqIntFunctionFactory: negated != 
  (null==99) = true), verified by a single-table WHERE c != 99 returning the null row and a wrapped-subquery oracle matching the direct form. My probe applied standard-SQL three-valued
  logic; QuestDB's is different and consistent. No leak.
  - No integration test proves a parallel consumer of EmptyTableRandomRecordCursor.newSymbolTable survives (test agent — FALSE). Covered twice: the direct unit test
  (EmptyTableRecordCursorFactoryTest:53-54, assertSame+valueOf) and the integration test testEmptyMasterFeedsSpliceJoin, which projects the master SYMBOL → QueryAssertion.testSymbolAPI:858
  calls cursor.newSymbolTable(...) per symbol column and runs the clone/thread-safety pass. Pre-fix, that path threw UOE — it's how the bug was found.
  - Concurrency race on the new SqlOptimiser instance fields (FALSE). SqlOptimiser is per-SqlCompilerImpl, checked out from SqlCompilerPool via atomic CAS — thread-confined per compile. The
  3 precompute call sites never clobber a field mid-read (model-order pre-reorder, exec-order post-reorder, separate moveWhereInsideSubQueries pass; recursion happens after the level's
  reads).
  - Shared EmptySymbolMapReader.INSTANCE from newSymbolTable is unsafe / leaks (FALSE). The class is stateless (no instance fields), not Closeable; the only free paths are Misc.*IfCloseable
  which skip it. keyOf(null)→VALUE_IS_NULL mirrors SymbolMapReaderImpl; no caller of the singleton (WalWriter, CoveringIndex, Switch) reaches keyOf(null) in a way that regresses.
  - EmptyTableNoSizeRecordCursor missing the same override (out of scope). Genuinely unused in production (no factory returns it) — not reachable, not this PR's concern.

  Cross-context coverage

  Out-of-diff callsite review ran fully and found zero out-of-diff bugs (not an underrun): the keyOf(null) contract change was traced through every EmptySymbolMapReader.INSTANCE caller
  (WalWriter, ExtraNullColumn*, CoveringIndex, Switch); newSymbolTable through AbstractDeferredValueRecordCursorFactory (a real page-frame factory); and the masterAlias=null clear through
  the full generateJoins blast radius (set non-null only for the root table; HORIZON/WINDOW followers are parser-forbidden or self-consistent). All verified safe.

  Summary

  Verdict: approve on correctness grounds. The splice/right/full-outer master-filter fix family is exhaustively closed — every master-only predicate shape (=, !=, <, >, IN, NOT IN, BETWEEN,
  regex, col=col, two-table equality, wrapped, reordered-CROSS, barrier-joined, stacked) is held post-join, confirmed by plan inspection and a clean ~2318-test suite post-merge. The three
  secondary fixes (masterAlias clear, keyOf(null), newSymbolTable) are correct and don't regress any caller. The upstream merge introduced no regression.

  - Findings: 0 Critical, 0 Moderate, 4 Minor (all previously declined or cosmetic).
  - Verification: ~12 draft findings examined; the 2 most significant (my own !=/NOT IN "leak" and the test-coverage gap) were both proven false positives via plan dumps and source tracing.
  0 in-diff bugs, 0 out-of-diff bugs.
  - Tradeoffs (already disclosed in the PR body): master-side filters on these joins lose index/interval scans and run single-threaded/interpreted post-join; a two-table master equality can
  become a cross join + post-filter. These are the documented cost of correctness.
  - The DO NOT MERGE label is your self-imposed gate; nothing in this pass blocks lifting it. PR-body wording polish remains your call.

bluestreak01 (dismissed):
Approve.

Level-3 review (10 review agents + per-finding verification + ran 27 of the PR's regression/parser/plan tests — all green). No blocking issues.

This is a correct, well-tested fix for a real wrong-results class of bug in SPLICE/RIGHT/FULL outer joins with a master-side WHERE filter (the two count 2→0 changes in SqlOptimiserTest are genuine fixes), plus a real pre-existing -ea SPLICE-as-master crash fix. Tests pin both plan placement (post-join Filter vs push-down) and data, with literal-vs-bind parity.

Verified the one Critical raised by the bot is stale: the current code defers the same-table a.x=a.y case (SqlOptimiser line ~1732) and the barrier single-table case falls through to the nulling-anchor else branch in assignFilters. Concurrency/recursion-state, the shared EmptySymbolMapReader.INSTANCE from newSymbolTable, the keyOf(null) sentinel change, and the parseWhereClause innerPredicate preservation were all traced across callers and verified safe.

Non-blocking follow-ups:
- M1 (Moderate): transitive-const registration when the outermost downstream nulling join is a SPLICE is result-neutral only via addTransitiveFilters skipping SPLICE plus the unwritten invariant that a SPLICE master is never fed by a row-removing operator. Not a live bug (no counterexample found), but worth a defensive assert/comment.
- M2: add coverage for function-equality WHERE on a nulling master, SPLICE as a downstream nulling join over an earlier inner-joined table, a materialized-view regression, and a.c1=a.c2 as an ON conjunct.
- Minor: precomputeNullingJoinAnchors could early-out when no nulling join exists at the level; criteriaHasCrossTableEquality boolean name lacks is/has prefix; add .noLeakCheck() to the two new EmptyTableRecordCursorFactoryTest cases for consistency.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 156 / 158 (98.73%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|152|154|98.70%|
|:large_blue_circle:|io/questdb/cairo/EmptySymbolMapReader.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/EmptyTableRandomRecordCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/model/QueryModel.java|1|1|100.00%|

bluestreak01 (approved):
Approve.

Independent level-3 pass on current HEAD (`47989e4d3c`): 10 fresh-context review agents (correctness, concurrency, cross-context, resource, performance, adversarial bug-hunt, adversarial performance, tests, quality, metadata) + per-finding source verification + execution on the branch.

The fix is correct. The core invariant holds factory-by-factory and order-by-order: a table is on the NULL-extended master side of a downstream SPLICE/RIGHT/FULL (or homogenized CROSS_RIGHT/CROSS_FULL) join iff it precedes that join — checked in model order pre-reorder (`analyseEquals`) and execution order post-reorder (`assignFilters`/`moveWhereInsideSubQueries`), with `hasNonEquiNullingJoin` covering the context-less CROSS that reorders last. `node.innerPredicate` cleanly separates ON conjuncts (push down) from WHERE predicates (held post-join) in both directions, and the `parseWhereClause` change preserves that origin across re-parse. The three secondary fixes (`keyOf(null)`→`VALUE_IS_NULL`, `newSymbolTable`→shared empty instance, `masterAlias` clear after SPLICE) and the SPLICE transitive-prune skip are all sound and traced safe across every out-of-diff caller.

Verification:
- 0 Critical, 0 Moderate blockers. 2 non-blocking Moderate notes: the `analyseRegex` transitive-const path is unguarded but result-neutral today (`~` is null-rejecting; SPLICE skipped) — worth a justifying comment; and `!=`/`NOT IN` master filters have plan-only coverage, no kept-rows data test. ~10 Minor (cold-compile-path micro-opts bounded by join count, comment voice, optional `Performance` label).
- 7 draft-finding categories downgraded as false positives after verification: stale-field-across-recursion (traced SAFE), orphaned JoinContext (arena pool, benign), EmptySymbolMapReader double-free (not Closeable), the keyOf/newSymbolTable/masterAlias/innerPredicate callers (all SAFE), the `!=` "leak" (row correctly kept under QuestDB `null != const` = TRUE), and the stale CodeRabbit same-table/barrier bypass (current code guards both). 0 out-of-diff bugs.
- Tests green on the branch: JoinTest 223, SqlOptimiserTest 174, SqlParserTest 1081, SqlCompilerImplTest 309, EmptySymbolMapReaderTest 1, EmptyTableRecordCursorFactoryTest 3, the modified ExplainPlan test. ~4000 differential-fuzz queries vs an independent oracle across INNER/LEFT/RIGHT/FULL/SPLICE × 2–3 tables: 0 divergences. (The full-ExplainPlanTest fork SIGABRT is `simulate_crash`, an unmodified deliberately-crashing test function — unrelated to this PR.)

Tradeoffs (master-side filters lose index/interval scans and run single-threaded/interpreted post-join; two-table master equality can degrade to cross-join + post-filter) are accurately and prominently disclosed in the description — the documented cost of correctness.

Non-blocking follow-ups: add the M1 regex-guard comment and an M2 `!=` kept-rows data test; optionally add a `Performance` label.

PR #7227, opened 2026-06-09
fix(core): fix crash and native memory overrun in covering index out-of-order commits

bluestreak01:
## Problem

Two latent bugs on the covering-index incremental seal path, both present since the posting index feature landed (#6861), both reached through the same trigger: a pure-append out-of-order commit on a partition with a sealed posting index and more than `DENSE_STRIDE` (256) symbol keys, with the O3 batch touching only some strides.

**1. Crash after dropping an INCLUDE column.** Dropping a column that a POSTING covering index INCLUDEs, then committing a pure-append out-of-order batch, crashes the commit:

```
java.lang.AssertionError: maxCompressedSize: unsupported column type -1
  at io.questdb.cairo.idx.CoveringCompressor.maxCompressedSize
  at io.questdb.cairo.idx.PostingIndexWriter.writeSidecarFixedStrideForColumn
  at io.questdb.cairo.idx.PostingIndexWriter.writeSidecarStrideData
  at io.questdb.cairo.idx.PostingIndexWriter.sealIncremental
```

The error escapes the commit as `data is committed but posting-index seal failed` (critical) and distresses the writer. The fuzz test `WalWriterFuzzTest#testConvertPartitionToParquetWithCoveringIndex` found it (seeds `6069995328240L, 1781013513457L`).

**2. Silent native-memory overrun with wide INCLUDE columns.** `sealIncremental` sizes the shared dirty-stride sidecar scratch at 8 bytes per merged value (`totalStrideVals * Long.BYTES`), but `writeSidecarFixedStrideForColumn` assembles one key's raw values into that scratch at `(1 << shift)` bytes each — 16 for UUID/LONG128/DECIMAL128, 32 for LONG256/DECIMAL256, all legal INCLUDE types. A symbol key holding more than half (16-byte types) or a quarter (32-byte types) of a dirty stride's merged values writes past the allocation through unbounded `Unsafe.copyMemory` — one hot symbol is enough. No DROP COLUMN involved. Code review of this PR found it; a guard-zone probe over the scratch reproduced it deterministically: a hot key with 4,001 rows and UUID+LONG256 INCLUDE columns writes 93,984 bytes past the end of a 34,048-byte allocation. The overrun is silent in a plain run because `compressSidecarBlock` reads the same out-of-bounds bytes straight back, so the sidecar output stays correct while adjacent native heap is overwritten.

## Root cause

**Tombstone crash.** `DROP COLUMN` on an INCLUDE column tombstones its cover slot: `tombstoneCoveredColumnInOtherIndexes` writes `-1` into the index's covering-column list. At the next seal, `mapCoveringColumnsForSeal` deliberately encodes the tombstone as `type=-1`/`shift=0`, and `openSidecarFiles` leaves the slot's sidecar mem null with no `.pcN` file. Every sidecar writer skips such slots — except `writeSidecarStrideData` on the incremental seal path. The `shift=0` sentinel routes the slot into the fixed-stride writer, which rejects type `-1`. Mutating O3 commits seal through `rebuildSidecarsByCopy`/`sealFull`, which skip tombstones correctly, and with every stride dirty `sealIncremental` falls back to the tombstone-safe `sealFull` — which is why the existing drop-column covering tests never caught it.

**Scratch undersizing.** Both full-seal twins size their sidecar scratch by value width (`writeSidecarForColumn`: `strideValCount * valueSize`; `writeSidecarForColumnStreaming`: `maxKeyCount * valueSize`); only the incremental path assumed 8 bytes per value. The first commit's audit note ("no further unguarded type dispatch") was accurate but incomplete — this is a size dispatch, not a type dispatch, in the same `coverCount` loop family.

## Fix

- `writeSidecarStrideData` now skips tombstoned slots, the same check `writeSidecarsPerColumn` and `writeSidecarsPerColumnStreaming` apply on the full-seal path. The skipped slot publishes end offset 0 via `captureCoverEndOffsets` and covered reads fall back to NULL, matching the on-disk shape the full seal has always produced for dropped INCLUDE columns.
- `sealIncremental` now sizes the dirty-stride sidecar scratch by the widest live fixed-size cover column via the existing `peakCoverColumnValueSize()` helper, mirroring the full-seal twins.

Effects:

- Pure-append O3 commits no longer crash on covering indexes with dropped INCLUDE columns; live cover slots write byte-identical sidecar output.
- Incremental seals with 16/32-byte INCLUDE columns no longer write past the scratch allocation. With the guard-zone probe in place, the regression workload trips the guard before the fix (93,984 bytes out of bounds) and records zero out-of-bounds writes after it.
- The tombstone check adds one list load and a predicted not-taken branch per cover slot per dirty stride; no allocations or I/O change.
- The scratch grows by the ratio of the widest cover column to 8 bytes (2x for UUID covers, 4x for LONG256) while the seal runs. The allocation is transient — freed at the end of the seal — and tables whose INCLUDE columns are all 8 bytes or narrower allocate exactly as before.
- Covered reads of a dropped INCLUDE column continue to return NULL (pre-existing design); the fix does not restore covered data for tombstoned slots.

## Test plan

- new `CoveringIndexTest#testDropCoveredColumnThenO3AppendIncrementalSeal` reproduces the tombstone crash deterministically (fails on master, passes here) and runs under all three posting encodings via the `Delta`/`Ef` subclasses
- new `CoveringIndexTest#testIncrementalSealWideIncludeHotSymbol` pins the hot-key workload with UUID and LONG256 INCLUDE columns and verifies covered reads across dirty and clean strides under all three encodings; a temporary guard-zone probe over the scratch confirms this test overruns by 93,984 bytes with the old sizing and stays in bounds with the new one (the overrun is silent in a plain run, so the test cannot fail on master by itself)
- `WalWriterFuzzTest#testConvertPartitionToParquetWithCoveringIndex` with the captured seeds completes the full fuzz run on this branch; on master it fails in under a second
- all `Covering*` test classes: 1228 tests pass
- all `Posting*` test classes pass, including `PostingIndexO3ConcurrencyFuzzTest`, which fuzzes covering-index adds
- `WalWriterFuzzTest` seeded siblings (`testConvertPartitionToParquet`, `testConvertPartitionToParquetAllNullChunkIndex`, `testConvertPartitionToParquetPostingSquashReseal`) pass


mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|4|4|100.00%|

bluestreak01:
## Code review (level 3: all agents + per-finding verification + empirical before/after)

**Verdict: approve the fix.** Both code changes are correct, minimal, and well-targeted, verified by static analysis and by running the regression tests on reverted-vs-fixed code. **No merge-blocking correctness findings in the diff.** The findings below concern the PR/commit/test *accuracy* and *test reliability*, plus minor coverage/wording nits.

### The two fixes (both confirmed real and correct)

1. **Tombstone crash** (`writeSidecarStrideData`, `PostingIndexWriter.java:6364`). Dropping a covered INCLUDE column sets `coveredColumnIndices[c] = -1`; `TableWriter.mapCoveringColumnsForSeal:7616-7621` encodes the slot as `type=-1/shift=0`. On a pure-append O3 incremental seal, the `shift=0` sentinel routed the slot into `writeSidecarFixedStrideForColumn` → `CoveringCompressor.maxCompressedSize(_, -1)`, which does an **unconditional `throw new AssertionError`** (`CoveringCompressor.java:804` — fires even without `-ea`). The new `coveredColumnIndices.getQuick(c) < 0` skip matches the tombstone convention used at 6 other sites in the file and on all sibling seal paths. Reproduced deterministically on master; passes with the fix.

2. **Native-heap overrun** (`sealIncremental`, `PostingIndexWriter.java:5266`/`5319`). The shared dirty-stride scratch was sized at 8 B/value, but `writeSidecarFixedStrideForColumn` fills it at `(1<<shift)` B/value (16 for UUID, 32 for LONG256). Now sized by `peakCoverColumnValueSize()`. Provably sufficient: the writer resets `rawOffset` per key, needing `maxKeyInStride * valueSize`, and `totalStrideVals * sidecarValueSize ≥ maxKeyInStride * valueSize` for every column. The `int * long` multiply promotes to `long` (no overflow regression).

### Moderate

**M1 — The "silent / cannot fail on master" claim is empirically false; the bug-2 regression guard is non-deterministic.**

The PR body, the commit body of `a819caa4`, and the test-2 comment (`CoveringIndexTest.java:11814-11824`) all state the overrun is *"silent in a plain run … so this test cannot fail on master by itself."* I verified this by reverting only the production file to master, recompiling, and running `testIncrementalSealWideIncludeHotSymbol`:

- **This machine, 5 runs: 3 failed, 2 passed.** Two failures were wrong query results (`count` off by 1–2 — corruption flipped a hot-key LONG256 so it missed the filter); one was `CairoException: [27] No space left [size=7955134456386289664]` surfacing as **"data is committed but posting-index seal failed"** — i.e. the overrun corrupted a size/offset field, producing the *same class of critical error* attributed to bug #1.
- Second machine (sub-agent): 2 of 4 failed, also varying wrong counts.
- With the fix: passes (authoritative `Tests run: 1, Failures: 0`).

So the overrun is UB whose manifestation is allocator/environment-dependent — silent on the author's machine, loudly corrupting (wrong results *and* a bogus-size seal failure) elsewhere. Consequences: (1) the narrative understates severity (not benign-silent — it can corrupt results and abort the seal) and overstates the test's limits; the commit body is the permanent squash record, so the wording is worth correcting **before merge**; (2) as a guard against reintroduction the test is probabilistic per CI run (~0% on a "silent" environment), so a future regression could pass a green CI.

Recommendation: reword the three "silent / cannot fail on master" statements to reflect the non-deterministic wrong-result / corrupted-size-seal failure, and make the guard deterministic — commit the guard-zone probe used during development (assert zero out-of-bounds writes), or loop the seal/read so corruption is near-certain. Not a blocker for the fix itself.

### Minor

- **m1 — Missing combined-path test.** No test drops a *wide* (UUID/LONG256) INCLUDE column and then O3-appends, exercising both fixes together (test 1 drops an 8-byte DOUBLE; test 2 has wide columns but no drop). The skip is type-agnostic so it would handle it, but it's unverified. Also, no test asserts the dropped column reads back as **NULL** via a covered query (test 1 selects only `ts, sym`).
- **m2 — Comment wording.** The test-1 comment and commit body say *"coveringColumnIndices entry -1"*; the field is `coveredColumnIndices` (covered, not covering). Cosmetic.
- **m3 — Inherited over-allocation (informational).** The new line sizes the scratch by `totalStrideVals` rather than the true high-water `maxKeyInStride`, matching the full-seal twin `writeSidecarForColumn:6154`. The buffer is transient and grow-only-reused; the PR correctly preserves parity. Tightening would mean changing both paths together — out of scope here.

### Downgraded (verified false positives / non-issues)

- **Integer overflow in `totalStrideVals * sidecarValueSize`** — safe: `sidecarValueSize` is `long`; values bounded (≤256 keys/stride, ≤32 B/value); `totalStrideVals` already clamped at `:5218`.
- **Leak via the new `continue`** — safe: at loop top, inside `try`, before any per-iteration alloc; workspaces freed in `finally`.
- **realloc-OOM leak/double-free on `incrSidecarBuf`** — safe: on throw the old pointer + matching size are retained and freed in `finally`.
- **Concurrency** — safe: seal/tombstone run single-threaded on the TableWriter thread (asserted `o3PartitionUpdRemaining == 0`); `incrSidecarBuf` is method-local.
- **Reader returns garbage for a tombstoned slot** — false: `ensureSidecarOpen` early-returns on `publishedEnd <= 0` (`AbstractPostingIndexReader.java:400`), read falls back to NULL — byte-identical to the full-seal shape.
- **`peakCoverColumnValueSize` could under-size some column** — false: it maxes over exactly the set that writes to the buffer (live + fixed-size).

### PR metadata
Title/labels/commits all good (Conventional Commits, repeats verb, ASCII-clean bodies, active voice). Test builder API compliant (`.returns(...)`, no `.returnsOnce`/`assertSql`; `.noLeakCheck()` justified by multi-statement `assertMemoryLeak` lambdas). Only the M1 wording in the body/commit needs attention.

### Summary
- Approve the fix; request the M1 wording correction and, ideally, a deterministic regression guard. The code is correct and should ship.
- Only runtime tradeoff: transient 2×/4× larger scratch when 16/32-byte INCLUDE columns are present (disclosed accurately); ≤8-byte cover sets unchanged. No correctness regressions.
- Cross-context pass confirmed no out-of-diff breakage: the incremental path was the *only* seal path missing the tombstone skip, and this PR closes that gap.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) — review-pr level 3</sub>

kafka1991:
@bluestreak01 

# Code review (level 3: all agents + per-finding verification)

**Verdict: approve.** Both production changes are correct on every traced path. Ten review agents produced ~19 draft findings; after source verification, 9 findings stand (6 in-diff, 3 out-of-diff/pre-existing) and 6 were dismissed as false positives. Nothing blocks the merge; the Moderate items below are worth addressing before or immediately after it.

**Verified-correct core claims** (each independently confirmed by 2–3 agents): the tombstone skip is the same check the full-seal twins apply, and the resulting on-disk shape is byte-equivalent to full seal because every component (`.pci` -1 entry, missing `.pcN`, footer end offset 0) comes from shared code; `publishToChain` clears `coverEndOffsetsCache` on every new-sealTxn publish (PostingIndexWriter.java:3737) before `captureCoverEndOffsets`, so the PR's "publishes end offset 0" claim holds; the scratch sizing is sufficient (per key, `rawOffset` resets to 0 at :6055, writing `count * (1 << shift) <= totalStrideVals * peak`); readers and the planner can never require a tombstoned slot; the pre-fix crash fired **without `-ea`** (explicit `throw new AssertionError` at CoveringCompressor.java:804) and is reachable from production WAL apply; all 14 other `coverCount` loops on seal/gen-write paths already skip tombstones; no leak or double-free on any path including realloc-OOM and the `savedSidecarBufs` ownership transfer.

## Critical

None.

## Moderate

**M1 — Scratch sized by stride sum where the consumer needs only the largest single key (in-diff, PostingIndexWriter.java:5319).**
`writeSidecarFixedStrideForColumn` holds at most one key's values at a time (`rawOffset = 0` per key, :6055), so the tight bound is `maxKeyCount * sidecarValueSize`, not `totalStrideVals * sidecarValueSize`. The streaming full-seal twin already sizes exactly that way (`sidecarBufSize = (long) maxKeyCount * valueSize`, :6207). For a uniform 256-key dirty stride with a LONG256 cover this is a 256x over-allocation (e.g. 8 GB where 32 MB suffices), 4x larger than even the old broken sizing — and `sealIncremental` has no RSS pre-flight (the full seal has one at :4154), so the realloc can trip `Unsafe.checkAllocLimit` and fail seals that tight sizing would not. The max is free to compute: the loop at :5314-5317 already touches every `strideKeyCounts[j]`. Three agents converged on this independently, including the fresh-context one. Suggested fix: track `max` alongside the sum in that loop and size by it.

**M2 — The overrun regression test cannot detect a regression (in-diff, CoveringIndexTest.java:11826).**
The PR discloses this honestly: `compressSidecarBlock`'s wide-type arm is a raw copy that reads the same out-of-bounds bytes back (verified, PostingIndexWriter.java:1643-1647), so `testIncrementalSealWideIncludeHotSymbol` passes on master too. Combined with unpinned fuzz seeds (see Minor), the memory-overrun fix ships with no test anywhere that fails if it is reverted. A cheap permanent guard exists: pass the scratch capacity into `writeSidecarFixedStrideForColumn` and assert `((long) count << shift) <= sidecarBufSize` (static message) before the assembly loop at :6053. CI runs with `-ea`, so this turns any future re-break into a deterministic failure of this exact test, and also guards the two full-seal call sites.

**M3 — Test 1 stops one assertion short of verifying the surviving cover (in-diff, CoveringIndexTest.java:10533).**
Auto-include of the designated timestamp defaults to true (PropServerConfiguration.java:1602), so after `DROP COLUMN price` the cover set is `[-1, ts]` and `SELECT ts, sym` can still be served by the covering index — but without `.withPlanContaining("CoveringIndex")` the test passes identically under a full scan, leaving the live slot's re-encoded sidecar (exactly what the fix writes) unverified. Two cheap variants are also missing: `INCLUDE (a, b)`, drop `b`, covered read of `a` (immune to the auto-include config), and an all-slots-tombstoned run (auto-include off — the file already toggles that property elsewhere), which would exercise the all-null `sidecarMems` branches.

**M4 — Pre-existing int overflow in the same callee family, follow-up recommended (out-of-diff, CoveringCompressor.java:802-806).**
`maxCompressedSize` computes `4 + count * ColumnType.sizeOf(...)` in int arithmetic: for LONG256/DECIMAL256 it wraps negative at count >= ~67.1M (134M for 16-byte types, ~178M for DOUBLE). A negative `compressBufSize` makes `compressBuf = 0` (PostingIndexWriter.java:6032-6033), and `compressSidecarBlock` then writes through address 0 — a JVM-aborting SIGSEGV. One symbol key holding 67M+ rows in one partition is within QuestDB scale and below the existing 2^31 stride-sum guard (:4112), and both seal paths share the code. This PR's second commit notes the first audit missed a "size dispatch, not a type dispatch" — this is the remaining size-*arithmetic* sibling in the same loop family. Not introduced by this PR; deserves a tracked follow-up (long arithmetic plus an explicit bail like :4112).

## Minor

- **Naming:** `sidecarValueSize` (:5266) reads as a buffer size; the full-seal path names the identical quantity `maxColValueSize` (:4154). Rename for symmetry.
- **Comments:** "The full-seal twins already size by value width" is fix-relative phrasing with unnamed referents — rephrase as a named cross-reference (e.g. `writeSidecarForColumnStreaming`'s `maxKeyCount * valueSize`). Qualify `TableWriter.mapCoveringColumnsForSeal` (the file's habit is to qualify cross-class references). The "guard-zone probe" sentence in the test comment describes debugging methodology that no longer exists anywhere in the tree. Both new tests place their `// Regression` comments above `@Test`; all ~17 existing ones sit inside the method body.
- **Fuzz seeds not pinned:** `testConvertPartitionToParquetWithCoveringIndex` uses unseeded `generateRandom(LOG)` while its adjacent sibling pins seeds (WalWriterFuzzTest.java:291); the finder seeds live only in a test comment. Low priority for bug (a) given the deterministic repro, but ties into M2 for bug (b).
- **Pre-existing follow-up candidates in the touched functions** (none introduced or worsened by this PR): the int `totalStrideVals` sum is unguarded on the incremental path while the full seal bails explicitly at 2^31 (:4112); a tiny `dirtyStridesAddr` leak window at :5168-5202 if `estimateMaxPerKey`/`addressOf` throws between the two try blocks; scratch and workspaces are allocated even when all covers are var-size or tombstoned (a `hasLiveFixedCover` gate is nearly free); per-dirty-stride malloc churn (workspaces + per-column `compressBuf`) that both full-seal twins hoist; and deeper seal hot-path items — multi-GB sidecar snapshot taken even when the all-dirty fallback discards it unread, snapshot-by-copy where an RO mmap of the immutable old `.pc` suffices, scalar bit-pack fallback used where the AVX2 native pack exists, O(N) stride min/max rescan over per-key-sorted data, and a triple copy for the wide types this PR targets (gather -> memcpy "compress" -> `putBlockOfBytes`). Worth an issue, not this PR.

- *Stale `coverEndOffsetsCache` publishes a non-zero extent for the tombstoned slot* — refuted: the seal always takes `publishToChain`'s new-entry branch (sealTxn advances before publish), which clears the cache (:3737) before `captureCoverEndOffsets` zero-fills it.
- *`tombstoneCover` can race an in-flight `sealIncremental` (use-after-free of `sidecarMems`)* — refuted: both run only under the single pool-owning writer thread; the one cross-thread seal (ColumnIndexerJob inline seal) executes while the writer thread parks on `indexLatch.await()`.
- *Realloc-OOM leaks or double-frees on the changed line* — refuted: `Unsafe.realloc` leaves the old block valid on both failure modes, and `sealIncremental`'s finally zeroes the shared `savedSidecarBufs` entries before `seal()`'s guarded finally runs.
- *PR body's "1228 tests pass" is inaccurate* — dropped as unverifiable: two independent static counts disagreed with each other (398 vs 402 `@Test` in the class), so the runtime figure can't be refuted statically.
- *The crash needed `-ea` to manifest* — refuted: it's an explicit `throw new AssertionError`, not an `assert`; the pre-fix bug crashed production regardless of JVM flags.
- *WAL tables unaffected since tests use BYPASS WAL* — refuted: WAL apply reaches the identical seal sweep (`processWalCommitFinishApply` -> `processO3Block` -> `finishO3Commit`), and the finder fuzz test is itself a WAL test.

## Metadata

Title, labels, body tone, commit message format, and test plan all comply with conventions. Every numeric claim in the PR body verified exactly — including the 34,048/93,984/128,032-byte overrun arithmetic (stride 0 holds 4,256 merged values once the NULL key-0 offset is accounted for) and "present since #6861" (the unguarded lines first appear in a297d7742d). `Fixes #NNN` correctly absent — no matching GitHub issue exists.

## Summary

- Approve. The two fixes are minimal, correct, and the failure-mode trade (clean `CairoException` on allocation limit instead of silent native-heap overwrite) is strictly better.
- Main tradeoff to weigh before merge: M1 — the sum-based sizing institutionalizes up to 256x over-allocation on an unguarded path when the tight per-key bound costs one extra compare in an existing loop.
- 19 draft findings -> 9 verified (M1–M4 + 5 minor groups), 6 false positives removed, 4 absorbed/deduplicated.
- Split: 6 in-diff, 3 out-of-diff (all pre-existing, none caused by this PR; M4 is the one deserving a tracked follow-up).

RaphDal (approved):

ideoma (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 50 / 50 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/GroupByHistogram.java|46|46|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|4|4|100.00%|

PR #7229, opened 2026-06-10
fix(core): fix posting index crash and wrong covered reads after rollback

bluestreak01:
A `TableWriter.rollback()` on a table with a covering POSTING index (`SYMBOL INDEX TYPE POSTING INCLUDE (...)`) leaves the index's covering sidecars inconsistent with its row-id data. Depending on the data that follows, the next O3 commit either fails with a spurious `CairoException: [28] No space left [size=<multi-exabyte garbage>]` and a DISTRESSED writer, or silently corrupts covered reads. Covered queries issued between the rollback and the next seal also return NULL values for the covered columns.

PR #7107's CI found the crash (`WalWriterFuzzTest#testCreateTableAsParquet`, build 240303, linux-fuzz1) on its non-WAL oracle table. The bug pre-exists on master and involves no parquet.

## Root cause

`rollbackValues()` -> `reencodeMonolithic()` rewrites the row-id index into a fresh `.pv` at a bumped sealTxn, but never writes the `.pc{c}` covering sidecars at that txn. The next gen flush lazily creates the missing `.pc` file in append layout (`openSidecarFilesForAppend`), and the next incremental seal snapshots that file assuming the sealed stride-indexed layout: the clean-stride copy interprets raw appended cover bytes as stride offsets and passes a garbage size to `putBlockOfBytes`.

## Changes (all in `PostingIndexWriter`)

- `rollbackValues()` returns early when no indexed value lies above the rollback point, mirroring `rollbackConditionally`'s existing guard. Previously every `TableWriter.rollback()` paid a full index reencode, a sealTxn bump and queued purge work even when it discarded nothing -- the common case, since uncommitted rows are not indexed yet.
- `reencodeMonolithic()` rebuilds the covering sidecars at the new sealTxn (via the same `writeSidecarsPerColumn` the seal path uses) when the name-based cover sources are configured. A published sealTxn therefore never advertises covered data without sealed `.pc` files: covered reads in the rollback-to-next-seal window stay correct, and the next seal can stay incremental.
- `seal()` validates each cover's sidecar layout on the read-only mmap before paying the snapshot copy, and falls back to `sealFull` (which rebuilds sidecars from column data) on the first live cover whose file is missing or structurally not a sealed file, without copying anything. Trusted covers copy only the sealed region (`sentinel + siSize` bytes): the file tail past it holds raw post-seal gen blocks that no snapshot consumer reads, while gen header offsets still validate against the real file length they legitimately point into. The validator also bounds the sealed end with an overflow-safe subtraction; a corrupt near-`Long.MAX_VALUE` sentinel previously overflowed past the `prev + siSize <= bufSize` form and marked the file trusted, turning one torn write into a negative-size allocation. The two layouts are deterministically distinguishable: an append-layout file's first gen block starts at exactly `PC_HEADER_SIZE`, below any valid sealed-region end. This also repairs on-disk states left behind by the bug before this fix, stale sidecars from abandoned seal txns, and rollbacks that run without cover sources configured.
- `sealIncremental()`'s clean-stride copy bounds every offset it reads and uses the stored stride-index sentinel as the last stride's upper bound instead of the snapshot length. This fixes a second, independent defect: post-seal gen flushes append raw cover blocks to the sealed `.pc` file, and every incremental seal whose last stride stayed clean swept those blocks into the new sealed sidecar -- the new regression test measures 196,608 bytes of sidecar growth for 160,000 bytes of appended cover data, compounding on every following incremental seal. Any residual mismatch now raises a descriptive `CairoException` instead of attempting a multi-exabyte allocation.
- `reencodeMonolithic()` and `truncate()` sync the `.pk` chain publish before recording the seal purge of the superseded files, mirroring `seal()`'s unconditional sync and `rebuildSidecarsByCopy`'s publish-then-sync. This fixes a third, pre-existing defect: the rollback paths published the new chain entry only to page cache while queueing the old `.pv`/`.pc` for unlink, and `PostingSealPurgeOperator` removes them with no ordering against `.pk` writeback. Under the default NOSYNC commit mode, a power loss after the unlink became durable but before the `.pk` page wrote back recovered a committed chain head pointing at a deleted `.pv` -- readers fail hard at open (`mapValueMem` has no missing-file tolerance, unlike the `ff.exists`-guarded sidecars) and only REINDEX repairs the index, strictly worse than NOSYNC's lose-recent-writes contract. On master every rollback was exposed; the early-out above narrows that to real discards, and the sync closes the remaining window.

## Costs and limitations

- Real-discard rollbacks (indexed rows above the committed boundary, e.g. crash-recovery eviction) now pay the sidecar rebuild during the rollback. No-op rollbacks -- the common case -- become strictly cheaper than before (no reencode at all).
- Real-discard rollbacks and index truncations also pay one synchronous `.pk` msync before queueing purge work -- the same barrier every seal already pays.
- The snapshot validation adds a few hundred comparisons per incremental seal, on the read-only mmap the snapshot maps anyway. A distrusted cover now skips the snapshot copy entirely instead of copying multi-GB files and freeing them unread.
- When validation distrusts a sidecar, that one seal degrades to `sealFull` for the affected partition.
- The rollback-time sidecar rebuild is gated on the name-based cover source (the addr-based pointers belong to an O3 job's processing window and may be stale by rollback time). A rollback without cover sources still leaves a sidecar-less txn; validation then forces `sealFull` at the next seal, but covered reads in that window return NULLs as before. All `TableWriter` rollback paths configure the name-based source before rolling back.
- The fd-based rollback branch (`reencodeMonolithic` with an empty partition path) keeps its existing behavior: it reuses `valueMem` without bumping sealTxn. Whether its sealed-sidecar/gen-0 misalignment is reachable needs a separate analysis. The new `.pk` sync is also path-based only: the fd-based branch never bumps sealTxn, so it queues no purge and needs no barrier.

## Test plan

- `testRollbackThenIncrementalSealMisreadsAppendModeSidecar` -- minimal crash reproduction: stage-and-rollback, then O3 append touching one stride. Fails on master with `Cairo [28] No space left [size=4294967297048576]`.
- `testReopenEvictionThenIncrementalSealRebuildsSidecar` -- real-discard variant: plants index rowids beyond the committed boundary (the state a crash between index publish and txn commit leaves), reopens the writer so `rollbackConditionally` runs the reencode, and asserts covered reads both inside the rollback-to-seal window and after the next commit. On master the in-window covered read returns `sum=null`.
- `testIncrementalSealDistrustsRecreatedSidecar` -- deletes the `.pc` at the published sealTxn (the pre-fix on-disk state), appends and commits; asserts a clean-stride key's covered sum, which master silently corrupts.
- `testIncrementalSealLastCleanStrideStopsAtSentinel` -- two O3-append commits; asserts the sealed sidecar grows by the compressed delta only. Master grows it by 196,608 bytes.
- `testIncrementalSealRejectsOverflowingSidecarSentinel` -- corrupts the sealed sidecar's stride-index sentinel with a near-`Long.MAX_VALUE` offset (monotonic, so only the sealed-end bound can reject it) and drives an incremental-shaped seal over it; the seal must distrust the file, reseal full and keep covered reads correct. With the overflowing bound the seal dies in a negative-size allocation; red-proven on this branch.
- `testRollbackRealDiscardSyncsKeyFileBeforePurgeUnlink` and `testRollbackConditionallyToZeroSyncsKeyFileBeforePurgeUnlink` -- attribute `msync` calls to the `.pk` mapping via a `FilesFacade` that tracks the file's fd and mapping range, and assert the rollback reencode and truncate paths sync `.pk` inside the rollback call, before the superseded files become purgeable. Both fail without the sync.
- The six tests above other than the sentinel-overflow one fail on master `15b6d72121` and pass with the fix.
- Green: `PostingIndexCriticalIssuesTest` (93), `CoveringIndexTest` (396), `PostingIndexNativeTest`, `PostingIndexOracleTest`, `PostingIndexConcurrencyTest`, `PostingIndexOomFallbackTest`, `PostingSealPurgeTest`, `PostingIndexO3ConcurrencyFuzzTest`, `O3MaxLagFuzzTest`, `DedupInsertFuzzTest`, `WalWriterFuzzTest`.


questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=240550)
**Tested Commit:** `0af39d94b248c8a3c8aff95577066da5bd769a9f`

bluestreak01:
## Code review (level 3 — adversarial, blocking)

A thorough multi-pass review (correctness, concurrency, performance/algorithmic optimality, resource management, tests, cross-context callers, plus two checklist-free adversarial passes), with every finding re-verified against source.

**Verdict: approve with minor changes.** The core fix is sound and well-tested. No correctness, concurrency, NULL-handling, or happy-path resource-leak defect. The on-disk-format reasoning (the sidecar-snapshot validator, the sentinel/overflow bound, the `.pv` → sidecars → `.pk` → purge sync ordering) holds up under adversarial analysis. 6 of the 7 new tests are genuinely red on master (re-verified by swapping master's `PostingIndexWriter.java` under the new test file). The findings below are robustness, efficiency, and test-coverage — not correctness.

Cross-context: every callsite of the changed public/override symbols (`rollbackValues`, `seal`, `truncate`) was cleared as SAFE. The `rollbackValues` early-out is provably unreachable from all `rollbackConditionally` callers (that guard already filters `getMaxValue() < row`), and safe from its one reachable POSTING caller, `TableWriter.rollbackIndexes()`. `SymbolMapWriter`'s `rollbackValues` callers use `BitmapIndexWriter`, so they're unaffected.

### Critical
None.

### Moderate

**M1 — `reencodeMonolithic` leaks the new `.pv`/`.pc` files if the sidecar rebuild throws.** `PostingIndexWriter.java:4637-4722`

The diff moves `switchToSealedValueFile(newTxn)` (4637) ahead of the new sidecar rebuild and `publishToChain` (4702). The post-switch window now contains `openSidecarFiles` (4658), `Unsafe.malloc` (4675), `writeSidecarsPerColumn` (4677 — per-cover `mmap` + FSST compression + sidecar disk writes) and the sidecar `sync` loop (4685). A throw there (ENOSPC on a sidecar extend, native OOM, an mmap/JNI failure) unwinds through only the two native-scratch `finally` blocks — there is **no `catch (Throwable)`** that unlinks the staged files.

- The sibling `rebuildSidecarsByCopy` (3951-4048) uses the *identical* early-switch ordering and handles exactly this: `catch (Throwable) { if (isSwitched) scheduleOrphanPurge(newSealTxn); else { Misc.free(sealValueMem); closeSidecarMems(); unlinkOrphanSealFiles(newSealTxn); } throw th; }`. `reencodeMonolithic` did not port that cleanup.
- On master, `reencodeMonolithic`'s post-switch window was only `storeFence + publishToChain`, so the leak window was negligible. This PR widens it into a wide, I/O-heavy window. The gap is pre-existing in kind but materially widened here.
- It is a **leak, not corruption**: the throw propagates before `rollbackToMaxValue`'s `recordPostingSealPurge(oldSealTxn)` (5187-5189), so the old `.pv` is not queued for purge and the chain head still references `oldSealTxn` — readers and recovery stay correct. But `.pv.{newTxn}`/`.pc{newTxn}` are orphaned, and recovery is a chain walk (no filesystem orphan scan), so they aren't reliably reclaimed. The writer is also left with inconsistent in-memory state (`sealTxn=newTxn`, `genCount=1`, chain head at `oldSealTxn`).
- The most likely trigger is disk-full during the rollback rebuild — leaking more disk on a disk-full path is a bad feedback loop.
- **Fix:** wrap the block from `switchToSealedValueFile` through `publishToChain` in the same `try/catch` as `rebuildSidecarsByCopy` (track `isSwitched`; `scheduleOrphanPurge(newTxn)` post-switch, `unlinkOrphanSealFiles(newTxn)` + close pre-switch). Add a fault-injection test (ENOSPC/OOM in `writeSidecarsPerColumn` during a real-discard rollback) asserting no orphaned `.pv`/`.pc` survives.

**M2 — Redundant O(keyCount) pass to compute `maxStrideTotal` on the rollback path.** `PostingIndexWriter.java:4662-4673`

The new loop sums per-stride row counts to size `strideValsAddr` — O(keyCount). But the method's own stride encode loop at 4552-4564 already computes the same per-stride total (`totalStrideValuesL`) for every stride and discards it. Tracking `maxStrideTotal = Math.max(maxStrideTotal, totalStrideValuesL)` inside that loop eliminates the separate pass (O(keyCount) → O(0) extra). At the ~11M-key scale the file's own comments cite, that is ~11M redundant `Unsafe.getInt` + adds per real-discard rollback (not rare under O3-heavy ingest). Cold-ish path, so Moderate, but it's a free, clean elimination.

### Minor

- **m1 — Dead `Math.min`.** `PostingIndexWriter.java:1407`: `sealedLen = Math.min(fileLen, sentinel + gen0SiSize)` can never select `fileLen` — the just-passed validator guarantees `sentinel + gen0SiSize <= fileLen`. Simplify to `long sealedLen = sentinel + gen0SiSize;`. Harmless but misleading.
- **m2 — Missing test for the covered-index *no-op* early-out.** The early-out (1289) is what prevents the bug in the common case, yet no new test exercises it for a *covered* POSTING index (the real-discard path is covered by `testReopenEviction...`; the no-op path is only incidentally line-covered by the non-covered `PostingIndexStressTest.testRollbackNoop`, which also passes on master). Low risk (the early-out is covering-agnostic), but a covered no-op-rollback-then-seal test would lock in the intent.
- **m3 — Redundant guard.** `PostingIndexWriter.java:4655` (and 1438): `if (sidecarMems.size() > 0) closeSidecarMems()` — `closeSidecarMems()` is already null-safe and a no-op on an empty list. Drop the guard. Not a bug.
- **m4 — Title bundles two symptoms.** Acceptable (it repeats the verb), but `...crash and wrong covered reads after rollback` reads as two release-note items. Optional: `fix(core): fix posting index crash and corruption after rollback`.
- **m5 — Latent `maxValue`-contract divergence (unreachable in production).** When `getMaxValue() < maxValue` *strictly*, the early-out returns without `setMaxValue`, whereas `BitmapIndexWriter.rollbackValues` sets it unconditionally. The sole production caller (`SymbolColumnIndexer.rollback(transientRowCount-1)`) always has `getMaxValue() == transientRowCount-1` for a symbol column, so the strict branch is unreachable. Worth a one-line comment noting the two `IndexWriter` impls now differ here.

### Must investigate before merge (not dismissable)

- **Enterprise CI failed** on the latest commit (`0af39d94b2`, build 240550). Per the project's own rule, this needs to be proven flaky/unrelated (with the proof reported) or fixed — not assumed. Please pull build 240550's failing job and confirm it's unrelated to this change.
- **Coverage: 12 uncovered new lines (66/78).** Verified breakdown: 8 are the `sealIncremental` structural-invalid `CairoException` (5491-5498), provably unreachable after a trusted snapshot (all four `||` conditions are false by the validator's guarantees) — acceptable as defensive. The other 4 (validator `bufSize < sealedRegionMin`; the first-entry/monotonic rejects at 3465/3470; the rollback `closeSidecarMems` at 4656) are low-risk gated branches; a corrupt-middle-stride-entry test would close two of them.

### Checked and cleared (false positives from the review pass)

- **`.returnsOnce(...)` on deterministic `count()`** (lines 5291/5586/5674/5764/5902): these are **pre-existing** test methods (e.g. `testSquashNonCoveringPostingWithMidStreamSpillFlush`), not added/modified by this PR — the diff adds 0 `returnsOnce` and 10 `.returns(`. Out of scope.
- **The validator's 128-slot gen-header scan**: bounding it by live `genCount` instead of `MAX_GEN_COUNT` would be a *correctness* regression (it must reject stale/append files with non-zero slots beyond the expected count). 128 longs off an already-resident mmap is negligible. Correct as written.
- **Concurrency / publication ordering**: `.pv` + sidecars are synced before the `.pk` chain publish, and `.pk` is synced before the superseded files are queued for purge. The early `switchToSealedValueFile` does not change what a concurrent reader sees (readers resolve file names only from a chain entry picked under the `.pk` seqlock; the chain still references `oldSealTxn` until publish). Single-writer confirmed.
- **The `switchToSealedValueFile` reorder is required, not a bug**: `writeSidecarsPerColumn` reads `valueMem.addressOf(0)` (the freshly-sealed gen 0), so the switch must precede it — matching `reencodeWithStrideDecoding`/`reencodeWithPerKeyStreaming`/`rebuildSidecarsByCopy`.
- **Resource double-free across the new `sealIncremental` throw**: safe — `sealIncremental` zeroes the shared `savedSidecarBufs`/`oldSidecarBufs` slots (5634) before `seal()`'s `finally` runs, so the outer `finally` no-ops.
- **Validator false-trust of an append-layout `.pc`**: not possible — `openSidecarFilesForAppend` jumps a fresh file to `PC_HEADER_SIZE` and `writeSidecarGenData` records the first block's offset (= `PC_HEADER_SIZE`) into a gen-header slot, which is `< sealedRegionMin`, so the gen-header check rejects it.

### Summary

- **2 Moderate, 5 Minor verified; ~6 draft findings dropped as false positives.** All verified findings are in-diff.
- **Tradeoffs the PR documents honestly**: real-discard rollbacks now pay a sidecar rebuild + one `.pk` msync; distrusted covers degrade one seal to `sealFull`; rollback without name-based cover sources still returns NULL covered reads in-window; the fd-based branch's reachability is left explicitly unanalyzed.
- **Highest-priority action:** M1 (orphan cleanup, for consistency with `rebuildSidecarsByCopy` and to avoid the disk-full feedback loop), then resolve the Enterprise CI failure. Everything else is optional polish.

nwoolmer (commented):
LGTM after discussion points on M1, M2, m1-3, m5 are handled

ideoma (dismissed):

bluestreak01:
## Code review (level 3 — re-review after consolidation)

Full mission-critical pass (correctness/resource, fresh-context adversarial, crash-consistency, tests/metadata, cross-context callers), every finding re-verified against source. This is a re-review: the branch was consolidated into one commit since the earlier review, so I first verified which of those points were actually addressed, then hunted the delta.

**Verdict: approve with minor changes.** No correctness, concurrency, NULL-handling, crash-consistency, or resource-leak defect survived verification. The on-disk-format reasoning (sidecar-snapshot validator, overflow-safe sentinel bound, the `.pv -> sidecars -> .pk -> purge` sync ordering) holds up adversarially; all five durability/concurrency claims were independently confirmed safe. The 8 new tests pass on the branch. The findings below are **test-quality and documentation** -- not correctness.

**Status of the earlier review's points (verified against current source):** M1 (orphan-cleanup leak) -- **fixed**, now with a dedicated fault-injection test (`testRollbackRealDiscardSidecarSyncFailureUnlinksStagedSealFiles`); M2 (redundant O(keyCount) pass) -- **fixed** (`globalMaxKeyCount` folded into the existing filter loop, 4466-4468); m1 (dead `Math.min`) -- **fixed** (now `sealedLen = sentinel + gen0SiSize`); m3 (redundant `sidecarMems.size()>0` guard) -- **fixed** (removed). m5 (the `maxValue`-contract divergence) is **still open** -- see m5 below, so the conditional LGTM isn't fully satisfied yet.

### Critical
None.

### Moderate

**M-A -- `testRollbackThenIncrementalSealUsesRebuiltSidecar` does not exercise the code it claims; it silently guards a different fix.** `PostingIndexCriticalIssuesTest.java:500-558` (in-diff)

The test stages **one uncommitted O3 row** and rolls it back, then asserts covered sums. Its docstring claims this "reencodes the .pv at a bumped sealTxn and must rebuild the matching sealed .pc sidecar," and that the next seal "copies the clean stride verbatim from the rollback-rebuilt sidecar." It does neither: this PR's own early-out (`rollbackValues`, line 1286: `if (getMaxValue() <= maxValue) return;`) short-circuits the rollback to a no-op here -- exactly the case the PR body describes (*"No-op rollbacks -- the common case, since uncommitted rows are not indexed yet."*). So `reencodeMonolithic`, and therefore `writeSidecarsPerColumnFromDecodedValues` (the headline new method of this PR), is **never called** in this test. Instrumentation confirms it: the guard fires (`curMax == reqMax == 3099`), no reencode runs, and the subsequent seal trusts the *original* merge-commit sidecar.

- The test would still pass if `writeSidecarsPerColumnFromDecodedValues` / `writeSidecarForColumnFromDecodedValues` were deleted entirely -- false coverage for the PR's central change. It remains a valid regression guard (red on master via the no-early-out reencode crash), but for the early-out, not the rebuild.
- The genuine rebuild coverage lives in `testReopenEvictionThenIncrementalSealRebuildsSidecar` (plants rowids above the committed boundary -> real discard -> reencode -> rebuild; asserts a dirty-stride and a clean-stride key).
- **Fix:** rewrite the docstring/inline comments to state this guards the *no-op early-out path*, **or** rework it to plant indexed rowids above the committed boundary like the eviction test so it actually reaches the rebuild.

**M-B -- The var-size cover-column rollback-rebuild path (`shift < 0`) is entirely untested.** `PostingIndexWriter.java:6439-6440` (in-diff), tests out-of-diff

`writeSidecarForColumnFromDecodedValues` routes var-size covers (STRING/VARCHAR/array INCLUDE columns) to `writeSidecarVarStrideData`. Every new test uses `INCLUDE (price)` (DOUBLE, `shift=3`) or a LONG cover; none exercises `shift < 0`. This is brand-new code on a corruption-sensitive path -- and the offset-feeding differs from the seal path (global offsets into `allValuesAddr` vs per-stride scratch), which is exactly where a subtle var-size bug would hide.

- **Fix:** add a real-discard (eviction-style) rollback test with a STRING or VARCHAR `INCLUDE` column, asserting covered string reads survive both in the rollback-to-seal window and after the next incremental seal.

### Minor

- **m1 -- Inverted stride labels in `testRollbackThenIncrementalSealUsesRebuiltSidecar` comments.** `PostingIndexCriticalIssuesTest.java:534-536`: `'k0'` is key id 299 -> stride 1 (the last stride), so touching `k0` dirties **stride 1**, not "stride 0 dirty, stride 1 clean". The two later tests state the mapping correctly. Comment-only.

- **m2 -- PR body cites a test name that doesn't exist.** The test plan lists `testRollbackThenIncrementalSealMisreadsAppendModeSidecar`; the actual method is `testRollbackThenIncrementalSealUsesRebuiltSidecar`. Update the body (and per M-A, also fix its claim that this test exercises the rebuild).

- **m3 -- `ServerMainCleanStartupTest` change is an unrelated drive-by.** `ServerMainCleanStartupTest.java:67-109` adds `.noMemoryUsageCheck()`/`.expectSize()` to three system-function queries -- unrelated to the posting index. It is *not* masking a regression this PR introduced (posting code can't affect WAL/writer-pool RSS; the 64 KiB RSS-delta bound is a legitimate false positive there), so it's harmless, but per "don't bundle unrelated fixes" it should be split out or called out in the PR body.

- **m4 -- The "validation then forces `sealFull` at the next seal" claim is imprecise.** `seal()` returns early at line 1315 (`genCount == 1 && gen0KeyCount >= 0`) **before** the snapshot validator runs, so a covered-but-sidecar-less sealTxn is only rescued at the next seal that appends a *new* gen, not the immediate next seal. The cross-context pass proved the bad precondition (rollback reaching `reencodeMonolithic` with `coverCount>0` but no name-based cover source) is **unreachable** in production -- all rollback paths reconfigure name-based covers first -- so this is documentation only. Cheap hardening: `assert !(coverCount > 0 && !hasCoverSources)` in the path-based rollback branch.

- **m5 -- `rollbackValues` no longer satisfies the `IndexWriter` post-condition `getMaxValue() == maxValue`** (still open from the earlier review). `PostingIndexWriter.java:1286`. `BitmapIndexWriter.rollbackValues` unconditionally `setMaxValue(maxValue)`; the new early-out returns without it, so when `getMaxValue() < maxValue` *strictly* the two impls diverge. The sole production caller (`SymbolColumnIndexer.rollback(transientRowCount-1)`) makes the strict-`<` branch unreachable for a symbol column, so it's latent -- but make it a deliberate, commented decision rather than silent.

- **m6 -- Dead `!pathBased` branch in `reencodeMonolithic` has no throw-cleanup.** `PostingIndexWriter.java:4495-4498`. fd-based open for posting indexes throws at construction, so this branch is unreachable (consistent with `switchToSealedValueFile`'s `assert partitionPath.size() > 0`). If ever revived, a mid-encode throw would leave `valueMem` half-overwritten with no restore. Drop the branch or guard it with the same assert.

### Downgraded (raised during the pass, dismissed after source verification)

- **"M1 orphan-cleanup leak still open"** -- *fixed.* `reencodeMonolithic` now wraps switch+publish in `try/catch(Throwable)` mirroring `rebuildSidecarsByCopy` (pre-switch: `Misc.free(sealValueMem)` + `closeSidecarMems` + `unlinkOrphanSealFiles` + state-restore; post-switch: `scheduleOrphanPurge`). `Misc.free` on the `final` reusable `sealValueMem` is idempotent -- no double-free. Red-proven by the new sync-failure test.
- **"Rebuilt sidecar layout may differ from the seal path"** -- *false.* `writeSidecarForColumnFromDecodedValues` produces a byte-identical sealed layout; the only difference is global offsets into `allValuesAddr` vs per-stride scratch, both resolving to the same surviving row-ids. Validated by `testReopenEviction...`'s clean-stride `k1` sum.
- **"`globalMaxKeyCount` mis-sized"** -- *false.* It's `max(survivingCount)`; `survivingValueCount==0` truncates and returns before the rebuild, so it's >=1 and >= every per-key count its workspaces serve.
- **"Power-loss window strands a dangling chain head"** -- *false.* New `.pv`+sidecars synced before `publishToChain`; `.pk` synced after publish; old files queued for purge only after return. `PostingSealPurgeOperator` unlinks both `.pv` and `.pc` under one scoreboard-gated entry, strictly after the `.pk` msync. `truncate()`'s added `keyMem.sync` makes the empty header durable before the old `.pv` is purgeable.
- **"Early `switchToSealedValueFile` exposes the new `valueMem` to readers"** -- *false.* The swap is writer-local; readers resolve `.pv` only from a chain entry read under the header seqlock, which `publishToChain` advances last.
- **"New `assertSql`/`returnsOnce` violations"** -- *none.* The diff adds 10 `.returns(` on deterministic `count()/sum()` queries and zero `assertSql`/`returnsOnce`. The `assertMemoryLeak`+`.noLeakCheck()` blocks are legitimate (each scope holds DDL + native allocs beyond the single builder call).
- **"Facade tests pass without the fix"** -- *false.* On master nothing syncs `.pk` in `truncate`/`reencodeMonolithic`; the `PkSyncCountingFacade`/`PcSyncFailingFacade` fd->mapping attribution is exact, and their tracking maps are Java-heap (no native leak under `assertMemoryLeak`).

### Summary

- **2 Moderate (both test-quality) + 6 Minor (documentation/hardening) verified; ~9 draft findings dropped as false positives.** All verified findings are in-diff (test files + one production comment/assert).
- **Cross-context:** every external callsite of `rollbackValues`/`seal()`/`truncate()`/`reencodeMonolithic` cleared as SAFE -- the early-out is provably safe at the one reachable production caller, and `BitmapIndexWriter`/`SymbolMapWriter` are unaffected.
- **Highest-priority actions:** (1) M-A -- fix the misleading test so the headline rebuild code is actually covered (or document that the eviction test is the real guard); (2) M-B -- add a var-size cover rollback test; then the Minor doc fixes. The earlier review's m5 is still owed.
- CI for the head commit (build 240658) was still **in progress** at review time; the earlier failed Enterprise build (240550) is stale/superseded. Per the repo rule, confirm the new Cairo/Griffin/Enterprise jobs land green before merge.

nwoolmer (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 240 / 279 (86.02%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|240|279|86.02%|

PR #7231, opened 2026-06-10
fix(qwp): fix unresponsive network I/O under heavy QWP ingestion

jerrinot:
A client streaming data fast over a QWP WebSocket could pin an HTTP worker thread, stalling other connections: REST queries and other WebSocket sessions saw elevated latency until the stream paused. this could also delay responsiveness to a shutdown signal.

The connection now returns to the dispatcher after each recv chunk instead of looping until the socket drains, so each worker rotates across all ready connections. Also: the HTTP keep-alive check no longer applies to upgraded connections on the receive path.

### DIFF
--- a/core/src/main/java/io/questdb/cutlass/http/HttpConnectionContext.java
+++ b/core/src/main/java/io/questdb/cutlass/http/HttpConnectionContext.java
@@ -238,6 +238,15 @@ public class HttpConnectionContext extends IOContext<HttpConnectionContext> impl
         LOG.debug().$("closed [fd=").$(fd).I$();
     }
 
+    public void drainRecvBuffer() {
+        try {
+            while (socket.recv(recvBuffer, recvBufferSize) > 0) {
+                // discard
+            }
+        } catch (Throwable ignored) {
+        }
+    }
+
     @Override
     public void fail(HttpRequestProcessorSelector selector, HttpException e) throws PeerIsSlowToReadException, ServerDisconnectException {
         LOG.info().$("failed to retry query [fd=").$(getFd()).I$();
@@ -439,15 +448,6 @@ public class HttpConnectionContext extends IOContext<HttpConnectionContext> impl
 
     public void resumeResponseSend() throws PeerIsSlowToReadException, PeerDisconnectedException {
         responseSink.resumeSend();
-    }
-
-    public void drainRecvBuffer() {
-        try {
-            while (socket.recv(recvBuffer, recvBufferSize) > 0) {
-                // discard
-            }
-        } catch (Throwable ignored) {
-        }
     }
 
     public void scheduleRetry(HttpRequestProcessor processor, RescheduleContext rescheduleContext) throws PeerIsSlowToReadException, ServerDisconnectException {
@@ -1157,8 +1157,12 @@ public class HttpConnectionContext extends IOContext<HttpConnectionContext> impl
         final HttpRequestProcessor processor = resolveResumeProcessor(selector);
         try {
             processor.resumeRecv(this);
-            // resumeRecv is not designed to complete normally (has a while-true loop). This line is unreachable.
-            return true;
+
+            // resumeRecv() returned normally -> the processor is ready to read from a websocket again.
+            // We don't return true, because that would make the infrastructure to call as immediately again
+            // and we would monopolize the thread. Instead, we use PeerIsSlowToWriteException as a signal
+            // to be registered for reading this gives the Worker thread a chance to run other processors.
+            throw PeerIsSlowToWriteException.INSTANCE;
         } catch (PeerIsSlowToReadException | PeerIsSlowToWriteException e) {
             // Need more data from/to peer
             throw e;
--- a/core/src/main/java/io/questdb/cutlass/http/HttpConnectionContext.java
+++ b/core/src/main/java/io/questdb/cutlass/http/HttpConnectionContext.java
@@ -1162,7 +1162,7 @@ public class HttpConnectionContext extends IOContext<HttpConnectionContext> impl
             // We don't return true, because that would make the infrastructure call us immediately again
             // and we would monopolize the thread. Instead, we use PeerIsSlowToWriteException as a signal
             // to be registered for reading. This gives the worker thread a chance to run other processors.
-            throw PeerIsSlowToWriteException.INSTANCE;
+            throw registerDispatcherRead();
         } catch (PeerIsSlowToReadException | PeerIsSlowToWriteException e) {
             // Need more data from/to peer
             throw e;
--- a/core/src/main/java/io/questdb/cutlass/http/HttpConnectionContext.java
+++ b/core/src/main/java/io/questdb/cutlass/http/HttpConnectionContext.java
@@ -1159,9 +1159,9 @@ public class HttpConnectionContext extends IOContext<HttpConnectionContext> impl
             processor.resumeRecv(this);
 
             // resumeRecv() returned normally -> the processor is ready to read from a websocket again.
-            // We don't return true, because that would make the infrastructure to call as immediately again
+            // We don't return true, because that would make the infrastructure call us immediately again
             // and we would monopolize the thread. Instead, we use PeerIsSlowToWriteException as a signal
-            // to be registered for reading this gives the Worker thread a chance to run other processors.
+            // to be registered for reading. This gives the worker thread a chance to run other processors.
             throw PeerIsSlowToWriteException.INSTANCE;
         } catch (PeerIsSlowToReadException | PeerIsSlowToWriteException e) {
             // Need more data from/to peer


jerrinot:
# PR #7231 Review — `fix(qwp): fix unresponsive network I/O under heavy QWP ingestion`

**Level 2.** Reviewed the current branch's PR (3 commits, one file: `core/src/main/java/io/questdb/cutlass/http/HttpConnectionContext.java`, +15/−11). Five review passes (correctness/cross-context, concurrency, tests, performance, plus inline metadata/quality), findings verified against source.

## What the change does (verified accurate)

`handleProtocolSwitchedRecv` (line 1156) now does `throw registerDispatcherRead();` after `resumeRecv()` returns normally, instead of `return true;`. `registerDispatcherRead()` returns the cached `PeerIsSlowToWriteException.INSTANCE` (`IOContext.java:101`), which `HttpServer.handleClientOperation` (`HttpServer.java:398-399`) maps to `registerChannel(context, READ)`. The old `return true` fed the in-worker loops (`handleClientOperation` do-while at 391-393, `busyRcvLoop` at 527), which re-invoked `handleClientRecv` immediately, pinning the worker on one fast client until the socket drained to EAGAIN. The new throw exits those loops after one chunk and re-arms for read so the worker rotates to other connections. The old comment (`resumeRecv ... has a while-true loop. This line is unreachable.`) was **factually wrong** — both QWP `resumeRecv` impls are linear and return normally when `0 < read < forceRecvFragmentationChunkSize` (the prod path, since the cap defaults to `Integer.MAX_VALUE`). The `drainRecvBuffer()` move is a pure alphabetical reorder (body byte-identical, method is live — called from both QWP processors) and **fixes** prior misordering (`close` < `drainRecvBuffer` < `fail`).

The "keep-alive no longer applies to upgraded connections on the receive path" claim is accurate and is an **emergent consequence of the same one-line change**, not a separate hunk: because the method now always throws, `handleClientRecv` never returns `true` for an upgraded connection, so `handleClientOperation` never reaches its `if (keepConnectionAlive()) {...} else throw disconnect` block on the recv path.

## Critical
None. The core change is correct and safe, and is consistent with the established dispatcher fairness idiom — `consumeChunked` (694), `consumeContent` (763), and the pgwire recv path all throw the same `registerDispatcherRead()` singleton to yield-and-re-arm. The no-stall guarantee holds on all three platforms: Linux re-arms via `EPOLL_CTL_MOD` (→ `ep_item_poll` re-polls readiness even under EPOLLET+ONESHOT), macOS uses `EV_ADD|EV_ONESHOT` without `EV_CLEAR` (level-triggered re-add), Windows uses level-triggered `select`. Partial frames straddling chunks survive the yield (`processWebSocketFrames` compacts residual bytes into the per-connection recv buffer), and a blocked ACK throws `PeerIsSlowToReadException` → WRITE registration, so no pending write is lost behind the READ throw.

## Moderate

**M1 — No regression test for the fairness behavior the PR exists to fix (in-repo gap).** Every QWP/WebSocket server test runs on a **single-worker** pool (`AbstractQwpWebSocketTest.java:242`, `RestartableQwpServer.java:130`, and others). With one worker, the old looping code and the new yielding code are *both* correct — they differ only in worker time-sharing. So **a revert to `return true;` would keep the entire suite green.** Multi-chunk *correctness* is well covered (randomized small fragmentation chunks in `QwpSenderE2ETest`/`QwpSenderFuzzTest`/`QwpIngressServerRestartFuzzTest` with exact row-count assertions), but the starvation/fairness property is unguarded. Add a guard: a 2-worker (or same-worker-affinity) test that streams a large multi-chunk QWP payload on one connection while a second connection (cheap REST `GET /exec?query=SELECT 1` or a second sender) must complete within a tight deadline; assert the second makes progress mid-stream and all rows land via `assertQuery(...).returns(...)` under `assertMemoryLeak`.

**M2 — Pre-existing, out-of-diff SEND-path keep-alive disconnect (not introduced here; recommend follow-up).** The recv-path fix doesn't extend to the send path. `handleClientSend` (1116-1150) for an upgraded connection does `proc.resumeSend(this); reset(); return true;`, and `resumeSend` *can* return normally. Back in `handleClientOperation`, `return true` → `if (keepConnectionAlive())` → with `http.server.keep.alive=false` the `else` branch throws `registerDispatcherDisconnect(DISCONNECT_REASON_KEEPALIVE_OFF)`, **tearing down a live WebSocket after a normal send.** Reachable only with the non-default `http.server.keep.alive=false` (`forceDisconnectOnComplete` is set only on connection-limit rejection, never on an established WebSocket). This PR doesn't touch `handleClientSend`, so it neither introduces nor worsens it — but the symmetry gap is worth a follow-up: `handleClientSend` should bypass the keep-alive disconnect for protocol-switched connections, mirroring the recv path. Not a blocker for this PR.

## Minor

- **m1 — PR description omits the tradeoff.** Per CLAUDE.md, descriptions must present negatives with equal weight. The fix adds, per readiness event, one `EPOLL_CTL_MOD` (or `kevent`/`select`) re-arm + two lock-free ring-queue hops vs. the old in-worker drain. In production (recv cap = `Integer.MAX_VALUE`) this is **per epoll wakeup, not per frame** — each `recv()` still drains the whole socket buffer — so the cost is modest, but it's real and should be stated alongside the responsiveness win.
- **m2 — Labels.** Consider adding `Performance` (and/or `REST API`) — this is an I/O latency/fairness fix, not only a functional bug.
- **m3 — Vestigial `boolean` return.** `handleProtocolSwitchedRecv` now never returns normally (every path throws); the `boolean` return type and the `return` at the call site (974) are dead. Cosmetic — could be `void`, though the current form is the pragmatic choice given `handleClientRecv` must return `boolean`. Leave or clean up.
- **m4 — Document the load-bearing invariant / note the bounded-loop option.** A one-line comment that yielding-with-buffered-data relies on the dispatcher re-arm re-polling readiness would protect the fix against future dispatcher refactors (e.g. switching Linux re-arm to DEL+ADD, or adding `EV_CLEAR` to kqueue). Separately, `IODispatcherLinux.java:245` already carries a TODO for a bounded in-worker loop count (process N chunks then yield) — a possible future refinement that would cut round-trips under a *fragmented* producer; N=1 here is near-optimal in production anyway.
- **m5 — No test for the keep-alive=false + upgraded-connection behavior** in either direction (lower priority than M1).

## Downgraded (false positives removed after source verification)

- **Stall / lost wakeup when yielding with data still in the kernel buffer** — Disproven. `EPOLL_CTL_MOD`/`EV_ADD`/`select` all re-report a socket that already has buffered RX; verified across all three dispatcher implementations and consistent with the existing `consumeChunked`/`consumeContent` pattern.
- **New throw skips resource/state cleanup** — Disproven. The old `return true` did no cleanup either; `reset()` deliberately preserves `resumeHandlerId`/`securityContext`/`isProtocolSwitched`, and `onConnectionClosed` runs only on the untouched disconnect/Throwable catch branches.
- **Pending ACK/WRITE lost behind the READ throw** — Disproven. A blocked send throws `PeerIsSlowToReadException` (→ WRITE registration); the READ throw is reached only when `resumeRecv` returned normally with nothing left to write.
- **Concurrent processing of the same fd / heartbeat race after re-arm** — Disproven. One-shot disarm + message handoff (the fd is `deleteRow`'d from `pending` while a worker owns it) prevents both; re-arm even refreshes the heartbeat/idle timestamps so an active stream isn't spuriously timed out.
- **Premise correction (not a finding):** the dispatcher control-flow exceptions are **cached singletons but not stackless** (no `fillInStackTrace` override). The throw is still cheap — the stack is captured once at class-init, never per-throw — so the conclusion (cheap, zero-alloc) stands; "cached singleton" is the accurate reason, not "stackless."

## Summary

**Approve** — the core fix is correct, safe, zero-allocation, and idiomatic for QuestDB's reactor. No critical or in-diff correctness issues. Two non-blocking follow-ups: (M1) add a fairness regression test — the bug is currently unguarded because all QWP server tests are single-worker; (M2) the analogous send-path keep-alive=false disconnect remains (pre-existing). Minors are description/label/cosmetic.

- ~11 draft findings assessed: **2 verified (1 in-repo test gap, 1 out-of-diff pre-existing)** + 5 minors; **4 false positives removed** + 1 premise correction.
- In-diff vs out-of-diff: in-diff findings are minor only (vestigial return, description, label, comment); the one substantive confirmed *bug* (M2) is **out-of-diff/pre-existing**; M1 is a missing-test gap. The diff itself is a single, well-targeted, correct change — the absence of in-diff critical findings is expected for a one-line dispatcher-yield fix, not a sign the cross-context pass underran (caller graph and all three dispatchers were walked explicitly).

mtopolnik (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 4 / 5 (80.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/http/HttpConnectionContext.java|4|5|80.00%|

PR #7232, opened 2026-06-10
feat(sql): add SHOW CREATE DATABASE for schema-only dumps

nwoolmer:
## What

Adds `SHOW CREATE DATABASE`, a schema-only logical dump (think `pg_dump --schema-only`) that emits round-trippable QuestDB DDL for every user object in the database, one statement per row.

Real `pg_dump` can't run against QuestDB — the `pg_*` catalogs are empty/absent and QuestDB DDL (`PARTITION BY`, `SYMBOL CAPACITY`, `timestamp()`, `DEDUP`, `TTL`, `IN VOLUME`, `BYPASS WAL`, storage policy) has no Postgres equivalent. So this emits QuestDB SQL instead.

## How

`ShowCreateDatabaseRecordCursorFactory` enumerates the engine's table tokens (skipping system/`sys.*` objects), classifies each as table / view / materialized view, orders them so a row-by-row replay always succeeds, and **delegates** per-object DDL to the existing `ShowCreate{Table,View,MatView}RecordCursorFactory` — no duplication of the tested DDL builders.

Emit order is the only real algorithm:
1. Tables
2. Materialized views (base tables already emitted)
3. Views (topologically sorted via the view graph so view-on-view replays)

Dependency ordering uses `engine.getViewGraph()` / `getMatViewGraph()`; matview deps are resolved by compiling the matview SQL and walking the plan so joined (non-base) tables are also ordered ahead of the matview. Sibling deps are name-sorted for deterministic output, and re-dumps are byte-idempotent (view/matview bodies are whitespace-trimmed before the `AS (...)` wrap).

### INCLUDE / EXCLUDE

`SHOW CREATE DATABASE [INCLUDE (categories) | EXCLUDE (categories) | INCLUDE ALL]`. Categories: `TABLES`, `VIEWS`, `MATERIALIZED_VIEWS` plus the `SCHEMA`/`ALL` umbrellas (enterprise adds the ACL categories). Filter-only semantics like `pg_dump -t` — cross-category deps may dangle by design. Carried on the query model (`IQueryModel.get/setShowCreateDatabaseInclude`), parsed in `SqlParser`.

## Wiring

- `IQueryModel.SHOW_CREATE_DATABASE = 18`; `SqlParser` `DATABASE` branch (no object name); `SqlParserCallback.generateShowCreateDatabaseFactory` default; `SqlOptimiser` switch case.
- Enterprise overrides the callback to wrap an ACL block (users/groups/service accounts, memberships, grants) — see the companion enterprise PR.

## Tests

`ShowCreateDatabaseTest` (13): empty db, ordering, system-object exclusion, replay-recreates-all, byte round-trip, view-on-view replay, matview multi-join dependency ordering, INCLUDE/EXCLUDE isolation per category, and parser negatives.

---

Tandem with the enterprise ACL PR: questdb/questdb-enterprise#1056 (merge this first).

---

## Hardening: concurrent DDL no longer fails the whole dump

`SHOW CREATE DATABASE` snapshots the table-token set up front, then produces each object's DDL afterward by delegating to the per-object `SHOW CREATE` factory. Previously, an object dropped, renamed or recreated in that window made the delegate throw — a table/matview "does not exist" `SqlException`, a view "could not read view definition" `SqlException`, or a `TableReferenceOutOfDateException` — which propagated out and aborted the entire dump with an error positioned at 0. On a live database under concurrent DDL the command could fail intermittently.

`appendObjectDdl` now runs a positive liveness pre-check before producing each object's DDL: it re-resolves the snapshotted name in the registry (synchronously purged on DROP for tables, views and materialized views alike) and skips the object up front, with a logged warning, if it is gone or its identity changed. The existing catch path remains as a backstop for the residual race between the pre-check and `getCursor`: a `TableReferenceOutOfDateException` skips directly, and on `SqlException`/`CairoException` it re-checks the registry to decide skip-vs-rethrow.

This pre-check also resolves an asymmetry the first cut left behind. A table or materialized view dropped mid-dump throws and was skipped, but a view dropped mid-dump did not: the view factory reads its DDL from an on-disk definition that lingers briefly after the DROP, so it succeeded and emitted stale DDL. The pre-check skips it as soon as it leaves the registry, so the dump is now consistent across object types and reflects the schema as of emit time.

Tradeoffs, stated plainly:
- This is best-effort, not a consistent snapshot. A skipped object simply does not appear in the dump, and if a skipped object was a dependency, a later object that references it may not replay cleanly. The dump no longer fails, but it is not guaranteed internally complete under concurrent DDL.
- The semantics shift toward "schema as of emit time" rather than "schema as of the initial snapshot": an object present at snapshot time but dropped before its emit is now omitted for every object type.
- The pre-check adds one registry lookup per emitted object.
- The skip path is deliberately narrow. The decision biases toward rethrow: cancellation and timeout (`CairoException.isCancellation()` / `isInterruption()`) always propagate, and a genuine error for an object that is still present also propagates, so real failures are never swallowed.

### Tests

`ShowCreateDatabaseTest` adds 10 cases driving a directly constructed dump factory whose per-object delegate is hooked at emit time:
- pre-emit liveness check skipping a table dropped before its emit, and a view dropped before its emit (the previously stale-DDL case), with the surviving objects still emitted
- backstop skips for a table dropped inside its own factory (at emit position zero) and a materialized view dropped likewise
- drop-and-recreate under the same name, where the stale snapshotted identity is skipped
- the `TableReferenceOutOfDate` skip branch
- rethrow branches: genuine `SqlException` and genuine non-cancellation `CairoException` for a still-present object, plus cancellation and timeout


nwoolmer:
## Level-3 review — findings

Reviewed the diff across correctness, cleanup, and altitude angles, verifying each finding against the source (not just static suspicion). No blocking bugs — model plumbing, `QueryModel.clear()` reset, per-object delegation, and the drop-race handling are all correct. Addressing the following:

**1. [med] Filtered/invisible dependency → silently unreplayable dump, no warning.** `topoEmit` only recurses into dependencies present in `visible`, but always emits the dependent. So e.g. `SHOW CREATE DATABASE INCLUDE (MATERIALIZED_VIEWS)` emits `CREATE MATERIALIZED VIEW mv ... FROM base` while `base` is filtered out — an unreplayable dump with no logged warning (unlike the drop-race path). Fix: log a warning when an emitted object depends on an object excluded from the dump.

**2. [low] `appendObjectDdl` reads only the first cursor row** (`if (cursor.hasNext())`). Correct today but bakes in a single-row assumption; switching to a `while` loop is free insurance.

**3. [low] No circuit-breaker between objects.** The whole dump is materialized eagerly in `cursor.of()`; the topo-walk + per-matview compile loop never checks for cancellation/timeout. Fix: check the circuit breaker per object.

**4. [low] `TableTokenCollector.val(Plannable)`/`val(ObjList)` don't collect factory tokens** the way `child()` does. If a nested cursor factory is ever reached via `val(...)` rather than `child(...)`, its base table is missed → matview ordered before its table → replay fails. Fix: unify the factory-aware traversal across `child`/`val`.

**5. [intentional, clarified] `putTrimmed`** now applies to existing `SHOW CREATE VIEW`/`MATERIALIZED VIEW` output. Verified semantically safe (the body is always a complete SELECT; trailing whitespace is never significant). Adding a clarifying comment.

**6. [doc] matview dependency collection** intentionally uses both a model walk (for inlined views) and a plan walk (for physical tables, incl. WHERE-subquery tables the model walk misses). Documenting why, so a future "simplification" doesn't regress it.

Fixes incoming on this branch.

nwoolmer:
Fixes pushed in `02a86e7` (branch also brought up to date with master):

- **#1** `topoEmit` now logs a warning when an emitted object depends on an object excluded from the dump (filtered category / system / unauthorized) — a partial, non-replayable dump is no longer silent.
- **#2** `appendObjectDdl` drains the per-object cursor instead of reading only the first row.
- **#3** circuit-breaker check per object in the topo-walk and emit loops, so a large schema dump stays cancellable even when no per-object factory throws.
- **#4** `TableTokenCollector` collects nested-factory tokens uniformly across `child()`/`val(Plannable)`/`val(ObjList)`, so a factory reached via `val()` still contributes its base table to the ordering.
- **#5/#6** clarifying comments on `putTrimmed` safety and on why both the model walk and plan walk are required for mat-view dependencies.

New tests: `testDumpHonorsCircuitBreaker` (asserts the dump consults the breaker per object via `CountingSqlExecutionCircuitBreaker`) and `testFilteredDumpEmitsDependentWithoutDependency` (pins the partial-dump contract). `ShowCreateDatabaseTest` 28/28 and `ShowCreateTableTest` 47/47 pass locally.

nwoolmer:
## Review fixes — full rundown

All findings from the level-3 review are addressed in `02a86e7` (branch is also up to date with master).

### Implemented
| # | Finding | Fix |
|---|---------|-----|
| 1 | Filtered/invisible dependency → silently unreplayable dump | `topoEmit` logs a warning when an emitted object depends on something excluded from the dump (filtered category / system / unauthorized) |
| 2 | `appendObjectDdl` read only the first cursor row | `if` → `while` to drain the cursor |
| 3 | No circuit-breaker between objects | `statefulThrowExceptionIfTripped()` per object in the topo-walk + emit loops |
| 4 | `val(Plannable)`/`val(ObjList)` didn't collect factory tokens like `child()` | Unified all three through a single null-safe `walk()` helper |
| 5/6 | `putTrimmed` behavior change; mat-view dual-walk rationale | Clarifying comments (both verified correct/intentional, no code change) |

### Deliberately NOT changed (with reasons)
- **Replace the plan-walk with `collectAllTableAndViewNames(.., false)`** — verified this is *not* a safe drop-in: the model walk only sees join/union/nested table names and misses tables referenced from WHERE sub-queries (`WHERE x IN (SELECT ... FROM t)`), which the plan walk reaches via nested cursor factories. Documented why the two-pass design is necessary instead.
- **Utf8→String→Utf8 round-trip** — left as-is; changing the `ObjList<CharSequence>` element type would break the `protected appendAclPrincipals/appendAclGrants` signatures the enterprise build overrides. Negligible cost on a cold path.
- **Refuted candidates** — shared-collector reentrancy (single-threaded, never re-entered mid-walk), rename liveness (handled by `TableToken.equals`), null base-table NPE (mat-views always carry a base) — checked against the code and excluded.

### Tests
- `testDumpHonorsCircuitBreaker` — asserts the dump consults the breaker per object via `CountingSqlExecutionCircuitBreaker`.
- `testFilteredDumpEmitsDependentWithoutDependency` — pins the partial-dump contract (dependent emitted, filtered-out dependency absent).

`ShowCreateDatabaseTest` 28/28 and `ShowCreateTableTest` 47/47 pass locally; core compiles clean.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 281 / 309 (90.94%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowCreateDatabaseRecordCursorFactory.java|211|239|88.28%|
|:large_blue_circle:|io/questdb/griffin/SqlParser.java|48|48|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlParserCallback.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowCreateTableRecordCursorFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/model/QueryModelWrapper.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/model/QueryModel.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowCreateViewRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowCreateMatViewRecordCursorFactory.java|3|3|100.00%|

bluestreak01 (approved):
Approved after a level-3 review (correctness, concurrency, performance/algorithmic optimality, resource management, tests, and cross-context caller impact), verifying every candidate finding against source.

No blocking issues. The concurrent-DDL handling (pre-emit liveness check + backstop catch that rethrows cancellation/timeout/genuine errors), the cycle-safe DFS topological sort (mark-before-recurse, name-sorted siblings, self-dep skip), and the per-object delegation to the existing SHOW CREATE factories are all correct and thoroughly tested (28 tests: ordering, replay round-trips, per-category filtering, cancellation/timeout, drop/recreate races, cursor contract).

Minor, non-blocking notes (optional follow-ups):
- TableTokenCollector overrides val(Plannable)/val(ObjList) to be factory-aware, but not the 2-arg val(Plannable, RecordCursorFactory). Not reachable today (both FilteredAsOfJoin callers also emit child(slaveFactory)), but it is a latent gap of the same class finding #4 closed. Cheap insurance: override it or pin the 'every val(_, factory) caller must also child(factory)' invariant with a comment.
- The dual model-walk + plan-walk for mat-view deps has no regression test for the one case that justifies the plan walk: a table referenced only via a WHERE sub-query (WHERE x IN (SELECT ... FROM t)). A future collapse to a single collectAllTableAndViewNames would pass all existing tests. Worth one targeted test.
- Trivial: the private static helpers are interleaved between protected-instance and private-instance methods (static/instance grouping); 'emitted' is really a visited-marker.

Several draft candidates were dropped as false positives after source verification (TableReferenceOutOfDate is converted to SqlException by generateSelectWithRetries; shared collector is single-threaded and not re-entered; getVarcharA==getVarcharB matches the existing ShowCreateTable pattern on a NoRandomAccess single-column cursor).

PR #7236, opened 2026-06-11
fix(core): prevent posting index out-of-memory during seal and rollback on skewed symbols

nwoolmer:
> This fixes an internal production out-of-memory incident on a memory-constrained
> instance ingesting a skewed symbol column. There is no public GitHub issue tracking
> it, hence no `Fixes #` reference.

## Summary

Related changes to the POSTING index seal/rollback machinery, prompted by a production out-of-memory on a memory-constrained instance ingesting a skewed symbol column. The deep reviews that followed also surfaced and closed several latent safety bugs in the existing posting code -- present at the merge-base, independent of the new streaming paths -- collected in section 5.

### 1. Fix the sealIncremental over-allocation (the reported OOM)

The WAL fast-lag commit appends one sparse generation per commit and fires an inline seal once the generation count reaches `MAX_GEN_COUNT`. The first seal compacts into a dense generation, so every subsequent seal takes the incremental branch (`sealIncremental`). Unlike the full seal, that branch had no RSS pre-flight and sized its per-stride merge/trial buffers to `DENSE_STRIDE (256) * preAllocPerKey` -- as if all 256 keys in a stride held the single hottest key's row count. On a skewed column a single hot symbol inflated the allocation by up to ~256x beyond the data the stride actually merges, exceeding the RSS limit and suspending the table.

`sealIncremental` now sizes its two directly-written buffers to the **actual** aggregate of the dirty strides (computed from per-key counts across gen 0 and the sparse gens; `mergeKeyValues` re-derives the same counts, so the bound is exact). A pre-flight defers to the full seal -- which streams per key -- when even the correctly sized buffers would breach the limit, or when a single stride exceeds 2^31 values.

### 2. Remove unreachable in-place fd-based seal/rollback code

The seal/rollback machinery carried `partitionPath.size() == 0` branches that rewrote `valueMem` in place. The fd-based open throws `UnsupportedOperationException` for posting indexes and O3 routes them through the path-based `openFromO3Context`, so those branches were unreachable -- and unsafe had they run (a throw mid-rewrite leaves a torn `.pv` with no recovery). Deleting them makes "no in-place rewrite" hold by construction; the live path-based seal/rollback is unchanged.

### 3. Stream the rollback for every cover shape and drop the whole-index decode

[...]

PR #7237, opened 2026-06-11
chore(core): remove unreachable in-place fd-based posting index seal and rollback code

nwoolmer:
## Summary

`PostingIndexWriter`'s seal and rollback machinery carried `partitionPath.size() == 0` ("fd-based") branches that rewrote `valueMem` in place — overwrite from offset 0, memmove the encoded output down, shrink the file — rather than staging a fresh `.pv` and switching to it. Investigation shows these branches are **unreachable**:

- The fd-based open `of(config, keyFd, valueFd, ...)` throws `UnsupportedOperationException` for posting indexes.
- O3 routes posting indexes through the path-based `openFromO3Context`; `O3CopyJob` takes the fd-based `of()` only for bitmap indexes.
- `partitionPath` is cleared only by `close()`, which also closes `keyMem`, so any writer that can seal or roll back has it set.

Had the in-place path ever run, a throw mid-rewrite would leave a torn `.pv` with no recovery — `reencodeMonolithic`'s restore logic (`keyCount`/`genCount`/`valueMemSize`/`maxValue` plus unlinking staged files) was gated on the path-based flag. Deleting the dead branches makes "no in-place rewrite" hold by construction.

## Changes

- Remove the fd-based in-place branches from `openSealValueFile`, `reencodeMonolithic`, `reencodeWithStrideDecoding`, and `reencodeWithPerKeyStreaming`; unwrap the path-based bodies from their now-always-true conditionals.
- Update stale comments that described the removed fd-based "mode-specific tail" as current.
- Keep the fd-based `of()` throw and the `switchToSealedValueFile` assert as the enforcement points for the invariant.

## Effects and scope

- No behaviour change on the live path-based seal/rollback: the retained code is byte-for-byte the previous path-based branch, only de-indented. The harmless defensive `partitionPath.size() == 0` guards in the purge-publish helpers and `unlinkOrphanSealFiles` are left in place.
- This is prerequisite cleanup for a follow-up that replaces `reencodeMonolithic`'s whole-index rollback buffer with per-key streaming; with the dead branch gone, that change only has the path-based case to handle.

## Test plan

- Full posting/covering suites pass unchanged: PostingIndexStressTest{,Delta,Ef}, PostingIndexOracleTest{,Delta,Ef}, PostingIndexCriticalIssuesTest, PostingIndexChainWriterTest, PostingIndexOomFallbackTest, PostingIndexO3ConcurrencyFuzzTest, PostingIndexConcurrencyTest, CoveringIndexTest, CoveringIndexMultiKeyOrderingTest, RssMemoryLimitTest, PostingIndexAdaptiveDeltaThresholdTest (~790 tests).

### DIFF
--- a/core/src/main/java/io/questdb/cairo/idx/PostingIndexWriter.java
+++ b/core/src/main/java/io/questdb/cairo/idx/PostingIndexWriter.java
@@ -1323,8 +1323,7 @@ public class PostingIndexWriter implements IndexWriter {
         final long newSealTxn = Math.max(1, chain.peekNextSealTxn());
 
         // Incremental seal: gen 0 dense covering every current key, all later
-        // gens sparse, and path-based (sealIncremental writes a fresh .pv via
-        // sealValueMem; fd-based seal must take sealFull's chunked path).
+        // gens sparse. sealIncremental writes a fresh .pv via sealValueMem.
         boolean isIncrementalCandidate = gen0KeyCount >= 0
                 && gen0KeyCount == keyCount
                 && partitionPath.size() > 0;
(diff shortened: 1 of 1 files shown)


nwoolmer:
Folding this into #7236 per review preference — the dead-code removal is the prerequisite first commit for the streaming-rollback change landing there, so keeping it all in one PR. Deleting the branch.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 25 / 25 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|25|25|100.00%|

PR #7238, opened 2026-06-11
fix(sql): fix cancel_query() spuriously cancelling the wrong query

jerrinot:
`CANCEL QUERY <id>` could cancel a different, unrelated query after the registry recycled a pooled query entry. The victim query failed with `cancelled by user` even though nobody cancelled it.

This PR fixes two user-visible effects of the same lifecycle race.

### cancel_query()

Cancellation now only applies to the active query that owns the requested id. A stale canceller holding a recycled registry entry fails the lifecycle check and reports the query as not found instead of cancelling the new owner of that entry.

Cost: one volatile lifecycle publish on query registration and one uncontended CAS when unregistering a query.

### query_activity()

`query_activity()` now snapshots registry entries optimistically and only returns a row if the entry still belongs to the same active query after the row has been copied. This prevents torn rows and wrong-query data while entries are being unregistered and recycled.

Tradeoff: a query in the `CANCELLING` state can transiently not appear in `query_activity()` while a canceller holds the lifecycle guard.

**notes for reviewers**
Each pooled Entry gets a volatile lifecycle word packing the owning query id and a state (IDLE -> ACTIVE -> (CANCELLING -> ACTIVE)* -> RETIRED -> IDLE).

cancel() may only act while it holds CANCELLING for the matching id; unregister() retires the entry after waiting out in-flight cancellers, so a stale canceller holding a recycled entry fails the CAS and reports "not found". The stress test in QueryRegistryLifecycleTest fails within a second without the guard. Cost: one volatile store per register() and one uncontended CAS per unregister().


jerrinot:
## Updated verdict: **approve** — the one pre-merge change is done.

The single thing I'd have asked for before merge (M1 — the lock-free snapshot comment asserting a false JMM invariant) is now fixed in `89819da2ea`. Everything else was already optional polish.

## Critical
None.

## Moderate
- ~~**M1** — `of()` comment claims plain-field visibility ordering the JMM doesn't provide.~~ **✅ Resolved** in `89819da2ea`. The comment now states the real mechanism (state-byte recheck anchors the window; the `(state, changedAtNs)` pair is best-effort; the lifecycle-word recheck rules out exposing a recycled entry).

## Minor (unchanged — all optional)
- **m1** — `unregister()` drops the entry + logs an "error" on the unreachable `retire()==false` branch. Prefer an `assert`/"defensive only" note. No leak.
- **m2** — `copy()` comment (`:159`) still slightly overstates the `catch` as the torn-string guard; the real guard is the `length()` recheck. *(Still open — you left this one; offered earlier.)*
- **m3** — `query_activity()` transiently hides a row mid-CANCELLING. Worth a one-line PR-description note.
- **m4** — `poolNameIsNull` / `principalIsNull` → `isPoolNameNull` / `isPrincipalNull` (naming convention).
- **m5** — "busy-wait"/"spin" comments overstate it (`Os.pause()` = `Thread.sleep(0)`, a yield).
- **m6** — Test gaps: no targeted test for the inner `beginCancel` re-acquire-fails branch; no deterministic test that `query_activity()` omits a stale row; WAL/admin throw paths only covered at higher layers. (The headline regression test is sound — verified to fail on `master`.)
- **m7** — No `Fixes #NNN`; consider a `Core` label.

## Downgraded (false positives) — unchanged
8 draft findings dropped after source verification: the `2^62` id-space comment (correct — equality-only), torn-string as a correctness bug (cosmetic, diagnostic-only), `Objects.equals` "regression" (actually a fix), pool leak-detection trip (dead counters), JVM-fault from the StringSink race (bounds-safe, grow-only), owner starvation in the retire spin (bounded set), PG cancel path "affected" (different method), and register/unregister double-call (pre-existing, balanced).

## Summary
- **Findings now: 0 Moderate open, 7 Minor open (all optional), 0 Critical.** The fix itself was already correct — no reachable crash, leak, deadlock, or data corruption across 10 agents + verification; liveness preserved; the regression test reproduces the bug on `master`.
- **Mergeable.** Of the remaining items, only **m2** is a like-for-like comment cleanup if you want full consistency with M1; the rest (m1, m3–m7) are nits, a doc note, and nice-to-have tests that don't block.

jerrinot:
a flaky test on linux-x86-graal. reproduciable. investigating.

bluestreak01 (changes requested):
**Requesting changes** — independent adversarial re-review (mission-critical pass). The fix is fundamentally sound: the `(queryId << 2) | state` lifecycle word, the two-phase guarded `cancel()`, the `retire()`-guarded recycle, and the optimistic snapshot-and-recheck in `query_activity()` correctly close the pooled-`Entry` reuse race on every normal path. No ABA (`idSeq` is monotonic), no lost updates, no deadlock/livelock (`CANCELLING` is always released in a `finally`), and the publication ordering is JMM-correct. The torn-read / use-after-free fears don't materialize (`StringSink` is heap-backed and grows monotonically; the tracker's native block is freed only at provider shutdown).

Three **Moderate** items, please. @jerrinot

---

### M1 — `register()`'s error/OOM catch recycles the Entry without `retire()`, leaving the exact wrong-query-cancellation hole the PR closes everywhere else

`QueryRegistry.java:230` adds `e.activate(queryId)` **before** `registry.put` (:231), but the `catch (Throwable th)` block (:237–:252) recycles with a bare `tlQueryPool.get().push(e)` and no `retire()`. `push()` → `Entry.clear()` (:393) does a *plain* `lifecycle = IDLE` store that doesn't wait an in-flight canceller out.

Verified path: `register(qOld)` activates then `registry.put` OOMs during resize — and in `ConcurrentLongHashMap.putVal` the node is inserted into the bin **before** `addCount(...)` triggers the allocating resize, so it's already visible to a concurrent `get()`. A canceller that grabbed `qOld` wins `beginCancel(qOld)` (CANCELLING), the catch's `clear()` clobbers it, the carrier re-pops `e` for `qNew`, the canceller's phase-1 `finally activate(qOld)` resurrects `ACTIVE(qOld)`, and its phase-2 `beginCancel(qOld)` then succeeds and flips the **unrelated qNew query's** cancelled flag — precisely the bug this PR exists to eliminate.

Reachability is very low (needs a genuine JVM `OutOfMemoryError` from the map resize — not the managed per-query limit — plus a concurrent canceller of the just-evicted id and a precise carrier re-pop), and it's **not a regression**. But it's a cheap, one-line hardening that completes the PR's own invariant — mirror `unregister()`:

```java
// in the catch, replace tlQueryPool.get().push(e);
if (e.retire(queryId)) {
    tlQueryPool.get().push(e);
} else {
    LOG.error().$("query lifecycle mismatch on register rollback [id=").$(queryId).I$();
}
```

`retire(qOld)` waits the canceller out (or CASes `ACTIVE→RETIRED` cleanly), after which the canceller's phase-2 `beginCancel` reads `RETIRED` and bails.

### M2 — `query_activity()` now materializes the full row (incl. the entire SQL text) *before* the principal filter

`QueryActivityFunctionFactory.java:102`: `hasNext()` calls `record.of(...)` — which copies `poolName`, `principal`, **and the full `query` string** plus all plain fields (`of()` :256–:281) — and only then evaluates `isAdmin || Objects.equals(record.getPrincipal(), principal)`. The old code filtered on the live principal first and copied nothing. So a non-admin polling `query_activity()` on a busy multi-tenant instance now snapshots every other user's full query text just to discard the row. It's allocation-free and never leaks the data (the row is rejected), so it's CPU-only — but it's avoidable work this PR introduced. Please snapshot+validate `principal` first for the non-admin case and copy `query`/`poolName`/the rest only on a match.

### M3 — PR description omits the second user-visible fix (`query_activity()`) and its tradeoff

The diff substantially rewrites the factory so `query_activity()` no longer returns torn / wrong-query rows under entry recycling — a distinct, user-visible fix the body doesn't mention. Per the "one section per fix" convention, please add a `query_activity()` section, and one line noting the tradeoff (a query in the `CANCELLING` state can transiently not appear in `query_activity()` while a canceller holds it). `Fixes #NNN` is correctly absent (no matching issue found).

---

### Minor (optional)
- **m4** — `copy()` comment (`:162`) attributes torn-read rejection to the `catch (IndexOutOfBoundsException)`, but the real guard is the `source.length() == len` recheck + the closing lifecycle/state recheck. For `StringSink` sources the buffer only grows, so the catch is dead (defensive only). Reword.
- **m5** — "busy-wait" comments (`:88–90`, `:119`) overstate `Os.pause()` (it's `Thread.sleep(0)`, a yield). Say "spins (yielding)".
- **m6** — `of()` reads the `memoryTracker` field twice (`:280–281`); cache one `MemoryTracker` local so both columns observe the same instance.
- **m7** — `retire()==false` drop drifts the pool's `leased` by +1 (inert today — `resetLeased()` has no callers; no native leak). Only reachable via M1; fixing M1 removes the trigger.
- **m8** — Test gaps worth a white-box test: the phase-2 `beginCancel` *failure* path (:111 → "not found"); the `authorizeSqlEngineAdmin()`-throws path asserting the entry is left `ACTIVE`; and a stale-rejection counter `> 0` in the stress test so it fails loudly if it degenerates into a no-race run. (Added tests are otherwise correct, leak-aware, and free of the removed/forbidden SQL-assertion APIs.)

### Dismissed after source verification
Torn-string segfault, `getMemoryUsed/Limit` use-after-free, spin deadlock/livelock, lifecycle ABA, new spurious permission/WAL throws, and the `poolNameIsNull`/`principalIsNull` naming (the `<noun>IsNull` form is the established codebase idiom) — all checked and cleared.

bluestreak01 (approved):
# Review of PR #7238 — `fix(sql): fix cancel_query() spuriously cancelling the wrong query`

**Reviewed at level 3** (full mission-critical pass: change-surface map, 5 parallel fresh-context agents across correctness/JMM/performance/tests/adversarial, plus per-finding source verification).

## Summary

**Verdict: approve.** This is a high-quality, correct fix for a real pooled-`Entry` reuse race. The lock-free lifecycle protocol (volatile CAS word packing `(queryId << 2) | state`) is sound; the two-phase `cancel()`, the `register()`/`unregister()` retire guards, and the `query_activity()` optimistic seqlock-style snapshot were all cleared under adversarial JMM analysis and traced to their real callers. Coverage is thorough and the concurrency tests are deterministic (latches/barriers, timed joins, liveness guards) with genuine fail-if-guard-removed efficacy.

- **Correctness & performance gate: PASS.** No confirmed correctness, resource, concurrency, or performance/IO defect on any reachable path. Every added cost (one volatile store per `register()`, one uncontended CAS per `unregister()`, per-row string copy + `loadFence` in the diagnostic `query_activity()`) lands on a per-statement or human-triggered-diagnostic path — never a per-row data path.
- **Test gate: PASS.** Not a claim-a-fix-without-a-regression-test case: the stress test `testStaleCancellerCannotTouchRecycledEntry` reproduces the bug and the per-branch tests fail if the corresponding guard is removed.
- **Findings:** 0 Critical, 1 Moderate (PR-body omission of the bundled submodule bump), several Minor (test-code polish + defensive-code notes). ~20 candidate concerns examined, **0 confirmed as defects**, all cleared as correct-by-design after source verification.
- **In-diff / out-of-diff / pre-existing split:** 0 / 0 / 0 confirmed defects. The only out-of-diff surface (`getEntry` consumed by `query_activity`, `cancel` consumed by `SqlCompilerImpl`/`cancel_query()`) integrates cleanly — `false` now correctly maps to the "query to cancel not found" user message.

## Coverage map

| # | Behavioral change | Test(s) | Failure link | Verdict |
|---|---|---|---|---|
| 1 | `cancel()` only cancels the owning active query; `false` for recycled/finished | `testStaleCancellerCannotTouchRecycledEntry` (stress), `testPhaseTwoCancelReturnsFalseWhenQueryFinishesDuringChecks`, `testCancelSetsCancelledFlagAndState`, `testCancelReturnsFalseForUnregisteredId` | stress asserts no odd-id flag set; phase-two asserts `cancel()==false` after owner unregisters mid-check | TESTED |
| 2 | Denied cross-user cancel reactivates entry before `authorizeSqlEngineAdmin()` throws | `testDeniedCrossUserCancelReactivatesEntry` | `assertActive()` fails if the `finally{activate}` is dropped (entry stuck CANCELLING) | TESTED |
| 3 | WAL cancel reactivates entry before throwing | `testWalCancelReactivatesEntryBeforeThrowing` | same reactivation guard; later `unregister` would spin/hang | TESTED |
| 4 | Canceller principal read once, outside the guard | `testSuccessfulCancelReadsCancellerPrincipalOnce`, `testSlowCancellerPrincipalDoesNotBlockUnregister` | `unregisterDone.await(5s)` times out if principal read moved inside guard | TESTED |
| 5 | `register()` activates before publish; rollback retires before pooling | `testRegisterRollbackRetiresEntryWhenListenerThrows` | asserts entry dropped, tracker released, recycled entry active for new id only | TESTED (single-threaded) |
| 6 | `unregister()` retire waits out in-flight canceller | `testStaleCancellerCannotTouchRecycledEntry`, `testSlowCancellerPrincipalDoesNotBlockUnregister` | stress detects any stale touch of a recycled entry | TESTED |
| 7 | `getLifecycle()`/`isActiveLifecycle()` bit-packing invariant | `testRecycledEntryReportsStaleLifecycle` | asserts active for new id, stale for old id after same-Entry recycle | TESTED |
| 8 | `query_activity` `of()` snapshot rejects recycled/torn rows | `testQueryActivitySnapshotUnderChurn`, `testQueryActivityRejectsRowWhenEntryRetiresDuringCopy`, `testQueryActivityRejectsShrinking/GrowingStringSnapshot`, `testQueryActivitySkipsCancellingEntry` | churn reader fails on any torn query text; reject tests isolate copy()/lifecycle recheck branches | TESTED |
| 9 | Non-admin principal filter moved into `of()` | `testNonAdminCanNotSeeOtherUsersCommands` (pre-existing) + churn tests | non-admin sees only own rows | TESTED |
| 10 | `Entry.getStateText()` removed | — | zero callers (grep-confirmed); `State.getText()` inlined | EXEMPT (dead-code removal) |

Coverage map: **10 behavioral changes, 9 tested, 0 UNTESTED, 1 exempt.**

## Moderate

**M1 — PR body omits the bundled `java-questdb-client` submodule bump (metadata).** `git diff origin/master...HEAD -- java-questdb-client` bumps the submodule `24fbbd5 -> 5cc68a9`, pulling in two QWP store-and-forward client commits (`fix(qwp): recover an SF slot whose segment tail is an unbacked mapped page (#64)` and `fix(qwp): fix a memory-access fault and worker leak during shutdown`) that are unrelated to query cancellation. Bundling is fine per the repo's consolidation policy, but the PR body describes only `cancel_query()` and `query_activity()` and never lists this client bump. *Behaviorally inert for the core fix; the concern is documentation only.* Add a one-line "Client submodule sync" section to the PR body naming what the bump contains, and confirm it is an intentional keep-current sync rather than a stray inclusion. (Store-and-forward checklist: the bumped commits are already-reviewed client *fixes* — no evidence they reintroduce a reconnect budget or producer-visible role/transport error, so no SF-contract violation; that verification properly belongs to submodule PR #64, which is why listing the bump here matters.)

## Minor

- **m1 — `copy()`'s `IndexOutOfBoundsException` catch is dead for the real sources** (`QueryActivityFunctionFactory.java`). The only genuinely mutable source is the `query` `StringSink`, whose heap `char[]` only ever grows (`checkCapacity` never shrinks), so `charAt(i)` for `i < capturedLen` can never throw; the meaningful guards are the `source.length()==len` recheck and the closing lifecycle recheck. Harmless defensive code exercised only by the adversarial-CharSequence tests. Optionally soften the "concurrent shrink can throw" comment since a `StringSink` never shrinks capacity.
- **m2 — Asymmetric owner-principal read inside the phase-1 CANCELLING guard** (`QueryRegistry.java`). The PR deliberately captures the *canceller* principal outside the guard, but `Chars.equals(entry.principal, cancellerPrincipal)` still dereferences the *owner* principal (`entry.principal`) inside the guard. A pathologically slow owner-principal `CharSequence` would extend the CANCELLING hold and make a concurrent `unregister()` busy-spin. *Not reachable with production inputs* — every `SecurityContext` in the tree returns a plain `String` principal — so this is a hardening note, not a live bug (and `testQueryActivitySkipsCancellingEntry` intentionally exploits it).
- **m3 — `retire()`/`beginCancel()` spin uses `Os.pause()` with no yield/park fallback**, and `cancel()` phase-2 runs `LOG.info(...)` inside the CANCELLING guard. Both are bounded in practice (guarded sections are minimal; `unregister` removes the map entry before spinning, so no new canceller can join) and the log tradeoff is documented. Acceptable as-is; noted for completeness.
- **m4 — Test reflection in `setPoolName()` is avoidable** (`QueryActivityFunctionFactoryTest.java`). `getDeclaredField("poolName")` + `setAccessible(true)` couples to a private field name, but `of()` reaches `copy()` identically through the public principal path, and the suite already demonstrates public injection via `CharSequencePrincipalSecurityContext`. Tests are correct; the reflection is just not unavoidable.
- **m5 — Whitebox count coupling in `testSuccessfulCancelReadsCancellerPrincipalOnce`.** Asserting `principalReads == 1` locks in an implementation detail. Defensible (guards the retire-spin-latency optimization) but brittle to a behavior-preserving refactor.
- **m6 — Minor test-helper duplication.** `PrincipalSecurityContext` is redefined in both test files; `QueryActivityFunctionFactoryTest` rolls its own `await(latch, message)` while the sibling uses `TestUtils.await`. A shared helper would reduce it.

## Downgraded (candidate concerns verified as correct-by-design)

- **"Wrong-query cancellation could still slip through"** — FALSE. `transitionFromActive` only succeeds when `lifecycle == (queryId<<2)|ACTIVE`; the id is packed into the CAS word and `idSeq` is monotonic (never reused below 2^62), so a recycled entry always carries a new id and a stale canceller's phase-2 `beginCancel` returns `false`.
- **"A still-active query's cancel could be lost (false negative)"** — FALSE. Phase-2 `beginCancel` fails only when the entry left `ACTIVE(id)`, which requires the query to have actually finished (`unregister` -> `retire`); a concurrent competing cancel serializes via the CANCELLING spin and still cancels.
- **"`IDLE=-1` sentinel could collide with a real lifecycle word"** — FALSE. `-1` has low bits `11b`=3, which `state` never takes, and every real word is non-negative.
- **"`query_activity` snapshot could expose a recycled/wrong-query row"** — FALSE. Volatile-acquire read -> plain copies -> `Unsafe.loadFence()` (necessary: a volatile *read* does not stop earlier plain loads from sinking past it) -> recheck `getLifecycle()==L0 && getState()==state`. Monotonic ids make ABA impossible, so a matching recheck provably means "same live query."
- **"Concurrent read of `entry.query` StringSink could fault / infinite-loop"** — FALSE. Heap `char[]` that only grows; `len` captured once bounds the loop; torn reads yield well-defined chars rejected by the length + lifecycle rechecks. No native memory, no OOB.
- **"Cross-thread `getMemoryUsed()/getMemoryLimit()` on a closed tracker is use-after-free"** — FALSE. `close()` returns the tracker to a pool; the native block is freed only at engine shutdown (`destroy()`/`drainPool()`), which joins workers first. Stale-but-valid number, never a fault; discarded by the lifecycle recheck anyway.
- **"`retire()` `Os.pause()` could deadlock/livelock"** — FALSE. Waiting is one-directional (retire waits for a canceller; a canceller never waits for retire), both `cancel()` guarded sections release via `finally{activate}`, and `unregister` removes the map entry before spinning, bounding the waiter set to a finite pre-remove canceller set.
- **"Torn plain-`long` display reads"** — non-issue: QuestDB targets 64-bit JVMs (atomic long access); worst case is one cosmetic `query_activity` cell.

---

*Review method: change-surface + callsite inventory built with `rg`/`git`; 5 fresh-context reviewer agents (correctness, JMM/concurrency, performance/adversarial-performance, tests/efficacy/test-quality/regression, and unguided adversarial) run in parallel; every candidate finding re-verified against current source.*

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 112 / 115 (97.39%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/QueryRegistry.java|42|44|95.45%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/activity/QueryActivityFunctionFactory.java|70|71|98.59%|

Issue #1057, opened 2021-05-24
Support for timezone conversions


postol:
**Is your feature request related to a problem? Please describe.**
Usually you store all your data with same timezone, let's say UTC. But then when you query the data you often need to work with datetimes / timestamps in a specific timezone. Currently it is not possible to convert such data easily between timezones especially because of daylight saving times.

**Describe the solution you'd like**
It would be great to support timezone conversions as other databases do where you can specify the target timezone as a string and database will convert the give datetime/timestamp to target timezone. For example postgresql has its at timezone keyword. Details how it works here: https://www.enterprisedb.com/postgres-tutorials/postgres-time-zone-explained

**Describe alternatives you've considered**
No real alternative unfortunately :(

**Additional context**

postol:
Can be implemented in any way ... the only important requirement is to convert given datetime/timestamp to target timezone. For example `source_datetime at timezone "Europe/Prague"` or `to_timezone(source_datetime, "Europe/Prague")`

bluestreak01:
Implemented in master. Will be released with 6.0.3

PR #7239, opened 2026-06-11
feat(wal): add WAL-apply suspension controls and ALTER TABLE REBASE WAL

ideoma:
Paired with the enterprise replication PR: questdb/questdb-enterprise#1057.

## What this adds

This PR adds a set of controls for managing **hard-suspended** WAL tables — tables whose WAL apply is intentionally stopped — and, on top of them, `ALTER TABLE <t> REBASE WAL` to rebuild a table under a fresh sequencer.

### Hard-suspend list

- `cairo.wal.apply.suspended.tables` — a reloadable, comma-separated list of table names that `ApplyWal2TableJob` skips, so their WAL transactions are never applied ("hard suspend"). The table stays put and accepts no further apply progress until it is removed from the list (or resumed). The config hot-swaps on reload.

### Deny writes to suspended tables

- `cairo.wal.apply.suspended.write.denied` — when true, writes to a hard-suspended table are rejected (a `CairoException`) instead of being queued into the WAL. This gives an operator a quiescent table to work on.

### Force non-structural ALTERs on a suspended table

- With write-deny on, a hard-suspended table would otherwise refuse every ALTER. This PR routes **non-structural** changes (and `FORCE DROP PARTITION`) through the WAL-bypass path so they apply directly to the table via `TableWriter` — mirroring the existing `FORCE DROP PARTITION` behaviour — keeping routine maintenance possible. **Structural** changes remain denied (they must stay versioned through the WAL).
- **Replication consequence (by design):** these force-applied changes bypass the WAL, so they are not captured by the sequencer and are therefore **not** propagated to replicas — the same property as the existing `FORCE DROP PARTITION`. A suspended table is being maintained out-of-band; re-baseline the replica with `REBASE WAL` afterwards if the change must reach it.

### Permissions

- `ALTER TABLE … SUSPEND WAL` and `RESUME WAL` both authorize through the same `SecurityContext.authorizeResumeWal` path (`authorizeSuspendWal` defaults to it), so **SUSPEND and RESUME are symmetric** — the same grant enables both and neither can be held without the other. Enterprise maps this pair to the single `RESUME WAL` permission. SUSPEND WAL no longer requires dev mode; it is now generally available under that authorization.
- `REBASE WAL` authorizes separately and more tightly: `authorizeRebaseWal` defaults to `authorizeSystemAdmin()`, **not** `authorizeResumeWal`. It is destructive (it discards pending WAL, mints a fresh `tableId`, and drops the old directory), so it is gated behind system-admin privilege rather than the table-level RESUME WAL grant. Enterprise requires the `SYSTEM ADMIN` permission for it, not `RESUME WAL`.

### ALTER TABLE REBASE WAL

`REBASE WAL` rebuilds a WAL table's storage under a fresh sequencer while keeping all of its applied data. It is a recovery tool for a table whose sequencer has drifted into an undesirable state (an extremely large or corrupt transaction log, or a table that needs re-baselining for replication).

- Preconditions: the table must be a WAL table, hard-suspended, and `cairo.wal.apply.suspended.write.denied=true`. Quiescence is achieved by precondition rather than locking — if either condition is missing the statement fails up front.
- It clones the applied table into a new directory via hard links (partition directories linked recursively, root files copied) into a `.rebase/` staging directory the startup scanner ignores, then renames it into place — so a crash mid-operation leaves no half-built live table.
- It resets `_txn` (sequencer txn 0) and `_meta` (new `tableId`, metadata version 0), seeds **two** empty transactions (so real data starts at seqTxn 3), swaps the name registry to the new directory, and drops the old one. Two seeds, not one: the replication uploader's rebase path starts at seqTxn 2, so a rebased-then-idle table needs a seqTxn 2 to settle on — otherwise the uploader busy-spins re-reading an empty txn range.
- Hard-linking preserves the data verbatim: column tops, column-version history, native and Parquet partitions, symbol columns, and every symbol index type.
- Materialized views: rebasing a **base table** invalidates dependent views (`invalid`; recover with `REFRESH MATERIALIZED VIEW <v> FULL`) rather than silently serving a stale incremental refresh; rebasing a **materialized view** itself re-registers it and lets it full-refresh from its base.

## Tradeoffs and limitations

- **Pending unapplied WAL is discarded** by a rebase. Only applied data is cloned; transactions sitting in the WAL but not yet applied are lost. This is intentional (suspension is required first) but is a real data-visibility consequence.
- **New `tableId`** after a rebase — open readers and cached factories must re-resolve; the rebased table is a different physical table reusing the logical name.
- **Materialized views fall back to a full refresh**, which can be expensive; incremental watermarks are not carried across a rebase.
- **Single volume only** — a table whose partitions span multiple volumes is rejected (with a late error) rather than rebased.
- A table left in `cairo.wal.apply.suspended.tables` makes no apply progress — that is the point, but it is easy to forget, and writes to it silently queue (or are rejected, with write-deny on).

## Test plan

- `WalWriterTest` — hard-suspended tables skip apply; write-deny rejects writes; non-structural ALTERs force-apply while structural ones stay denied.
- `WalWriterTest` — rebase keeps data, resets the sequencer, and stays writable; rebase discards a pending structural change; rebase is rejected for non-WAL tables, when not suspended, and when write-deny is off.
- `WalWriterTest#testRebaseWalComplexTablePreservesDataAndIndexes` — a table with column tops, several column-version ALTERs, mixed native and Parquet partitions, and symbol columns of every index type is rebased and verified to return the identical full dataset, remain queryable by each index (plan still shows an index scan, over Parquet partitions too), keep its Parquet partitions, and stay writable.
- `MatViewTest` — rebasing a materialized view re-registers and refreshes it; rebasing a base table invalidates the dependent view, which `REFRESH ... FULL` recovers, after which it tracks new base data again.


bluestreak01:
## Code review — @ideoma (tandem with questdb/questdb-enterprise#1057)

Reviewed #7239 + the enterprise PR together as one changeset. Verdict: **request changes**. Findings that originate in core are below; two were reproduced locally.

### Critical

**C2 — `ALTER TABLE … REBASE WAL` over the HTTP `/exec` endpoint throws NullPointerException.** `parseRebaseWal` produces `CompiledQuery.TABLE_REBASE` (=38, `isExecutedAtParseTime`), but `JsonQueryProcessor` registers an executor for every other parse-time ALTER type (`TABLE_RESUME`, `TABLE_SUSPEND`, `TABLE_SET_TYPE`, `ALTER_STORAGE_POLICY`, …) and **not** `TABLE_REBASE` (`JsonQueryProcessor.java:108-143`). Dispatch is unguarded:
```java
queryExecutors.getQuick(cc.getType()).execute(state, cc, ...);  // getQuick(38) == null → NPE
```
The completeness guard gives false confidence:
```java
assert this.queryExecutors.size() == (CompiledQuery.TYPES_COUNT + 1);
```
passes anyway, because `EMPTY` (39) extends the `ObjList` to size 40, leaving index 38 silently `null` — so even with `-ea` the hole is invisible. The rebase already ran at parse time, so the operation *succeeds* but the HTTP client gets a 500/NPE. Untested because the test helpers compile+execute in-process and never hit `/exec`. PGWire is fine (dispatches parse-time types via a `default`); `SqlValidationProcessor` falls to its `default` and mislabels REBASE as `"UNKNOWN"`.
Fix: `this.queryExecutors.extendAndSet(CompiledQuery.TABLE_REBASE, sendConfirmation);` + a regression test over HTTP. Consider making the assert detect interior nulls, not just size.

**C1 (root cause is here) — REBASE WAL's synchronous old-dir `rmdir` breaks replication.** `rebaseWalTable0` does `ff.rmdir(oldDir, true)` right after `dropTable(oldToken, false)` (`CairoEngine.java:2452-2460`), deleting `txn_seq` while the enterprise replication uploader still needs it to upload the drop. Normal `DROP TABLE` defers dir removal to the async purge job (gated on `upload.pending`); rebase bypasses that. On a replication primary the uploader then retries forever on the missing `txn_seq` and the old table's drop never reaches replicas. **Reproduced** in the enterprise repo: `ReplicationTest#testRebaseNewTableStallsOnReplica` fails deterministically (64.7s timeout) and is the root cause of the red enterprise CI. Full details on questdb/questdb-enterprise#1057. Fix: drop the old table through the coordinated path normal `DROP TABLE` uses instead of an unconditional synchronous `rmdir`.

### Moderate

**M1 — Incomplete rollback after the registry swap in `rebaseWalTable0`.** `CairoEngine.java:2380-2410`: the empty-seed commit runs *after* `rebaseSwap` (`swapped = true`). If `getWalWriter`/`commitRebaseSeed` throws (IO error), the `catch` runs `dropTable(newToken)` + `purgeToken(newToken)` + `rmdir(newDir)` — but the forward name mapping already points at `newToken` and is never restored, leaving `verifyTableName(t)` resolving to a deleted directory and the old dir orphaned. Reorder so the seed commit precedes `rebaseSwap`, or restore the old token in the rollback when `swapped`.

**M4 — REBASE WAL (and now production SUSPEND WAL) are gated only by the RESUME_WAL permission.** `SecurityContext.authorizeRebaseWal`/`authorizeSuspendWal` default to `authorizeResumeWal`. REBASE WAL is destructive (discards pending WAL, resets the table, drops the old dir), and this PR removes the dev-mode gate on SUSPEND WAL (`SqlCompilerImpl.java:5140`), making it generally available. Mapping a destructive, identity-changing recovery op to the same privilege as "resume a stalled table" is arguably under-privileged — confirm this is intended and document it. (The ACL plumbing itself is consistent across all `SecurityContext` implementations — verified.)

### Minor
- `SqlValidationProcessor` lacks a `TABLE_REBASE` case → EXPLAIN/validation reports the `"UNKNOWN"` tag.

### Verified and dismissed (false positives)
- `writeRebaseNewMarker` does create the marker (`Files.touch` = `openRW`+close).
- The startup-migration `first_txn` path is correct (`process_startup_table` honours the rebase marker — handled in the ENT Rust uploader).

The remaining ENT-side findings (Rust `cargo fmt`, replica suspend-aborts-all-downloads, per-poll JNI status fan-out, reloadability test gap, `REBASE_NEW` doc inconsistency) are on questdb/questdb-enterprise#1057.

*Tandem review at level 3. 14 findings reported across both PRs (3 Critical, 5 Moderate, ~4 Minor); 5 draft findings dropped as false positives after source verification. Three findings reproduced locally.*

ideoma:
## Code review — tandem #1057 + questdb/questdb#7239 (level 2)

Reviewed both PRs as a single changeset. Independent multi-agent sweep across correctness, concurrency, resource management, Rust panic-safety, ACL, materialized-view suspend handling, tests, and performance; every reported item verified against live source, and the two substantive findings re-scoped against the operating model (REBASE WAL is a last-resort, very-infrequent admin op; "no online recovery" acceptable; operator performs FS-level cleanup + restart).

**Verdict: ENT — approve once `cargo fmt` is run; OSS — approve.** A single CI-hygiene fix (`cargo fmt`) is the only thing blocking green. Everything else is low / minor / optional.

### Blocking (must fix before merge)

**B1 — [ENT] `cargo fmt --check` fails → "Rust Test and Lint" CI job red (in-diff).** `questdb-ent/rust/qdb-ent/src/wal/downloader/mod.rs:3153` — the `check_data_consistency(...)` test call (a site edited to thread the new `is_replication_disabled` closure) is not rustfmt-clean (rustfmt wants it expanded one-arg-per-line). Ran directly: exit 1. `clippy --all-targets -- -D warnings` is clean. **Fix: `cargo fmt` and commit.** This is the only merge blocker.

### Low / operational

**L1 — [OSS] `REBASE WAL` seed I/O failure leaves the old table directory behind (in-diff; recoverable by restart ± `rm`).** `CairoEngine.rebaseWalTable0`: after the registry swap (`dropTable(oldToken)` → `oldTableDropped=true`, line 2731; then `registerName(newToken)`), `commitRebaseSeed()` (line 2762) does real I/O and can throw on a full disk / FS error. The inner catch correctly does **not** delete the new dir (its hard-links are the only copy of the data) but re-throws **past the entire old-table teardown** (lines 2816-2846: `writeRebaseSourceMarker`, remove `_txn`/`_meta`, `tableSequencerAPI.dropTable`, `notifyPoolsTableDropped`), leaving the old dir registry-DROPPED with `_txn`/`_meta`/`txn_seq` intact, no `_rebase_source` marker, no sequencer Dropped entry, old tracker stuck `hardSuspended` (in-memory).

Fully recoverable and safe — verified end-to-end:
- **No data loss.** New dir holds the data via hard-linked partition column files (`link(2)`); the live name resolves to the new dir.
- **No corruption from manual `rm`.** `rm -rf` of the old dir only unlinks its entries; shared inodes survive via the new dir's links. QuestDB never reads the old dir after the swap, and destructive table ops are copy-on-write at the partition `nameTxn` level (never rewrite a hard-linked original in place).
- **OSS self-heals on restart.** `DefaultWalDirectoryPolicy.isInUse()==false` (no dir-level pending marker), so once `hardSuspended` clears on restart, `WalPurgeJob` + `ApplyWal2TableJob` reclaim the old dir automatically.
- **ENT is not self-healing but inert.** `UploadWalDirectoryPolicy`'s `txn_seq/.pending/upload.pending` keeps `WalPurgeJob.hasPendingTasks=true`, so the operator must `rm` the old dir + restart. Verified **no replica wipe** (old index entry stays `deleted_at=None`; the `TxnLogReader::last() -> Option` "never infer Dropped" guard holds; the replica only deletes on `deleted_at.is_some()`), **no clean-shutdown block** (the unseeded-new-dir actor Err-retries on a `drain.cancelled()`-racing path), and **no other-table starvation / index corruption**. The unseeded new table (`max_txn` 0/1) is correct and writable; the ENT uploader throttle-retries until a write reaches seqTxn 2.

Low severity: the trigger (mid-seed I/O failure on a last-resort op) is rare and recovery is the FS-ops + restart the operating model accepts. **Optional near-free hardening:** make the post-swap teardown crash-safe (run it on the seed-failure path before re-throwing) — converts ENT's "operator must `rm` + restart" into automatic recovery, and `WalWriterTest.testRebaseWalSeedFailureCommitsNewDirUnseeded` (today asserts only the *new* dir) could then also assert old-dir reclamation.

### Minor

**M1 — [OSS] A `SUSPEND WAL` landing during an in-flight mat-view refresh leaves the view in-memory-invalid (completeness inconsistency, in-diff).** `MatViewRefreshJob`. The PR guards the writer-*acquire* window (up-front `isViewWriteSuspended` gate + `isTableSuspendedError` in each **outer** catch) but not the *commit* window. `checkHardSuspended()` is enforced at `TableSequencerImpl.nextTxn`/`nextStructureTxn` (418/356), which every held-writer commit reaches via `getSequencerTxn()` → `sequencer.nextTxn()` (proven by the PR's own `testWalApplySuspendDeniesSequencerCommit`). So a suspend landing during the refresh body throws into the **inner** catch (`fullRefresh:801` and the `rangeRefresh`/`refreshIncremental`/`refreshDependentViewsIncremental` equivalents), which calls `refreshFailState` without checking `isTableSuspendedError`. `refreshFailState` runs `markAsInvalid()` (in-memory) before the `resetMatViewState` mint re-throws. Residual: the view is left in-memory-invalid while persisted `view_status` still reads `valid`, and the WAL-retention clamp (gated on `!invalid`) is skipped. **Low severity:** views are suspended to perform force-alters and the documented recovery is `RESUME WAL` + `REFRESH … FULL`, which recomputes from the base table (not base WAL) — so the skipped clamp is mooted, no data is actually lost, and the refresh worker survives. *Fix (~3 lines each):* check `isTableSuspendedError(th)` and skip before `refreshFailState` in the four inner catches.

Other minor nits:
- **[OSS]** `rebaseWalTable0` precondition checks `getTxnTracker(oldToken).isHardSuspended()` (runtime only, ~2620) while the write-block guarantee is keyed on `isWalApplySuspended` (runtime **or** config set); a config-only-suspended table is refused rebase. Conservative/safe.
- **[ENT]** `WalUploader.Listener.getReplicationStatus(long,int)` lacks the `try/finally { Path.clearThreadLocals(); }` its downloader twin / `segmentUploaded` carry — not a confirmed leak today, add for symmetry.
- **[ENT]** `process_startup_table` (`uploader/mod.rs:~4350`) stores `last_uploaded_txn = meta.last_txn` without `unpack_rebase_source(...).1`; safe today (only `.is_none()`/compare), add a defensive unpack since release builds disable overflow checks.
- **[OSS]** `commitRebaseSeed` seed-2 failure can leave `max_txn=1`; the ENT uploader handles it with a throttled Err-retry (no busy-spin). Worth a one-line post-condition note.
- **[ENT]** `EntSecurityContextBase.asValidationContext()` member-ordering matches a case-sensitive sort, not the case-insensitive convention (codebase split 2-2). Trivial.
- **[ENT]** #1057 carries only `tandem`; `WAL` / `Replication` / `Security` labels would fit.

### Verified clean (false positives dismissed after source verification)

- **Concurrency — all clear:** the `TableNameRegistryRW` metadata-cache-lock reorder (no shared mutex → no deadlock/inconsistency), the `AbstractMultiTenantPool` empty-entry sweep (gated to `Long.MAX_VALUE`, value-conditional remove), the lock-free first-node fast path in the concurrent hash maps (faithful JDK `get()` mirror), `SeqTxnTracker.hardSuspended` volatile (sequencer `checkHardSuspended` backstops under the per-table lock), config-reload of the disabled-tables set (atomic instance swap), `TABLE_DIR_NAME` (carrier-local).
- **Resource management — no leaks/double-frees:** `cloneTableDirForRebase` (`findFirst`/`findClose` paired on every throw; double `openSmallFile` re-maps cleanly), `rebaseWalTable0` `oldWriter` (single close on every path), the static `createSequencerFiles` refactor, the JNI listener GlobalRef/JMethodID lifecycle.
- **Rust panic/overflow/wipe-replica — none:** every `last_txn` reader unpacks or filters `REBASE_SOURCE_FLAG` before arithmetic; `Index.java` uses `toLong()`; `TxnLogReader::last() -> Option` handled at both sites without inferring "Dropped"; `summary.rs` `.unwrap()` sound; the `min_first_txn < first_txn` stall is correct for `rebase_new` and doesn't misfire on legitimate `first_txn>0`.
- **ACL — no bypass:** verified all 13 `SecurityContext` implementors + interface + test mocks; REBASE (SYSTEM ADMIN), symmetric SUSPEND/RESUME (shared RESUME WAL), the `DispatchingSecurityContext` default-delegation hole explicitly closed, the replica break-glass admits only non-structural alters and is unreachable by non-admins, dropping SUSPEND's dev-mode gate *added* an authorize check.
- **Test standards & logging — clean:** no forbidden `assertSql` query-helper, no `.returnsOnce` on deterministic queries, `assertMemoryLeak()` present, all new `LOG` chains closed.
- **Performance — clear:** `isWalApplySuspended` is allocation-free on the per-apply-tick path (lock-free CHM read + cached config set); the consolidated status JNI call scales with working tables, not table count; the clone loop is single-pass; the pool sweep is cold-gated.

### Withdrawn from the initial pass

- **No runtime-reload test for `replication.disabled.tables` / `pause.upload.on.suspended.tables`** — withdrawn; the codebase does not carry a reload test per reloadable knob (boot-only is the norm).
- **No PGWire REBASE/SUSPEND test; `REBASE WAL INTO` parser-rejection untested** — withdrawn; PGWire `/exec` compilation is generic/protocol-agnostic (PGWire dispatch is covered by `WriteFenceEntryPointMatrixTest`'s `TABLE_REBASE` classification), and the path-traversal validation is present in `parseRebaseWal` and re-checked server-side.

### Summary

- **ENT #1057: approve once `cargo fmt` is run** — B1 is the only merge blocker (one-command fix). The Rust/ACL/replication machinery is otherwise solid.
- **OSS #7239: approve** — no blocking correctness issue. L1 (seed-failure dir, recoverable + safe) and M1 (mat-view suspend-during-refresh, recovers via REFRESH FULL) are real but correctly low/minor under the operating model; the substantive machinery verified clean.

In-diff/out-of-diff: B1, L1, M1 are all in-diff, with L1's recovery confirmed at out-of-diff sites (`WalPurgeJob`, `ApplyWal2TableJob`, downloader/cleaner).

ideoma:
## Code review — tandem #1057 + questdb/questdb#7239 (level 2)

Reviewed both PRs as a single changeset — independent multi-agent sweep across correctness, concurrency, performance, resource management, Rust panic-safety, ACL, mat-view suspend handling, tests, and PR conventions; every reported and dismissed item verified against live source. The three prior-resolved Criticals (rmdir-race wedging the uploader, `cargo fmt` red, hard-suspend aborting all downloads) are confirmed still fixed and are not re-reported.

**Verdict: ENT #1057 — approve. OSS #7239 — approve.** No Critical or Moderate findings survived verification. The high-risk machinery (the `ConcurrentHashMap` lock-free fast path, the registry lock-order move, the read-only-fence force-bypass exemption, the Rust `REBASE_SOURCE_FLAG` packing, `TxnLogReader::last()→Option`, and the full ACL surface) is solid. Remaining items are Minor/optional cleanups, none blocking.

### Critical
None.

### Moderate
None. (Two findings raised during review were withdrawn after verification — see **Downgraded**.)

### Minor

- **[ENT] Ungated `_rebase_source` `stat()` on the hot upload path.** `questdb-ent/rust/qdb-ent/src/wal/uploader/mod.rs`, `TableEventProcessor::advance_uploads` — `rebase_source_marker_file(...).exists()` runs a `stat` syscall (plus a `PathBuf` alloc) on **every upload advance for every actively-ingesting table**, returning false for ~100% of tables. The sibling `is_rebase_new` stat directly above is correctly gated on `last_uploaded_txn.is_none()`, and the startup twin gates on the index first — only this site is ungated. Two perf passes flagged it independently. Fix: gate it (e.g. only when `table_status == Dropped`, or cache it once like the construction-time `index_rebase_source` flag). Lowest-effort real win in the set.
- **[ENT] `WalUploader.Listener.getReplicationStatus(long, int)` is missing `try/finally { Path.clearThreadLocals(); }`** that its `WalDownloader` twin (`WalDownloader.java:289`) and `segmentUploaded` carry. No active leak today (the impl allocates no `Path`), but it's a dormant per-thread native-leak risk on Rust-owned tokio threads and breaks the symmetry every sibling upcall maintains. One-line fix.
- **[ENT] `EntPropServerConfiguration.java:163`** — the new `private final CharSequenceHashSet replicationDisabledTables` is placed in the non-final mutable-field block and is out of alphabetical order (`D` sorts before the `replicationPrimary*` entries). Belongs in the `private final` block near `:105` (the sibling `pauseUploadOnSuspendedTables:101` is placed correctly).
- **[OSS] `CairoEngine.rebaseWalTable` javadoc** references a non-existent `_rebase.state` marker. The real markers are `_rebase_new` / `_rebase_source`. Comment-only.
- **[Rust] `process_startup_table` (`uploader/mod.rs`)** stores `last_uploaded_txn = meta.last_txn` without `unpack_rebase_source(...).1`. Verified safe today (only feeds `.is_none()`, Debug-format, and a comparison that correctly evaluates false for the flagged value) — but release builds disable overflow checks, so a future `.next()`/`+` would wrap silently. A defensive unpack or `debug_assert` matching `TableEventProcessor::new` would future-proof it.
- **[ENT] Optional test + labels.** No negative test exercises the `REBASE WAL INTO '<dir>'` `invalid rebase target directory` rejection (path-traversal `..`/`/`/`\`, non-numeric tableId) — the guard is correct, only the rejection is unexercised. And #1057 carries only `tandem`; `replication` + `New feature` + `security` would fit (`WAL`/`storage` don't exist in the ENT repo).

### Downgraded (raised then dismissed after source verification)

**Two Moderate findings, withdrawn:**

- **[OSS] `rebaseWalTable0` non-atomic registry swap orphaning rebased data under a concurrent same-name CREATE** — *Original claim:* the `lockedToken == null` path (`CairoEngine.java:2751`) is reachable because `createTable` takes `lockTableName` before `lockTableCreate` and `rebaseWalTable0` frees the name (`dropTable`) before its own `lockTableName`, so a racing `CREATE TABLE t` can strand the rebased dir. *Why dismissed:* the race is thin, and the outcome is recoverable with no data loss — stop, `rm` the empty dir the concurrent CREATE left, restart, and `reloadFromRootDirectory` re-discovers the rebased dir under the name. That is the same FS-ops + restart operating model REBASE (a last-resort admin op) already assumes. Low, not a blocker.
- **[OSS] `commitRebaseSeed` partial seed-failure skips the old-dir teardown** — *Original claim:* if the second seed throws, the throwable bypasses `writeRebaseSourceMarker` / `tableSequencerAPI.dropTable` / `removeQuiet(_txn,_meta)`, leaking the old dir and mis-recording it on replicas. *Why dismissed:* already covered by a prior review, and the leak framing is wrong — old-dir reclamation is driven by the **registry drop alone** (`tableNameRegistry.dropTable` → reverse-map entry dropped → `WalPurgeJob.forAllWalTables` sweeps it like any dropped table, exactly as the in-code comment states). The explicit teardown is an optimization, not load-bearing for cleanup. Requires a rare mid-seed I/O failure on a last-resort op; recoverable.

**False positives verified against source:**

- **`std.ConcurrentHashMap/IntHashMap/LongHashMap.computeIfAbsent` lock-free fast path** — concern: a core data structure with the widest blast radius changes behavior. Dismissed: faithful OpenJDK port — matches a plain first node only (`fh == h`; TreeBin/MOVED/RESERVED are negative and fall through to the lock), volatile-reads `f.val`, `keyEquals` consistent with `get()` and the locked rescan, which still re-walks the chain. No behavior change; a genuine win on the hot `getTxnTracker`/metadata-cache read path.
- **`TableNameRegistryRW.dropTable`/`renameToNew` metadata-cache-lock reorder** — concern: lock-order inversion / visibility window. Dismissed: registry holds no monitor and the cache reconcile reads the registry outside its rwLock (no inversion); re-hydration is gated by `isTableDropped` and cache ops guard on token-equality (no stale repopulation).
- **`TableNameRegistryStore.reloadFromTablesFile` `token = null` branch** — concern: the rest of the loop misbehaves when `token` is now null. Dismissed: the loop reconstructs the token from the record and doesn't depend on the removed value; a normal drop replay still releases the name.
- **`OperationDispatcher` force-bypass / `isWriteRefused`** — concern: the read-only-fence exemption is an ACL bypass on a replica. Dismissed: structural changes can never reach the bypass (`isForceableWhenSuspended() == !isStructural()`), and `isForceAlterAllowed()` is correctly false in OSS and under `cairo.read.only`.
- **ACL surface** — all `SecurityContext` impls + `DispatchingSecurityContext` explicit forwards verified; REBASE→`SYSTEM_ADMIN`, SUSPEND/RESUME symmetric via the `authorizeSuspendWal`→`authorizeResumeWal` default; replica break-glass admits only non-structural alters and is admin-gated.
- **Rust `REBASE_SOURCE_FLAG` (high bit of `last_txn`)** — every reader unpacks or `is_rebase_source()`-short-circuits before arithmetic (downloader retain, `check_data_consistency`, summary, cleaner, indexer, uploader); Java `Index.java` uses `toLong()`. No overflow, no bogus download.
- **Rust `TxnLogReader::last() → Option`** — removes a real JVM-abort `unwrap()`; both call sites handle `None` without inferring Dropped (no replica wipe).
- **`AbstractMultiTenantPool.releaseAll` empty-entry sweep** — gated to `deadline == Long.MAX_VALUE` (pool-close / `@TestOnly`), value-conditional remove; the demote `WriterPool` doesn't extend this class. No live hazard.
- **`SeqTxnTracker.hardSuspended` quiescence** — the TOCTOU between the suspend check and the writer acquire is benign (the exclusive `TableWriter` serializes rebase vs apply; a mid-flight commit throws at the sequencer); the seeds run on the new token's fresh, unsuspended tracker.
- **`getReplicationStatus` under config reload** — atomic whole-config swap; the disabled-tables set is read-only after publication. No torn read on the Rust tokio threads.
- **Mat-view suspend handling** — the up-front `isViewWriteSuspended` gate plus the inner-catch `isTableSuspendedError` skips are correct; the documented residual (a suspend landing in the commit window leaving the view in-memory-invalid) recovers via `RESUME WAL` + `REFRESH … FULL`, which recomputes from the base table — no data loss.

### Summary

- **ENT #1057: approve** — Minor cleanups only (most useful: gate the `_rebase_source` stat; the `Path.clearThreadLocals` symmetry fix).
- **OSS #7239: approve** — no blocking correctness issue; one Minor javadoc fix.
- 2 Moderate findings raised and withdrawn after verification; ~10 draft findings dismissed as false positives. All surviving findings are in-diff (the withdrawn M1's trigger reached an out-of-diff lock-ordering contract in `createTable`). The non-atomic-swap and partial-seed paths are real but recoverable under REBASE's last-resort + FS-ops/restart operating model.

bluestreak01 (dismissed):
## Approve — @ideoma · tandem #7239 + questdb/questdb-enterprise#1057 (level 3)

Independent level-3 verification as one changeset (full detail on the enterprise PR). No blocking correctness issue on the core side:
- `rebaseWalTable0` defers old-dir reclamation to `WalPurgeJob` (no synchronous `rmdir`), rollback guarded by `oldTableDropped`, resources close on all paths.
- `TABLE_REBASE` is registered in `JsonQueryProcessor` (HTTP `/exec` NPE fixed); `ConcurrentHashMap/Int/Long` fast path is a faithful OpenJDK `computeIfAbsent` re-port; `OperationDispatcher` force-bypass denies structural changes and exempts the read-only fence only for authorized non-structural force-alters; SUSPEND/RESUME symmetric, REBASE→`SYSTEM ADMIN`.
- `TableNameRegistryStore` conditional `OPERATION_REMOVE` replay and `WalWriter.commitRebaseSeed` (two empty seeds) verified sound.

Non-blocking nits (carried on #1057): `SqlValidationProcessor.java:348` maps `TABLE_REBASE` to the `"UNKNOWN"` tag (cosmetic; the C2 fix updated `JsonQueryProcessor` but not this sibling); unrelated `SqlKeywords.java` reordering churn beyond the new `isRebaseKeyword`; `WalWriter.commitRebaseSeed` 2nd-seed failure can leave `max_txn=1` (rare mid-seed I/O failure) — worth a post-condition note.

The one Moderate from the tandem review (config-list vs tracker `getReplicationStatus` gap) lives entirely on the enterprise side, so it does not block this PR. Approving.

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 372 / 459 (81.05%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|16|37|43.24%|
|:large_blue_circle:|io/questdb/cairo/SecurityContext.java|2|4|50.00%|
|:large_blue_circle:|io/questdb/cairo/wal/CheckWalTransactionsJob.java|1|2|50.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|21|36|58.33%|
|:large_blue_circle:|io/questdb/cairo/wal/WalWriter.java|5|8|62.50%|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|104|138|75.36%|
|:large_blue_circle:|io/questdb/cairo/wal/WalUtils.java|86|97|88.66%|
|:large_blue_circle:|io/questdb/cairo/wal/seq/TableSequencerImpl.java|23|23|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/JsonQueryProcessor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/PropertyKey.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/seq/TableSequencerAPI.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/pool/AbstractMultiTenantPool.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoConfigurationWrapper.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/std/ConcurrentLongHashMap.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlKeywords.java|33|33|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableNameRegistryStore.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/AbstractOperation.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/PropServerConfiguration.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/OperationDispatcher.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/std/ConcurrentIntHashMap.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/seq/SeqTxnTracker.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/std/ConcurrentHashMap.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableNameRegistryRW.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/CompiledQueryImpl.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/ApplyWal2TableJob.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoConfiguration.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/AlterOperation.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoException.java|5|5|100.00%|

PR #7240, opened 2026-06-11
fix(ilp): fix data corruption in concurrent ILP over TCP

jerrinot:
Concurrent ILP clients sending to WAL tables over TCP could silently corrupt each other's data: values from one connection leaked into rows written by another. The corruption affects DECIMAL columns ingested from numeric or string values, LONG256 columns, and binary-format values cast to SYMBOL columns. The server reports no error, so the problem only shows up in query results. Deployments with a single network IO worker or a single connection are unaffected, as are ingestion over HTTP and QWP over WebSocket.

**details for reviewers**
LineTcpMeasurementScheduler shared one LineWalAppender across all network IO workers, so concurrent appendToWal calls raced on its Decimal256, Long256Impl and DirectUtf8Sink scratch buffers. The scheduler now keeps one appender per network IO worker, indexed by worker id like the existing tableNameSinks; LineWalAppender owns its scratch sink and implements QuietCloseable, and LineHttpProcessorState frees the appender it already owned per connection. This restores the invariant the HTTP path and QwpWalAppender already follow: one appender per thread of execution. The cost is a handful of scratch objects + extra instances of LineWalAppender per worker.


jerrinot:
## Review: PR #7240 — `fix(ilp): fix data corruption in concurrent ILP over TCP` (level 3)

### Verdict: **Approve** — clean, correct, root-caused fix. No blocking issues found.

The fix is exactly right: `LineWalAppender` held mutable scratch state (`Decimal256`, `Long256Impl`, `DirectUtf8Sink`) but the scheduler shared one instance across all network IO worker threads, so concurrent `appendToWal` calls corrupted each other's conversions. Moving to one appender per IO worker — indexed by worker id, mirroring the existing `tableNameSinks[]` pattern — restores thread-confinement, which is the correct invariant.

### Verification performed (all CONFIRMED safe)

**Worker-id indexing is bounded and proven (in-diff, `LineTcpMeasurementScheduler:336`).** `walAppenders` is sized `networkSharedPoolSize` and indexed by `netIoJob.getWorkerId()`. Worker ids are assigned `0..networkSharedPoolSize-1` at `createNetworkIOJob(dispatcher, i)`, each `NetworkIOJob` is bound to a single thread (`assert this.workerId == workerId` in `run()`), and `tableNameSinks[netIoJob.getWorkerId()]` already uses this identical index on the same path. No out-of-bounds possible. ✓

**Thread-confinement is complete.** I swept the rest of the concurrent `scheduleEvent`→`appendToWal` path for other shared mutable scratch: `tableNameSinks` is per-worker, the shared `path`/`tableUpdateDetailsUtf16` are touched only under `tableUpdateDetailsLock.writeLock()`, and the appender no longer references any scheduler-level mutable state. No remaining race. ✓

**Resource cleanup on all paths.** `walAppenders` is freed via `Misc.freeObjList` (null-safe; verified). The constructor's `catch (Throwable t) { close(); throw t; }` handles a partially-populated list correctly. `QuietCloseable extends Closeable`, so `ObjList<LineWalAppender>` works with `freeObjList`. Both `new LineWalAppender(...)` callsites use the new 5-arg signature — no stale callers. ✓

**HTTP path has no use-after-free (out-of-diff lifecycle check).** `LineHttpProcessorState` is stored in a `LocalValue` keyed by connection context; per-request reuse calls `clear()`, and `close()` (which now frees the appender) runs only at final context disposal. The appender owns and frees its own `DirectUtf8Sink`; the state no longer creates one, so no double-free and no leak. The old `utf8Sink` field has zero remaining references in the file. ✓

**Regression test is valid.** `LineTcpWalDecimalConcurrencyTest` overrides `getWorkerCount()→16`, and the receiver is created with that pool as the network IO pool, so `networkSharedPoolSize=16` — multiple IO workers, which is required to reproduce the cross-worker race. 16 senders × 25k rows writing distinct per-variant DECIMAL values to a shared column will detect value bleed via the variant WHERE-clause check. Uses `.returns(...)` (not `returnsOnce`), and `runInContext` wraps the multi-statement body in `assertMemoryLeak`, so the `.noLeakCheck()` on the inner `assertQuery` is the legitimate shared-scope case, not the anti-pattern. ✓

### Minor (non-blocking)

1. **`QwpSenderE2ETest.java` diff is inflated by alphabetical method reordering.** Roughly half the diff is `testDecimal`/`testDecimalRescale`/`testIntColumnIntoIPv4TranslatesNullSentinel`/`testDeferredCommit*` being deleted and re-inserted at new positions to keep methods sorted. It's test-only and compliant with the alphabetical-member convention, but it makes the genuinely new content (`testConcurrentSenders_sameTable_doubleToDecimal`) hard to spot. The `.returnsOnce(...)` calls in those moved methods are pre-existing and merely relocated, so not findings — but worth knowing the only *new* QWP test is the one concurrency test.

2. **The TCP regression test is inherently probabilistic** (race detection). That's standard and acceptable, but confirm you've observed it actually fail on `master` without the fix — otherwise it could pass for the wrong reason and silently rot. Worth a one-line note in the PR test plan.

3. **PR metadata is compliant.** Conventional Commits title, verb repeated, end-user impact framing, level-headed description that correctly scopes affected vs. unaffected deployments. Labels `Bug`/`ILP` fit. No issue reference, which is fine if none exists. Optionally consider adding `WAL` to labels since the corruption is WAL-table-specific.

### Summary
- **Verdict: approve.** Correct root cause, minimal blast radius, follows the established per-worker pattern, good regression coverage.
- No regressions or tradeoffs beyond a handful of extra scratch objects per IO worker, which the PR description already discloses honestly.
- **Draft findings: 9 examined, 0 confirmed as bugs, 9 cleared after source verification** (worker-id bounds, freeObjList null-safety, HTTP use-after-free, double-free of utf8Sink, stale constructor callers, QuietCloseable typing, remaining shared scratch on concurrent path, test worker-count validity, leak-check wrapping).
- **In-diff vs out-of-diff:** the highest-risk checks were out-of-diff (HTTP `LocalValue` lifecycle, caller worker-id range, `freeObjList` null contract) and all came back safe — the diff did not quietly break any callsite.

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/LineWalAppender.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/LineTcpMeasurementScheduler.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/LineHttpProcessorState.java|1|1|100.00%|

PR #7222, opened 2026-06-09
perf(sql): speed up ORDER BY ... LIMIT N queries

kafka1991:
## Summary

The change targets a query slowdown: `SELECT * ... WHERE <interval + symbol filter> ORDER BY timestamp, rpt_seq LIMIT N` over a wide Parquet partition (an order-book table whose bid/ask books are `DOUBLE[][]` array columns) runs in milliseconds at LIMIT 1000 but degrades to seconds at LIMIT 6000, and intermittently at LIMIT 10000. A local reproduction (50M-row sort target, same array shape, single Parquet partition) shows the same access pattern but not the customer's degradation multiples or the intermittency; those point at environment factors - disk read performance under mmap and concurrent load - that multiply the cost of every Parquet row-group decode, which on this schema is dominated by the two wide array columns. The working conclusion: the effective lever is reducing how much data Parquet decode touches. The PR does that on three axes.

- The top-N build does zero random-access decodes (the forward scan still decodes filter and sort-key columns). The encoded-sort cursor replaces LimitedSizeLongTreeChain for fixed-width sort keys: it collects (encoded key, rowId) pairs in a flat native buffer during the forward scan and sorts them natively. The tree chain did one recordAt() per comparison, O(K log K) random-access decodes per query, each one a full row-group decode on a cache miss. The buffer stays O(limit): once it reaches max(2 x limit, 4096) entries (clamped to the sort memory cap) the cursor sorts, truncates to the top limit, and rejects subsequent rows with a single native compare against the boundary entry.
- The build phase narrows decoded columns on async-filtered scans. The sort cursor declares its sort-key columns to the base cursor (setParentUsedColumns); the async filter's worker-side late materialization (AsyncFilterAtom) is the only consumer of that declaration today, so for the order-book schema's filtered query, build-phase decode drops from the full projection (including both array columns) to the filter symbol plus the two key columns. Plain page-frame and Java-filtered scans ignore the declaration and still decode the full projection during build; wiring those paths up is a follow-up.
- The emit phase decodes only the rows it returns. After sorting, the cursor declares the exact emit row set to the base cursor (setRecordAtRows); Parquet frames decoded during emit materialize just those rows via the fill-with-nulls row filter instead of whole row groups. A density gate falls back to full decode when the declared rows cover half or more of a row group, where row filtering would not pay. The declaration is pull-based: cursors that cannot benefit (native partitions, non-page-frame bases) never pull, so they pay nothing.

[...]

PR #7241, opened 2026-06-12
perf(sql): speed up parallel ORDER BY ... LIMIT N queries with encoded sort keys

kafka1991:
Depends on #7222 — this branch builds on `encode_sort_on_light_topk`, so the diff includes that PR's commits until it merges. The parallel top-K work is the commits on top of that branch.

## Summary

The parallel top-K factory (`Async Top K` plans: `ORDER BY ... LIMIT N` over page-frame tables) previously collected candidates into a per-worker `LimitedSizeLongTreeChain` driven by a compiled `RecordComparator`: every scanned row paid a tree insertion whose comparisons re-position the record across frames. This PR routes encodable sort keys (fixed-width, byte-comparable keys up to 32 bytes — the same set #7222 encodes on the serial path) through flat per-worker `EncodedTopKBuffer`s of `(encoded key, rowId)` entries instead:

- Each worker appends encoded entries and keeps its buffer at O(limit) through sort-and-truncate compaction; once a key threshold is known, non-qualifying rows drop with a native key compare.
- The owner merges the worker buffers through the same threshold filter, runs one final native sort (parallel above `cairo.sql.sort.encoded.parallel.threshold`), and emits rows by rowId.
- Full-frame scans decode only the sort-key columns; filtered scans late-materialize only the sort-key columns the filter pass did not decode.

On top of the basic path switch:

- `SortKeyEncoder.encodeFixed8Frame` batch-encodes a whole frame when the key is a single fixed-width-8 column (the common `ORDER BY col LIMIT N` shape): the column base address, type dispatch and direction transform hoist out of the per-row loop, and `EncodedTopKBuffer.fastRejectsKey` drops a row with one unsigned compare before writing the entry once compaction publishes a threshold. Multi-column keys and column-top frames fall back to the per-row encoder.
- Per-worker encoders share the owner's symbol rank maps. Previously every one of the workerCount+1 encoders sorted the whole symbol dictionary in `init()` and held its own native copy of identical maps.
- `AsyncTopKRecordCursor.copyParquetRowIdsTo` counts parquet rowIds before allocating, matching the serial cursor.

Non-encodable keys (STRING/VARCHAR, non-static SYMBOL, keys over 32 bytes) keep the existing tree-chain path unchanged.

## Measured results

The reproduction of the customer scenario that motivated #7222: a 200M-row order-book-shaped table in a single DAY partition (identifiers anonymized):

```sql
CREATE TABLE books (
    sym       SYMBOL,
    category  SYMBOL,
    seq       LONG,
    sub_seq   LONG,
    action    LONG,
    src       VARCHAR,
    ingest_ts TIMESTAMP_NS,
    bids      DOUBLE[][],
    asks      DOUBLE[][],
    timestamp TIMESTAMP
) TIMESTAMP(timestamp) PARTITION BY DAY;

INSERT INTO books
SELECT
    rnd_symbol('s0','s1','s2','s3','s4','s5','s6','s7','s8','s9') sym,
    'C1'                                                          category,
    x                                                             seq,
    x                                                             sub_seq,
    rnd_long(0, 2, 0)                                             action,
    'incremental'                                                 src,
    (1000L * to_timestamp('2026-05-28T13:00:00','yyyy-MM-ddTHH:mm:ss')::long
        + (x - 1) * 3600000L)::timestamp_ns                      ingest_ts,
    rnd_double_array(2, 0, 0, 2, 10)                              bids,
    rnd_double_array(2, 0, 0, 2, 10)                              asks,
    (to_timestamp('2026-05-28T13:00:00','yyyy-MM-ddTHH:mm:ss')::long
        + (x - 1) * 360L)::timestamp                             timestamp
FROM long_sequence(200_000_000) x;

-- parquet variant only:
ALTER TABLE books CONVERT PARTITION TO PARQUET WHERE timestamp > 0;
```

```sql
SELECT *
FROM books
WHERE timestamp IN '2026-05-28'
  AND timestamp BETWEEN '2026-05-28T13:55' AND '2026-05-28T14:05'
  AND sym = 's0'
ORDER BY category, seq
LIMIT 100_000;
```

The sort key (`SYMBOL`, `LONG`) is a two-column 16-byte encoded key, so the query takes the per-row encoder, not the single-column batch path.

| Partition format | master                                | this PR          |
|------------------|---------------------------------------|------------------|
| Parquet          | 380 ms to 200 s (unstable run to run) | 19.76 ms (stable) |
| Native           | 29 ms                                 | 9.72 ms          |

The master jitter profiles to `AsyncTopKRecordCursor.mergeChains` → `LimitedSizeLongTreeChain.put` → `recordAt` → `PageFrameMemoryPool.navigateTo` → `ParquetPartitionDecoder.decodeRowGroup`: the single-threaded tree merge re-positions records through the comparator, and on parquet each jump can decode a row group, so the cost tracks decode-cache behavior rather than row count. The encoded path never calls `recordAt` during build or merge — it compares native keys and touches each emitted row once, by rowId.

This is one query shape on one machine; the native-partition gain is correspondingly smaller since native frames make `recordAt` cheap.

## Tradeoffs

- Peak native memory on the encoded parallel path is up to workerCount times the serial encoded cursor: each worker buffer independently holds up to ~max(2*limit, 4096) entries and carries the full sort budget. The `cairo.sql.sort.key.max.bytes` / `cairo.sql.sort.light.value.max.bytes` caps still bind on each buffer.
- When the budget caps fire on the parallel path over encodable keys, the error now names EncodedSort and both config keys instead of the RedBlackTree message.
- Worker compaction sorts no longer check the circuit breaker mid-sort; breaker checks stay at frame granularity, so a tripped query exits at the next frame boundary.

## Test plan

- `ParallelTopKFuzzTest#testParallelTopKEncodedTypes` (new): checks the parallel encoded path against the serial tree-chain reference for all 20 fixed-width-8 key types plus DECIMAL128/256, ASC and DESC, filtered scans, full-row emission on unique keys, and parquet frames; verified across repeated random parameter combinations (parallel/JIT/parquet).
- `OrderByMemoryCapTest`: the per-worker budget cap fires with the EncodedSort hint on encodable keys; a new VARCHAR-key test keeps the tree-chain fallback's RedBlackTree hint covered.
- Existing suites pass: `OrderByEncodeSortTest`, `OrderByEncodeSortFuzzTest`, `LimitTest`, `TopKFilterCompilationLeakTest`.


kafka1991:
# Code Review: parallel ORDER BY ... LIMIT N with encoded sort keys

**Review level 3** (full mission-critical pass: all 11 review agents, per-finding source verification). Rust safety (Agent 8) was not applicable — this PR's own delta (8 files vs base branch `encode_sort_on_light_topk`) contains no `.rs` files; the Rust/`.dylib` churn in the GitHub diff belongs to the #7222 dependency, below the merge-base.

## Verdict: Approve with minor changes. No blocking findings.

This is a well-built PR. The encoded batch encoders are bit-identical to the per-row encoders, the threshold/merge selection is mathematically sound, the shared-rank-map concurrency is race-free, native memory lifecycle is leak-free on all traced paths (constructor error paths, reopen, double-clear), and all four out-of-diff `SortKeyEncoder` callsites remain correct. The fresh-context adversarial pass found no wrong-results bug. The strongest signal is in **test gating** and **merge-path performance** — none of it blocking.

---

## Moderate

### 1. The centerpiece test skips itself on ~half of CI runs

`ParallelTopKFuzzTest.testParallelTopKEncodedTypes` opens with `Assume.assumeTrue(enableParallelTopK)`, and `data()` picks **one random `{parallel, JIT, parquet}` triple per CI run**. So whenever the draw yields `enableParallelTopK == false` (≈50%+ of runs), the entire new encoded-path test — the only differential oracle for this PR's new code — does not execute.

The fix is clean and specific to this test: the test's own `assertTopKMatch` already calls `ctx.setParallelTopKEnabled(true)` for the actual run and `false` for the reference run (`setParallelTopKEnabled` is a direct context-level override of the config property). The test therefore **functions regardless of the `enableParallelTopK` parameter** — the `Assume` gate is unnecessary and only suppresses coverage. Drop the `Assume.assumeTrue(enableParallelTopK)` (JIT and parquet still vary per run via the param), so the new path runs on every invocation.

While there, two smaller coverage gaps in the same test:

- **ASC/DESC is coin-flipped per type per run** (`rnd.nextBoolean()`), and each direction uses a different `xorMask`/transform in `encodeFixed8Batch`. A direction-specific encoding bug in one type could pass intermittently. Iterate both directions deterministically (`for (boolean desc : {false, true})`) — cheap, removes the blind spot.
- **No empty-result and no LIMIT-larger-than-rowcount case on the encoded path.** These exercise distinct new branches: `mergeChains` with `emitStartAddr == emitEndAddr` (the `if (emitStartAddr < emitEndAddr)` false branch, `setRecordAtRows` skipped) and `emitCount = Math.min(count, lo)` clamping down to `count`. Add one of each.

Coverage that *is* present and was verified (by instrumenting the source): all 20 fixed-8 types via the batch encoder, decimal128/256 generic, multi-column generic, filtered scans, parquet frames, SYMBOL-with-NULL, full-row emission on unique keys, compaction firing at 4096, and the column-top-as-sort-key fallback via `testParallelTopKIntrinsicsFilter`'s `ORDER BY colTop`.

---

## Minor

### 2. `mergeFrom` skips the cheap fast-reject before copying each entry (perf)

`EncodedTopKBuffer.mergeFrom` routes every worker entry through `beginAppend()` → `Vect.memcpy(entrySize)` → `endAppend()`, paying an unconditional `memcpy` and the full multi-word `isBeyondThreshold` compare on every entry, including ones a single leading-word `compareUnsigned` would reject. On the single-threaded owner merge (up to `workerCount × compactionTrigger` entries before its own compactions), reading the source's leading key word and calling `fastRejectsKey` first — skipping both the `memcpy` and the `beginAppend` bookkeeping on rejection — is a free constant-factor win with no correctness or memory-bound impact.

Note: do **not** follow the tempting "bulk-copy all workers then sort once" path — the intermediate compaction in `mergeFrom` is what bounds owner memory to O(limit); removing it would let the owner exceed `maxEntries`. The fast-reject is the safe optimization. A bounded k-way merge of the (mostly-sorted) worker runs is a larger, legitimate future optimization but out of scope here.

### 3. The async final merge-sort does not poll the circuit breaker

`AsyncTopKRecordCursor.mergeChains` runs `ownerTopK.sort()` (potentially a large sort for big `lo`) with no circuit-breaker check before or after, and `EncodedTopKBuffer.sort()`/`compact()` hold no breaker reference. A query cancelled after the last frame runs the full merge sort before noticing. This is partially acknowledged in the PR's "Tradeoffs" (worker compactions) and is **not a regression** — the old tree-chain merge also did no per-entry breaker check. But the serial path *does* poll after its final sort (`EncodedSortLimitedLightRecordCursor` line 237); mirroring that with one `statefulThrowExceptionIfTrippedNoThrottle()` around the merge sort would close the responsiveness gap cheaply. (The serial refactor itself is fine — it still polls per-row and post-sort.)

### 4. Dropped defensive try/catch in the refactored serial cursor constructor

`EncodedSortLimitedLightRecordCursor`'s constructor now does `new SortKeyEncoder(...)` then `new EncodedTopKBuffer(...)` with no try/catch, where the old inline code freed both on partial failure. This is **safe today only because both allocate natively-deferred** (`keepClosed=true`, no malloc in the constructor) — a throw between the two `new`s leaks nothing. It silently becomes a leak if either type ever switches to eager allocation. Either restore the guard or add a one-line comment recording the deferred-allocation assumption.

### 5. BOOLEAN batch path assumes raw `{0,1}` byte storage

`SortKeyEncoder.batchIntegral1` for BOOLEAN reads the raw stored byte and xors, whereas the per-row `encodeFixed8` normalizes via `record.getBool()` (`== 1`). They diverge only if a BOOLEAN column stores a byte other than 0/1, which storage never produces (`putBool` writes exactly `(byte)(v?1:0)`), so this is unreachable. Worth a one-line comment noting the {0,1}-only assumption.

### 6. PR is under-labeled

Has `SQL` + `Performance`; the change touches parallel page-frame execution and parquet frames. Consider adding `Core`. Also: because the diff currently carries #7222's commits, ensure #7222 squash-merges first so this PR's squashed message doesn't absorb #7222's work.

The PR title, description tone (honest "one query shape on one machine" caveat, both partition formats shown, a fair Tradeoffs section), `Depends on #7222` dependency expression, and plain-bullet test plan are all compliant.

---

- **"Encoded keys differ from the tree-chain reference on duplicate sort keys."** Intentional and legal top-K tie-break-by-rowId; entries are unique by rowId so no distinct row is ever wrongly dropped; the test projects only the sort column for duplicate-key shapes to keep the oracle deterministic.
- **"Serial cursor lost its per-compaction circuit-breaker check."** False — `runBuild` polls per-row and `buildAndSort` polls after the final sort; the intermediate check effectively moved to per-row, equivalent.
- **"`init(RecordCursor)` → `init(SymbolTableSource)` breaks the 3 serial callers."** False — `RecordCursor extends SymbolTableSource`; widening, all callsites compile and behave identically, and they construct 2-arg encoders so the new `assert !hasBorrowedRankMaps` always holds.
- **"Shared rank maps double-free / use-after-free / data race."** False — owner owns the maps and frees once; borrowing encoders' `close()` is a no-op for maps; workers read them read-only after the dispatch barrier; `clear()` is idempotent.
- **"Constructor partial-failure leaks native memory."** False — all native buffers use deferred allocation; final fields are null-or-assigned before any throw; `close()→clear()` routes through null-safe `Misc.free*`.
- **"`mergeFrom` could overflow `maxEntries` during merge."** False — `compactionTrigger ≤ maxEntries` forces compaction back to `limit` before the cap; for `limit ≥ maxEntries` every worker already threw during its own build.
- **"Bulk-append + single sort would speed up the merge."** Rejected — would break the O(limit) memory bound; the intermediate compaction is load-bearing.

---

## Summary

- **Verdict:** Approve; address the test gating (#1) before merge, the rest are optional.
- **Tradeoffs the PR is honest about:** peak native memory scales with `workerCount`; circuit-breaker granularity coarsens to frame boundaries; the encoded-budget error message changes wording. All disclosed in the PR body.
- **Findings:** ~30 draft observations across 10 agents → 6 reported (1 Moderate, 5 Minor), 7 false positives dropped after source verification. **All 6 reported findings are in-diff; 0 out-of-diff** — the cross-context pass ran the full callsite inventory for the widened `SortKeyEncoder` API and the refactored `EncodedTopKBuffer` and confirmed every external callsite safe, so the zero out-of-diff count reflects genuine backward compatibility, not an underrun.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 354 / 373 (94.91%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyEncoder.java|98|113|86.73%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortLimitedLightRecordCursor.java|18|19|94.74%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedTopKBuffer.java|70|73|95.89%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKRecordCursorFactory.java|39|39|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKAtom.java|81|81|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKRecordCursor.java|48|48|100.00%|

bluestreak01 (approved):
Approved after a level-3 mission-critical review.

No correctness, concurrency, resource-leak, or out-of-diff (P0) bugs found. Verified from source: the parallel top-K gate only fires for FIRST N with no offset (so isFirstN=true and emitCount=min(count,lo) are correct), the batch encoders produce byte-identical keys to the per-row encodeFixed8 for every fixed-8 type ASC+DESC incl NULL/NaN/null-symbol, isSupported correctly rejects non-static SYMBOL, late-mat parquet buffers are full-frame indexed with NULL fill (consistent with the per-row path), the buffer close/reopen/clear and double-free lifecycle is sound, and shared rank maps are read-only and fully built before dispatch. The distributed merge is provably correct (per-worker top-N union contains the global top-N; total order by (key,rowId)).

Non-blocking follow-ups:
- perf: the multi-column / generic per-row encode pays the full key write per scanned row with no leading-word fast-reject, even though the threshold is valid for multi-column keys — largest available win (also present in the serial path).
- test: the equivalence fuzz test asserts only 'Async'+'Top K' in the plan, which is identical for the encoded and tree-chain async modes, so it could pass vacuously if isEncoded regressed (partly mitigated by the EncodedSort cap test). Consider an encoded plan marker.
- stray artifacts to revert: questdbr.dll file-mode flip 100755->100644 (content identical) and the THIRD_PARTY_LICENSES.txt date-only bump.
- remove the now-stale 'Depends on #7222' note (merged).

PR #7242, opened 2026-06-12
perf(sql): speed up ORDER BY on varchar, UUIDs and wide multi-column keys

kafka1991:
## What

The encoded sort - full sort (`EncodedSortRecordCursorFactory` / `EncodedSortLightRecordCursorFactory`), serial top-K (`EncodedSortLimitedLightRecordCursorFactory`) and parallel top-K (`AsyncTopKRecordCursorFactory`), all radix-sorting byte-comparable keys - previously engaged only when all ORDER BY columns were fixed-width and the combined key fit 32 bytes. Everything else - VARCHAR, STRING, non-static SYMBOL, UUID, LONG256, LONG128, or fixed-width keys past 32 bytes - fell back to the red-black tree sort (`LongTreeChain` + compiled comparator), which performs O(log n) comparator calls per inserted row and, on the light path, a `recordAt` random access per comparison.

This PR extends the encoded sort to every column type that ORDER BY accepts and to keys of any width, across both the full-sort and the (serial and parallel) top-K paths. The tree path remains available behind the `cairo.sql.orderby.sort.enabled` kill switch.

## How

- `SortKeyEncoder` normalizes each sort column into a byte-comparable segment:
  - VARCHAR: UTF-8 bytes shifted by +1 with a 0x00 terminator; a 0xFF byte (which only non-UTF-8 input can produce) would wrap to the terminator, so the encoder detects it per value and rejects the sort; matches `Utf8s.compare`.
  - STRING and non-static SYMBOL: UTF-16BE code-unit bytes with 0x00/0x01 escaping; matches `Chars.compare`, including its divergence from VARCHAR order on supplementary-plane characters.
  - UUID and LONG256: unsigned big-endian with an order-preserving +1 remap that moves the all-`Long.MIN_VALUE` null pattern to zero, so null sorts first without a marker byte; single-column UUID/LONG256 keys stay on the inline fixed path (16/32 bytes).
  - LONG128: signed hi + unsigned lo, mirroring `Long128.compare`.
  - Static SYMBOL keeps the existing rank-map encoding inside variable keys.
- Keys that stay fixed-width and fit 32 bytes use the existing FIXED_8..32 entries unchanged. Wider or variable keys use a new VARIABLE entry: a 16-byte inline prefix, the key length, an offset into a key heap holding the full normalized key, and the rowId - 40 bytes per entry.
- The native kernel gains an `encoded_var` entry type: radix passes cover the two prefix words; partitions that exhaust the prefix finish with pdqsort whose comparator memcmps the key tails in the heap, skipping the 16 bytes the prefix already decided. The fixed-key template instantiations compile to the same code as before.
- Top-K (`ORDER BY ... LIMIT`) builds the same entries in an `EncodedTopKBuffer` that stays O(limit): it sorts-and-truncates at a compaction trigger, and rejects a row up front when its leading word - plus the length, for variable keys - is already beyond the kept boundary, so a losing row skips the full key encode and the key-heap write. The parallel reduce (`AsyncTopKRecordCursorFactory`) reads fixed-width, UUID/LONG256/LONG128 and VARCHAR/STRING keys straight from the page frame's column and aux vectors per frame and encodes only the survivors.
- The cursors own the key heap (`MemoryCARWImpl`), account it against `cairo.sql.sort.key.max.bytes`, and free it right after the native sort - the emit phase reads only rowIds. Keys of up to 16 bytes never touch the heap, and a sort where every key fit the prefix skips the offset-to-pointer rebase pass entirely.

The PR also fixes a pre-existing bug this work exposed: `generateCastFunctions` had no LONG128 case and silently skipped the column, misaligning the union cast-function list. A UNION that requires casting (e.g. one carrying a SYMBOL column) and that also contains a LONG128 column would throw `UnsupportedOperationException`, or read values through a neighbouring column's function.

## Benchmark

`SELECT SearchPhrase FROM hits WHERE SearchPhrase IS NOT NULL ORDER BY SearchPhrase LIMIT $limit` on the ClickBench `hits` table. SearchPhrase is a variable-length key, so this exercises the parallel encoded top-K added here:

| LIMIT | this PR | master |
|---|---|---|
| 10 | 37 ms | 52 ms |
| 1000 | 40 ms | 58 ms |
| 100_000 | 85 ms | 100 ms |

The gap is largest at small limits, where the threshold pre-reject skips the full key encode and key-heap write for most rows; it narrows as the limit grows and more rows survive to be encoded. This is the shape the encoded path is built for; the memory and failure-mode tradeoffs below still apply (a large or unbounded variable sort can exceed the key-heap budget and error rather than fall back).

## Tradeoffs

- Variable keys materialize into the key heap: a sort on long strings pays the copy and the memory budget; exceeding `cairo.sql.sort.key.max.bytes` raises `LimitOverflowException` instead of degrading to the tree sort. STRING content inflates to 3 bytes per ASCII character (UTF-16 high-byte escaping); VARCHAR stays at byte length + 2.
- String data sharing long common prefixes degrades to comparison sort within the tied partitions - bounded by the same O(n log n) comparisons the tree fallback would perform, but the radix win shrinks.
- For the newly covered types, rows with identical keys now emit in rowId order (matching the existing encoded path) instead of the tree chain's insertion order. SQL leaves tie order unspecified; forward scans produce identical output either way.
- The native library must be rebuilt: `Vect.sortEncodedVarEntries` is a new JNI symbol.
- A VARCHAR sorted by the encoded path must be valid UTF-8. QuestDB stores VARCHAR bytes verbatim (ILP and CSV ingestion do not validate), so a value may contain a raw `0xFF` byte. The encoder detects it per value and raises an error rather than emit a wrong order; set `cairo.sql.orderby.sort.enabled=false` to sort such data through the tree path. The detection scope differs between a full sort and a top-K: the full sort encodes every key, so it always errors on a `0xFF` byte; a top-K (`ORDER BY ... LIMIT k`) only examines the bytes it needs to threshold-reject a row, so a losing row whose `0xFF` sits past those bytes is dropped without being fully encoded, and the query succeeds. The result is never wrong - a rejected row genuinely does not belong in the top-K, and any row that ties on the examined prefix is deferred to the full encode, which then errors - but whether invalid-UTF-8 data raises an error can depend on the presence of a `LIMIT` and the data distribution.

## Test plan

- OrderByEncodeSortTest: new cases for VARCHAR/STRING asc and desc with nulls, empty strings, 16-byte prefix boundary lengths and 20-byte shared prefixes; UUID/LONG256 values straddling the null bit pattern; supplementary-plane characters asserting STRING (UTF-16) and VARCHAR (UTF-8) each keep their own comparator order; fixed keys over 32 bytes; mixed fixed/variable keys in both column orders; the non-random-access UNION path; key-heap budget overflow; plan assertions. Every case runs against both the encoded path and the tree fallback with identical expected output.
- OrderByEncodeSortFuzzTest: adds UUID, LONG256, LONG128 and four string column flavours, widens the fixed-key targets to 64 bytes, and adds light and non-light rounds mixing random fixed columns with shuffled-in string columns; the tree-path output is the oracle.
- ParallelTopKFuzzTest: exercises the parallel top-K over the newly encoded types - single-column UUID, LONG256, LONG128, DECIMAL128/256, STRING and VARCHAR, asc and desc, with varying LIMIT, over native and Parquet frames - with the legacy top-K as the oracle.


kafka1991:
## Code review (level 3, blocking pass)

Large, intricate change at three sensitive boundaries: the native radix-sort kernel (`ooo.cpp`), a new JNI symbol (`Vect.sortEncodedVarEntries`), and the in-memory sort-key format. It is also unusually well-defended: every new type path is fuzz-tested against the red-black-tree sort as oracle (`OrderByEncodeSortFuzzTest`, `ParallelTopKFuzzTest`), with unique trailing keys to make tie order total. The core design is internally consistent. **No correctness-breaking (Critical) issues.**

Note: the PR still carries the **`DO NOT MERGE`** label.

### Critical
None found.

I specifically verified, and could not break, the highest-risk invariants:

- **Heap-offset vs absolute-pointer lifecycle.** `sortEncodedVarEntries` rebases `entry.key += heapAddr` once, then sorts; `encoded_var::operator<` / `encoded_tail_less` dereference `key+16` only after rebase. `sort()` leaves absolute pointers; `compact()` -> `compactVarHeap()` restores fresh offsets. Worker buffers in `AsyncTopK` are only ever `compact()`ed (never finally `sort()`ed) before `mergeVarFrom` reads them by offset — so the merge reads offsets, not stale pointers. Correct.
- **NULL encodings** for VARCHAR/STRING/SYMBOL (null = 1 byte vs empty = marker+terminator, so null sorts first asc) and the order-preserving `+1` remap that relocates the UUID/LONG256 null bit-pattern to zero (incl. the `lo == -1` carry into `hi`, and the 256-bit increment chain). Traced against the test fixtures whose ids deliberately straddle the null pattern — all correct.
- **Native comparator == `isVarBeyondThreshold` == `fastRejectsVarEntry`** define the same total order; `compareMemory` (byte-reversed SWAR) matches `memcmp` on the heap tail; `thresholdTailMem` is sized/read within bounds.
- **Split-VARCHAR prefix-6 reject** (`fastRejectsVarEntryPrefix6` + `varcharLeadingWordFromPrefix` + `VAR_PREFIX6_MASK`) correctly defers a 6-byte tie to the full encode — exactly what `testParallelTopKVarcharSplitPrefixCollision` regresses.
- **LONG128 union-cast fix.** Confirmed real: the outer `switch (toTag)` in `generateCastFunctions` has no `default`, so pre-PR a required LONG128 cast added no function and shifted every later column. The identity-`Long128Column` fix mirrors `UUID` and is covered by `UnionAllCastTest.testLong128Long128`.
- **Concurrency.** Per-worker encoder/buffer/heap are distinct objects; the owner runs as slot `-1`; rank maps are built before dispatch and read-only during it; merge is post-barrier single-threaded; native parallel partitions are disjoint over a read-only heap. Clean.

### Moderate

**1. Invalid-UTF-8 VARCHAR errors inconsistently between full-sort and top-K (behavioral).**
The top-K paths only examine the bytes they actually touch before a row is threshold-rejected: the split path validates only the 6-byte inline prefix (`varcharLeadingWordFromPrefix`), and the inline/parquet/`varcharLeadingWordFromBytes` paths only the first 8 bytes when `size >= 8`. A row carrying `0xFF` *past* the examined bytes that loses on its (valid) prefix is dropped without ever being fully encoded, so no error fires.

Consequence: on identical data, `SELECT ... ORDER BY v` (full sort) **always** errors with `invalid UTF-8`, while `SELECT ... ORDER BY v LIMIT k` may **succeed** — and even within top-K the outcome depends on the limit and data distribution. This is **not a wrong-result bug**: a row rejected on its valid prefix genuinely does not belong in the top-K, and any row that ties on the prefix is deferred to the full encode (which then throws). It is a usability/consistency wart that adding a `LIMIT` silently changes whether the query errors. Recommend documenting the asymmetry in the tradeoffs section, or accepting it as a known limitation — but as a conscious decision, not incidental.

### Minor

**2. Missing top-K coverage for invalid UTF-8 (test gap, ties to #1).** `testOrderByVarcharInvalidUtf8` only exercises the full-sort paths. There is no test for the top-K behavior (`... LIMIT k`) with a `0xFF` row — neither the "row survives and errors" case nor the "row rejected and succeeds" case, and none for the parallel/split-storage path. Given #1, this path deserves an explicit test pinning the intended behavior.

**3. `long256Key` / `uuidKey*` recompute the null/below-null classification and the `+1` remap once per word (perf).** `encodeLong256` and `appendLong256Row` call `long256Key(..., word, ...)` four times; each re-runs `Long256Impl.isNull`, `isBelowLong256Null`, and the full 256-bit increment chain on its own copy — a 4x redundant constant factor on the LONG256 encode path. `uuidKeyHi` + `uuidKeyLo` do the same 2x. Computing the adjusted words once and extracting the lanes would remove the redundancy. Per-row, on less-common types; not blocking.

**4. `encodeStringRow` / `variableRowRejected` scan the string head twice (perf).** `fastRejectsVarEntry(utf16LeadingWord(value, ...), utf16Len(value), ...)` walks the leading chars once for the word and again for the length class. Both are bounded (~4 chars), but a single pass returning both would halve the prefix scan on the STRING/SYMBOL reject path.

**5. `runBuildTimestampEarlyStop` doesn't roll back the heap for the stopping row's variable key (trivial).** The early-stop `return` skips `entries.endAppend()`; for a VARIABLE key the last `encodeCurrentRow()` already advanced the key heap, leaving those bytes orphaned. They are never read (the entry isn't counted) and are reclaimed at `clear()`/`close()`, so this is a single-key transient, not a true leak — worth a one-line note or an `endAppend`-free rollback for cleanliness.

### Downgraded (investigated, dismissed)

- **License header `/*+****` on changed files** — not a PR finding. `master` already uses `/*+****` on ~112/118 sampled files; the PR diff changes exactly one header (`EncodedTopKBuffer.java`), aligning it *to* the repo norm. Pre-existing, out of scope.
- **Double-free of the borrowed `keyHeap`** — false. `SortKeyEncoder.close()` only frees rank maps (and only when not borrowed); it never frees the `keyHeap` it received via `setKeyHeap` (owned by the buffer/cursor).
- **`mergeVarFrom` reading absolute pointers as offsets** — false. Worker buffers are only `compact()`ed during the parallel phase (offsets restored); the single final `sort()` runs only on `ownerTopK`, after merge.
- **`compareMemory` / `thresholdTailMem` over-read** — false. Reads bounded by `minLen - KEY_PREFIX_BYTES <= thresholdLen - 16 <= capacity`.
- **`isSupported` true for a type `encode` can't handle** — false. `isEncodable` returns true only for STRING/VARCHAR/SYMBOL (all handled) or `fixedColumnByteWidth >= 0`; every such type has a branch in `encodeFixedColumn`. Anything else falls back to the tree sort safely.

### Summary

**Approve once `DO NOT MERGE` is cleared** — no blocking issues. The encoding, null handling, threshold/merge logic, native-kernel consistency, JNI lifecycle, resource cleanup, and the LONG128 cast fix are correct, and the SQL-test assertions follow the builder API strictly (28 `assertQuery` / 24 `.returns` / 4 `.failsWith`, zero `assertSql` / `returnsOnce` / `getPlan`). Tradeoffs (key-heap memory, ASCII -> 3-byte STRING inflation, error-vs-fallback for invalid UTF-8) are real and largely documented.

The one judgment call is #1: the full-sort-vs-top-K asymmetry on invalid UTF-8 should be a documented, deliberate choice, with a test (#2) pinning it. Items #3-#5 are minor constant-factor / cleanliness improvements.

- Draft findings: 9 raised -> 5 verified, 4 dismissed as false positives.
- In-diff vs out-of-diff: 5 in-diff; 0 confirmed out-of-diff (the cross-context sweep over every caller of the changed symbols — `EncodedTopKBuffer` ctor, `isSupported`, `sortEncodedVarEntries`, `encodeFrame`, `SortKeyType.fromKeyLength`, the new `VarcharTypeDriver` helpers, `generateCastFunctions` — found every caller updated and correct).

kafka1991:
## Code review (level 3, blocking pass)

Independent level-3 pass over this change (native C++ kernel, the new `Vect.sortEncodedVarEntries` JNI symbol, and the in-memory sort-key format), corroborating the earlier review and adding findings beyond it. Each finding below was verified against current source.

### Critical

**None.** Three independent deep passes (native kernel, Java encoding, fresh-context adversarial) plus the cross-context sweep could not break the highest-risk invariants. Verified safe:

- **Heap offset->pointer rebase lifecycle.** `sortEncodedVarEntries` rebases `key += heapAddr` exactly once; no caller re-sorts an already-rebased array (`compact()` restores offsets via `compactVarHeap`; cursors re-encode; worker buffers are only `compact()`ed, never finally `sort()`ed, before the offset-based merge). `key` is never dereferenced when `heapAddr==0`.
- **memcmp bounds & strict weak ordering** in `encoded_var::operator<` / `encoded_tail_less` — reads bounded by `len <= heap-bytes-present`; the (k1, k2, tail, len, rowId) tie-break is a valid total order with unique rowIds.
- **Radix->pdqsort handoff** — every path that exhausts the two prefix words invokes the tail pdqsort for `encoded_var`; no prefix-tied partition is left unsorted.
- **Four VARCHAR leading-word encoders are bit-identical** to each other and to the heap bytes, and `fastRejectsVarEntry`/`fastRejectsVarEntryPrefix6`/`isVarBeyondThreshold` define the same order as the native comparator (numerically reproduced).
- **NULL/empty/below-null remaps** for VARCHAR/STRING/SYMBOL/UUID/LONG256/LONG128 — correct and collision-free.
- **No native-memory leak or double-free** on any path (happy, mid-build throw, circuit-breaker trip, overflow, re-execution, partial-init close, async worker-init partial failure); `MemoryCARWImpl.close()` is idempotent/reusable; the encoder never frees the borrowed heap.
- **Concurrency** — per-worker encoder/buffer/heap are distinct; rank maps read-only during dispatch; merge is post-barrier single-threaded; worker/owner buffers sort serially (`parallelThreshold=Long.MAX_VALUE`) so no nesting onto the worker pool.

### Moderate

**1. Invalid-UTF-8 VARCHAR errors inconsistently between full-sort and top-K (behavioral).**
The top-K reject paths validate only the bytes they touch before threshold-rejecting a row: `varcharLeadingWordFromPrefix` checks only the 6-byte inline prefix; the inline/Parquet/`varcharLeadingWordFromBytes` paths only the first 8 bytes. A row carrying `0xFF` *past* those bytes that loses on its valid prefix is dropped without a full encode, so no error fires. On identical data, `SELECT ... ORDER BY v` (full sort) **always** errors with "invalid UTF-8", while `... ORDER BY v LIMIT k` may **succeed**, and the outcome within top-K depends on the limit and data distribution. Not a wrong-result bug (a rejected row genuinely does not belong in the top-K; a prefix tie defers to the full encode, which then throws), but adding a `LIMIT` silently changing whether a query errors is a wart. The tradeoff is documented in the PR body; it should also be pinned by a test (Minor 5) and accepted as deliberate.

**2. `generateCastFunctions` outer `switch (toTag)` still has no `default` — the same silent-misalignment bug class the PR fixes for LONG128 remains open for every other un-cased type.**
The PR correctly adds `case LONG128` (a genuine fix: a required cast on a LONG128 union column previously added no cast function and shifted every later column). But the outer switch has no trailing `default`, so any *other* union-output tag without a case (CURSOR, VARARG, RECORD, generic GEOHASH, regclass, regprocedure, `text[]`, PARAMETER, UNDEFINED) would reproduce the identical silent column-misalignment. The inner VARCHAR/ARRAY switches in the same method already use `default -> assert false`. Adding it to the outer switch makes any future un-cased type fail loudly rather than corrupt the cast-function list. Reachability of those tags as projected union columns was not fully proven, so this is hardening rather than a live bug — but it is cheap, in a method the PR already edits, and prevents the exact class of bug this PR just fixed.

### Minor

**3. Full-sort variable path does not pre-size the key heap (perf).**
In `EncodedSortLightRecordCursor.buildAndSort` / `EncodedSortRecordCursor.buildAndSort`, `entryMem` is pre-sized to the known row count, but the key heap is `close()`d at build start and re-grown lazily by page-doubling from the 128 KB default page. On a large `ORDER BY varchar` (every key spills, no threshold reject on the full-sort path) this is ~log2(totalKeyBytes/128KB) reallocations, each copying the live heap. Growth is amortized O(1) so it is constant-factor, not algorithmic — but the cursor already knows `estimatedSize`; pre-`extend()`-ing the heap to e.g. `min(maxEntryMemBytes, estimatedSize * KEY_PREFIX_BYTES)` would collapse the doubling cascade, mirroring the existing entryMem pre-size.

**4. Per-row encode micro-redundancies on the variable path (perf).** Non-blocking constant factors:
- Surviving top-K rows read+normalize the leading word twice (reject helper, then `encodeVariable`/`appendKeyBytes`); bites only the survivor fraction and the cold warm-up window.
- `encodeVariable` pads keys <16 bytes to 16 on the heap, reads two words back, byteswaps, then discards via `jumpTo(start)` — the prefix words could be assembled in-register for the single-column shapes (the reject side already does this).
- `fastRejectsVarEntryPrefix6` recomputes `thresholdEntry[0] & VAR_PREFIX6_MASK` per row; cache it at `compact()` time.
- `compactVarHeap` copies every live key byte twice (gather into `scratchHeap`, bulk copy back); swapping the buffer references + a `setKeyHeap` would save the second full copy per compaction.

**5. Test coverage gaps.**
- **No top-K invalid-UTF-8 test** (ties to Moderate 1): `testOrderByVarcharInvalidUtf8` covers only full-sort. Missing: top-K "survives -> errors", top-K "rejected -> succeeds", and the parallel/split-storage variant. This is the one behavioral asymmetry with zero coverage.
- **Non-static SYMBOL variable path** is covered by a single 5-row ASC-only test (`testOrderByCastSymbolColumn`) — no DESC, no LIMIT/top-K, no parallel. (The fuzz `col_sym` is a *static* table symbol -> fixed 4-byte path, so it does not reach `KeyShape.SYMBOL`.)
- **No deterministic Parquet variable-key sort test**: serial Parquet fuzz selects only fixed columns; the Parquet VARCHAR/STRING slice encode path runs only when `ParallelTopKFuzzTest`'s random `parquet=true` axis is drawn, and only under LIMIT.
- **Serial top-K variable-key heap overflow** is untested (light/non-light/window/per-worker-parallel are covered).
- Minor mislabel: `OrderByEncodeSortFuzzTest.VAR_COLUMNS` lists `col_sym` "to force the variable key path", but a static table symbol does not — single-VAR-column iterations that draw `col_sym` silently test a fixed key under a "variable" label.

**6. SampleBy fill encoded sort is not gated by the global kill switch (pre-existing, informational).**
`SqlCodeGenerator.java:4125/4135` route through `cairo.sql.sampleby.fill.sort.strategy` (default `LIGHT_ENCODED`), independent of `cairo.sql.orderby.sort.enabled`. So the kill switch (the documented escape hatch for the invalid-UTF-8 case) does not force the SampleBy fill sort onto the tree path. Practically harmless (the fill key is always the timestamp, never VARCHAR), but worth confirming it is intentional.

### Downgraded (investigated, dismissed)

- **Prior review's perf finding on `long256Key`/`uuidKey*` recomputing the null/+1 remap per word** — does not apply to current source: those helpers do not exist; `encodeLong256`/`encodeUuid`/`appendLong256Row`/`appendUuidRow` run the classification + increment chain **once per value**.
- **Prior review's perf finding on `encodeStringRow`/`variableRowRejected` scanning the head twice via `utf16LeadingWord`** — `utf16LeadingWord` does not exist; `utf16Rejects` computes the leading word and length in a single fused loop.
- **Prior review's note that `runBuildTimestampEarlyStop` orphans the stopping row's heap key** — resolved in current diff via the added `entries.discardLastAppend()`.
- **Double-rebase / double-add of `heapAddr`** — false; every caller feeds offsets and the only re-sort path (`compact`) restores offsets first.
- **memcmp / `thresholdTailMem` over-read** — false; reads bounded by `minLen-16 <= thresholdLen-16 <= capacity`.
- **Double-free of the borrowed `keyHeap`** — false; `SortKeyEncoder.close()` frees only rank maps.
- **`mergeVarFrom` reading absolute pointers as offsets** — false; worker buffers are in offset form at merge time.
- **`isSupported` returns true for a type `encode` cannot handle** — false; `isEncodable` exactly matches the ORDER BY type-acceptance gate; non-encodable types fall back to the tree path safely.
- **AsyncTopKAtom missing kill-switch guard** — this PR *adds* the guard, fixing a latent bypass from #7241 where the parallel encoded top-K ignored `cairo.sql.orderby.sort.enabled=false`. Net improvement.

### Summary

**Approve, with two recommended follow-ups (neither blocking):** (1) a test pinning the documented invalid-UTF-8 full-sort-vs-top-K asymmetry, and (2) a `default -> assert false` on the `generateCastFunctions` outer switch to close the same bug class the PR just fixed. No correctness, memory-safety, leak, or concurrency defect found — the encoder/comparator consistency (native <-> Java), NULL handling, threshold/merge logic, JNI rebase lifecycle, resource cleanup, and the LONG128 cast fix are all correct, and the SQL test assertions follow the builder API strictly (zero `assertSql`/`returnsOnce`/`getPlan` in new code).

- Draft findings: 18 raised -> 11 verified (2 Moderate, ~9 Minor across perf/tests/metadata), 7 dismissed (4 of which were prior-review items that no longer match current source).
- In-diff vs out-of-diff: all 11 verified findings are in-diff; the cross-context sweep over every caller of the changed symbols found zero broken out-of-diff callsites.
- Tradeoffs (all documented in the PR body): STRING content inflates to ~3 bytes/ASCII char in the key heap; a large/unbounded variable sort errors rather than degrading to the tree sort; invalid-UTF-8 VARCHAR now errors on the encoded path (mitigated by the kill switch).

bluestreak01 (approved):
Approve (level-3 blocking pass).

No correctness, memory-safety, leak, concurrency, or JNI-boundary defect found. Verified the highest-risk invariants against current source: the heap offset->pointer rebase lifecycle (single rebase; workers only compact()/never finally sort() before the offset-based merge; owner does the lone final sort), native<->Java comparator equivalence, memcmp bounds, radix->pdqsort handoff for prefix-tied partitions, the fast-reject deferral logic (never over-rejects, thanks to the terminator/+1-shift invariant), NULL/below-null remaps for VARCHAR/STRING/SYMBOL/UUID/LONG256/LONG128, MemoryCARWImpl close()/re-grow reuse, and the LONG128 union-cast fix. CI is fully green (Azure Cairo/Griffin/Other/cover shards across arm64/x64/graal, Rust lint, enterprise CI, coverage 92.75%, gitleaks), and the new Vect.sortEncodedVarEntries symbol is in the committed native libs.

This PR has also absorbed the two substantive follow-ups from the earlier reviews: the top-K invalid-UTF-8 test (testOrderByVarcharInvalidUtf8TopK) and the generateCastFunctions outer 'default: assert false'. SQL test assertions follow the builder API strictly (zero assertSql/returnsOnce/getPlan in new code).

Non-blocking nits only:
- OrderByEncodeSortFuzzTest.VAR_COLUMNS lists col_sym as 'forces the variable key path', but a static table SYMBOL encodes to a fixed 1-4 byte rank; an iteration drawing only col_sym tests a fixed key under the 'variable' label (assertion still valid vs the tree oracle, just mislabeled coverage).
- Constant-factor perf polish on the variable encode path: leading word computed twice for survivors; compactVarHeap copies live key bytes twice per compaction; fastRejectsVarEntryPrefix6 recomputes the masked threshold word per row; the full-sort variable path doesn't pre-size the key heap.

The invalid-UTF-8 full-sort-vs-top-K asymmetry is documented and pinned by a test - accepted as deliberate.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 688 / 745 (92.35%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|3|4|75.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedTopKBuffer.java|121|140|86.43%|
|:large_blue_circle:|io/questdb/cairo/VarcharTypeDriver.java|10|11|90.91%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyEncoder.java|439|472|93.01%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortRecordCursor.java|30|31|96.77%|
|:large_blue_circle:|io/questdb/griffin/engine/window/EncodedWindowSortBuffer.java|25|26|96.15%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortLightRecordCursor.java|35|36|97.22%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKAtom.java|13|13|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyType.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortLightRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortLimitedLightRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortLimitedLightRecordCursor.java|5|5|100.00%|

PR #7243, opened 2026-06-12
fix(core): fix snapshot restore accepting truncated parquet partitions when _pm sidecar exists

bluestreak01:
**Tandem:** questdb/questdb-enterprise#1059 — enterprise side (backup re-upload of in-place-regenerated parquet partitions); it pins this branch via the submodule. Merge together.

Snapshot/checkpoint restore validates parquet partitions against `_txn` before generating missing `_pm` sidecars. The committed-size check previously ran only when `_pm` was absent: a restored `_pm` short-circuited the partition before any validation. A snapshot can pair `_txn` with a stale or truncated `data.parquet` and a matching old `_pm` - for example when the parquet file is regenerated in place between the moments the two files were captured - and the restore then completed silently, leaving a partition that reads garbage at query time.

The restore agent now opens `data.parquet` and checks its length against the committed size from `_txn` for every parquet partition, regardless of whether `_pm` exists, and only then decides whether `_pm` needs generating.

Effects:

- A snapshot whose parquet partition is shorter than the committed size in `_txn` now fails the restore with the existing "restored parquet file is shorter than committed size" diagnostic, even when `_pm` was captured alongside it.
- A parquet partition missing `data.parquet` now fails the restore even when `_pm` exists; previously the partition was skipped silently.
- Restores that previously appeared to succeed with such partitions now fail loudly. The partition was unreadable either way; the failure surfaces at restore time rather than at query time, and the restore can be retried against an intact snapshot.
- Each parquet partition restore performs one extra open/length call when `_pm` already exists.

Also a comment-only fix in `TableWriter.o3ConsumePartitionUpdateSink`: the parquet branch described itself as a native partition "mutated in place and then encoded back to parquet"; it consumes the in-place parquet O3 update, which appends row groups to the existing `data.parquet` at a constant directory and name txn.

Test plan:

- new `CheckpointTest#testCheckpointRestoreRejectsTruncatedParquetFileWithPmPresent` covers the `_pm`-present case
- existing `testCheckpointRestoreRejectsTruncatedParquetFile` and `testCheckpointRestoreFailsOnMissingParquetFile` pass unchanged, with unchanged diagnostics


RaphDal (dismissed):

ideoma (dismissed):

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=241773)
**Tested Commit:** `f23788ee82bf996e8b15733d837b093ddacfe6d8`

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 89 / 107 (83.18%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/DatabaseCheckpointAgent.java|2|4|50.00%|
|:large_blue_circle:|io/questdb/cairo/TableSnapshotRestore.java|87|103|84.47%|

PR #7244, opened 2026-06-12
perf(sql): push down filters into joins wrapped in views and subqueries

ideoma:
Filtering a view or subquery that wraps a join by a column of the join's master table now pushes the equality predicate into the master scan and propagates it across the equi-join keys into the slave scans, turning full table scans into keyed lookups. Previously the filter reached only the master table; the same query written inline was already optimized, so this aligns the two.

Covers INNER and LEFT joins, table/column aliases, and bind parameters as well as literal constants. Constants pinned to the slave side are still not propagated to the master, and a WHERE on a left-joined slave column still runs as a post-join filter (both unchanged).

## Example

A view that encapsulates joins, queried by the master key:

```sql
CREATE VIEW entity_stats AS (
  SELECT e.id AS entity_id, e.created_at, e.status,
         related_event.created_at AS terminal_at,
         decision_event.payload AS decision_payload
  FROM entity_terminal e
  LEFT JOIN events decision_event
    ON decision_event.entity_id = e.id
   AND decision_event.created_at = e.created_at
   AND decision_event.event_type = 'decision'
  LEFT JOIN events related_event
    ON related_event.entity_id = e.id
   AND related_event.id = e.last_event_id
   AND related_event.event_type = 'terminal'
);

SELECT * FROM entity_stats WHERE entity_id = 'abc';
```

Before, only `entity_terminal` used an index; both `events` joins were full scans. Now `entity_id = 'abc'` is propagated through `events.entity_id = e.id` into both `events` scans:

```
Hash Left Outer Join Light
  ...
    Index forward scan on: id           filter: id='abc'        -- master (already worked)
    Index forward scan on: entity_id    filter: entity_id='abc' -- slave (now pushed)
    Index forward scan on: entity_id    filter: entity_id='abc' -- slave (now pushed)
```

A bind parameter behaves the same way and is also pushed to the slave scans:

```sql
SELECT * FROM entity_stats WHERE entity_id = :id;   -- filter: entity_id=:id::string
```

## Test plan

- New EXPLAIN-plan and result-correctness tests in ExplainPlanTest (view/subquery, INNER/LEFT, aliases, bind params)
- Ran join, optimiser, where-clause, view, codegen, group-by and sample-by suites (~3,500 tests) with no regressions


ideoma:
## Code review — push down filters into joins wrapped in views and subqueries

A blocking-style review of the optimiser change (`deriveTransitiveFiltersFromPushedPredicate` + its call site in `moveWhereInsideSubQueries`). I traced the new method end-to-end, walked its single call site, ran broad regression suites, and added empirical probes comparing the view path against the equivalent inline query.

### Verdict

Approve. The change is correct and produces plans/results identical to the equivalent inline query — which is the established baseline the PR explicitly aims to match. No correctness, concurrency, or resource issues found. The notes below are follow-ups, not blockers; several are already addressed by the latest changes on the branch.

Regression status: `ExplainPlanTest` (546), `SqlOptimiserTest` (174), `JoinTest` (183) all green.

### Critical
None.

### Moderate

1. **NULL-constant guard coverage — addressed.** Added `testLeftJoinFilterWithNullConstantThroughView`. Writing it corrected an assumption from the first pass: the `nullCount > 0` guard does **not** fire for `WHERE key = NULL` — the literal folds to a deferred symbol filter (`=0`) before reaching the guard, so the constant *is* pushed onto the slave. This is not a bug: EXPLAIN confirms the inline query produces a byte-identical plan, and both return zero rows. The pushdown is lossless because `slave.key = master.key` can never match a NULL master key. The guard stays a faithful mirror of `analyseEquals`, which folds `= NULL` the same way.

2. **Headline pushdown paths assert the plan but not the rows — still open.** The INNER master-key case (`testInnerJoinFilterPushedThroughView`) and the bind-variable case (`testLeftJoinFilterByBindVariablePushedThroughView`) are plan-only. A `.returns(...)` on an INNER pushdown plus an executed bind-variable query would close this; the shared-by-reference const node makes the bind-variable execution path the most worthwhile to check at row level.

3. **Internal WHERE + consumer WHERE stacking — addressed.** Added `testLeftJoinFilterStacksWithSubQueryInternalWhere`: a subquery carrying `WHERE a.av='keep'` plus a consumer `WHERE k='x'`. Confirms the pushed predicate AND-stacks with the pre-existing internal filter rather than clobbering it, while still deriving `bkey='x'` on the slave.

### Minor

4. **Self-join and ASOF regression tests — addressed.** Added `testSelfJoinFilterPushedThroughView` (same table, same column name on both instances; confirms `addTransitiveFilters` disambiguates by name **and** join-model index) and `testAsOfJoinFilterPushedThroughView`. The ASOF test pins down a genuinely subtle point: the ASOF slave *does* receive the pushdown (the pinned master is index 0 / inner, so the guard passes), identical to inline, and the result is correct.

5. **Redundant second tree traversal — still open (negligible).** `deriveTransitiveFiltersFromPushedPredicate` re-walks `node` via `traverseNamesAndIndices` after `moveWhereInsideSubQueries` already walked it. The lhs/rhs split is new information, the path only fires for single-table `=` predicates pushed into a 2+-join nested model, and it runs at compile time. Not worth restructuring.

6. **Const-map clearing asymmetry — addressed.** Dropped the redundant trailing `constNameTo*.clear()` and added a comment recording why no post-clear is needed: `optimiser.clear()` resets the maps per query, `optimiseJoins` (the only other reader) runs in an earlier pass, and each invocation clears before its own `put`.

7. **Single-hop propagation ceiling — still open.** A one-line comment on `addTransitiveFilters` noting it propagates a constant one hop (so `A.k=B.k AND B.k=C.k` reaches B but not C, matching inline) would document the known limitation.

### Dismissed after verification

These were considered and ruled out against the source: guard-ordering short-circuit (`||` evaluates the cheap check first), ASOF divergence from inline (inline pushes to ASOF slaves too), self-join name collision (matched on name **and** index), shared-buffer corruption of the literal collectors, const-map leakage downstream, non-equality operator slipping through `joinOps`, double derivation across recursion, and a re-traversal `SqlException` aborting a previously-compiling query.

### Summary

Correct, well-scoped, and consistent with the inline baseline. Of the seven follow-ups, four are addressed (#1, #3, #4, #6) and three remain open and non-blocking (#2 row-level tests for INNER/bind-var, #5 redundant traversal, #7 documentation comment). The most useful remaining addition is a row-level assertion on the INNER and bind-variable pushdown paths.

<sub>Automated review via Claude Code.</sub>

ideoma:
## Code review — push down filters into joins wrapped in views and subqueries

Blocking-style review (level 2) of the optimiser change (`deriveTransitiveFiltersFromPushedPredicate` + its single call site in `moveWhereInsideSubQueries`). Traced the new method against the inline (non-view) baseline it replicates, walked all readers/writers of the shared const-maps, and ran adversarial queries against a live build comparing the view path to the equivalent inline query.

### Verdict

**Approve.** A tightly-scoped, compile-time-only change that faithfully replicates the inline transitive-pushdown path. No correctness, concurrency, resource, or performance issues survived verification. Results are unchanged; plans become keyed lookups. The only findings are test-coverage gaps, and they're non-blocking because every plan-only test asserts a plan byte-identical to the independently-trusted inline baseline.

### Critical
None.

### Moderate

1. **Headline INNER-join pushdown has no row-level assertion.** `testInnerJoinFilterPushedThroughView` (distinct `ta`/`tb`, both sides become keyed lookups) is plan-only. A wrong-but-plausible plan that drops/duplicates rows would pass. `testSelfJoinFilterPushedThroughView` exercises inner transitive propagation but only across two instances of the *same* table/column, not the distinct-table shape. *Fix:* add a `…ReturnsCorrectRows` sibling that inserts rows and `.returns(...)` with one surviving and one excluded master row.

2. **The bind-variable pushdown path is never executed** (most valuable addition). `testLeftJoinFilterByBindVariablePushedThroughView` is plan-only. This matters more than a normal plan gap: the derived slave filter shares the bind-variable `ExpressionNode` **by reference** with the master predicate — `deriveTransitiveFiltersFromPushedPredicate` stores `node.rhs` into `constNameToNode`, and `addTransitiveFilters` reuses it via `node.rhs = constNameToNode.get(name)`. One bind-variable node feeds two filter scans; a plan-only test renders `:kp::string` twice but never proves both scans read the bound value at execution. *Fix:* a `.returns(...)` test that `setStr("kp","x")`, executes for real, and (ideally) re-binds to a second value to prove the shared node re-reads.

3. **Missing guard-boundary test classes.** No test covers: a non-`=` operator must *not* push (`WHERE k > 'x'`, locking the `joinOps.get(node.token) != JOIN_OP_EQUAL` guard); multiple constants pushed together (`WHERE k1='x' AND k2='y'`); a three-table chain (single-hop propagation); RIGHT JOIN. Live runs of the multi-predicate and three-table cases matched inline — these are lock-ins, not suspected bugs.

### Minor

4. **Latent name/index desync is defused only by the `functionCount` guard.** `literalCollectorANames/BNames` are `ObjList` (no dedup) while `literalCollectorAIndexes/BIndexes` are `IntHashSet` (dedup); the method reads `getQuick(0)` from both. It's safe today only because the `functionCount > 0` early-return excludes any input carrying two literals on one side. A one-line comment on that guard ("also keeps names/indexes in sync — single literal per side") would prevent a future edit from silently introducing a wrong-column pushdown.

5. **Single-hop propagation ceiling is undocumented.** A one-liner noting `addTransitiveFilters` propagates a constant exactly one hop (`A.k=B.k AND B.k=C.k` reaches B but not C, matching inline) would record the known limitation.

### Dismissed after verification

- **Const-map corruption from the mid-pass clear/repopulate** — safe. `addTransitiveFilters` is the only reader of `constNameToIndex/Node/Token`; `optimiseJoins` fully precedes `moveWhereInsideSubQueries`, and nothing reads the maps afterward.
- **Wrong-slave pushdown on self-joins / name collisions** — safe. Match is on name **and** model index; a missing name returns `-1` (`NO_ENTRY_VALUE`), never equal to a valid `>=0` index.
- **Non-lossless pushdown onto LEFT/ASOF slaves** — safe. Propagation fires only along the actual equi-join key; the `joinBarriers` guard on the pinned table mirrors `analyseEquals`; the equi-key forces matched slaves to share the filtered value, preserving NULL-extension.
- **Shared const-node corrupted by in-place rewrite** — safe. The shared node is always CONSTANT/BIND_VARIABLE; `literalRewritingVisitor` only mutates `LITERAL` tokens.
- **Recursion double-derivation** — safe. Slave models are single-table and early-return on `getJoinModels().size() < 2`.
- **`assertMemoryLeak` + `.noLeakCheck()` redundant-wrapper anti-pattern** — not present. Wrapped tests run multiple statements; plan-only tests correctly use the builder's `.ddl(...)`. No `assertSql`/`returnsOnce` anywhere.

### Summary

13 draft findings → 5 verified (3 Moderate, 2 Minor), 8 dismissed as false positives after source verification. All findings in-diff; zero out-of-diff (justified — the new method is `private` with one call site, and the cross-context pass traced every reader/writer of the shared maps and the reused `addTransitiveFilters` with no broken contract). Purely additive optimisation: no path is made slower, compile-time cost is a single extra gated tree walk. The highest-value action is item #2 — execute the bind-variable pushdown, since the shared-by-reference const node is the one genuinely non-obvious mechanism and is currently proven only at the plan level.

<sub>Automated review via Claude Code (level 2).</sub>

bluestreak01:
## Code review (level 3) — re-review at `f231a73` with the const-map fix

A re-review of the optimiser change at the current HEAD, which now includes the stale-const-map fix. Because that fix was authored in the prior review pass, I treated myself as the most-anchored possible reviewer: I re-derived the const-map invariant from scratch, ran three independent adversarial passes, and **settled two contested points with live EXPLAIN/execution probes** rather than reasoning alone.

### Verdict

**Approve.** The Critical regression flagged in the first pass is fixed on this HEAD, and the fix is verified sound from four independent angles. Remaining items are non-blocking test-coverage suggestions; the most important one (bind-variable execution) was empirically de-risked, and a reviewer-vs-reviewer contradiction (NULL pushdown) was resolved in the code's favor.

### Critical

**None.** The previously-confirmed Critical — `deriveTransitiveFiltersFromPushedPredicate` reading stale `constNameTo*` entries left by an IN-lambda's join pass, injecting a phantom slave filter and dropping rows — is **fixed**, and the fix verifies:

- **Reader-side-reset invariant (re-mapped, not trusted).** The only reader of `constNameToIndex/Node/Token` is `addTransitiveFilters` (`SqlOptimiser.java:1499`), reachable from exactly two callsites — `optimiseJoins:5814` and `deriveTransitiveFiltersFromPushedPredicate:2973` — each immediately preceded by `clearConstNameMaps()` (5785, 2969). `mergeConstIntoPostJoinWhereClause` (4988) is **not** a reader (it touches `constWhereClause`).
- **Recursion ordering is safe.** Within `optimiseJoins(n>1)`: clear (5785) → populate via `processJoinConditions` (5802–5806) → read via `addTransitiveFilters` (5814) → only *then* recurse (5818+). A child's entry-clear cannot wipe a parent's maps before the parent reads them.
- **The removed union-recursion clear is fully subsumed** by the new entry-clear, which protects nested and union children uniformly; `n==1` models never reach a reader.
- **Pipeline order confirms it:** `optimiseExpressionModels`(11928) → `optimiseJoins`(11950) → `moveWhereInsideSubQueries`(11962); both downstream readers clear-before-read, closing both the introduced regression and the pre-existing master-side variant.
- **`derive` faithfully mirrors `analyseEquals`** — operand selection, the `joinBarriers` outer/asof guard (2961 ↔ 1634/1713), the `nullCount` guard, plus a stricter `functionCount>0` guard.

Cross-context check: the rename `clearForUnionModelInJoin → clearConstNameMaps` leaves **zero references to the old name** anywhere (no compile break); the maps are `private final` with no external access; the optimiser is thread-confined (one per pooled `SqlCompilerImpl`), so the instance-field maps cannot race.

### Moderate (non-blocking — test coverage only)

1. **The bind-variable pushdown path is asserted but never *executed*.** `testLeftJoinFilterByBindVariablePushedThroughView` is plan-only (`.assertsPlan`), despite `setStr("kp","x")`. This is the highest-value gap because the derived slave filter shares the bind `ExpressionNode` *by reference* with the master predicate (`derive` stores `node.rhs`; `addTransitiveFilters` reuses it at 1517) — one node feeds two scans. **I executed it live:** `kp='x'` returns exactly `x/ax/bx`, then re-binding `kp='y'` returns exactly `y/ay/by`. So both scans read the bound value *and* the shared node correctly re-reads on re-bind. **Conclusion: the mechanism is correct; this is a coverage gap, not a bug.** Worth adding a `.returns()` + re-bind test to lock it in.

2. **The headline INNER pushdown has no distinct-table row-level assertion.** `testInnerJoinFilterPushedThroughView` is plan-only. Inner transitive propagation is exercised at row level only by `testSelfJoinFilterPushedThroughView` (same table/column twice) and the LEFT `…ReturnsCorrectRows`. A distinct-table INNER `.returns(...)` would close the residual gap. Low risk given existing coverage.

### Minor

3. **Single-hop propagation ceiling is undocumented.** `addTransitiveFilters` propagates a constant exactly one hop (`A.k=B.k AND B.k=C.k` reaches B, not C, matching inline). A one-line comment would record the known limitation.

4. The comment on `testLeftJoinFilterWithNullConstantThroughView` calls the `=0` filter a "deferred symbol filter," but the live plan renders it `filter: akey=0` *without* the `deferred: true` marker (unlike the `='x'` case). The substantive claim is correct; only the word "deferred" is slightly imprecise.

### Out of scope (pre-existing master bug — unchanged by this PR)

- `col=const` pushed pre-join through FULL/RIGHT joins NULL-extends ghost rows on master. This PR's guard faithfully mirrors `analyseEquals` and does **not** worsen it; the PR doesn't claim to fix it. Worth a separate issue.

### Dismissed after live verification (false positive)

- **"`derive` bails on `nullCount>0`, so the NULL-constant slave gets no filter; the test comment is misleading"** — **FALSE.** Live EXPLAIN of `WHERE k = NULL` shows the slave scan getting `filter: bkey=0`, byte-identical to inline. `= NULL` on a symbol folds to symbol-key `=0` (token `"0"`, not `"null"`) *before* `derive` runs, so `nullCount` stays 0 and the constant *is* pushed onto the slave. The comment is substantively correct.

### Summary

2 Moderate + 2 Minor verified; 1 draft finding dropped as a false positive after live verification. All in-diff (0 out-of-diff — the new method is `private` with one callsite, the renamed method has zero external callers, and the const-maps are `private final`). The fix is sound; no new defects introduced by the rename, the new entry-clear, or the removed union-clear.

<sub>Automated review via Claude Code (level 3).</sub>

bluestreak01 (approved):
Approving after a level-3 re-review of HEAD 4d09333.

The stale-const-map regression from the earlier pass is fixed. The reader-side-reset invariant holds (the sole reader, `addTransitiveFilters`, is clear-before-read at both callsites); the `optimiseJoins` recursion order (clear -> populate -> read -> recurse) keeps a child's entry-clear from wiping a parent's maps; and the removed union-recursion clear is fully subsumed by the new entry-clear. The rename to `clearConstNameMaps` leaves no references to the old name, the maps are `private final`, and the optimiser is thread-confined per pooled compiler.

Two points were settled empirically rather than by reasoning:
- `WHERE k = NULL` on a symbol column folds to symbol-key `=0` before the derive pass, so the constant is pushed onto the slave (`bkey=0`), byte-identical to the inline query.
- The bind-variable path executes correctly, and the shared-by-reference const node re-reads on re-bind rather than freezing on the first value.

4d09333 adds the two row-level tests that were the remaining non-blocking coverage gaps: the INNER distinct-table case with a `withPlanContaining` fragment that guards the transitive push (a pure row check would pass either way), and the executed bind-variable path with a re-bind.

The pre-existing FULL/RIGHT-join ghost-row behavior on master is unchanged by this PR and worth a separate issue.

<sub>Automated review via Claude Code (level 3).</sub>

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 24 / 27 (88.89%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|24|27|88.89%|

PR #7247, opened 2026-06-12
chore(ilp): fix QWP symbol corruption under orphan adoption

mtopolnik:
## Summary

Concurrent QWP ingestion into a table's `SYMBOL` column can store the
wrong symbol for a row after a client sender restart (orphan-adoption
replay). The stored value is a valid-but-wrong symbol from the same
column — e.g. a row whose symbol should be `sym_4` reads back as
`sym_10`.

## Root cause

The per-connection `ClientSymbolCache` maps a client's symbol-dictionary
index (`clientSymbolId`) to the table's symbol key, and is invalidated
only when the column's symbol-count **watermark** changes
(`ClientSymbolCache.checkAndInvalidate`).

QWP orphan-adoption replays an earlier sender's self-sufficient frames
(a full symbol dictionary from id 0) into the adopting sender's
connection, ahead of its own frames. The two senders assign
`clientSymbolId`s in independent first-appearance order, so the replayed
dictionary **remaps** existing `clientSymbolId`s to different strings.
Once the table's symbol set is fully populated the watermark stops
moving, so the watermark check never fires again and a stale cache hit
returns the prior sender's key — the row is stored under the wrong
symbol.

The connection dictionary (`connectionSymbolDict`) is overwritten
correctly per frame; only the cache goes stale. Concurrency makes it
common: multiple senders writing the same small symbol set saturate the
watermark quickly and adopt one another's frames often, while a single
sender almost never trips it.

## Fix

- `QwpMessageCursor` flags a symbol-dictionary **redefinition** — an
  existing `clientSymbolId` overwritten with a *different* string. An
  identical re-send (the common dict-from-0 case within one sender) or a
  pure append is not a redefinition, so the cache stays warm in the
  steady state.
- `QwpIngressProcessorState.processMessage()` clears the connection
  symbol cache for a message that carries such a remap, before applying
  its rows, forcing the affected symbols to re-resolve against the
  refreshed dictionary.

## Testing

- New `QwpSymbolDecoderTest#testSymbolDictRedefinitionDetected` unit test
  covering the remap, identical re-send, and pure-append cases.
- Reproduced via the Go client's QWP restart-replay fuzz scenario (3
  concurrent producers with sender restarts and orphan adoption),
  repeated many times: corrupted within ~14–19 repetitions before the
  fix; **320 repetitions clean** afterward (120 with server bounces +
  200 without).

## Notes

Surfaced by the Go client's QWP ingress-oracle restart-replay fuzz test.
The bug is concurrency- and adoption-gated: a single producer is
effectively immune, while ≥2 concurrent producers writing a shared
symbol column trip it readily.


bluestreak01 (approved):
Approved.

Verified the fix end-to-end: the connection symbol dictionary is shared across all symbol columns on the connection, so a remapped clientSymbolId can poison any column's cache; clearing the entire connection cache on a detected remap is the correct (and necessary) response. The redefinition detection (previous != null && !previous.equals(symbol)) correctly distinguishes a true remap from an identical re-send or a pure append, and the flag is reset per message. UDP ingress is unaffected (it decodes with a null dict, so the symbol cache is never engaged).

Non-blocking follow-ups:
- Consider a deterministic regression test driving processMessage() with a remap and asserting the stored symbol value, in addition to the unit test that covers detection. The end-to-end path is currently guarded only by the randomly-seeded oracle fuzz tests.
- Title: prefer fix(ilp): over chore(ilp): since this is a data-corruption fix and the squashed title becomes the release-notes line. Adding Bug/ILP labels would match.
- Minor: in QwpSymbolDecoderTest, the new test method and decodeDeltaDict helper are slightly out of alphabetical/grouping order.

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/qwp/protocol/QwpMessageCursor.java|8|8|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=241516)
**Tested Commit:** `68fc7506f3ef49800db58def80361174a6a70fe6`

PR #7249, opened 2026-06-12
fix(core): fix unexpected logout during HTTP session rotation

### LINKED ISSUE #7235: HttpSessionStoreImpl: rotation grace period defeated by torn read of sessionId+rotateAt (spurious 401 on in-flight requests)
## Summary

`HttpSessionStoreImpl` can **evict a session's old ID during the rotation grace period**, because `SessionInfo.rotate()` updates two coupled fields non-atomically and `evictExpiredSessions()` reads them non-atomically under a *different* lock. An eviction sweep that interleaves between the two writes sees the **new** `sessionId` together with the **stale** `rotateAt`, computes an eviction time in the past, and drops the old ID immediately instead of after the grace period — so an in-flight request still carrying the old session cookie can be spuriously rejected as unauthenticated (HTTP 401).

Found with controlled-concurrency testing (Fray); reproduces deterministically within a few dozen explored interleavings.

## Root cause

When a session rotates, [`SessionInfo.rotate`](https://github.com/questdb/questdb/blob/089e8a2894a7da3cbbd73c50a7479e737b8f387d/core/src/main/java/io/questdb/cutlass/http/HttpSessionStore.java#L158-L161) writes two **logically-coupled** fields in sequence (under the session's `tryLock`):

```java
void rotate(String newSessionId, long nextRotationAt) {
    this.sessionId = newSessionId;   // write 1
    this.rotateAt  = nextRotationAt; // write 2
}
```

Both fields are `volatile`, so this is **not** a visibility problem — it is a non-atomic update of a pair.

Meanwhile [`evictExpiredSessions`](https://github.com/questdb/questdb/blob/089e8a2894a7da3cbbd73c50a7479e737b8f387d/core/src/main/java/io/questdb/cutlass/http/HttpSessionStoreImpl.java#L218-L239) — `synchronized(this)`, a **different lock** than the session's `tryLock` — reads the same pair non-atomically while deciding whether an entry is an evictable "old rotated ID":

```java
} else if (!Chars.equals(sessionId, sessionInfo.getSessionId())) {                   // reads NEW sessionId
    final long evictRotatedAt = sessionInfo.getRotateAt() - rotatedSessionEvictionTime; // reads STALE rotateAt
    if (evictRotatedAt < currentMicros) {
        iterator.remove();                                                            // evicts the old ID
    }
}
```

Interleaved between write 1 and write 2, eviction observes:
- `getSessionId()` = **new** ID → the old map key `!= sessionId` → classified as an "old rotated ID", eligible for eviction;
- `getRotateAt()` = **stale** (pre-rotation) value → `evictRotatedAt = oldRotateAt - grace` lands **in the past** → `evictRotatedAt < currentMicros` is true → the old ID is removed **immediately**, ~`rotatedSessionEvictionTime` earlier than the grace period intends.

This defeats the documented guarantee on the old ID:

> Old session IDs remain valid for `rotatedSessionEvictionTime` after rotation to allow clients with in-flight requests to transition smoothly.

### Interleaving

- Thread A (`verifySessionId`, rotating): `rotate()` executes write 1 (`sessionId = new`) and is paused before write 2.
- Thread B (`verifySessionId`, eviction due): runs `evictExpiredSessions`, visits the old-ID entry, sees `sessionId = new` (old key) + `rotateAt = stale` → `evictRotatedAt` in the past → removes the old ID.
- Thread A: completes write 2.

The session object stays alive (the new ID still resolves, `invalid == false`); only the old ID key is gone — confirmed by the harness diagnostic `evicted=true, sessionStillLive=true, si0.invalid=false`.

## Reproduction (Fray harness, minimised)

Create a session, advance a deterministic clock past `rotateAt` (rotation due) but below `expiresAt` (not expired), then fire several concurrent `verifySessionId` calls on the current ID. Oracle: every concurrent request on the valid session resolves to a non-invalid `SessionInfo`, exactly one rotation occurs, and the old ID remains resolvable during the grace period.

```java
clock.set(1_000_000_000L);
HttpSessionStoreImpl store = new HttpSessionStoreImpl(cfg);   // cfg injects an advancing test clock
store.setTokenGenerator(() -> "s" + counter++);              // deterministic ids: s0, s1, ...
store.createSession(alice, createCtx);
String s0 = createCtx.getSessionIdSink().toString();
SessionInfo si0 = store.getSession(s0);
clock.set(si0.getRotateAt() + 1);                           // rotation due, not expired

// 3 concurrent in-flight requests on the current id
// each: got[i] = store.verifySessionId(s0, reqCtx[i]);
// ... start, join ...

assert store.getSession(s0) == si0;   // FAILS: old id evicted mid-rotation, though the session is alive
```

Under Fray this fails within a few dozen iterations with `store.getSession(s0) == null` while `getSession(newId) == si0` and `!si0.isInvalid()` — i.e. premature eviction of a live session's old ID, not a destroy/expiry.

## Impact

Sole consumer is [`HttpConnectionContext.configureSecurityContext`](https://github.com/questdb/questdb/blob/089e8a2894a7da3cbbd73c50a7479e737b8f387d/core/src/main/java/io/questdb/cutlass/http/HttpConnectionContext.java#L586-L637):

```java
if (authenticator.authenticate(headerParser)) principalContext = authenticator;  // header credentials
else if (sessionInfo != null)                 principalContext = sessionInfo;     // cookie session
else return false;                                                                // auth fails -> 401
```

- **Cookie-reliant clients (the Web Console / browser UI)** send only the session cookie. When the old ID is prematurely evicted, `verifySessionId` returns `null`, `authenticate(headerParser)` is `false` → `configureSecurityContext` returns `false` → **the in-flight request is rejected (401)**. It does not even reach the "client sent an evicted id → create a new session" recovery (that code is after the `return false`).
- **Credentialed REST clients** (Basic-auth header per request) are unaffected — `authenticate()` succeeds; the only effect is a fresh session being minted.

User-visible symptom: an **intermittent, spurious 401 / "logged out"** on a Web Console request around a rotation boundary (rotation every `sessionTimeout/2`), most likely when the browser fires concurrent requests and some are in flight on the old cookie. Transient and self-correcting — the next request uses the new cookie.

### Severity: low

- **Fail-closed, not fail-open** — the bug wrongly *denies* a valid session; it never grants access. **No security impact** (no hijack, no privilege escalation, no cross-session leakage).
- Transient and self-correcting; the session object stays alive.
- Reachability requires an eviction sweep to coincide with a rotation's narrow torn-read window while in-flight requests still use the old ID — rare per session, but surfaces as hard-to-diagnose intermittent auth blips on busy multi-user instances (e.g. auto-refreshing dashboards).

## Suggested fix

Swap the two writes in `rotate()` so `rotateAt` is written **before** `sessionId`:

```java
void rotate(String newSessionId, long nextRotationAt) {
    this.rotateAt  = nextRotationAt; // write 1
    this.sessionId = newSessionId;   // write 2
}
```

Then a torn read observes `(sessionId = old, rotateAt = new)`: eviction sees `sessionId == mapKey` → not an "old rotated ID" → skips it (safe). Once both writes land, state is consistent and the old ID evicts at the correct grace-period time. (Alternatives: read both fields under the session lock in `evictExpiredSessions`, or fold the pair into a single atomically-swapped object.)

---

Affected code: `core/src/main/java/io/questdb/cutlass/http/HttpSessionStoreImpl.java` + `HttpSessionStore.java` (introduced in #6277), `master` @ `089e8a2`.

brunocalza:
## Summary

Fixes https://github.com/questdb/questdb/issues/7235

During session id rotation, `SessionInfo.rotate()` updates the coupled sessionId and rotateAt fields in sequence, while the eviction sweep reads them under a different lock. An eviction that interleaves between the two writes observes the new sessionId together with the stale rotateAt, computes an eviction time in the past, and drops the old session id immediately instead of keeping it valid through the grace period. A request still in flight on the old cookie is then rejected with HTTP 401, so the user's Web Console session appears to log out.

## Implementation
I opted for a more robust fix than the one suggested by the issue. The issue suggests swapping the order of the writes. Although that can resolve the issue, it is not an explicit fix. I opted for adding a new record [`RotationInfo`](https://github.com/questdb/questdb/compare/master...brunocalza:questdb:bcalza/fix-session-rotation-concurrency?expand=1#diff-d43384907da08c7628215dfe16746478d1ffb07517296d5a0ff6f99c064e3a20R181). `rotate()` uses `RotationInfo` to do a single write of both sessionId and rotateAt. The sweep reads that record as one snapshot. 

## Testing
- Existing HttpCookieTest and HttpCookieConcurrentTest pass
- I set up a [Fray](https://github.com/cmu-pasta/fray) test outside this PR, and it confirmed the fix resolves the issue
- I manually tested using the web console, setting the session timeout to 1m, and checked the [log](https://github.com/questdb/questdb/compare/master...brunocalza:questdb:bcalza/fix-session-rotation-concurrency?expand=1#diff-29c1fdffd4b97dd6edb034d242c0e0ae4ed38dd395aae25d7db3ee31bc7bf2d1R232) to confirm the flow the code is triggered
- I added a unit [test](https://github.com/questdb/questdb/compare/master...brunocalza:questdb:bcalza/fix-session-rotation-concurrency?expand=1#diff-092a994df764052c78a97b13c48dadfa08f82104aae2407cf09cca9fbc3d3b30R89). It is not related to this fix. It is just to make sure [these](https://github.com/questdb/questdb/compare/master...brunocalza:questdb:bcalza/fix-session-rotation-concurrency?expand=1#diff-29c1fdffd4b97dd6edb034d242c0e0ae4ed38dd395aae25d7db3ee31bc7bf2d1R228-R229) lines are covered.

### DIFF
--- a/core/src/main/java/io/questdb/cutlass/http/HttpSessionStore.java
+++ b/core/src/main/java/io/questdb/cutlass/http/HttpSessionStore.java
@@ -1,6 +1,7 @@
 package io.questdb.cutlass.http;
 
 import io.questdb.cairo.security.PrincipalContext;
+import io.questdb.std.Chars;
 import io.questdb.std.ConcurrentHashMap;
 import io.questdb.std.Misc;
 import io.questdb.std.ObjList;
@@ -89,16 +90,14 @@ public interface HttpSessionStore {
         private final String principal;
         private volatile long expiresAt;
         private volatile boolean invalid = false;
-        private volatile long rotateAt;
-        private volatile String sessionId;
+        private volatile RotationInfo rotationInfo;
 
         public SessionInfo(@NotNull String sessionId, String principal, @NotNull ConcurrentHashMap<ReadOnlyObjList<CharSequence>> groupsByEntity, byte authType, long expiresAt, long rotateAt) {
-            this.sessionId = sessionId;
+            this.rotationInfo = new RotationInfo(sessionId, rotateAt);
             this.principal = principal;
             this.groupsByEntity = groupsByEntity;
             this.authType = authType;
             this.expiresAt = expiresAt;
-            this.rotateAt = rotateAt;
         }
 
         @Override
@@ -122,11 +121,11 @@ public interface HttpSessionStore {
         }
 
         public long getRotateAt() {
-            return rotateAt;
+            return rotationInfo.rotateAt;
         }
 
         public String getSessionId() {
-            return sessionId;
+            return rotationInfo.sessionId;
         }
 
         public void invalidate() {
@@ -143,21 +142,27 @@ public interface HttpSessionStore {
 
         @Override
         public String toString() {
+            final RotationInfo rotationInfo = this.rotationInfo;
             final StringSink sink = Misc.getThreadLocalSink();
             sink.put("SessionInfo [principal=").put(principal)
                     .put(", groups=").put(getGroups())
                     .put(", authType=").put(authType)
                     .put(", expiresAt=").put(expiresAt)
-                    .put(", rotateAt=").put(rotateAt)
-                    .put(", sessionId=").put(sessionId)
+                    .put(", rotateAt=").put(rotationInfo.rotateAt)
+                    .put(", sessionId=").put(rotationInfo.sessionId)
                     .put(", invalid=").put(invalid)
                     .put("]");
             return sink.toString();
         }
 
+        boolean isEvictableOldSessionId(@NotNull CharSequence mappedSessionId, long evictionGracePeriod, long currentMicros) {
+            final RotationInfo rotationInfo = this.rotationInfo;
+            return !Chars.equals(mappedSessionId, rotationInfo.sessionId)
+                    && rotationInfo.rotateAt - evictionGracePeriod < currentMicros;
+        }
+
         void rotate(String newSessionId, long nextRotationAt) {
-            this.sessionId = newSessionId;
-            this.rotateAt = nextRotationAt;
+            rotationInfo = new RotationInfo(newSessionId, nextRotationAt);
         }
 
         boolean tryLock() {
@@ -166,6 +171,14 @@ public interface HttpSessionStore {
 
         void unlock() {
             lock.set(false);
+        }
+
+        /**
+         * Immutable (sessionId, rotateAt) pair. The two values are logically coupled:
+         * rotation must replace them in a single volatile store, so concurrent readers
+         * never observe the id of one rotation generation with the deadline of another.
+         */
+        private record RotationInfo(String sessionId, long rotateAt) {
         }
     }
 }
--- a/core/src/main/java/io/questdb/cutlass/http/HttpSessionStoreImpl.java
+++ b/core/src/main/java/io/questdb/cutlass/http/HttpSessionStoreImpl.java
@@ -198,7 +198,9 @@ public class HttpSessionStoreImpl implements HttpSessionStore {
      *       still pointing to a rotated session), and grace period has expired</li>
      * </ol>
      *
-     * <p>For old rotated IDs, the eviction time is calculated as:
+     * <p>For old rotated IDs, {@link SessionInfo#isEvictableOldSessionId(CharSequence, long, long)}
+     * makes the decision over an atomic snapshot of the (sessionId, rotateAt) pair, with the
+     * eviction time calculated as:
      * {@code evictRotatedAt = sessionInfo.rotateAt - rotatedSessionEvictionTime}
      * <p>
      * This works because {@code rotateAt} stores when the <i>next</i> rotation should occur.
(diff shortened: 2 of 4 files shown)


bluestreak01 (approved):
Approved.

The fix is correct, complete, and minimal. Folding the coupled `sessionId` and `rotateAt` into an immutable `RotationInfo` record behind a single volatile reference closes the torn read in `evictExpiredSessions` at the JMM level (single volatile store of an all-final record = safe publication). The eviction predicate is preserved byte-for-byte, the per-request path stays allocation-free (the one allocation in `rotate()` is on a cold path, at most once per rotationPeriod per session), and all 9 callsites of the changed symbols were verified SAFE with zero out-of-diff breakage.

Non-blocking follow-ups:
- Add labels (Bug, Core, REST API) to match the PR scope.
- The torn-read race itself has no deterministic in-repo regression test (understandable — it needs a controlled scheduler like Fray/jcstress). Consider a guard comment at rotate()/isEvictableOldSessionId() tying back to the atomic-pair invariant, since the getters still allow a future caller to re-introduce the tear by reading sessionId and rotateAt separately.

PR #7251, opened 2026-06-13
chore(core): stop returning a reader to the pool when its refresh fails

RaphDal:
## Summary

Under concurrent table reads, a reader pulled from the pool occasionally fails with `NullPointerException: ... roTxMemBase is null`, and the run can finish holding leaked reader file descriptors. It shows up when many threads acquire readers for the same table while that table is being written to or altered, so pooled readers are frequently refreshed and a refresh sometimes fails. This change closes the race that causes both symptoms.

## Root cause

The cleanup for a failed reader refresh is split across two methods, and the first one returns the reader to the pool instead of disposing it.

When `ReaderPool.R.refresh()` (or `refreshAt()`) sees `goActive()` throw, its `catch` calls `close()`:

```java
public void refresh(ResourcePoolSupervisor<R> supervisor) {
    this.supervisor = supervisor;
    try {
        goActive();
    } catch (Throwable ex) {
        close();    // R.close() -> goPassive() + returnToPool()
        throw ex;
    }
}
```

For a still-pooled reader, `R.close()` is not a dispose: it calls `goPassive()` then `returnToPool(this)`, which sets `allocations[i] = UNALLOCATED` and returns `true`, leaving the reader open and still referenced at `entry.getTenant(i)`.

`AbstractMultiTenantPool.get0()` then runs its own cleanup on the rethrow:

```java
} catch (Throwable th) {
    tenant.goodbye();           // pool = null, entry = null
    tenant.close();             // pool == null now -> super.close() disposes the reader
    e.assignTenant(i, null);    // slot finally cleared here
    Unsafe.arrayPutOrdered(e.allocations, i, UNALLOCATED);
    throw th;
}
```

So `close()` runs twice on a failed refresh. The first call re-publishes the slot as available; the second tears the reader down. In between — for the whole duration of `super.close()` — `allocations[i]` is `UNALLOCATED` while `entry.getTenant(i)` still points at the reader being freed.

A concurrent `get0()` that CAS-acquires that slot reads the still-assigned tenant and calls `refresh()` on it. `TableReader.close()` frees `txFile` (nulling `roTxMemBase`) several statements before `freeTempMem()` flips `isOpen()` to `false`, so the reader still reports open while its `_txn` mapping is already gone, and the reload dereferences it:

```
java.lang.NullPointerException: Cannot invoke "...MemoryMR.getLong(long)" because "this.roTxMemBase" is null
    at io.questdb.cairo.TxReader.unsafeReadVersion(TxReader.java:715)
    at io.questdb.cairo.TableReader.readTxnSlow(TableReader.java:1569)
    at io.questdb.cairo.TableReader.goActive(TableReader.java:593)
    at io.questdb.cairo.pool.ReaderPool$R.refresh(ReaderPool.java:263)
    at io.questdb.cairo.pool.AbstractMultiTenantPool.get0(AbstractMultiTenantPool.java:316)
```

The same double-ownership corrupts the allocation/tenant bookkeeping (a reader can be assigned to a slot the pool considers free, or trip the `"double close"` guard on return), which orphans open readers and leaks their file descriptors.

## The fix

Make `get0()` the sole disposer on a failed refresh: drop the failure-path `close()` from `R.refresh()` and `R.refreshAt()`.

```java
public void refresh(ResourcePoolSupervisor<R> supervisor) {
    this.supervisor = supervisor;
    goActive();
}
```

Because `get0()`'s `catch` calls `goodbye()` before `close()`, that `close()` goes straight to `super.close()` (it does not call `returnToPool()`), and `allocations[i]` stays owned by the failing thread until the slot is cleared. There is no interval in which the reader is both released and still assigned, so a concurrent `get()` can no longer acquire a reader that is being torn down.

`refresh()` and `refreshAt()` are called only from `get0()`, which already disposes the tenant on any throw, so the removed `close()` was redundant. `supervisor.onResourceReturned(...)` is still invoked (inside the `close()` that `get0()` runs), so reader leak tracking is unchanged.

## Behavior change and risks

- Happy path (refresh succeeds) is unchanged.
- The only behavioral difference is on the refresh-failure path, where disposal now happens once, in `get0()`, instead of a return-to-pool followed by a dispose.
- The change deletes a cleanup branch, so the risk is a caller that depended on `refresh()`/`refreshAt()` self-disposing on failure. Both are invoked only from `get0()`, which owns that cleanup, so no such caller exists in the tree.
- This is a concurrency fix; the timing window is small, which is why it surfaced only under heavy concurrent load. A full CI run is worthwhile before merge.

## Scope: the same shape in `WalWriterPool`

`WalWriterPool.WalWriterTenant.refresh()` has the same textual shape this PR removes — `try { goActive(); } catch (Throwable ex) { close(); throw ex; }` — and `WalWriterPool` extends the same `AbstractMultiTenantPool`, so `get0()` is already its disposer too. It is left unchanged here because, unlike the reader, it is not a reachable race:

- `WalWriterTenant.close()` returns the writer to the pool only on its `!isDistressed()` branch. A distressed writer is disposed in place via `super.close()` + `expelFromPool()`, and `expelFromPool()` clears the tenant before releasing the slot, so there is no released-but-still-assigned window.
- `WalWriter.goActive()` marks the writer distressed on every `CairoException` (file/mmap errors, metadata validation, sequencer reads, version mismatch). `EntryUnavailableException` extends `CairoException`, so it is covered too. The only declared non-Cairo exception on that path, `AlterTableContextException`, is thrown solely under `if (!contextAllowsAnyStructureChanges)`, and the WAL replay calls `apply(metaWriterSvc, true)`, so it is unreachable.

So every normal failure of a WAL writer refresh sets `distressed` and takes the race-free disposal path. The reader has no such guard — `R.close()` always returns to the pool on failure — which is exactly why the race is reachable for readers and not for WAL writers. Dropping the `close()` from `WalWriterPool.refresh()` would be defense-in-depth against an unexpected non-`CairoException` (a bug or `Error`) reopening the window, not a bug fix, so it stays out of scope for this PR.

## Test plan

- Adds `ReaderPoolTest.testConcurrentGetRefreshFailureDoesNotHandOutDisposedReader`, a deterministic reproduction of the `refresh()`/`get()` path. It uses a single-slot pool so two threads contend for the same reader, an injected `openRO(_meta)` failure to make one thread's refresh throw, and a `FilesFacade` that parks that thread inside its disposal (after `_txn` is freed) and releases the second thread to contend for the just-released slot. The test fails on the current code with the NPE above and passes with this change.
- Adds `ReaderPoolTest.testConcurrentGetCopyOfRefreshFailureDoesNotHandOutDisposedReader`, the matching reproduction for the `refreshAt()`/`getCopyOf()` path, which this PR changes identically. It uses a two-slot pool — a checked-out source reader holds one slot while a newer-txn reader returned to the other slot is the refresh target — so `getCopyOf()` takes `goActiveAtTxn()`'s downgrade path; an injected `openRO` failure on the symbol-offset (`.o`) file makes `refreshAt()` throw, and the same disposal gate lets a concurrent `get()` contend for the slot. It fails on the unfixed `refreshAt()` with the same NPE and passes with this change.
- The full `ReaderPoolTest` suite passes (33 tests), including the failure-path and race tests `testGetReaderFailure`, `testGetAndCloseRace`, and `testReaderDoubleClose`.
- Reverting only the `ReaderPool.java` change reproduces both failures, confirming the tests are tied to this fix. Each test pins its own half — re-adding the failure-path `close()` to `refreshAt()` alone reproduces the second test's NPE, leaving the first test green.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/pool/ReaderPool.java
+++ b/core/src/main/java/io/questdb/cairo/pool/ReaderPool.java
@@ -259,23 +259,17 @@ public class ReaderPool extends AbstractMultiTenantPool<ReaderPool.R> {
         @Override
         public void refresh(ResourcePoolSupervisor<R> supervisor) {
             this.supervisor = supervisor;
-            try {
-                goActive();
-            } catch (Throwable ex) {
-                close();
-                throw ex;
-            }
+            // Do NOT close() here on failure: get0() owns disposal via goodbye()+close().
+            // Closing here returnToPool()'s the reader (slot released, tenant still assigned)
+            // before get0() disposes it, letting a concurrent get() acquire a torn-down reader.
+            goActive();
         }
 
         @Override
         public void refreshAt(@Nullable ResourcePoolSupervisor<R> supervisor, R srcReader) {
             this.supervisor = supervisor;
-            try {
-                goActiveAtTxn(srcReader);
-            } catch (Throwable ex) {
-                close();
-                throw ex;
-            }
+            // See refresh(): get0() owns disposal on failure; closing here re-publishes the reader.
+            goActiveAtTxn(srcReader);
         }
 
         @Override
--- a/core/src/test/java/io/questdb/test/cairo/pool/ReaderPoolTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/pool/ReaderPoolTest.java
@@ -368,6 +368,197 @@ public class ReaderPoolTest extends AbstractCairoTest {
         });
     }
 
+    // Covers the refreshAt()/getCopyOf() half of the same fix as
+    // testConcurrentGetRefreshFailureDoesNotHandOutDisposedReader. PR #7251 removed the failure-path
+    // close() from BOTH R.refresh() and R.refreshAt(); this exercises the refreshAt() branch.
+    //
+    // getCopyOf(src) requires src to be a pooled R holding its own slot, and get0()'s copyOf branch
+    // only calls tenant.refreshAt(supervisor, copyOfTenant) when the CAS-acquired slot already holds a
+    // returned tenant. So this uses a 2-slot pool: src holds slot 0 (kept checked out for the whole
+    // test) while a second reader R1 - opened at a newer txn and returned to the pool - holds slot 1.
+    // getCopyOf(src) lands on slot 1, sees R1.txn > src.txn, takes goActiveAtTxn()'s downgrade branch
+    // (init() -> openSymbolMaps()), and the injected fault makes the symbol-offset (.o) open throw. If
+    // refreshAt() were to close() on that failure it would returnToPool() the half-dead reader before
+    // get0() disposes it, letting a concurrent get() acquire a reader whose _txn mapping is already
+    // gone and dereference it - the "roTxMemBase is null" NPE. A correct pool keeps the slot owned by
+    // the disposing thread until it is fully gone, so the contender only ever sees
+    // EntryUnavailableException and never an NPE.
+    @Test
+    public void testConcurrentGetCopyOfRefreshFailureDoesNotHandOutDisposedReader() throws Exception {
+        final AtomicReference<TableReader> pooledReader = new AtomicReference<>();
+        final AtomicLong disposingThreadId = new AtomicLong(-1);
+        final AtomicBoolean faultSymbol = new AtomicBoolean();
+        final AtomicBoolean gateFired = new AtomicBoolean();
+        final AtomicReference<Throwable> refreshError = new AtomicReference<>();
+        final AtomicReference<Throwable> contenderError = new AtomicReference<>();
+        final CountDownLatch readerReleased = new CountDownLatch(1);
+        final CountDownLatch contenderDone = new CountDownLatch(1);
+
+        final TestFilesFacade ff = new TestFilesFacade() {
+            boolean isCalled = false;
+
+            @Override
+            public void munmap(long address, long size, int memoryTag) {
+                final TableReader r = pooledReader.get();
+                // The disposing thread frees _txn (nulling roTxMemBase) before it stops being open.
+                // The first munmap once isOpen() is false (the _cv unmap) is past that point: park
+                // here so the contender grabs a reader whose _txn mapping is already gone.
+                if (r != null
+                        && Thread.currentThread().threadId() == disposingThreadId.get()
+                        && !r.isOpen()
+                        && gateFired.compareAndSet(false, true)) {
+                    readerReleased.countDown();
+                    try {
+                        contenderDone.await(10, TimeUnit.SECONDS);
+                    } catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
+                    }
+                }
+                super.munmap(address, size, memoryTag);
+            }
+
+            @Override
+            public long openRO(LPSZ name) {
+                // Fault only the open of the symbol-offset (.o) file on the disposing thread, reached
+                // via init() -> openSymbolMaps() inside goActiveAtTxn()'s downgrade. exists() is left
+                // to pass so only a real file's open fails.
+                if (faultSymbol.get()
+                        && Thread.currentThread().threadId() == disposingThreadId.get()
+                        && Utf8s.endsWithAscii(name, ".o")) {
+                    isCalled = true;
+                    return -1;
+                }
+                return super.openRO(name);
+            }
+
+            @Override
+            public boolean wasCalled() {
+                return isCalled;
+            }
+        };
+
+        final CairoConfiguration poolConfig = new DefaultTestCairoConfiguration(root) {
+            @Override
+            public @NotNull FilesFacade getFilesFacade() {
+                return ff;
+            }
+
+            @Override
+            public int getPoolSegmentSize() {
+                return 2;
+            }
+
+            @Override
+            public int getReaderPoolMaxSegments() {
+                return 1;
+            }
+
+            @Override
+            public long getSpinLockTimeout() {
+                return 250;
+            }
+        };
+
+        assertWithPool(pool -> {
+            final String tableName = "copyOfRefreshFailure";
+            TableModel model = new TableModel(configuration, tableName, PartitionBy.DAY)
+                    .col("sym", ColumnType.SYMBOL)
+                    .timestamp("ts");
+            AbstractCairoTest.create(model);
+            final TableToken tableToken = engine.verifyTableName(tableName);
+
+            // Keep the table's scoreboard memory alive across the disposal so the contender's
+            // acquireTxn() reads a live scoreboard and fails on the null _txn, not a freed scoreboard.
+            final TxnScoreboard scoreboardHold = engine.getTxnScoreboardPool().getTxnScoreboard(tableToken);
+            TableReader src = null;
+            try {
+                // T1: one row in the first partition.
+                try (TableWriter w = newOffPoolWriter(configuration, tableName)) {
+                    TableWriter.Row r = w.newRow(1);
+                    r.putSym(0, "foo");
+                    r.append();
+                    w.commit();
+                }
+
+                // src opens at T1 and takes slot 0; keep it checked out for the whole test so the
+                // contender is forced onto slot 1 and never onto slot 0.
+                src = pool.get(tableToken);
+
+                // T2: a second row in a NEW partition, so a copyOf onto a T2 reader must downgrade.
+                try (TableWriter w = newOffPoolWriter(configuration, tableName)) {
+                    TableWriter.Row r = w.newRow(Micros.DAY_MICROS + 1);
+                    r.putSym(0, "bar");
+                    r.append();
+                    w.commit();
+                }
+
+                // R1 opens at T2 and takes slot 1, then returns to the pool (stays open, R1.txn == T2).
+                final TableReader r1 = pool.get(tableToken);
+                pooledReader.set(r1);
+                r1.close();
+
+                final Thread contender = new Thread(() -> {
+                    try {
+                        if (!readerReleased.await(10, TimeUnit.SECONDS)) {
+                            return;
+                        }
+                        final long deadline = configuration.getMillisecondClock().getTicks() + 2000;
+                        while (configuration.getMillisecondClock().getTicks() < deadline) {
+                            try (TableReader r = pool.get(tableToken)) {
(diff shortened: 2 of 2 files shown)


RaphDal:
# Review of PR #7251 — `chore(core): stop returning a reader to the pool when its refresh fails`
                                                                                                                                                                                                                        
**Reviewing at level 3** (full mission-critical pass: complete change-surface map, 5 targeted agents covering all mandated dimensions, per-finding source verification, plus an empirical build+test run in both
directions). Agents 3/8/11 (performance, Rust, adversarial-performance) are N/A by construction — the production change is a pure deletion with no hot path, no allocation surface, and no `.rs` files — so I folded their
checks inline rather than spawn redundant agents.
                                                                                                                                                                                                                        
## What the change does
                                                                                                                                                                                                                        
In `ReaderPool.R.refresh()` and `refreshAt()`, the failure-path `try { goActive(); } catch (Throwable ex) { close(); throw ex; }` is replaced with a bare `goActive();` / `goActiveAtTxn(srcReader);`. The deleted `close()`
was the bug: for a still-pooled reader it called `returnToPool()` (releasing `allocations[i]`) while `entry.getTenant(i)` still pointed at the open-but-being-disposed reader. A concurrent `get0()` could CAS-acquire that
slot, see the still-assigned tenant, call `refresh()→goActive()`, and dereference the already-freed `txFile` → `NullPointerException: roTxMemBase is null`, plus bookkeeping corruption that orphaned FDs. The fix makes
`get0()`'s catch the sole disposer, which keeps the slot owned by the failing thread for the entire teardown (`assignTenant(i, null)` runs *before* the slot is released).
                                                                                                                                                                                                                        
## Critical
                                                                                                                                                                                                                        
**None.** The fix is correct, minimal, and verified along every axis:
- **Sole caller / no orphaned contract** — `refresh`/`refreshAt` on `ReaderPool.R` are invoked only from `get0()` (`AbstractMultiTenantPool.java:314,316`); no reflection, no subclass override. Happy path is byte-for-byte│
unchanged (deleted code lived in a `catch`).
- **Slot ownership + ordering** — `get0()` catch (`AbstractMultiTenantPool.java:319-322`) holds the slot across `goodbye(); close(); assignTenant(i,null); release`, and `arrayPutOrdered` (release-store) publishes the
prior `tenants[i]=null` to any thread that later CAS-acquires (acquire). No residual interleaving in `releaseAll`/`notifyDropped`/`expelFromPool`/`lock` exposes a mid-dispose reader.
- **Supervisor accounting unchanged** — `onResourceReturned` still fires exactly once on the failure path (now in `get0()`'s `close()`), matching pre-change behavior; no false-positive/negative for the FD-leak tracker.
- **No leak on partial failure** — a throw part-way through `goActiveAtTxn→init→openSymbolMaps` leaves partially-opened resources reachable from the `columns`/`symbolMapReaders` lists; `super.close()` frees them
idempotently. The scoreboard `incrementTxn`/`releaseTxn` pair stays balanced.
- **Empirically validated** — both new tests pass on this branch and fail on reverted-`ReaderPool.java` with the exact `roTxMemBase is null` NPE.

## Moderate
                                                                                                                                                                                                                        
**1. (Out-of-diff, scope) `WalWriterPool.refresh()` retains the identical pattern; its safety is incidental, not structural.** `WalWriterPool.java:158-165` still has `try { goActive(); } catch { close(); throw; }`. It
avoids the race only because `WalWriter.goActive()` sets `distressed = true` inside its `catch (CairoException e)` (`WalWriter.java:425`), and the distressed branch of `WalWriterTenant.close()` disposes in place via
`expelFromPool()` (which clears the tenant before releasing the slot) instead of `returnToPool()`. Two agents independently confirmed this holds for the entire *realistically reachable* failure surface (all
`CairoException`s; the `AlterTableContextException` path is genuinely unreachable because WAL replay passes `contextAllowsAnyStructureChanges=true`). **But** `goActive()` only sets `distressed` for `CairoException` — an
unchecked `RuntimeException`/`Error` escaping deep inside `applyMetadataChangeLog` (e.g. from genuinely corrupt sequencer metadata) would skip it and reopen the same window. Note that
`applyStructural`/`applyNonStructural` deliberately use `catch (Throwable) + distressed=true`, whereas `applyMetadataChangeLog` catches only `CairoException`.
                                                                                                                                                                                                                        
The PR explicitly acknowledges and scopes this out, calling it defense-in-depth rather than a bug fix. That is a defensible judgment for a focused fix. **Not a blocker.** Optional follow-up: either apply the same 2-line
change to `WalWriterPool.refresh()` (and harmlessly to the no-op `ViewWalWriterPool.refresh()`) so all six pools are race-free *by construction*, or add a one-line comment at `WalWriterPool.refresh()` documenting that
its safety depends on `goActive()` marking the writer distressed on every throw.
                                                                                                                                                                                                                        
**2. PR title type contradicts the label and the release-notes convention.** The title is `chore(core): …` but the PR is labeled **Bug** and fixes a real concurrency defect (NPE + leaked FDs under concurrent reads). Per
CLAUDE.md the title description is copied to release notes; `chore` undersells a user-facing fix and may drop it from the fixes section. The description "stop returning a reader to the pool when its refresh fails" is
also implementation-focused, where the convention asks for end-user impact. Consider `fix(core): fix NullPointerException and file-descriptor leak under concurrent table reads`. (Flagging per conventions; final PR-type
call is yours.)
                                                                                                                                                                                                                        
## Minor
                                                                                                                                                                                                                        
- **Test boolean flag naming.** `faultSymbol` / `faultMeta` / `gateFired` (`ReaderPoolTest.java:390-391,579-580`) don't use the `is…`/`has…`/`should…` prefix from CLAUDE.md. Borderline — they mirror the existing
`AtomicBoolean`/`boolean isCalled` style in the file's `TestFilesFacade` subclasses.
- **Tests don't positively assert post-disposal serviceability.** On the *fixed* pool the contender never acquires within the parked window (it only ever sees `EntryUnavailableException`), so
`Assert.assertTrue(r.isOpen())` at `:508`/`:672` is effectively dead on the happy path. Consider a final `try (TableReader r = pool.get(token)) { Assert.assertTrue(r.isOpen()); }` after both threads join, to prove the
slot was cleanly reclaimed and the pool remains usable. Purely additive.
- **CI cost.** Each test deterministically spends ~2000ms because the disposer parks until the contender's full retry deadline. The 2000ms `deadline` could be trimmed (e.g. 500ms) without weakening the race window.
Optional.
                                                                                                                                                                                                                        
## Downgraded (considered and dismissed)
                                                                                                                                                                                                                        
- *"A `detached` reader hitting `refresh()` would skip disposal via `R.close()`'s detached branch."* — False positive: a detached reader still owns its slot, so `get0()`'s CAS can never re-enter `refresh()` on it. Cannot│
occur in old or new code.
- *"If `super.close()` throws inside `get0()`'s catch, the slot/tenant clear is skipped."* — Pre-existing property of the unchanged `get0()` catch; not introduced or worsened by this PR, and `super.close()` only does
null-safe `Misc.free` over validated handles.
                                                                                                                                                                                                                        
## Summary
                                                                                                                                                                                                                        
**Verdict: approve.** This is a correctly diagnosed, minimally scoped concurrency fix with strong, deterministic regression tests that I confirmed reproduce the bug on master (RED, exact NPE) and pass with the fix
(GREEN). No correctness, concurrency, resource-leak, or visibility issue survives verification.
                                                                                                                                                                                                                        
- **Regressions/tradeoffs:** none found. The change is strictly safer (it also closes latent `notifyDropped`/`releaseAll` double-dispose windows that shared the same released-but-still-assigned gap).
- **Draft findings:** ~6 distinct items verified, 2 dropped as false positives. Of the verified items, none are blocking — 2 moderate (1 out-of-diff scope note on `WalWriterPool`, 1 PR-title convention), 3
minor/optional.
- **In-diff vs out-of-diff:** the 3 minor findings are in-diff (test file); the production code (`ReaderPool.java`) is clean. The one substantive code observation is out-of-diff (`WalWriterPool`), reached via a full
sweep of all five sibling `PoolTenant` implementations — the cross-context pass did not underrun (the other four siblings — `Sequencer`, `TableReaderMetadata`, `SqlCompiler`, `ViewWalWriter` — are race-free because their│
`refresh()` never self-closes).
                                                                                                                                                                                                                        
The only thing genuinely worth a decision before merge is whether to fold the `WalWriterPool` hardening into this PR or leave it as the documented follow-up the description already proposes.

Note:
- Minor issues were addressed in 13b681c897498eaec843106372f3dcdb246ecae8
- Moderate issue 1: unreachable (see PR's body)
- Moderate issue 2: the issue wasn't found by a customer nor was it found during normal operation

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/pool/ReaderPool.java|2|2|100.00%|

bluestreak01 (approved):
Approve.

Reviewed at level 3 (full mission-critical pass): complete change-surface map, parallel fresh-context review agents (correctness, concurrency, resource management, cross-context caller impact, tests, conventions, plus a no-checklist adversarial pass), and per-finding source verification.

The fix is correct, minimal, and well-targeted. `refresh()`/`refreshAt()` are called only from `get0()`, whose `catch (Throwable)` already owns disposal, so the removed failure-path `close()` was redundant and was the bug: for a still-pooled reader it ran `returnToPool()`, republishing the slot (`allocations[i]=UNALLOCATED`) while `entry.getTenant(i)` still pointed at the open-but-being-disposed reader, letting a concurrent `get0()` refresh a reader whose `txFile`/`roTxMemBase` was already freed (the `roTxMemBase is null` NPE + orphaned FDs). Making `get0()` the sole disposer keeps the slot owned by the failing thread (`assignTenant(i,null)` before the release-store), closing the window. Verified: single caller, correct release-store/acquire ordering, supervisor `onResourceReturned` still fires exactly once, no partial-open leak or double-free, and the change incidentally hardens the notifyDropped/lock/releaseAll paths. The two regression tests are deterministic, leak-clean, hang-free, and reproduce on reverted code.

No blocking issues. Non-blocking notes:
- WalWriterPool.refresh() keeps the identical pattern but is safe in production (goActive() marks the writer distressed on every CairoException, routing failures through the race-free expelFromPool() path). The PR correctly scopes this out as defense-in-depth; optionally apply the same 2-line change or document the dependency.
- Title type `chore` is inconsistent with the `Bug` label and the user-visible NPE + FD leak fix; `fix(core):` with an end-user-facing description would surface it in release notes.

Recommend a full CI run before merge since it's a concurrency change.

PR #7253, opened 2026-06-13
fix(core): fix POSTING-indexed symbol queries returning wrong rows or crashing

bluestreak01:
## What

The backward POSTING index reader keeps a per-key cache of the
`(generation, position)` entries that a full index walk resolves, and replays
it on subsequent lookups of the same key. The cache-build guard keys off
`isEFMode`, but the "this generation holds no values for the key" early returns
in `loadSparseGenByPrefixSum` reset `isFlatMode` and the block count and leave
`isEFMode` untouched.

So when a walk passes through an Elias-Fano-encoded generation (which sets
`isEFMode`) and then a lower generation that does not hold the key — reachable
when that generation's split-block bloom filter false-positives the key, so it
is not skipped early — the stale `isEFMode` makes the guard cache a spurious
entry. The entry points at the lower generation but carries the position from
the EF generation.

## Impact

On the next lookup of the same key, replaying the spurious entry:

- returns row ids belonging to a different key (wrong query results), when the
  stale position is in range for the lower generation; or
- reads a file offset far past the mapped value file and crashes the process
  with a SIGSEGV in `readDeltaBlockMetadata`, when it is not.

This surfaced as a fuzz crash in `WalWriterFuzzTest.testWalWriteRollbackHeavy`:
a rollback-heavy WAL workload accumulates many small generations, and a
backward (`ORDER BY ts DESC`) scan over a POSTING-indexed symbol column trips
the leak. Which outcome occurs — wrong rows or crash — depends on the cached
position relative to the lower generation's key count.

The forward reader (`PostingIndexFwdReader`) is unaffected: its equivalent
early returns call `clearBlockState()`, which already resets `isEFMode`.

## Fix

Clear `isEFMode` at the start of each generation-load method
(`loadSparseGenByPrefixSum`, `loadSparseGenDirect`, `loadDenseGenerationCached`)
so a prior generation's mode cannot leak into the cache-add guard or a later
iteration. `readDeltaBlockMetadata` still sets the flag whenever the generation
actually has data, so correctly-encoded EF generations are unchanged.

## Tradeoffs and scope

- The change is three single-field assignments. It does not change behaviour
  for lookups that never hit the leak, and adds no measurable cost on the index
  read path.
- It does not touch the writer or the on-disk format. Indexes written before
  this change are read correctly; no rebuild is required.
- The forward reader is left as-is because it already resets the flag. Only the
  backward reader carried the partial reset.

## Test plan

- New `PostingIndexBwdCachePoisonTest` deterministically reproduces the leak: it
  builds two sparse generations (a 128-key generation that excludes the query
  key but bloom-false-positives it, found by replicating the writer's SBBF, and
  an EF-encoded generation holding only the query key), then asserts a cached
  replay returns the same rows as the initial build. Red before the fix
  (`[2002,2001,2000]` vs `[2002,2001,2000,1001,1000]`), green after.
- Regression suites pass: `PostingIndexOracleTest{,Ef,Delta}`,
  `CoveringIndexTest`, `PostingIndexConcurrencyTest`,
  `PostingIndexStressTest{,Ef,Delta}`.
- `WalWriterFuzzTest.testWalWriteRollbackHeavy` passes; the deterministic unit
  test is the primary proof since the fuzz seed varies per run.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/idx/PostingIndexBwdReader.java
+++ b/core/src/main/java/io/questdb/cairo/idx/PostingIndexBwdReader.java
@@ -463,6 +463,9 @@ public class PostingIndexBwdReader extends AbstractPostingIndexReader {
         private void loadDenseGenerationCached(int gen) {
             this.isCurrentGenDense = true;
             this.bufferRangeChecked = false;
+            // See loadSparseGenByPrefixSum: clear the EF flag so a previous
+            // gen's EF mode cannot leak past this gen's empty early returns.
+            this.isEFMode = false;
             long genFileOffset = genLookup.getGenFileOffset(gen);
             long genDataSize = genLookup.getGenDataSize(gen);
             int genKeyCount = genLookup.getGenKeyCount(gen);
@@ -617,6 +620,16 @@ public class PostingIndexBwdReader extends AbstractPostingIndexReader {
         private void loadSparseGenByPrefixSum(int gen) {
             this.isCurrentGenDense = false;
             this.bufferRangeChecked = false;
+            // Clear the EF flag up front so a previous gen's EF mode never
+            // survives into this gen's load. The early-return "no data for this
+            // key" paths below reset encodedBlockCount and isFlatMode but not
+            // isEFMode; without this, a stale isEFMode makes the build-mode
+            // cache-add guard (hasNext -> advanceToPrevRelevantGen) record a
+            // bogus (gen, posInGen) entry for a gen that has no values for the
+            // key. Replaying that entry then reads the wrong key's posting list
+            // (or an out-of-bounds offset). readDeltaBlockMetadata sets the flag
+            // correctly when this gen actually has data.
+            this.isEFMode = false;
             computePerColumnSidecarOffsets(gen);
             long genFileOffset = genLookup.getGenFileOffset(gen);
             long genDataSize = genLookup.getGenDataSize(gen);
@@ -683,6 +696,10 @@ public class PostingIndexBwdReader extends AbstractPostingIndexReader {
         private void loadSparseGenDirect(int gen, int idx) {
             this.isCurrentGenDense = false;
             this.bufferRangeChecked = false;
+            // See loadSparseGenByPrefixSum: clear the EF flag so a previous
+            // gen's EF mode cannot leak past this gen's empty (totalValueCount
+            // == 0) early return and make hasNext re-decode the prior gen.
+            this.isEFMode = false;
             computePerColumnSidecarOffsets(gen);
             long genFileOffset = genLookup.getGenFileOffset(gen);
             long genDataSize = genLookup.getGenDataSize(gen);
--- a/core/src/test/java/io/questdb/test/cairo/PostingIndexBwdCachePoisonTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/PostingIndexBwdCachePoisonTest.java
@@ -1,0 +1,185 @@
+/*******************************************************************************
+ *     ___                  _   ____  ____
+ *    / _ \ _   _  ___  ___| |_|  _ \| __ )
+ *   | | | | | | |/ _ \/ __| __| | | |  _ \
+ *   | |_| | |_| |  __/\__ \ |_| |_| | |_) |
+ *    \__\_\\__,_|\___||___/\__|____/|____/
+ *
+ *  Copyright (c) 2014-2019 Appsicle
+ *  Copyright (c) 2019-2026 QuestDB
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ ******************************************************************************/
+
+package io.questdb.test.cairo;
+
+import io.questdb.cairo.idx.PostingIndexBwdReader;
+import io.questdb.cairo.idx.PostingIndexUtils;
+import io.questdb.cairo.idx.PostingIndexWriter;
+import io.questdb.cairo.idx.SplitBlockBloomFilter;
+import io.questdb.cairo.sql.RowCursor;
+import io.questdb.std.LongList;
+import io.questdb.std.Misc;
+import io.questdb.std.Unsafe;
+import io.questdb.std.str.Path;
+import io.questdb.test.AbstractCairoTest;
+import org.junit.Assert;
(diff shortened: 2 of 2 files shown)


bluestreak01:
## Code review — level 3 (mission-critical pass)

Adversarial review across correctness, concurrency, resource management, performance, tests, cross-context caller impact, and a fresh-context pass. Every finding was verified against source; the bug was independently red-proven by reverting the fix.

**Verdict: approve.** This is a correct, complete, minimal fix for a genuinely mission-critical bug (wrong query results *and* a process-killing SIGSEGV), backed by a deterministic, leak-clean, red-proven regression test. The fix itself has no blocking issues. Two non-blocking findings below, one of which is pre-existing and unrelated to the change.

### What the fix does (verified correct & complete)

The backward reader keeps a per-key cache of `(gen, posInGen)` entries. The cache-add guard in `advanceToPrevRelevantGen` (`PostingIndexBwdReader.java:326`) is `if (encodedBlockCount > 0 || isFlatMode || isEFMode)`. The "no data for this key" early returns in `loadSparseGenByPrefixSum` reset `encodedBlockCount` and `isFlatMode` but historically **not** `isEFMode`. When a walk crosses an EF-encoded gen (sets `isEFMode`) then descends into a lower gen that SBBF-false-positives the key (so it isn't skipped early) but whose prefix sum reports the key absent, the stale `isEFMode` trips the guard → a bogus cache entry carrying a stale `posInGen` → on the next lookup, wrong rows (stale ordinal in range) or an out-of-bounds read → SIGSEGV (stale ordinal out of range). The fix clears `isEFMode` at the top of all three generation-load methods (lines 468, 632, 702).

Completeness verified directly: the three guard flags are each provably reset on every empty exit now; clearing at method top is sufficient because the only writer of `isEFMode = true` (`readDeltaBlockMetadata:777`) runs strictly after every early return, and no helper called before an early return touches the flag. The forward reader's "unaffected" claim checks out — its early returns funnel through `clearBlockState()` (which already resets `isEFMode`); that asymmetry (FWD uses `clearBlockState()`, BWD hand-rolls the resets and missed one flag) is the entire root cause.

**Real-SQL reachability confirmed** (not just the white-box test): `TableReader.getIndexReader` returns a cached, reused reader whose `genLookup` cache persists across `getCursor` calls; a single partition spanning more than one page frame issues `getCursor(symbolKey)` two or more times on that one reader within a single pass, so `WHERE sym='x' ORDER BY ts DESC` (and `LATEST ON` re-reads) hits the replay path. EF is the production default (ADAPTIVE picks EF when it encodes smaller). So this fixes a bug reachable by ordinary descending indexed queries on multi-generation indexes.

### Critical

None. (The PR fixes a Critical-severity bug; the fix introduces none.)

### Moderate

**M1 — Pre-existing O(rank) sidecar-base linear scan on the covering-index hot path. Not introduced by this PR; optional follow-up.**

`PostingIndexBwdReader.java:683-691` (`loadSparseGenByPrefixSum`) and the identical `:730-738` (`loadSparseGenDirect`):

```java
if (coverCount > 0) {
    int sidecarBase = 0;
    for (int i = 0; i < start; i++) {                 // start = active-key rank of requestedKey within the gen
        sidecarBase += Unsafe.getInt(countsBase + (long) i * Integer.BYTES);
    }
    this.sidecarOrdinal = sidecarBase + totalValueCount;
}
```

`start` is the requested key's active-key ordinal within the sparse gen, bounded only by the gen's active-key count — for a high-cardinality symbol it can be large. The scan is O(start) per relevant sparse gen, runs only for covering indexes (`coverCount > 0`), and runs on both the cache-build and the cache-replay paths — so the per-key cache amortizes gen *location* but not this sidecar-base cost, and a DESC covering scan re-pays it per page frame per key. Optimal is O(1): either a persisted cumulative-value-count prefix array (the writer already walks `countsBase` at flush time, so emitting a running sum is nearly free) or, without a format change, memoizing the base per gen across replay. The dense analogue at `:597-601` is correctly bounded to O(stride)=O(256) and is fine. This is pre-existing — the diff neither introduced nor worsened it — and should not block this PR; worth a separate optimization ticket.

### Minor

**m1 — `isFlatMode` reset asymmetry in `loadDenseGenerationCached` (benign latent trap).** The fix hoists `isEFMode = false` to line 468 (before all early returns), but `isFlatMode = false` remains at line 484, *after* the first early return (`requestedKey >= genKeyCount`, lines 473-477). That one early return can leave `isFlatMode` stale. It is benign today: dense loads return via `advanceToPrevRelevantGen:299` and never reach the cache-add guard at 326, and `isFlatMode`'s only reader (`hasNext:265`) is gated by `flatRemaining`, which is provably 0 at advance time. For parity and future-proofing, hoist `this.isFlatMode = false;` next to the `isEFMode` reset at 468 (or adopt the forward reader's `clearBlockState()`-in-early-returns discipline so the flags stay in lockstep).

**m2 — Optional coverage nice-to-haves (non-blocking).** The test is strong (deterministic, pins exact rows `[2002,2001,2000]` rather than just `pass1 == pass2`, red-proven). Two optional additions: (a) a deterministic SQL-level `assertQuery("... ORDER BY ts DESC").returns(...)` end-to-end test, since the cited `WalWriterFuzzTest.testWalWriteRollbackHeavy` uses an unseeded random and only probabilistically reproduces; (b) a one-line comment documenting the FWD/BWD `clearBlockState` asymmetry that makes the forward reader immune. Leaving the crash variant (SIGSEGV) untested is reasonable — a JVM-aborting test is inherently fragile and the wrong-rows variant exercises the identical root cause.

### Standards check (clean)

The new test asserts on drained `RowCursor` output via raw JUnit `Assert.assertEquals` — it is a white-box reader unit test, not an `assertQuery` SQL-result test, so the builder-API rules do not apply. ASCII-only comments, no banner comments, correct member ordering (static fields → `@Test` → alphabetical private helpers), `assertMemoryLeak`-wrapped, balanced native alloc/free in `findBloomFalsePositiveKey` (captured before `try`, freed in `finally`), pooled-cursor native buffers reclaimed at `reader.close()`.

### Summary

- 0 out-of-diff *broken* callsites. The cross-context pass ran wide (`TableReader` reader reuse, the Bwd/Fwd page-frame cursor drivers, and the consumer factories: `SymbolIndexRowCursorFactory`, `LatestByValue(s)IndexedFiltered`, `CoveringIndexRecordCursorFactory`, `PostingIndexDistinct`, `LatestByAllIndexed`) and found every callsite safe after the fix, with production reachability proven — the zero count is genuine, because this change makes previously-broken paths correct.
- No regressions or tradeoffs. Three boolean stores on a cold gen-load path are perf-neutral. No on-disk format change; indexes written before the fix read correctly.
- One pre-existing optimization (M1) and one cosmetic parity tidy (m1) are worth tracking but neither blocks merge.

*Reviewed at level 3 with multi-agent analysis; findings verified against source.*

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexBwdReader.java|3|3|100.00%|

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexBwdReader.java|3|3|100.00%|

PR #7255, opened 2026-06-13
perf(core): speed up checkpoint/snapshot restore for tables with many parquet partitions

bluestreak01:
## Impact

This PR makes two independent improvements to checkpoint/snapshot restore.

Tables with many parquet partitions restore faster: each partition's
parquet-metadata sidecar (`_pm`) is now mapped and checksum-verified once instead
of up to twice mapped / three times verified, and the previously serial
per-partition validation now runs in parallel across the recovery pool. This
benefit scales with parquet-partition count and is largest on cold, object, or
network storage, where the old serial per-partition I/O dominated restore
startup.

Tables with few partitions but several indexed symbol columns rebuild their
bitmap indexes faster: the native (non-parquet) index rebuild now parallelizes
across (partition, column) work items rather than whole partitions, so a
non-partitioned table no longer serializes all of its indexed columns onto one
thread. Tables with neither many parquet partitions nor multiple indexed symbol
columns are unaffected.

No end-to-end wall-clock benchmark was measured. The improvements are
characterized below by the per-partition reduction in map/checksum operations,
the removal of the O(N) serial stat/validation loop, and the restored
column-level parallelism on the native index path.

## What

On the snapshot/checkpoint restore path, the `_pm` parquet-metadata sidecar of every parquet partition was mapped and full-CRC-verified more than once.

On the index-rebuild path (`rebuildPartitionColumnIndexes=true`, the enterprise-restore default) each `_pm` was mapped twice and CRC-verified three times per partition:

1. serial `generateMissingParquetMetaFiles` → `isParquetMetaResolvable`: `openAndMapRO` + `resolveFooter` (CRC) + `munmap`
2. parallel rebuild task: `openAndMapRO` + `resolveFooter` (CRC) on a fresh per-task reader
3. `ParquetPartitionDecoder.of(long,long,...)` inside that task: re-binds the same mapping and runs `resolveFooter` (CRC) a third time (its internal `ParquetMetaFileReader.of(long,long)` clears the `checksumVerified` flag, forcing a re-verify)

The serial validation pass (1) was introduced recently (#7243), so for the ~100k-partition tables this targets, restore paid ~100k redundant serial map/CRC/unmap cycles at startup. On the checkpoint-recovery path (`rebuildPartitionColumnIndexes=false`, the OSS default) the serial pass was the only validation but ran single-threaded over every partition.

## Change

Collapse the work to a single `_pm` map + single footer resolve (single CRC) per parquet partition, fully parallel on both paths:

- Decoder reuse: feed the decoder via `ParquetPartitionDecoder.of(ParquetMetaFileReader, ...)`, which shallow-copies the already-resolved + CRC-verified reader state instead of re-resolving the same mapping (eliminates CRC #3).
- Bounded worker pool, one `_pm` map per partition: `prepareParquetPartitions` submits one worker per recovery-pool thread (`workerCount = min(threadCount, partitionCount)`), each pulling parquet partitions from a shared cursor (dynamic load balancing) and reusing a single set of native scratch objects (`ParquetMetaFileReader` / decoder / `RowGroupBuffers` / `DirectIntList` / index writers) across every partition it handles, with per-partition resets between partitions. `processParquetPartition` maps `_pm` once, validates it, regenerates from `data.parquet` when missing/stale/torn/zero-length, and — when indexes are requested and the partition has rows — reuses that same mapping to rebuild bitmap indexes. The committed-size truncation check on `data.parquet` runs inside the worker (before regeneration), so the calling thread issues no per-partition stats — removing the O(N) serial `ff.length()` loop that otherwise becomes the Amdahl tail on cold/object/network storage. This also parallelizes the previously-serial checkpoint-recovery validation.
- Shared regeneration: extracted `regenerateParquetMetaFile` and `mapResolvableParquetMeta` so the missing/stale/torn/zero-length sidecar paths share one implementation; removed the now-unused serial `parquetMetaReader` field.

### Cost per parquet partition

| Path | Before | After |
|---|---|---|
| `rebuild=true` (enterprise default) | 2 maps + 3 CRCs, serial validation | 1 map + 1 CRC, parallel |
| `rebuild=false` (checkpoint default) | 1 map + 1 CRC, serial loop | 1 map + 1 CRC, parallel |

## Native bitmap-index rebuild: per-column parallelism

The non-parquet (native) bitmap-index rebuild distributed a whole partition to
each worker (`workerCount = min(threadCount, partitionCount)`) and rebuilt that
partition's indexed columns serially on a single thread. For non-partitioned or
low-partition-count tables with several indexed symbol columns, effective
parallelism was capped at `partitionCount` rather than
`partitionCount x indexedColumns` -- a non-partitioned table rebuilt every
indexed column on one thread, regardless of pool size.

The work is now flattened into a single `(partitionIndex, colIdx)` list built on
the calling thread (skipping empty partitions, parquet partitions, and columns
absent from a partition). Workers pull individual items from a shared cursor and
reuse one `Path` each, so the per-worker `Path`/scratch amortization that
motivated this PR is preserved (one `Path` per worker, not one per item) while
column-level parallelism is restored. Worker count is keyed on the work-item
count, matching the symbol-rebuild phase, which already used per-column
granularity.

The list is partition-major, so a worker pulling consecutive items usually stays
within one partition; it caches the last partition's resolved
timestamp/rowcount/nameTxn and only re-resolves on a partition change, avoiding
redundant `txWriter` lookups. The many-parquet-partition target is unaffected:
that path is handled by `prepareParquetPartitions`, and on the native path a high
partition count already saturates the pool.

## Behavior change: fail-fast on a truncated capture is traded for parallelism

Worth calling out explicitly, because it is a real semantic change rather than a pure refactor.

Previously the serial `generateMissingParquetMetaFiles` ran before the symbol and native-bitmap rebuild phases were submitted, so a short/missing `data.parquet` aborted the restore before any sibling work started. The committed-size check now runs inside the parquet workers, so a truncated capture is detected at the single `finalizeParallelTasks()` drain rather than strictly ahead of the other phases.

This is bounded and safe, not a correctness regression:

- The path-bearing diagnostic and the abort are preserved — the restore still fails on a truncated capture, naming the offending partition.
- The first failing worker trips the shared abort latch that every phase's worker loop polls, so sibling workers across all three phases short-circuit at their next item boundary instead of running their whole queue.
- The per-partition committed-size check still guards regeneration, so a truncated `data.parquet` is never fed to `ParquetMetadataWriter.generate`.

Net: on a truncated capture the extra cost is the in-flight items already running when the latch trips — wasted I/O on an already-doomed restore, not corruption. Strict pre-submission fail-fast and cross-phase parallelism are mutually exclusive; this PR chooses parallelism for a path that runs once per restore and only does the extra work on bad input headed for operator intervention anyway.

## Preserved behavior

- Durability: fsync of regenerated `_pm` + partition-dir fsync, partial-write cleanup on failure.
- Path-bearing truncation diagnostics for short/missing `data.parquet`, with the restore still aborting (see the fail-fast note above for the change in *when* it surfaces).
- Self-healing of stale/missing/torn/zero-length sidecars on both restore paths (including `rebuild=false`).
- Abort/drain machinery and cross-table reader-reuse safety.
- No public API or call-signature changes (all edits are private methods/fields), so enterprise restore is unaffected at the boundary.

## Tests

All green:

- `CheckpointTest` — 108/108, including a new deterministic test, `testCheckpointRestoreRebuildsBitmapIndexesAcrossParquetPartitionsReusingWorkerScratch`, that pins the recovery pool to 2 workers and rebuilds 5 all-parquet partitions so at least one worker reuses its scratch across ≥2 partitions (pigeonhole) — the cross-partition reuse path, which the default 4-thread floor never exercises with ≤2 partitions. Each partition carries a distinct indexed-symbol distribution, and the test asserts per-partition `WHERE sym = …` results plus per-partition index-file rebuild, so any state leaking from one partition's rebuild into the next fails. The existing arms still cover stale, stale-without-index-rebuild (`rebuild=false`), truncated, zero-length, committed-size, cannot-create-`_pm`, rejects-truncated-parquet, missing-parquet-with-`_pm`-present, and the drain/interrupt/multi-table cases.
- `PostingSealPurgeTest` — 17/17
- `PostingIndexCriticalIssuesTest` — 95/95
- `CoveringIndexTest` — 398/398
- `IndexBuilderTest` — 23/23
- `CheckpointFuzzTest` — 3/3 (restore over fuzzed data shapes; re-run after the native per-column change)
- `CheckpointWindowsTest` — skipped (Windows-gated)

`assertMemoryLeak` wrappers passing confirms the mmap/munmap and reader lifecycles stay balanced.


bluestreak01:
## Review of PR #7255 — single-map `_pm` + parallel restore

**Level 3 review** (all review passes — correctness, concurrency, performance/algorithmic optimality, resource management, tests, code quality, cross-context caller impact, plus two fresh-context adversarial passes — with per-finding verification against source). Reviewed the working tree of `perf/snapshot-restore-pm-single-map`. Diff is self-contained in `TableSnapshotRestore.java` and `CheckpointTest.java`.

Every candidate bug from the first pass (double-munmap on the `openAndMapRO` throw path, the 3-phase concurrency hazard, cross-partition scratch state-bleed, abort-latch error reordering, the `assertMemoryLeak`+`assertQuery` "anti-pattern") was traced to source and **disproved**. No correctness, concurrency, or resource bug survived verification.

### Critical
**None.** The three headline mechanisms are verified sound:
- **Single `_pm` map + CRC**: `ParquetPartitionDecoder.of(ParquetMetaFileReader,…)` shallow-copies the already-resolved/CRC-verified reader (`checksumVerified` carried over), so the footer is resolved and CRC'd exactly once (was up to 2 maps + 3 CRCs).
- **mmap/fd lifecycle on all paths** (happy, both early returns, every throw): clear-then-munmap ordering is honored for both the worker `metaReader` and the decoder's internal reader; the suspected double-munmap when `openAndMapRO` throws `size>file` inside the regeneration revalidation block leaks nothing and double-frees nothing (`parquetMetaAddr` stays 0, `openAndMapRO` self-cleans).
- **Concurrency of the 3 now-overlapping phases**: every `txWriter`/`columnVersionReader`/`tableMetadata` getter read concurrently is a pure `binarySearchBlock`/`getQuick`/immutable read (no mutable cache); the three phases write disjoint files (parquet dirs vs native dirs vs table-root symbol files); per-worker scratch is never shared; the abort-latch trip on failure cannot skip work on a successful restore nor reorder the first reported error.

### Moderate
**M1 — New private methods are not in alphabetical order (enforced convention).** The instance-method block was alphabetical before; this breaks it:
- `regenerateParquetMetaFile` (line 822) precedes `mapResolvableParquetMeta` (876) — `r` before `m`.
- `mapResolvableParquetMeta` (876) sits after `prepareParquetPartitions`/`processParquetPartitions` — `m` should precede `p`.
- `processParquetPartition` (1122) is stranded inside the `rebuild*` block, far from `processParquetPartitions` (740).

(The new `threadCount` field *is* correctly placed.) Suggested order: `…copyFile, mapResolvableParquetMeta, prepareParquetPartitions, processParquetPartition, processParquetPartitions, rebuildBitmapIndexes…, rebuildSymbolFiles…, regenerateParquetMetaFile, removePartitionDirsNotAttached, submitParallelTask`.

**M2 — Test coverage gaps on the new branches.** The new `…ReusingWorkerScratch` test is well-built (the pigeonhole over 2 pinned workers × 5 parquet partitions is valid; distinct per-partition A/B/C distributions catch state bleed; builder-API compliant). But it only covers **`rebuildIndexes=true`**, whereas the OSS checkpoint-recovery default is **`rebuildIndexes=false`** (`PropServerConfiguration:1025`). Untested new paths:
- cross-partition scratch reuse on the validation-only (`rebuild=false`) path (null rebuild buffers, relying on `metaReader.clear()` between partitions);
- `_pm` **regeneration** interleaved with reuse in one worker (the test keeps every `_pm` intact, so the regenerate branch never runs inside a reuse worker);
- native per-column work-list reuse with ≥2 items per worker (the `cachedPartitionIndex` hit/miss branches — only 1 native item in the test).

All reachable with the same pinned-pool technique. Verified safe by inspection, but unguarded.

**M3 — Per-worker amortization stops at `Path`; the heavy native-backed objects stay per-item** (pre-existing pattern, consciously kept). `rebuildBitmapIndexForNativePartitionColumn` does `new SymbolColumnIndexer(...)` per (partition,column), each a **4 MB `Unsafe.malloc`** (`SymbolColumnIndexer:60-61`) + free; `rebuildSymbolFilesForColumns` does `new SymbolMapUtil()` per symbol column though its own class doc says the instance is reusable. The rewrite amortizes the cheap `Path` per worker but reallocates these per item. Caveat: `SymbolColumnIndexer`'s index type is fixed at construction, so a per-worker hoist needs a small per-`indexType` cache, not one instance. Worth addressing since amortization is the PR's thesis; not a regression.

**M4 — `perf`-typed PR with no measured wall-clock benchmark.** Win characterized only by op-count reduction with an explicit disclaimer. The mechanistic argument is sound and the disclaimer is honest, but the `perf` claim has no figure behind it — one synthetic restore on a high-parquet-partition table would close this.

### Minor
- **m1 — By-timestamp binary searches where the raw index is in hand.** `getPartitionRowCountByTimestamp(ts)` / `getPartitionNameTxnByPartitionTimestamp(ts)` (lines 769, 980-981, 1277) do O(log P) searches returning the identical masked value as the O(1) `getPartitionSize(i)` / `getPartitionNameTxn(i)`. Worse, `partitionRowCount` at line 769 is computed unconditionally but unused when `rebuildIndexes=false` (the OSS default) — a wasted search per parquet partition on the path being optimized.
- **m2 — `columnTop` computed twice** (calling-thread filter `:1297`, then again in the worker `:993`); a parallel `LongList` aligned to `nativeIndexWork` would carry it forward.
- **m3 — Partition-metadata cache largely ineffective**: the shared cursor interleaves workers ~`workerCount` apart, so with 1–3 indexed columns/partition the cache mostly misses. Fixing m1 to O(1) getters makes it unnecessary.
- **m4 — Redundant `ff.exists(data.parquet)` at line 1192** after `ff.length(...)` at 1157 already proved existence.
- **m5 — `new StringSink()` + `new long[]` per partition** in `rebuildParquetPartitionIndexes` (1620, 1701) not folded into worker scratch (pre-existing; static method shared with O3).
- **m6 — No LPT ordering.** Cursors pull in index order, not size order; one worker can run the largest parquet partition alone in the tail. `getPartitionParquetFileSize(i)` is O(1), so a largest-first scan would shrink makespan for size-skewed tables.
- **m7 — Stale doc comments** in `CheckpointTest.java` (≈lines 1006, 1163, 2343, 2493, 2626) still name the removed `generateMissingParquetMetaFiles`.
- **m8 — Boolean params lack `is…`/`has…` prefix** (`rebuildIndexes`, `doRebuild`, `regenerate`, `resolved`) — though `rebuildIndexes` mirrors the unchanged public `rebuildPartitionColumnIndexes`.
- **m9 — Diagnostic non-determinism** (documented behavior change): a truncated capture's reported partition is now the first *failing future in submission order*, not the lowest partition index. Still path-bearing; restore still aborts.
- **m10 — PR metadata polish**: the title scopes to "many parquet partitions" but the second improvement (native per-column parallelism, benefiting few-partition/many-indexed-column tables) is buried; optional `storage` label.

### Downgraded (investigated, dismissed after source verification)
- **Double-munmap / leak when `openAndMapRO` throws in `mapResolvableParquetMeta`** → throw happens before any mapping; caller's `parquetMetaAddr` stays 0; outer finally skips munmap.
- **Concurrency hazard from 3 phases overlapping** → all concurrent reads are pure; phases write disjoint files; old code already ran symbol+bitmap concurrently.
- **Cross-partition scratch state bleed** → `openAndMapRO`→`clear()`, `decoder.of`→`destroy()`, and `rebuildParquetPartitionIndexes` `clear()`s reset every reused object per partition; the new test exercises this.
- **Abort-latch reordering the reported error** → a latch-bailing worker returns with nothing to throw, so the earliest-submitted thrower is still reported.
- **`assertMemoryLeak` wrapping `assertQuery`** → the wrapper holds many statements, so the outer scope is the correct idiom and inner `.noLeakCheck()` is appropriate. All assertions use `.returns(...)`, none use `.returnsOnce(...)` or the removed `assertSql`.
- **`indexWriters` ObjList not freed in worker outer-finally** → holds only Java refs and is always empty there (`rebuildParquetPartitionIndexes` frees+clears its native contents every call).

### Summary
**Approve with minor changes.** Careful, well-documented refactor; correctness, concurrency, and native-resource handling hold up under adversarial tracing, and the sole production caller (`DatabaseCheckpointAgent`) is verified SAFE across the per-table drain, error-handling, and fail-fast-residue dimensions. The one thing worth fixing before merge is **M1 (alphabetical member ordering)**. **M2** (test the `rebuild=false` reuse path — the OSS default) and **M3** (amortize the 4 MB indexer / `SymbolMapUtil` per worker) are worth doing but non-blocking.

Honest tradeoffs: the **fail-fast timing change** (truncated capture surfaces at the drain, not before sibling work) is real but safe — restore still aborts, the truncated file is never fed to regeneration, partial writes live only in a doomed/retried table dir; and the **perf claim is unmeasured** (M4).

Findings: ~16 verified (4 Moderate, ~12 Minor); 6 candidate bugs dropped as false positives after verification. In-diff: all findings; out-of-diff: 0 — justified because every edited symbol is private, all public signatures are unchanged, and the production caller was verified SAFE.

<sub>🤖 Generated with an automated level-3 review.</sub>

bluestreak01:
Addressed the partition-metadata follow-ups in `TableSnapshotRestore` (commit 8951d99192). All changes are value-preserving; `CheckpointTest` (108) and `CheckpointFuzzTest` (3) pass.

- **m1 — O(log P) by-timestamp lookups where the index is in hand:** switched to the O(1) by-index getters `getPartitionSize(i)` / `getPartitionNameTxn(i)` at the 3 sites (parquet worker loop, native-rebuild worker, native work-list build). The by-timestamp variants ran a binary search to return the identical masked value; on the validation-only (`rebuildIndexes=false`) default path the parquet-loop lookup was also wasted entirely. Left the directory-scan lookup as by-timestamp since it parses the timestamp from a dir name (no index available) and relies on the not-found semantics.

- **m5 — per-partition allocations in `rebuildParquetPartitionIndexes`:** folded the `new StringSink()` and `new long[]` into caller-owned per-worker scratch, reused across every partition a worker handles (mirroring the existing `rowGroupBuffers`/`parquetColumns`/`indexWriters` scratch). `columnTops` is now a reusable `LongList`.

- **m3 — partition-metadata cache largely ineffective:** removed the per-worker `cachedPartitionIndex` cache in `rebuildBitmapIndexesForNativePartitions`. The shared cursor disperses a worker's items ~workerCount apart in the partition-major list, so it mostly missed for the typical 1–3 indexed columns/partition, and the getters are O(1) after m1. Now resolves metadata per item.

- **m2 — `columnTop` computed twice:** the calling-thread filter already computes `columnTop` per (partition, column); now carried forward via a `LongList` index-aligned with `nativeIndexWork` instead of re-running `getColumnTop` (an O(log V) search) per item in the workers.

bluestreak01:
## Review of PR #7255 — single-map `_pm` + parallel restore

**Level 3 review** (correctness, concurrency, performance/algorithmic optimality, resource management, tests, code quality, cross-context caller impact, plus two fresh-context adversarial passes — with per-finding source verification). Reviewed the working tree at HEAD `6942e0e3`, including the two follow-up commits (`8951d99` O(1)-getter rewrite + `6942e0e` test hardening). Diff is self-contained in `TableSnapshotRestore.java` and `CheckpointTest.java`.

Every candidate correctness/concurrency/resource bug was traced to source and **disproved**. No blocker survived verification.

### Critical
None.

### Moderate
None. (The performance items below are real but are once-per-restore, `threadCount`-bounded, or pre-existing — none rises to Moderate.)

### Minor
- **m1 — `rebuildParquetPartitionIndexes` makes 3 passes over `columnCount` where 1 suffices (pre-existing, not introduced).** Each pass calls `getIndexedParquetColumnIndex` (which does an O(P_cols) linear `getColumnId` scan + an O(log V) `getColumnTop`); pass 3 (`:1735-1744`) re-calls `getColumnTop` a 4th time per indexed column. Pass 2 already holds `writerIndex` and can `columnTops.add(...)` inline, deleting pass 3 entirely (3–4× fewer metadata reads/binary searches per parquet partition). The PR only changed `columnTops` from `long[]` to a reused `LongList` here, so this is a pre-existing inefficiency surfaced by the review — a follow-up, not a regression.
- **m2 — Parquet/symbol worker count is sized off the wrong denominator.** `prepareParquetPartitions` uses `workerCount = min(threadCount, partitionCount)` (`:712`) and `rebuildSymbolFiles` uses `min(threadCount, columnCount)` (`:1387`), not the parquet-partition / symbol-column counts. A 100k-native / few-parquet table spawns up to `threadCount` parquet workers that mostly drain-and-`continue`, and — when `rebuildIndexes=true` — each still allocates `RowGroupBuffers`/`DirectIntList`/`ParquetPartitionDecoder` scratch it never uses. Bounded by `threadCount` and once-per-restore, so low cost, but the phase could be skipped when no parquet partition exists.
- **m3 — No size-ordered (LPT) scheduling.** The shared cursor pulls partitions/items in index order, so the largest parquet partition can land last and run alone in the makespan tail (common in time-series, where the newest partition is largest and last). `getPartitionParquetFileSize(i)` is O(1); a descending-size pre-sort gives the classic 4/3-OPT makespan bound at negligible cost.
- **m4 — `getColumnNameTxn` not precomputed in the native-bitmap phase (asymmetric with `columnTop`).** `rebuildBitmapIndexesForNativePartitions` precomputes `columnTop` into an index-aligned list (good) but still calls `columnVersionReader.getColumnNameTxn(partitionTimestamp, writerIndex)` per item (O(log V)) inside the worker, though it depends only on values known when the work list is built.
- **m5 — Zero-row parquet partition is untested.** `doRebuild = rebuildIndexes && partitionRowCount > 0` (`:781`) gates the "valid `_pm`, no index rebuild" branch; no test constructs a 0-row parquet partition. Low risk (hard to construct via `CONVERT TO PARQUET`), optional.
- **m6 — Member ordering.** New private methods follow call-flow grouping (`prepareParquetPartitions` → `processParquetPartitions` → `regenerateParquetMetaFile` → `mapResolvableParquetMeta` → …) rather than the alphabetical-within-visibility convention. Instance fields *are* correctly alphabetical (the `threadCount` insertion preserved order). Minor style nit.
- **m7 — Commit messages use Conventional-Commit prefixes** (`perf(core): …`, `test(core): …`); the convention reserves that prefix for the PR title and asks for plain-English commit titles. PR title/labels are correct.
- **m8 — PR title scope.** The title scopes to "many parquet partitions"; the second improvement (restored column-level parallelism on the native index path, benefiting few-partition/many-indexed-column tables) is understated. Optional.

### Process note
- Enterprise **coverage check passes** (215/246, 87.40%), which also confirms enterprise compiles against the changed `static rebuildParquetPartitionIndexes(...)` signature. If the umbrella "Enterprise CI" status is still red, confirm the failing sub-check is unrelated before merge.

### Downgraded (investigated, dismissed after source verification)
- **Double-munmap / use-after-unmap on the reused decoder + shared `metaReader`** → `decoder.of(reader)` shallow-copies into the decoder's own `ParquetMetaFileReader` with its own lazy native handle; `clear()` never munmaps; teardown is `decoder.close()`→munmap(parquetAddr) (inner finally) then `metaReader.clear()`→munmap(parquetMetaAddr) (outer finally). Each mapping is unmapped exactly once; clear-then-munmap honored on every arm of `mapResolvableParquetMeta` (success/stale/regenerate-fail).
- **3-phase overlap concurrency hazard** → write domains are disjoint (parquet partition dirs / table-root symbol map / native partition dirs); symbol⇄bitmap overlap already existed pre-PR; all shared getters (`AbstractRecordMetadata`/`TableReaderMetadata`/`TxReader`/`ColumnVersionReader`) are pure reads with no lazy mutation.
- **Abort latch causing a silent missing-index "success"** → impossible: `submitParallelTask` only sets the latch *and rethrows*; `finalizeParallelTasks` always rethrows when any future failed. A latch-skipping sibling co-occurs with a reported failure; in-flight items finish at a clean boundary (no torn index from the abort itself).
- **Cross-partition scratch state bleed** → `metaReader.clear()`, `decoder.of()/close()`, and `columnNamesSink/columnTops/parquetColumns/indexWriters.clear()` reset every reused object per partition; exercised by the two new pigeonhole tests.
- **`getPartitionSize(i)` vs `getPartitionRowCountByTimestamp(ts)` divergence** → both read `attachedPartitions[idx + PARTITION_MASKED_SIZE_OFFSET] & PARTITION_SIZE_MASK`; the O(1)-by-index swap is value-preserving.
- **Pigeonhole test non-determinism / drains-test flake** → sound: on a successful restore the cursor dispenses all N items exactly once across 2 pinned workers (⌈N/2⌉ ≥ 3 or ≥ 2), so the reuse path runs in every interleaving; the column reorder makes `sym_slow` work-item index 0 (grabbed before the latch can trip), and the removed early abort-check means the column opens unconditionally once grabbed.

### Summary
**Approve.** Careful, exceptionally well-documented refactor. Correctness, concurrency, native-resource handling, and the fail-fast timing tradeoff hold up under adversarial tracing and per-finding source verification; the sole production caller (`DatabaseCheckpointAgent`, serial-per-table with a draining `finalizeParallelTasks` before each table reload) is verified SAFE, and the two changed/new signatures (`rebuildParquetPartitionIndexes`, `decoder.of(ParquetMetaFileReader,…)`) each have exactly one verified in-diff caller. The follow-up commits' O(1)-getter rewrite and index-aligned `columnTops` list are value-preserving.

- **Findings:** 8 verified (all Minor); 6 candidate bugs investigated and dropped as false positives after source verification.
- **In-diff vs out-of-diff:** all findings in-diff; **0 confirmed out-of-diff bugs** — justified because the diff is small and self-contained, public signatures are unchanged, and the cross-context pass walked the single caller of each changed/new symbol and verified both SAFE.
- **Honest tradeoffs:** the documented fail-fast timing change (truncated capture surfaces at the drain, after sibling work may have rebuilt valid unrelated files) is real but safe — restore still aborts, the truncated file is never fed to `ParquetMetadataWriter.generate`, and partial writes live only in a doomed/retried table dir; the perf claim is reasoned but unmeasured, and the parquet phase spends `threadCount` workers/scratch even when few/zero parquet partitions exist (m2).

Nothing here blocks merge. m1 (fuse the 3-pass parquet rebuild) and m2/m3 (size workers to parquet count + LPT ordering) are the highest-value optional follow-ups.

<sub>🤖 Generated with an automated level-3 review.</sub>

bluestreak01:
Tightened the comments and javadoc in `TableSnapshotRestore` and `CheckpointTest`.

The originals restated context, narrated the former behavior, and spelled out step-by-step proofs that added bulk without adding information. This pass keeps the rationale behind each non-obvious decision (use-after-free risk on the shared readers, abort-latch semantics, mapping `_pm` once, the pigeonhole reasoning in the worker-scratch-reuse tests) and removes the repetition around it.

Comments only; no executable code changed. A diff filter confirmed every changed line is a comment or javadoc line. Net: ~136 fewer lines (225 insertions / 361 deletions).

bluestreak01:
@RaphDal done making comments more concise

RaphDal (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 206 / 235 (87.66%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableSnapshotRestore.java|206|235|87.66%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=244917)
**Tested Commit:** `50d09ee082fa31a7ee0625aa0b37cdc9c40a569e`

PR #7256, opened 2026-06-14
fix(core): fix NATIVE_INDEX_READER leak when a posting-index row cursor is closed after its reader

bluestreak01:
## What

Fixes a native-memory leak (`NATIVE_INDEX_READER` tag) in the POSTING covering-index read path, and adds a deterministic regression test.

## Symptom

`PostingIndexCriticalIssuesTest.testMultiThreadedO3CoveringLatestByConcurrentReaderFuzz` fails **non-deterministically** with:

```
java.lang.AssertionError: Memory usage by tag: NATIVE_INDEX_READER, difference: 512 expected:<0> but was:<512>
	at io.questdb.test.tools.TestUtils$LeakCheck.close(TestUtils.java:2655)
	at io.questdb.test.cairo.PostingIndexCriticalIssuesTest.testMultiThreadedO3CoveringLatestByConcurrentReaderFuzz(...)
```

It is a fuzz/concurrency flake: the seed only fixes the data shape, not the thread interleaving, so it cannot be reproduced on demand from the seed alone.

## Root cause

`PostingIndexBwdReader` / `PostingIndexFwdReader` pool idle row cursors in `freeCursors` / `freeNullCursors`. To reuse the decode scratch, the pooling `close()` path frees the covering caches and the EF rank directory but **deliberately retains the cursor's `blockBufferAddr`** — a `NATIVE_INDEX_READER` allocation of up to `BLOCK_CAPACITY` longs (64 × 8 = **512 bytes**). That block buffer is only ever reclaimed when the **reader's** own `close()` drains the cursor pool (`releaseResources()` per pooled cursor).

So if a cursor returns to the pool **after** its owning reader was already closed — e.g. a reader-thread cursor that outlives a concurrent reseal/reload — `cursor.close()` re-pools it (with its retained block buffer) into a reader whose `freeCursors` list will never be drained again. The 512-byte block buffer leaks.

`blockBufferAddr` is only allocated on the non-constant-delta decode paths (packed/EF/flat), which is exactly why the leaked size is a clean 512 bytes.

## Fix

Guard the re-pool with the reader's `isOpen()` state in both readers' `Cursor.close()` and `NullCursor.close()`:

```java
if (!isPooled && isOpen() && free*Cursors.size() < MAX_CACHED_FREE_CURSORS) {
    ... // re-pool, retaining the block buffer for reuse
}
releaseResources(); // reader already closed -> free the block buffer now instead of leaking
```

When the owning reader is already closed, the cursor releases its native buffers immediately rather than pooling into a dead reader. The block-buffer reuse optimisation is preserved for the normal (reader-open) case.

## Test

Adds `testRowCursorClosedAfterReaderDoesNotLeakBlockBuffer` — a deterministic, single-threaded regression that:

1. builds a single key with irregular row-id gaps (forces a non-zero bit-width block that allocates `blockBufferAddr`),
2. drives a cursor to completion so the decode block buffer is allocated,
3. closes the reader while the cursor is still checked out,
4. then closes the cursor.

Before the fix this reproduces the **exact** CI signature (`NATIVE_INDEX_READER, difference: 512`) in ~60 ms with no threads.

## Verification

- New test: red before the fix, green after.
- Full `PostingIndexCriticalIssuesTest`: **96 tests, 0 failures** (including the concurrent fuzz tests).

### DIFF
--- a/core/src/main/java/io/questdb/cairo/idx/AbstractPostingIndexReader.java
+++ b/core/src/main/java/io/questdb/cairo/idx/AbstractPostingIndexReader.java
@@ -103,6 +103,9 @@ public abstract class AbstractPostingIndexReader implements IndexReader {
     protected int genCount;
     protected int keyCount;
     protected RecordMetadata metadata;
+    // Assertion-only stamp of the thread that last checked out a cursor; see
+    // assertStampOperatingThread() / assertSameOperatingThread().
+    private long assertOperatingThreadId = -1L;
     // Last successfully observed seqlock value of the chain header's active
     // page. Used by reloadConditionally to detect any publish (appendNewEntry
     // or extendHead — both republish the header) and skip the picker walk
@@ -1075,6 +1078,27 @@ public abstract class AbstractPostingIndexReader implements IndexReader {
 
     protected static long varBlockOffsetsSize(int count, boolean longOffsets) {
         return (long) (count + 1) * (longOffsets ? Long.BYTES : Integer.BYTES);
+    }
+
+    // Single-owner tripwire (assertion-only). A posting reader and its pooled
+    // cursors are driven by exactly one thread at a time: the thread that owns the
+    // enclosing TableReader between pool acquire/release. getCursor() stamps that
+    // thread via assertStampOperatingThread() and every cursor close() checks it
+    // here. A posting cursor whose close() runs after the reader was released to the
+    // pool and re-acquired by another thread -- the lifecycle hazard that
+    // CoveringIndexRecordCursorFactory.CoveringCursor.close() avoids by freeing the
+    // row cursor BEFORE the frame cursor -- trips this assert instead of silently
+    // re-pooling into / racing a concurrently-reloaded reader. Never relied upon for
+    // correctness: the isOpen() guard in each cursor close() is the actual leak
+    // mitigation, and this stamp is only written under -ea.
+    protected boolean assertSameOperatingThread() {
+        final long owner = assertOperatingThreadId;
+        return owner == -1L || owner == Thread.currentThread().threadId();
+    }
+
+    protected boolean assertStampOperatingThread() {
+        assertOperatingThreadId = Thread.currentThread().threadId();
+        return true;
     }
 
     protected void ensureSidecarOpen(int c) {
--- a/core/src/main/java/io/questdb/cairo/idx/PostingIndexBwdReader.java
+++ b/core/src/main/java/io/questdb/cairo/idx/PostingIndexBwdReader.java
@@ -177,7 +177,14 @@ public class PostingIndexBwdReader extends AbstractPostingIndexReader {
 
         @Override
         public void close() {
-            if (!isPooled && freeCursors.size() < MAX_CACHED_FREE_CURSORS) {
+            // Only return to the idle pool while the owning reader is still open.
+            // The pool retains blockBufferAddr (NATIVE_INDEX_READER) for reuse and
+            // relies on the reader's close() draining freeCursors to reclaim it; a
+            // cursor that re-pools after the reader was closed (e.g. a reader-thread
+            // cursor outliving a concurrent reseal/reload) would never be drained
+            // again and would leak its block buffer. When the reader is closed,
+            // release everything immediately instead.
+            if (!isPooled && isOpen() && freeCursors.size() < MAX_CACHED_FREE_CURSORS) {
                 isPooled = true;
                 closeCoveringResources();
                 if (efRankDirAddr != 0) {
@@ -920,7 +927,9 @@ public class PostingIndexBwdReader extends AbstractPostingIndexReader {
 
         @Override
         public void close() {
-            if (!isPooled && freeNullCursors.size() < MAX_CACHED_FREE_CURSORS) {
+            // See Cursor.close(): never re-pool into a closed reader, otherwise the
+            // retained blockBufferAddr (NATIVE_INDEX_READER) leaks.
+            if (!isPooled && isOpen() && freeNullCursors.size() < MAX_CACHED_FREE_CURSORS) {
                 isPooled = true;
                 closeCoveringResources();
                 if (efRankDirAddr != 0) {
--- a/core/src/main/java/io/questdb/cairo/idx/PostingIndexBwdReader.java
+++ b/core/src/main/java/io/questdb/cairo/idx/PostingIndexBwdReader.java
@@ -92,6 +92,7 @@ public class PostingIndexBwdReader extends AbstractPostingIndexReader {
 
     @Override
     public RowCursor getCursor(int key, long minValue, long maxValue, int[] requiredCoverColumns) {
+        assert assertStampOperatingThread();
         reloadConditionally();
 
         // See PostingIndexFwdReader.getCursor: clamp the index-walked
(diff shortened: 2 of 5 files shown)


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 42 / 44 (95.45%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/http/HttpSessionStore.java|6|8|75.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexBwdReader.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/std/Unsafe.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/MemoryConfigurationImpl.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/PropServerConfiguration.java|24|24|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpSessionStoreImpl.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexFwdReader.java|2|2|100.00%|

bluestreak01:
## Review of PR #7256 — Level 3 (full mission-critical pass)

11-agent fanout + per-finding source verification. Two reviewer agents independently built and ran the suite: the 4 new tests pass green under `-ea`, and the full `PostingIndexCriticalIssuesTest` (96–99 tests) passes including the concurrent fuzz.

### Verdict: Approve with minor follow-ups

The fix is correct, minimal, and well-reasoned. It has two parts that work together:
1. **`isOpen()` re-pool guard** in all four posting-cursor `close()` paths — the always-on leak mitigation.
2. **`CoveringCursor.close()` reorder** (free `currentRowCursor` before `frameCursor`) — the actual cross-thread correctness fix, ensuring the row cursor re-pools into a still-open reader on the owning thread before the `TableReader` returns to the pool.

Verified: **no double-free / use-after-free** (the free-sets in `releaseResources()` vs `closeCoveringResources()` are disjoint and idempotent), the reorder matches the existing `CoveringPageFrameCursor.close()` precedent, and the new `assert` tripwire is a true production no-op that **cannot false-trip** on any legitimate query — including the parallel `LATEST BY` path, which uses raw index addresses and never calls `getCursor()`.

### Critical
None.

### Moderate

**M1 — The headline concurrency fix (the `close()` reorder) has no deterministic regression test.** (in-diff, test file)
`CoveringIndexRecordCursorFactory.java:509-510`. The 4 new tests construct `PostingIndex*Reader` directly and never instantiate `CoveringIndexRecordCursorFactory`, so they exercise only the `isOpen()` guard. Reverting **only** the reorder (keeping the guard) leaves all 4 tests green while reintroducing the flaky fuzz leak. Single-threaded, the reorder is in fact untestable (the guard masks the leak in both orders); its sole value is the cross-thread race, whose only coverage is the fuzz test the PR itself states "cannot reproduce the race on demand." Recommendation: either add a targeted concurrency test (return the frame cursor to the pool + reload the reader on another thread between the two frees, assert no corruption), or state explicitly in the PR body that the reorder is concurrency-only and not deterministically reproducible.

### Minor

- **m1 — `isOpen()` leak-correctness now hinges on `getFd()` reliably reaching -1.** (out-of-diff, pre-existing) `MemoryCMRImpl.close()` (`:84-86`) resets `fd=-1` only when `ff.close(fd)` succeeds. On a rare `close()` syscall failure, `munmap` still runs but `fd` stays ≥ 0, so `isOpen()` returns `true` on an unmapped, drained reader and the next `cursor.close()` would re-leak the block buffer. Exceptional and pre-existing, but the PR newly leans on `isOpen()`. If hardening, gate on `pageAddress != 0` or a dedicated closed flag.
- **m2 — Doc tension on what protects production.** `AbstractPostingIndexReader.java:1091-1093` says "the `isOpen()` guard ... is the actual leak mitigation," while the `Cursor.close()` comments correctly say the *ordering* provides cross-thread safety and `isOpen()` is "NOT a concurrency primitive." A maintainer reading the abstract-class comment in isolation could drop the ordering. Align the abstract-class comment.
- **m3 — The single-owner assert is an unreliable detector.** `assertOperatingThreadId` is a plain non-volatile `long`; in the cross-thread close it targets there is no happens-before edge, so it can read a stale value and not fire. Documented as "never relied upon," but its prominence overstates the protection. If it *does* fire under `-ea`, the `AssertionError` skips `releaseResources()` and leaves `frameCursor` unfreed for that run (acceptable — a fired tripwire is itself a failure).
- **m4 — PR metadata.** No labels are set — add `Bug`, `Core`, `Flaky Test`. The title leads with the internal memory-tag name `NATIVE_INDEX_READER` rather than user-facing impact (titles should speak to end-user impact). The body uses bold emphasis on numbers/words (`**512 bytes**`, `**96 tests, 0 failures**`, `**non-deterministically**`), against the level-headed-tone guideline.
- **m5 — Test comment ambiguity.** `PostingIndexCriticalIssuesTest.java:~2256` says "BLOCK_CAPACITY longs = 512 bytes." Two `BLOCK_CAPACITY` constants exist (`PostingIndexUtils`=64 → 512 B, the intended one; `BitpackUtils`=128). The number is right; qualify the name.
- **m6 — Comment verbosity.** The ~19-line rationale block in `PostingIndexBwdReader.Cursor.close()` is duplicated in condensed form at the other three sites that already say "See Cursor.close()". Housekeeping only.

### Downgraded (false positives, dismissed after source verification)
- **Member ordering of `assertOperatingThreadId`** — Correct as written. It is the first **private** field (the protected group ends at `metadata:105`) and sorts alphabetically before `chainSequence`.
- **`Thread.threadId()` compile risk** — Dismissed. Already used pervasively in core; build targets Java 25.
- **Double-free / use-after-free between pool drain and later `close()`** — Dismissed. Checked-out cursors aren't in `freeCursors`; pooled cursors hold no live reference after `Misc.clear`; all frees are guarded by `addr != 0` and zero-after-free. `closeCoveringResources()` touches only cursor scratch, never the reader's `keyMem`.
- **`assertSame/assertStampOperatingThread` boolean naming without `is/has`** — Dismissed. Follows established assert-helper precedent (`TableWriter.assertColumnPositionIncludeWalLag()`).
- **Branch commit titles using Conventional-Commits prefixes** — Dismissed. PR squash-merges; branch history is throwaway and the squash title correctly uses `fix(core):`.

### Summary
- **Verdict:** approve; address M1 (or document the coverage gap) and the PR-metadata nits (m4) before merge.
- **No regressions or correctness hazards found** in the change itself. Verified by a passing build + full test class run.
- **Findings:** ~16 draft → **1 Moderate + 6 Minor verified, 5 dropped as false positives.**
- **In-diff vs out-of-diff:** 6 in-diff, 1 out-of-diff (the pre-existing `MemoryCMRImpl` fd-reset edge). The cross-context pass ran a full callsite inventory across all posting-index consumers (covering, DISTINCT, `SymbolIndexRowCursorFactory`, `LatestBy*Indexed*`, parallel `LatestByAllIndexed`) and found every callsite SAFE.

<sub>🤖 Generated with adversarial level-3 review (pi review-pr skill)</sub>

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexBwdReader.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/AbstractPostingIndexReader.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexFwdReader.java|5|5|100.00%|

PR #7257, opened 2026-06-14
fix(qwp-ws): fix flushAndGetSequence() returning a stale FSN after an empty flush

### LINKED ISSUE #7142: QWP WS: flushAndGetSequence() contract conflicts with implementation and drain() semantics
## Problem

`Sender.flushAndGetSequence()` documents per-call semantics: return the highest FSN published by this call, and `-1` when this call publishes nothing. `QwpWebSocketSender.flushAndGetSequence()` repeats that contract.

Current implementation returns the global cursor watermark instead:

```java
return cursorEngine != null ? cursorEngine.publishedFsn() : -1L;
```

So an empty flush after a previous publish returns the previous FSN, not `-1`.

## Why this is subtle

The current `Sender.drain(timeout)` default depends on the watermark behavior:

```java
return awaitAckedFsn(flushAndGetSequence(), timeoutMillis);
```

If `flushAndGetSequence()` is changed to obey the formal per-call contract without also changing `drain()`, then `drain()` after a prior `flush()` with unacked frames can return immediately because the empty flush returns `-1`.

`QwpWebSocketSender.close()` already uses the watermark explicitly (`cursorEngine.publishedFsn()`), so close-drain semantics are separate.

## Suggested resolution

Pick and document one contract:

1. Keep the formal per-call contract: return `-1` when this call publishes nothing, and override QWP WS `drain()` to wait for the current published watermark after flushing.
2. Or keep watermark semantics: update `Sender` and `QwpWebSocketSender` docs to say the method returns the current highest published FSN after flush, not necessarily one published by this call.

A red test for option 1 is straightforward: publish one row, assert first `flushAndGetSequence()` returns `0`, then call `flushAndGetSequence()` again with no buffered data and expect `-1`; current behavior returns `0`.

adi1719:
Fixes #7142

`QwpWebSocketSender.flushAndGetSequence()` documents a per-call contract --
return the highest FSN published by *this* call, or `-1` when the call
publishes nothing -- but it returned the global published-FSN watermark
instead. An empty flush after a previous publish returned the previous FSN
rather than `-1`, violating the documented contract.

This change enforces the per-call contract by snapshotting the published FSN
before and after the flush work, returning the new FSN only when this call
actually advanced it, and `-1` otherwise.

Returning `-1` on an empty flush is observable: callers that read
`flushAndGetSequence()` directly after a no-op flush now get `-1` instead of
the previous watermark. `drain()` callers are unaffected (see below).

`drain()` behavior is unchanged. The default `Sender.drain()` derives its
wait target from `flushAndGetSequence()`, so once an empty flush returns
`-1`, the default would short-circuit immediately even with prior
unacknowledged frames still in flight. To keep `drain()`'s existing
watermark semantics, `QwpWebSocketSender` now overrides `drain()` to flush
and then wait on the current published-FSN watermark directly (the same
approach close-time drain already uses). This override preserves the prior
behavior rather than fixing a separately shipped `drain()` bug.

Changes:
- `QwpWebSocketSender.flushAndGetSequence()`: snapshot publishedFsn
  before/after flush, return the new FSN only when it advanced, else `-1`
- `QwpWebSocketSender.drain()`: new override using watermark semantics to
  preserve drain() behavior under the new `flushAndGetSequence()` contract
- `Sender.drain()`: Javadoc noting implementations may override with
  watermark semantics
- Tests: deterministic regression test in the client module (`CloseDrainTest`
  with a delayed-ACK server) plus e2e coverage for the `-1`-on-empty contract
  and drain-after-flush


bluestreak01 (changes requested):
@adi1719 — review below. Requesting changes (non-blocking on correctness; blocking on test effectiveness).

## What the PR actually changes

The visible core diff is only a submodule pointer bump + 2 new tests. The real change is in the **`java-questdb-client` submodule** (commit `9f26e6c`):

- `QwpWebSocketSender.flushAndGetSequence()` — snapshots `publishedFsn()` before/after the flush and returns the new FSN only when this call advanced it, else `-1` (was: unconditionally returned the global watermark).
- `QwpWebSocketSender.drain(long)` — new override using watermark semantics (`flush()`, then await the current `publishedFsn()`), so empty-flush `-1` doesn't short-circuit drain.
- `Sender.java` — Javadoc only.
- Core repo — 2 new JUnit tests.

The fix is **correct and minimal**. The `drain()` override is behaviorally identical to the *pre-PR* QWS drain (which already waited for the watermark because the old `flushAndGetSequence()` returned it), so this is a faithful, coordinated fix of the contract described in #7142. No correctness regression at any of the 7 verified call sites. The first new test (`...ReturnsNegativeOneWhenNothingPublished`) is a genuine, deterministic regression guard.

## Critical
None.

## Moderate

### M1 — `testDrainWaitsForPriorUnackedFrames` is not a reliable regression guard for the half of the fix it claims to cover
`core/.../QwpWebSocketSenderObservabilityTest.java:82-101`

This test is the *only* coverage for the `drain()` override — the subtler, more important half of #7142 — and it does not deterministically fail on the un-fixed code. Trace the **buggy** path (drain *not* overridden ⇒ `Sender.drain()` default = `awaitAckedFsn(flushAndGetSequence(), t)`):

1. `flush()` (line 91) publishes FSN 0; the ACK round-trips asynchronously.
2. `drain(10_000)` (line 94) → inner `flushAndGetSequence()` finds nothing buffered → returns `-1` → `awaitAckedFsn(-1, …)`. Since `cursorEngine.ackedFsn() >= -1` is always true (`QwpWebSocketSender.java:757`), it returns `true` **immediately, without waiting**.
3. `assertTrue("drain() must wait...", drained)` (line 95) → **passes on buggy code too**. This assertion is completely non-discriminating.
4. `getAckedFsn() >= 0` (lines 98-99) is the only assertion that *could* catch the bug, but it samples the ack state a few microseconds after `flush()` against a **real in-process server with no controllable ack timing** (`AbstractQwpWebSocketTest.runInContext` builds a live `HttpServer`/`QwpIngressHttpProcessor`). Whether the loopback ack has landed by line 98 is a pure race — so this assertion is **timing-dependent**, not deterministic, and can false-pass on the buggy code.

Net: a green run carries almost no signal that the override is present. The test asserts an end-state (`acked`) the system reaches anyway after `flush()`, not the discriminating behavior (that `drain()` *blocked* until the watermark was acked). It does *not* flake on correct code, so this is a "weak guard," not a "flaky test."

**Fix:** add a deterministic test in the submodule's `CloseDrainTest` (which has `DelayingAckHandler`/`SilentHandler` mock servers), shaped like the existing close-drain tests:

```java
DelayingAckHandler(600);
sender.table("foo").longColumn("v", 1L).atNow();
sender.flush();                         // publish FSN 0; ack delayed 600ms
long t0 = System.nanoTime();
boolean drained = sender.drain(5_000);  // inner empty flush → -1 on buggy code
long elapsedMs = (System.nanoTime() - t0) / 1_000_000;
assertTrue(drained);
assertTrue("drain must wait for prior unacked frame", elapsedMs >= 300);
```

On buggy code `drain` short-circuits at `elapsed≈0` → fails deterministically; on fixed code it waits ~600ms → passes. Note the existing `CloseDrainTest.testDrainBlocksUntilAckArrivesAndReturnsTrue` (`CloseDrainTest.java:185-214`) does *not* cover this: it never calls `flush()` first, so the inner `flushAndGetSequence()` publishes the row and returns a real FSN — the empty-flush `-1` short-circuit is never exercised.

### M2 — The submodule change ships with no co-located test
The submodule commit `9f26e6c` touches only `Sender.java` + `QwpWebSocketSender.java` — zero test files. Both regression tests live in the core e2e suite, where the harness cannot control ack timing (see M1). The natural, controllable home for the `drain()` regression test is the submodule. This compounds M1: the submodule's own CI can pass without ever exercising the drain short-circuit path.

## Minor

### m1 — Trailing whitespace introduced
`QwpWebSocketSender.java:1359` — the PR adds a blank line containing 8 trailing spaces (between `checkConnectionError();` and `long afterFsn = ...`). Strip it.

### m2 — Dead sub-condition `afterFsn != -1L`
`QwpWebSocketSender.java:1361`: `return (afterFsn > beforeFsn && afterFsn != -1L) ? afterFsn : -1L;`. Since `publishedFsn()` is `>= -1` by invariant and monotonic within a call, `afterFsn > beforeFsn` already implies `afterFsn >= 0`. The `&& afterFsn != -1L` clause can never flip the result — harmless but dead. Simplify to `afterFsn > beforeFsn ? afterFsn : -1L`.

### m3 — PR metadata / conventions
- **No labels.** A `fix(qwp-ws)` bug fix should carry GitHub labels (e.g. `Bug` + an ILP/client label).
- **`Closes #7142` at the bottom** of the body; convention is `Fixes #NNN` at the **top**.
- **Title leans on internal method names** ("enforce per-call contract for `flushAndGetSequence()` and override `drain()`") rather than end-user impact. Borderline — these are public client-API methods — but it reads as implementation-internal.

### m4 — Submodule commit message violates commit conventions
`9f26e6c` title: `fix(ws): implement watermark semantics for drain method in QwpWebSocketSender` — 77 chars (convention: ≤50), uses a Conventional-Commits `fix(ws):` prefix (convention: plain-English titles, no CC prefix in commit messages), and has **no long-form body**.

## Downgraded (false positives / non-issues after verification)

- **"`flushAndGetSequence()` before/after delta could be corrupted by a concurrent publisher."** Dismissed: `publishedFsn` is a volatile single-writer field written only by the user thread (`SegmentRing.java:40-49`, writes at ctor and `appendOrFsn`). Background drainers and the I/O thread never write it; `autoFlushIntervalNanos` is assigned but never consumed. The publish during flush runs on the *same* user thread. The before/after pair is race-free.
- **"A flush that only emits a deferred commit frame wrongly returns an FSN instead of `-1`."** Dismissed: `sendCommitMessage()` seals and publishes a real commit frame via `sealAndSwapBuffer()`→`appendBlocking`, so the call genuinely published a frame; returning its FSN is consistent with "highest FSN published by this call." (Also irrelevant to the new tests, which run with `deferCommit=false`.)
- **"Extra volatile reads on the flush path are a perf regression."** Dismissed: the per-row auto-flush path goes through `flushPendingRows` directly, not `flushAndGetSequence`; the added reads land only on explicit flush/drain and are ~1ns against a segment-seal + network RTT. The 50µs park-poll in `awaitAckedFsn` is the correct choice for a client blocking-await (a condition var would pessimize the lock-free I/O ack thread).
- **"Caller `QwpSenderE2ETest.java:4781` breaks on the new `-1`."** Dismissed: every `assertCoercionError` path buffers a valid row and flushes (non-empty), so it gets a real FSN exactly as before; the `-1` path is never hit there. All 7 callers verified non-breaking.

## Summary

**Verdict: request changes** (non-blocking on correctness; blocking-quality on test effectiveness).

The actual code fix is correct, minimal, and well-reasoned: the `flushAndGetSequence()` per-call contract is now honored, the `drain()` override exactly restores the pre-PR watermark behavior, and no caller regresses. The contract half is covered by a deterministic test. **The weakness is the test for the `drain()` half** — its only meaningful assertion is timing-dependent and its boolean assertion is non-discriminating, so it does not deterministically fail on the un-fixed code, and the submodule ships with no co-located test at all (M1, M2). Add a deterministic `DelayingAckHandler` + elapsed-time test in `CloseDrainTest` before merge. Plus minor cleanups: trailing whitespace (m1), dead condition (m2), labels/`Fixes`/commit-message conventions (m3, m4).

- Findings: 8 raised, 6 verified/confirmed, 4 dismissed as false positives.
- In-diff vs out-of-diff: all 6 confirmed findings in-diff; 0 out-of-diff (full callsite inventory walked — blast radius genuinely contained).

adi1719:
@bluestreak01  Could you please review the PR again.
Thanks

bluestreak01 (approved):
Approve.

Re-reviewed at level 3 after the latest updates. The fix is correct, minimal, and race-free: `publishedFsn` is volatile single-writer on the user thread, so the before/after snapshot in `flushAndGetSequence()` is exact, and the `drain()` override is behaviorally identical to the pre-PR watermark drain (no caller regressions).

Verified the prior review feedback is addressed:
- Added `CloseDrainTest.testDrainAfterFlushWaitsForPriorUnackedFrames` (deterministic). I ran it both ways: passes on the fix, and with the `drain()` override removed it fails with 'returned in only 0ms (expected >= 300ms)' — a genuine regression guard.
- Dead `afterFsn != -1L` guard removed; trailing whitespace gone; `Fixes #7142` now at the top; submodule commit message has a proper plain-English title and long-form body.

Callsite inventory walked: QwpWebSocketSender is the only FSN-tracking sender overriding flushAndGetSequence(); PooledSender delegates correctly; all flushAndGetSequence() callers flush non-empty buffers and get a real FSN, so none hit the new -1 path.

Minor (non-blocking): `drain()` should sit alphabetically between `doubleColumn` and `floatColumn` per the member-ordering convention; the PR is missing labels (suggest Bug + ILP).

PR #7260, opened 2026-06-14
fix(sql): fix internal error when a dotted name follows an operator

bluestreak01:
## Problem

Our demo box was being probed by vulnerability scanners with requests like:

```
tables()/.env
tables()/.git/HEAD
```

These are sent to `/exec?query=...` and reach the SQL compiler. `tables()/.env` is **not** a path lookup — the parser reads it as the division expression `tables() / .env` in the FROM clause.

While building the AST, the expression parser handles the `.env` dotted name by peeking the operator stack and casting the qualifier node's token to `GenericLexer.FloatingSequence`:

```java
final GenericLexer.FloatingSequence fsA = (GenericLexer.FloatingSequence) en.token;
```

When the stack top is a **pending operator** (`/`), its token is a plain `java.lang.String`, so the unchecked cast throws a raw **`ClassCastException`** instead of a `SqlException`. The result is that a malformed scanner probe surfaces as an **internal/critical error** (500-class, stack-traced in the logs) rather than an ordinary "bad query" `400`.

Reproduced directly against the compiler:

| query | before | after |
|---|---|---|
| `tables()/.env` | `ClassCastException` | `SqlException: '.' is unexpected here` |
| `tables()/.git/HEAD` | `ClassCastException` | `SqlException: '.' is unexpected here` |
| `tables() / .env` | `ClassCastException` | `SqlException: '.' is unexpected here` |
| `select 1/.env from long_sequence(1)` | `ClassCastException` | `SqlException: '.' is unexpected here` |

(The crash-vs-clean behavior even varied with pooled parser state across queries, which is why some variants happened to look fine.)

## Fix

`ExpressionParser` — guard the fast-path concat so it only runs when the qualifier token is actually a `FloatingSequence`; otherwise reject cleanly with `'.' is unexpected here`. This is general: it fixes **every** malformed `<expr> / .name` shape, not just the specific names scanners use.

## Tests

- New regression test `ExpressionParserTest#testDottedNameAfterOperatorFailsCleanly`. It fails with the original `ClassCastException` when the fix is reverted (verified), and passes with the fix.
- Full parser suites green: `ExpressionParserTest` (296), `SqlParserTest` (1081), `ExpressionParserFuzzTest`, `SqlParserUpdateTest` — **1410 tests, 0 failures**.
- Valid dotted names are unaffected: `(a.b).c`, `"x".y`, `tables().f(...)`, qualified columns/tables, etc.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/ExpressionParser.java
+++ b/core/src/main/java/io/questdb/griffin/ExpressionParser.java
@@ -2013,10 +2013,17 @@ public class ExpressionParser {
                                 cse.put(en.token).put(GenericLexer.unquoteIfNoDots(tok));
                                 opStack.push(expressionNodePool.next().of(
                                         ExpressionNode.LITERAL, cse.toImmutable(), Integer.MIN_VALUE, en.position));
-                            } else {
+                            } else if (en.token instanceof GenericLexer.FloatingSequence) {
                                 final GenericLexer.FloatingSequence fsA = (GenericLexer.FloatingSequence) en.token;
                                 // vanilla 'a.b', just concat tokens efficiently
                                 fsA.setHi(lexer.getTokenHi());
+                            } else {
+                                // 'en' is not a valid qualifier for a dotted name. This happens for
+                                // malformed input such as "tables()/.env", where the top of the stack
+                                // is a pending operator ('/') rather than an identifier. Reject it
+                                // cleanly instead of letting the cast above fail with an internal
+                                // ClassCastException (which would surface as a 500/critical error).
+                                throw SqlException.$(lastPos, "'.' is unexpected here");
                             }
                         } else if (prevBranch == BRANCH_DOT_DEREFERENCE) {
                             argStackDepth++;
--- a/core/src/test/java/io/questdb/test/griffin/ExpressionParserTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/ExpressionParserTest.java
@@ -901,6 +901,18 @@ public class ExpressionParserTest extends AbstractCairoTest {
     }
 
     @Test
+    public void testDottedNameAfterOperatorFailsCleanly() {
+        // A dotted name whose qualifier is a pending operator (not an identifier) used to
+        // blow up with an internal ClassCastException, e.g. the demo-box scanner probes
+        // "tables()/.env" and "tables()/.git/HEAD" (parsed as the division tables() / .env).
+        // It must now fail with a clean SqlException instead.
+        assertFail("tables()/.env", 10, "'.' is unexpected here");
+        assertFail("tables()/.git/HEAD", 10, "'.' is unexpected here");
+        assertFail("tables() / .env", 12, "'.' is unexpected here");
+        assertFail("1/.env", 3, "'.' is unexpected here");
+    }
+
+    @Test
     public void testDotDereferenceFunction() throws SqlException {
         x("a.b 1 2 3 f .", "(a.b).f(1,2,3)");
     }


bluestreak01:
## Review — `fix(sql): fix internal error when a dotted name follows an operator`

**Verdict: Approve with minor changes.** The core change is correct, minimal, and regression-free. It builds and the new regression test passes (`ExpressionParserTest#testDottedNameAfterOperatorFailsCleanly`).

By construction the change can only convert a previously-thrown `ClassCastException` (a 500/internal error) into a `SqlException` (a clean 400): any input that reached the old `else` and succeeded necessarily had `en.token instanceof FloatingSequence`, so it still runs the identical `setHi`. I verified that no production code anywhere catches `ClassCastException`, so no caller relied on the old behavior, and every reachable context (REST `/exec`, `/exp`, pgwire, ILP, materialized views) handles the new `SqlException` correctly — pgwire and `/exp` actually improve (real message + position instead of a raw cast string).

No Critical findings.

### Moderate

**1. The fix is asymmetric — the quoted variant `tables()/."env"` still mis-parses silently (out-of-diff, pre-existing).** `core/src/main/java/io/questdb/griffin/ExpressionParser.java:2007`

The unchanged sibling branch `if (Chars.isQuoted(tok) || en.token instanceof NameAssemblerCharSequence)` runs *before* the new guard and never validates that `en` is a real qualifier. For `tables()/."env"` the quoted `tok` short-circuits into that branch, which does `opStack.pop()` (discarding the pending `/` operator) and `cse.put(en.token)` → concatenates the operator token `"/"` with `env`. **Confirmed empirically**: this throws *nothing* and parses to RPN `tables /env` — silently dropping the division and producing garbage, which is arguably worse than the CCE this PR fixes. Same root cause and same probe family, just on the quoted path. It is **pre-existing and out of scope** (not a regression), but the PR's stated goal — "fixes *every* malformed `<expr> / .name` shape" — overstates the result. Suggested follow-up: gate the first branch on `en` also being a valid qualifier and route operator/constant qualifiers to the same `'.' is unexpected here` error, then add a `tables()/."env"` test case. Not blocking for the unquoted scanner probes the PR actually targets.

**2. Missing PR labels.** The PR has zero labels. A `fix(sql):` PR should carry **SQL** and **Bug**; given the vulnerability-scanner context, **Security** is also warranted.

### Minor

**3. New test method is alphabetically misplaced.** `ExpressionParserTest.java:904` — `testDottedNameAfterOperatorFailsCleanly` was inserted inside the `testDotDereference*` cluster. `testDotted...` sorts after the entire `testDot[A-Z]*` group and before `testDouble*`, so its correct slot is between `testDotStar` (999) and `testDoubleArrayDereference` (1004).

**4. Error position vs. message wording.** `SqlException.$(lastPos, "'.' is unexpected here")` points the caret at the identifier *after* the dot (col 10 for `tables()/.env`), but the message names the `'.'`, which sits at `lastPos - 1` (col 9). The whitespace guard at line 2000 guarantees `lastPos - 1` is exactly the dot when the throw is reached, and the sibling `"'[' is unexpected here"` (line 1173) points *at* the `[`. Per the error-position convention, `lastPos - 1` would mark the offending character the message names. (Tightening it requires updating the asserted positions to 9/9/11/2.) Low priority.

**5. `instanceof` without a pattern variable.** `ExpressionParser.java:2016-2017` does `else if (en.token instanceof GenericLexer.FloatingSequence)` immediately followed by a manual `(GenericLexer.FloatingSequence) en.token` cast. Folding to `instanceof GenericLexer.FloatingSequence fsA` drops the redundant cast (line 805 already does exactly this). Bytecode/perf identical — readability only.

**6. PR description tone.** Prefer a level-headed tone with no bold emphasis on numbers; the body bolds `**1410 tests, 0 failures**` and several words, and uses passive voice ("Our demo box was being probed", "These are sent to `/exec`"). De-bold the metrics and switch to active voice.

**7. Optional extra coverage.** The parser-level test is sufficient. Optionally worth adding: a SQL-compiler-level `assertException("tables()/.env", 10, ...)` exercising the full `/exec` compile path, and one or two other-operator variants (`*`, `||`).

### Checked and dismissed (non-findings)

- *"`instanceof` + cast adds a redundant type check (perf)"* — the old `else` already executed a `checkcast`; the new `instanceof` is the same test hoisted one line earlier, and HotSpot folds the dominated `checkcast`. Zero added cost on the once-per-compile path, zero allocation on the success path.
- *"Missing `assertMemoryLeak()`"* — the file has zero `assertMemoryLeak` usages; `testParseExpression` allocates no native memory. The new test matches the file convention.
- *"Builder-API assertion rules violated"* — N/A; this is a parser-level `assertFail` test, not an `assertQuery`/`assertSql` query-result test.
- *Sibling unguarded cast at `ExpressionParser.java:~1111`* — only runs after a literal/constant branch (tokens always `FloatingSequence`); could not construct a reaching input, so unproven and out of scope.

### Summary

Correct, minimal, verified by build + passing test; no Critical issues, no regressions. The change is strictly safer (CCE→SqlException) and improves error reporting in pgwire and `/exp` as a side effect. The honest caveat is the quoted-variant `tables()/."env"` (Moderate #1), which the "fixes every shape" claim doesn't actually cover — pre-existing, out of scope, suggested as follow-up.

<sub>Reviewed at level 3: all applicable agents (Rust agent N/A), with build, regression-test run, an empirical quoted-variant probe, and per-finding source verification.</sub>

nwoolmer (approved):

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/ExpressionParser.java|3|3|100.00%|

PR #7266, opened 2026-06-16
chore(core): prevent a JVM crash when an async writer command races a writer close

RaphDal:
## Summary

A thread publishing an async writer command (`CairoEngine.getWriterOrPublishCommand`) can serialize into a `TableWriter`'s command queue at the same moment another thread closes that writer and frees the queue. The publisher then writes into freed/zeroed native memory and the JVM aborts with `SIGSEGV` at address `0x0`. This change makes the writer pool wait for in-flight publishers before it frees a writer's command queue, so the publish and the close can no longer overlap.

The race surfaced in CI as a forked-VM crash in the enterprise `StoragePolicyFuzzTest#testConcurrentInsertsAndStoragePolicyChanges` (Linux, `-XX:+UseParallelGC`), but the defect and the fix are entirely in OSS `WriterPool` / `TableWriter`. The enterprise storage-policy jobs only exercise the path more aggressively than async `ALTER`/`UPDATE` normally do.

## Root cause

`addCommandToWriterQueue` captures `e.writer` and then calls `writer.publishAsyncWriterCommand(...)`, which runs `command.serialize(task)` to write into a `TableWriterTask` buffer inside the writer's `RingQueue`. Nothing stopped the owner thread from closing that writer in between:

- `closeWriter` runs `w.close()` -> `TableWriter.doClose()` -> `Misc.free(commandQueue)` -> `RingQueue.close()`.
- `RingQueue.close()` calls `TableWriterTask.close()` (which sets `appendPtr = 0` then `appendLim = 0`) and frees the backing buffer.
- A publisher reading the half-updated state `appendPtr == 0, appendLim != 0` passes `checkCapacity` (`0 + 1 > appendLim` is false) and executes `Unsafe.putByte(0, c)`, which faults. This matches the reported `si_addr: 0x0` and the `putByte <- serialize <- publishAsyncWriterCommand <- addCommandToWriterQueue` stack.

A prior change (#7121/#6417) added an `entries.get(...) != e` check inside the spin loop, but that only guards the case where `e.writer` is observed null while waiting. It does not cover the reported case, where the publisher captures a non-null writer and the close frees the queue during `serialize`.

## The fix

`WriterPool.Entry` gains a `commandPublisherCount` of threads currently serializing a command into its queue. The publisher and the closer coordinate through it:

- The publisher increments the count (a full-fence `getAndAddLong`), re-reads `e.writer` with a volatile load, and bails with `EntryUnavailableException` (the caller retries) if a close has nulled it; otherwise it publishes and decrements in a `finally`.
- `drainCommandPublishers` nulls `e.writer` (volatile) and spins until the count reaches zero before the writer is freed. The publisher's fenced increment paired with the closer's fenced count read forms a Dekker-style handshake: the closer either observes the in-flight count and waits, or the publisher observes the nulled writer and bails -- the two cannot both slip through.

The post-publish `tick()` in `addCommandToWriterQueue` now re-reads the writer under ownership and skips a null one, so a publisher that wins the owner CAS after a racing close does not tick a freed writer.

### Scope: all close paths, not only the distressed one

The reported crash is the distressed-close path, but the same capture-then-free window exists on every path that frees a pooled writer's queue while a publisher holds a captured reference: a publisher captures `e.writer` while the owner is busy, the owner returns the writer cleanly (`owner` -> `UNALLOCATED`), and a `lock`, an idle-expiry, a pool-close re-grab, or a `goodbye()` hand-off then frees it. To avoid leaving those as latent crashes, `drainCommandPublishers` runs from the shared choke points: `closeWriter` (lock / expiry / distressed / pool close), the `returnToPool` pool-closed re-grab branch, and `Entry.goodbye()`.

One path still frees the queue without passing through any of those pool choke points: the WAL drop-table purge (`ApplyWal2TableJob.purgeTableFiles` -> `TableWriter.destroy()`) calls `TableWriter.doClose()` directly, bypassing the pool's close path entirely. To cover it without special-casing the caller, `doClose()` invokes a `LifecycleManager.onBeforeClose()` pre-free hook before it frees the command queue. `WriterPool.Entry` implements that hook as `drainCommandPublishers`; `DefaultLifecycleManager` (non-pooled writers, and the pool paths that swap to it before `close()`) no-ops, so the hook only does work on the `destroy()` path. The now-orphaned entry is evicted when the writer's own `close()` later reaches `returnToPool`, after `doClose()` has released the table lock, so no window exists where the entry is gone while its lock is still held. This makes `destroy()` safe by construction rather than relying on the caller to drain first.

## Tradeoffs and costs

- The async-command publish path gains two atomic add operations and one extra volatile read per publish. This path handles `ALTER`/`UPDATE`/storage-policy commands, not per-row ingestion, so the cost is off the hot path.
- Every writer close now performs a volatile store and a fenced count read. When no publisher is in flight (the common case) it reads zero and returns immediately. `doClose()` also gains one virtual call to the `onBeforeClose()` hook, which is a no-op for the default lifecycle manager.
- While a close waits for the count to drain, it spins with `Os.pause()`. The wait is bounded: `publishAsyncWriterCommand` either completes or throws on a full queue, and the publisher always decrements in a `finally`, so the drain cannot deadlock or spin unboundedly.
- A command published into a writer that is about to be freed is dropped when the queue is freed. This is pre-existing behavior for a distressed/closing writer, unchanged by this PR; callers that need the result re-drive.
- The fix widens, by the duration of the drain plus close, the existing window in which a concurrent async publisher observes `e.writer == null` and spins/retries. This was already possible because the close nulled the writer; the writer is now nulled slightly earlier (before the free instead of after).
- The `destroy()` path now evicts its pool entry by scanning `entries` by identity (the table dir name is no longer reachable through the nulled `e.writer`). This is O(entries) but runs only on a WAL table drop, a rare path.

## Test plan

- New `WriterPoolTest#testAsyncCommandPublishDrainsBeforeDistressedClose` parks a publisher inside `serialize` and asserts a concurrent distressed close blocks in the drain until the publisher finishes, rather than freeing the buffer underneath it.
- New `WriterPoolTest#testAsyncCommandPublishDrainsBeforeDestroy` does the same for a concurrent `TableWriter.destroy()` (the WAL drop-table purge path), asserting it blocks in the `doClose()` pre-free hook before freeing the queue.
- New `WriterPoolTest#testAsyncCommandPublishConcurrentWithWriterClose`, `#testAsyncCommandPublishConcurrentWithWriterDestroy`, and `#testAsyncCommandPublishConcurrentWithLockAndIdleClose` stress the publish path against, respectively, distressed close, out-of-band `destroy()`, and the lock-close / idle-expiry (`releaseInactive()`) triggers.
- Each new test fails with the fix reverted: the deterministic tests fail their negative assertion (the close/destroy returns while the publisher is still parked mid-`serialize`), and the stress tests surface the use-after-free as an `async command/event queue buffer overflow` or a `SIGSEGV` with `si_addr: 0x0` on the `serialize -> publishAsyncWriterCommand -> addCommandToWriterQueue` stack. With the fix they pass.
- `WriterPoolTest` (31 tests, including the existing lock/expiry/distress race tests) passes.
- `WalTableSqlTest`, `WalTableFailureTest`, `WalPurgeJobTest`, and `WalAlterTableSqlTest` pass, covering the real WAL drop-table purge / `destroy()` path and the production async `ALTER` publish path.
- For the enterprise reproduction, loop the whole fuzz class (it is parameterized, so run it via `test.include` rather than `-Dtest=Class#method`):
  ```bash
  mvn test --batch-mode -Dtest.include="%regex[.*StoragePolicyFuzzTest.*]" \
    -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false -pl questdb-ent -am
  ```


bluestreak01 (approved):
Approving after a thorough concurrency review.

The core fix is correct: the Dekker handshake between the publisher (seq-cst `getAndAddLong(count,+1)` then volatile read of `writer`) and the closer (`putObjectVolatile(writer,null)` then volatile read of `count`) is sound, and SC rules out the "publisher serializes AND closer frees" overlap. Every `doClose -> Misc.free(commandQueue)` is preceded by a drain — explicitly on the three pool paths that swap to DefaultLifecycleManager, and via the onBeforeClose hook on the destroy() path. No deadlock (count released in finally; publish throws rather than blocks on a full queue), and the destroy() orphan entry is evicted O(1) via the writer's trailing close in ApplyWal2TableJob. Tests cover all four drain sites deterministically plus three stress paths and the writer_pool() null-token regression.

Non-blocking follow-ups:
- Title type: `chore(core)` should be `fix(core)` — this is a user-facing JVM-crash (use-after-free) fix labeled Bug, and the squash title feeds release notes.
- Description "Tradeoffs" bullet says the destroy()-path eviction is an O(entries) scan; the code is actually O(1) `entries.remove(e.dirName, e)` (the inline comment is correct). Please update the description.
- Enterprise CI is reported red on HEAD, and the repro is an enterprise test — worth confirming that failure is unrelated/green before merge.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 40 / 40 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/LifecycleManager.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/WriterPoolRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/pool/WriterPool.java|37|37|100.00%|

PR #7267, opened 2026-06-16
perf(sql): reduce memory footprint of cached parallel aggregations and joins

puzpuzpuz:
## Summary

Parallel `GROUP BY`, `TOP K`, and ASOF/window (horizon) joins run their filter through a per-query `AsyncFilterContext`, which holds an owner plus one-per-worker set of `DirectLongList`s for matched row ids (and, under a JIT filter, column addresses). These lists are accounted under the `NATIVE_OFFLOAD` memory tag.

Under a JIT filter the row-id lists are pre-sized to the full page frame - up to `cairo.sql.page.frame.max.rows` longs (8 MB each at the 1M default) - and `DirectLongList` capacity only ever grows. `AsyncFilterContext.clear()`, called at
cursor teardown, freed the page-frame memory pools but left these lists at their peak size; they were freed only by `close()`, on query-cache eviction. As a result an idle or cached aggregation/join factory kept `1 + workerCount` peak-sized buffers allocated for its entire cached lifetime, and a box running many such factories concurrently accumulated a large, persistent `NATIVE_OFFLOAD` footprint that did not subside while idle.

`clear()` now calls `resetCapacity()` on the owner and per-worker lists, mirroring what `PageFrameReduceTask.clear()` already does for the shared reduce-task pool. Idle/cached factories return their row-id buffers to the 256-long initial capacity and re-grow on the next reuse. This is safe because `clear()` runs only from `PageFrameSequence.reset()` at cursor teardown, guarded by the assert that no reduce tasks are in flight.

## Impact

- Lower steady-state `NATIVE_OFFLOAD` memory for workloads with many cached or long-lived parallel aggregations/joins: idle factories no longer pin peak-sized (up to 8 MB per list, `1 + workerCount` lists) row-id buffers.
- Tradeoff: a cached factory reused for another large scan pays one reallocation to grow its row-id list back up, instead of reusing a retained buffer. This is the same tradeoff the shared reduce pool already makes, and is small relative to the per-frame reduce work.
- No behavioral or result change; only the lifetime of internal buffers changes.

Note: these lists are allocated directly against the `NATIVE_OFFLOAD` tag, outside any per-query `MemoryTracker`, so a per-query memory limit still does not bound them. That is unchanged here.

## Safety fix: clear() after close()

Review surfaced a native leak that the new `resetCapacity()` call introduces on its own. `DirectLongList.resetCapacity()` re-mallocs a *closed* list: a closed list has `capacity == 0`, which differs from `initialCapacity`, so `setCapacityBytes()` runs a fresh `Unsafe.realloc(0, 0, initialCapacity)` that nothing then frees. Before this PR `clear()` never touched the lists, so calling `clear()` after `close()` was a harmless no-op; the new `resetCapacity()` turns it into a leak.

The 4 horizon-join factories reach that path on a failed `getCursor()`. `AsyncHorizonJoinRecordCursor.of()` assigns `frameSequence` before the throwing `getPageFrameCursor()` / `Function.init()` calls, so a throw leaves the cursor half-open. The factory's `_close()` then frees `frameSequence` (which closes the `AsyncFilterContext` and its lists) before the half-open cursor's `close()` re-runs `frameSequence.reset() -> atom.clear() -> filterCtx.clear() -> resetCapacity()` on the freed lists. The leaked memory is the owner row-id + data + aux lists, ~2.3 KB per failed query at the default initial capacity (the re-malloc is to initial, not peak) - bounded per failure but unbounded across repeated failures, and it trips `assertMemoryLeak`. The 3 GROUP BY/TopK factories free `cursor` first and never hit it.

Two complementary changes:

- `AsyncFilterContext.resetCapacity()` skips closed lists (`getCapacity() > 0`). This neutralizes any `clear()`-after-`close()`, including a factory double-`close()`, and preserves the shrink for open grown lists.
- The 4 horizon-join factories now free `cursor` before `frameSequence`, matching the GROUP BY/TopK factories, whose `cursor.close()` likewise depends on `frameSequence` being alive. `clear()` therefore always runs while the lists are still open, so it can never resurrect freed buffers on the teardown path.

The 2 `AsyncWindowJoin` factories share the same `frameSequence`-before-`cursor` ordering but are unaffected: their atoms hold no `AsyncFilterContext`, and `AsyncWindowJoinRecordCursor.of()` assigns `frameSequence` last with no throwing call after it, so the half-open state is not reachable. Left unchanged.

## Test plan

- `AsyncFilterContextTest` grows the owner plus per-worker row-id lists to a full frame (~40 MB), calls `clear()`, and asserts the list capacities and the `NATIVE_OFFLOAD` memory both return to their initial level. Verified it fails without the fix (`expected:<256> but was:<1000000>`).
- `AsyncFilterContextTest.testClearAfterCloseDoesNotResurrectLists` grows a context, calls `close()` then `clear()`, and asserts `NATIVE_OFFLOAD` returns to the pre-construction baseline. Verified it fails without the guard (2304 bytes leaked) and passes with it.
- `HorizonJoinTest` (165 tests) re-run green after the cursor/frameSequence reorder.
- Ran the context-path suites: `HorizonJoinTest`, `ParallelGroupByFuzzTest` (including the factory-reuse case), and `ParallelTopKFuzzTest` - all green.

## Also in this PR: fix flaky CheckpointTest drain test

This branch also stabilizes `CheckpointTest.testCheckpointRestoreDrainsParallelTasksOnTaskFailure`, which reddened the `mac-griffin` CI shard on this PR. The failure is unrelated to the memory-footprint work above; it is bundled here because it is what makes this branch's CI go green.

The test asserts both that the snapshot-restore drain does not abandon in-flight parallel tasks (`started == finished`) and that the drain actually waited for one (`finished > 0`). The second assertion only holds if a `sym_slow` bitmap-rebuild task is in flight at the instant `sym_fail`'s rebuild fails. Nothing guaranteed that: `rebuildTableFiles()` flips the abort flag on the first task failure, and any `sym_slow` task that has not yet passed its top-of-task abort check then short-circuits without opening `sym_slow.d`. Under scheduling pressure every `sym_slow` task short-circuited, so `finished` stayed 0 and the assertion failed. The engine behavior is correct; the test over-asserted.

The fix gates the `sym_fail.d` open so the rebuild fails only after at least one `sym_slow` task has started (is parked in its sleep). The drain then has a genuine in-flight task to await, so `started == finished` and `finished > 0` hold deterministically on any recovery-pool size and under load. The gate wait is bounded, so it cannot hang the test. Net change vs master is the gate plus a comment update; the first commit's single-thread + buffer-column harness is a throwaway deterministic reproduction that the fix commit reverts.

- Pre-fix reproduction harness failed 15/15 with the exact CI assertion.
- Post-fix: 20/20 green on an idle box, 10/10 green with all cores saturated.
- Full `CheckpointTest` (107 tests) re-run green.


puzpuzpuz:
## Code review

A thorough adversarial pass over this change (correctness, concurrency, resource lifecycle, performance, tests, standards), verifying each concern against source rather than the diff alone.

**Verdict: LGTM, approve with non-blocking suggestions.** The change is correct, concurrency-safe, leak-free, and optimal for its cold (per-query-teardown) path, and it faithfully mirrors the in-tree `PageFrameReduceTask.clear()` precedent.

The core safety property — that `clear()` reallocs these lists only after all reduce tasks have finished — is **control-flow guaranteed** in production, not merely asserted: every cursor calls `frameSequence.await()` immediately before `reset()`, and `await()` spins on a `volatile`-backed `doneLatch` / `reduceFinishedCounter` that each worker decrements as its last action (happens-before edge). The blast radius is identical to the pre-existing `Misc.free(ownerMemoryPool)` already in `clear()`, so no new hazard is introduced.

### Moderate

**1. No end-to-end test exercises the real teardown path; the regression is invisible to existing leak machinery.**
The two new tests construct `AsyncFilterContext` directly and call `clear()`. They never go through the production wiring — `PageFrameSequence.reset()` / `UnorderedPageFrameSequence.reset()` -> `atom.clear()` -> `filterCtx.clear()` — for any of the 5 atom types (`AsyncGroupByAtom`, `AsyncGroupByNotKeyedAtom`, `AsyncTopKAtom`, `BaseAsyncHorizonJoinAtom`, `BaseAsyncMultiHorizonJoinAtom`). This matters more than usual:

- `getMemUsedByFactories()` **excludes** `NATIVE_OFFLOAD` (`AbstractCairoTest.java:247`), so the standard per-query factory-leak check is blind to this tag.
- End-of-test full-leak checks only fire after factory `close()`, which frees these lists regardless of the fix.

Net: `ParallelGroupByFuzzTest`, `ParallelTopKFuzzTest`, and `HorizonJoinTest` structurally cannot catch a regression where an idle/cached factory keeps peak buffers. If a future change drops a `filterCtx.clear()` call from one atom or alters the `reset()` path, nothing fails. The shipped code is correct and the wiring was verified by hand, so this is purely future-regression protection — non-blocking. Suggested addition: one integration test that runs a JIT-filtered parallel GROUP BY / TOP K / horizon join, closes the cursor while keeping the factory cached, and asserts `getMemUsedByTag(NATIVE_OFFLOAD)` subsided.

### Minor

**2. The `clear()` comment slightly over-claims the precedent** (`AsyncFilterContext.java:139-142`). "...as `PageFrameReduceTask.clear()` does" — `PageFrameReduceTask` has no *per-worker* lists; it resets only its three owner-equivalent lists, whereas the new code also resets per-worker lists. Consider "...mirroring the per-task reset in `PageFrameReduceTask.clear()`."

**3. `testClearShrinksGrownColumnAddressLists` uses an unrealistic growth size** (`AsyncFilterContextTest.java:70-75`). It grows the data/aux address lists to `setCapacity(100_000)`, but in production those lists only ever grow to `columnCount` entries (one `add()` per column in `populateJitAddresses`) — tens of longs, never frame-sized. Only the row-id list reaches ~8 MB, so `testClearShrinksGrownRowIdLists` is the load-bearing test. The column test is still mechanically valid (`resetCapacity` is capacity-agnostic), but a realistic size or a clarifying comment would avoid implying the column-address lists are a major consumer.

**4. Redundant shrink-then-free at TopK eviction** (`AsyncTopKAtom.java:204-208`). `close()` calls `clear()` (now reallocs the 6 lists down to initial) and then `Misc.free(filterCtx)` (frees them) — the shrink is wasted work undone one line later. Negligible (once per factory eviction, cold path) and the `close()`-delegates-to-`clear()` structure predates this PR; `AsyncGroupByAtom.close()` avoids it by freeing `filterCtx` directly. Informational only.

**5. Untested edge cases** (low value). Idempotent double-`clear()` and `slotCount == 0` aren't tested. Both are provably safe — the `if (this.capacity != capacity)` no-op guard makes a second `clear()` allocation-free, and empty per-worker `ObjList`s iterate zero times — so this is optional; a one-line double-`clear()` assertion would cheaply lock in reuse safety.

### Verified safe (concerns investigated and dismissed)

- **Use-after-free / data race from realloc while workers run** — `clear()` is reached only via `reset()`, which every cursor calls strictly after `await()` drains all tasks through the `volatile doneLatch` / `reduceFinishedCounter` happens-before edge. Guaranteed by control flow, not just the `-ea` assert.
- **Residual `pos`/`size()` after shrink corrupts the next read** — every consumer calls `rows.clear()` / `dataAddresses.clear()` before reading, and `applyCompiledFilter` re-sizes `filteredRows` to `frameRowCount`; no path reads stale size.
- **8 MB memcpy on every teardown** — `realloc` to a 2 KB target cannot copy more than 2 KB; `setCapacityBytes` adds no explicit copy. Cost is malloc-small + free-big.
- **Leak / double-free / tag mis-accounting** — the `capacity` field always tracks the live allocation, net `NATIVE_OFFLOAD` delta over the lifecycle is 0, `close()` is `if (address != 0)`-guarded, and the tag is immutable across realloc.
- **NPE on null/empty data-aux lists (non-JIT path)** — scalar helper null-guards; `ObjList` helper iterates `size()` (0 for the never-populated per-worker lists).
- **Member-ordering / style** — the two `private static` helpers after `toPlan()` are consistent with `AsyncGroupByAtom` / `PageFrameReduceTask` / `AsyncGroupByRecordCursorFactory`.
- **Realloc churn / doubling chain on regrow** — `filteredRows` regrows via a single `setCapacity(frameRowCount)` jump-to-target, not a doubling chain; address lists double only from 16 -> `columnCount` on tiny buffers, once per query.

### Summary

The change does what it claims with 100% line coverage on the new code, and the description's stated tradeoff (one reallocation when a cached factory is reused on a large scan) is the same one the shared reduce pool already makes. The only item worth acting on before merge is the Moderate test-coverage gap, and even that is non-blocking since the shipped code is verified correct — it only protects the fix from future regressions that `NATIVE_OFFLOAD`'s exclusion from the leak check would otherwise hide.

mtopolnik:
# Review of PR #7267

**Level 2.** One production method changed (`AsyncFilterContext.clear()` + two private helpers), one new test file. Note: the PR was updated *during* this review (commit `aa8493c8b2` at 09:36Z added the column-address test and tightened the comment); findings below reflect the **current** head.

The core change is correct and well-targeted in isolation. But the new `resetCapacity()` breaks an implicit contract that unchanged caller code relies on, producing a native-memory leak on an error path. That is the blocking finding.

## Critical

### 1. `resetCapacity()` resurrects already-freed `DirectLongList`s → `NATIVE_OFFLOAD` leak on the horizon/ASOF-window-join error-teardown path  *(out-of-diff)*

**Mechanism (in-diff):** `AsyncFilterContext.clear()` now calls `resetCapacity()` on six `DirectLongList`s. `DirectLongList.resetCapacity()` → `setCapacityBytes(initialCapacity)` (`DirectLongList.java:148-150`), and since a *closed* list has `capacity == 0 != initialCapacity`, the guard at `DirectLongList.java:226` passes and it executes `Unsafe.realloc(0, 0, initialCapacity, NATIVE_OFFLOAD)` — i.e. a fresh `malloc` on an already-closed list. Nothing ever frees it (the owning `filterCtx` is already closed and won't close again).

Pre-PR, `clear()` never touched these lists, so calling `clear()` after `close()` was a harmless no-op. **The PR converts that benign re-entrancy into a leak.** The violated contract (2.5c): *`AsyncFilterContext.clear()` must be safe to invoke after `close()`* — it no longer is.

**Reachable trigger (out-of-diff, all 4 horizon factories):**
The 4 join factories free `frameSequence` **before** `cursor` in `_close()` (e.g. `AsyncHorizonJoinRecordCursorFactory.java:593-594`; same in `AsyncHorizonJoinNotKeyedRecordCursorFactory`, `AsyncMultiHorizonJoinRecordCursorFactory`, `AsyncMultiHorizonJoinNotKeyedRecordCursorFactory`). The 3 GROUP BY/TopK factories free `cursor` first and are **safe**.

1. `getCursor()` (`AsyncHorizonJoinRecordCursorFactory.java:208-211`, no try/catch) calls `cursor.of(frameSequence, ctx)`.
2. `cursor.of()` assigns `this.frameSequence = frameSequence` (`AsyncHorizonJoinRecordCursor.java:253`) while `isOpen == true`, then calls the **throwing** `slaveFactory.getPageFrameCursor(...)` (line 258) or `Function.init(...)` (line 264). A throw there leaves the cursor **half-open**: `isOpen == true`, `frameSequence != null`.
3. The exception propagates out of `getCursor()`; the caller closes the **factory** → `_close()`:
   - `Misc.free(frameSequence)` → `UnorderedPageFrameSequence.close()` (`:170-174`): `reset()` (→ `clear()` on **valid** lists, fine) then `Misc.free(atom)` → `BaseAsyncHorizonJoinAtom.close()` → `Misc.free(filterCtx)` (`:340`) → `AsyncFilterContext.close()` **frees all six lists** (address=0).
   - `Misc.free(cursor)` → `cursor.close()` (`:103-118`): `isOpen==true && frameSequence!=null` → `frameSequence.await()` (no-op) → `frameSequence.reset()` → `atom.clear()` → `filterCtx.clear()` → `resetCapacity()` on the **freed** lists → re-`malloc` → **leak**.

Every link verified against source. `filterCtx` is `final` and `clear()`'s pre-`resetCapacity` statements are idempotent `Misc.*` calls, so control reaches the resurrection.

**Severity / blast radius:** Per failed query the leak is bounded — the lists are re-malloc'd to *initial* capacity (256 longs ≈ 2 KB each for row-id, 16 longs ≈ 128 B for column), so ~tens of KB for `(1 + workerCount)` lists, not the 8 MB peak. But it is unbounded across repeated failures and **will trip `assertMemoryLeak`** in any test that exercises a post-`frameSequence`-init `getCursor()` failure on these join types. That no existing test catches it is itself the coverage gap (see Moderate #2). For mission-critical zero-leak code this blocks.

**Fix (primary, minimal, robust):** make the helper skip already-closed lists, mirroring `DirectLongList.reopen()`'s own `if (address == 0)` self-guard:
```java
private static void resetCapacity(@Nullable DirectLongList list) {
    if (list != null && list.getCapacity() > 0) { // a closed list has capacity 0; don't resurrect it
        list.resetCapacity();
    }
}
```
This preserves the optimization on the intended teardown path (lists are open there) and neutralizes every clear-after-close path — including the more speculative factory double-`close()` vector (`AbstractRecordCursorFactory.close()` has no idempotency guard), which would hit the same resurrection even on a GROUP BY factory.

**Alternative (also fixes this specific path):** reorder the 4 horizon `_close()` bodies to `Misc.free(cursor)` before `Misc.free(frameSequence)`, matching the GROUP BY/TopK factories. Less robust than the guard (doesn't cover double-close) and touches 4 files.

## Moderate

### 2. No test covers the error/clear-after-close path
Both new tests drive `clear()` only on the happy path (construct → grow → `clear()`). Nothing exercises `clear()` after `close()`, nor a failed `getCursor()` on a horizon join — exactly the path that leaks. Existing suites (`HorizonJoinTest`, `ParallelGroupByFuzzTest`, `ParallelTopKFuzzTest`) execute the new code but assert only query correctness; the regression this PR *fixes* (idle pinning) is invisible to leak checks, and the regression it *introduces* (Critical #1) has no covering test. Add a test that grows a context, calls `close()` then `clear()` (and ideally a horizon-join `getCursor()` forced to fail after `frameSequence` init), asserting no `NATIVE_OFFLOAD` leak.

### 3. Steady-state-large cached workloads pay a free+remalloc of peak buffers every execution
`clear()` runs once per cursor teardown (verified: only reachable via `reset()`, which `toTop()` does *not* call). That's correct and off the per-row path. But shrink-to-initial with no hysteresis means a cached factory that **repeatedly** runs large JIT-filtered scans frees `(1 + workerCount)` row-id buffers (up to ~8 MB each) at every teardown and re-`malloc`s + faults them back in on the next run — `O(workerCount)` allocator calls and peak-byte fault-in per execution. The PR description frames this as "the same tradeoff the shared reduce pool already makes," but `PageFrameReduceTask` deliberately uses a high-water **hysteresis** — it only shrinks for the last `frameQueueCapacity` frames ("we let these sizes stick until producer starts to wind down", `PageFrameReduceTask.java:334-337`). The two policies are not the same. The chosen policy is reasonable for the stated *idle-footprint* goal, but the description should state the asymmetry honestly rather than equate the two; if steady-state-large cached latency matters, an idle-gated or high-water shrink would avoid the churn.

## Minor

- **Comment / PR-description accuracy:** the in-code comment "as `PageFrameReduceTask.clear()` does" and the PR body's "same tradeoff the shared reduce pool already makes" both gloss over the hysteresis difference in #3. Tighten the wording.

## Summary

- **Verdict: request changes** — one Critical native-memory leak introduced by the PR, on a reachable error path.
- The optimization itself is sound (shrink runs once per teardown, off the hot path; re-grow is a single exact realloc). The problem is that `resetCapacity()` resurrects closed lists, breaking the "`clear()` is safe after `close()`" contract that the horizon/ASOF-join teardown ordering depends on. One-line guard fixes it.

puzpuzpuz:
Verified the Critical finding end to end against the current head — it is correct and reachable. Confirmed each link:

- `DirectLongList.resetCapacity()` on a closed list (`capacity == 0 != initialCapacity`) passes the guard at `DirectLongList.java:226` and re-mallocs via `Unsafe.realloc(0, 0, initialCapacity)`.
- All 4 horizon-join factories free `frameSequence` before `cursor` in `_close()`; the 3 GROUP BY/TopK ones free `cursor` first and are safe — matches your reading.
- `getCursor()` has no try/catch, and `cursor.of()` assigns `this.frameSequence` (`AsyncHorizonJoinRecordCursor.java:253`) before the throwing `getPageFrameCursor()` / `Function.init()` calls, so a throw leaves the cursor half-open (`isOpen == true`, `frameSequence != null`). `reset()` / `clear()` carry no guards and `atom` / `filterCtx` are `final`, so the second `reset()` from `cursor.close()` reaches the resurrection.

Reproduced it empirically with a direct `close()`-then-`clear()` unit test: 2304 bytes leaked without the fix, clean with it.

One refinement to the blast radius: only the **owner** lists resurrect, not `(1 + workerCount)`. `AsyncFilterContext.close()` frees the per-worker lists via `Misc.freeObjList`, whose `freeObjList0` nulls each slot in place (`setQuick(i, null)`). `clear()`'s per-worker `resetCapacity` loop then sees `null` and skips them. Only the owner row-id + data + aux fields (not nulled by `Misc.free`) get re-malloc'd — ~2.3 KB at the default initial capacity (256 + 16 + 16 longs), and the re-malloc is to initial, not peak, so it is independent of both `workerCount` and `page.frame.max.rows`. Still a real per-failure leak, unbounded across repeated failures, and it trips `assertMemoryLeak`, so the blocking verdict stands.

Fixed with two complementary changes:

- `abbcbb0a9a` — the one-line guard (`getCapacity() > 0`) in `resetCapacity()`, plus `testClearAfterCloseDoesNotResurrectLists` covering the clear-after-close path (Moderate #2). The guard neutralizes any `clear()`-after-`close()`, including the factory double-`close()` vector.
- `dcd6038df7` — the 4 horizon-join factories now free `cursor` before `frameSequence`, matching the GROUP BY/TopK factories (whose `cursor.close()` likewise depends on `frameSequence` being alive). This fixes the structural defect — previously `cursor.close()` ran `await()`/`reset()` against an already-closed `frameSequence` — so `clear()` always runs while the lists are open and can never resurrect freed buffers on the teardown path. `HorizonJoinTest` (165 tests) green after the reorder.

I also checked the rest of the codebase for the same ordering. The 2 `AsyncWindowJoin` factories free `frameSequence` before `cursor` too, but they are unaffected: their atoms hold no `AsyncFilterContext`, and `AsyncWindowJoinRecordCursor.of()` assigns `frameSequence` last with no throwing call after it, so the half-open state is not reachable there. Left unchanged. The 7 `AsyncFilterContext` holders (4 horizon + 3 GROUP BY/TopK) are the complete affected set, and all are now correct.

mtopolnik (dismissed):

puzpuzpuz:
@mtopolnik thanks for the review!

bluestreak01 (approved):
Approve.

Level 3 review (full pass: change-surface map, all agent perspectives, per-finding source verification). Re-verified the whole change from source rather than trusting the thread.

Core safety properties confirmed:
- clear() is reachable only via reset(), which asserts no reduce task in flight; every cursor calls await() before reset(), giving a happens-before edge through the volatile reduceFinishedCounter/doneLatch. The realloc in resetCapacity() cannot race a worker.
- The getCapacity() > 0 guard is load-bearing: only the owner lists (2304 bytes) could resurrect (per-worker slots are nulled in place by Misc.freeObjList); a closed list reports capacity 0 and is skipped. It also neutralizes factory double-close (AbstractRecordCursorFactory.close() has no idempotency guard).
- The cursor-before-frameSequence reorder is correct, consistent across all 4 horizon cursor types (all assign this.frameSequence before the throwing getPageFrameCursor()/Function.init()), and also fixes a separate pre-existing use-after-free where cursor.close() would reset() an already-freed frameSequence. Guard fixes the leak; reorder fixes the UAF; both are present.
- All 5 atom types call filterCtx.clear(), so the optimization fires everywhere.
- Window-join factories correctly left unchanged: their atoms hold no AsyncFilterContext and the half-open cursor state is unreachable (of() assigns masterFrameSequence after both throwing calls).
- CheckpointTest gate is sound: recovery pool floor is 4 threads, so the bounded sym_fail.d gate cannot self-deadlock, and it does not mask the engine bug it tests.

Previously-blocking Critical leak is resolved in-diff and regression-tested (testClearAfterCloseDoesNotResurrectLists). Tests use assertMemoryLeak and the builder API correctly.

Non-blocking minor notes:
1. Description equates the unconditional teardown shrink with the reduce pool's tradeoff, but PageFrameReduceTask.collected() uses high-water hysteresis (PageFrameReduceTask.java:334-338); the policies differ. Impact is limited since these lists are per-query, not per-frame, so churn only hits a cached factory re-running large JIT scans back-to-back. Worth stating the asymmetry honestly.
2. AsyncTopKAtom.close() calls clear() (reallocs the 6 lists down) then frees filterCtx one line later - redundant cold-path work, pre-existing structure.
3. End-to-end teardown test covers only the GROUP BY path; TopK/horizon happy-path wiring relies on the shared clear() plus the direct unit tests.

All findings in-diff; no out-of-diff bugs found.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 17 / 17 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncFilterContext.java|13|13|100.00%|

PR #7268, opened 2026-06-16
chore(core): stop truncating the `_pm` sidecar on a rolled-back partition update

RaphDal:
## Summary

This is a defense-in-depth hardening change, not a customer-reported bug fix, so it lands as a `chore`, not a `fix`. It carries no user-visible behaviour and no `_pm` format change.

The `SIGBUS` (JVM exit 134) it closes was never actually observed, in the wild or in CI. CI did surface a `SIGBUS`, but a different one: the investigation it triggered turned up two independent candidate causes, and the other candidate was the real one (addressed separately). This `_pm` truncate-on-rollback path is the second candidate. It can produce the same crash signature, but only through a narrow race — a transaction rollback truncating `_pm` while a concurrent reader has already mapped past the committed head, before that reader runs the MVCC footer walk that resolves the committed size. The odds are very low (the rollback window must overlap a reader's pre-resolution mmap), but the path is genuinely reachable, so this change closes it.

## Addressing review feedback

This revision folds the review changes onto the same branch (CI shares a lineage, so splitting would only multiply CI cost):

- **Rollback fsync no longer masks a failed update.** The rollback makes the leftover `_pm` tail durable with `ff.fsyncAndClose`, which throws on an fsync fault. That call sat in the rollback `catch` block ahead of `o3BumpErrorCount`, so under a non-`NOSYNC` commit mode an fsync fault during rollback would skip the error count, the worker latch would still be released, and the failed update would commit as a success with no table suspend. The rollback now swallows and logs the `_pm` fsync error so `o3BumpErrorCount` always runs; the fd is still closed (no leak), and a non-durable dead tail is harmless because `_txn` is unchanged and the next update overwrites it.

- **The `data.parquet` truncate no longer masks a failed update either.** The same rollback `catch` opens `data.parquet` via `TableUtils.openRW` to truncate it back to its pre-merge size, and `openRW` throws on an FS fault — sitting ahead of `o3BumpErrorCount` exactly as the `_pm` fsync did, so an FS fault on the data-file truncate would skip the error count and commit the failed update as a success. That IO now lives in the same swallow-and-log guard, so `o3BumpErrorCount` always runs and the table suspends. When `openRW` cannot open the file the data file keeps an uncommitted tail past the committed size; that is harmless because `_txn` is unchanged, so readers ignore it and the retried update overwrites it from the committed base — the property the `_pm` tail already relies on. This is symmetric with the `_pm` fsync guard above and does not change the common rollback path, where the truncate still succeeds and leaves no data-file tail.

- **`resolveFooter(Long.MAX_VALUE)` foot-gun replaced by a dedicated method.** Callers previously passed `Long.MAX_VALUE` to `resolveFooter` to mean "resolve the physically-last footer, skip MVCC matching", which under the dead-footer-tolerant writer can resolve an orphaned footer on a live `_pm`. `resolveFooter(long)` now does MVCC matching only (no sentinel branch), and a separate `resolveLastFooter()` method makes the "physically-last, no matching" intent explicit — safe only when no rolled-back update can have left a dead tail (a freshly staged or read-only `_pm`). The boundary/attach read paths in `TableWriter` derive the committed parquet-data size from `ff.length(data.parquet)` — the same value `_txn` field 3 stores, and the same source the existing `data.parquet`-only branch already used — and call `resolveFooter`, so they walk back to the committed footer; the magic value can no longer be passed to the matching API by mistake.

- **`PartitionUpdater.of`/`create`** rename the misleading `parquetFileSize` parameter to `parquetMetaFileSize` (it carries the `_pm` parse anchor, the Rust `parquet_meta_file_size`), with a javadoc separating it from `appendBase` and `existingParquetFileSize`.

## Problem

A parquet partition keeps two files: `data.parquet` and `_pm` (a QuestDB metadata sidecar — header + column descriptors + row-group blocks + a backward-linked footer chain). An in-place O3 update grows `_pm`, appends a new footer, and patches the header field `parquet_meta_file_size` (offset 0). On a transaction failure (OOM, I/O error, a concurrent drop, …) the rollback restored the old header **and `ff.truncate`'d `_pm` back to its pre-update size**.

Every exposed reader maps `_pm` at the *raw header value*, read before any MVCC matching:

- `ParquetMetaFileReader.openAndMapRO` reads `parquet_meta_file_size` at offset 0 and `mapRO`s that many bytes; `resolveFooter` runs only afterwards.
- `SHOW PARTITIONS` and a plain `SELECT` over a parquet partition both map `_pm` at the header size on a SQL thread with **no writer lock**. A read txn pins the logical MVCC version; it does not stop a raw `ff.truncate`.

So a reader that mapped `[0, M)` at the speculative header `M` and is reading near `M` loses pages `[N, M)` when the rollback truncates to `N` → `SIGBUS`. `data.parquet` is not exposed: its readers size their mmap from the footer-derived committed size, never the raw physical length. `_pm` is the only file whose mmap size comes raw from a mutable header — that asymmetry is the whole bug, and it is why only `_pm` needs the treatment.

## The change

No `_pm` format change, no new footer field, no `_txn` change.

### Never truncate `_pm`; publish the header last

The rollback no longer truncates `_pm` or rewrites its header. Two pieces make that safe:

- **The header patch moved to the end of the update.** `updateFileMetadata` (Rust `end()`) writes the new footer and blocks but leaves the header untouched; a new `commitParquetMeta` step patches `parquet_meta_file_size` at offset 0 — the MVCC commit signal — only *after* `updateParquetIndexes`, as the last `_pm` write before the `_txn` commit. Any failure during the update therefore leaves the header at the committed head `N`, with the speculative footer sitting past it as an orphaned "dead" footer that no reader maps and the MVCC walk never reaches.
- **The next update overwrites the dead footer.** Because the header still reads `N`, the next in-place update writes from `N` and reclaims the dead tail in place. No truncate, no header restore, no torn read for a concurrent reader (which only ever mapped `[0, N)`).

### Append base = the `_pm` header

An in-place update writes at the **append base** = the committed `parquet_meta_file_size` header (offset 0), read native-side from the held fd (the writer holds the table lock, so it is stable). It equals the parse anchor — the `_txn`-resolved committed head — in the common case, so the write is a contiguous overwrite from `N` with a plain cumulative CRC.

The two diverge only in the **crash window**. The header and `_txn` live in separate files and cannot commit atomically, and a reader resolves via `_txn` then walks back from the header, so the header must be patched first. A crash after the header patch (`M`) but before the `_txn` commit (`N`) leaves a durable, uncommitted footer at `[N, M)`. On restart the header is `M`; the next update parses the committed head `N`, appends past `M` (rule: never overwrite `[0, header)`, which a reader may have mapped), and the new footer's cumulative CRC folds in the dead bytes `[N, M)`. This crash window is the only path that appends past a dead footer, and the only reason the CRC-resume-over-dead-bytes machinery exists.

### Read path — validate the resolved footer

The backward walk is unchanged. One change: the once-per-open CRC verification now runs **after** the walk, keyed on the resolved committed size, not before it on the mapped size. The physically-last footer may now be an orphaned dead footer; only the resolved footer is guaranteed committed. A footer's cumulative CRC covers the whole file through its CRC field, so a reader resolving a footer that ends past a dead region naturally re-hashes that region — the writer folds the same bytes in.

### Dead-footer remnant — orphaned, not proactively reclaimed

A dead footer arises only in the crash window. A plain failed update self-heals: the header never advanced past the committed head, so the next update overwrites the unpublished tail in place (above). A crash-window remnant is not overwritten — the next in-place update appends *past* it (it sits inside `[0, header)`), so it stays embedded as orphaned space that no reader's MVCC walk reaches. There is no dedicated `_pm` reclaim trigger: the remnant is dropped whenever a rewrite fires for an existing reason — the data-side unused-bytes ratio/limit, a schema change, or a single-row-group partition — which regenerates a compact `_pm` from scratch. The remnant is bounded per crash and harmless while it persists.

## Files

| File | Change |
| --- | --- |
| `core/rust/qdbr/src/parquet_write/update.rs` | `end()` reads the `_pm` header for the append base and writes the footer there, no longer patching the header. `sync_parquet_meta` becomes `commit_parquet_meta(sync)`: patches the header (publish), then fsyncs when `sync` is set (skipped in NOSYNC). |
| `core/rust/qdbr/src/parquet_metadata/convert.rs` | `update_parquet_metadata` takes an `append_base` (the header), validates it against the parse anchor, and calls `finish_appending_at`. |
| `core/rust/qdb-parquet-meta/src/writer.rs` | `finish()` delegates to `finish_appending_at(append_base)`; new blocks, the new footer, and the CRC carry are positioned at the append base, `prev` stays the parse anchor, and the cumulative CRC folds in any dead tail between them. |
| `core/rust/qdbr/src/parquet_write/jni.rs` | Renames the `syncParquetMeta` JNI entry to `commitParquetMeta` and threads the `sync` flag. |
| `core/src/main/java/io/questdb/cairo/O3PartitionJob.java` | Reorders to `updateFileMetadata` → `updateParquetIndexes` → `commitParquetMeta` → `_txn` commit. Rollback no longer truncates `_pm` or restores its header (keeps the `data.parquet` truncate and the `_pm` fsync). Both rollback IO steps — the `data.parquet` truncate (its `TableUtils.openRW` throws on an FS fault) and the `_pm` fsync — swallow and log their error so `o3BumpErrorCount` always runs and the table suspends. |
| `core/src/main/java/io/questdb/cairo/ParquetMetaFileReader.java` | New `resolvedFileSize` field + getter (the committed head from the walk); CRC verification moved after the walk onto the resolved footer; `resolveFooter` now takes only a committed parquet-data size as the MVCC token (the `Long.MAX_VALUE` "physically-last, no matching" sentinel is removed), and a new `resolveLastFooter()` method resolves the physically-last footer explicitly. The shared footer validation/commit and the footer-offset bound check are factored into private helpers used by both. |
| `core/src/main/java/io/questdb/cairo/TableWriter.java` | The boundary/attach `_pm` read paths (`readPartitionMinMaxTimestamps`, `readPartitionSizeMinMaxTimestamps`) derive the committed parquet-data size from `ff.length(data.parquet)` instead of passing `Long.MAX_VALUE` to `resolveFooter`. |
| `core/src/main/java/io/questdb/griffin/engine/table/parquet/PartitionUpdater.java` | `syncParquetMeta()` becomes `commitParquetMeta(boolean sync)`; the `parquetFileSize` parse-anchor parameter is renamed `parquetMetaFileSize`. |

## Tradeoffs

- A failed or crash-interrupted in-place update now leaves `_pm` **larger** than its committed size (an orphaned dead-footer tail) rather than truncating it back. A plain failed update's tail is overwritten by the very next update; a crash-window tail persists, orphaned and unread, until an unrelated rewrite (the data-side unused-bytes ratio/limit, a schema change, or a single-row-group partition) regenerates a compact `_pm`. The growth is bounded and rare (`ApplyWal2TableJob` suspends the table on a non-OOM `CairoException` and caps OOM retries), but `_pm` is no longer minimal-on-disk immediately after a failure.
- The crash-window append-past makes the new footer's cumulative CRC re-hash the dead bytes between the committed head and the append base. That is a bounded one-time cost on the rare crash-recovery path, not the steady state.
- The header patch now always publishes regardless of commit mode; only the fsync is gated on `commitMode != NOSYNC`, matching the previous `_pm` durability behaviour (NOSYNC still skips the fsync).

## Compatibility / downgrade

No `_pm` format change, no new footer feature bit, no `_txn` change. A footer this code writes is byte-identical in layout to one written today, and in the common case it lands at the same offset (the committed head) as before. Only after a crash do a footer's `prev` and physical position differ, and an orphaned dead footer is never reachable by any reader — old or new — walking the chain. An older binary that still truncates is unaffected; this tree simply never truncates, so an older reader sees a longer file with an orphaned tail it never reaches.

## Testing

New / retargeted coverage:

- `ParquetWriteTest.testUpdateRollbackDoesNotTruncateParquetMeta` (headline): a full-SQL in-place O3 update fails inside `updateParquetIndexes` (the 2nd `data.parquet` `openRO`, after `updateFileMetadata` and before `commitParquetMeta`). Asserts `_pm` is left **grown, not truncated**, the table suspends, the committed snapshot is still readable through the unpatched header (no `SIGBUS`), and after `RESUME WAL` the retry overwrites the dead tail with correct data.
- `ParquetWriteTest.testRollbackFsyncFailureStillSuspends`: under `SYNC` commit mode, the same in-place update fails AND the rollback `_pm` `fsyncAndClose` throws an injected fsync fault. Asserts the table still suspends (the swallowed fsync error does not skip `o3BumpErrorCount`) and `RESUME WAL` recovers. Confirmed to fail without the fix.
- `ParquetWriteTest.testRollbackDataTruncateFailureStillSuspends`: the same in-place update fails AND the rollback's `data.parquet` truncate `openRW` throws an injected fault. Asserts the fault fires once and the table still suspends (the swallowed error does not skip `o3BumpErrorCount`), and `RESUME WAL` recovers. Confirmed to fail without the fix (with the swallow reverted the table is not suspended).
- `PartitionUpdaterTest.testUpdateAppendsPastDeadTail`: end-to-end through `PartitionUpdater` — splices a dead tail and a dirty-ahead header, runs an incremental update, and asserts it appends strictly past the dead tail, with `openAndMapRO` + `resolveFooter` accepting the result (its cumulative CRC spans the tail).
- `ParquetMetaFileReaderTest.testResolveFooterSkipsOrphanedDeadFooter`: the backward walk reaches the new footer then the committed one without reading the orphaned dead footer, the dead footer's version is unreachable, and the resolved-footer CRC (spanning the dead region) validates.
- `ParquetMetaFileReaderTest.testDirtyAheadHeaderResolvesToCommittedHead`: a dirty-ahead header (`M` past the committed footer `N`) drives `openAndMapRO` + `resolveFooter`; `getResolvedFileSize()` reports `N` while `getFileSize()` is `M`, and the file is left un-truncated at `M`.
- `ParquetMetaFileReaderTest.testResolveLastFooterTakesPhysicallyLastFooter`: on that same dirty-ahead `_pm`, `resolveFooter(committedSize)` walks back to the committed footer while `resolveLastFooter()` resolves the physically-last (dead) one, pinning the contract between the two methods.
- `O3ParquetStaleReaderTest.testConcurrentReaderSurvivesFailedInPlaceUpdate`: a reader pinned on the committed snapshot keeps calling `canSkipRowGroup` with no crash across a failed in-place update. With no truncate and no header churn during the failed update, the `SIGBUS` is structurally unreachable; this guards that property.
- Rust (`qdb-parquet-meta` `writer.rs`, `qdbr` `convert.rs`): append-past unit tests representing the crash window — the new footer's `prev` points at the parse anchor, new blocks land at the append base, the committed + dead region is byte-unchanged, and a reader resolving the new size verifies the cumulative CRC (which spans the dead bytes).

After the `Long.MAX_VALUE` removal, every OSS `resolveFooter` caller passes a committed parquet-data size and matches the committed footer through the MVCC walk; no production path can resolve an orphaned dead footer. The boundary/attach callers (ATTACH, replace-commit drop-neighbour recompute) source that size from `ff.length(data.parquet)`, which equals the `_txn` field-3 token and is always present in those contexts (a detached partition is copied with its `data.parquet`; an in-table parquet partition has both files). Test call sites that simply open a freshly built single-footer `_pm` now call `resolveLastFooter()` (the direct replacement for the old `Long.MAX_VALUE`); the dedicated MVCC tests still pass explicit committed sizes to exercise the walk, match, and no-match paths.

Regression (all green on this revision): `ParquetMetaFileReaderTest` (52), `ParquetWriteTest` (40, incl. all three rollback tests), `PartitionUpdaterTest` (10), `ParquetMetaCrcCorruptionTest` (3), `O3ParquetStaleReaderTest` (7), `O3ParquetMergeStrategyTest` (21), `Mig941Test` (18), `AlterTableDetachPartitionTest` (112, the ATTACH/`ff.length` path), `AlterTableConvertPartitionTest` (39, incl. the NOSYNC `_pm`-durability test), `ShowPartitionsTest` (76), `ParquetRowGroupPruningTest` (120), `TimeFrameCursorTest` (42). No Rust source changed, so the Rust crate suites are unaffected.

```bash
cd core/rust/qdbr && cargo fmt && cargo check --all-targets && cargo clippy --all-targets && cargo test --lib
cd ../qdb-parquet-meta && cargo fmt && cargo check --all-targets && cargo clippy --all-targets && cargo test --lib
mvnd -pl core -Dtest=PartitionUpdaterTest,ParquetMetaFileReaderTest,ParquetWriteTest,O3ParquetStaleReaderTest test -P build-rust-library
```


bluestreak01 (changes requested):
@RaphDal — code review (level 3, blocking). The dead-footer / append-past / publish-header-last design is sound and well-tested: I verified the Rust CRC-resume logic and ran the new Rust unit tests green, traced the MVCC walk, confirmed the JNI parameter order matches across the boundary, and confirmed every OSS read path resolves the footer before use. Two issues block merge.

## Critical

### C1 — Enterprise CI is red and the read-path contract change is the prime suspect (out-of-diff, enterprise)

`ParquetMetaFileReader.readPartitionMeta()`, `canSkipRowGroup()`, and `getOrCreateNativeReaderPtr()` now key the native reader on `resolvedFileSize` and guard with `assert resolvedFileSize != 0` (`ParquetMetaFileReader.java:303, 426, 648-650`). `resolvedFileSize` is **0 until `resolveFooter()` runs**, so every caller must now `resolveFooter()` first.

- **OSS is safe** — every OSS caller resolves the footer first (`TableReader.java:1396`, `TableSnapshotRestore.java:775/956`, `O3PartitionJob.java:143`, `ShowPartitionsRecordCursorFactory.java:415`, `ParquetPartitionDecoder.of(...)`), and `of(ParquetMetaFileReader other)` now copies `resolvedFileSize` (`ParquetMetaFileReader.java:624`).
- **Enterprise is the exposure.** The PR's own test comment (`ParquetMetaFileReaderTest.java:1422-1423`) names the only `readPartitionMeta` caller: `StoragePolicyJob.readParquetMetaSidecar`, "invokes after openAndMapRO." `StoragePolicyJob` is not in OSS (enterprise/tiered storage). Before this PR `readPartitionMeta` used `fileSize` (set by `openAndMapRO`), so `openAndMapRO` → `readPartitionMeta` worked **without** `resolveFooter`. After this PR it asserts — and QuestDB ships with `-ea` (`core/pom.xml:36`, `docker-entrypoint.sh:124`, `systemd.service:25`), so it throws `AssertionError` in production and CI.

The enterprise CI bot reports failure on the exact tested commit `8431d32f` (build 242696). This red CI plus a concrete, identified contract break is blocking. Either the enterprise caller adds a `resolveFooter()` before `readPartitionMeta()`, or `readPartitionMeta` resolves internally / falls back. Please root-cause the enterprise CI failure before merge rather than dismissing it.

### C2 — Rollback `fsyncAndClose` introduces a throw that skips `o3BumpErrorCount`, turning a failed in-place update into a silent commit (in-diff)

`O3PartitionJob.java:679` replaced the old non-throwing `ff.fsync(fd); ff.close(fd)` with `ff.fsyncAndClose(fd)`, which **throws `CairoException` on fsync failure** (`FilesFacadeImpl.java:184-192`). This call sits in the outer rollback `catch (Throwable th)` block, and `tableWriter.o3BumpErrorCount(...)` is the *last* statement of that block (`O3PartitionJob.java:686`), after the throw site, with no protecting `finally`.

Failure trace (commit mode `SYNC`/`ASYNC` — not the default `NOSYNC`):
1. In-place parquet O3 update fails → outer `catch (th)` runs.
2. Rollback truncates `data.parquet`, reopens `_pm`, calls `ff.fsyncAndClose(fd)` → fsync errors → **throws**.
3. `o3BumpErrorCount(th)` (`:686`) is **skipped**; `o3ErrorCount` stays 0.
4. The outer `finally` still runs `o3CountDownDoneLatch()` / `o3ClockDownPartitionUpdateCount()` (`:714-715`), releasing the main thread.
5. The escaping exception is swallowed: `o3ProcessPartitionSafe` only logs (`TableWriter.java:8588-8593`); the worker path (`AbstractQueueConsumerJob.run` → `doRun`) has no error-count hook either.
6. The committing thread sees `o3ErrorCount == 0`, treats the partition as successful (`TableWriter.java:9513-9519`), and commits `_txn` with the partition-update sink written in the `finally` (post-truncate file size / pre-failure row count). **No table suspend; the failed update is committed as success with inconsistent partition metadata.**

This is a regression: on `master` the rollback fsync return was ignored (never threw), so `o3BumpErrorCount` always ran and the table suspended. The PR comment at `:677-678` reasons only about fd leaks (correct on that point) but misses that the throw escapes the recovery block. Reachability is bounded (non-default `SYNC`/`ASYNC` *and* an fsync fault during rollback), but `SYNC` is what durability-conscious deployments run, and an fsync fault is plausible exactly when the original update failed for I/O reasons.

Suggested fix: keep `o3BumpErrorCount` reachable — move it into a `finally`, or swallow/log the rollback fsync error (as `master` did) while still closing the fd.

## Moderate

- **M1 — The C2 regression is untested.** `testUpdateRollbackDoesNotTruncateParquetMeta` and `testConcurrentReaderSurvivesFailedInPlaceUpdate` inject `openRO` failures, where the rollback fsync *succeeds*, so `o3BumpErrorCount` runs and the table suspends (the tests assert suspension). Add a `FilesFacade` that fails fsync during rollback under `SYNC` mode and assert the table still suspends.
- **M2 — `resolveFooter(Long.MAX_VALUE)` is now a latent foot-gun with only a javadoc guard** (`ParquetMetaFileReader.java:670-679`). No OSS caller passes `Long.MAX_VALUE` (verified — only tests do), so not a current bug, but it now silently resolves an orphaned dead footer on a live `_pm`, enforced only by a comment. Consider an explicit guard or a separate, clearly-named "no MVCC matching" entry point.

## Minor

- **Misleading Java parameter name.** `PartitionUpdater.of(...)`/`create(...)` still name the parse-anchor argument `parquetFileSize` (`PartitionUpdater.java:128, 244`) even though O3 passes `getResolvedFileSize()` and the Rust side was clarified to `parquet_meta_file_size`. Pre-existing, but this signature was edited here — a rename to `parseAnchor`/`existingParquetMetaFileSize` would match the Rust naming.
- `Storage`/`WAL` labels would also fit (this is the O3 commit/rollback path).

- "`createNativeReader` switching `fileSize` → `resolvedFileSize` breaks stale readers" — false; it's a correctness *improvement* (keys the native reader on the resolved snapshot `N` consistently with `rowGroupCount`; mmap `[0, M) ⊇ [0, N)` so in-bounds).
- "Post-walk validation bounds-checks against mapped `M` not resolved `N`" — pre-existing (intermediate MVCC footers), backed by the now-resolved-footer CRC check.
- "`createNativeReader(addr, 0)` reads garbage if `resolveFooter` skipped" — guarded by assert + `-ea`; all OSS callers resolve first (enterprise case is C1).
- "`vec![0u8; append_base]` + `read_exact(append_base)` reads past EOF" — false; `openAndMapRO` mapped `append_base` bytes, so physical ≥ append_base.
- "`append_base < parse_anchor` panics via JNI" — false; guarded in `update_parquet_metadata` (returns `InvalidValue`), and the walk guarantees `resolved ≤ header`.
- "CRC resume over the dead tail is wrong" — false; verified by the new Rust unit tests (ran green) and the `[HEADER_CRC_AREA_OFF, new_crc_field)` contiguity.
- `existing_parquet_file_size` gate rename — verified semantically identical to `master` (both gate on the parquet *data* file size); only the misleading field name was fixed.

## Summary

Requesting changes. Findings verified vs dropped: **6 confirmed** (2 Critical, 2 Moderate, 2 Minor) vs **7 draft hypotheses dropped** as false positives. In-diff / out-of-diff: **5 in-diff, 1 out-of-diff** (C1). Tradeoffs are honestly disclosed in the PR body; the main un-disclosed one is C2 — the rollback now propagates an exception it previously suppressed.

<sub>Automated level-3 review.</sub>

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 270 / 287 (94.08%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|3|6|50.00%|
|:large_blue_circle:|qdbr/src/parquet_write/jni.rs|7|9|77.78%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|8|9|88.89%|
|:large_blue_circle:|qdb-parquet-meta/src/writer.rs|58|62|93.55%|
|:large_blue_circle:|qdbr/src/parquet_metadata/convert.rs|124|130|95.38%|
|:large_blue_circle:|qdbr/src/parquet_write/update.rs|28|29|96.55%|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileReader.java|39|39|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionUpdater.java|3|3|100.00%|

RaphDal:
@bluestreak01 - addressed your feedback and made another review:

# Code Review — PR #7268 (level 3)

**`chore(core): stop truncating the `_pm` sidecar on a rolled-back partition update`**

Reviewed at branch HEAD `474734ce45`.

Eleven review roles ran across six parallel agents (correctness, cross-context
callers, concurrency, resource, Rust-safety, tests, fresh-context adversarial,
adversarial-performance) plus independent tracing of the load-bearing
invariants. Every draft finding was re-verified against source; both Rust crates
build and their unit suites run green (134 + 70); `core` compiles clean offline.

The design — never truncate `_pm`, publish the header last via
`commit_parquet_meta`, append past an orphaned dead footer with a cumulative CRC
that folds in the dead bytes, and resolve + CRC-verify the *committed* footer
rather than the physically-last one — is **correct and well-tested**. The
rollback-vs-reader `SIGBUS` race it targets is genuinely closed, and the prior
blocking item is resolved.

## Critical

**None.** No correctness, concurrency, resource, Rust-safety, or test-validity
defect blocks merge.

The prior automated review's in-diff Critical — **C2** (rollback
`fsyncAndClose`/`openRW` throwing past `o3BumpErrorCount`, committing a failed
update as success) — is **resolved and verified**. Both rollback IO steps now
swallow-and-log (`O3PartitionJob.java:657-670` data-truncate, `:684-697` `_pm`
fsync); `o3BumpErrorCount` (`:701`) sits outside both. The consumer side
confirms it: `o3ConsumePartitionUpdateSink()` runs only under
`o3ErrorCount.get() == 0 && success` (`TableWriter.java:9518`), and on a bumped
error count the commit instead throws `"bulk update failed and will be rolled
back"` (`:9522`). `testRollbackFsyncFailureStillSuspends` and
`testRollbackDataTruncateFailureStillSuspends` assert the fault fired *and* the
table suspended, and fail without the swallow.

## Moderate

**None.**

## Minor

- **`resolveLastFooter()` has no production caller**
  (`ParquetMetaFileReader.java:714`; grep confirms test / migration-readback
  only). Correct today, but a future caller invoking it on a live `_pm` with a
  dead footer would silently CRC-accept rolled-back row groups. Consider
  `@TestOnly` / package-private or a guard. (in-diff)
- **Redundant header re-patch on the full-create path** (`update.rs`
  `commit_parquet_meta`). The full-create branch already writes a correct header
  in `finish()` (`writer.rs:248-250`); `commit_parquet_meta` then
  unconditionally re-seeks offset 0 and rewrites the identical 8 bytes. Harmless
  idempotent double-write, micro-cost on the cold full-rewrite / first-create
  path. Optional: gate the patch to the incremental branch. (in-diff)
- **`validateAndCommitFooter` bounds the resolved footer's column-name reads
  against `this.fileSize` (mapped `M`), not `currentSize` (resolved `N`)** (the
  `nameOffset > parquetMetaFileSize` check). When the walk settles on an older
  footer, an offset in `(N, M]` passes the bound. Pre-existing (master used
  `this.fileSize`), within the mapping (no `SIGBUS`), and the resolved footer's
  CRC covers `[0, N)` so a CRC-valid committed footer can't carry such a
  pointer — but a one-line intent comment would help. (pre-existing/out-of-diff)
- **Test coverage gaps (low):** (a) no test exercises `commitParquetMeta(false)`
  / `commit_parquet_meta(sync=false)` — the NOSYNC header-patch-without-fsync
  path (only the `true` path is covered; the sole delta is a skipped
  `sync_data`); (b) `testUpdateAppendsPastDeadTail`'s header comment frames the
  spliced tail as an MVCC-"orphaned dead footer" but the body splices garbage and
  asserts only append-past + resolves — the real orphaning assertions live in the
  Rust `update_appends_past_dead_footer` and reader
  `testResolveFooterSkipsOrphanedDeadFooter`; soften the comment.
- **Non-ASCII in a few *new* test comments** — em dashes (`—`) at
  `ParquetMetaFileReaderTest.java:741,805` and `→` in `ParquetWriteTest.java`.
  These are comments, not `LOG` strings, so not a strict CLAUDE.md violation, but
  QuestDB prefers ASCII (`--`, `->`); the new *production* comments correctly use
  `--`.
- **`commitParquetMeta(boolean sync)` parameter** (`PartitionUpdater.java:80,223`)
  — CLAUDE.md asks for an `is.../has...` prefix on booleans; `isSync`/`shouldSync`
  would conform. Trivial.

## Downgraded (draft hypotheses dismissed after source verification)

- "`createNativeReader` switching `fileSize`→`resolvedFileSize` breaks pruning /
  stale readers" — false; it's a correctness *improvement* (keys the native
  reader on the resolved committed head `N`, consistent with `rowGroupCount`;
  mmap `[0,M) ⊇ [0,N)`), guarded by `testDirtyAheadCanSkipRowGroupUsesCommittedFooter`.
- "`assert resolvedFileSize != 0` throws in production / reads garbage under
  `-da`" — false; every OSS prune/decode caller resolves first, the skip loop is
  gated by `rowGroupCount` (set *with* `resolvedFileSize`), and under `-da` the
  Rust `from_file_size(_,0)` returns `Err` (no panic/abort).
- "`vec![0u8; append_base]` + `read_exact` reads past EOF" / "`append_base <
  parse_anchor` panics via JNI" — false; `openAndMapRO` already bounds
  `append_base ≤ physical length`, and `convert.rs:587-604` guards the range
  before any slicing (`update_rejects_invalid_append_base` covers it).
- "CRC resume over the dead tail is wrong / off-by-one" — false;
  `[old_crc_field, append_base)` + new bytes equals exactly what
  `verify_checksum` re-hashes over `[HEADER_CRC_AREA_OFF, resolved_crc_field)`;
  verified by Rust units (green) and independent traces.
- "Removed `let existing = reader.data();` breaks reused-offset math" — false;
  all offset reads are driven by the parse anchor / committed footer length, not
  `existing.len()`.
- "Write/publish reordering lets a reader see the new header but not the appended
  footer" — false; the footer load is address-dependent on the header value (no
  load-load reorder on x86-64/aarch64), same-fd write ordering + MAP_SHARED
  holds, and CRC-on-resolved-footer is the backstop.
- "Resource leak / fd double-close on rollback" — false; `fsyncAndClose`
  closes-before-throw (`FilesFacadeImpl.java:184-192`), the Rust `_pm` fd is
  dropped by `partitionUpdater.close()` before Java reopens by path, and the
  `finally` munmaps once after `clear()`.
- "`commit_parquet_meta` writes `u64::MAX` when size is `-1`" — false; only
  called after a successful `updateFileMetadata()` in the same try; an `end()`
  failure jumps to the catch.
- "`ff.length` token resolves the dead footer in the dirty-ahead case" — false;
  it resolves the *committed* footer (a fix for master's `Long.MAX_VALUE` latent
  bug).

## Non-defect observations (acceptable, correctly disclosed)

- The crash-window dead region `[N, M)` in `_pm` is **not** counted toward the
  `unusedBytes` rewrite threshold (that threshold tracks `data.parquet` dead
  space). It is cleared only when a full `_pm` rewrite fires for an unrelated
  reason. This is exactly what the PR's Tradeoffs section discloses; bounded per
  crash, harmless while it persists.
- Under NOSYNC, a power-loss crash between the header patch and the `_txn` commit
  can leave the header durable while the appended footer bytes are not
  (`header > physical` → `_pm` unreadable until regeneration). This exposure is
  **pre-existing** — master had the same header-last ordering and the same NOSYNC
  non-durability — and NOSYNC explicitly trades durability for speed. Not changed
  by this PR.

## PR metadata

- Title is valid Conventional Commits; **`chore` is correct** (defense-in-depth,
  never observed in the wild — not a customer bug). Labels
  `Core`/`WAL`/`rust`/`storage` fit. Tradeoffs section is honest and
  level-headed. No `Fixes #` needed.

## Summary

**Verdict: approve — non-blocking suggestions only.** The headline design is
correct; the rollback-vs-reader `SIGBUS` race is closed (no `_pm` truncate
remains on any concurrent path); the read path resolves and CRC-verifies the
committed footer on every OSS callsite; and the prior blocking C2 is fixed and
guarded by tests that fail without it.

- **Findings:** ~20 draft hypotheses across six agents → **0 Critical, 0
  Moderate, 6 Minor** (all trivial/optional); the rest dropped as false
  positives.
- **Cross-context:** all eight OSS `resolveFooter` callsites verified; confirmed
  on `master` that only the two boundary callers ever used the removed
  `Long.MAX_VALUE` sentinel, so the change is a no-op for the rest.
- **Scope note:** scoped to OSS. The `readPartitionMeta` /
  `assert resolvedFileSize != 0` contract (resolve before use) is satisfied by
  every OSS caller; `readPartitionMeta` has no OSS caller.
- **Load-bearing properties** all hold: no `_pm` truncate on any concurrent
  path, CRC-resume over the dead tail, publish-header-last ordering, and
  C2/`o3BumpErrorCount` reachability.

bluestreak01 (dismissed):
Approve — level-3 review, non-blocking suggestions only.

Ran the full level-3 pass (change-surface map, six parallel fresh-context review agents, per-finding source verification). The design is correct and well-tested: never truncate `_pm`, publish the header last via `commitParquetMeta`, append past an orphaned dead footer with a cumulative CRC that folds in the dead bytes, and resolve + CRC-verify the committed footer rather than the physically-last one. The rollback-vs-reader SIGBUS race is genuinely closed (the only remaining `_pm` truncate is `Mig941.java:212`, under the exclusive lock).

Both prior blocking criticals are resolved and verified:
- C1 (enterprise CI red / `readPartitionMeta` `assert resolvedFileSize != 0`): current HEAD `474734ce` is fully green including enterprise-ci (buildId 243842). Every OSS caller resolves the footer before use.
- C2 (rollback `fsyncAndClose`/`openRW` throwing past `o3BumpErrorCount`): both rollback IO steps swallow-and-log; `o3BumpErrorCount` always runs (`O3PartitionJob.java:701`); consumer at `TableWriter.java:9518` only consumes the sink under `o3ErrorCount==0 && success`. Guarded by tests that fail without the fix.

Findings: 0 Critical, 1 Moderate, 7 Minor — all non-blocking.

Moderate (worth a one-line disclosure, not blocking): the boundary/attach recompute paths (`TableWriter.java:11455,11500`) derive the MVCC token from `ff.length(data.parquet)` instead of the `_txn` committed size. Verified this is equal-or-better than master in every crash state (master used `Long.MAX_VALUE`→last footer), and the only new behavior is a fail-loud `CairoException` on a physically-inconsistent `data.parquet`, reachable only via power-loss/double-fault before WAL self-heal — never a SIGBUS or silent corruption.

Minor (optional): prefer a returning guard over `debug_assert!(new_file_size > 0)` in `commit_parquet_meta`; `resolveLastFooter()` has no production caller (consider `@TestOnly`); `_pm` dead-tail growth is invisible to the data-side rewrite trigger (disclosed in Tradeoffs); redundant 8-byte header re-patch on the full-create path; small coverage gaps (NOSYNC `commitParquetMeta(false)`, stacked dead regions); a few non-ASCII em-dashes in new test comments.

Tradeoffs are honestly disclosed. Nice work.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 280 / 295 (94.92%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|qdbr/src/parquet_write/jni.rs|7|9|77.78%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|13|15|86.67%|
|:large_blue_circle:|qdb-parquet-meta/src/writer.rs|58|62|93.55%|
|:large_blue_circle:|qdbr/src/parquet_metadata/convert.rs|124|130|95.38%|
|:large_blue_circle:|qdbr/src/parquet_write/update.rs|28|29|96.55%|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileReader.java|39|39|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/parquet/PartitionUpdater.java|3|3|100.00%|

bluestreak01 (approved):
Approve — independent level-3 review, non-blocking suggestions only.

Verified against source (not just the diff): both Rust crates build, clippy is clean, and the three new append-past/invalid-base unit tests pass. The shipped linux-x86-64 libquestdbr.so exports commitParquetMeta and no longer exports syncParquetMeta, so the binary matches the JNI rename (no UnsatisfiedLinkError risk).

Design is correct and very well tested: never truncate `_pm`, publish the header last via commitParquetMeta, append past an orphaned dead footer with a cumulative CRC that folds in the dead bytes, and resolve + CRC-verify the committed footer rather than the physically-last one. The rollback-vs-reader SIGBUS race is genuinely closed.

Prior blocking C2 confirmed fixed: both rollback IO steps (data.parquet truncate at O3PartitionJob.java:657-670 and `_pm` fsync at :684-697) swallow-and-log, o3BumpErrorCount (:701) runs unconditionally, and the consumer at TableWriter.java:9527 only consumes the partition-update sink under o3ErrorCount==0 && success (else it throws at :9533). fsyncAndClose closes the fd on its failure path, so no fd leak. Both regression tests assert the fault fired and the table suspended.

Findings: 0 Critical, 1 Moderate (disclosed tradeoff: boundary/attach reads now derive the MVCC token from ff.length(data.parquet) and fail loud instead of silently taking the last footer in a rare double-fault/crash-window state — equal-or-better than master and fail-loud rather than SIGBUS), 4 Minor (resolveLastFooter has no production caller; redundant 8-byte header re-patch on the full-create path; debug_assert vs returning guard in commit_parquet_meta; a few non-ASCII chars in new test comments). All non-blocking.

Title, labels (Core/WAL/rust/storage), and the analytical PR body conform to conventions.

PR #7269, opened 2026-06-16
fix(core): regenerate _pm when restored data.parquet exceeds the committed size

bluestreak01:
**Tandem PR:** https://github.com/questdb/questdb-enterprise/pull/1072 (Enterprise) — must merge together.

## Summary

A snapshot/checkpoint can capture a parquet partition whose on-disk `data.parquet`
is a **later generation** than the committed parquet size recorded in `_txn`. The
restore path trusted the partition's existing `_pm` sidecar in that state and left
it stale, so the first read/merge after restore decodes a column chunk that lies
past the committed size:

```
CairoException: error in ParquetPartitionDecoder.decodeRowGroup:
File out of specification: Column chunk range N..M exceeds data length N
```

On a replica that replays the WAL over the restored partition this is fatal: the
`ApplyWal2TableJob` apply fails and the table is **suspended**, breaking
replication. This is the failure seen in the enterprise
`ReplicationFuzzTest.testFuzzWithSnapshotDroppedTableReplica`.

## How a snapshot captures a torn partition

An in-place O3 update of a parquet partition (`partitionMutates=false`) grows
`data.parquet`/`_pm` to a new generation and only then advances the committed
parquet size in `_txn`. The data write and the `_txn` commit live in separate
files and cannot commit atomically, so a checkpoint that lands in between captures
`data.parquet` at the later generation (size `M`) paired with `_txn` still at the
earlier committed size `N` (`N < M`). For a replica the window is wider still: the
snapshot carries the earlier committed `_txn` while the replica's sequencer is
already ahead, so it **replays** the O3 update over the restored partition.

## Root cause

`TableSnapshotRestore.generateMissingParquetMetaFiles` validates `data.parquet`
against the committed size, then decides whether to trust the existing `_pm`:

```java
if (isParquetMetaResolvable(tablePath, parquetFileSize)) { continue; }
```

`isParquetMetaResolvable` only checks that the sidecar **resolves a footer** at the
committed size. A `_pm` from the later generation still resolves a footer at the
committed head, so the stale sidecar was kept — and then mis-read at the committed
size, where a later-generation column chunk lies past it. The existing guard
already rejects an *undersized* `data.parquet` (`onDiskSize < committed`); it had
no symmetric handling for an *oversized* one (`onDiskSize > committed`), which is
exactly the torn snapshot.

## The change

Only trust an existing `_pm` when `data.parquet` is **exactly** the committed size.
When the file is longer (`onDiskSize > parquetFileSize` — the snapshot captured the
partition mid in-place rewrite), regenerate `_pm` from the committed size, which
the size check above has already validated:

```java
if (onDiskSize == parquetFileSize && isParquetMetaResolvable(tablePath, parquetFileSize)) {
    continue;
}
// otherwise remove + regenerate _pm from the committed size
```

`onDiskSize == committed` is the normal consistent invariant, so regeneration fires
only for genuinely torn partitions; it is correctness-preserving (it rebuilds a
compact `_pm` for the committed footer).

## Testing

- Reproduced deterministically end-to-end in enterprise replication (paired PR):
  a primary snapshot of a parquet partition mid in-place O3 rewrite, restored on a
  replica that replays the update — **RED** (table suspended, `File out of
  specification`) before this change, **GREEN** after.
- `CheckpointTest` (107 tests, incl. the `_pm` regenerate / reject / stale-sidecar
  cases) and `ParquetMetaGenerateTest` — all green.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
+++ b/core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java
@@ -699,21 +699,36 @@ public class TableSnapshotRestore implements QuietCloseable {
 
             tablePath.trimTo(partitionDirLen).concat(TableUtils.PARQUET_METADATA_FILE_NAME).$();
             if (ff.exists(tablePath.$())) {
-                if (isParquetMetaResolvable(tablePath, parquetFileSize)) {
+                // Trust an existing _pm only when data.parquet is EXACTLY the
+                // committed size. When the file is longer (onDiskSize >
+                // parquetFileSize), the snapshot captured the partition mid in-place
+                // O3 rewrite: a later generation appended row groups and a new
+                // footer past the committed point and rewrote _pm to describe that
+                // later generation, while _txn still records the earlier committed
+                // size. resolveFooter() can still resolve a footer at the committed
+                // size, so the stale sidecar would be silently kept and then
+                // mis-read at the committed size -- a column chunk of the later
+                // generation lies past parquetFileSize, surfacing as
+                // "File out of specification" on the first merge/read and suspending
+                // a replica that replays over the restored partition. Regenerate
+                // from data.parquet at the committed size, which the size check
+                // above has already validated.
+                if (onDiskSize == parquetFileSize && isParquetMetaResolvable(tablePath, parquetFileSize)) {
                     tablePath.trimTo(plen);
                     continue;
                 }
                 // The sidecar does not resolve a footer at the committed parquet
-                // size: a stale _pm paired with an in-place regenerated
-                // data.parquet, a torn copy, or a partial file left by a crashed
-                // restore. Trusting it would defer the failure to the first read
-                // of the partition, so remove it and regenerate from
-                // data.parquet, which the committed-size check above has already
-                // validated.
+                // size, or data.parquet is longer than the committed size (a stale
+                // _pm paired with an in-place regenerated data.parquet, a torn
+                // copy, or a partial file left by a crashed restore). Trusting it
+                // would defer the failure to the first read of the partition, so
+                // remove it and regenerate from data.parquet, which the
+                // committed-size check above has already validated.
                 if (!ff.removeQuiet(tablePath.$())) {
                     throw CairoException.critical(ff.errno()).put("cannot remove unresolvable _pm file [path=").put(tablePath).put(']');
                 }
-                LOG.info().$("removed unresolvable _pm of restored parquet partition for regeneration [path=").$(tablePath).I$();
+                LOG.info().$("removed stale/unresolvable _pm of restored parquet partition for regeneration [path=").$(tablePath)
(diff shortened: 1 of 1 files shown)


bluestreak01:
## Code review (tandem with the enterprise regression PR)

The fix is minimal, conservative, and correctness-preserving: `onDiskSize == committed` is the normal consistent invariant, so regeneration fires only for genuinely torn partitions, and it rebuilds `_pm` at the committed size which the `onDiskSize < committed` guard above has already validated. The reorder also skips the `_pm` mmap on the regenerate path (small win). LOG chain is properly closed with `.I$()`. No concurrency concern (the restore loop is single-threaded at startup).

### Verification (decisive)
I drove the enterprise replication reproduction both ways on Linux:

- **Without this fix → RED**, exact failure:
  ```
  ParquetPartitionDecoder.decodeRowGroup: File out of specification:
  Column chunk range 8529..8607 exceeds data length 8529
  -> ApplyWal2TableJob job failed, table suspended
  ```
- **With this fix → GREEN**; the replica replays the O3 merge over the restored partition and matches the primary row-for-row.

Confirmed the root cause mechanism end to end: the gen-2 `_pm` still resolves a footer at the committed size via the MVCC chain walk in `resolveFooter`, which is exactly why the old `isParquetMetaResolvable`-only guard kept the stale sidecar.

### Moderate — coverage gap (worth a focused OSS test)
The 3/3 coverage gate is satisfied **incidentally**: consistent-restore tests cover the new `if`-true path, and `testCheckpointRestoreRegeneratesStalePmSidecar` covers the regenerate LOG line via an *unresolvable* sidecar — a path that was already correct before this PR. The **actual** fixed branch (`_pm` resolves a footer at committed **but** `onDiskSize > committed` → regenerate) is exercised only by the enterprise E2E test, not by any OSS unit test.

Suggest a focused `CheckpointTest` case: do a real in-place O3 insert to grow `data.parquet`/`_pm` to gen-2, then restore the gen-1 `_txn` over it, and assert `_pm` is regenerated and the partition reads at the committed size. That pins the exact branch at the OSS layer so OSS CI guards it against future refactors of `generateMissingParquetMetaFiles`.

### Checked and dropped (false positives)
- *"Regenerating `_pm` from the committed size on an oversized `data.parquet` reads a torn footer"* — no: in-place O3 append leaves bytes `[0, committed)` intact, `ParquetMetadataWriter.generate` reads at the committed size, and the `onDiskSize < committed` guard already validated it. Proven by the GREEN run.
- *"Short-circuiting `isParquetMetaResolvable` could leak the mmap when oversized"* — no: it isn't called at all when oversized; when called it munmaps in its own `finally`.

### Verdict
**Approve.** Optional but recommended: add the focused OSS regression test above.

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableSnapshotRestore.java|3|3|100.00%|

ideoma (approved):

Issue #4, opened 2014-08-18
Parameterize TxListener.onCommit()


bluestreak01:
it would be useful to priovide rowids to onCommit() to the listener knows how much data to process.

Issue #1076, opened 2021-06-01
Calling flush on LineProtoSender sometimes causes io.questdb.network.NetworkError: [10053] send error


sarunas-stasaitis:
**Describe the bug**

```
io.questdb.network.NetworkError: [10053] send error
	at io.questdb.std.ThreadLocal.initialValue(ThreadLocal.java:36)
	at java.base/java.lang.ThreadLocal.setInitialValue(ThreadLocal.java:195)
	at java.base/java.lang.ThreadLocal.get(ThreadLocal.java:172)
	at io.questdb.network.NetworkError.instance(NetworkError.java:47)
	at io.questdb.cutlass.line.tcp.LineTCPProtoSender.sendToSocket(LineTCPProtoSender.java:55)
	at io.questdb.cutlass.line.LineProtoSender.send(LineProtoSender.java:275)
	at io.questdb.cutlass.line.LineProtoSender.flush(LineProtoSender.java:156)
	at lt.energyadvice.qdbtest.MainQuestDB.main(MainQuestDB.java:73)
	at org.codehaus.mojo.exec.ExecJavaMojo$1.run(ExecJavaMojo.java:254)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

**To Reproduce**
```
import io.questdb.cutlass.line.LineProtoSender;
import io.questdb.cutlass.line.tcp.LineTCPProtoSender;
import io.questdb.network.Net;
import org.apache.commons.csv.CSVFormat;
import org.apache.commons.csv.CSVRecord;

import java.io.File;
import java.io.FileInputStream;
import java.io.InputStreamReader;
import java.nio.charset.StandardCharsets;
import java.time.Instant;
import java.time.LocalDateTime;
import java.time.ZoneOffset;
import java.util.concurrent.atomic.AtomicLong;

import static lt.energyadvice.qdbtest.Util.dtf;
import static lt.energyadvice.qdbtest.Util.sleep;

public class MainQuestDB {

    private static final AtomicLong count = new AtomicLong();

    public static void main(String[] args) throws Exception {

        File file = new File("path_to_csv");

        String hostIPv4 = "127.0.0.1";
        int port = 9009;
        int bufferCapacity = 256*1024*1024;

        System.out.println("Connected");

        String dbName = "ksImport" + System.currentTimeMillis();

        Thread t = new Thread(() -> {
            long prev = 0;
            while (true) {
                long l = count.get();
                System.out.println(l - prev);
                prev = l;
[...]

Issue #14, opened 2014-08-29
Messaging Patterns


sirinath:
Is it possible to have messaging pattern abstraction implementation on top of NFSdb?

Issue #13, opened 2014-08-27
Export tool


bluestreak01:
Create utility to export nfsdb to a delimited format, consumable by third-party import tools.

This is more formal to-do item following conversation under issue #11

bluestreak01:
it took a while to do this, but i feel like it is properly done now. Import and export are both very fast, zero-gc and can handle unlimited amount of data.

There are some examples here: 

https://github.com/NFSdb/nfsdb/tree/master/nfsdb-examples/src/main/java/org/nfsdb/examples/io/

sirinath:
Cool

PR #7273, opened 2026-06-17
fix(core): fix the web console showing "admin" instead of the authenticated user

glasstiger:
## Problem

`current_user()` and `session_user()` always returned the hardcoded default `admin`, regardless of which
user actually authenticated. The web console reads `current_user` on page load to show the logged-in
user, so it displayed `admin` for everyone — e.g. a configured `http.user`, or a PGWire user.

The root cause is that the allow-all and read-only security contexts — the ones used when per-user
authorization is not enforced — reported a fixed principal (`Constants.USER_NAME`) from `getPrincipal()`
instead of the authenticated principal.

## Fix

`AllowAllSecurityContext` and `ReadOnlySecurityContext` now carry a `principal` field and expose
`forPrincipal(principal)`, which returns a context that reports the authenticated user. The security
context factories (`AllowAllSecurityContextFactory`, `ReadOnlySecurityContextFactory`,
`ReadOnlyUsersAwareSecurityContextFactory`) pass the authenticated principal through it.

- An anonymous request allocates nothing: `forPrincipal` returns the singleton unchanged for a
  null/empty principal, or one that already matches the default name.
- An authenticated principal takes the derivation path. The HTTP path re-derives the security context on
  every request (PGWire/LineTCP derive once per connection), so derived contexts are cached per principal
  — in a `ConcurrentHashMap` — and after the first request for a given user the HTTP path is
  allocation-free again. Concurrently active principals coexist in the cache instead of evicting one
  another.
- The cache is bounded at 256 principals. The bound is a **soft** one: it is checked before the insert,
  not inside it, so racing inserts can overshoot it by up to the number of concurrent threads. That is
  deliberate — the alternative is a lock or a CAS retry loop on the hot path to enforce a limit whose
  exact value carries no meaning. Past the bound, an unseen principal degrades to allocate-per-call
  rather than growing the map without limit.
- `forPrincipal` is `final` and creates instances via the overridable `newPrincipalContext`. That alone
  does **not** preserve a subclass's runtime type: a subclass of a concrete context that does not
  override `newPrincipalContext` inherits its supertype's implementation and is silently downgraded,
  dropping every `authorize*` and identity override it declared — a context that DENIES an operation
  would start ALLOWING it. An assertion now rejects that, so the requirement is enforced rather than
  merely documented (QuestDB runs with `-ea`). `DenyAllSecurityContext` and the materialized-view refresh
  context override `newPrincipalContext` to return `this`, so they are never re-derived per principal and
  never downgraded to a read-allowing context.
- The `@Transient` principal is copied before it is retained.

## `current_user()` / `session_user()` resolve once per cursor

Both functions read the principal off the `SqlExecutionContext` in `getStrA`/`getStrB`, i.e. **once per
row**. That was already the case before this PR; making the principal a per-connection value rather than
a shared constant makes it worth fixing, because the chain is now longer (a proxy hop and a map lookup on
Enterprise).

They are now `isRuntimeConstant()` functions that snapshot the principal in `init()` — the standard
`now()` idiom. A 1000-row query drops from 1004 to 5 security-context reads for `current_user()` and from
1000 to 1 for `session_user()`. The result is unchanged: the security context cannot change mid-cursor,
and `init()` runs on every `getCursor()`, so a context re-derived between executions of a cached select is
still picked up.

## Testing

- `SecurityContextFactoryPrincipalTest` — factory wiring, the anonymous/default fallback, the
  `@Transient` principal copy (asserted on the factory path as well as on `forPrincipal`, by handing in a
  mutable buffer and then mutating it), the 256-principal soft bound degrading to allocate-per-call, and
  that the deny-all / read-only / settings-read-only restrictions survive the derivation. Runtime-type
  preservation is asserted on the exact runtime class (an `instanceof` assertion cannot detect a
  downgrade, because a downgraded plain `AllowAllSecurityContext` satisfies it), and a subclass that drops
  `newPrincipalContext` is rejected. Concurrency stress tests exercise the per-principal cache: under
  contention every caller reports its own principal, distinct principals are retained without thrashing one
  another, racing first-derivations of the same principal converge on a single cached instance, and no
  context leaks across principals when requests alternate between principals.
- `UserFunctionsTest` — `current_user()` / `session_user()` resolve once per cursor and not per row
  (asserted by scale-invariance: the security-context read count must not grow between a 10-row and a
  10,000-row query), and both refresh on every `getCursor()`, so a cached select does not serve a stale
  principal to the next execution.
- `MatViewTest` — `forPrincipal` does not downgrade the materialized-view refresh context, so its
  view-scoped `authorizeInsert` survives the per-principal derivation.
- `HttpSecurityTest` / `PGSecurityTest` — `current_user()` / `session_user()` reflect the configured HTTP
  and PGWire users (HTTP also covers the bare `current_user` form the web console issues on load), while
  anonymous requests still report the default.

---

Tandem with questdb/questdb-enterprise#1076 (enterprise mirror + submodule bump); this one lands first.


bluestreak01 (dismissed):
Approving — level 3 tandem review (with ENT mirror questdb-enterprise#1076). No in-diff or out-of-diff defects. The lock-free single-entry forPrincipal cache is JMM-correct (final principal + volatile publish + local re-check via Chars.equalsNc), so a racing slot can only cost an allocation, never leak another principal's context; DenyAllSecurityContext.newPrincipalContext returns this, so the 3 == DenyAllSecurityContext.INSTANCE sentinels are safe. The fix is security-improving — it also closes the logout/session-leak path in HttpConnectionContext. Verified locally: SecurityContextFactoryPrincipalTest 19/19 green. Non-blocking: title condition is imprecise ('when authentication is enabled' — the real trigger is authorization-not-enforced AND a non-default user authenticates); add Postgres Wire / REST API / UI / Security labels; drop the unrelated whitespace-only reindentation churn (SqlException, Concurrent*HashMap, WhereClauseParserTest) from a security-sensitive diff. Lands first, before ENT bumps the submodule.

bluestreak01 (dismissed):
Approved — level-3 tandem review passed (with ENT questdb-enterprise#1076). Lock-free per-principal cache is JMM-correct and identity-leak-free, re-parenting blast radius is zero, and forPrincipal is allocation-optimal for its call frequency. New suites green: SecurityContextFactoryPrincipalTest (27), HttpSecurityTest / PGSecurityTest / MatViewTest new methods. This one lands first.

bluestreak01 (changes requested):
@glasstiger

# Review: PR #7273 — `fix(core): display correct username in web console when authentication is enabled`

**Level 3 (full mission-critical pass).** 14-agent equivalent coverage (Agent 8 Rust skipped — no `.rs` files; Agents 12–14 run — test code + fix PR), all findings source-verified, plus direct analysis by the lead.

## What the PR does

`current_user()` / `session_user()` always returned the hardcoded `admin` because the allow-all and read-only security contexts (used when per-user authorization is *not* enforced) reported the fixed `Constants.USER_NAME` from `getPrincipal()`. The fix introduces `AbstractPrincipalAwareSecurityContext`, which carries a `principal` field and a `final forPrincipal(CharSequence)` returning either `this` (null/empty/matching principal) or a per-principal-cached derived context that reports the authenticated user. `AllowAll`/`ReadOnly` are refactored onto extracted `AbstractAllowAll`/`AbstractReadOnly` bases; `DenyAll` is re-parented and pins `newPrincipalContext → this`; the three OSS factories thread the principal through `forPrincipal`.

## Critical

None. No confirmed correctness, data-integrity, resource-safety, or concurrency defect, in-diff, out-of-diff, or pre-existing, across the changed files, their callsite inventory, and the cross-context exposures.

Key adversarial checks that came back clean (all source-verified):

- **Lock-free copy-on-write cache is JMM-correct.** `principalContextCache` is `volatile`; the write path (`synchronized`, `addPrincipalContext`) builds a *brand-new* `CharSequenceObjHashMap` (`putAll(old)+put(new)`, rehash included) and only then executes the volatile store — textbook safe publication. `CharSequenceObjHashMap.get()`/`keyIndex()`/`valueAt()` are pure reads (verified against `AbstractCharSequenceHashSet`: no mutation of `keys`/`values`/`mask`/`free`/`list`), so a lock-free reader can only observe a fully built, immutable map. A racing first-derivation costs at most an extra allocation, never a wrong-principal or null-for-present-key result.
- **No identity-comparison breakage.** Repo-wide grep: zero production `== / != AllowAllSecurityContext.INSTANCE` or `ReadOnlySecurityContext.INSTANCE`, zero `instanceof` of these types. Replacing the singleton with a derived instance breaks no caller. Long-lived retainers (`CopyImportContext`, `TableUpdateDetails.ownSecurityContext`) hold immutable, no-op-`clear()` contexts — safe to share/retain.
- **DenyAll re-parenting is behavior-preserving.** `AbstractReadOnlySecurityContext` holds the old `ReadOnlySecurityContext` method bodies verbatim; DenyAll's own `authorize*` throwing overrides still apply; `newPrincipalContext → this` guarantees deny-all is never downgraded. `DenyAllSecurityContext.INSTANCE` identity (relied on at `HttpConnectionContext:587`, `LineTcpConnectionContext:246,308`, etc.) is preserved because no factory calls `forPrincipal` on DenyAll. `PrincipalOnlySecurityContext` and the 3 test subclasses still compile and behave.
- **`Chars.equalsNc(principal, getPrincipal())` argument order is correct** — `principal` is guaranteed non-null (short-circuited earlier) and is the `@NotNull` arg; a null `getPrincipal()` yields `false` (no NPE), deriving a context. Covered by `testForPrincipalWithNullCurrentPrincipalDoesNotThrow`.
- **`admin` = default *and* anonymous sentinel is not a conflation bug** — the factory selects AllowAll-vs-ReadOnly by name *before* `forPrincipal`, and these contexts are identity-only, so a real user named `admin` and an anonymous request are observationally identical.
- **Read hot path is allocation-free** — anonymous/short-circuit returns `this`; cache hit is a volatile read + content-hash lookup of the flyweight (no `String` materialization). `@Transient` principal is copied via `Chars.toString` only on the cold miss path, before retention.

## Moderate

**M1 — `addPrincipalContext` rebuilds the copy-on-write map at default capacity, forcing repeated rehashing on the write path.** `core/src/main/java/io/questdb/cairo/security/AbstractPrincipalAwareSecurityContext.java:130`

```java
final CharSequenceObjHashMap<SecurityContext> next = new CharSequenceObjHashMap<>();
if (cache != null) {
    next.putAll(cache);   // grows 32 -> 64 -> ... -> 512 slots, ~5 rehashes as it approaches 255 entries
}
```

The no-arg constructor starts at 16 slots (`MIN_INITIAL_CAPACITY`, load factor 0.5 → 32-slot backing array); `putAll` of up to 255 existing entries triggers ~5 `rehash()` doublings each rebuild. Pre-sizing eliminates the intermediate rehashes:

```java
final CharSequenceObjHashMap<SecurityContext> next =
        new CharSequenceObjHashMap<>(cache != null ? cache.size() + 1 : 1);
```

*Severity justification (why Moderate, not Critical):* this is off the per-request read path entirely — it runs only on the arrival of a first-seen authenticated principal, at most `MAX_CACHED_PRINCIPALS` (256) times per singleton over the whole JVM lifetime; total churn is on the order of 256² element-copies, once, spread across process life. It is a genuinely cold path structurally bounded by a small constant, so it cannot affect a result, leak, or cost measurable CPU on any hot path. It is a real (if microsecond-scale), free, zero-risk optimization on the sole allocating path, so per QuestDB's "use the best known approach" standard it should be closed here rather than left sub-optimal.

## Minor

- **First-256-wins cache admission with no eviction (design note).** Once 256 distinct principals are cached, later principals degrade to allocate-per-call permanently. In OSS this is effectively unreachable (static-credential auth yields a small fixed principal set), and the degrade is graceful (bounded per-call cost equal to the pre-cache baseline, no memory growth), so it is not a DoS. Worth one extra sentence in the `forPrincipal` javadoc making the "first-256-wins, no eviction" semantics explicit, rather than only "further principals degrade to allocate-per-call." `AbstractPrincipalAwareSecurityContext.java:37,123-126`
- **Test constant duplication.** `SecurityContextFactoryPrincipalTest.CACHE_CAP = 256` duplicates the `private MAX_CACHED_PRINCIPALS`; the "must match" comment documents the coupling but a future production change would silently mis-size the test. Acceptable (the production constant is private), noted only.
- **Test method ordering.** `SecurityContextFactoryPrincipalTest` has two alphabetical runs (a second batch restarts at `testForPrincipalCapDegradesToAllocatePerCall`), mildly violating the alphabetical-members guideline. JUnit is order-agnostic; no functional impact.
- **Unrelated churn.** `CancelQueryFunctionFactoryTest` only drops an unused `SqlExecutionContext` import (verified genuinely unused). Harmless, but unrelated to the security fix.
- **Pre-existing, unchanged:** `DenyAllSecurityContext.isSystemAdmin()` inherits `true` while every `authorize*` denies — a pre-existing quirk carried over verbatim from the old `ReadOnly` parent, not touched by this PR.

- *"`forPrincipal` on a return-`this` subclass (DenyAll / mat-view) still caches a wasted entry."* — Dismissed: no production code calls `forPrincipal` on those subclasses (factories only call it on the true singletons), so their caches stay `null`; the machinery is dormant, not wasteful.
- *"Cross-request auth bypass via cache reuse."* — Dismissed: `HttpConnectionContext.reset()` restores `DenyAllSecurityContext.INSTANCE` and `configureSecurityContext` re-authenticates every request; a cached context is never reused for a different request without re-auth.
- *"Retaining a factory context in `CopyImportContext` across an async import is unsafe."* — Dismissed: derived contexts are immutable (`final principal`, no-op `clear()`, no mutable state), so retention/sharing is safe; behavior unchanged from the pre-PR singleton.

## Coverage map

15 behavioral changes, 15 tested, 0 UNTESTED (consistent with the reported 162/162 = 100% line coverage). The M1 change is an optimization of already-tested behavior, so it adds no new behavioral row.

| # | Change (symbol / branch) | Test | Failure link | Verdict |
|---|---|---|---|---|
| 1 | `forPrincipal` null/empty/self → `this` | `testForPrincipalShortCircuitsReturnThis`, `...EmptyStringKeepsSingleton`, `testAnonymousRequestReportsDefaultPrincipal` | asserts `assertSame(this/INSTANCE)` and `"admin"`; fails if a context is derived | TESTED |
| 2 | `forPrincipal` new principal → derived, reports principal | `testAllowAll/ReadOnlyFactoryReportsConfiguredPrincipal`, `...ForPrincipalCachesDerivedContext` | `assertNotSame(INSTANCE)` + `assertEquals("foo", getPrincipal())`; fails if singleton/`admin` returned | TESTED |
| 3 | Per-principal cache retains distinct principals | `testForPrincipalRetainsManyDistinctPrincipals`, `...DoesNotLeakAcrossPrincipals`, `...ConcurrentDistinctPrincipalsRetainedNoThrash` | `assertSame` on re-request; fails on eviction/thrash | TESTED |
| 4 | `@Transient` principal copied before retention | `testForPrincipalCopiesTransientPrincipal`, `testAllowAllFactoryCopiesTransientPrincipal` | mutates source buffer, asserts cached value intact | TESTED |
| 5 | 256-cap degrade → allocate-per-call | `testForPrincipalCapDegradesToAllocatePerCall` | fills to cap, asserts `assertNotSame` beyond cap yet correct principal | TESTED |
| 6 | Concurrent COW publish/convergence | `testForPrincipalConcurrent{ReportsOwnPrincipal,AlternatingPrincipalsNeverLeak,SamePrincipalReusesCachedContext,FirstDerivationConverges}` | `AtomicInteger errors` + `assertSame` convergence; captures spawned-thread failures | TESTED |
| 7 | null `getPrincipal()` no NPE | `testForPrincipalWithNullCurrentPrincipalDoesNotThrow` | derives without NPE, asserts derived principal | TESTED |
| 8 | `AllowAll.newPrincipalContext` keeps type + settingsReadOnly | `testForPrincipalPreservesRuntimeType`, `testSettingsReadOnlyForPrincipalStaysSettingsReadOnly` | `instanceof AllowAll` + settings denial | TESTED |
| 9 | `ReadOnly.newPrincipalContext` keeps type + read-only | `testReadOnlyForPrincipalRetainsAndStaysReadOnly`, `testReadOnlyFactoryReportsConfiguredPrincipal` | write denied + principal reported | TESTED |
| 10 | `DenyAll.newPrincipalContext → this` | `testDenyAllForPrincipalStaysDenyAll` | `assertSame(INSTANCE)` + deny on `authorizeHttp` | TESTED |
| 11 | DenyAll re-parent preserves deny behavior | `testDenyAllForPrincipalStaysDenyAll` + existing HTTP/PG security suites | deny paths still throw | TESTED |
| 12 | Mat-view refresh keeps view-scoped insert | `MatViewTest.testRefreshExecutionContextForPrincipalKeepsViewScopedInsert` | `assertSame` + permitted view insert / denied non-view insert | TESTED |
| 13 | HTTP `current_user` reflects configured user (call + bare form) | `HttpSecurityTest.testStaticHttpAuthenticatorFactory_principalReflectsConfiguredUser`, `testAnonymousRequestReportsDefaultPrincipal` | dataset `[["foo"]]` vs `[["admin"]]`; fails on revert | TESTED |
| 14 | PGWire `current_user()/session_user()` reflect user | `PGSecurityTest.testCurrentUserReflectsConfiguredPrincipal` | RO user reads back `"user"`; fails on revert | TESTED |
| 15 | ReadOnlyUsersAware default/ILP branch reports principal | `testReadOnlyUsersAwareFactoryDefaultInterfaceReportsConfiguredPrincipal` | principal reported + anonymous → singleton | TESTED |

Regression efficacy (Agent 14): the three load-bearing tests (#13, #14, #2) each **fail** when the production hunk is reverted (factory returns singleton / `getPrincipal()` returns `admin`) — verified by reasoning against the revert. No new `assertSql(...)`/`getPlan(...)` and no new deterministic `.returnsOnce(...)` introduced.

## PR metadata

- Title is Conventional-Commits-valid and repeats the verb. As already noted, "when authentication is enabled" is slightly imprecise — the real trigger is *authorization-not-enforced AND a non-default user authenticates* — but it reads well for release notes; non-blocking.
- Labels present and correct: Bug, UI, Postgres Wire, REST API, Security, tandem. The earlier-flagged unrelated whitespace churn (SqlException / Concurrent*HashMap / WhereClauseParserTest) has already been dropped from the branch.
- Description lists the fix, the caching design, the cap degradation, and the tandem ENT PR (#1076). Level-headed and complete.

## Summary

- **Verdict: request changes** — on a single, trivial, verified item: **M1**, pre-size the copy-on-write map in `addPrincipalContext` (one line). There are **zero** correctness, data-integrity, resource-safety, or concurrency defects; the lock-free per-principal cache is JMM-correct and identity-leak-free, the DenyAll re-parenting has zero blast radius, and the read hot path is allocation-optimal. M1 is a cold-path (≤256 lifetime rebuilds per singleton), bounded, microsecond-scale inefficiency — hence Moderate, not Critical — but per QuestDB's zero-known-inefficiency standard and the "fix it while we're in there" doctrine it should be closed in this PR rather than left sub-optimal. Everything else is genuinely Minor/cosmetic.
- **Correctness & performance gate:** the only open non-Minor finding is M1 (cold + bounded); no finding touches a reachable hot path.
- **Test gate:** passes — every behavioral change is tested with a real failure link; the fix has regression tests that fail on revert.
- Coverage map: 15 behavioral changes, 15 tested, 0 UNTESTED.
- Findings: 3 draft false positives removed; verified findings = 1 Moderate + 5 Minor.
- Split: 0 in-diff correctness defects, 0 out-of-diff, 0 pre-existing bugs (1 pre-existing unchanged quirk noted, non-actionable).

Requested change for this PR: apply the one-line pre-size in **M1**. Once that lands, this is an approve.

glasstiger:
# Code review — tandem PR: enterprise #1076 + OSS questdb/questdb#7273

**Level 3, tandem mode — full mission-critical pass.** Both diffs read in full; change-surface and callsite inventory built with grep; both modules compiled clean (`compile test-compile`, main + test); focused analysis passes across correctness, concurrency, performance, resource, configuration, cross-context caller impact, and tests, each source-verified.

**Net: 0 Critical, 5 Moderate (all non-blocking coverage-hardening), 2 Minor; 1 false positive dropped. Zero out-of-diff defects. Verdict: approve (ENT) / approve (OSS).** This is an exceptionally well-engineered and well-tested security fix.

## Critical
**None.** No confirmed correctness, security, data-integrity, resource, concurrency, or performance/IO defect — in-diff, out-of-diff, or pre-existing. The load-bearing invariants were each source-verified:

- **`forPrincipal` cache is identity-leak-safe.** The ≤256 cap is enforced by the CAS admission counter (a map insert requires a successful bump < 256, so the map can never exceed the cap); the lock-free read is JMM-correct (`volatile` map + `final principal` + volatile `ConcurrentHashMap` Node publication), so a racing slot can only cost an extra allocation, never a wrong-principal context. Every production subclass overrides `newPrincipalContext` (or returns `this`), so the deny→allow downgrade is impossible even without `-ea`; the assertion is a belt-and-suspenders guard.
- **ENT dispatch proxy identity routing is consistent.** For the shared anonymous proxy, `getPrincipal()` / `getSessionPrincipal()` / `getAssumedServiceAccount()` all resolve to the delegate ("admin") — `Chars.toString(null) == null` drives the `getSessionPrincipal` routing. The `AtomicReference`→plain-`volatile DelegateSnapshot` change is stale-safe (active factory swapped *before* the generation bump; immutable `{delegate, generation}` pair; last-writer-wins), pinned by the freshness suite's torn-pair / straddling-flip / settle-convergence tests.
- **ILP bypass (fix #4).** `SecurityContextFactory.ILP == SecurityContext.AUTH_TYPE_JWK_TOKEN == 2`, so re-keying the bypass on `SecurityContextFactory.ILP` is a robustness fix, not a behavior change. The two owner-grant guards compose correctly: `isIdentitylessOwnerContext` skips the singleton-identity case (survives an admin rename), and `grantIfExisting` is the backstop that refuses to fabricate an ACL entity for an unknown name (verified against a raw non-Ent `AllowAllSecurityContext` too — `tryToEntSecurityContext` returns null there, and `grantIfExisting` still prevents fabrication).
- **`AccessListModel` refactor is exactly behavior-preserving.** `resolveGrantOrRevokeAccessList(name, verify, !verify)` reproduces the old `grantOrRevoke` for both `verify` values (verify=true throws "Grantee does not exist" before the create branch; verify=false creates-if-missing); `grantIfExisting` (`false, false`) returns null for missing entities → no fabrication, while the external-user / external-alias throws still fire first. All under `synchronized`, so the existence-check + grant is atomic vs a concurrent `dropEntity`.
- **Config reject.** Channel-parity with the OSS `getString` resolution is exact across all four channels (`http.password` property / `QDB_HTTP_PASSWORD` env / `http.password.file` / `QDB_HTTP_PASSWORD_FILE`), using the now-public shared suffix constants so the literals cannot drift. No silent no-auth channel survives, and no normal config triggers a spurious boot failure (`overrideOssDefaults` never sets these keys). `http.user` is non-sensitive so the secret-file channel is correctly probed only for the password. Runs before `super(...)` so it preempts the OSS pair-validation misdirection.
- **`current_user()`/`session_user()` runtime-constant change** is the exact `now()` idiom: `init()` re-runs on every `getCursor()` (no cross-user staleness on a reused select-cache factory), `isThreadSafe` sharing has a happens-before edge via the reduce queue, and the runtime-constant classification is strictly *safer* than before (it avoids a latent compile-time `getStrA` read on a not-yet-initialized null context in the symbol-key paths). Reachability expansions (SAMPLE BY time zone, ANCHOR rejection, mat views) are all benign.
- **Re-parenting / submodule bump.** No `AllowAllSecurityContext`/`ReadOnlySecurityContext` identity-`==`/`instanceof` exists in production; only `DenyAllSecurityContext` comparisons exist and are safe (it pins `newPrincipalContext→this` and is never routed through `forPrincipal`). The full reactor compile confirms no broken caller across the signature changes (`grantIfExisting` on every `AccessListWriter` impl, the new `getInstance` overload, the four-arg `DispatchingSecurityContextFactory` constructor, the removed `isPrincipalIndependent`).

## Moderate — coverage hardening (non-blocking)
The behaviors below are source-verified correct and covered transitively; each item is a missing *direct* test to pin against future regressions, not an untested user-visible behavior. None blocks merge.

1. **[OSS] Cross-user re-snapshot × parallel-Async execution are never crossed in one test.** `UserFunctionsTest.testUserFunctionsRefreshOnEveryCursor` proves per-user re-snapshot only for a single-threaded plain select; `testUserFunctionsAreSharedAcrossParallelWorkers` proves worker-sharing only with one user + fresh factories. The production-critical shape — a shared select-cache factory reused by user A then user B *while running as a parallel Async filter / GROUP BY* — is covered only transitively. Suggested: compile one Async-filter factory, run as alice, re-run the same instance as bob, assert bob's identity.
2. **[ENT] Non-empty OIDC group defensive-copy + forwarding through the dispatch proxy has no positive test** — only the empty→null collapse is pinned. The per-name copy loop is largely pre-existing (this PR only cached `groupCount` and added the collapse), so risk is low. Suggested: a named connection with a non-empty group list; mutate the source list; assert the forwarded copy is unaffected.
3. **[ENT] Named ACL-disabled connection surviving PRIMARY→REPLICA→PRIMARY with pooling enabled** — all flip tests target the anonymous *pooled* proxy; the named (unpooled) path is flip-tested only with pooling off. Wiring is identical regardless of the flag, so low risk.
4. **[ENT] Named-user `current_user` over HTTP on a read-only instance** — `ReadOnlyInstanceCurrentUserTest` covers named users only over PGWire (HTTP tests only the anonymous default); mitigated by OSS `HttpSecurityTest`.
5. **[OSS] `WHERE indexed_symbol = current_user()`** has no regression test — verified by source to stay a per-row filter (never mis-folded into a deferred symbol key), but nothing pins that against future `WhereClauseParser` changes.

## Minor
1. **[ENT] PR title** carries a 💥 emoji — slightly informal vs the level-headed-tone convention, though it usefully flags the breaking change. Cosmetic.
2. **[ENT, pre-existing] `DispatchingSecurityContextFreshnessTest.entHttpAuthenticatorReadsStoresFromEngineEveryCall`** asserts field *absence* via reflection (name-brittle; tests structure, not the store-visibility behavior it guards). Inherited by this PR, not introduced (only comments were removed); defensible since "field absent" has no public-API equivalent. A behavioral test (swap store, authenticate, observe) would be more robust.

## Downgraded (false positive)
- **"The single-mode Ent factories route through `forPrincipal` without an ILP guard"** — **disproven.** Both `EntAllowAllSecurityContextFactory` and `EntReadOnlySecurityContextFactory` have `if (interfaceId == SecurityContextFactory.ILP) return …INSTANCE;` immediately above the `forPrincipal` line; the guard is present and consistent with the OSS counterparts.

## Coverage map (key behavioral changes; TESTED unless noted)

| Change | Test (verified) | Fails if regressed? |
|---|---|---|
| [OSS] `AllowAll/ReadOnly.getPrincipal()` → configured principal | `SecurityContextFactoryPrincipalTest.testAllowAll/ReadOnlyFactoryReportsConfiguredPrincipal` | asserts "foo" ≠ "admin" |
| [OSS] `forPrincipal` cache: cap / no-overshoot / retain / no-leak / downgrade-reject / transient-copy | 20+ `testForPrincipal*` incl. 11 concurrency | barrier forces 8 racers → 1 retained; etc. |
| [OSS] `DenyAll` / MatViewRefresh `newPrincipalContext → this` | `testDenyAllForPrincipalStaysDenyAll`, `MatViewTest.testRefresh…KeepsViewScopedInsert` | derived context would deny the view insert |
| [OSS] factories route principal, ILP exempt | `testAllowAll/ReadOnly/UsersAware*` + `*IlpDoesNotRouteTransportCredential` | ILP returns bare singleton |
| [OSS] current_user/session_user per-cursor snapshot | `UserFunctionsTest.testUserFunctionsResolveOncePerCursorNotPerRow` | `principalCalls == 1` at 10 & 10k rows |
| [OSS] refresh on every getCursor | `testUserFunctionsRefreshOnEveryCursor` (alice→bob→alice) | stale snapshot serves prior user |
| [OSS] isThreadSafe shared across workers | `testUserFunctionsAreSharedAcrossParallelWorkers` | count grows with worker count |
| [OSS] HTTP/PG current_user reflects user; anon → default | `HttpSecurityTest`, `PGSecurityTest.testCurrentUserReflectsConfiguredPrincipal` | reverts to "admin" |
| [OSS] settings-read-only survives derivation | `SettingsEndpointTest.testSettingsReadOnlyForAuthenticatedUser` | authed user regains /settings write |
| [OSS] `SECRET_FILE_*_SUFFIX` made public | — | **EXEMPT** (visibility only; consumed by the ENT reject) |
| [ENT] reject http.user/http.password (property/env/file, alone/paired/blank) | `EntPropServerConfigurationTest` (16 tests) | OSS pair-validation message wins instead |
| [ENT] anonymous proxy pooling on/off by ACL | `EntFactoryProviderSecurityWiringTest`, `DispatchingSecurityContextFactoryTest` | pooled/unpooled identity wrong |
| [ENT] dispatch volatile snapshot + identity routing | `DispatchingSecurityContextFreshnessTest`, `…FactoryTest.namedConnectionReportsItsOwnAuthType` | torn/stale delegate; wrong auth type |
| [ENT] ILP bypass keyed on `ILP`; no phantom entity | `LineTcpAclDisabledTest`, `…RenamedAdminTest`, `LineTcpAclTest.testReplicaIlpBypassIsGatedOnLineAclBeingDisabled` | phantom "admin"/keyid ACL entity minted |
| [ENT] replica factory `assert isAclEnabled`, lineAcl ILP bypass | `testAnonymousIlpResolvesOnBothFactorySlots`, `testAuthenticatedIlpDoesNotMasqueradeAsAclUser` | replica ILP dies on "Invalid user name" |
| [ENT] `grantIfExisting` + `isIdentitylessOwnerContext` | `AccessControlConcurrentTest.testGrantIfExisting*`, `LineTcpAclDisabledRenamedAdminTest` | phantom entity / atomicity vs DROP |
| [ENT] `EntValidationSecurityContext.newPrincipalContext → this` | `testValidationContextForPrincipalStaysValidationContext` | validation context downgraded |
| [ENT] same-role retry preserves stateful context | `EntCairoEngineSwitchRoleIntegrationTest.sameRoleRetry…` | assumed account / owner lost |
| [ENT] non-empty OIDC group copy through proxy | *(none direct — Moderate #2)* | **UNTESTED** (pre-existing copy loop) |
| [OSS] cross-user × parallel-Async; `WHERE sym = current_user()` | *(none direct — Moderate #1, #5)* | **UNTESTED** (source-verified safe) |

## Summary
- **Correctness & performance gate: PASS** — zero confirmed correctness/security/resource/concurrency/performance findings on any reachable path (in-diff, out-of-diff, or pre-existing). Full reactor compile is green; the callsite inventory (singleton comparisons, factory constructors, `grantIfExisting` impls, `forPrincipal` callers, the removed `isPrincipalIndependent`) is clean.
- **Test gate: PASS** — all four headline fixes have revert-would-fail regression tests; new user-visible behavior has error-path (write-denied) and NULL/anonymous-path coverage; the SQL-assertion gate is fully clean (zero new query-result `assertSql`, zero `returnsOnce`, no lone `assertQuery`-in-`assertMemoryLeak`). One efficacy nuance worth keeping in mind for future edits: the `EntDdlListener` owner-grant hunk is pinned **only** by the `RenamedAdmin` test variant (with the default admin name, the old `owner == adminUser` screen already suppresses the grant, so the plain variant passes even with that hunk reverted).
- **Coverage map: ~20 behavioral changes, ~18 tested, 1 EXEMPT, ~2 UNTESTED** (the Moderate coverage-hardening gaps above — behavior source-verified correct).
- **In-diff / out-of-diff / pre-existing split:** 0 out-of-diff defects, 1 pre-existing Minor (the brittle-reflection test).
- **Merge order** (from the PR body, worth honoring): OSS #7273 lands first; after it squash-merges, re-bump the submodule in #1076 to the squashed `master` commit before merging — `ci/submodule_is_mergeable.py --strict` now gates on exactly that.

The five Moderate and two Minor items are optional hardening/polish — none is a mis-filed correctness or performance defect. Nothing blocks merge.

<sub>🤖 Level-3 tandem review by Claude Code</sub>

glasstiger:
## Code review — `fix(core): fix the web console showing "admin" instead of the authenticated user`

Reviewed at **level 3, tandem mode** (OSS #7273 + Enterprise mirror questdb/questdb-enterprise#1076) — full mission-critical pass across correctness, concurrency, performance, resource management, tests, code quality, and cross-context caller impact, plus fresh-context and adversarial passes. This comment covers the OSS half; ENT-side findings are on the enterprise PR.

**Verdict: approve.** The fix is confirmed end-to-end: `AllowAllSecurityContext`/`ReadOnlySecurityContext` are re-parented under a new `AbstractPrincipalAwareSecurityContext` that reports the authenticated principal via a per-principal `forPrincipal` cache, and `current_user()`/`session_user()` snapshot the principal once per cursor instead of per row.

### Critical
None. Verified in depth:
- **The per-principal cache cannot leak principal A's context to B** — keys are content-hashed (`Chars.hashCode`/`Chars.equals`), the stored value is created for and reports its own key, the `@Transient` principal is copied to an immutable `String` before storage, and the 256 bound is a **hard** cap (CAS admission before insert, self-healing on failed derivation, no overshoot under contention).
- **JMM / concurrency** — the map is a case-sensitive JDK-CHM fork; `computeIfAbsent(key, token, biFn)` runs the mapping at most once per key under the bin lock and stores nothing on null; the lazy-map init CAS discards the losing thread's throwaway map cleanly. No monitor serialization on the hot path (pinned by `testForPrincipalNeverSerializesOnTheInstanceMonitor`).
- **`current_user()`/`session_user()`** now `isRuntimeConstant()`+`isThreadSafe()` snapshot in `init()` (runs per `getCursor()`), which is behavior-preserving (the principal is invariant per connection) and a clear win over the old per-row `getPrincipal()` read; no compile-time folding, so a cached plan reused by another connection still reports the right user.
- **`getAutoCreateOwner()`** default (`= getPrincipal()`) and the `LineTcpMeasurementEvent:356` switch from `getPrincipal()` are value-identical for authenticated users; the null-owner ILP path is null-safe (`addColumnName` writes `principalLen=0`).
- **Re-parenting** breaks no `instanceof`/cast anywhere (grep-verified both repos); `DenyAllSecurityContext`, mat-view and live-view refresh contexts correctly override `newPrincipalContext`→`this` so they are never downgraded.

### Minor / latent (optional)
- **`forPrincipal` downgrade guard is assert-based** — `AbstractPrincipalAwareSecurityContext.newCheckedPrincipalContext`. All current subclasses conform (verified across both repos), so this is a guard for *future* subclasses that override `authorize*` but forget `newPrincipalContext` (which would silently downgrade a DENY context to an ALLOW one). Per QuestDB convention `assert` is the accepted mechanism for internal invariants a user cannot trigger, so this is a note rather than a defect — but the surrounding comments state "QuestDB runs with `-ea`" as a production guarantee; if production JVMs don't set `-ea`, consider softening that framing (or a non-assert throw for this specific security-relevant invariant).
- **PR description describes a superseded cache design** — the body calls the 256 bound "soft… racing inserts can overshoot… the alternative is a CAS retry loop," but the code implements exactly a hard-cap CAS retry loop (`testForPrincipalConcurrentCacheCapDoesNotOvershoot` asserts no overshoot). Inline docs are correct; the body just drifted.

### Adjacent (pre-existing; not this PR)
- **`ConcurrentHashMap.computeIfAbsent(key, token, fn)` stores the raw key on the collision/tree paths** (`:1000,1014`), not `maybeCopyKey`-copied like the single-`Function` overload — a future caller passing a `CloneableMutable` flyweight key could alias a reused buffer. `forPrincipal` is immune because it pre-copies to an immutable `String` (that copy is load-bearing precisely for this). Worth a one-line contract note so a future caller isn't bitten. Pre-existing (commit `5d42ab699e`).

### Coverage
Excellent. `SecurityContextFactoryPrincipalTest` (26 methods incl. 8 concurrency stress tests) covers cache hit-reuse, distinct-principal retention, the 256-cap→allocate-per-call degradation, the `@Transient` copy on both the cached and over-cap legs, first-derivation/last-slot convergence, failed-derivation reservation release, exact-runtime-type preservation, and the downgrade rejection. `UserFunctionsTest` pins per-cursor resolution (scale-invariant 10 vs 10 000 rows), refresh-on-every-getCursor, and parallel-worker sharing. `HttpSecurityTest`/`PGSecurityTest`/`SettingsEndpointTest`/`MatViewTest`/`LiveViewRefreshSqlExecutionContextTest` cover the user-visible surfaces and the settings-read-only + view-scoped-insert survival. No SQL-assertion-rule violations. The regression tests genuinely fail on revert (verified counterfactually).

### Summary
- **Verdict: approve.** No Critical, no blocking coverage gap; the two items above are a latent-hardening note and PR-description drift.
- Tandem with questdb/questdb-enterprise#1076; that PR carries one Moderate ENT-side item (an ILP owner-grant guard that misses the OSS bypass singleton on the QWP-over-UDP path — net-neutral vs base, non-blocking) plus a few Minor ENT items.

<sub>🤖 Automated level-3 tandem review via Claude Code.</sub>

glasstiger:
# Code review — `fix current_user() reporting "admin"; reject http.user; stop the ILP ACL-bypass minting phantom ACL entities`

**Level 3, tandem review** (ENT questdb-enterprise#1076 + OSS questdb/questdb#7273) — full mission-critical pass: 13 fresh-context agents (correctness, concurrency, performance ×2 incl. adversarial, resource, tests, code-quality, PR-metadata, cross-context caller impact, fresh-context adversarial, and the three test-code agents; Rust agent skipped — no `.rs` files), every finding source-verified. Same report posted to both PRs.

**Verdict: ENT → approve · OSS → approve.** Both the correctness gate and the test gate pass. All 13 agents plus independent hand-verification converged on **zero Critical and zero user-visible Moderate findings**. Everything below is a redundant-guard coverage nit or cosmetic polish; nothing blocks merge.

**Submodule provenance:** `questdb` — OFF-DEFAULT, in scope (OSS #7273 branch head; the pin is re-bumped to squashed `master` before ENT merges, CI-gated by `submodule_is_mergeable.py` — merge mechanics, not a finding). No nested `java-questdb-client` move.

## What the change does (verified end-to-end)
When authorization isn't enforced (ACL disabled / allow-all / read-only), the security contexts reported the hard-coded default `admin` from `getPrincipal()`/`getSessionPrincipal()`, so `current_user()`, `session_user()`, and `SHOW CREATE … OWNED BY` reported `admin` for every authenticated user — worse on read-only instances via a shared-proxy fast path. The fix threads the authenticated principal through a lock-free per-principal `forPrincipal` cache, removes the read-only shared-proxy path, rejects the inert-but-dangerous `http.user`/`http.password` at boot (they left REST/console unauthenticated while the operator believed otherwise), and closes an ILP ACL-bypass that fabricated phantom ACL entities (latent privilege escalation after an admin rename).

## Critical
**None.** No confirmed, reachable defect with a user-visible symptom — in-diff or out-of-diff — across the changed files, their callsite inventory, and cross-context exposures. The load-bearing properties were each proven independently:
- **Cache security invariant** ("principal A's context never handed to B"): holds because identity derives from the map key, not the reservation counter; the cap is exactly 256 (CAS gate; null mappings store nothing); the `@Transient` principal is pre-copied to a `String` before storage (required — `ConcurrentHashMap`'s collision/tree paths don't copy the key).
- **No ACL pollution / privilege escalation**: three consistent guards — synchronous create via `isIdentitylessOwnerContext` (unwraps the dispatch proxy through `resolveCurrentDelegate`), async ILP column-add via the blank serialized `getAutoCreateOwner()`, and `grantIfExisting` which fabricates nothing for an unknown name. Verified against default-admin, renamed-admin, and real-colliding-user scenarios (three distinct ILP paths: direct context, dispatch-proxy unwrap, real TCP end-to-end).
- **No out-of-diff breakage**: every changed/removed symbol walked two caller-levels up across both repos — all SAFE. `isPrincipalIndependent` fully removed (compiles); all four `AccessListWriter` implementors define `grantIfExisting`; `DenyAll` sentinel identity comparisons unaffected; ~20 test subclasses inherit a concrete `newPrincipalContext` and none are `forPrincipal`-routed (the `-ea` downgrade assert is never reached).
- **Concurrency**: the `active`-before-`generation` volatile ordering makes a stale delegate stamp "only ever too old" (one redundant re-resolve, never a stale role); the `current_user()`/`session_user()` snapshot publishes to parallel workers via the same `MCSequence` cursor edge `now()` relies on.

## Moderate
**M1 — `grantIfExisting`'s built-in-admin branch is untested `[incomplete-hardening]`.** `AccessListModel.java:421` (`if (adminUser != null && Chars.equals(adminUser, entityName)) return false;`) has no test that calls `grantIfExisting(<adminUser>, …)` and asserts `false`. Reachable defensive guard with **no distinct user-visible outcome** — it is redundant with the null-resolution path directly below it: the built-in admin is never an ACL entity, so `resolveGrantOrRevokeAccessList(admin, false, false)` returns `null` → `false` regardless. Non-blocking. *Optional fix:* one unit test asserting `grantIfExisting(ADMIN_USER, …)` returns `false` and mutates nothing.

## Minor (all non-blocking)

**Production — member ordering (CLAUDE.md: alphabetical within kind):**
- `EntReplicaSecurityContextImpl.java:208` — `getAssumedServiceAccount()` should sort before `getAuthType()` (203). `DispatchingSecurityContext` got the same pair right — internal inconsistency.
- `AccessListModel.java:606` — private `applyGrantOrRevoke()` should sort to the top of the private group (before `getAccessList()` at 557), not sit beside its logical parent `grantOrRevoke`.

**Test efficacy (each backed by a discriminating test elsewhere — no real coverage hole):**
- `DispatchingSecurityContextProxyTest.authTypeIsPreservedAcrossSwitch:226-257` — name/comment promise an auth-type-across-switch check; the only assertion is on `getSessionPrincipal()`. Auth-type reporting is covered by `DispatchingSecurityContextFactoryTest.namedConnectionReportsItsOwnAuthType`.
- `EntSecurityContextFactoryPrincipalTest.testLegacyConcreteSuperclassesArePreserved:112-117` — the two `assertSame(x, x)` calls are runtime-vacuous; the real (legitimate) guard is the compile-time typed assignment.
- `EntCairoEngineSwitchRoleIntegrationTest.sameRoleRetry…:314-315` — `assertEquals("admin", …)` are default-principal sanity checks that can't reach the fix (the method's assumed-SA assertions are effective).
- `LineTcpAclDisabledTest.testAuthenticatedIlpAutoCreateDoesNotGrantAclPermissions` — `!contains("ilp_key_id")` is structurally always true (bypass owner is `admin`, never the key id) → passes with the fix reverted. Defense-in-depth; the bug is covered by the renamed-admin suite.

**Test-code quality:**
- `DispatchingSecurityContextFreshnessTest:293-320` reinvents a `PrincipalContext` double instead of the PR's own new `CredentialsPrincipalContext` (its sibling test uses it).
- Duplicated read-only-user config across `AclDisabledCurrentUserTest`/`ReadOnlyInstanceCurrentUserTest` (hoist to the shared base, as the assert-helpers already were).
- Lowercase SQL keywords in `LineTcpAclDisabledTest`/`LineTcpAclDisabledRenamedAdminTest` — contradicts CLAUDE.md and the PR's own uppercase conversion of `AbstractAccessControlTest`.
- Repeated owner-grant assertion tail (~4×) in `LineTcpAclDisabledRenamedAdminTest` → add an `assertNoOwnerGrant(store, principal, token)` helper (the column counterpart already exists).
- Member-ordering nits across several new/changed test files; one bare `false` arg missing the `/*isAnonymousProxyPoolingEnabled*/` comment its siblings carry.

**PR metadata:**
- ENT title carries an emoji + `breaking change 💥 -` prefix (0 of ~120 recent ENT titles use emoji/"breaking"); it lands verbatim in release notes. Suggest dropping both — the breaking signal is already well carried by the `## Breaking change` body section + `docs/upgrade.md`.
- OSS title under-scopes to "the web console"; the fix also corrects `current_user()`/`session_user()` for PGWire and REST callers.
- Optional labels: ENT `rbac` fits and is unapplied; OSS `non-breaking-tandem` fits the non-breaking OSS half.

## Downgraded (raised in analysis, dismissed on verification — not findings)
- **`forPrincipal` non-LRU cap tail**: once 256 lifetime-distinct principals are cached on a singleton, further principals allocate-per-call. Bounded, O(1), documented; unreachable in normal deployments; ILP key-ids are deliberately excluded so a device fleet can't saturate it. Net-**positive** vs base (base allocated per-request unconditionally).
- **4-arg `getInstance` default allocates an anonymous `PrincipalContext`**: not on any production path — all five production factories override the overload; only 2-arg-only test/custom factories hit it.
- **OOM-during-CHM-node-alloc counter leak**: if node allocation throws `OutOfMemoryError` after the mapping returns, one reservation isn't rolled back — but it only saturates the cap marginally *early* (safe direction), never exceeds it, never affects identity.
- **`http.user` → boot failure**: the intended, documented breaking change (fail-closed on a setting that left the server unauthenticated). Confirmed intended via both PR bodies + `docs/upgrade.md`.

## Adjacent findings
None. The review visited the callsite inventory and cross-context exposures and found no pre-existing bug this PR walks past.

## Coverage map
All 19 behavioral changes are TESTED with a verified failure link (assertions read, not trusted by name); all 6 dangerous mutation-spot-check lines are caught; no SQL-test-assertion-rule violations (zero new `assertSql`/`returnsOnce`; the four new `assertQuery` calls terminate in `.assertsPlanContaining`).

| Behavioral change | Test (verified) | Verdict |
|---|---|---|
| current_user()/session_user() reflect auth user, ACL-off | `AclDisabledCurrentUserTest` (rouser≠admin); OSS `HttpSecurityTest`/`PGSecurityTest`/`UserFunctionsTest` | TESTED |
| OWNED BY reflects auth user on read-only instance | `ReadOnlyInstanceCurrentUserTest` (OWNED BY 'rouser') | TESTED |
| user fns once/cursor + refresh every getCursor + thread-safe + optimizer paths | OSS `UserFunctionsTest` (scale-invariance; alice→bob→alice; Async plan; runtime-const arg) | TESTED |
| forPrincipal cap/copy/coexist/converge/no-leak/reject-downgrade | OSS `SecurityContextFactoryPrincipalTest`; ENT `EntSecurityContextFactoryPrincipalTest` | TESTED |
| DenyAll/validation/matview/liveview not downgraded | OSS `MatViewTest`/`LiveViewRefreshSqlExecutionContextTest`; ENT principal test | TESTED |
| ENT pooling (named→own; anon ACL-off→shared/iface; anon ACL-on→own; re-resolve on flip; no spurious ASA) | `DispatchingSecurityContextFactoryTest`/`…FreshnessTest`/`…IsolationTest` | TESTED |
| Boot rejects http.user/http.password (alone/pair/env/secret-file; no false-positive on blank) | `EntPropServerConfigurationTest` (14 methods) | TESTED |
| ILP bypass mints no ACL entity (default+renamed admin; direct/proxy/TCP-e2e); real colliding user gets nothing; gated on line ACL | `LineTcpAclDisabledTest`, `LineTcpAclDisabledRenamedAdminTest`, `LineTcpAclTest` | TESTED |
| getAutoCreateOwner() null for Ent singletons | `EntSecurityContextFactoryPrincipalTest` | TESTED |
| grantIfExisting: grant existing / false-unknown-no-fabricate / throw external / **false-admin** | `AccessControlConcurrentTest`, `AccessListTest`, `LineTcpAclDisabledRenamedAdminTest` | TESTED except the **admin** branch → **M1** |
| writes denied on read-only instance (storage fence) | `ReadOnlyInstanceCurrentUserTest`/`AclDisabledCurrentUserTest` | TESTED |
| replica session-principal / assumed-SA survive ASSUME | `EntSecurityContextBaseTest`, `ReplicationAclTest` | TESTED |
| switchRole live flip: connections survive, delegate re-resolves | `ConnectionSurvivalTest`, `EntCairoEngineSwitchRoleIntegrationTest`, `…FreshnessTest` | TESTED |

## Summary
- **Verdict — ENT: approve · OSS: approve.**
- **Correctness gate: PASS.** Zero open Critical. Severity audited both directions: no Critical to downgrade; no Moderate/Minor promotable (M1 has no completable symptom sentence — redundant guard, no user-visible delta; every "weak test" is backed by a discriminating test elsewhere).
- **Test gate: PASS.** Every user-visible behavioral change ships a regression test with a verified failure link; all 5 fixed bugs fail-without-fix (confirmed by reasoning against base source; empirical revert not run — full reactor build exceeded budget). The lone untested item (M1) is a redundant internal branch and does not block.
- **Coverage map: 19 behavioral changes, 19 tested** (M1 is a sub-branch of an otherwise-tested symbol; no UNTESTED user-visible rows).
- **Findings: 0 Critical, 1 Moderate, ~13 Minor.** Draft findings verified; 4 raised-then-dismissed items moved to Downgraded.
- **Split: 0 in-diff Critical / 0 out-of-diff-breakage / 1 incomplete-hardening / 0 adjacent.** The cross-context pass explicitly cleared out-of-diff breakage — a genuine clean bill, not an under-run.
- **Regressions/tradeoffs:** one intended, documented cost — named ACL-disabled PGWire/ILP connections on a read-only instance now allocate one dispatch proxy per *connection* (not per request) where the old shared proxy allocated none; that shared proxy is exactly what reported `admin`. Net allocation is *reduced* on the ACL-enabled HTTP path.

The change is unusually careful — the test suite anticipates precisely the failure modes an adversarial reviewer would probe (cross-cursor snapshot staleness, renamed-admin collision across three ILP paths, the JWK-key-id-out-of-cache decision, `-ea` downgrade rejection). Nothing blocks merge.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) — level-3 tandem review</sub>

ideoma:
#### [PR Coverage check]

:heart_eyes: **pass** : 201 / 203 (99.01%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/pgwire/ReadOnlyUsersAwareSecurityContextFactory.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/cairo/security/AbstractPrincipalAwareSecurityContext.java|44|45|97.78%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/SessionUserFunctionFactory.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/SecurityContext.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/security/AbstractReadOnlySecurityContext.java|61|61|100.00%|
|:large_blue_circle:|io/questdb/cairo/security/AbstractAllowAllSecurityContext.java|60|60|100.00%|
|:large_blue_circle:|io/questdb/cairo/lv/LiveViewRefreshSqlExecutionContext.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/LineTcpMeasurementEvent.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/security/AllowAllSecurityContext.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/security/AllowAllSecurityContextFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/security/DenyAllSecurityContext.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/CurrentUserFunctionFactory.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/security/ReadOnlySecurityContext.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/security/ReadOnlySecurityContextFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshSqlExecutionContext.java|1|1|100.00%|

bluestreak01 (approved):

ideoma:
#### [PR Coverage check]

:heart_eyes: **pass** : 201 / 203 (99.01%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/pgwire/ReadOnlyUsersAwareSecurityContextFactory.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/cairo/security/AbstractPrincipalAwareSecurityContext.java|44|45|97.78%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/SessionUserFunctionFactory.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/SecurityContext.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/security/AbstractReadOnlySecurityContext.java|61|61|100.00%|
|:large_blue_circle:|io/questdb/cairo/security/AbstractAllowAllSecurityContext.java|60|60|100.00%|
|:large_blue_circle:|io/questdb/cairo/lv/LiveViewRefreshSqlExecutionContext.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/LineTcpMeasurementEvent.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/security/AllowAllSecurityContext.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/security/AllowAllSecurityContextFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/security/DenyAllSecurityContext.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/CurrentUserFunctionFactory.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/security/ReadOnlySecurityContext.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/security/ReadOnlySecurityContextFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshSqlExecutionContext.java|1|1|100.00%|

PR #7275, opened 2026-06-17
fix(core): retry mat view refresh on transient errors instead of invalidating

bluestreak01:
## Summary

A materialized view's incremental refresh can fail transiently in two ways: the base-table reader pool is exhausted (`EntryUnavailableException`, "table busy"), or a refresh query runs out of memory. Previously both were treated as fatal — the view was invalidated (busy immediately; OOM after in-call step-reduction retries), and a busy failure cascaded up the dependent-view chain until someone ran a manual `REFRESH MATERIALIZED VIEW ... FULL`.

This PR bundles three related changes that harden the refresh path against a misbehaving view and give operators the tools to take one out of circulation without bringing the database down:

1. **Transient-error retry** — route "table busy" and OOM through one non-blocking, capped, timer-driven retry path, and expose a queryable `retrying` status. Genuine errors (bad SQL, type mismatch, dropped base table, etc.) still invalidate immediately.
2. **Refresh block list** — a startup config that names views the refresh job must never refresh, as an escape hatch for a view whose refresh crashes or destabilises the database.
3. **In-flight cancellation (verified, no new code)** — confirm and regression-test that a running refresh is already cancellable via `cancel_query()` and that its circuit breaker honours the cancel.

## Problem (production incident)

A materialized view's incremental refresh acquires a reader on its base table (`engine.getReader(baseTableToken)` in `MatViewRefreshJob.refreshIncremental0`). When the reader pool is exhausted, that call throws `EntryUnavailableException` — `table busy [reason=pool size exceeded]`. Reader-pool exhaustion is transient and load-driven: under heavy concurrent query + ingestion load, a momentary spike could permanently invalidate an otherwise-healthy chain of views.

A 4-level chain `market_data -> bbo_1s -> bbo_1m -> bbo_1h -> bbo_1d` was invalidated at the exact instant the base table's reader pool was exhausted:

```
ApplyWal2TableJob job ejected [table=market_data~29, ... time=1328ms]
ReaderPool could not get, busy [table=market_data~29, thread=204, retries=10]
E MatViewRefreshJob marking materialized view as invalid [view=bbo_1m~34, reason=base materialized view refresh failed]
E MatViewRefreshJob marking materialized view as invalid [view=bbo_1h~35, reason=base materialized view is invalidated]
E MatViewRefreshJob marking materialized view as invalid [view=bbo_1d~36, reason=base materialized view is invalidated]
```

### 1. Defer transient "table busy" instead of invalidating

On `EntryUnavailableException` during incremental refresh, schedule a per-view backoff deadline (`MatViewState.refreshRetryAfterMicros`) and return without invalidating. While the backoff is active the incremental paths skip the view (`isRefreshDue` gate), so a busy base table is not hammered on every commit. `MatViewTimerJob` re-enqueues an incremental refresh once the backoff elapses — the only wakeup path for `immediate` views, which have no timer of their own. The view then catches up from its own `lastRefreshBaseTxn`.

### 2. Defer out-of-memory refreshes too, without blocking the worker

`insertAsSelect` keeps its interval step-halving (the actual OOM mitigation) but no longer calls `Os.sleep` between attempts, so the single refresh worker that drains the whole task queue is not blocked. Once the step can no longer shrink, the OOM propagates to the incremental caller and is deferred like a busy error rather than invalidating immediately. `full` and `range` refresh still invalidate on OOM, preserving their truncate-then-empty semantics.

### 3. Cap consecutive retries, then invalidate (bounds WAL retention)

Each consecutive transient failure that does not advance the refresh watermark bumps a per-view counter (`MatViewState.refreshRetryCount`). Once it exceeds `cairo.mat.view.refresh.busy.retry.limit` (default 10) the view is invalidated. This bounds a real retention problem: a valid-but-deferred view holds `WalPurgeJob` at its un-advanced `lastRefreshBaseTxn` (`getSafeToPurgeUpToTxn` only releases WAL for invalid views), so under genuinely persistent exhaustion an uncapped retry would pin base-table WAL without bound. The counter resets on any successful refresh, so a view making slow forward progress is never wrongly capped.

### 4. New `retrying` status

`materialized_views.view_status` returns `retrying` while a view is in a transient-refresh backoff window, with precedence `invalid` > `retrying` > `refreshing` > `valid`. This gives operators a queryable signal that a view is deferring, rather than it silently reporting `valid` while falling behind.

### 5. Non-blocking timer wakeup

`MatViewRefreshJob` notifies the timer of the retry deadline via a `RETRY` task; `MatViewTimerJob` tracks the earliest pending deadline so its per-tick fast path is a single compare and the full-fleet scan runs only when a retry is actually due. The sweep rides the existing `mat-view-refresh` worker pool (no new thread) and does not spin.

## Changes: refresh block list

`cairo.mat.view.refresh.block.list` is a comma-separated list of materialized view names that the refresh job must never refresh. Listed views are skipped by every refresh path — incremental (both the dependent-views fan-out and the single-view path), full, and range — without being invalidated. View names match case-insensitively, consistent with how QuestDB treats mat view names elsewhere.

This is the operator escape hatch for the crash-loop scenario: if a view's refresh crashes the JVM (or otherwise destabilises the database), the database replays the refresh on restart and crashes again. Naming the view in the block list lets the database start and stay up while the offending view is left untouched.

The option is read once at startup; changing it requires a restart. The default is empty, so behavior is unchanged unless configured.

**Tradeoff:** a blocked view grows stale, and because it never advances its last refreshed base txn it can keep pinning the base table's WAL retention until it is dropped or removed from the list. This applies equally to `immediate`, `timer`, and `manual` views: the block-list skip in `refreshDependentViewsIncremental` short-circuits *before* the `incrementRefreshIntervalsSeq()` bump (its only caller), so a blocked `timer`/`manual` view never signals `MatViewTimerJob` to enqueue an intervals-update task. Its cached intervals therefore stop advancing just like an `immediate` view's, and `getSafeToPurgeUpToTxn` pins base-table WAL just as hard. (The behavior is safe — it only over-retains WAL, no data loss — but the block list is not a softer escape hatch for non-immediate views.)

## Cancelling an in-flight refresh

The block list stops *future* refreshes, but it does not stop a refresh that is already running when the operator reacts. That one is already cancellable, and this PR adds a regression test to lock the behavior in — there is no production code change for cancellation.

How it works: the refresh job compiles the view query through `SqlCompiler.compile()`, which wraps the SELECT factory in a `QueryProgress`. `QueryProgress.getCursor()` registers the query in the `QueryRegistry` (so the refresh appears in `query_activity()`) and wires the registry entry's cancelled flag into the refresh execution context's circuit breaker. `MatViewRefreshSqlExecutionContext` exposes the cancellable simple circuit breaker, and the refresh reports `isWalApplication() == false`, so `cancel_query()` does not refuse it. A cancel sets the flag; the running cursor trips on its next circuit breaker check (`sleep()`, group-by, and sample-by cursors all probe it) and the refresh aborts with `cancelled by user`, leaving the view `invalid`.

Operator commands:

```sql
-- find the running refresh; its query text is the view's SQL
SELECT query_id, query, worker_pool, state
FROM query_activity()
WHERE query LIKE '%<snippet of the view SQL>%';

-- cancel it; the circuit breaker trips and the cursor aborts
SELECT cancel_query(<query_id>);   -- or: CANCEL QUERY <query_id>;
```

Notes and limits:
- The cancelled refresh is left `invalid` (status `invalid`, `invalidation_reason` contains `cancelled by user`), so it does not auto-rerun on the next base commit.
- Registration spans an open cursor, which the refresh opens per interval, so on a multi-interval refresh there is a brief unregistered window between intervals. If a cancel lands in that gap it returns "query not found"; re-issue it.
- Cancellation cannot help a refresh that hard-crashes the JVM. For that failure mode, use the block list plus a restart.

## Taking a bad view out of circulation (full workflow)

When a single view's refresh is misbehaving — burning resources, looping, or crashing the database — take it out of circulation without disturbing the other views on the same base table:

1. **Identify** the offending view and its status:
   ```sql
   SELECT view_name, view_status, invalidation_reason, last_refresh_start_timestamp
   FROM materialized_views;
   ```
2. **Stop the in-flight refresh** if one is running (see above):
   ```sql
   SELECT query_id FROM query_activity() WHERE query LIKE '%<view SQL snippet>%';
   SELECT cancel_query(<query_id>);
   ```
   The view goes `invalid`, so the refresh job stops re-driving it on every base commit.
3. **Prevent replay across restarts.** Cancellation only affects the running process; if the refresh crashes the JVM, the database replays it on restart. Add the view to the block list and restart so it is never refreshed again until you remove it:
   ```
   cairo.mat.view.refresh.block.list=bad_view
   ```
4. **Investigate / fix**, then **return the view to service** by removing it from the block list (restart) and running a clean rebuild:
   ```sql
   REFRESH MATERIALIZED VIEW bad_view FULL;
   ```
   The full refresh catches up the whole backlog accumulated while the view was out of circulation. If the view is beyond repair, `DROP MATERIALIZED VIEW bad_view;` instead, which also releases the base-table WAL retention the stale view was pinning.

Note the division of labor: `cancel_query()` is the runtime, no-restart tool that stops a running refresh; the block list is the across-restart tool that keeps a crash-looping refresh from being replayed. They are complementary, and step 2 + step 3 together cover both the running and the replayed refresh.

## Config

```
cairo.mat.view.refresh.busy.retry.timeout=1000   # ms, backoff between retries (default 1s)
cairo.mat.view.refresh.busy.retry.limit=10       # consecutive retries before invalidating
cairo.mat.view.refresh.block.list=               # comma-separated view names to never refresh (default empty)
```

## Tradeoffs and behavior changes

- A genuinely unrecoverable OOM view now spends up to `limit x backoff` (about 10s at defaults) in `retrying` before invalidating, where it previously invalidated as soon as in-call step-reduction was exhausted. The terminal signal is delayed for that case.
- The OOM path no longer sleeps between in-call step-halvings. The halving is CPU-bound and bounded by `cairo.mat.view.max.refresh.retries`, and the per-round backoff now comes from the timer, but a sustained OOM does slightly more query work per deferred round.
- `view_status` gains a new value, `retrying`. Tooling that enumerates the status set may need updating.
- Each deferred OOM round rebuilds the refresh context and re-halves the step from scratch (no step persistence across rounds): correct, but redundant CPU under sustained OOM. Left as a future optimization.
- `cairo.mat.view.refresh.oom.retry.timeout` only gated the removed `Os.sleep` and is now inert. The key is retained for config compatibility.
- The `busy.retry.*` config names now also govern OOM deferral, so "busy" is a mild misnomer.
- A blocked view grows stale and can pin base-table WAL retention until it is dropped or unblocked (see the block-list tradeoff above). The block list is read only at startup, so changing it requires a restart.
- A cancelled refresh leaves the view `invalid`; recovery is an explicit `REFRESH ... FULL` (or `DROP`).

## Tests

- `MatViewRefreshBusyRetryTest` injects `EntryUnavailableException` from `getReader`; asserts the view reports `retrying` (not invalidated), then the timer re-drives it to `valid` once the reader is available, including a backoff-deadline gating case.
- `MatViewRefreshRetryTest.testOom` / `testOomParallelSql` updated for the new flow: OOM reports `retrying`, repeated re-drives exhaust the limit and invalidate, then `REFRESH ... FULL` recovers to `valid`.
- `MatViewRefreshBlockListTest` covers the block list: a listed view is skipped by incremental, full, and range refresh without being invalidated, catches up after unblocking, case-insensitive matching, and an unlisted view on the same base table keeps refreshing.
- `MatViewRefreshCancelTest` (no product code) regression-guards cancellation: a `QueryRegistry.Listener` cancels the refresh at registration; one test asserts the cancel trips the circuit breaker and leaves the view `invalid` with a `cancelled by user` reason and an un-advanced row count, the other asserts a later `REFRESH ... FULL` succeeds, proving the breaker resets with no stale cancelled flag. The pre-existing `MatViewTest#testSimpleCancelRefresh` already exercises the same path end-to-end via `CANCEL QUERY`.
- `ServerMainTest#testShowParameters` covers the new config keys.

Verified green locally: `MatViewRefreshBusyRetryTest`, `MatViewRefreshRetryTest`, `MatViewRefreshBlockListTest`, `MatViewRefreshCancelTest`, `MatViewTest` (190, including `testSimpleCancelRefresh`), `ServerMainTest#testShowParameters`, `PropServerConfigurationTest` (99), `PropertyKeyTest`.


bluestreak01:
## Review (level 3 — adversarial, blocking pass)

All 11 review dimensions + per-finding source verification. **18 draft findings verified, 5 dropped as false positives.** All confirmed findings are in-diff except one out-of-diff *interaction* (WAL purge retention). No Critical correctness, concurrency, or resource-leak defects — the core mechanism is sound and data-safe.

### Verdict: Request changes (non-blocking)

Reclassifying `EntryUnavailableException` as transient is correct and well-reasoned. The three things most likely to be wrong all hold:
- **Rollback-before-rethrow**: every throw site inside `insertAsSelect`'s outer `try` is nested in the per-iteration `catch(Throwable)` that runs `walWriter.rollback()` first; the outer-catch `Misc.free(factory)` acts on an already-nulled local → no double-free, no dirty WAL state.
- **No data loss**: a deferred view stays *valid* with an un-advanced `lastRefreshBaseTxn`, so `WalPurgeJob.getSafeToPurgeUpToTxn` (line 486) retains the base WAL it needs; catch-up is driven from the view's own watermark; intermediate `REPLACE_RANGE` commits are idempotent on re-drive.
- **No livelock**: `MatViewTimerJob.processRefreshRetries` is the guaranteed wakeup path for immediate views and re-checks `isDropped/isInvalid/isPendingInvalidation`.

What should be addressed before merge is one performance regression, the test gaps around the actual incident scenario, and PR metadata.

---

## Critical
None.

## Moderate

### M1 — `processRefreshRetries` injects an O(V) full-fleet scan into the shared refresh hot-loop *(in-diff: `MatViewTimerJob.java:285-303,347`)*
`runSerially()` calls `processRefreshRetries()` on **every** tick. It unconditionally does `matViewGraph.getViews(retrySink)` (full iteration of `definitionsByTableDirName`, O(V)) **plus** a second `getViewState()` hash lookup per view into `stateByTableDirName` (another O(V) of string hashing), even though a pending retry exists only after a rare transient `EntryUnavailableException` — the common case is *zero* pending retries.

The cost lands on a hot loop: `MatViewRefreshJob` and `MatViewTimerJob` share `mvWorkerPool` (`ServerMain.java:535-537`), and `Worker.loopBody` (`Worker.java:429-430`) sets `ticker=0` (no sleep, immediate re-spin) whenever any assigned job returns `true`. So while refreshes are flowing, the O(V) scan runs between refresh tasks; pre-PR the steady-state tick was O(1) (`processExpiredTimers` peeks the priority-queue head). At thousands of views this is a needless CPU drain on exactly the large-fleet/heavy-load deployments the feature targets.

Fix (cheapest → cleanest):
- early-return guarded by an `AtomicInteger pendingRetryCount` (or a shared `volatile long` earliest-deadline) bumped in `scheduleRefreshRetry` / decremented in `clearRefreshRetry`/`markAsValid` → common case O(1); **or**
- piggyback the existing deadline-sorted `timerQueue`: enqueue a one-shot retry timer via `timerTaskQueue` (the same cross-thread channel as ADD/REMOVE/UPDATE) and let `processExpiredTimers` handle it — removes `processRefreshRetries` entirely.

Secondary: if any scan remains, iterate `stateByTableDirName.values()` directly instead of tokens-then-`getViewState`, dropping the V redundant lookups and the `retrySink` indirection.

### M2 — Undisclosed tradeoff: uncapped busy-retry pins base WAL under *sustained* exhaustion *(in-diff behavior; out-of-diff interaction `WalPurgeJob.java:479-490`)*
Unlike the OOM path (capped by `getMatViewMaxRefreshRetries`, then invalidates) and unlike pre-PR invalidate-on-busy (which set `invalid=true` and *released* WAL retention), the busy path re-arms `scheduleRefreshRetry` on every attempt with **no cap**. Because a valid-but-deferred view holds `WalPurgeJob` back at its un-advanced `lastRefreshBaseTxn`, genuinely persistent pool exhaustion (a reader/writer leak, chronic over-subscription, misconfig) means the view retries forever, keeps reporting `valid`, and base-table WAL accumulates without bound — with no operator-visible signal. Defensible correctness-over-purge tradeoff, but the PR body omits it. Consider capping consecutive busy deferrals (fall back to invalidation after N) or surfacing a "deferred/retrying" status in `materialized_views`, and at minimum document the tradeoff.

### M3 — Test does not cover the incident it fixes *(in-diff: `MatViewRefreshBusyRetryTest.java`)*
- Sets `CAIRO_MAT_VIEW_REFRESH_BUSY_RETRY_TIMEOUT=0`, so `isRefreshDue` is always true and the **non-zero-backoff blocking/skip behavior is never exercised** — an inverted comparison in `isRefreshDue` (`MatViewState.java:481-484`) would still pass.
- The **dependent-view chain cascade** (`market_data → bbo_1s → bbo_1m → bbo_1h → bbo_1d`) — the literal production incident — is **not tested**. The test has a single immediate view directly on the base, so the cascade-non-invalidation and the `continue` that skips a busy view without advancing `minExaminedToTxn` are uncovered.
- Failure is injected only at `getReader` (base reader pool). The **writer-pool outer-catch path** (`refreshIncremental` ~1641, `refreshDependentViewsIncremental` ~1521) and the **`insertAsSelect` mid-scan rethrow / partial-commit idempotency** are untested.
- The preserved **full/range invalidate-on-busy** branch has no negative test, and `markAsValid()` clearing the retry marker is untested.

### M4 — PR metadata
- **No labels.** A `fix(core):` mat-view PR should carry `Bug`, `Core`, `Materialized View`.
- Body references "observed in production" but no `Fixes #NNN`. If a tracking issue exists, reference it at the top.
- The tradeoff (M2) and chain-coverage gap (M3) should be stated in the body (tone is otherwise good, though the heavy `**bold**` on prose words is borderline against the "level-headed" guideline).

## Minor

- **m1 — Success paths leave a stale retry marker** *(`MatViewState.java`)*: `refreshSuccess` and `refreshSuccessNoRows` do not clear `refreshRetryAfterMicros` (only `markAsValid`/`clearRefreshRetry` do). If a base-commit-driven incremental refresh succeeds after the deadline elapses but before the timer sweep, the next `processRefreshRetries` tick clears the stale deadline and enqueues **one redundant no-op incremental refresh** (opens a WAL writer + reader for a `fromBaseTxn==toBaseTxn` early return). Self-limiting but real wasted work. Fix: clear the marker in `refreshSuccess`/`refreshSuccessNoRows`.
- **m2 — Member ordering (CLAUDE.md alphabetical-within-kind)**: `MatViewRefreshJob` field `busyRetryTimeoutUs` (line 84, `b`) sits after the `c*` fields; `isTableBusy` is `private static` but placed among private *instance* methods (line 804); `tryScheduleBusyRetry` (`t`) precedes `handleErrorRetryRefresh` (`h`). `MatViewState`: `isRefreshDue` (481) precedes `isPendingInvalidation` (486); `clearRefreshRetry`/`scheduleRefreshRetry` are placed after `markAsValid` instead of in alphabetical position.
- **m3 — Modern Java 17** *(`MatViewRefreshJob.java:1146-1150`)*: `if (th instanceof EntryUnavailableException) { ... throw (EntryUnavailableException) th; }` uses an explicit cast while the next branch uses a pattern variable. Prefer `if (th instanceof EntryUnavailableException eue) { throw eue; }`.
- **m4 — Per-view native clock read in the dependent-views loop** *(`MatViewRefreshJob.java:1491`)*: `isRefreshDue(microsecondClock.getTicks())` calls the JNI clock once per dependent view where the value is effectively constant across the pass. Hoist one `getTicks()` above the loop.
- **m5 — User `REFRESH ... INCREMENTAL` during a backoff window is silently deferred** *(`MatViewRefreshJob.java:1596-1599`)*: hits the `isRefreshDue` gate, returns `false` without re-enqueueing, and only completes via the timer re-drive (~1s default). Eventually consistent, but the explicit command has no immediate effect. Consider exempting user-initiated refreshes.
- **m6 — Scope asymmetry**: `tryScheduleBusyRetry` is wired only into the incremental paths; a period view whose base is momentarily busy still cascade-invalidates via `rangeRefresh`→`refreshFailState`. Reasonable scoping, but worth calling out as a known limitation.
- **m7 — Benign clobber race + comment nit**: `MatViewTimerJob`'s `clearRefreshRetry()` is a non-atomic read-then-clear not under the view lock, so it can clobber a freshly re-armed deadline — bounded to one extra attempt, never a stuck view. The comment "the only path that wakes up immediate views" (`MatViewTimerJob.java:282`) is slightly inaccurate: a subsequent base commit also wakes an immediate view once the backoff elapsed.

## Downgraded (false positives — dismissed after source verification)

- **"`insertAsSelect` rethrow double-frees the factory / leaves dirty WAL."** False: the inner per-iteration catch nulls `factory` and rolls back before the throw reaches the outer catch; `EntryUnavailableException` is `NON_CRITICAL` (not OOM) so it skips the retry branch and rethrows cleanly.
- **"Skipped busy view loses base txns / data gap."** False: `notifyBaseRefreshed` is a wakeup/dedup mechanism; the per-view `lastRefreshBaseTxn` is authoritative and `WalPurgeJob` retains the needed WAL because the view stays valid.
- **"Reader/WAL-writer/factory leak on busy paths."** False: traced all three (getReader-in-header, mid-scan rethrow with detach/attach `finally`, getWalWriter-in-header) — all close/re-attach on every path.
- **"`.noLeakCheck()` inside `assertMemoryLeak` is the banned anti-pattern."** False: the `assertMemoryLeak` lambda holds many statements (DDL + inserts + drains + multiple asserts) sharing one leak scope with live tables/views, so deferring the per-query check is the legitimate use.
- **"View stuck in `refreshing` status during backoff is a bug."** False: this is the PR's intended behavior ("shown as refreshing/pending"); the start/finish timestamps reconcile on re-drive.

## Summary
- **Verdict:** Request changes — sound, data-safe core; address the O(V) timer-scan regression (M1), add the cascade / non-zero-backoff / writer-pool tests (M3), and fix metadata/tradeoff disclosure (M2, M4). No blocking correctness, concurrency, or leak defect.
- **Tradeoffs:** O(1)→O(V) per-tick timer cost on a shared hot loop (M1); uncapped retry can pin base WAL under sustained exhaustion (M2) — both honestly weighed against the real win of not spuriously invalidating healthy view chains.
- **Counts:** 18 verified, 5 false positives dropped. All confirmed findings in-diff except the M2 WAL-retention interaction in `WalPurgeJob`.

<sub>🤖 Generated with adversarial level-3 review (11 agents + per-finding source verification)</sub>

bluestreak01:
## Review (level 3 — adversarial, blocking pass)

All 11 review dimensions + 2 fresh-context adversarial agents + per-finding source verification. **23 draft findings examined, 7 confirmed (all Minor), 16 dropped/cleared as false positives or already-fixed.** All confirmed findings are in-diff.

> Note: the earlier level-3 review on this PR targeted a smaller version. The current head has grown to address every blocking item from that pass — the O(V) per-tick timer scan is replaced by a `(deadline,view)` min-heap, the unbounded busy-retry is capped, the cascade/non-zero-backoff/writer-pool test gaps are filled, the block list + cancellation tests were added, and labels are present. This review is of the current head.

### Verdict: Approve with minor nits

The core mechanism is sound and, most importantly, **data-safe**. I independently verified the load-bearing invariant: `WalPurgeJob.getSafeToPurgeUpToTxn` (line 486) pins `safeToPurgeTxn` at each **valid** view's own `max(lastRefreshBaseTxn, refreshIntervalsBaseTxn)`, *not* the `notifyBaseRefreshed` fleet watermark. So a deferred-but-valid view always retains the base WAL it still needs (no data loss), and the new retry **cap** is exactly what bounds the resulting WAL retention — invalidation at line 485 flips `invalid` and releases it. REPLACE_RANGE commits are idempotent and the watermark only advances on the terminal `commitMatView`, so a mid-loop retriable throw persists no partial progress and re-drives cleanly.

The three things most likely to be wrong all hold:
- **`throw (RuntimeException) th` in `insertAsSelect` (line 1190)** — safe: guarded by `isRetriableRefreshError`, which is true only for `EntryUnavailableException extends CairoException` or `isCairoOomError` (requires `instanceof CairoException`); `CairoException extends RuntimeException`. A real `java.lang.OutOfMemoryError` is not a `CairoException`, so it never reaches the cast.
- **Off-latch CAS clear vs under-latch re-arm** — sound: `clearRefreshRetry` CASes on the exact observed deadline; a concurrent under-latch `scheduleRefreshRetry`/`resetRefreshRetry` makes the CAS lose, and the re-arm queued its own heap entry. Worst case is one wasted `enqueueIncrementalRefresh` that the `isRefreshDue` gate then skips.
- **No stuck immediate view** — every armed deadline enqueues a RETRY task on the unbounded `mp.ConcurrentQueue` (same channel as ADD/REMOVE/UPDATE), the single `MatViewTimerJob` drains it, and the per-view cap guarantees termination by invalidation.

## Critical
None.

## Moderate
None.

## Minor

- **m1 — `refreshFail` leaves stale retry state on a now-invalid view** *(in-diff: `MatViewState.java:633-637`, `509-514`)*. `markAsValid` and `resetRefreshRetry` clear `refreshRetryAfterMicros`/`refreshRetryCount`, but `refreshFail`→`markAsInvalid` does not. Path: an immediate view that previously deferred (deadline `d1`, count=1) is re-driven by a **base commit** after `d1` elapses (passes the `isRefreshDue` gate, which — unlike the timer — does not pre-clear the deadline) and then fails **non-retriably** → inner catch → `refreshFailState`. The view becomes `invalid` while still carrying `refreshRetryAfterMicros=d1, refreshRetryCount=1`. Benign in every path I traced (`getViewStatus` checks `invalid` first; `processRefreshRetries` drops the heap entry via its `isInvalid()` guard; `isRefreshDue` is never consulted on invalid views; `markAsValid` clears on recovery), **but** it contradicts the exact invariant the PR's own `testResetState…`/`testInsertFailure…` tests assert ("an invalid view carries no pending retry deadline") — those tests only cover the retriable-after-invalid case, not this one. Fix: clear the two fields in `refreshFail` (or in `tryScheduleRetry`'s non-retriable early return).

- **m2 — Member ordering (alphabetical-within-kind)**. `MatViewState`: `isRefreshDue` (495) precedes `isPendingInvalidation` (500); the new `clearRefreshRetry`/`incrementRefreshRetryCount` sit *after* `markAsValid` instead of before it. `MatViewRefreshJob`: `busyRetryTimeoutUs` (b) is declared after `columnFilter` (c). `PropServerConfiguration`: `matViewRefreshBusyRetryLimit` is placed after `matViewRefreshIntervalsUpdatePeriod` (BusyRetry… sorts before Intervals…).

- **m3 — User-initiated `REFRESH MATERIALIZED VIEW x FULL` on a blocked view is silently skipped** *(in-diff: `MatViewRefreshJob.java:708-712`)*. Logs INFO and returns; the operator gets no error. This is intentional (the block list must stop *every* path, including the explicit command, because the view's refresh may crash the DB) and it's documented, but an explicit command no-opping with only a log line is a footgun. Consider surfacing it in the view's `invalidation_reason`/status, or at minimum keep the documentation prominent.

- **m4 — `busy.retry.timeout=0` permits a bounded tight re-drive/re-arm spin** *(config edge)*. With the timeout at 0 (which the OOM tests use), a re-armed deadline is immediately due every tick, and `processRefreshRetries` returning `true` keeps the shared `mat-view-refresh` worker re-spinning until the cap invalidates. Bounded by `busy.retry.limit` and absent at the 1000ms default, so not a default-config defect — worth a note that 0 is a hot-spin setting.

- **m5 — Per-view native clock read in the fan-out loop** *(in-diff: `MatViewRefreshJob.java:1598`)*. `isRefreshDue(microsecondClock.getTicks())` calls `Os.currentTimeMicros()` (JNI) once per immediate dependent view. Hoisting one read above the loop is the obvious micro-opt — **but** there's a fair counterpoint that, because each iteration can take milliseconds, a fresh per-view read is actually *more* accurate for the due-check, so this is closer to a wash than a clear win. Optional.

- **m6 — Cosmetic status flap to `refreshing` during a re-drive window**. `insertAsSelect` bumps `lastRefreshStartTimestampUs` and does not restore it on a retriable rethrow; after the timer clears the deadline and before the re-drive completes, `view_status` can briefly read `refreshing` instead of `retrying`. Self-correcting, never sticks.

- **m7 — PR metadata**. Body references a production incident but has no `Fixes #NNN`. If a tracking issue exists, reference it at the top. (Labels are correct; title is good Conventional Commits.)

## Downgraded (false positives / already-resolved — dismissed after source verification)

- **"`(RuntimeException) th` can ClassCastException on a real OOM."** False — `isCairoOomError` requires `instanceof CairoException`; the cast only ever sees `CairoException` subtypes (all `RuntimeException`).
- **"Block-list read is an unsynchronized data race."** False — `matViewRefreshBlockList` is a `final` field populated in the config constructor and never mutated in production (only `contains` reads). Final-field safe publication covers it; tests mutate it only by rebuilding the whole config.
- **"Block list isn't case-insensitive."** False — `contains`→`keyIndex`→`Chars.lowerCaseHashCode`/`equalsLowerCase`; matching is case-insensitive on both add and probe.
- **"A deferred immediate view can get permanently stuck."** False — unbounded RETRY queue + capped retries guarantee re-drive-or-invalidate.
- **"Partial progress then throw wrongly counts toward the cap."** False — the watermark advances only at the terminal `commitMatView`, reached only on full completion, immediately followed by `resetRefreshRetry`; a mid-loop throw persists nothing, so not resetting the counter is correct.
- **"Deferred view loses base txns / data gap."** False — `getSafeToPurgeUpToTxn` retains WAL per-valid-view-watermark; `notifyBaseRefreshed` is only the clean/dirty wakeup.
- **Prior M1 (O(V) per-tick scan)** — resolved: replaced by the `peek()`-fast-path min-heap (O(1) common case, O(k log n) due entries). **Prior M2 (uncapped WAL pinning)** — resolved: capped then invalidated. **Prior M3 (test gaps)** — resolved: cascade, non-zero backoff, writer-pool, range-vs-period, and concurrent-re-arm are all now covered.

## Summary
- **Verdict:** Approve — data-safe, leak-free, livelock-free core; well-tested (22 new/updated mat-view tests, no banned `assertSql`/`returnsOnce` patterns). The one genuinely actionable item is the trivial m1 clear in `refreshFail` to make the code hold the invariant its own tests assert; everything else is ordering/metadata/config-edge polish.
- **Tradeoffs (honestly weighed in the body):** the OOM path no longer sleeps between in-call step-halvings (more CPU per round, bounded); a blocked view can pin base-table WAL until dropped/unblocked (documented escape-hatch cost); a terminal OOM now spends ~`limit×backoff` (~10s) in `retrying` before invalidating.
- **Counts:** 7 verified (all Minor), 16 dropped as false positives or already-fixed-since-prior-review. All confirmed findings in-diff; the one out-of-diff interaction (`WalPurgeJob.getSafeToPurgeUpToTxn`) was traced and is the *correct* basis of the data-safety guarantee, not a defect.

<sub>🤖 Generated with adversarial level-3 review (11 dimensions + 2 fresh-context adversarial agents + per-finding source verification)</sub>

bluestreak01:
## Review (level 0 — single-pass, diff + targeted source verification)

Note: this PR already received two level-3 adversarial passes, and the current head addressed every blocking item from them (O(V) timer scan -> min-heap, uncapped retry -> capped, test gaps filled). This pass independently re-verifies the load-bearing safety facts and the residual nits rather than re-deriving the whole thing.

### What I independently verified (and it holds)

- **`throw (RuntimeException) th` in `insertAsSelect` is provably safe.** Guarded by `isRetriableRefreshError(th)`, true only for `EntryUnavailableException` (`extends CairoException`) or `CairoException.isCairoOomError(th)` — and `isCairoOomError` is `t instanceof CairoException && ((CairoException) t).isOutOfMemory()` (`CairoException.java:138`). `CairoException extends RuntimeException` (`CairoException.java:38`). A real `java.lang.OutOfMemoryError` is not a `CairoException`, so it never reaches the cast. No `ClassCastException` is reachable.
- **The RETRY wakeup is never dropped.** `notifyRefreshRetry` enqueues onto `engine.getMatViewTimerQueue()`, an unbounded `ConcurrentQueue` — same channel as ADD/REMOVE/UPDATE. So an armed `refreshRetryAfterMicros` always has a heap entry to re-drive it; no immediate view can get silently stuck.
- **NULL handling is correct.** `refreshRetryAfterMicros` uses `Numbers.LONG_NULL` as the "no pending retry" sentinel, and `isRefreshDue` distinguishes it from a real deadline.
- **Retry counter resets on every forward-progress path** — `resetRefreshRetry()` after `insertAsSelect` success in both `refreshIncremental0` and `rangeRefresh`, on the already-up-to-date early return, and on invalidation. A slow-but-progressing view is never wrongly capped.
- **Data safety / WAL retention** rests on `WalPurgeJob.getSafeToPurgeUpToTxn` pinning at each *valid* view's own watermark, so a deferred-but-valid view retains the base WAL it needs; the new cap bounds the resulting retention.
- **Test assertions are clean:** no banned `assertSql`/`returnsOnce`; `.noLeakCheck()` is used only inside helpers called from genuinely multi-statement `assertMemoryLeak` lambdas (legitimate).

### Critical
None.

### Moderate
None.

### Minor

- **m1 — Member ordering (alphabetical within kind).** Verified against the current head:
  - `MatViewState.java:520` — `isRefreshDue` is declared before `isPendingInvalidation` (520 vs 525); `P` sorts before `R`.
  - `MatViewState.java:565-591` — new `clearRefreshRetry` / `incrementRefreshRetryCount` / `resetRefreshRetry` / `scheduleRefreshRetry` are grouped after `markAsValid` instead of in alphabetical position.
  - `MatViewRefreshJob.java:87` — `busyRetryTimeoutUs` (`b`) is declared after `clusteredIntervals`/`columnFilter` (`c`).
  - `PropServerConfiguration.java` — `matViewRefreshBusyRetryLimit` is placed after `matViewRefreshIntervalsUpdatePeriod`, and `...BusyRetryTimeout`/`...Limit` are swapped (`BusyRetryLimit` < `BusyRetryTimeout` < `Intervals`).

- **m2 — Stale retry state survives a cascade `markAsInvalid` (benign).** `refreshFail` now clears the retry fields (good — closes the prior-review m1), but a view invalidated via the base-table/parent cascade goes through `invalidateView`->`markAsInvalid` directly, which does not clear `refreshRetryAfterMicros`/`refreshRetryCount`. Verified benign on every path: `getViewStatus` checks `invalid` before `retrying`, `processRefreshRetries` drops the heap entry via its `isInvalid()` guard, `isRefreshDue` is never consulted on an invalid view, and `markAsValid` clears both on recovery. Worth clearing them in `markAsInvalid` purely so the "an invalid view carries no pending retry" invariant holds uniformly (which is what the new tests assert for the `refreshFail` path).

- **m3 — `cairo.mat.view.refresh.busy.retry.timeout=0` is a hot-spin setting.** At 0 backoff a re-armed deadline is immediately due every tick and `processRefreshRetries` returning `true` keeps the shared `mat-view-refresh` worker re-spinning until the cap invalidates. Bounded by `busy.retry.limit` and absent at the 1000 ms default, so not a default-config defect — worth a one-line note in the config doc that 0 means "spin until cap".

- **m4 — PR metadata.** Labels and title are correct. The body documents the incident in detail but carries no `Fixes #NNN` — reference the tracking issue at the top if one exists. Tone is analytical and the tradeoffs section is even-handed; the only nit is the heavy inline bold on prose labels.

### Downgraded (false positives — dismissed after source verification)

- **"`throw (RuntimeException) th` can `ClassCastException` on OOM."** False — `isRetriableRefreshError` only admits `CairoException` subtypes.
- **"A lost RETRY task strands a deferred immediate view."** False — `timerTaskQueue` is an unbounded `ConcurrentQueue`; `enqueue` never drops.
- **"`.noLeakCheck()` inside `assertMemoryLeak` is the banned anti-pattern."** False — the wrapping lambdas hold many statements sharing one leak scope.
- **`String.split(",")` in `parseMatViewRefreshBlockList`** — not a finding: startup-only config parse, off every data path.

### Summary

**Verdict: Approve.** The core mechanism — reclassifying `EntryUnavailableException` and tracked-memory OOM as transient, deferred, capped retries — is sound, data-safe, leak-free, and well-tested. I confirmed the three highest-risk facts (cast safety, reliable wakeup, retry-counter reset) directly against source.

Actionable items are all Minor: member-ordering nits (m1), a cosmetically-incomplete invalidation-path cleanup (m2), a config-edge note (m3), and an optional `Fixes #NNN` (m4). None block merge.

**Tradeoffs (disclosed in the body):** a terminal OOM now spends ~`limit x backoff` (~10 s) in `retrying` before invalidating; the OOM path no longer sleeps between in-call step-halvings; a blocked view can pin base-table WAL until dropped/unblocked. All documented with their costs.

**Counts:** all verified findings in-diff (Minor only); 4 candidate findings dropped as false positives. The single cross-context interaction (`WalPurgeJob.getSafeToPurgeUpToTxn`) is the correct basis of the data-safety guarantee, not a defect.

<sub>Generated with a level-0 review pass (diff + targeted source verification).</sub>

bluestreak01:
## Review (level 3 — adversarial, blocking pass)

Change-surface map + 11 review dimensions (4 fresh-context agents incl. adversarial correctness/performance) + per-finding source verification. **6 findings verified (all Minor), 5 dropped as false positives.** No Critical/Moderate correctness, concurrency, resource, or data-loss defects.

Note: this PR has materially evolved since the earlier level-3 comment. Both `Moderate` findings from that pass are **resolved here**: the O(V) per-tick full-fleet timer scan is gone (replaced by a deadline min-heap that is O(1) on the common no-retry tick), and busy-retry is now capped by a per-view consecutive-failure counter that invalidates (releasing WAL retention) once exceeded.

### Verdict: Approve (only Minor nits; no blocking defects)

The reclassification of `EntryUnavailableException`/Cairo-OOM as transient-and-retriable is correct, and the highest-risk areas all hold up under tracing:

- **Data safety / WAL retention (out-of-diff, verified):** `WalPurgeJob.getSafeToPurgeUpToTxn` (`WalPurgeJob.java:494-525`) retains base-table WAL using each non-invalid view's `max(lastRefreshBaseTxn, refreshIntervalsBaseTxn)`. A deferred or block-listed view stays **valid** with an un-advanced watermark, so its WAL is retained and it catches up later — no data loss. `notifyBaseRefreshed` only re-enqueues a task; it does **not** drive purge. Invalidation (cap exceeded) flips `invalid=true` and releases retention.
- **Concurrency:** `refreshRetryAfterMicros` is written under the view latch (plain volatile) and cleared off-latch only via `clearRefreshRetry`'s CAS-on-observed-deadline (`MatViewState.java:565`). Deadlines are strictly increasing, so the CAS cannot mis-clear a freshly re-armed backoff; every arm/re-arm queues its own `RETRY` heap entry on an unbounded `ConcurrentQueue` (no lost wakeup, no double-drive). `retryQueue`/`retryEntryPool` are single-threaded (one `MatViewTimerJob`, `SynchronizedJob`).
- **No double-free / leak:** `throw (RuntimeException) th` in `insertAsSelect`'s outer catch is type-safe (`EntryUnavailableException` and OOM are both `CairoException extends RuntimeException`; `isCairoOomError` requires `instanceof CairoException`). The inner loop already nulled `factory` and rolled back the WAL writer before propagating; `WalWriter`/`TableReader` are all try-with-resources. `cloneInstance()` re-runs the full constructor, so per-worker clones read the new config.
- **Defer-vs-invalidate split is unambiguous:** `periodRefresh = (rangeFrom == LONG_NULL)`; the grammar mandates `RANGE FROM` (`SqlCompilerImpl.java:3545`) and the period timer is the lone `LONG_NULL` caller (`MatViewTimerJob.java:267`). Both halves are tested.

---

### Critical
None.

### Moderate
None. (Prior pass's M1 O(V) timer scan and M2 uncapped-retry WAL pin are both resolved in this version.)

### Minor

- **m1 — Redundant per-view JNI clock read in the dependent-views fan-out** *(in-diff: `MatViewRefreshJob.java:1822`)*. Inside the `for (v…)` loop of `refreshDependentViewsIncremental`, `isRefreshDue(microsecondClock.getTicks())` reads the native clock once per immediate dependent view, even though in the steady state (`refreshRetryAfterMicros == LONG_NULL`) the value is discarded — O(V) unused native clock reads per base commit. The single-view paths (1649, 1975) read once and are fine. Fix: hoist `final long nowMicros = microsecondClock.getTicks();` above the loop.

- **m2 — A view can transiently report `refreshing` after an OOM-defer if it is suspended before the timer re-drive** *(in-diff: `MatViewRefreshJob.java` insertAsSelect ~1016, gates 1972/1963; `MatViewsFunctionFactory.getViewStatus`)*. The OOM defer path bumps `lastRefreshStartTimestampUs` and does **not** restore it before the retriable throw (restore only runs on the two non-throwing early returns). Status reads `retrying` while the deadline is set, but `processRefreshRetries` clears the deadline and enqueues a refresh **without** checking `isViewWriteSuspended`; if the re-driven `refreshIncremental` then returns at the suspend gate without restoring the timestamp, the view shows `refreshing` (start>finish, retrying cleared) until a real refresh runs post-resume. Cosmetic/observability only, self-heals, no data impact; narrow reachability (OOM defer + WAL-suspend during the backoff). Optional fix: restore the start timestamp on the retriable-throw path.

- **m3 — Member ordering (CLAUDE.md alphabetical-within-kind)** *(in-diff)*. Fields are fixed, but: `MatViewRefreshJob.tryScheduleRetry` (t, 892) sits before `handleErrorRetryRefresh` (h, 922); `MatViewState.isRefreshDue` (520) precedes `isPendingInvalidation` (525); and `clearRefreshRetry`/`incrementRefreshRetryCount`/`resetRefreshRetry`/`scheduleRefreshRetry` (565-591) sit after `markAsValid` instead of in alphabetical position. (The new `private static` helpers among instance methods are consistent with the file's existing pattern.)

- **m4 — `cairo.mat.view.refresh.busy.retry.timeout=0` collapses the retry budget** *(behavioral)*. With timeout 0 the deadline is `now`, so a persistently-failing view is re-driven every tick and burns the whole `busy.retry.limit` almost instantly, then invalidates. Tests use this deliberately and the default is 1000ms, but an operator setting 0 expecting "fast infinite retries" gets near-immediate invalidation. Worth a one-line note in the config doc.

- **m5 — `Fixes #NNN` missing** *(metadata)*. The body describes a concrete production incident but carries no `Fixes #NNN`. If a tracking issue exists, reference it at the top. (Title format, verb repetition, tone, and labels are good.)

- **m6 — Minor test-coverage gaps** *(in-diff tests)*. `testTimerClearRefreshRetryDoesNotClobberConcurrentReArm` only simulates the off-latch clear/re-arm interleaving single-threaded (the real cross-thread race is unexercised). `NoOp`/`Forwarding` `notifyRefreshRetry` and the `state==null`/`isDropped`/`isPendingInvalidation` drop branches in `processRefreshRetries` have no direct coverage (defensive, low risk). Trivial: `parseMatViewRefreshBlockList` splits on `,`, so a view name containing a comma can't be block-listed.

---

### Downgraded (false positives — dismissed after source verification)

- **"The retry counter resets without progress for period views, weakening the WAL-retention cap."** Dismissed — a reset needs a *non-throwing* `insertAsSelect` between failures, but a persistently-OOMing complete period is unrefreshed work, so every re-drive re-hits the OOM and the counter accumulates to the cap. Reset fires only on genuine no-work/success.
- **"`throw (RuntimeException) th` can throw `ClassCastException`."** Dismissed — only reached for `CairoException` subtypes, all `RuntimeException`; a `java.lang.OutOfMemoryError` is not matched.
- **"Removing `Os.sleep` creates a tight OOM spin."** Dismissed — each OOM `continue` advances `i` and halves `naturalStep` under `i < maxRetries && naturalStep > 1`, then defers; bounded, resources freed per iteration.
- **"Block-list tests pass vacuously / block list isn't read live."** Dismissed — `CairoTestConfiguration.getDelegate()` rebuilds `PropServerConfiguration` from overrides on change, so tests drive the real path; production is correctly startup-only (and documented).
- **"`markAsValid`/`refreshFail` plain writes race the timer CAS and lose a clear."** Dismissed — `volatile long` is atomic and every interleaving against CAS-on-observed-deadline resolves correctly (guarded by the regression test).

---

### Summary
- **Verdict:** Approve. Sound, data-safe, well-tested — the design (deadline min-heap + CAS-on-observed-deadline off-latch clear + capped consecutive-failure counter) is well-guarded, and the suite covers the incident scenario (cascade non-invalidation, non-zero backoff, writer-pool path, staggered deadlines, retry cap, counter reset, period-vs-user range divergence, m1/m3 stale-invalid windows).
- **Tradeoffs (disclosed in the body):** blocked/deferred views over-retain base WAL until they advance/recover/invalidate; deferred OOM rebuilds the step each round; `view_status` gains `retrying`.
- **Counts:** ~13 candidates examined, **6 verified (all Minor)**, **5 false positives dropped**. All 6 verified findings are in-diff; the one significant out-of-diff interaction (WalPurgeJob retention) was verified **correct/safe**.

<sub>🤖 Generated with adversarial level-3 review (change-surface map + 4 fresh-context agents + per-finding source verification)</sub>

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=245785)
**Tested Commit:** `0c9c5d50af7738920156fcd67531436039d3d073`

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 153 / 167 (91.62%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/mv/ForwardingMatViewStateStore.java|0|2|00.00%|
|:large_blue_circle:|io/questdb/cairo/mv/NoOpMatViewStateStore.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewTimerTask.java|6|7|85.71%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|60|66|90.91%|
|:large_blue_circle:|io/questdb/PropServerConfiguration.java|18|19|94.74%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewTimerJob.java|36|38|94.74%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewState.java|18|18|100.00%|
|:large_blue_circle:|io/questdb/cairo/DefaultCairoConfiguration.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/PropertyKey.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/MatViewsFunctionFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewStateStoreImpl.java|3|3|100.00%|

PR #7276, opened 2026-06-17
test(qwp): assert proper treatment of empty and NULL arrays on egress

mtopolnik:
## What

Bug discovered in Java client: it didn't accept empty arrays from server on egress. The tandem PR in client repo addresses that. This PR contributes an end-to-end test for the behavior, and must also bump the client module pointer to the `main` branch after merging that PR.

## Changes

- `QwpEgressBootstrapTest.testEmptyAndNullArrayColumns`: ingest a regular, an empty (`ARRAY[]`), and a NULL `DOUBLE[]` into a booted server and read them back via `QwpQueryClient`, asserting the empty array decodes as a non-null 0-element value distinct from NULL. Without the decoder fix this reproduces the egress rejection (`ARRAY dim 0 must be >= 1: 0`).

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -9,9 +9,9 @@ jobs:
     if: ${{ github.event.pull_request.head.repo.full_name == github.repository }} # Only run on non-forked PRs
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v5
       - name: Use Node.js 22.x
-        uses: actions/setup-node@master
+        uses: actions/setup-node@v5
         with:
           node-version: 22.x
       - name: install danger
--- a/.github/workflows/gitleaks.yml
+++ b/.github/workflows/gitleaks.yml
@@ -12,7 +12,7 @@ jobs:
     env:
       HAS_GITLEAKS_LICENSE: ${{ secrets.GITLEAKS_LICENSE != '' }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           fetch-depth: 0
       - uses: gitleaks/gitleaks-action@ff98106e4c7b2bc287b24eaf42907196329070c7
--- a/.github/workflows/glibc_smoke_test.yml
+++ b/.github/workflows/glibc_smoke_test.yml
@@ -24,36 +24,37 @@ jobs:
       # Use Amazon Linux 2 image - uses GLIBC 2.26
       image: amazonlinux:2
       volumes:
-        - /node20217:/node20217
-        - /node20217:/__e/node20
+        - /nodeglibc217:/nodeglibc217
+        - /nodeglibc217:/__e/node24
     steps:
       - name: Install dependencies
         run: yum install -y curl git tar gunzip xz wget
       - name: Print GLIBC version
         run: ldd --version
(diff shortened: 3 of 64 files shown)


mtopolnik:
# Review: PR #7276

**Level 2.** Scope: one new test method in `QwpEgressBootstrapTest` (in-diff) + a `java-questdb-client` submodule pointer bump that locks in the client-side decoder fix (out-of-diff, in the submodule). I reviewed both the test and the client fix it ships.

The code is essentially correct. The decoder fix is sound, the test exercises real server behavior, and resource handling is clean.

## Critical 

None. 

## Minor

### M1 — `testEmptyAndNullArrayColumns` is in the wrong alphabetical position *(in-diff)*
It's inserted at line ~208, right after `testArrayColumns` (grouped by topic). Alphabetically it belongs between `testDropRecreateRecompilesStaleFactory` and `testEmptyResultSet` (i.e. just before line 709): `testEmpty**A**ndNull…` < `testEmpty**R**esultSet`. This violates the project rule ("insert new methods in the correct alphabetical position", "never group methods by logical category") — and your own recorded preference to sort test methods alphabetically rather than grouping by category. Move it down to the `testEmpty…` cluster.

### M3 — Nit: optional multi-dimensional empty-array coverage
The e2e test covers the 1-D empty array. The decoder fix also added the per-dimension cap that matters for multi-dim frames with a `0` in one dimension and a large value in another. That hostile path is already covered by the client-side `QwpResultBatchDecoderHardeningTest`, so this is not a gap — but an end-to-end `ARRAY[]::DOUBLE[][]` case (if SQL-creatable) would exercise the multi-dim empty encoding through the real server. Optional, low value.

### M4 — Nit: unbounded index into fixed-size result arrays
`isNull[count[0]++]` writes into length-3 arrays with no bound. If a server bug ever streamed >3 rows, this throws `ArrayIndexOutOfBoundsException` inside `onBatch` — which the outer `catch (Throwable)` converts to `errMsg`, so the test still fails (just with a murkier message than the explicit `assertEquals(3, count[0])`). Not a real risk; noting for completeness.

## Summary

**Verdict: approve the code**

- The decoder fix is correct and complete; the test exercises genuine server behavior (server emits empty `DOUBLE[]` inline as non-null `nDims=1, dim[0]=0`; storage preserves empty-vs-null), is deterministic (monotonic designated timestamps + single partition), is a valid regression (fails on the old decoder with "ARRAY dim 0 must be >= 1: 0"), and adds value over the client-only unit test by validating the server↔client contract end to end.
- Findings: **1 critical (release sequencing, already on the author's checklist), 4 minor/nits.** All in-diff or metadata. 6 draft concerns verified and dropped as false positives.
- No correctness, concurrency, resource-leak, or performance issues in the test or the fix.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 5 / 8 (62.50%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/std/ConcurrentLongHashMap.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/std/ConcurrentIntHashMap.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/std/ConcurrentHashMap.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/SqlException.java|5|5|100.00%|

bluestreak01 (approved):

PR #7277, opened 2026-06-17
chore(build): migrate GitHub Actions workflows off the deprecated Node 20 runtime

RaphDal:
## Background

GitHub is [deprecating Node.js 20 on Actions runners](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/).
During the current brownout the runner ignores an action's declared `node20` runtime and forces every JavaScript action onto its bundled Node.js 24. This has two distinct consequences for our workflows:

- Jobs that run inside a container with **old glibc** fail hard. The runner's Node.js 24 binary needs glibc >= 2.28, but these containers ship older glibc, so `actions/checkout` (and any other JS action) aborts before doing any work with `libc.so.6: version 'GLIBC_2.27' not found`.
- Every other workflow still runs, but each JS action now emits a Node.js 20 deprecation annotation. Those become hard failures once GitHub removes the Node.js 20 fallback entirely.

Two runs that surfaced this:

- `Build and Push Release Rust Libraries` (`rebuild_rust.yml`) - the `build-all-linux-x86-64` job failed at `Run actions/checkout@v4`.
- `GLIBC Smoke Test` (`glibc_smoke_test.yml`) on PR #7268 - the `amd64` job failed at the same step.

### 1. Node runtime for old-glibc container jobs

Three jobs build portable binaries inside containers whose glibc predates the Node.js 24 requirement:

- `glibc_smoke_test.yml` - `amd64` (`amazonlinux:2`, glibc 2.26)
- `rebuild_rust.yml` - `build-all-linux-x86-64` (`manylinux2014_x86_64`, glibc 2.17)
- `rebuild_native_libs.yml` - `build-all-linux-x86-64` (`manylinux2014_x86_64`, glibc 2.17)

Each already carried a hack that downloaded a glibc 2.17-compatible Node.js 20 and mounted it over `/__e/node20`. The runner no longer looks there - it now uses `/__e/node24` - so the override stopped taking effect.

The fix follows the same mechanism, updated for the new runtime path:

- The container now mounts the override directory over both `/__e/node20` and `/__e/node24`, so the custom binary is used regardless of which path the runner selects.
- The install step downloads Node.js 22.22.3 instead of 20.9.0. Node.js 24 has no glibc 2.17 build on `unofficial-builds.nodejs.org`; the newest available glibc 2.17 build is Node.js 22 (the current LTS line). A single glibc 2.17 build covers both the glibc 2.17 and glibc 2.26 containers.

Tradeoff: JS actions in these three jobs run on Node.js 22 rather than 24. That is an intentional gap - it is the newest runtime with a glibc 2.17 build, and the actions we run there (`checkout`, `cache`) execute correctly on it. The jobs in glibc 2.28 containers (`almalinux:8`, `manylinux_2_28_aarch64`, `debian:buster`, the async-profiler builders) already run on the stock Node.js 24 and need no override.

### 2. Action version bumps

Every first-party action was bumped to the lowest major that ships a `node24` runtime, which clears the deprecation warnings while keeping the version jump as small as possible:

| Action                                         | Before        | After | Runtime |
| ---------------------------------------------- | ------------- | ----- | ------- |
| `actions/checkout`                             | v4 / `master` | v5    | node24  |
| `actions/cache`, `cache/save`, `cache/restore` | v4 / v3       | v5    | node24  |
| `actions/setup-java`                           | v4            | v5    | node24  |
| `actions/setup-python`                         | v5            | v6    | node24  |
| `actions/setup-node`                           | v3 / `master` | v5    | node24  |
| `actions/setup-go`                             | v4            | v6    | node24  |
| `actions/setup-dotnet`                         | v3            | v5    | node24  |
| `actions/upload-artifact`                      | v4            | v6    | node24  |
| `actions/download-artifact`                    | v4            | v7    | node24  |

Notes and tradeoffs:

- `cache/save@v3` and `cache/restore@v3` were the most outdated - they targeted Node.js 16, which GitHub removed earlier - so they were already at risk independent of the Node.js 20 deprecation.
- `upload-artifact` (v4 -> v6) and `download-artifact` (v4 -> v7) are larger jumps because v5/v6 of those actions kept Node.js 20 as the default. The intermediate releases carry no behavioral change for our usage: the only real break is `download-artifact` v5 changing the output path for single downloads **by ID**, and both call sites download **by name**. v6/v7 require Actions runner >= 2.327.1, which GitHub-hosted runners satisfy.
- `setup-go` moves v4 -> v6, and its `go-version` is raised from 1.24 to 1.25 in the same step (the PGWire stable Go client scenarios) so the Go toolchain stays on a maintained release alongside the action bump.
- `danger.yml` pinned `actions/checkout` and `actions/setup-node` to floating `@master`. Both are now pinned to `@v5`, which makes those runs reproducible and removes a moving dependency, at the cost of needing a manual bump for future majors.
- Third-party actions pinned by commit SHA (`ad-m/github-push-action`, `gitleaks/gitleaks-action`, `actions-rust-lang/setup-rust-toolchain`, `questdb/r-actions/setup-r`) were left untouched.

### 3. Distinct log artifact names in the GLIBC smoke test

The `amd64` and `aarch64` jobs in `glibc_smoke_test.yml` both uploaded their failure logs as an artifact named `logs`. `actions/upload-artifact` v4 and newer treat artifact names as immutable and unique within a run, so a run where both jobs failed would abort the second upload on a name collision. Each job now uploads under a distinct name (`logs-amd64`, `logs-aarch64`).

## Risks

- The three container jobs now depend on Node.js 22 staying available as a glibc 2.17 build on `unofficial-builds.nodejs.org`. This is the same external dependency the previous Node.js 20 hack relied on, not a new one.
- Running actions on Node.js 22 in those jobs is a deliberate compromise. Once a glibc 2.17 Node.js 24 build appears, or once the containers move to glibc 2.28+, the override can be dropped and the jobs can use the stock runtime.
- The bumps move several actions across multiple majors at once. The artifact-action changelogs were reviewed (above); the remaining bumps are runtime-only releases.

## Test plan

- Re-run the two workflows that failed and confirm the previously-failing
  container jobs now get past `Run actions/checkout` and complete:
  - `Build and Push Release Rust Libraries` (manual `workflow_dispatch`)
  - `GLIBC Smoke Test` (re-trigger on a PR that touches
    `core/src/main/resources/io/questdb/bin/**`)
- Confirm the Node.js 20 deprecation annotations no longer appear on the
  PR-triggered workflows: `pgwire_stable.yml`, `r_test.yml`,
  `parquet_compat.yml`, `danger.yml`, `gitleaks.yml`, `rust_license_check.yml`.
- Exercise the manually-triggered rebuild workflows on a throwaway branch to
  confirm the cache save/restore and artifact upload/download steps still pair
  up: `rebuild_native_libs.yml`, `rebuild_rust_test.yml`,
  `rebuild_async_profiler.yml`, `rebuild_win64svc.yml`.
- Already validated locally: every workflow file still parses as YAML, and each
  bumped tag exists and resolves to a `node24` runtime.

### DIFF
--- a/.github/workflows/rebuild_native_libs.yml
+++ b/.github/workflows/rebuild_native_libs.yml
@@ -74,7 +74,6 @@ jobs:
       image: quay.io/pypa/manylinux2014_x86_64
       volumes:
         - /nodeglibc217:/nodeglibc217
-        - /nodeglibc217:/__e/node20
         - /nodeglibc217:/__e/node24
     steps:
       - name: Install tools, most are needed to build nasm
@@ -98,8 +97,7 @@ jobs:
         # Q: Why do we need this hack at all? A: Many GitHub actions, e.g. actions/checkout, run on Node.js.
         # GitHub deprecated Node.js 20 and now forces actions onto its Node.js 24 build, which needs a newer glibc
         # than manylinux2014 has. Node.js 24 has no glibc 2.17 build, so we install the newest glibc 2.17-compatible
-        # release (Node.js 22 LTS) and mount it over both /__e/node20 and /__e/node24, so it is used no matter which
-        # runtime the runner picks.
+        # release (Node.js 22 LTS) and mount it over /__e/node24, the path the runner now resolves every action to.
         run: |
           curl -LO https://unofficial-builds.nodejs.org/download/release/v22.22.3/node-v22.22.3-linux-x64-glibc-217.tar.xz
           tar -xf node-v22.22.3-linux-x64-glibc-217.tar.xz --strip-components 1 -C /nodeglibc217
--- a/.github/workflows/rebuild_native_libs.yml
+++ b/.github/workflows/rebuild_native_libs.yml
@@ -25,7 +25,7 @@ jobs:
         os: [macos-14, macos-15-intel]
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           submodules: true
       - name: Install toolchains (CXX/NASM)
@@ -54,14 +54,14 @@ jobs:
           cp core/target/classes/io/questdb/bin-local/libquestdb.dylib core/src/main/resources/io/questdb/bin/darwin-x86-64/
       - name: Save darwin-aarch64 Libraries to Cache
         if: ${{ matrix.os == 'macos-14' }}
-        uses: actions/cache/save@v3
+        uses: actions/cache/save@v5
         with:
           path: |
             core/src/main/resources/io/questdb/bin/darwin-aarch64/libquestdb.dylib
           key: nativelibs-armosx-${{ github.sha }}
       - name: Save darwin-x86-64 Libraries to Cache
         if: ${{ matrix.os == 'macos-15-intel' }}
-        uses: actions/cache/save@v3
+        uses: actions/cache/save@v5
         with:
           path: |
             core/src/main/resources/io/questdb/bin/darwin-x86-64/libquestdb.dylib
@@ -73,8 +73,9 @@ jobs:
     container:
       image: quay.io/pypa/manylinux2014_x86_64
       volumes:
-        - /node20217:/node20217
-        - /node20217:/__e/node20
+        - /nodeglibc217:/nodeglibc217
+        - /nodeglibc217:/__e/node20
+        - /nodeglibc217:/__e/node24
     steps:
       - name: Install tools, most are needed to build nasm
         run: |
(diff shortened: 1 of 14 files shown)


.github/workflows/rebuild_native_libs.yml
```diff
@@ -73,8 +73,9 @@ jobs:
     container:
       image: quay.io/pypa/manylinux2014_x86_64
       volumes:
-        - /node20217:/node20217
-        - /node20217:/__e/node20
+        - /nodeglibc217:/nodeglibc217
+        - /nodeglibc217:/__e/node20
```
jovfer: do we need to keep node20?

RaphDal: Good catch - removed in 3f8253f0d1. Now that every action is bumped to a node24-native version, nothing in these container jobs declares the node20 runtime, so the runner only ever resolves `/__e/node24`. The `/__e/node20` mount was dead weight carried over from the pre-bump hack, so all three workflows now mount only `/__e/node24`.

jovfer (approved):

bluestreak01 (approved):

PR #7280, opened 2026-06-18
perf(sql): evaluate runtime-constant subexpressions once per query instead of per row

puzpuzpuz:
## What

A subexpression whose value is invariant across the rows of a single cursor (a "runtime constant") but that is not a compile-time constant is currently re-evaluated for every row when it appears inside a per-row context such as a `CASE` branch or a filter predicate.

A typical example is a time-window threshold:

```sql
SELECT sum(CASE WHEN ts >= dateadd('d', -30, to_timezone(now(), 'Asia/Kolkata'))
                THEN amount ELSE 0 END)
FROM trades;
```

`now()` is already cached, but `to_timezone(...)` performs a timezone-offset lookup and `dateadd(...)` re-derives on top of it, both repeated on every row even though the result is the same for the whole scan. Because QuestDB does no scalar common-subexpression elimination, each textual occurrence is recomputed independently.

This change evaluates such a subexpression once at cursor open and reuses the cached value for every row.

## How

- `RuntimeConstFunction` wraps a runtime-constant subtree, evaluates it once in `init()`, caches the result in a primitive field (no per-row allocation), and serves it from every getter. Fixed-width scalar types (timestamp, long, int, double, date, boolean, ipv4, geo, uuid, ...) are cached; getters for other types delegate to the argument. The wrapper delegates `toPlan()` to its argument, so `EXPLAIN` output is unchanged.
- `FunctionParser.visit()` folds at boundaries only. When the function being built is itself runtime constant, nothing is wrapped (its enclosing boundary will wrap the larger subtree); when it is not, each runtime-constant argument is a maximal runtime-constant subtree and gets wrapped. This wraps the topmost node only and never nests.
- Trivial runtime-constant leaves (bind variables, `now()`) already cache their value, so the fold skips them. This avoids adding per-row indirection to common predicates such as `col > $1`.

The existing once-per-query `WHERE ... -> interval scan` path is unaffected: it relies on the same `isRuntimeConstant()` contract and reads the bound after `init()`, exactly as the wrapper does.

## Scope and limitations

- This is a per-row CPU optimization only. It does not change result sets, memory footprint, or the parallel/serial plan choice. Its magnitude is secondary: in aggregation-heavy queries the hash-map probe and the accumulator updates dominate, so the threshold recompute is typically a low-single-digit fraction of filter/projection CPU rather than a step change. Queries without runtime-constant subexpressions in per-row contexts see no change.
- Only fixed-width scalar types are folded. Variable-length and object types (varchar, string, array, interval, long256, decimal128/256) are left untouched for now and can be added later.
- Folding is applied to function arguments at boundaries, not to a wholly runtime-constant projection root (e.g. `SELECT dateadd('d', -1, now()) FROM t`), which has no enclosing boundary and is still computed per row. The common case - thresholds inside `CASE`/filters - is covered.
- Scalar common-subexpression elimination (sharing identical subtrees) is out of scope here; two identical subtrees are still folded independently.

## Test plan

- New `RuntimeConstFunctionTest`:
  - proves the wrapped subtree is evaluated exactly once per cursor via a call counter that stays at 1 across 100 row reads
  - proves only the maximal subtree is wrapped (no double-wrapping below it)
  - proves trivial runtime-constant leaves are not wrapped
  - proves a wholly runtime-constant expression with no enclosing boundary is left alone
  - end-to-end: a folded bind-variable threshold inside a `CASE` returns identical results to the literal-threshold reference
  - end-to-end: the wrapper does not leak into `EXPLAIN` output
- Existing suites pass unchanged: `FunctionParserTest`, `WhereClauseParserTest` (interval-scan path), `ExplainPlanTest` (plan transparency, 548 tests), `SqlCodeGeneratorTest`, `ParallelFilterTest`, `ParallelGroupByFuzzTest`, bind-variable and CASE/timestamp-arithmetic suites - roughly 2300 tests across the affected paths.


puzpuzpuz:
I'm debugging the `QwpEgressCreditFlowTest` test failure separately.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 174 / 176 (98.86%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/RuntimeConstFunction.java|161|163|98.77%|
|:large_blue_circle:|io/questdb/griffin/FunctionParser.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/json/JsonExtractFunction.java|1|1|100.00%|

bluestreak01 (approved):
Approving after a level-3 review.

Verified locally: build compiles; RuntimeConstFunctionTest (18), FunctionParserTest (120), WhereClauseParserTest (846), and all 4 json suites (45) pass.

Core mechanism is sound. Checked the high-risk paths:
- Interface-default dispatch resolves correctly (UnaryFunction/RuntimeConstFunction defaults win over the inherited Function defaults), so close()/toTop()/offerStateTo() delegate to the arg and isRuntimeConstant/shouldMemoize/toPlan are correct. No resource leak.
- Parallel filter: shared path is read-only after init() (happens-before via the reduce queue); non-thread-safe path clones per worker and each clone re-inits. Safe.
- JIT serializes the AST, not the function tree, so the wrapper can't affect JIT correctness.
- WHERE -> interval-scan path compiles the bound standalone, so it gets the raw runtime-const function (never wrapped). Unaffected.
- Audited every isRuntimeConstant() override in core: json_extract was the sole violator and it is correctly fixed; rnd/interval/geo are safe.

Non-blocking follow-ups:
- PR body should describe the bundled json_extract correctness fix in its own section and add the Bug label.
- Consider a defensive guard documenting/asserting the 'isRuntimeConstant() true => getters must not touch the record' contract, since correctness relies on it globally.
- Minor: boolean field named 'value', isComposite member ordering, and the description's 'delegate to the argument' wording (cross-type numeric getters actually route through the native getter).

PR #7282, opened 2026-06-18
fix(sql): fix wrong SAMPLE BY first()/last() results when the time column is not selected

ideoma:
## What changed

A `SAMPLE BY` query that aggregates with `first()` / `last()`, filters on an
indexed `SYMBOL` column, and uses `ALIGN TO FIRST OBSERVATION` could return a
bucket timestamp where an aggregated value was expected. After this change such
queries return the correct values.

The wrong results appeared only when the designated timestamp column was left
out of the `SELECT` list, and only for the buckets in the middle of the result.
The first and last buckets were always correct, which made the problem easy to
miss. There was no error and no crash, so an affected query silently produced a
plausible-looking timestamp in place of the value.

## Reproduction

```sql
CREATE TABLE t (sym SYMBOL INDEX, val LONG, ts TIMESTAMP) TIMESTAMP(ts) PARTITION BY DAY;
INSERT INTO t
SELECT ('s' || (x % 5)::STRING)::SYMBOL, (x * 3)::LONG,
       timestamp_sequence('2024-01-01T00:00:00.000000Z', 600_000_000)
FROM long_sequence(60);

SELECT sym, first(val) fv, last(val) lv
FROM t WHERE sym = 's1' SAMPLE BY 1h ALIGN TO FIRST OBSERVATION;
```

Before the fix, `lv` for the middle buckets came back as the bucket-start
timestamp (e.g. `1704070800000000`) instead of the aggregated value (e.g. `33`).
With the fix, every bucket reports the expected value.

Adding the timestamp to the projection, switching to `ALIGN TO CALENDAR`, or
dropping the indexed-symbol filter all avoided the problem, because each of
those routes the query away from the affected execution path.

## Root cause

`SampleByFirstLastRecordCursorFactory` emits the middle (non-boundary) rows
through its data record. The data record special-cased the bucket-timestamp
column using the base-table timestamp index, while the column it was handed is a
projection index. The two indexes coincide only when the timestamp is projected
at the same position it occupies in the base scan. When the timestamp was not
projected, the base index pointed at an unrelated projected column (the trailing
aggregate) and overwrote it with the bucket timestamp. The boundary rows, which
travel through a separate record that already used the projection index, were
unaffected, which is why the first and last buckets stayed correct.

The fix makes the data record use the same projection index as the boundary
record. When the timestamp is not projected, no column matches the special case
and every aggregate reports its stored value.

## Scope and risk

- The change is limited to one comparison on the read path of this factory; it
  is an index comparison only, with no extra work, so query performance is
  unchanged.
- Queries that already projected the designated timestamp were correct before
  and remain correct; their plans and results do not change.
- Factory selection is unchanged: the same queries route to this factory as
  before.

## Test plan

- Added `SampleByTest#testSampleByFirstLastReturnsValuesWhenTimestampNotProjected`,
  which fails on master and passes with the fix. It asserts both the
  timestamp-projected form (a regression guard that was already correct) and the
  timestamp-omitted form (the bug).
- Ran the `SampleByTest#testSampleByFirstLast*` and
  `ExplainPlanTest#testSampleByFirstLast*` suites to confirm factory selection
  and plans are unchanged.
- Ran the full `SampleByTest` (313 tests) and `SampleByNanoTimestampTest`
  (279 tests) classes; all pass.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/groupby/SampleByFirstLastRecordCursorFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/groupby/SampleByFirstLastRecordCursorFactory.java
@@ -980,12 +980,15 @@ public class SampleByFirstLastRecordCursorFactory extends AbstractRecordCursorFa
                 public long getLong(int col) {
                     long pageAddress = pageAddresses[col];
                     if (pageAddress > 0) {
-                        if (col != timestampIndex) {
+                        if (col != groupByTimestampIndex) {
                             return Unsafe.getLong(pageAddress + (getRowId(firstLastIndexByCol[col]) << 3));
                         }
-                        // Special case - timestamp the sample by runs on
-                        // Take it from timestampOutBuff instead of column
-                        // It's the value of the beginning of the group, not where the first row found
+                        // Special case - the projected designated timestamp the sample by runs on.
+                        // Take it from the sample period buffer instead of the column, so it reports
+                        // the value of the beginning of the group, not where the first row was found.
+                        // groupByTimestampIndex is the projection-space index of that column (-1 when
+                        // it is not projected); comparing against the base-table timestampIndex would
+                        // alias an unrelated projected column whenever the two index spaces disagree.
                         return samplePeriodAddress.get(getRowId(TIMESTAMP_OUT_INDEX) - prevSamplePeriodOffset);
                     } else {
                         return Numbers.LONG_NULL;
--- a/core/src/test/java/io/questdb/test/griffin/engine/groupby/SampleByTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/engine/groupby/SampleByTest.java
@@ -6483,6 +6483,56 @@ public class SampleByTest extends AbstractCairoTest {
     }
 
     @Test
+    public void testSampleByFirstLastReturnsValuesWhenTimestampNotProjected() throws Exception {
+        assertMemoryLeak(() -> {
+            execute("CREATE TABLE t (sym SYMBOL INDEX, val LONG, ts TIMESTAMP) TIMESTAMP(ts) PARTITION BY DAY");
+            execute("""
+                    INSERT INTO t
+                    SELECT ('s' || (x % 5)::STRING)::SYMBOL AS sym,
+                           (x * 3)::LONG AS val,
+                           timestamp_sequence('2024-01-01T00:00:00.000000Z', 600_000_000) AS ts
+                    FROM long_sequence(60)""");
+
+            // Reference: with the designated timestamp projected, first()/last() are correct.
+            // This selects the same SampleByFirstLast factory.
+            assertQuery(
+                    "SELECT ts, sym, first(val) fv, last(val) lv FROM t WHERE sym = 's1' SAMPLE BY 1h ALIGN TO FIRST OBSERVATION"
+            ).timestamp("ts").noRandomAccess().returns("""
+                    ts\tsym\tfv\tlv
+                    2024-01-01T00:00:00.000000Z\ts1\t3\t18
+                    2024-01-01T01:00:00.000000Z\ts1\t33\t33
+                    2024-01-01T02:00:00.000000Z\ts1\t48\t48
+                    2024-01-01T03:00:00.000000Z\ts1\t63\t63
+                    2024-01-01T04:00:00.000000Z\ts1\t78\t78
+                    2024-01-01T05:00:00.000000Z\ts1\t93\t108
+                    2024-01-01T06:00:00.000000Z\ts1\t123\t123
+                    2024-01-01T07:00:00.000000Z\ts1\t138\t138
+                    2024-01-01T08:00:00.000000Z\ts1\t153\t153
+                    2024-01-01T09:00:00.000000Z\ts1\t168\t168
+                    """);
+
+            // Without the designated timestamp projected, the data-record path used to alias the
+            // trailing last(val) onto the bucket-start timestamp for the middle buckets. The first
+            // and final buckets travel through the cross-frame record and were always correct.
+            assertQuery(
+                    "SELECT sym, first(val) fv, last(val) lv FROM t WHERE sym = 's1' SAMPLE BY 1h ALIGN TO FIRST OBSERVATION"
+            ).noRandomAccess().returns("""
+                    sym\tfv\tlv
+                    s1\t3\t18
+                    s1\t33\t33
+                    s1\t48\t48
+                    s1\t63\t63
+                    s1\t78\t78
+                    s1\t93\t108
+                    s1\t123\t123
+                    s1\t138\t138
+                    s1\t153\t153
+                    s1\t168\t168
+                    """);
+        });
+    }
+
+    @Test
     public void testSampleByFirstLastWithNonTsOrFilteredSymbolColumn() throws Exception {
         Rnd rnd = TestUtils.generateRandom(LOG);
         setProperty(PropertyKey.DEBUG_CAIRO_COPIER_TYPE, rnd.nextInt(4));


ideoma:
## Review (level 2)

I verified the fix empirically (reverted the one line → reproduced the bug; restored → confirmed green), traced the index spaces through the constructor and `SqlCodeGenerator`, audited every sibling use of `timestampIndex`, and ran an adversarial pass over the surrounding factory that turned up one verified pre-existing bug in the same file.

### The fix is correct

The change is `col != timestampIndex` → `col != groupByTimestampIndex` in `SampleByDataRecord.getLong()`.

- `col` and `groupByTimestampIndex` are both **projection-space**; `timestampIndex` is **base-table space** (it comes from `getTimestampIndex()` → `factory.getMetadata()` and is compared against `baseMetadata.getColumnIndex(...)` in `buildFirstLastIndex`). The old comparison mixed the two spaces.
- Empirical proof: with the one line reverted, the new test's **non-projected** assertion (`SELECT sym, fv, lv`) fails — `lv` (projection index 2) is aliased to the bucket-start timestamp (`1704070800000000` instead of `33`) for the middle buckets — while the **projected** assertion passes. Restoring the fix makes both pass. This matches the root-cause narrative exactly.
- The new behavior is strictly an improvement: it special-cases *exactly* the projected designated-timestamp key column (`groupByTimestampIndex`, or nothing when it is `-1`). It can never turn a correct result wrong, because `col` and `groupByTimestampIndex` are always in the same projection space by construction.
- Audited every other `timestampIndex` read in the file: all are correctly in base/frame space (column-type lookup, key detection, `frameMemory.getPageAddress` which takes a base/frame index). No sibling instances of the same confusion near `getLong`.
- Only `getLong`/`getTimestamp` need the special case — the designated timestamp is LONG-backed and is never read through `getInt`/`getDouble`/`getSymA`/etc. `getTimestamp` correctly delegates to `getLong`. The cross-frame record path was already correct (`saveFirstLastValuesToCrossFrameRowBuffer` already uses `groupByTimestampIndex`), which is why the boundary buckets were always right.

PR metadata, commit message, and test style all conform. No correctness, concurrency, resource, or performance concerns in the diff (a single int-field comparison; identical cost).

### Moderate (out-of-diff, pre-existing — worth bundling)

**`toPlan()` uses a projection-space index where a base-table index is required — wrong EXPLAIN names, and an `AssertionError` crash.**

`SampleByFirstLastRecordCursorFactory.toPlan()` calls `sink.putBaseColumnName(i)` (the keys loop and the values loop) where `i` iterates the *projection* columns, but `putBaseColumnName` resolves against the **base factory's** metadata (`BasePlanSink.getBaseColumnName`). When two aggregates reference the **same** base column (the canonical OHLC shape `first(price), last(price)`), the base metadata has fewer columns than the projection, so the trailing index either names the wrong column or runs off the end.

Verified by direct reproduction (`printSql` renders the plan; `execute("explain ...")` alone does not trigger `toPlan`):

- `select sym, first(i), last(i) from a ... sample by 1h align to first observation` → plan prints `values: [first(i), last(ts)]` — `last(i)` is mislabeled `last(ts)`.
- `select sym, first(ts), last(ts) from a ... sample by 1h align to first observation` → `java.lang.AssertionError: index out of bounds, 2 >= 2` (EXPLAIN crashes; QuestDB runs with `-ea`).

This is **not introduced by this PR** (`toPlan` is untouched by the diff), but it is the same root-cause family (projection index vs. base index) in the same file. The existing `ExplainPlanTest.testSampleByFirstLast` only uses distinct columns, so it never exercised the duplicate-column path.

Fix — route both calls through the existing mapping:

```java
// both the keys loop and the values loop
sink.putBaseColumnName(queryToFrameColumnMapping[i]);
```

`queryToFrameColumnMapping[i]` holds the base-factory-metadata index (built from the same `baseMetadata` that `putBaseColumnName` resolves against), so it is correct for both loops. Add an `ExplainPlanTest` case with two aggregates over one column (e.g. `first(i), last(i)` and `first(ts), last(ts)`) to lock it in.

### Minor

- The new test omits an OHLC-shaped case (two aggregates over the same column with the timestamp not projected). The current fix is fully covered without it, but if the `toPlan` fix above is bundled, a `first(val), last(val)` variant would guard both at once.
- The single `assertMemoryLeak` wraps two `assertQuery(...).returns(...)` calls, so the builder's own internal leak check runs nested inside the outer one. Harmless and justified here (DDL + INSERT + two assertions share the scope) — just noting the redundancy.

### Summary

**Approve** the fix as-is — minimal, correct, empirically verified to fix the reported bug, strictly improving, and well-tested. No blocking issues, no regressions.

- 1 verified out-of-diff bug (pre-existing `toPlan` defect, same file), 2 minor notes.
- The diff itself is clean: 0 in-diff issues.
- Recommendation: merge the timestamp fix; consider adding the two-line `toPlan` correction + an OHLC EXPLAIN test to this same branch, since it shares the file, the root cause, and the CI run.

<sub>🤖 Reviewed with Claude Code</sub>

ideoma:
## Review (level 2) — independent pass

**Verdict: Approve.** The one-line fix is correct, minimal, and empirically confirmed. There is one pre-existing, out-of-diff bug in the same file worth bundling, but it does not block this PR.

### The fix is correct

The change is `col != timestampIndex` -> `col != groupByTimestampIndex` in `SampleByDataRecord.getLong()` (`SampleByFirstLastRecordCursorFactory.java:983`). I verified the index-space mismatch from the source rather than the description:

- **`col` is projection-space** — it indexes `pageAddresses[col]` / `firstLastIndexByCol[col]`, both sized to `columns.size()`.
- **`groupByTimestampIndex` is projection-space** — `buildFirstLastIndex` sets it to the projection loop index `i` (line 274), and leaves it `-1` when the timestamp is not projected.
- **`timestampIndex` is base/frame-space** — also used at `metadata.getColumnType(timestampIndex)` (base metadata) and `frameMemory.getPageAddress(timestampIndex)` (frame space). The old comparison mixed the two spaces; they coincide only when the timestamp is projected at the same ordinal it occupies in the base scan.

When the timestamp is not projected, `groupByTimestampIndex == -1`, so `col != -1` is always true and nothing is special-cased — correct, since the timestamp isn't in the result. The new behavior can never turn a correct result wrong, because `col` and `groupByTimestampIndex` are same-space by construction.

**Completeness:** only `getLong`/`getTimestamp` carry the special case (`getTimestamp` delegates to `getLong`), which is right — the designated timestamp is LONG-backed and never read via `getInt`/`getShort`/`getSymA`/etc. The cross-frame path (`saveFirstLastValuesToCrossFrameRowBuffer`) already used `groupByTimestampIndex`, which is why the boundary buckets were always correct. The remaining `timestampIndex` reads in the file are all legitimately base/frame-space; no sibling instances of the same confusion were left behind.

**Empirical confirmation:** the new regression test passes on this branch (`duration_ms=284`, `fd reuse count=0, mmap reuse count=0`, no leak). I also hand-checked every expected value against the generator (`val = x*3`, 10-min spacing, `sym='s1'` => x in {1,6,11,...,56}); all ten buckets are right, and the single-row middle buckets are exactly the ones that returned the bucket-start timestamp before the fix.

### Test quality — clean

`testSampleByFirstLastReturnsValuesWhenTimestampNotProjected` uses `.returns(...)` (not `returnsOnce`), `.noRandomAccess()` matches the factory (`recordCursorSupportsRandomAccess()` returns `false`), and the single `assertMemoryLeak` wrapper is legitimate (DDL + INSERT + two assertions share one scope). It guards both the timestamp-projected and timestamp-omitted forms.

### Moderate (out-of-diff, pre-existing — reasonable to bundle, not blocking)

**`toPlan()` indexes base metadata with a projection-space index -> mislabeled EXPLAIN names and an `AssertionError` crash for duplicate-column aggregates.**

`toPlan()` (lines 189 and 204) calls `sink.putBaseColumnName(i)` where `i` iterates the projection columns, but during `toPlan` the factory is on top of `factoryStack`, so `putBaseColumnName(i)` resolves via `getBaseFactory().getMetadata().getColumnName(i)` — the base factory metadata (same `baseMetadata` from `SqlCodeGenerator.java:7830`). The spaces diverge whenever two aggregates reference the same base column (the OHLC `first(price), last(price)` shape) or the projection reorders columns.

Confirmed by trace:
- `select sym, first(i), last(i) ... sample by 1h align to first observation` -> base metadata `[sym, i, ts]`, so projection index 2 (`last(i)`) resolves to base col 2 = `ts` => plan prints `last(ts)`.
- `select sym, first(ts), last(ts) ...` -> base metadata `[sym, ts]` (2 cols), projection index 2 => `AssertionError: index out of bounds, 2 >= 2` (and QuestDB runs with `-ea`).

The existing `ExplainPlanTest.testSampleByFirstLast` only uses distinct columns, so base order coincidentally matches projection order and the path is never exercised.

This is **not introduced by this PR** (`toPlan` is untouched), so it does not block the merge — but it is the same root-cause family (projection index vs. base index) in the same file with the same CI lineage, so it's a sensible add to this branch. Fix:

```java
sink.putBaseColumnName(queryToFrameColumnMapping[i]);   // both the keys and values loops
```

`queryToFrameColumnMapping[i]` is built from the same `baseMetadata`, so it is the correct index. Add an `ExplainPlanTest` case with two aggregates over one column (`first(i), last(i)` and `first(ts), last(ts)`) to lock it in.

### Minor

- The regression test omits an OHLC-shaped case (two aggregates over the same column with the timestamp not projected, e.g. `first(val), last(val)`). The current fix is covered without it, but such a case would guard both the data-path fix and, if bundled, the `toPlan` fix at once.

### Summary

- Approve. The fix is correct, minimal, strictly an improvement, and verified by index-space trace plus a green regression test.
- 3 draft findings raised, 3 verified, 0 false positives. In-diff vs out-of-diff: 0 in-diff issues, 1 out-of-diff (the pre-existing `toPlan` bug) + 1 minor test-coverage note.
- Recommendation: merge the timestamp fix as-is; optionally fold in the two-line `toPlan` correction plus an OHLC EXPLAIN test on this same branch, since it shares the file, the root cause, and the CI run.

<sub>Reviewed with Claude Code</sub>

jovfer (approved):

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFirstLastRecordCursorFactory.java|1|1|100.00%|

PR #7283, opened 2026-06-18
chore(core): improve Parquet interoperability with external readers and Iceberg tooling

RaphDal:
This branch bundles fourteen Parquet fixes. Fixes 1–13 concern how QuestDB's Parquet output interoperates with external readers (pyarrow, Spark, DuckDB, Trino) and Iceberg tooling; Fix 14 is a separate, pre-existing correctness bug in QuestDB's own `_pm` metadata sidecar (bloom-filter row-group pruning) that the interoperability work does not touch. All center on `core/rust/qdbr/`, and several touch the vendored `parquet2` / `parquet-format-safe` forks. Fixes 1–3, 8, and 9 are independent; Fixes 4 and 5 are the statistics-truncation pair (truncate, then mark the bound inexact) and Fixes 6 and 7 are the page-index pair, each landing together. Fixes 10 through 13 close four min/max statistics gaps — dict-encoded IPv4 (Fixes 1 and 9 together), opaque-`Binary` all-`0xFF` max bounds (Fix 1 alone), the multi-page chunk-level statistics reduce for unsigned IPv4 columns, and the chunk reduce plus the page index for the opaque-`Binary` unbounded-max sentinel (Fix 1 again) — that the `column_orders` declaration turned from latent into reader-visible. Every change is write-time only — files already on disk are untouched until the next conversion, export, or O3 merge rewrites them. That rewrite re-encodes and recomputes statistics for changed groups, but the O3 merge also copies unchanged source row groups verbatim (a zero-decode raw copy that reuses their chunk-level statistics), so rewriting a legacy file written before Fixes 10–13 exposes such a copied group's pre-fix min/max as reader-authoritative under the newly-declared `column_orders` without recomputing it. This is a pre-existing data condition rather than a new hazard: QuestDB's own `_pm` pruning already reads those same legacy bounds, recomputing stats for a copied group would defeat the zero-decode copy, and reachability is narrow — IPv4-dict needs a non-default `RLE_DICTIONARY` export override, and the opaque-`Binary` all-`0xFF` case requires a row-group max of eight `0xFF` bytes exceeding nine bytes.

## Fix 1: Declare Parquet column orders so external readers use min/max statistics

QuestDB writes per-row-group `min_value`/`max_value` but omitted the file-level `column_orders` declaration. The spec leaves min/max **undefined** without it, so conformant readers report `has_min_max=False` and skip predicate pushdown, while PyIceberg `add_files` drops the file into a null partition. The fix emits one `TypeDefinedOrder` per leaf column (a derivable constant) from all three live footer builders plus `into_thrift`, which also retrofits the declaration onto pre-fix files when they are next merged. The async `FileStreamer` path got the same one-line change for consistency but is not compiled in this build, so it is verified by inspection only.

## Fix 2: Name nested-array list elements canonically for Iceberg tooling

Exporting a nested array (`DOUBLE[][]`) named the outer LIST element node `list` instead of `element`, yielding `bids.list.list.list.element` instead of the canonical `bids.list.element.list.element`. The file is spec-compliant and reads correctly everywhere, but PyIceberg `add_files` does a literal path string lookup and raises `KeyError`. The one-line fix renames the intermediate LIST-annotated group to `element` for every `dim >= 2`. Pure field-name change: no data, type, or repetition-level change, and 1D arrays were already canonical.

## Fix 3: Carry the column id in QuestDB metadata and prefer it over the Parquet field_id

QuestDB stamps each Parquet `field_id` with its internal writer index, which is load-bearing for QuestDB's own column mapping but collides with an Iceberg table's field IDs, mis-mapping columns in external tools. This relocates the authoritative id into the QuestDB-private metadata (`QdbMetaCol` gains an optional `id`) and reads it in preference to `field_id` via a shared `resolve_column_id` helper. Purely additive and bidirectionally compatible — the `field_id` is still written, so nothing on the wire changes yet. It is the prerequisite that lets a separate enterprise change strip `field_id` for Iceberg without breaking QuestDB's reads.

## Fix 4: Bound String/Symbol/Varchar min/max statistics to a fixed size

QuestDB wrote text-column min/max at full length, so a single multi-megabyte string bloated the footer and the `_pm` sidecar. The fix bounds String/Symbol/Varchar min/max to 64 bytes on a UTF-8 codepoint boundary: the min truncates down to a prefix (valid floor), the max truncates and advances to the next Unicode scalar (valid ceiling), falling back to the full value when it cannot. Bounds stay conservative but become looser for long strings; opaque `Binary` keeps its <=9-byte min prefix and rounded-up 8-byte max ceiling, and fixed-length types keep their exact bounds.

## Fix 5: Mark truncated min/max statistics as inexact

Building on Fix 4, QuestDB now sets the upstream Parquet `Statistics.is_min_value_exact` / `is_max_value_exact` thrift fields so a reader can tell a truncated bound from an exact one. The vendored `parquet-format-safe` `Statistics` struct gains optional bool fields 7 and 8, and the `parquet2` binary-stat path carries them through serialization and row-group/page reduction. `BinaryMaxMinStats::into_parquet_stats` writes `false` for any bound that Fix 4 widened — both a UTF-8 text bound it truncated and an opaque-`Binary` bound it clamped (min) or rounded up (max); exact short text and fixed-length byte-array stats leave the fields absent, relying on Parquet's default exact semantics. Readers that ignore the fields still get valid superset bounds.

## Fix 6: Declare the sorted timestamp's page-index boundary order

The per-column page index hardcoded `boundary_order = UNORDERED` for every column, including the designated timestamp the partition is physically sorted by, so readers forfeited binary search over its monotonic page bounds. The fix derives `boundary_order` from the file's `sorting_columns` (`ASCENDING`/`DESCENDING` for a listed column, `UNORDERED` otherwise) and threads it through `serialize_column_index`. The order is taken from QuestDB's own `SortingColumn` declaration rather than recomputed from page values, which would mis-order types like unsigned IPv4. `UNORDERED` was spec-legal, so this is a missed optimization, not a correctness fix.

## Fix 7: Preserve the page index when rewriting or appending partitions

The primary writer emits a page index (`ColumnIndex` + `OffsetIndex`), but the O3 merge-rewrite and append paths did not, so a partition lost its index on first O3 merge and appended files ended up partially indexed. A partial index is worse than none: arrow-rs `with_page_index(true)` fails the whole read with `missing offset index`. A single `write_page_index` helper now serves all three paths and composes the page index with Fix 6's boundary order; copied row groups carry their source index rebased by the data `offset_delta`. An all-or-nothing gate keeps every file uniformly indexed or fully unindexed — a legacy source row group lacking an `OffsetIndex` makes the whole output stay unindexed rather than mixed. The append/update path also keeps the source file's statistics setting rather than the current runtime config, so a `statistics.enabled` flip between the original write and a later append cannot leave the footer with statistics on some row groups but not others. It reads that setting from whether the source carries chunk-level statistics, not from its `ColumnIndex` presence: an `ALTER TABLE ADD COLUMN` partition rewrite raw-copies the data columns with their statistics intact but emits no page index, so a stats-bearing source can legitimately have no `ColumnIndex`, and tying the two together would drop footer `min_value`/`max_value` on every group appended after such a rewrite, forfeiting row-group pruning for QuestDB's own `_pm` and external readers alike. `ColumnIndex` emission is gated separately, on whether the source is itself uniformly `ColumnIndexed`. The append/update path re-emits the unchanged source row groups verbatim from a footer cache, so they keep their own `ColumnIndex`; the gate therefore runs in the other direction too. When a newly appended group cannot carry a `ColumnIndex` at all — an opaque-Binary page with an unbounded (max-less) max, per Fix 13 — the cached source groups have their `ColumnIndex` stripped (their `OffsetIndex` retained), keeping the footer uniformly `OffsetIndex`-only instead of advertising a `ColumnIndex` on the cached groups but not the appended one. The cost is that the previously indexed cached groups forfeit `ColumnIndex` pruning, the same conservative direction as Fixes 11 and 13.

## Fix 8: Populate Parquet page encoding statistics

QuestDB left `ColumnMetaData.encoding_stats` unset even though the page headers carried enough to derive it (audit Finding 6). `build_column_chunk()` now accumulates one `PageEncodingStats` per page-body `(page_type, encoding)` pair — dictionary pages as `DICTIONARY_PAGE`, V1 as `DATA_PAGE`, V2 as `DATA_PAGE_V2` — and the updater's synthetic null chunks set matching stats. Level-encoding `RLE` is excluded since `PageEncodingStats` describes the page body, not levels; empty chunks stay `None`. Readers that ignore the field are unaffected.

## Fix 9: Add the legacy UINT_32 converted type for IPv4

QuestDB writes IPv4 as `INT32` with the modern unsigned logical type but left the deprecated `ConvertedType` unset, so legacy readers that consult only `ConvertedType` saw a bare signed `INT32` (audit Finding 7). The fix writes `PrimitiveConvertedType::Uint32` alongside the existing `Integer(UInt32)` logical type. The physical type, data encoding, statistics, and QuestDB's own reads are unchanged; modern readers see no difference.

## Fix 10: Compute dict-encoded IPv4 min/max statistics with unsigned ordering

The Plain and Delta IPv4 encoders compare min/max as unsigned `u32`, but the `RLE_DICTIONARY` path compared them as signed `i32`. On its own this was latent — QuestDB's own `_pm` pruning compares IPv4 unsigned — but Fix 1 now declares `column_orders`, so a conformant external reader trusts those bounds and compares them unsigned too. For an IPv4 column written with the non-default `RLE_DICTIONARY` encoding and values straddling the i32 sign bit (any address `>= 128.0.0.0`), the signed bounds were byte-wrong under unsigned comparison, and a reader could prune the row group holding the matches and silently drop rows from a query like `WHERE ip > '200.0.0.0'`. The fix threads an `UNSIGNED_STATS` flag through the dictionary encoder so the IPv4 stats use unsigned ordering, matching the Plain/Delta paths; signed types (Geo*, decimals, the designated timestamp) keep signed ordering. The default IPv4 encoding is Plain, which was already correct, so only the explicit dictionary-encoding override was affected.

## Fix 11: Drop the opaque-Binary max statistic when its prefix cannot be bounded

QuestDB clamps an opaque `Binary`/BLOB column's min/max statistic to a <=9-byte prefix, then rounds the max up to a valid 8-byte ceiling by incrementing its big-endian prefix (`binary_upper_bound`, unchanged since #4460). When that prefix is eight `0xFF` bytes the increment overflows, and the code fell back to emitting the clamped 9-byte prefix as the max. For a real column value longer than nine bytes that shares the all-`0xFF` prefix — a true max of eight `0xFF` bytes then `0xAA 0xBB`, stored as eight `0xFF` bytes then `0xAA` — that stored prefix is byte-wise *less than* the real value, an invalid upper bound. On its own this was latent, but Fix 1 now declares `column_orders`, so a conformant reader trusts the bound and could prune the row group for a predicate like `WHERE blob > ...`, silently dropping the row holding the real maximum. It is the same latent-stat-becomes-authoritative mechanism Fix 10 addresses for IPv4, left open for opaque Binary; `is_max_value_exact = false` does not rescue it, because the spec still requires an inexact `max_value` to be a valid bound. The fix has `binary_upper_bound` return `None` in the all-`0xFF` case, so the writer omits `max_value` entirely and clears `is_max_value_exact` with it: a missing max reads as unbounded, so readers never prune on it. The tradeoff is that QuestDB forfeits max-based row-group pruning for that column in that one case — correct but looser, the same direction as Fix 4. The trigger is narrow (a row-group maximum that begins with eight `0xFF` bytes and exceeds nine bytes), the only affected path is opaque `BINARY`/BLOB with statistics enabled, and the flawed fallback predates this PR — Fix 1 is what makes it reader-visible.

## Fix 12: Reduce multi-page IPv4 chunk statistics with unsigned ordering

Fix 10 made the per-page IPv4 statistics unsigned across all three encoders, but a column chunk spanning more than one data page also carries chunk-level (row-group) `min_value`/`max_value` in the footer, which `parquet2` derives by reducing the per-page bounds. That reduction (`reduce_primitive`) compared `INT32`/`INT64` values with signed ordering regardless of the column's logical type, so a multi-page IPv4 chunk lost the unsigned ordering its individual pages already had. For an address straddling the i32 sign bit (any `>= 128.0.0.0`, stored as a negative i32), the chunk max was understated — a chunk holding `200.0.0.0` on one page and `100.0.0.0` on another reduced to a chunk max of `100.0.0.0`. Fix 1's `column_orders` makes a conformant reader trust and unsigned-compare that bound, so it could prune the row group and silently drop rows from a query like `WHERE ip > '199.0.0.0'` — the same latent-becomes-authoritative mechanism as Fix 10, left open at the chunk-reduce step. The per-page `ColumnIndex` stayed unsigned-correct, so the file was internally inconsistent: page-level pruning right, row-group-level pruning wrong. The fix reduces `INT32`/`INT64` chunk statistics with unsigned ordering when the column's logical or converted type marks it unsigned (IPv4's `UInt32`), matching the per-page statistics; the signed-versus-unsigned choice is made once per column and threaded as a const generic, mirroring the per-page `StatsUpdater<T, UNSIGNED_STATS>` split. Signed types and floating-point reductions are unchanged. Among QuestDB's types only IPv4 (`UInt32`) actually diverges — `CHAR` is `UInt16` but its values are always non-negative as an i32, and there is no unsigned 64-bit column — and the chunk must span more than one data page, reachable with a large partition or a reduced `data_page_size`. A short IPv4 column at the smallest default row group writes a single page, which was already correct.

## Fix 13: Carry the opaque-Binary unbounded-max sentinel through the chunk reduce and page index

Fix 11 drops an opaque `Binary`/BLOB page's `max_value` when its all-`0xFF` prefix has no short upper bound, leaving the min present and the max absent to mean "unbounded — readers must not prune." Two downstream structures ignored that sentinel. The per-page `ColumnIndex` requires a max for every non-null page, so the max-less page made the writer reject the whole file with `OutOfSpec("max value of a page is required")` — an opaque-Binary value at least nine bytes long that begins with eight `0xFF` bytes, with statistics enabled, could not be written at all, even on a single page. The writer now omits the `ColumnIndex` for the file (keeping the `OffsetIndex`) when any page carries an unbounded max, the same all-or-nothing degradation Fix 7 applies when a source lacks a `ColumnIndex`. Separately, a column chunk spanning more than one data page carries chunk-level `max_value` in the footer, which `parquet2` derives by reducing the per-page bounds; `reduce_binary` treated an absent max as "no contribution," so an unbounded page reduced against a smaller bounded max on another data page yielded that smaller bound — byte-wise less than the real all-`0xFF` value, the understated bound Fix 11 removes, re-introduced at the chunk level. With Fix 1's `column_orders` a conformant reader trusts it and prunes the row group, dropping the all-`0xFF` row from a query like `WHERE blob > ...`. The reduce now distinguishes a page that contributed a value but no max (a present min with an absent max) from an all-null page that contributed nothing (both absent), and drops the chunk max when any contributing page omitted its own. The write failure triggers for any such column with statistics on; the understated chunk max additionally requires the column to span more than one data page. The UTF-8 text and fixed-length byte-array paths are unaffected, since their per-page max is absent only for an all-null page, where dropping it is correct. The tradeoff is looser pruning: the file forfeits page-level pruning across all its columns, on top of the row-group max pruning Fix 11 already forfeits for the one column — the same conservative direction as Fixes 4 and 11. The async `FileStreamer` path is not compiled in this build, so it is left unchanged, as with the other page-index fixes.

## Fix 14: Inline the surviving column's own bloom filter into the `_pm` sidecar after DROP COLUMN

QuestDB's `_pm` metadata sidecar inlines a copy of each column's bloom filter for row-group pruning. When an O3 insert rewrites an already-Parquet partition after a `DROP COLUMN`, `copy_row_group_with_null_columns` extracted the row group's bloom bitsets in the source (pre-drop) column order but rebuilt the column chunks in the target (post-drop) order, then handed both to the `_pm` writer at the same row-group index. The writer indexes the bitset vector positionally, so each surviving column received the bloom filter of the column that used to occupy its new slot: dropping a leading bloomed column from `[a, b, c, ts]` gives `b` column `a`'s bloom, `c` gets `b`'s, and `c`'s bloom spills onto the bloom-less timestamp. The read-side row-group skipper then probed an equality/`IN` filter against the wrong column's bloom, could conclude a present value was absent, and silently skipped a row group that held matching rows — fewer rows returned than exist, with no error or log line. The Parquet data file stayed correct, since its inline blooms are rebased in place, so external tools reading the file directly saw nothing wrong; only QuestDB's own `_pm`-driven pruning was poisoned.

The fix builds the bitset vector in target order, in lockstep with the column chunks, reusing the same field-id → target-position map already used to remap the chunks: each survivor's bitset moves to its new slot and a dropped column's bitset is discarded, so the writer's positional indexing inlines each column's own bloom. The data-file blooms need no change. This is a pre-existing bug whose write-side producer and read-side consumer have coexisted since the `_pm` sidecar was introduced, and it is independent of the `column_orders` work in Fixes 1–13 — it affects QuestDB's own reads regardless of any external reader. Reachability is narrow: bloom filters are opt-in per column (`… PARQUET(… BLOOM_FILTER)` or `COPY … bloom_filter_columns`), a `DROP COLUMN` must shift a bloomed column's position, an O3 insert must rewrite the already-Parquet partition, and a read must apply an equality/`IN` filter on the shifted column. Default tables carry no bloom filters and are unaffected. A regression test rewrites a four-column partition with blooms on three columns, drops the leading one, and asserts the regenerated `_pm` inlines each survivor's own bloom at its new position and attaches none to the timestamp.

## Bundled dependency updates

This branch also bumps `criterion` (0.4 -> 0.8.2) and refreshes the lockfile. The motivating reason is tooling: the older criterion pinned transitive dependencies that prevented `cargo nextest` from running the crate's tests, and the bump unblocks it. As a side effect the bump also drops the unmaintained `atty` crate (RUSTSEC-2021-0145), so `cargo deny check advisories` stays green for the affected lockfiles. The `THIRD_PARTY_LICENSES.txt` timestamp churn is an automated cargo-deny regeneration artifact. The `qdb-sqllogictest` lockfile is deliberately left at its `master` state: an earlier round of incidental `cargo update` drift there (unrelated transitive crates, no advisory or `parquet` relevance) was reverted to keep the diff scoped to the Parquet work.

It also bumps `memmap2` to 0.9.11 across the `qdb-core`, `qdbr`, and `qdb-parquet-meta` crates to clear [RUSTSEC-2026-0186](https://rustsec.org/advisories/RUSTSEC-2026-0186). The previously resolved versions (0.9.9 / 0.9.10) were vulnerable; after the bump `cargo deny check advisories` reports `advisories ok`. The version requirements in the two declaring manifests move to `0.9.11`; no API surface changed, so the crates compile unmodified.

## Test plan

- `cargo test --manifest-path core/rust/qdbr/Cargo.toml --lib` — 1084 pass, 3 ignored
- `cargo test --manifest-path core/rust/qdbr/Cargo.toml --test encode_statistics` — 12 pass
- `cargo test --manifest-path core/rust/qdbr/Cargo.toml --test decode_pm_e2e` — 16 pass
- `PARQUET2_IGNORE_PYARROW_TESTS=1 cargo test --manifest-path core/rust/parquet2/Cargo.toml --lib` — 185 pass
- `PARQUET2_IGNORE_PYARROW_TESTS=1 cargo test --manifest-path core/rust/parquet2/Cargo.toml --test it` — 82 pass
- `cargo fmt`, `cargo check --all-targets`, `cargo clippy --all-targets`, `cargo test --lib` from `core/rust/qdbr` — clean with zero warnings

## PR type

Fixes 10–13 prevent conformant external readers from silently dropping rows once Fix 1 declares `column_orders`, so there is a genuine correctness dimension. The branch nonetheless ships as `chore(core):` rather than `fix(core):`: the net user-facing change is interoperability, no file already on disk is corrupted, and the affected bounds stay latent until a future rewrite republishes them under `column_orders`. These are internally-discovered audit findings rather than customer-reported defects, consistent with reserving the `fix` conventional-commit type for customer-reported bugs. The tradeoff is that the corrections land outside the release-notes 'Fixes' section; the `Bug` label still flags the correctness dimension for triage.

Fix 14 is the exception to that latency. It is a live defect in QuestDB's own `_pm` pruning that already returns fewer rows than exist on the narrow DROP-COLUMN-then-O3-rewrite path, and a `_pm` written on that path persists the wrong bloom on disk until the partition is rewritten again — not a bound that stays dormant until `column_orders` republishes it. It is still an internally-discovered bug rather than a customer report, so it too lands as `chore(core):` under the same convention; the `Bug` label carries the correctness signal. If the team would rather surface this in the release-notes 'Fixes' section given that it affects live queries, `fix(core):` is the defensible alternative.

## Summary by CodeRabbit

* **New Features**
  * Parquet output now includes richer metadata for column ordering and page indexes, improving compatibility with external readers and pruning.
  * String statistics handling is more precise for UTF-8 data, and IPv4 values are now ordered correctly using unsigned comparisons.

* **Bug Fixes**
  * Fixed several Parquet export/update cases where min/max statistics, page indexes, and nested array schema paths could be inconsistent.
  * Improved handling of binary maxima so unbounded values are no longer reported incorrectly.

* **Tests**
  * Updated and expanded coverage for Parquet export, metadata, statistics, and decoding behavior.


RaphDal:
# Code Review — PR #7283

`chore(core): improve Parquet interoperability with external readers and Iceberg tooling`

Branch `rd/fix-parquet-stats` · +4067/−558 · 10 bundled Parquet-interoperability fixes, almost entirely in `core/rust/`.

Reviewed at level 2: nine parallel agents (correctness, concurrency, performance ×2, resource management, test coverage, code quality, Rust safety, PR metadata) plus inline cross-context verification of every changed `pub` symbol's callsites and the JNI/Java boundary.

## Verdict

**Approve.** This is a carefully built PR. The page-index rebasing, the all-or-nothing index gating, the UTF-8 codepoint-boundary truncation, and the unsigned-IPv4 dictionary statistics are all correct on close inspection and backed by strong end-to-end tests that read back through the independent Arrow reader (including `with_page_index(true)`, which structurally rejects mixed-index files).

- **No Critical findings** — no reachable JVM-aborting panic, no resource leak, no data-corruption path, no concurrency defect, no algorithmic regression on the per-row hot path.
- The one moderate code finding — `ParquetFile::end` Mode::Update lacked the `all_indexable` guard that Mode::Write has — **has been fixed in this branch** (`core/rust/parquet2/src/write/file.rs`). The Mode::Update branch now also requires every new group to be individually indexable before writing the page index, the same all-or-nothing rule Mode::Write applies, so the file can never mix indexed and unindexed row groups. Reachable behavior is unchanged (validated: `cargo fmt`/`check`/`clippy` clean, parquet2 index tests and all 25 qdbr append/rewrite tests pass).

The remaining items below are style, comments, and test hardening. None block merge.

## Minor

- **Em-dash in a newly-added test comment — `core/rust/qdbr/src/parquet_write/update.rs:3050`.** `// Bloom filter on col1 (index 1) — the LAST column.` is the only non-ASCII byte the PR adds; every other new doc uses ASCII `--`. The strict-ASCII rule is formally scoped to log messages, so this is cosmetic, but it breaks the author's own consistency. Replace `—` with `--`.

- **`next_utf8_string` allocates a `Vec<char>` plus a `String` — `core/rust/qdbr/src/parquet_write/util.rs:300-311`.** Warm path: runs at most twice per column-chunk and only when a text min/max exceeds 64 bytes — not per row. An in-place `&[u8]`-tail approach would avoid both allocations (and the O(k²) rebuild on the degenerate all-U+10FFFF carry path). Optional cleanup, not a throughput concern.

- **`increment_encoding_stats` lacks the rationale comment its sibling has — `core/rust/parquet2/src/write/column_chunk.rs:307`.** The linear `.find()` is the correct choice (≤3 distinct page_type/encoding pairs — a `HashMap` would be slower), exactly as `push_unique` argues a few lines up. Add the matching one-line justification so a future reader does not "optimize" it into a map.

- **Document the cross-file Update-path coupling.** `update.rs:269` derives `write_statistics` from the source's **ColumnIndex** presence, while `file.rs:1027` gates on the source's **OffsetIndex** presence — two different fields, in two files, that must stay jointly consistent for the Update path to produce a uniform footer. The behavior is correct today (and `write_page_index`'s internal `emit_column_index` re-check is the real safety net), but a one-line comment tying the two together would prevent a future desync. The same applies to the direct index `m.schema[i]` at `convert.rs:135`, which is safe only because of the length guard at `convert.rs:99-106`; a comment linking them would prevent a live panic if a future multi-leaf column type were ever added.

- **`all_indexable` indexes parallel vecs by `i` — `core/rust/parquet2/src/write/file.rs:983`** (and the new sibling guard in the Mode::Update branch, written to match it). `.iter().enumerate().all(|(i,_)| !page_specs[i].is_empty() || copied_page_index[i].is_some())` ignores the iterated item; a `.zip(...).zip(...)` reads cleaner. Trivial; kept index-by-`i` in the new guard intentionally for consistency with the existing one.

## Test hardening (Low — none block merge)

- **The async `FileStreamer` `column_orders` change ships with no test assertion** (`core/rust/parquet2/src/write/stream.rs:195`). The PR's "verified by inspection only" is accurate; the existing async test asserts only data + per-column stats. It is not compiled into the OSS build (no `async` feature on the `parquet2` dependency), so this is benign, but it is the one production change with zero coverage.
- **`reduce_exact_flag`'s `_ => None` arm** (both inputs exact → stays exact across a reduce) is only implicitly exercised. A one-line addition to the `binary()` test in `parquet2/src/write/statistics.rs` (two pages, both flags `None`, assert result `None`) closes it.
- **`build_copied_page_index`'s offset-rebase arithmetic and its corrupt-input guards** (negative offset/length, overflow) are covered only end-to-end, or not at all. The guards are correctly `Result`-returning rather than `unwrap` — the right choice for JNI-reachable corrupt-file input — just uncovered.
- **The incremental `_pm` append branch / `sync_parquet_meta`** is untested; every `_pm` test uses the full-create branch (`existing_parquet_meta_file_size = -1`).

## Notes for the team (not findings)

- `_pm` deliberately does **not** round-trip the new `is_min/max_value_exact` flags (`convert.rs::apply_thrift_stats` reads only `min_value`/`max_value`). This is correct: `_pm` pruning needs only valid superset bounds, and the truncated min/max are valid floors/ceilings; the exact flags are footer-only hints for external readers. Confirm this matches intent.
- Tradeoffs are honestly disclosed in the PR body (looser stats bounds for long strings, larger footers, the all-or-nothing index degradation on legacy sources, the inspection-only async path).

## Cross-context coverage

Out-of-diff breakage is zero, and this is not an underrun. Verified inline: every changed `pub` symbol's callsites were grepped (`Statistics::new`, `serialize`/`write_column_index`, `write_raw_row_group_*`, `resolve_column_id`, `into_thrift`, `type_defined_column_orders`); the changed symbols live in vendored crates (`parquet2`, `parquet-format-safe`) consumed only by `qdbr` (a closed world); Rust's compiler enforces signature consistency (rebuilt `.so` + clean `cargo check`); and the two semantic-contract changes (`into_thrift` now always emitting `column_orders`; `boundary_order` derived from `sorting_columns`) were traced to all callers and the JNI/Java boundary (`O3PartitionJob`, `PartitionUpdater`).

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=245134)
**Tested Commit:** `05ac88e4a4770da80a9a3af4fdb663fff6feca02`

bluestreak01:
@RaphDal — two **moderate** items from a deeper review pass. Neither is a code-correctness blocker; both are worth a decision before merge.

### 1. Verbatim-copied legacy row groups + newly-declared `column_orders`

`copy_row_group` reuses a source row group's chunk-level `Statistics` verbatim — `into_thrift()` keeps `meta_data.statistics`, and `adjust_column_chunk_offsets` only rewrites offsets, not stats. Fix 1 now stamps `column_orders = TypeDefinedOrder` on the rewritten footer. So when an O3 merge rewrites a **legacy** file (written before Fix 10/11) and copies an *unchanged* group, that group's old, possibly-invalid min/max becomes reader-authoritative for the first time.

- Fixes 10/11 only correct the **fresh-encode** path; they can't recompute stats for verbatim-copied groups.
- Reachability is narrow:
  - IPv4-dict (Fix 10) is effectively **unreachable** for storage partitions — `encoding_map` returns `Plain` for IPv4; `RLE_DICTIONARY` needs an explicit export override. Legacy partition files have correct unsigned Plain bounds.
  - The opaque-Binary all-`0xFF` case (Fix 11) **is** reachable for normal `BINARY`/BLOB columns, but only when a row-group max begins with eight `0xFF` bytes and exceeds 9 bytes — astronomically rare, and external range-pruning on BLOB is unusual.
- The PR body says files "are untouched until the next... O3 merge rewrites them," which reads as "the rewrite applies the fix." For copied-unchanged groups it instead *exposes* the old bound.

Suggestion: one sentence in the PR body acknowledging this and confirming it's acceptable (it largely mirrors the `_pm`-flags note — a pre-existing data condition that QuestDB's own `_pm` pruning is already subject to). No code change needed unless you want the copy path to recompute/clear stats for copied groups, which would defeat the zero-decode raw copy.

### 2. PR type `chore` undersells reachable correctness fixes

Fixes 10 and 11 prevent conformant external readers from **silently dropping rows** (the PR body's own framing), and Fix 1 is what makes them reader-visible. Shipping under `chore(core):` keeps these out of the release-notes "Fixes" section, and the PR currently lacks the `Bug` label. Consider `fix(core):` + add `Bug`. Judgment call — the net user-facing change is interoperability, so `chore` is defensible — but given the silent-data-loss framing, `fix` is the better fit.

---

Everything else verified clean: append-mode offset handling preserves the source page index, the all-or-nothing index gating is sound and well-tested, `convert.rs` direct indexing is guarded by the length check, Fix 9 doesn't alter IPv4 type inference, the async `stream.rs` path isn't compiled (parquet2 default excludes `async`), and out-of-diff caller breakage is zero.

One external item to confirm separately: the enterprise CI failure (build 245247) — likely just the enterprise equivalent of the byte-size test assertions this PR updated in `ExpParquetExportTest`/`CopyExportTest`.

bluestreak01 (changes requested):
see PR comment

RaphDal:
Thanks @bluestreak01
I've updated the PR body to acknowledge both:

1. Verbatim-copied legacy row groups + newly-declared column_orders
Marking this as acceptable. The fresh-encode path always writes correct bounds, so only legacy groups copied verbatim by an O3 rewrite carry old stats forward, and only in the narrow cases you noted (IPv4-dict needs a
non-default export override; the opaque-Binary all-0xFF case is astronomically rare). It's the same pre-existing condition QuestDB's own `_pm` pruning already lives with, and recomputing stats for a copied group would
defeat the zero-decode raw copy. Not aware of any external reader hitting it in practice.

2. PR type
Keeping it as chore. None of these were reported by an external user/customer. They're preemptive interoperability fixes for external readers (pyarrow/Spark/DuckDB/Trino), which hadn't been a focus area until now. I've
kept the Bug label so the correctness dimension still shows up for triage.

RaphDal:
# Code Review — PR #7283: `chore(core): improve Parquet interoperability with external readers and Iceberg tooling`

**Level 2** review (9 parallel agents + batched verification). **Verdict: approve.**

This is a large (13-fix, ~4.4k-line) on-disk-format PR, and it is unusually careful — defensive against corrupt input, heavily and meaningfully tested, with an honest, tradeoff-forward description. The four highest-risk hypotheses (round-trip breakage, reader intolerance, invalid sort-order bounds, offset-rebasing corruption) were all investigated against source and **disproved**. No Critical or Moderate findings; the rest are Minor.

## Critical

None.

## Minor

### m1 — Redundant per-chunk statistics serialization on the footer path (in-diff, perf)

`core/rust/parquet2/src/write/indexes/serialize.rs:73` — `pages_support_column_index` calls the full `serialize_statistics()` (a `Vec<u8>` clone for the 13 byte-array/decimal stat types) on every fresh data page purely to test `min.is_none() || max.is_some()`, then when the gate passes `serialize_column_index` (line 25) serializes the same pages *again*. `PageWriteSpec.statistics` is already the typed form, so the presence check needs no serialization. Scale is modest — ~1 redundant clone per column chunk at file finalize under default sizing (1 page/chunk), growing to per-page only with large row groups / small `data_page_size`. Cleanly removable: return `true` immediately for non-`ByteArray` `physical_type()`, and for `ByteArray` downcast to `&BinaryStatistics` and read the `Option` fields directly.

### m2 — `result_parquet_meta_size` set before the `_pm` write in the full-create branch (in-diff)

`update.rs:1030` sets the size *before* `write_all` (1031–1034); the incremental branch (1092) correctly sets it *after* the write. Unreachable on a correct caller (the Java driver aborts on the throw and never commits), and `commit_parquet_meta` only `debug_assert!`s the size — pure defense-in-depth/consistency. Move line 1030 after the write to match the incremental branch.

### m3 — New `assert!` const-folds to `assert!(true)` (in-diff)

`core/rust/parquet2/src/write/statistics.rs:639`: `assert!(IP_LOW > 0 && IP_HIGH < 0, ...)` — both operands are `const i32`, so clippy flags it as optimized-out and the `0xC800_0000_u32 as i32` cast already documents the intent. Delete it (or make the values non-const if a runtime guard is actually wanted).

### m4 — Verbose production comments exceed the "few and short" house style (in-diff)

The page-index all-or-nothing rationale is restated three times (`parquet2/src/write/file.rs` doc + inline blocks in the `Mode::Update` branch), and the "pyarrow/Spark/DuckDB/Trino/Iceberg" reader-list + truncation rationale are spelled out at paragraph length in `mod.rs`, `statistics.rs`, and `util.rs`. They're accurate but collectively are the long-block pattern the convention warns against; trim each to 1–2 lines and let the (excellent) PR body carry the deep rationale. (Test-comment verbosity weighted more leniently.)

### m5 — `schema[i]` indexing vs the safe `.get()` (in-diff, consistency)

`core/rust/qdbr/src/parquet_metadata/convert.rs:135` uses `qdb_meta.and_then(|m| m.schema[i].id)` (panics if `schema` is shorter than `columns`), whereas the parallel read path `meta.rs:215` uses `.and_then(|m| m.schema.get(index))`. No *new* panic surface (the adjacent pre-existing line 138 already indexes `[i]`), but the two new `resolve_column_id` callsites diverge in style; aligning convert.rs onto `.get(i)` is a cheap JVM-abort guard.

### m6 — PR body's `Bug`-label claim (metadata) — RESOLVED

The body's closing sentence says "the `Bug` label still flags the correctness dimension for triage." The `Bug` label has now been applied (labels: Bug, Enhancement, Compatibility, Core, rust, storage), so the body and the label set are consistent — Fixes 10–13 carry a genuine row-dropping correctness dimension that justifies it. Title format, tone (level-headed, tradeoffs given equal weight), the `chore` type choice, and the absence of `Fixes #NNN` are all correct for this PR.

### m7 — `encodePartition` hardcodes ASCENDING sort order (out-of-diff fragility)

`core/rust/qdbr/src/parquet_write/jni.rs:542-543` sets `SortingColumn::new(timestamp_index, false /*ascending*/, false)` with no direction parameter, unlike the streaming path (jni.rs:1048) and the update path (update.rs:240), which both derive direction. Correct *today* because this path serializes on-disk partition storage, which QuestDB always lays out ascending-by-designated-timestamp (the read side asserts it). Now that Fix 6 makes `boundary_order` consume this declaration, the correctness of the emitted ASCENDING bound rests on that storage invariant rather than the actual data — worth a comment so a future descending-storage change doesn't silently emit a wrong bound.

### m8 / m9 — Two small test gaps

- **m8:** No end-to-end test for the *single-page* opaque-Binary unbounded-max write — i.e. the exact original-bug scenario (Fix 13's `OutOfSpec` write failure on one page). The multi-page reduce is tested e2e and the gate is unit-tested, so behavior is sound; add a single-page `finish()`-succeeds test in `core/rust/qdbr/tests/encode_statistics.rs`.
- **m9 (informational):** No *new* round-trip test for the Fix 2 array rename; the untouched `testNdArray` in `ParquetTest.java` already guards it (the reader is name-agnostic), so this is optional.

## Downgraded (investigated, not real)

- **Concern A — Fix 2 rename breaks QuestDB's own nested-array reads:** FALSE. The reader matches LIST levels structurally (`max_rep_level`/`max_def_level`, `GroupConvertedType::List`), never by node name — zero `== "list"`/`"element"` comparisons in read logic. Triple-confirmed (manual grep + tests agent + correctness agent), and the untouched `testNdArray` round-trip exercises the renamed path in CI.
- **Concern B — readers/code intolerant of `column_orders` now being `Some`:** FALSE. No production code reads/asserts/counts `column_orders`; the one index getter `column_order(i)` is dead code; the parser uses panic-safe `.zip()`; leaf count is correct for nested arrays; both finalize paths regenerate it fresh (no doubling).
- **Concern C — ASCENDING/DESCENDING boundary_order is an invalid bound:** FALSE. The global timestamp sort is enforced Java-side (`TableWriter` sort), pages are sequential row-range slices, so per-page min/max are genuinely monotonic across pages and row groups; DESCENDING is emitted only for descending data. Verified end-to-end against the external Arrow reader.
- **Concern D — page-index offset rebasing corrupts the copied index:** FALSE. `offset_delta = new_offset − rg_start` is correct and applied uniformly (the row group is copied as one contiguous blob); `checked_add` guards overflow; negative offset/length are rejected before any cast/allocation; the all-or-nothing gate is airtight on both Write and Update paths.
- **Other non-findings (verified):** `binary_upper_bound`'s `[..8].unwrap()` (caller-guarded, unreachable), `write_page_index` zip "truncation" (lengths always equal), `reduce_int_primitive` downcast/`unwrap` (non-empty + uniform-type guaranteed by `reduce`), the `i32` index-length cast (needs ~16M pages/chunk), concurrency races (write path is single-threaded with disjoint per-column bloom locks), resource leaks (RAII + correct error-path fd handling; no partial footer ever committed). The const-generic `UNSIGNED_STATS` threading and `reduce_int_primitive` are deliberately monomorphized/inlined as intended.

## Summary

- **Verdict: approve.** No blocking issues. The one pre-merge item — the `Bug`-label inconsistency (m6) — has been resolved (the label is now applied); the remaining findings are genuinely minor / optional.
- **Tradeoffs (as the PR itself states):** Fixes 4/11/13 deliberately loosen bounds (truncated/omitted max → less external-reader pruning) in exchange for never emitting an invalid bound. The conservative all-or-nothing page-index degradation (a file goes fully unindexed rather than mixed) is part of the same direction and is consistent throughout.
- **Findings: 9 verified (all Minor), 0 false positives carried into the report.** All are in-diff except m7 (`encodePartition`, out-of-diff); m6 is a PR-metadata fix. Plus 4 out-of-diff *hypotheses* (Concerns A–D) investigated and cleared. The cross-context pass did not underrun — the out-of-diff hot spots (array reader, `column_orders` consumers, sort-order monotonicity, the Java fd lifecycle) were all reached and traced.
- **Note (out of scope, pre-existing — proven via the diff, not assumed):** `update_parquet_metadata` indexes `bloom_bitsets` positionally against the *reordered* row groups while it's stored in push order (convert.rs:657/671); the diff shows this loop unchanged, so it's not this PR's regression, but it's worth a separate look if O3 ever drives non-ascending replace/insert with bloom-enabled columns.

bluestreak01 (changes requested):
@RaphDal — adversarial level-3 review (full source verification + cross-context pass). Net: **approve with changes** — no blocking defect, but the two Moderate items below should be addressed (or consciously waived) before merge. Excellent, unusually thorough fixes and tests.

---

## Critical
None confirmed. (The append-path item below is real but benign in practice → Moderate.)

## Moderate

### M1 — Append/update can emit a footer with mixed `ColumnIndex` presence
`core/rust/parquet2/src/write/file.rs` — `ParquetFile::end` `Mode::Update` (~1027–1062) + `write_page_index` `emit_column_index` (~301–314).

In append/update mode, `write_page_index` runs over the **newly written** groups only; unchanged source groups are re-emitted verbatim from the footer cache (`RowGroupSource::Cached`), keeping their original `column_index_offset`. So `emit_column_index` inspects only the new groups. Trigger chain (all required):
1. Append to a fully-`ColumnIndex`ed source ⇒ `ParquetUpdater::new` adopts `write_statistics = true` (`update.rs:271–282`).
2. A newly appended group has an opaque `BINARY`/BLOB column whose row-group max prefix is eight `0xFF` bytes and exceeds 9 bytes ⇒ Fix 13 emits `max_value = None` ⇒ `pages_support_column_index()` returns false ⇒ `emit_column_index = false`.
3. New groups get `OffsetIndex` only; cached groups keep `ColumnIndex` + `OffsetIndex` ⇒ the footer advertises a `ColumnIndex` on some row groups but not others — the exact state `write_page_index`'s own doc comment promises never happens.

**Why Moderate, not Critical:** the `OffsetIndex` stays uniform, so the arrow-rs `with_page_index(true)` "missing offset index" hard-failure that Fix 7 targets is *not* triggered. `ColumnIndex` is an optional per-chunk optimization conformant readers tolerate as absent, so the practical effect is only looser pruning on the new groups — and this is still a net improvement over master (where appended groups previously had neither index, i.e. a partial `OffsetIndex`). It nonetheless contradicts the stated all-or-nothing `ColumnIndex` invariant and is the one unguarded direction the new tests don't cover. Suggest folding the cached groups' `ColumnIndex` presence into `emit_column_index` (or stripping cached `ColumnIndex` when a new group can't be indexed), plus a test for *indexed source + new unbounded-max Binary page*.

### M2 — `pages_support_column_index` serializes every page's stats twice at finalize
`core/rust/parquet2/src/write/indexes/serialize.rs:70–77`

The presence gate calls `serialize_statistics(stats)` for **every data page of every column** of every fresh row group — which `clone()`s the min and max `Vec<u8>` (`statistics/binary.rs:48–58`, `primitive.rs`) — purely to read two `Option::is_some()` flags, then discards the result. If the file then emits the index, `serialize_column_index` serializes the same pages **again**. That is `pages × columns × 2` serializations + throwaway min/max clones per file finalize (thousands for a large export / O3 merge). The presence check needs no serialization: add an allocation-free trait accessor (e.g. `Statistics::is_column_index_representable()` → `min_value.is_none() || max_value.is_some()` per concrete type), eliminating the first pass. Footer-time, not per-row, but real and fully avoidable.

## Minor
- **m1** — `build_copied_page_index` (`update.rs:1283,1314`) allocates `vec![0u8; oi_length/ci_length]` up to the i32 cap (~2 GB) from footer-supplied lengths before `read_exact`; only negative lengths are rejected. On a corrupt/truncated footer this zero-allocates a large buffer before the read fails. Low reachability (trusted partition files), but since the code already guards corrupt footers, bounding `length` against remaining file size — and reusing one scratch buffer instead of a fresh zero-filled `vec` per column — is cheap.
- **m2** — `into_thrift` (`metadata/file_metadata.rs:101`) now always emits `column_orders` even when statistics are off. Harmless (`TYPE_DEFINED_ORDER` is the default and readers only consult it when stats exist), but slightly broader than the spec's "present when min/max written".
- **m3** — CHAR (`UInt16`) takes the signed per-page stats path but the unsigned chunk-reduce. Benign (values always `[0,65535]` as i32, so signed == unsigned), but worth a one-line comment so a future 16-bit-signed type doesn't inherit the mismatch unnoticed.

- **`testBasics` size `44690` not updated** — passes an empty `params`, so `/exp` defaults to **CSV**; parquet footer growth doesn't apply. Correct as-is.
- **CodeRabbit underscore nit** — already addressed; every PR-added 5-digit size literal uses underscores. The only bare 5-digit literals left (`44690`, `10091`) are pre-existing, untouched lines.
- **`copied_page_index` vector desync** — `add_row_group` is the shared helper; all five parallel vectors stay length-synced across `write`/`replace`/`insert`/`write_raw_row_group_with_index`.
- **`reduce_int_primitive` empty-input panic** — `reduce` returns early on empty stats before dispatch; the peek is unreachable when empty.
- **`binary_upper_bound[..8]` / `utf8_floor_len value[end]` OOB** — only reached when `len ≥ 9` / `len > max_len`; provably in-bounds.
- **`convert_from_parquet` `m.schema[i]` panic** — guarded by the `meta.schema.len() != col_count` check before the loop.
- **`num_values: row_count as i32` truncation** — pre-existing (#6875), not this PR; unrealizable scale.
- **`next_char`/`truncate_max_utf8` validity & overflow, `has_unbounded_max` dropping a valid TEXT/fixed-len max** — verified: floor/ceiling logic is byte-wise sound for all Unicode incl. carries and the surrogate gap; the `(min=Some, max=None)` sentinel is produced only by the opaque-Binary all-`0xFF` branch.

## Cross-context checks (all SAFE)
- **Fix 10/12 unsigned ordering complete & exclusive:** every `encode.rs` dispatch enumerated — IPv4 is the only `true`; Geo*/Date/Timestamp/Long/Char/Byte/Short/Float/Double all `false`; the removed `StatsUpdater<i64,true>` impl has no remaining caller.
- **QuestDB's own reader:** `_pm` build (`convert.rs:387`) skips the max branch on `max_value = None` ⇒ treated as unbounded; new `column_orders`/`is_*_value_exact` parsed without affecting QuestDB reads.
- **`boundary_order` index-space-safe:** ASCENDING/DESCENDING only when `primary.column_idx == leaf column_idx`; every QuestDB column maps to one leaf, so a non-sorted column is never declared ordered.

---

**Summary:** approve with changes. Draft findings ~16 → **5 verified** (2 Moderate, 3 Minor), **11 dropped** as false positives after source verification. In-diff vs out-of-diff: **5 in-diff, 0 out-of-diff** (diff is large but self-contained; cross-context callers explicitly cleared). Please address M1 and M2 (or waive consciously) before merge.

<sub>Reviewed at level 3 with the QuestDB `review-pr` skill (adversarial, blocking, mission-critical).</sub>

RaphDal:
# Code Review — PR #7283: `chore(core): improve Parquet interoperability with external readers and Iceberg tooling`

**Level 2** review (7 parallel agents + independent lead verification), 2026-06-26.

**Verdict: Approve.** No Critical or Moderate code defects found across 7 review dimensions (correctness, concurrency, performance ×2, resource/panic safety, tests, code quality/metadata) plus an independent lead pass over the stats core, thrift codegen, gating truth table, and decimal/fixed-len ordering. Only Minor items remain. The `DO NOT MERGE` label is still on the PR — that is the author's gate, not a code issue.

This is a large (13-fix, ~4.4k-line) on-disk-format PR, and it is unusually careful: defensive against corrupt input, heavily and meaningfully tested end-to-end against the external `arrow`/`parquet` crate reader, with an honest, tradeoff-forward description. The four silent-row-drop fixes (10–13) that `column_orders` makes reader-authoritative are correct, conservative (floor/ceiling bounds), panic-safe at the JNI boundary, and backed by external-reader e2e tests that replay the pruning and assert real rows survive.

---

## Critical

None.

---

## Moderate

None. The four silent-row-drop fixes (10–13), the thrift wire codegen (fields 7/8), the all-or-nothing OffsetIndex/ColumnIndex gating, the `column_orders` leaf-count, the UTF-8/opaque-Binary stats truncation, and the unsigned-ordering threading were each verified correct against source, with strong unit + external-reader (Arrow) regression coverage.

---

## Minor

1. **`parquet2/write/file.rs:1142-1159` + update.rs accounting — stripped ColumnIndex bytes not counted as unused.** *(in-diff)* When an append strips a cached group's ColumnIndex (the `append_unbounded_max_binary_strips_cached_column_index` path), the orphaned ColumnIndex bytes become dead space but are never added to the `_pm` `unused_bytes` total. Effect: `_pm` under-reports free space by a few KB, marginally delaying compaction. No correctness/read impact. *Fix (optional): add the cleared `column_index_length` to the updater's unused-bytes counter.*

2. **`parquet2/write/stream.rs` `FileStreamer::end` — no Fix-13 gate (latent).** *(in-diff)* The async streamer writes the ColumnIndex unconditionally and would error (`"max value of a page is required"`) on an unbounded-max sentinel page. It is **not reachable**: the `async` feature is off for qdbr's parquet2 dependency and all writes route through the gated `FileWriter`/`ParquetFile`. The PR documents this. *Fix (optional): apply the same `pages_support_column_index` gate so the path cannot regress if async is ever enabled.*

3. **Test coverage gaps (all Minor, non-blocking).** *(in-diff)*
   - No Java end-to-end O3-merge / `ALTER TABLE ADD COLUMN` integration test that inserts into a partitioned parquet table and reads it back through an external reader — Fix 7/13 are covered only at the Rust `ParquetUpdater` level (which is strong, but the full TableWriter→JNI→updater path has no external-read assertion).
   - No e2e for an all-null column or 0-row/empty partition through the full writer (unit-only).
   - Fix 5 primitive (int/double) `is_*_value_exact == None` is asserted only in a parquet2 unit, not end-to-end.

4. **Hardening notes (not bugs).** *(in-diff)*
   - A `debug_assert!(new_groups_indexable || !source_offset_indexed)` in the Mode::Update branch of `ParquetFile::end` would pin the "no mixed OffsetIndex" invariant under `-ea` if a future caller ever copies row groups during append (today it cannot — `O3PartitionJob` only copies under `isRewrite`).
   - Fix 6 makes ColumnIndex pruning **load-bearing** on `sorting_columns[0].descending` matching the physical page order. True for all four current sort sources; a future path that declares a sort it does not physically honor would cause silent wrong pruning. Worth a comment/guard at the `SortingColumn` declaration sites.
   - `allow_column_index` / `ascending` booleans lack the `is_`/`has_` prefix — idiomatic Rust and in the vendored fork, so defensible; note only.

---

## Downgraded (false positives — dismissed after source verification)

- **Decimal stats use byte-wise (unsigned) order, inconsistent with the signed Decimal sort order Fix 1 declares** — *Dismissed.* Decimals don't use `BinaryMaxMinStats`; `encode_decimal` → `decimal_segments_to_page` uses `MaxMin<Decimal>` with `Decimal::ord`, which is a correct signed comparison (Decimal8/16/32/64 = `self.0.cmp`; Decimal128/256 = derived `Ord` over `(i64, u64, …)` limbs = signed-high/unsigned-low). Per-page (signed), chunk reduce (`ord_binary_signed`), and reader (Decimal logical type) all agree. UUID/Long128/Long256 have no logical type → byte-wise on both sides → also self-consistent.
- **`has_min_value`/`has_max_value` are dead code** — *Dismissed.* Used at `indexes/serialize.rs:74` as the per-page ColumnIndex-validity gate (`pages_support_column_index`), the heart of Fix 13.
- **"C1" review-label left in a shipped comment (`decode.rs:4530`)** — *Dismissed.* The line is pre-existing (not in this PR's diff), so out of scope for this PR.
- **`ExpParquetExportTest.java:132` (44690) wasn't bumped like its ~30 siblings** — *Dismissed.* That call uses the `(url, len, param, sql)` overload with empty `params`, so it sends no `fmt=parquet` and exports CSV; `column_orders` growth correctly does not touch it. Pre-existing, identical on master.
- **Mixed offset-indexed/unindexed footer (arrow `with_page_index(true)` fails)** — *Dismissed.* Both the update.rs review and the parquet2 review independently proved it unreachable: the only `None`-page-index raw-copy runs solely under `is_rewrite` (Mode::Write, where the whole index is dropped uniformly), and the Mode::Update copy returns `None` only when the source also lacks an OffsetIndex (forcing `source_offset_indexed = false`). The two conditions are mutually exclusive.

---

## PR metadata

Compliant with CLAUDE.md conventions. The title is valid Conventional Commits and reads well for release notes. `chore` over `fix` is defensible (internally-discovered audit findings, not customer-reported bugs) and explicitly justified in the body. The tone is level-headed and foregrounds the tradeoffs (looser pruning, the pre-existing legacy-copied-group condition exposed by `column_orders`) with the same weight as the improvements. Labels match scope; the `Bug`+`chore` pairing is intentional (correctness dimension flagged for triage). The dependency bumps (criterion dev-only 0.4→0.8.2 dropping `atty`; `memmap2`→0.9.11 for RUSTSEC-2026-0186) are correctly scoped. No `Fixes #NNN` is needed (no linked issue). The `DO NOT MERGE` label is still present.

---

## Summary

**Verdict: Approve.** Across 7 parallel review dimensions plus an independent lead pass, **no Critical or Moderate code defects were found**. The hardest parts — the four silent-row-drop corrections that `column_orders` makes reader-authoritative (IPv4 signed/unsigned per-page + chunk reduce; opaque-Binary all-`0xFF` unbounded-max sentinel carried through both the chunk reduce and the page index), the thrift field 7/8 wire format, and the all-or-nothing OffsetIndex/ColumnIndex gating across the rewrite/append/ADD-COLUMN paths — are correct, panic-safe at the JNI boundary, and backed by external-reader (Arrow) e2e tests that replay the pruning and assert real rows survive.

- **Findings: 8 candidates verified, 5 dropped as false positives** (2 from lead leads — decimal ordering and the dead-code hypothesis — and 3 from agents).
- **In-diff vs out-of-diff: all confirmed findings are in-diff; zero out-of-diff bugs.** This is a genuine result — the cross-context pass actively traced the Java JNI callers (`O3PartitionJob.java`) and the enterprise `_pm` consumers and confirmed the changed contracts hold (the mixed-footer hazard is unreachable; `QdbMetaCol.id` is additive and backward-compatible).
- **Tradeoffs (correctly disclosed by the PR, not regressions to fix):** looser row-group/page pruning in the narrow opaque-Binary all-`0xFF` and unbounded-max cases, and the conservative `reduce_exact_flag` that can under-claim exactness — both safe directions (bounds stay valid; only a reader optimization is occasionally forfeited).

RaphDal:
@bluestreak01 - Thanks for the review, I addressed your feedback in 9e8e6c21dd2cf18b72be3a455d14e50917743359 and made a fresh review in https://github.com/questdb/questdb/pull/7283#issuecomment-4809205372

bluestreak01 (dismissed):
Approving after a full mission-critical review (all 11 review agents + per-finding verification).

The central invariant holds: no Parquet min/max statistic becomes byte-wise tighter than the real data once `column_orders` is declared. Min truncates down to a prefix, max rounds up or is omitted as unbounded, unsigned ordering is threaded through both per-page and chunk-reduce paths, and the page-index all-or-nothing gating is internally consistent and verified end-to-end against an independent Arrow reader. The cross-context callsite audit confirmed every trait impl, struct literal, and changed signature compiles. No correctness regressions, no resource leaks, no new reachable JNI panics.

Non-blocking items worth a look (none gate the merge):
- M1 (latent): `boundary_order_for_column` stamps ASCENDING/DESCENDING on the primary sort column unconditionally, while this PR adds non-order-preserving max-widening (truncate_max_utf8, binary_upper_bound). Not reachable today since the only sort column is the designated INT64 timestamp (never byte-truncated), but a future String/Binary sort key could silently break the declared monotonicity. A debug_assert/comment tying Fix 6 to Fix 4 would be cheap insurance.
- M2 (perf): build_copied_page_index does per-column unbuffered seek+read on the O3-copy path; one bulk read per row group or a BufReader would cut the syscall count on wide tables.
- Minor: reduce_exact_flag over-clears the exact bit across multi-page chunks (conservative, safe); a couple of foldable double-passes and a 'pages' variable shadow in write_page_index; a few low-risk test gaps (nested-array column_orders count, FileStreamer async, dict-IPv4 + multi-page chunk reduce).

The strongest draft finding (adjust_column_chunk_offsets unchecked +=) was disproven as pre-existing (byte-identical on master; release wraps rather than panics) and is out of scope.

PR description is exemplary in surfacing tradeoffs, and bundling the 13 related fixes is correct since Fix 1 is what makes Fixes 10-13's bounds reader-authoritative.

RaphDal:
# Code Review — PR #7283

**Title:** `chore(core): improve Parquet interoperability with external readers and Iceberg tooling`
**Branch:** `rd/fix-parquet-stats` · +5423/−525 · 60 files · 14 bundled Parquet fixes, almost entirely `core/rust/`
**Review level:** 3 (full mission-critical pass: 7 parallel review agents + independent per-fix source verification and JNI/Java read-side tracing)

**Verdict: Approve.** No Critical, no Moderate. Everything below is Minor or test-hardening.

> Note: this file supersedes an earlier level-3 review that raised a Critical `ord_binary`
> multi-page prefix bug (C1, "app"/"apple" understating the chunk max). That is now fixed on
> this branch — `ord_binary` (`statistics.rs:173-183`) uses `a.cmp(&b)` (full lexicographic
> compare with the length tiebreaker), and commit `a62021bea8` added the
> `reduce_binary_prefix_bounds_apply_length_tiebreaker` test asserting min=`"app"`/max=`"apple"`.

The central safety invariant holds: no min/max statistic ever becomes byte-wise tighter than
the real data once `column_orders` is declared, and nothing consumes a truncated bound as an
exact value.

---

## Critical

None.

---

## Moderate

None. The four highest-risk hypotheses were each traced to source and disproved:

- **Truncated/omitted text & binary min/max (Fix 4/5) → silent wrong results.** Disproved.
  These stats feed only *conservative* row-group pruning — a loose superset bound makes a value
  test "inside" more readily, so it can only prune *less*, never drop a matching row
  (`row_groups.rs:1266/1537`, `skip.rs`). The one stat surfaced as an exact *value* is the
  designated timestamp (`rowGroupMin/MaxTimestamp` → `ParquetTimestampFinder`), which is INT64
  and never truncated. `min()`/`max()` aggregates decode the column; no pushdown reads a stat
  as an answer. The read side also requires *both* min and max present (`skip.rs:262/281`,
  `row_groups.rs:1285`), so an omitted opaque-Binary max never mis-prunes.

- **Mixed page-index footer → arrow-rs `with_page_index(true)` hard read failure.** Unreachable.
  The only producer of an unindexed copied group (`copyRowGroup`) is gated behind
  `O3PartitionJob.java:521 if (isRewrite)` — rewrite is Mode::Write with a uniform
  all-or-nothing gate. In Mode::Update every new group is freshly encoded
  (`new_groups_indexable` always true), and a copied group would anyway trip a clean
  `ordinal == None` error (`update.rs:1092`) before any footer commits.

- **M1 — append an unbounded-max Binary page to a fully-ColumnIndexed source → mixed ColumnIndex.**
  Prevented. `file.rs:1142-1159` strips the cached groups' `ColumnIndex` (flipping Cached→Fresh
  so they re-serialize with the pointer cleared, OffsetIndex retained). Tested in true update
  mode by `append_unbounded_max_binary_strips_cached_column_index` under `with_page_index(true)`.

- **Fix 14 (the one live defect) correctness.** Confirmed empirically — reverting the bloom
  remap makes the regression test fail with the stray bloom landing on the timestamp
  (`[0,1,2]` vs `[0,1]`); write-side target order matches the read-side probe (`skip.rs:215`).

---

## Minor (none block merge)

1. **Weak assertion in the new bloom test** — `ParquetRowGroupPruningTest.java`,
   `testBloomFilterDroppedColumnO3Rewrite` (~L522). The final
   `Assert.assertTrue(getRowGroupsSkipped() > 0)` is weaker than its comment ("proving the copied
   blooms stay live"): for `WHERE b = 2095`, row group `[2050,2080]` is pruned by min/max alone,
   so `> 0` passes even if the copied bloom were broken. The real regression coverage is the
   *positive* `b = 2100` lookup (a misattributed bloom would prune it and return empty). Tighten
   to an exact skip count, or soften the comment.

2. **`build_copied_page_index` per-column seek+read** — `update.rs:1300-1301,1341-1342` (perf).
   O(4·columns) syscalls per copied row group on the O3-rewrite path; the source lays each group's
   OffsetIndex and ColumnIndex out contiguously, so this collapses to ~2 reads/group. Real but
   syscall-count-only (index bytes are KB, page-cached, dwarfed by the MB–GB data copy).

3. **`reduce_vec8` clones the challenger unconditionally** — `statistics.rs:31` (perf).
   `Some(ord(x, y.clone(), max))` clones the incoming page's `Vec<u8>` every fold step even when
   the accumulator wins; comparing by reference and cloning only on a win roughly halves the
   allocations. Small values (≤65 bytes, truncated), finalize path.

4. **`next_utf8_string` double allocation** — `util.rs:305-316` (perf). Builds a `Vec<char>` then
   a `String`; an in-place tail approach avoids both. Runs at most per-page-chunk and only when a
   text max exceeds 64 bytes.

5. **Empty-source `source_column_indexed` asymmetry** (= CodeRabbit #1) — `file.rs:1048`. For a
   0-row-group source, `source_offset_indexed` is `true` (vacuous `.all()`) but
   `source_column_indexed` is `false` (the `!is_empty()` guard), so the first append forgoes a
   `ColumnIndex` it could carry. Output stays uniform and valid — a missed optimization on a
   marginal path, not a correctness bug.

6. **Stripped `ColumnIndex` bytes not counted as unused** — the M1 strip path (`file.rs:1142-1159`)
   orphans the cached `ColumnIndex` bytes without adding them to `accumulated_unused_bytes`, so
   `_pm` under-reports free space by a few KB and marginally delays compaction. No read/correctness
   impact.

7. **Cross-context test coverage** — Fix 7/13/14 are covered strongly at the Rust `ParquetUpdater`
   level and one Java DROP-COLUMN+O3 path, but there is no Java end-to-end test for
   `ALTER TABLE ADD COLUMN` + O3-merge readback, nor for the IPv4-unsigned / opaque-Binary type
   paths through the full `TableWriter→JNI→updater` chain. Worth a follow-up, not blocking.

**Trivial nits:** `increment_encoding_stats` (`column_chunk.rs`) lacks the rationale comment its
sibling `push_unique` has (both correctly use a linear find over ≤3 pairs);
`adjust_column_chunk_offsets` uses unchecked `+=` while `build_copied_page_index` uses
`checked_add` for the same delta (harmless at realistic scale); `into_thrift`
(`file_metadata.rs:101`) always emits `column_orders` even when statistics are off (harmless,
slightly broader than the spec's "present when min/max written").

---

The four **open CodeRabbit findings** all resolve to non-blockers — this is the main new signal
for the author:

- **`binary_upper_bound` panics on <8-byte input** (CodeRabbit #3) — **False.** The sole caller
  guards `max.len() > SIZEOF_I64` (`util.rs:162`) before calling it, so `max_value[..8]` is always
  in bounds.
- **Async `stream.rs` missing the Fix-13 ColumnIndex gate** (CodeRabbit #4) — **Dead code.**
  `stream.rs` is `#[cfg(feature = "async")]`; parquet2's `default` features exclude `async` and
  qdbr enables no extra features, so it is not compiled.
- **`append_base` allocation before validation** (CodeRabbit #2) — corrupt-`_pm`-header-only.
  `append_base` is a `u64` offset Java derives from the `_pm` header under the table write lock;
  a bad value requires on-disk corruption, which the project deliberately does not harden the
  `_pm` read path against. Note only: the sibling `build_copied_page_index` *does* `try_reserve`
  its footer-parsed lengths, so matching that here would be a consistency improvement — author's
  call.
- **`pages_support_column_index` serializes stats twice** (CodeRabbit M2 / earlier passes) —
  **Stale.** The current code uses O(1) accessors (`has_min_value()`/`has_max_value()`/
  `null_count()`, `serialize.rs:71-84`), not `serialize_statistics()`.

Also dismissed after source verification:

- `convert.rs:135 schema[i]` panic — guarded by the length check at `convert.rs:99-106`.
- The `column_order(i)` getter's unguarded `data[i]` — zero callers, dead code.
- An earlier review note that "`_pm` ignores the exact flags" — **outdated**. `apply_thrift_stats`
  reads them correctly at `convert.rs:378-379`, which fixes a latent "always claims exact" bug
  (the bounds could already be inexact from the pre-existing opaque-Binary rounding). The exact
  bit itself (`MIN_EXACT`/`MAX_EXACT` in `qdb-parquet-meta/types.rs`, not in this diff) is
  pre-existing `_pm` format, so this is a population fix, not a format change.
- The prior review's **C1 `ord_binary` prefix bug** — fixed on this branch (see the note at the
  top). Current `ord_binary` compares full `Vec<u8>` via `a.cmp(&b)`, and the `"app"`/`"apple"`
  test guards it.

---

## Per-fix verification summary

- **Fix 1 `column_orders`** — one `TypeDefinedOrder` per leaf column; read-side `parse_column_orders`
  uses panic-safe `.zip()`.
- **Fix 2 LIST `element` rename** — pure field-name change; tests assert canonical
  `bids.list.element.list.element` paths.
- **Fix 3 `resolve_column_id`** — correct precedence (non-negative qdb id > field_id > -1);
  `convert.rs:135` guarded.
- **Fix 4/5 truncation + exact flags** — max bound `P+1` provably ≥ every full value whose 9-byte
  clamp is ≤ the stored max; thrift fields 7/8 spec-correct on both read and write, unknown-field
  skip preserves compatibility.
- **Fix 8 encoding_stats** — linear find over ≤3 (page_type, encoding) pairs, correct; empty → None.
- **Fix 9 UINT_32 converted type** — additive; physical type/encoding/stats/reads unchanged.
- **Fix 10 dict-encoded IPv4 unsigned** — IPv4 is the only `UNSIGNED_STATS=true` across all three
  encodings; `ord_unsigned` casts i32→u32 / i64→u64; stats computed over dictionary *values*, not
  RLE index codes; matches the pre-existing unsigned reader (`row_groups.rs:1283`). The blanket
  `StatsUpdater<i64, UNSIGNED>` impl was removed, so unsigned-i64 is now a compile error.
- **Fix 12/13 chunk reduce** — unbounded-max sentinel propagation correct; `reduce_exact_flag`
  conservative (over-clears, never wrongly claims exact); unsigned/​signed chunk-reduce split
  matches per-page.
- **Fix 14 bloom reorder** — target-order bitset vector built in lockstep with the column chunks
  via the same field-id → position map; empirically validated.

---

## Summary

- **Verdict: approve.** ~20 draft candidates → **7 verified Minor**, the rest dropped as false
  positives / stale / dead-code after source verification.
- **In-diff vs out-of-diff: 0 out-of-diff bugs.** A real result, not an underrun — the
  cross-context pass reached and cleared the read-side pruning consumers (`skip.rs`,
  `row_groups.rs`), the O3 JNI callers (`O3PartitionJob`), the sort-order declaration sites
  (`jni.rs`, `update.rs`), and the timestamp-finder path. All verified findings are in-diff or
  test/coverage.
- **Tradeoffs (disclosed by the PR, not regressions):** Fixes 4/11/13 deliberately loosen or omit
  bounds in narrow cases in exchange for never emitting an invalid one — the same conservative
  direction as the all-or-nothing page-index degradation.
- The `chore` type + `Bug` label pairing, level-headed tradeoff-forward description, and
  Conventional-Commits title are all consistent with house conventions.

bluestreak01 (dismissed):
Approving after a full level-3 mission-critical review (all 11 review dimensions + per-finding source verification + an independent re-run of the Rust test suite).

The central safety invariant holds and was verified from source, not from the description: once column_orders is declared, no min/max statistic becomes byte-wise tighter than the real data. Min truncates down to a prefix (valid floor), max rounds up or is dropped-as-unbounded (valid ceiling), unsigned ordering is threaded through both the per-page StatsUpdater and the chunk-level reduce, and the page index degrades uniformly all-or-nothing. Confirmed end-to-end against an independent Arrow reader via the encode_statistics tests that replay external-reader pruning and assert real rows survive.

Independently verified:
- cargo check --all-targets clean; qdbr clippy zero warnings.
- Green: util truncation (35), convert _pm exact-flags + resolve_column_id (8), parquet2 statistics reduce/serialize gating/boundary order/types (23), update.rs page-index + bloom-remap + column_orders (28), encode_statistics (13), decode_pm_e2e (16).

The four highest-risk hypotheses were each traced to source and disproved: the mixed offset-indexed footer is unreachable (both raw-copy paths run only under O3PartitionJob isRewrite -> Mode::Write, where the all_indexable gate degrades uniformly); the text/Binary truncation only ever loosens bounds; the unsigned-reduce dispatch matches the per-page path and makes unsigned-i64 a compile error; and the Fix 14 bloom remap is index-aligned via the field-id -> target-position map.

No Critical, no Moderate, zero out-of-diff bugs. Non-blocking items worth a look (none gate the merge):
- Latent: boundary_order stamps ASCENDING/DESCENDING on the primary sort column unconditionally while max-widening is non-order-preserving; safe today only because the sort column is the INT64 timestamp. A debug_assert/comment tying Fix 6 to Fix 4 would be cheap insurance.
- Latent: async stream.rs lacks the sync path's pages_support_column_index gate; harmless because the async feature is not compiled in this build.
- Perf: build_copied_page_index does per-column unbuffered seek+read (syscall-count only); next_utf8_string double-allocates.
- Test: testBloomFilterDroppedColumnO3Rewrite's final getRowGroupsSkipped() > 0 is weaker than its comment (the b=2095 row group is prunable by min/max alone); the positive b=2100/5000 lookups carry the real regression coverage.

CodeRabbit's open findings resolve to non-blockers: binary_upper_bound's <8-byte panic is guarded by its sole caller; the append_base allocation is pre-existing (PR #7268, byte-identical on master) and corrupt-_pm-only; the async stream.rs gate is dead code.

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

PR #7284, opened 2026-06-18
feat(sql): parallel covered-index decode for aggregation and filter

nwoolmer:
## What

Moves covered-index column decode for aggregation and filter queries off the single dispatch thread onto the async reduce-worker pool, eliminating the dispatcher serialization that made the covered plan 3–4× slower than it needed to be. The covered decode now **parallelizes 3–4×**; for selective predicates (the covering-index use case) the covered plan is far faster than a full scan, and `count(*) WHERE sym = '<lit>'` is answered from posting-index metadata with **no decode**.

## How

- **Single choke point.** A covered page frame is a metadata-only decorator over its base frame. `PageFrameAddressCache.add()` is the one place that stashes per-frame covered metadata; every consumer reads covered-ness back from the cache (mirrors parquet metadata handling).
- **Worker decode arm.** `CoveredColumnDecoder` + per-pool `CoveringBuffers` decode covered columns from the sidecar and synthesize the symbol key, on both the flyweight and record fast-paths. All column types incl. var-size (VARCHAR/STRING/BINARY/ARRAY). The fixed-width layout is a single shared `writeFixedWidthCovered` used by both the worker and eager paths so they cannot drift.
- **Wrapper propagation.** Projection (`SelectedPageFrame`) and null-pad (`ExtraNullColumnPageFrame`) wrappers propagate covered metadata through their column remap.
- **Concurrent-read-safe posting reader.** Detached per-worker cursors; a freeze before dispatch (in both the base `PageFrameSequence` and `UnorderedPageFrameSequence`) makes `reloadConditionally()` a no-op and `openRequiredSidecars()` a no-op while frozen, so workers never mutate shared reader state; assert-not-extend at gen-load sites.
- **Metadata-only single-key production.** O(genCount) `countMatchesClamped` + `selectKthMatch` random access into the EF/flat/delta posting index for the chunk row-id spans, replacing the O(rows) dispatch-thread traverse. Multi-key stays eager.
- **count() metadata fast path.** `SingleKeyCoveringCursor.size()` now uses `countMatchesClamped` (exact per-gen first/last coverage), so `CountRecordCursorFactory`'s `size()` short-circuit returns an O(genCount) answer that decodes nothing and never walks postings. (Previously it used a primitive that false-bailed on freshly-resealed partitions and fell back to an O(rows) traverse.)

## Performance

Two complementary views, both warm JMH (`org.questdb.CoveredIndexDecodeBenchmark`, in-process, shared engine), 20M rows, 8 workers, filtered key present at exactly the stated fraction of rows. (Plus a cold-cache note at the end.)

### 1. Parallelism — the PR's core contribution

Moving covered decode onto the reduce-worker pool. `A` = parallel covered (8 workers), `B` = serial covered (1 worker); **B/A is the speedup** (at 10% selectivity, ms):

| shape | A parallel | B serial | **speedup** |
|---|--:|--:|--:|
| sum | 2.15 | 7.26 | **3.37×** |
| multi-agg | 2.25 | 6.76 | **3.00×** |
| first/last | 2.08 | 7.61 | **3.66×** |
| residual filter | 3.04 | 10.42 | **3.43×** |
| group by (symbol) | 7.12 | 27.06 | **3.80×** |
| group by (varchar) | 12.62 | 33.33 | **2.64×** |
| count | 0.009 | 0.009 | 1.0× (metadata, nothing to parallelize) |
| filter projection | 40.67 | 37.12 | 0.91× (output-bound: draining the rows dominates) |

The decode-heavy shapes get **~2.6–3.8×** off the old serial path. This is harness-independent (both sides measured identically) and is the regression this PR removes.

### 2. Covered vs a warm parallel scan — selectivity sweep

Whether a covering-index plan beats a full scan is a selectivity question. Covered speedup over the scan (scan ÷ covered; **>1× = covered faster**):

| selectivity | sum | multi-agg | first/last | residual | filter-proj | gb (sym) | gb (vchar) | count |
|---|--:|--:|--:|--:|--:|--:|--:|--:|
| **0.1%** | **18.7×** | **18.9×** | **18.7×** | **15.5×** | **3.8×** | **6.6×** | **6.8×** | **614×** |
| **1%**   | **3.5×**  | **3.6×**  | **3.7×**  | **2.85×** | **3.4×** | **1.50×** | **1.43×** | **189×** |
| **5%**   | 1.10×     | **1.24×** | 1.10×     | 0.68×     | **1.61×** | 0.47×   | 0.62×   | **80×** |
| **10%**  | 0.80×     | 0.96×     | 0.73×     | 0.55×     | **1.49×** | 0.36×   | 0.30×   | **60×** |
| **25%**  | 0.77×     | 1.01×     | 0.66×     | 0.46×     | **1.44×** | 0.18×   | 0.31×   | **35×** |
| **50%**  | 0.40×     | 0.59×     | 0.36×     | 0.25×     | **1.45×** | 0.19×   | 0.36×   | **15×** |

`count()` (metadata, decode-free) and **filter projection** (covered reads only matching rows' columns; the scan reads every row's) win at *every* selectivity. Aggregations win below a **~5%** crossover, residual below **~4%**, keyed group-by below **~2%** (the vectorized scan group-by is very efficient). Above the crossover a full scan is faster — where the planner ideally costs in the scan (a cost-model follow-up; `count()` routing is already handled here).

### Cold cache — bytes read from disk

The warm wall-clock above understates covered on cold/slow storage, because on NVMe the covered path's *fixed CPU* (reopen every partition's readers + traverse the posting index + decode) dwarfs the I/O it saves. The hardware-independent truth is **bytes actually fetched from the block device** (`/proc/self/io:read_bytes`, which counts mmap page faults), measured with the page cache evicted before each query (5M rows):

| shape | sel | COV disk | REF disk | covered reads |
|---|--:|--:|--:|--:|
| sum | 1% / 10% / 50% | 14 / 21 / 41 MB | 60 MB | **4.3× / 2.9× / 1.5× less** |
| filter_project | 1% / 10% / 50% | 43 / 67 / 135 MB | 160 MB | **3.7× / 2.4× / 1.2× less** |
| first_last | 1% / 10% / 50% | 14 / 21 / 41 MB | 60 MB | **4.3× / 2.9× / 1.5× less** |
| group by (symbol) | 1% / 10% / 50% | 23 / 32 / 61 MB | 80 MB | **3.5× / 2.5× / 1.3× less** |

REF (full scan) reads a **constant** number of bytes regardless of selectivity; covered **scales with selectivity** — it reads only the matching rows' columns, packed per symbol. Covered reads strictly *fewer* bytes at every point. The eviction is genuinely cold (REF reads exactly `sym`+`px` = 5M×12B = 60 MB; warm would read ~0).

So once storage is bandwidth-bound, cold time ≈ `fixed_cpu + bytes/B` and covered wins by ~the byte ratio. Projecting `sum` @ 1% from the measured bytes + measured NVMe overhead:

| device | COV | REF | covered |
|---|--:|--:|--:|
| NVMe ~2.5 GB/s | 70 ms | 27 ms | 0.4× (fixed CPU dominates) |
| SATA/network ~100 MB/s | ~200 ms | ~600 ms | **~3×** |
| HDD ~30 MB/s | ~530 ms | ~2000 ms | **~3.8×** |

> An earlier version of this section claimed the cold advantage was "~neutral on NVMe". That conflated wall-time (fixed-CPU-bound on fast storage) with I/O. The byte measurement above is the correct like-for-like: covered reads 1.2–4.3× fewer bytes, so it wins on any bandwidth-bound (cold/slow) device by that factor.

Reproduce: `mvn -pl benchmarks -am package` then run `org.questdb.CoveredIndexDecodeBenchmark` (warm) with the JVM flags in the class javadoc; cold byte counts via the `/proc/self/io` probe described in the PR thread.

## Review

Two deep ("level-3") review rounds since the initial implementation:
- **Concurrency**: wired the freeze into the async (non-vectorized) keyed group-by path (`UnorderedPageFrameSequence`) — confirmed reachable, then verified by stress test that the gap was latent (eager warm + snapshot-stable reader) rather than a live UAF; made `openRequiredSidecars` a frozen no-op.
- **Correctness/guards**: >2GB ARRAY/STRING/BINARY overflow guards on both decode paths; `selectKthMatch`/`countMatchesClamped` skip a gen fully below `minValue` (perf, with white-box + negative-control tests); hard error on a covered re-decode row-count mismatch.
- **Memory**: covered decode buffers use global `NATIVE_INDEX_READER` accounting (an earlier per-query-tracker charge was reverted — it mis-charged a recycled tracker because covered buffers are freed at the next query, after the owning tracker is recycled).

## Test plan

- [x] `CoveringIndexTest` — 398/398 (equivalence oracle; worker output byte-equivalent to the serial path)
- [x] `CoveringIndexParallelDecodeTest` — 24/24 (incl. count metadata fast path; gated perf)
- [x] `CoveringIndexParallelGroupByReachabilityTest` — 3/3 (async group-by routing + concurrent-writer regression)
- [x] `PostingReaderConcurrentReadTest` — 6/6 (concurrency safety, incl. detached bwd cursor)
- [x] `PostingReaderSelectKthMatchTest` — 14/14 (random-access primitive + fully-below skip)
- [x] `ParallelGroupByMemoryTrackerTest` — 7/7 (per-query tracker unaffected)
- [x] Merged master; main compiles; full suite green with `-ea`


nwoolmer:
## Level-3 deep review

A multi-lens adversarial review of this branch (concurrency, cheap-path algorithm, native memory, per-type decode, wrapper integration, tests). The two headline findings were verified against source, not just statically flagged.

**Verdict: request changes.** 2 verified defects (one concurrency-CRITICAL, one correctness-HIGH) plus one latent-HIGH, all concentrated in the new warm/decode path. The green test run does **not** contradict this — the test gaps line up exactly with the bugs (see below).

### Findings

| # | Sev | Location | Issue |
|---|-----|----------|-------|
| 1 | **CRITICAL** | `AbstractPostingIndexReader.populateCacheForKey:505` + `PostingIndexFwdReader:406` | Frozen readers race on `putCacheEntries` |
| 2 | **HIGH** | `AbstractPostingIndexReader.populateCacheForKey:522-531` | Cache poisoning → ~1% silent wrong rows |
| 3 | **HIGH (latent)** | `PageFrameMemoryPool.publishAddresses:1671` / `:1578` | `includeIdx<0 ⟹ symbol-key` → wrong value + OOB |
| 4 | MEDIUM | `CoveringIndexRecordCursorFactory:1578` (`int cheapChunkBase`) | Overflow → silent row loss past 2³¹ matches |
| 5 | LOW | `PageFrameAddressCache.add:164` | New unconditional O(cols) probe on hot non-covered path |
| 6 | LOW | `CoveredColumnDecoder.writeBinary/writeString:206/181` | `int` truncation on >2 GB var values → OOB |
| 7 | NIT | `CoveredColumnDecoder.writeArray:142` | Non-double ARRAY decodes as zeros (pre-existing, shared w/ eager path) |

#### 1 — CRITICAL: frozen readers race on the genLookup cache write
The safety model is "warm → freeze → workers only *read* the shared reader." Two holes break it: `warmForKeys` (the full, correct warm) has **no production caller** — production warms via `populateCacheForKey`, which **bails on `genCount <= 1`** (`:505`). For a **single sparse generation** the cache stays cold, so a worker's detached cursor cache-misses, walks the gen, and at exhaustion runs `genLookup.putCacheEntries(...)` (`PostingIndexFwdReader:406`) — gated only on `!isCacheReplayMode`, **not `frozen`/`isDetached`**. Two workers on the same reader write concurrently → torn `IntLongHashMap`/`DirectLongList` writes during the window the design declares immutable → wrong results or crash. `putCacheEntries` doesn't bump `cacheVersion`, so the existing guard doesn't catch it.
**Fix:** warm single-sparse-gen via the metadata path (don't bail at `genCount<=1`), **and** make a detached/frozen cursor that reaches the cache write fail loud (defense-in-depth).

#### 2 — HIGH: `populateCacheForKey` poisons the cache for absent-in-range keys
The cursor records **no entry** when `start == prefixSum[k] == prefixSum[k+1] == end` (`PostingIndexFwdReader:757`); `selectSparseKeyCount` likewise guards `start >= end` (`:1609`). But `populateCacheForKey` reads only `start` and uses `counts[start] > 0` (`:528`). For a key in `[minKey,maxKey]` but **absent** from a sparse gen, `start == end` yet `counts[start]` is the *next* key's count (`> 0`) → spurious `packCacheEntry` pointing at a **different key's** postings. The SBBF pre-check only filters ~99% (1% FPP), so ~1% slip through → probabilistic silent wrong rows. The equivalence test masks this by putting every key in every gen.
**Fix:** read `end = prefixSum[k+1]` and `if (start == end) continue;` — mirror the cursor.

#### 3 — HIGH (latent): symbol-key conflation in the worker decode
`publishAddresses`/`ensureColumnBuffers` classify columns purely by `coveredIncludeIdx[q] < 0` ⟹ "symbol key, point at the 4-byte `symAddr`", never consulting `isColumnCovered`. A synthetic null-pad column (`ExtraNullColumnPageFrame`, which correctly reports `includeIdx=-1` for a non-symbol column) would be published pointing at `symAddr` → wrong value, and for a wider-than-4-byte type a native OOB read. Currently not planner-reachable (the null-pad wrapper is only emitted for window-join padding consumed via the record path), so latent — but the assumption is unsound and untested.
**Fix:** gate the `symAddr` branch on the column actually being the symbol key, not on the `includeIdx<0` catch-all.

#### Lower severity
- **4 (MEDIUM):** `cheapChunkBase`/`chunkBase` are `int` accumulators over a `long` match domain → overflow → silent under-count past 2³¹ (extreme scale). Make them `long` / fall back when `total > Integer.MAX_VALUE`.
- **5 (LOW):** `add()` adds a second unconditional per-column loop for every frame incl. non-covered. Fold into the existing address loop.
- **6 (LOW):** `(int)(Long.BYTES + len)` / `charCount*2` truncate for >2 GB BINARY/STRING → OOB. Widen + guard.
- **7 (NIT):** non-vanilla/non-double ARRAY decodes as zeros — pre-existing, shared with the eager writer.

### Why the green test run didn't catch this
The test gaps map 1:1 onto the defects:
- **Parallel type coverage is DOUBLE/LONG/VARCHAR/SYMBOL only** — STRING/BINARY/ARRAY/LONG256/UUID/GEO* never flow through the worker decode (hides #3/#7).
- **The concurrency test is a 2-thread, single-rendezvous, one-pass, ~15-row smoke check** with no iteration loop and no single-sparse-gen case (would pass through #1); backward-reader concurrency untested.
- **The equivalence oracle puts every key in every gen** (masks #2).
- **The perf/scan-parity verdict is computed but never asserted** — the headline number has no automated backing.
- The prior ARRAY page-frame oracle was **weakened to assert placeholder zeros** and not relocated.
- New suite uses twin-diff + self-rerun, **no hardcoded-literal assertions** (vs. 455 in the base test).

### Verified clean
Native memory is leak/UAF/double-free-free (balanced allocs incl. OOM paths; offset-based aux + post-loop base capture defeat dangling-flyweight risk); the `gen ⊆ valueMemSize` invariant + assert-not-extend hold at all 6 sites; freeze/unfreeze are exception-symmetric and inert for non-covered queries; the `bindGeneration` guard mirrors parquet; fixed-width + VARCHAR-split + null-sentinel decode is byte-correct; wrapper `columnCrossIndex` remap and default-getter safety check out. The architecture is sound — the defects are localized.

<sub>🤖 Produced with Claude Code (multi-agent deep review).</sub>

nwoolmer:
## Fixes pushed — addressing the review

All 7 findings are fixed in `455771e`, each with tests:

| # | Sev | Fix |
|---|-----|-----|
| 1 | CRITICAL | `populateCacheForKey` now warms single-sparse-gen layouts (dropped the `genCount<=1` bail) so workers replay read-only; detached worker cursors never write the shared cache; `putCacheEntries` asserts `!frozen` (defense in depth). |
| 2 | HIGH | `populateCacheForKey` uses `start != end` (mirroring `loadSparseGenByPrefixSum` / `selectSparseKeyCount`), not `counts[start] > 0` — an SBBF false-positive on an absent in-range key no longer caches a spurious entry pointing at another key's postings. |
| 3 | HIGH | `CoveringPageFrame` reports its key column as `COVERED`; the worker decode publishes a non-covered (null-pad) column as NULL instead of mis-binding it to the 4-byte symbol-key buffer. |
| 4 | MEDIUM | cheap-path chunk accumulator widened to `long` (no silent row loss past 2³¹, most reachable via the IS NULL prefix). |
| 5 | LOW | covered-source probe folded into the existing address-cache loop (no second per-column pass for non-covered / parquet frames). |
| 6 | LOW | >2GB BINARY/STRING guarded before the int size cast (was an OOB risk). |
| 7 | NIT | covered ARRAY copies real element data via a shared `ArrayTypeDriver.appendArrayData` (strided for non-vanilla arrays), never zero-fills; the eager and worker paths now agree. |

### Tests
- **RED→GREEN** regression for the #2 cache-poisoning bug (absent in-range key + SBBF false-positive — confirmed it fails with the bug reintroduced, passes with the fix).
- Single-sparse-gen warm guard + a **6-thread × 300-iteration** concurrent detached-decode guard (production warm + freeze) for #1.
- **All-types** worker-decode equivalence vs a non-indexed twin: LONG/INT/SHORT/BYTE/BOOLEAN/FLOAT/DATE/UUID/LONG256/IPv4/STRING/VARCHAR/BINARY/ARRAY + per-type nulls, through both the record fast-path and `first`/`last`.
- `sym IS NULL` null-prefix equivalence; the gated perf verdict is now **asserted**, not just printed.
- Updated the white-box source/metadata tests for the corrected `COVERED` semantics.

Full covering + posting suite green (434 tests, incl. the `CoveringIndexTest` equivalence oracle); non-covered async-pipeline smoke green.

<sub>🤖 Fixes implemented with Claude Code.</sub>

nwoolmer:
### Second-round level-3 review + master merge

Merged latest `master` and ran a deep review of the parallel covered-index decode path. Addressed the findings below (commits `cbce8a2023`, `c1514e5487`, `aa71fb8c47`).

**Concurrency / correctness**
- **`writeArray` >2GB guard** — `CoveredColumnDecoder.writeArray` cast the framed size to `int` with no overflow guard (its `writeString`/`writeBinary` siblings already had one). A >2GB covered ARRAY would truncate and write past the buffer. Added the `long`-computed guard; mirrored into the eager `write*ToFrame` twins.
- **`openRequiredSidecars` frozen no-op** — it zeroed/rewrote the reader-level `coveredAvailable[]` on every `getDetachedCursor`, i.e. from N workers concurrently. Now a no-op while frozen (warm already populated it), with an `-ea` invariant check.
- **`UnorderedPageFrameSequence` freeze** — the async (non-vectorized) keyed GROUP BY / TopK / HorizonJoin path did not `freezeCoveredReaders()` before dispatch, unlike the base `PageFrameSequence`. Wired freeze/unfreeze in to match.

**Memory accounting**
- Covered decode buffers now charge the per-query `MemoryTracker` (each `CoveringBuffers` captures it at decode, like `RowGroupBuffers` captures its allocator) and are tracked via a `coveredCachedBytes` counter. Documented that they're query-lifetime (can't LRU-evict pinned `first()`/`last()` pointers) so the tracker is the enforced ceiling.

**Lower-severity** — guarded the `long`→`int` frame-size cast and cumulative var-data overflow; added the OOM `releaseResources` guard to `getDetachedCursor` (fwd+bwd); `count == rowCount` assert after the detached re-decode; per-gen `<= Integer.MAX_VALUE` asserts in `selectKthMatch`/`countMatchesClamped`; doc/comment fixes.

**On the freeze gap (F1) — severity note**
I confirmed the async keyed GROUP BY over a covering index *does* route covered frames through `UnorderedPageFrameSequence` (new `CoveringIndexParallelGroupByReachabilityTest`; a throwaway instrumented dispatch counter observed it directly). However, a negative control — removing the freeze and stressing the path with a concurrent O3 writer + 4 workers (aggressive variant, 400 iters × 3 runs) — did **not** reproduce corruption: the eager production warm makes worker decode replay read-only and the query reader is snapshot-stable with `valueMem` pre-extended. So the missing freeze is a **latent/defensive gap, not a live use-after-free**. The freeze is kept as correct hardening (consistency with the base path, keeps the frozen-only `openRequiredSidecars` no-op effective, future-proofs).

**Tests** — covered-byte accounting (tracked then released to zero); F1 reachability (async-groupby-over-covering plan shape) + a concurrent-writer regression. Full covering suite green with `-ea` (`CoveringIndexParallelDecodeTest` 22, `CoveringIndexTest` 398, `PostingReaderConcurrentReadTest` 5, `PostingReaderSelectKthMatchTest` 12, `ParallelGroupByMemoryTrackerTest` 7).

> The >2GB ARRAY/STRING/BINARY guards are not unit-tested (a real multi-GB value is impractical in CI); they mirror the prior round's reviewed string/binary guards.

nwoolmer:
### Follow-up review round — LOW/NIT fixes + an M1 correction

Two more deep-review passes since the last update; pushed `e8f3119eb2` and `ee119a5000`.

**LOW — `selectKthMatch`/`countMatchesClamped` fully-below skip.** A generation whose postings for the key are entirely below `minValue` contributes nothing (the cursor filters them all), but the metadata path tripped the `exactMin < minValue` MIXED check and forced the whole (key, partition) onto the O(rows) traverse. Added the symmetric `exactMax < minValue -> continue` to both methods (the counterpart of the existing fully-above-clamp skip); a straddling gen still bails. Verified with white-box tests using a real cursor as the oracle, plus a negative control (fix removed -> the new test fails `expected:<50> but was:<sentinel>`).

**NITs.**
- `count == rowCount` promoted from an `-ea` assert to a hard `CairoException`, so a release build rejects an under-decode instead of aggregating uninitialized buffer tail.
- Extracted the byte-identical fixed-width covered switch into `CoveredColumnDecoder.writeFixedWidthCovered`, called by both the worker decode and the eager merge — the fixed-width layout can no longer drift between the two paths.
- The backward reader's `getDetachedCursor` (symmetry-only, unused by the forward-only pipeline) now has concurrent-path coverage via `testTwoDetachedBwdCursorsSameKeyConcurrent`.

**Correction — reverted the per-query MemoryTracker charging for covered buffers (the earlier "M1").** This turned out to be unsafe and I had previously mis-assessed it as fine. Covered decode buffers are query-lifetime and freed only at the *next* query's `clear()`/`of()` (the reduce task is reset on reuse, not at the owning query's teardown) — i.e. after the owning query's tracker has been recycled. Charging that tracker then decremented a recycled block and tripped `PerQueryMemoryTracker.acquire()`'s `used == 0` guard (surfaced as `tracker recycled with used=15000` once the `selectKthMatch` change routed more frames through worker decode). Covered buffers now use global `NATIVE_INDEX_READER` accounting, matching the eager covering path; the `coveredCachedBytes` observability counter stays. Net effect: per-query memory limits do not account covered decode memory — a documented limitation shared with the eager path, rather than a buggy charge. `testCoveredAsyncGroupByTrackerBalancedAcrossManyQueries` guards it.

All green with `-ea`: CoveringIndexParallelDecode 22, CoveringIndex 398, PostingReaderSelectKthMatch 14, PostingReaderConcurrentRead 6, ParallelGroupByReachability 3, ParallelGroupByMemoryTracker 7.

nwoolmer:
## Level-3 review — parallel covered-index decode

Reviewed at `1fe6178f82` (merged up to `master`). This is a second deep pass, run as four independent adversarial tracks — JMM/concurrency, query-path edge cases, test-coverage adequacy, and code-quality + validation of the latest fixes — each instructed to *break* the code and to verify against the actual implementation rather than comments/javadoc.

**Verdict: no new correctness bugs found.** The concurrency model holds up to JMM-level scrutiny, the high-risk SQL edge cases are correctly guarded, and the three fixes from the previous round all validate. Remaining items are test-coverage gaps and one design/maintainability note — all non-blocking.

### Fixes from the prior round — validated ✅
- **`LONG128` covered decode** (added to the shared `writeFixedWidthCovered` 16-byte case): PASS and *complete*. Enumerated the covered reader's decodable type set (`AbstractPostingIndexReader` ~`:3014-3030`) and diffed it against the decoder's switch — the set difference is now empty. `LONG128`'s `getCoveredLong128Lo/Hi` are byte-identical to UUID. `TIMESTAMP_NANO`/`INTERVAL` are not covered-storable (`CoveringCompressor` rejects them), so they aren't a gap.
- **Loud `default` throws** (worker + eager paths): PASS — unreachable for any legitimately-handled type today, correct as a forward-compat backstop, actionable messages incl. the tag.
- **`symBytes` int-cast guard**: PASS — and no second unguarded narrowing in the covered allocator (`colCap` is `long`; `initDataCap` is guarded by the symmetric `rowCount*32` assert; var-data >2GB throws `CairoException`).

### Concurrency / JMM — verified sound
The plain (non-`volatile`) `frozen` field is safe because the happens-before chain is real and correctly typed: freeze is program-order-before the reduce-queue publish on the dispatch thread (`reducePubSeq.done()` → `Unsafe.putOrderedInt` release-store), and the worker reads it after its `Sequence` dequeue (`Unsafe.getIntVolatile` acquire-load on the same flag slot). The **keystone** is that detached worker cursors are structurally read-only against shared reader state — `advanceToNextRelevantGen()` gates the only shared mutation (`putCacheEntries`) on `!isDetached`, so warm-completeness is belt-and-suspenders rather than load-bearing. Inline/work-stealing reduce on the dispatch thread is per-pool CAS-exclusive (`PerWorkerLocks.acquireSlot`), pooled `PageFrameReduceTask`/buffers are cleared via the `sameQueryExecution` check + `bindGeneration`, and unfreeze is gated behind an `AtomicInteger`/latch drain that even a throwing worker advances (`finally subSeq.done(...)`). Could not construct a failing interleaving.

### Query-path edge cases — verified sound
The highest-risk trap — a `count()` metadata fast-path over-counting when a **residual filter** is present — does not fire: the metadata `size()` lives only on `SingleKeyCoveringCursor`, and any predicate beyond the single indexed-symbol equality wraps the factory in `Filtered`/`AsyncFilteredRecordCursorFactory`, whose `size()` returns `-1`, forcing `CountRecordCursorFactory` onto the iterating `calculateSize`. Interval predicates only bound `[rowLo,rowHi)` (counted correctly); `count(col)` is excluded from the fast path at codegen (`paramCount==0`). Also confirmed correct: descending/negative-LIMIT stays on the traverse (never the forward `selectKthMatch`), multi-key routing can't be mis-classified (decided by `nKeyValues`, deduped before the k-way merge), `IS NULL`/columnTop null-prefix is handled with the null bound left *unclamped* vs `entryMaxValue`, empty/zero-row/absent-key returns clean zeros, and LATEST ON disables the page-frame path entirely. The `countMatchesClamped`/`selectKthMatch` pair share byte-identical gen-walk predicates, so count and k-th-match cannot desync.

### Findings (non-blocking)

**1. (med) `count()` metadata fast path is only smoke-tested on a fresh partition.** The white-box `selectKthMatch` tests cover resealed / multi-gen / null-prefix / MIXED-bail, but the **end-to-end SQL `count()`** path is asserted only against one freshly-loaded table. `size()` uses its own slack-bound logic that *false-bails* where `countMatchesClamped` doesn't (the documented divergence at `AbstractPostingIndexReader` ~`:262`), so a resealed-partition or multi-gen `count() WHERE sym=k` that diverged from the true row count wouldn't be caught. Suggest: force a reseal/O3-shrink so `entryMaxValue` < data max, then assert `count() WHERE sym=k` equals the same query forced through a full scan.

**2. (med) The dispatch / `UnorderedPageFrameSequence` freeze has no red-on-break (negative-control) test.** As the PR description itself notes, removing the freeze does not fail the concurrent-writer test (eager warm makes worker decode read-only on the tested paths). So the PR's central invariant has positive coverage only. Hard to close without a fault-injection hook, and the design is safe today — but worth calling out so a future change that makes a reader advance mid-query isn't silently uncaught.

**3. (med) `first()`/`last()` over a covered **ARRAY**/BINARY column isn't exercised.** The zero-copy stable-pointer aggregation (raw pointer retained into a covered buffer for the merge phase) is the riskiest lifetime case; only VARCHAR/STRING go through it today. A `first(bin_col), last(arr_col)` over a covered query spanning more than `frameQueueCapacity` frames would catch a premature-free/dangling pointer.

**4. (med, design) Six covered-specific methods on the generic `PageFrame` interface fail *silent* if a wrapper forgets to delegate.** `getColumnSource`/`getCovered*` default to non-covered sentinels (`DIRECT` / `VALUE_NOT_FOUND`), so a future `PageFrame` wrapper placed above a covering frame that doesn't propagate them would be recorded as non-covered by `PageFrameAddressCache.add` and silently produce wrong columns — no fail-loud. Complete today (only `SelectedRecordCursorFactory` + `ExtraNullColumnCursorFactory` wrap covering frames, both delegate correctly). Worth a guarding assertion, or narrowing the surface so the obligation is explicit.

**5. (low) GEO (4 widths), DECIMAL (6 widths), CHAR, TIMESTAMP decode only single-threaded.** They go through the record path at 1 worker; the all-types parallel oracle covers LONG/INT/SHORT/BYTE/FLOAT/DATE/UUID/LONG256/IPv4/STRING/VARCHAR/BINARY/ARRAY. Same code line regardless of worker count, so low risk — but the 16/32-byte and sub-byte-null layouts are the most layout-sensitive and get zero multi-worker coverage. Cheap to add to `testParallelCoveredDecodeAllTypesMatchReference`.

**6. (low) `assert` vs `throw` asymmetry on allocation-size overflow.** `symCap`/`initDataCap` overflow is `assert`-guarded (silent if `-ea` off) while var-data overflow throws `CairoException`. Both are bounded by the page-frame row cap in practice; a `throw` on the alloc-size paths would be more defensible for the same class of overflow.

**7. (low, nit) EXPLAIN gives no signal of the decode strategy.** `toPlan` is unchanged and decode-strategy-agnostic, so there's no way to tell from a plan whether a query took the metadata-only single-key path vs the eager multi-key merge. An `attr` hint would help observability/regression-pinning.

### Acceptable omissions (noted, no action needed)
`LONG128` has no SQL surface and shares UUID's code line; DECIMAL64/32/16/8 share the LONG/INT/SHORT/BYTE lines that *are* parallel-tested; the backward-reader `selectKthMatch` is forward-only-tested because production descending always traverses; `@TestOnly coveredRowsWrittenForTesting` is `assert`-gated (zero production cost). The equivalence oracle is byte-exact (`assertSqlCursors`), so a wrong-but-plausible decoded value would not pass.

---
*Methodology: 4 parallel adversarial review tracks against `1fe6178f82`, findings cross-checked against the implementation; the three prior-round fixes re-validated with `CoveringIndexTest` 398/398 + `CoveringIndexParallelDecodeTest` 23/23.*

bluestreak01 (changes requested):
## Level-3 review

Full mission-critical pass. The PR has already been through two prior level-3 rounds that caught and fixed the serious defects (frozen-cache race, cache poisoning, symbol-key conflation, >2GB guards, the `UnorderedPageFrameSequence` freeze gap). I independently re-traced the load-bearing invariants and confirmed those are fixed and did not surface a new Critical.

### Critical
None. The invariants hold:

- **valueMem is safe for the new metadata reads.** `countMatchesClamped` / `selectKthMatch` / `select*KthValue` read gen data directly via `valueMem.addressOf(offset)` without the `valueMem.extend(...)` the cursor path used. Safe because `mapValueMem`/`reloadConditionally` always map valueMem to the full `valueMemSize` (covers every gen), and `TableReader.getIndexReader()` always reloads/`of()`s before these reads. No OOB.
- **Per-worker pool isolation.** `CoveringBuffers`/`coveringByFrame` (a non-thread-safe `IntObjHashMap`) is per-`PageFrameReduceTask`/per-slot pool; one worker owns a task+pool at a time. No cross-worker race.
- **Null-prefix equivalence.** For `sym IS NULL` (index key 0), the metadata (null-prefix + gen walk), the detached `NullCursor` (`hasNext()` emits nulls then `super.hasNext()` walks key-0 postings), and the eager path all agree; the detached-cursor `nullCount = min(columnTop, maxValue+1)` matches `getCursor` byte-for-byte. Any divergence trips the hard `CairoException` row-count mismatch in `CoveringBuffers.decode` — fail-loud, never silent-garbage. Correct design.
- **Detached cursors never write shared state** (`!isDetached` guard at both `advanceToNext/PrevRelevantGen`; `putCacheEntries` asserts `!frozen`); the cache is warmed before freeze on both the cheap (`populateCacheForKey`) and traverse (natural-exhaustion) paths.

### Moderate

**M1 — Covered-decode native memory is not capped by the per-query `MemoryTracker`, and lingers on shared reduce-task pools across queries.**
`CoveringBuffers` allocates with `MemoryTag.NATIVE_INDEX_READER` and a `null` tracker. Per the tracker checklist these are data-scaled, structurally unbounded allocations (grow with covered frame count x rows x columns) — normally a "wire it" case. The PR deliberately leaves them on global accounting and documents why: they're freed lazily at the next query's `of()`/`clear()`, after the owning query's tracker is recycled, so charging them would decrement an unrelated workload's tracker and trip the `used==0` guard. The reasoning is sound and the eager covering path has the identical behavior (not a regression). Worth surfacing honestly:
- The per-query memory ceiling does not protect a covering query. A low-selectivity `sym = '<lit>'` over many partitions can allocate large uncapped `NATIVE_INDEX_READER` memory; only global RSS accounting applies.
- The "freed at next query" lifetime means a completed query's covered buffers stay resident on the shared reduce-task pools until those tasks are reused. After one big covering query on an otherwise-idle system, that memory lingers.

Not a bug, but it deserves a one-line note in the user-facing docs that covered-index queries are outside the per-query memory limit, and ideally a follow-up to bound the cross-query retention.

**M2 — Freeze/unfreeze asymmetry + no `frozen` reset in `of()`/`reopen()` (defensive gap).**
The vect group-by path unfreezes in a `finally` (robust). The async filter/group-by path instead unfreezes in `PageFrameSequence.reset()` / `UnorderedPageFrameSequence.reset()`, which `AsyncFilteredRecordCursor.close()` calls after `collectCursor(true)` and `await()` and outside the `finally`. In normal error handling the worker error surfaces during iteration (`fetchNextFrame` -> `task.buildError()`), and `collectCursor`/`await` are bookkeeping/wait ops that don't normally throw, so `reset()` is reliably reached — hence Moderate, not Critical. But `AbstractPostingIndexReader.of()` and `PostingGenLookup.reopen()` do not reset `frozen` (only `setFrozen(false)` clears it). So if `reset()` is ever skipped, the reader stays frozen permanently: `reloadConditionally()` becomes a no-op and a reused reader silently misses writer republishes -> stale reads on a later query. Cheap self-healing fix: set `frozen = false` in `of()` (and `reopen()` / `genLookup.reopen()`). It removes the reliance on `reset()` being outside a `finally`.

### Minor
- **`System.out.println` in test code** (`CoveringIndexParallelDecodeTest.java:203, 2298, 2305`). All in the `-Dcovering.perf`-gated harness/report helpers and paired with `LOG.advisory()`, so acceptable, but redundant with the log. Prefer log-only.
- **Assert with side effect:** `assert (coveredRowsWrittenForTesting++) >= 0;` in `writeCoveredRow`. The increment is skipped when `-ea` is off. Test-only and single-threaded so harmless, but folding a mutation into an `assert` is a smell — make the increment unconditional (the field is already `volatile`) or gate it explicitly.
- **Two parallel copies of the covered layout** (`CoveredColumnDecoder.write*` for workers vs `CoveringIndexRecordCursorFactory.write*ToFrame` for the eager merge). The shared `writeFixedWidthCovered` and `ArrayTypeDriver.appendArrayData` de-dup the fixed-width and array paths, but the var-size (VARCHAR/STRING/BINARY) sink logic is still duplicated. Documented as a "drift risk"; consider extracting the var-size writers behind `VarDataSink` too.
- **Diff noise:** the PR diff includes `.pi/skills/review-pr/SKILL.md` and `.claude/skills/review-pr/SKILL.md`. These are a merge-base artifact (branch behind `master`, which updated the skill), not authored changes — they vanish on rebase/merge. No action needed.

### Downgraded (investigated, not findings)
- "Metadata reads bypass `valueMem.extend` -> OOB" — false: valueMem is always mapped to full `valueMemSize`.
- "coveringByFrame races across workers" — false: pools are per reduce slot, single-worker.
- "Leaked freeze causes UAF in current query" — false: eager warm makes worker decode read-only and the query reader is snapshot-stable; residual risk is only stale reads on a later query (M2).
- "Non-covered/null-pad column mis-bound to the 4-byte symbol buffer" — fixed in this PR: `publishAddresses` gates on `coveredColumn[q]` and publishes non-covered columns as NULL.
- "`TestUtils.assertSql` banned-helper usage in tests" — false: all matches are `TestUtils.assertSqlCursors` (twin-diff equivalence oracle); result assertions use `assertQuery(...).returns(...)`.

### Summary
No Critical or High defects survived an independent trace; the two prior level-3 rounds did their job, and the fail-loud-on-divergence design (hard `CairoException` on any metadata/cursor mismatch) is the right posture. Requesting changes for the two Moderate items: (M1) document that covered-index queries sit outside the per-query memory limit and consider bounding cross-query retention; (M2) reset `frozen = false` in `of()` for a self-healing guarantee. Everything else is Minor.

Tradeoffs are documented honestly in the PR body: filter-projection regresses ~0.9x on the parallelism view, aggregations lose to a full scan above ~5% selectivity, and the cold-cache section retracts an earlier "neutral on NVMe" claim. The planner does not yet cost the crossover — deferred, as stated.

<sub>Reviewed at level 3 (inline multi-lens pass; per-finding source verification).</sub>

bluestreak01 (dismissed):
Approving after a level-3 review (all 19 production files, 6 test files, full cross-context pass, per-finding source verification).

Verified the highest-risk areas against source with no correctness, concurrency, or resource defects surviving scrutiny:
- Concurrency model (freeze + warm + detached per-worker cursors, happens-before via the dispatch sequence) is sound; freeze/unfreeze is wired on all reachable parallel consumers (async filter, unordered group-by, vect group-by) and self-heals in reader.of().
- Metadata primitives (countMatchesClamped / selectKthMatch) are gen-skip identical, null-prefix consistent, and guarded by a hard-error re-decode row-count check.
- Memory-tracker reconcile balances (charge at alloc, reverse at teardown, global-only lazy free); breach check fires before native realloc.
- Resource cleanup covered on error paths (prep + parked cursor lifecycle, CoveringBuffers, freeze reset ordering).

Dropped 6 draft findings as false positives (unguarded cast is provably safe, IntObjHashMap frame-0 freed, no missed freeze consumer, tracker accounting balances, no silent over-fill truncation).

Non-blocking nits to fold in when convenient:
- Remove the '// -----' banner section heading in AbstractPostingIndexReader.java (CLAUDE.md bans banner comments).
- Re-sort the new public methods (countMatchesClamped, populateCacheForKey, selectKthMatch, warmForKeys) into alphabetical position.
- Add SQL / Performance / New feature labels.
- The .claude/.pi SKILL.md edits are unrelated to covered-index decode and would be cleaner as a separate change.

bluestreak01 (approved):
## Level-3 review — approve

Independent level-3 pass (correctness, concurrency, performance, resource/memory, cross-context, tests, code-quality), per-finding source verification. No correctness, concurrency, or resource defect survived an independent trace — the three prior deep rounds did their job, and the fail-loud-on-divergence design (hard `CairoException` on any covered re-decode row-count mismatch) is the right posture.

### Verified clean
- **Concurrency race on genLookup cache** — fixed: `putCacheEntries` gated on `!isDetached` + asserts `!frozen`; freeze/unfreeze wired on all three parallel consumers (async filter, unordered group-by, vect group-by).
- **Cache poisoning on absent-in-range keys** — fixed: `populateCacheForKey` uses `start != end`, mirroring the cursor predicate.
- **Symbol-key conflation** — fixed: `publishAddresses` gates on `coveredColumn` and publishes non-covered columns as NULL.
- **>2GB var-size overflow (STRING/BINARY/ARRAY)** — guarded in `long` before the int cast on both decode paths.
- **Per-query memory-tracker capping (newest code, `b6b353b870`)** — sound: charge at alloc (breach throws), `reconcileCovered()` reverses the outstanding charge from `used` at tracker close, buffers freed later on global-only accounting. Malloc/reconcile deltas match; `PerQueryMemoryTracker` is the only subclass so reconcile is never skipped; null-tracker path guarded; `init()` backstop + `used==0` assert still catch real leaks.
- **valueMem OOB, of() self-heal (frozen reset), selectKthMatch/countMatchesClamped gen-walk parity, cross-context callers (count() size(), the two frame wrappers, PageFrameAddressCache.add)** — all traced clean.

### Non-blocking follow-ups
- Strip the unrelated `.claude/skills/review-pr/SKILL.md` / `.pi/skills/review-pr/SKILL.md` edits — the branch is ahead by 2 commits on these files, so they would land on master (not a stale artifact that vanishes on merge).
- Confirm enterprise CI green on current HEAD (`251b4bbfeb`); the last red run was on the older `259341e9a9`, likely repaired by the subsequent navigateCoveredRecord init fix.
- Add coverage for first()/last() over a covered ARRAY/BINARY (stable-pointer aggregation); the CoveringIndexTest ARRAY oracle was weakened to placeholder zeros.
- Minor: remove the `// -----` banner comment (AbstractPostingIndexReader ~1461, CLAUDE.md bans banners); re-sort countMatchesClamped/selectKthMatch/populateCacheForKey/warmForKeys into alphabetical position; use public getEntryMaxValue() instead of reflection in PostingReaderSelectKthMatchTest; add SQL/Performance/New feature labels.

<sub>Reviewed at level 3 (inline multi-lens pass; per-finding source verification).</sub>

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 964 / 1097 (87.88%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/IndexReader.java|0|3|00.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrame.java|2|6|33.33%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexBwdReader.java|21|44|47.73%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexFwdReader.java|21|44|47.73%|
|:large_blue_circle:|io/questdb/cairo/arr/ArrayTypeDriver.java|7|10|70.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ExtraNullColumnCursorFactory.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/std/MemoryTracker.java|13|15|86.67%|
|:large_blue_circle:|io/questdb/griffin/engine/table/CoveringIndexRecordCursorFactory.java|149|165|90.30%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|250|275|90.91%|
|:large_blue_circle:|io/questdb/cairo/idx/AbstractPostingIndexReader.java|256|279|91.76%|
|:large_blue_circle:|io/questdb/cairo/sql/CoveredColumnDecoder.java|144|152|94.74%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameAddressCache.java|75|77|97.40%|
|:large_blue_circle:|io/questdb/cairo/sql/async/PageFrameSequence.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/async/UnorderedPageFrameSequence.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/vect/GroupByRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingGenLookup.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/std/PerQueryMemoryTracker.java|2|2|100.00%|

PR #7285, opened 2026-06-18
fix(core): fix out-of-order rows after a lag commit crosses a partition boundary

### LINKED ISSUE #7278: O3 commit-with-lag + rollback across a partition boundary corrupts row order on the next commit
## Summary

When out-of-order ingestion uses a commit-with-lag (`TableWriter.ic()`) whose appended data crosses a partition boundary, the writer's `maxTimestamp` is left lower than the data actually committed on disk. After a `rollback()`, the next out-of-order (O3) commit reads that stale `maxTimestamp` as the existing-data upper bound and physically reorders rows in the affected partition. The row count stays correct and no rows are lost or duplicated, but the on-disk timestamp order is wrong — so `SELECT *` (read in storage order) returns rows out of order / wrong values at given positions.

It is deterministic and single-threaded (not a race). It is also long-standing: it reproduces at `586df7eb0a` (the parent of #7107), so it predates the recent O3/Parquet work. It was surfaced by the existing `O3MaxLagFuzzTest#testRollbackFuzzParallel` on a CI run (`O3MaxLagFuzzTest` random seeds `10956520017506302L, 1781697443410L`).

## Impact

- Silent data corruption: a partition's tail ends up out of timestamp order; ordered / `SELECT *` reads return wrong results positionally.
- `maxTimestamp` (the table high-water mark) is under-reported, which can also affect interval scans, `latest on`, and the fast `max(ts)` metadata path.
- Scope: non-WAL O3 ingestion that commits with lag, crosses a partition boundary, rolls back, then ingests more out-of-order data. Purely in-order ingestion and the WAL apply path are unaffected.

## Reproduction

Deterministic. Add this test to `O3MaxLagFuzzTest` and run it — it fails on `assertXY`. The column data is fixed because `AbstractO3Test` seeds `SharedRandom` with a constant `new Rnd()`, so only the three numeric params matter:

```java
@Test
public void testRollbackLagPartitionSwitchRegression() throws Exception {
    // 57091 rows, partition by DAY. The in-order prefix crosses the 1970-01-01 -> 1970-01-02
    // boundary, so a lag-commit + rollback leaves maxTimestamp below the committed data.
    executeWithPool(2, (engine, compiler, sqlExecutionContext, timestampTypeName) ->
            testRollbackFuzz0(engine, compiler, sqlExecutionContext, 57_091, 151, 1_545_848));
}
```

(`testRollbackFuzz0` is the existing helper in that class: it builds `x`, `z = x order by f`, an empty `y`, inserts `z limit 151` with `ic()`, rolls back, inserts the rest with `ic()`, commits, then `assertXY` compares `x` and `y`.) It also reproduces with `executeWithPool(1, ...)`, confirming a logic bug rather than a worker race.

Observed failure:

```
Row 53563 column i[INT]
  expected:<53563  ibm    ... ts=1970-01-01T22:59:58.710576Z ...>
  but was: <53577  googl  ... ts=1970-01-01T23:00:20.352448Z ...>
```

All 57,091 rows are present (no loss or duplication); ~6 rows in the 1970-01-01 partition tail are physically out of timestamp order.

## Root cause

`TableWriter.o3MoveUncommitted()` shifts the active partition's uncommitted tail into the O3 buffer and calls the one-arg `TxWriter.resetToLastPartition(committedTransientRowCount)`, which restores `maxTimestamp` from the durable `_txn` value (`TX_OFFSET_MAX_TIMESTAMP_64`). If the current commit already sealed an earlier partition via `switchPartition` (its rows are committed but not yet persisted), that durable value is below the real committed max. `processO3Block()` then sets `maxTimestamp = max(reset, o3CommittedMax)`, which cannot recover the lost value. On the next O3 commit, the too-low `maxTimestamp` is used as `dataTimestampHi` for the partition, so its prefix/merge/suffix split misplaces the rows above it.

Concretely in the reproducer: the commit fills the 1970-01-01 partition (max ts `23:53:23`), then a 1970-01-02 row seals it; `o3MoveUncommitted` empties the new active partition (`committedTransientRowCount == 0`); `maxTimestamp` resets to the durable value instead of `23:53:23`; the rollback persists/reloads the stale value; the next bulk O3 commit reorders the rows above `22:59:57`.

brunocalza:
## Summary

Fixes https://github.com/questdb/questdb/issues/7278

When a non-WAL writer ingests out of order and the in-order prefix crosses a partition boundary, the partition switch seals the earlier partition in memory but doesn't persist `_txn`. On the next lag commit, `o3MoveUncommitted` reclaims the active partition's uncommitted in-order tail into the O3 buffer and **resets `maxTimestamp` to the durable `_txn` value**. That value predates the switch, so the commit registers `maxTimestamp` from the committed O3-buffer rows only, even though the sealed partition holds a row with a higher timestamp. The high-water mark ends up below committed data: a rollback reloads it, and the next O3 commit reorders the partition tail.

## Fix

Make sure `maxTimestamp` resets to the max timestamp of the last sealed partition. We introduce a new variable to keep track of that information. Whenever a partition switch happens, we update that information. 

https://github.com/questdb/questdb/blob/870ff7ac5f16f9521eec3f46ded7f6c42e4adf05/core/src/main/java/io/questdb/cairo/TxWriter.java#L504-L506

https://github.com/questdb/questdb/blob/870ff7ac5f16f9521eec3f46ded7f6c42e4adf05/core/src/main/java/io/questdb/cairo/TxWriter.java#L839-L844

## Alternative solutions considered
- Make use of `prevMaxTimestamp`. Using it would make the regression test pass, but it would not be correct. There are cases where `prevMaxTimestamp` does not match the max timestamp of the last sealed partition.
- Whenever we reset `maxTimestamp` to the last partition, we read that value directly from the partition. It adds a disk read and is more complicated than simply having a variable

## Testing
- [`testRollbackLagPartitionSwitchRegression`](https://github.com/questdb/questdb/compare/master...brunocalza:questdb:bcalza/fix-o3-rollback?expand=1#diff-78da7596913cf1d05ef92b95e49917456fba0f6fb3d09b12270aa9de669c389eR67) that reproduces the bug is added, confirming the bug is fixed.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TxWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TxWriter.java
@@ -47,6 +47,7 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
     private int lastRecordBaseOffset = -1;
     private long lastRecordStructureVersion = -1;
     private long lastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
+    private long prevLastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
     private long prevMaxTimestamp;
     private long prevMinTimestamp;
     private long prevPartitionTableVersion = -1;
@@ -117,6 +118,7 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
         if (transientRowCount == 1 && txPartitionCount > 1) {
             // we have to undo creation of partition
             txPartitionCount--;
+            lastSealedPartitionMaxTimestamp = prevLastSealedPartitionMaxTimestamp;
             fixedRowCount -= prevTransientRowCount;
             transientRowCount = prevTransientRowCount + 1; // When row cancel finishes 1 is subtracted. Add 1 to compensate.
             attachedPartitions.setPos(attachedPartitions.size() - LONGS_PER_TX_ATTACHED_PARTITION);
@@ -157,6 +159,7 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
         lastRecordBaseOffset = -1;
         prevRecordBaseOffset = -2;
         lastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
+        prevLastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
     }
 
     @Override
@@ -311,6 +314,8 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
     public void removeAllPartitions() {
         maxTimestamp = Long.MIN_VALUE;
         minTimestamp = Long.MAX_VALUE;
+        lastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
+        prevLastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
         prevTransientRowCount = 0;
         transientRowCount = 0;
         fixedRowCount = 0;
@@ -387,6 +392,7 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
     public void resetTimestamp() {
         recordStructureVersion++;
         lastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
+        prevLastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
         prevMaxTimestamp = Long.MIN_VALUE;
         prevMinTimestamp = Long.MAX_VALUE;
         maxTimestamp = prevMaxTimestamp;
@@ -503,6 +509,7 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
 
     public void switchPartitions(long timestamp) {
         recordStructureVersion++;
+        prevLastSealedPartitionMaxTimestamp = lastSealedPartitionMaxTimestamp;
         lastSealedPartitionMaxTimestamp = maxTimestamp;
         fixedRowCount += transientRowCount;
         prevTransientRowCount = transientRowCount;
@@ -523,7 +530,6 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
     }
 
     public void truncate(long columnVersion, ObjList<? extends SymbolCountProvider> symbolCountProviders) {
-        lastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
         removeAllPartitions();
         if (!PartitionBy.isPartitioned(partitionBy)) {
             attachedPartitions.setPos(LONGS_PER_TX_ATTACHED_PARTITION);
@@ -555,6 +561,7 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
         this.prevPartitionTableVersion = partitionTableVersion;
         this.txPartitionCount = 1;
         this.lastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
+        this.prevLastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
         if (baseVersion >= 0) {
             this.readBaseOffset = getBaseOffset();
             this.readRecordSize = getRecordSize();
@@ -666,6 +673,7 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
         prevMinTimestamp = minTimestamp;
         prevMaxTimestamp = maxTimestamp;
         lastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
+        prevLastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
 
         prevRecordStructureVersion = lastRecordStructureVersion;
         lastRecordStructureVersion = recordStructureVersion;
--- a/core/src/main/java/io/questdb/cairo/TxWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TxWriter.java
@@ -46,6 +46,7 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
     private TableWriter.ExtensionListener extensionListener;
     private int lastRecordBaseOffset = -1;
     private long lastRecordStructureVersion = -1;
+    private long lastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
     private long prevMaxTimestamp;
     private long prevMinTimestamp;
     private long prevPartitionTableVersion = -1;
@@ -155,6 +156,7 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
         prevRecordStructureVersion = -2L;
         lastRecordBaseOffset = -1;
         prevRecordBaseOffset = -2;
+        lastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
     }
 
     @Override
@@ -365,6 +367,10 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
         resetLagAppliedRows();
     }
 
+    public void resetPartitionParquetFormat(long timestamp) {
+        setPartitionParquetFormat(timestamp, -1, false);
+    }
+
     public void resetPartitionParquetGenerated(int partitionIndex) {
         resetPartitionParquetGeneratedByRawIndex(partitionIndex * LONGS_PER_TX_ATTACHED_PARTITION);
     }
@@ -374,16 +380,13 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
         attachedPartitions.setQuick(indexRaw + PARTITION_PARQUET_FILE_SIZE_OFFSET, -1L);
     }
 
-    public void resetPartitionParquetFormat(long timestamp) {
-        setPartitionParquetFormat(timestamp, -1, false);
-    }
-
     public void resetStructureVersionUnsafe() {
         txMemBase.putLong(readBaseOffset + TX_OFFSET_STRUCT_VERSION_64, 0);
     }
 
     public void resetTimestamp() {
         recordStructureVersion++;
+        lastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
         prevMaxTimestamp = Long.MIN_VALUE;
         prevMinTimestamp = Long.MAX_VALUE;
         maxTimestamp = prevMaxTimestamp;
@@ -500,6 +503,7 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
 
     public void switchPartitions(long timestamp) {
         recordStructureVersion++;
+        lastSealedPartitionMaxTimestamp = maxTimestamp;
         fixedRowCount += transientRowCount;
         prevTransientRowCount = transientRowCount;
         long partitionTimestampLo = getPartitionTimestampByTimestamp(maxTimestamp);
@@ -519,6 +523,7 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
     }
 
     public void truncate(long columnVersion, ObjList<? extends SymbolCountProvider> symbolCountProviders) {
+        lastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
         removeAllPartitions();
         if (!PartitionBy.isPartitioned(partitionBy)) {
             attachedPartitions.setPos(LONGS_PER_TX_ATTACHED_PARTITION);
@@ -549,6 +554,7 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
         this.baseVersion = getVersion();
         this.prevPartitionTableVersion = partitionTableVersion;
         this.txPartitionCount = 1;
+        this.lastSealedPartitionMaxTimestamp = Long.MIN_VALUE;
         if (baseVersion >= 0) {
             this.readBaseOffset = getBaseOffset();
             this.readRecordSize = getRecordSize();
@@ -596,12 +602,12 @@ public final class TxWriter extends TxReader implements Closeable, Mutable, Symb
         return maskedSize;
     }
 
-    private static long updatePartitionHasParquetGenerated(long maskedSize, boolean parquetGenerated) {
-        return updatePartitionFlagAt(maskedSize, parquetGenerated, PARTITION_MASK_PARQUET_GENERATED_BIT_OFFSET);
-    }
-
     private static long updatePartitionHasParquetFormat(long maskedSize, boolean isParquetFormat) {
         return updatePartitionFlagAt(maskedSize, isParquetFormat, PARTITION_MASK_PARQUET_FORMAT_BIT_OFFSET);
+    }
+
+    private static long updatePartitionHasParquetGenerated(long maskedSize, boolean parquetGenerated) {
+        return updatePartitionFlagAt(maskedSize, parquetGenerated, PARTITION_MASK_PARQUET_GENERATED_BIT_OFFSET);
     }
 
     private static long updatePartitionIsReadOnly(long maskedSize, boolean isReadOnly) {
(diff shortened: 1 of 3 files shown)


jerrinot:
Hello @brunocalza,

many thanks for this PR! I had a look and it appears that even with this fix there are still issues. This test is failing on your branch:
```java
@Test
    public void testCancelAfterPartitionSwitchDoesNotPoisonLagMax() throws Exception {
        executeWithPool(0, (engine, compiler, sqlExecutionContext, timestampTypeName) -> {
            // Use a 6-hour O3 lag. On commit, rows newer than maxO3Timestamp - 6h
            // stay uncommitted in lag. The test arranges for 1970-01-02T12:00:00Z
            // to stay in lag, so the committed max timestamp must remain at
            // 1970-01-02T00:00:00Z unless the cancelled partition switch poisons
            // the writer's cached max timestamp state.
            engine.execute("CREATE TABLE x (i INT, ts " + timestampTypeName + ") TIMESTAMP(ts) PARTITION BY DAY WITH maxUncommittedRows=2, o3MaxLag=21600s", sqlExecutionContext);
            final TimestampDriver driver = timestampType.getDriver();
            final long firstPartitionO3Ts = driver.parseFloorLiteral("1970-01-01T00:00:00.000000Z");
            final long firstPartitionCommittedTs = driver.parseFloorLiteral("1970-01-01T12:00:00.000000Z");
            final long secondPartitionExpectedMaxTs = driver.parseFloorLiteral("1970-01-02T00:00:00.000000Z");
            final long secondPartitionLagTs = driver.parseFloorLiteral("1970-01-02T12:00:00.000000Z");
            final long cancelledThirdPartitionTs = driver.parseFloorLiteral("1970-01-03T00:00:00.000000Z");

            try (TableWriter writer = TestUtils.getWriter(engine, "x")) {
                // Build an in-order transaction that spans two partitions. The first row
                // is later than the future O3 row, so O3 lag remains positive even after
                // the cancelled partition switch is correctly rolled back. The last row in
                // the second partition is intentionally inside that future lag window.
                Row row = writer.newRow(firstPartitionCommittedTs);
                row.putInt(0, 1);
                row.append();

                row = writer.newRow(secondPartitionExpectedMaxTs);
                row.putInt(0, 2);
                row.append();

                row = writer.newRow(secondPartitionLagTs);
                row.putInt(0, 3);
                row.append();

                // Open the third partition and then cancel the row. After this cancellation
                // the writer should behave as if the third partition switch never happened,
                // including any in-memory max timestamp state used by later O3 lag commits.
                row = writer.newRow(cancelledThirdPartitionTs);
                row.putInt(0, 4);
                row.cancel();

                // Insert an out-of-order row into the first partition. This forces ic() to
                // move the uncommitted second-partition rows into the O3 path, which is where
                // the cancelled partition switch can leak a stale sealed-partition max.
                row = writer.newRow(firstPartitionO3Ts);
                row.putInt(0, 5);
                row.append();

                // maxO3Timestamp is secondPartitionLagTs, so that row stays in lag while
                // rows up to secondPartitionExpectedMaxTs are committed. The committed
                // table max after ic() must therefore be secondPartitionExpectedMaxTs,
                // not the lag row.
                writer.ic();

                Assert.assertEquals(secondPartitionExpectedMaxTs, writer.getMaxTimestamp());
            }
        });
    }
```
Just paste it to `O3MaxLagTest` Could you please have another look?

bluestreak01 (changes requested):
@brunocalza — thanks for the fix. Adversarial review below (level 3). The core change correctly resolves #7278, but it reintroduces the same failure class in the opposite direction on an out-of-diff path (`rowCancel`) that the fix doesn't reset for. **Requesting changes** on that basis.

The fix itself is sound for its target: `switchPartitions()` records the sealed partition's max into `lastSealedPartitionMaxTimestamp`; `resetToLastPartition(long)` floors the reloaded `maxTimestamp` with `Math.max(durableMax, lastSealedPartitionMaxTimestamp)`. I verified that when a real (non-undone) switch happens the active partition is brand-new so `committedTransientRowCount == 0` and `lastSealed` is exactly the correct remaining max; and when no switch happens `lastSealed == MIN_VALUE`, so `Math.max` collapses to the pre-existing durable value (byte-for-byte old behavior). It is O(1), allocation-free, adds no new mmap read (the `getLong(TX_OFFSET_MAX_TIMESTAMP_64)` call pre-existed), and does not touch the WAL path.

---

## Critical

### C1 — `rowCancel()` undoing a partition switch leaves `lastSealedPartitionMaxTimestamp` stale → persists `maxTimestamp` ABOVE committed data (inverse of #7278)

**Files:** `TxWriter.java:506` (set), `TxWriter.java:113-135` `cancelRow()` (missing reset), `TableWriter.java:14201-14235` `rowCancel()` undo branch, consumed at `TxWriter.java:839-843`, folded at `TableWriter.java:9503`.

**Status:** CONFIRMED code-path divergence (verified by reading source). End-to-end corruption not independently reproduced — caveat below.

**Trace (every link code-verified):**

1. P0 has durably-committed rows (max `T0`). A new transaction appends in-order rows to P0 up to `T2` (durable still `T0`).
2. A row crosses into P1 → `switchPartition` → `switchPartitions()` sets `lastSealedPartitionMaxTimestamp = maxTimestamp = T2` (`TxWriter.java:506`). Then `updateMaxTimestamp` sets `maxTimestamp = T_boundary`, `prevMaxTimestamp = T2`.
3. That first P1 row is cancelled (`row.cancel()` → `rowCancel()`, `dirtyTransientRowCount == 1` partition-undo branch). It re-opens P0 with `maxTimestamp = prevMaxTimestamp = T2` and calls `txWriter.cancelRow()`. **Neither `rowCancel()` nor `TxWriter.cancelRow()` clears `lastSealedPartitionMaxTimestamp`, which stays `T2`** — yet P0 is the active partition again, so nothing is actually sealed.
4. More in-order rows go to P0 (`T3, T4 > T2`, still inside P0's day). `lastSealed` stays `T2`.
5. An out-of-order row flips to O3 mode; `ic(o3MaxLag)` → `o3Commit` → `o3MoveUncommitted` (`TableWriter.java:8557`). P0 has durable rows so `committedTransientRowCount > 0`; the uncommitted `T1..T4` move into the O3 buffer, and `resetToLastPartition(committed, Math.max(durable=T0, lastSealed=T2)) = T2` sets `maxTimestamp = T2` — **but the only rows physically remaining in P0 are the durable ones, max `T0`.**
6. `TableWriter.java:9503` folds `updateMaxTimestamp(Math.max(getMaxTimestamp()=T2, o3TimestampMax_committed))`. When `o3MaxLag` holds the `T2..T4` rows in lag (trivially satisfied since `T4 - T2` is small), the committed O3 max is `< T2`, so the persisted `maxTimestamp` stays `T2` — **above the max row actually on disk.**

The pre-fix code seeded with durable `T0` here and let line 9503 raise it to the true committed max — i.e. **the old code was correct in this case; the fix breaks it.** The over-reported value is silent (the only nearby assert, `TableWriter.java:9644`, is WAL-only and would pass anyway since `T2`/`T4` share a partition), self-corrects on the next clean commit, but is persisted to `_txn` — so a `rollback()` reloads the too-high value, the exact rollback-after-lag-commit condition this PR exists to fix, now failing in the opposite direction. `removeAllPartitions()` (called from `rowCancel`'s no-partitions branch and `reset()`) has the same gap.

**Reachability:** `row.cancel()` is live non-WAL production code — `InsertRowImpl.java:91` (SQL INSERT), `CairoTextWriter.java`/`CopyImportTask.java` (CSV import), `LineTcpMeasurementEvent.java:181,307` (non-WAL ILP). Cancelling a row that happens to be the first of a new partition, then continuing the transaction with a lag commit, reaches this path. Narrow but not impossible.

**Fix:** reset `lastSealedPartitionMaxTimestamp = Long.MIN_VALUE` in `TxWriter.cancelRow()` (at minimum the `txPartitionCount > 1` undo branch that reverses the seal), and defensively in `removeAllPartitions()`. Add a regression test: durably commit to P0, cross the boundary with one row, cancel it, append more in-order, then `ic(lag)` + `rollback` + re-ingest and assert `max(ts)` and full contents.

**Caveat:** I confirmed the invariant violation (`maxTimestamp` persisted above the on-disk max) by reading every link, but did not build a running reproduction. Please confirm with the test above, or prove the boundary-row-cancel + lag path is unreachable, before dismissing. Given it is a silent reintroduction of the #7278 corruption class and the remedy is one line, I'm treating it as blocking.

---

## Moderate

**M1 — The new test does not assert the invariant the fix targets.** `testRollbackLagPartitionSwitchRegression` validates only row contents (`assertXY`) and row count. It never asserts the `maxTimestamp` high-water-mark that #7278 explicitly calls out — no `max(ts)` query, no `LATEST ON`, no `assertMaxTimestamp(...)` (the helper exists at `AbstractO3Test.java:188/195`). In this bug the under-reported max also reorders rows so `assertXY` catches it transitively, but a future regression corrupting only the metadata high-water-mark would pass silently. Add a direct `max(ts)` / `assertMaxTimestamp` assertion after the final commit.

**M2 — Coverage gaps on the field lifecycle.** Not covered: the cancel-after-switch path (C1); multiple partition switches in one `ic()` (the data spans only ~1.02 days = exactly one boundary); a `switch → truncate` sequence (the reset at `TxWriter.java:526`); the mixed case where a real switch occurred but `durableMax > lastSealed`.

---

## Minor

- **m1 — Three unused imports** in `O3MaxLagFuzzTest.java:29,32,33` (`io.questdb.cairo.TableReader`, `io.questdb.cairo.sql.Record`, `io.questdb.test.cairo.TestTableReaderRecordCursor`) — dead code from an earlier reader-based draft. Line 33 also breaks import order (wedged between two `io.questdb.cairo.sql.*` imports). No checkstyle in the build so CI won't fail, but please delete them.
- **m2 — PR has no labels.** Add `Bug`, `Core`, `storage` (matching #7278). Correctly not `WAL` or `Performance`.
- **m3 — PR body omits two tradeoffs:** the CI cost of the deterministic 57,091-row regression test, and the unrelated alphabetical method-reordering churn bundled into the diff (`resetPartitionParquetFormat` moved up; `updatePartitionHasParquetFormat`/`Generated` swapped — both legitimate ordering corrections, verified as byte-identical pure moves).
- **m4 — Comment nit:** `O3MaxLagFuzzTest.java:67` writes "57091 rows" without the underscore separator the code uses (`57_091`).

---

## Cleared after verification (false positives)

- **Concurrency:** no `volatile`/fence needed — `TxWriter` is single-writer-owned, the field is `private` and read strictly after the `o3DoneLatch` fork-join barrier; workers never touch it.
- **Performance:** allocation-free, O(1), no new mmap read/syscall; caching in a field is the cheapest correct option (the sealed partition's max-data ts is not recoverable from `attachedPartitions`, which stores partition start, not max).
- **Resource management:** primitive `long`, nothing to close; the test closes its writer on all paths and honors `assertMemoryLeak`.
- **Cross-context o3 downstream:** the non-cancel `o3MoveUncommitted` → `O3PartitionJob` merge split consumes a byte-identical `dataTimestampHi` (when `committedTransientRowCount > 0` in the legitimate flow, `lastSealed == MIN_VALUE`); a higher value only routes more data through the safe general merge path.
- **Member ordering / naming:** new field is alphabetically placed; the two method relocations restore correct ordering and are pure moves.

---

### Summary
- **Verdict:** request changes — the fix correctly resolves #7278 but reintroduces the same corruption class (persisting `maxTimestamp` too *high*, silently, surviving rollback) on the `rowCancel`-undoes-a-switch path it doesn't reset for.
- 6 findings verified (1 Critical, 2 Moderate, 3 Minor); 8 draft findings cleared as false positives after source verification.
- The blocking finding (C1) is **out-of-diff** (`rowCancel`/`removeAllPartitions` interacting with the new field's contract).

<sub>🤖 Adversarial level-3 review.</sub>

brunocalza:
Thanks @jerrinot and @bluestreak01 for reviewing it. Nice catch! Looks like, by introducing this new variable, when a partition is cancelled we're not rolling back the value of `lastSealedPartitionMaxTimestamp`. The solution is to introduce a new variable to keep track of the previous last sealed partition max timestamp, so when we cancel the row we can rollback to the previous value, similar to `maxTimestamp` and `prevMaxTimestamp`.  I'll explore this path and push another commit. 

Aside: Adding new variables (we're adding 2) makes be hesitate a bit because the number of states grows considerably, and we might actually be introducing new bugs, but tbh I'm not seeing alternatives. Happy to discuss any other approaches if you have ideas.

brunocalza:
I pushed another commit with the fix I proposed: having another variable ([`prevLastSealedPartitionMaxTimestamp`](https://github.com/questdb/questdb/pull/7285/changes/d2ca61d1c850eb4b278e6497c8b492a46e64c9f1#diff-9d8c8429f72093a8e4ebdb5177035187f875a194722b731014357c76b1199d5eR50)) to keep track of the previous value, and [resetting `lastSealedPartitionMaxTimestamp` on `cancelRow()`](https://github.com/questdb/questdb/pull/7285/changes/d2ca61d1c850eb4b278e6497c8b492a46e64c9f1#diff-9d8c8429f72093a8e4ebdb5177035187f875a194722b731014357c76b1199d5eR121). 

A new test [`testCancelPartitionSwitchMaxTimestamp`](https://github.com/questdb/questdb/pull/7285/changes/d2ca61d1c850eb4b278e6497c8b492a46e64c9f1#diff-6d3b05a80271bfec09dbc0803749871be083840eb30af84a96844c78e4b887bdR101) is added with two sub tests:
- The first is a rewrite of @jerrinot 's test, asserting that the txn max timestamp comes from the o3 buffer
- The second one asserts that the txn max timestamp comes from the previous sealed partition. This test is important because it verifies that resetting `lastSealedPartitionMaxTimestamp = Long.MIN_VALUE` in `TxWriter.cancelRow()` is not enough. We must keep track of the previous. 

Summary of the points raised by @bluestreak01 
- C1 is fixed
- M1 is addressed with testCancelPartitionSwitchMaxTimestamp
- M2 is addressed with testCancelPartitionSwitchMaxTimestamp
- m1 and m4 done
- m2, I have no permissions
- m3 is irrelevant imo

bluestreak01 (approved):
Approving after a level-3 adversarial review.

The fix correctly resolves #7278, and the follow-up commit closes the inverse-direction regression (C1) without reopening it. Verified by reading every set/reset/use site of the two new fields across partition switch, row cancel, multi-switch, lag commit, rollback, truncate, and the WAL paths, plus an empirical build + test pass.

Key points (all source-verified):
- `switchPartitions()` pushes one history level; `resetToLastPartition(long)` floors with `Math.max(durableMax, lastSealed)`. With no switch, `lastSealed == MIN_VALUE` so behavior is byte-for-byte the old path.
- Single-level history is sufficient: a second back-to-back partition-undo cancel is unreachable (after an undo `transientRowCount >= 2`, so the `== 1` guard can't fire without an intervening switch that refreshes `prevLastSealed`).
- Fast-path `commit()` not resetting the fields is safe (monotonic `recordStructureVersion` forces the slow path whenever `lastSealed != MIN_VALUE`).
- `Math.max` cannot over-report (both operands are committed timestamps; the O3 fold re-raises the active-partition max).
- Concurrency: writer-thread-local, read before the O3 fork / after the join barrier.

Testing:
- `O3MaxLagTest#testCancelPartitionSwitchMaxTimestamp` and `O3MaxLagFuzzTest#testRollbackLagPartitionSwitchRegression` pass.
- Full `O3MaxLagTest` (34 tests) passes — no regression from the adaptive-lag shift.
- Revert experiment: disabling the fix makes both new tests fail, confirming they are genuine regression guards.

Non-blocking nits: a few untested-but-proven-safe lifecycle paths (multi-switch+cancel, switch->truncate, real switch where durableMax>lastSealed, cancel->ic->rollback->re-ingest) would be nice follow-up assertions; PR could use `Bug`/`Core`/`storage` labels; the bundled alphabetical method reordering is unrelated churn (verified byte-identical).

Issue #1077, opened 2021-06-01
Hasura support or compatibility


rrjanbiah:
**Is your feature request related to a problem? Please describe.**

Long time ago, I opened https://github.com/hasura/graphql-engine/issues/4307 in Hasura tracker. Basically, was looking for Hasura + high performant PostgreSQL variant. Also, posted similar issues in others:

- https://github.com/MaterializeInc/materialize/issues/3496
- https://github.com/greenplum-db/gpdb/issues/10909
- https://github.com/hasura/graphql-engine/issues/678#issuecomment-754649678

Now, having stumbled upon QuestDB through HN, thought of opening the same ticket. Looks like it is already possible in Quest and in that case, some documentation would attract the community.

PR #7286, opened 2026-06-18
ci(build): route Maven builds through the self-hosted Reposilite cache

sklarsa:
Ports the self-hosted Maven cache integration from `questdb-enterprise` (PR #1077, green) to OSS CI.

## What this does
A probe step selects the private Reposilite cache (`http://maven.internal:8081`) when reachable on the Hetzner private network, otherwise falls back to Maven Central. Self-hosted `hetzner-incus*` pools route through the cache; MS-hosted mac/windows agents fail the probe and behave exactly as before (graceful fallback, no behavior change).

Mechanism (same as enterprise): the probe sets two job-scoped vars:
- `MAVEN_SETTINGS_FILE` — hetzner mirror settings on a cache hit, empty settings on a miss
- `MAVEN_KEEPALIVE_OPT` — empty on hit (keep-alive on), `-Dhttp.keepAlive=false` on miss

Every reachable Maven task injects `-s $(MAVEN_SETTINGS_FILE)`. The HttpClient-transport tasks that carried a literal `-Dhttp.keepAlive=false` (test-pipeline CoverageReport, snapshot) swap it for `$(MAVEN_KEEPALIVE_OPT)`.

## Why keep-alive is handled differently here than in enterprise
The heavy compile/test tasks resolve via the **wagon** transport (`-Dmaven.resolver.transport=wagon`), unlike enterprise (HttpClient). Validated against the live cache on a throwaway container — full reactor resolve, 184 jars + 291 poms, TCP connections via `/proc/net/snmp`:

| settings | transport | TCP conns | served from |
|----------|-----------|-----------|-------------|
| empty | HttpClient | 5 | Maven Central |
| hetzner | HttpClient | 5 | maven.internal |
| hetzner | HttpClient + `keepAlive=false` | **950** | maven.internal |
| hetzner | **wagon + ttl=30** | **5** | maven.internal |
| hetzner | wagon + `keepAlive=false` | 5 | maven.internal |

So: wagon honors the `-s` mirror, and already pools connections via the existing `ttlSeconds=30` (keep-alive flag is a no-op there) → no keepAlive toggle on those tasks. On the HttpClient path, `keepAlive=false` is a 190× connection-storm amplifier → swapped. `mirrorOf` is `central` (the only remote repo id in the OSS poms).

## Scope
- New: `ci/maven-settings-hetzner.xml`, `ci/maven-settings.xml`
- Probe + `-s`/keepAlive edits: `ci/templates/steps.yml`, `ci/templates/compat-steps.yml`, `ci/templates/aux-job.yml`, `ci/test-pipeline.yml`, `ci/snapshot-pipeline.yml`
- `github-release-pipeline.yml` left unchanged (MS-hosted only, cannot reach the cache)

### DIFF
--- a/ci/maven-settings-hetzner.xml
+++ b/ci/maven-settings-hetzner.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<settings xmlns="http://maven.apache.org/SETTINGS/1.2.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.2.0 https://maven.apache.org/xsd/settings-1.2.0.xsd">
+
+  <mirrors>
+    <mirror>
+      <id>hetzner-reposilite</id>
+      <name>Hetzner Reposilite Maven Cache</name>
+      <url>http://maven.internal:8081/releases/</url>
+      <mirrorOf>central</mirrorOf>
+    </mirror>
+
+    <!--
+      Maven 3.8.1+ ships a built-in "maven-default-http-blocker" mirror that refuses
+      any plain-HTTP repository. Our internal cache speaks HTTP over the private Hetzner
+      network (maven.internal is private-only, not reachable publicly), so we neutralise
+      the blocker by pointing its mirrorOf at a repo id that never matches. Remove this
+      block once the cache is fronted by HTTPS.
+    -->
+    <mirror>
+      <id>maven-default-http-blocker</id>
+      <name>Disable HTTP blocker for trusted internal mirror</name>
+      <url>http://0.0.0.0/</url>
+      <mirrorOf>dummy-disabled</mirrorOf>
+    </mirror>
+  </mirrors>
+
+</settings>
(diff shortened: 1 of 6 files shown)


bluestreak01 (approved):
Approving. The Reposilite cache routing mechanism is sound: every job consuming $(MAVEN_SETTINGS_FILE) runs the probe first in the same job and sets the variable on both branches (graceful fallback), mirrorOf=central covers the full dependency+plugin graph (no non-central repos, no third-party SNAPSHOTs), and the wagon-vs-HttpClient keep-alive handling matches the actual task wiring. No correctness/blocking issues.

Minor follow-ups (non-blocking):
- PR description references ci/snapshot-pipeline.yml and a 'snapshot' task that don't exist in OSS (copied from the enterprise PR) — please trim the Scope/keep-alive sections.
- aux-job.yml's Javadoc task relies on the probe defined in the compat-steps.yml it includes; consider hoisting the probe into one shared template to remove the cross-template coupling and the 3x duplication.
- MAVEN_KEEPALIVE_OPT is set but unused in steps.yml/compat-steps.yml (all wagon transport) — only CoverageReport consumes it.
- Add the CI label.

PR #7287, opened 2026-06-19
ci(build): retry Maven wagon transfers on cache connect failure

sklarsa:
## Problem
Since master moved onto the self-hosted Reposilite Maven cache, a large share of builds fail at dependency resolution with:
```
Could not transfer artifact ... from/to hetzner-reposilite (http://maven.internal:8081/releases/):
Connect to maven.internal:8081 [10.0.0.13] failed: Connection timed out
```
The cache is healthy and near-idle (the exact "timed-out" jar serves in 0.15s). The cache is a **single IP** reached over the Hetzner vSwitch; under the fleet's cold-build connection bursts a fraction of TCP **SYNs are dropped**, and because the mirror has no mid-build failover, one dropped SYN kills the whole build.

## This change (client-side backstop)
Add `-Dmaven.wagon.http.retryHandler.count=3` to `MAVEN_RUN_OPTS` so the wagon transport **re-attempts a transfer on connect failure** instead of failing outright. The loss is intermittent, so a retry typically succeeds.

Applied to all three wagon `MAVEN_RUN_OPTS` (`test-pipeline.yml`, `test-hosted-pipeline.yml`, `test-fuzz.yml`), kept identical to avoid drift. The hosted mac/windows agents resolve from Central, so it's a harmless no-op there.

> Note: `-Dmaven.wagon.http.connectionTimeout` was deliberately **not** added — testing showed it does not bound the TCP connect in this Maven/wagon version (a blackholed connect still waited out the OS default), so it would be misleading.

## Root-cause fix (separate, already applied on the cache box)
The real fix is on the cache box itself, via the `azure-ci-runners` maven-cache role:
- **Reposilite switched to host networking** — removes the docker-proxy/iptables-DNAT hop that every connection traversed.
- **Raised `net.ipv4.tcp_max_syn_backlog` (512→8192), `net.core.netdev_max_backlog` (1000→16384), `somaxconn` (→8192)** — so SYN bursts are absorbed instead of dropped.

This PR is the belt-and-suspenders client side.

### DIFF
--- a/ci/test-fuzz.yml
+++ b/ci/test-fuzz.yml
@@ -17,7 +17,9 @@ variables:
   includeTests: "%regex[.*[^o].class]"
   MAVEN_CACHE_FOLDER: $(HOME)/.m2/repository
   MAVEN_OPTS: "-Dmaven.repo.local=$(MAVEN_CACHE_FOLDER) -Xmx3072m -XX:+UseParallelGC"
-  MAVEN_RUN_OPTS: "-Dmaven.repo.local=$(MAVEN_CACHE_FOLDER) -Dmaven.resolver.transport=wagon -Dmaven.wagon.httpconnectionManager.ttlSeconds=30"
+  # Wagon retries the transfer on connect failure so a dropped SYN to the single-IP cache
+  # re-attempts instead of failing the build. See test-pipeline.yml for the full rationale.
+  MAVEN_RUN_OPTS: "-Dmaven.repo.local=$(MAVEN_CACHE_FOLDER) -Dmaven.resolver.transport=wagon -Dmaven.wagon.httpconnectionManager.ttlSeconds=30 -Dmaven.wagon.http.retryHandler.count=3"
   MAVEN_VERSION: "version"
   MAVEN_VERSION_OPTION: "Default"
   # Full reactor (the fuzz regex spans core and compat), but no fuzz test serves the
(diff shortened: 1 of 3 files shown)


sklarsa:
```
  Summary

  - Verdict: approve. This is a low-risk, correctly-targeted CI change. -Dmaven.wagon.http.retryHandler.count=3 is a
  valid wagon-http property; I verified it lands on the real mvn command line via options: "... $(MAVEN_RUN_OPTS)"
  in ci/templates/steps.yml (and aux-job.yml/compat-steps.yml), and that -Dmaven.resolver.transport=wagon is already
  present to activate the wagon transport so the property takes effect. The retry is idempotent-safe (artifact
  GETs) and a genuine no-op on the hosted mac/windows agents that resolve from Central.
  - Tradeoff worth noting: a retry count of 3 will mask the symptom on genuinely-down cache scenarios by stretching
  each failed resolve up to ~4× the OS connect timeout before the build finally fails — so a real cache outage
  produces slower red builds, not faster feedback. Acceptable given the root cause (intermittent SYN loss) is being
  fixed on the cache box in parallel.
  - Only actionable item: add the CI label to the PR.
  - Findings: 0 critical, 0 moderate, 2 minor (1 whitespace-claim nit, 1 confirmatory note), plus 1 metadata gap
  (missing CI label). No out-of-diff findings — appropriate, since the change introduces no symbols with callsites;
  the cross-context concern here is "where is MAVEN_RUN_OPTS consumed," which I traced and confirmed.

  Want me to add the CI label to the PR?
```

bluestreak01 (approved):

PR #7288, opened 2026-06-19
fix(core): fix table corruption when a range-replace commit adds partitions above the last one

puzpuzpuz:
## What

Fixes a storage-engine corruption that could occur when a WAL `REPLACE_RANGE` commit in O3 mode appends brand-new partitions above the table's previous last partition.

## Root cause

In `TableWriter.processO3Block`, replace mode derived `partitionTimestampHi` from `o3TimestampMax`, which is the replace-range high boundary rather than the highest timestamp actually written. For an open-ended range the boundary is `Long.MAX_VALUE - 1`, and `getCurrentPartitionMaxTimestamp` overflows on it. As a result `partitionTimestampHi` stayed at the stale pre-commit ceiling, `finishO3Commit` skipped switching the writer's active column files to the new last partition, and the next commit reused the previous partition's column descriptors. That commit then read the old partition's data, producing rows dated below the new partition's floor or suspending the table.

## Fix

Derive `partitionTimestampHi` from the actual highest written timestamp (`replaceMaxTimestamp`) in replace mode. The non-replace path is unchanged. In the replace path where no rows were written (`replaceMaxTimestamp == Long.MIN_VALUE`), `partitionTimestampHi` is now left unchanged rather than being recomputed from the (possibly overflowing) range boundary; the existing no-rows / truncate replace-range tests cover this case and pass.

## Effects and tradeoffs

- Positive: range-replace commits that append partitions above the last partition no longer corrupt the table or leave it suspended; a subsequent full rebuild reads correct data.
- Scope: the change touches only the replace-mode branch of `processO3Block`. Non-replace O3 commits are unaffected.
- No measurable performance impact: the fix swaps one argument to an existing call and adds no new work on the commit path.

## Test plan

- Adds `WalWriterReplaceRangeTest.testReplaceRangeAddsPartitionsAboveLastThenRebuilds`: commit A appends partitions above the last via `REPLACE_RANGE [oldMax+1us, +inf)`, then commit B does a full rebuild via `REPLACE_RANGE [min, +inf)`. The result is compared against a plain-commit oracle table.
- Verified the new test fails without the fix (table suspends with "bulk update failed and will be rolled back") and passes with it.
- Full `WalWriterReplaceRangeTest` class passes (58 tests), including the existing no-rows and truncate replace-range cases that exercise the changed branch.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -9497,11 +9497,17 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
             } // end while(srcOoo < srcOooMax)
 
             // at this point we should know the last partition row count
-            partitionTimestampHi = Math.max(partitionTimestampHi, txWriter.getCurrentPartitionMaxTimestamp(o3TimestampMax));
-
             if (!isCommitReplaceMode()) {
+                partitionTimestampHi = Math.max(partitionTimestampHi, txWriter.getCurrentPartitionMaxTimestamp(o3TimestampMax));
                 txWriter.updateMaxTimestamp(Math.max(txWriter.getMaxTimestamp(), o3TimestampMax));
             } else if (replaceMaxTimestamp != Long.MIN_VALUE) {
+                // Derive partitionTimestampHi from the actual highest written timestamp, not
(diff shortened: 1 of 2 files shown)


puzpuzpuz:
Closed in favor of #7289

PR #7289, opened 2026-06-19
fix(core): fix table corruption when a range-replace commit adds partitions above the last one

puzpuzpuz:
## What

Fixes a storage-engine corruption that could occur when a WAL `REPLACE_RANGE` commit in O3 mode appends brand-new partitions above the table's previous last partition.

## Root cause

In `TableWriter.processO3Block`, replace mode derived `partitionTimestampHi` from `o3TimestampMax`, which is the replace-range high boundary rather than the highest timestamp actually written. For an open-ended range the boundary is `Long.MAX_VALUE - 1`, and `getCurrentPartitionMaxTimestamp` overflows on it. As a result `partitionTimestampHi` stayed at the stale pre-commit ceiling, `finishO3Commit` skipped switching the writer's active column files to the new last partition, and the next commit reused the previous partition's column descriptors. That commit then read the old partition's data, producing rows dated below the new partition's floor or suspending the table.

## Fix

Derive `partitionTimestampHi` from the actual highest written timestamp (`replaceMaxTimestamp`) in replace mode. The non-replace path is unchanged. In the replace path where no rows were written (`replaceMaxTimestamp == Long.MIN_VALUE`), `partitionTimestampHi` is now left unchanged rather than being recomputed from the (possibly overflowing) range boundary; the existing no-rows / truncate replace-range tests cover this case and pass.

## Effects and tradeoffs

- Positive: range-replace commits that append partitions above the last partition no longer corrupt the table or leave it suspended; a subsequent full rebuild reads correct data.
- Scope: the change touches only the replace-mode branch of `processO3Block`. Non-replace O3 commits are unaffected.
- No measurable performance impact: the fix swaps one argument to an existing call and adds no new work on the commit path.

## Test plan

- Adds `WalWriterReplaceRangeTest.testReplaceRangeAddsPartitionsAboveLastThenRebuilds`: commit A appends partitions above the last via `REPLACE_RANGE [oldMax+1us, +inf)`, then commit B does a full rebuild via `REPLACE_RANGE [min, +inf)`. The result is compared against a plain-commit oracle table.
- Verified the new test fails without the fix (table suspends with "bulk update failed and will be rolled back") and passes with it.
- Full `WalWriterReplaceRangeTest` class passes (58 tests), including the existing no-rows and truncate replace-range cases that exercise the changed branch.

---

## Also in this PR: de-flake `testFuzzConcurrentSleeps`

This branch also de-flakes an unrelated CI failure that surfaced on its own run.

### Symptom

`ServerMainSleepTest.testFuzzConcurrentSleeps` failed on `windows-other-1` (1 of 14 platforms) with `cancelled sleep runner did not exit`. The test is on `master` unchanged and shares no code path with the `TableWriter` fix; it exercises the PGWire `sleep()` / `TimerCont` cancellation machinery.

### Root cause

The test drove 64 client threads against a server with only 2 PG/HTTP/shared workers — a 32x oversubscription. On a small, heavily loaded Windows agent that starved the PG accept/dispatch path and the CPU enough that a cancelled sleep's circuit-breaker trip could not land inside the runner's 30s join window. Stale comments in the test still cited an original `12 threads * 25 iters` sizing, indicating the thread count had been cranked up after the test was written.

### Fix

- Reduce client threads 64 → 16 (still 8x the worker count, a wide margin for the parallelism gap the test asserts).
- Raise iterations per thread 8 → 16 so total coverage holds at 256 iterations; every scenario bucket is still hit many times.
- Refresh the two stale `12x25` comments.

The rarer interleavings the fuzz targets come from concurrency, not raw thread count, so the test keeps its bug-catching power.

### Tradeoff

Lowering the thread count reduces peak load on the server, so a genuine carrier-pinning regression now has a smaller (though still 8x) parallelism margin to surface against. The change does not add coverage for the `sleep()` subsystem; it only makes the existing stress test survive constrained CI agents.

### Test plan

- `testFuzzConcurrentSleeps` passes locally with the new sizing: `happy=311, cancelled=37, dropped=25, failures=0`, observed parallelism 9.4 (vs. the 2-worker pool), confirming carrier-freeing is still demonstrated and every scenario bucket is exercised.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -9497,11 +9497,17 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
             } // end while(srcOoo < srcOooMax)
 
             // at this point we should know the last partition row count
-            partitionTimestampHi = Math.max(partitionTimestampHi, txWriter.getCurrentPartitionMaxTimestamp(o3TimestampMax));
-
             if (!isCommitReplaceMode()) {
+                partitionTimestampHi = Math.max(partitionTimestampHi, txWriter.getCurrentPartitionMaxTimestamp(o3TimestampMax));
                 txWriter.updateMaxTimestamp(Math.max(txWriter.getMaxTimestamp(), o3TimestampMax));
             } else if (replaceMaxTimestamp != Long.MIN_VALUE) {
+                // Derive partitionTimestampHi from the actual highest written timestamp, not
+                // o3TimestampMax (the replace-range high boundary, which is Long.MAX_VALUE - 1 for
+                // an open-ended range and overflows getCurrentPartitionMaxTimestamp). Otherwise a
+                // replace that appends partitions above the previous last partition leaves
+                // partitionTimestampHi stale, finishO3Commit skips switching the active partition,
+                // and the next commit reuses the previous partition's column descriptors.
+                partitionTimestampHi = Math.max(partitionTimestampHi, txWriter.getCurrentPartitionMaxTimestamp(replaceMaxTimestamp));
                 txWriter.updateMaxTimestamp(replaceMaxTimestamp);
             }
         } catch (Throwable th) {
--- a/core/src/test/java/io/questdb/test/ServerMainSleepTest.java
+++ b/core/src/test/java/io/questdb/test/ServerMainSleepTest.java
@@ -441,8 +441,18 @@ public class ServerMainSleepTest extends AbstractBootstrapTest {
             // Os.sleep, ratio is bounded by workerCount; with TimerCont, ratio
             // can approach clientThreads. The bigger the gap, the less
             // ambiguous the failure mode.
-            final int clientThreads = 64;
-            final int iterationsPerThread = 8;
+            //
+            // Kept modest on purpose: 16 client threads is still 8x the worker
+            // count -- a wide enough margin for the parallelism gap -- while
+            // avoiding the connection/CPU oversubscription that, on small and
+            // heavily loaded CI agents (Windows in particular), starved the PG
+            // accept/dispatch path so badly that a cancelled sleep's breaker
+            // trip could miss its 30s join window. Do not crank these back up to
+            // chase coverage; the rarer interleavings come from concurrency, not
+            // from raw thread count, and 16x16=256 iterations already hit every
+            // scenario bucket many times over.
+            final int clientThreads = 16;
+            final int iterationsPerThread = 16;
 
             try (final ServerMain serverMain = ServerMain.create(root, new HashMap<>() {{
                 put(PropertyKey.SHARED_WORKER_COUNT.getEnvVarName(), String.valueOf(workerCount));
@@ -534,8 +544,9 @@ public class ServerMainSleepTest extends AbstractBootstrapTest {
                     t.start();
                 }
 
-                // Hard upper bound: 12 threads * 25 iters * worst-case ~600ms = ~3 min.
-                // Allow plenty of headroom; the test timeout still bounds the run.
+                // Hard upper bound: 16 threads * 16 iters * worst-case ~600ms, run
+                // in parallel, stays well under this. Allow plenty of headroom;
+                // the test timeout still bounds the run.
                 Assert.assertTrue(
                         "fuzz did not complete in time, seeds=" + seed0 + "L, " + seed1 + "L",
                         doneLatch.await(150, TimeUnit.SECONDS)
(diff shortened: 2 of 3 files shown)


puzpuzpuz:
## Code review (level 3 — mission-critical pass)

**Verdict: Approve.** The fix is correct, minimal, and well-targeted; the regression test genuinely reproduces the bug (verified to fail on the parent commit and pass with the fix). No Critical or Moderate issues. One worthwhile Minor item (the PR title).

Reviewed with correctness + cross-context, fresh-adversarial, test, and standards/perf/concurrency/resource agents, each finding source-verified. The full 58-test `WalWriterReplaceRangeTest` suite passes, and reverting the one-line fix makes the new test fail with "table is suspended".

### What the change does (verified correct)

The change in `TableWriter.processO3Block` (lines 9499–9512) splits a previously-unconditional `partitionTimestampHi` recompute into the two commit branches. Three cases, all verified:

1. **Non-replace** (`o3Commit` caller, line 7912): byte-for-byte identical — `dedupMode` is only set to `REPLACE_RANGE` on the WAL apply path (line 10647), so this caller always takes the unchanged branch.
2. **Replace with rows written**: `partitionTimestampHi` now derives from `replaceMaxTimestamp` (the true highest written timestamp, line 9341) instead of `o3TimestampMax`. Confirmed `o3TimestampMax` is the replace-range *high boundary* (`Long.MAX_VALUE - 1` for open-ended ranges), and `getCurrentPartitionMaxTimestamp` = `ceil(...) - 1` genuinely overflows on it, leaving `partitionTimestampHi` stale → `finishO3Commit` (6629–6633) skips the active-partition switch → next commit reuses stale column descriptors. This is the real bug; the fix is correct.
3. **Replace with no rows** (`replaceMaxTimestamp == Long.MIN_VALUE`): the block is now a no-op. Verified safe — when a replace removes/truncates the last partition, `txWriter.maxTimestamp` is re-derived in `o3ConsumePartitionUpdateSink` (8275) and `partitionTimestampHi` self-heals via `closeActivePartition → isLastPartitionClosed() → finishO3Commit`'s `openPartition` (8771). The old `Math.max`-based recompute could only ever *raise* `partitionTimestampHi`, which in case 3 was either a no-op (overflow) or a latent bug (bumping it to an empty partition); dropping it is correct.

Cross-context check found **no out-of-diff bugs**: both `processO3Block` callers, all consumers of `partitionTimestampHi`, the next-commit invariant assert (9650–9651), and the self-heal path were walked and shown consistent.

### Critical
None.

### Moderate
None.

### Minor

1. **PR title is truncated/garbled.** The title ends mid-phrase: `...adds partitions above the last on`. Per the repo conventions the description is copied verbatim into release notes and must read well standalone — "the last on" reads as broken. Suggest: `fix(core): fix table corruption when a range-replace commit adds partitions above the last partition`.

2. **No `Fixes #NNN` reference.** The body and test comment cite the bug as "reachable via live-view BACKFILL + O3" but reference no GitHub issue. Compliant *if* there is no tracked issue — add the reference if one exists.

3. **Case-3 coverage is by existing tests only, with a precise caveat.** The new test covers case 2. Case 3's *behavioral change* (finite-range, no-rows replace) is covered by the existing `testRemovesLastPartitionNoRowsAdded` / `testReplaceDeletesAppendsNothingToLast...` / truncate tests, which exercise the changed branch. The open-ended (`Long.MAX_VALUE`) + no-rows sub-case is not explicitly tested — but it is behaviorally **identical** between old and new code (both no-op, since the old `Math.max` saw the same overflow), so no new test is required for this change. Optional: state this precisely in the PR body rather than "existing tests cover this case".

### Downgraded (false positives caught during verification)

- *"`newPartitionRows` is misplaced in member ordering"* — Dismissed. The following method `insertRowWithReplaceRange` (2086) is `private void` (instance), not static; it belongs to a separate member group. Within the `private static` group the new helpers are correctly alphabetized, with `newPartitionRows` properly last.
- *"Case-3 dropping the recompute is a regression"* — Dismissed. Proven safe by source trace (self-heal via `openPartition`) and empirically (full suite green; reverted-fix test fails as expected).

### Summary

- 0 Critical, 0 Moderate, 3 Minor (only the title garble is worth acting on before merge, for release-note hygiene). 4 draft findings verified, 2 dropped as false positives.
- No regressions or tradeoffs: `getCurrentPartitionMaxTimestamp` call count is `≤` before (0 in case 3); no new allocations, no concurrency or resource impact (single-writer field, scalar arg swap); comment is ASCII-clean and accurate.

bluestreak01 (approved):
Approve — level 3 (mission-critical) review.

The fix in `TableWriter.processO3Block` is correct, minimal, and tightly scoped. Deriving `partitionTimestampHi` from `replaceMaxTimestamp` (the true highest written ts) instead of `o3TimestampMax` (the replace-range high boundary, `Long.MAX_VALUE-1` for open-ended ranges, which overflows `getCurrentPartitionMaxTimestamp`) is the right fix: it stops `finishO3Commit` from skipping the active-partition switch and reusing stale column descriptors on the next commit. Verified all three cases (non-replace identical; replace-with-rows fixed; replace-no-rows safely drops a recompute that was previously a no-op/over-raise). Cross-context walk of every `partitionTimestampHi` consumer and both `processO3Block` callers found no out-of-diff breakage, and the 9650 invariant holds. No concurrency or resource issues. The new `testReplaceRangeAddsPartitionsAboveLastThenRebuilds` is deterministic, oracle-compared, and a faithful regression guard.

No Critical or Moderate issues. Minor, non-blocking:
1. PR body says the de-flake keeps 'total coverage holds at 256 iterations', but it actually halved (64x8=512 -> 16x16=256). Suggest restating the 512->256 reduction explicitly as the cost of lowering oversubscription.
2. Optional: add a no-rows REPLACE_RANGE test with an open-ended (+inf) high boundary — the exact case the fix's no-rows branch covers (safe by construction today).
3. Add `Fixes #NNN` if a tracking issue exists.

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|2|2|100.00%|

PR #7291, opened 2026-06-19
chore(core): allow ADD INDEX over a parquet partition where the column is absent

RaphDal:
## Summary

`ALTER TABLE ... ADD INDEX` on a SYMBOL column can suspend a WAL table when the column is absent from one of the table's parquet partitions. The crash is an over-strict assertion in `TableWriter.indexParquetColumn`; the surrounding production code already handles the case correctly, so this only bites builds that run with assertions enabled (tests and CI).

## Root cause

`TableWriter.indexParquetColumn` reads the indexed column's columnTop and asserts it is either `0` or `partitionSize`:

```java
final long columnTop = columnVersionWriter.getColumnTop(timestamp, columnIndex);
assert columnTop == 0 || columnTop == partitionSize : ...;   // too strict
if (columnTop > -1 && partitionSize > columnTop) {           // already handles -1
    ...
}
```

`ColumnVersionReader.getColumnTop()` returns `-1` to mean "this column does not exist in the partition" (see its javadoc). That value occurs whenever the indexed column was added in a *later* partition than the one being indexed, so the older partition has no record for it. The `if (columnTop > -1 ...)` guard on the next line is written precisely to skip that case (there is nothing to index), but the assertion above it never allowed for `-1` and so trips.

The reproducing sequence (hit by the Enterprise backup parquet fuzz test):

1. add a SYMBOL column — the column-version record is written only on the then-last partition;
2. convert an *older* partition (which predates the column) to parquet;
3. `ADD INDEX` on that column.

When indexing reaches the older parquet partition, `getColumnTop` returns `-1`, the assertion fails inside `ApplyWal2TableJob`, and the WAL apply path suspends the table:

```
C ApplyWal2TableJob job failed, table suspended [seqTxn=10,
  error=java.lang.AssertionError: parquet partition indexed SYMBOL columnTop=-1
        is neither 0 nor partitionSize=1000 (columnIndex=9)]
  at TableWriter.indexParquetColumn(TableWriter.java:7220)
  at TableWriter.addIndex(TableWriter.java:858)
  at AlterOperation.applyAddIndex(...)
```

## Fix

Add `-1` to the permitted set in the assertion and update the comment to document all three legal columnTop values (`-1`, `0`, `partitionSize`). The runtime behavior is unchanged: the existing `if (columnTop > -1 && partitionSize > columnTop)` guard already skips both `-1` (column absent) and `partitionSize` (column all-NULL), indexing only when `0 <= columnTop < partitionSize`.

## Impact and scope

- This is an assertion-only defect. With assertions disabled (the default for production servers) the guard already does the right thing and the partition is correctly skipped, so there is no data-correctness change and no production behavior change.
- With assertions enabled (`-ea`, i.e. tests and CI) the assertion previously suspended the table. After this change `ADD INDEX` over such a parquet partition completes normally.
- The change does not touch the parquet metadata (`_pm`) read path, the convert-to-parquet path, or `ColumnVersion{Reader,Writer}`. It is a single assertion/comment edit.

## Test plan

- The Enterprise `BackupTest.testBackupFuzzyTableWithParquetPartitions` fuzz test surfaced this non-deterministically (failing seed: `10430439030179876L, 1781860299323L`); it passes once the assertion accepts the `-1` sentinel.
- `ParquetTest.testAddIndexOverParquetPartitionWithAbsentColumn` is a targeted, deterministic OSS regression test that reproduces the failure under `-ea`: it creates a WAL table, adds a SYMBOL column while a later partition is the last one (so an earlier partition predates the column), `CONVERT PARTITION TO PARQUET` on that earlier partition, then `ADD INDEX` on the column. It asserts the table is not suspended, that `sym = 'a'` returns the indexed rows through a pinned `Index forward scan on: sym` plan (so a silently-degraded full scan cannot mask a broken index by returning the right rows for the wrong reason), and that a full read keeps the parquet partition's NULL values alongside the later rows. The test fails on the parent commit and passes with the assertion fix.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -7208,18 +7208,23 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
         final long partitionSize = txWriter.getPartitionRowCountByTimestamp(timestamp);
         final long columnTop = columnVersionWriter.getColumnTop(timestamp, columnIndex);
 
-        // Invariant: on a parquet partition the indexed SYMBOL's columnTop
-        // is either 0 (column has data from row 0) or partitionSize (column
-        // is all-NULL in this partition). zeroColumnTopsAfterParquetRewrite
-        // collapses every intermediate 0 < columnTop < partitionSize down
-        // to 0 at convert time, so the merged decode below can rely on
-        // columnTop == 0 whenever it executes. A future ATTACH PARQUET /
-        // restore path that bypasses that routine must restore this
-        // invariant before reaching here, or the rowLo formula at the
-        // decodeRowGroup call would truncate the covered columns.
-        assert columnTop == 0 || columnTop == partitionSize
+        // Invariant: on a parquet partition the indexed SYMBOL's columnTop is
+        // one of three values. 0 (column has data from row 0):
+        // zeroColumnTopsAfterParquetRewrite collapses every intermediate
+        // 0 < columnTop < partitionSize down to 0 at convert time, so the merged
+        // decode below can rely on columnTop == 0 whenever it executes.
+        // partitionSize (an explicit column-version record marks the column
+        // all-NULL in this partition). -1 (getColumnTop's "column does not exist
+        // in this partition" sentinel, returned when the column was added in a
+        // later partition and has no record here). The guard below skips both -1
+        // and partitionSize -- there is nothing to index -- so only 0 reaches the
+        // decode. A future ATTACH PARQUET / restore path that bypasses
+        // zeroColumnTopsAfterParquetRewrite must restore this invariant before
+        // reaching here, or the rowLo formula at the decodeRowGroup call would
+        // truncate the covered columns.
+        assert columnTop == -1 || columnTop == 0 || columnTop == partitionSize
                 : "parquet partition indexed SYMBOL columnTop=" + columnTop
-                + " is neither 0 nor partitionSize=" + partitionSize
+                + " is none of -1, 0 or partitionSize=" + partitionSize
                 + " (timestamp=" + timestamp + ", columnIndex=" + columnIndex + ")";
 
         if (columnTop > -1 && partitionSize > columnTop) {
--- a/core/src/test/java/io/questdb/test/griffin/ParquetTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/ParquetTest.java
@@ -284,8 +284,11 @@ public class ParquetTest extends AbstractCairoTest {
                     engine.getTableSequencerAPI().isSuspended(engine.verifyTableName("x")));
 
             // The index is queryable: sym='a' returns the indexed rows in ts order.
+            // Pin the plan so a silently-degraded full scan cannot mask a broken
+            // index by returning the right answer for the wrong reason.
             assertQuery("x where sym = 'a'")
                     .noLeakCheck()
+                    .withPlanContaining("Index forward scan on: sym")
                     .timestamp("ts")
                     .returns("""
                             ts\tsym
--- a/core/src/test/java/io/questdb/test/griffin/ParquetTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/ParquetTest.java
@@ -241,6 +241,76 @@ public class ParquetTest extends AbstractCairoTest {
     }
 
     @Test
+    public void testAddIndexOverParquetPartitionWithAbsentColumn() throws Exception {
+        // ADD INDEX on a SYMBOL column must not suspend a WAL table when the column
+        // predates one of its parquet partitions. sym is added while 2024-01-02 is
+        // the last partition, so the older 2024-01-01 partition has no column-version
+        // record and getColumnTop returns -1 there (not 0, not partitionSize).
+        // 2024-01-01 is converted to parquet AFTER the column exists, so the parquet
+        // file physically carries sym (all-NULL) -- findParquetColumnIndex finds it and
+        // indexParquetColumn proceeds to the columnTop check. This previously tripped an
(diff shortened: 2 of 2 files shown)


RaphDal:
**Level 2.** Verdict up front: **Approve.** This is a minimal, correct, assertion-only fix with no
production behavior change, backed by a genuine red→green regression test. Every claim was verified
against source and by running the test both ways; one substantive agent finding (an alleged `IS NULL`
correctness bug) was empirically refuted.

## What the change actually does (verified)

The diff widens one debug assertion in `TableWriter.indexParquetColumn` from
`columnTop == 0 || columnTop == partitionSize` to also permit `-1`, and rewrites its comment. The
runtime guard immediately below — `if (columnTop > -1 && partitionSize > columnTop)` — is unchanged.

Confirmed against source:

- `ColumnVersionReader.getColumnTop` javadoc + `getColumnTopByIndexOrDefault`
  (`ColumnVersionReader.java:126,136`): returns `-1` only when there is no column-version record **and**
  the column was added in a *later* partition. `ColumnVersionWriter extends ColumnVersionReader` without
  overriding it — same semantics on the writer.
- The introducing commit #7171 shipped the `columnTop > -1` guard **together with** the assertion that
  never allowed `-1`. The guard was written to skip `-1` from day one; the assertion was simply
  inconsistent with its own guard. This PR makes them agree.
- `zeroColumnTopsAfterParquetRewrite(…, false)` at the CONVERT path (`TableWriter.java:1625`, `:13978`)
  leaves an absent column's `-1` untouched (`midColTop` is false, `zeroAllColumns` is false) — exactly
  the state the test creates.
- The relaxed assertion still fires for any value that is **not** `-1`, `0`, or `partitionSize` (i.e. an
  uncollapsed `0 < columnTop < partitionSize`), so it keeps its diagnostic value.

## Critical

None.

## Moderate

None.

## Minor

1. **Optional test strengthening (not a gap).** No existing test covers `IS NULL` over a
   `columnTop == -1` (added-later) parquet partition — the siblings
   `testIndexAllNullSymbolChunkOnParquetPartition{,AfterO3Rewrite}` only exercise `columnTop == 0`. I
   added a `WHERE sym IS NULL` assertion with the index plan pinned and ran it: it **passes**, returning
   all 3 NULL rows (2 from the skipped parquet partition + 1 native) via `Index forward scan on: sym`.
   Adding it would lock in the `BitmapIndexFwdReader` NullCursor synthesis for this specific case. Purely
   additive; the current test is already sufficient for the fix.

1. **"`WHERE sym IS NULL` returns 0 of the 3 NULL rows via the index" (alleged latent correctness
   bug).** **Refuted empirically.** The query returns all 3 NULL rows through the index path (verified
   twice: data-only, then with `withPlanContaining("Index forward scan on: sym")` — both green). The
   trace behind the claim missed `BitmapIndexFwdReader.getCursor:72`: for the NULL key (`sym=0`), when
   `columnTop > 0`, the reader's `NullCursor` synthesizes the implicit-NULL rows `[minValue, columnTop)`
   that aren't physically in the index. The skipped parquet partition's rows resolve through that path,
   so indexed `IS NULL` is correct. There is no follow-up bug here — the open question is answered.

2. **"columnTop == -1 could coincide with real (non-NULL) data in the parquet, so skipping drops index
   rows."** Refuted with traces: every O3 worker path that physically writes a column upserts a concrete
   `columnTop >= 0` (`updateO3ColumnTops`, `TableWriter.java:13829`; the partition-update sink is memset
   to `-1` at `:9335` and only overwritten on write), and CONVERT encodes an absent column as physically
   all-NULL (`TableUtils.produceParquetFromNative`). `getColumnTop == -1` is mutually exclusive with the
   partition holding any non-NULL symbol value. Skipping is provably correct.

3. **Comment "only 0 reaches the decode."** Accurate, given `zeroColumnTopsAfterParquetRewrite`
   guarantees no `0 < columnTop < partitionSize` survives. Descriptive, not a defect.

## PR metadata

- **Title:** Conventional Commits OK. `chore` is correct — this is an internal fuzz-test-surfaced
  assertion defect, not a customer-reported bug. Scope `core` OK; description repeats the verb and reads
  well standalone.
- **Labels** (Core, Flaky Test, WAL, storage): appropriate.
- **No `Fixes #NNN`:** correct — surfaced by the Enterprise `BackupTest` fuzz, no OSS issue.
- **Body:** root cause / fix / impact-scope / test plan, level-headed tone, correctly states "no
  production behavior change." Good. (Branch commit titles use `type(scope):` prefixes against the
  plain-English convention, but the branch is squash-merged and throwaway — not worth touching.)

## Summary

- **Verdict: Approve.** Correct, minimal, no production behavior change (assertions-off already skipped
  `-1`); under `-ea` it stops a spurious `AssertionError` from suspending WAL tables.
- **Empirical proof obtained:** test passes with the fix (`Tests run: 1, Failures: 0`); on the parent
  assertion it fails (`Failures: 1` — the `columnTop=-1` AssertionError suspends the table and
  `assertFalse(isSuspended)` catches it). Genuine regression test.
- **Findings:** the only substantive draft finding (the `IS NULL` claim) was dropped as a false positive
  after running it; one optional test-strengthening remains. No Critical, no Moderate.
- **In-diff vs out-of-diff:** all in-diff; **0 out-of-diff, correctly** — the changed symbol is a
  `private` method and the change only widens an assertion's accepted set, introducing no contract that
  any caller depends on (both callers verified: `partitionSize` is always a real partition there). The
  trivial blast radius is real, not an under-run of the cross-context pass.

No blocking changes requested. The one optional touch-up is folding in the `WHERE sym IS NULL` assertion
already confirmed green.

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|

ideoma (approved):

PR #7292, opened 2026-06-19
test(core): de-flake HTTP export file-descriptor leak checks

RaphDal:
## Summary

`ExpParquetExportTest` shares one static `testHttpClient` across all 77 methods.
`testParquetExportWithPivot` turned HTTP keep-alive on for that shared client, ran a size-asserting export, and only turned it back off *after* the assertion:

```java
testHttpClient.setKeepConnection(true);
testHttpClient.assertGetParquet("/exp", 1105, "... PIVOT ...");   // size assert
engine.execute("INSERT INTO monthly_sales ...");
testHttpClient.setKeepConnection(false);                          // skipped if the assert throws
testHttpClient.assertGetParquet("/exp", 1370, "... PIVOT ...");
```

When that first assertion fails, the method throws before re-clearing the flag, so the shared client is left in keep-alive mode with its socket still open. `TestUtils.assertMemoryLeak` calls `skipChecks()` on any throwable, so the failing pivot method does not run its own fd leak check and the dirty state goes unreported. Every later method that reuses `testHttpClient` then takes the `if (!keepConnection) httpClient.disconnect()` path as a no-op, so the open socket survives across test boundaries.

The per-test fd-count leak check counts every descriptor in the JVM, including the shared client's, so it trips on whichever *later* method happens to run while a socket is lingering -- not on the method that actually left the mess. Because JUnit runs methods in a clock-seeded random order, which method trips is non-deterministic, which is why this looked like a rare, order-dependent leak.

## How the +1 / -1 pair arises

The original failures came in two flavours: a method ending with one descriptor too few, and another ending with one too many. That is a single socket being handed down the chain, not two independent leaks:

- A method that disconnects the shared client explicitly (e.g. `testExportDropConnection`, which calls `client.disconnect()`) inherits a lingering socket in its baseline and closes it -> ends one **lower** (`-1`).
- A method that opens a socket through `assertGet` while keep-alive is still dirty and never disconnects it (e.g. an error-path test such as `testParquetExportBloomFilterFppBoundaryOne`) -> ends one **higher** (`+1`).

The culprit method (the one that fails on size and leaves the flag dirty) is a *third* method; its own leak check is skipped, so it never shows up as an fd failure -- it shows up among the size failures instead.

## Why it surfaced on the parquet-stats branch and not on master

The pivot assertion only fails when the exported size drifts. On `rd/fix-parquet-stats` the parquet writer emits extra statistics bytes, so the pivot's expected `1105`/`1370` were stale (the size-fix commit later updated them to `1147`/`1422`). During that window the pivot failed on size, left the shared client dirty, and the cross-test churn produced the fd-leak failures. On `master` the sizes match, the pivot passes, the flag is reset, and the suite is clean. The fragility is latent on master and triggers whenever a keep-alive method that uses the shared client fails for any reason.

## Fix

The pivot was the inconsistency. This class already establishes a convention: keep-alive multi-request tests use a *local* `TestHttpClient` in try-with-resources, precisely so their connection state cannot leak -- `testParquetExportDescendingPreservesDictionaryEncoding`, `testParquetExportPageFrameBackedPreservesDictionaryEncodingAcrossPartitions`, and `testParquetExportPageFrameDescendingAfterRecompile` all do this. The pivot was the only keep-alive test that used the shared static client instead.

The fix makes the pivot follow that convention -- it runs its two keep-alive requests on a local client:

```java
try (TestHttpClient client = new TestHttpClient()) {
    client.setKeepConnection(true);
    client.assertGetParquet("/exp", 1105, "... PIVOT ...");
    engine.execute("INSERT INTO monthly_sales ...");
    client.setKeepConnection(false);
    client.assertGetParquet("/exp", 1370, "... PIVOT ...");
}
```

Try-with-resources closes the local client -- disconnect plus free -- on every exit path, including an assertion failure, so a failure can no longer strand keep-alive state or a socket on the shared client. No other method sets keep-alive on the shared client, so this removes the cross-test bleed at its source. This is the whole fix; it does not touch production code.

A more defensive alternative would also reset the shared client in `setUp()` before each method's leak baseline, as a guardrail against any future test that repeats the original mistake. That is not included here -- the convention the fix restores is what the other keep-alive tests already rely on -- but it is a one-line addition if a guardrail is wanted.

## Test plan

- Reproduced deterministically: ran `ExpParquetExportTest` against the larger-output `rd/fix-parquet-stats` library with `master`'s (smaller) expected sizes, so the pivot assertion fails. Repeated full-class runs produced `cached file descriptors` leak-check failures on assorted neighbour methods (e.g. `testExportDropConnection` at start=4/end=3), and an open-site trace pinned the lingering descriptors to client sockets opened by `assertGet`.
- Captured both directions in one deterministic run: a pivot-style culprit left the client dirty, a `disconnect`-style neighbour ended at -1 (closed the inherited socket), and an `assertGet` error-path neighbour ended at +1 (left a fresh one) -- the two sockets being consecutive allocations, i.e. the same slot churning.
- With the fix and the same stale-size setup, the pivot still fails on size (as expected -- those sizes are deliberately stale here) but, because it now fails on a local client, no fd leak-check failure occurs over repeated full-class runs.
- `ExpParquetExportTest` is green on this branch (correct sizes), unchanged by the fix: 77 run, 0 failures.


RaphDal:
# Code Review: PR #7292 — `test(core): de-flake HTTP export file-descriptor leak checks`

**Level 2.** Reviewed at: full change-surface map, fresh-context adversarial agent +
targeted fd-semantics verification agent, all findings verified against source. The diff
is 17 lines, test-only, in one file (`HttpQueryTestBuilder.java`).

## Verdict: Approve — safe to merge

The fix is mechanically sound. `awaitFileDescriptorQuiescence` **only reads two counters
and sleeps — it has zero side effects on fd state.** Therefore it cannot convert a passing
leak check into a failure or a real leak into a pass; `LeakCheck.close()` re-evaluates the
same condition authoritatively afterward. The barrier can only change *timing* — delaying
when `close()` samples so this test's own async teardown (engine/reader-pool fd release)
settles before the next test starts. That is a legitimate, low-risk way to stop one test's
teardown tail from bleeding into a neighbor's leak-check window.

Both named flaky methods funnel through the patched method (`testParquetExportBloomFilterFppBoundaryOne`
via `.run()`; `testExportDropConnection` via `.runNoLeakCheck()`, both wrapped in
`assertMemoryLeak`). No subclasses of `HttpQueryTestBuilder`. Imports, member ordering, and
naming conform. No SQL/query assertions are added, so the test-assertion-API rules don't apply.

## Critical

None.

## Moderate

None.

## Minor

**1. The PR body's "captured at the same point in the lifecycle" is imprecise (no functional
impact).** The barrier samples its baseline at `runNoLeakCheck` *entry* and waits at *exit*
(`HttpQueryTestBuilder.java:117-118, 343`), whereas `LeakCheck` samples in its constructor and
asserts in `close()`, which run *additional* cleanup (`Path.clearThreadLocals()`,
`Misc.free(O3PartitionJob.THREAD_LOCAL_CLEANER)`, `CLOSEABLE.forEach(Misc::free)`) before each
fd snapshot (`TestUtils.java:2601-2613, 2630-2642`). I verified those three steps are
**fd-neutral** — they free only native/`Unsafe` memory, and the O3 path's fds are
`Files.detach()`-ed off the counters before donation to Rust (`O3PartitionJob.java:276-302`,
`FdCache.java:142-152`). So the imprecision is harmless: the barrier and the assert read the
same fd quantity. Suggest softening the claim in the PR body to "the barrier waits on the same
fd counters the leak check asserts; the intervening `LeakCheck.close()` cleanup does not touch
fd counts" rather than "same instant."

**2. The "~5 s" bound is a lower bound.** `Os.sleep(1)` is `Thread.sleep(1)`
(`Os.java:231-243`), which the Linux scheduler rounds up; 5000 iterations is realistically
5–30 s under CI load. This cost lands on every *non-settling* case — genuine leaks (disclosed)
and any transient that never returns to the exact baseline. For green runs it's ~0 (settling
tests return on iteration 0), so CI impact is negligible in practice, but the PR body should
say "≥5 s" not "~5 s."

**3. Exact-`==` wait only drains *positive* churn.** The loop returns only on
`count == baseline` (`HttpQueryTestBuilder.java:456`). If a test's activity drives the count
*below* its own entry baseline (e.g. evicting a cached descriptor inherited from an earlier
test), the barrier can never satisfy `==`, spins to the full timeout, and `close()` then fails
on its own baseline anyway. This is a real limitation of the heuristic — it absorbs the "this
test opened fds that close shortly after" direction but not the below-baseline eviction
direction. It's consistent with the PR's "a genuine leak never settles" disclosure; worth one
sentence extending that disclosure to the below-baseline case.

**4. Silent timeout — no diagnostic.** On hitting 5000 iterations the method just returns
(`HttpQueryTestBuilder.java:455-460`), leaving no trace. When the subsequent
`LeakCheck.close()` then fails, there's nothing in the log to distinguish "barrier drained,
real leak" from "barrier timed out." A single `LOG.info()` on timeout (ASCII only, per
CLAUDE.md) would make post-mortems cheaper. Optional.

## Downgraded (false positives / disproven)

- **"Lifecycle mismatch causes spurious ~5 s stalls on passing tests" (fresh agent, rated
  HIGH)** — Dismissed. Rested on the hypothesis that `LeakCheck.close()`'s cleanup releases the
  fd the barrier waits for. The targeted verification agent traced all three cleanup steps and
  proved they cannot decrease either fd count (native-memory-only; O3 fds already detached from
  the counters). The barrier waits on the async engine/reader-pool fd drain instead, which is
  the correct target. Reduced to the wording nit in Minor #1.
- **Non-atomic two-count read in the loop condition** — Dismissed as a practical issue.
  `getOpenFileCount()` (AtomicLong) and `getOpenCachedFileCount()` (synchronized) are read
  without a shared lock, but `workerPool.halt()` joins all workers before the barrier runs
  (`HttpQueryTestBuilder.java:336`), and a torn read merely costs one extra poll iteration — it
  never makes the outcome worse than no barrier. Latent only; not worth changing for the current
  callers.

## Summary

- **Verdict: Approve / safe to merge.** No correctness risk: the barrier is side-effect-free on
  fd state, so `LeakCheck.close()` remains authoritative and a real leak still fails (slower,
  bounded). The efficacy rests on a sound positive-churn-drain mechanism plus the author's
  empirical validation (7 green runs vs. reproduced-before); that's appropriate for a flake fix.
- **Tradeoffs (well-disclosed by the author):** symptom-level fix at the harness boundary rather
  than teaching `LeakCheck` itself to drain+retry; bounded busy-wait slows genuine-leak
  reporting; scope limited to `HttpQueryTestBuilder` tests. The PR body is exemplary in
  surfacing these.
- **Verification:** 6 draft findings → 4 verified (all Minor: wording, latency magnitude,
  exact-`==` limitation, missing timeout log), 2 dropped as false positives (the HIGH
  lifecycle-stall claim and the non-atomic-read race).
- **In-diff vs out-of-diff: 4 in-diff, 0 out-of-diff.** Zero out-of-diff is correct here, not an
  underrun: the change is purely additive harness drainage with no observable contract change —
  both callers and the absence of subclasses were verified.

The only actionable asks are documentation tweaks to the PR body (Minor #1–#3) and an optional
timeout log line (Minor #4). None block merge.

RaphDal:
# Code Review: PR #7292 — `test(core): de-flake HTTP export file-descriptor leak checks`

**Reviewed at level 2.** (Level 3 exists for a full 11-agent pass; this is a 17-line
test-only change, so the fan-out was scaled to the risk surface: full inline
change-surface verification plus one fresh-context adversarial agent.) Findings verified
against source on both sides of the socket boundary.

## Verdict: Approve — safe to merge

The fix is correct, complete, and resource-safe. It moves the only keep-alive test that used
the shared static `testHttpClient` onto a local `TestHttpClient` in try-with-resources, so a
failed assertion can no longer strand a keep-alive socket on the shared client and bleed a
descriptor into a neighbouring test's fd leak check.

**Verified facts behind the approval:**

- **fd released on every exit path, including the first-assertion-throws case.** When the
  first `assertGetParquet` throws while `keepConnection == true`, `TestHttpClient`'s `finally`
  skips `disconnect()` (`TestHttpClient.java:347-349`) — but try-with-resources still calls
  `close()` -> `Misc.free(httpClient)` -> `HttpClient.close()` (calls `disconnect()` first,
  `HttpClient.java:100`) -> `Misc.free(socket)` -> `socket.close()` releases the OS fd. The
  `if (bufLo != 0)` guard (`HttpClient.java:101`) and an idempotent `socket.close()` (fd
  already `-1`) make the close/disconnect double-call safe — no double-free.
- **Root cause is real and fully scoped.** Every other `setKeepConnection(true)` in the file
  is on a local client (lines 596, 1231, 1635, 1890, 2330 — several shadow the static field
  name with a local var in try-with-resources). The pivot was the only keep-alive test on the
  shared client, so this removes the bleed at its source. `testExportDropConnection`
  (line 448) is a victim, not a co-culprit: it calls `client.disconnect()` (line 463) and
  never sets keep-alive.
- **Leak check still has teeth and fires at the right time.** `run()` ->
  `assertMemoryLeak(() -> runNoLeakCheck(...))` (`HttpQueryTestBuilder.java:110-111`) wraps the
  lambda; the local client closes *inside* the lambda, before the fd snapshot. On success the
  full `LeakCheck.close()` runs (including the OS fd-count check), so coverage is not weakened.
- **Keep-alive is still genuinely exercised.** Both requests go through the same local `client`
  to matching host/port, so the second `newRequest` reuses the open socket
  (`HttpClient.java:121-124`) rather than reconnecting — the same fd spans both requests.
- **Port is correct.** `getExportTester()` never sets a port; server and `TestHttpClient` both
  default to 9001, so the omitted `client.port = testHttpClient.port` (which other tests do
  redundantly) is a no-op here.
- **`@AfterClass assert NATIVE_HTTP_CONN == 0` (line 91) is unaffected** — the local client's
  header buffer is freed by `close()` within the lambda.

## Critical
None.

## Moderate
None.

## Minor (all optional, none blocking)

**1. The shared static `testHttpClient` is still a latent recurrence vector — consider the
guardrail.** This PR fixes the one method that misused the shared client, but the class still
hands a single mutable client (carrying `keepConnection` + an open socket) to all 77 methods,
and the only thing preventing a repeat is an unenforced convention a future contributor won't
know about. The author's own documented alternative — one line in `setUp()`:

```java
testHttpClient.setKeepConnection(false);
```

would defend the entire class against this whole bug class for free, by resetting the baseline
before each method's leak check. The PR explicitly weighs and declines this, which is a
defensible call for a minimal flake fix; including it would harden against a repeat of exactly
the fragility that just cost a flake, but it does not block.

**2. The inline comment (lines 2245-2249) runs to 5 lines** — longer than the project's
concise-comment convention (deep rationale belongs in the PR body, which already covers this in
full). Could collapse to ~2 lines, e.g. "Use a local client (not the shared testHttpClient) so
try-with-resources closes the keep-alive socket on every path, including assertion failure — a
stranded socket otherwise bleeds into another test's fd leak check." Pure style.

**3. Housekeeping: a stale inline PR review comment describes the abandoned approach.** The
earlier self-review comment on the PR reviews `awaitFileDescriptorQuiescence` in
`HttpQueryTestBuilder.java` — that barrier (commit `07febc2eef`) was fully reverted by the
local-client refactor (`c72160c59e`); `grep` confirms no `Quiescence` code remains and
`git diff master...HEAD` touches only `ExpParquetExportTest.java`. The **PR body is current and
matches the diff**; only that one inline comment is stale. Worth resolving/deleting it so it
doesn't mislead reviewers (its two findings — the HIGH "lifecycle stall" and the "non-atomic
read" — applied only to the deleted barrier and are moot).

## Downgraded (false positives / not applicable)

- **All findings from the stale inline self-review** ("lifecycle-stall ~5s stalls",
  "non-atomic two-count read", "silent timeout", "exact-`==` only drains positive churn") —
  these targeted the `awaitFileDescriptorQuiescence` barrier, which is **no longer in the PR**.
  Not applicable to the current diff.
- **Adversarial agent's "HTTP status never validated, only byte size" note** — pre-existing
  behavior of the 3-arg `assertGetParquet` across the whole file; not introduced or changed by
  this diff. Out of scope.

## Summary

- **Verdict: Approve / safe to merge.** Zero correctness or resource risk. The fix releases
  all fds/native memory on every exit path, preserves keep-alive coverage and the leak-check
  teeth, and removes a genuine, order-dependent cross-test fd bleed at its source.
- **Tradeoffs (well-disclosed by the author):** symptom-scoped fix that fixes the one offending
  method rather than hardening the shared-client pattern class-wide; the PR body documents the
  one-line `setUp()` guardrail as the deliberate omission (Minor #1).
- **Verification:** inline pass + 1 fresh-context adversarial agent both found **0
  correctness/resource issues**. 3 minor items, all optional (a guardrail suggestion, a
  comment-length nit, and a stale-comment cleanup).
- **In-diff vs out-of-diff: 0 actionable findings either way.** The changed code is a `@Test`
  method body with no callers, so the callsite inventory is empty by construction; the
  meaningful out-of-diff check was the shared-client contract (does any *other* method leave it
  dirty?) — verified clean. Zero out-of-diff findings is correct here, not an underrun.
- **Not independently run:** the suite was not re-executed. The diff leaves the size assertions
  (1105/1370) unchanged, so the author's "green on this branch, 77 run/0 failures" plus the
  triggered `azp run macwin` cover the runtime result; the change cannot affect exported byte
  sizes.

bluestreak01 (approved):
Approve. Clean, minimal, test-only fix that eliminates the cross-test file-descriptor bleed at its root.

Verified against source:
- TestHttpClient.close() -> Misc.free(httpClient) disconnects the socket, so try-with-resources tears it down on every exit path including assertion failure.
- assertMemoryLeak (TestUtils.java:858) calls skipChecks() on any throwable, confirming the culprit's own fd check is masked, as described.
- After this change, ZERO call sites set keep-alive on the shared static client - all remaining setKeepConnection calls operate on local clients (lines 1231/1635/2330 are local shadows, not the static field). The fix removes the bleed at its source, not just for the pivot's happy path.
- The local client is created and closed inside the leak-checked run(...) lambda, so no new leak is introduced.
- Fix mirrors the three existing keep-alive convention tests in the same file.

Minor (non-blocking):
- Title scope test(core) would be more accurate as test(http); the branch commit already uses test(http).
- The shared static client remains a latent foot-gun for any future keep-alive test; the PR explicitly defers the optional setUp() guardrail, which is reasonable since no caller now sets keep-alive on it.

PR #7297, opened 2026-06-20
fix(core): fix O3 lag commit recording maxTimestamp below the committed max across partitions

bluestreak01:
## What

Fixes a data-ordering bug in the non-WAL out-of-order (O3) commit path where an out-of-order commit with lag (`TableWriter.ic()`) could leave the table's `maxTimestamp` **below** the maximum timestamp actually committed to disk. A later O3 commit then merged against that stale boundary and left a **single timestamp inversion** in a designated-timestamp table, so queries (which scan in storage order) returned rows out of order.

## How it was found

CI build `243069` failed in `O3MaxLagFuzzTest.testRollbackFuzzParallel` — a randomized, time-seeded fuzz test (so flaky run-to-run, and unrelated to the PR it ran on). Recovering the seed from the log gave the exact parameters (`nTotalRows=57091, lim=151, microsBetweenRows=1545848`), which reproduce the failure deterministically, single-threaded, and **without** the rollback in the original test. The corruption: all rows present and distinct, but exactly one `ts` inversion.

## Root cause

When the uncommitted rows of a lag commit span **more than the active partition**, `o3MoveUncommitted()` pulls only the active partition's rows into the sorted O3 batch and empties it. The in-order rows left on disk in the previous (now last) partition stay committed but are **not** part of the sorted O3 batch, so the committed `maxTimestamp` was recomputed solely from the O3 batch boundary (`o3TimestampMax`) and could end up lower than the true on-disk maximum.

Concretely, after the first `ic()`:

```
writer.getMaxTimestamp() = 1970-01-01T22:59:57.164728Z   (recorded)
max(ts) committed on disk = 1970-01-01T23:53:23.253480Z  (actual)
```

The next O3 commit then used the too-low `maxTimestamp` as the last partition's existing-data upper bound (`dataTimestampHi`) and mis-merged, placing the higher-`ts` rows physically before the lower-`ts` incoming rows.

## Fix

In `o3MoveUncommitted()`, when the active partition is emptied while uncommitted rows remain in an earlier partition, read that partition's actual max timestamp (via the existing `readPartitionMinMaxTimestamps()`) and fold it into the committed `maxTimestamp` computation in `processO3Block()`, so the writer's `maxTimestamp` is never below what is committed on disk.

- The new logic only activates in this specific multi-partition-uncommitted-span scenario; the common path is unchanged.
- The WAL commit path is unaffected (it uses a separate lag mechanism and passes `Long.MIN_VALUE`).

## Test

Adds `O3MaxLagFuzzTest.testRollbackRegression3`, pinning the failing parameters from the randomized test. It **fails before** the fix (`Row 53563 column i expected 53563 but was 53577`) and **passes after**.

## Validation

Regression test fails without the fix / passes with it, plus no regressions across the O3/WAL suites:

- `O3Test` (148), `O3FailureTest` (53)
- `O3MaxLagTest` (33), `O3MaxLagRetentionTest` (12)
- `O3SplitPartitionTest` (13), `O3SquashPartitionTest` (18)
- `O3MaxLagFuzzTest` (5), `WalWriterTest` (78)


bluestreak01:
## Code review (level 3, adversarial)

**Verdict: Approve with minor changes.** The core fix is correct and well-targeted. Multi-agent review + per-finding source verification confirm the logic; no correctness blockers. Findings are confined to one style-rule violation, a performance optimization opportunity with a real robustness tradeoff, test-coverage gaps on the fix's untested branches, and PR metadata.

The fix is sound: `o3MoveUncommitted()` empties the active partition via `txWriter.resetToLastPartition(0)`, which resets `maxTimestamp` to the persisted committed value (`TX_OFFSET_MAX_TIMESTAMP_64`, `TxWriter.java:832-834`). When an in-order append run crossed a partition boundary this batch, the previous partition (`count-2`) holds appended-but-uncommitted rows whose on-disk max exceeds both that persisted value and the lag-trimmed `o3TimestampMax`. Folding the freshly-read prev-partition max into `updateMaxTimestamp` (L9539-9542) restores the invariant. This is **not** a no-op precisely because `resetToLastPartition` discards the live in-memory max.

### Critical
None.

### Moderate

**M1 — Avoidable disk read on the commit path (performance)**
`TableWriter.java:8573-8576`. On every qualifying commit (`rowsAdded > transientRowCount`) the fix opens and reads a file to recover `count-2`'s max: native = `openRO` + 2× `readNonNegativeLong` + `close`; parquet = `mmap` the `_pm` footer + parse + `munmap` (worst case, purely to recover one `long`).

The value is resident in memory in the common case: `newRow` calls `switchPartition(...)` **before** `updateMaxTimestamp(timestamp)` (L2590/L2598), so at the boundary switch `txWriter.getMaxTimestamp()` equals the outgoing (`count-2`) partition's max. Capturing it into a field at switch time would be O(1), zero syscalls.

Tradeoff: the disk read is more robust — it recomputes from ground truth and is immune to a stale capture if `count-2` were modified on disk by an intervening mid-batch merge; the in-memory approach needs per-commit invalidation discipline. This path is per-commit, not per-row, and is gated. Recommendation: consider the in-memory capture (especially to avoid the parquet mmap) with a clear invalidation rule, or document why the disk read was chosen. Not a merge blocker.

**M2 — Test covers only one of the fix's branches**
`testRollbackRegression3` exercises only the native-partition, `prevSize > 0`, exactly-2-partition, micros-timestamp case (`testRollbackFuzz0` hardcodes a micros `timestamp` column and discards the parameterized `timestampTypeName`). Untested branches of the new code:
- **Previous partition is PARQUET** — `readPartitionMinMaxTimestamps` routes to entirely different read code (`readParquetMetaMinMaxTimestamps`/`readParquetMinMaxTimestamps`). Strongest gap; add an `ALTER TABLE ... CONVERT PARTITION TO PARQUET` on the prior partition before the boundary-spanning `ic()`.
- `prevSize == 0` guard branch (the no-op path — see m3).
- Uncommitted rows spanning > 2 partitions.
- Split previous partition.

Also worth adding: a minimal **pool=0, no-rollback** repro matching the PR body's stated single-threaded reproduction (the current test piggybacks on the rollback fixture at pool=2). pool=2 is deterministic here (fixed-seeded data, and the `updateMaxTimestamp` the fix touches runs in the single-threaded post-join section), so not flaky — but a minimal repro isolates the trigger.

### Minor

**m1 — Member-ordering violation**
`TableWriter.java:366`. `o3MoveUncommittedMaxTimestamp` was inserted between `o3MasterRef` (L363) and `o3MemColumns1` (L367), but `o3Mo...` sorts **after** `o3Mem...` (4th char `o` > `e`). Coding standard requires same-kind members sorted alphabetically. Correct position: after `o3MemColumns2` (L368), before `o3NullSetters1` (L369).

**m2 — New `CairoException.critical` throw site on the commit path doesn't set `o3InError`**
`TableWriter.java:8573`. `readPartitionMinMaxTimestamps` can throw (read failure, or the "data does not match partition directory name" validation originally written for ATTACH PARTITION). Verified: no fd/mapping leak (fd closed in finally; `path` restored in two nested finallys; mmap munmapped on all paths), and the throw fires before any state mutation, so `finishO3Append(0)` leaves the writer closeable and recoverable via `rollback()`. Unlike the adjacent epoch-check throws, this one does not set `o3InError = true` — assessed benign. It does expand the commit failure surface (this path previously did no I/O); flagged for awareness.

**m3 — Fix silently no-ops if `count-2` is empty (low reachability)**
`TableWriter.java:8567-8569`. The `prevSize > 0` guard means a trailing/intermediate empty partition at `count-2` leaves the field at `Long.MIN_VALUE` and the bug unfixed for that layout. `reconcileOptimisticPartitions()` normally prunes trailing empties, so likely unreachable in steady state, but the fix assumes `count-2` is non-empty. A defensive note or test would help. (Handling only `count-2` is *correct* for non-empty layouts even with 3+ partition spans, since partitions are timestamp-ordered and `count-2` holds the global remaining max.)

**m4 — Redundant work in `readPartitionMinMaxTimestamps`**
`TableWriter.java:11491`. The caller needs only `attachMaxTimestamp`, but the helper also reads min and runs two `getPartitionTimestampByTimestamp` validation binary searches. Cheap and the validation has safety value; subsumed if M1's read is eliminated.

**m5 — PR metadata**
- Title is internal-facing and long (~92 chars); the description is copied to release notes and should speak to user impact. Suggested: `fix(core): fix out-of-order timestamps after O3 commit with lag spanning partitions`.
- No labels. Add `Bug`, `Core`, `storage`. Do not add `WAL` (WAL path is explicitly unaffected); `Fixes #NNN` is correctly absent (found via CI fuzz, no tracked issue).
- Optional: field/param names diverge (`o3MoveUncommittedMaxTimestamp` vs `committedDataMaxTimestamp`); the param comment bridges them, acceptable.

### Downgraded (false positives / non-issues)
- "Reading only `count-2` misses the true max for >2-partition spans" — Dropped. Partitions are timestamp-ordered; after the active partition is emptied, `count-2` is the highest-timestamp partition still holding data, so its max is the global committed-on-disk max.
- "`srcOooMax == 0` early-return path doesn't apply the fix" — Dropped. Nothing is persisted there, the writer stays in O3 append mode, and held-back lag rows carry the highest timestamps, so the next commit corrects `maxTimestamp`. Self-correcting.
- "`partitionTimestampHi` still derived from `o3TimestampMax`" — Dropped. Self-heals: recomputed from `getMaxTimestamp()` on the next `openLastPartition`/`finishO3Commit`.
- "Non-volatile field is a concurrency hazard" — Dropped. TableWriter is single-threaded per instance; the field is reset at `o3MoveUncommitted` entry, set/read on the writer thread before any worker fan-out, never referenced by any O3 worker job.
- "Resource leak when the new read throws" — Dropped. fd closed in finally, mappings munmapped, `path` restored; writer left consistent.
- "WAL apply path behavior changed" — Dropped. `processWalCommitFinishApply` passes `Long.MIN_VALUE`, which folds out of `Math.max`; byte-for-byte identical to pre-PR.
- Ascending-order assumption inside `count-2` — Verified safe. O3 rows go to memory (`newRowO3`), never disk; on-disk native partitions are always timestamp-sorted.

### Summary
- The core data-ordering fix is correct and minimal; the WAL path is provably unaffected; no resource or concurrency regressions.
- Tradeoffs: the fix adds per-commit file I/O (M1) and a new commit-path throw site (m2) in exchange for robustness — defensible, but the I/O is avoidable in the common case.
- In-diff vs out-of-diff: effectively all findings in-diff; **0 broken out-of-diff callsites**. Justified, not an underrun: both `processO3Block` callsites and all `getMaxTimestamp()` consumers (~19 files: readers, `ApplyWal2TableJob`, mat-view refresh, `O3PartitionJob`) were walked — the change only *raises* `maxTimestamp` to the correct value in the buggy case, fixing the downstream merge-boundary consumer rather than breaking any caller.

<sub>Reviewed at level 3 (full mission-critical pass).</sub>

ideoma:
https://github.com/questdb/questdb/pull/7297#issuecomment-4760031909 M1:   This path never reads the _pm file. count-2 is always a native partition here: it's the non-WAL O3 path, and the gate (rowsAdded > transientRowCount) only fires when count-2 just received in-order native appends — so readPartitionMinMaxTimestamps() always takes the native  branch. Keeping the read via readPartitionMinMaxTimestamps() is fine — this condition is hit very rarely, so it's not a hot path worth optimizing.

ideoma (approved):

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|16|16|100.00%|

PR #7298, opened 2026-06-20
fix(sql): concurrent NPE in CoveringIndexRecordCursorFactory.getCursor (#7294)

### LINKED ISSUE #7294: Intermittent NullPointerException in CoveringIndexRecordCursorFactory.getCursor under concurrent queries (window function over UNION ALL with posting/covering index)
### Describe the bug

Queries that combine a **window function** over a **`UNION ALL`** whose first branch scans a table through a **posting/covering index** (`SYMBOL INDEX TYPE POSTING INCLUDE (...)`, introduced in 9.4.0) fail **intermittently** with:

```
java.lang.NullPointerException: Cannot invoke
"io.questdb.cairo.sql.Function.init(io.questdb.cairo.sql.SymbolTableSource, io.questdb.griffin.SqlExecutionContext)"
because the return value of "io.questdb.std.ObjList.getQuick(int)" is null
```

The failure only happens when **multiple such queries run concurrently** (over PG wire in our case). The exact same query always succeeds when executed alone, which strongly suggests a race condition. The failure distribution is non-deterministic: repeating the same batch of queries fails on different queries each round.

Our real-world trigger is Grafana opening a dashboard: every panel fires its query at the same time, so a fraction of panels randomly fail with HTTP 500 ("error querying the database: pq: Cannot invoke ...").

**9.4.2 does not fix it.** The release notes mention "Fixes for the posting and covering index from 9.4.0", but we measured an even higher failure rate on 9.4.2 (~14%) than on 9.4.1 (~5%), with a byte-identical stack trace.

### Steps to reproduce

Schema (table populated by ILP, a few hundred rows are enough; ~250 rows in our case):

```sql
CREATE TABLE signals_boolean (
    ts TIMESTAMP,
    name SYMBOL INDEX TYPE POSTING INCLUDE (source, value, ts),
    source SYMBOL,
    value BOOLEAN
) TIMESTAMP(ts) PARTITION BY DAY TTL 180 DAYS
DEDUP UPSERT KEYS(ts, name, source);
```

Failing query (Grafana macros expanded to literal timestamps; `name` values exist in the table):

```sql
SELECT ts, last_value(v) ignore nulls over (order by ts) AS lv
FROM (
  SELECT ts, 1 AS v
  FROM signals_boolean
  WHERE name IN ('signal.a','signal.b','signal.c','signal.d','signal.e')
    AND ts <= cast(1781184000000000 as timestamp)
  UNION ALL SELECT cast(1781042400000000 as timestamp), cast(null as int)
)
```

Driver: run **~20 instances of this query concurrently over PG wire**, repeat a few rounds. With this minimal query we observe ~2% failures; with our production variant (6 window columns over 6 `CASE` pivots, 3-branch `UNION ALL`, outer time filter + `ORDER BY`) the failure rate rises to **~14% on 9.4.2** and ~5% on 9.4.1 (n = 2890 executions, 5 rounds × 578 queries × 20 concurrent).

Important negative results (all 0 failures in 100 concurrent executions each):
- The bare indexed scan (`SELECT ... WHERE name IN (...)`) alone: **OK**.
- The window function directly over the indexed scan (no `UNION ALL`): **OK**.
- The `UNION ALL` without the window function: **OK**.

So the minimal trigger we found is: **window function → UNION ALL → covering-index scan + constant-row branch, under concurrency**.

### Expected behavior

The query either succeeds consistently or fails deterministically with a meaningful error — no intermittent NPE depending on concurrent load.

### What we observed instead

Stack trace (identical on 9.4.1 and 9.4.2):

```
java.lang.NullPointerException: Cannot invoke "io.questdb.cairo.sql.Function.init(io.questdb.cairo.sql.SymbolTableSource, io.questdb.griffin.SqlExecutionContext)" because the return value of "io.questdb.std.ObjList.getQuick(int)" is null
	at io.questdb.cairo.sql.Function.init(Function.java:100)
	at io.questdb.griffin.engine.table.CoveringIndexRecordCursorFactory.getCursor(CoveringIndexRecordCursorFactory.java:196)
	at io.questdb.griffin.engine.table.VirtualRecordCursorFactory.getCursor(VirtualRecordCursorFactory.java:124)
	at io.questdb.griffin.engine.union.AbstractSetRecordCursorFactory.getCursor(AbstractSetRecordCursorFactory.java:73)
	at io.questdb.griffin.engine.union.UnionAllRecordCursorFactory.getCursor(UnionAllRecordCursorFactory.java:32)
	at io.questdb.griffin.engine.union.AbstractSetRecordCursorFactory.getCursor(AbstractSetRecordCursorFactory.java:73)
	at io.questdb.griffin.engine.union.UnionAllRecordCursorFactory.getCursor(UnionAllRecordCursorFactory.java:32)
	at io.questdb.griffin.engine.window.CachedWindowRecordCursorFactory.getCursor(CachedWindowRecordCursorFactory.java:203)
	at io.questdb.griffin.engine.table.FilteredRecordCursorFactory.getCursor(FilteredRecordCursorFactory.java:75)
	at io.questdb.griffin.engine.table.SelectedRecordCursorFactory.getCursor(SelectedRecordCursorFactory.java:126)
	at io.questdb.griffin.engine.table.VirtualRecordCursorFactory.getCursor(VirtualRecordCursorFactory.java:124)
	at io.questdb.griffin.engine.orderby.EncodedSortLightRecordCursorFactory.getCursor(EncodedSortLightRecordCursorFactory.java:71)
	at io.questdb.griffin.engine.QueryProgress.getCursor(QueryProgress.java:303)
	at io.questdb.cutlass.pgwire.PGPipelineEntry.msgExecuteSelect(PGPipelineEntry.java:1474)
	at io.questdb.cutlass.pgwire.PGPipelineEntry.msgExecute(PGPipelineEntry.java:630)
	at io.questdb.cutlass.pgwire.PGConnectionContext$PGConnectionBatchCallback.postCompile(PGConnectionContext.java:1601)
	at io.questdb.griffin.SqlCompilerImpl.compileBatch(SqlCompilerImpl.java:480)
	at io.questdb.cairo.pool.SqlCompilerPool$C.compileBatch(SqlCompilerPool.java:150)
	at io.questdb.cutlass.pgwire.PGConnectionContext.msgQuery(...)
```

The error occurs both with `cache=false` and `cache=true` in the `QueryProgress` log line, and with `jit=true`.

`EXPLAIN` of the failing query shows the branch goes through the covering index:

```
CachedWindow
  Union All
    VirtualRecord
      CoveringIndex on: name with: ts, source, value
        filter: name IN [...]
    VirtualRecord (constant row)
      long_sequence count: 1
```

### Workarounds we verified

Both fully suppress the error (0 failures in thousands of concurrent executions):

1. **Classic symbol index** — `name SYMBOL INDEX CAPACITY 256` instead of the posting index. The plan becomes `FilterOnValues` + `Index forward scan on: name` and works flawlessly (tested on 9.4.1, 0/2890).
2. **Defeating the index at query level** — `WHERE cast(name as varchar) IN (...)`. The plan becomes `Async Filter` (full scan) and works flawlessly (tested on 9.4.1, 0/1734).

This confines the defect to the `CoveringIndexRecordCursorFactory` code path.

### Environment details (no dedicated form field)

- WAL table, partitioned by day.
- DEDUP upsert keys `(ts, name, source)`.
- ~10–60 distinct `name` symbols per table, low row counts (hundreds).
- Client: PG wire protocol via the Grafana QuestDB datasource plugin, ~20 concurrent connections.

### QuestDB version:

 9.4.2

Ubuntu 24.04.3 LTS (Docker)

xfs

### Full Name:

Tarquinio De Nicola

### Affiliation:

Lan System

- [x] Yes, I have

### Additional context

_No response_

bluestreak01:
## What

Fixes #7294 — an intermittent `NullPointerException` in
`CoveringIndexRecordCursorFactory.getCursor` that occurs under concurrent queries.

```
java.lang.NullPointerException: Cannot invoke
  "io.questdb.cairo.sql.Function.init(io.questdb.cairo.sql.SymbolTableSource, io.questdb.griffin.SqlExecutionContext)"
  because the return value of "io.questdb.std.ObjList.getQuick(int)" is null
    at io.questdb.cairo.sql.Function.init(Function.java:100)
    at io.questdb.griffin.engine.table.CoveringIndexRecordCursorFactory.getCursor(...)
    at io.questdb.griffin.engine.table.VirtualRecordCursorFactory.getCursor(...)
    at io.questdb.griffin.engine.union.UnionAllRecordCursorFactory.getCursor(...)
    at io.questdb.griffin.engine.window.CachedWindowRecordCursorFactory.getCursor(...)
```

It reproduces for the query shape in the report: a **window function** over a
**`UNION ALL`** whose first branch is an **IN‑list scan served by the posting/covering
index** (`SYMBOL INDEX TYPE POSTING INCLUDE (...)`). It only manifests when several such
queries run concurrently (e.g. all panels of a Grafana dashboard firing at once over PG wire).

## Root cause

`CoveringIndexRecordCursorFactory` stored a **direct reference** to
`intrinsicModel.keyValueFuncs`:

```java
this.keyValueFuncs = keyValueFuncs; // intrinsicModel.keyValueFuncs
```

That `ObjList` is **pooled** — it belongs to the compiler's `WhereClauseParser`
(`ObjectPool<IntrinsicModel>`). `SqlCompiler`s are themselves pooled and shared across
threads/connections, so the lifecycle of that list is **not** tied to the factory:

1. Thread A compiles the query, borrows a pooled compiler, builds the covering factory
   (holding a reference to that compiler's pooled `keyValueFuncs`), and returns the
   compiler to the pool. The factory is cached (prepared statement / window operator).
2. Thread B borrows the **same** pooled compiler and recompiles. `models.next()` →
   `IntrinsicModel.clear()` → `keyValueFuncs.clear()`, and `ObjList.clear()` runs
   `Arrays.fill(buffer, null)` **before** resetting `pos = 0`.
3. Thread A's `getCursor()` calls `Function.init(keyValueFuncs, …)`, which reads a stale
   `size()` (> 0) and then `getQuick(i)` on an already‑nulled slot → **NPE**.

This is why it is intermittent and concurrency‑only: a single query run in isolation never
has another thread clear the shared list between compile and execute. The
window‑over‑`UNION ALL` shape simply widens the window between the factory being
built/cached and `getCursor()` running, so it surfaces there while a bare scan effectively never did.

## Fix

Keep our **own copy** of the list in the constructor:

```java
this.keyValueFuncs = keyValueFuncs != null ? new ObjList<>(keyValueFuncs) : null;
```

- The copy holds the **same `Function` instances**, which this factory already owns and
  frees in `close()` (`Misc.freeObjList`). The pooled `IntrinsicModel` only ever *clears*
  its list (`keyValueFuncs.clear()`), never frees the functions — so there is **no
  double‑free** and no change in ownership.
- This restores the contract already honored by `FilterOnValuesRecordCursorFactory`, whose
  key‑value list parameter is `@Transient` and is consumed at construction rather than retained.

Minimal, additive, non‑breaking: one logical line in production code plus a regression test.

## Test

`PostingIndexConcurrentQueryNpeTest` recreates the issue's schema (posting/covering index
with an `INCLUDE` list), inserts ~250 rows, and runs the reported window‑over‑`UNION ALL`
query concurrently. It also:

- asserts the query is actually served by the covering index (`CoveringIndex` in the plan),
  so the buggy path is genuinely exercised; and
- asserts the multi‑key IN‑list covering scan still matches the `/*+ no_covering */`
  reference results (proves the defensive copy preserves behavior).

### Verification (compiled against this tree)

| Code | Result |
| --- | --- |
| **before** the fix | reproduces the exact NPE on essentially every run (verified 9/9) |
| **after** the fix | passes reliably, ~1–2s (verified 11/11) |

Red before, green after.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/table/CoveringIndexRecordCursorFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/table/CoveringIndexRecordCursorFactory.java
@@ -141,13 +141,18 @@ public class CoveringIndexRecordCursorFactory implements RecordCursorFactory {
         int[] requiredIncludeIndices = buildRequiredIncludeIndices(queryColToIncludeIdx);
 
         int[] symInclCols = findSymbolIncludeCols(queryColToIncludeIdx, metadata);
-        if (keyValueFuncs != null) {
-            this.resolvedKeys = new IntList(keyValueFuncs.size());
-            int multiKeyCapacity = keyValueFuncs.size();
+        // Read the owned defensive copy (this.keyValueFuncs), never the pooled parameter. The two are
+        // content-identical here (compiling thread owns the model exclusively during construction), but
+        // the copy is the reference this factory owns and frees in close(); using it consistently avoids
+        // a refactor hazard if the defensive copy above is ever changed or removed.
+        final ObjList<Function> keyValueFuncsCopy = this.keyValueFuncs;
+        if (keyValueFuncsCopy != null) {
+            this.resolvedKeys = new IntList(keyValueFuncsCopy.size());
+            int multiKeyCapacity = keyValueFuncsCopy.size();
             if (reader != null) {
                 SymbolMapReader smr = reader.getSymbolMapReader(indexColumnIndex);
-                for (int i = 0, n = keyValueFuncs.size(); i < n; i++) {
-                    Function f = keyValueFuncs.getQuick(i);
+                for (int i = 0, n = keyValueFuncsCopy.size(); i < n; i++) {
+                    Function f = keyValueFuncsCopy.getQuick(i);
                     int key = f.isRuntimeConstant() ? SymbolTable.VALUE_NOT_FOUND : smr.keyOf(f.getStrA(null));
                     resolvedKeys.add(key);
                 }
--- a/core/src/main/java/io/questdb/griffin/engine/table/CoveringIndexRecordCursorFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/table/CoveringIndexRecordCursorFactory.java
@@ -128,7 +128,16 @@ public class CoveringIndexRecordCursorFactory implements RecordCursorFactory {
         this.latestBy = latestBy;
         this.latestByFilter = latestByFilter;
         this.queryColToIncludeIdx = queryColToIncludeIdx;
-        this.keyValueFuncs = keyValueFuncs;
+        // Defensive copy. The caller passes intrinsicModel.keyValueFuncs, which is a
+        // POOLED ObjList owned by the compiler's WhereClauseParser (ObjectPool<IntrinsicModel>).
+        // SqlCompilers are pooled and shared across threads/connections, so when another
+        // thread borrows the same compiler and recompiles, models.next() -> IntrinsicModel.clear()
+        // -> keyValueFuncs.clear() nulls the backing array (Arrays.fill BEFORE pos=0). A concurrent
+        // getCursor() on this still-cached factory would then read a stale size() (> 0) and a null
+        // slot in Function.init(...), producing the intermittent NPE in issue #7294. We keep our own
+        // list of the same Function instances -- which this factory owns and frees in close() (the
+        // pooled model only clears references, never frees) -- to decouple from the model's lifecycle.
+        this.keyValueFuncs = keyValueFuncs != null ? new ObjList<>(keyValueFuncs) : null;
         int[] requiredIncludeIndices = buildRequiredIncludeIndices(queryColToIncludeIdx);
 
         int[] symInclCols = findSymbolIncludeCols(queryColToIncludeIdx, metadata);
--- a/core/src/test/java/io/questdb/test/cairo/covering/PostingIndexConcurrentQueryNpeTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/covering/PostingIndexConcurrentQueryNpeTest.java
@@ -1,0 +1,172 @@
+/*+*****************************************************************************
+ *     ___                  _   ____  ____
+ *    / _ \ _   _  ___  ___| |_|  _ \| __ )
+ *   | | | | | | |/ _ \/ __| __| | | |  _ \
+ *   | |_| | |_| |  __/\__ \ |_| |_| | |_) |
+ *    \__\_\\__,_|\___||___/\__|____/|____/
+ *
+ *  Copyright (c) 2014-2019 Appsicle
+ *  Copyright (c) 2019-2026 QuestDB
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ ******************************************************************************/
+
+package io.questdb.test.cairo.covering;
+
+import io.questdb.cairo.security.AllowAllSecurityContext;
+import io.questdb.cairo.sql.Record;
+import io.questdb.cairo.sql.RecordCursor;
+import io.questdb.cairo.sql.RecordCursorFactory;
+import io.questdb.griffin.SqlCompiler;
+import io.questdb.griffin.SqlExecutionContextImpl;
+import io.questdb.test.AbstractCairoTest;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Reproduces issue #7294: intermittent NullPointerException in
+ * CoveringIndexRecordCursorFactory.getCursor under concurrent queries.
+ * <p>
+ * Root cause: {@link io.questdb.griffin.engine.table.CoveringIndexRecordCursorFactory}
+ * keeps a <b>direct reference</b> to {@code IntrinsicModel.keyValueFuncs}, which is a
+ * <b>pooled</b> {@link io.questdb.std.ObjList} owned by the compiler's
+ * {@link io.questdb.griffin.WhereClauseParser} (an {@code ObjectPool<IntrinsicModel>}).
+ * SqlCompilers are pooled and shared across threads/connections. When another thread
+ * borrows the same pooled compiler and recompiles, {@code models.next()} calls
+ * {@code IntrinsicModel.clear()} -> {@code keyValueFuncs.clear()}, which runs
+ * {@code Arrays.fill(buffer, null)} <i>before</i> resetting {@code pos = 0}.
+ * <p>
+ * A concurrent {@code getCursor} on the still-cached factory calls
+ * {@code Function.init(keyValueFuncs, ...)} which reads {@code size()} (stale, &gt; 0)
+ * and then {@code getQuick(i)} (already nulled), producing exactly the reported:
+ * <pre>
+ * java.lang.NullPointerException: Cannot invoke
+ *   "io.questdb.cairo.sql.Function.init(...)" because the return value of
+ *   "io.questdb.std.ObjList.getQuick(int)" is null
+ * </pre>
+ * <p>
+ * The query mirrors the issue: a window function over a UNION ALL whose first branch
+ * is an IN-list scan served by the posting/covering index.
+ */
+public class PostingIndexConcurrentQueryNpeTest extends AbstractCairoTest {
+
+    // Window function over a UNION ALL whose first branch is an IN-list covering-index scan.
+    private static final String FAILING_QUERY = "SELECT ts, last_value(v) ignore nulls over (order by ts) AS lv\n" +
+            "FROM (\n" +
+            "  SELECT ts, 1 AS v\n" +
+            "  FROM signals_boolean\n" +
+            "  WHERE name IN ('signal.a','signal.b','signal.c','signal.d','signal.e')\n" +
+            "    AND ts <= cast(1781184000000000 as timestamp)\n" +
+            "  UNION ALL SELECT cast(1781042400000000 as timestamp), cast(null as int)\n" +
+            ")";
+
+    @Test(timeout = 120_000)
+    public void testConcurrentWindowOverUnionAllWithCoveringIndex() throws Exception {
+        assertMemoryLeak(() -> {
+            // Schema from the issue (TTL/DEDUP omitted -- irrelevant to the race; the
+            // posting/covering index with an INCLUDE list is what matters).
+            execute("CREATE TABLE signals_boolean (" +
+                    "  ts TIMESTAMP," +
+                    "  name SYMBOL INDEX TYPE POSTING INCLUDE (source, value, ts)," +
+                    "  source SYMBOL," +
+                    "  value BOOLEAN" +
+                    ") TIMESTAMP(ts) PARTITION BY DAY");
+
+            // ~250 rows across the 5 signal.* keys used by the IN list.
+            execute("INSERT INTO signals_boolean " +
+                    "SELECT (1781000000000000 + x * 100000000)::timestamp, " +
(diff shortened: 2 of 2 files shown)


bluestreak01:
## Review verdict: Approve (on the merits)

Reviewed at level 3 (full pass: change-surface map, parallel review agents, per-finding verification). The fix is correct, minimal, and complete. Posting as a comment since GitHub blocks formal self-approval.

**Root cause + fix verified**
- The race is real and exactly diagnosed: `CoveringIndexRecordCursorFactory` aliased `intrinsicModel.keyValueFuncs`, a list pooled by the compiler's `WhereClauseParser`. A sibling thread recompiling on the same pooled compiler runs `IntrinsicModel.clear() -> keyValueFuncs.clear()`, which `Arrays.fill(buffer, null)`s before resetting `pos`, so a concurrent `getCursor()` reads a stale `size() > 0` and a null slot in `Function.init(...)` -> NPE. The defensive copy decouples the factory from the pooled list and covers both `getCursor` and `getPageFrameCursor` via the single constructor change.
- No double-free / no new leak: the copy holds the same Function instances; `SqlCodeGenerator` never frees `keyValueFuncs` and `IntrinsicModel.clear()` only nulls references, so the factory's `close()` is the sole freer. The fix also removes a latent double-free (a recompile could previously refill the shared list and the stale factory's `close()` would free the next query's functions).
- Complete: `CoveringIndex` was the only factory aliasing the pooled list. Siblings already copy-and-own (`LatestByDeferredListValuesFilteredRecordCursorFactory`) or are `@Transient` (`FilterOnValues`/`FilterOnExcludedValues`). The fix matches the established precedent.
- Performance: compile-path O(k) arraycopy, once per cached factory; never per row.

**Non-blocking polish (optional)**
- Title could repeat the verb / lead with end-user impact, e.g. `fix(sql): fix intermittent NPE on concurrent indexed queries`.
- Add labels (Bug, SQL, regression).
- Consider annotating the `keyValueFuncs` param `@Transient` and reading `this.keyValueFuncs` (not the param) after the copy for clarity.
- Test exercises `getCursor` only (not `getPageFrameCursor`); both are covered by the single fix, so this is a coverage gap, not a correctness gap. Worth confirming the test is RED on the parent commit with the fix reverted, so it's trusted as a regression guard.

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/CoveringIndexRecordCursorFactory.java|7|7|100.00%|

core/src/main/java/io/questdb/griffin/engine/table/CoveringIndexRecordCursorFactory.java
```diff
@@ -128,17 +128,31 @@ public CoveringIndexRecordCursorFactory(
         this.latestBy = latestBy;
         this.latestByFilter = latestByFilter;
         this.queryColToIncludeIdx = queryColToIncludeIdx;
-        this.keyValueFuncs = keyValueFuncs;
+        // Defensive copy. The caller passes intrinsicModel.keyValueFuncs, which is a
```
puzpuzpuz, line 131: nit: I'd either remove this comment and add `@Transient` annotation for `keyValueFuncs` or make it much more compact.

puzpuzpuz:
## Code review (level 3)

**Verdict: approve.** The fix is correct, complete, and follows an established codebase precedent. No Critical or blocking issues. A few minor/moderate items below are at the author's discretion.

### What the fix does (verified)

The factory stored a direct reference to `intrinsicModel.keyValueFuncs` — a **pooled** `ObjList<Function>` owned by the compiler's `WhereClauseParser` (`ObjectPool<IntrinsicModel>`). `SqlCompiler`s are pooled and shared across threads. When a sibling thread borrows the same pooled compiler and recompiles, `IntrinsicModel.clear()` → `keyValueFuncs.clear()` runs `Arrays.fill(buffer, null)` **before** `pos = 0` (`ObjList.java:105-109`), so a concurrent `getCursor()` reads a stale `size()>0` and a nulled slot → the reported NPE in `Function.init`. The defensive copy decouples the factory from the pooled list. Verified:

- **The list copy is sufficient.** The `Function` instances inside are freshly `new`-allocated per compile (`StrConstant` for literals; `IndexedParameterLinkFunction`/`NamedParameterLinkFunction` for bind vars, via `FunctionParser.createBindVariable`). `IntrinsicModel.clear()` only nulls list slots, never frees or mutates the function objects, so a concurrent recompile allocates *different* instances and the copy's referenced objects are never touched.
- **No double-free / use-after-free.** Ownership is unchanged — the pre-PR `close()` already did `Misc.freeObjList(keyValueFuncs)`; the copy holds the same instances and frees each exactly once. (It also incidentally hardens `close()`, which pre-fix could have freed a sibling thread's functions out of the cleared pooled list.)
- **Both runtime paths fixed.** `getCursor` (196-203), `getPageFrameCursor` (282-290), `toPlan` (380-381), and `close` (179) all read the field, which is now the copy. `keyValueFuncs` is a parameter only in the constructor.
- **Out-of-diff blast radius is clean.** Every other consumer of `intrinsicModel.keyValueFuncs` in `SqlCodeGenerator` is safe: `FilterOnValuesRecordCursorFactory` and `LatestByValuesIndexedFilteredRecordCursorFactory` mark the param `@Transient` and consume it at construction; `LatestByDeferredListValuesFilteredRecordCursorFactory` already does the **identical** `new ObjList<>(...)` copy (lines 70-71). `CoveringIndex` was the sole outlier — this PR brings it in line with the established pattern.

### Moderate

**1. Test covers only the record-cursor path; the parallel GROUP BY page-frame path (`getPageFrameCursor`) has no regression coverage.** The same constructor copy repairs both `getCursor` *and* `getPageFrameCursor` (lines 282-290 had the identical unguarded `Function.init(keyValueFuncs, ...)` pre-fix). The window-over-`UNION ALL` query forces the serial record-cursor path; the page-frame path (parallel/async GROUP BY) is never exercised under concurrency. The production code there is provably fixed by the same mechanism, so this is **not a shippable bug** — but half the repaired surface lacks a regression test. Consider adding a concurrent variant that routes through the page-frame path, e.g. `SELECT name, count() FROM signals_boolean WHERE name IN ('signal.a','signal.c','signal.e') GROUP BY name`, asserting the plan still shows `CoveringIndex`.

### Minor

**2. New test SQL omits the numeric-underscore convention.** Numbers ≥ 5 digits should use `_` separators: `1781000000000000`, `1781184000000000`, `1781042400000000`, `100000000` → `1_781_000_000_000_000`, etc. At minimum the INSERT literals; `FAILING_QUERY` has a reproduction-fidelity argument for staying verbatim.

**3. Cast-style inconsistency in the test.** Prefer `expr::TYPE`. The INSERT uses `(...)::timestamp` (good) but `FAILING_QUERY` uses `cast(... as timestamp)` / `cast(null as int)`. Partly justified by mirroring the issue text.

**4. PR title convention.** Prefer repeating the verb and speaking to user impact rather than the internal class name — e.g. `fix(sql): fix intermittent NPE in concurrent covering-index queries`. This line becomes the squash commit / release note. (`Fixes #7294` is correctly at the top of the body; labels are appropriate.)

**5. `done.await()` is unbounded.** Backstopped by `@Test(timeout = 120_000)` and the `finally` countdown, so it can't hang CI — but a stuck thread yields an opaque 120s timeout instead of surfacing `firstError`. `done.await(60, TimeUnit.SECONDS)` + assert would fail cleaner.

**6. `keyValueFuncsCopy` local is redundant** — identical to `this.keyValueFuncs` at that point; harmless, the comment justifies it as a refactor-hazard guard.

### Dismissed after verification (false positives)

- *Double-free of shared functions* — `close()` already owned/freed them pre-PR; copy frees the same instances once.
- *Per-function-object race survives the list copy* — functions are freshly allocated per compile; `clear()` never touches the objects.
- *Residual read of the pooled param at runtime* — all runtime reads are of the field (= copy).
- *Test could flake / false-red with the fix* — the copy fully decouples per-thread factories; no shared mutable state remains.
- *Factory leaked if `getCursor` throws* — `try (RecordCursorFactory f = factory; RecordCursor cursor = f.getCursor(ctx))` still closes `f`.

### Summary

No regressions or tradeoffs. The change adds one O(n) copy of a small IN-list-sized list on the compile path — negligible, not a data path. Findings: 7 verified (all in-diff), 5 dropped as false positives; out-of-diff blast radius genuinely contained.

puzpuzpuz (approved):

core/src/main/java/io/questdb/griffin/engine/table/CoveringIndexRecordCursorFactory.java
bluestreak01: i think this is correct behaviour, in all cases the value that's passed here is reusable obj list on the intrinsic model. We cannot store a reference to that. A copy of the list is what's genuinely needed here.

puzpuzpuz: The thing is that the comment is very lengthy while the code is self-explaining. CC tends to add very verbose comments even for obvious code maybe that's because of our skill.

PR #7299, opened 2026-06-20
test(core): stop posting-index fuzz from mixing replace commits with Parquet

bluestreak01:
## What

Disable `replaceProb` in `PostingIndexO3ConcurrencyFuzzTest.testCoveringPostingParquetO3SpillFuzz` so the fuzz never combines **replace-range commits** with **Parquet partitions**.

## Why

That test was the only one enabling both `partitionToParquetProb` (0.5) and `replaceProb` (0.1).

Replace-range commits (`WAL_DEDUP_MODE_REPLACE_RANGE`) are a **materialized-view-only** operation in production: they are issued exclusively through `WalWriter.commitMatView` from `MatViewRefreshJob`, and are **never** applied to a regular WAL table. The fuzz harness simulates them on plain WAL tables via the test-only `WalWriter.commitWithParams` entry point (`FuzzRunner` lines ~392 and ~1155).

When the fuzz combined replace with Parquet conversion, a generated replace commit eventually landed on a partition that had been converted to Parquet. `TableWriter` correctly rejects this:

```
[0] commit replace mode is not supported for Parquet partitions
    at io.questdb.cairo.TableWriter.processO3Block(TableWriter.java:9457)
```

which suspends the table and trips `FuzzRunner.applyWal`:

```
java.lang.AssertionError: Table is suspended
    at io.questdb.test.cairo.fuzz.FuzzRunner.applyWal(FuzzRunner.java:414)
```

This is a **seed-dependent, flaky** failure for a state that cannot occur in production (regular tables never receive replace commits; the replace+Parquet combination is unreachable). The `TableWriter` guard is behaving correctly — the issue is purely that the fuzz models a non-production scenario.

Observed on a CI run (build 244219) of an unrelated PR:
`PostingIndexO3ConcurrencyFuzzTest.testCoveringPostingParquetO3SpillFuzz` → `AssertionError: Table is suspended`.

## Change

`replaceProb`: `0.1` → `0.0` in the Parquet test only. Replace and Parquet are now mutually exclusive across the suite:

| Test | partitionToParquetProb | replaceProb |
|------|------------------------|-------------|
| `testCoveringPostingO3NativeSpillFuzz` | 0.0 | 0.15 |
| `testCoveringPostingParquetO3SpillFuzz` | 0.5 | **0.0** (was 0.1) |
| `testCoveringPostingSquashSpillFuzz` | 0.0 | 0.2 |

Native-partition replace coverage is fully retained in the two native tests. No production code changed.

### DIFF
--- a/core/src/test/java/io/questdb/test/cairo/fuzz/PostingIndexO3ConcurrencyFuzzTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/fuzz/PostingIndexO3ConcurrencyFuzzTest.java
@@ -134,7 +134,16 @@ public class PostingIndexO3ConcurrencyFuzzTest extends AbstractFuzzTest {
                 0.1,
                 0.0,
                 0.8,
-                0.1,   // replaceProb
+                0.0,   // replaceProb -- DISABLED: replace-range commits are a mat-view-only
+                       //   operation in production (WalWriter.commitMatView via
+                       //   MatViewRefreshJob); they are never issued against a regular WAL
+                       //   table. With partitionToParquetProb>0 above, enabling replace here
+                       //   makes the fuzz apply a replace commit onto a Parquet partition --
+                       //   an unsupported, production-unreachable state that suspends the
+                       //   table ("commit replace mode is not supported for Parquet
+                       //   partitions"). Replace and Parquet must stay mutually exclusive;
+                       //   native-partition replace coverage lives in
+                       //   testCoveringPostingO3NativeSpillFuzz and testCoveringPostingSquashSpillFuzz.
                 0.0,
                 0.01,
                 0.1,   // setParquetEncodingProb
--- a/core/src/test/java/io/questdb/test/cairo/fuzz/PostingIndexO3ConcurrencyFuzzTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/fuzz/PostingIndexO3ConcurrencyFuzzTest.java
@@ -135,15 +135,15 @@ public class PostingIndexO3ConcurrencyFuzzTest extends AbstractFuzzTest {
                 0.0,
                 0.8,
                 0.0,   // replaceProb -- DISABLED: replace-range commits are a mat-view-only
-                       //   operation in production (WalWriter.commitMatView via
-                       //   MatViewRefreshJob); they are never issued against a regular WAL
-                       //   table. With partitionToParquetProb>0 above, enabling replace here
-                       //   makes the fuzz apply a replace commit onto a Parquet partition --
-                       //   an unsupported, production-unreachable state that suspends the
-                       //   table ("commit replace mode is not supported for Parquet
-                       //   partitions"). Replace and Parquet must stay mutually exclusive;
-                       //   native-partition replace coverage lives in
-                       //   testCoveringPostingO3NativeSpillFuzz and testCoveringPostingSquashSpillFuzz.
+                //   operation in production (WalWriter.commitMatView via
+                //   MatViewRefreshJob); they are never issued against a regular WAL
+                //   table. With partitionToParquetProb>0 above, enabling replace here
+                //   makes the fuzz apply a replace commit onto a Parquet partition --
+                //   an unsupported, production-unreachable state that suspends the
+                //   table ("commit replace mode is not supported for Parquet
+                //   partitions"). Replace and Parquet must stay mutually exclusive;
+                //   native-partition replace coverage lives in
+                //   testCoveringPostingO3NativeSpillFuzz and testCoveringPostingSquashSpillFuzz.
                 0.0,
                 0.01,
                 0.1,   // setParquetEncodingProb


bluestreak01:
## Review (level 3, adversarial / mission-critical pass)

Test-only change: one positional argument (`replaceProb`, position 17) flipped `0.1 → 0.0` in `PostingIndexO3ConcurrencyFuzzTest.testCoveringPostingParquetO3SpillFuzz`, plus an explanatory comment. No production code touched.

### Verdict: Approve

The change is correct, minimal, and conservative. Its central claim — that replace-range + Parquet is production-unreachable — holds under verification, and the flake is fully closed across the suite. All findings below are documentation/metadata polish; none block merge.

### Critical
None.

### Moderate

**1. PR has zero labels — add them.** CLAUDE.md requires labels consistent with the PR. This fixes a self-described "seed-dependent, flaky" fuzz test, so it needs at minimum `Flaky Test`, plus `Core` (the `test(core)` scope) and `WAL` (exercises WAL replace commits); `Materialized View` is defensible since the rationale hinges on MV-only replace semantics.

### Minor

**2. PR body + commit message mislabel `WalWriter.commitWithParams` as "test-only" — it is production.** The body says replace commits are "simulated on plain WAL tables via the test-only `WalWriter.commitWithParams` entry point." `commitWithParams` (`WalWriter.java:345`) is a **production** method, called from `MatViewRefreshJob.java:376, 971, 1055`. Relatedly, "issued exclusively through `WalWriter.commitMatView`" is imprecise — `commitWithParams` is the primary path (3 sites) and `commitMatView` only 1. The **code comment in the diff is accurate**; only the PR-body/commit-message prose is wrong.

**3. Body tone: bold emphasis on a number and absolute phrasing.** CLAUDE.md: "don't use superlatives or bold emphasis on numbers." The change table bolds **0.0**, and the body bolds **never** / **materialized-view-only**. De-bold, and soften "production-unreachable" / "cannot occur" to "unreachable under current production code paths" — the claim is evidence-backed but stated absolutely.

**4. Body omits the tradeoff for balance.** It stresses "Native-partition replace coverage is fully retained" but never states the negative with equal weight: after this change **no fuzz test exercises a replace commit against a Parquet partition**, so the `TableWriter` guard at line 9451 loses fuzz coverage (it remains covered only by targeted unit tests). One caveat sentence fixes it.

**5. No regression lock against re-enabling the combination.** Nothing fails if a future edit sets `replaceProb > 0` alongside `partitionToParquetProb > 0` on any fuzz test — only the inline comment and the `0.0` literal protect it. An assertion in `FuzzRunner.setFuzzProbabilities` forbidding `replaceInsertProb > 0 && (partitionToParquetProb > 0 || setTableFormatProb > 0 || createWalAsParquet)` would lock the invariant suite-wide. Optional hardening.

### Observations (non-blocking, out of scope for a test PR)

- **Defense-in-depth gap in production code.** Replace+Parquet is unreachable *only* because `compileAlterTable` calls `checkMatViewModification(tableToken)` at `SqlCompilerImpl.java:2293` before the `CONVERT PARTITION TO PARQUET` dispatch. The CONVERT-PARTITION compile site (`alterTableDropConvertDetachOrAttachPartition` / `ofConvertPartition`) and the execution path (`AlterOperation`/`TableWriter`) have **no** local `isMatView()` guard — unlike `SET FORMAT PARQUET`, which has its own at line 1680. If a future refactor routes CONVERT PARTITION around that entry check, a mat view could acquire Parquet partitions and the next replace-range refresh would suspend the view — the exact scenario this fuzz modeled. A belt-and-suspenders guard at the convert-partition site is worth considering as production hardening.
- **Residual coverage nuance.** A replace commit on a *native* partition of a *mixed-format* table (one also holding Parquet partitions) is now untested in this suite, since the native sibling tests use `partitionToParquetProb=0`. The harness can't easily restore this (replace ranges are widened ±~7h and generated up front, before partition format is known), so it's acceptable — recording for completeness.

### Downgraded (false positives, verified and dismissed)

- **"Argument-order swap in `AbstractFuzzTest` forwarding is a bug."** `AbstractFuzzTest.java:365-369` forwards `setTableFormatProb, addCoveringIndexProb` while declaring them in the opposite order; `FuzzRunner`'s terminal method declares the same swapped order — the two swaps cancel, so values land correctly by name (`addCoveringIndexProb=0.6`, `setTableFormatProb=0.0`). Fragile but correct.
- **"`replaceProb=0.0` may not fully disable replace generation."** Dismissed: `replaceInsertProb` is the sole gate (`FuzzTransactionGenerator.java:312`, `rnd.nextDouble() < replaceInsertProb`); `nextDouble() ∈ [0,1)` so `< 0.0` is always false, and both apply paths are gated by `transaction.hasReplaceRange()` (`FuzzRunner.java:393, 1155`).
- **"Flipping 0.1→0.0 shifts the RNG stream / changes seeds."** Dismissed: `nextDouble()` is called unconditionally at the decision point regardless of the threshold, so no seed divergence is introduced there.

### Summary

- **Approve.** Correct, well-scoped, test-only.
- The central "production-unreachable" claim is **verified sound**: the only production source of replace-range commits is `MatViewRefreshJob`, and mat-view tables cannot acquire Parquet partitions (blocked by `checkMatViewModification` at `SqlCompilerImpl.java:2293`; `SET FORMAT PARQUET` additionally blocked at 1680). The pre-existing comment `DedupInsertFuzzTest.java:362` ("Replace commits not yet supported with Parquet") confirms this is an established, deliberate invariant.
- The flake is **fully closed**, not merely relocated: no other fuzz test in the tree simultaneously enables Parquet conversion and replace commits.
- **Findings: 0 code bugs (in-diff or out-of-diff).** All confirmed findings are documentation/metadata (1 moderate, 4 minor) plus 2 production-hardening observations out of scope for this test PR. 3 draft concerns were verified and dropped as false positives.
- **Tradeoffs/regressions:** the only real cost is reduced fuzz coverage of the `TableWriter` replace-on-Parquet guard, which the body should acknowledge — there is no correctness regression.

<sub>🤖 Generated with the QuestDB review-pr skill (level 3).</sub>

PR #4817, opened 2024-07-24
 feat(sql): add table_storage() functionality

siddharth0815:
Exposes `table_storage()` function which gives the following results for each table : 

- tableName
- walEnabled
- totalDiskSize(including partitions)
- total partition count (excluding attached and detachable partitions)
- total row count
- partitionBy

NW: added link to relevant issue, won't close it yet: https://github.com/questdb/questdb/issues/3816

PR #7301, opened 2026-06-21
feat(sql): add all_tables_partitions() function

### LINKED ISSUE #3816: Table partitions for all tables in the QuestDB

I need to get information about partitions for all tables in QuestDB.

I'd like to have a function like `table_partitions`, for example `all_tables_partitions`, which returns information about partitions for all tables in the database. For example:
```
all_tables_partitions()
```

Alternative can be a function `tables_partitions` to take array of tables `string[]` as an argument and returning information about partitions for each table in the array. For example:
```
tables_partitions(['table_1', 'table_2', 'table_3'])
```
or
```
tables_partitions(SELECT name FROM tables())
```

Right now, I get the list of tables and use `concat` with `union all`  to create a query:
```
select concat(
  'select ''',name,''' as table_name,''',
  walEnabled,''' as walEnabled,''',
  partitionBy,''' as partitionBy,count(*) partition_count,sum(numrows) row_count,(sum(disksize)) as size_b from table_partitions(''',
  name,''') group by table_name union all')
from tables()
order by name;
```
The final query looks like this:
```
select 'table_1' as table_name,'false' as walEnabled,'DAY' as partitionBy,count(*) partition_count,sum(numrows) row_count,(sum(disksize)) as size_b from table_partitions('table_1') group by table_name union all
select 'table_2' as table_name,'false' as walEnabled,'DAY' as partitionBy,count(*) partition_count,sum(numrows) row_count,(sum(disksize)) as size_b from table_partitions('table_2') group by table_name union all
select 'table_3' as table_name,'false' as walEnabled,'DAY' as partitionBy,count(*) partition_count,sum(numrows) row_count,(sum(disksize)) as size_b from table_partitions('table_3') group by table_name
```
I think this is too tedious and complicated.

### Additional context.

_No response_

amaan14999:
Fixes #3816

### What

Adds `all_tables_partitions()` - a zero-argument function that returns the same per-partition columns as `table_partitions()` for every non-system table in the database, prefixed with a `tableName` column.

### Why

Issue #3816 requests a way to inspect partitions across all tables without manually concatenating `table_partitions()` calls per table. 
PR #4817 introduced `table_storage()`, which addresses the aggregated summary use case from the issue. This PR addresses the remaining per-partition use case: returning individual partition rows for every table, so users can filter, sort, or aggregate as needed.

### How

- New `AllTablesPartitionsFunctionFactory` in the `functions.catalogue` package, registered in `function_list.txt`
- Cursor iterates all tables via `engine.getTableTokens()`, skips system tables, opens a `TableReader` per table, and walks all partitions including detached and attachable ones
- Non-partitioned tables emit a single row with `partitionBy=NONE` and `name=default`
- Output columns (17 total): `tableName`, then the same 15 columns as `table_partitions()` (`index`, `partitionBy`, `name`, `minTimestamp`, `maxTimestamp`, `numRows`, `diskSize`, `diskSizeHuman`, `readOnly`, `active`, `attached`, `detached`, `attachable`, `hasParquetGenerated`, `isParquet`, `parquetFileSize`)
- `TIMESTAMP_MICRO` used for min/max timestamp columns to accommodate both partitioned and non-partitioned tables in a single result set
- Path lifecycle managed at the factory level, matching the pattern in `ShowPartitionsRecordCursorFactory`

### Test plan

7 test cases covering:

- Single partitioned table (DAY)
- Single non-partitioned table
- Multiple partitioned tables
- Mixed partitioned and non-partitioned tables
- Empty database (no user tables, only system tables filtered)
- `ORDER BY` on `tableName`
- `WHERE` filter on `tableName`

All related existing tests pass: `ShowPartitionsTest`, `ShowTableStorageTest`, `TablesFunctionFactoryTest`.


amaan14999:
@puzpuzpuz @nwoolmer. Hey folks. I have raised the PR to resolve #3816. Do Let me know if any changes  are required from my end. Thanks.

PR #7304, opened 2026-06-22
fix(sql): fix internal error when subtracting a literal from a timestamp

puzpuzpuz:
## Problem

Subtracting a `CHAR` or `STRING` literal from a timestamp, e.g.

```sql
SELECT * FROM t WHERE timestamp > now() - '1' LIMIT 1
```

failed with an internal error. `'1'` is a single-character `CHAR` literal; it was implicitly bound to the `LONG` operand slot of the `timestamp - long` operator (`SubTimestampFunctionFactory`, signature `-(Nl)`), but the factory
read the right operand with `getTimestamp()` instead of `getLong()`. Calling `getTimestamp()` on a `CharFunction` throws `UnsupportedOperationException`, which surfaced to users as an internal error. The crash also occurred on the interval-optimizer path (`RuntimeIntervalModel`), as in the production stack trace that motivated this change.

## Fix

`SubTimestampFunctionFactory.Func.getTimestamp()` now reads the right operand with `getLong()`, mirroring the symmetric `AddLongToTimestampFunctionFactory` (`+(NL)`), which already did this. The change is behaviour-preserving for every `LONG`-compatible operand: `TimestampFunction.getLong()` returns the timestamp, `IntFunction.getTimestamp()` returns the long, and so on. It only changes the outcome for operands whose `getTimestamp()` is unsupported.

The other timestamp arithmetic factories were audited and are already correct: `AddLongToTimestampFunctionFactory` reads its count via `getLong()`, and the `dateadd`/`datediff` factories read the stride via `getInt()` and the period via `getChar()`.

## Effects and tradeoffs

- Positive: `now() - '1'` and similar queries no longer crash; the result is now consistent with `now() + '1'` (subtracting the numeric value of the operand).
- Behaviour change: `-` and `+` now produce identical, well-typed cast errors for non-numeric literals. `now() - '3 day'` now reports `inconvertible value: '3 day' [STRING -> LONG]` instead of the previous `[STRING -> TIMESTAMP]`. This is the same message `+` has always produced.
- Bug exposure: the old `-` path silently produced a meaningless value for a valid timestamp string (e.g. `now() - '2020-01-01T00:00:00.000000Z'` computed `now_micros - parsed_micros` and reinterpreted it as a timestamp). That case is now a clean cast error, matching `+`. Any query relying on that accidental behaviour will now error instead.

## Test plan

- Added `SubTimestampFunctionFactoryTest` covering the char, string, long, null, nanosecond-timestamp, and interval-scan (`now() - '1'`) paths.
- Updated the two `IntervalFilterTest` leak tests to expect the new, consistent `STRING -> LONG` error message.
- Ran `SubTimestampFunctionFactoryTest`, `IntervalFilterTest`, `TimestampQueryTest`, `TimestampAddFunctionFactoryTest`, `TimestampDiffFunctionFactoryTest`, and `WhereClauseParserTest` locally; all
  pass.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/functions/math/SubTimestampFunctionFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/functions/math/SubTimestampFunctionFactory.java
@@ -73,8 +73,12 @@ public class SubTimestampFunctionFactory implements FunctionFactory {
 
         @Override
         public long getTimestamp(Record rec) {
+            // The right operand is a LONG (signature "-(Nl)"), so read it with getLong().
+            // Calling getTimestamp() here breaks for operands whose getTimestamp() is unsupported,
+            // e.g. a CHAR constant implicitly cast to LONG ("now() - '1'"), which threw
+            // UnsupportedOperationException. This mirrors AddLongToTimestampFunctionFactory.
             long l = left.getTimestamp(rec);
-            long r = right.getTimestamp(rec);
+            long r = right.getLong(rec);
 
             if (l != Numbers.LONG_NULL && r != Numbers.LONG_NULL) {
                 return l - r;
--- a/core/src/test/java/io/questdb/test/griffin/IntervalFilterTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/IntervalFilterTest.java
@@ -879,7 +879,7 @@ public class IntervalFilterTest extends AbstractCairoTest {
             assertException(
                     "select * from x where ts > now() - '3 day' order by ts desc",
                     0,
-                    "inconvertible value: `3 day` [STRING -> TIMESTAMP_NS]"
+                    "inconvertible value: `3 day` [STRING -> LONG]"
             );
         });
     }
@@ -899,7 +899,7 @@ public class IntervalFilterTest extends AbstractCairoTest {
             assertException(
                     "select * from x where ts > now() - '3 day'",
                     0,
-                    "inconvertible value: `3 day` [STRING -> TIMESTAMP_NS]"
+                    "inconvertible value: `3 day` [STRING -> LONG]"
             );
         });
     }
--- a/core/src/test/java/io/questdb/test/griffin/engine/functions/math/SubTimestampFunctionFactoryTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/engine/functions/math/SubTimestampFunctionFactoryTest.java
@@ -1,0 +1,106 @@
+/*+*****************************************************************************
+ *     ___                  _   ____  ____
+ *    / _ \ _   _  ___  ___| |_|  _ \| __ )
+ *   | | | | | | |/ _ \/ __| __| | | |  _ \
+ *   | |_| | |_| |  __/\__ \ |_| |_| | |_) |
+ *    \__\_\\__,_|\___||___/\__|____/|____/
+ *
+ *  Copyright (c) 2014-2019 Appsicle
+ *  Copyright (c) 2019-2026 QuestDB
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ ******************************************************************************/
+
+package io.questdb.test.griffin.engine.functions.math;
+
+import io.questdb.test.AbstractCairoTest;
+import org.junit.Test;
+
+public class SubTimestampFunctionFactoryTest extends AbstractCairoTest {
+
+    @Test
+    public void testNonNumericCharThrows() throws Exception {
+        // A non-numeric CHAR cannot be converted to LONG; the user gets a typed cast error,
+        // not an internal UnsupportedOperationException.
+        assertMemoryLeak(() -> assertException(
+                "SELECT '2020-01-01T00:00:00.000000Z'::timestamp - 'a'",
+                0,
+                "inconvertible value"));
+    }
+
+    @Test
+    public void testNowMinusCharInIntervalScan() throws Exception {
+        // Reproduces the production crash: "WHERE ts > now() - '1'" pushed into the interval
+        // optimizer (RuntimeIntervalModel) evaluated the CHAR operand via getTimestamp() and
+        // threw UnsupportedOperationException. now() is fixed via setCurrentMicros() to keep
+        // the result deterministic.
+        assertMemoryLeak(() -> {
+            execute("CREATE TABLE x (ts TIMESTAMP) TIMESTAMP(ts) PARTITION BY DAY");
+            execute("INSERT INTO x VALUES (8::timestamp), (9::timestamp), (10::timestamp), (11::timestamp)");
+            setCurrentMicros(10);
+            // now() is 10 micros, now() - '1' is 9 micros, so ts > 9 keeps rows 10 and 11.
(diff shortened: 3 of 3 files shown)


puzpuzpuz:
## Code review

The one-line fix (`right.getTimestamp(rec)` -> `right.getLong(rec)` in `SubTimestampFunctionFactory.Func.getTimestamp`) is correct and verified. The bug existed (`CharFunction.getTimestamp()` throws `UnsupportedOperationException` at `CharFunction.java:240`; `getLong()` -> `castCharToNumber` returns the digit value), the fix aligns `-` with the long-standing `+` operator (`AddLongToTimestampFunctionFactory`), and all 33 tests pass locally (SubTimestampFunctionFactoryTest 7/7, IntervalFilterTest 26/26 across MICRO+NANO). No correctness, leak, or concurrency defects found. The findings below are about disclosure accuracy and test completeness, not shipped-behavior correctness.

### Critical
None.

### Moderate

**M1 - The description and commit body overstate behavior-preservation.** Both claim the change is *"behaviour-preserving for every LONG-compatible operand."* That is false for two reachable operand types, because the change swaps which base-class getter runs and several getters diverge:

- **DATE constant** (`SELECT ts - '2020-01-01'::date`): `DateFunction.getLong()` returns **raw millis** (`DateFunction.java:153-155` -> `getDate()`), but the old `getTimestamp()` returned **micros** (`DateFunction.java:218-221` -> `MicrosTimestampDriver.fromDate`, i.e. x1000). This is a silent 1000x result change for a query that did not crash before. DATE->LONG is a valid overload (`ColumnType.java:902`, row 7) and a date literal is a constant, so it binds to the `-(Nl)` slot. Mentioned nowhere in the PR.
- **VARCHAR constant** (`ts - '1000000'::varchar`): `getLong()` parses as long (`VarcharFunction.java:163-165`), old `getTimestamp()` parsed as a nanos timestamp (`:246-248`). Same class of change as STRING, but only STRING is disclosed in "Effects and tradeoffs".

Both new behaviors do match `+` (which already uses `getLong()`), so they are defensible as consistency - but they should be disclosed as behavior changes, not described as behavior-preserving. Suggest correcting that sentence in the body and commit message and listing the DATE and VARCHAR cases alongside STRING.

**M2 - Test gaps for the operand types the fix actually changes.** The new test only exercises CHAR (and LONG/NULL no-op cases). It does not cover the other types the fix changes - including two that are crash repairs in the exact same bug family as the headline CHAR fix:

- **BYTE** (`ts - 1::byte`) - old threw `UnsupportedOperationException` (`ByteFunction.java:213-214`), now returns the value. Untested.
- **SHORT** (`ts - 1::short`) - old threw `UnsupportedOperationException` (`ShortFunction.java:213-214`), now returns the value. Untested.
- **DATE** (`ts - '...'::date`) - the silent x1000 change from M1. Highest-value missing test: it locks in a behavior that changed silently; without a pinned test a future refactor can flip it again unnoticed.
- **VARCHAR** (`ts - '...'::varchar`) - parse-semantics change from M1. Untested.
- **STRING positive case** (`ts - '1000000'::string` returning a value) - only the error path is covered, and only incidentally in `IntervalFilterTest`.

Suggest adding direct `.returns(...)` tests for BYTE/SHORT/DATE/VARCHAR and a positive STRING case, so the "mirror `+`" contract is explicit rather than implicit.

### Minor

**m1 - Redundant `assertMemoryLeak` wrapper around a lone `assertQuery`** (lines 79, 87, 94, 102). The `assertQuery(...)` builder runs its own memory-leak check by default, so wrapping a single builder chain in `assertMemoryLeak(() -> ...)` double-wraps it. Drop the wrapper for the four single-statement tests (`testTimestampMinusCharConstant`, `testTimestampMinusLongConstant`, `testTimestampMinusNullChar`, `testTimestampNanoMinusCharConstant`) and call `assertQuery(...)` directly. The two interval-scan tests legitimately keep the wrapper (DDL + INSERT + assertion in one scope).

**m2 - Number literal missing underscores** (line 87). `... - 1000000` should be `... - 1_000_000` per the >=5-digit convention.

**m3 - `testTimestampMinusNullChar` is misnamed.** `- null` binds to a folded LONG null constant, not a null CHAR. The test is correct, but the name implies null-CHAR coverage that does not exist. An actual null CHAR (`cast(null as char)`) would instead throw `inconvertible value [CHAR -> LONG]` (see m4).

**m4 - Pre-existing, not regressions (noted for completeness):**
- `ts - cast(null as char)` throws `ImplicitCastException` rather than yielding NULL, because a null char is `(char)0` and `castCharToNumber` rejects it before the `LONG_NULL` guard (`CharFunction.java:281-288`). Old code threw `UnsupportedOperationException`; `+` behaves identically.
- `l - r` has no overflow guard. Identical to `+`.
- Signature asymmetry: `+(NL)` (non-const) vs `-(Nl)` (const `l`). Consequence: `ts - long_column` routes to `SubLongFunctionFactory -(LL)` and returns a plain LONG, whereas `ts + long_column` returns a TIMESTAMP. Pre-existing and unchanged; it means "mirrors `AddLongToTimestampFunctionFactory`" is true only for the runtime getter, not the signature.

### Summary

**Approve.** The fix is correct, minimal, well-commented, removes real crashes (CHAR/BYTE/SHORT), and is verified by passing tests. No blocking issues.

The two items worth addressing before merge: document the DATE/VARCHAR behavior changes (M1) and add tests for DATE/BYTE/SHORT/VARCHAR (M2). The rest are minor.

Cross-checks performed and cleared: the interval-scan regression test genuinely hits the optimizer (plan is `Interval forward scan`; both `WhereClauseParser.getTimestampFromConstFunction:285` and `RuntimeIntervalModel.getTimestamp:358` evaluate `getTimestamp` on the whole `now() - '1'` expression); `ts - long_column` is unaffected (routes to `-(LL)`); and a repo-wide grep found no other test relying on the old behavior.

<sub>Reviewed with Claude Code (`/review-pr`, level 3).</sub>

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/math/SubTimestampFunctionFactory.java|1|1|100.00%|

bluestreak01 (approved):
Approving. The one-line fix (`right.getTimestamp(rec)` -> `right.getLong(rec)` in `SubTimestampFunctionFactory.Func.getTimestamp`) is correct and verified against source: the bug was real (`CharFunction.getTimestamp()` throws `UnsupportedOperationException`; `getLong()` works), and the fix aligns `-` with the long-standing `+` operator (`AddLongToTimestampFunctionFactory`). I confirmed no new crash is introduced - every constant operand type bindable to the LONG slot (BYTE/SHORT/INT/LONG/CHAR/STRING/VARCHAR/DATE/TIMESTAMP) has a non-throwing `getLong()`. NULL handling and the interval-optimizer path are correct. No correctness, leak, or concurrency defects.

Non-blocking follow-ups (worth folding into this branch):

Moderate:
- The PR body and commit message claim the change is "behaviour-preserving for every LONG-compatible operand". That is inaccurate for two reachable constant types: DATE (`getLong()`=millis vs old `getTimestamp()`=micros, a silent 1000x change) and VARCHAR/STRING (long-count parse vs nanos-timestamp parse). `FunctionParser` inserts no cast wrapper for the LONG slot, so these raw constants reach the getter. The new behaviour matches `+` (good), but it is a behaviour change, not behaviour-preserving - please correct the wording and list DATE/VARCHAR/STRING.
- Test gaps for the operand types the fix actually changes: BYTE and SHORT constants are crash-repairs in the same family as the headline CHAR fix (their `getTimestamp()` throws, `getLong()` works) and are untested; DATE (the silent 1000x change) and a positive numeric STRING/VARCHAR case are also untested. Add direct `.returns(...)` cases so the 'mirror +' contract is explicit.

Minor:
- Four single-statement tests (`testTimestampMinusCharConstant`, `testTimestampMinusLongConstant`, `testTimestampMinusNullChar`, `testTimestampNanoMinusCharConstant`) wrap a lone `assertQuery(...).returns(...)` in `assertMemoryLeak(...)`; the builder already leak-checks by default, so the outer wrapper is redundant - drop it (keep it on the multi-statement interval-scan tests).
- No `Fixes #NNN` reference despite citing a production crash.

PR #793, opened 2021-01-28
feat(sql): now() function in select queries

ideoma:

PR #7305, opened 2026-06-22
perf(sql): speed up filters that wrap the designated timestamp in a function

### LINKED ISSUE #7279: No-op casts on designated timestamp prevent interval intrinsics and partition pruning
## Problem

A no-op cast around the designated timestamp column prevents the planner from recognizing timestamp interval intrinsics. This can turn predicates that should prune partitions into full table scans.

Example:

```sql
SELECT *
FROM trades
WHERE cast(timestamp as timestamp) in today();
```

Observed plan shape keeps the predicate as a row filter and scans frames:

```text
Async Filter workers: 46
  filter: timestamp in today()
    PageFrame
        Row forward scan
        Frame forward scan on: trades
```

Without the cast, the planner recognizes the timestamp intrinsic and uses an interval scan/partition pruning.

This also affects production-style predicates such as:

```sql
WHERE CAST(timestamp AS TIMESTAMP) >= dateadd('m', -10, now())
  AND CAST(timestamp AS TIMESTAMP) < now()
```

## Expected behavior

A cast from the designated timestamp column to the same timestamp precision is a no-op and should be treated the same as the bare column for timestamp intrinsic extraction. These should all be equivalent for pruning:

```sql
timestamp in today()
cast(timestamp as timestamp) in today()
timestamp::timestamp in today()
```

For `timestamp_ns` designated timestamps, the exact same rule should apply to `timestamp_ns`. Cross-precision casts, for example `timestamp::timestamp_ns` on a microsecond designated timestamp, should not be treated as no-op.

## Suspected cause

Timestamp intrinsic extraction runs on the raw `ExpressionNode` AST before `FunctionParser` folds no-op casts. At that stage, `CAST(timestamp AS TIMESTAMP)` is still represented as a `cast` function node, so `WhereClauseParser` does not see a literal designated timestamp column and leaves the predicate as a normal filter.

This is separate from, but related to, the no-op cast/JIT filter issue in https://github.com/questdb/questdb/pull/7048. That PR is about JIT filtering; this issue is about timestamp intrinsics and partition pruning.

## Impact

Queries over partitioned tables can lose interval scans and partition pruning when users or generated SQL wrap the designated timestamp in an explicit same-type cast. On large tables this can change a bounded time-window query into a full table scan.

kafka1991:
Fixes #7279.

## What

QuestDB prunes partitions and builds interval scans for predicates directly on the designated timestamp (`WHERE ts >= '2024-01-01'`). Until now, wrapping that timestamp in *any* function defeated this: `WHERE date_trunc('day', ts) >= '2024-01-01'`, `WHERE cast(ts AS timestamp) > $1`, or `WHERE dateadd('h', 1, ts) < now()` all fell back to a full table scan plus a row filter.

This change lets the optimizer push such predicates back onto the timestamp axis when the wrapping function is monotonic. A predicate `g(ts) OP bound` is recognized, the queried value range is inverted layer by layer back to a range on `ts`, and the result drives the usual interval scan / partition pruning.

## How

- A small `MonotonicTimestampFunction` contract: a function exposes its timestamp argument and inverts an output interval onto its input interval, returning a soundness grade.
- The grade decides what the optimizer may do:
  - **EXACT** — the inverse is exact; the interval replaces the predicate (the row filter is dropped).
  - **SUPERSET** — the inverse soundly contains the true preimage but may be wider (DST, calendar clamping); the interval prunes and the original predicate stays as a residual filter.
  - **NONE** — cannot bound; the predicate stays a plain row filter.
- The grade can depend on the bound. For a named-timezone function the zone offset is only locally constant; the inverse checks (via `getNextDST`) that no DST transition falls in a `±24h`/`±48h` window around the bound and its preimage. When the window is transition-free the offset is constant and the function collapses to a constant shift, so the inverse is **EXACT**; near a transition it stays **SUPERSET**. The runtime-deferred path probes with an unbounded interval, which always spans a transition and so stays conservative.
- `WhereClauseParser` recognizes a chain of these functions over the designated timestamp. Constant bounds are inverted at compile time; runtime bounds (`now()`, bind variables) are deferred to a `TimestampMonotonicInverter` evaluated when the scan opens, mirroring the existing `$now`/`$today` machinery.
- Recognition is fail-safe: any compile/parse failure falls back to a row filter, so a query that did not previously use this path is never broken.

## Functions covered

EXACT (interval replaces the predicate, row filter dropped):

- no-op / precision casts (`cast(ts AS timestamp[_ns])`, `ts::long`), `ts + n` / `ts - n`, `date_trunc` / `timestamp_floor` (single unit, stride, and explicit origin; fixed units), `timestamp_ceil`, `dateadd` with a fixed unit, `year`, fixed-offset `to_timezone` / `to_utc`.
- Named-timezone `to_timezone` / `to_utc`, fixed-unit `dateadd`, and local-return `timestamp_floor` — but only when the bound provably lies inside a single DST segment (no transition in the surrounding window). Fixed-offset zones (e.g. `'+02:00'`) are always EXACT.

SUPERSET (prune + keep residual filter):

- `dateadd` by month/year (calendar day clamping).
- The named-timezone functions above when the bound is near a DST transition (spring-forward gap, fall-back overlap, or a zone's first transition).
- `timestamp_floor_utc` (UTC-return) with a named timezone.

## Additional fixes (bundled)

- **`getNextDST` before a zone's first transition.** `AbstractTimeZoneRules.getNextDST(utcEpoch)` returned `Long.MAX_VALUE` ("no further transition") for any instant before a zone's first historic DST transition, instead of returning that first transition. This is a pre-existing latent bug: it is unreachable for mainstream zones (whose first transition predates 1970, before which QuestDB rejects data) but reachable for zones whose first DST transition is after 1970 (e.g. `Antarctica/Troll`, 2005). The new named-timezone EXACT path relies on `getNextDST` to detect transitions, which is what surfaced it — a query around such a transition could wrongly grade EXACT and return extra rows. Other `getNextDST` callers (`SAMPLE BY`, materialized-view interval iteration) were latently affected for these zones as well. Fixed in `dstFromHistory`.
- **Overflow hardening** of the fixed-unit `dateadd` inverse (a round-trip guard that could never fire) and the `timestamp_ceil` inverse (an `n*bucket - 1` underflow at the long extreme), plus a `timestamp == null` compile-path short-circuit and minor naming/ordering cleanups.
- **Domain-ceiling-aware overflow check for constant-shift inverses.** `dateadd('<fixed>', +stride, ts)`, `ts + n`, and a positive fixed-offset `to_timezone` / `to_utc` add the same constant to every timestamp. For `< / <= bound` (open lower, finite upper) a positive shift can wrap `ts + shift` past `Long.MAX_VALUE` for a `ts` near the type maximum, splitting the preimage into two disjoint ranges that a single interval cannot represent, so a naive prune would silently drop the wrapped rows. Rather than decline unconditionally, `shiftWrapsIntoRange` now consults the write-time domain ceiling (new `TimestampDriver.getMaxDesignatedTimestamp`, enforced by `TableWriter`/`WalWriter.validateBounds`): micros are capped at `9999-12-31`, ~284000 years below the wrap point, so no realistic stride overflows and these predicates prune exactly (the filter is dropped). Nanoseconds are not capped below the `long` range, so a positive shift is still treated as wrapping and the predicate stays a row filter. The ceiling bounds only the *innermost* shift (the one directly on the designated timestamp); an outer shift in a multi-function chain keeps the conservative decline, because an inner shift can inflate its input past the ceiling, so two per-layer-safe shifts could still sum to an overflow (`shiftInputCeiling`).

## Tradeoffs and limitations

- SUPERSET predicates still run the original row filter, so the win there is partition pruning, not filter elimination. The interval is widened by a safety margin (up to 48h for timezone cases, to cover offset jumps) — correct, but it can scan slightly more than the minimum.
- The named-timezone EXACT upgrade only helps bounds that are away from a DST transition; bounds near a transition, `timestamp_floor_utc`, and calendar `dateadd` remain SUPERSET. The realistic win is for the common case of a bound nowhere near spring-forward / fall-back.
- The actual speedup depends entirely on selectivity and partitioning: a predicate that already spans most partitions sees little benefit; a narrow range over a large table can skip most of the data.
- Recognition compiles the wrapped operand speculatively at query-compile time, a small one-off cost on the first compilation of each distinct query (compiled queries are cached); on failure it is discarded.
- Open-lower comparisons with a positive constant shift on a **nanosecond** column — `dateadd('<fixed>', +stride, ts) < bound`, `ts + n < bound`, positive fixed-offset `to_timezone` / `to_utc` — fall back to a row filter with no pruning. The nanosecond domain is not capped below the `long` range, so the forward shift can wrap for a `ts` near the type maximum, splitting the preimage; a single interval cannot represent that, so the inverse declines (see the domain-ceiling check above). The same predicates on microsecond columns prune exactly, and `>=` / `>` prune on both, so the `<` / `>=` asymmetry on nano columns is easy to miss.
- Multi-layer chains of *positive* constant shifts with an open lower bound (e.g. `dateadd('d', 1, dateadd('h', 2, ts)) < bound`) fall back to a row filter, even when the composed shift does not overflow. Only the innermost shift may use the domain ceiling; an outer shift stays conservative because a preimage split by a composed overflow cannot be represented as a single interval. Single-layer predicates (the common case) and chains where the shift sits innermost (e.g. `date_trunc('day', dateadd('d', 1, ts)) < bound`) still prune. This is a conservative correctness choice, not a wrong result.
- Deliberately not covered, falling back to a row filter: floor with a *runtime* timezone/offset, `dateadd` with a runtime stride, `cast(ts AS date)` / `cast(ts AS double)`, and calendar (month/year/week) strides for floor. These are uncommon and add risk and complexity out of proportion to their value.

## Test plan

- `MonotonicTimestampPruningTest` covers each function family across comparison, `=`, `BETWEEN`, `IN`, both operand orderings, constant and runtime bounds, EXACT vs SUPERSET, NULL bounds, empty results, and the non-prunable fall-back. New DST-edge cases for the named-timezone functions: winter/summer EXACT (filter dropped), spring-forward gap and fall-back overlap SUPERSET (filter kept, colliding rows preserved), first-transition (`Antarctica/Troll`) SUPERSET, fixed-offset floor EXACT, and `timestamp_floor_utc` SUPERSET.
- `TimeZoneRulesMicrosTest.testGetNextDSTBeforeFirstTransition` directly guards the `getNextDST` fix.
- Regression: `SampleByTest`, `TimeZoneIntervalIteratorTest`, `TimezoneFloorTimestampSamplerTest`, `TimeZoneRules{Micros,Nanos,Millis}Test`, `WhereClauseParserTest`, `ExplainPlanTest`, `SqlOptimiserTest`, and the date/cast function-factory unit tests pass — confirming the `getNextDST` fix is safe for all its callers and that interval models and plans for existing predicates are unaffected.


bluestreak01 (approved):
Approving after a thorough level-3 review (full diff read, ownership/leak lifecycle traced through RuntimeIntervalModelBuilder/IntrinsicModel/SqlCodeGenerator, inversion math verified against QuestDB's actual ceil/floor/addMonths/getNextDST semantics, compiles clean, MonotonicTimestampPruningTest 73/73 green).

Verdict: approve. The core correctness paths all hold up:
- EXACT/SUPERSET/NONE inversion contract and the "SUPERSET interval superset of the true preimage, so it may only over-scan, never wrongly prune" invariant.
- Ownership transfer via the new isOwnershipTransferred flag: build() deep-copies both lists so clear() can't corrupt a built model; clear() frees dynamic functions exactly when build() was not called; intersectMonotonicTimestamp frees the inverter when the set is already empty. No leak, no double-free.
- The getNextDST fix is strictly more correct and safe for all six callers (SAMPLE BY, mat-view iteration, the new paths); it only changes zones whose first transition is post-1970.
- Precision-cast inverses (widening/narrowing) and timestamp_ceil EXACT inverse verified exact given ceil = floor + bucket and designated-timestamp >= 0.

Non-blocking follow-ups (can be addressed here or in a later PR):
- Minor: the calendar dateadd('M'/'y') SUPERSET inverses (TimestampAddFunctionFactory and TimestampAddWithTimezoneFunctionFactory) lack the overflow guard their fixed-unit siblings have. Impact is negligible because the forward addMonths/addYears overflows at the same extreme stride, so any query that trips it is already semantically undefined; worth closing for consistency by returning NONE on overflow.
- Minor: no test exercises a runtime (now()/bind) bound on a nanosecond column through TimestampMonotonicInverter.resolveBound's precision normalization.

Metadata is clean: Conventional-Commits title with repeated verb, Fixes #7279, analytical tone with an explicit tradeoffs/limitations section, and Bug/SQL/Performance labels matching scope.

bluestreak01 (changes requested):
@kafka1991 Thorough, careful work — the EXACT/SUPERSET grading, the runtime-deferred inverter's memory ownership, and the bundled `getNextDST` fix all hold up under review, and I verified the touched suites pass locally (`MonotonicTimestampPruningTest` 73/73, `WhereClauseParserTest` 846/846 incl. the unchanged `dateadd` interval-model assertions, plus `TimeZoneRules{Micros,Nanos,Millis}Test`, `TimeZoneIntervalIteratorTest`, `TimezoneFloorTimestampSamplerTest`).

I'm requesting one change before merge: **add a randomized differential fuzz test for the timestamp-function pruning.**

### Why
The EXACT path *drops* the residual row filter, so an interval that is even slightly too tight there is silent wrong-results, and the floor-EXACT machinery (`tryFloorNamedTzExact` -> `floorPreimageShifted` + the four-point span computation) is intricate enough that hand-picked cases can miss a {zone, unit, stride, bound, precision} combination. The margin argument is sound (real zone offsets are bounded by ~+/-14h, well under the 24h/48h widening, and EXACT is gated on `getNextDST` proving a transition-free window) — a fuzz test would turn that argument into a checked invariant, which is the standard guard we use for pruning optimizations.

### What it should do
Differential check: run `SELECT ... WHERE g(ts) OP bound` and compare the result against a baseline that is forced **not** to use the pushdown (e.g. the same predicate kept as a plain row filter, or a brute-force scan), asserting identical rows. Randomize over:
- function family: `date_trunc`/`timestamp_floor` (single unit, stride, explicit origin), `timestamp_ceil`, `dateadd` (fixed + calendar units, named/fixed-offset tz), `to_timezone`/`to_utc` (named + fixed-offset), no-op/precision casts, `year`
- a pool of named zones including DST-edge ones (`Europe/London`, `Antarctica/Troll`, a Samoa-style date-line zone), fixed offsets, and bounds placed both near and far from transitions (spring-forward gap, fall-back overlap, first transition)
- comparison vs `=` vs `BETWEEN` vs `IN`, both operand orderings, constant and runtime (`now()`/bind-var) bounds, micro and nano columns, empty/NULL bounds

You can extend the existing fuzz infra under `core/src/test/java/io/questdb/test/griffin/fuzz/` or add a self-contained `assertMemoryLeak()` differential loop in `MonotonicTimestampPruningTest` with a logged seed for reproducibility — either is fine.

Everything else looks good; happy to re-approve once the fuzz coverage is in.

kafka1991:
@bluestreak01 Thanks for the thorough review! Added the differential fuzz test: `MonotonicTimestampPruningTest.testFuzzDifferentialPushdown`.

It runs `g(ts)` (pushed down) against `g(ts2)` — an identical non-designated copy the optimizer can't rewrite — and asserts identical rows. Self-contained `assertMemoryLeak()` loop with a logged seed, randomizing over all the dimensions you listed: function families (floor/ceil/`date_trunc`/`dateadd`/`to_timezone`/`to_utc`/precision casts/`year`, with chains), DST-edge zones + fixed offsets, `= BETWEEN IN`, both operand orderings, constant/`now()`/bind/NULL bounds, micro and nano columns.

The branch also bundles a precision fix in `CastTimestampToTimestampFunctionFactory`: the narrowing cast `cast(nanoTs AS timestamp)` resolves its bound in the micro domain, which drops the sub-microsecond bits the comparison still sees. It now widens the interval by a microsecond and grades SUPERSET (filter kept) instead of EXACT, so sub-microsecond bounds no longer over- or under-match; the widening micro -> nano branch stays exact.

Full `MonotonicTimestampPruningTest` (74) and `WhereClauseParserTest` (846) pass; fuzz clean across ~50 seeds.

bluestreak01 (changes requested):
@kafka1991 — adversarial level-3 review. Careful, high-quality work overall (the EXACT/SUPERSET/NONE grading, the runtime-deferred inverter ownership, the `getNextDST` fix, and the differential fuzz all hold up), but there is one **confirmed data-loss bug** plus a related one that share a single root cause. Requesting changes.

## Critical

### C1 — Unguarded round-up overflow in the EXACT floor/ceil inverses silently drops rows (filter is dropped)

The EXACT floor inverses round the upper (and the non-aligned lower) bound *up* with `ceil`/`add` and **no overflow guard**, then return `EXACT`, which drops the residual row filter. When a constant bound lies within one bucket of the timestamp domain max, the round-up wraps past `Long.MAX_VALUE` to a large negative number, the committed interval becomes garbage, and rows silently vanish.

Confirmed sites:
- `core/src/main/java/io/questdb/griffin/engine/functions/date/TimestampFloorFunctions.java:85` (`ceil.ceil(lo)`) and `:89` (`ceil.ceil(hi)`) — `date_trunc(...)` / `timestamp_floor(unit, ts)`
- `core/src/main/java/io/questdb/griffin/engine/functions/date/TimestampFloorFunctions.java:176`/`:179` (`add(floor(...), addUnit, stride)`) — `timestamp_floor(stride, ts)`
- `core/src/main/java/io/questdb/griffin/engine/functions/date/TimestampFloorOffsetFunction.java:96`/`:99` — `timestamp_floor(stride, ts, origin)`

Code-path trace (`<=`, reachable from plain SQL):
1. `analyzeMonotonicTimestamp` (`WhereClauseParser.java:1306–1323`): `intervalScratch.of(Long.MIN_VALUE, hiConst)`, `foldInvert(chain, ...)`.
2. `TimestampFloorFunction.invertTimestampInterval`: `c = ceil.ceil(hi)` overflows negative; `hi = c - 1` (negative); `io.of(LONG_NULL, negative)`; returns `EXACT`.
3. Back in the caller: `intervalScratch.getLo() > intervalScratch.getHi()` is `Long.MIN_VALUE > negative` → **false**, so it is *not* flagged empty; `model.intersectIntervals(Long.MIN_VALUE, negativeHi)` then `node.intrinsicValue = TRUE` (**row filter dropped**).

Exact arithmetic (verified numerically), nanosecond column:
```
H         = date_trunc('day','2262-04-11') nanos = 9_223_286_400_000_000_000   (representable)
ceilDD(H) = 9_223_372_800_000_000_000  > Long.MAX  ->  wraps to -9_223_371_273_709_551_616
interval  = (-inf, -9_223_371_273_709_551_617]   (NOT flagged empty: MIN_VALUE < that)
a 2024 row (1.718e18) is NOT <= that negative hi  ->  WRONGLY DROPPED
```

Concrete repros:
```sql
CREATE TABLE t (ts TIMESTAMP_NS) TIMESTAMP(ts) PARTITION BY DAY;
INSERT INTO t VALUES ('2024-06-15T12:00:00Z'), ('2262-04-10T08:00:00Z');

-- upper bound near domain max: should return BOTH rows, returns EMPTY
SELECT * FROM t WHERE date_trunc('day', ts) <= '2262-04-11';

-- '>=' near max wraps the lower bound negative -> interval [neg, +inf] returns ALL rows
-- AND drops the filter, so non-matching rows leak too
SELECT * FROM t WHERE date_trunc('day', ts) >= '2262-04-11T08:00:00.000Z';
```

Reachability: bound within one bucket of the type max. For **nanosecond** columns that is ≈ year 2262 — a plausible open-ended "max sentinel" in a `<=` / `BETWEEN` guard; for **year/month buckets the window widens to ~1 year** (e.g. `date_trunc('year', ts) <= '2262-01-01'`). Micro columns trigger at ≈ year 294247 (effectively unreachable). The differential fuzz (`testFuzzDifferentialPushdown`) uses mid-range anchors, so it cannot reach this.

This is an oversight, not by design: the sibling inverses all guard the symmetric case — `TimestampCeilFunctionFactory` added `mulOverflows`/`prod==MIN`; `Offset`/`AddLong`/`Sub`/fixed-unit `dateadd` return `NONE` on shift overflow. Only the floor sites round up unguarded.

Suggested fix: detect the round-up overflow (e.g. `if (c < hi) return NONE;` after `ceil`, and the analogous `add` overflow check) and return `NONE`, matching the sibling inverses.

## Moderate

### M2 — Calendar `dateadd('M'/'y')` and tz-`dateadd` SUPERSET inverses can mis-prune at domain extremes (same root cause)

`TimestampAddFunctionFactory.java:149`/`:152` compute the SUPERSET inverse with `addMonths`/`addYears` and **no overflow guard**:
```java
lo = periodAddFunction.add(periodAddFunction.add(lo, -stride), -1);
hi = periodAddFunction.add(periodAddFunction.add(hi, -stride), 1);
```
With a negative stride and a near-max bound, the inner `add(hi, -stride)` (positive shift) wraps past the domain max. Although SUPERSET keeps the residual filter, **partition pruning runs before the filter**, so a wrapped interval that is *not* a true superset wrongly excludes partitions → silent data loss. Example: `dateadd('y', -5, ts) <= '2260-01-01'` on a nano table computes `addYears('2260-01-01', +5) = 2265` (exceeds the nano max ≈ 2262) → wrapped `hi`. The forward function stays well-defined for those `ts`, so the "forward is also undefined" argument does not fully cover the negative-stride case. `TimestampAddWithTimezoneFunctionFactory` clamps off the *post-overflow* value and does not reliably catch a wrap. Recommend the same overflow→`NONE` hardening as C1.

## Minor

- **Member ordering** (`AbstractTimestampFloorFromOffsetFunctionFactory.java`): `floorPreimageShifted` (`:386`) placed after `floorWithTz` (`:361`) though alphabetically it precedes it; `invertFloorNamedTz`/`invertFloorSuperset` (`:184`/`:205`) sit before the private statics — confirm against the visibility-tier convention.
- **DRY / maintainability**: the unit-classification ladders are duplicated across 5 classes, and two methods named `isFixedUnit` have **divergent sets** — `TimestampAddFunctionFactory:184` includes `'w'`, `TimestampFloorOffsetFunction:117` does not. The divergence is intentional (dateadd inverts week, floor does not) but the identical name is a trap. Extract a shared helper (e.g. on `TimestampDriver`). `ceilDiv` is also duplicated in `CastTimestampToTimestampFunctionFactory` and `TimestampCeilFunctionFactory`.
- **Naming**: `WhereClauseParser.cannotAccumulateTimestampFunction` is a boolean method with a `cannot…` prefix instead of `is…/has…`.
- **Test coverage** (add alongside the C1 fix): no test near the timestamp domain max (the gap that hid C1); no deterministic test for a nano-precision **runtime** bound through `TimestampMonotonicInverter.resolveBound` (only the seeded fuzz exercises it); `timestamp_ceil`/`SubTimestamp` covered for a single operator only; the `to_timezone`/`to_utc`/cast `NONE`-on-overflow grade is untested.
- **Perf (compile-time, low)**: `f(ts) OP g(ts)` (both sides reference the timestamp) speculatively compiles each side up to twice during analysis. Bounded by 2×, compile-time only.

## Verified sound (dropped from the draft after source verification)

- **Named-tz FLOOR EXACT** (`tryFloorNamedTzExact` + 4-point span): the `getNextDST` window includes the computed preimage bounds padded by 24h ≥ max real offset (~14h); any transition inside the true preimage is detected; open probe → conservative SUPERSET. (Its `add` round-up is *accidentally* caught by the `shift` guard for typical offsets — fragile; fold into the C1 fix.)
- **`invertZoneOffsetShift`/`tryZoneOffsetExact`**: `getOffset(localBound)` is sound because EXACT requires a transition-free ±24h window and UTC↔local differ by < 24h.
- **`timestamp_ceil`, micro↔nano cast, `year()`**: math verified exact/sound; guards present; `year()` out-of-range maps to empty/unbounded (never `NONE`).
- **`getNextDST`/`dstFromHistory` cross-context**: all callers verified SAFE — SAMPLE BY (`AbstractNoRecordSampleByCursor`, `SampleByFirstLastRecordCursorFactory`) and mat-view (`TimeZoneIntervalIterator`) use it as a forward boundary; the returned value is always strictly `> utcEpoch`. The old `Long.MAX_VALUE` was a latent bug for post-1970 first-transition zones.
- **Ownership/leaks**: head/loBound/hiBound freed exactly once on every path; `isOwnershipTransferred` prevents the pooled-model double-free and the no-build leak; `close()` frees only head+bounds (chain freed transitively). No leak, no double-free.
- **Concurrency**: `AbstractTimeZoneRules` effectively immutable + `ConcurrentIntHashMap`; `WhereClauseParser` scratch is per-compile single-threaded; the `tempMonotonicChain` re-entrancy snapshot is present and correct.
- **`extract()` signature change**: all callers updated (SqlCodeGenerator ×2, WhereClauseParserTest ×5); no stale callers repo-wide.
- **Per-row cost**: no inverse touches per-row `getTimestamp`; all added work is compile-time or scan-open.

## Summary

Request changes. One **Critical** data-loss bug (C1): the EXACT `date_trunc`/`timestamp_floor` inverses overflow when a constant bound is within one bucket of the timestamp domain max (realistically reachable on nanosecond columns ≈ year 2262, widening to ~1 year for year/month buckets), dropping the row filter and silently returning empty or extra rows. One **Moderate** (M2) shares the root cause in the calendar/tz `dateadd` SUPERSET inverses. Both are fixed by the same small overflow-guard pattern (`return NONE`) the sibling inverses already use, plus a near-domain-max regression test. Everything else holds up: 8 of 11 draft findings were dropped as false positives after verification, and the bundled `getNextDST` fix is a genuine correctness improvement that is safe for all its callers.

kafka1991:
@bluestreak01 Thanks for the adversarial pass — all confirmed findings are addressed.

**C1 (EXACT floor/ceil round-up overflow):** the three floor inverses now detect the round-up wrap and return `NONE` (fall back to a row filter), the same `if (c < hi) return NONE;` guard the sibling inverses already use:
- `TimestampFloorFunction` — `ceil.ceil` on both bounds
- `TimestampFloorWithStrideFunction` and `TimestampFloorOffsetFunction` — the `add(floor(...), addUnit, stride)` paths

Verified by execution: `date_trunc('day', ts) <= '2262-04-11'` on a nanosecond column now returns both rows (was empty), and the `>=`-near-max case no longer leaks rows with the filter dropped.

**M2 (calendar `dateadd('M'/'y')` + tz-`dateadd` SUPERSET):** same overflow→`NONE` hardening — a wrapped `addMonths`/`addYears` now returns `NONE` instead of committing a non-superset interval. The tz variant keeps its margin-clamp-to-open only for legitimate (non-wrapped) near-extreme widening. This covers the negative-stride case you flagged (`dateadd('y', -5, ts) <= '2260-01-01'`, where the forward function stays well-defined).

**Tests:** added near-domain-max regressions on both micro and nanosecond columns (incl. the `2262-04-11` / `2260-01-01` repros), a deterministic nano-precision runtime bound through `TimestampMonotonicInverter.resolveBound`, plus `timestamp_ceil` / `SubTimestamp` / `dateadd('y')` coverage. `MonotonicTimestampPruningTest` 86/86, `WhereClauseParserTest` 846/846, `TimestampAdd*FunctionFactoryTest` green.

**Minors:** member ordering fixed (`floorPreimageShifted` now precedes `floorWithTz`); the `f(ts) OP g(ts)` speculative double-compile is short-circuited in `resolveScalarBound`. Two I left as-is with a note: `cannotAccumulate…` follows the existing `cannotCastTo`/`cannotResolveAliasLocally` naming precedent, and the shared `isFixedUnit`/`ceilDiv` extraction is a cross-cutting cleanup better done as a follow-up (the `'w'` divergence is intentional) — happy to fold it in if you'd prefer.

Ready for re-review.

kafka1991:
## Level-3 review

Full mission-critical pass (parallel correctness/concurrency/resource/perf/test/cross-context/adversarial agents + per-finding source verification). The branch HEAD already contains `0cc49f7275` "Guard timestamp pushdown against forward overflow" (the C1/M2 fixes), which I re-verified independently rather than trusting. This is the high-stakes class level 3 exists for: the **EXACT** grade *drops the residual row filter*, so an interval one tick too tight is silent data loss.

### Critical
**None.** I tried to break the C1/M2 fixes and could not. Independent differential checks (~110k+: `g(ts)` pushed down vs. the identical predicate kept as a row filter on a non-designated copy) across DST-edge zones, date-line skips, leap day, micro/nano columns, all operators, BETWEEN/IN, NULL/runtime bounds — zero mismatches. Manual re-derivation agrees:

- **C1 fix correct & complete.** `ceil` is an *exclusive* ceiling (`ceilHH(x)=floorHH(x)+HOUR`, `ceilDD(boundary)=boundary+1day`; identity for the smallest unit), so `floor(ts) <= hi` → `ts <= ceil(hi)-1` is exact, and the new `c < lo` / `c <= hi` guards return `NONE` on every round-up that wraps `Long.MAX_VALUE`. All three floor sites and the ceil site are guarded.
- **M2 fix correct.** Calendar `dateadd('M'/'y')` and tz-`dateadd` SUPERSET inverses detect the wrapped `addMonths`/`addYears` (`addOverflows` / `w>=m` / `w<=m`) and return `NONE`. The negative-stride case `dateadd('y',-5,ts) <= '2260'` is caught.
- **`getNextDST`/`dstFromHistory` fix sound for all callers.** `binarySearch(SCAN_UP)` returns `-1` only when `utcEpoch < transitions[0]`, so returning `transitions[0]` (guarded by `size()>0`) preserves the strict-increase invariant every consumer (SAMPLE BY, mat-view `TimeZoneIntervalIterator`, the new EXACT/SUPERSET gates) relies on. Mainstream pre-1970-first-transition zones are unaffected.

### Moderate (non-blocking; recommend addressing here)

**M1 — The differential fuzz still never visits the timestamp domain extremes — the exact blind spot that hid C1.** `testFuzzDifferentialPushdown` uses anchors in 2005–2024 and a jitter of at most ±400 days, so it cannot reach the year-2262 / near-`Long.MAX` region. The overflow guards are locked only by the ~10 hand-written near-domain-max regressions (which are good and genuinely fail without the fix), not by the randomized harness. Three invertible functions have no near-max test at all: `OffsetTimestampFunction` (fixed-offset `to_timezone`/`to_utc` wrap), `TimestampAddConstConstVarConst` (tz-`dateadd` near max), `SubTimestampFunctionFactory`. Adding occasional domain-edge anchors/bounds to the fuzz would turn the margin argument into a checked invariant.

**M2 — DRY trap with divergent semantics in the unit-classification ladders.** This PR adds four near-identical unit predicates: `isNotFixedStrideUnit` (`AbstractTimestampFloorFromOffsetFunctionFactory.java:458`), two `isFixedAlignedUnit` (`TimestampCeilFunctionFactory.java:149`, `TimestampFloorOffsetFunction.java:127`), and `isFixedDurationUnit` (`TimestampAddFunctionFactory.java:212`). The first three test `{s,m,h,d,T,U,n}`; `isFixedDurationUnit` **silently also includes `'w'`** with no comment. `{s,m,h,d,T,U,n}` is exactly `!isFixedAlignedUnit`, yet the names differ. A future "unification" could silently break dateadd's week handling or floor/ceil's week exclusion. The constant-shift inversion body is also copy-pasted 4× (`AddLong`, `Offset`, `Sub`, fixed-unit `dateadd`) and `ceilDiv` 2×. Suggest one positively-named helper plus a documented week-inclusive variant, and a shared `MonotonicTimestampFunction.invertConstantShift(io, shift)`.

### Minor
- **Perf/consistency:** the three `AllConst*` floor classes in `AbstractTimestampFloorFromOffsetFunctionFactory` recompute `isFloorExactlyInvertible(...)` on every `invertTimestampInterval` call, whereas the two sibling offset-floor classes cache it as a constructor `final boolean`. Cache for uniformity (scan-open, low impact).
- **Perf/compile-time:** for `f(ts) OP g(ts)`, `analyzeMonotonicTimestamp` fully compiles the operand before `resolveScalarBound` rejects the ts-referencing bound — two wasted `parseFunction` compiles. A cheap `isFunc(bound) && referencesTimestamp(bound)` pre-check would avoid them. Uncommon predicate, compile-time only.
- **Defensive parity:** `resolveScalarBound`'s CONSTANT branch (`WhereClauseParser.java:2953,2966`) lacks the `b == LONG_NULL → BOUND_EMPTY` guard its function-const sibling has (`:2988`). Confirmed no functional impact (the only literal mapping there, `-9223372036854775808`, types as DOUBLE and never reaches the long-parse path); add for symmetry.
- **Naming:** `WhereClauseParser.java:1251` `boolean boundIsFunc` → `isBoundFunc` per the `is.../has...` rule.
- **Test gaps (low):** nothing locks the runtime-timezone / runtime-stride fall-back to a plain filter (safe-by-construction today, but a future refactor could slip through); no fall-back test for a non-invertible-typed *operand* like `cast(ts AS date)`.

### Out-of-diff / pre-existing (flagging, not attributing to this PR)
- **Double-`close()` on the invalid-named-timezone path:** `AbstractTimestampFloorFromOffsetFunctionFactory.newInstance` frees `timestampFunc` then throws, and `FunctionParser` frees the same arg again. Pre-existing (`ddbbd0ae60`), benign because `close()` is idempotent; worth a separate cleanup.
- **`WHERE ts < -9223372036854775808`** on a bare designated column returns all rows (the `-1` adjustment overflows `Long.MIN`→`Long.MAX`). Pre-existing since #793 (2022), untouched here.

### Downgraded (draft findings dismissed after source verification)
- *"C1/M2 still drop rows / overflow unguarded"* → fixed; guards present at every site.
- *"Reversed form `'2022-01-03' > date_trunc('day', ts)` returns opposite rows"* → false positive; `analyzeGreater` passes the const as `hiBoundNode`, plan and rows correct (disproved at runtime).
- *"`TimestampMonotonicInverter.io` / `tempMonotonicChain` race under parallel scan"* → false positive; both touched only on the single-threaded compile and scan-open paths (parser is per-thread from `SqlCompilerPool`; `evaluate` runs once at cursor-open before workers dispatch).
- *"Inverter leaks the chain / double-frees bounds when loBound==hiBound"* → false positive; chain elements are sub-functions of `head` (freed transitively), and the `hiBound != loBound` guard plus `isOwnershipTransferred` make every free exactly-once on all paths.
- *"New pushdown double-applies with the pre-existing `dateadd` `and_offset` optimizer"* → false positive; disjoint inputs (virtual-projected literal vs. direct base-table wrap), cannot both fire.

### PR metadata
Clean. Conventional-Commits title with repeated verb, `Fixes #7279` at top, level-headed tone with an explicit tradeoffs/limitations section, `Bug`/`SQL`/`Performance` labels match scope. No test-assertion violations: every new assertion uses the `assertQuery(...).returns(...)` builder; no `assertSql`, no `returnsOnce`, no redundant `assertMemoryLeak` wrapper.

### ⚠️ Process gate
The `questdb-butler` bot reported **Enterprise CI failed** on tested commit `efdf3589da`, which predates the latest fix commit `0cc49f7275` — so it may already be stale. Please **re-run enterprise CI on the current HEAD and confirm green before merge**; if it stays red, investigate rather than assume it's the stale commit or a flake.

### Summary
**Approve, pending a green enterprise-CI re-run on HEAD.** No Critical or correctness findings survived verification; the previously-flagged C1/M2 data-loss bugs are correctly and completely fixed, and the bundled `getNextDST` fix is a genuine, safe improvement. Remaining items: 2 Moderate (fuzz domain-edge coverage; the DRY / divergent-`'w'` maintainability trap) and a handful of Minor cleanups — none blocking. Substantive verified findings are all in-diff (5 Minor / 2 Moderate); the 2 out-of-diff items are pre-existing; ~a dozen draft hypotheses (leak/race/double-apply/wrong-rows) were dropped as false positives after tracing source. The main tradeoff is inherent and well-documented in the PR body: EXACT drops the filter (max win, needs exact inversion), SUPERSET only prunes partitions and over-scans by a safety margin.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) — level-3 review</sub>

bluestreak01 (approved):
Approve — independent level-3 review (full diff read at HEAD 1982304b5c, inversion arithmetic re-derived by hand, ownership/leak lifecycle traced, compiled and ran suites locally rather than trusting prior conclusions).

## Verification
- MonotonicTimestampPruningTest 96/96, WhereClauseParserTest 846/846, TimeZoneRulesMicrosTest 6/6, TimeZoneIntervalIteratorTest 40/40 — all green
- testFuzzDifferentialPushdown run 3x with fresh random seeds, zero differential mismatches
- Empty-interval semantics traced through AbstractIntervalPartitionFrameCursor (empty list -> zero-partition scan -> empty result); ownership through RuntimeIntervalModelBuilder; all six getNextDST callers

## Critical: none
The overflow data-loss guards are correct and complete at every site — floor (c<lo / c<hi), stride/offset floor (c<=hi, bound<lo), floorPreimageShifted, ceil (mulOverflows + prod==Long.MIN), and the calendar/tz dateadd SUPERSET inverses (addOverflows / w>=m / w<=m). Re-derived numerically for nano bounds near 2262; the near-max regressions genuinely fail without the guards. The EXACT-drops-the-filter invariant holds. The getNextDST/dstFromHistory fix preserves the strict-increase contract for all callers and is inert for mainstream zones.

## Moderate (non-blocking)
- dateadd('<fixed>', +stride, ts) < / <= bound now falls back to a full row filter (no partition pruning) — correct (the old pruning was unsound near the domain max), but a real perf change for a common query shape that the "Tradeoffs and limitations" section does not mention. Worth a line there.

## Minor (non-blocking)
- Numbers.ceilDiv has no guard against b==0 / a==Long.MIN_VALUE; safe for current callers but invites misuse as a shared utility.
- cannotAccumulateTimestampFunction uses a cannot... prefix rather than is.../has... (existing cannotCastTo precedent).
- The fuzz's near-Long.MAX jitter only fires for nano columns; micro domain-max is covered only by the hand-written regressions.
- f(ts) OP g(ts) compiles the operand up to twice (compile-time only).

Dropped as false positives after source tracing: runtime-NULL-bound full-scan, inverter aliasing the shared chain, ownership double-free on the FALSE-sibling path, = becoming an interval, non-identity multi-value IN mis-prune, getNextDST breaking SAMPLE BY / mat-view.

Note: the earlier stale enterprise-CI result predates the fix commits — confirm a green enterprise-CI run on the current HEAD before merge rather than assuming staleness.

bluestreak01 (changes requested):
@kafka1991 Reversing my earlier approval to request one change before merge.

## M1 — document (and, where cheap, narrow) the `dateadd(+stride) < bound` pruning loss

`dateadd('<fixed>', +stride, ts) < bound` and `<= bound` now fall back to a full row filter with no partition pruning. This is *correct* — the previous AST-rewrite (`ts < bound - stride`, filter dropped) was unsound near the domain max, since a `ts` whose `ts + stride` overflows negative satisfies the predicate but was pruned away — and `shiftWrapsIntoRange` rightly returns `NONE` for the open-lower / finite-upper / positive-shift case (`WhereClauseParserTest.testDateaddLessThan` now asserts `hasIntervalFilters()==false`). Within a single-interval model, `NONE` is the only sound choice.

The problem is the user-visible effect: for realistic data (nowhere near year 294247) the old pruning was effectively correct, so a common query like `WHERE dateadd('d', 1, ts) < '2020-01-01'` on a large table now does a full scan where it previously pruned. `>=` / `>` still prune, so this asymmetry is easy to miss.

### What I'm asking for
1. Add an explicit bullet to the PR body's "Tradeoffs and limitations" section calling out that `dateadd(+stride) < / <= bound` (open-lower comparisons with a positive fixed stride) no longer prunes and falls back to a row filter, and why (forward-overflow wrap can split the preimage). Right now the section lists the SUPERSET/timezone tradeoffs but not this one, and it's the most likely to surprise someone who relied on the old `dateadd` pushdown.
2. If it's cheap, add a one-line code comment at the `shiftWrapsIntoRange` call site (or on the method) stating that this is what drops pruning for `dateadd(+) < bound`, so a future reader doesn't "optimize" it back into an unsound single-interval prune.

No code-correctness change is required — the fallback itself is right. This is about making the tradeoff discoverable in the PR narrative and the source, given the EXACT path drops the filter and the asymmetry between `<` and `>=` is non-obvious.

Everything else in my level-3 pass holds up (overflow guards complete, getNextDST fix safe for all callers, ownership/leak lifecycle sound, fuzz clean across fresh seeds). Happy to re-approve once the tradeoff is documented.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 769 / 870 (88.39%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/MillisTimestampDriver.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/AddLongToTimestampFunctionFactory.java|5|7|71.43%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/math/SubTimestampFunctionFactory.java|5|7|71.43%|
|:large_blue_circle:|io/questdb/griffin/model/IntrinsicModel.java|3|4|75.00%|
|:large_blue_circle:|io/questdb/griffin/WhereClauseParser.java|209|250|83.60%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/TimestampCeilFunctionFactory.java|22|26|84.62%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/AbstractTimestampFloorFromOffsetFunctionFactory.java|87|101|86.14%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|55|63|87.30%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/TimestampAddWithTimezoneFunctionFactory.java|55|63|87.30%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/YearFunctionFactory.java|22|25|88.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/TimestampAddFunctionFactory.java|23|26|88.46%|
|:large_blue_circle:|io/questdb/griffin/model/RuntimeIntervalModelBuilder.java|15|17|88.24%|
|:large_blue_circle:|io/questdb/griffin/model/TimestampMonotonicInverter.java|54|59|91.53%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/cast/CastTimestampToTimestampFunctionFactory.java|22|24|91.67%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/TimestampFloorOffsetFunction.java|30|32|93.75%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/MonotonicTimestampFunction.java|65|67|97.01%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/TimestampFloorFunctions.java|78|79|98.73%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/cast/CastTimestampToLongFunctionFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/model/RuntimeIntervalModel.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/NanosTimestampDriver.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/datetime/CommonUtils.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/MicrosTimestampDriver.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/ToUTCTimestampFunctionFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/std/datetime/AbstractTimeZoneRules.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/Numbers.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/ToTimezoneTimestampFunctionFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/OffsetTimestampFunction.java|2|2|100.00%|

bluestreak01 (approved):
Approving after a level-3 review (full diff read against actual source, ownership/leak lifecycle traced through analyzeMonotonicTimestamp/TimestampMonotonicInverter/RuntimeIntervalModelBuilder, inversion math verified against ceil/floor/addMonths/getNextDST semantics, plus local runs: MonotonicTimestampPruningTest 99/99, WhereClauseParserTest 846/846, TimeZoneRulesMicrosTest 6/6, TimeZoneIntervalIteratorTest 40/40, ExplainPlanTest#testDateaddIntrinsic green).

The two data-loss bugs from the earlier thread are fixed and covered:
- C1 (unguarded ceil/add round-up in the floor/ceil EXACT inverses) - all five sites now guard the round-up (if (c < lo)/(c <= hi) return NONE), TimestampCeil guards with mulOverflows + prod==MIN. Domain-max tests (testDateTruncDayNanoNearMaxOverflow, testDateTruncDay*Overflow*, testTimestampCeilNearMaxBound, testFloorTimezoneFixedOffsetNanoNearMaxOverflow) all green.
- M2 (calendar dateadd('M'/'y') overflow) - now guarded with addOverflows on both ends and declines the open-lower/finite-upper positive-stride wrap.

Core correctness holds up:
- EXACT/SUPERSET/NONE contract: the runtime-deferred path always keeps the residual filter, and the compile-time EXACT drop is only taken where the inverse is provably exact (guarded overflow, transition-free +/-24h/48h window for named-tz, exact algebra for shift/floor/ceil/year/cast).
- Ownership via isOwnershipTransferred + the boundIsFunc chain snapshot: build() vs clear() free exactly once, intersectMonotonicTimestamp frees the inverter on an already-empty set; no leak, no double-free (testRuntimeBoundWithFalseSibling exercises it under assertMemoryLeak).
- getNextDST fix is strictly more correct and safe for all six callers (only affects instants before a zone's first transition; no infinite loop, dispatch guards the OOB at the historic tail).

Non-blocking follow-ups (fine here or later):
- Unrelated TableWriter churn (151/151 pure method movement + o3MemColumns field reorder) is scope creep in a hot file; consider dropping it.
- Naming: cannotAccumulateTimestampFunction uses a cannot... prefix instead of is.../has....
- Minor DRY on the duplicated unit->char ladders in TimestampFloorFunctions/TimestampFloorOffsetFunction.
- Worth a release note: the getNextDST fix changes SAMPLE BY / mat-view bucketing for data spanning a post-1970 first transition (e.g. Antarctica/Troll).

PR #7306, opened 2026-06-22
test(core): stabilize flaky QWP credit-flow and worker-pool CI tests

puzpuzpuz:
## Summary

This PR stabilizes two flaky tests that intermittently failed on the `mac-other` CI job while passing on every Linux and Windows run. Both are test-only changes; neither investigation found a production bug. Each fix has its own section below.

---

### Root cause

`QwpEgressCreditFlowTest.testCreditFlowCompletesLargeQuery` intermittently hung for the full 20-minute global test timeout on the `mac-other` CI job (build 244560). The hang is catastrophic slowness, not a dispatcher stall.

`setUpFragmentationChunks` draws `sendChunk`/`recvChunk` randomly from `[1, 500]`. `HttpResponseSink.sendBuffer` parks the connection once per `sendChunk` bytes (throwing `PeerIsSlowToReadException` to re-register for write), so a byte-level draw streams the test's 500k-row (~12 MB) result one byte at a time -- roughly 12 million park/resume round-trips through the network dispatcher.

- The hung run's logged seed (`2878987929496L, 1782130185058L`) deterministically produces `sendChunk=1` (recomputed with the real `Rnd`).
- The two periodic thread dumps captured 10 minutes apart show the server actively in `Net.send` (`QwpEgressUpgradeProcessor.resumeSend` -> `HttpResponseSink.sendBuffer`) and the client reading the trickle in `ioWait` -- both sides progressing, not deadlocked.
- Reproduced locally on Linux: 30k rows at `sendChunk=1` streams at ~8,670 rows/s, so the 500k-row payload takes ~58s on a fast Linux runner. On the slower, heavily oversubscribed `mac-other` runner the same work crosses the 20-minute cap.

A lost wakeup is ruled out on both sides:

- Server dispatcher: instrumented the macOS kqueue path on a prior CI run and logged 1,150 arm/fire READ events with zero arm-without-fire -- the dispatcher fired READ continuously right up to the timeout instant.
- Client receive loop (`QwpEgressIoThread`): polls with a finite 100 ms timeout and reads before waiting (`recvOrTimeout` calls `socket.recv()` first and arms a fresh kqueue READ filter only on a zero read). Any dropped wakeup self-heals within 100 ms, so it cannot hang for minutes.

The macOS-only appearance is incidental: the mechanism is platform-independent (round-trips = payload / chunk); macOS just lost the lottery first with an unlucky seed on the slowest runner.

### Fix

Floor `sendChunk`/`recvChunk` at 64 in `QwpEgressCreditFlowTest.setUp()`. Fragmentation still splits every batch into thousands of partial sends, so the partial-send and credit suspend/resume paths stay covered -- credit cycling is driven by the advertised credit budget, not the chunk size. Worst-case wall-clock drops by ~64x: the forced-`sendChunk=1` large-query case now completes in ~1s locally instead of ~58s.

Tradeoff: these large-payload credit-flow tests no longer exercise sub-64-byte fragmentation chunks. The small-payload QWP test classes (`QwpEgressFuzzTest` and the exhaustive suite) keep the full `[1, 500]` range, where a byte-level chunk finishes quickly and remains valuable edge-case coverage.

---

### Root cause

`testHaltDuringStartAddLoopIsHeldOffNotReadTorn` flaked on the same `mac-other` job (build 244884) while passing on every other run. The test is a witness for the WorkerPool safe-publish fix: it parks `start()` mid-add holding `workersLock`, fires a concurrent `halt()`, and checks that `halt()`'s first pass is held off (so worker 0 keeps running) rather than reading the half-built list and signalling worker 0.

The witness counted worker 0's free-running job-loop ticks over a fixed 500 ms wall-clock window and asserted `> 100`. The witness job always returns `false`, so the worker's idle backoff (`Os.pause` -> `Os.sleep`) governs its tick *rate*. On a slow or contended runner `Os.sleep` overshoots and the worker -- even though it is correctly held off -- ticks only a few dozen times in the window, so the count underflows 100 and the assertion flakes RED even though `halt()` was held off.

Reproduced deterministically by advancing the worker into its `Os.sleep(10)` backoff zone before measuring: `tickDelta` drops to 49 and the test fails with the exact CI assertion message and a matching ~10 s duration (CI: 10130 ms, local repro: 10156 ms). `halt()` is provably blocked on `workersLock` the entire time, so worker 0 is never signalled -- this is a witness-measurement flake, not a production bug. The safe-publish in `WorkerPool` works.

### Fix

Replace the fixed-window tick count with a poll for the tick counter to advance past a small margin (32) within a generous deadline (5 s). Held off, worker 0 clears the margin near-instantly regardless of sleep granularity; signalled (the un-fixed tree), its loop exits and the counter freezes, so the poll times out at the deadline -- RED. The margin sits well above the un-fixed tree's couple of in-flight ticks, so the two outcomes never overlap.

Tradeoff: a genuine regression now takes up to the 5 s deadline to surface in the witness instead of a fixed 500 ms, but the test already rides out a ~10 s halt timeout (only 1 of 4 workers spawns in this scenario), so green-path wall-clock is unchanged.

## Test plan

- `QwpEgressCreditFlowTest` passes locally (4 tests, a few seconds); forcing `sendChunk=1` before the floor still completes the 500k-row test in ~1s
- `WorkerPoolBootFailureTest` passes locally (8 tests); the previously-failing deep-backoff scenario now passes, and the witness still fails RED on a simulated un-fixed tree (the safe-publish removed), confirming it still catches the regression it guards
- Linux and Windows CI expected green; the previously-flaky `mac-other` job expected green


mtopolnik:
#### [PR Coverage check]

:disappointed: **fail** : 1 / 10 (10.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/network/IODispatcherOsx.java|1|10|10.00%|

bluestreak01 (approved):
Approving. Test-only stabilization of two CI flakes; both functional changes verified against source.

- QWP chunk floor [1,500]->[64,500] in QwpEgressCreditFlowTest.setUp(): JUnit runs the superclass @Before draw before the subclass @Before floor, so the floor correctly lands on top of the random draw. Credit cycling is orthogonal to the HTTP send-fragmentation chunk, so suspend/resume and credit-cycle coverage is retained; only sub-64-byte fragmentation for these large-payload tests is dropped (still covered by the fuzz/exhaustive suites). Sibling subclasses are unaffected.
- Worker-pool witness poll in testHaltDuringStartAddLoopIsHeldOffNotReadTorn: verified against Worker backoff defaults (sleepThreshold=10000, sleepMs=10, napThreshold=7000). Held-off worker 0 clears the 32-tick margin near-instantly; a signalled worker freezes and times out at 5s. The 32 margin sits well above the un-fixed tree's ~1 in-flight tick (halt signals within microseconds), so no false-GREEN. ~31x more headroom than the old 200-ticks/s fixed-window check.

Minor (non-blocking): the GREEN poll could still time out on an extremely oversubscribed runner sustaining <6.4 ticks/s for 5s; bumping the deadline to 10s would cost nothing on the green path. The sibling metrics-scrape witness still uses a 500ms wall-clock bound, but that's out of scope and its slow-runner failure mode is a false-GREEN on the un-fixed tree, not a CI flake.

PR #7308, opened 2026-06-22
fix(core): RSS pre-flight the covered-sidecar snapshot in incremental posting seal

nwoolmer:
## Summary

Follow-up to #7236. That PR added a streaming-fallback RSS pre-flight to the posting-index seal's **stride-merge buffers** (in `reencodeAllGenerations`) and to rollback — but the incremental seal still takes an **unguarded covered-column sidecar snapshot** *before* that pre-flight ever runs, so a skewed symbol with a covered (`INCLUDE`) column can still OOM and suspend the table.

### The remaining hole

`PostingIndexWriter.seal()` snapshots each covered column's entire sealed sidecar into native memory at the top of the incremental path:

```java
// PostingIndexWriter.seal()  (~line 1526)
savedSidecarBufs[c] = Unsafe.malloc(sealedLen, MemoryTag.NATIVE_INDEX_READER);
Unsafe.copyMemory(mapped, savedSidecarBufs[c], sealedLen);
```

This runs **before** `sealIncremental → reencodeAllGenerations` executes the stride pre-flight, so it is never covered by it. On a skewed symbol a single partition's covered sidecar is multi-GB, and the eager `malloc` trips the global RSS limit during a WAL fast-lag commit (the `MAX_GEN_COUNT` auto-seal inside `commit()`):

```
C TableWriter posting-index fast-lag commit failed
  io.questdb.cairo.CairoException: [-1] global RSS memory limit exceeded
  [usage=4409420634, RSS_MEM_LIMIT=7217901090, size=4331646224, memoryTag=62]
  at PostingIndexWriter.seal(PostingIndexWriter.java:1526)
  at PostingIndexWriter.flushAllPending(PostingIndexWriter.java:3236)
  at PostingIndexWriter.commit(PostingIndexWriter.java:511)
  at TableWriter.publishPostingIndexesForLastPartitionFastLag(TableWriter.java:11173)
```

(observed on a nightly built from a commit that already contains #7236 — a single 4.03 GiB snapshot on top of 4.11 GiB resident against a 6.72 GiB limit.)

### The fix

Add the **same** headroom pre-flight before the snapshot copy, re-reading `getRssMemUsed()` per cover so it accounts for snapshots already taken in this loop. When the copy would not fit the live headroom, clear the incremental candidacy and fall through to `sealFull` — which rebuilds every sidecar by streaming per-key from the column files and needs **no** snapshot (peak bounded by the largest single key: ~525–790 MiB in the production incident vs the 4+ GiB snapshot). Snapshots already taken for earlier covers are freed by the existing `finally`.

This reuses the established `isIncrementalCandidate = false; break` escape (same as the existing "snapshot not trusted" branch).

## Test

`PostingIndexOomFallbackTest#testSealIncrementalDefersToFullSealWhenSnapshotExceedsRssHeadroom` reproduces the exact production throw pre-fix (`seal()` @ 1526, `memoryTag=62`, *global RSS memory limit exceeded*) and passes post-fix — building a dense gen-0 covered sidecar, dirtying one stride to trigger the incremental seal, then sealing under a headroom below the snapshot but far above the per-key streaming peak. Adds `isLastSealSnapshotDeferredForTesting()` so the assertion pins the **snapshot** guard (distinct from the later stride pre-flight, which also clears the incremental flag).

Verified locally against current `master`:
- `PostingIndexOomFallbackTest` (22), `PostingIndexCriticalIssuesTest` (116), `PostingSealPurgeTest` (22), `PostingIndexStressTest` (55), `CoveringIndexTest` (398), `PostingIndexO3ConcurrencyFuzzTest` (3) — all green.


nwoolmer:
## Self-review (high-effort, multi-angle)

Ran a high-effort review (7 finder angles: line-by-line, removed-behavior, cross-file tracer, reuse, simplification, efficiency, altitude — each verified). **No correctness bugs found.** The deferral path is control-flow-equivalent to the existing `!isTrusted` fallback, `sealFull` has its own RSS pre-flight (so deferring can't silently re-OOM), the new field is reset consistently in `seal()`/`close()`/`of()`, and no caller branches on incremental-vs-full.

Two non-blocking cleanup/altitude items (addressed in follow-up commit):

1. **Duplicated headroom idiom** — `rssLimit > 0 ? Math.max(0, rssLimit - getRssMemUsed()) : Long.MAX_VALUE` is now hand-copied across ~4 seal/rollback sites. Extracting a single `rssHeadroom()` helper removes the risk of the copies drifting apart (a missed copy silently reintroduces the OOM this PR fixes).

2. **Altitude** — this is the 3rd–4th independent RSS guard on the seal path, each re-deriving headroom piecemeal. A *full* single up-front pre-flight isn't feasible (snapshot size is read from each cover's mmap'd sentinel; stride-merge sizes are only known after the snapshot-fed merge), but routing all sites through the shared helper collapses the duplicated derivation.

**Refuted (for transparency):** a "snapshot-then-defer wastes the malloc" candidate — verified **pre-existing** (the snapshot malloc was unconditional before this PR; the diff only wraps it in an `else`, strictly *reducing* snapshot allocations), not introduced here.

nwoolmer:
## Self-review #2 (high-effort, post-refactor)

Re-ran the high-effort multi-angle review over the full PR (both commits), focused on the `rssHeadroom()` refactor in commit 2. **No findings** — `[]`.

Verified across all 7 angles (line-by-line, removed-behavior, cross-file, reuse, simplification, efficiency, altitude):

- **Behavior-preserving at all 4 routed sites.** The helper `rssHeadroom(limit, used) = limit > 0 ? Math.max(0, limit - used) : Long.MAX_VALUE` reproduces each deleted expression exactly:
  - Snapshot guard (~1556): no-arg `rssHeadroom()` reads the same fresh `(limit, used)` pair, same left-to-right order, as the original inline ternary.
  - Rollback (~4665) and sealIncremental (~5688): both sit inside `if (limit > 0)` guards, so the helper's re-check always takes the `>0` branch — the added ternary is inert, value identical to the original bare `Math.max(0, limit - used)`.
  - reencode fast-vs-streaming (~4823): keeps the `rssLimit`/`rssUsed` locals (still consumed by the diagnostic log at ~4863), so the headroom and the logged figures share one consistent snapshot.
- **No dropped use / no orphan local.** The removed `rssLimit` local at the snapshot site was referenced only by its own ternary; the adjacent `LOG.info` logs `snapshotBytes`/`headroom`, never `rssLimit`.
- **Right home & abstraction.** No pre-existing headroom helper exists anywhere in `core/src/main` (checked `Unsafe`, `TableWriter`, `CairoConfiguration`, `MemoryUsageFormatter`); the only other `limit - used` derivation is `Unsafe.checkAllocLimit` (the allocation guard itself). With zero external re-derivers, `private static` on `PostingIndexWriter` is correct — promoting to `Unsafe` would be speculative public API. The 2-arg/0-arg split is warranted (callers that need the raw figures for diagnostics pass their own snapshot; the snapshot site wants a fresh read).
- **No read regression / no syscalls.** `getRssMemLimit`/`getRssMemUsed` are static side-effect-free `getLongVolatile` reads; per-cover invocation in the snapshot loop is *necessary* (each prior cover's snapshot malloc bumps usage), and the refactor introduces no extra volatile reads at any site.
- **Tests:** full posting + covering suite green (OomFallback 22, CriticalIssues 116, SealPurge 22, Stress 55, CoveringIndex 398).

bluestreak01 (approved):
Approve.

Verified the fix reproduces the exact production throw: disabling the new guard (`if (false)`) makes `testSealIncrementalDefersToFullSealWhenSnapshotExceedsRssHeadroom` fail with the identical incident — `global RSS memory limit exceeded ... memoryTag=62` at `Unsafe.checkAllocLimit -> malloc -> PostingIndexWriter.seal()` (the covered-sidecar snapshot copy). Restored, the test passes; full `PostingIndexOomFallbackTest` is 22/22 green and the new test is stable across 4 reruns.

Correctness checks:
- The guard `sealedLen > rssHeadroom()` fires exactly when `checkAllocLimit` would throw (`used + size > limit`) — same boundary, and `getRssMemUsed()` tracks the malloc counter so the per-cover re-read correctly accounts for snapshots already taken in the loop.
- The deferral is control-flow-equivalent to the existing `!isTrusted` fallback; `sealFull` is the reference path and produces identical sealed output (test round-trips every value).
- Resource safety: partial snapshots for earlier covers are freed by the existing `finally`; `sealFull` never takes ownership, so no leak/double-free/use-after-free on the defer path.
- The `rssHeadroom()` refactor is behavior-preserving at all 4 sites (3 guarded by `limit > 0`, 1 already had the ternary); no headroom idiom remains hand-rolled.
- Cross-context: `seal()`'s callers (TableWriter, IndexBuilder, SymbolColumnIndexer, TableSnapshotRestore) don't branch on incremental-vs-full; the `isLastSeal*` flags are test-only, so the change is transparent.

Non-blocking nits (optional):
- The test comment says "~14 MiB sealed sidecar" but the real snapshot is ~4.7 MiB (sequential LONGs ALP-compress), so the defer margin is ~382 KB rather than ~10 MiB — empirically stable, but the comment overstates it. The streaming sealFull peak here is only ~48 KB, so the headroom could be dropped to widen the margin.
- PR title/labels are implementation-focused; consider `Bug`/`Core` labels and a more user-facing description (prevents OOM/table-suspend when ingesting skewed symbols with INCLUDE columns).

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=245515)
**Tested Commit:** `3ccfdb47fb88c567630186185f2e49a6edb9bdc6`

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 27 / 27 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|27|27|100.00%|

PR #7309, opened 2026-06-22
chore(core): fix native memory leak when a sharded parallel GROUP BY trips the per-query memory limit

RaphDal:
## Summary

A high-cardinality parallel keyed GROUP BY that crosses the sharding threshold and then trips the per-query memory limit can leak a single partially-allocated shard map. `GroupByMapFragment.reopenShards()` allocated each shard's native backing before registering the shard in the fragment's `shards` list, so a reopen that threw mid-way left the half-initialized map untracked and unfreed.

This surfaced as an intermittent failure of `ParallelGroupByMemoryTrackerTest.testParallelKeyedGroupByFailsOnLargeKeySet` on the Windows `griffin-base` CI job:

```
Memory usage by tag: NATIVE_UNORDERED_MAP, difference: 3104 expected:<0> but was:<3104>
```

## Root cause

When sharding is first enabled, `reopenShards()` lazily creates 256 shard maps and opens each one, but added the map to the tracked list only *after* a successful `reopen()`:

```java
final Map shard = MapFactory.createUnorderedMap(configuration, keyTypes, valueTypes, true, false);
shard.setMemoryTracker(memoryTracker);
shard.reopen();      // allocates the shard's native backing
shards.add(shard);   // registered only after reopen() returns
```

`UnorderedVarcharMap.reopen()` performs three native allocations, in order:

| #   | allocation                       | tag                        | charged to per-query tracker? |
| --- | -------------------------------- | -------------------------- | ----------------------------- |
| 1   | `memStart` (hash table, 3072 B)  | `NATIVE_UNORDERED_MAP`     | yes                           |
| 2   | `keySink` (DirectByteSink, 32 B) | `NATIVE_UNORDERED_MAP`     | no -- global counter only     |
| 3   | allocator chunk index            | `NATIVE_GROUP_BY_FUNCTION` | yes                           |

When the per-query limit trips on allocation #3, `reopen()` throws after allocations #1 and #2 have already succeeded but before `shards.add(shard)` runs. The half-initialized shard is a local variable that never enters the `shards` list, so `GroupByMapFragment.close()` -- reached on the breach cleanup path via `frameSequence.reset()` -> `atom.clear()` -> `shardingCtx.clear()` -- never frees it.

The per-query tracker block is recycled to a pool at query end, but the global `NATIVE_UNORDERED_MAP` counter retains the orphan's bytes:

- `memStart` = entrySize(24) x initialKeyCapacity(128) = 3072 B
- `keySink`  = 32 B (DirectByteSink minimum)
- total      = 3104 B, exactly the asserted leak

(`initialKeyCapacity` is `ceilPow2(64 / 0.7) = 128`; entrySize is `align8(16-byte key + 8-byte LONG value) = 24`.)

The sibling method `GroupByShardingContext.reopenDestShard()` already does this correctly: it stores the destination shard into `destShards` *before* calling `reopen()`, so a partial reopen there is still freed at close. `reopenShards()` was the only place that registered the map after opening it.

### Why it was intermittent and surfaced on Windows

The leak occurs only when the limit trips precisely on a shard's allocator-chunk allocation (#3), after that same shard's `memStart` (#1) and `keySink` (#2) have succeeded. Because `memStart` is the first and largest tracker-charged allocation per shard, the trip usually lands there instead, in which case nothing was allocated for that shard and nothing leaks. `keySink` (#2) is global-only and never trips. Which allocation crosses the limit depends on the cumulative per-query usage at sharding time, which varies with how the owner and four worker threads interleave. The bug is therefore a timing-dependent flake that happened to land on this Windows run rather than a Windows-specific defect.

## The fix

Register each shard in the `shards` list before opening it, mirroring `reopenDestShard()`:

```java
final Map shard = MapFactory.createUnorderedMap(configuration, keyTypes, valueTypes, true, false);
shards.add(shard);
shard.setMemoryTracker(memoryTracker);
shard.reopen();
```

`GroupByMapFragment.close()` already tolerates a never-opened or partially-opened map (`Map.close()` no-ops on an unallocated `memStart` and frees whatever `keySink`/allocator backing exists), so the orphaned partial shard is now freed on the same cleanup path as a fully-opened one. The free debits both the global counter and the still-bound per-query tracker symmetrically, exactly as the existing fragment and destination-shard frees do.

`GroupByMapFragment` backs both `AsyncGroupByAtom` (parallel keyed GROUP BY) and the horizon-join atoms (`AsyncHorizonJoinAtom`, `AsyncMultiHorizonJoinAtom`), so the fix covers both. Both bind the per-query tracker the same way -- from `executionContext.getMemoryTracker()` -- so when a per-query limit is configured, either path can trip mid-shard-open and leak. When no limit applies the tracker is null, and the only trigger would then be a genuine RSS-level OOM during shard allocation.

## Tradeoffs and costs

- Removes the native memory leak on the parallel GROUP BY breach path and stabilizes the flaky test.
- No behavior change on the success path: when every shard opens cleanly, the resulting `shards` list is identical to before.
- The reordering relies on `Map.close()` being safe to call on an unopened or partially-opened map. That contract already holds and is exercised by the lazy (`openOnInit=false`) map lifecycle elsewhere; this change widens its use to the shard-open error path.
- The production change is two reordered lines plus a comment in `reopenShards()`, plus a `@TestOnly` constructor on `GroupByMapFragment` used only by the regression test; it does not alter the hot path or any allocation sizes.

## Test plan

- New `GroupByMapFragmentTest` adds a deterministic white-box regression test. `testShardOpenBreachReleasesPartialShardMap` builds one fragment, binds a `MemoryTracker` whose limit equals exactly one shard's `memStart`, and calls `shard()` on the owning thread; this forces the breach onto the shard's allocator-chunk allocation (#3 above) rather than the parallel interleaving the end-to-end test depends on. It asserts the breach surfaces with `isOutOfMemory()` and that `close()` returns the per-query counter to zero, all wrapped in `assertMemoryLeak`. A second case, `testSuccessfulShardReleasesAllShards`, opens all 256 shards under a generous limit and asserts `close()` releases every tracked byte.
- Reverting the fix makes `testShardOpenBreachReleasesPartialShardMap` fail with `expected:<0> but was:<3072>` -- the orphaned shard's `memStart`, still charged to the tracker -- and also trips the `NATIVE_UNORDERED_MAP` leak check; with the fix both new tests pass.
- The white-box test exists because the end-to-end `ParallelGroupByMemoryTrackerTest.testParallelKeyedGroupByFailsOnLargeKeySet` reproduces the leak only when the breach lands on allocation #3, a minority of runs that depends on thread interleaving. That test still passes (7 tests, 0 failures) and remains as broad coverage.
- The numbers match the analysis: the orphaned shard is a 3072 B hash table plus a 32 B key sink, so the breach leaves 3072 B on the tracker (the key sink is global-only) and 3104 B on the global `NATIVE_UNORDERED_MAP` counter -- the same 3104 B the original Windows failure reported.
- `mvnd -pl core -Dtest=GroupByMapFragmentTest,ParallelGroupByMemoryTrackerTest test` passes (9 tests, 0 failures).


RaphDal:
# Review: PR #7309 — `chore(core): fix native memory leak when a sharded parallel GROUP BY trips the per-query memory limit`

**Reviewed at level 2** (full change-surface map, parallel Agents 1–7 + 11, batched verification). Level 3 exists for the full mission-critical pass.

**Verdict: Approve.** The 2-line production fix is correct, complete, and minimal; the regression test is a genuine guard (verified by reverting → `expected:<0> but was:<3072>`). No blocking issues. One pre-existing latent fragility on the same error path is worth a decision, plus minor test-comment nits.

The fix reorders `GroupByMapFragment.reopenShards()` so `shards.add(shard)` precedes `shard.reopen()`. The leak it closes is real and reachable in production: when a sharded GROUP BY breaches the per-query limit during the 256-shard open, the half-allocated shard (memStart 3072 B + keySink 32 B) was an orphaned local never entering `shards`, so `close()` never freed it — a permanent native leak on the global `NATIVE_UNORDERED_MAP` counter, accumulating across breaches. This is the 3104 B the Windows CI flake reported.

## Critical
None.

## Moderate

### M1 — Pre-existing latent fragility on the breached path: `shards` is left inconsistent and is never cleared (currently unreachable, but unaddressed by a PR that hardens exactly this path)

*(in unchanged code — `GroupByMapFragment.close()` and the `reopenShards()` else-branch)*

`close()` resets `sharded = false` (line 120) but never clears the `shards` `ObjList` (lines 123-126 free the maps and leave them in the list). After a partial-fill breach, `shards.size()` stays at a partial `k` (1..255). If the same fragment were re-sharded, `reopenShards()` would take the else branch (`size != 0`, line 204) which assumes all 256 exist, and `shards.getQuick(k)` would throw `AssertionError` under `-ea` (`ObjList.java:189`) / NPE without it.

**This is not reachable today** (see Downgraded D1): any throw from `reopen()` is a memory breach → a non-cacheable `CairoException` → the owning factory is unconditionally closed and freed, never reused. So the fragment is destroyed, not re-entered. It is also pre-existing (before the fix `k=i`, after `k=i+1`; both partial). I flag it because the PR specifically hardens this breach path's *accounting*, yet leaves the adjacent data-structure invariant (`reopenShards` else-branch requires `shards.size()==NUM_SHARDS`) violated after the same breach, guarded only by the cache layer's close-on-OOM behavior, not by anything in `GroupByMapFragment` itself.

Options for the author (pick one — neither is required for merge):

- Clear `shards` **only on the failure path** so a re-entry rebuilds via the `size==0` branch. Do **not** clear unconditionally in `close()` — on the success path the else branch intentionally reuses the 256 retained (closed) Map objects, and clearing would force a 256-object re-create every query (minor allocation regression).
- Or add a one-line comment stating that a fragment whose `shard()` threw is never re-sharded (the factory is closed on breach), documenting the invariant the safety currently rests on.

## Minor

### m1 — `@TestOnly` constructor is a production test seam
The new public 5-arg constructor (`GroupByMapFragment.java:102-111`) exists only for the test. You have a standing preference against test seams in production. Counterweights (verified): `@TestOnly` is idiomatic here (123 production files, including constructors); the test cannot reach the package-private 8-arg constructor (tests live in `io.questdb.test.*`) and cannot deterministically force the breach via a real query (that's the nondeterministic path this PR is escaping). It's a defensible, conventional choice — raising it only so you can make the call.

### m2 — Test comment imprecisions (`GroupByMapFragmentTest.java`)
- Line 65 "Measure one shard map's memStart": `getHeapSize()` returns `(memLimit-memStart) + allocator.allocated()`. It equals memStart (3072) only because `allocator.allocated()==0` right after `reopen()` (the allocator reopens its chunk *index*, mallocs no data chunk). Correct today, but the comment would silently break if a chunk were ever pre-allocated.
- Class javadoc lists "memStart **and the key sink** under `NATIVE_UNORDERED_MAP`, then its key-arena chunk index" — accurate on tags, but implies the key sink is tracker-charged. It is global-only (`DirectByteSink.inflate` → `Unsafe.recordMemAlloc`, not the tracker), so it never trips the per-query limit. The inline comment at lines 65-68 already states this correctly; the class text could mislead.

### m3 — Test coverage gaps (non-blocking; the chosen case is the right one)
- The breach-on-memStart (#1) case (nothing allocated yet → nothing to leak, and the add-before-reopen ordering is irrelevant) is untested.
- The `if (map.size() > 0)` entry-copy block in `shard()` (lines 174-185) is never exercised here (the owner `map` is never populated); it's covered by the broader `ParallelGroupBy*` suite, not this unit.

### D1 — "Reuse-after-OOM crash is reachable in production" (initial hypothesis) — DROPPED
The owning factory is unconditionally closed on any memory breach, never cached: `CairoException.cacheable` defaults false and `setCacheable(true)` is called **only** for authorization "Write permission denied" errors (`ReadOnlySecurityContext`, `MatViewRefreshSqlExecutionContext`), never for OOM. `JsonQueryProcessor.java:751` gates caching on `ex.isCacheable()` → OOM forces `queryCacheable=false` → factory closed; pgwire `PGPipelineEntry` frees the errored factory likewise. Closing the factory frees the fragment, so the partial-`shards` instance never survives to a second `shard()`. The data-structure smell is real (kept as M1) but the crash is not reachable.

### D2 — "Tracker malloc/free desync could trip the balance assert" — DROPPED
malloc and free use the same tracker for memStart/keySink/allocator. The tracker is never nulled on the fragments/shards before their free: `AsyncGroupByAtom.clear()` frees the shards first and nulls only its own field last; `close()` frees `shardingCtx` first and nulls only the *pooled allocators* (whose backing outlives the query), deliberately. No alloc-under-T / free-under-null path exists.

### D3 — "Other allocate-before-register leak sites" — DROPPED
`reopenMap()` operates on the always-registered `map` field (plus a try/catch around `setBatchEmptyValue`); `reopenDestShard()` already registers before `reopen()`; the entry-copy loop allocates only through already-registered shards. None can orphan.

### D4 — Transient local test failure `expected:<0> but was:<3072>` — DROPPED as a build artifact
Stale `core/target/classes` made `mvnd` incremental run the test against pre-fix bytecode. A clean compile is deterministic (5/5 green). This is the known stale-`.class` footgun, not a defect. (One agent's identical-looking result *was* a deliberate revert-test and correctly proves the guard works.)

## PR metadata
- Title `chore(core): fix native memory leak ...` — correct: internal flaky-CI fix (not a customer-reported bug) → `chore`; verb repeated in the description. Reads well standalone.
- Labels Bug / SQL / Core — appropriate.
- No `Fixes #NNN` — fine; it fixes an internal flaky test, not a filed issue.
- Body is analytical and level-headed, with an accurate allocation table and an honest tradeoffs section (it documents the `@TestOnly` constructor and the reliance on `Map.close()` tolerating a partial map). Accurate — it claims a *leak* (real and reachable), not a crash.

## Summary
- **Verdict: approve.** The leak fix is correct, complete, and concurrency-safe (each fragment is single-writer via `PerWorkerLocks` slot mutex; `shards` is read only after a merge-ring barrier); zero perf/algorithmic delta; success-path `shards` list byte-identical to before. The regression test is deterministic and revert-proven.
- **Tradeoff/gap to weigh (M1):** the PR hardens the breach path's *memory accounting* but leaves the `shards`-not-cleared invariant fragility in place — harmless today only because the cache layer closes the factory on OOM. Optional to address now or document.
- **Verification:** 8 draft findings → 3 confirmed actionable (M1 nuanced, plus minor test nits), 4 dropped as false positives (D1–D4), 1 (core fix) confirmed correct. Of the actionable items, the fix and test nits are in-diff; M1 lives in unchanged code (`close()` / `reopenShards` else-branch) reachable from the changed method's contract. The diff is small and fully covered; no wider cross-context callers were left unexamined (`shard()` has 4 call sites — keyed group-by reduce/collect and two horizon-join factories — all verified to follow the same single-writer + breach-closes-factory pattern).

puzpuzpuz:
# Review (level 3): native memory leak fix for sharded parallel GROUP BY

**Verdict: request changes.** The leak fix itself is correct, minimal, concurrency-safe, and the regression test genuinely guards it (reverting the reorder fails it with `expected:<0> but was:<3072>`, and trips the `NATIVE_UNORDERED_MAP` leak check). The concern is that the PR hardens the *accounting* on the shard-open breach path while leaving a reachable *crash* on that same path, and the reorder converts the shard-index-0 breach from a leak into that crash.

## Critical

### C1 — Reuse after a shard-open breach reads `shards` out of bounds

**Where:** the crash mechanism lives in unchanged code (`reopenShards()` else-branch, `GroupByMapFragment.java:204-213`, plus `close()` never resetting `shards`, `:113-127`); the regression is introduced by the in-diff reorder at `:198-203`.

**The violated invariant:** the else-branch assumes `shards.size() == NUM_SHARDS` (it loops `i = 0..255` doing `shards.getQuick(i)`). Nothing enforces that after a partial breach. `shards` grows via `add()` (`new ObjList<>(NUM_SHARDS)`), unlike the sibling `destShards`, which is pre-sized with `setPos(NUM_SHARDS)` + `set()` and is therefore never partial (`GroupByShardingContext.java:116,270`). When `reopen()` trips the per-query limit mid-loop, `shards` is left with `1..255` entries; a later call takes the else-branch and `shards.getQuick(partialSize)` reads past the end. `ObjList.getQuick` asserts `index < pos` (`ObjList.java:189`): `AssertionError` under `-ea` (CI/dev), or a null slot then NPE in a no-`-ea` build.

**What the diff changes (the shard-0 regression):**
- breach on shard 0, **before** the fix (`add` after `reopen`): the throw precedes `add`, so `shards.size() == 0`, and a later `shard()` takes the safe `size==0` rebuild branch — no crash (just the leak this PR fixes).
- breach on shard 0, **after** the fix (`add` before `reopen`): `shards.size() == 1`, and a later `shard()` takes the else-branch -> `shards.getQuick(1)` out of bounds.

So for shard 0 the PR trades a leak for a crash. For shard >= 1 the crash is pre-existing (both orderings leave `1 <= size < 256`), but it sits unaddressed on a path this PR otherwise hardens.

**Why it is reachable in production (PGWire statement cache).** Verified each link against source:
1. A sharded keyed GROUP BY breaches the per-query limit inside `reopenShards()` (the minority-of-runs case described in the PR body; the orphaned-shard leak is the common symptom). The worker `shard()` throws; `dispatchAndAwait()` re-throws it out of cursor consumption.
2. PGWire's `outCursor` catches the `Throwable`, records the error message, and does **not** free the factory or clear `selectIsCacheable` (`PGPipelineEntry.java:2640-2666`).
3. `selectIsCacheable` is only ever assigned from `cq.isCacheable()`; a runtime OOM never resets it (`PGPipelineEntry.java:205/375/3447/3469/3481`). A plain `SELECT k, count(*) FROM t GROUP BY k` is cacheable.
4. `syncPipeline()` captures `isError` at `:1398` **before** `msgSync()` (`:1414`) sets it, then gates caching on that stale value: `if (!isError) cacheIfPossible(...)` (`:1471`).
5. `cacheIfPossible()` runs: frees the cursor (-> `atom.clear()`, which **retains** the partial `shards` list via `freeObjListAndKeepObjects`, `GroupByShardingContext.java:132-137`), sets `factory = null`, and `tasCache.put(sqlText, tas)` (`PGPipelineEntry.java:267-278`). The OOM'd factory is now cached.
6. A later execution of the same SQL reuses the cached factory -> `atom.reopen()` does not rebuild `shards` (`AsyncGroupByAtom.java:431-448`) -> reduce calls `fragment.shard()` -> `reopenShards()` sees the partial size -> else-branch -> out of bounds. The exception is caught in `outCursor` (logged `critical` as a questdb-bug/data-corruption if it surfaces as `AssertionError`) and re-cached, so the statement stays poisoned until cache eviction.

It is not a JVM abort (the assertion is caught), but it is a spurious failure of a valid query plus a poisoned statement cache. HTTP is not a vector (`JsonQueryProcessor` closes the factory when `!ex.isCacheable()`), and the within-query path is not a vector (`dispatchAndAwait()` throws before the owner reaches `mergeShards()/shardAll()`, so a fragment that breached during reduce is never re-sharded in the same query). The blast radius includes horizon joins, which bind the same tracker and drive the same `reopenShards()` (see M1).

This corrects D1 from the earlier review ("reuse-after-OOM unreachable, factory closed on OOM"): that holds for HTTP but misses the PGWire stale-`isError` caching path above. The bug was also surfaced independently by a no-context adversarial pass given only the diff.

**Recommended fix (cleanest option): split `reopenShards()`'s lazy branch into create-then-open.**

```java
if (size == 0) {
    // Phase 1: create + register + bind tracker. No native allocation here, so the
    // per-query limit cannot trip; every shard is in the list before any open.
    for (int i = 0; i < NUM_SHARDS; i++) {
        final Map shard = MapFactory.createUnorderedMap(configuration, keyTypes, valueTypes, true, false);
        shards.add(shard);
        shard.setMemoryTracker(memoryTracker);
    }
    // Phase 2: open. The only place a per-query OOM can fire; on a breach the list is
    // already full (size == NUM_SHARDS), so reuse never indexes past the end.
    for (int i = 0; i < NUM_SHARDS; i++) {
        shards.getQuick(i).reopen();
    }
}
```

This works because `createUnorderedMap(..., openOnInit=false)` is lazy and allocates nothing (`UnorderedVarcharMap.java:216-218,241-245`), and `setMemoryTracker()` only stores a reference, so the per-query limit (checked only on a tracker-charged `malloc` with `size > 0`, `Unsafe.java:615-622`) can only trip in phase 2 -- by which point all 256 shards are registered. It makes `shards.size() in {0, NUM_SHARDS}` a structural invariant rather than a violation repaired after the fact, eliminates the shard-0 regression, retains the shard objects across a breach->reuse cycle, and subsumes the current `add`-before-`reopen` reorder. (Phase 1 can still throw a deterministic type-validation `CairoException` -- but that fires on shard 0 every run, leaving `size == 0`, not a breach-dependent partial -- or a JVM-heap `OutOfMemoryError`, which is unrecoverable and equally unhandled today.) An optional `assert shards.size() == 0 || shards.size() == NUM_SHARDS;` at the top of the else-branch documents and guards the invariant under `-ea`.

## Moderate

### M1 — PR description is inaccurate about the horizon-join path

The body states the horizon-join path "does not bind a per-query tracker, so its only trigger would be a genuine RSS-level OOM." This is not the case: `BaseAsyncHorizonJoinAtom.init()` binds `memoryTracker = executionContext.getMemoryTracker()` (`:472`), and `AsyncHorizonJoinAtom.reopen()` / `AsyncMultiHorizonJoinAtom.reopen()` propagate it via `shardingCtx.setMemoryTracker(...)` (`:255` / `:237`), exactly like `AsyncGroupByAtom`. So when a per-query limit is configured the horizon-join path binds the same tracker, shared the same leak (the fix correctly covers it), and is equally exposed to C1. Worth correcting the sentence.

### M2 — No test covers reuse after a sharded breach (the C1 path)

The new `GroupByMapFragmentTest` calls `shard()` exactly once per fragment, so it never exercises the else-branch or any reuse. The existing suite does not either: `testParallelKeyedGroupByOpenFailureReleasesAllocations` re-runs 5x but with 5 keys (`x % 5`), below the sharding threshold, so it breaches on the non-sharded open and never shards; `testParallelKeyedGroupByFailsOnLargeKeySet` shards but runs once. Add a white-box case that calls `shard()` (breach), `close()`, then `shard()` again on the same fragment and asserts no `AssertionError`/NPE. Without the C1 fix it throws on the second `shard()`; with it, it passes. (The suite already tests this reuse pattern for the non-sharded case, so the sharded one is the missing sibling.)

## Minor

- **m1** — test comments and the class javadoc (`GroupByMapFragmentTest.java`, around lines 64-68 and 89) describe the key sink as if it were tracker-charged. It is global-only (`DirectByteSink` -> `Unsafe.recordMemAlloc`, never the tracker), which is exactly why `tracker.getUsed()` reflects only memStart (3072) and the arithmetic is robust. The class javadoc also references `ParallelGroupByMemoryTrackerTest` as `griffin.engine.groupby.*`, but it lives at `io.questdb.test.griffin.*`.
- **m2** — the `@TestOnly` constructor javadoc warns only against `reopenMap()`, but a second `shard()` on a test fragment also reaches the else-branch and would NPE on the null `shardStats`; this is mooted once C1 is fixed. The seam itself is fine and idiomatic.
- **m3** — title uses `chore` while the `Bug` label is applied and this fixes a real native leak reachable whenever a per-query limit trips mid-shard-open; `fix(core):` reads more accurately in release notes. Consider adding `Flaky Test` (it stabilizes a flake) and possibly `regression`. `Performance` is not warranted. The allocation arithmetic in the body (entrySize 24, keyCapacity 128, memStart 3072, keySink 32, total 3104) is correct.

## Verified correct

- **Leak fix / tracker accounting** — malloc and free use the identical bound tracker for the partial shard's `memStart`; the tracker is never nulled on a shard between open and free; `close()` is idempotent; shard maps free under the live tracker before it returns to the pool (the `QueryProgress.unregisterAndCleanup` ordering), in correct contrast to the pooled allocators which null first. The per-query counter balances to 0.
- **Concurrency** — each fragment is single-writer (per-slot `PerWorkerLocks`); the only cross-thread reads of `shards` are in the merge phase, fenced behind the `dispatchAndAwait` volatile-latch barrier. A half-opened shard is only ever observed by its owning thread.
- **Performance** — neutral. `shards` is pre-capacity'd to 256 so `add()` never reallocs; the reordered statements touch disjoint objects with no data dependency; success-path outcome is byte-identical.

## Summary

The leak fix is correct and the test guards it. The blocking item is C1: the PR leaves (and for shard 0 newly introduces) a reachable crash on the breach path it hardens, via PGWire statement-cache reuse, affecting both parallel GROUP BY and horizon joins. The two-loop split closes it structurally, subsumes the current reorder, and is small enough to belong on this branch. Also worth doing on the same branch: the M2 reuse test and the M1 description correction.

RaphDal:
@puzpuzpuz - thank you for the review, addressed your feedback in https://github.com/questdb/questdb/pull/7309/commits/e9cc18b70e9a8ad14732c0d4735c1fb9d458bf7b

puzpuzpuz:
# Re-review (level 3): two-phase shard-open fix — C1 verification

Reviewing the post-feedback state (commit `e9cc18b`). This pass verifies the two-phase split actually closes the C1 reuse crash, then sweeps for anything the restructure introduced or left behind.

**Verdict: approve.** The two-phase `create-then-open` split is correct and **fully resolves C1** for both the parallel GROUP BY and the horizon-join atoms. The new `testShardReuseAfterBreachReopensAllShards` is a genuine guard. No blocking issues remain. The items below are a description/code-comment drift (worth fixing before the squash-merge) plus minor hardening and a title nit.

## How C1 is closed (verified)

After a Phase-2 breach the `shards` list is already full (`size == NUM_SHARDS`), because Phase 1 registers all 256 shards before any `reopen()` and **Phase 1 performs zero native allocation**, so the per-query limit cannot trip there:
- `createUnorderedMap(..., openOnInit=false)` allocates nothing (`memStart==0`, `keySink` kept closed, allocator chunk index unallocated), and `setMemoryTracker()` only stores references.
- `MapFactory.createUnorderedMap` picks the map type purely from the (identical) key/value types, so it cannot succeed on shard 0 yet throw on shard *i>0*; any deterministic ctor throw (`validateBatchAddressable`, unsupported value type) fires on shard 0, before the first `add()`, leaving `size==0`.

So `shards.size() ∈ {0, NUM_SHARDS}` holds on the per-query path, and every consumer that does `getShards().getQuick(shardIndex)` (the `reopenShards()` else-branch on reuse, `GroupByShardingContext.mergeShard`, `getShardMap`) stays in-bounds. The shard-0-breach-becomes-crash regression that the simple reorder introduced is gone.

Cross-context callers of `shard()` — keyed GROUP BY reduce/`shardAll`, plus `AsyncHorizonJoinRecordCursorFactory` and `AsyncMultiHorizonJoinRecordCursorFactory` reduce — all rely on exactly this invariant and are all SAFE under it.

### Verification performed
- Independent clean build (`core`): `GroupByMapFragmentTest` — `Tests run: 3, Failures: 0`.
- Revert-test (restore the pre-fix interleaved order): `testShardOpenBreachReleasesPartialShardMap` fails `expected:<0> but was:<3072>`; `testShardReuseAfterBreachReopensAllShards` fails `expected:<256> but was:<0>` — both regressions fire, so the tests are load-bearing.
- Leak balance on the breach path: `close()` frees the partial shard, the opened shards, the never-opened shards, and the owner `map`; `Map.close()` no-ops on a never-opened lazy map; `tracker.getUsed()` returns to 0.
- Tracker malloc/free symmetry holds on both the breach path and the close-then-reuse path (the else-branch rebinds `setMemoryTracker(t2)` on closed maps before `reopen()`); atom teardown frees fragments under the live tracker before nulling the pooled allocators. No alloc-under-A/free-under-B desync.
- Concurrency: `reopenShards()` is single-writer per fragment (per-slot `PerWorkerLocks`); `shards` is read cross-thread only after the `dispatchAndAwait` barrier.
- `OrderedMap` (non-VARCHAR keys) shares the lazy/reopen/close contract, so the map-type-agnostic fix covers it too (the deterministic test exercises the `UnorderedVarcharMap` path; `OrderedMap` coverage stays indirect via the parallel suite).

## Moderate

### M1 — PR description no longer matches the merged code

The body still documents the **superseded simple reorder**, not the two-phase split that actually landed:

- The `## The fix` code block shows a single loop with `shards.add(shard); shard.setMemoryTracker(...); shard.reopen();` — i.e. the interleaved version, which is precisely the one that turns a shard-0 breach into the out-of-bounds reuse crash. The merged code is the two-loop `create+register` / `open` split.
- "The production change is two reordered lines plus a comment in `reopenShards()`" understates it: `reopenShards()` was restructured into two phases, plus a `@TestOnly newShardStats()` helper was added alongside the `@TestOnly` constructor.
- The `Root cause` / `Why it was intermittent` framing presents the leak as the whole story; there's no mention of the reuse out-of-bounds that motivated the restructure.
- The `Test plan` lists only `testShardOpenBreachReleasesPartialShardMap` and `testSuccessfulShardReleasesAllShards`; it omits `testShardReuseAfterBreachReopensAllShards`, the regression test for the reuse path.

PRs here are squash-merged, so this body becomes the permanent `master` commit message. Worth updating it to describe the two-phase split, the reuse crash it prevents, and all three tests. (The horizon-join sentence was already corrected, so the body is in a mixed state.)

## Minor

- **m1 — stale comment, `GroupByShardingContext.java:78-80`.** It still says the tracker is "Null when no per-query limit applies (the shared horizon-join path never binds one)". The horizon-join path *does* bind a tracker: `BaseAsyncHorizonJoinAtom.init():472` / `BaseAsyncMultiHorizonJoinAtom:548` capture `executionContext.getMemoryTracker()`, propagated via `shardingCtx.setMemoryTracker(...)` at `AsyncHorizonJoinAtom:255` / `AsyncMultiHorizonJoinAtom:237`. This is the exact claim the diff corrected in the sibling `GroupByMapFragment` field comment — the same correction should apply here.

- **m2 — invariant is correct-by-construction but not guarded.** `shards.size() ∈ {0, NUM_SHARDS}` is guaranteed only for the per-query-limit path. A Java-heap `OutOfMemoryError` mid-Phase-1 (on `new UnorderedVarcharMap`) would leave a partial list and reintroduce the out-of-bounds on reuse; so would any future change that adds a throwing operation to Phase 1. This is not reachable today via the per-query limit (Phase 1 allocates no native memory), and a Java-heap OOM mostly means the query is already dead — but the optional `assert shards.size() == 0 || shards.size() == NUM_SHARDS;` at the top of the else-branch (raised earlier) is cheap insurance and documents the invariant. The Phase-1 comment's absolute "the per-query limit cannot trip" is accurate as written, but it's specifically the *per-query* limit, not every possible throw.

- **m3 — title `chore(core)` vs `fix(core)`.** With the `Bug` label and a leak that's reachable in production whenever a per-query memory limit is configured and a sharded GROUP BY / horizon join breaches mid-shard-open (plus, pre-fix, the PGWire prepared-statement reuse crash), `fix(core)` reflects the user impact and lands it in the release notes; `chore` typically won't. Consider a `Flaky Test` label too, since it stabilizes the Windows `griffin-base` flake.

## Summary

- The two-phase split is correct, complete, and concurrency-safe; it closes C1 structurally rather than papering over it, and subsumes the original reorder. Success-path behavior is byte-identical to before. Zero perf/algorithmic delta.
- Verification: 3/3 tests green on an independent build, revert-proven; leak balance, tracker symmetry, concurrency, and OrderedMap coverage all re-checked and clear.
- Nothing blocking. One description-vs-code drift (M1) worth fixing before merge since the body becomes the `master` commit; the rest are a stale comment, an optional defensive assert, and the title.

puzpuzpuz (approved):

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/GroupByMapFragment.java|9|9|100.00%|

PR #7311, opened 2026-06-23
ci(build): bound cache connect timeout and retry connect failures

sklarsa:
## Problem

Self-hosted CI builds still intermittently fail with `Connect to maven.internal:8081 ... Connection timed out`, even after the cache was fronted by nginx. Diagnosis on the cache box and a runner container showed the failure is a **connect-phase** event: under the CI fleet's burst (a single PR fans out ~24 self-hosted jobs that resolve the Maven tree at once), an occasional fresh TCP connect to the single-IP cache has its SYN dropped and retransmitted, riding the OS SYN-retransmit budget (`tcp_syn_retries=6` → **~127s**) into a dead build.

Two things made this fatal rather than self-healing:

1. **No connect timeout** — wagon inherited the ~127s OS default, so a black-holed connect hung out the whole build.
2. **`retryHandler.count=3` did not retry it** — Apache HttpClient's `DefaultHttpRequestRetryHandler` puts `ConnectException` on its built-in non-retryable list (`{InterruptedIOException, UnknownHostException, ConnectException, SSLException}`), so connect-phase failures were never retried.

## Fix

- **`ci/maven-settings-hetzner.xml`** — add a `<server>` `<httpConfiguration>` for the cache mirror with `connectionTimeout=10000` (and `readTimeout=120000`), so a stuck connect fails fast (~10s) instead of ~127s. Scoped to the cache mirror only.
- **`MAVEN_RUN_OPTS`** (test-pipeline / test-fuzz / test-hosted) — add `-Dmaven.wagon.http.retryHandler.nonRetryableClasses=java.io.NotSerializableException`, which overrides HttpClient's default non-retryable set so `ConnectException` **is** retried. A retried connect uses a fresh source port and almost always succeeds.

Net effect: a black-holed connect becomes a ~10s fast-fail + retry instead of a ~127s fatal hang.

## Validation

Verified empirically in a throwaway runner container (Maven 3.8.7, same as CI) against a black-holed mirror IP: with the default handler the connect failure was **not** retried; with `nonRetryableClasses` overridden it **was** retried, and `connectionTimeout` bounded each attempt to ~10s. (See PR thread for the before/after wagon `-X` output.)

This is the client-side half of the fix. The cache-side half (enlarged nginx upstream keepalive pool + `proxy_next_upstream` retry) ships separately in the infra repo and reduces the drop rate at the source.

### DIFF
--- a/ci/maven-settings-hetzner.xml
+++ b/ci/maven-settings-hetzner.xml
@@ -26,4 +26,27 @@
     </mirror>
   </mirrors>
 
+  <servers>
+    <!--
+      Bound the connect/read timeouts for the cache mirror. The cache is a single private
+      IP fronted by nginx; under the CI fleet's connection bursts a fresh TCP connect can
+      occasionally have its SYN dropped and retransmitted. Wagon's default connect timeout
+      inherits the OS SYN-retransmit budget (~127s with tcp_syn_retries=6), so a single
+      black-holed connect rides that full window and kills the build. A short connectionTimeout
+      makes a stuck connect fail fast (in seconds) so the wagon retry handler can reconnect
+      on a fresh source port instead of hanging. The server id MUST match the mirror id above.
+    -->
+    <server>
+      <id>hetzner-reposilite</id>
+      <configuration>
+        <httpConfiguration>
+          <all>
+            <connectionTimeout>10000</connectionTimeout>
+            <readTimeout>120000</readTimeout>
+          </all>
+        </httpConfiguration>
+      </configuration>
+    </server>
+  </servers>
+
 </settings>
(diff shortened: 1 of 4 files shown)


sklarsa:
**Empirical validation** (throwaway runner container, Maven 3.8.7 = CI version, against a black-holed mirror IP):

| | wall-clock | connect error |
|---|---|---|
| **baseline** (no connectionTimeout) | **135s** — rides the full OS SYN-retransmit budget | `java.net.ConnectException: Connection timed out` |
| **fix** (connectionTimeout=10s) | **11s** — bounded by the 10s connect timeout | `java.net.SocketTimeoutException: Connect timed out` |

This confirms the `connectionTimeout` is applied: a black-holed connect now fails in ~10s instead of ~127s.

The `nonRetryableClasses` override making `ConnectException` retryable is confirmed by wagon 3.5.3 source — `getRetryHandler()` calls `new DefaultHttpRequestRetryHandler(count, requestSentEnabled, getNonRetryableExceptions())`, whose third arg **replaces** HttpClient's default non-retryable set `{InterruptedIOException, UnknownHostException, ConnectException, SSLException}`. Against this *permanent* black-hole every attempt fails by design (the IP never answers); in the real *transient*-drop case the retried connect uses a fresh source port and succeeds.

bluestreak01 (approved):

Issue #1083, opened 2021-06-03
Add all contribs 🎉 


bsmth:
Adding latest contributors 🎉

bsmth:
@all-contributors bot, please add @rrjanbiah for bug reports

Thanks for your contribution!

bsmth:
@all-contributors bot, please add @rrjanbiah for bug reports

Thanks from the QuestDB team!

bsmth:
@all-contributors bot, please add @sarunas-stasaitis for bug reports

Thanks from the QuestDB team!

bsmth:
@all-contributors bot, please add @RiccardoGiro for bug reports

Thanks for your contribution!

bsmth:
@all-contributors bot, please add @duggar for bug reports

Thanks from the QuestDB team!

bsmth:
@all-contributors bot, please add @postol for bug reports

Thanks from the QuestDB team!

bsmth:
@all-contributors bot, please add @nikochiko for feature request and feedback

Thanks for your contribution!

bsmth:
@all-contributors bot, please add @nikochiko for feature request and feedback

Thanks for your contribution!

bsmth:
@all-contributors bot, please add @petrjahoda for bug reports

Thanks from the QuestDB team!

bsmth:
@all-contributors bot, please add @t00 for bug report

Thanks from the QuestDB team!

bsmth:
@all-contributors bot, please add @snenkov for user testing, feature requests and bug reports

Thanks for your continued feedback and ideas!

bsmth:
Thanks everyone for your awesome feedback 🙌🏻

snenkov:
Wow that's cool! Thank you 🙏

rrjanbiah:
@bsmth Honestly, it is too much for my single ticket #1077 :-) Thank you!

bsmth:
> @bsmth Honestly, it is too much for my single ticket #1077 :-) Thank you!

Not at all, all contributions are helpful for getting feedback from the community!

PR #7315, opened 2026-06-23
chore(core): refresh materialized views correctly after a base-table truncate and on read-only nodes

jovfer:
Tandem: https://github.com/questdb/questdb-enterprise/pull/1083

## What

Materialized-view incremental refresh did not always converge a still-valid view to its base table.
This corrects the behaviour around base-table truncates, read-only / role-switch nodes, and the
role-promote hydrate window. Each change is gated on a condition that does not arise on a normal
single-node primary (a truncate in the scanned range, a read-only/authorization refusal, a WAL-read
failure, or the promote gate), so single-node behaviour is unchanged.

### Base-table truncate

1. **Treat a TRUNCATE in the scanned WAL range as an invalidating barrier.** Incremental interval
   planning scanned the base-table WAL range but skipped a `TRUNCATE` there (it carries no data
   interval), so `refreshIntervalsBaseTxn` could advance past the truncate without invalidating —
   leaving a still-valid view holding stale pre-truncate rows. `WalTxnRangeLoader` now surfaces an
   in-scan `hasTruncate` flag over the exact scanned bounds; incremental interval planning
   (`MatViewRefreshJob.updateRefreshIntervals0`) and the hydrate-time WAL-gap classifier
   (`CairoEngine.loadMatViewIntoStore`) reproduce the existing primary `"truncate operation"`
   invalidation. Gated on a previously-refreshed view (`lastRefreshTxn > -1` / `lastRefreshBaseTxn > -1`),
   so a never-refreshed view is not over-invalidated.

2. **Do not advance the refresh watermark past a truncate barrier.** A refresh that detected the
   truncate enqueued the INVALIDATE but then let the same run's no-rows commit advance the persisted
   `lastRefreshBaseTxn` past the truncate. If that INVALIDATE was later lost (for example discarded
   across a role switch), the load-time gap scan started after the truncate and never re-detected it.
   `RefreshContext` now carries a `hasTruncateBarrier` flag so the truncate path stops without the
   watermark-advancing no-rows commit — keeping the durable baseline honest for the load-time backstop.

3. **Tolerate a missing WAL in the hydrate truncate scan.** `hasBaseTableTruncateInWalGap` let a
   missing/purged-WAL read throw escape `loadMatViewIntoStore`'s logging-only catch, skipping
   `enqueueIncrementalRefresh` and leaving the view silently unscheduled after a promote-hydrate. It
   now catches the read failure and treats it as no-truncate-found, so the normal incremental refresh
   is still enqueued (the refresh path already falls back to a full refresh on the same read failure).

### Read-only / role-switch nodes

4. **Do not re-enqueue a read-only-refused refresh while the node is read-only.**
   `handleErrorRetryRefresh` re-enqueued a refusal unconditionally, so a drain-to-empty loop could
   spin. The re-enqueue is gated on `!isReadOnlyMode()`; a read-only refusal still never invalidates
   the view (the defer-and-retry-later contract is preserved).

5. **Defer `invalidateView`'s inner retry loop on a read-only refusal.** When a node flipped read-only
   after `invalidateView` passed its top-of-method guard, the inner writer-acquire loop refused on
   every pass and never threw — spinning at 100% CPU while holding the view-state lock for the
   replica's lifetime. It now defers on an authorization refusal (mark pending + re-enqueue + return),
   the same as the top guard; the rename-retry case is unchanged. The re-enqueue is bounded by
   `markAsPendingInvalidation` (the re-dequeued task hits the top guard as a no-op).

6. **Suspend refresh-task execution during the role-promote read-only window.** New
   `CairoEngine.isMatViewRefreshSuspended()` (default `false`); `MatViewRefreshJob.processNotifications()`
   skips dequeuing any task while it is set. Enterprise sets it for the promote window — the real store
   is hydrated and the catch-up kickstart enqueued, but writes are not yet open — so a refresh worker
   cannot consume-and-drop the kickstart on the still-read-only engine. OSS behaviour is unchanged: the
   gate is never set.

### Unrelated, same branch

7. **`WorkerPool`: count down not-yet-added workers on a concurrent halt.** When `start()`'s
   worker-add loop observed `closed` mid-add and broke out, the `halted` latch was already sized for
   the full worker count, so a `halt()` blocked on it waited out its whole timeout. It now counts the
   latch down for every not-yet-added worker before breaking.

## Why

A still-valid materialized view must converge to its base table; retaining stale pre-truncate rows, or
leaving a view unscheduled/stuck at seed rows after a promote, is incorrect derived data. An unbounded
read-only re-enqueue can wedge a refresh drain, and the inner-loop spin holds the view-state lock for
the node's lifetime.

## Test plan

- `MatViewTest` (whole class) green (294, 1 pre-existing skip), including the truncate-boundary anchors
  `testBaseTableInvalidateOnTruncate` and `testBaseTableTruncateDoesNotInvalidateFreshMatView` (the
  over-invalidation guard), plus new RED-first cases:
  `testTruncateBarrierDoesNotAdvanceRefreshBaseTxnPastTruncate` (watermark hold) and
  `testHydrateTruncateScanThrowStillSchedulesRefresh` (missing-WAL hydrate scan).
- The new barrier reuses the existing primary invalidation reason `"truncate operation"`, so the
  boundary tests continue to assert it.
- The Enterprise tandem (#1083) carries the promote-window and inner-loop witnesses and the heavy
  SwitchFuzz convergence soak.


bluestreak01:
## Tandem code review — #1083 (ENT) ⇄ #7315 (OSS)

Reviewed as one logical change (the ENT pin `7e861e7→5dcec72` carries the two OSS mat-view commits). Findings are prefixed **[ENT]** / **[OSS]**. The TRUNCATE-detection assumption (`walId > 0` for a WAL-table truncate) was confirmed: `WalWriter.truncateSoft()` → `events.truncate()` + `getSequencerTxn()` commits with the writer's positive walId, so the loader's `if (walId > 0)` scan sees it.

---

## Critical

### C1 — [OSS] `handleErrorRetryRefresh` now spins `invalidateView`'s inner retry loop forever on a read-only race  *(in-diff change → out-of-diff break)*

**File:** `core/src/main/java/io/questdb/cairo/mv/MatViewRefreshJob.java`
- In-diff: `handleErrorRetryRefresh` (≈L836–892) — the authorization branch now **always returns `true`**, including when `stateStore == null`.
- Out-of-diff break: `invalidateView` inner loop, the catch at **L1380**:
  ```java
  while (true) {
      viewToken = engine.getUpdatedTableToken(viewToken);
      try (WalWriter walWriter = engine.getWalWriter(viewToken)) {
          setInvalidState(...); break;
      } catch (CairoException ex) {
          if (!handleErrorRetryRefresh(ex, viewToken, null, null)) {  // stateStore == null
              throw ex;
          }
      }
  }
  ```

**Contract violated:** pre-PR, `handleErrorRetryRefresh(..., stateStore=null, ...)` on an authorization error returned **`false`** (git-confirmed against pin `7e861e7`; the old comment even says *"we cannot requeue here, so fall through to today's semantics" → return false*). The L1380 caller relies on that `false` to `throw ex` and unwind. The PR changed it to **`true`**, so `if (!true)` never throws.

**Reachable code path:**
1. A shared refresh-pool worker (always registered — `EntServerMain.setupMatViewJobs` runs regardless of role) dequeues an `INVALIDATE` task (e.g. the new "truncate operation" barrier, or any base-table truncate/drop/update). `invalidate()` → `invalidateView(view, reason, true)` (force=true).
2. Top-of-method `if (engine.isReadOnlyMode())` is **false** (node still PRIMARY) → proceeds, `tryLock()` succeeds, enters `while(true)`.
3. A demote runs concurrently. `EntCairoEngine.switchRole` **Step 1 flips read-only before** the mat-view quiesce drain (`EntCairoEngine.java:1167-1192`).
4. `EntCairoEngine.getWalWriter()` (override at `EntCairoEngine.java:732`) now throws `CairoException.authorization()`.
5. `handleErrorRetryRefresh` sees `isAuthorizationError()` → returns `true` → `if (!true)` → **no throw** → loop → `getWalWriter` throws again (still read-only) → … **non-terminating loop**.

**Impact:** the worker spins at 100% CPU holding `viewState`'s lock for the entire replica lifetime (only a re-promote lets `getWalWriter` succeed and break). Each demote that loses this race burns another shared-pool worker; repeated occurrences exhaust the pool and stop all mat-view refresh. The pre-PR fallback was a bounded throw-and-unwind (the `finally` released the lock). This is a **new regression**, and the existing `invalidateOnDemoteDefersAndNeverEscapesTheWorker` test does *not* cover it — it sets read-only *before* enqueuing, so it only hits the top-of-method guard, never the inner-loop race. Its own comment documents the very throw behavior this PR removes.

**Suggested fix:** the inner catch must distinguish read-only (defer, don't retry) from rename (`isTableDoesNotExist`, retry). E.g. in the catch, if `engine.isReadOnlyMode()` / `ex.isAuthorizationError()`, do `viewState.markAsPendingInvalidation(); stateStore.enqueueInvalidate(viewToken, invalidationReason); return;` (mirroring the top-of-method guard) instead of looping; keep the retry only for the rename case.

---

### C2 — [OSS] Truncate barrier is defeated across a demote: the detecting refresh advances `lastRefreshBaseTxn` *past* the truncate, blinding the load-path backstop

**File:** `MatViewRefreshJob.java` `updateRefreshIntervals0` truncate branch (≈L1916–1929); downstream `findRefreshIntervals` (L525–548) → `insertAsSelect`/`refreshSuccessNoRows` (≈L935–949); `CairoEngine.hasBaseTableTruncateInWalGap` (CairoEngine.java:2399–2411).

**Trace** (`R < T < L`; view previously refreshed at `R`, truncate at `T`, base last txn `L`):
1. Incremental refresh on the PRIMARY: `updateRefreshIntervals0` scans `(R, L]`, `hasTruncate()` true → `enqueueInvalidate(view, "truncate operation")`, `intervals.clear(); return intervals` (**empty, non-null**).
2. `findRefreshIntervals`: `cachedIntervals.size()==0` → treated as *"no data, nothing to refresh"*, `refreshIntervals=null`, but `refreshContext.toBaseTxn = L`. With `minTs(MAX) > maxTs(MIN)`, `intervalIterator` stays `null`.
3. `insertAsSelect` (`intervalIterator==null`): `commitBaseTxn = L > lastRefreshBaseTxn(R)` → `refreshSuccessNoRows(..., baseTableTxn=L, ...)` **persists `lastRefreshBaseTxn=L` and marks the view valid** (`resetMatViewState(L, …, invalid=false, …)`).
4. The view is now on-disk **valid at `L` (past the truncate)** with only a *queued* INVALIDATE. A demote sets read-only (Step 1) and quiesce-drains: the INVALIDATE is processed read-only → `invalidateView` top guard defers (`markAsPendingInvalidation` in-memory + re-enqueue), and the deferred state is **discarded at the NoOp swap** — it never persists `invalid`.
5. On promote, `hasBaseTableTruncateInWalGap` scans `(L, lastBaseTxn]`. The truncate at `T < L` is **not in range** → not detected → the view stays **valid with stale pre-truncate rows** — exactly the over-count failover bug this PR claims to fix.

**Why the tests miss it:** `demoteThenPromoteMustNotRetainPreTruncateMatViewRows` *suppresses* the refresh jobs and explicitly does **not** drain after the truncate, so steps 1–3 never run and `lastRefreshBaseTxn` stays at `R` where the load-path backstop works. Production (shared pool running) does run the refresh, hitting the unprotected path. The fix's two halves don't compose: the refresh advances the watermark relying on the enqueued INVALIDATE, while the load-path backstop relies on the watermark *not* having advanced.

**Suggested fix:** when `updateRefreshIntervals0` hits the truncate barrier, the refresh must **not** advance `lastRefreshBaseTxn` — abort the watermark commit (a distinct "stop, do not commit" signal rather than returning empty intervals that `insertAsSelect` watermark-commits), so the backstop remains effective even if the INVALIDATE is lost across a demote. Add a test with the shared pool *enabled* (refresh advances past truncate) + demote/promote.

---

## Moderate

**M1 — [OSS] No OSS test for any of the three new OSS code paths.** `WalTxnRangeLoader.hasTruncate()`, `updateRefreshIntervals0`'s truncate branch, and `CairoEngine.hasBaseTableTruncateInWalGap` have zero OSS coverage (the cited anchors `testBaseTableInvalidateOnTruncate`/`testBaseTableTruncateDoesNotInvalidateFreshMatView` exercise only the pre-existing `ApplyWal2TableJob` apply-time invalidation, not the new barrier). The interval-planning barrier is OSS-reproducible without any role switch — it belongs in `MatViewTest`. At minimum add a direct `WalTxnRangeLoader` unit test (range with a truncate → `hasTruncate()==true`; range without → `false`).

**M2 — [ENT] New tests give false confidence on the demote/promote case.** Because `createEntServerMainWithoutMatViewJobs` suppresses the refresh pool, `demoteThenPromoteMustNotRetainPreTruncateMatViewRows` only validates the "refresh never advanced" sub-case and masks C2; nothing exercises C1's inner-loop race. The suppressed-pool seam is fine for determinism but should be complemented by a pool-enabled variant.

---

## Minor

- **m1 — [OSS]** `hasBaseTableTruncateInWalGap` allocates a fresh `Path` + `WalTxnRangeLoader` (native `DirectLongList` + `WalEventReader`) + `LongList` **per view, per boot/promote-hydrate**, and builds+discards the full `intervals` union just to read one boolean. Cold path, but on a promote with many lagging views this is per-view native alloc/free + a full WAL-gap scan. Hoist a single reusable loader into the `loadMatViewIntoStore` caller loop.
- **m2 — [OSS]** `chore(core):` / `chore(ent):` for a data-correctness bug fix; should be `fix(...)` (mirrors the `Bug` label).
- **m3 — [ENT]** `AbstractReplicationTest.tearDown` calls `stopWatchdog()` *before* `super.tearDown()`/`objectStore.stop()`, so a hang during teardown (not the test body) is never captured — a gap given the watchdog targets a fork/upload hang that could surface during stop.
- **m4 — [ENT]** `AbstractReplicationTestWatchdogTest` asserts on the dump snapshot the instant it becomes non-empty; if `PrintStream` chunks the encoded write to the `ByteArrayOutputStream`, `contains(" at ")` could read a header-only prefix. Practically safe (8KB buffer ≫ header), but re-reading once after a short settle would remove the theoretical flake.
- **m5 — [ENT]** `SwitchFuzzTest.getSoakIterations`/`isRunAllFuzzTests` were inserted after `sendIlpHttp` (`s` < `g`/`i`), breaking alphabetical method ordering. Trivial.

---

## Downgraded (ruled out during source verification)

- *"`updateRefreshIntervals0` NPEs on `stateStore.enqueueInvalidate`"* — `this.stateStore = engine.getMatViewStateStore()` is always non-null.
- *"TRUNCATE has no `walId>0`, scan never flags it"* — disproved: `truncateSoft()` commits with the writer's positive walId.
- *"INVALIDATE early-defer self-feeds the demote drain like the refresh path"* — disproved: `markAsPendingInvalidation()` makes the re-enqueued task a no-op on the second pass; bounded.
- *"Load-path lower bound off-by-one vs interval planning misses a truncate"* — disproved: the load-path bound is *earlier* (wider scan), conservative-safe.
- *"`hasTruncate` stale-read if `load()` throws"* — disproved: both call sites read it only immediately after a successful `load()`.
- *"`WalTxnRangeLoader.close()` leaks `walEventReader`"* — pre-existing pattern, unchanged; `load()`'s try-with-resources closes the reader.

---

## Summary

**Verdict: request changes (both PRs).** The two production fixes are individually well-reasoned and the truncate-detection mechanism is sound, but:

- **C1** is a new liveness regression (non-terminating, lock-holding spin) introduced by the control-flow simplification in `handleErrorRetryRefresh`, undetected by any test.
- **C2** shows the truncate fix is **incomplete** across the very demote/promote scenario it targets — the detecting refresh advances the base watermark past the truncate and blinds the load-time backstop, so a lost (demote-discarded) INVALIDATE leaves a stale-valid view after failover.

Both should block merge. The highest-value findings (C1, C2) are out-of-diff: small in-diff changes break unchanged callers (`invalidateView`'s retry loop; the `insertAsSelect` watermark commit + load-path backstop). The enterprise test additions (watchdog, soak profiling, deterministic regressions) are reasonable and the RED→GREEN claims are consistent with the code, but the suppressed-pool seam hides C2.

<sub>🤖 AI-assisted review (level 3, tandem). Confirm findings against source before acting.</sub>

jovfer:
## Consolidated review — mat-view truncate / role-switch refresh fix

Two independent tandem reviews (round-2 + level-3, every finding source-verified at OSS `90df2559` / ENT `28da0e02`) converge on the items below. The three production fixes (truncate-as-barrier watermark hold, read-only inner-loop defer, missing-WAL hydrate tolerance) and the promote-suspend gate are individually sound, and the two earlier Criticals (inner-loop spin; watermark advancing past a truncate) are verified **fixed**.

### Verdict
- **OSS #7315 — request changes** (one Critical + one in-diff Moderate).
- **ENT #1083 — approve** after one test-harness minor.

### Critical [OSS] — pending-invalidation "zombie"
A truncate INVALIDATE can be permanently swallowed on a steady-state, multi-worker primary, leaving a view **`view_status='valid'` while serving stale pre-truncate rows** — the exact failure this change targets, now reachable with no role switch.
- **Trap (pre-existing):** `invalidateView`'s top guard skips the method when `isPendingInvalidation()` (`MatViewRefreshJob.java:1354`); `pendingInvalidation` is cleared only by `markAsValid()` ← `fullRefresh`, never by an incremental/range `refreshSuccess`, and `initFromReader` does not reset it.
- **Path A (this change's trigger):** the new in-lock `enqueueInvalidate` (`MatViewRefreshJob.java:1954`) runs while the refresh holds the view lock; with the default ≥2 refresh workers, another worker dequeues it inside the hold window → `tryLock` fails → `markAsPendingInvalidation` + re-enqueue → the re-dequeue hits the top guard → no-op forever.
- **Path B (gap in the new defer):** on a commit-fence refusal (writer acquired, then a demote flips read-only), `setInvalidState` has already set in-memory `invalid=true` before the fenced commit throws; the catch then also sets `pendingInvalidation`, leaving `invalid && pending` in memory but `valid` on disk.
- The green soak does not cover this — its frequent role switches rebuild the store and reset the in-memory flag.
- **Fix being applied:** finalize the barrier invalidation inline on the primary (the producer already holds the view lock + writer) instead of enqueuing a task it cannot run under its own lock; clean up the commit-fence defer so it never leaves `invalid && pending`; plus a pool-enabled (≥2 worker) regression test asserting the view ends *actually* invalid, not stuck `pending && valid`.

### Moderate [OSS], in-diff — period mat-views bypass the watermark-hold
For **period** views the barrier early-return is skipped: the period branch (`MatViewRefreshJob.java:629-649`) builds a non-null interval iterator, so `insertAsSelect` never reaches the `hasTruncateBarrier` guard (which lives inside `if (intervalIterator == null)` at `:932`), and `lastRefreshBaseTxn` still advances past the truncate → on promote the load-time backstop scans past it → stale-valid period view. **Fix being applied** (hoist the barrier check above the period branch) + a period-view test. Every existing test uses non-period `sample by 1h`.

### Tracked deferrals (out-of-diff, same bug class — recorded, not silent)
- **[OSS]** DROP/DETACH PARTITION and data UPDATE / structural ALTER share the identical WAL-gap-skip loss; the barrier closes only TRUNCATE (`WalTxnRangeLoader.java:149`).
- **[OSS]** A truncate whose gap-WAL is purged during the pending-invalidation window can still leave a stale-valid view (the deferred "Option D").
- **[OSS]** A third barrier site — `DatabaseCheckpointAgent.updateMatViewRefreshIntervals` advances the watermark past a truncate on incremental backup without consulting the new `hasTruncate()` primitive (a cheap future wiring).
- **[OSS] perf**, in-diff: on promote, the new load-time `hasBaseTableTruncateInWalGap` scan and the scheduled refresh re-scan the same `(R,L]` gap per lagging view.

### Tests [OSS] — under-assertion (being strengthened)
`testTruncateBarrierDoesNotAdvanceRefreshBaseTxnPastTruncate` asserts only that the watermark didn't move, never that the view ends invalidated/converged; there is no direct `WalTxnRangeLoader.hasTruncate()` unit test and no pool-enabled contention test.

### Minor [ENT] — being fixed
The replication test harness disarms its hang-watchdog *before* teardown, leaving the teardown/object-store-stop hang window (the exact incident the watchdog exists for) uncovered. Moving the disarm after teardown.

### Verified fixed / false-positive (no action)
Both prior Criticals; the bundled `WorkerPool` unstarted-worker halt-accounting change (correct, RED-verified — `countDownUnstartedWorkers` covers exactly the never-added tail, latch sum = workerCount); the promote-suspend gate "stuck-suspended on a writable primary" concern (all four set/clear paths paired after `setCurrentRole(PRIMARY)`, hydrate-unwind + demote-entry clear defensively, demote clears before the quiesce drain); the per-worker `hasTruncate`/`refreshContext` "data race" (per-instance via `cloneInstance`); native-memory-leak-on-throw; and purge-horizon-advancing-past-truncate.

<sub>Consolidated from two AI-assisted tandem reviews; every finding verified against source. In-diff items (the Critical, the period bypass, tests, the ENT watchdog) are being addressed in follow-up commits; out-of-diff items are tracked for follow-up.</sub>

jovfer:
### Update — in-diff review feedback addressed

Pushed (OSS `38833451`+`ab0f2294`, ENT `c6f0325f`+`c4e60835`):

- **Critical (pending-invalidation "zombie") — fixed.** The truncate barrier now invalidates the view **inline** under the lock + WalWriter the refresh already holds, instead of enqueuing an INVALIDATE a second worker could park — eliminating the race at its root. The dependent/chained-view cascade is restored so it stays symmetric with the apply-time and load-time invalidation paths. Path B (commit-fence refusal) now rolls back the in-memory `invalid` flag and defers cleanly, never leaving `invalid && pending`. (The guard-relaxation alternative was rejected: it caused an infinite spin in the refresh drain loop.)
- **Period mat-views — fixed.** The watermark-hold is now honored for period views too (the `hasTruncateBarrier` early-return was hoisted above the period-iterator branch).
- **Tests strengthened.** The barrier test now also asserts the view ends `view_status='invalid'` (the user-visible half), plus a direct `WalTxnRangeLoader.hasTruncate()` unit test, a **pool-enabled** 2-worker contention witness, and a chained-view (view-on-view) regression test. [ENT] the replication test-harness watchdog is now disarmed after teardown.
- **Tracked deferrals** (recorded as code comments + an internal note, same bug class, signed-off): DROP/DETACH PARTITION + data UPDATE / structural ALTER barrier scope; a truncate whose gap-WAL is purged during the pending window (the "Option D" residual); the promote double-scan; and the incremental-backup checkpoint barrier site.

CI is re-running.

jovfer:
# Tandem code review (round 10, from scratch) — ENT #1083 ⇄ OSS #7315

**Level 3, tandem mode.** Reviewed as one logical change: the ENT pin bump
`1f78f2ead → ff9ca62f0` carries the OSS mat-view refresh fixes; the ENT side wires the
promote refresh-suspend gate and adds tests/diagnostics. Heads reviewed: **ENT
`3ab045d73`**, **OSS `ff9ca62f0`** (local checkout matches both PR heads exactly).

This round was done fresh, ignoring prior review rounds. 11-agent pass (no `.rs` files → Rust
agent N/A; PR-metadata folded inline), every finding source-verified in Step 3b.

The two production fixes (TRUNCATE-as-barrier with watermark hold; read-only inner-loop
defer + promote-suspend gate) and the WorkerPool halt-latch fix are individually sound. The
earlier-round Criticals (inner-loop spin; watermark advancing past a truncate) are
**re-verified fixed** at these heads — see Downgraded. No new Critical. Two Moderate (one a
behavioral trade-off, one a test/doc coverage loss), the rest Minor.

---

## Critical

None.

---

## Moderate

### M1 — [Cross-repo] A drain-budget-expiry demote **abort** can drop an in-flight single-view refresh that is then not auto-recovered *(in-diff change → out-of-diff behavior regression)*

**OSS (the change):** `core/.../mv/MatViewRefreshJob.java:856` — `handleErrorRetryRefresh`
now re-enqueues a read-only-refused refresh only `if (stateStore != null && !engine.isReadOnlyMode())`.
**ENT (the affected path):** `EntCairoEngine.java:1534-1562` (`drainWriterPool`,
`restorePrimaryOnTimeout=true`) + `EntLifecycleOrchestrator.java:331-354` (clean refusal).

**Contract changed:** pre-PR, a read-only refusal with `stateStore != null`
**re-enqueued unconditionally** (git-confirmed against `1f78f2ead`). The PR gates the
re-enqueue on `!isReadOnlyMode()` so a read-only refusal during a demote no longer self-feeds
the quiesce drain — this is the correct fix for the demote-quiesce hang (CI 244815/244851,
pinned by `demoteQuiesceIsBoundedUnderSelfRequeue`). But the same guard also changes the
**abort** path.

**Reachable trace:**
1. An early demote flips the node read-only (Step 1), then `drainWriterPool(...,
   restorePrimaryOnTimeout=true)` waits for busy CLIENT writers. The mat-view refresh pool is
   a separate pool and keeps running (it is NOT the promote-suspend window — suspend is
   promote-only).
2. During that read-only drain window a refresh worker dequeues a **single-view**
   `INCREMENTAL_REFRESH` (e.g. a `tryLock`-retry re-enqueue at `MatViewRefreshJob.java:1739`,
   or a leftover kickstart). `refreshIncremental` → `engine.getWalWriter(viewToken)` throws
   authorization (read-only) → catch at `:1772` → `handleErrorRetryRefresh(..., stateStore,
   null)` defers but, being read-only, **does not re-enqueue** → the task is dropped.
   `refreshIncremental` (single-view) has no `notifyBaseRefreshed` re-arm.
3. The drain budget expires with the client writer still live → `drainWriterPool` restores
   PRIMARY (`EntCairoEngine.java:1545`) and throws `RoleSwitchDrainBudgetExpiredException`
   **before** `switchMatViewMachinery(REPLICA)` runs — so the store stays primary-shaped and
   was never swapped to NoOp.
4. The orchestrator takes the clean-refusal branch (`EntLifecycleOrchestrator.java:340`):
   publishes DEGRADED and `role.set(priorRole=PRIMARY)` — **no store rebuild, no
   re-kickstart, no refresh re-enqueue**.

Net: the node is back to PRIMARY with the same store, and the dropped single-view refresh is
gone. The view lags by one refresh cycle. It recovers only on the **next base-table commit**
(which enqueues a base-table task that refreshes all dependents), a restart (boot reload
re-kickstarts), or `REFRESH FULL`.

**Why it is bounded (and why this may be an acceptable trade-off):** under QuestDB's normal
continuous-ingestion workload the next base write self-heals the view near-instantly; the view
is never marked invalid (no user-visible "invalid" state), and there is no data loss — only
transient staleness. The trigger is narrow: a demote that *aborts* via budget expiry (an
exceptional refusal that keeps the node PRIMARY) coinciding with a single-view incremental
in-flight in the drain window. Per this repo's role-switch triage rubric, new-path staleness
that is operator-recoverable is acceptable. The pre-PR alternative (unconditional re-enqueue)
was strictly worse — it hung the completing-demote quiesce drain.

**The residual concern:** for a **quiesced / low-write-rate base table**, the view stays stale
until a restart or `REFRESH FULL`, and nothing surfaces it. No test covers the abort-path
refresh-drop. This is an undocumented behavioral regression on an existing path (the
drain-budget refusal predates this PR).

**Suggested resolution (author's call):** either (a) on the clean-refusal abort at
`EntLifecycleOrchestrator.java:340-354`, re-kickstart an incremental refresh for the views
(the node is writable PRIMARY again, so it would succeed), or (b) explicitly document that an
aborted demote can leave a view lagging until the next base write / `REFRESH FULL`, and accept
it. If (b), a one-line note in the PR body under the read-only-defer section is enough.

### M2 — [ENT] Deleting the STRAND/ER004 deterministic witness + a now-false javadoc claim *(in-diff)*

**Files:** `SwitchFuzzTest.java` (deleted `runDemoteMidCatchupIteration` +
`testRegression_demoteMustNotLeaveStoreDataAheadOfIndex`);
`GracefulFailoverLossDirectionTest.java:413-418` (rewritten javadoc);
`UploaderTeardownBudgetBoundedTest.java` (the cited replacement).

The PR deletes the only deterministic STRAND-direction test (zero-drain-budget demote
mid-catchup → assert settles REPLICA, i.e. local index never left ahead of remote → strict
row-count equality) and rewrites `GracefulFailoverLossDirectionTest`'s javadoc to claim
`UploaderTeardownBudgetBoundedTest` (with the soak's PITR dimension) now "guards the STRAND
direction -- the store must never be left with uploaded data ahead of its remote index (which
would raise ER004)."

**Verified false:** `UploaderTeardownBudgetBoundedTest` is a **bounded-teardown budget** test.
It hangs the upload, demotes with a 1.5s budget, asserts the demote settles to REPLICA within
a **45s wall-clock ceiling** (`DEMOTE_WALL_CLOCK_CEILING_MS`), then **releases the wedge so the
downloader can settle** (its own javadoc, line 50). It contains zero references to ER004,
`index.msgpack`, "remote index", or strand, and it deliberately *avoids* stranding (it releases
the hang so the downloader reconciles). It proves bounded reap latency + no row loss — not the
strand invariant.

So: (a) the javadoc is factually wrong about what `UploaderTeardownBudgetBoundedTest` covers,
and (b) deterministic strand coverage is downgraded to probabilistic — the `testSoak` PITR path
still hits the strand backstop via force-evict budgets + `waitForRole(REPLICA)`, but only
across fresh seeds, and the PITR data invariant is a `>=` row bound (not the deleted test's
strict equality). This is a test/documentation regression, not a shipping bug.

**Suggested fix:** restore a deterministic strand test (re-add `runDemoteMidCatchupIteration`
as a non-seeded `@Test`, the same way the other pinned seeds were converted), or correct the
`GracefulFailoverLossDirectionTest` javadoc to stop crediting `UploaderTeardownBudgetBoundedTest`
with STRAND coverage and instead cite the soak's force-evict `waitForRole(REPLICA)` backstop as
the (probabilistic) guard.

---

## Minor

- **m1 — [OSS] Double WAL-gap scan + per-view native alloc on promote/boot.**
  `CairoEngine.hasBaseTableTruncateInWalGap` (CairoEngine.java:2409) allocates a fresh `Path` +
  `WalTxnRangeLoader` (native `DirectLongList` + `WalEventReader`) + `LongList` per lagging
  view and runs a full WAL-gap scan to read one boolean; the subsequently-enqueued incremental
  refresh re-scans the identical `(lastRefreshBaseTxn, baseTableLastTxn]` range. So a
  no-truncate promote scans each lagging view's gap twice and allocates a loader per view. Cold
  path (boot / REPLICA→PRIMARY promote), bounded to lagging views, and already self-documented
  as a tracked optimization (CairoEngine.java:50-55, 2413-2417). It does inflate promote
  latency, the exact metric this PR family targets. Cheap mitigation: hoist one reusable
  `WalTxnRangeLoader` (and the caller's already-live `WalEventReader`/`Path`) into the
  `buildViewGraphs`/`hydrateMatViewStateStore` loop and thread it into
  `hasBaseTableTruncateInWalGap` — kills the per-view native alloc. Non-blocking.

- **m2 — [ENT] `matViewRefreshSuspended = true` is set outside the hydrate-unwind `try`.**
  `EntCairoEngine.java:1931` sets the gate before the `try {` at `:1949`. A throw between them
  (`createConcurrentQueue` 1935 / `new MatViewStateStoreImpl` 1940) would leave the gate set and
  bypass the unwind clear at 1963. Verified NOT a live leak: the store delegate only becomes
  `MatViewStateStoreImpl` at 1941, and there is no throwing statement between 1941 and the `try`
  entry — so any realistic throw leaves the delegate NoOp, and a same-direction promote retry
  rebuilds (no early-return at 1921); a demote also clears defensively at 1975. Cosmetic
  defense-in-depth: move the set to just before `hydrateMatViewStateStore()` inside the try.

- **m3 — [OSS] Stale test name in a comment.** `MatViewRefreshJob.java:2015` cites
  `testRefreshIntervalsDoNotAdvanceBeforeFirstDataRefresh`; the test actually added by this PR
  is `testRefreshIntervalsAdvanceOnlyAfterFirstDataRefresh` (the other cited test,
  `testReusedReaderResetsStaleStateWhenLaterBlocksAbsent`, exists and is correct). Align the
  cited name.

- **m4 — [OSS] Checkpoint-restore can resume a view past a truncate (acknowledged, but the
  comment overstates it).** `DatabaseCheckpointAgent.java:665-669` advances the checkpoint base
  txn without consulting `walTxnRangeLoader.hasTruncate()`. The comment says restoring "resumes
  the view past the truncate with stale pre-truncate rows." In practice the new boot-time
  backstop catches the common case: checkpoint-create advances only `refreshIntervalsBaseTxn`
  and leaves `lastRefreshBaseTxn` unadvanced, and `loadMatViewIntoStore`'s backstop scans from
  `lastRefreshBaseTxn` (pre-truncate) on the post-restore boot → detects + invalidates before
  any resumed refresh. The genuine residual is only the WAL-purged-gap case (same residual the
  refresh path already tracks). Suggest softening the comment and, ideally, wiring `hasTruncate()`
  into this path (the primitive now exists) — deferral is acceptable if listed in the PR body.

- **m5 — [OSS] TRUNCATE-only barrier; DROP/DETACH PARTITION and rows-affected UPDATE share the
  gap-skip loss (acknowledged, pre-existing).** `WalTxnRangeLoader.java:149-159` flags only
  `WalTxnType.TRUNCATE`. DROP/DETACH PARTITION and rows-affected UPDATE also delete base rows
  and commit as non-data txns the interval scan skips, so a mat-view can retain rows deleted by
  `ALTER TABLE ... DROP PARTITION` in the unscanned gap. This is **pre-existing** (the scan
  skipped all non-data txns before this PR; the PR strictly improves matters by adding TRUNCATE
  detection) and explicitly scoped as a tracked follow-up. Given DROP PARTITION is the common
  retention operation, weigh whether at least DROP/DETACH PARTITION should ship in this PR.

- **m6 — [OSS] Standalone `UPDATE_REFRESH_INTERVALS` leaves `hasTruncateBarrier=true` dangling
  on the shared `RefreshContext`.** `updateRefreshIntervals0` (MatViewRefreshJob.java:2049) sets
  the flag but the standalone caller never reads/resets it. Verified correct: every subsequent
  consumer first resets it (incremental/range/full via `findRefreshIntervals`→`clear()`; the
  next `updateRefreshIntervals0` via its top reset at 1982; INVALIDATE never reads it) and
  `refreshContext` is per-instance (no cross-thread sharing). Fragile-but-correct; optionally
  reset to false after the standalone mint for clarity.

- **m7 — [OSS] Over-claiming comment on the post-dequeue gate re-check.**
  `MatViewRefreshJob.java:1452-1454` says the post-dequeue read is "guaranteed to observe the set
  gate." That happens-before guarantee holds only for the hydrate-enqueued kickstart (whose
  enqueue the promoter ordered after the gate-set); a task enqueued *before* the promote carries
  no HB edge, so its re-check is best-effort (volatile, practically immediate). Not a bug — such
  a task hits the read-only refusal and defers — but scope the comment to the kickstart.

- **m8 — [ENT] Member ordering.** `getSoakIterations()`/`isRunAllFuzzTests()` were inserted
  after `sendIlpHttp()` in `SwitchFuzzTest.java`, out of alphabetical order — mitigated by the
  fact that the file's whole `private static` method block is already unsorted, so the PR
  follows local convention rather than regressing sorted code. Trivial.

- **m9 — [ENT] Informational: per-PR soak sensitivity drops 100→3.** `LIGHT_SOAK_ITERS=3` on
  the PR path; the heavy 100 runs only under `QDB_RUN_ALL_FUZZ_TESTS` in the scheduled pipeline.
  This is a deliberate, documented cost split (the deterministic regressions remain as
  per-PR canaries and the scheduled job keeps full coverage), but the per-PR fuzz gate is
  materially weaker. Worth a conscious nod; not a blocker.

---

## Downgraded (ruled out during source verification)

- **Prior-round C1 — inner-loop read-only spin in `invalidateView`** — FIXED & verified. The new
  `if (ex.isAuthorizationError())` branch at `MatViewRefreshJob.java:~1404` intercepts the
  read-only refusal (mark-valid rollback + pending + enqueue + return) BEFORE the inner-loop
  `handleErrorRetryRefresh(..., null, null)` call at `:1424`. Confirmed no read-only-but-non-auth
  exception reaches 1424 (the read-only chokepoint throws `CairoException.authorization()`), so
  the `handleErrorRetryRefresh` always-return-true change cannot re-introduce the spin.
- **Prior-round C2 — refresh advances `lastRefreshBaseTxn` past the truncate** — FIXED & verified.
  `updateRefreshIntervals0` mints the invalidation inline under the held lock+writer and sets
  `hasTruncateBarrier`; `insertAsSelect` (`:935`, placed above the `intervalIterator == null`
  branch so period views are covered) returns without the watermark-advancing no-rows commit.
- **WorkerPool latch double/under-count** — FALSE POSITIVE. `halted` is sized `workerCount`;
  spawned workers `[0,i)` each count down once on exit; `countDownUnstartedWorkers(i)` covers
  exactly the never-spawned `[i, workerCount)`. Add+spawn are one monitor-guarded loop iteration
  gated by the same `closed` check, so the partition is disjoint. Both halt-before-start and
  start-before-halt races traced clean; `SOCountDownLatch.countDown` clamps at 0.
- **`reenqueueRefreshTask` aliasing the reused task** — FALSE POSITIVE. `ConcurrentQueue.enqueue`
  does `item.copyTo(slot.item)` (value copy); `MatViewRefreshTask.copyTo` copies all fields.
- **`MatViewStateReader.of(BlockFileReader)` clear()-first breaks a caller** — FALSE POSITIVE.
  All 5 callers (WalUtils:301, DatabaseCheckpointAgent:309/646, MatViewsFunctionFactory:207 —
  which already hand-called `clear()`) are populate-then-consume; none accumulates across `of()`.
  The clear() defaults match the block-absent sentinels and the always-present first block
  repopulates the core fields. This is the fix, not a regression.
- **`hasBaseTableTruncateInWalGap` off-by-one / over-broad catch** — FALSE POSITIVE. Probe range
  `(lastRefreshBaseTxn, baseTableLastTxn]` is a superset of the refresh scan range, so it cannot
  miss a truncate the refresh would see; the `CairoException`→false catch routes a genuine read
  failure to the refresh's own full-refresh fallback (safe outcome).
- **Resource leaks (Path/loader/WalEventReader, intervalIterator on the barrier early-return,
  walWriter on the inline-mint rollback, realStore on the hydrate unwind)** — FALSE POSITIVE.
  All traced to guaranteed frees on happy + error paths; the barrier early-return skips factory
  acquisition entirely (factory still null); the iterator is a pooled job-lifetime field with no
  native state.
- **Unclosed LOG chains** — none. Every new `LOG.*` chain terminates with `.$()`/`.I$()` and uses
  `.$`/`.$ts`/`.$safe` (no `.put` in a chain); all-ASCII.

---

## Summary

**Verdict: approve with two Moderate items to weigh — ENT: approve, OSS: approve.**

The production logic is sound and the earlier Criticals are re-verified fixed at the current
heads. Neither Moderate is a shipping correctness defect:

- **M1** is a deliberate behavioral trade-off — the read-only re-enqueue guard that fixes the
  demote-quiesce hang also drops a single-view refresh on the (exceptional) drain-budget-expiry
  demote-abort, leaving a view transiently stale until the next base write / `REFRESH FULL`. No
  data loss, self-healing under continuous ingestion. Either re-kickstart on the clean-refusal
  abort, or document it.
- **M2** is a test/doc coverage loss — a deterministic STRAND/ER004 witness was deleted and the
  replacement javadoc credit (`UploaderTeardownBudgetBoundedTest`) is factually wrong (it is a
  teardown-budget test that releases the wedge, not a strand witness). Restore the test or fix
  the comment.

Findings: **13 reported (0 Critical, 2 Moderate, 9 Minor + 2 informational), ~9 draft findings
dropped as false positives** after source verification. In-diff vs out-of-diff: **12 in-diff, 1
out-of-diff behavior regression (M1, an in-diff guard changing the unchanged abort path)**. The
out-of-diff cross-context pass (Agent 9) walked the full callsite inventory of every changed
symbol (5 `of()` callers, all `handleErrorRetryRefresh` callers, the sole `tryDequeue`/`reenqueue`
consumers, all `switchMatViewMachinery` callers) — all SAFE except the M1 interaction.

<sub>AI-assisted review (level 3, tandem, from-scratch). Confirm findings against source before acting.</sub>

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 56 / 84 (66.67%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/mv/ForwardingMatViewStateStore.java|0|2|00.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewStateStoreImpl.java|0|2|00.00%|
|:large_blue_circle:|io/questdb/cairo/mv/NoOpMatViewStateStore.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|22|45|48.89%|
|:large_blue_circle:|io/questdb/cairo/mv/WalTxnRangeLoader.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/mp/WorkerPool.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewStateReader.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|20|20|100.00%|

jovfer:
@bluestreak01 JFYI: ENT CI failure in the comment above was caused by infra timeout and green after restart.

bluestreak01 (approved):
Approved. Reviewed as the production half of tandem ENT#1083 at level 3 (11-agent pass + per-finding source verification). No Critical/P0 blockers. Verified: the truncate-barrier in updateRefreshIntervals0 (inline mint under view-lock+WalWriter, rollback-on-fence-refusal, dependent cascade only on success), the load-time backstop in hasBaseTableTruncateInWalGap, the read-only defer paths in handleErrorRetryRefresh/invalidateView, the processNotifications suspend-gate ordering (ConcurrentQueue release/acquire edge holds), MatViewStateReader.of(BlockFileReader) clear() fixing a real reused-reader carry-forward bug, and the WorkerPool countDownUnstartedWorkers latch fix. Non-blocking follow-ups: checkpoint-restore ignores hasTruncate() (DatabaseCheckpointAgent — stale-valid if gap WAL is purged), barrier scoped to TRUNCATE only (DROP/DETACH PARTITION + deleting UPDATE share the gap-skip class), cold-path double scan / per-view allocation, and the inaccurate 'same range' comment (probe range is actually wider — behaviorally safe). Consider fix(core) over chore(core) given the bug label.

PR #7318, opened 2026-06-23
fix(sql): fix wrong results from queries on renamed Parquet columns

### LINKED ISSUE #7316: WalWriterFuzzTest.testCreateTableAsParquet() deterministic failure with given seeds
### Summary

`WalWriterFuzzTest.testCreateTableAsParquet` fails with:

```
java.lang.AssertionError: Actual cursor does not have record at 0
  TestUtils.assertEquals(:294) <- assertSqlCursors(:1040/:1021)
    <- FuzzRunner.checkIndexRandomValueScan(:913) <- assertRandomIndexes(:423) <- runFuzz
  random seeds: 1504463487934207962L, 2622578810447144450L
```

Root cause: an equality predicate on a SYMBOL column returns wrong (missing)
rows when the data is in a Parquet partition, because the partition was written
with WAL-local symbol keys instead of the table's global keys.

### To reproduce

In `WalWriterFuzzTest.testCreateTableAsParquet`, pin the seed:

```java
Rnd rnd = new Rnd(1504463487934207962L, 2622578810447144450L);
```

Reproduces on macOS/APFS (CI failed on linux/ZFS), on roughly 4 of 5 runs - the
remaining runs pass because the outcome depends on concurrent WAL-apply
ordering. The failing check compares `... where "new_col_5" = 'YV'` between the
Parquet WAL table (`_wal`) and the non-WAL reference (`_nonwal`); `_wal` returns
0 rows where `_nonwal` returns 6.

### It is not the index, not JIT, not data loss

Diagnostics captured at the failure point (`new_col_5` is the symbol column
`c2`, after a `c2 -> new_col_5` rename):

| query on the Parquet `_wal` table | rows |
| --- | --- |
| `count()` total | 5025 (equals the reference) |
| `new_col_5 = 'YV'` (key compare) | **0** |
| `cast(new_col_5 as string) = 'YV'` (string compare) | **6** |
| `new_col_5 = 'YV'` with JIT disabled | **0** |

- The plans are filter scans (no index is used), and the failure reproduces
  identically for all four index types (BITMAP, POSTING, POSTING DELTA,
  POSTING EF), so it is not the index implementation (#6861 is exonerated).
- JIT-off still returns 0, so it is not a JIT codegen bug.
- The totals match and the 6 rows are physically present (the string scan finds
  them), so there is no data loss.
- The same equality on the native `_nonwal` table returns 6. The defect is
  Parquet-specific.

### Mechanism

- WAL writers assign per-segment **local** symbol keys (`WalWriter.localSymbolIds`);
  these are remapped to **global** table keys on apply
  (`TableWriter.processWalCommitBlock_remapSymbols` ->
  `Vect.remapSymbolColumnFromManyAddresses`).
- QuestDB stores Parquet symbol columns as `LocalKeyIsGlobal`
  (`parquet/qdb_metadata.rs:83`, enforced at `parquet_read/decode.rs:1329`): the
  stored integer key is asserted to equal the global key. The Parquet file also
  carries its own local string dictionary.
- The block-apply override at `TableWriter.java:10101-10121` replaces a symbol
  column with its remapped (global-key) buffer only when the block introduced
  **new** symbols (the `-1` indicator at line 10107); otherwise it keeps the raw
  WAL-local buffer (line 10096).
- When a symbol column reaches the Parquet writer with un-remapped local keys,
  the file ends up with local keys plus a locally-consistent string dictionary,
  but tagged `LocalKeyIsGlobal`. Therefore:
  - `cast(sym as string)` resolves each row via the local string dictionary, so
    it returns the correct rows.
  - `sym = 'literal'` resolves the literal to the global key and compares the
    local stored keys, so it matches nothing.

This is the symbol-column analog of #7310, which fixed a different failure mode
(var-column pointers -> table suspension) in the same `processWalCommitBlock`
Parquet path. The intermittency follows from concurrent WAL-apply ordering
deciding whether the local keys happen to coincide with the global keys; the
`c2 -> new_col_5` rename perturbs symbol/column assignment.

### Suggested fix direction

Ensure symbol columns destined for a Parquet partition are remapped to global
keys on the block-apply path even when the block adds no new symbols (or do not
tag the column `LocalKeyIsGlobal` when the stored keys are not global). One
piece is not yet hard-confirmed: that the block-apply path is the only trigger
(versus the row-by-row O3 apply). That is the next probe.

bluestreak01:
Fixes #7316

## Summary

Equality and IN filters on a renamed column of a Parquet partition could
return too few rows (often zero), silently dropping valid data. The defect
is in Parquet row-group bloom-filter pushdown, not in any index or in the
bloom filter itself.

The pushdown resolves the filtered column against the Parquet file's column
names. Those names are frozen when the partition is converted to Parquet, so
a later `RENAME COLUMN` leaves them stale. When some other column already
carries the query's current name (common after a chain of renames), the
pushdown looks up that name, lands on the wrong Parquet column, checks that
column's bloom filter, gets a false negative, and skips the whole row group
before any row is scanned. The rows are never returned.

The normal data-decode path was never affected: it maps columns by stable
column id (`getColumnId`, which equals the table writer index), so it always
reads the correct column regardless of renames. Only the pruning path used
name resolution, which is why a plain `col::string = 'x'` (cast suppresses
pushdown) returned the rows while `col = 'x'` did not.

## Root cause walkthrough

Reproduced from `WalWriterFuzzTest.testCreateTableAsParquet`
(seeds `1504463487934207962L, 2622578810447144450L`). The fuzz table renamed
`c2 -> new_col_5` while a different column had earlier been named `new_col_5`.
In the Parquet metadata:

- index 0: `c2` (id=1) -- the actual symbol column holding `'YV'`
- index 10: `new_col_5` (id=15) -- an unrelated column

The query `WHERE "new_col_5" = 'YV'` resolved the name `new_col_5` to Parquet
index 10 and tested that column's bloom filter, which does not contain `'YV'`,
so the row group was pruned and the query returned 0 rows. The reference
non-WAL table returned the 6 matching rows.

## Fix

Make the native-table pushdown resolve columns by stable id, the same way the
data-decode path already does:

- `PushdownFilterExtractor.PushdownFilterCondition` carries the column's stable
  writer index (column id).
- `ParquetMetaFileReader` gains `getColumnIndexById()`, mirroring
  `PageFrameMemoryPool.buildColumnIdMap`'s `id < 0 ? positional` fallback for
  external files without field ids. Native-table partitions reach the pruning
  path through `ParquetPartitionDecoder.metadata()`, which returns a
  `ParquetMetaFileReader`, so this is the only resolver that runs by id.
- `ParquetRowGroupFilter.prepareFilterList` takes a `resolveByColumnId` flag.
  Native-table cursors (`FwdTableReaderPageFrameCursor`,
  `BwdTableReaderPageFrameCursor`) pass `true` and resolve by id.
  `read_parquet()` cursors pass `false` and keep name resolution, matching how
  `read_parquet()` already projects external files by name in
  `canProjectMetadata`.

## Cleanup

`prepareFilterListImpl` previously carried a by-id branch on
`ParquetFileDecoder.Metadata` as well, but no caller ever reached it: the only
callers of the `Metadata` overload are the two `read_parquet()` cursors, and
both pass `resolveByColumnId=false`. The commit drops the unreachable
`ParquetFileDecoder.Metadata.getColumnIndexById()` overload and collapses the
dead ternary into a direct `ParquetMetaFileReader` call, guarded by an assert
that documents the invariant.

## Effects and tradeoffs

- Correctness: filters on renamed columns of Parquet partitions now match the
  same rows as native partitions. This is the data-loss fix.
- `read_parquet()` behaviour is unchanged: it keeps name-based resolution, which
  is the right semantics for arbitrary external files (no QuestDB column ids,
  columns matched by name).
- Pruning effectiveness is unchanged for the common (non-renamed) case: when the
  name and id point at the same column the result is identical; only the lookup
  key changed. Row groups are still pruned for genuinely absent values (covered
  by tests that assert the prune still fires).
- Negligible cost: `getColumnIndexById` is a linear scan over the partition's
  column list, run once per partition during filter-list preparation, the same
  shape as the existing `getColumnIndex(name)` scan it replaces. No extra work
  on the per-row or per-row-group path.
- Scope limited to the row-group pruning column resolution; the bloom filter
  format, statistics, and decode paths are untouched.

## Test plan

- `ParquetRowGroupPruningTest#testBloomFilterSymbolRenamedColumn` renames a
  symbol column into a name previously held by another column, then asserts
  equality on the bloom-filter path returns the right rows and that an absent
  value still prunes.
- Six additional rename-scenario tests cover the min/max-statistics path, which
  shares the same column resolution: `testRenamedColumnMinMaxEquality`,
  `testRenamedColumnRange`, `testRenamedColumnBetween`, `testRenamedColumnInList`,
  `testRenamedColumnOrEquality`, and `testRenamedColumnIsNull` (the last covering
  the null-count path for `IS [NOT] NULL`). Each swaps two columns' names so the
  stale parquet name targets the wrong column's statistics, then asserts matching
  rows survive and genuinely-absent values still prune.
- Verified that all six new tests, plus the existing bloom test, fail when the
  by-id resolution is reverted to by-name, and pass with the fix.
- `WalWriterFuzzTest#testCreateTableAsParquet` passes on the reported seeds and
  across several fresh random-seed runs.
- `ParquetRowGroupPruningTest` (127), `ReadParquetFunctionTest` (110),
  `ParquetTest`, sqllogic `parquet.ParquetTest` (46), and
  `AlterTableRenameColumnTest` (30) all pass.


bluestreak01:
## Review (level 3, mission-critical pass)

Full pass: change-surface map, all agent dimensions (correctness, concurrency, performance/optimality, resources, tests, code quality, metadata, cross-context, adversarial), and per-finding source verification. No Rust files in the diff.

### Change summary (verified)
The bug: native-table Parquet row-group pruning resolved the filtered column against the **frozen Parquet column name**. After `RENAME COLUMN` that name is stale; when another column already carries the query's current name, pruning checked the *wrong* column's bloom filter / stats and dropped the whole row group, causing silent data loss. The decode path was unaffected because it resolves by stable column id (`writerIndex`).

The fix threads the column's stable `writerIndex` into `PushdownFilterCondition` and adds `getColumnIndexById()` so native-table pruning resolves by id (`resolveByColumnId=true`), while `read_parquet()` keeps name resolution (`false`). The resolution chain checks out end-to-end:

- `PartitionEncoder.populateFromTableReader` writes `columnId = getWriterIndex()` into the `_pm` metadata at conversion (frozen, never changes on rename).
- Native pruning now calls `ParquetMetaFileReader.getColumnIndexById(writerIndex)`, reading the same `parquetMetaReader.getColumnId(i)` that the decode path's `PageFrameMemoryPool.buildColumnIdMap` already consumes, keyed by the same `metadata.getWriterIndex()`. **Pruning resolution == decode resolution.**

The fix is also strictly broader than the reported symptom: it additionally repairs the plain-rename case (old code silently lost pruning) and the drop-then-re-add-same-name case (old code could wrongly prune).

### Critical
None.

### Moderate
None.

### Minor

1. **Dead code: `ParquetFileDecoder.Metadata.getColumnIndexById` is currently unreachable** (`ParquetFileDecoder.java:484`). The only callers of the `ParquetFileDecoder.Metadata` overload of `prepareFilterList` are the two `read_parquet()` cursors, and both pass `resolveByColumnId=false`, so they always take the `getColumnIndex(name)` branch. The active fix path uses `ParquetMetaFileReader.getColumnIndexById` (native tables go through `ParquetPartitionDecoder.metadata()`, which returns `ParquetMetaFileReader`). The `Metadata` version is defensible as symmetric API but is not exercised — consider a short comment noting it's there for symmetry, or drop it. Non-blocking.

2. **Test coverage is narrow relative to the fix's reach** (`ParquetRowGroupPruningTest.java:817`). `testBloomFilterSymbolRenamedColumn` exercises only SYMBOL equality (bloom path). The same resolution change governs min/max-stats range filters, `IN`, OR-of-equalities, `BETWEEN`, and `IS [NOT] NULL` on renamed columns — none are tested for the rename scenario. They all funnel through the one `prepareFilterList` resolution, so the single test does cover the resolution fix, but a range/min-max test on a *renamed numeric* column would guard the broader contract cheaply. The test itself is well-formed: `.returns(...)` (not `returnsOnce`), the `assertMemoryLeak` + `.noLeakCheck()` pairing is legitimate (the lambda holds many statements sharing one leak scope), and the prune-still-fires assertion (`getRowGroupsSkipped() > 0`) is correctly reset beforehand and robust against `.returns()`'s double pass.

### Downgraded (false positives, verified and dismissed)

- **"Constructor signature change breaks callers"** — all 7 `new PushdownFilterCondition(...)` sites are in `PushdownFilterExtractor` and updated; a repo-wide grep (incl. tests/benchmarks) found none elsewhere.
- **"`writerIndex` is lost through `GenericRecordMetadata.copyOfNew`"** — `copyColumns` shallow-copies the `TableColumnMetadata` references, so `getWriterIndex()` returns the original stable id at all three `extractAndCompile` callsites.
- **"Pruning resolution may diverge from the decode path"** — both read the same `parquetMetaReader.getColumnId(i)` keyed by the same `getWriterIndex()`; results are identical (ids are unique for native tables, so the hashmap-last-wins vs linear-scan-first-wins distinction is moot).
- **"`getWriterIndex` throws / is invalid for `read_parquet` metadata"** — returns `-1` (those `TableColumnMetadata` have `writerIndex=-1`); harmless because `read_parquet` passes `false` and never reads it.
- **"`getColumnIndexById` linear scan is a perf regression"** — identical O(columns) shape to the `getColumnIndex(name)` scan it replaces, run once per partition; the decode path's O(1) map isn't warranted for a handful of conditions.
- **"`id < 0` positional fallback could mismatch for native tables"** — native conversion always writes `writerIndex >= 0`; the fallback only matters for external `read_parquet` files, which resolve by name anyway.
- **Concurrency** — the new field is a `final int` set at compile time and read-only thereafter; no new shared mutable state.

### Summary
- **Verdict: approve.** The fix is correct, minimal, and well-targeted; it aligns the pruning column-resolution with the already-correct decode path, and the description's effects/tradeoffs are accurate.
- No regressions or tradeoffs identified: pruning effectiveness is unchanged for non-renamed columns, `read_parquet()` semantics are preserved, and the cost is the same per-partition scan as before.
- Draft findings: ~9 considered, 2 verified (both Minor), 7 dropped as false positives after source verification.
- In-diff vs out-of-diff: 2 in-diff, 0 out-of-diff. The zero out-of-diff count is justified — every consumer of the changed contract was traced (decode path in `PageFrameMemoryPool`, conversion in `PartitionEncoder`, metadata copy in `GenericRecordMetadata`, and all three `SqlCodeGenerator` callsites) and each upholds the new id-based resolution by construction.

<sub>Generated with a level-3 automated review pass.</sub>

jovfer (approved):
LGTM, if we OK to ignore dead code (minor 1 in the review above)

bluestreak01:
@jovfer thanks, dead code was removed

image

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 22 / 24 (91.67%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/ParquetMetaFileReader.java|4|5|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ParquetRowGroupFilter.java|7|8|87.50%|
|:large_blue_circle:|io/questdb/griffin/engine/table/PushdownFilterExtractor.java|11|11|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=245231)
**Tested Commit:** `6f6f0b4f0488a1f8b0f78e3bdfa456696528ac57`

PR #7319, opened 2026-06-24
chore(sql): keep the per-query memory tracker balanced across aggregate re-execution

RaphDal:
## Summary

A parallel GROUP BY / aggregate query could release its per-query `MemoryTracker` back to the provider pool with a non-zero `used` counter. Because trackers are pooled and recycled, the next query that acquired the dirty block started mis-accounted: its `query_activity.memory_used` was over-reported and its per-query memory limit was correspondingly weakened, both by a small fixed amount. This is a per-query *attribution* error, not a leak — global RSS stays balanced and the divergence is bounded — but it violates the `used == 0`-at-recycle invariant the tracker pool relies on everywhere else.

The fix stops charging the allocator's bookkeeping index to the per-query counter, so each query's tracker comes back clean and the index can stay retained across cursors with no per-cursor churn.

## Root cause

The defect is in the GROUP BY allocator's chunk-index lifecycle.

`FastGroupByAllocator` keeps a small native hash map (`chunks`, a `DirectLongLongHashMap`, ~256 bytes at its initial capacity) that indexes the data chunks it hands out to group-by function state. The allocator bound the per-query tracker to both the data chunks *and* this index, and `clear()` (cursor close) retained the index via `restoreInitialCapacity()`. The retained index therefore stayed charged to the query that first grew it. When that query released its tracker to the pool, the tracker's `used` still held the index's bytes; the next query that recycled the block inherited the dirty counter.

The residue left charged at recycle is the index at its initial capacity (16 bytes x initial capacity = 256 bytes) per allocator, times the number of allocators an atom reopened (owner plus any per-worker allocators): `clear()` shrinks the index back to initial capacity before the tracker is released, so the dirty amount is this fixed size regardless of the query's data volume. The live index itself grows with the distinct-chunk count during execution (see Tradeoffs).

## Fix

`core/src/main/java/io/questdb/griffin/engine/groupby/FastGroupByAllocator.java`:

- `setMemoryTracker()` no longer binds the per-query tracker to the chunk index. The index is allocator bookkeeping whose footprint is bounded by the data chunks it indexes — about 16 bytes per chunk (chunks are >= 128 KiB), under ~0.02% of the tracked data — so it stays on the global counter, the kind of negligible structure the per-query tracker deliberately leaves out. Only the data chunks (query-proportional, unbounded) remain charged to the per-query tracker.
- Because the index is no longer on the per-query counter, `clear()` retains it across cursors as before (`restoreInitialCapacity()`), and `malloc()` / `free()` need no extra guards. The retained-index optimization is preserved.

The per-query tracker now accounts only for the data chunks, which `clear()` already frees at cursor close while the tracker is bound, so the tracker comes back to the pool clean. Comment-only updates in the five aggregate/join atoms (`AsyncGroupByAtom`, `AsyncGroupByNotKeyedAtom`, `BaseAsyncHorizonJoinAtom`, `BaseAsyncMultiHorizonJoinAtom`, `AsyncWindowJoinAtom`) describe the new split: data chunks on the per-query tracker, the index on the global counter. The `GroupByAllocator` interface javadoc and the `GroupByAllocatorFactory` javadoc are corrected to match, stating the chunk index stays on the global counter and is never bound to the per-query tracker.

`core/src/main/java/io/questdb/std/PerQueryMemoryTracker.java`:

- `init()` asserts `getUsed() == 0` at the recycle boundary, as a permanent regression guard.

The fix lands in `FastGroupByAllocator`, so it covers every consumer: the keyed and not-keyed async GROUP BY atoms, the horizon / multi-horizon join atoms, the window join atom, and the synchronous GROUP BY / SAMPLE BY paths.

## Tradeoffs

- The chunk index no longer counts against the per-query memory limit. On paths where the index was the only per-query-tracked allocation at cursor open — parallel GROUP BY and the window joins, whose maps and slave buffers are allocated lazily during the reduce rather than at open — a pathologically tiny limit (smaller than the first map/data allocation) no longer trips *at cursor open*; it trips on the first drain instead. Realistic limits are unaffected, because they are crossed by the data chunks / maps during execution regardless. Paths that allocate a map at cursor open (synchronous GROUP BY, ASOF joins, sync horizon joins) still breach at open. The breach-at-open tests for the lazy-map paths were updated to expect the breach during execution; the rest are unchanged. Because the index is data-proportional but tiny (about 16 bytes per >= 128 KiB chunk, under ~0.02% of the tracked data), `query_activity.memory_used` and the per-query limit now exclude it; the data chunks they still count dominate by orders of magnitude.
- The `init()` assertion ships enabled in production: the standard launcher (`core/src/main/bin/questdb.sh`), the Docker entrypoint (`core/docker-entrypoint.sh`), and the AMI systemd unit (`pkg/ami/marketplace/assets/systemd.service`) all run the server with `-ea`. The `used == 0`-at-recycle invariant is therefore a hard check, not only a test-time one. The intended payoff is fail-fast detection of any future tracker-bound path that releases a tracker dirty; the cost is that such a regression would throw `AssertionError` from `PerQueryMemoryTracker.init()` on the query-registration path under `-ea`, surfacing as a failed query / WAL apply / mat-view refresh rather than the bounded, self-correcting attribution drift it replaces. (The Windows service wrapper in `win64svc/` does not pass `-ea`.) An audit of every per-query-tracker-bound structure found the invariant already holds today, so the assert guards against regressions rather than masking a known violation.

## Test plan

- The recycle invariant is covered by the existing suites, which fail on the old code (`tracker recycled with used=256` / `used=1280`) and pass on the fix: `ParallelParquetMemoryTrackerTest` (parallel GROUP BY over a `CONVERT PARTITION TO PARQUET` partition, with a `WorkerPool` and `assertInTree` pinning the async factory) and `ParallelGroupByMemoryTrackerTest` (keyed and not-keyed, non-parquet). A standalone parquet recycle test added earlier was dropped: its parquet machinery was not load-bearing (the leak is the storage-independent chunk index), one of its three cases routed to the synchronous factory and did not reproduce the bug, and it duplicated the suites above.
- `PerQueryMemoryTracker.init()` asserts the `used == 0`-at-recycle invariant. Because the standard launcher ships `-ea`, the check is live in production as well as under test (see Tradeoffs), so any future regression on a tracker-bound path fails fast at the recycle boundary.
- `PerQueryMemoryTrackerTest.testAcquireAssertsCleanRecycleOnDirtyRelease` directly pins the guard: it charges a tracker, releases it dirty, and asserts the next `acquire()` throws `AssertionError` carrying `tracker recycled with used=256` — a focused negative test for the invariant itself, independent of the GROUP BY paths.
- The breach-at-open tests on the lazy-map paths (`ParallelGroupByMemoryTrackerTest`, `ParallelHorizonJoinMemoryTrackerTest`, `ParallelWindowJoinMemoryTrackerTest`, `SyncWindowJoinMemoryTrackerTest`) now drive the breach to the first drain and assert the per-query out-of-memory there; they continue to verify that the breach releases allocations and the factory stays reusable. The breach-at-open tests on map-at-open paths (`GroupByAllocatorMemoryTrackerTest`, `SyncHorizonJoinMemoryTrackerTest`, `JoinMemoryTrackerTest`) are unchanged and still trip at cursor open.
- Verified green with the probe active (no `recycled with used`, no crash): `ParallelGroupByMemoryTrackerTest`, `ParallelParquetMemoryTrackerTest`, `ParquetMemoryTrackerTest`, `GroupByAllocatorTest` (including `testCanBeUsedAfterClear`), all four join tracker suites, `GroupByAllocatorMemoryTrackerTest`, `SampleByTest`, `GroupByTest`, and `ParallelGroupByFuzzTest` (whose `recordPerQueryMemAlloc` balance assertion is live on the default tracker).


RaphDal:
# Review: PR #7319 — `chore(sql): keep the per-query memory tracker balanced across aggregate re-execution`

**Reviewed at level 2.**

**Verdict: Approve.** No Critical or blocking findings. This is a tightly-scoped, correct fix; the bulk of the diff is comments and tests. The one thing worth a conscious maintainer decision is the production-enabled assertion (see Moderate #1), which the PR already documents.

The behavioral delta is exactly one line of production logic — `FastGroupByAllocator.setMemoryTracker()` no longer calls `chunks.setMemoryTracker(tracker)` — plus the new `assert getUsed() == 0` in `PerQueryMemoryTracker.init()`. Everything else is comment/javadoc edits and test changes.

---

## Verification scope (what was actually checked, beyond the diff)

The diff is small but its blast radius is large (the new assert runs on **every query** — see below), so the review covered far more than the changed files: the full `setMemoryTracker` inventory (~60 owning classes), the async atoms, sharding context, top-K, latest-by, parquet page-frame pool, window functions, join cursors, sort/order-by cursors, union/distinct, and the mat-view/WAL workload boundaries.

- **Memory model** (`Unsafe.java`): `malloc/free/realloc(size, tag, tracker)` *always* updates global RSS; the per-query counter is updated *additionally* only when a tracker is bound. Moving the index to a `null` tracker keeps it on global RSS (no leak) and removes it from the per-query counter — exactly the intent. No double-count.
- **The index is now permanently on the global counter**: `DirectLongLongHashMap.memoryTracker` defaults to `null`, and the removed line was its only binder (grep-confirmed). Its malloc/free are now consistently balanced on global RSS.
- **Atom lifecycle**: `clear()` frees data chunks while the allocator's tracker is still bound (it's nulled only in `close()`), so the per-query tracker returns to the pool at `used == 0`. Confirmed across `AsyncGroupByAtom`, `AsyncGroupByNotKeyedAtom`, `BaseAsyncHorizonJoinAtom`, `BaseAsyncMultiHorizonJoinAtom`, and `AsyncWindowJoinAtom` (including its second `ownerTemporaryAllocator` family and the mid-reduce `clearTemporaryData()` recycle).
- **The `init()` assert's blast radius — the PR's "audit holds today" claim is verified true for OSS.** The bug class is "a tracker-charged native structure *retained across cursors* (not freed at teardown)." That pattern — `restoreInitialCapacity()` in a teardown path — exists in exactly **two** places repo-wide: the now-fixed `FastGroupByAllocator.chunks`, and `SortKeyMaterializingRecordCursor.rowIdToOrdinal` (constructed with **no** tracker, never bound → already on the global counter). Every other tracker-bound structure is *closed* (fully freed) under the bound tracker at cursor teardown:
  - async-atom maps (verified `OrderedMap.clear()` frees no native memory, but the atoms call `Misc.free` / `close`), top-K heaps, latest-by maps;
  - parquet buffers — `PageFrameMemoryPool.clear()` calls `releaseParquetBuffers()` (frees all cached buffers, zeroes `cachedBytes`) *before* nulling the tracker;
  - window-function `map` / `memory` / `dequeMemory` (freed in `reset()`); own-`MemoryARW` window classes `.close()` them;
  - join key maps / chains / sink targets (`Misc.free` at cursor close); `RecordChain` / `RecordArray` / tree chains;
  - Rosti (`pRosti`) and encoded-sort key heaps are deliberately **not** tracker-bound at all.
  - Mat-view refresh (`MatViewRefreshJob`) and WAL apply (`OperationExecutor`) null the tracker only after the inner SQL has run and freed its allocators.
  - Ordering proof: cursor `close()` (→ `base.close()`) strictly precedes tracker release (`QueryProgress.RegisteredRecordCursor.close()` → `QueryRegistry.unregister()` → `tracker.close()`).
- **`init()` runs on every query**, not just memory-limited ones: `QueryRegistry` acquires a tracker even when the limit is 0 (`QueryRegistry.java:173-187`, "so that accounting stays on and query_activity.memory_used reports live usage for every query"). This cuts both ways — the existing test suite exercises the assert broadly (a common leak would turn the whole suite red, strongly corroborating the audit), and a *future* regression fails every query under `-ea`, not just limited ones.
- **Concurrency**: no new exposure. The index moving to the global RSS counter (`LongAdder` + atomic RMW) is touched by the same N workers that already updated it for data chunks; each worker owns a distinct allocator/index (thread-confined); the reduce/post-aggregation latch barrier and the drain-on-breach (`UnorderedPageFrameSequence` "still wait for in-flight tasks ... to avoid data races") ensure all worker frees are visible and complete before `clear()` and before recycle.
- **Tests**: the breach-at-drain rewrites are sound and non-vacuous — `while (cursor.hasNext()) {}` + `Assert.fail` can only pass on a genuine breach (empty result → `Assert.fail` fires; a global-RSS breach → wrong message → the `assertContains("query memory limit exceeded", "workload=QUERY")` fails). The rewrite is *necessary*: with the index off the tracker, the lazy-map paths no longer breach at open. The new `testAcquireAssertsCleanRecycleOnDirtyRelease` correctly pins the assert directly (pool `pop()` → `init()` asserts before resetting `used`; cleanup credits the leak back and recycles cleanly).
- **Quantitative claims all verified**: 256 B initial index (`ceilPow2(11) = 16` × 16 B); production default chunk = `128*1024` (`PropServerConfiguration.java:1689`); 16 B/entry → ~0.012–0.017 %; `used=1280` = 1 owner + 4 workers × 256 B. All four `-ea` claims confirmed with file:line (`questdb.sh:360`, `docker-entrypoint.sh:124`, `pkg/ami/.../systemd.service:25`; absent in `win64svc/src/questdb.c`).

---

## Critical

None.

---

## Moderate

1. **The `init()` assertion ships enabled in production (`-ea`) — accept consciously.** This is the PR's central design decision, not a defect, and the Tradeoffs section documents it honestly: any *future* tracker-bound path that releases dirty will throw `AssertionError` on the query-registration path (failed query / WAL apply / mat-view refresh) rather than producing bounded, self-correcting drift. The audit confirms no such path exists in OSS today, so the assert guards against regressions rather than masking a live violation. Because it runs on *every* query, the failure mode is broad if a future change reintroduces the pattern — but that is exactly the fail-fast the PR intends, and the dedicated negative test plus the existing suites lock it in. Worth an explicit maintainer sign-off, nothing more.

2. **Recycle-invariant coverage is indirect for non-GROUP-BY workloads.** The shipped assert + `ParallelGroupByMemoryTrackerTest` / `ParallelParquetMemoryTrackerTest` cover the GROUP BY and parquet paths directly, and `testAcquireAssertsCleanRecycleOnDirtyRelease` covers the guard itself. Mat-view refresh and WAL apply rely on the same allocators and are covered transitively (and by the assert running on their trackers), but there is no dedicated recycle test for those two workloads. Acceptable given the shared code; a one-line note in the test plan that these are covered transitively would help future readers.

---

## Minor

1. **`Bug` label vs `chore(...)` title.** Mildly inconsistent — `chore` signals an internally-discovered correctness fix (consistent with the "fix = customer bug only" convention), while `Bug` implies a tracked defect. Defensible either way; `SQL` and `Core` labels fit.
2. **The "~0.02 %" figure is the production-default ratio.** With a non-default small chunk size (e.g. the 16 KiB embedded `DefaultCairoConfiguration` default), the index-to-data ratio rises to ~0.1 % — still negligible, and the PR honestly frames it as "about 16 bytes per >= 128 KiB chunk," but the percentage is config-dependent.

---

- **`realloc()` slow path doesn't free the old pointer** (`FastGroupByAllocator.java:180-184`) — not a tracker imbalance: the old pointer is interior to a chunk that stays a `chunks` entry and is freed under the tracker at `_close()`. Pre-existing intra-arena fragmentation, unchanged by this PR.
- **`close()`-without-`clear()` leaving live data chunks charged** — unreachable: teardown is `frameSequence.close()` → `reset()` → `atom.clear()` (frees under bound tracker) → only then `atom.close()`; the construction error path runs `close()` before any `reopen()`, so the lazy index is unopened and `_close()` is a no-op under the `chunks.isOpen()` guard.
- **`SortKeyMaterializingRecordCursor.rowIdToOrdinal` retains via `restoreInitialCapacity()` at `close()`** — safe: never tracker-bound, so it retains on the global counter (same shape as the fixed index).
- **`PageFrameMemoryPool` retains tracker-charged parquet buffers across cursors** — false: `clear()` calls `releaseParquetBuffers()` (frees every cached buffer, `cachedBytes = 0`) before nulling the tracker.
- **New concurrency exposure from the index moving to the shared global counter** — none: same counters, same workers, same barriers as the data chunks already used.

---

## Summary

- **Verdict: Approve.** The fix is correct on every path through `FastGroupByAllocator` and its consumers; the chunk index moves cleanly to the global counter with no new imbalance, no RSS leak, and no concurrency exposure. The new recycle assert is sound, correctly positioned, and its "invariant holds today" claim is verified across the OSS tracker-bound surface.
- **Regressions / tradeoffs:** the assert is a deliberate, documented fail-fast that ships under `-ea` in production; a future retain-while-charged regression would fail queries rather than drift. That is the intended behavior, with the negative test and existing suites as the guard.
- **Draft findings:** ~10 candidates raised across 5 review agents + independent verification; **0 confirmed as bugs**, 5 explicitly dismissed after source verification (above). The remainder resolved to the 2 Moderate / 2 Minor non-blocking notes.
- **In-diff vs out-of-diff:** 0 in-diff bugs, 0 out-of-diff bugs. The out-of-diff zero is backed by an extensive cross-context pass (the assert's whole-repo blast radius), not an underrun — the change is a pure *restriction* (removing a tracker binding) whose only risk surface is the new assert, which was audited across the full `setMemoryTracker` inventory.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 46 / 51 (90.20%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/sql/NetworkSqlExecutionCircuitBreaker.java|4|9|44.44%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinLightNoKeyRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/ReadParquetRecordCursor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/ShowStandardConformingStringsCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/PgExtensionFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/PgClassFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/AllTablesFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/MemoryMetricsRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/ShowDateStyleCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/ShowSearchPathCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/ShowMaxIdentifierLengthCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/FilesFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/ExportFilesFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/ShowTimeZoneFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/ShowTransactionIsolationLevelCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/AtomicBooleanCircuitBreaker.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/KeywordsFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/ShowDefaultTransactionReadOnlyCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/TableWriterMetricsRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/ShowParametersCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/PgDatabaseFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/InformationSchemaColumnsFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/PerQueryMemoryTracker.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/LtJoinNoKeyRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/CopyExportException.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/TablesFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/ImportFilesFunctionFactory.java|1|1|100.00%|

bluestreak01 (approved):
Approve — reviewed at level 3 (full mission-critical pass with per-finding source verification).

The production behavioral delta is two lines: `FastGroupByAllocator.setMemoryTracker()` no longer binds the chunk index to the per-query tracker, and a new `assert getUsed() == 0` in `PerQueryMemoryTracker.init()`. The rest is comment/javadoc and test changes.

Verified against source:
- Fix correctness: chunk-index ops are now consistently global-counter-tracked (`DirectLongLongHashMap.memoryTracker` stays null, the removed line was its only binder); data chunks stay on the per-query tracker and are freed at `clear()` under the bound tracker, so the tracker recycles clean.
- Atom lifecycles balance: all five atoms free tracker-bound data chunks under the still-bound tracker before nulling, and rebind in `reopen()` before any allocation; `AsyncWindowJoinAtom` temp-allocator family + mid-reduce `clearTemporaryData()` recycle confirmed.
- Assert audit holds: the retained-across-cursors pattern exists only in the now-fixed `chunks` index and `SortKeyMaterializingRecordCursor.rowIdToOrdinal` (never tracker-bound -> global, won't trip).
- Quantitative claims confirmed: 256 B initial index, 128 KiB prod chunk (PropServerConfiguration.java:1689), -ea in questdb.sh/docker-entrypoint.sh/systemd.service and absent in win64svc, FastGroupByAllocator is the sole GroupByAllocator impl.
- Tests: dropped ParquetGroupByTrackerRecycleTest correctly absent; breach tests insert 1,000 rows so the drain genuinely breaches (non-vacuous); the new negative test pins the guard directly and cleans up with no leak.

No Critical or Moderate-blocking findings. Two non-blocking notes worth a conscious sign-off:
1. The init() assert ships under -ea in production; because the tracker pool is shared across QUERY / MAT_VIEW_REFRESH / WAL_APPLY, a future dirty-release on any workload would throw AssertionError on whichever workload next recycles that pooled block (possibly an unrelated user query). Documented in Tradeoffs; the audit shows no such path exists today.
2. The rewritten breach tests now accept a breach at open OR drain (the catch wraps both), so they no longer pin the breach location -- intended and necessary, but slightly less specific.

Minor: Bug label vs chore() title; the ~0.02% ratio is the production-default figure (config-dependent).

PR #7321, opened 2026-06-24
feat(qwp): add per-query SYMBOL dict reset flag

kafka1991:
> 🔗 **Tandem PR — merge together with its Enterprise counterpart:**
> **Enterprise:** questdb/questdb-enterprise#1112

## Why

**This matters specifically for the Python client's dataframe egress** (`to_polars()` / `to_pandas()`) — that is the only place it changes anything meaningful, and it helps to see why those paths are fast in the first place.

A `SYMBOL` column arrives on the wire as a dictionary (the distinct string values) plus per-row integer codes. The dataframe consumers turn that into a pandas/polars **Categorical** the most efficient way there is: they **reuse the QWP wire dictionary directly as the Categorical's dictionary** — the per-row codes are adopted zero-copy and the wire dictionary becomes the categories as-is, with no re-interning, no re-encoding, and no per-row remap. That direct reuse is what makes `SYMBOL → dataframe` maximally efficient.

The catch is that the QWP dictionary is **connection-scoped**: it accumulates every symbol seen across all queries on a connection (resetting only at a soft cap). Because the dataframe reuses that dictionary directly, a result that uses few symbols but lands on a connection whose dictionary was grown by an earlier high-cardinality query inherits a **bloated Categorical** — categories the result never references. The very zero-copy reuse that makes the path fast is what drags the cross-query bloat in.

## What

To fix that, this PR lets a client scope the dictionary to a single query. It adds an optional, per-query `query_flags` field to the QWP egress `QUERY_REQUEST`; its first bit, `QUERY_FLAG_RESET_DICT`, asks the server to reset the connection-scoped SYMBOL dictionary *before* this query streams. The directly-reused dictionary then holds exactly this query's symbols — keeping the zero-copy speed without the bloat. Support is advertised via a new `SERVER_INFO.capabilities` bit, `CAP_QUERY_FLAGS`.

The flag is opt-in; in the Python client it is defaulted on only for the dataframe egress path.

### Why `to_arrow()` / `__arrow_c_stream__` don't use this (they ship a compact dictionary instead)

The zero-copy reuse above only works for **registry-aware** consumers — `to_polars()` / `to_pandas()` understand the QWP connection-scoped global codes and adopt them directly. Generic Arrow consumers can't: `to_arrow()` hands out a pyarrow Table and `__arrow_c_stream__` feeds any Arrow-compatible library (DuckDB, Polars, pandas-via-pyarrow, …), and the Arrow C Data Interface has no concept of a connection-scoped (or delta) dictionary shared across batches. Each `DictionaryArray` must be self-contained, so a consumer that receives batches with different dictionaries unifies them itself.

Those paths therefore ship a **compact** dictionary: each batch carries only the distinct symbols it actually references (local codes). Since a generic consumer unifies per-batch dictionaries regardless, compact is strictly better there — far less data on the wire (on a wide-symbol result the Arrow C-stream payload was ~14× smaller than carrying the full connection dictionary per batch) and **already immune to cross-query bloat**, since a batch never carries a symbol it doesn't use. `QUERY_FLAG_RESET_DICT` would be a no-op for them, so they don't set it.

In short: full connection dictionary + per-query reset is for the registry-aware dataframe consumers (zero-copy reuse, kept lean); a compact per-batch dictionary is for generic Arrow consumers (self-contained, minimal, bloat-free by construction).

### Why per-query (not a connection-level setting)

The dictionary is a connection-scoped *resource*, but "reset it before this query" is a *per-request intent*. Per-query keeps each request free to choose its own policy — the client is pooled (many queries share one connection) and will grow modes beyond dataframes (raw row scans, custom Arrow consumers, mixed workloads) that may want different policies on the same connection. `query_flags` is a varint, so future per-query options get a home **without another protocol change**.

### Why it's backward compatible (no version bump)

- `query_flags` is an **optional trailer** after the bind section, detected by payload length — a baseline client leaves `p == limit` and it defaults to 0.
- A client appends it **only when the server advertised `CAP_QUERY_FLAGS`**, so an older server never sees a trailing byte it can't parse.
- old server + new client → no cap → no trailer (unchanged); new server + old client → no trailer → `queryFlags = 0` (unchanged); new ↔ new → honoured.

### Reuses the existing reset path

A set flag just ORs `RESET_MASK_DICT` into `computeCacheResetMask`; the dict is cleared at the query boundary and a `CACHE_RESET` frame is emitted before the first `RESULT_BATCH` — exactly what the client already handles. An empty-dict guard skips the redundant frame when there's nothing to reset.

## Client

Client side (Python `py-questdb-client` + its `c-questdb-client` submodule): https://github.com/questdb/py-questdb-client/pull/130

Python `Client.query(..., reset_symbol_dict=True)` defaults the flag on for the dataframe egress path; Rust/FFI default off, and the flag only reaches the wire when the server advertises `CAP_QUERY_FLAGS`.


kafka1991:
## Code review notes

Reviewed the QWP `query_flags` trailer change end-to-end (decoder, processor state, upgrade processor, frame writer) and against the callsites. The core protocol logic is correct: the non-flag path is byte-for-byte unchanged, the trailer parse is bounds-safe and backward-compatible, and the dict reset stays in client/server lockstep. A few items worth addressing:

### Moderate

**Stale javadoc on `QwpEgressConnSymbolDict.clear()` (out-of-diff, `QwpEgressConnSymbolDict.java:108-112`)**

The comment still reads:

> Clears the dict completely. Called on connection drop / reset, **never on per-query boundaries** (conn-ids allocated in earlier queries stay live so the client's view is consistent).

This PR makes exactly a per-query-boundary clear of this dict reachable: `applyCacheResetForUpcomingQuery` -> `applyCacheReset` -> `connSymbolDict.clear()` (`QwpEgressProcessorState.java:367-372`) now fires when a query carries `QUERY_FLAG_RESET_DICT`. The behavior is safe (the staged `CACHE_RESET` frame keeps the client in lockstep), but the "never on per-query boundaries" claim is now false and misleading. Please update it to describe the new opt-in per-query reset path.

### Minor

**1. `RESET_DICT` on a non-SELECT defers the client-visible reset to the next SELECT.**
When a non-SELECT (DDL/INSERT/UPDATE) carries the flag on a connection with a non-empty dict, the server clears its dict immediately and stages the mask, but `executeNonSelect` returns via `sendExecDone` without ever calling `emitPendingCacheReset` (only reachable from `streamResults`). So the `CACHE_RESET` frame goes out on the *next* SELECT, not "before this query" as the constant's javadoc states. Verified non-corrupting (no RESULT_BATCH is sent in between, so client/server stay in lockstep, and the official client only defaults the flag on for the SELECT-only dataframe path). Worth a sentence in the `QUERY_FLAG_RESET_DICT` javadoc clarifying the reset is applied at the query boundary and surfaces on the next result-producing query. A test exercising the non-SELECT-then-SELECT deferred-emit sequence would lock down the subtlest correctness path in the change.

**2. `queryFlags` init placement (`QwpEgressRequestDecoder.java:204`).**
`queryFlags` is zeroed at the trailer, not at method entry. If `decodeQueryRequest` throws before that line (truncated SQL, bad bind), the field retains the previous request's value. Harmless today (the only reader runs after a successful decode), but inconsistent with `reset()` which clears it, and would bite a future reader added before the success check. Move `queryFlags = 0` up next to `sql.clear()` at the top.

**3. PR metadata.**
- Title description should be verb-led per repo convention (release notes copy it verbatim): e.g. `feat(qwp): add per-query SYMBOL dict reset flag`.
- No labels set — add ones consistent with the change (e.g. `New feature`/`Enhancement`, `Core`).

### Verified safe (no action needed)

- **Backward compat:** the baseline client writes nothing after the bind payload, so existing frames leave `p == limit` and `queryFlags` defaults to 0. Unconditional trailer parse is fine.
- **Concurrency:** the decoder is per-connection and the IO dispatcher guarantees one worker per fd; the new `queryFlags` field adds no shared state.
- **Native memory:** `clear()` retains the heap/entries buffers (freed only in `close()`), so repeated forced resets cause no native churn or leak.
- **Empty-dict guard:** server-dict-empty implies client-dict-empty under the lockstep invariant; the fail-then-empty sequence preserves the previously staged mask.
- **Byte-value overlaps** (`QUERY_FLAG_RESET_DICT`/`RESET_MASK_DICT`/`CAP_ZONE`/`CAP_QUERY_FLAGS`): distinct namespaces, never conflated; `CAP_QUERY_FLAGS` doesn't touch the `CAP_ZONE` bit, so SERVER_INFO size is unaffected.
- **Malformed varint trailer:** bounds- and 10-byte-capped; violations throw `QwpParseException` -> clean `QUERY_ERROR`.

bluestreak01:
@kafka1991 The decode + state-mutation paths are well covered, but the new `QUERY_FLAG_RESET_DICT` path has no **end-to-end wire test**. Could you add coverage before merge?

The current tests stop at:
- decode only (`QwpEgressRequestDecoderTest`)
- decode -> state mutation, no wire (`QwpEgressQueryFlagsResetTest`)
- capability advertisement (`QwpEgressServerInfoTest`)

The actual over-the-wire `CACHE_RESET` emission is exercised **only for the cap-triggered path** (`QwpEgressCacheResetWireTest`), never for the flag path. The two paths share the `emitPendingCacheReset` machinery, so risk is low — but the subtlest correctness path is untested:

1. **Flag path emits CACHE_RESET before the first RESULT_BATCH.** A SELECT with `query_flags=RESET_DICT` on a connection with a non-empty dict must put `CACHE_RESET` on the wire before any `RESULT_BATCH`, with the next batch's `deltaStart=0`.
2. **Deferred emit: non-SELECT-then-SELECT.** A non-SELECT carrying the flag clears the server dict immediately but returns via `sendExecDone` without calling `emitPendingCacheReset`, so the staged `CACHE_RESET` surfaces on the *next* result-producing query. This non-SELECT -> SELECT deferred sequence is the trickiest path in the change and currently has zero coverage. A regression here would silently desync the client/server dict view.

`QwpEgressCacheResetWireTest` already has the harness to assert frame ordering on the wire — extending it (or a sibling) to drive the flag instead of the cap should be straightforward.

Separately, the empty-dict guard (`testFlagSetButEmptyDictIsNoOp`) is covered at the state level but a wire-level assertion that *no* `CACHE_RESET` frame goes out would round it off.

kafka1991:
@bluestreak01 Added the end-to-end wire coverage in `QwpEgressQueryFlagsResetWireTest` (a sibling of `QwpEgressCacheResetWireTest`, caps left at production defaults so the flag is the only reset trigger). It drives the flag through `client.execute(..., resetSymbolDict=true)` and asserts both the server `cacheResetDictCount()` metric and full client-side symbol resolution after the flush:

- `testResetFlagFiresCacheResetBeforeBatch` — a flagged SELECT on a non-empty dict fires exactly one `CACHE_RESET` and still resolves every symbol, proving the frame lands before the first `RESULT_BATCH` (the next delta restarts at 0).
- `testResetFlagOnNonSelectDefersToNextSelect` — a non-SELECT carrying the flag clears the dict and stages the reset via `EXEC_DONE` without emitting; the next SELECT surfaces the deferred `CACHE_RESET` and resolves all symbols, with the counter advancing once, on the non-SELECT. This is the deferred non-SELECT -> SELECT path you flagged.
- `testResetFlagOnEmptyDictEmitsNoFrame` — the flag on an empty dict emits no `CACHE_RESET` frame (counter unchanged) — the wire-level complement to the state-level empty-dict guard.
- `testResetFlagScopesEachQuery` — across a run of flagged SELECTs, the first (empty) is a no-op and each later one resets the dict the previous left behind.

All four pass. I also dropped the headless `QwpEgressQueryFlagsResetTest`, since the decode -> state path it covered is already exercised by `QwpEgressRequestDecoderTest` + `QwpEgressProcessorStateCacheResetTest`, and the full decode -> reset path is now covered end-to-end over the wire. Pushed in d874d0c.

kafka1991:
## Code review (level 3, full pass)

Reviewed the QWP `query_flags` trailer change end-to-end across the decoder, processor state, upgrade processor, frame writer, capability negotiation, and the bumped client submodule — plus the callsites outside the diff. Verdict below; substantive items first.

### Verdict: approve, with one cross-repo follow-up

The protocol logic is correct end-to-end: the non-flag path is byte-for-byte unchanged, the trailer parse is bounds-safe and backward/forward compatible, the dict reset stays in client/server lockstep, there are no native leaks, and there are no concurrency hazards.

### Moderate

**M1. Enterprise `QwpServerInfoProvider.getCapabilities()` must also advertise `CAP_QUERY_FLAGS` (out-of-diff, cross-repo).**
The OSS `DefaultQwpServerInfoProvider.getCapabilities()` now returns `CAP_QUERY_FLAGS`, and `writeServerInfoFrame` (`QwpEgressUpgradeProcessor.java:685`) passes the capabilities word straight to the wire — there is **no auto-OR** of `CAP_QUERY_FLAGS` anywhere (verified `QwpEgressFrameWriter.java:178`; same manual contract as `CAP_ZONE`). `CairoConfiguration.getQwpServerInfoProvider()` defaults to the OSS provider, but enterprise substitutes its own implementation that overrides `getCapabilities()`. If that override is not updated to OR in `CAP_QUERY_FLAGS`, an enterprise `SERVER_INFO` won't advertise it, the client's `resolveQueryFlags` won't append the trailer, and the feature **silently no-ops against enterprise servers** (no crash — the shared egress decoder still honors a trailer if one arrives; the gap is purely the un-advertised capability). The interface javadoc was updated to "every implementation should advertise it," so the intent is clear — this just needs a matching enterprise change landed alongside. Not blocking for the OSS merge.

### Minor

**m1. `markCacheResetDict()` counts at apply-time, not emit-time (informational, pre-existing).** The metric is incremented when the mask is *staged* (`QwpEgressUpgradeProcessor.java:1311`), not when `emitPendingCacheReset` ships the frame. On the deferred non-SELECT path (or a flagged query that errors), the reset is counted even though the `CACHE_RESET` frame goes out on a later query — and if the connection drops in between, it's counted with no frame ever reaching the client. Identical to the pre-existing cap path, so not a regression, but the name reads as "frames emitted." Worth a one-line doc note.

**m2. Optional hardening: post-`EXEC_DONE` shrink can emit a duplicate `QUERY_ERROR`.** On the non-SELECT branch, `state.getBatchBuffer().resetForNewQuery()` (`:1118`) runs *after* `executeNonSelect` already sent `EXEC_DONE` (`:828`). If that shrink threw (near-impossible — it's a shrinking `Unsafe.realloc`), the outer `catch (Throwable)` would emit a `QUERY_ERROR` for the same `requestId` after the client already received `EXEC_DONE`. Not a leak or crash, just a redundant frame. Wrapping that best-effort housekeeping call in its own try/catch would make it airtight.

**m3. Test coverage gaps (low priority).** Coverage is otherwise thorough (22/22 new lines; decode + state + wire). Two shared-machinery paths lack a dedicated test:
- A flagged SELECT that **errors during compile/streaming with a reset mask already staged** — confirm the deferred mask still emits on the next SELECT. (The logic is correct: the `catch` path never clears `pendingCacheResetMask`.)
- **Two flagged non-SELECTs back-to-back** (second sees an empty dict, no-op, no double-count).

Both are nice-to-haves; the deferral machinery is already well covered by `testResetFlagOnNonSelectDefersToNextSelect`.

**m4. Boolean naming (very minor).** `forceDictReset` / `cacheResetApplied` / `capExceeded` don't use the `is`/`has` prefix from the coding guidelines, though this matches existing practice for locals.

### Verified safe (investigated, not findings)

- **Trailer backward/forward compat:** `query_flags` is the first-ever optional trailing field, so no pre-existing field leaves stray bytes; a baseline frame leaves `p == limit` and defaults to 0. The varint decode is bounds-checked against `limit` (malformed trailer throws `QwpParseException` -> clean `QUERY_ERROR`); `p` cannot exceed `limit` after the bind loop. Client appends the trailer only when `CAP_QUERY_FLAGS` is advertised. All four old/new x client/server combos hold.
- **Deferred non-SELECT emit:** the staged `pendingCacheResetMask` survives because the next query's `applyCacheResetForUpcomingQuery` returns early on `mask == 0` without clobbering it. The wire test is a **strong** regression guard despite Q1/Q3 returning identical symbol sets: the client enforces `deltaStart == connDictSize` (`QwpResultBatchDecoder.java:749`) and throws on mismatch, so a missing deferred reset would surface Q3's `deltaStart=0` against a stale `connDictSize=4` -> decode error -> `onError` -> `Assert.fail`.
- **Empty-dict guard:** server-dict-empty is a reliable proxy for client-dict-empty under the lockstep + pending-mask invariant.
- **Shrink-ownership refactor** (`clearConnKeyMaps()` vs `resetForNewQuery()`): correct, and it removes a pre-existing double-shrink (old code ran `resetForNewQuery` in both `applyCacheReset` and `beginStreaming`, the second `updatePeakUsage()` over-shrinking against already-zeroed peaks). No path resets the dict without the single shrink eventually running; no unbounded scratch growth.
- **`requestId` seeding before decode:** bounds-safe (`length >= 9` gate), offset/encoding match the decoder's raw LE `Unsafe.getLong(payload+1)`.
- **Concurrency:** one worker per fd; `queryFlags` is a per-connection pooled field, zeroed at decode entry and in `reset()`, read once into a local before any park point. No shared mutable state, correctly non-volatile.
- **Native memory:** `clear()` / `clearConnKeyMap(s)` do zero native work (Java `Arrays.fill` only); the dict's native heap/entries are retained and freed only in `close()`. Per-query forced resets cause no native churn or leak. No new tracker-worthy allocation site is introduced (scratch/dict buffers are connection-lived and self-capped, correctly not wired to a per-query `MemoryTracker`).
- **Stale `clear()` javadoc** (raised earlier): already fixed in the current diff.
- **PR metadata:** verb-led Conventional Commits title; `New feature` + `Core` labels; analytical body covering the backward-compat matrix and the `to_arrow()` tradeoff.

### Summary

- 1 Moderate (out-of-diff, the enterprise provider) + 3 Minor + 1 very-minor retained; ~8 draft findings verified as false positives / pre-existing and dropped.
- In-diff vs out-of-diff: 3 in-diff (all Minor), 1 out-of-diff (M1, highest value). The out-of-diff item is exactly the contract the change quietly depends on at a callsite the diff doesn't touch.
- No functional regressions. The shrink refactor is a small net improvement. The only deliberate tradeoff is the one the description documents (full-dict-plus-reset for registry-aware consumers vs. compact per-batch dicts for generic Arrow consumers).

bluestreak01 (dismissed):
Approve.

Reviewed the QWP `query_flags` trailer change end-to-end (decoder, processor state, upgrade processor, capability advertisement, frame ordering) and against the callsite inventory. The protocol logic is correct: the non-flag path is byte-for-byte unchanged, the optional trailer is bounds-safe and backward-compatible (length-detected, defaults to 0, gated by CAP_QUERY_FLAGS), masking never sign-extends, and the dict reset stays in client/server lockstep across the SELECT, deferred non-SELECT, failed-compile, and repeated-non-SELECT paths. The empty-dict guard and the apply-time metric are consistent and don't double-count. Test coverage at the protocol level is thorough.

No blocking issues. A few minor, non-blocking suggestions:

- A flagged/cap SELECT that throws before `beginStreaming` no longer opportunistically shrinks the scratch buffers (the shrink re-add covers only the non-SELECT success branch). Verified harmless — failed/non-streaming queries can't grow the buffers and the next SELECT shrinks — but the success-only asymmetry is worth a one-line comment so it doesn't read as a missed case.
- New booleans `capExceeded` / `cacheResetApplied` would read better with the `is` prefix per the repo convention.
- `markCacheResetDict()` counts at apply time, not frame-emit time, so the metric isn't a 1:1 count of CACHE_RESET frames sent — worth a doc line if it's user-facing.
- Optional: a single end-to-end test that grows the dict with a high-cardinality SYMBOL query then asserts the server-side `connSymbolDict.size()` is lean after a flagged low-cardinality query would directly lock down the bloat-elimination the feature exists for (currently covered only compositionally).
- Consider adding a `Performance` label given the body's framing.

bluestreak01 (dismissed):

bluestreak01:
## Code review (level 3, tandem)

Reviewed the `query_flags` trailer change end-to-end — decoder, processor state, upgrade processor, capability negotiation, the shrink refactor, and the bumped client submodule — plus the callsites outside the diff and the enterprise tandem side.

### Verdict: approve

The protocol logic is correct end-to-end. I independently verified the subtle paths and they hold:

- **Deferred non-SELECT / failed-SELECT lockstep** — the staged `pendingCacheResetMask` survives because the next query's `applyCacheResetForUpcomingQuery` returns early on `mask == 0` *without* clobbering it; `emitPendingCacheReset` then ships it before the first `RESULT_BATCH` with `deltaStart=0`. Traced non-SELECT→SELECT, flagged→flagged, and back-to-back non-SELECTs.
- **Empty-dict guard** — every `connSymbolDict.clear()` is paired with a staged pending mask (or is a connection teardown that also resets the client), so server-dict-empty is a sound proxy for client-dict-empty.
- **Shrink refactor** (`clearConnKeyMaps` vs `resetForNewQuery`) — correct, and a net improvement: it removes the pre-existing double `resetForNewQuery` (apply-time + `beginStreaming`). No path resets the dict without the single shrink eventually running; no unbounded scratch growth.
- **Trailer parse** — bounds- and 10-byte-capped; a lone `0x80` throws `QwpParseException` → clean `QUERY_ERROR`. Baseline frame leaves `p == limit` → defaults to 0. Backward/forward compat holds across all four old/new × client/server combos.
- **request-id pre-seed** — `length >= 9` gate keeps the peek in bounds; endianness matches the decoder and the LE test builder; `requestId` is a method-local.
- **Concurrency** — one worker per fd; `queryFlags` is a per-connection pooled field zeroed at decode entry and in `reset()`, read into a local before any park point; non-volatile is correct.
- **Native memory** — `clear()`/`clearConnKeyMap(s)` do zero native work; the dict heap is retained and freed only in `close()`. Per-query forced resets cause no churn or leak.

Test coverage (decode + state + wire) is thorough. A couple of minor, non-blocking items:

### Minor

**1. Duplicate `QUERY_ERROR` risk on the non-SELECT reset path (near-impossible).** `state.getBatchBuffer().resetForNewQuery()` runs *after* `executeNonSelect` already sent `EXEC_DONE`. If that shrinking `Unsafe.realloc` threw, the outer `catch (Throwable)` would emit a second frame for a request the client already saw completed. Not a leak/crash (factory/cursor are null on this branch). Optional: wrap the best-effort shrink in its own try/catch.

**2. `markCacheResetDict()` counts at apply-time, not emit-time (doc-only, pre-existing pattern).** On the deferred/failed paths (or a connection drop before the next SELECT), the metric increments even though the `CACHE_RESET` frame ships later or never. Identical to the cap path, so not a regression, but the name reads like "frames emitted" — worth a one-line doc note.

**3. Shrink-asymmetry comment.** A flagged/cap SELECT that throws before `beginStreaming` no longer opportunistically shrinks the scratch buffers. Verified harmless (a query that never streamed can't have grown them, and the next SELECT shrinks), but a one-line comment at the deferred site would stop it reading as a missed case.

**4. Boolean naming (very minor).** `forceDictReset` / `cacheResetApplied` / `capExceeded` don't use the `is`/`has` prefix, though this matches existing local practice.

### Tandem reminder (cross-repo)

The enterprise counterpart must advertise `CAP_QUERY_FLAGS` for this to actually take effect on enterprise servers — `EntQwpServerInfoProvider` is currently constructed with `capabilities = 0` and only ORs in `CAP_ZONE`, so the client's `resolveQueryFlags` never appends the trailer and the feature silently no-ops there. The client gates the trailer strictly on the advertised bit, so without the enterprise change (and a matching cap-advertisement test) the feature is dead on enterprise. Flagging it here so the two PRs merge in lockstep.

bluestreak01:
## Review — level 3, tandem (OSS side; ENT counterpart: questdb/questdb-enterprise#1112)

Reviewed together with the enterprise PR as one logical change, including the `java-questdb-client` bump. Built a full change-surface map and callsite inventory with `rg`, verified every draft finding against source, and ran the new tests locally: 72 decoder/state/serverinfo tests, all 7 `QwpEgressQueryFlagsResetWireTest` e2e wire tests, and the client-side `QwpQueryClientQueryFlagsTest` cap-gating coverage check out.

Verified in particular:
- Trailer parsing is backward/forward compatible as claimed (absent ⇒ 0, malformed ⇒ clean `QwpParseException`, high bits preserved, no bleed across pooled decodes).
- The empty-dict guard is safe: `connSymbolDict.clear()` only ever executes alongside `clearConnKeyMaps()`/`batchBuffer.reset()`, so dict-empty ⇒ key-maps-empty holds on every path.
- The `applyCacheReset` split (clear-only, shrink owned by `beginStreaming` / the explicit non-SELECT site) is correct at its single production callsite — and quietly fixes a latent double-shrink on the pre-existing cap path.
- Deferred `CACHE_RESET` emission stays in lockstep across the non-SELECT, failed-SELECT, and disconnect paths (`state.clear()` drops mask and dict together).
- The `requestId` pre-seed matches the decoder's byte order (`Unsafe.getLong`, native LE) and is pinned by the raw-wire malformed-trailer test.
- Mid-stream reset is impossible: the `isStreamingActive()` gate rejects a second QUERY_REQUEST before decode.
- Capability bits don't collide (`CAP_ZONE=0x01`, `CAP_QUERY_FLAGS=0x02`) and both `QwpServerInfoProvider` implementations (OSS default + enterprise) advertise the new bit.

### Critical
None.

### Moderate
None.

### Minor
1. `QwpEgressRequestDecoder.decodeQueryRequest` — bytes after the `query_flags` trailer are silently ignored (no `p == limit` check after the trailer). Deliberate forward-compat posture consistent with the pre-existing laxity after the bind section, but it also swallows a buggy client's double-encoded trailer. Consider a strict end-of-frame check; a future optional field must be gated by its own capability bit anyway.
2. `QwpEgressRequestDecoder.java` — the final `p += varintScratch.bytesRead;` is a dead store (last read of `p`). Harmless cursor-pattern consistency; drop it or leave a comment.
3. `java-questdb-client`: `QwpEgressMsgKind.QUERY_FLAG_RESET_DICT` is declared `byte`, but the server-side mirror is deliberately `int` with a javadoc explaining the sign-extension hazard for bit 7+. Fine for `0x01`, but the mirror should adopt `int` now so the trap never materializes.
4. Style: `QwpEgressServerInfoTest.testOssServerAdvertisesQueryFlagsCapability` uses an inline FQCN for `QwpEgressMsgKind` instead of an import; `QwpEgressQueryFlagsResetWireTest` similarly uses inline `java.util.Set`/`java.util.HashSet` FQCNs throughout.

### Summary
**Approve.** Well-engineered: opt-in, capability-gated, byte-identical baseline frames, and the trickiest paths (deferred emission through EXEC_DONE and the error path) are explicitly designed and tested. Findings are minor polish only. One tandem-process note lives on the ENT PR: the enterprise submodule pointer must be re-aimed at the post-squash master SHA after this merges.

bluestreak01 (dismissed):

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 26 / 26 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/qwp/codec/QwpResultBatchBuffer.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressUpgradeProcessor.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/codec/DefaultQwpServerInfoProvider.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/codec/QwpColumnScratch.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressRequestDecoder.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressProcessorState.java|4|4|100.00%|

PR #6879, opened 2026-03-17
chore(core): replace /*** with /*+** in copyright headers

mtopolnik:
Change the opening line of the copyright notice from `/***` to `/*+**`. This removes 20,000+ inspection warnings from the project.

PR #7322, opened 2026-06-24
fix(sql): fix view reads failing due to incomplete dependency tracking

glasstiger:
A view's dependency map records every base table the view can scan, along
with the columns referenced from each. It is built once at `CREATE` /
`ALTER VIEW` and consulted at read time to authorize the scan of each
base table (`AbstractPartitionFrameCursorFactory.authorizeSelect`). Two
independent gaps let a view read fail; this branch addresses both.

## Fix 1: capture base tables reached only through a sub-query

The collector (`SqlUtil.collectColumnReferencesFromExpression`) walked
nested, join and union models, but never descended into sub-queries
embedded in expressions -- `WHERE col IN (SELECT ... FROM t)`,
`EXISTS (...)`, scalar sub-selects. A base table reachable only through
such a sub-query was left out of the map.

When that table is later scanned through the view, `authorizeSelect`
looks it up in the incomplete map, gets `null`, and the read fails. The
same gap can also cause a base table the view legitimately uses to be
treated as not-covered-by-the-view, producing an incorrect permission
denial.

`collectColumnReferencesFromExpression` now recurses into
`expr.queryModel`, so tables and columns referenced inside expression
sub-queries are recorded too -- at any depth, including inside CTEs and
inside the body of a referenced view.

The change only affects dependency collection at `CREATE` / `ALTER VIEW`
(the method is reachable from nowhere else), so it does not touch any
query execution path. It does not retroactively repair already-stored
maps: a view created before this change keeps its map until it is next
created/altered, so an existing view that already trips this bug still
needs a `CREATE OR REPLACE` / `ALTER VIEW` to rebuild its map.

## Fix 2: fail safe when a base table is missing from the dependency map

Fix 1 closes the known collector gap at the source, but the lookup site
stays fragile. The collector is a hand-rolled walker that must mirror
every clause capable of carrying a table reference; any residual or
future gap re-introduces the same failure. A base table can also go
missing for a non-bug reason -- it was renamed after the view was
created, so the map still holds the old name.

`authorizeSelect` previously dereferenced the looked-up column set
directly, so a `null` (missing entry) aborted the read with a
`NullPointerException`. The NPE already failed closed -- it aborted the
read rather than leaking data -- so this is a robustness/UX issue, not a
security hole.

The guard now treats a `null` column set as "this base table is not
covered by the view" and falls back to requiring explicit per-column
SELECT on the base table, the secure default the method already applies
for non-covered columns. Any residual gap thus becomes correct,
merely-stricter authorization instead of a crash.

Tradeoff: the fallback can mask a real collector gap by surfacing it as
a stricter denial rather than a crash. A caller holding only view-level
SELECT (not base-table SELECT) would be denied where a complete map
would have allowed the read. That is still strictly safer than an NPE
and never leaks data, but it makes the underlying inconsistency less
obvious. Keeping the collector correct (Fix 1) remains the primary
defense; the guard is the safety net.

## Test plan

- `ViewDependencySubqueryTest` asserts the dependency map captures the table for:
  - a table referenced only via `WHERE ... IN (SELECT ...)`
  - a table referenced only via `WHERE ... NOT IN (SELECT ...)`
  - a sub-query nested two levels deep inside a CTE
  - a sub-query that reaches a table through another view
  - a read-back of such a view
  - these five fail without Fix 1 and pass with it
- `ViewDependencyGapAuthorizationTest` simulates a missing entry by removing it from the live dependency map and checks that:
  - an allow-all caller reads cleanly instead of hitting the NPE
  - a caller granted view SELECT but denied base-table SELECT is denied rather than crashed
  - with the entry present the base-table check is suppressed, so the same denied caller still reads the view
  - the two missing-entry cases fail with the NPE when the guard is reverted and pass with it
- `ViewQueryTest` passes unchanged


bluestreak01 (approved):
Approved.

Verified both fixes by reverting each and observing the precise failure it prevents:
- Reverting Fix 1 (SqlUtil sub-query recursion) fails all 5 ViewDependencySubqueryTest cases.
- Reverting Fix 2 (null guard) reproduces the exact NPE at AbstractPartitionFrameCursorFactory.authorizeSelect:139 in the 2 missing-entry tests.
- With both present, 8/8 new tests pass; ViewQueryTest (38) and ViewModificationTest (4) pass unchanged, so Fix 1 does not over-collect.

Both fixes are correct, narrowly scoped, and fail-closed. Fix 1 only records tables/columns the view definition genuinely reads (definer-rights semantics, no over-grant); Fix 2's fallback is the strictest per-column base-table SELECT path, so it can only deny, never over-allow. No correctness, concurrency, resource, or security findings.

Minor (non-blocking) suggestions:
- ViewDependencyGapAuthorizationTest.java uses the old /*** copyright banner; repo convention (post #6879) and the sibling new file use /*+**.
- Residual collector gaps remain for SAMPLE BY / LATEST BY / FILL / LIMIT sub-queries (backstopped by Fix 2, as the PR notes) - worth a TODO.
- Consider an explicit EXISTS / scalar-subselect test (the comment names them but only IN/NOT IN are exercised), and asserting captured column sets, not just table-level membership.

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/AbstractPartitionFrameCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlUtil.java|2|2|100.00%|

PR #7324, opened 2026-06-24
chore(core): fix file-descriptor leak when the server shuts down with a query parked

### LINKED ISSUE #7296: sleep() over PGWire: a dropped connection's socket can leak during server shutdown (ServerMainSleepTest flake)
## Summary

`ServerMainSleepTest.testSleepCancelledByConnectionDrop` failed in CI on `linux-x86-graal` with a leaked file descriptor reported by `assertMemoryLeak`:

```
ServerMainSleepTest.testSleepCancelledByConnectionDrop:108 -> assertMemoryLeak:118
expected: cached fds 2, OS fds 1
actual:   cached fds 3, OS fds 2   (extra fd: 0x80097f9f / osFd=121, NON_CACHED == a socket)
```

This is a **hypothesis** write-up: the mechanism below is derived from reading the code and from eliminating alternatives by instrumentation, but I was **not** able to reproduce the failure locally (~400+ iterations across idle, `stress-ng --cpu 48`, and single-core `taskset` pinning — all green). The race window appears to be sub-microsecond on a fast multi-core box and only loses on a slow, contended agent.

Observed once on PR #7280 (build 244186), whose changes are unrelated to PGWire/network/shutdown — the const-folding PR only wraps scalar fixed-width runtime-constant subexpressions, and `sleep()` is a cursor/table function that is never wrapped.

### 1. A graceful client `conn.close()` mid-`sleep()` is not detected by the server

The parked `sleep()` body probes the SQL circuit breaker on each wake interval:

- `SleepFunctionFactory.sleep()` calls `executionContext.getCircuitBreaker().statefulThrowExceptionIfTrippedNoThrottle()`
- → `NetworkSqlExecutionCircuitBreaker.testConnection(fd)`
- → `NetworkFacadeImpl.testConnection()`:

```java
final int nRead = Net.peek(fd, buffer, bufferSize);
return nRead < 0;
```

`peek` reports "disconnected" only on `nRead < 0` (a hard socket error). The PG JDBC driver's `conn.close()` sends a `Terminate ('X')` message and then closes gracefully (FIN), so a non-blocking `peek` on the server side returns `> 0` (the buffered `X`) or `0` (EOF) — **never `< 0`**. The breaker therefore never trips, and the parked `sleep(60)` runs until the server shuts down.

(Note: the test/class comment claiming "the server's next wake probe sees the broken FD via the SQL circuit breaker and unwinds the body" does not hold for the graceful-close case — the body is actually unwound by shutdown, not by drop detection.)

### 2. The only reaper for this connection is server shutdown

`ServerMain.close()`:

```java
engine.signalClose();          // 1
if (initialized) {
    workerPoolManager.halt();  // 2
}
freeOnExit.close();            // 3 (frees PG server / dispatcher)
```

`CairoEngine.signalClose()` → `timerShards.shutdown()` drains the timer shards and, per entry, calls `TimerCont.shutdown()` → `cont.shutdown()` + `cont.scheduleResume()`. This enqueues the parked sleep continuation onto the worker continuation queue **while the worker pools are still RUNNING**, then returns. It does **not** wait for those resumes to actually be drained.

### 3. The race: the resume can be stranded by `halt()`

- `Worker`'s outer driver only remounts queued continuations `while (lifecycle.get() == WorkerLifecycle.RUNNING)` (`Worker.java:243`).
- `Worker.halt()` simply flips lifecycle to `HALTED` (`Worker.java:150`).
- `WorkerPool.halt()` waits for the worker **thread to exit** (via the halt latch), but exiting does not imply the scheduled resume was processed.

If the worker observes `HALTED` before it dequeues the resumed sleep continuation, the continuation is never remounted, the body never unwinds, and `IODispatcher.disconnect()` is never called for that connection.

### 4. Nothing else frees a checked-out context, so the socket leaks until JVM exit

- `IOContextFactoryImpl.close()` only sets a `closed` flag; it does not free **checked-out** contexts (the parked connection's context is not in the pool).
- A context's socket is freed only when `AbstractIODispatcher.doDisconnect()` runs for it, which happens only when a queued disconnect is processed — and no disconnect was ever queued in the stranded case.

So the `PGConnectionContext` and its socket fd are abandoned for the lifetime of the JVM. Because `assertMemoryLeak` baselines fds at the start of each test, only the test in which the race fires reports the leak; subsequent tests re-baseline against the elevated count, which is consistent with the single observed failure.

### Alternative ruled out

I instrumented `AbstractIODispatcher.disconnect()` to log when a disconnect is dropped because the dispatcher is already `closed` (the `cursor < 0` branch). It never fired across 80 stressed iterations. `bullyUntilClosed()` returns a valid cursor whenever the queue has room, so a late `disconnect()` would still enqueue — but `close()` runs its single `processDisconnects()` drain up front and never re-drains, so that path is not the cause here. By elimination, the leak is the stranded-resume path (no `disconnect()` is ever issued).

## Practical impact

This is a **shutdown-only** leak — the OS reclaims the fd on process exit, and nothing leaks during steady-state operation. The user-visible effect is limited to CI flakiness in `assertMemoryLeak`. A secondary observation worth noting: a client that drops its connection mid-`sleep()` is not detected and the server holds that connection (and worker-free parked sleep) for the full requested duration (up to the 24h cap), rather than reaping it promptly.

## Possible directions (not yet validated)

- **Shutdown sequencing:** have `signalClose()`/`timerShards.shutdown()` wait until all scheduled continuation resumes have been drained (continuation queue empty / parked conts completed) before worker pools halt, so resumes cannot be stranded.
- **Test hardening:** make `testSleepCancelledByConnectionDrop` reap the dropped connection deterministically while the server is still running (e.g. cancel the sleep query found via `query_activity()` and poll until it disappears) before closing `ServerMain`, removing the dependency on the racy shutdown drain.
- **Drop detection (separate concern):** consider whether a parked `sleep()` should detect a peer that has half-closed (`peek` returning `0`) once any buffered bytes are consumed, so dropped connections are reaped promptly rather than held for the full duration.

## Reproduction status

Not reproduced locally. Repro would likely need a slow/oversubscribed environment (the failing agent was `linux-x86-graal`) or a deliberately widened window between `scheduleResume()` and `halt()`.

## References

- Failing test: `core/src/test/java/io/questdb/test/ServerMainSleepTest.java` (`testSleepCancelledByConnectionDrop`)
- `core/src/main/java/io/questdb/network/NetworkFacadeImpl.java` (`testConnection`)
- `core/src/main/java/io/questdb/griffin/engine/functions/date/SleepFunctionFactory.java` (wake-probe loop)
- `core/src/main/java/io/questdb/cairo/CairoEngine.java` (`signalClose`)
- `core/src/main/java/io/questdb/mp/continuation/TimerShards.java` / `TimerCont.java` (shutdown drain + `scheduleResume`)
- `core/src/main/java/io/questdb/mp/Worker.java` (`halt`, RUNNING-gated remount loop)
- `core/src/main/java/io/questdb/network/IOContextFactoryImpl.java` (`close` / `done`)
- `core/src/main/java/io/questdb/network/AbstractIODispatcher.java` (`disconnect` / `close` / `doDisconnect`)

RaphDal:
Fixes #7296

## Summary

When `ServerMain` (or an embedded engine) closes while a query is parked on a worker continuation -- `sleep(...)`, or any query suspended on the query-continuation wake interval -- the server-side socket of that query's connection could be leaked. The continuation's resume is scheduled at shutdown but can be stranded in the worker resume queue if the worker pool halts before the resume runs; the connection context is then closed by neither the dispatcher nor the continuation machinery, so its file descriptor stays open for the lifetime of the JVM.

A single leaked descriptor at process exit is reclaimed by the OS and is harmless. The leak matters for long-lived JVMs that start and stop the engine repeatedly -- embedded deployments and the test harness -- where it accumulates across restarts. It is also what makes `ServerMainSleepTest.testSleepCancelledByConnectionDrop` flake on its `assertMemoryLeak` check: the leak check samples the descriptor count after the server closes and intermittently sees the stranded socket.

The change touches one production file, `core/src/main/java/io/questdb/mp/Worker.java`, and only the shutdown path; normal operation is unchanged.

## Root cause

1. `sleep()` parks the worker-loop continuation through `TimerCont` (`SleepFunctionFactory.java:201-220`). When the query is dispatched, its PG connection context is removed from the dispatcher's `pending` set (`IODispatcherLinux.java:99`), so while the query is parked the dispatcher no longer tracks that context.
2. At shutdown, `CairoEngine.signalClose()` drains the timer shards (`CairoEngine.java:2031`) before the worker pools halt. The drain sets each parked continuation's shutdown flag and calls `scheduleResume()`, pushing it onto the worker pool's `ContinuationQueue` so a still-running worker remounts it, the body observes `isShuttingDown()`, and it unwinds and disconnects. This ordering is intentional (`ServerMain.java:205` runs `signalClose()` before `workerPoolManager.halt()` at `223`).
3. The drain only *schedules* the resumes; it does not wait for them to run. `WorkerPool.halt()` then flips the worker lifecycle to `HALTED`, and the worker's loop exits (`Worker.loopBody:399`, outer driver `Worker.run:243`) **without draining the `ContinuationQueue`** -- the queue is drained inside the running loop (`Worker.loopBody:435`), not on the way out. Whether the worker drains the scheduled resume before it observes `HALTED` is a race.
4. When the worker loses that race, the parked continuation is stranded in the queue. It never unwinds, so it never disconnects. Its connection context is not in the dispatcher's `pending` set or any of its queues, so `AbstractIODispatcher.close()` (which sweeps only those) does not close it; and the shared-singleton dispatcher job keeps the no-op `closeInstance()` (`Job.java:64-77`), so the continuation framework does not free it either. The socket leaks.

The race is why the symptom is intermittent and load- and platform-sensitive: a quiet, fast machine usually drains the resume before halting, while a loaded CI agent (or a weaker memory model) often halts first.

## Fix

After a worker's `RUNNING` loop exits at halt, drain the `ContinuationQueue` before the worker thread stops. `Worker.drainShutdownContinuations()` remounts each parked continuation and runs it to completion, mirroring the outer driver's existing handoff walk, so the body observes `isShuttingDown()` and unwinds -- releasing the connection context (and any other checked-out resource) exactly as it would have during normal operation. This makes `signalClose()`'s documented intent ("workers remount them, the bodies observe isShuttingDown(), and unwind") hold deterministically instead of depending on a race.

A dequeued continuation can still be transiently mounted on the carrier that scheduled it. `TimerShards.register()` takes a synchronous shutdown path that calls `scheduleResume()` while the continuation is still executing the `register()` call -- one `suspend()` before it parks (`SleepFunctionFactory.java:210-211`). A peer worker that dequeues it during its own drain cannot remount it yet: `Continuation.run()` throws while the continuation is mounted on another carrier. Abandoning it there would re-introduce the same leak on a multi-worker pool -- the continuation would end up parked off-queue, drained by nobody. Instead, `mountForeignCont()` hands the continuation back to the queue; the drain re-dequeues and retries until the registering carrier reaches `suspend()` and unmounts, which it does within nanoseconds. The drain therefore runs every scheduled continuation to completion regardless of which carrier scheduled it or which worker observes it, rather than only when no peer steals a still-mounted continuation in that window. A phantom resume (a refused `suspend()` whose entry is already on the queue) is still dropped, not re-enqueued.

Properties:

- The drain runs only after the worker's running loop has exited, i.e. only at shutdown. The normal hot path is byte-for-byte unchanged; there is no per-connection or per-request bookkeeping.
- It reuses the existing remount machinery (`mountForeignCont`, the handoff walk), so it inherits the same continuation-lifecycle invariants rather than introducing new ones. The only behavioral change to `mountForeignCont()` is on its halt branch: it re-enqueues a still-mounted continuation instead of abandoning it.
- It generalizes beyond `sleep()` to any timer-parked continuation stranded at shutdown, on single- and multi-worker pools alike.

## Tradeoffs and limits

- Draining parked continuations during halt slightly extends shutdown when queries are parked: each stranded continuation runs its (short) unwind on the halting carrier. This is bounded by the existing `WorkerPool` halt timeout, which still proceeds if a worker does not stop in time.
- Retrying a continuation that is still mounted on its registering carrier is a brief spin on the resume queue, bounded by that carrier reaching `suspend()` -- nanoseconds, since the parked sleep is not carrier-pinned. A carrier that never unmounts (a genuinely wedged worker, which already blocks its own halt) does not make this worse: the same `WorkerPool` halt timeout proceeds and logs rather than letting the drain spin without end.
- A continuation that re-parks deep without honoring the `isShuttingDown()` loop-discipline contract would still be left mounted for the halt-timeout backstop rather than drained. No continuation in the OSS tree does this today; the contract is documented on `TimerCont.isShuttingDown()`.
- A test-only wait (poll the descriptor count before the leak check) was considered and rejected: the descriptor genuinely leaks, so a test-side wait would hide a real defect rather than fix it.

## Verification

- A reused-JVM reproduction -- boot `ServerMain`, run `sleep(60)`, force-close the connection, close the server, repeat under scheduling jitter -- leaked a server socket (decoded: a non-cached accepted descriptor) on 885 of 1000 iterations before the change and on 0 of 500 after it. (This was a temporary investigation harness and is not part of the branch.)
- `WorkerContinuationTest.testHaltDrainsContinuationScheduledJustBeforeHalt` schedules a continuation's resume while the single worker is in its backoff sleep, then halts. A handshake job makes the worker observably `RUNNING` first, so halt cannot win the `BORN -> RUNNING` race and skip the drain; each resume records -- from its stack -- whether it actually ran through `drainShutdownContinuations()`, so a drain regression cannot pass as a false green where the in-loop dequeue happens to cover. It fails without the fix and passes with it.
- `WorkerContinuationTest.testHaltReEnqueuesContinuationStillMountedOnAnotherCarrier` reproduces the multi-carrier window: a holder thread keeps a continuation mounted while it is scheduled for resume, then the pool worker is halted on a separate thread. With the fix the worker re-enqueues and retries, so `halt()` blocks until the continuation is released and drained (and the resume is confirmed to run through `drainShutdownContinuations()`); without the fix the worker abandons the still-mounted continuation and `halt()` returns promptly, which the test detects within the first iteration. It fails without the fix and passes with it.
- `WorkerContinuationTest.testHaltDropsPhantomResumeInsteadOfRetrying` guards the other side of the re-enqueue rule: a phantom resume (a stale queue entry left by a refused suspend, the continuation still mounted on its polling carrier) must be dropped, not retried. The holder keeps the continuation mounted and never releases it, and the entry is marked park-refused; halt must still complete promptly. With park-refused handling removed the drain spins on the never-unmounting continuation and `halt()` blocks until the timeout -- which the test detects -- so it confirms the re-enqueue path does not swallow phantoms.
- The continuation framework, worker-pool lifecycle, and sleep/wait_wal suites pass, including `WorkerContinuationTest`, `TimerContTest`, `TimerShardsTest`, `WorkerPoolManagerTest`, `WorkerPoolBootFailureTest`, `UnequalWorkerPoolTest`, `SleepFunctionFactoryTest`, `ServerMainSleepTest` (with the concurrent-sleep fuzz), and `ServerMainWaitWalTableTest`.

## Test plan

- Run the continuation/worker/sleep suites listed above and confirm they pass.
- Run `ServerMainSleepTest` repeatedly and confirm the `testSleepCancelledByConnectionDrop` leak check no longer trips.
- Optionally, exercise an embedded boot/close loop with a parked `sleep()` and confirm `Files.getOpenFileCount()` returns to its baseline after each close.

### DIFF
--- a/core/src/main/java/io/questdb/mp/Worker.java
+++ b/core/src/main/java/io/questdb/mp/Worker.java
@@ -287,7 +287,11 @@ public class Worker extends Thread {
                         // pool instead of leaking them in ownedJobClones until halt.
                         recycleJobList(cont);
                         cont = handoff;
-                        mountForeignCont(cont);
+                        if (!mountForeignCont(cont)) {
+                            // Dropped as a phantom or re-enqueued for the shutdown drain;
+                            // the cont is no longer ours to inspect. Stop walking the chain.
+                            break;
+                        }
                         if (cont.isDone()) {
                             recycleJobList(cont);
                             break;
@@ -301,11 +305,7 @@ public class Worker extends Thread {
                     // next cont runs shares no live state with any parked or spent cont.
                     currentJobsGen = mintNextGen(currentJobsGen);
                 }
-                // Lifecycle flipped to HALTED. signalClose() may have scheduled resumes for
-                // continuations parked on suspending queries (e.g. sleep) so their bodies
-                // observe isShuttingDown() and unwind on a live carrier, releasing checked-out
-                // resources such as connection contexts. The RUNNING loop can exit before
-                // draining them; finish here so they are not stranded with resources unreleased.
+                // Drain any continuations scheduled for resume after the RUNNING loop exited.
                 drainShutdownContinuations();
             }
         } catch (Throwable e) {
@@ -338,8 +338,11 @@ public class Worker extends Thread {
      * RUNNING loop exits. Mirrors the outer driver's handoff walk: each parked cont is
      * remounted and run to completion so its body observes {@code isShuttingDown()} and
      * unwinds, releasing any checked-out resource (e.g. a connection context). A cont
-     * parked long enough to reach here is past the benign mount-race window, so
-     * {@link #mountForeignCont} runs it rather than abandoning it.
+     * scheduled in the narrow window before its registering carrier reached suspend() is
+     * still transiently mounted there, so its remount throws and {@link #mountForeignCont}
+     * re-enqueues it; this loop re-dequeues and retries until that carrier unmounts
+     * (nanoseconds later), rather than stranding the cont off-queue with its resource
+     * unreleased.
      */
     private void drainShutdownContinuations() {
         if (continuationQueue == null) {
@@ -348,18 +351,30 @@ public class Worker extends Thread {
         final ContinuationQueue.ResumeTask scratch = new ContinuationQueue.ResumeTask();
         WorkerContinuation cont;
         while ((cont = continuationQueue.tryDequeue(scratch)) != null) {
-            mountForeignCont(cont);
+            if (!mountForeignCont(cont)) {
+                // Phantom resume (dropped), or re-enqueued because its registering carrier
+                // still has it mounted. A re-enqueue lands back on this queue, so pause and
+                // let the loop re-dequeue until the carrier unmounts -- bounded, since a cont
+                // on the resume queue is one suspend() away from parking.
+                Os.pause();
+                continue;
+            }
             while (!cont.isDone()) {
                 WorkerContinuation handoff = cont.takeHandoff();
                 if (handoff == null) {
-                    // Parked deep without a handoff; leave it for the halt-timeout backstop.
+                    // Re-parked deep without a handoff; nothing more to run here. In-tree
+                    // suspending functions throw on isShuttingDown() rather than re-park, so
+                    // this is not reached by them.
                     break;
                 }
                 recycleJobList(cont);
                 cont = handoff;
-                mountForeignCont(cont);
+                if (!mountForeignCont(cont)) {
+                    cont = null;
+                    break;
+                }
             }
-            if (cont.isDone()) {
+            if (cont != null && cont.isDone()) {
                 recycleJobList(cont);
             }
         }
@@ -540,24 +555,44 @@ public class Worker extends Thread {
      *       a cont that has already completed via another path. cont.isDone() is
      *       the structural test.</li>
      * </ol>
-     * Lifecycle bail-out: if this worker is halting, abandon the spin so the pool
-     * can shut down. The cont stays mounted on its carrier; when that carrier
-     * finishes its body the cont becomes done and is naturally disposed of.
+     * Lifecycle handling: if this worker is halting, it does not keep spinning to
+     * completion here -- a slow registering carrier could stall the pool's shutdown.
+     * Instead it re-enqueues the cont so the shutdown drain
+     * ({@link #drainShutdownContinuations}) retries it once that carrier unmounts,
+     * which happens within nanoseconds since a cont on the resume queue is one
+     * {@code suspend()} from parking. Returning the cont to the queue rather than
+     * abandoning it keeps a cont caught in the benign-mount-race window from being
+     * stranded off-queue with its checked-out resources unreleased.
+     *
+     * @return {@code true} if the cont ran on this carrier and is now parked or done,
+     * so the caller owns it and may inspect its handoff slot or recycle it;
+     * {@code false} if the cont was not run here and the caller must not touch it --
+     * either a phantom resume still owned by its polling carrier, or a cont
+     * re-enqueued for the shutdown drain to retry.
      */
-    private void mountForeignCont(WorkerContinuation cont) {
-        if (cont.consumeParkRefused() || cont.isDone()) {
-            return;
+    private boolean mountForeignCont(WorkerContinuation cont) {
+        if (cont.consumeParkRefused()) {
+            return false;
+        }
+        if (cont.isDone()) {
+            return true;
         }
         while (true) {
             try {
                 cont.run();
-                return;
+                return true;
             } catch (IllegalStateException e) {
-                if (cont.isDone() || cont.consumeParkRefused()) {
-                    return;
+                if (cont.isDone()) {
+                    return true;
+                }
+                if (cont.consumeParkRefused()) {
+                    return false;
                 }
                 if (lifecycle.get() != WorkerLifecycle.RUNNING) {
-                    return;
+                    // Halting: hand the cont back to the queue so the drain retries it
+                    // once its registering carrier unmounts, instead of stranding it.
+                    continuationQueue.put(cont);
+                    return false;
                 }
                 Os.pause();
             }
--- a/core/src/main/java/io/questdb/mp/Worker.java
+++ b/core/src/main/java/io/questdb/mp/Worker.java
@@ -250,9 +250,10 @@ public class Worker extends Thread {
                     cont.run();
                     if (cont.isDone()) {
                         // loopBody returned, which only happens when this worker
-                        // observed lifecycle HALTED. Exit the pool.
+                        // observed lifecycle HALTED. Break to drain any parked conts
+                        // before exiting the pool.
                         recycleJobList(cont);
-                        return;
+                        break;
                     }
                     // Walk the handoff chain. Each parked body, just before yielding,
                     // may have stashed a dequeued foreign cont in its handoff slot.
@@ -300,6 +301,12 @@ public class Worker extends Thread {
                     // next cont runs shares no live state with any parked or spent cont.
                     currentJobsGen = mintNextGen(currentJobsGen);
                 }
+                // Lifecycle flipped to HALTED. signalClose() may have scheduled resumes for
+                // continuations parked on suspending queries (e.g. sleep) so their bodies
+                // observe isShuttingDown() and unwind on a live carrier, releasing checked-out
+                // resources such as connection contexts. The RUNNING loop can exit before
+                // draining them; finish here so they are not stranded with resources unreleased.
+                drainShutdownContinuations();
             }
         } catch (Throwable e) {
             ex = e;
@@ -323,6 +330,38 @@ public class Worker extends Thread {
             // does not survive across engine restarts in long-running JVMs.
             // No-op if bind() was never reached (lifecycle CAS failed).
             CarrierIdentity.unbind();
+        }
+    }
+
+    /**
+     * Drains continuations scheduled for resume during shutdown, run after the worker's
+     * RUNNING loop exits. Mirrors the outer driver's handoff walk: each parked cont is
+     * remounted and run to completion so its body observes {@code isShuttingDown()} and
+     * unwinds, releasing any checked-out resource (e.g. a connection context). A cont
+     * parked long enough to reach here is past the benign mount-race window, so
+     * {@link #mountForeignCont} runs it rather than abandoning it.
+     */
+    private void drainShutdownContinuations() {
+        if (continuationQueue == null) {
+            return;
+        }
+        final ContinuationQueue.ResumeTask scratch = new ContinuationQueue.ResumeTask();
+        WorkerContinuation cont;
+        while ((cont = continuationQueue.tryDequeue(scratch)) != null) {
+            mountForeignCont(cont);
+            while (!cont.isDone()) {
+                WorkerContinuation handoff = cont.takeHandoff();
+                if (handoff == null) {
+                    // Parked deep without a handoff; leave it for the halt-timeout backstop.
+                    break;
+                }
+                recycleJobList(cont);
+                cont = handoff;
+                mountForeignCont(cont);
+            }
+            if (cont.isDone()) {
+                recycleJobList(cont);
+            }
         }
     }
 
--- a/core/src/test/java/io/questdb/test/mp/WorkerContinuationTest.java
+++ b/core/src/test/java/io/questdb/test/mp/WorkerContinuationTest.java
@@ -24,29 +24,38 @@
 
 package io.questdb.test.mp;
 
+import io.questdb.mp.Worker;
 import io.questdb.mp.WorkerPoolConfiguration;
 import io.questdb.mp.continuation.WorkerContinuation;
+import io.questdb.std.Os;
 import org.junit.Assert;
 import org.junit.Test;
 
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
 public class WorkerContinuationTest {
 
     @Test
     public void testHaltDrainsContinuationScheduledJustBeforeHalt() throws Exception {
-        // Regression: a continuation scheduled for resume just before the pool halts must
-        // still be resumed so its body can unwind and release any checked-out resource,
-        // instead of being stranded in the resume queue. This is the worker-level shape of
-        // the sleep()-at-shutdown server socket leak: the parked query continuation was
-        // scheduleResume()'d by engine close but the worker exited without draining it.
-        for (int i = 0; i < 5; i++) {
-            // Single worker that drops into a long backoff sleep after one idle iteration,
-            // so when we schedule the resume it is asleep (past its last empty dequeue) and
-            // exits its RUNNING loop on halt without a further dequeue -- isolating the
-            // shutdown drain from the normal in-loop dequeue.
+        // Regression: a continuation scheduleResume()'d just before the pool halts must still be
+        // resumed -- by the worker's shutdown drain -- so its body can unwind and release any
+        // checked-out resource, instead of being stranded in the resume queue. This is the
+        // worker-level shape of the sleep()-at-shutdown server socket leak: the parked query
+        // continuation was scheduled by engine close but the worker exited without draining it.
+        //
+        // A handshake job makes the worker observably RUNNING before we schedule/halt, so halt()
+        // cannot win the BORN->RUNNING race and skip the whole run() body (and the drain). The
+        // resume is then enqueued while the single worker is parked in its idle backoff sleep,
+        // past its in-loop dequeue, so the RUNNING loop exits via the top-of-loop check without
+        // dequeuing the cont and only drainShutdownContinuations() can pick it up. Each resume
+        // records whether it actually came through the drain (vs the in-loop dequeue), so a
+        // regression in the drain cannot pass as a false green where the in-loop dequeue covers.
+        boolean drainWasExercised = false;
+        for (int attempt = 0; attempt < 8; attempt++) {
+            CountDownLatch workerRunning = new CountDownLatch(1);
             TestWorkerPool pool = new TestWorkerPool(new WorkerPoolConfiguration() {
                 @Override
                 public String getPoolName() {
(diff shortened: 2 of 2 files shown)


RaphDal:
# Code Review: PR #7324 — `chore(core): fix file-descriptor leak when the server shuts down with a query parked`

**Reviewed at level 2** (Agents 1–7 + 11; Agent 8 skipped — no Rust; 9/10 skipped per level).
Fixes #7296 · `Worker.java` (+87/−13), `WorkerContinuationTest.java` (+256) · Labels: Bug, Core, Flaky Test.

8 review agents dispatched (correctness, concurrency, performance, resources, tests, code-quality,
metadata, adversarial-performance); every load-bearing claim verified against source inline. ~10
low/moderate suggestions verified; **1 notable draft finding dropped as a false positive** (the
multi-worker "starvation/livelock", refuted at source) plus many SAFE confirmations.

> Note: this file replaces an earlier `PR-7324-review.md` that reviewed a **pre-`4040387525`**
> revision of the branch. That older review's central finding — `mountForeignCont` abandoning a
> still-mounted cont on the halt branch instead of re-enqueuing it, leaving a residual multi-worker
> leak — was **addressed by the re-enqueue/retry logic that commit `4040387525` added** and that this
> review covers. Its metadata finding (missing `Fixes #7296` / `Flaky Test` label) is likewise now
> resolved.

---

## Verdict: **Approve — no blocking issues.**

The fix is correct and load-bearing. Verified end-to-end:

- **The bug is real.** Old `Worker.run()` exited the RUNNING loop via `return` (on `cont.isDone()`)
  or fell through (on the lifecycle check) and **never drained the `ContinuationQueue`**. Any cont
  `scheduleResume()`'d during the shutdown window after the worker stopped dequeuing was stranded.
  Because the parked PG context was already removed from the dispatcher's `pending` set
  (`IODispatcherLinux.java:98-99`), `AbstractIODispatcher.close()` could not sweep it either — so
  the socket fd leaked for the JVM's lifetime.
- **The fix works.** `return`→`break` routes both loop-exit paths into the new
  `drainShutdownContinuations()`, which remounts each stranded cont. The body resumes, observes
  `isShuttingDown()`, throws `CairoException`, and unwinds into PGWire's `catch(Throwable)`
  (`PGServer.java:151`) → `dispatcher.disconnect()` → fd freed exactly once. The
  `mountForeignCont` halt branch now **re-enqueues** a still-mounted cont (`continuationQueue.put`)
  for the drain to retry, instead of abandoning it — closing the multi-worker residual that an earlier
  revision left open.
- **Halt stays bounded.** `WorkerPool.halt()` → `halt(DEFAULT_HALT_TIMEOUT_NANOS=30s)` →
  `halted.await(remaining(deadline))`; no production path reaches an unbounded wait, so the drain
  cannot make shutdown unkillable even against a genuinely wedged carrier.
- **No lost cont, no double-run** (`ContinuationQueue` is an exactly-once MPMC queue; re-enqueue
  keeps net count at 1), **no double-close/double-recycle** (`takeJobs()` nulls the field;
  cross-worker `snapshotPool`/`ownedJobClones` accounting closes each clone exactly once via its
  minting worker).
- **Hot path untouched.** `loopBody`'s per-tick loop is byte-for-byte unchanged; the boolean refactor
  of `mountForeignCont` adds at most one register-branch to a warm (per-wake, not per-tick) walk.
- **Tests are sound and not flaky** — verified empirically (3/3 pass, test 1 looped 12× clean, the
  `drainShutdownContinuations` stack-walk assertion confirmed non-vacuous on GraalVM CE 25).

---

## Critical
None.

---

## Moderate (non-blocking; out-of-diff, pre-existing)

### M1 — Drain leak-freedom rests on an unstated, unenforced invariant
*Trigger out-of-diff (`CairoEngine.signalClose`/`TimerShards`); relied on by the in-diff drain.*

The correctness and concurrency passes independently converged here. Both the drain's **termination**
and its **no-leak** guarantee depend on:

> every parked cont is registered in a timer shard at the instant `signalClose()` →
> `timerShards.shutdown()` snapshots it, plus the per-waiter `state` CAS that turns a later
> cross-pool `tryFire` into a no-op.

Pools halt sequentially with a shared deadline (`WorkerPoolManager.halt`), and the WAL-apply (write)
pool keeps running while the query pool drains and exits. A write-pool `fireWaiters` →
`scheduleResume` onto the *already-drained, already-exited* query queue would be a lost cont — exactly
the class this PR closes. It is safe **today** only because the two in-tree suspending functions
(`sleep`, `wait_wal_table`) both `register()` in a timer shard immediately before `suspend()`, so
`signalClose()` cancels them before any pool halts; **no concrete leak is constructible**, and this is
**not a regression** (master leaked unconditionally — this is strictly safer).

The risk is latent: a future suspending primitive that parks without a live timer-shard entry (e.g. a
cont registered only on `SeqTxnTracker`) would silently reopen the window.

**Suggested:** add a one-line comment stating the invariant in `drainShutdownContinuations` /
`signalClose`; consider halting the write/apply pool *before* the query pool as defense-in-depth.
Flagging for awareness, not as a merge blocker.

---

## Minor

### m1 — Drain has no per-cont try/catch *(in-diff)*
`drainShutdownContinuations` (Worker.java:353–380) processes conts in a bare `while`. A resumed cont
throwing a non-`IllegalStateException` that escapes `loopBody` — reachable only under non-default
`shared.worker.haltOnError=true` — unwinds out of the drain and **abandons the rest of the queue**
(their fds leak until process exit; `haltLatch` still counts down via the `finally`). Cheap hardening:
wrap the per-cont body in `try/catch` that logs and continues.

### m2 — `Os.pause()` re-enqueue comment misstates the bound *(in-diff)*
The re-enqueue spin (Worker.java:354–360, 591–595) is commented as bounded "nanoseconds." That's the
*typical* case. `Os.pause()` is `Thread.sleep(0)` (Os.java:215) — a scheduler **yield** (the right
primitive: a pure `onSpinWait` could livelock under core-pinning) — and the true worst-case bound is
the 30s halt timeout. With multiple workers draining one shared queue, a re-enqueued cont ping-pongs
between drainers until its registering carrier unmounts. Suggest the comment note (a) `Os.pause()` is a
yield (the mechanism that lets the registering carrier make progress), and (b) the hard bound is the
halt timeout, not nanoseconds.

### m3 — `mountForeignCont` check-order asymmetry *(in-diff)*
Pre-loop checks `consumeParkRefused()`→`isDone()` (574/577); the catch checks
`isDone()`→`consumeParkRefused()` (585/588). Benign today (a park-refused cont is still mounted+running
so cannot be done; and foreign conts carry no job snapshot, so the divergent recycle is a no-op) — but
a latent trap if a cont with an attached snapshot is ever routed here. Make both sites check
`isDone()` first, or add a one-line "order is intentional/benign" note.

### m4 — Duplicated rationale in comments *(in-diff)*
The `drainShutdownContinuations` (336–345) and `mountForeignCont` (558–571) Javadocs restate the
"re-enqueue / one suspend() from parking / nanoseconds" reasoning the inline comments at 355–358 and
592–593 already carry. Per the repo norm ("few and short comments; deep rationale in the PR body"),
collapse one side to a pointer.

### m5 — Tests: assert thread death after the `finally` joins *(in-diff)*
Tests 2/3 do `holder.join(...)`/`halter.join(...)` in `finally` without asserting the thread died. A
future wedge would silently leak a non-daemon thread rather than fail. Add
`assertFalse(holder.isAlive())` / `assertFalse(halter.isAlive())` after the joins.

### m6 — Test 2's holder thread is dead weight; its comment overstates it *(in-diff)*
In `testHaltDropsPhantomResumeInsteadOfRetrying`, the phantom is dropped at `mountForeignCont`'s top
via `consumeParkRefused()` (line 574) **before** any `cont.run()` (line 582) — so the holder's mount is
never contended and does not gate what the test asserts. The test would behave identically with just
`markParkRefused()` + `scheduleResume()`. Tighten the comment (it implies the holder gates the drop) or
drop the holder for a minimal test.

### m7 — Explain the 55.56% coverage rather than leave it bare *(PR body)*
The uncovered new lines are the drain's **inner handoff walk** (Worker.java 363–375) plus ISE arms
(577–578, 586, 589). That inner walk is the structural mirror of the outer driver and is unreached
because no in-tree resumed cont re-dequeues+hands off during the drain (the same reason the
deep-repark path is unreachable — both `sleep` and `wait_wal_table` throw on `isShuttingDown()`).
Acceptable to leave uncovered, but state this in the PR body so the 55% isn't read as untested risk.

### m8 — FD-release symptom is proven only out-of-diff *(PR body)*
The three new unit tests use synthetic conts (no socket) and assert the *drain/resume/phantom*
mechanics — the worker-level proxy. The end-to-end fd-release is covered by the pre-existing
`ServerMainSleepTest.testSleepCancelledByConnectionDrop` (confirmed **not** in this PR's diff).
Defensible (this PR's purpose is to make that test reliable, and the test plan does reference it), but
the body could state the split explicitly: new tests = drain branches in isolation; fd symptom =
existing `ServerMainSleepTest`.

### m9 — Metadata nits *(PR metadata)*
- The `Bug` label sits oddly against the deliberate `chore(` type. #7296 is an internal flaky-test/leak
  (team-authored, "Flaky Test" labeled), so `chore` is the *correct* type per repo convention; `Core`
  + `Flaky Test` already cover it — consider dropping `Bug`.
- Branch commit titles (`fix(worker): ensure continuations are properly drained…`, 95 chars;
  `chore(core): …`, 84 chars) use Conventional-Commits prefixes, exceed 50 chars, and have empty
  bodies. Throwaway under squash, and the squashed PR title that lands is compliant — cosmetic only.

---

- **Multi-worker drain "starvation/livelock" (drafted Medium, in-diff) — REFUTED.** The claim rested
  on "`Os.pause()` does not yield the core, so a spinning peer starves the carrier it waits on."
  `Os.pause()` is `Thread.sleep(0)` (Os.java:215, verified) — it **yields**; and the sleep body calls
  `suspend()` the statement *after* `register()`, so the registering carrier unmounts within
  nanoseconds, not "after the entire unwind" (the unwind happens *after* resume, on the drainer).
  Bounded + yielding + multi-core ⇒ no starvation. The performance and resource passes independently
  reached "bounded and fine." The only valid residual is the comment nit (m2).
- **"Deep re-park strands a cont" — documented + verified unreachable.** Both in-tree suspending
  functions throw `CairoException` on `isShuttingDown()` rather than re-park
  (`SleepFunctionFactory.java:217`, `WaitWalFunction.java:108`). Already disclosed in the PR's
  "Tradeoffs and limits." Not a live bug.
- **SAFE confirmations** (not findings): `ContinuationQueue` exactly-once MPMC delivery; queue
  happens-before via the slot `sequenceNumber` volatile; `parkRefused` CAS single-consumer;
  lifecycle-read consistency between both callers; halt boundedness; no zero-GC/hot-path regression
  (`loopBody` byte-for-byte unchanged). Corroboration of the design's soundness.

---

## Summary

- **Verdict: approve.** No Critical or blocking issues; the leak fix is correct, verified end-to-end,
  and strictly safer than master. The hot path is untouched.
- **Tradeoffs:** shutdown can spin (bounded by the 30s halt timeout, yielding via `Thread.sleep(0)`)
  and is slightly longer when conts are parked — already disclosed in the PR body.
- **Findings:** ~10 low/moderate suggestions verified; **1 notable draft finding dropped as a false
  positive** (multi-worker starvation, refuted at source) alongside many SAFE confirmations. The single
  most useful follow-up is the M1 invariant comment.
- **In-diff vs out-of-diff:** 8 in-diff (mostly Minor doc/test polish), 1 out-of-diff/pre-existing (the
  cross-pool invariant — surfaced by the correctness + concurrency + resource passes, so cross-context
  coverage did not underrun).

### Most useful changes before merge (all optional)
1. **M1** — one-line invariant comment in `drainShutdownContinuations`/`signalClose`.
2. **m1** — per-cont `try/catch` in the drain loop (cheap robustness under `haltOnError=true`).
3. **m2** — fix the `Os.pause()`/"nanoseconds" comment to state it's a yield bounded by the halt timeout.
4. **m3** — make the two `isDone()`/`consumeParkRefused()` checks consistent.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 33 / 46 (71.74%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/mp/Worker.java|22|35|62.86%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/vm/MemoryCMRImpl.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/FastGroupByAllocator.java|1|1|100.00%|

bluestreak01 (approved):
Approved at level 3 (full review). The fix correctly closes the #7296 FD leak; the break+drain, the boolean mountForeignCont re-enqueue/drop protocol, and the multi-worker MPMC drain are internally consistent and terminate under the verified shutdown ordering (signalClose/timerShards.shutdown joins timer threads before worker pools halt). No reachable correctness/concurrency/resource bug found in-tree.

One non-blocking note for a follow-up: the shutdown drain has no internal deadline and relies on the bounding invariant 'carriers unmount promptly or set parkRefused'. The PR body/comments describe the WorkerPool halt timeout as a backstop, but halt() only logs and proceeds on timeout — it does not stop a spinning worker thread, so a never-unmountable cont would leave a hot orphan + potential UAF after the 30s timeout. Unreachable for in-tree suspending functions today; consider a defensive bound or a corrected comment.

PR #7325, opened 2026-06-24
test(core): stop SampleByConfigTest flaking on the post-close RSS check

RaphDal:
## Summary

`SampleByConfigTest` validates the `cairo.sql.sampleby.default.alignment.calendar` wiring through a real `ServerMain`, asserting a `SAMPLE BY 1d` query via `QueryAssertion`. It leaves the post-close memory-usage check enabled. That check bounds how much native memory a cursor may retain after it closes (64 KiB) by reading a process-global counter, so on a live server it can trip on memory held by unrelated concurrent work rather than by the cursor under test. Its nanosecond sibling `SampleByNanoTimestampConfigTest` -- the same assertion against `TIMESTAMP_NS` data -- exhibited the resulting intermittent failure and already carries this opt-out on master:

```
SampleByNanoTimestampConfigTest.testSampleByConfigDefaultAlignmentCommented
AssertionError: cursor is allowed to keep up to 64 KiB of RSS after close. This cursor kept 1593 KiB.
  at QueryAssertion.assertFactoryMemoryUsage(QueryAssertion.java:1416)
```

`SampleByConfigTest` drives the identical live-`ServerMain` `SAMPLE BY` assertion with the same exposure and is the remaining test of the pair still without the opt-out. This PR adds `.noMemoryUsageCheck()` to it. This matches the opt-out already merged for the identical `SAMPLE BY` assertions in `ServerMainForeignTableTest` (#7307, #7323) and the pattern in `ServerMainCleanStartupTest`.

## Root cause

`QueryAssertion.assertFactoryMemoryUsage()` compares `getMemUsedByFactories()` taken before the cursor opens against the same value after it closes, and fails if the difference exceeds 64 KiB. `getMemUsedByFactories()` sums `Unsafe.getMemUsedByTag()` over the "factory" tags -- every `NATIVE_*` tag except an explicit deny-list. That deny-list excludes the heavy background holders (`NATIVE_TABLE_WRITER`, `NATIVE_TABLE_READER`, `NATIVE_TABLE_WAL_WRITER`, `NATIVE_O3`, `NATIVE_JOIN_MAP`, ...), so writer/reader/WAL memory does not enter the sum. What does enter it are query-execution tags such as `NATIVE_GROUP_BY_FUNCTION`, `NATIVE_FAST_MAP`, `NATIVE_FUNC_RSS` and `NATIVE_HTTP_CONN`.

That bound is sound for `AbstractCairoTest`, where the test owns an isolated engine and essentially nothing but the cursor under test moves those tags between the two readings. `SampleByConfigTest` is an `AbstractBootstrapTest` running a real `ServerMain`. On a live server, concurrent query workers and async HTTP connections allocate and free those same tags between the snapshot and the post-close reading, so the delta is not attributable to the cursor -- it is noisy in both directions (a concurrent allocation inflates it; a concurrent free can mask a real retention). The cursor itself is clean: across repeated runs the `SAMPLE BY` factory releases all but a few hundred bytes of native memory on close.

This is unrelated to the per-query memory-tracker work. `Unsafe.malloc` records the global per-tag counter unconditionally and only *additionally* records to a bound per-query tracker, so changing per-query tracker accounting cannot move the totals this check reads. The two counters are independent.

## Why dropping the check here loses no coverage

- On a live `ServerMain` the check cannot attribute memory to the cursor in either direction, so it does not provide reliable leak protection in this test -- only flake.
- `SampleByConfigTest` already sets `.noLeakCheck()`, which disables the stricter total-memory-balance assertion for the same reason (a live server legitimately retains memory). The RSS check is the same class of assertion, narrower and noisier; `.noMemoryUsageCheck()` makes the two consistent.
- Per-cursor retention for the `SAMPLE BY` / GROUP BY allocator path stays covered, reliably, by the isolated-engine suites: `SampleByNanoTimestampTest` runs 77 `SAMPLE BY` queries with this exact 64 KiB check enabled and 95 `assertMemoryLeak` blocks, and `SampleByTest` plus the parallel GROUP BY memory-tracker suites cover the rest. There the global counter is attributable to the cursor, so a real retention regression fails deterministically.

## Changes

- `SampleByConfigTest`: `.noMemoryUsageCheck()` added in `testSampleByConfigDefaultAlignmentOption()`; all five `@Test` methods funnel through this helper. Its nanosecond sibling `SampleByNanoTimestampConfigTest` already received the same one-line opt-out on master, so this is the last of the pair to get it.

The query results, timestamp column, random-access support and reported size are still asserted in all cases via `.returns(...)`; only the RSS measurement is skipped.

## Tradeoffs

- The post-close RSS check no longer runs for this live-server `SAMPLE BY` assertion. That is the intended effect: the check is not meaningful against a live `ServerMain`, because the counter it reads is not isolated to the cursor. The meaningful coverage stays in the `AbstractCairoTest`-based suites.
- The narrower alternatives -- raising the 64 KiB threshold or retrying the reading -- keep a check that structurally cannot attribute memory to the cursor on a shared server. The observed noise reaches the MB range, so any workable threshold would render the check meaningless; these would move the flake rather than remove it.

## Test plan

- `SampleByConfigTest` -- 5 tests pass.


RaphDal:
# Review: PR #7325 — `test(core): stop SampleByConfigTest flaking on the post-close RSS check`

**Level 2.** One-line, test-only change. Verification done inline rather than via the full agent fleet, which would add no signal here.

## Verification performed

All of the PR's factual claims check out against the source:

- `.noMemoryUsageCheck()` → `memoryUsageCheck=false` → `snapshotMemoryUsage()` early-returns leaving `memoryUsage=-1` → `assertFactoryMemoryUsage()` no-ops (`if (memoryUsage > -1)`). Every other assertion (`returns`, `timestamp`, `supportsRandomAccess`, `expectSize`, variable-column, second-pass, calculateSize) still runs. The opt-out disables **only** the RSS check.
- `getMemUsedByFactories()` sums `Unsafe.getMemUsedByTag()` — a process-global per-tag counter, so the cross-talk argument on a live `ServerMain` is real.
- Established pattern verified: `ServerMainForeignTableTest` (3 sites) uses the identical `.noLeakCheck().noMemoryUsageCheck()....returns(...)` idiom on a live-server `SAMPLE BY`.
- Coverage-preservation claim verified: `SampleByNanoTimestampTest extends AbstractCairoTest` (isolated engine) keeps the RSS check enabled (0 opt-outs).
- Net PR surface is genuinely one line (sibling/added-then-removed branch commits net out via the master merge).

## Critical

None.

## Moderate

None.

## Minor

- **No remaining native-memory assertion in `SampleByConfigTest` after this change (informational, not a defect).** With both `.noLeakCheck()` (pre-existing) and `.noMemoryUsageCheck()` (this PR) disabled, the test now asserts only query results, timestamp, random-access support, and size — it performs zero native-memory checking. This is the correct and intended outcome for a live-`ServerMain` config-wiring test, and the GROUP BY/SAMPLE BY retention path stays covered deterministically in the isolated-engine suites (`SampleByNanoTimestampTest` et al.). Worth keeping in mind only so a future reader doesn't mistake this for a memory-coverage test. No change required.

None — no speculative findings were drafted to dismiss. The substantive risk for this kind of change ("does silencing the check hide a real leak?") was checked directly: the silenced check is non-attributable on a shared live server (global counter, already noisy enough that `.noLeakCheck()` was disabled for the same reason), and the attributable version of the same check runs with the engine isolated in `SampleByNanoTimestampTest`.

## Summary

- **Verdict: approve.** Correct, minimal, well-scoped flaky-test fix. It matches the already-merged #7307/#7323 opt-out in `ServerMainForeignTableTest` and brings `SampleByConfigTest` in line with its already-fixed sibling `SampleByNanoTimestampConfigTest`.
- **Builder-API checklist: clean.** Keeps `.returns(...)` (full battery, not `.returnsOnce(...)`); no `assertSql(...)`; not the `assertMemoryLeak`-wrapping-a-single-`assertQuery` anti-pattern.
- **PR metadata: clean.** Conventional Commits title (`test(core):` is the right type for a test-only change), accurate and level-headed body with an explicit Tradeoffs section, correct `SQL` + `Flaky Test` labels. No `Fixes #NNN` needed (not an issue fix).
- **No regressions or tradeoffs beyond the one the PR already documents** (the non-meaningful live-server RSS check stops running here; meaningful coverage is unaffected).
- 6 PR claims verified, 0 dropped as false positives. 1 minor (informational) finding, in-diff; 0 out-of-diff findings — expected for a one-line diff whose single changed call was traced through `QueryAssertion` (the out-of-diff file) and its sibling callsites.

> Note left out of the review per scope: the `questdb-butler` "Enterprise CI Failed" comment targets a stale commit (`fdd69da5ea`, before the branch's `master` merge) and cannot be caused by adding `.noMemoryUsageCheck()` to an OSS test — it's not a blocker for this PR.

bluestreak01 (approved):

PR #7327, opened 2026-06-25
fix(http): prevent HTTP context reuse crashes and tighten multipart upload parsing

jerrinot:
HTTP connections and contexts are reused across requests. Several HTTP parser/response objects were not fully resetting or reopening their per-request state, which could make a later request inherit stale state from an earlier one. In the worst case, a reused context could hit closed native parsing/compression buffers and crash the server when handling otherwise valid requests.

This PR fixes the reusable HTTP lifecycle state and hardens multipart upload parsing around boundary and `Content-Disposition` edge cases.

**User Impact**
- Prevents server crashes during HTTP context reuse when later requests parse quoted header parameters such as `charset="utf-8"`.
- Prevents gzip response state from leaking across keep-alive requests, so clients only receive gzip responses when the current request advertises gzip support.
- Prevents stale parsed charset and cookie metadata from being observed on reused header parsers.
- Prevents multipart upload bodies from losing bytes when body data resembles a multipart boundary but is not one.
- Allows uploaded filenames such as `a;b.csv` and `a=b;c.csv` to parse correctly.

**What Changed**
- Reopen the header parser’s quoted-value sink when pooled HTTP contexts are reused.
- Clear per-request gzip negotiation state in `HttpResponseSink`.
- Clear `charset` and mapped cookie state in `HttpHeaderParser.clear()`.
- Fix multipart false-boundary replay so body bytes are preserved when a boundary-like sequence is followed by non-boundary suffixes.
- Keep multipart resume pointers anchored to the receive buffer during retry paths.
- Parse `Content-Disposition` parameters with quote-aware delimiter handling, including semicolons and equals signs inside quoted values.
- Reset `Content-Disposition` parameter state consistently after known and unknown parameters.

### DIFF
--- a/core/src/main/java/io/questdb/cutlass/http/HttpHeaderParser.java
+++ b/core/src/main/java/io/questdb/cutlass/http/HttpHeaderParser.java
@@ -120,6 +120,7 @@ public class HttpHeaderParser implements Mutable, QuietCloseable, HttpRequestHea
         this.query = null;
         this.headerName = null;
         this.contentType = null;
+        this.charset = null;
         this.boundary = null;
         this.contentDisposition = null;
         this.contentDispositionName = null;
@@ -139,7 +140,9 @@ public class HttpHeaderParser implements Mutable, QuietCloseable, HttpRequestHea
         this.contentLength = -1;
         this.cookieList.clear();
         this.cookiePool.clear();
+        this.cookies.clear();
         this.ignoredCookieCount = 0;
+        this.sink.clear();
         // do not clear the pool
         // this.pool.clear();
     }
@@ -347,6 +350,7 @@ public class HttpHeaderParser implements Mutable, QuietCloseable, HttpRequestHea
             this.hi = headerPtr + bufferSize;
         }
         boundaryAugmenter.reopen();
+        sink.reopen();
     }
 
     public int size() {
--- a/core/src/main/java/io/questdb/cutlass/http/HttpHeaderParser.java
+++ b/core/src/main/java/io/questdb/cutlass/http/HttpHeaderParser.java
@@ -629,6 +629,9 @@ public class HttpHeaderParser implements Mutable, QuietCloseable, HttpRequestHea
 
         boolean expectFormData = true;
         boolean swallowSpace = true;
+        boolean valueQuoted = false;
+        boolean valueEscaped = false;
+        boolean valueStart = false;
 
         DirectUtf8String name = null;
 
@@ -640,7 +643,31 @@ public class HttpHeaderParser implements Mutable, QuietCloseable, HttpRequestHea
                 continue;
             }
 
-            if (p > hi || b == ';') {
+            if (name != null && valueStart) {
+                valueStart = false;
+                if (b == '"') {
+                    valueQuoted = true;
+                    valueEscaped = false;
+                    continue;
+                }
+            }
+
+            if (valueQuoted) {
+                if (valueEscaped) {
+                    valueEscaped = false;
+                    continue;
+                }
+                if (b == '\\') {
+                    valueEscaped = true;
+                    continue;
+                }
+                if (b == '"') {
+                    valueQuoted = false;
+                    continue;
+                }
+            }
+
+            if (p > hi || (b == ';' && !valueQuoted)) {
                 if (expectFormData) {
                     this.contentDisposition = csPool.next().of(_lo, p - 1);
                     _lo = p;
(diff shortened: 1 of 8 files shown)


jerrinot:
# Review: PR #7327 — `fix(http): prevent HTTP context reuse crashes and tighten multipart upload parsing` (level 3)

Full 11-agent pass (Agent 8 Rust skipped — no `.rs` files). **6 review agents dispatched, all reported; ~6 draft findings verified, 1 false positive dropped. All findings are in-diff; the cross-context pass ran wide (Agent 9, 49 tool calls) and found 0 broken callsites — out-of-diff is genuinely zero because every change is a net fix to a previously-broken contract.**

## Critical
None.

## Moderate
None. Every production change was verified correct and each fixes a real pre-existing defect:
- **`HttpHeaderParser.reopen()` → `sink.reopen()`**: fixes a SIGSEGV — a pooled context parsing a quoted `charset="utf-8"`/`filename="..."` wrote into a freed `DirectUtf8Sink` (`impl==0`). Lifecycle verified leak-free/double-free-free.
- **`clear()` → `charset=null` / `cookies.clear()` / `sink.clear()`**: fixes stale charset and a `getCookie()` use-after-recycle (mapped `HttpCookie` slots were recycled while the map still referenced them) across keep-alive reuse.
- **`HttpResponseSink.clear()` → `deflateBeforeSend=false`**: fixes gzip leaking into a following non-gzip keep-alive request. `reset()` runs `clear()` per request and negotiation runs later, so it can't suppress same-request gzip.
- **Multipart false-boundary replay (new `*_AFTER_CR/DASH` states)**: byte-exact reconstruction of `<boundary>+<CR|dash>` from the stable boundary buffer; preserves bytes the old code dropped. Retry semantics correct (re-delivery via `onRequestRetry` + state resume; no double/lost delivery).
- **`POTENTIAL_BOUNDARY` `handleIncomplete` true→false**: fixes heap corruption — old path set `resumePtr=boundary.lo()` then `memmove(recvBuffer, boundary.lo(), buf-boundary.lo())` with a garbage length. Only `BOUNDARY_NO_MATCH` reaches it, so no legitimate "wait for more bytes" case is lost.

## Minor
1. **`HttpHeaderParser.java:656-658` (in-diff)** — a dangling trailing backslash in a quoted value (`...; name="a\`) hits the `valueEscaped` `continue`, which skips the `p > hi` terminator; the loop then exits leaving `contentDispositionName`/`Filename` `null`. Master threw `HttpException("unclosed quote")`. Both still reject the request (`onPartBegin` → "invalid value..." disconnect), so it's a less-precise error path on malformed input only — not a crash or data loss. Optional: treat a dangling escape as end-of-value.
2. **Missing tests (Low)** — `name=""` (empty quoted value) and whitespace-only value are unasserted; both behave correctly (empty/literal-space). Worth adding given the `unquote()` len==0 boundary sits right next to it.
3. **`HttpHeaderParserTest.java:393,424`** — `testContentTypeReuseClearsCharset` / `testCookiesReuseClearsMappedCookies` allocate native memory but aren't wrapped in `assertMemoryLeak()`. Freed in `finally`, and matches the file's existing convention — consistency nit only.
4. **`HttpHeaderParser.parseContentDisposition` booleans** `valueQuoted`/`valueEscaped`/`valueStart` lack the `is`/`has` prefix the guidelines mandate; they match the pre-existing locals (`expectFormData`, `swallowSpace`), so low-priority. The `name != null` guard at line 646 is redundant (`valueStart` is only set together with `name`).
5. **PR description** omits the one behavioral tradeoff: a false-boundary replay that throws `NotEnoughLines` now fails the import instead of (incorrectly) waiting. Per the level-headed-tone convention, one sentence noting it would round out the description.

- *"Backslash-escape branch (L656-662) is dead/uncovered"* (Agent 5 initial) — disproved by a live probe; `testContentDispositionQuotedFilenameWithSemicolonAndEscapedQuote` exercises it. Stale `jacoco.exec` from `surefire:test` without recompile.
- *"HttpMultipartContentParser only 86% covered"* (mtopolnik bot) — stale; measured before the suffix-state tests landed (`c165454548`, `2b046be59b`). Current HEAD is 100% line coverage.

## Summary
**Verdict: approve.** This is a clean set of HTTP lifecycle/parser fixes; correctness, resource, performance, concurrency, and cross-context passes are all clean, and the strict SQL-assertion checklist passes (no new `assertSql`/`returnsOnce`; the `returnsOnce("count\n24\n")`→`.expectSize().noRandomAccess().returns(...)` upgrade is correct). Only minor, non-blocking items remain. The two-step replay adds at most one extra single-byte `onChunk` per (astronomically rare) false boundary — bounded, not an amplification vector.

jerrinot:
a flaky(?) tests, exploring

bluestreak01 (approved):
Approving after a level-3 mission-critical review (full 5-agent fanout + per-finding source verification).

All five production hunks fix real pre-existing defects, each verified against source and callsites:
- reopen() -> sink.reopen(): fixes a native fault when a reused pooled context parses a quoted charset/filename into a freed DirectByteSink.
- clear() -> charset=null / cookies.clear() / sink.clear(): fixes stale-charset reads and a getCookie() use-after-recycle across keep-alive reuse.
- HttpResponseSink.clear() -> deflateBeforeSend=false: stops gzip leaking into a following non-gzip keep-alive response (clear() runs before the new request's setDeflateBeforeSend, so same-request gzip is preserved).
- Multipart false-boundary replay (*_AFTER_CR/DASH, firstDashRead removed): byte-exact reconstruction of dropped CR/dash bytes from the stable boundary buffer; retry re-delivery consistent with the existing onChunkWithRetryHandle contract.
- POTENTIAL_BOUNDARY handleIncomplete true->false: fixes heap corruption (resumePtr would otherwise point into the boundary buffer and memmove a garbage length). The rethrown NotEnoughLines lands in failProcessor() as a graceful disconnect.

No criticals, no blockers. Cross-context callsite walk found 0 broken callsites. Only minor, non-blocking items: a dangling-backslash quoted value leaves name/filename null (malformed input, still rejected); a few untested quoted-value edge cases (name="", whitespace-only, dangling backslash); and the PR description could note the one tradeoff (a false-boundary replay that throws NotEnoughLines now fails the import instead of waiting).

SQL-assertion checklist passes: no new assertSql/returnsOnce; the returnsOnce->.expectSize().noRandomAccess().returns(...) upgrade is correct for a deterministic count() query.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 72 / 72 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/http/HttpResponseSink.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpHeaderParser.java|35|35|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/HttpMultipartContentParser.java|36|36|100.00%|

PR #7329, opened 2026-06-25
fix(core): fix query cancellation being lost when it races query startup

jovfer:
Fixes a live production bug that affects any node serving the PostgreSQL wire protocol: a query cancellation can be silently lost, and the affected query's per-query timeout is defeated along with it, when the cancel races query startup. There is no GitHub issue; fixing directly.

## Root cause

`NetworkSqlExecutionCircuitBreaker.cancel()` always sets the `powerUpTime == Long.MIN_VALUE` sentinel but only flips the per-query cancelled flag when that flag is already attached. `QueryRegistry.register()` publishes the query and fires its listener *before* it binds the per-query flag. A `CancelRequest` landing in that window sets only the sentinel; `register()` then binds a fresh flag whose value is `false`, so the cancel survives only in the sentinel.

The stateful poll path used by essentially all query execution (`statefulThrowExceptionIfTripped*`) runs `testTimeout()` then `testCancelled()`, and neither consulted the sentinel:

- `testTimeout()` computes `now - powerUpTime`; `now - MIN_VALUE` overflows to a large negative runtime, so it never trips.
- `testCancelled()` only checked the flag, which is now `false`.

Net effect: the cancel is dropped, and because `powerUpTime` stays at `MIN_VALUE` the per-query timeout is defeated too. The query runs to natural completion ignoring both cancel and timeout, pinning a worker; async/parallel queries can pin several.

## Fix

- `testCancelled()` checks `isCancelled()` (the sentinel) first, so both stateful paths abort a racing cancel as `queryCancelled()`.
- `getState()` classifies the sentinel as `STATE_CANCELLED` instead of mislabelling it `STATE_TIMEOUT`.
- `PGConnectionContext.prepareForNewQuery()` clears the sentinel (via the new `clearCancelSentinel()`) before every query on a connection, bounding it to a single query.

The sentinel is unforgeable: only `cancel()` writes `MIN_VALUE`. Within a query it is honoured by `testCancelled()` / `getState()`; across queries it cannot leak because `prepareForNewQuery()` clears it before the next query consults the breaker. The *guarded* resets `if (!isTimerSet()) resetTimer()` do not clear it (`isTimerSet()` treats `MIN_VALUE` as set), which is why the explicit per-query clear exists; that clear is a no-op for a real running timer, so it leaves the `query_timeout` window intact.

## Scope and tradeoffs

The change is deliberately minimal: the `testCancelled()` sentinel check and the `getState()` label on the breaker, plus a new `clearCancelSentinel()` called once from `PGConnectionContext.prepareForNewQuery()` to bound the sentinel to a single query. It does not reorder `QueryRegistry.register()` (an alternative the design discussion floated; the sentinel fix subsumes the window without touching the hot registration path or the listener's registration-only contract), does not touch `cancel()` / lifecycle / error text, and does not alter throttling. The optional `testTimeout()` overflow guard is omitted because the `testCancelled()` fix makes it redundant; the pre-existing (now unreachable) `isCancelled()` branch inside `testTimeout()` is left as-is rather than broadening the diff. The `ServerMainSleepTest` cancel-fuzz oracle is left unchanged.

## Test plan

- New deterministic regressions in `DefaultPGCircuitBreakerRegistryTest` drive the race ordering directly on the breaker (no fuzz/timing): a sentinel-only cancel with no flag attached, a cancel followed by a fresh `false` flag (the exact production race), and the `getState()` label. On master all three fail for the expected reasons (nothing thrown / `STATE_TIMEOUT`). Two further tests pin the cross-query no-leak property: a guarded `if (!isTimerSet()) resetTimer()` does not clear the sentinel, `clearCancelSentinel()` does, and a real running timer is left untouched. The class is 13/13 green.
- `PGJobContextTest#testRunQueryAfterCancellingPreviousInTheSameConnection` and `#testCancelQueryThatReusesCircuitBreakerFromPreviousConnection` (same-connection and cross-connection breaker reuse) stay green with the per-query clear.
- `ServerMainSleepTest#testFuzzConcurrentSleeps`: `cancelAttempts=60, cancelObserved=60, cancelNormal=0, cancelNormalBeforeDeadline=0, failures=0` — every cancel observed, zero missed, zero runaways.
- `ServerMainSleepTest` full class: 7/7 green; the timeout and connection-drop paths are unaffected.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/sql/NetworkSqlExecutionCircuitBreaker.java
+++ b/core/src/main/java/io/questdb/cairo/sql/NetworkSqlExecutionCircuitBreaker.java
@@ -147,6 +147,12 @@ public class NetworkSqlExecutionCircuitBreaker implements SqlExecutionCircuitBre
 
     @Override
     public int getState(long millis, long fd) {
+        // A cancelled breaker carries the powerUpTime == MIN_VALUE sentinel, which the overflow-safe
+        // timeout predicate below would otherwise report as STATE_TIMEOUT. Classify it as cancelled so
+        // callers reading getState() see an honest reason.
+        if (isCancelled()) {
+            return STATE_CANCELLED;
+        }
         if (clock.getTicks() - timeout > millis) {
             return STATE_TIMEOUT;
         }
@@ -267,7 +273,12 @@ public class NetworkSqlExecutionCircuitBreaker implements SqlExecutionCircuitBre
     }
 
     private void testCancelled() {
-        if ((cancelledFlag != null && cancelledFlag.get()) || engine.isClosing()) {
+        // isCancelled() (the powerUpTime == MIN_VALUE sentinel) must come first: cancel() sets the
+        // sentinel unconditionally but only flips cancelledFlag when it is already attached. A cancel
+        // that races QueryRegistry.register() before it binds the per-query flag leaves only the
+        // sentinel, and testTimeout()'s now - MIN_VALUE arithmetic overflows without tripping, so this
+        // is the single place that reliably turns such a cancel into a thrown queryCancelled().
+        if (isCancelled() || (cancelledFlag != null && cancelledFlag.get()) || engine.isClosing()) {
             throw CairoException.queryCancelled(fd);
         }
     }
--- a/core/src/main/java/io/questdb/cairo/sql/NetworkSqlExecutionCircuitBreaker.java
+++ b/core/src/main/java/io/questdb/cairo/sql/NetworkSqlExecutionCircuitBreaker.java
@@ -112,13 +112,10 @@ public class NetworkSqlExecutionCircuitBreaker implements SqlExecutionCircuitBre
     }
 
     public void clearCancelSentinel() {
-        // A cancel that raced QueryRegistry.register() before the per-query flag was attached leaves
-        // only the powerUpTime == MIN_VALUE sentinel. A guarded "if (!isTimerSet()) resetTimer()" cannot
-        // clear it (isTimerSet() is true for MIN_VALUE), so the connection clears it explicitly at the
-        // start of every query. Clearing only the sentinel, never a real running timer, keeps a
-        // query_timeout window intact for a named portal paginated across Sync.
-        if (powerUpTime == Long.MIN_VALUE) {
-            powerUpTime = Long.MAX_VALUE;
+        // Drop a cancel left by a prior, finished query so it cannot trip the next one on this reused
+        // breaker. Guarded per-query resets cannot clear it (isTimerSet() reports MIN_VALUE as "set").
+        if (isCancelled()) {
+            unsetTimer();
         }
     }
 
--- a/core/src/main/java/io/questdb/cairo/sql/NetworkSqlExecutionCircuitBreaker.java
+++ b/core/src/main/java/io/questdb/cairo/sql/NetworkSqlExecutionCircuitBreaker.java
@@ -111,6 +111,17 @@ public class NetworkSqlExecutionCircuitBreaker implements SqlExecutionCircuitBre
         timeout = defaultMaxTime;
     }
 
+    public void clearCancelSentinel() {
+        // A cancel that raced QueryRegistry.register() before the per-query flag was attached leaves
+        // only the powerUpTime == MIN_VALUE sentinel. A guarded "if (!isTimerSet()) resetTimer()" cannot
+        // clear it (isTimerSet() is true for MIN_VALUE), so the connection clears it explicitly at the
+        // start of every query. Clearing only the sentinel, never a real running timer, keeps a
+        // query_timeout window intact for a named portal paginated across Sync.
+        if (powerUpTime == Long.MIN_VALUE) {
+            powerUpTime = Long.MAX_VALUE;
+        }
+    }
+
     @Override
     public void close() {
         buffer = Unsafe.free(buffer, bufferSize, memoryTag);
--- a/core/src/main/java/io/questdb/cutlass/pgwire/PGConnectionContext.java
+++ b/core/src/main/java/io/questdb/cutlass/pgwire/PGConnectionContext.java
@@ -1281,6 +1281,11 @@ public class PGConnectionContext extends IOContext<PGConnectionContext> implemen
 
     private void prepareForNewQuery() {
         LOG.debug().$("prepare for new query").$();
+        // Bound the cancel sentinel to a single query: a cancel for a prior, already-finished query
+        // can leave powerUpTime == MIN_VALUE on this reused per-connection breaker, and the guarded
+        // per-query resets do not clear it. prepareForNewQuery() runs once before each query (every
+        // Sync, both protocols), at which point any sentinel belongs to a finished query.
+        circuitBreaker.clearCancelSentinel();
         Misc.clear(bindVariableService);
         freezeRecvBuffer = false;
         sqlExecutionContext.setCacheHit(false);
--- a/core/src/test/java/io/questdb/test/cutlass/pgwire/DefaultPGCircuitBreakerRegistryTest.java
+++ b/core/src/test/java/io/questdb/test/cutlass/pgwire/DefaultPGCircuitBreakerRegistryTest.java
@@ -170,6 +170,48 @@ public class DefaultPGCircuitBreakerRegistryTest extends AbstractCairoTest {
     }
 
     @Test
+    public void testClearCancelSentinelClearsStaleCancel() throws Exception {
+        // A guarded "if (!isTimerSet()) resetTimer()" cannot clear the sentinel because isTimerSet()
+        // is true for MIN_VALUE. clearCancelSentinel() is the unconditional per-query clear that bounds
+        // the sentinel to a single query.
+        assertMemoryLeak(() -> {
+            try (NetworkSqlExecutionCircuitBreaker cb = newCircuitBreaker()) {
+                cb.resetTimer();
+                cb.cancel();
+                // A guarded reset does NOT clear the sentinel (the breaker still reports the cancel):
+                if (!cb.isTimerSet()) {
+                    cb.resetTimer();
+                }
+                expectQueryCancelled(cb::statefulThrowExceptionIfTripped);
+
+                // The connection's per-query entry clears the stale sentinel:
+                cb.clearCancelSentinel();
+                // Now a guarded reset arms a fresh timer, and the breaker is OK:
+                if (!cb.isTimerSet()) {
+                    cb.resetTimer();
+                }
+                cb.statefulThrowExceptionIfTripped(); // must not throw
+                Assert.assertEquals(SqlExecutionCircuitBreaker.STATE_OK, cb.getState());
+            }
+        });
+    }
+
+    @Test
+    public void testClearCancelSentinelKeepsRunningTimer() throws Exception {
+        // clearCancelSentinel() must not disturb a legitimately running timer (e.g. a named portal
+        // paginated across Sync); it clears only the cancel sentinel.
+        assertMemoryLeak(() -> {
+            try (NetworkSqlExecutionCircuitBreaker cb = newCircuitBreaker()) {
+                cb.resetTimer();
+                cb.clearCancelSentinel(); // no-op: powerUpTime is a real timer, not the sentinel
+                Assert.assertTrue(cb.isTimerSet());
+                cb.statefulThrowExceptionIfTripped(); // must not throw
+                Assert.assertEquals(SqlExecutionCircuitBreaker.STATE_OK, cb.getState());
+            }
+        });
+    }
+
+    @Test
     public void testSentinelCancelReportedByGetState() throws Exception {
         // A cancel that lands before QueryRegistry binds the per-query flag leaves only the
         // powerUpTime == MIN_VALUE sentinel. getState() must report it as cancelled instead of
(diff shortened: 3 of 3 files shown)


jovfer:
## Code review (level 3 — full multi-agent pass + per-finding source verification)

**Verdict: approve.** The fix is correct, minimal, and delivers cancellation for non-parallel `SELECT`, parallel-filter `SELECT`, and async `GROUP BY` over PG wire. `UPDATE` is unaffected (runs on the simple flag-only breaker — no regression, pre-existing scope). No correctness, concurrency, resource, or performance regressions found; the hot-path cost is one extra L1-hot volatile load + a never-taken branch, zero allocation. 12 draft findings verified, 2 dropped as false positives.

### How the fix actually works (verified mechanism)

The functional fix is **entirely in `testCancelled()`**. The `getState()` change is inert in production (see m1). The per-query breaker is consulted on the **main thread** right after `QueryRegistry.register()` binds the flag:

- non-parallel scan: `PageFrameRecordCursorImpl.hasNext()` per page frame;
- **parallel filter**: `AsyncFilteredRecordCursorFactory.getCursor():144` at cursor open (after `QueryProgress.register():289`, before dispatch);
- async `GROUP BY`: `AsyncGroupByRecordCursor.buildMap():198`.

All three reach `testCancelled()` and now throw `queryCancelled` on the sentinel.

### Moderate

**M1 — The cross-query safety rationale is imprecise; the no-leak invariant is fragile and undocumented.**
The PR body says *"resetTimer() overwrites it at every query start, so the change ... cannot leak across queries."* That mechanism is not literally true: `isTimerSet()` is `powerUpTime < MAX_VALUE`, and `MIN_VALUE < MAX_VALUE`, so a sentinel reads as "timer set" and every **guarded** reset `if (!isTimerSet()) resetTimer()` *skips* clearing a stale sentinel — `SqlCompilerImpl.java:3110/3445/4075` and `PGPipelineEntry.outCursor():2620`.

There is **no reachable false-positive today** (verified): fresh `SELECT`/`EXPLAIN` reset *unconditionally* before `getCursor()` (`PGPipelineEntry.java:1004/1661`, `cursor == null`), wiping any stale sentinel; row-looping writes (`UPDATE`/`INSERT AS SELECT`/`CTAS`) use the flag-only `AtomicBooleanCircuitBreaker`; DDL/compilation never consult the network breaker; cross-connection reuse is cleared by `PGConnectionContext.clear()` + the per-connection secret. But the conclusion rests on a coincidence, not the stated reason. For a mission-critical breaker, recommend either hardening (treat `MIN_VALUE` as "not set" so guarded resets re-arm — e.g. `isTimerSet()` returns `powerUpTime != MIN_VALUE && powerUpTime < MAX_VALUE`, or clear the sentinel unconditionally at statement entry) or at minimum correcting the rationale to state the real invariant.

**M2 — Missing the regression test for the PR's central safety claim.** No test asserts that a sentinel does not leak across queries (i.e. that `resetTimer()` clears it). Adding one (pushed to this branch).

### Minor

**m1 — The `getState()` half of the diff is inert in production.** The only production callers of `NetworkSqlExecutionCircuitBreaker.getState()` are the async reduce jobs (`PageFrameReduceJob.java:268`, `UnorderedPageFrameReduceJob.java:147`) on worker-local wrapper breakers whose `powerUpTime` is freshly reset by `SqlExecutionCircuitBreakerWrapper.init()` (so `isCancelled()` is always false there); `UpdateOperation` uses the simple breaker. The new `STATE_CANCELLED` branch only fires in the unit test. Fine to keep for defensiveness/contract honesty, but the async benefit the PR text implies comes from `testCancelled()` on the main thread, not from `getState()`.

**m2 — `checkIfTripped()` not updated for consistency.** `getState()`/`testCancelled()` now check `isCancelled()` explicitly, but `checkIfTripped(millis, fd)` still detects a sentinel only via the `clock.getTicks() - timeout > millis` overflow when `millis == MIN_VALUE`. Works for every realistic timeout (and `testCancelHappyPath` relies on it), but brittle — a future overflow guard on that predicate would silently stop it seeing cancels. Consider adding `isCancelled()` for symmetry.

**m3 — No parallel/async or deterministic end-to-end PG-wire race test.** The deterministic breaker-unit tests are a reasonable substitute, but nothing drives the sentinel-only ordering through the PG layer. `ServerMainSleepTest#testSleepCancelledByStatementCancel` cancels after a 300ms sleep (flag already attached — passes on master), and `testFuzzConcurrentSleeps` explicitly tolerates a missed cancel and is single-threaded.

**m4 — Pre-existing dead branch left in place.** `testTimeout(long):293-294`'s `if (isCancelled())` is unreachable (`now - MIN_VALUE` overflows). Pre-existing, deliberately left to keep the diff narrow — acceptable; a follow-up could delete it now that the new comment documents the overflow.

**m5 — Test placement nit.** The sentinel tests live in `DefaultPGCircuitBreakerRegistryTest` but never touch the registry; a `NetworkSqlExecutionCircuitBreakerTest` would fit better. Not worth churn.

Conventions, member ordering, comments (ASCII/active-voice/accurate race description), naming, labels, title, and commit message all check out. No banned query-result `assertSql(...)`; `assertMemoryLeak()` + try-with-resources are correct (`expectQueryCancelled` catches inside the try, so `close()` runs on the throw path — no native leak). All three tests genuinely fail on master and pass with the fix (arithmetic verified).

### Downgraded (false positives, verified and dropped)

- **"Parallel async FILTER loses the racing cancel."** Dropped. The streaming `hasNext()` loop never checks the per-query breaker, but `AsyncFilteredRecordCursorFactory.getCursor():144` calls `statefulThrowExceptionIfTrippedTimeThrottled()` at open, *after* `register()` binds the flag — so the sentinel is caught before dispatch. The sentinel-only state exists only pre-register; post-open the flag is bound and the worker/flag path handles it.
- **"`getState()` reclassification reaches `UpdateOperation`."** Corrected. `UpdateOperation.withContext():180` binds `getSimpleCircuitBreaker()` (an `AtomicBooleanCircuitBreaker`, which never returns `STATE_TIMEOUT` and has no sentinel), so the reclassification does not reach it.

*Reviewed with the multi-agent `/review-pr` harness; findings independently source-verified.*

jovfer:
Code review — PR #7329 fix(core): fix query cancellation being lost when it races query startup

Reviewing at level 2 (agents 1–7 + 11; agent 8 skipped — no .rs files). Level 3 exists for the full 11-agent + per-finding pass if you want it for this breaker, but the change is small and concentrated, so L2 is appropriate.

Verdict: approve with minor changes. No correctness, concurrency, resource, or performance regression found. The sentinel-first testCancelled() fix is sound, the cross-query no-leak property holds, and the getState() change is inert-but-honest in production. The findings below are one moderate test-coverage gap, a few minor nits.

Critical

None.

Moderate

1. The production wiring at PGConnectionContext.java:1288 has no regression test — removing it keeps the suite green. [in-diff]
Every new unit test calls cb.clearCancelSentinel() directly on the breaker, so they pin the method's contract but not the wiring into prepareForNewQuery(). The only E2E that runs a query after a cancel — PGJobContextTest.testRunQueryAfterCancellingPreviousInTheSameConnection:9743 — uses a fresh SELECT count(*) follow-up, which clears any stale sentinel via the unconditional resetTimer() at PGPipelineEntry:1004/1661 whether or not line 1288 exists (verified). So the second hunk of the fix is effectively untested.

Context that sets the severity: the wiring is defense-in-depth. Agent 1 and the prior level-3 review both established there is no reachable false-positive today even without it — fresh queries reset unconditionally, row-writing statements use the sentinel-less AtomicBooleanCircuitBreaker, and compilation/DDL never consult the network breaker. The one path it would defend (a named portal resuming through the guarded reset at PGPipelineEntry.outCursor():2620 after a sentinel-only cancel) isn't proven reachable as a true positive. So this is correct hardening of the invariant the level-3 M1 flagged as fragile — it just lacks a test that would fail if it regressed.
Recommendation: add a regression test that drives a sentinel-leaving cancel followed by a query through the real prepareForNewQuery() path (or a focused PGConnectionContext-level test), so the wiring can't be silently deleted. If you'd rather not, a one-line code/PR note that the clear is belt-and-suspenders would at least record intent.

Minor

2. testClearCancelSentinelKeepsRunningTimer doesn't prove the timeout still fires. [in-diff] DefaultPGCircuitBreakerRegistryTest.java:200. It asserts isTimerSet() and STATE_OK after clearCancelSentinel() on a running timer, but never advances the clock, so it doesn't demonstrate the PR's "preserves the query_timeout window" claim behaviorally. Stronger: setTimeout(small) → resetTimer() → advance clock past timeout → clearCancelSentinel() → expect queryTimedOut.

3. Misleading "must come first" comment in testCancelled(). [in-diff] NetworkSqlExecutionCircuitBreaker.java:284. The condition is an || of three side-effect-free reads that all throw the identical queryCancelled(fd), so operand order is irrelevant to correctness — isCancelled() must be checked, not first (first-position is only a short-circuit micro-opt). Contrast getState():158 where ordering genuinely matters because the branches return different STATE_* values, and whose comment is accurate. Reword to "must check isCancelled() (the sentinel), not only the flag:".

4. Two trivial PR-body inaccuracies. (a) The test-plan narrative describes 5 of the 6 new tests — testSentinelClearedByResetTimer isn't mentioned; reconcile the count. (b) "the pre-existing (now unreachable) isCancelled() branch inside testTimeout()" — that branch was already unreachable before this PR (the now - MIN_VALUE overflow), so "now" slightly misattributes it to the fix.

5. (informational, no action) The getState() hunk has no live production consumer. Its new STATE_CANCELLED-from-sentinel branch never fires in production: the 2-arg callers (PageFrameReduceJob:268, UnorderedPageFrameReduceJob:147) run on a wrapper whose init()→resetTimer() keeps powerUpTime positive, and UpdateOperation:124 uses the sentinel-less simple breaker. Keep it for contract honesty, but the actual fix is entirely in testCancelled(). (Matches the existing level-3 review's m1.) The optional single-volatile-read fusion in the throttled path (agents 3/11) is dominated by the existing currentTimeMillis()/recv and not worth the readability cost — leave it.

Downgraded (investigated, dropped as non-issues)

- "The internal parallel-query shared breakers that call cancel() (GroupBy/LatestBy/AsyncGroupBy/merge-shard/topK/Copy) now spuriously honor a leaked sentinel." Dropped — all are AtomicBooleanCircuitBreaker, which has no powerUpTime sentinel (cancel() sets only the flag). The testCancelled()/getState()/isCancelled() changes are confined to NetworkSqlExecutionCircuitBreaker. This was the largest blast-radius lead and it is fully closed.
- "getState() reclassification reaches UpdateOperation." Dropped — UpdateOperation binds getSimpleCircuitBreaker() (AtomicBooleanCircuitBreaker), whose getState() returns only STATE_CANCELLED/STATE_OK.
- "clearCancelSentinel() can drop a live cancel." Dropped — it runs in prepareForNewQuery() before register()/resetTimer() of the next query, so it only clears prior-query sentinels; live cancels arrive after registration and are carried by the flag. The read-then-write is non-atomic vs a concurrent cancel(), but every interleaving is benign (the next query's reset reconciles it).
- "HTTP path is missing clearCancelSentinel()." Dropped — no caller invokes cb.cancel() on an HTTP NetworkSqlExecutionCircuitBreaker, so its sentinel is never set.

Summary

- One-line verdict: approve — minor changes only; nothing blocking.
- Verification: ~9 draft findings; 5 confirmed (1 moderate, 4 minor/informational), 4 cross-context leads dropped as false positives after source verification.
- In-diff vs out-of-diff: 5 in-diff/metadata, 0 out-of-diff. The zero is well-supported, not an underrun: the diff is 4 small hunks, and agents 1/4/5/11 each walked the full callsite inventory (every cancel() caller, every getState() caller, both prepareForNewQuery() callers) and confirmed the behavior change cannot reach any out-of-diff caller incorrectly.
- Tradeoffs/regressions: none. Hot-path cost is one extra L1-hot volatile load in testCancelled() per consult (a per-frame/throttled site, not per-row — the per-row statefulThrowExceptionIfTripped() is untouched), zero allocation. The change correctly hardens the no-leak invariant that the prior level-3 review's M1 called fragile.
- Net: fix the moderate test gap (#1) if you want the wiring protected, optionally tighten test #2 and the comment #3, correct the two body nits (#4), and push the branch so CI runs the simplified code.

bluestreak01 (approved):
Approve — level-3 review (all agents, per-finding source verification). The fix is correct, minimal, and closes the documented startup race: a PG-wire CancelRequest landing after resetTimer() but before QueryRegistry.register() binds the per-query flag previously left only the powerUpTime==MIN_VALUE sentinel, which testTimeout()'s now-MIN_VALUE overflow swallowed and testCancelled() never consulted. Honoring the sentinel in testCancelled() (and getState()), plus a per-query clearCancelSentinel(), is the right mechanism.

Verified: overflow arithmetic; PG SELECT exec ordering (resetTimer → getCursor → register → setCancelledFlag); breaker-instance identity; member ordering; test semantics. No correctness, concurrency, resource, or performance regressions. Hot-path cost is one extra L1-hot volatile load + a predicted-false branch, zero allocation, gated behind existing throttling.

Non-blocking notes:
- The fix narrows but cannot fully close the race: a cancel in [prepareForNewQuery..resetTimer] is still overwritten by resetTimer(); this is inherent to the PG cancel protocol (CancelRequest carries no query id) and is pre-existing, not a regression. No test pins the adverse half of that window.
- The user-visible wiring (prepareForNewQuery → clearCancelSentinel) has no end-to-end PG-wire assertion; the new tests are unit-on-breaker only, and the existing fuzz oracles explicitly accept a missed cancel.
- Minor: checkIfTripped(long,long) lacks the symmetric isCancelled() guard (still trips via overflow); getState(long,long) cancel branch reads the live field, ignoring millis; comment at PGConnectionContext:1287 doesn't hold for suspended portals.

False positives checked and dropped: parallel shared breakers (all AtomicBooleanCircuitBreaker, no sentinel), getState reaching UpdateOperation (AtomicBoolean), clearCancelSentinel dropping a live cancel (benign), HTTP path missing the clear (cancel() never called there). All findings in-diff/test-scope; 0 confirmed out-of-diff bugs.

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/sql/NetworkSqlExecutionCircuitBreaker.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGConnectionContext.java|1|1|100.00%|

Issue #52, opened 2016-06-23
Problems in Windows 10


jacuesta:
Hi, i would like to see the possible use of this library in our programs.
I've got a problem as I run the examples provided.

Exception in thread "main" java.lang.UnsatisfiedLinkError: C:\Users\Usuario\AppData\Local\Temp\libquestdb3846739714947417735.dll: %1 no es una aplicación Win32 válida

Both the Os and the JDK are 64bits. I reckon the origin of the error is that. Could you provide us a 64 bit version of the dll. 

Thanks in advance
Juan

bluestreak01:
Hi Juan,

the DDL is 64bit bit already. It is compiled on Windows XP though, i'll have to check on windows 10 and get back to you.

Regards,
Vlad

bluestreak01:
the error reads a bit suspicious though, DLL is indeed not a win32 application.

bluestreak01:
Another idea: do you have anti-virus running and could be preventing java from accessing DLL?

jacuesta:
Yes, i've got mcaffy running. I will try stopping it and let you know.
Thanks.

jacuesta:
Hi Vlad,
It was my mystake. I did a save as with the right mouse button on the link and I didn't realize the file was not a dll one.

In order to be able to run the examples in a Windows platform I had to modify the Os.java 

                if (type == LINUX){
                    tempLib = File.createTempFile(lib, ".so");
                }else if (type == OSX){
                    tempLib = File.createTempFile(lib, ".dylib");
                }else if (type == WINDOWS){
                    tempLib = File.createTempFile(lib, ".dll");
                }

I'll carry on testing the library.

Regards,
Juan

bluestreak01:
great, thanks Juan

PR #7331, opened 2026-06-25
chore(core): OIDC sign-in via device flow - OSS tandem

glasstiger:
The OSS half of the OIDC device-flow tandem. The feature code lives in the client PR linked below; this
repository's contribution is the `java-questdb-client` submodule bump — plus the one assertion that
belongs on this side, that the bumped client still ingests against a real server.

## What's here

**Submodule bump** — `java-questdb-client`, so OSS CI builds and runs the end-to-end tests against the
new client.

**`LineHttpSenderTokenProviderServerTest`** — real-server coverage for `httpTokenProvider`, the ingestion
half of the device-flow work. Until now nothing here exercised the token provider at all: the branch was
a submodule pointer and no other change. The client's own provider tests all drive a mock HTTP server,
which proves how a request is *built* but not that a provider-sourced token survives a real server's HTTP
stack and lands rows.

- **`testRotatingTokenProviderIngestsAgainstARealServer`** — pins that a rotating provider is queried
  *per request* rather than captured once. The provider hands out a different token every time it is
  asked; a sender that captured one token at build time would ingest happily here, so the pull count is
  what separates the two. Five flushes must produce at least five pulls, and all 50 rows must land.
- **`testRowsAfterAProviderFailureStillParseServerSide`** — drives the documented not-yet-signed-in
  shape, which is what `OidcDeviceAuth::getToken` looks like before `signIn()`: `build()` succeeds, the
  deferred pull surfaces the provider's own error at the first row, and the pending stamp is retried by a
  later row. The client's own test stops at *"the buffer is non-empty"* after recovery. What only a real
  server adds is that the request built around that failed stamp is not corrupt — the bytes still parse,
  and exactly the rows written after recovery arrive, so the failed row contributed nothing and left no
  partial line behind to corrupt the ones that followed.

## What this cannot cover

OIDC token **validation** is an Enterprise feature — open-source QuestDB has no `acl.oidc.*` and no
bearer-token ACL — so an open-source server accepts the request whatever the `Authorization` header says,
and no assertion here can prove the server *honoured* the credential. That belongs in the Enterprise
tandem, and the class javadoc says so rather than leaving the gap implied.

## What the bump now carries

A sixth review round landed eight commits on the client branch, so this bump is no longer the same client it
was. The client PR describes them in full; the two that reach an open-source user of this repository, neither
of them OIDC-specific, are:

- **`FileTokenStore`'s staleness floor rose from four to six times `httpTimeoutMillis`.** A refresh spends six
  independently bounded timeout budgets, not four -- `HttpClient` grants the TLS handshake its own budget
  rather than continuing the TCP connect's -- so the old floor could let a peer steal a live holder's lock
  mid-refresh. `build()` now rejects `httpTimeoutMillis(120_000)` paired with a default store, which it used
  to accept. Only callers who opted into token persistence can reach it, and the default 30s timeout is
  unaffected.
- **A desktop stack that cannot initialise no longer fails `signIn()`.** `java.awt.AWTError` is neither an
  `Exception` nor a `LinkageError`, so it escaped both guards on the best-effort browser launch. The stock
  `accessibility.properties` on several Linux distributions produces it.

The rest are a CPU busy-spin on the store-and-forward drainer's shutdown path, an interrupt swallowed by the
token store's rename retry, documentation of what `tlsConfig` covers, two regression tests, and comment and
member-order housekeeping.

## Tandem

Client: https://github.com/questdb/java-questdb-client/pull/52
Ent: https://github.com/questdb/questdb-enterprise/pull/1090


sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=262912)
**Tested Commit:** `2541836b68dbd39113619f0dfa5514c122f93753`

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=263183)
**Tested Commit:** `121412d56af53f91bc33370e4a857d1a44fbe62e`

glasstiger:
## Level-3 tandem review — all findings addressed

Reviewed as one logical change across the three repos. Findings were raised against
`java-questdb-client@09eb8b5b`, `questdb@39c2267473`, `questdb-enterprise@b5adec109`; all six are now
fixed at `java-questdb-client@3b313d15`, `questdb@4fdce5fb90`, `questdb-enterprise@487219f0af`.

**Verdict: approve** — all three PRs. No Critical findings were admitted; the test gate passed at the
reviewed revision and the four Moderate items are now closed. (Independent of the open
`CHANGES_REQUESTED` review on #52, which is its author's to resolve.)

**Submodule provenance.** `questdb` 271ec07c → 39c22674 and `java-questdb-client` 2489b243 → 09eb8b5b are
both **OFF-DEFAULT** feature-branch commits, so all three diffs are this PR's work — nothing is an
upstream sync whose contents could be waved through.

### Findings, and where each landed

| # | Severity | Finding | Fix |
|---|---|---|---|
| 1 | Moderate | The 2xx drain-abort disconnect (`!drained` in `flush0`) was pinned by no test | `java-questdb-client@bfc66a9a` |
| 2 | Moderate | `isRetryable()` unasserted on the unreadable-error-body wrapper | `java-questdb-client@1edc09ce` |
| 3 | Moderate | A third user-observable change missing from the PR body's own numbered list | this PR's description |
| 4 | Moderate | A new private helper orphaned `recvOrDie`'s javadoc in an exported package | `java-questdb-client@3b313d15` |
| 5 | Minor | Reflection into `QwpSidecarMain.err` where a same-package call reaches it | `questdb-enterprise@4332ee6f3` |
| 6 | Minor | New members broke the alphabetical order `CLAUDE.md` requires | `questdb-enterprise@eb5e8b518` |

Each production-behaviour claim was verified counterfactually rather than by a green run:

- Removing `!drained ||` sends a **committed batch three times over two flushes** — the second flush lands
  on a socket still carrying the first response's chunk-size digits, its header read never completes, the
  whole-call bound throws, and `flush0` classifies that as transport and retries.
- Hardcoding `retryable` at the wrapper's throw site reddens one test per direction: `false` breaks the new
  503 case, `true` breaks the new 401 assertion. Both were needed — `assertFalse` alone cannot distinguish
  "classified permanent" from "not classified at all", since the three unclassified constructors also
  report `false`.
- The javadoc claim was checked with a standalone `javadoc` run over both shapes: with a private helper
  between the comment and the method, the method renders with **no description at all**.
- De-reflecting the ERR test kept its teeth: making `sanitize()` a no-op still reddens it with a raw LF.
- Both member-ordering moves are pure code motion — the multiset of non-blank lines is identical to the
  base revision in both files, and file lengths are unchanged.

### Checked and found clean

Recorded so these are not re-litigated:

- **Exported API is purely additive.** A whole-tree diff of every `public`/`protected` member under
  `io.questdb.client`, base vs head, is **+77 / −0** — nothing removed or retyped. All 11 base
  `QwpWebSocketSender.connect` overloads and both `createLineSender` arities are byte-identical and pinned
  by `ExportedApiCompatibilityTest`, which also pins `Response.recv(int)` as `default` rather than abstract.
- **No out-of-diff breakage.** `questdb` and `questdb-enterprise` **production** code has zero
  `io.questdb.client.*` imports; only tests use `Response`/`HttpClient`, and the full OSS Cairo/Griffin/Other
  matrix was green at the reviewed revision — executed evidence against "the bounded `recv()` breaks an
  existing consumer".
- **Store-and-forward contract holds.** Credential-unavailable is retried indefinitely in the running loop
  with no wall-clock budget and no terminal latch; the rotating-401 gate increments its counter only on
  `QwpAuthFailedException`, so a transient class cannot consume the terminal budget.
- **Test-assertion rules.** No new `assertSql`/`returnsOnce` anywhere. The three `serverMain.assertSql(...)`
  calls in the new OSS test bind to `TestServerMain.assertSql(String,String)` — the sanctioned live-server
  wrapper, not the removed query-result helper.

### Examined and deliberately omitted

Two serious-sounding candidates were dropped at the admission gate rather than reported:

- A native-memory/fd leak via `PoolHousekeeper.stop()`'s new interrupt reaching `QwpQueryClient.close()`'s
  `InterruptedException` arm. That arm is **byte-identical to base**, and the new producer's reachability
  rests on an unexecuted race whose load-bearing step — an *idle*, non-borrowed worker's I/O thread taking
  ≥2 s to exit — could not be established (`reapIdle()` only touches `available`). Two independent lines of
  analysis both returned `producer: unknown`.
- `JsonLexer.wipe()` not zeroing its native `cache` stash. Disproved by the code's own stated scope at
  `OidcDeviceAuth.java:2223-2226`: the same token bytes are deliberately left in the HTTP receive buffers,
  so zeroing the stash changes nothing observable.

### Status

CI is still running at the new heads; nothing has failed so far, and `Validation Javadoc` has already
passed. The fixes are test-only apart from the javadoc move, and were verified locally — 81 client tests
across 12 classes and 15 enterprise tests, all green. Two cosmetic items inside `QwpSidecarMain` were left
untouched on purpose, as they were not among the reported findings: a fully-qualified
`java.util.concurrent.atomic.AtomicInteger` beside an imported `AtomicBoolean`, and `io.questdb.client.*`
imports interleaved into the `java.io.*` block.

<sub>Review and fixes by Claude Opus 5.</sub>

glasstiger:
## Code review — OIDC device flow (RFC 8628), level 3, tandem

Reviewed as one logical change across three repos:
`java-questdb-client#52` (`d0f924f2..903e3fde`), `questdb#7331` (`d55ccfd4..2d9b1fbd`),
`questdb-enterprise#1090` (`2aa857c3..9da1d0db`).

**Revised.** Findings were raised against `java-questdb-client@903e3fde`. The four Moderate items are
now fixed at `java-questdb-client@1f8c3ac0` (`25a80648..1f8c3ac0`); the three Minor items are still
open. Everything below reflects that.

### Critical

None.

### Moderate — all four fixed

**1. The interrupt escalation never ran when the closing thread carried an interrupt.**
`impl/PoolHousekeeper.java:72-105`, `impl/SenderPool.java:1711-1724` — in-diff.
Both put the new `if (thread.isAlive()) { thread.interrupt(); thread.join(...); }` inside the same
`try` whose first statement is `thread.join(STOP_TIMEOUT_MILLIS)`. `Thread.join(millis)` consults the
*calling* thread's flag before it looks at whether the target is alive, so a caller arriving
interrupted made that join throw at 0 ms and skip the escalation — in the one case it is most needed.
It compounded: `stop()`'s catch re-asserted the flag, so by the time `QuestDBImpl.close()` reached
`senderPool.close()` the second escalation was *guaranteed* to skip too. One carried flag disabled
both, and `close()` returned with the recoverer still holding its store-and-forward slot flock.

Measured on the project JDK, reproducing `stop()`'s exact shape:
`join THREW InterruptedException after 0 ms; escalation ran: false; target still alive: true; target interrupted: false`.
A `close()` from a task cancelled by `shutdownNow()`, or from a `finally` on a cancelled thread,
arrives exactly this way. The same file names the hazard 20 lines below the bug, and the branch
already applies clear-and-restore at four *callee* sites — but not at the two caller joins.

Fixed in **`25a80648`**: both sites now clear the flag for the duration and restore it before
returning. `testCloseBreaksARecoveryDelegateStuckInACredentialPullWhenTheCallerIsInterrupted` is the
sibling of the existing clean-caller test, differing only in setting the flag before `close()`, and
also asserts the flag is handed back rather than consumed. Reverting either site turns it red.

**2. `noteAckProgress()`'s rotating-credential resets had no test that could fail.**
`qwp/client/sf/cursor/BackgroundDrainer.java:897-905` — in-diff, coverage gap.
`capabilityGapAttempts = 0` is pinned by `BackgroundDrainerMidDrainCapabilityGapTest`. The two lines
beside it — `dynamicCredentialAuthAttempts` and `firstDynamicCredentialAuthFailureNanos` — were not:
deleting both left **2059 tests green**. Unreset, those counters span every session rather than the
current run of rejections, so rejections that were never consecutive accumulate to the threshold and
an orphan slot quarantines and reports `DATA_LOSS` over a credential outage that ack progress had
already shown to be transient — behind a `.failed` sentinel nothing in production clears.

Fixed in **`990f6902`**. This needed *two* tests, not one: the quarantine gate is an `AND`, so
dropping the attempt reset leaves the dwell short and the `OR` inside the gate still satisfied, while
dropping the anchor leaves the attempt count low. `GrantsAFreshRideOut` sets the dwell to 1 ms so only
the attempt count decides; `RestartsTheDwellAnchor` keeps the attempt count legitimate and puts 800 ms
of real wall clock in the delivering session against a 300 ms dwell, so only the anchor decides — the
margin is the sleep, not backoff timing. Deleting either line turns exactly its own test red.

**3. The cross-process lock steal could collapse to a bare delete undetected.**
`cutlass/auth/FileTokenStore.java:1369-1400` — in-diff, coverage gap.
`stealIfStale`'s own comment says a bare `deleteIfExists(lock)` "would admit two holders at once", yet
replacing the whole capture/verify/restore with exactly that left **249 tests green**. The two
`testRestoreCapturedLock*` cases look like coverage but invoke `restoreCapturedLock` *directly by
reflection*, so they pin the helper and not its call site;
`testConcurrentStealersLeaveExactlyOneWinner` asserts only "no stealer threw" and "the lock is gone",
both of which a bare delete satisfies.

Fixed in **`67ffa18b`**. Worth a second opinion: counting winners in the concurrency test — the fix I
first proposed — does **not** work, because capture-then-delete and a bare delete are
observationally identical once the lock is gone. The only arm that discriminates is the one where the
capture takes a *live* lock a peer recreated in the gap, and no amount of concurrency forces that
interleaving. So this adds a `@TestOnly beforeCaptureHook` that runs in that gap and nowhere else —
null in production, one null check on the lock-stealing path, idiomatic here (`SenderPool` carries
several such hooks), but it does reverse the fifth round's "no production seam" call. The test plants
a peer's live lock and asserts it survives byte for byte, because `releaseLock` verifies the owner
stamp before deleting. The bare-delete mutation now turns it red.

**4. The drain loop's comments described code that is not there.**
`BackgroundDrainer.java:1093-1094` and `:1120-1121` — in-diff.
The loop head said re-entry happens "ONLY when a mid-drain reconnect sweep hit a durable-ack
CAPABILITY gap"; this diff added a second re-entry class 57 lines below (`|| loop.authTerminal() != null`).
Anyone auditing whether that loop can spin without bound reads the "ONLY" and stops. Separately, the
newly added comment cited `QwpWebSocketSender.startCursorSendLoop`, a name that appears nowhere in
main or test except that comment — the real site is `ensureConnected()`.
Fixed in **`1f8c3ac0`** (comments only).

### Minor — still open

**5. `ExportedApiCompatibilityTest` ships a claim its own commit message disowns.**
`test/compat/ExportedApiCompatibilityTest.java:52` states *"There is no japicmp or revapi gate on this
build, so this test is the gate."* It pins 13 signatures and only checks that literally-listed names
are **present**, so it structurally cannot detect an unlisted removal. Commit `30529143`'s message
puts it accurately ("a targeted pin, not a general compatibility gate"); the shipped sentence does
not. Also `assertSignaturesPresent` skips any array entry not starting with `methodName + "("`,
guarded only by `expected > 0`, so a typo in one entry silently drops it while the test stays green.

**6. A regression test's precondition guard is vacuous.**
`OidcDeviceAuthPersistenceTest.java:1563` asserts
`assertEquals("the refresh must have run, or this test proves nothing...", 0, device.get())`.
`device` increments only under `if (DEVICE_PATH.equals(path))`, and the test's only production call is
`auth.getToken()` on a seeded entry, which never runs the device flow — so it is 0 by construction. A
token-endpoint counter is the guard that message describes.

**7. Import block disorder in the ENT sidecar.**
`questdb-ent/src/test/java/com/questdb/e2e/QwpSidecarMain.java:20-31` — five `io.questdb.client.*`
imports wedged into the `java.io.*` block, `OidcDeviceAuth` separated from its four `cutlass.auth`
siblings by a blank line, and `AtomicInteger` fully qualified at its field declaration while
`AtomicBoolean` is imported at line 38.

### Coverage map

**Test gate: passes.** Zero admitted Critical coverage gaps. The two admitted Moderate gaps (findings
2 and 3) are closed; both were established by executed mutation at the reviewed revision rather than
by inference, and both now have a test that goes red under that same mutation.

Every high-risk shared-infrastructure change has a named test with a real failure link, verified by
recorded search: `ResponseTest#testRecvHonoursTotalTimeoutWhenNoApplicationBytesArrive`,
`ChunkedResponseTest#testRecvHonoursTotalTimeoutWhileChunkSizeDribbles` plus all three chunk-size
overflow residues and the zero-padded false-positive case,
`LineHttpSenderErrorResponseTest#testQuestDbRowErrorRendersTheDecodedNewlineAsAnEscape`,
`HttpClientConstructorLeakTest`, `StringSinkWipeTest`, `DesktopFreeModulePathTest`.

Regression sweep after the fixes: `FileTokenStoreTest` 70, `OidcDeviceAuthTest` 130,
`OidcDeviceAuthPersistenceTest` 48, both `BackgroundDrainerMidDrain*` 5+5,
`SenderPoolSfTokenProviderTest` 4, plus the close-lifecycle and `QueryWorkerTest` suites — all green.

### Summary

**Verdict — client #52: approve. OSS #7331: approve. ENT #1090: approve.**

Both gates pass and the four Moderate items are closed. The three Minor items are optional.

- **Submodule provenance:** `questdb` → `2d9b1fbd08`: **OFF-DEFAULT**, in scope.
  `java-questdb-client` → `903e3fde`: **OFF-DEFAULT**, in scope. No upstream sync; all three diffs are
  this change.
- **Pointer bump still needed:** the fixes land in the client at `1f8c3ac0`, so OSS #7331's
  `java-questdb-client` pointer (and in turn ENT #1090's `questdb` pointer) need re-pinning to pick
  them up.
- **Tandem gate: passes.** Both counterparts exist on the matching `ia_oidc_device_flow` branch,
  cross-link bidirectionally, and run their covering suites — OSS real-server ingestion via the
  submodule bump, ENT HA-failover and `kill -9` python e2e via
  `ci/build-and-test.yaml` → `build-and-test-e2e.yaml` (`not fuzz`; neither new e2e file carries a
  skip or fuzz marker).
- **Split:** 7 in-diff, 0 out-of-diff-breakage. A `javap -p` diff of *both* built revisions across all
  27 exported packages found zero removed or retyped public/protected members — the only two
  disappearances are package-private — `module-info` exports are byte-identical, and no OSS or ENT
  **main** source imports `io.questdb.client.*` at all.
- **Java 8 floor: clean.** `onSpinWait` goes through the prescribed `src/main/java8/Compat` shim,
  `Files.readAllBytes(Path)` is Java 7, and there is no `var`, text block, enhanced switch or
  `instanceof` pattern outside a comment. No committed binaries in any of the three repos.
- **Severity distribution:** 0 Critical, 4 Moderate (all fixed), 3 Minor (open).

**Tradeoffs, all disclosed by the PR and all accurate as checked.** The `JsonLexer` escape decoding
changes the most common ILP error message from `\n` to `\u000a`; running both compiled revisions over
the exact server-shaped body confirms no OSS or ENT assertion breaks, and the same change makes a
server-escaped `\"` render correctly. The whole-call `recv` bound tightens existing non-OIDC flushes,
but head's per-call budget is never smaller than base's per-read budget and all five cross-repo
callers run on the 600 s default.

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=264398)
**Tested Commit:** `1c38e2c69738931c1e57797b819b12ca41c19f62`

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=264631)
**Tested Commit:** `f08ef2be8d50a7dcb11cb2fcf8d593b6c28f770f`

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=265205)
**Tested Commit:** `6c6cfd3c3285e78426f85d229aeb1f5a5599c714`

glasstiger:
## Tandem review — client #52 / OSS #7331 / ENT #1090

Reviewed as one logical change at client `3e534eb8`, OSS `bc26767e`, ENT `eda397ea`.

**Submodule provenance.** `questdb` and `java-questdb-client` pointers are both OFF-DEFAULT (branch-only), so all three diffs are this change's work — no upstream content is being attributed here.

**Verdict: approve on all three.** No Critical findings. Three Moderate items and a Minor bundle remain open below; none blocks. The test gate passes (one admitted coverage gap, Moderate).

Compile compatibility is settled: public signature sets diffed base-vs-head show zero removals, both new interface methods are `default`, and `mvn -pl core test-compile` exits 0. Full client suite is green at **3452 tests, 0 failures, 0 errors, 5 skipped**.

---

### Resolved since the first pass

Two findings from the initial review are fixed in `3e534eb8` (**disclosure: I wrote that commit**, so it warrants a second pair of eyes rather than my own sign-off):

1. **An unreadable `.store.lock` permanently wedged the token store.** `stealIfStale` returned as soon as `readLockHolder` threw, so a directory, a symlink or a mode-000 file at the lock name was never reclaimed — `CREATE_NEW` reports EEXIST for all three exactly as for a peer's live lock, and nothing else sweeps that name. `load()`/`save()` then failed permanently, `OidcDeviceAuth` degraded to "continuing without persistence", and every process start re-ran the interactive device flow — a hard failure for the headless `getToken()` consumer persistence exists for. Reachable without an attacker: a run under a different uid killed while holding the lock leaves a 0600 file this uid cannot read.

   The fix splits the shapes. A directory or symlink is a squatter no wait turns into a lock, so it is displaced on sight (mirroring `markUntrusted`'s handling of a squatted `.untrusted` name), capturing atomically first so a peer's real lock created in the gap is restored rather than deleted. An unreadable regular file may be a live holder's, so it is not displaced: `stealIfStale` carries "stamp unreadable" as a third state, ages it on the full staleness window, and folds readability into the capture-verify.

2. **The non-POSIX trust check ignored the `.untrusted` sentinel.** `restrictToOwner()`'s `UnsupportedOperationException` branch returned a bare `true`, so a directory a POSIX peer had marked untrusted and not finished sweeping read as trusted on Windows — entries adopted, mark never cleared. It now evaluates the sentinel, matching its sibling `canUseShortDirectoryLockLease()` and `design/oidc-token-persistence.md:325` ("whatever the permission bits say").

Four regression tests were added. The three that pin the wedge fail against the previous code with the reported `OidcAuthException` and pass now; the fourth pins the safety property the ageing buys — a *fresh* unreadable lock may be a live holder's and must survive. The non-POSIX branch remains untested (needs a Windows agent or a synthetic `FileSystemProvider`), as the class javadoc already records.

---

### Open — Moderate

#### M1. `clearCache()` is not published as the token-store waiter

**Problem:** `clearCache()` enters `TokenStore.inLock` without publishing itself, so `close()` cannot interrupt it.
**Net impact:** `close()` blocks for a peer's full refresh budget, against its own documented bound.
**Evidence:** Static. Sole `publishTokenStoreWaiter()` call site is `OidcDeviceAuth.java:2271`; `clearCache()` calls the store at `:488`; the contract is claimed at `:504-505`.

`close()` states: *"Only while an operation is waiting for a peer instance's shared token-store lock is its thread interrupted, abandoning that wait immediately."* `clearCache()` is such an operation and isn't covered — `close()` reads `tokenStoreWaiterThread` as null (`:529`) and then blocks on the plain `lock.lock()` at `:534`. `FileTokenStore.inLock`'s wait is `lockInterruptibly()` with no budget.

Provenance: `f99e9c1b` removed `publishLockHolder()` from five sites including `clearCache()`, re-added it only in `tryRefreshCoordinated()`, and strengthened this javadoc in the same diff. `OidcDeviceAuthPersistenceTest:399-401` pins the `getToken()` shape; there is no `clearCache()` counterpart.

*Suggested fix:* a tightly scoped publish/clear pair around `:488`. Note `FileTokenStore.clear` captures `wasInterrupted == false` at `:369`, so a naive pair leaves a stray flag on the caller.

#### M2. `drainOnClose()`'s re-asserted interrupt defeats `close()`'s own interrupt-neutrality

**Problem:** A consumed interrupt is re-asserted, so the next wait fails at 0 ms.
**Net impact:** `QuestDB.close()` reports a stopped I/O thread as not stopped and skips the drainer graceful window.
**Evidence:** Reproduced 6/6 at head (`LineSenderException: cursor I/O thread did not stop…`, `isSlotLockReleased()==false` on return). `CountDownLatch.await(t,u)` throws at 0 ms with a preset flag **even when the latch is already at zero** — measured.

`QwpWebSocketSender.java:3959-3963` → `:1435`/`:1452`. `close()` clears a carried flag once at `:1295`, but `drainOnClose()`'s `finally` re-asserts on every exit, and `close0()` then runs `cursorSendLoop.close()` and `drainerPool.close()` on the same stack with no intervening clear. The flag's provenance is lost: `PoolHousekeeper.stop():115` interrupts to break a *credential pull*, and `BackgroundDrainerPool.close()` then reads it as caller intent and skips both the 2.5 s graceful and 0.5 s stop windows — documented there as deliberate for an *application* interrupt. `PoolHousekeeper.stop()`'s own comment states the invariant broken here: *"The flag must not outlive the interrupt's target."*

Not Critical: for the identical trigger, base produced the same failed-stop throw **plus** a full-core busy-spin, so head is better there; only the producer is new. Consequences are bounded — `SenderPool.reapIdle` swallows the throw, cleanup completes ~50 ms later via `delegateClose`, and a retired slot has three recovery paths. No data loss.

*Suggested fix:* clear once more at the top of `close0()`'s teardown and let `close()`'s outermost `finally` restore. A second route exists — an interrupt landing after the final iteration's `Thread.interrupted()` check is never consumed — so removing the re-assert alone does not close it.

#### M3. Coverage: nothing pins the new 401 terminal as orphan-only

**Problem:** No test asserts the rotating-credential 401 terminal is unreachable from the live drainer.
**Net impact:** A future change moving it onto the foreground path would drop a live producer silently.
**Evidence:** `connectWithDurableAckRetry` callers are only `BackgroundDrainer.java:1075` and `:1246`; `ReconnectPolicy.ORPHAN` is constructed only at `:1143`.

Every other aspect of the ride-out is pinned well — attempt threshold without dwell, constant-credential immediate quarantine, the clamped dwell at the call site, one test per transient arm proving it rewinds the dwell but not the attempt count, both `noteAckProgress` resets separated, credential tagging across all four shapes. The orphan-only scoping is the one property enforced purely by construction, and a regression there is the SF data-loss class the carve-out exists to prevent.

*Suggested test (~60 lines, seams already exist):* in `WebSocketTokenProviderTest`, arm `TestWebSocketServer.setRejectWithStatus(401, …)` (`:299`) after the first ACK on a live WS `Sender` with `httpTokenProvider`, sleep past `reconnectMaxDurationMillis`, then assert the sender still accepts rows, no `.failed` sentinel appears under `sf_dir`, and no `DATA_LOSS` reaches the handler.

---

### Open — Minor

- `FileTokenStoreTest.java:2551` — `store.inLock(key, …)`'s boolean is discarded and the sole assertion lives inside the lambda; `inLock` has three early returns that never enter it. The file's own convention captures the boolean at 14 sibling sites.
- `OidcDeviceAuthPersistenceTest.java:381` — asserts `Thread.State.WAITING` as a proxy for "queued on the shared process lock"; any untimed park satisfies it. The stronger form already ships here (`FileTokenStoreTest:3146 awaitWaitingInside`, `OidcDeviceAuthTest:4298 awaitInside`).
- `BackgroundDrainerMidDrainAuthRejectTest.java:396` and `:494` — byte-identical `durableAckFrame`/`okFrame` pairs in one compilation unit; hoist to file scope.
- `WebSocketTokenProviderTest.java:706` — `connectionsAccepted` written, never read (copied from `ReconnectTest`, where it *is* read), along with the `firstClient` branch feeding it.
- `BackgroundDrainerMidDrainAuthRejectTest.java:176` — comment says the session "sleeps 800ms before it acks"; `01fe9c99` moved the sleep after the ack, and the helper's javadoc at `:429` says the opposite. No efficacy impact.

---

### Coverage

Verified COVERED with real failure links: `validateToken` CR/LF and both boundary pairs; the snapshot-before-validate contract at all three production pull sites via the deterministic `HandOffCharSequence`; POSIX 0600/0700 with a negative control preventing vacuous passes; cross-process lock contention, stale-lock takeover and crash-during-save (unit plus the two-sided ENT kill-9 e2e); the interrupt escalation across three modes including caller-already-interrupted and interrupt-during-join; `requires static java.desktop` degrading on a genuinely desktop-free module path; the `JsonLexer` value cap driven through the production instance; and token secrecy, including reflective sweeps with positive preconditions so an empty buffer cannot pass. Windows/non-POSIX arms are an accepted gap, disclosed in the class javadoc and the design doc's open questions.

### Notes

Three behaviours existing non-OIDC ILP-over-HTTP users can observe — JSON escape decoding reaching the error-response parser, whole-call read bounds, and a `FileTokenStore` lock-hold multiple that turns a previously-accepted config into a `build()` failure — are all disclosed in the PR body with their pinning tests named, which is why none is filed above.

Java 8 compatibility verified by compilation: 421 classes at `-source 8 -target 8`, zero errors; `--release 8` flags only pre-existing `sun.misc` usage.

*Reviewed with Claude Code.*

glasstiger:
## Level 3 tandem review

Reviewed in tandem with questdb/java-questdb-client#52 and questdb/questdb-enterprise#1090.

### Findings

No admitted findings attributable to the OSS diff.

The tandem review has one non-blocking Moderate coverage finding in the client PR and one in the enterprise PR; each is posted on its owning PR.

### Coverage map

The OSS test gate passes with 0 admitted coverage gaps. The real-server token-provider coverage and the full GitHub/Azure test matrices are green.

### Summary

- Client PR #52: **approve**
- OSS PR #7331: **approve**
- Enterprise PR #1090: **approve**
- Correctness gate: pass
- Test gate: pass
- Severity on this PR: 0 Critical, 0 Moderate, 0 Minor
- Attribution on this PR: 0 in-diff, 0 out-of-diff breakage
- `questdb` pointer in enterprise: **OFF-DEFAULT — in scope**
- `java-questdb-client` pointer in OSS: **OFF-DEFAULT — in scope**

Both tandem submodule contents were independently reviewed because their PRs were explicitly supplied. No other submodule contents were expanded.

bluestreak01 (approved):

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=266058)
**Tested Commit:** `15ffa163f0f0810ea49246b355c78fbc7e6e0574`

PR #7332, opened 2026-06-26
fix(core): fix POSTING-indexed tables suspending WAL apply after out-of-order writes

RaphDal:
## What this fixes

A POSTING-indexed column can carry a column-version record on a partition while having no rows there (`columnTop >= partition size`). Whether such a row-less column has a `.pk` key file depends on the partition: on a *historic* partition it has none (the seal sweep `sealPostingIndexForPartition` skips row-less columns and never builds one), while on the *active* partition `ADD COLUMN ... INDEX TYPE POSTING` builds an empty `.pk` even with no rows.

Several per-partition code paths iterate posting columns guarded only against the *absent* case (`columnTop == -1`), not the *row-less* case. Two distinct faults follow:

- Reaching a row-less column with no `.pk` and opening/linking it throws `CairoException: index does not exist` / `index files do not exist`. Both fixed sites hit this.
- Conversely, blanket-skipping *every* row-less column is wrong where one can legitimately have a `.pk`. `restorePostingIndexersToLastPartition` re-points indexers at the active partition, so skipping a row-less column that has a `.pk` there silently strands its index — rows stay in the base data but become invisible to indexed predicates, with no exception and no table suspension.

This branch fixes the two sites: `restorePostingIndexersToLastPartition` discriminates by `.pk` existence (so it skips the no-`.pk` case and re-points the `.pk`-present case), and the parquet-switch link mirrors its historic-partition sibling. It also aligns a third guard so it cannot silently rot.

## Fix 1 — restoring posting indexers after an O3 seal

`restorePostingIndexersToLastPartition()` re-points every posting indexer at the last open partition `L` after the O3 seal sweep, so later appends and rollbacks target `L`. Its guard only caught absent columns (`columnTop < 0`), so a row-less column with no key file fell through to `PostingIndexWriter.of(isInit=false)`, which fails at `!ff.exists(keyFile)` (`PostingIndexWriter.java:1093`):

```
CairoException: index does not exist [path=.../<col>.pk.<txn>]
```

User-visible impact of that crash:

- On a WAL table, `ApplyWal2TableJob` catches the exception and suspends the table mid-apply. The data is committed, but apply stops and the table needs manual resume.
- Under `BYPASS WAL`, the commit throws straight back to the caller (the INSERT fails).

A row-less column on `L` (`columnTop >= partition size`) may or may not have a `.pk`, and the two cases need opposite handling, so `columnTop` alone cannot decide:

- **No `.pk`** — e.g. a last-partition split tail the seal sweep never built a key file for. Re-pointing throws `index does not exist`, so restore must skip.
- **`.pk` present** — `ADD COLUMN ... INDEX TYPE POSTING` on the active partition `L` builds an empty `.pk` and stamps `columnTop == partition size`. Restore must re-point this indexer to `L`. Skipping it leaves the indexer where the seal sweep last pointed it — an older partition `P` that an O3 write back-filled `L`'s column into — and the next in-order append then reads `L`'s data but writes into `P`'s `.pk`/`.pv`. The row never reaches `L`'s index: invisible to `WHERE col = ...`, no exception, no suspension.

So restore discriminates on the key file's actual presence, not on row-less-ness — keep the `columnTop == -1` skip, and skip a row-less column only when its `.pk` is absent:

```java
long partitionSize = txWriter.getPartitionRowCountByTimestamp(lastOpenPartitionTs);
...
long columnTop = columnVersionWriter.getColumnTop(lastOpenPartitionTs, colIdx);
if (columnTop == -1) {
    continue;
}
...
if (columnTop >= partitionSize) {
    boolean keyFileExists = ff.exists(
            keyFileName(metadata.getColumnIndexType(colIdx), path.trimTo(plen), colName, colNameTxn));
    path.trimTo(plen);
    if (!keyFileExists) {
        continue;
    }
}
```

The accessor also moves from `getColumnTopQuick` to `getColumnTop`: `getColumnTopQuick` returns `0` (never `-1`) for an absent column, so the `columnTop == -1` arm was dead and an absent column — one whose home partition was dropped so it no longer exists on `L` — would have fallen through the same way. `getColumnTop` returns `-1` for that case. Restore reads `partitionSize` once above the loop (it is loop-invariant), matching the sibling sweeps.

The change is in the method body, so both callers benefit: the O3 seal sweep (`sealPostingIndexesForO3Partitions`) and the squash path.

## Fix 2 — linking index files when switching a partition to parquet

`linkPartitionIndexFiles()` (reached from the public `switchNativePartitionWithParquet()` API) hard-links each indexed column's index files from the source native partition into the new partition directory. Its guard was `getColumnTop(...) == -1` only — the comment said "no data in partition for this column", but a row-less column also has no data there while `getColumnTop` returns `partition size` (≥0, not `-1`), so it was not skipped. Control then reached `linkColumnIndexFiles()`, whose `linkFile(.pk)` throws when the key file was never built:

```
CairoException: index files do not exist [path=.../<partition>]
```

The guard now mirrors the parallel helper for the same feature, `copyOrRebuildColumnIndexes()` (which already carries the full `colTop == -1 || colTop >= partitionRowCount` guard):

```java
final long partitionSize = txWriter.getPartitionRowCountByTimestamp(partitionTimestamp);
...
final long columnTop = columnVersionWriter.getColumnTop(partitionTimestamp, columnIndex);
if (columnTop == -1 || columnTop >= partitionSize) {
    continue;
}
```

Unlike Fix 1, the blanket row-less skip is correct here: the switched partition is historic, where a row-less posting column never has a `.pk` (the seal sweep skips it and `ADD COLUMN ... INDEX` only builds a `.pk` on the active partition), so there is never a key file to link.

Severity note: the in-OSS `ALTER TABLE ... CONVERT PARTITION TO PARQUET` path goes through `convertPartitionNativeToParquet` → `copyOrRebuildColumnIndexes`, which was already guarded, so this site is reachable only through the public `switchNativePartitionWithParquet` API. It is the same absent-only guard gap as Fix 1's crash, but lower-severity within OSS than the WAL-suspend path.

## Root cause

Both fixed sites iterate posting columns and were guarded only against the absent case (`columnTop == -1`), not the row-less case (`columnTop >= partition size`). But the correct response to a row-less column differs by what the site does:

- **Fix 2 and the seal/copy siblings** (`sealPostingIndexForPartition`, `copyOrRebuildColumnIndexes`, `linkColumnIndexFiles`) operate on historic partitions, where a row-less posting column never has a `.pk`. There the row-less blanket skip is correct; Fix 2 simply brings `linkPartitionIndexFiles` into line with `copyOrRebuildColumnIndexes`.
- **Fix 1** re-points the live writer at the *active* last partition `L`, where a row-less posting column added via `ADD COLUMN ... INDEX` *does* have a `.pk`. A blanket row-less skip there is over-broad — it strands that indexer (silent index data loss) — so restore discriminates on the `.pk`'s presence instead.

The row-less-without-`.pk` state itself arises from a multi-partition, same-transaction O3 write interleaved with `ADD COLUMN`: the column-version bookkeeping stamps a `columnTop` onto a partition the column has no rows in, and the seal then skips building its `.pk`.

## Effects and tradeoffs

- Fix 2 only adds a skip; on the historic partition it touches, a row-less posting column has no `.pk` to link. Columns that have rows (and a key file) are unaffected.
- Fix 1 adds a skip for row-less columns with no `.pk`, and now correctly *re-points* row-less columns that do have one (added via `ADD COLUMN ... INDEX` on the active partition). Columns that have rows are unaffected. The re-point path is the same `configureFollowerAndWriter` call already used for every populated column.
- Fix 1's accessor change (`getColumnTopQuick` -> `getColumnTop`) is behavior-equivalent for every column present on the last partition (both return the same top); it differs only for an absent column, which previously fell through and now is correctly skipped.
- Cost: Fix 1 adds one `ff.exists(...)` per row-less posting column on the restore path (off the hot write path); both fixes add one `getPartitionRowCountByTimestamp(...)` lookup, now hoisted out of the per-column loop. Negligible.
- Narrow scope: no data migration, no format change. Outside the row-less/absent interleaving, behavior is unchanged.

## Audit

Beyond the observed failure, I swept every posting-index open/link/restore site to find siblings of the same family. Fix 2 came out of that sweep. The remaining `PostingIndexWriter.of(isInit=false)` and `configureFollowerAndWriter` sites are each protected by one of: the full two-arm guard, a preceding `createIndexFiles(...)` that materializes the `.pk`, an explicit `ff.exists` check, a `skipForPosting` gate, or the O3 worker's self-correcting `isInit = (row == 0)`. I investigated one adjacent lead — `mapCoveringColumnsForSeal` opening a covered column's `.d` without an absence skip — but left it out: it operates on a covered data file rather than a posting sidecar, and I could not show that a covering posting index ever has a covered column absent on a resealed partition.

## Test plan

- `PostingIndexCriticalIssuesTest#testO3SealRestoreSkipsRowlessPostingColumnWithoutKeyFile` (Fix 1, no-`.pk` arm). Builds a partition where a later-added POSTING column is row-less, forces a last-partition split so the seal sweep restores indexers there, and uses a scoped `FilesFacade` to report that column's `.pk` as absent — the production state. Fails before the fix with `index does not exist [.../extra.pk.1]` at `restorePostingIndexersToLastPartition`; passes after. It then reads the row-less column itself — `WHERE extra = 'a'` returns 0 and `WHERE extra IS NULL` returns 21 — confirming the recovered row-less posting column stays queryable.
- `PostingIndexCriticalIssuesTest#testO3SealRestoreRowlessNoKeyFileDoesNotSuspendWalTable` (Fix 1, no-`.pk` arm, WAL). The WAL counterpart of the test above: the same row-less keyless `extra` column and last-partition split, but on a `WAL` table applied through `drainWalQueue()`. Before the fix `ApplyWal2TableJob` catches the `index does not exist` throw and suspends the table mid-apply (error `.../2024-01-01/extra.pk.1`); after the fix apply completes, the table is not suspended, all 21 rows are present, and the `sym` index serves reads. This is the regression guard for the headline WAL-suspend symptom (proven to suspend-and-fail with the skip block neutralized, pass with it in place).
- `PostingIndexCriticalIssuesTest#testO3SealRestoreSkipStrandsRowlessIndexerWithKeyFile` (Fix 1, `.pk`-present arm). Plain SQL: adds a POSTING column on the active partition (row-less there, with a `.pk`), back-fills it into an older partition via an O3 write (which points its indexer at the older partition), then appends in order to the active partition. Asserts the appended row is visible to `WHERE c = ...`. With a blanket row-less skip the read returns 0 (silent data loss); with the existence-based skip it returns the row. The test also passes on master — master's `getColumnTopQuick` path already re-points the column — so it is not a before/after regression for the shipped diff; it is a forward guard pinning the existence-check arm against a future simplification to a blanket row-less skip.
- `TableWriterTest#testSwitchNativePartitionWithParquetSkipsRowlessPostingColumn` (Fix 2). Adds a posting column to a historic partition, stamps it row-less (`columnTop == partition size`, no `.pk`), and switches that partition to parquet. Fails before the fix with `index files do not exist [.../2013-03-04]` at `linkColumnIndexFiles`; passes after. Also asserts positively that the live `sym` column's `.pk` *was* linked into the new parquet partition dir while `extra`'s is absent — so an over-broad guard that dropped the live column too would be caught (it would link nothing yet still return `SWITCH_OK`).
- Regression sweep, all green: `PostingIndexCriticalIssuesTest` (118), `TableWriterTest` (157), `PostingSealPurgeTest` (22); plus `CoveringIndexTest` (398), `PostingIndexOracleTest` (16), `PostingIndexConcurrencyTest` (12), `PostingIndexNativeTest` (5), `PostingIndexO3ConcurrencyFuzzTest` (3). These counts are from the prior full sweep and predate the latest test additions/edits; I re-ran the four added/changed methods (`testO3SealRestoreSkipsRowlessPostingColumnWithoutKeyFile`, `testO3SealRestoreSkipStrandsRowlessIndexerWithKeyFile`, `testO3SealRestoreRowlessNoKeyFileDoesNotSuspendWalTable`, `testSwitchNativePartitionWithParquetSkipsRowlessPostingColumn`) green individually this round, but a confirming full re-run before merge is advisable.

Run the new/changed tests:

```bash
mvnd -pl core \
  -Dtest='PostingIndexCriticalIssuesTest#testO3SealRestoreSkipsRowlessPostingColumnWithoutKeyFile+testO3SealRestoreSkipStrandsRowlessIndexerWithKeyFile+testO3SealRestoreRowlessNoKeyFileDoesNotSuspendWalTable,TableWriterTest#testSwitchNativePartitionWithParquetSkipsRowlessPostingColumn' \
  -DfailIfNoTests=false test
```

### DIFF
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -7485,12 +7485,15 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
     private void linkPartitionIndexFiles(long partitionTimestamp, long partitionNameTxn, int partitionDirLen, int newPartitionDirLen) {
         try {
             final int columnCount = metadata.getColumnCount();
+            final long partitionSize = txWriter.getPartitionRowCountByTimestamp(partitionTimestamp);
             for (int columnIndex = 0; columnIndex < columnCount; columnIndex++) {
                 if (!ColumnType.isSymbol(metadata.getColumnType(columnIndex)) || !metadata.isIndexed(columnIndex)) {
                     continue;
                 }
-                // no data in partition for this column
-                if (columnVersionWriter.getColumnTop(partitionTimestamp, columnIndex) == -1) {
+                // Absent (columnTop == -1) or row-less (columnTop >= partition size) columns
+                // have no index files in this partition. Same guard as copyOrRebuildColumnIndexes.
+                final long columnTop = columnVersionWriter.getColumnTop(partitionTimestamp, columnIndex);
+                if (columnTop == -1 || columnTop >= partitionSize) {
                     continue;
                 }
                 linkColumnIndexFiles(
@@ -12349,10 +12352,11 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
                 }
                 // Absent (columnTop == -1) or row-less (columnTop >= partition size) columns
                 // have no .pk file here, so there is no index to restore. Same guard as
-                // sealPostingIndexForPartition.
+                // sealPostingIndexForPartition -- use getColumnTop (sentinel -1), not
+                // getColumnTopQuick (returns 0 for an absent column, so the -1 arm is dead).
                 long partitionSize = txWriter.getPartitionRowCountByTimestamp(lastOpenPartitionTs);
-                long columnTop = columnVersionWriter.getColumnTopQuick(lastOpenPartitionTs, colIdx);
-                if (columnTop < 0 || columnTop >= partitionSize) {
+                long columnTop = columnVersionWriter.getColumnTop(lastOpenPartitionTs, colIdx);
+                if (columnTop == -1 || columnTop >= partitionSize) {
                     continue;
                 }
                 CharSequence colName = metadata.getColumnName(colIdx);
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -12347,10 +12347,12 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
                 if (indexer == null) {
                     continue;
                 }
-                // columnTop == -1 means the column does not exist on this partition at
-                // all (no .pk file), so skip those.
+                // Absent (columnTop == -1) or row-less (columnTop >= partition size) columns
+                // have no .pk file here, so there is no index to restore. Same guard as
+                // sealPostingIndexForPartition.
+                long partitionSize = txWriter.getPartitionRowCountByTimestamp(lastOpenPartitionTs);
                 long columnTop = columnVersionWriter.getColumnTopQuick(lastOpenPartitionTs, colIdx);
-                if (columnTop < 0) {
+                if (columnTop < 0 || columnTop >= partitionSize) {
                     continue;
                 }
                 CharSequence colName = metadata.getColumnName(colIdx);
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -12340,6 +12340,7 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
         path.trimTo(pathSize);
         setPathForNativePartition(path, timestampType, partitionBy, lastOpenPartitionTs, currentNameTxn);
         int plen = path.size();
+        long partitionSize = txWriter.getPartitionRowCountByTimestamp(lastOpenPartitionTs);
         try {
             for (int colIdx = 0; colIdx < columnCount; colIdx++) {
                 if (metadata.getColumnType(colIdx) <= 0 || !metadata.isColumnIndexed(colIdx)
@@ -12350,17 +12351,26 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
                 if (indexer == null) {
                     continue;
                 }
-                // Absent (columnTop == -1) or row-less (columnTop >= partition size) columns
-                // have no .pk file here, so there is no index to restore. Same guard as
-                // sealPostingIndexForPartition -- use getColumnTop (sentinel -1), not
-                // getColumnTopQuick (returns 0 for an absent column, so the -1 arm is dead).
-                long partitionSize = txWriter.getPartitionRowCountByTimestamp(lastOpenPartitionTs);
+                // A column added after this partition (columnTop == -1) has no .pk here.
                 long columnTop = columnVersionWriter.getColumnTop(lastOpenPartitionTs, colIdx);
-                if (columnTop == -1 || columnTop >= partitionSize) {
+                if (columnTop == -1) {
                     continue;
                 }
                 CharSequence colName = metadata.getColumnName(colIdx);
                 long colNameTxn = columnVersionWriter.getColumnNameTxn(lastOpenPartitionTs, colIdx);
+                // A row-less column (columnTop >= partition size) has a .pk only if ADD COLUMN
+                // INDEX TYPE POSTING built one here on the active partition; a split tail whose
+                // seal never built one has none. Discriminate on the .pk's presence: skipping a
+                // present .pk strands the indexer on an older partition (silent index loss),
+                // while opening an absent one throws "index does not exist".
+                if (columnTop >= partitionSize) {
+                    boolean keyFileExists = ff.exists(
+                            keyFileName(metadata.getColumnIndexType(colIdx), path.trimTo(plen), colName, colNameTxn));
+                    path.trimTo(plen);
+                    if (!keyFileExists) {
+                        continue;
+                    }
+                }
                 indexer.configureFollowerAndWriter(
                         path.trimTo(plen), colName, colNameTxn,
                         getPrimaryColumn(colIdx), columnTop,
--- a/core/src/main/java/io/questdb/cairo/TableWriter.java
+++ b/core/src/main/java/io/questdb/cairo/TableWriter.java
@@ -12351,7 +12351,8 @@ public class TableWriter implements TableWriterAPI, MetadataService, Closeable {
                 if (indexer == null) {
                     continue;
                 }
-                // A column added after this partition (columnTop == -1) has no .pk here.
+                // Defensive: columnTop == -1 means the column is absent on this partition
+                // (no .pk). Normally unreachable on the last partition; kept as a guard.
                 long columnTop = columnVersionWriter.getColumnTop(lastOpenPartitionTs, colIdx);
                 if (columnTop == -1) {
                     continue;
--- a/core/src/test/java/io/questdb/test/cairo/PostingIndexCriticalIssuesTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/PostingIndexCriticalIssuesTest.java
@@ -760,6 +760,54 @@ public class PostingIndexCriticalIssuesTest extends AbstractCairoTest {
         });
     }
 
+    @Test
+    public void testO3SealRestoreSkipsRowlessPostingColumnWithoutKeyFile() throws Exception {
+        // A POSTING-indexed column with no rows in a partition (columnTop >= partition
+        // size) has no .pk key file there: the O3 seal sweep skips row-less columns
+        // (sealPostingIndexForPartition), so it never builds one. The seal-sweep tail
+        // restorePostingIndexersToLastPartition used to open that column's index anyway
+        // and threw "index does not exist", which suspends a WAL table mid-apply (here,
+        // BYPASS WAL, it surfaces as the INSERT throwing).
+        //
+        // A last-partition split shrinks 2024-01-01 while `extra` stays row-less there
+        // (columnTop above the shrunk row count). The FilesFacade reports `extra`'s key
+        // file as absent on the day partition the seal-sweep restores to -- the exact
+        // production state -- but only once ADD COLUMN has built it (so the legitimate
+        // openNewColumnFiles path during ADD COLUMN is unaffected), and only for the
+        // day partition (not the split tail, whose dir carries a "...T..." suffix).
+        node1.setProperty(PropertyKey.CAIRO_O3_PARTITION_SPLIT_MIN_SIZE, 1);
+        node1.setProperty(PropertyKey.CAIRO_O3_LAST_PARTITION_MAX_SPLITS, 20);
+        node1.setProperty(PropertyKey.CAIRO_O3_MID_PARTITION_MAX_SPLITS, 20);
+        final AtomicBoolean keyFileHidden = new AtomicBoolean(false);
+        ff = new TestFilesFacadeImpl() {
+            @Override
+            public boolean exists(LPSZ path) {
+                if (keyFileHidden.get()
+                        && Utf8s.containsAscii(path, "extra.pk")
+                        && !Utf8s.containsAscii(path, "2024-01-01T")) {
+                    return false;
+                }
+                return super.exists(path);
+            }
+        };
+        assertMemoryLeak(ff, () -> {
+            execute("CREATE TABLE t_rowless (ts TIMESTAMP, sym SYMBOL INDEX TYPE POSTING, v LONG) TIMESTAMP(ts) PARTITION BY DAY BYPASS WAL");
+            execute("INSERT INTO t_rowless(ts, sym, v) SELECT ('2024-01-01T00:00:00.000000Z'::timestamp + x * 3600000000L)::timestamp, 'a', x FROM long_sequence(20)");
+            // extra has no rows on 2024-01-01 (columnTop == partition size).
+            execute("ALTER TABLE t_rowless ADD COLUMN extra SYMBOL INDEX TYPE POSTING");
+
+            keyFileHidden.set(true);
+            // O3 insert forces a last-partition split; the seal sweep then restores the
+            // posting indexers to the shrunk 2024-01-01, where `extra` is row-less.
+            execute("INSERT INTO t_rowless(ts, sym, v) VALUES ('2024-01-01T18:30:00.000000Z', 'a', 999)");
+            keyFileHidden.set(false);
+
+            // The table survived: all rows present, and the `sym` posting index serves reads.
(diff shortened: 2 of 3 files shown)


RaphDal:
# Review: PR #7332 — `chore(core): keep POSTING-indexed tables queryable and WAL apply running after out-of-order writes`

**Level 2** (full change-surface map; agents 1–7 + 11; batched verification). *Level 3
exists for the full 11-agent mission-critical pass — not warranted here given the
contained, private-method diff.*

Reviewed: 29 lines in `TableWriter.java` (two private methods) + 4 tests. The fix guards
two posting-index sites against the *row-less* column case (`columnTop >= partition size`),
where they previously only guarded the *absent* case (`columnTop == -1`).

## Critical

**None.** No correctness, concurrency, resource, or reachability bug survived verification.

## Moderate

**None.** The one substantive candidate — the no-`.pk` skip leaving an indexer stranded —
was traced to a self-heal and verified benign (see Downgraded).

## Minor

1. **Comment verbosity vs the project's concise-comment standard** *(in-diff)* —
   `TableWriter.java:12362–12366` is a 5-line inline block, and several new test comments
   run 10–13 lines (e.g. `PostingIndexCriticalIssuesTest#testO3SealRestoreSkipStrandsRowlessIndexerWithKeyFile`).
   They document genuinely non-obvious logic, but the per-branch failure-mode detail
   ("strands the indexer on an older partition (silent index loss)" / "throws index does not
   exist") is deep rationale that already lives in PR.md. The adjacent 2-line guards
   (`12354–12355`, `7493–7494`) are the right length. Consider trimming the prose to the
   *what*, leaving the *why* in the PR body.

2. **Marginally redundant `path.trimTo(plen)` at `TableWriter.java:12370`** *(in-diff)* —
   for the `.pk`-present branch, line 12376 (`configureFollowerAndWriter(path.trimTo(plen), …)`)
   trims again with no intervening mutation. The code is **correct**; this is a cosmetic
   redundancy at most. ⚠️ **Trap to avoid:** the obvious "simplification" of dropping 12376's
   trim and passing bare `path` is **unsafe** — 12376's trim is load-bearing because
   `writer.of()` leaves `path` dirty (it appends the `.pk` name), so a populated column
   following any processed column relies on 12376 to reset the buffer. If you touch this at
   all, remove 12370 (which only cleans before the `continue`), never 12376. Leaving it as-is
   is fine.

3. **Defense-in-depth test gap: the self-heal path is untested** *(test)* — no test does a
   *subsequent in-order append with a non-null value* into the skipped keyless row-less
   column. It would pass trivially today (the mandatory `openPartition` in `finishO3Commit`
   re-creates the `.pk` and re-points the indexer), but the skip's safety depends entirely on
   the `restore`→`openPartition` ordering at `TableWriter.java:6612`/`6635`. A regression test
   that appends-then-reads after the no-`.pk` seal would pin that ordering against a future
   refactor. Optional, but the invariant is subtle enough to be worth locking in.

4. **Branch commit-message hygiene** *(commits, squash-mitigated)* — commits 1 & 3 carry a
   `fix(core):` Conventional-Commits prefix (only the *PR* title should) and exceed 50 chars;
   all four have empty bodies; 2 & 4 ("more fixes", "Address minor feedbacks") are
   non-descriptive. CLAUDE.md treats branch history as throwaway and the squash will take the
   compliant PR title — so this is informational. Just ensure the final squash body carries the
   real rationale.

5. **Pre-existing double `getRecordIndex` per column** *(optional, not introduced here)* —
   both methods resolve the same `(partition, col)` record twice (`getColumnTop` then
   `getColumnNameTxn`), each an O(log N)+linear-scan lookup; worst-case quadratic in
   last-partition column-version records. This is the established file-wide idiom
   (`copyOrRebuildColumnIndexes` does the identical double-resolve) and the PR neither
   introduces nor worsens it. Both methods are cold paths (once per O3 seal / parquet switch).
   A single-lookup fuse via the existing public API (`getRecordIndex` +
   `getColumnTopByIndexOrDefault` + `getColumnNameTxnByIndex`) is a clean opportunistic win if
   you're in the neighborhood — not a blocker.

## Downgraded (false positives / verified benign)

- **Stranded no-`.pk` indexer → silent rollback failure** (Agent 4). *Benign.* `finishO3Commit`
  runs `restore` (6612) then `openPartition` (6635) whenever the last partition was closed —
  and the keyless-row-less state only arises from a split/mutate, which always calls
  `closeActivePartition`. `openPartition` → `createIndexFiles` re-creates the absent `.pk`
  (verified: the `allowDestructiveRecovery=false` posting branch falls through and creates it)
  and `configureFollowerAndWriter` re-points the indexer. The squash caller is symmetric
  (`openLastPartition` precedes restore). The skip never strands an indexer that later gets
  indexed. Verified at 6612/6635, createIndexFiles, and corroborated by Agent 1's F6 + Agent
  5's green-on-HEAD run.
- **`partitionSize == -1` makes the guard skip everything** (Agents 1/5). *Unreachable.* Fix 1's
  early `currentNameTxn == MIN_VALUE` return and Fix 2's `partitionIndex < 0` throw +
  dir-exists check both share the same `findAttachedPartitionRawIndexByLoTimestamp` lookup as
  `getPartitionRowCountByTimestamp`, so the partition is provably attached (size ≥ 0) before
  the guard runs.
- **TOCTOU on `.pk` between `ff.exists` and `writer.of`** (Agent 2). *Not a bug.* Straight-line
  code on the single writer thread (exclusive `WriterPool` lock); O3 workers joined
  (`o3PartitionUpdRemaining == 0`); column/partition purge jobs are scoreboard-gated and only
  target superseded name-txns, never the live current version.
- **Re-point leaks the prior writer / double-opens** (briefing lead). *Not a bug.*
  `PostingIndexWriter.of(isInit=false)` calls `close()` as its first statement
  (PostingIndexWriter.java:1074); `configureFollowerAndWriter` closes on throw. The borrowed
  `.d` fd is set to `-1` without `ff.close`, so it never double-closes.
- **`boolean keyFileExists` violates the is/has naming rule** (Agent 6). *Not a finding.* The
  identical name is the established in-file precedent at `TableWriter.java:2957` for the same
  `ff.exists(keyFileName(...))` rebind idiom Fix 1 is modeled on; renaming here would diverge
  from its sibling. Local consistency wins.
- **Discrimination checks the wrong file** (correctness lead). *Verified sound.*
  `IndexFactory.keyFileName(POSTING, …)` delegates to `PostingIndexUtils.keyFileName`, the
  exact name `writer.of()` opens, with the same `colNameTxn` — a stale `.pk` carries a
  different txn and correctly fails to match.
- **Missing tests for `partitionSize == 0` and the live `columnTop == -1` arm** (Agent 5).
  *Defensive-only.* Both are unreachable states (active partition holds the max-ts row; a
  last-partition column's add-partition is always ≤ it), so the absent arm is the documented
  "normally unreachable" defensive guard.

## Summary

**Verdict: Approve.** This is a correct, well-scoped, well-tested fix. No Critical or Moderate
issues; the five Minor items are comment length, a cosmetic redundant trim, an optional
defense-in-depth test, squash-throwaway commit hygiene, and a pre-existing perf idiom — none
blocking.

- **Verification:** 8 agents launched (correctness, concurrency, performance, resources, tests,
  code quality, PR metadata, adversarial performance); ~13 draft findings → **5 Minor verified,
  8 downgraded** as false-positive/benign after source verification. Correctness, concurrency,
  resources, and performance came back clean. Agent 5 independently **reverted to master and
  confirmed 3 of the 4 tests are genuinely red-on-master** (the 4th is a deliberate forward
  guard, as the PR states) — strong evidence the regression coverage is real.
- **In-diff vs out-of-diff:** all 5 findings are in-diff or branch metadata; **out-of-diff
  findings: 0.** This is *not* an underrun — the cross-context pass was substantial and
  load-bearing: the `openPartition`/`createIndexFiles` self-heal (8804–8841, 5260–5298), the
  `rollbackIndexes`/`rollbackValues` no-op guards, the three sibling guards
  (`copyOrRebuildColumnIndexes`, `sealPostingIndexForPartition`), and the sole public entry
  point (`switchNativePartitionWithParquet`) were all traced and confirmed **safe**. The most
  important out-of-diff result is positive: `openPartition`'s `createIndexFiles` is what makes
  the no-`.pk` skip safe, and it holds.
- **Tradeoffs (level-headed):** Fix 1 adds one `stat(2)` per row-less posting column on a cold
  restore path and one hoisted partition-size lookup per call — negligible, and correctly the
  *only* way to distinguish the two row-less sub-states (column-version metadata cannot). Fix 2
  is a pure additional skip that strictly reduces I/O. No data migration, no format change.
  Fix 2's only OSS reach is the public `switchNativePartitionWithParquet` API (no internal
  caller), so within OSS it is lower-severity than Fix 1's WAL-suspend path — which the PR body
  already states plainly.

bluestreak01 (approved):
Approve. Reviewed at level 3 (full pass: correctness, concurrency, performance, resources, tests, cross-context, and adversarial agents, plus per-finding source verification).

No Critical or Moderate code-correctness issues. Both fixes align with proven sibling guards (copyOrRebuildColumnIndexes, sealPostingIndexForPartition):
- Fix 1 (restorePostingIndexersToLastPartition): discriminates row-less columns on the .pk's actual presence — skips the no-.pk split-tail case, re-points the .pk-present ADD COLUMN case. Verified the no-.pk skip is self-healed within the same finishO3Commit: a last-partition split calls closeActivePartition, so isLastPartitionClosed() is true and openPartition(L) re-creates the .pk and re-points the indexer before any subsequent append — no stranding, no silent loss.
- Fix 2 (linkPartitionIndexFiles): brings the parquet-switch guard in line with copyOrRebuildColumnIndexes (skip absent or row-less). Benign: a row-less column carries an empty index and the path preserves column tops.

Verified safe: getColumnTopQuick->getColumnTop swap (identical for every column present on L, strictly safer for the absent case), partitionSize >= 0 guaranteed on both paths, no TOCTOU (single writer thread, purge jobs target only superseded versions), no writer leak on re-point or error paths.

Out-of-diff: 0 bugs — both restore callers (O3 seal, squash) and the public switchNativePartitionWithParquet entry traced and confirmed SAFE.

Non-blocking follow-ups (optional): (1) a regression test appending-then-reading into a row-less keyless column after the no-.pk skip to pin the restore->openPartition ordering; (2) coverage for Fix 2's row-less-with-.pk-on-historic-then-parquet sub-case.

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|12|12|100.00%|

PR #7333, opened 2026-06-26
fix(sql): LATEST ON with an indexed key and a residual filter could return the wrong row

nwoolmer:
## Problem

A query of the form

```sql
SELECT ... FROM t
WHERE sym = 'X' AND venue = 'Y' AND ts <= '...'
LATEST ON ts PARTITION BY sym;
```

where `sym` is an indexed `SYMBOL` and `venue = 'Y'` is a residual filter, could return a row belonging to a **different** `sym` value. The selected row was internally consistent (all columns from one real row) — just the wrong one.

## Root cause

The index-backed *filtered* `LATEST ON` cursors subtracted the page frame's `partitionLo` from the row id a **second** time:

```java
recordA.setRowIndex(cursor.next() - partitionLo); // double subtraction
```

`BitmapIndexBwdReader`/`BitmapIndexFwdReader` cursors already return a frame-relative row id (`result - minValue`, where `minValue == partitionLo`). The extra subtraction positioned the record `partitionLo` rows too early whenever the matched row fell in a page frame with `partitionLo > 0` — i.e. a partition large enough to span multiple page frames. With small/single-frame partitions `partitionLo == 0`, so the bug was invisible in tests and on small tables, and only surfaced on busy real tables.

The non-filtered variants (`LatestByValueIndexedRowCursorFactory`, `LatestByValuesIndexedRecordCursor`) and `PageFrameRecordCursorImpl` already use `cursor.next()` directly, which is the correct convention.

## Fix

Remove the redundant `- partitionLo` in:
- `LatestByValueIndexedFilteredRecordCursor` (single key value)
- `LatestByValuesIndexedFilteredRecordCursor` (`IN` list; two sites — `setRowIndex` and `Rows.toRowID`)

No stored data is affected — the index is correct; only query-time row positioning was wrong.

## Tests

Adds two regression tests to `LatestByTest` (run for MICRO and NANO timestamps) that force small page frames via `changePageFrameSizes(1, 8)` so the single partition splits and the matched row lands in a frame with `partitionLo > 0`. They fail without the fix and pass with it. `LatestByTest` 120/0 and `ParallelLatestByTest` 16/0 green.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/table/LatestByValueIndexedFilteredRecordCursor.java
+++ b/core/src/main/java/io/questdb/griffin/engine/table/LatestByValueIndexedFilteredRecordCursor.java
@@ -121,7 +121,11 @@ class LatestByValueIndexedFilteredRecordCursor extends AbstractLatestByValueReco
 
             try (RowCursor cursor = indexReader.getCursor(symbolKey, partitionLo, partitionHi)) {
                 while (cursor.hasNext()) {
-                    recordA.setRowIndex(cursor.next() - partitionLo);
+                    // cursor.next() already returns a frame-relative row index (BitmapIndex*Reader
+                    // subtracts minValue == partitionLo). Subtracting partitionLo again here positioned
+                    // the record partitionLo rows too early whenever the match fell in a page frame with
+                    // partitionLo > 0, returning a neighbouring row (often a different symbol).
+                    recordA.setRowIndex(cursor.next());
                     if (filter.getBool(recordA)) {
                         isRecordFound = true;
                         return;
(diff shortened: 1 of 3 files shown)


nwoolmer:
Auto-closed by a branch rename — continued in #7334 (same fix plus deferred and column-top regression tests).

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByValuesIndexedFilteredRecordCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByValueIndexedFilteredRecordCursor.java|1|1|100.00%|

PR #7334, opened 2026-06-26
fix(sql): fix LATEST ON returning the wrong row with an indexed key and residual filter

nwoolmer:
_Supersedes #7333 (auto-closed by a branch rename; same content + extra tests)._

This branch carries two related fixes to the index-backed `LATEST ON` cursors: a row-positioning correctness bug, and a separate early-exit regression for duplicate keys.

### Problem

```sql
SELECT ... FROM t
WHERE sym = 'X' AND venue = 'Y' AND ts <= '...'
LATEST ON ts PARTITION BY sym;
```

where `sym` is an indexed `SYMBOL` and `venue = 'Y'` is a residual filter, could return a row belonging to a **different** `sym` value. The returned row was internally consistent (one real row) — just the wrong one.

### Root cause

The index-backed *filtered* `LATEST ON` cursors subtracted the page frame's `partitionLo` from the row id a **second** time:

```java
recordA.setRowIndex(cursor.next() - partitionLo); // double subtraction
```

`IndexReader.getCursor(key, minValue, maxValue)` already returns a row id relative to `minValue` (its javadoc: "index value cursor, relative to the minValue"), and `minValue == partitionLo` at every call site (the `BitmapIndexBwdReader`/`BitmapIndexFwdReader` cursors compute `result - minValue`). The extra subtraction positioned the record `partitionLo` rows too early whenever the matched row fell in a page frame with `partitionLo > 0` — i.e. a partition large enough to span multiple page frames. With single-frame partitions `partitionLo == 0`, so it was invisible in tests and on small tables, and only surfaced on busy real tables.

The non-filtered variants (`LatestByValueIndexedRowCursorFactory`, `LatestByValuesIndexedRecordCursor`) and `PageFrameRecordCursorImpl` already use `cursor.next()` directly.

### Change

Remove the redundant `- partitionLo` in:
- `LatestByValueIndexedFilteredRecordCursor` (single key value);
- `LatestByValuesIndexedFilteredRecordCursor` (`IN` list; both `setRowIndex` and `Rows.toRowID`).

The deferred variant (`LatestByValueDeferredIndexedFilteredRecordCursorFactory`) delegates to the same cursor, so it is fixed too. No stored data is affected — the index is correct; only query-time row positioning was wrong.

### Problem

For a query that resolves the same key both statically and as a runtime constant — e.g. `sym IN ('a', :sym)` with `:sym = 'a'` — the multi-value indexed cursors stopped using their early exit and full-scanned every remaining frame (opening older partitions) even after the only unique key had already been found in the latest frame.

### Root cause

The cursors stop once `found.size() >= keyCount`, where `keyCount = symbolKeys.size() + deferredSymbolKeys.size()`. A deferred key that duplicated a static key was counted twice in `keyCount`, but `found` is a set, so `found.size()` could never reach the inflated `keyCount` — the early exit never fired.

### Change

- `AbstractDeferredTreeSetRecordCursorFactory.initRecordCursor()` now adds a resolved deferred key only when it is not already present in `symbolKeys`, so the two key sets are disjoint and `keyCount` is exact.
- The now-redundant per-frame `!symbolKeys.contains(symbolKey)` guard is removed from the deferred loop in both `LatestByValuesIndexedFilteredRecordCursor` and `LatestByValuesIndexedRecordCursor` (`addFoundKey` is already idempotent via `found.keyIndex()`).
- The frame loop condition is reordered to `found.size() < keyCount && (frame = frameCursor.next()) != null`, so once all keys are found the cursor short-circuits before pulling — and opening — the next (older) page frame.

## Tests

Adds regression tests to `LatestByTest` (MICRO + NANO).

For Fix 1, force small page frames via `changePageFrameSizes(1, 8)` so the partition splits and the matched row lands in a frame with `partitionLo > 0`:
- single-value filtered cursor;
- `IN`-list filtered cursor;
- subquery (`IN (SELECT ...)`) filtered cursor;
- deferred path (bind-variable key);
- a column with `columnTop > 0`;
- `IN`-list with a not-found / `NULL` element (filtered and non-filtered).

For Fix 2, inject a `FilesFacade` whose `openRO` fails for the older partition, proving the cursor never opens it once the only unique key has been found in the latest partition (single and filtered variants).

All fail without the corresponding fix and pass with it. `LatestByTest` 124/0, `ParallelLatestByTest` 16/0.


jovfer:
## Codex PR review

### Moderate

1. Missing split-frame regression for the subquery-backed path.

`WHERE sym IN (SELECT ...)` reaches the changed multi-value cursor through:

`SqlCodeGenerator.java:6571` -> `LatestBySubQueryRecordCursorFactory.java:81` -> `LatestByValuesIndexedFilteredRecordCursor.java:137`

The PR tests literal `IN`, single deferred key, and column-top cases, but the existing subquery tests use default page frames and do not reproduce `partitionLo > 0`. Please add a sibling regression using `changePageFrameSizes(1, 8)` with a query such as `sym IN (SELECT 'g1' UNION ALL SELECT 'g2')` plus the residual filter.

2. Duplicate literal/deferred keys disable early exit in the multi-value cursor.

In `core/src/main/java/io/questdb/griffin/engine/table/LatestByValuesIndexedFilteredRecordCursor.java:158`, `keyCount` includes all deferred keys, but duplicates already present in `symbolKeys` are skipped later at line 186. For a query like `sym IN ('a', :sym)` with `:sym = 'a'`, `found.size()` can never reach `keyCount`, so the cursor scans every remaining frame instead of stopping once the unique key is found.

Suggested fix: dedupe static and deferred keys before computing `keyCount`, or build one combined unique key collection for both iteration and early-exit accounting.

### Minor

1. PR metadata needs cleanup.

The PR has no labels despite SQL bug scope. Please add at least `SQL` and `Bug`. The title should also repeat the verb after `fix(sql):`, and the generated-attribution footer should be removed from the public PR body.

2. New public tests are placed after private helper methods.

`core/src/test/java/io/questdb/test/griffin/LatestByTest.java:1589`

Move the new public `@Test` methods above the private helper block to preserve the class member ordering convention.

3. Added SQL literals should use underscore grouping.

The new tests use `1000000` instead of `1_000_000` at:

- `LatestByTest.java:1602`
- `LatestByTest.java:1632`
- `LatestByTest.java:1668`
- `LatestByTest.java:1702`
- `LatestByTest.java:1711`

4. The new comments describe the contract too narrowly.

`LatestByValueIndexedFilteredRecordCursor.java:124` and `LatestByValuesIndexedFilteredRecordCursor.java:142` mention `BitmapIndex*Reader`, but these cursors consume generic `IndexReader` implementations, including posting indexes. Please reference the `IndexReader.getCursor()` contract instead: returned row ids are relative to `minValue`.

jovfer:
Reviewing PR #7334 at level 3.

### Critical

None.

### Moderate

1. **[in-diff] Multi-key indexed cursors still probe keys that were already found.**

`LatestByValuesIndexedRecordCursor.buildTreeMap()` and `LatestByValuesIndexedFilteredRecordCursor.buildTreeMap()` iterate the full `symbolKeys` / `deferredSymbolKeys` sets on every frame, then rely on `addFoundKey()` to discard already-found keys with `found.keyIndex()`.

Code path: `SqlCodeGenerator` creates these cursors for indexed `LATEST ON ... PARTITION BY sym` with `IN (...)` / subquery keys. With a large key list where one key is late or missing, the cursor keeps paying `O(frames * totalKeys)` hash probes instead of iterating only unresolved keys.

Suggested fix: keep a reusable remaining-key list and remove/swap keys as soon as they are found. The `symbolKeys.contains(symbolKey)` guard in the deferred loop is also redundant now that deferred keys are deduped during `initRecordCursor()`.

Files:
- `core/src/main/java/io/questdb/griffin/engine/table/LatestByValuesIndexedRecordCursor.java:160`
- `core/src/main/java/io/questdb/griffin/engine/table/LatestByValuesIndexedFilteredRecordCursor.java:179`
- `core/src/main/java/io/questdb/griffin/engine/table/LatestByValuesIndexedFilteredRecordCursor.java:186`

2. **[metadata] PR metadata is stale for the current branch.**

The title/body still describe only the `partitionLo` row-positioning fix, but the branch now also includes the duplicate static/deferred key early-exit fix and two related regressions. The title also violates the QuestDB title convention because the description after `fix(sql):` does not repeat the verb.

Suggested fix: update the title/body to cover both fixes, remove the generated attribution footer, and clean up commit messages if the branch is not squash-merged.

### Minor

1. New `@Test` methods are appended after private helper methods, breaking class member ordering.

`core/src/test/java/io/questdb/test/griffin/LatestByTest.java:1588`

2. Added SQL test strings use lowercase SQL keywords and several ungrouped `1000000` literals.

Files:
- `core/src/test/java/io/questdb/test/griffin/LatestByTest.java:1594`
- `core/src/test/java/io/questdb/test/griffin/LatestByTest.java:1602`
- `core/src/test/java/io/questdb/test/griffin/LatestByTest.java:1632`
- `core/src/test/java/io/questdb/test/griffin/LatestByTest.java:1668`
- `core/src/test/java/io/questdb/test/griffin/LatestByTest.java:1702`
- `core/src/test/java/io/questdb/test/griffin/LatestByTest.java:1711`

3. The new production comments describe the contract too narrowly.

The code consumes generic `IndexReader` cursors, but the comments name `BitmapIndex*Reader`. Reference the `IndexReader.getCursor()` contract instead: returned row ids are relative to `minValue`.

Files:
- `core/src/main/java/io/questdb/griffin/engine/table/LatestByValueIndexedFilteredRecordCursor.java:124`
- `core/src/main/java/io/questdb/griffin/engine/table/LatestByValuesIndexedFilteredRecordCursor.java:142`

### Downgraded

- Missing subquery and duplicate-deferred-key regressions: dismissed; the current diff adds both.
- Duplicate static/deferred key early exit still broken: dismissed; deferred keys are deduped against static keys and the frame loop now checks `found.size() < keyCount` before advancing.
- Row-id fix could break callers: dismissed; `IndexReader.getCursor()` returns row ids relative to `minValue`, and the inspected callsites preserve that contract.
- SQL assertion builder violations: dismissed; no new `assertSql`, `getPlan`, or deterministic `returnsOnce()` usage was added.

### Summary

Verdict: needs discussion. No critical correctness, concurrency, or resource-management issues found. Verified findings: 5. False positives removed: 4. Split: 4 in-diff findings, 1 metadata finding.

nwoolmer:
## Level-3 review

**Verdict: correct, complete, and well-tested — approve.** Both fixes are sound, the root-cause analyses hold up against the sources, and the fix is complete across the whole indexed `LATEST ON` cursor family (no other latent instances repo-wide). I re-ran the suites and addressed the two suggestions in `26b60d5964`.

### Fix 1 — verified correct
Confirmed the invariant the fix relies on:
- `BitmapIndexBwdReader.Cursor.next()` returns `next - minValue` (`idx/BitmapIndexBwdReader.java:138`), and every call site passes `minValue == partitionLo`. So the cursor result is **already frame-relative** — the second subtraction was the bug.
- `NullCursor extends Cursor` and inherits the same `next - minValue`, so the **null bucket is frame-relative too** — the `IN (..., null)` path is correct.
- The fix **aligns the filtered variants with the already-correct non-filtered ones**: `LatestByValueIndexedRowCursorFactory:52` (`indexReaderCursor.next()` directly) and `LatestByValuesIndexedRecordCursor` (`Rows.toRowID(frameIndex, cursor.next())` directly).

**Completeness:** I swept all of `core/src/main` for the antipattern (any `getCursor(...)` result with a subtraction: `setRowIndex(... -`, `toRowID(... -`, `next() -`, `- partitionLo`, `- minValue`). The only `partitionHi - partitionLo` hits are the non-indexed full-scan cursors that compute a frame-relative row *count* and feed `setRowIndex(row)` with an already-0-based index — those are correct. **No other instance of the double-subtraction bug exists.** The four-way matrix (single/multi × filtered/non-filtered) is now uniformly correct.

### Fix 2 — verified correct
- Root cause confirmed: a deferred key duplicating a static key was counted twice in `keyCount` while `found` is a set, so `found.size()` could never reach the inflated `keyCount` and the early exit never fired.
- The factory-level dedup runs **before** `cursor.of()` resets `keyCount = -1` and before `buildTreeMap()` caches it — ordering is correct.
- **Both creators of the now-unguarded cursors are safe:** the only subclass of the deduping abstract factory is `LatestByValuesIndexedFilteredRecordCursorFactory` (creates both the filtered and non-filtered multi-value cursors with deduped keys). The other creator, `LatestBySubQueryRecordCursorFactory`, passes `deferredSymbolKeys = null`, so the `if (deferredSymbolKeys != null)` block is skipped and dropping the per-frame guard is a no-op there.
- The loop reorder `found.size() < keyCount && (frame = frameCursor.next()) != null` correctly short-circuits before pulling/opening the next (older) frame — independently proven by the `openRO`-fails-on-`1970-01-01` negative-control test.
- Even if the invariant were ever violated, `addFoundKey` is idempotent via `found.keyIndex()`, so the worst case is a perf regression, never wrong results.

### Suggestions — addressed in `26b60d5964`
1. **Enforce the dedup invariant.** Since a violation degrades silently to a full scan (no test would catch it), added `assert keysDisjoint(symbolKeys, deferredSymbolKeys)` where `keyCount` is computed in both multi-value cursors. Assertions are `-ea` in the suite, so this is now actively exercised.
2. **Dedup the test facade.** Extracted the two identical `openRO` `FilesFacade`s into a shared `failOpenForPartition(...)` helper.

### Tests — re-run green (assertions enabled)
- `LatestByTest`: **136 run / 0 failures / 0 errors** (4 pre-existing `Assume` skips). All 10 new regression methods ran in both MICRO and NANO parameterizations.
- `ParallelLatestByTest`: **16 / 0**.

### Risk
Low. Changes are localized to query-time row positioning and loop control in the indexed cursors; no stored-data/format impact (the index itself was always correct). The fix narrows behavior (correct row, fewer partitions opened) rather than broadening it.

<sub>🤖 review assisted by Claude Code</sub>

nwoolmer:
## Level-3 review (refresh — full diff @ `0e61fd79`, rebased on latest master)

Re-reviewed the complete change set end-to-end after merging the latest `master`. **Verdict: correct, complete, well-tested — ready to merge.**

### Master merge (`0e61fd79`)
Merged `origin/master`, which advanced by one commit: `f8cf9e468e` — *fix(core): fix POSTING-indexed tables suspending WAL apply after out-of-order writes (#7332)*. That commit touches only `TableWriter.java`, `PostingIndexCriticalIssuesTest.java`, and `TableWriterTest.java` — **disjoint from this PR's files**, so the merge is clean (no conflicts, no semantic interaction with the `LATEST ON` cursors). Re-ran the suites on the merged tree: **`LatestByTest` 136/0/0** (4 pre-existing `Assume` skips), **`ParallelLatestByTest` 16/0**.

### Fix 1 — wrong row with indexed key + residual filter — correct
- The fix removes a second `- partitionLo`. Verified the invariant it relies on: `BitmapIndexBwdReader.Cursor.next()` returns `next - minValue` (`idx/BitmapIndexBwdReader.java:138`) and every call site passes `minValue == partitionLo`, so the cursor result is **already frame-relative**. `NullCursor` inherits the same `next()`, so the null bucket is frame-relative too.
- Brings the filtered cursors in line with the already-correct non-filtered ones (`LatestByValueIndexedRowCursorFactory:52`, `LatestByValuesIndexedRecordCursor` addFoundKey).
- **Completeness:** swept all of `core/src/main` for the antipattern (any `getCursor(...)` result with a subtraction). The only `partitionHi - partitionLo` hits are non-indexed full-scan cursors computing a frame-relative row count — correct. No other instance exists; the single/multi × filtered/non-filtered matrix is now uniformly correct.

### Fix 2 — duplicate static/deferred key defeats early exit — correct
- Root cause confirmed: a deferred key duplicating a static key inflated `keyCount` while `found` is a set, so `found.size()` could never reach `keyCount` and the early exit never fired (full scan of older partitions).
- Factory-level dedup runs before `cursor.of()` resets `keyCount = -1`, so `keyCount` is computed from the deduped set. Both creators of the now-unguarded cursors are safe: the deduping `LatestByValuesIndexedFilteredRecordCursorFactory` (both cursor types) and `LatestBySubQueryRecordCursorFactory` (passes `deferredSymbolKeys = null`).
- Loop reorder `found.size() < keyCount && (frame = frameCursor.next()) != null` short-circuits before opening the next older frame — proven by the `openRO`-fails-on-`1970-01-01` negative-control test.
- `addFoundKey` is idempotent via `found.keyIndex()`, so even a violated invariant degrades to a perf regression, never wrong results.

### Hardening (commit `26b60d59`, reviewed here)
- `assert keysDisjoint(symbolKeys, deferredSymbolKeys)` added where `keyCount` is computed in both multi-value cursors — makes the dedup invariant explicit and catches a future direct caller. Assertions are `-ea` in the suite, so this is actively exercised by the duplicate-key tests (and confirmed not to trip across the full suite).
- The two Fix-2 tests' duplicated `openRO` `FilesFacade` extracted into a shared `failOpenForPartition(...)` helper.

### Tests
10 new regression methods, each parameterized MICRO + NANO. Coverage: single/multi/deferred/subquery/columnTop/not-found/null for Fix 1; single + filtered duplicate-deferred-key early-exit for Fix 2. Good design touches — small page frames (`changePageFrameSizes(1, 8)`) to force `partitionLo > 0`, and an `openRO`-failing facade as a clean negative control. All pass; each fails without its corresponding fix.

### Risk
Low. Query-time row positioning and loop control only; no stored-data/format impact (the index was always correct). The change narrows behavior — correct row, fewer partitions opened.

### CI
Full Azure matrix (Cairo A/B, Griffin, Other A/B across arm64 / x64-zfs / x86-graal), Rust/lint, coverage runs, and all GitHub builds were green on the pre-merge head; the merge only adds disjoint master code, re-verified locally.

<sub>🤖 review assisted by Claude Code</sub>

bluestreak01:
## Level-3 review

Independent full pass, verified against the sources (not a rehash of the earlier reviews).

### Change surface verified

Two logically distinct fixes:
- **Fix 1** — remove the double `- partitionLo` in the index-backed *filtered* `LATEST ON` cursors.
- **Fix 2** — dedup deferred keys against static keys so the multi-value early exit fires; reorder the frame loop to short-circuit before opening the next (older) frame.

Complete cursor-creator inventory, all safe:
- `LatestByValuesIndexedFilteredRecordCursorFactory` (extends the deduping `AbstractDeferredTreeSetRecordCursorFactory`, uses its inherited `initRecordCursor` -> deferred keys deduped).
- `LatestBySubQueryRecordCursorFactory` passes `deferredSymbolKeys = null` -> deferred loop skipped, `keysDisjoint(symbolKeys, null)` trivially true.
- `LatestByValueIndexedFilteredRecordCursorFactory` / `LatestByValueDeferredIndexedFilteredRecordCursorFactory` -> single-value cursor (Fix 1 only).

### Critical

None.

Fix 1 is correct. `IndexReader.getCursor(key, minValue, maxValue)` javadoc states the returned rows are "relative to the minValue", and `BitmapIndexBwdReader.Cursor.next()` returns `next - minValue` (`BitmapIndexBwdReader.java:138`) with `minValue == partitionLo` at every callsite. So `cursor.next()` is already 0-based frame-relative; the old `- partitionLo` produced `result - 2*partitionLo` — a negative local row id whenever `partitionLo > 0`, positioning the record into the previous frame's memory (a different symbol's row). The null-bucket path holds too: `NullCursor.hasNext()` sets `next = nullCount` (partition-relative) and inherits `next() = next - minValue`, so the implicit-columnTop-null branch is also frame-relative. No other instance of the pattern survives repo-wide (the row-cursor-factory variants already call `next()` directly). The four-way matrix is now uniformly correct.

Fix 2 is correct. A deferred key duplicating a static key inflated `keyCount` while `found` is a set, so `found.size()` never reached `keyCount`. Dedup at factory init runs before `buildTreeMap()` caches `keyCount`, and both the loop reorder and the dedup are needed for the negative-control test to pass. Even a violated invariant degrades to a full scan, never wrong rows (`addFoundKey` is idempotent via `found.keyIndex()`), so the added `assert keysDisjoint(...)` is a correct hardening.

### Moderate

**M1 — Multi-value cursors re-probe already-found keys on every frame.** `buildTreeMap()` in `LatestByValuesIndexedRecordCursor` and `LatestByValuesIndexedFilteredRecordCursor` iterates the full `symbolKeys` + `deferredSymbolKeys` sets per frame and relies on `found.keyIndex()` to skip already-resolved keys — `O(frames x totalKeys)` hash probes rather than `O(frames x remainingKeys)`. `found.keyIndex()` is O(1) and the key count is bounded by the IN-list size, but the work is genuinely redundant: once a key is resolved it is probed again on every subsequent (older) frame. A reusable remaining-key list (remove/swap keys as they are found, iterate only the unresolved remainder) collapses this to the minimal traversal. The structure predates this PR, but this PR is already reworking the exact same loops and key-set accounting, so it is the right place to fold in the fix rather than open a separate PR.

### Minor

1. New SQL test literals skip the underscore separator (CLAUDE.md requires `1_000_000` for 5+ digit numbers): bare `1000000` at `LatestByTest.java:1602, 1632, 1668, 1702, 1711, 1779, 1816, 1848`, while line 1742 correctly uses `1_000_000` — an inconsistency within the same PR.
2. Production comments name `BitmapIndex*Reader` but the cursors consume the generic `IndexReader` interface (`LatestByValueIndexedFilteredRecordCursor.java:124`, `LatestByValuesIndexedFilteredRecordCursor.java:142`, `LatestByValuesIndexedRecordCursor.java`). Reference the `IndexReader.getCursor()` contract ("rows relative to minValue") instead of the concrete class.
3. The `keysDisjoint` static helper sits among instance methods; CLAUDE.md groups members by kind (static vs instance).
4. Coverage micro-gap: the columnTop test uses a non-null key; the null-bucket + columnTop + `partitionLo > 0` positioning path is exercised only indirectly. A `sym IS NULL` variant over split frames with a columnTop would close it.
5. New test SQL uses lowercase keywords while several neighbors in the same file use UPPERCASE (CLAUDE.md permits mixed case, style-only).

### Downgraded (false positives / already addressed)

- Title verb repetition / labels — the current title repeats the verb and labels are present; metadata is clean.
- Missing subquery / duplicate-deferred-key regressions — both are in the current diff.
- Fix could break other `getCursor` callers — every other consumer already treats `next()` as frame-relative, so the fix aligns rather than diverges.
- Loop reorder changes behavior at `keyCount == 0` — verified benign (opens zero frames instead of one; identical empty result).
- Concurrent factory reuse race on `deferredSymbolKeys` — the clear+rebuild in `initRecordCursor` predates this PR; no new sharing introduced.
- SQL builder-API violations — all new tests use `.returns(...)`; the `assertMemoryLeak` wrappers hold multi-statement lambdas, so they are legitimate.

### Summary

Verdict: approve on correctness — both fixes are verified correct and the double-subtraction is fully eradicated across the indexed `LATEST ON` cursor family. No correctness, concurrency, or resource issues.

@author — please fold **M1** into this PR before merge: switch the two multi-value cursors' per-frame loops to iterate only the unresolved keys (a reusable remaining-key list) instead of re-probing every key on every frame. The minor items (especially the `1000000` -> `1_000_000` inconsistency) are worth cleaning up in the same push.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 64 / 66 (96.97%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByValuesIndexedFilteredRecordCursor.java|32|33|96.97%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByValuesIndexedRecordCursor.java|28|29|96.55%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AbstractDeferredTreeSetRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByValueIndexedFilteredRecordCursor.java|1|1|100.00%|

bluestreak01 (approved):
Level-3 review complete. Both fixes verified correct at the source level (IndexReader frame-relative contract; swap-remove early-exit). Empirically confirmed the regression tests fail on revert (10/10) and pass with the fix (10/10). No blocking findings.

PR #7261, opened 2026-06-14
chore(core): engine seams for hot in-place primary/replica role switch

jovfer:
Tandem: [questdb/questdb-enterprise#1024](https://github.com/questdb/questdb-enterprise/pull/1024)

Core-side (OSS) half of the **hot in-place primary/replica role switch**. This PR provides the engine
seams the QuestDB Enterprise replication layer uses to flip a running node between primary and replica
with **no JVM restart**; the enterprise orchestration, SQL/REST switch surfaces, replication transport,
and tests live in the tandem PR above.

## Merge order

This is the core half of **questdb/questdb-enterprise#1024**. The two must land in lockstep:

1. **Merge this OSS PR first.**
2. The enterprise PR then re-pins its `questdb` submodule to the squashed master SHA and merges.

Merging the enterprise PR before this one would leave its build pointing at an orphaned submodule commit.

## What this adds (core seams)

- A shared/exclusive **role-switch lock** on `CairoEngine` (`getRoleSwitchReadLock()` /
  `getRoleSwitchWriteLock()`) and an `isReadOnlyMode()` hook. These are no-ops / uncontended in a plain OSS
  deployment - the enterprise layer flips the read-only flag under the write lock when a node demotes.
- A **write fence at externalization.** Every write path re-checks the read-only flag under the role-switch
  read lock at the moment it externalizes, so a demote that lands mid-operation refuses-and-rolls-back (or
  never mints the seqTxn for) the unreplicated write rather than acking it:
  - `INSERT` / `INSERT...SELECT` commit (`InsertOperationImpl` / `InsertAsSelectOperationImpl`),
  - WAL `UPDATE` / `ALTER` via the shared `OperationDispatcher.applyFenced` chokepoint both HTTP `/exec`
    and pg-wire funnel through (the fix for an acked-write-loss gap - a WAL `UPDATE`/`ALTER` could
    previously mint a seqTxn after a demote and be silently superseded),
  - CTAS / DDL materialization (`JsonQueryProcessor` / `PGPipelineEntry` fenced execution).
- `ReadOnlyStatementGate` - the compile-time statement-class classification of what a read-only replica
  refuses, with the deliberate replica-admin allow-list (REPAIR/VACUUM/SNAPSHOT/RESUME) documented.
- Lifecycle / shutdown hardening used by the switch teardown: bounded boot-join and close rendezvous,
  shutdown-hook deregistration, and native-handle safety on the close path.

## Tested

[...]

PR #7335, opened 2026-06-26
fix(core): prevent a crash on shutdown during materialized view hydration

jovfer:
tandem: https://github.com/questdb/questdb-enterprise/pull/1091

## What

`CairoEngine.hydrateMatViewStateStore()` rebuilds the live materialized-view state store by scanning every view off disk, holding native `Path` / `BlockFileReader` / `WalEventReader` / `MemoryCMR` and writing into `matViewStateStore`. The loop had no `isClosing()` check. If `close()` (a SIGTERM / pod eviction) arrives while the loop runs, teardown can free `matViewStateStore` and the native pools while the loop is still reading and writing them — a use-after-free in the `panic=abort` Rust cdylib, which aborts the JVM.

This method is a public engine seam with no in-tree caller; the enterprise hot primary/replica role switch invokes it (a promote rebuilds and hydrates a real store before opening writes), in the same family as the seams in #7261. A shutdown landing inside that promote/hydrate window is the trigger.

## Change

Poll `isClosing()` once per view at the top of the hydration loop and throw a `CairoException` if the engine is closing. `signalClose()` sets `closing` before teardown frees the engine, so the check observes shutdown promptly and the caller unwinds the half-built store. The check is a single volatile read per view on a cold boot/promote path, so there is no steady-state cost. The poll lives in the loop body rather than inside `loadMatViewIntoStore`, whose `catch (Throwable)` would otherwise swallow it.

## Test plan

- The enterprise tandem PR adds `CloseDuringPromoteHydrateRaceTest`, which parks inside a promote hydrate, lands a concurrent `close()`, and asserts the cascade observes `closing` and unwinds: RED without this poll (the loop runs to completion and the promote advances to PRIMARY), GREEN with it (the node stays at its prior role, `close()` returns cleanly, native-memory delta bounded).
- Materialized-view and role-switch suites pass on a clean build.


jovfer:
# Review — PR #1091 (ENT) + #7335 (OSS), tandem, level 3 (round 3, re-run)


**Verdict: ENT approve, OSS approve.** No Critical, no P0 out-of-diff bug. The close-vs-cascade
serialization is correct and every changed contract is handled at every production callsite. Findings:
1 disclosed by-design residual (with a cheap hardening), 1 worthwhile test-assertion gap, and a set of
comment/doc/naming nits.

10 dimensions run; ~24 draft observations → **2 Moderate + 8 Minor confirmed, 3 pre-existing noted,
~7 dropped** as false-positive/benign. All confirmed findings are in-diff or about-the-diff; the
cross-context pass cleared every out-of-diff callsite (Contracts A/B/C) as SAFE — a genuine clear, not
an under-run, because each changed symbol has a tiny fully-enumerated caller set.

---

## The fix, verified end to end

- `ServerMain.close():205` runs `engine.signalClose()` (sets `volatile closing=true`) **before**
  `freeOnExit.close():225`. FreeOnExit is LIFO; the engine is registered first (`:102`) so it is freed
  **last**, after the orchestrator's `close()` → `awaitInFlightWork()` rendezvous and the reverse-topo
  stop loop. So in the happy path the cascade unwinds before the engine's native free.
- The cascade observes `closing` at the per-view poll (`CairoEngine.java:580`) → throws
  `CairoException.nonCritical()` (a `RuntimeException`); `switchMatViewMachinery`'s `catch(Throwable)`
  (`EntCairoEngine.java:2020`) unwinds (delegate→NoOp `:2021`, queue→NoOpQueue `:2028`,
  `Misc.free(realStore)` `:2029`, clears `matViewRefreshSuspended` `:2032`) and rethrows → the
  orchestrator's per-component `catch(Throwable)` (`EntLifecycleOrchestrator.java:364`) → DEGRADED /
  role UNKNOWN; `completedRole` stays at the prior role; a same-direction retry heals.
- The new ACL/heal `throwIfClosing` guards throw `IllegalStateException` on the same path.
- **OSS poll: exactly one production caller** (`EntCairoEngine.switchMatViewMachinery(PRIMARY)`,
  `:2019`), inside the unwinding try/catch; no OSS production caller; `MatViewTest.java:3461` runs on a
  non-closing engine. `CairoException` is already imported (52 uses) — no new import. So the OSS change
  cannot break any OSS path.
- **M1**: `awaitInFlightWork` now bounds the rendezvous to the in-flight per-request budget
  (`closeRendezvousBudgetMs` → `activeRoleSwitchTimeoutMs`, clamped `[30s,600s]`) instead of the
  configured default — a fix, not a regression; the no-switch-in-flight path (`:436`) is byte-identical.

---

## Critical

None.

---

## Moderate

### M-1 — [ENT/OSS, in-diff, disclosed residual] budget-expiry still yields a genuine UAF + double-free; `MatViewStateStoreImpl.close()` is not idempotent

This is the residual the PR body's Tradeoffs already disclose ("C1 narrows, does not eliminate").
Reported as Moderate because it is the one point where the guarantee is probabilistic, a cheap
hardening exists, and the body under-states the scope.

If `awaitInFlightWork()`'s budget expires while the cascade is still inside a single
`loadMatViewIntoStore` (or a single ACL reload call), `LifecycleOrchestrator.close():157-162` logs and
proceeds; FreeOnExit then runs `CairoEngine.close():674` → `Misc.free(matViewStateStore)` while the
cascade is live. The wrapper delegate at that instant is the half-built `realStore` (swapped in at
`EntCairoEngine.java:2010`, **before** hydrate at `:2019`), so:
1. close-thread frees `realStore`'s `MatViewState`s via `wrapper.close()` → `realStore.close()`;
2. the cascade is still reading/writing those handles in the hydrate loop → **native UAF**;
3. the cascade's unwind then runs `Misc.free(realStore)` `:2029` → **double-free**.

`MatViewStateStoreImpl.close()` (`questdb/core/.../mv/MatViewStateStoreImpl.java:130-134`) iterates
`stateByTableDirName.values()` and `Misc.free(state)` each but **never clears the map**, so the second
free re-visits every state (the `@TestOnly clear():122` clears *after* close; production `close()` does
not). The per-view poll bounds the abort to one view-load, so this needs a single view-load — or a
single `reloadAccessLists`/`loadSecrets` call — to exceed the remaining budget (≥30s). Very unlikely;
**this PR strictly shrinks the window**; not a regression, not blocking.

Suggestions (optional, non-blocking):
- Make `MatViewStateStoreImpl.close()` idempotent (clear/null the map after freeing) — converts step 3
  from a double-free into a safe no-op; defensible as a standalone hardening.
- Broaden the body's residual from "a single pathological **view-load**" to also name the unpolled
  `reloadAccessLists`/`loadSecrets` calls (`EntCairoEngine.java:1317-1323`, `:1927-1933`) — same class.

### M-2 — [ENT, in-diff, test] hydrate witness asserts only `completedRole`, not the post-abort unwind invariants

`CloseDuringPromoteHydrateRaceTest` is now honestly self-labeled (the relabel addressed the prior
round) and `completedRole == REPLICA` (`:170-172`) is a faithful RED/GREEN discriminator for the OSS
poll (verified: removing the poll lets `super.hydrateMatViewStateStore()` complete and the cascade
reach `completedRole = PRIMARY` at `EntCairoEngine.java:1368`). But it asserts **only** that. It does
not assert the other two invariants the unwind establishes:
- `engine.isMatViewRefreshSuspended() == false` (cleared at `:2032`; public accessor at `:721` — one
  cheap line);
- the store/queue delegates were restored to NoOp (`:2021`, `:2028`).

A regression dropping the `matViewRefreshSuspended = false` clear or the `NoOpQueue` restore would leave
mat-view refresh **permanently wedged** after an aborted promote, yet this test stays GREEN. Add both.
(The `closeError == null` and `memDelta < 1MB` checks are self-labeled secondary sanity and do not
discriminate the poll — fine to keep, but they are not the witness.)

---

## Minor

### m-1 — [ENT, in-diff, comment] heal-path comment over-generalizes "the C1 poll never runs"
`EntCairoEngine.java:1919-1924` (and the heal test's prose) justify F1 by asserting
`switchMatViewMachinery(PRIMARY)` "early-returns (store already primary-shaped), so the C1 poll never
runs." True only for an **already-primary** store (the clean heal the test exercises). Reachable
counterexample: a demote that swapped the store to NoOp at `switchMatViewMachinery(REPLICA)` (`:1252`)
then aborted at `throwIfClosing` (`:1268`) leaves `completedRole == PRIMARY` with a NoOp store; a
same-direction promote heal then routes to `republishRoleState(PRIMARY)` → `switchMatViewMachinery
(PRIMARY)` where the `instanceof MatViewStateStoreImpl` early-return (`:1990`) does **not** fire and the
heal **does** hydrate (with C1 active). Strictly safer (not a hole), but scope the comment/test
rationale to the already-primary heal.

### m-2 — [ENT, in-diff, comment] `CLOSE_RENDEZVOUS_CEILING_MS` comment is stale after M1
`EntLifecycleOrchestrator.java:41-44` attributes the 600s cap to "a misconfigured or stale budget
**supplier**" and claims close "can never wait longer than the operator's k8s termination grace
allows." Post-M1 the rendezvous budget is the in-flight per-request `activeRoleSwitchTimeoutMs`, which
can legitimately reach 600s — far past a typical 30s grace (then SIGKILL; safe for the in-process UAF
but defeats graceful drain). Cap value still correct; reword the framing.

### m-3 — [ENT, in-diff, comment] dropped/inaccurate re-interrupt rationale; promote interrupt is inert
The M1 refactor deleted the inline comment explaining why `cascade.interrupt()` is re-issued every 50ms
(`:443-446`), and the retained class javadoc (`:424-427`) describes `drainWriterPool`'s catch
inaccurately (it *re-sets* the flag and `break`s — `EntCairoEngine.java:1575-1579` — it does not "clear
and resume sleeping"). Also: on the **promote** path there is no `Thread.sleep` to interrupt (only the
demote drains), so the interrupt is set-and-ignored there and the abort is driven entirely by the
`isClosing()` poll/guards. Restore a correct rationale and note the interrupt is the demote mechanism,
inert on promote.

### m-4 — [ENT, comment] misplaced "volatile because" field comment, now load-bearing for M1
`EntLifecycleOrchestrator.java:53-58`: this comment documents `activeSwitchTimeoutMs` (the `AtomicLong`
at `:60`, the "-1 when timeout omitted" budget) but physically sits **above** `activeSwitchTargetRole`
(the `AtomicReference<Role>` at `:59`), and says *"volatile because …"* for a field that is an
`AtomicLong`, not a volatile primitive. Pre-existing, but M1's new read path
(`closeRendezvousBudgetMs → activeRoleSwitchTimeoutMs → activeSwitchTimeoutMs.get()`) makes the field's
cross-thread contract load-bearing. Move the comment above `:60` and change "volatile because" → "an
AtomicLong because". (Caught by the prior round; verified still present.)

### m-5 — [ENT, test] per-reload-step guards #2–#4 and the ACL-disabled promote-close are un-witnessed
`CloseDuringPromoteAclReloadRaceTest` parks at the `onBeforePrimaryAclReload` seam **before** the first
guard, so only guard #1 ever fires on each of the cascade and heal; deleting any of the step-2/3/4
guards (`EntCairoEngine.java:1318/1320/1322`, `:1928/1930/1932`) keeps every test GREEN. And all three
tests boot ACL-enabled, so the ACL-disabled fast-path guards (`:1194`, `:1200`) are uncovered. A much
smaller defense-in-depth gap than before (the prior round's "no abort coverage at all" is resolved);
non-blocking.

### m-6 — [OSS, in-diff, comment] the poll comment names enterprise internals
`CairoEngine.java:581-586` names `switchMatViewMachinery` and "panic=abort cdylib" — enterprise/native
concepts an OSS reader cannot navigate to. The OSS *body* framing ("public engine seam with no in-tree
caller; the enterprise role switch invokes it") is fine; genericize the *in-code* comment (e.g. "an
out-of-tree caller drives this during a role switch and frees the store on close").

### m-7 — [ENT, test] two new orchestrator tests omit the `test` prefix
`EntLifecycleOrchestratorTest.java`: `closeRendezvousBudgetClampsToFloorAndCeiling` and
`closeRendezvousBudgetUsesPerRequestTimeoutNotConfiguredDefault` break the file's prevailing `test…`
naming (e.g. `testCurrentRoleReflectsLatestSwitchRoleCall`). The third,
`awaitInFlightWorkBoundsToPerRequestBudgetViaCloseRendezvous`, likewise. In-file consistency nit.

### m-8 — [OSS, metadata] OSS title uses internal jargon for a release note
`fix(core): prevent a crash on shutdown during materialized view hydration` — "hydration" is the
internal name of the disk-scan step. Conventional Commits is satisfied; for the release note, prefer
user-facing phrasing ("…while materialized views are loading"). The ENT title ("during a primary
promote") is fine — "promote" is replication-user vocabulary. Soft nit.

---

## Pre-existing (NOT introduced by this PR — noted for awareness)

- **P-1 [ENT]** `switchMatViewMachinery` sets `matViewRefreshSuspended=true` at `:2000`; if
  `createConcurrentQueue` (`:2004`) or `new MatViewStateStoreImpl` (`:2009`) throws **before** the
  `try` at `:2018`, the clearing catch (`:2032`) never runs → gate stuck true with a real (pure-Java)
  queue feeding a NoOp store. Pure-Java, no native leak; a refresh-wedge nit. Motivates M-2's asserts.
- **P-2 [ENT]** No `throwIfClosing` between the live-pointer publish (`:1326-1342`) and hydrate
  (`:1350`) on the cascade PRIMARY path: a close during the final `loadSecrets` publishes primary
  pointers, then aborts at hydrate — a logical torn-pointer state, **not a UAF** (every pointer
  references an un-freed store; freed after the rendezvous, nulled at `:395-397`), moot on a closing
  node. Pre-existing publish-before-hydrate ordering.
- **P-3 [ENT]** `nanoTime() < deadlineNanos` (`:442`) mis-evaluates only within ~10 min of
  `Long.MAX_VALUE` (astronomically unlikely). Canonical form: `deadlineNanos - System.nanoTime() > 0`.

---

## Downgraded (false positives / verified-benign)

- **Rendezvous `cascade.interrupt()` corrupting a later native op (loadSecrets/hydrate).** Dropped:
  QuestDB file I/O is native (`io.questdb.std.Files`), non-interruptible; locks are `ReentrantLock.lock()`
  (non-interruptible); the flag is discarded when the shutdown executor terminates the thread; a
  hypothetical spurious throw lands in the orchestrator's `catch(Throwable)` → UNKNOWN (safe on a
  closing node).
- **Two-atomic budget read (`switchInFlight` then `activeSwitchTimeoutMs`) returns wrong/unbounded.**
  Dropped: always clamped `[30s,600s]`; the only interleaving reads `-1` → configured-default fallback
  while the cascade is already finishing (so `awaitTermination` returns at once); `executor.shutdown()`
  + `submitSwitch`'s `isClosed()` check prevent a new switch intervening.
- **`Misc.free(realStore)` in the abort catch double-frees with `engine.close()`'s
  `Misc.free(matViewStateStore)` on the normal abort.** Dropped: the catch swaps the delegate to NoOp
  (`:2021`) *before* freeing realStore (`:2029`), so `engine.close()` frees only the NoOp singleton.
  The double-free materializes **only** in the M-1 budget-expiry residual.
- **The new guard throw escapes to the executor UncaughtExceptionHandler and crashes/exits the JVM.**
  Dropped: the sole production driver `EntEngineInitEnvelope.switchRole` (`EntServerMain.java:954-977`)
  catches+rethrows into the orchestrator's per-component `catch(Throwable)` → DEGRADED/UNKNOWN; nothing
  reaches the UEH.
- **M1 breaks the SIGTERM upload drain / other `activeRoleSwitchTimeoutMs` consumers.** Dropped:
  `ReplicationEnvelope.stop()` reads the configured default independently and floors/ceilings with the
  same constants; `contextFor` sharing the method is consistent (both want the in-flight per-request
  budget); the no-switch path is byte-identical.
- **Empty token bucket / non-mat-view token / `MatViewTest` direct call.** Dropped: empty bucket does
  zero native work; the poll runs before the `isMatView` filter so it fires for any token; `MatViewTest`
  never sets `closing`, so the poll is a constant-false branch there.
- **Unclosed LOG chain / `.put()` mid-chain / member ordering / banner comments / non-ASCII logs.**
  Dropped: every touched LOG chain closes with `.$()`/`.I$()`; the OSS poll throws via
  `CairoException.nonCritical().put(...)` (terminated by `throw`, not a LOG); all three new members sit
  in their correct alphabetical+visibility slot (verified against `.idea/codeStyles`); no `// ===`/`//
  ---` banners; new log/exception strings are ASCII (the lone non-ASCII em-dash at
  `EntLifecycleOrchestratorTest.java:355` is pre-existing, in an `Assert.fail` string, not a LOG).

---

## Summary

- **ENT #1091: approve. OSS #7335: approve.** The UAF fix is correct and well-scoped. The branch has
  materially improved since the previous r3 report (ACL/heal abort coverage, M1 wiring witness, honest
  test relabeling, accurate body count all landed). Six+ independent agents plus direct source tracing
  found no Critical and no blocking correctness/concurrency/resource defect.
- **Blocking: none.** Cheap and worthwhile before merge: M-2 (add the two post-abort assertions) and the
  M-1 body-scope broadening; consider the M-1 `MatViewStateStoreImpl.close()` idempotency as a follow-up.
- **Tradeoffs (disclosed, intended):** the longer worst-case close rendezvous (up to 600s for an
  authorized long promote), the DEGRADED/UNKNOWN torn node after an interrupted promote (needs a retry
  to heal), and the per-view UAF residual (narrowed, not eliminated). The headline is honest if framed
  as "bounds/narrows a critical UAF window," which the body's Tradeoffs already do.
- **Findings: 2 Moderate + 8 Minor confirmed, 3 pre-existing noted, ~7 dropped.** In-diff/out-of-diff:
  all confirmed are in-diff; the cross-context pass cleared every out-of-diff callsite as SAFE — a true
  clear (OSS `hydrateMatViewStateStore`: 1 production caller; ENT `engine.switchRole(Role,long)`: 1
  production driver; `activeRoleSwitchTimeoutMs`: 2 consumers), not an under-run.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 1 / 2 (50.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|1|2|50.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=249897)
**Tested Commit:** `c31a135d5dff1aafbdc672fdf6cb5ce69a11fcaa`

jovfer:
Superseded by #7361 — the hydrate-abort commit (`0a83347e68`) has been merged into `sm_role_switch_teardown_repl_fixes` (merge `8bfa699391`) and now rides #7361, because its enterprise tandem (questdb-enterprise#1091) was bundled into questdb-enterprise#1109: the two enterprise changes interact in the same role-switch close paths and must be validated in one CI matrix.

The level-3 review's C-2 hardening (idempotent `MatViewStateStoreImpl.close()` — clear the state map after the free loop, plus a double-close unit test) will be added on #7361 as requested. Authorship is preserved in branch history and will be credited with `Co-authored-by` on the squash.

PR #7336, opened 2026-06-26
fix(sql): avoid intermediate overflow in regr_r2() denominator (#7328)

### LINKED ISSUE #7328: regr_r2() denominator overflows/underflows like corr() did
`regr_r2()` computes the coefficient of determination as `(sumXY * sumXY) / (sumX * sumY)` over Welford sums of squared deviations, at `RegressionR2FunctionFactory.java:77`. This has the same intermediate over/underflow exposure that #7313 fixes for `corr()`, and additionally has no clamp to `[0, 1]`:

- For large-magnitude inputs (near `+/-1e153`), `sumX * sumY` overflows to `+Infinity`, so the result becomes `0` instead of the true r-squared. The numerator `sumXY * sumXY` is itself a square that can overflow independently.
- For small-magnitude inputs (near `+/-1e-150`), `sumX * sumY` underflows to `0.0`, producing `Infinity` / `NaN`.
- The result is not clamped, so the two-multiplication denominator can let it drift slightly above `1.0`.

Reproduce:

```sql
SELECT regr_r2(y, x) FROM t;   -- returns 0; should be 1.0
```

Suggested fix: mirror the `corr()` fix in #7313. The cleanest option is to compute the Pearson ratio with the overflow/underflow-safe split-sqrt denominator (and `[-1, 1]` clamp), then square it, which lands in `[0, 1]` naturally. The existing SQL:2003 short-circuits (`sumX == 0` -> NaN, `sumY == 0` -> 1.0) should be preserved.

Found while reviewing #7313; left out of that PR to keep it scoped.

kapil971390:
Fixes #7328.

## What

`regr_r2()` computes the coefficient of determination as:

```
(sumXY * sumXY) / (sumX * sumY)
```

where `sumX`, `sumY`, `sumXY` are Welford sums of squared deviations. The denominator product `sumX * sumY` overflows to `+Infinity` for inputs near `±1e153`, and underflows to `0.0` for inputs near `±1e-150`. Both produce wrong results:

| Input magnitude | `sumX * sumY` | Old result | Correct result |
|---|---|---|---|
| `±1e153` | `+Infinity` | `NaN` | `1.0` |
| `±1e-150` | `0.0` | `NaN` / `Infinity` | `1.0` |

## Fix

Mirror the approach taken for `corr()` in #7313:

1. Detect overflow (`product == +Infinity`) or underflow (`product == 0.0`) of `sumX * sumY`
2. Fall back to `sqrt(sumX) * sqrt(sumY)` — each factor stays well below overflow
3. Clamp the final result to `[0, 1]` to absorb sub-ULP rounding drift from the split-sqrt path

Normal-magnitude inputs keep the original single-sqrt path and bit-exact results.

## Manual analysis and verification

**How the bug was found:** Reviewed `RegressionR2FunctionFactory.java` alongside the merged `corr()` fix in #7313. The PR description for #7313 explicitly noted that `regr_r2()` had the same denominator exposure and was left out to keep that PR scoped (tracked as #7328). Confirmed by reading:

- `RegressionR2FunctionFactory.java` line 77: `return (sumXY * sumXY) / (sumX * sumY);` — direct product, no guard
- `CorrGroupByFunctionFactory.java` post-fix: uses `Math.sqrt(sumY * sumX)` — same exposure but the PR description explains the split-sqrt approach

**Manual reproduce steps (before writing the fix):**

Simulated the Welford accumulator arithmetic to confirm the overflow/underflow paths:

1. **Overflow** — two rows `(1e153, 1e153)` and `(-1e153, -1e153)`:
   - After Welford update: `sumX ≈ 2e306`, `sumY ≈ 2e306`, `sumXY ≈ 2e306`
   - `sumX * sumY = +Infinity` (IEEE 754 overflow)
   - Old: `(2e306)² / Inf = Inf / Inf = NaN` — **wrong**, should be `1.0`
   - Fixed: `sqrt(2e306) * sqrt(2e306) = 2e306` as denom → `r² = 1.0` ✓

2. **Underflow** — two rows `(1e-150, 1e-150)` and `(-1e-150, -1e-150)`:
   - After Welford update: `sumX ≈ 2e-300`, `sumY ≈ 2e-300`, `sumXY ≈ 2e-300`
   - `sumX * sumY = 0.0` (IEEE 754 underflow)
   - Old: `sqrt(0.0) = 0.0` → division by zero → `NaN` — **wrong**, should be `1.0`
   - Fixed: `sqrt(2e-300) * sqrt(2e-300) = 2e-300` as denom → `r² ≈ 1.0` ✓

3. **Normal case** — dataset `{(1,2),(2,4),(3,5),(4,4),(5,5)}`:
   - `sumX = 10, sumY = 6, sumXY = 6` — product `60` is finite and non-zero
   - Old: `36 / 60 = 0.6`, Fixed: `6 / sqrt(60) → 0.6` — **unchanged** ✓

## Tests added

Three new test methods in `RegressionR2FunctionFactoryTest`:

- `testRegrR2OverflowDenominator` — inputs near `±1e153`, expects `1.0`
- `testRegrR2UnderflowDenominator` — inputs near `±1e-150`, expects `round(..., 10) = 1.0`
- `testRegrR2ClampAboveOne` — moderate large-magnitude inputs, verifies result never exceeds `1.0`

All existing tests remain unchanged.

## Trade-offs

Same as noted in #7313 for `corr()`: the split-sqrt path uses two `sqrt` roundings instead of one, so for underflow-range inputs the result can be `0.9999999999999998` rather than exactly `1.0` (within 2 ULP of the true value). The `[0, 1]` clamp handles the overflow tail where rounding can overshoot.


kapil971390:
**Reproduce — BEFORE fix (bug):**

Screenshot shows the Welford accumulator simulation for overflow and underflow cases returning wrong results (NaN) before any fix was written.

questdb-7328-regr_r2-overflow

---

**After fix:**

Same cases with the split-sqrt fallback return correct results.

questdb-7328-regr_r2-overflow-after

bluestreak01 (changes requested):
@kapil971390 — adversarial review below. The fix is functionally correct and does resolve #7328, but I'm requesting changes on a few items before merge.

**Verdict:** No Critical correctness/concurrency/resource/security findings. The diff is self-contained — `RegressionR2Function.getDouble()` is a `private static` method with no external callers, subclasses, or overrides (zero out-of-diff blast radius). The issues are an unnecessarily broad normal-path rewrite that contradicts the PR's own "bit-exact" claim, one ineffective test, and several inaccurate comments + metadata.

## Moderate

### M1 — The normal (finite) path is rewritten unnecessarily; the "bit-exact" claim is false
`RegressionR2FunctionFactory.java:84-95`

Old code: `(sumXY * sumXY) / (sumX * sumY)`. New code routes the common finite case through `r = sumXY / sqrt(product); r2 = r*r`. These are mathematically equal but **not** bit-identical (sqrt + divide + square = 3 roundings vs. multiply + divide = 2). Verified by exact double-precision simulation:

- `testRegrR2KnownDataset` data: old `0.6` → new `0.6000000000000001`
- minimal repro `sumX=2, sumY=1, sumXY=1`: old `0.5` → new `0.4999999999999999`
- random sweep: **~60% of finite-path inputs change** in the last ULP.

The PR description says *"Normal-magnitude inputs keep the original single-sqrt path and bit-exact results."* This is doubly wrong: the original path had **no sqrt at all**, and the results are **not** bit-exact. The rewrite is also gratuitous — by Cauchy–Schwarz `sumXY² ≤ sumX·sumY = product`, so whenever `product` is finite the old numerator provably cannot overflow (`|sumXY| ≤ sqrt(product) < 1.34e154`). The only inputs the fix actually rescues are `product == +Inf` and `product == 0.0`.

Recommended minimal form — confines the change to the broken tails, keeps the original bit pattern + fewer ops on the hot path, and still clamps uniformly:

```java
final double product = sumX * sumY;
final double r2;
if (product == 0.0 || Double.isInfinite(product)) {
    final double r = sumXY / (Math.sqrt(sumX) * Math.sqrt(sumY)); // rare path only
    r2 = r * r;
} else {
    r2 = (sumXY * sumXY) / product; // bit-exact with pre-PR behavior
}
return r2 > 1.0 ? 1.0 : r2;
```

If you keep the current shape, please correct the description to drop "bit-exact"/"unchanged" and state the finite-path result changes sub-ULP for most inputs.

### M2 — `testRegrR2ClampAboveOne` does not exercise the clamp it is named for
`RegressionR2FunctionFactoryTest.java:321-331`

The test is named/commented as the dedicated guard for `r2 > 1.0 ? 1.0 : r2`, but for its inputs the **unclamped** value is *exactly* `1.0` (verified), so the clamp branch is never taken — it passes byte-identically with or without the clamp. The clamp is in fact exercised by `testRegrR2OverflowDenominator` (unclamped `1.0000000000000004`, which would fail the `1.0` assertion without the clamp). Either give this test inputs whose unclamped `r2 > 1.0`, or delete it as redundant. Note: no test exercises the clamp on the *finite* `sqrt(product)` path.

### M3 — The cited precedent (`corr()` fixed in #7313) is not present in this tree
`CorrGroupByFunctionFactory.java:113` still does the **unguarded** `return sumXY / Math.sqrt(sumY * sumX);` — no split-sqrt, no overflow guard, no clamp — so `corr()` retains the *identical* overflow/underflow bug this PR fixes (and `regr_r2 = corr²`, same denominator). Either #7313 is still unmerged (then the "post-fix" quote describes un-fixed code and is misleading), or the reference is wrong. Consider bundling the symmetric `corr()` fix here, or correct the description.

## Minor

- **Inaccurate comment** (`testRegrR2OverflowDenominator`, line 289): "Before the fix this returned 0 (0.0 / Infinity)." False — the old numerator `sumXY²=(2e306)²` also overflows to `+Inf`, so the old result was `Inf/Inf = NaN`, not `0`. (Your own summary table correctly says `NaN`.)
- **Inaccurate comment** (`testRegrR2UnderflowDenominator`, line 305): "dividing by sqrt(0.0) = 0…". The old code had no `sqrt`; it was `0.0/0.0 = NaN`. The `NaN` conclusion is right but the mechanism described belongs to the new code.
- **Comment overstates the clamp** (line 83): "clamped to [0, 1]" but only the upper bound is enforced. Harmless (`r2 = r*r ≥ 0`), but the comment should say "clamped to ≤ 1".
- **Missing labels:** this PR has no labels; a `fix(sql)` PR should carry at least `SQL` and `Bug`.

## Verified false positives (dismissed after source check)

- *"Split fallback `sqrt(sumX)*sqrt(sumY)` can itself overflow"* — unreachable; at that magnitude `sumX`/`sumY` already overflow upstream, degrading to `NaN → NULL`, never a silent wrong finite answer.
- *"sqrt of a negative product/moment"* — unreachable; Welford M2 is provably ≥ 0 through `aggregate()` and `merge()`.
- *"`assertMemoryLeak` + `.noLeakCheck()` wrapper"* — not a violation; the bodies hold multiple statements and match the file's convention.
- *"`regr_slope`/`regr_intercept` need the same coverage"* — not owed; their denominator is `sumX` alone, not the product. Only `corr()` shares the exposure (see M3).

---
*9 draft findings verified, 4 dropped as false positives. All verified findings in-diff; 0 out-of-diff (single private-static method, confirmed via repo-wide search).*

kapil971390:
Thanks for the thorough review. All three moderate items addressed in the latest commit:

**M1 — Restored bit-exact finite path**
The normal-magnitude path now uses the original `(sumXY * sumXY) / product` formula — no sqrt, same rounding as pre-PR. The split-sqrt fallback is confined to the two broken tails (`product == 0.0 || Double.isInfinite(product)`). The PR description claim of "bit-exact / unchanged" was wrong for the previous shape; it's accurate now.

**M2 — testRegrR2ClampAboveOne now actually exercises the clamp**
Verified via Python simulation that the overflow inputs (`1e153, 1e153` / `-1e153, -1e153`) produce an unclamped r² of `1.0000000000000004`. The test now asserts `regr_r2(y, x) = 1.0` (exact equality) against those inputs, so it fails without the clamp and passes with it. The redundant moderate-magnitude variant was removed.

**M3 — Comments corrected**
- Overflow test comment: was "returned 0 (0.0 / Infinity)" → now "sumXY² also overflows to +Inf, giving Inf/Inf = NaN" (matches the summary table).
- Underflow test comment: was "dividing by sqrt(0.0)" → now "(sumXY * sumXY) / 0.0 = Infinity or NaN" (the old code had no sqrt).
- Production comment: "clamped to [0, 1]" → "clamped to ≤ 1.0" (only the upper bound is enforced).

kapil971390:
Could a maintainer please add the `Bug` label to this PR? I don't have write access to apply labels from the contributor side.

kapil971390:
Correction on my 2026-06-27 reply: I mislabeled the comment-correction items as "M3" — those were the **Minor** section items. I never actually replied to the real **M3** (corr() precedent). Addressing it properly now:

**M3 — corr() precedent**

You were right that the reference was misleading at review time — `CorrGroupByFunctionFactory.java` was still unguarded then. Since the review, #7313 (the corr() overflow/underflow fix, same split-sqrt approach, by @Sergey Minaev) was merged into upstream `master` at `3b0fc98`. This branch has since been synced with `master`, so `corr()` now goes through the same guarded path via the shared `Numbers.corrFromSums()` helper (`core/src/main/java/io/questdb/std/Numbers.java:663`) — confirmed `3b0fc98` is an ancestor of this branch's current HEAD.

So the PR description's "Mirror the approach taken for corr() in #7313" is accurate now — it wasn't at review time, and you were right to flag it.

PR #7340, opened 2026-06-29
feat(core): add REPLICA ONLY index modifier to skip index build on replicating primaries

nwoolmer:
## What

Adds a per-index **`REPLICA ONLY`** modifier. The index is part of the replicated schema (identical on every node) but its physical materialization is node-local and driven by the replication role: replicas/standalone build and use it; replicating **primaries skip the build** (no per-row index maintenance on the apply path) and queries transparently fall back to full scans. Works for BITMAP, POSTING/POSTING_DELTA/POSTING_EF, and covering (`INCLUDE(...)`) indexes via the shared `IndexType` dispatch.

This is the OSS-core half; the enterprise role wiring lives in questdb/questdb-enterprise.

## How

- **Syntax**: trailing `REPLICA ONLY` on the index clause in `CREATE TABLE` and `ALTER TABLE ... ADD INDEX` (contextual keywords, not reserved).
- **Persistence**: new `_meta` flag bit 7 (`META_FLAG_BIT_REPLICA_ONLY`), threaded through reader/writer/sequencer metadata, WAL metadata transactions, and metadata caches so the flag is identical on every node. The sequencer payload is checksummed (version + count + flag bytes) and only accepts canonical 0/1 flags, with absent-section back-compat.
- **Role seam**: `CairoConfiguration.skipReplicaOnlyIndexes()` (default `false` in OSS) drives the write-path skip, planner index-eligibility (`isColumnIndexActive`), and the checkpoint-rebuild gate.
- **Read path**: the planner treats a skipped replica-only column as un-indexed (full scan); the reader tolerates a transiently-absent replica-only index file as recoverable (non-critical) rather than corruption.
- **Reconcile**: build on demote / purge on promote, reopen-independent via an engine role-generation counter + writer self-heal at the WAL-apply choke point. Presence is validated per partition across every required sidecar (`.k/.v`, `.pk/.pv`, covering `.pc`); failed purges are propagated (the generation is not acknowledged until absence is verified).
- **Role-sensitive plan cache**: a compiled factory that references a replica-only column is wrapped so its cursor execution revalidates the engine role generation (recompile on a role flip); `bumpRoleGeneration()` also flushes the query cache. `touch_table()` now reads the role live at execution.
- **ATTACH**: on a non-skipping node, a partition whose replica-only index sidecars are absent (e.g. a backup produced by a skipping primary) has its index **rebuilt synchronously** at the final table path before the partition is published.
- **Introspection**: `SHOW CREATE TABLE` re-emits `REPLICA ONLY`; `table_columns()` gains `indexReplicaOnly`.

## Tradeoffs / disclosures

- **Full-scan cost on primaries**: skipping index maintenance lowers primary write cost but makes filtered/LATEST-BY/index-hint queries full-scan on the primary.
- **Transient query degradation**: during an incomplete materialization window (mid restore/promote/reconcile) an index-selecting query on a non-skipping node may raise a **non-critical, recoverable** `replica-only index not materialized` error; it succeeds once the next WAL-apply reconcile completes. The fuzz oracles assert this contract (non-critical errno) and require the query to converge and match an index-free reference table per partition.
- **Posting seal-floor assertion** was relaxed from strict-monotonic to entry-floor (intra-entry generations are non-monotonic by design; see `POSTING_INDEX_CHAIN_DESIGN.md` §4.5.1). A previously-quarantined multi-column fuzz seed (`0xABCD`) now runs to completion with the quarantine removed. Broader posting `.pk`-absent planner/reader hardening is tracked separately in #7349 and is out of scope here.

## Testing

New/updated test classes cover parser round-trips (all index types), `_meta` persistence, sequencer payload corruption/bit-flip/non-canonical bytes, write-path skip, planner fallback (WHERE, single- and multi-key LATEST BY, ASOF index-hint, DISTINCT/ORDER BY/window shapes, aliased projections, explicit NULL), reader missing-file tolerance, reconcile build/purge across role flips (including older/partial sidecars and failure-injected purge), non-skipping ATTACH rebuild of an older sidecar-less partition, `UPDATE`/`touch_table()` on both roles, stale-factory-reuse across role transitions, checkpoint-rebuild gating (native + parquet), and single-/multi-column fuzz under role flips with a strict per-partition convergence oracle.


bluestreak01 (changes requested):
## Review — Level 3 (mission-critical pass) — **Request changes**

Two verified **Critical** defects, both the *same failure class* the PR set out to eliminate: reading the schema `indexed` flag and then touching physical index files without the node-role gate. The core design (role-driven node-local materialization behind `skipReplicaOnlyIndexes()`, `isColumnIndexActive` planner gating, reconcile self-heal, reader tolerance, backward-compatible `_meta`/sequencer/`extraInfo` formats) is sound and unusually well-documented, and the great majority of the ~40 changed physical-index callsites are correctly guarded — but not all.

Pipeline: 12 review agents + per-finding source verification. ~34 draft findings; 2 dropped as false positives.

> ⚠️ **C1 is a working-tree regression, not visible in `gh pr diff`.** The branch HEAD is 6 commits ahead of origin and includes a local merge (`4214cbce4f`) of master `39f902479b7`. `gh pr diff` *shows* the guard being added, but the checked-out tree lost it in that merge. Review against the working tree, not the diff.

---

## Critical

### C1 — `resealParquetCoveringForPartition` lost its `indexers` bounds guard in the local master merge → the PR's own covering-posting reseal test asserts/crashes under `-ea`

**`core/src/main/java/io/questdb/cairo/TableWriter.java:12966` (loop guard) + `:12979` (`indexers.getQuick(colIdx)`).**

Every other per-column posting/indexer loop the PR touched carries `colIdx >= indexers.size()` — fast-lag `:5201`, `publishPostingIndexesForLastPartitionFastLag :11611`, `restorePostingIndexersToLastPartition :13089`, `sealPostingIndexForPartition :13420`. `resealParquetCoveringForPartition` does **not**: its guard (`:12966-12968`) is only `getColumnType<=0 || !isColumnIndexed || !isPosting`, then it calls the unchecked `indexers.getQuick(colIdx)` at `:12979`. `ObjList.getQuick` is `assert index < pos; return buffer[index]` — under `-ea` (always on; every test) `colIdx >= indexers.size()` throws `AssertionError`; in a no-assert build it is an OOB read (AIOOBE/suspend, or a stale pooled indexer driven against the wrong column).

`git blame -L 12966,12968` attributes the guard-less loop to master `39f902479b7` (which refactored it to `instanceof`), so the merge dropped the PR's guard.

Reachable today on a skipping primary with a covering replica-only POSTING index: `sealPostingIndexForPartition` → for a parquet partition `return resealParquetCoveringForPartition(...)` (`:13385`, delegates *before* its own guarded loop) → `hasCoveringPostingIndex()` reads metadata only (true for replica-only) → the loop reaches the replica-only column → `indexers.getQuick(colIdx)`.

The PR's own `TableWriterReplicaOnlySkipTest.testParquetCoveringPostingResealReplicaOnlyDoesNotCrash` (`:503`) creates `x (s symbol index type posting include (v) replica only, v double, ts timestamp)` — `s` is column 0, the only indexed column, so on a skipping primary `indexers.size()==0` and `getQuick(0)` asserts `0 < 0`. The test comment already says: *"A replica-only covering-posting column on a skipping primary has no wired indexer -> out-of-bounds get without the bounds guard."* CI hasn't caught it because the merged HEAD isn't pushed.

**Fix:** restore the guard, matching the three siblings:
```java
if (metadata.getColumnType(colIdx) <= 0 || !metadata.isColumnIndexed(colIdx)
        || colIdx >= indexers.size()
        || !IndexType.isPosting(metadata.getColumnIndexType(colIdx))) {
    continue;
}
```
*(Verified by source read + `git blame` + `ObjList.getQuick` semantics + the test schema; not run through a full `mvn` build — the trace is deterministic and the failing test is in the PR.)*

### C2 — `FrameImpl.getContiguousFileFrameColumn` uses raw `isColumnIndexed` → partition **squash** crashes (suspend) and **split** wrongly materializes a replica-only BITMAP index on a skipping primary (out-of-diff, untested)

**`core/src/main/java/io/questdb/cairo/frm/file/FrameImpl.java:204`.**
```java
boolean isIndexed = metadata.isColumnIndexed(columnIndex);              // raw — NOT role-aware
int indexBlockCapacity = isIndexed ? metadata.getIndexValueBlockCapacity(columnIndex) : 0;
```
`FrameImpl` gets the writer's `TableWriterMetadata`, whose `isColumnIndexed()` is schema-true for a REPLICA ONLY column even on a skipping primary. The PR added `isColumnIndexActive(...)` for exactly this and guarded O3PartitionJob/openPartition/processO3Block/copyOrRebuildColumnIndexes/linkPartitionIndexFiles/rebuildPartitionIndexFiles — but not the frame split/squash path, and `FrameImpl` has no access to `skipReplicaOnlyIndexes()`.

`squashPartitions` (`TableWriter.java:13945`) and `splitLastPartition` (`:8695`) both pass the writer `metadata` to the frame factory; `ContiguousFileColumnPool.create` builds an indexed frame column whenever `indexBlockCapacity > 0`:
- **Squash — crash/suspend:** RW target, `isEmpty=false` → `indexWriter.of(..., 0)` → `BitmapIndexWriter.of` sets `init=false` → `if (!ff.exists(keyFile)) throw CairoException.fileNotFound("index does not exist")` (`BitmapIndexWriter.java:301-304`). The replica-only `.k` never existed → **table suspended.**
- **Split — invariant violation:** `createRW` → `isEmpty=true` → `init=true` → it **builds `.k/.v` for the replica-only column** (the work the feature offloads), then leaves it unmaintained; `replicaOnlyIndexFilesPresent()` probes the *last* (split-off) partition and is fooled into "materialized".

No test drives squash/split for a replica-only index. **Fix:** thread `skipReplicaOnlyIndexes()` into `FrameImpl` and gate `:204` on `isColumnIndexActive(columnIndex, skip)` (or suppress the indexed frame-column treatment for replica-only columns in split/squash while skipping). Add a skipping-primary + `SYMBOL INDEX REPLICA ONLY` + O3-split-then-`squashPartitions()` regression test.

---

## Moderate

- **M1 — `changeColumnType` not role-aware and drops `replicaOnly`** (`TableWriter.java:1312`, out-of-diff). `(SymbolColumnIndexer) indexers.get(columnIndex)` runs when `IndexType.isIndexed(indexType)` with no skip gate and no flag carry-over; on a skipping primary `ObjList.get` throws AIOOBE (or NPE / materializes the index). Inconsistent with the `changeSymbolCapacity` fix the PR made. Reachability of an indexed replica-only symbol column surviving a TYPE change as still-indexed is narrow, but add the same defensive gate + flag-carry for consistency.
- **M2 — Undocumented ordering contract between the role flip and `bumpRoleGeneration()`** (`CairoEngine.java:496`, `TableWriter.java:1584`). If the enterprise caller makes the generation bump visible *before* the skip-flag flip on the net-final transition, a WAL apply reconciles against the stale flag, stores `lastReconciledRoleGen=gen`, and never re-fires — index left built-when-should-be-purged until a writer reopen. Every OSS test does flag-then-bump but nothing enforces it. Document the contract on `bumpRoleGeneration()`/`getRoleGeneration()` ("publish the skip-flag BEFORE bumping; the bump is the release fence").
- **M3 — `replicaOnlyIndexFilesPresent()` judges whole-table materialization from the last partition only** (`~:12446`). If an older partition lacks the index while the last one has it (partial restore, or C2's split), reconcile skips the all-partitions rebuild and queries over the old partition hit the "not materialized" error. Probe more robustly or document the assumption.
- **M4 — Test gap: `PostingGenLookup` assertion relaxation has no deterministic test.** The weakened `assert txnAtSeal >= 0 && txnAtSeal >= genSealFloor` (live under `-ea`) is only exercised probabilistically by the multi-column fuzz. Add a boundary unit test: a legit 1-txn fast-lag dip passes; a >1-txn / recycled-gen regression still trips.
- **M5 — Test gap: sequencer optional-section cross-version framing untested.** No test for a pre-feature `_meta` (absent section), bad checksum, or `replicaOnlyCount != columnCount` — the guarded fallback-to-`false` paths. This is a WAL/replication cross-version format.
- **M6 — Test efficacy: the multi-column fuzz `.pk.`-substring quarantine can mask a genuine replica-only regression.** `ReplicaOnlyIndexMultiColumnFuzzTest` swallows any `"could not open, file does not exist"` + `".pk."`; columns `c`/`e` are replica-only posting/covering, so a real replica-only reconcile fault matches and is tolerated (and the whole-scenario catch bypasses that seed's file-presence oracle). Scope the quarantine to assert the faulting column is `replicaOnly==false`.

---

## Minor

- **Reconcile BUILD error path** leaves `denseIndexers` referencing the freed stale indexer if `writeIndex` throws (`:12327-12345`). Not reachable at runtime as analyzed, but the invariant is held by external control-flow, not the method. One-line hardening: `populateDenseIndexerList()` right after nulling `stale` (or in the `catch`).
- **Member-ordering (alphabetical-within-kind):** `TableColumnMetadata` field `replicaOnlyIndex` + getter `isReplicaOnlyIndex()` (after `symbolCacheFlag` — but the *setter* is placed correctly, so it's internally inconsistent); `RecordMetadata` (`isColumnIndexActive` should precede `isColumnIndexed`); `TableStructure.isReplicaOnlyIndex`; `CairoConfiguration.skipReplicaOnlyIndexes` + `CairoConfigurationWrapper` override (splits the `is*` run); `CairoEngine.bumpRoleGeneration()` (after `buildViewGraphs`); `TableWriter` fields `replicaOnlyIndexAbsentAtOpen`/`replicaOnlyReconcileWalPathSnapshot`.
- **Perf:** `PostingGenLookup.snapshotMetadata` does an unconditional native `keyMem.getLong` feeding only the assert, on the posting read/seqlock-retry path — move behind the assertion. Per-column `skipReplicaOnlyIndexes()` re-reads in O3/openPartition/processO3Block loops — hoist to a `final boolean`. Reconcile calls `populateDenseIndexerList()` once per column instead of once after the loop; plus a double `ff.exists` per present replica-only column at open (both role-flip/open only — low impact).
- **Test-code quality:** `indexFilesExist`/`matchesIndexFile` (+`deleteIndexFiles`/`forEachIndexFile`) copy-pasted across **7** new test classes — hoist to a shared `AbstractReplicaOnlyIndexTest`/`TestUtils`. `AlterOperationAddIndexFormatTest.CapturingMetadataService` hand-stubs ~25 methods instead of `implements MetadataServiceStub`. Home-grown single-pass `assertSql`/`assertContents` helpers resurrect the removed weaker compare — use `assertQuery(q).returns(expected)` for deterministic row assertions. `ReplicaOnlyIndexMultiColumnFuzzTest` class javadoc is a ~90-line investigation log (move to PR body/issue). `System.out.println` reproducers (6 sites). Boolean fields without `is/has`. 5+ digit literals without `_` (`1000000`, `86400000000` in `ReplicaOnlyReconcileTest`).
- **Test coverage nits:** `IsColumnIndexActiveTest` asserts only 2 of 4 truth-table arms; `SnapshotReplicaOnlyRebuildTest.testNonSkipNodeRebuildsAllIndexes` has no pre-assertion that files were absent before rebuild; `testReconcileIfRoleChangedTriggerOnPooledWriter` never asserts the 2nd call was a no-op; the `addColumn` sink `replicaOnlyIndex` arg has only indirect coverage.
- **By-design note:** the reader tolerance surfaces a *retryable query error* ("not materialized … retry shortly"), not a silent full-scan fallback, during the restore/promote window on a non-skipping node. The PR body's "queries transparently fall back to full scans" overstates it for that window — document.
- **PR metadata:** add scope labels (`Core`, `New feature`, `SQL`, `WAL`, `storage`, `Compatibility`); replace "All green." with a checkable status; add a Tradeoffs/limitations subsection (primary-side full-scan latency, relaxed `PostingGenLookup` assertion, cross-version binary formats constraining rolling upgrades); consider `feat(core): add REPLICA ONLY index modifier` (verb-first); drop the generated-with trailer.
- **Style:** an existing ASCII `// remove indexer - skip` was changed to an em-dash, and new comments/javadoc add em-dashes/arrows — QuestDB leans ASCII; revert to hyphens.

---

## Downgraded (false positives / not defects)

- **"Reader errno tolerance is dead code / masks corruption."** Verified live and correctly discriminated: `TableUtils.openRO` throws `CairoException.critical(ff.errno())` with real ENOENT; `Files.isErrnoFileDoesNotExist` matches; non-ENOENT and all normal-index failures rethrow critical. Negative-control tests pin this.
- **"`reloadColumnAt`/`getIndexReader` double-free or leak on tolerance."** `close()`/`Misc.free` are idempotent and `getAndSetQuick` nulls the slot atomically.
- **"WalWriter 5-arg `addIndex` drops `replicaOnly` → flag lost."** Intentional and correct: the authoritative flag flows through `SequencerMetadataService.addIndex` → `SequencerMetadata` → `syncToMetaFile` optional section (verified end-to-end).

---

## Summary

- **Request changes.** C1 (a working-tree merge regression that makes the PR's own `testParquetCoveringPostingResealReplicaOnlyDoesNotCrash` throw under `-ea`) and C2 (untested squash-crash / split-mis-materialize because `FrameImpl.java:204` was missed by the `isColumnIndexActive` sweep) both fail the test gate.
- Coverage map: mainline (parser, `_meta` bit 7, planner gating, reader tolerance, reconcile lifecycle, introspection, `extraInfo` compat) is genuinely well tested; gaps cluster on squash/split, `changeColumnType`, the relaxed assertion, and the sequencer cross-version framing.
- ~34 findings drafted, 2 dropped as false positives → **2 Critical, 6 Moderate, ~14 Minor.** Split: C1 is a branch-HEAD merge regression (flag against the working tree, not the diff); C2/M1 are out-of-diff callsites; M2 is in-diff (contract documentation); the rest in-diff.
- Please fix C1 + C2 (with the squash/split regression test) and M1's gate on this branch, document M2, and close the M4/M5 test gaps + tighten M6 — none of this should be split into a follow-up.

<sub>🤖 Level-3 adversarial review via pi review-pr skill (12 parallel review agents + per-finding source verification).</sub>

nwoolmer:
@bluestreak01 — thanks, both Criticals and all six Moderates addressed. Rebased on master (`8f07ffdbed`), full suite green; enterprise-ci green on the merged head.

**C1 (`resealParquetCoveringForPartition` lost its `indexers` bounds guard in the master merge)** — restored, matching the three sibling loops (`colIdx >= indexers.size()`). `24e7a89c5e`. The PR's own `testParquetCoveringPostingResealReplicaOnlyDoesNotCrash` now exercises it under `-ea`.

**C2 (`FrameImpl.getContiguousFileFrameColumn` raw `isColumnIndexed` → squash crash / split materializes a replica-only index)** — threaded `skipReplicaOnlyIndexes()` through `FrameFactory` into `FrameImpl` and gated the indexed-frame-column decision on `isColumnIndexActive(columnIndex, skip)`. `47495c514f`. Added a skipping-primary + `SYMBOL INDEX REPLICA ONLY` + O3-split-then-squash regression test (`TableWriterReplicaOnlySkipTest`).

**M1 (`changeColumnType` not role-aware, drops `replicaOnly`)** — same commit: gated on `isColumnIndexActive` and carry the `replicaOnly` flag, mirroring the `changeSymbolCapacity` fix.

**M2 (role-flip ↔ `bumpRoleGeneration()` ordering)** — documented the contract on `bumpRoleGeneration()`/`getRoleGeneration()`: publish the skip flag **before** bumping; the bump is the release fence the reconcile path acquires against. `47495c514f`.

**M3 (`replicaOnlyIndexFilesPresent()` last-partition assumption)** — documented the "reconcile is all-partitions-or-none, so the last partition is representative" assumption at the probe. Same commit.

**M4 (`PostingGenLookup` seal-floor assertion had no deterministic test)** — added `PostingGenLookupSealFloorTest` (boundary: a legit 1-txn fast-lag dip passes; a >1-txn / recycled-gen regression trips).

**M5 (sequencer optional-section cross-version framing untested)** — added `SequencerReplicaOnlyMetadataTest` (absent section, bad checksum, `replicaOnlyCount != columnCount`).

**M6 (fuzz `.pk.`-substring quarantine could mask a real regression)** — scoped `ReplicaOnlyIndexMultiColumnFuzzTest` to assert the faulting column is `replicaOnly == false` before tolerating.

bluestreak01 (changes requested):
@nwoolmer

## Review — PR #7340 `feat(core): REPLICA ONLY index modifier` — Level 3 (mission-critical pass) — **Request changes**

Pipeline: 5 parallel fresh-context review agents (correctness/gating audit, on-disk/wire back-compat + concurrency, performance, tests/coverage, unanchored adversarial) + per-finding source verification. Reviewed against the checked-out working tree (`feat/index-skip-primary`, HEAD == origin == `430166bd3f`), not just `gh pr diff`.

The core design — role-driven node-local materialization behind `skipReplicaOnlyIndexes()`, `isColumnIndexActive` planner gating, reconcile self-heal at the WAL-apply choke point, reader missing-file tolerance, and backward-compatible `_meta` bit-7 / sequencer optional-section / ALTER `extraInfo` formats — is sound, unusually well-documented, and heavily tested. I re-verified the prior review's C1 (`resealParquetCoveringForPartition` guard at `TableWriter.java:13003`) and C2 (`FrameImpl` `isColumnIndexActive` gate) fixes: both correct and complete. The great majority of the ~40 physical-index callsites are correctly gated.

But the same failure class the PR exists to eliminate — reading the raw schema `indexed` flag and then touching physical index files / the wired `indexers` list without the node-role gate — survives on the **ATTACH PARTITION path** (missed by the prior review, no test coverage), and the PR **ships a known, deterministic, reachable query-time crash quarantined in its own fuzz test**.

~14 draft findings; the two attach defects and the fuzz quarantine are the blockers.

---

## Critical

### C1 — `attachPartitionCheckSymbolColumn` requires physical index files with no role gate → ATTACH suspends the table on a skipping primary
**`core/src/main/java/io/questdb/cairo/TableWriter.java:4506-4525`** (missed-pattern; in-scope). **UNTESTED.**

`attachPartitionCheckSymbolColumn` reads the raw schema flag and demands the sidecars exist:
```java
byte indexType = metadata.getColumnIndexType(columnIndex);
if (IndexType.isIndexed(indexType)) {                    // :4507 - no role gate
    keyFileName(indexType, partitionPath.trimTo(pathLen), columnName, columnNameTxn);
    if (!ff.exists(partitionPath.$())) {
        throw CairoException.fileNotFound().put("Index key file does not exist ...");  // :4510
    }
    ... valueFileName ...; if (!ff.exists(...)) throw "Index value file does not exist"  // :4523
}
```
**Reachable trace:** `ALTER TABLE ... ATTACH PARTITION LIST '...'` passes `partitionSize < 0`, so `forceRenamePartitionDir = true` (`:1032`). When the attached partition lacks `_dmeta` (backups / older / filesystem-copied partitions - the code's own comment at `attachPrepare` acknowledges this), `attachPrepare` returns `false` → `attachValidateMetadata` (`:1055-1056`) → `attachPartitionCheckSymbolColumn` (`:4730`). On a skipping primary a replica-only symbol column reports `getColumnIndexType == indexed` but the invariant guarantees it has **no** `.k/.v` (or `.pk/.pv`) files → `fileNotFound` thrown. This runs during WAL apply of the ATTACH (WAL tables support ATTACH PARTITION - see `WalAlterTableSqlTest`), so the throw **suspends the table**. Also fails on a non-skipping replica attaching a primary's backup whose replica-only sidecars were never captured.

**Fix:** mirror the sibling gate in `copyOrRebuildColumnIndexes` (`:5367`) / `linkPartitionIndexFiles` (`:7755`): skip the existence check when `metadata.isColumnReplicaOnlyIndex(columnIndex) && configuration.skipReplicaOnlyIndexes()`, and tolerate a transiently-absent replica-only sidecar on a non-skipping node (the `!ff.exists(keyFileName(...))` tolerance branch). Add a `TableWriterReplicaOnlySkipTest` case: skip=true, `SYMBOL INDEX REPLICA ONLY`, detach then re-attach a `_dmeta`-less (backup-style) partition.

### C2 — `attachPrepare` rebuild branch materializes a replica-only index on a skipping primary (invariant violation + wasted IO)
**`core/src/main/java/io/questdb/cairo/TableWriter.java:4698-4701`** → `rebuildAttachedPartitionColumnIndex:12039` → `RebuildColumnBase.reindexColumn:163`. **UNTESTED.**

```java
} else if (isIndexedNow
        && (!wasIndexedAtDetached || indexValueBlockCapacityNow != indexValueBlockCapacityDetached
        || indexTypeNow != indexTypeAtDetached)) {           // :4698 - no role gate
    detachedPath.trimTo(detachedPartitionRoot);
    rebuildAttachedPartitionColumnIndex(partitionTimestamp, partitionSize, columnName);  // builds .k/.v
}
```
`rebuildAttachedPartitionColumnIndex` calls the `reindexColumn(ff, cvReader, metadata, columnIndex, …)` overload, which calls `doReindex` **directly, bypassing `isSupportedColumn`** (the gate that protects the REINDEX-all path at `RebuildColumnBase.java:293/311`). **Reachable trace:** detach a partition *before* `ADD INDEX … REPLICA ONLY` (so `wasIndexedAtDetached == false`), add the replica-only index, then re-attach on a skipping primary → `isIndexedNow == true`, `!wasIndexedAtDetached == true` → the branch builds physical `.k/.v` on a node that has no wired indexer for the column. This violates the invariant `indexer wired + files present <=> !skipReplicaOnlyIndexes()`: the skipping primary never maintains those files (stale/partial index), and `reconcileReplicaOnlyIndexes`'s last-partition-only probe (`:12442`) won't detect stray files on a non-last attached partition, so they persist unmaintained. Queries stay correct only because the planner treats the column inactive; the cost is wasted index materialization (an extra `doReindex` pass over the partition) and a latent stale-index window. A capacity/type change on a replica-only index triggers the same branch.

**Fix:** guard the branch (or `rebuildAttachedPartitionColumnIndex`) with `!(metadata.isColumnReplicaOnlyIndex(colIdx) && configuration.skipReplicaOnlyIndexes())`, mirroring `rebuildPartitionIndexFiles:12071`. The removal branch (`!isIndexedNow && wasIndexedAtDetached`, `:4671`) is fine - it only unlinks via tolerant `removeFileOrLog`.

### C3 — the PR quarantines a deterministic, reachable POSTING-reader query crash in its own fuzz test instead of fixing it
**`core/src/test/java/io/questdb/test/cairo/ReplicaOnlyIndexMultiColumnFuzzTest.java:118-142` (javadoc) + `:171-184` (quarantine).** Root cause lives in the posting-index reader/planner (`AbstractPostingIndexReader.of` via `TableReader.getIndexReader` / `SymbolIndexRowCursorFactory`). **Tagged pre-existing / newly-unmasked; in-scope per the no-deferral policy.**

The test's own javadoc documents that with the PR's seal-assert relaxation, seed `0xABCD` runs to ~op#327 and then **deterministically** throws a query-time `could not open, file does not exist: .../<partition>/<col>.pk.<colTxn>` on column `d_r229` - a **non-replica-only** POSTING index churned through `CONVERT PARTITION TO PARQUET` + drop/re-add/rename. "The planner emits an index scan but the per-partition posting `.pk` is absent for that partition, and the reader HARD-FAULTS." The test then tolerates exactly this signature and only asserts `strictlyCompleted >= 3` of 4 seeds (`:183`), so the crash ships green and "belongs to the posting-index owner."

This is a legitimate SQL sequence (ADD INDEX / CONVERT / DROP / re-ADD / RENAME) reproducible without the replica-only flag, i.e. a genuine reachable query-time crash. A confirmed reachable bug found in visited code is not deferred to a follow-up or a subsystem owner via a silent test quarantine - it is fixed here, and the fix gets its own regression test. Either the planner must not choose an index scan when the per-partition posting `.pk` is absent, or the posting reader must tolerate per-partition `.pk` absence (the way the replica-only reader path already does). If the maintainers judge that fixing the posting subsystem genuinely destabilizes this PR, that must be an explicit, named risk-acceptance decision - not a `strictlyCompleted >= 3` gate. As written, the quarantine also silently drops all coverage for seed `0xABCD` past op#327.

### C4 — reinvented single-pass `assertSql` helper on deterministic results defeats the `.returns(...)` battery
**`core/src/test/java/io/questdb/test/cairo/TableWriterReplicaOnlySkipTest.java:159-162`** and **`TableWriterReplicaOnlyNoSkipTest.java:64-67`** define `private void assertSql(String expected, String query)` doing `printSql` + `TestUtils.assertEquals`, called on deterministic queries at Skip `:140/:153/:155`; and the same file asserts deterministic results via raw `printSql` + `TestUtils.assertEquals` at `:189, :229, :293, :386, :416, :454, :493, :531`. Sibling offenders: `ReplicaOnlyReconcileTest:274`, `ReplicaOnlyIndexReaderToleranceTest:125/:207`, `ReplicaOnlyIndexPlannerTest:61/:90/:111`.

The removed `assertSql` was replaced by `assertQuery(sql).returns(expected)` specifically because the single-pass print/compare skips the second cursor pass, `calculateSize()` cross-check, variable-column check, and factory-property assertions. Reinventing it under the same name circumvents the removal on the exact new full-scan-fallback behavior these tests validate. All these queries are deterministic (no `rnd_*`, no `now()`), so `.returns(...)` applies. **Fix:** fold each into `assertQuery(sql).noLeakCheck()[.assertsPlanNotContaining(...)].returns(expected)` - the author already uses this form at `TableWriterReplicaOnlySkipTest:599`, so the pattern is in-file. Note: the `ReplicaOnlyReconcileTest` / `ReplicaOnlyIndexPlannerTest` cases run the plan check and the result check as two separate calls, so the results never receive the `.returns()` battery at all - folding them fixes both.

---

## Moderate

- **M1 - `createIndexReaderAt` redundant `ff.exists()` stat on the success path.** `TableReader.java:1085-1092`: a replica-only column runs an `ff.exists(keyFileName)` pre-check before `IndexFactory.createReader(...)`, but the `createReader` call is already wrapped in a `catch (CairoException)` that converts file-not-found to the identical non-critical error (`:1109-1124`). On the success path this is a strictly redundant extra `stat(2)` per index-reader open (per partition, per query, for replica-only index scans on a non-skipping node - not per-row). Behaviorally inert (the catch backstop alone preserves correctness), so Moderate; the tradeoff (clean early-out without partially initializing an `IndexReader`) is defensible, but the extra syscall on the read path is avoidable.

- **M2 - last-partition-only materialization probe can strand queries on older partitions.** `replicaOnlyIndexFilesPresent()` (`TableWriter.java:12442`) probes only the last partition and `reconcileReplicaOnlyIndexes` treats it as whole-table truth. After a *partial* external restore (last partition's index present, an older one's absent), open-time reconcile skips the rebuild and a query touching the older partition hits the `createIndexReaderAt` non-critical "not materialized, retry shortly" - which does **not** self-heal on a read-only/idle table (reconcile fires only on a WAL *insert* apply). Behaviorally inert for data correctness (retryable error, never wrong results or a crash) and documented in-comment as a deliberate tradeoff, so Moderate; worth confirming the enterprise operational runbook expects it.

- **M3 - replica-only reader tolerance can mask genuine corruption.** The three `catch (CairoException e)` blocks in `TableReader` (`:407, :1118, :1932`) convert *any* `FILE_DOES_NOT_EXIST` during index open into the benign retryable error for a replica-only column. For a *partially-materialized* replica-only POSTING index (`.pk` present, a `.pv.<sealTxn>` gen file missing = real corruption), this reclassifies corruption as "not materialized." Behaviorally bounded (the next insert-apply reconcile rebuilds), so Moderate, but it broadens the masking window beyond "never built"; consider narrowing the tolerance to the `.pk`-absent case.

- **M4 - `TableColumnMetadata` member ordering.** `TableColumnMetadata.java:48-49` declares `symbolCacheFlag` before `replicaOnlyIndex`, and the getter/setter (`:233/:268`) follow the same order. Per CLAUDE.md, private instance fields sort alphabetically, so `replicaOnlyIndex` should precede `symbolCacheFlag`. Behaviorally inert.

- **M5 - `indexFilesExist` / `matchesIndexFile` copy-pasted across 7 test classes.** The ~40-line index-file scanner is verbatim in `TableWriterReplicaOnlySkipTest`, `...NoSkipTest`, `ReplicaOnlyReconcileTest`, `ReplicaOnlyIndexReaderToleranceTest`, `ReplicaOnlyIndexFuzzTest`, `ReplicaOnlyIndexMultiColumnFuzzTest`, `SnapshotReplicaOnlyRebuildTest`. Inert for correctness but a maintenance liability: the core `assertFalse(indexFilesExist(...))` invariant the whole suite rests on would have to be fixed in 7 places, and divergence silently weakens it. Hoist a single `ReplicaOnlyIndexTestUtils` helper.

---

## Minor

- **`O3PartitionJob.java:3408` and `:3987`** recompute `tableWriter.getConfiguration().skipReplicaOnlyIndexes()` inside the per-column loop; it is loop-invariant (role cannot flip mid-O3-open under the writer lock). Hoist to a local. Negligible - a couple of virtual dispatches per column at partition-open time, off the per-row path.
- **`PostingGenLookupSealFloorTest`** guards `testNegativeSealTrips`/`testRegressionBelowFloorTrips` with `Assume.assumeTrue(...desiredAssertionStatus())`, so they no-op without `-ea`. Correct for an assert-only guard, but means the relaxed seal-floor invariant has zero production protection (a >1-txn seal regression is undetected with assertions off). Confirm that is the intended risk posture.

---

## Downgraded (verified, not defects / not OSS blockers)

- **Adversarial "O3 dynamic skip-flag read desync mid-commit"** (`O3PartitionJob:3408/:3987` re-reading the flag while `O3Basket` was sized earlier from `denseIndexers`) - **not reachable in OSS** (flag is a compile-time constant `false`; `bumpRoleGeneration` never called). Only reachable if the enterprise runtime flips the role mid-O3-commit without quiescing WAL apply → flag for the enterprise integration review, not an OSS finding.
- **Back-compat "mixed-version cluster drops the sequencer replica-only section"** - verified: an *old* node rewriting sequencer `_meta` emits only 3 optional sections, losing the replica-only flags until a new node rewrites. A *functional* regression (a primary may build an index it should skip), never on-disk corruption, and identical to the pre-existing covering/index-type optional-section evolution model → acknowledged tradeoff, not a defect.
- **`PostingGenLookup` seal-floor relaxation "empirically not formally justified"** - verified assertion-only (no production behavior change under `-ea` off); the added `PostingGenLookupSealFloorTest` pins the boundary. Noted as Minor risk-posture above.
- No pure false positives among the reported blockers; every Critical traced to reachable source. The back-compat + concurrency agent independently cleared the `_meta` bit-7 round-trip, sequencer optional-section framing, ALTER `extraInfo` layout, and the `roleGeneration` `AtomicLong` publish-before-bump memory ordering - concur with all four.

---

## Coverage map

| Behavioral change | Test | Verdict |
|---|---|---|
| CREATE TABLE `REPLICA ONLY` (bitmap/posting/covering) + parse error | `CreateTableReplicaOnlyIndexTest.*` | TESTED |
| ALTER ADD INDEX `REPLICA ONLY`; DROP clears flag | `AlterTableAddReplicaOnlyIndexTest.*` | TESTED |
| Write-path skip on skipping primary (insert/O3/parquet<->native/reindex/truncate/drop-index/symbol-cap/split-squash/change-type/posting-seal) | `TableWriterReplicaOnlySkipTest.*` (skip=true) | TESTED |
| No-skip node builds index | `TableWriterReplicaOnlyNoSkipTest` | TESTED |
| `isColumnIndexActive` truth table | `IsColumnIndexActiveTest` | TESTED |
| Planner full-scan fallback | `ReplicaOnlyIndexPlannerTest.*` (skip=true) | TESTED |
| Reader missing-file tolerance + open-race + negative control | `ReplicaOnlyIndexReaderToleranceTest.*` | TESTED |
| Reconcile build-on-demote / purge-on-promote / cached-reader / pooled-writer | `ReplicaOnlyReconcileTest.*` | TESTED |
| Checkpoint-rebuild gate | `SnapshotReplicaOnlyRebuildTest.*` | TESTED |
| `_meta` bit-7 persistence + cold reload | `TableUtilsReplicaOnlyTest`, `SequencerReplicaOnlyMetadataTest` | TESTED |
| Sequencer optional-section absent/bad-checksum/count-mismatch | `SequencerReplicaOnlyMetadataTest.*` | TESTED |
| ALTER `extraInfo` old-3-slot vs new-4-slot | `AlterOperationAddIndexFormatTest.*` | TESTED |
| SHOW CREATE TABLE re-emit + `table_columns().indexReplicaOnly` | `ReplicaOnlyIndexIntrospectionTest.*` | TESTED (skip=true introspection not asserted) |
| Multi-shape reconcile fuzz under role flips | `ReplicaOnlyIndexFuzzTest`, `...MultiColumnFuzzTest` | TESTED but **quarantines C3** |
| **ATTACH PARTITION on replica-only column (skip + non-skip, `_dmeta` present & absent)** | **none** | **UNTESTED → C1, C2** |
| **Non-replica-only POSTING `.pk`-absent query after convert/drop/re-add churn** | quarantined in `...MultiColumnFuzzTest` | **UNTESTED → C3** |

Coverage map: 16 behavioral changes, 14 tested, 2 UNTESTED (the attach path; the quarantined posting crash). Minor gaps: `table_columns()`/`SHOW COLUMNS` on a skipping primary (user sees `indexed=true` while physically absent) is asserted only at skip=false; `changeColumnType` replica-only-survival path is documented unreachable via SQL.

---

## Summary

**Verdict: request changes.** Four Critical findings block approval.

- **Correctness/performance gate:** fails - C1 (table-suspending ATTACH), C2 (invariant-violating index materialization + wasted IO on a skipping primary), and C3 (a shipped, deterministic, reachable query-time POSTING-reader crash) are all reachable on the feature's target deployment. C1/C2 are the exact failure class the prior review already found+fixed elsewhere (C1/C2/M1/split-squash) - the ATTACH path was simply missed and has zero test coverage. The fixes are mechanical: mirror the `copyOrRebuildColumnIndexes`/`rebuildPartitionIndexFiles` gate + tolerance.
- **Test gate:** fails - the attach behaviors are UNTESTED (regression tests are writable in OSS via the skip=true config factory the suite already uses), and C3 is a fix PR's own fuzz test masking a confirmed bug via a `strictlyCompleted >= 3` allowance rather than fixing it or making the risk-acceptance explicit. C4 weakens the result-assertion battery on the exact new behavior.
- In-diff / out-of-diff / pre-existing split: C1/C2 are missed-pattern defects in changed-adjacent code (the PR gates every *other* physical-index touchpoint but not attach); C3 is pre-existing-in-posting-subsystem but unmasked and quarantined by this PR; C4 is in-diff test code.
- Findings: ~14 draft → 4 Critical + 5 Moderate + 2 Minor verified; 0 pure false positives; 3 agent "residual risks" reclassified as enterprise-integration items or documented format-evolution tradeoffs.
- Not blocking: do **not** split this PR - the expectation is to fix C1-C4 (each with a regression test) on this branch and re-run CI. The design is strong; these are gaps at the edges of an otherwise carefully-gated change.

bluestreak01 (changes requested):
@nwoolmer

Reviewing PR #7340 at level 3.

## Critical

1. **[in-diff] Reconciliation treats one key file as proof that every partition has a complete index**

   `core/src/main/java/io/questdb/cairo/TableWriter.java:12339-12350,12450-12470`

   `replicaOnlyIndexFilesPresent()` checks only the last partition and only `.k`/`.pk`, not `.v`/`.pv`. An older partition can lack sidecars—or the last partition can retain `.pk` while losing `.pv`—and reconciliation still declares the index materialized.

   The planner then selects the index, and `TableReader.java:1095-1119` repeatedly raises “replica-only index not materialized.” Writer reopen and ordinary WAL apply repeat or skip the same probe, so they do not repair it.

   **Fix:** validate every partition and every required live sidecar, or maintain authoritative per-partition materialization state. Add older-partition and `.pk`-present/`.pv`-missing regressions.

2. **[in-diff] ATTACH accepts an unusable index on non-skipping nodes**

   `core/src/main/java/io/questdb/cairo/TableWriter.java:4506-4518`

   The new condition returns when a replica-only key is missing even when `skipReplicaOnlyIndexes()==false`. ATTACH then publishes the partition without rebuilding its index. If it is older than the current last partition, finding 1 prevents later reconciliation from detecting it.

   **Fix:** tolerate absent sidecars only on skipping nodes. Non-skipping nodes must rebuild before publishing ATTACH, or record a mandatory per-partition rebuild. Test an older `_dmeta`-less partition on a non-skipping node.

3. **[in-diff] Failed purge is acknowledged, allowing a stale index to return incomplete rows**

   `core/src/main/java/io/questdb/cairo/TableWriter.java:12404-12425,12666-12682,1595-1600`  
   `core/src/main/java/io/questdb/cairo/idx/PostingIndexUtils.java:1422-1470`

   Purge ignores `removeAllSealedFiles()`’s retry result, while `removeFileOrLog()` logs key deletion failure and returns normally. The writer nevertheless records the generation as reconciled.

   Rows written during the skipping window do not update the surviving index. On promotion, the surviving key makes reconciliation accept the stale index, so indexed queries can omit those rows.

   **Fix:** propagate deletion failure, do not acknowledge the generation until absence is verified, and add a failure-injected unlink/retry test.

4. **[out-of-diff callsite] Cached and prepared query plans retain the old role**

   `core/src/main/java/io/questdb/griffin/WhereClauseParser.java:180,2220`  
   `core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java:5076,6532,8609,10228,10726,10819`  
   `core/src/main/java/io/questdb/cairo/CairoEngine.java:556-573,2330-2342`  
   Callers: `JsonQueryProcessor.java:191-204`, `PGPipelineEntry.java:1669-1693`

   The changed contract makes index eligibility role-dependent, but factories capture that role only during compilation. `bumpRoleGeneration()` neither changes table metadata versions nor invalidates factories. A factory compiled on a replica remains an index plan after promotion purges the sidecars.

   `TouchTableFunctionFactory.java:76,88-99,179-180` has the same defect explicitly: it stores the role flag for the function’s lifetime.

   **Fix:** include role generation in factory validity, invalidate all cached/prepared factories on a role change, or defer eligibility to cursor execution. Add factory-reuse tests across both role transitions.

5. **[in-diff] Sequencer “checksum” does not protect the new payload**

   `core/src/main/java/io/questdb/cairo/wal/WalUtils.java:431-438`  
   `core/src/main/java/io/questdb/cairo/wal/seq/SequencerMetadata.java:507-525`

   The stored value covers only the pre-existing schema checksum plus a salt. It does not cover the replica-only count or flag bytes. The reader also interprets every nonzero byte as `true`.

   A payload bit flip can silently change role/index semantics while the checksum remains valid.

   **Fix:** checksum the version, count, and payload; accept only `0` and `1`; retain absent-section compatibility. Add payload-bit-flip and noncanonical-byte tests.

6. **[in-diff] The fuzz suite deliberately converts a deterministic production failure into success**

   `core/src/test/java/io/questdb/test/cairo/ReplicaOnlyIndexMultiColumnFuzzTest.java:132-195,220-240`

   Seed `0xABCD` deterministically reaches a normal POSTING index whose `.pk` is absent while the planner selects it. The test catches that critical exception, terminates the seed early, and reports success without running the remaining operations or final assertions.

   **Fix:** repair the POSTING lifecycle defect and remove the quarantine. This confirmed bug must be fixed in this PR, not deferred to an owner.

7. **[in-diff] Both fuzz oracles may skip query correctness indefinitely**

   `ReplicaOnlyIndexFuzzTest.java:265-303`  
   `ReplicaOnlyIndexMultiColumnFuzzTest.java:543-587`  
   `ReplicaOnlyIndexTestUtils.java:105-110`

   The tests accept any exception whose text contains `replica-only index not materialized`; they do not verify non-critical errno or eventual recovery. Their final check only proves that some matching sidecar exists in some partition, not that every partition is complete, and never reruns the filtered-query oracle.

   **Fix:** assert the exact non-critical error contract, force reconciliation, then require the query to return and match the reference table. Verify all required sidecars per partition.

8. **[in-diff] Reconciliation uses avoidable quadratic bookkeeping**

   `TableWriter.java:12330-12425,12434-12447`  
   `TableWriter.java:9251-9265`

   For `R` replica-only columns and `A` absent columns, list membership costs `O(R*A)`; rebuilding `denseIndexers` after every changed index costs `O(K*C)`. Writer opening also duplicates existence probes for present indexes.

   Column counts are not structurally bounded by a small constant, so the cold-path exception for performance findings does not apply.

   **Fix:** maintain a replica-only count/list, collect presence once, consume it directly, and rebuild `denseIndexers` once after the loop. Target `O(C+R)` bookkeeping.

9. **[pre-existing] Parquet snapshot restore performs `O(P*T²)` metadata work**

   `core/src/main/java/io/questdb/cairo/TableSnapshotRestore.java:566-570,1579-1671`

   Three passes repeatedly call a linear parquet-column lookup. The PR’s new gates are placed after that lookup, so even skipped replica-only columns pay for it.

   **Fix in this PR:** build one writer-ID-to-parquet-index map per partition, gate before lookup, and collect all scratch lists in one pass. Existing `O3PartitionJob` and `PageFrameMemoryPool` demonstrate the appropriate `IntIntHashMap` pattern.

10. **[in-diff] New tests reintroduce the banned `assertSql` surface**

    `TableWriterReplicaOnlySkipTest.java:157-163`  
    `TableWriterReplicaOnlyNoSkipTest.java:57-62`

    The local wrappers eventually invoke the builder, but the review standard explicitly prohibits new query-result helpers named `assertSql`. Blanket `sizeMayVary()` and inferred properties also weaken per-query factory assertions.

    **Fix:** use direct `assertQuery(...).returns(...)` calls with the precise factory assertions each query warrants.

11. **[in-diff] Fourteen behavioral rows remain untested**

    The missing regression coverage includes:

    - ASOF/index hints, DISTINCT posting, multi/prefix LATEST BY, indexed ORDER BY, and window metadata copies.
    - `UPDATE` and `touch_table()`.
    - Stale factory reuse across role changes.
    - Older/partial sidecars and failed purge.
    - Non-skipping ATTACH.
    - Parquet snapshot restore.
    - `changeColumnType()` propagation/build gate.
    - Actual SQL `NULL`.
    - Strict fuzz convergence.
    - Role-publication/concurrent WAL apply.
    - Malformed ALTER `REPLICA`.
    - Truncated/corrupted sequencer payload.

    These are user-visible SQL, error, concurrency, storage-lifecycle, or bug-fix paths, so the coverage gate classifies them as Critical.

## Moderate

1. **The type-change regression test does not exercise the claimed change**

   `TableWriterReplicaOnlySkipTest.java:587-622` converts indexed `SYMBOL` to unindexed `VARCHAR`. `indexType` is therefore `NONE`, making both new production branches at `TableWriter.java:1300-1303,1322-1323` irrelevant. The test passes with those hunks reverted.

   This is behaviorally and performance-wise inert today because the reachable conversion always produces an unindexed replacement, but the test must stop claiming it pins those branches.

2. **The defensive ATTACH rebuild test is mutation-ineffective**

   `TableWriterReplicaOnlySkipTest.java:669-708` explicitly documents that it passes without the production gate because reindexing targets a path that does not exist yet. The gate is currently behaviorally inert, so this is test-maintainability rather than a runtime defect.

3. **PR metadata is incomplete**

   The title needs a user-facing verb, for example:  
   `feat(core): add REPLICA ONLY indexes to offload primary index maintenance`

   The body must disclose the relaxed posting assertion, transient query failures, full-scan tradeoff, deterministic quarantined POSTING failure, and lost fuzz coverage. The statement “All green” is misleading while a deterministic exception is accepted.

   This is behaviorally and performance-wise inert because it changes metadata only.

## Minor

- Boolean names repeatedly omit `is...`/`has...`; representative examples are `CairoColumn.java:47`, `TableWriter.java:896`, and `WhereClauseParser.java:132`.
- `TableWriter.java:361-367` places `replicaOnly...` fields before `lifecycleManager`, violating alphabetical ordering.
- Added tests contain 74 unseparated large numeric literals across 41 lines; for example `1000000` and `86400000000` in `TableWriterReplicaOnlySkipTest.java`.
- `ReplicaOnlyIndexMultiColumnFuzzTest.java:236` prints during a successful, quarantined run.
- Internal “Task 12/13” references remain in production/test comments; `TableWriterReplicaOnlySkipTest.java:53` even says the planner guard has not landed although this PR contains it.

## Downgraded

- **Concurrent generation bump can lose a transition:** dismissed; sampling the older generation causes the next invocation to reconcile again.
- **Changed `MetadataCacheTest.returnsOnce()` calls are deterministic misuse:** dismissed; concurrent table rename/ALTER makes the result genuinely unstable.
- **Any surviving posting value file causes stale-index acceptance:** narrowed; the dangerous case requires the authoritative `.pk`/`.k` to survive.
- **Protocol true-value propagation is untested:** dismissed; `AlterOperationAddIndexFormatTest` verifies `replicaOnly=true` through serialization.
- **The PR introduced the whole quadratic snapshot algorithm:** dismissed; the quadratic traversal is pre-existing, although the PR repeats unnecessary lookups inside it.
- **Writer open always performs two full scans:** narrowed; empty tables and some skipping states return early.
- **Commit history must be rewritten:** dismissed. This repository squash-merges PRs; only the final squash message must follow commit-message rules.

## Coverage map

Legend: H=happy path, E=error path, N=NULL, B=boundary, C=concurrency, R=resource cleanup. For parser/metadata rows, N/C/R are N/A because they do not consume row values, mutate shared runtime state, or own resources.

| # | Change | Test and failure link | Dimensions | Verdict |
|---|---|---|---|---|
| 1 | CREATE grammar and missing `ONLY` | `CreateTableReplicaOnlyIndexTest` create variants and `testReplicaWithoutOnlyErrors`; parser/model regression changes metadata or error assertion | H✓ E✓ B✓ N/C/R— | TESTED |
| 2 | ALTER positive propagation | `AlterTableAddReplicaOnlyIndexTest.testAlterAddReplicaOnlyIndex`; metadata assertion loses flag on regression | H✓ B✓ E✗ N/C/R— | TESTED |
| 3 | Malformed ALTER `REPLICA` | No ALTER equivalent of the CREATE malformed-input test | E✗ B✗ N/C/R— | **UNTESTED** |
| 4 | `_meta` bit/cache propagation | `TableUtilsReplicaOnlyTest.testReplicaOnlyFlagRoundTrips`; read-back flag fails | H✓ B✓ N/C/R— | TESTED |
| 5 | ALTER `extraInfo` compatibility | Six `AlterOperationAddIndexFormatTest` methods; old/new slot assertions fail | H✓ E✓ B✓ N/C/R— | TESTED |
| 6 | Sequencer valid/absent/bad-sentinel/count mismatch | Five `SequencerReplicaOnlyMetadataTest` methods; reload/fallback assertions fail | H✓ E✓ B✓ N/C/R— | TESTED |
| 7 | Payload corruption/truncation/noncanonical bytes | Existing tests do not mutate payload under a valid sentinel | E✗ B✗ N/C/R— | **UNTESTED** |
| 8 | WHERE, one-key LATEST BY, ordinary alias fallback | Three `ReplicaOnlyIndexPlannerTest` methods; plan/result assertions fail | H✓ B✓ E— N✗ C/R— | TESTED |
| 9 | ASOF, DISTINCT, multi/prefix LATEST, ORDER BY, window copies | No skipped-replica tests reach these branches | H✗ B✗ N✗ C/R— | **UNTESTED** |
| 10 | Skip build across WAL/non-WAL/O3/conversion/reindex/truncate/drop | Focused `TableWriterReplicaOnlySkipTest` methods; suspension/file/result assertions fail | H✓ E✓ B✓ N partial C— R✓ | TESTED |
| 11 | UPDATE and `touch_table()` | No skipped-replica tests | H✗ E✗ N✗ B✗ C/R— | **UNTESTED** |
| 12 | Sequential reconcile build/purge/open/cached reader | Five `ReplicaOnlyReconcileTest` methods; files/result/criticality assertions fail | H✓ E✓ B✓ C partial R✓ N✗ | TESTED |
| 13 | Factory reuse across role changes | No test reuses one compiled factory across promote/demote | H✗ E✗ C✗ B✗ N/R— | **UNTESTED** |
| 14 | Older/partial sidecar repair | No older-partition or `.pk`-present/`.pv`-missing test | H✗ E✗ B✗ R✗ N/C— | **UNTESTED** |
| 15 | Purge deletion failure | No failure-injected unlink/retry test | E✗ C✗ R✗ B✗ H— N— | **UNTESTED** |
| 16 | Missing-metadata ATTACH on skipping node | `testAttachPartitionReplicaOnlyIndexMissingMetadataDoesNotRequireIndexFiles`; pre-fix attach throws | H✓ E✓ B✓ R✓ N/C— | TESTED |
| 17 | Missing-sidecar ATTACH on non-skipping node | No older-partition non-skipping test | H✗ E✗ B✗ R✗ N/C— | **UNTESTED** |
| 18 | Reader total absence/corruption/open race/cached reload | Reader-tolerance and reconcile tests; criticality/reload assertions fail | H✓ E✓ B✓ C✓ R✓ N— | TESTED |
| 19 | Partial posting/covering sidecars | No `.pk` present with missing `.pv`/`.pc` test | E✗ B✗ R✗ H— N/C— | **UNTESTED** |
| 20 | Native snapshot restore gating | Two `SnapshotReplicaOnlyRebuildTest` methods; file-presence assertion fails | H✓ E✓ B✓ R✓ N/C— | TESTED |
| 21 | Parquet snapshot restore gating | Native-only tests do not reach three parquet loops | H✗ E✗ B✗ R✗ N/C— | **UNTESTED** |
| 22 | Split/squash role-aware frames | `testSplitAndSquashReplicaOnlyIndexDoesNotMaterializeOrDistress`; split count/files/results fail | H✓ E✓ B✓ R✓ N partial C— | TESTED |
| 23 | `changeColumnType()` propagation/build gate | Existing test passes with production hunks reverted | H nominal; no failure link; N/B/C/R✗ | **UNTESTED** |
| 24 | Posting seal floor assertion | Three `PostingGenLookupSealFloorTest` methods; exact/deeper/negative boundaries fail under `-ea` | H✓ E✓ B✓ N/C/R— | TESTED |
| 25 | SHOW CREATE / `table_columns()` | Three `ReplicaOnlyIndexIntrospectionTest` methods; emitted schema/result fails | H✓ B✓ N/C/R— | TESTED |
| 26 | Actual SQL NULL across roles | Tests use `"null-ish"` or implicit column tops, not explicit NULL role comparisons | N✗ H✗ B✗ C/R— | **UNTESTED** |
| 27 | Fuzz correctness and convergence | Both fuzz suites can skip query or terminate a seed while passing | H nominal; E✗ B✗ C✗ R✗ N✗ | **UNTESTED** |
| 28 | Role publication and concurrent WAL apply | Only sequential flips are tested | C✗ E✗ B✗ H partial N/R— | **UNTESTED** |
| 29 | Defensive ATTACH rebuild gate | Verified behaviorally inert because downstream reindex currently no-ops | All N/A: no behavioral delta | EXEMPT |

## Summary

**Verdict: request changes.**

- Coverage map: **28 behavioral changes: 14 tested, 14 UNTESTED**, plus 1 verified no-behavior exemption.
- Verified findings: **16**; false positives removed: **7**.
- Split: **14 in-diff, 1 out-of-diff callsite, 1 pre-existing**.
- Principal regressions: stale or partial indexes can become planner-visible, cached plans retain the previous role, failed purge can produce incomplete results, and one deterministic production failure is deliberately quarantined.
- Tradeoff not adequately disclosed: skipping index maintenance reduces primary write cost but introduces primary full scans and currently exposes transient or permanent query failures during incomplete materialization.
- My targeted Maven run could not reach execution because the current checkout failed test compilation on missing QWP/HTTP symbols. I did not use that environment failure to dismiss any finding.

nwoolmer:
@bluestreak01 Thanks for the follow-up review. I worked through the reported correctness failures and committed the fixes as `583df9b2` (`fix(core): harden replica-only index lifecycle`).

## Actions taken

1. **Whole-table materialization was inferred from one key file**
   - Replaced the last-partition/key-only probe with validation of every partition.
   - BITMAP materialization now requires both its `.k` key file and `.v` value file.
   - POSTING requires its `.pk` key file and the matching live `.pv` value generation.
   - Covering POSTING additionally validates every required `.pc<N>` covered-column sidecar.
   - Added a regression that removes only an older partition's index and verifies open-time reconciliation repairs it before indexed queries run.

2. **Non-skipping ATTACH could publish an unusable replica-only index**
   - Missing sidecars are tolerated during detached-path validation only because that path is not yet a reliable rebuild target.
   - After the partition reaches its final table path, a non-skipping node synchronously rebuilds missing replica-only indexes before publishing the attachment.
   - Skipping nodes continue to omit them intentionally.

3. **Failed purge could be acknowledged, leaving stale planner-visible files**
   - Index removal now reports failure when any posting generation, covering sidecar, key, or bitmap value file survives unlink.
   - Reconciliation throws and does not advance `lastReconciledRoleGen` when purge is incomplete, so the operation remains retryable rather than accepting stale state.

4. **Role-sensitive cached query state**
   - `bumpRoleGeneration()` now publishes the standard query-cache flush event after the role-generation release.
   - Added a runtime defensive check in `TableReader`: a cached reader whose physical implementation no longer matches current metadata (POSTING vs BITMAP) is closed and recreated.
   - Metadata column transitions now close cached index readers instead of transferring potentially stale reader implementations between shifted dense slots.

5. **Sequencer replica-only payload was not checksummed**
   - The checksum now covers the section count and every flag byte.
   - The reader accepts only canonical `0` and `1` values and applies the section only when count, payload, and checksum all agree.
   - Added payload-bit-flip and noncanonical-byte regressions while retaining absent-section/count-mismatch compatibility tests.

6. **Deterministic `.pk` failure was quarantined in fuzz**
   - Root cause was not an absent physical index: after DROP/re-ADD and dense-column shifts, a cached POSTING reader could occupy a BITMAP column's slot and reopen that column using `.pk` rather than `.k` filenames.
   - Fixed reader lifecycle/type validation as described above.
   - Removed the exception quarantine, accepted-seed threshold, reproducer parser, risk-acceptance comments, and success-path `System.out` output.
   - All four fixed seeds now execute all 500 operations and all final correctness/file-presence assertions strictly, including `0xABCD`.

7. **Absent/all-top columns across parquet/native lifecycle**
   - A re-added column absent from an older parquet schema now receives a valid all-NULL index rather than metadata being published with no physical index.
   - Parquet-to-native rebuild indexes the NULL prefix and physical suffix correctly.
   - Native-to-parquet conversion preserves valid empty/all-top sidecars.
   - Added focused DROP/re-ADD, ADD POSTING, parquet/native conversion, TRUNCATE, metadata reload, and indexed-query coverage.

8. **Rebuild error-path indexer consistency**
   - `denseIndexers` is repopulated immediately after a stale indexer is removed, so a failed rebuild cannot leave it referencing a closed indexer.

## Validation

Passed locally:

- strict `ReplicaOnlyIndexMultiColumnFuzzTest` (4 seeds x 500 operations; no quarantine)
- `TableWriterReplicaOnlySkipTest`
- `ReplicaOnlyReconcileTest`
- `SequencerReplicaOnlyMetadataTest`
- focused ATTACH and parquet/POSTING lifecycle regressions
- `mvn -pl core -DskipTests package`
- `git diff --check`

The commit will be pushed to `feat/index-skip-primary` immediately after this response.

nwoolmer:
Addressed the 2026-07-11 (level 3) round. After merging master and re-baselining, most items were already closed by the two post-review commits (`583df9b2`, `393771a6`); the genuinely-open ones are handled in `01bce242`.

**Correctness**

- **C2 (ATTACH on a non-skipping node)** — was still broken: the post-rename self-heal called an `EMPTY`-rooted `IndexBuilder` with the already-freed detached `ColumnVersionReader`, so it silently no-oped and a reattached older partition stayed index-less (planner picks the index, reader raises "not materialized" indefinitely). Now rebuilt synchronously at the FINAL table path via a dedicated writer-lifetime `IndexBuilder` driven by `columnVersionWriter`, before the partition is published. Regression: `ReplicaOnlyReconcileTest#testNonSkippingAttachRebuildsMissingOlderPartitionIndex` (fails → passes).
- **C4 (touch_table role capture)** — `touch_table()` now reads `skipReplicaOnlyIndexes()` live at execution instead of capturing it at construction. (The cached/prepared SELECT plan-cache path was already covered by `RoleGenerationRecordCursorFactory` in `393771a6`.)
- **C1, C3, C5, C8, C9** — verified already fixed in the post-review commits (per-partition/all-sidecar presence probe; propagated purge-deletion failure gated on verified absence; checksummed sequencer payload with canonical 0/1 + absent-section compat; snapshot-once / rebuild-`denseIndexers`-once; single-pass parquet restore with an `IntIntHashMap` gate-before-lookup). No further change needed.

**Tests / coverage**

- **C7** — both fuzz oracles now assert the degradation error is **non-critical** and, at the post-reconcile quiescent point, require the filtered query to **succeed and match the index-free reference for every partition** (strict per-partition completeness, not "any sidecar exists").
- **C6** — confirmed the `0xABCD` quarantine is gone; the seed now runs to completion (500 ops) green under the strengthened oracle.
- **C10** — confirmed the two writer tests use fluent `assertQuery(...).returns(...)`; no `assertSql`/blanket `sizeMayVary()`.
- **C11** — added: non-skipping ATTACH rebuild, non-skip `touch()` reads the active index, multi-key `LATEST BY` full-scan on a skipping primary.
- **Moderate** — softened the `changeColumnType` and skip-node ATTACH test comments that over-claimed which production branches they pin (the reachable SYMBOL→unindexed conversion / the skip-node no-op are genuinely inert; the effective side is now covered by the non-skip ATTACH test above).
- **Minors** — dropped stale internal "Task NN" references from production/test comments; fixed the alphabetical field ordering; verb-first title; body now discloses the relaxed posting seal-floor assertion, the transient non-critical degradation window, the primary full-scan tradeoff, and the #7349 posting-hardening scope, and drops "All green".

Local run (surefire authoritative): `TableWriterReplicaOnlySkipTest` 19, `TableWriterReplicaOnlyNoSkipTest` 2, `ReplicaOnlyReconcileTest` 12, `ReplicaOnlyIndexPlannerTest` 7, `ReplicaOnlyIndexReaderToleranceTest` 4, `ReplicaOnlyIndexFuzzTest` 1, `ReplicaOnlyIndexMultiColumnFuzzTest` 1 — all green; `TouchTableFunctionTest` 6 and `AlterTableAttachPartitionTest` 80 pass (no regressions).

Note on **posting `.pk`-absent naming boolean minors** (`is/has`) and the numeric-separator churn: left as-is to keep the diff focused; happy to fold those in if you'd like them in this PR.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 571 / 639 (89.36%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/wal/seq/SequencerMetadataService.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/RoleGenerationRecordCursorFactory.java|21|65|32.31%|
|:large_blue_circle:|io/questdb/cairo/sql/RecordMetadata.java|2|3|66.67%|
|:large_blue_circle:|io/questdb/cairo/CairoColumn.java|3|4|75.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowColumnsRecordCursorFactory.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/griffin/SqlParser.java|19|22|86.36%|
|:large_blue_circle:|io/questdb/cutlass/parquet/ParquetExportMode.java|9|10|90.00%|
|:large_blue_circle:|io/questdb/cairo/TableSnapshotRestore.java|30|32|93.75%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|269|281|95.73%|
|:large_blue_circle:|io/questdb/cairo/wal/seq/SequencerMetadata.java|19|20|95.00%|
|:large_blue_circle:|io/questdb/cairo/TableReader.java|36|37|97.30%|
|:large_blue_circle:|io/questdb/cairo/frm/file/FrameImpl.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalWriterMetadata.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableReaderMetadata.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/seq/TableSequencerImpl.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/SQLSerialParquetExporter.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/WhereClauseParser.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/TableMetadata.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriterMetadata.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableColumnMetadata.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|26|26|100.00%|
|:large_blue_circle:|io/questdb/cairo/frm/file/FrameFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|34|34|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractRecordMetadata.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlKeywords.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/cairo/IndexBuilder.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowCreateTableRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/AlterOperationBuilder.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/TouchTableFunctionFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingGenLookup.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableUtils.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalUtils.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/ExportQueryProcessor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CreateTableOperationImpl.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableStructure.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/UpdateOperatorImpl.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/model/CreateTableColumnModel.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/pool/SequencerMetadataPool.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/AlterOperation.java|2|2|100.00%|

PR #7144, opened 2026-05-23
feat(qwp): add deferred commit for transactional ingestion

bluestreak01:
Fixes #7143

Tandem with questdb/java-questdb-client#32.

## Summary

Adds `FLAG_DEFER_COMMIT` (`0x01`) to the QWP ingress protocol. When a
client sets this flag on a message, the server appends rows to WAL
writers but skips the commit. The commit fires on the next message
without the flag. This allows clients to split a logical batch across
multiple messages that individually fit within the server's recv buffer,
removing the hard coupling between `http.recv.buffer.size` and maximum
transaction size.

### Server changes

- `QwpConstants`: add `FLAG_DEFER_COMMIT = 0x01`.
- `QwpIngressProcessorState`: add `isDeferCommit()` to read the flag
  from the message header, and `clearMessageState()` to reset
  per-message parsing state without rolling back WAL rows.
- `QwpIngressUpgradeProcessor.handleBinaryMessage()`: skip `commit()`
  when `FLAG_DEFER_COMMIT` is set; call `clearMessageState()` instead
  of `clear()` in the finally block so accumulated WAL rows survive.
  On error, the full `clear()` still fires to roll back all deferred
  rows.

### Client changes (java-questdb-client submodule)

- `QwpConstants`: add `FLAG_DEFER_COMMIT = 0x01`.
- `QwpWebSocketEncoder`: add `setDeferCommit()` to set/clear the flag.
- `QwpBufferWriter` / all implementations: add `patchByte()`.
- `QwpWebSocketSender.flushPendingRows(boolean deferCommit)`: when
  `deferCommit` is true, the encoded message carries `FLAG_DEFER_COMMIT`.
  When the batch exceeds `serverMaxBatchSize`, `flushPendingRowsSplit()`
  sends each table as a separate deferred message with the last one
  committing (or deferring, if the caller requested defer).
- `QwpWebSocketSender.sendCommitMessage()`: sends an empty QWP message
  (0 table blocks, no defer flag) to trigger the server-side commit
  when all pending rows were already sent via deferred auto-flushes.
- `Sender` builder: add `transaction=on|off` connect-string key and
  `transactional(boolean)` builder method. When enabled, auto-flush
  sends with `FLAG_DEFER_COMMIT` and only explicit `flush()` triggers
  the commit.

### Protocol spec update

- Updated the flags byte table (bit 0 is now `FLAG_DEFER_COMMIT`).
- Added "Deferred commit" section covering wire semantics, error
[...]

Issue #10, opened 2014-08-23
Inherited fields support for POJOs


ValeryNo:
Hi guys, first of all - thanks for your hard work!

It would be great if you can support inherited class fields for POJOs.
Here is my case - in addition to raw trades data I want to store pre-calculated 1/3/5/60 mins quotes.
Currently, JournalMetadataBuilder.parseClass uses getDeclared fields method to find class fields, so there is no way I can write 1 QuoteInterval class with all required fields/methods and inherit all 1,3,5,60 minutes ones from it, I'll have to maintain an exact copy of the same logic 4 times.

I think such support would be beneficial. Though, in my particular case, if I would be able to partition data by interval field and query it somehow by key + interval value that would make my life easier. But as far as I understand at this point I can only fetch a slice of data based on from-to interval, but there is no way to fetch data by a partition based on other custom criteria (like interval value in my case, which would be int representing # of minutes).

As for my proposed inherited fields request here are the changes I made in my local NFSDB copy for now.
Sorry, I haven't used patch util for a while, so posting my changes as a text here. They're super easy anyway.

```
private void parseClass() throws JournalConfigurationException {
```
#1

```
    //Field[] classFields = modelClass.getDeclaredFields();
    List<Field> classFields = getAllFields(modelClass);
```
#2

```
    //this.nameToIndexMap = new TObjectIntHashMap<>(classFields.length, Constants.DEFAULT_LOAD_FACTOR, -1);
    this.nameToIndexMap = new TObjectIntHashMap<>(classFields.size(), Constants.DEFAULT_LOAD_FACTOR, -1);
```
#3 - recursive fields list calculation

```
public static List<Field> getAllFields(Class<?> type)
{
    return getAllFields(new LinkedList<Field>(), type);
}

private static List<Field> getAllFields(List<Field> fields, Class<?> type)
{
    Collections.addAll(fields, type.getDeclaredFields());

    if (type.getSuperclass() != null)
    {
        fields = getAllFields(fields, type.getSuperclass());
    }

    return fields;
}
```

PR #7341, opened 2026-06-29
feat(qwp): add a connect timeout and tolerant client startup

bluestreak01:
Tandem PRs:
- Client: https://github.com/questdb/java-questdb-client/pull/60
- Enterprise: https://github.com/questdb/questdb-enterprise/pull/1105

## Overview

This PR advances the `java-questdb-client` submodule to bring in an
application-level connect timeout for the client transports plus a round of
QuestDB-facade configuration work, and fixes the one parent-repo test that the
facade API change broke.

Since the original description, the branch has grown a second theme: the
server-side half of the client's failover contract. The QWP ingress now treats
an in-place demote as a TRANSIENT role change (reconnect-eligible close, never
a client-terminal `SECURITY_ERROR`), defers that close on durable-ack
connections until committed work is durably uploaded (exactly-once replay),
and the e2e error-path suites are migrated to the client's NACK policy v2.
See "Server-side QWP ingress" below.

The rest of the diff lives in the submodule (connect timeout, facade config,
Invariant B, NACK policy v2, M8–M11 hardening — see the client PR); the parent
repo carries the pointer bumps, the server-side ingress work, the e2e
migrations, and CI. The sections below describe what lands here.

## Connect timeout (the headline change)

Adds a real, cross-platform connect timeout for the HTTP and WebSocket (QWP)
transports. Until now a connect to a black-holed or firewalled host blocked on
the OS-level TCP connect timeout (often 60-120s), because the socket was created
blocking and only switched to non-blocking *after* connect, and no knob existed
to clamp it.

- A new native primitive switches the socket to non-blocking *before* connect,
  so `connect()` returns `EINPROGRESS` immediately, then polls for writability
  bounded by the caller's budget and confirms the outcome via `SO_ERROR`. A
  distinct return code (`CONNECT_TIMEOUT`, -3) lets the Java layer raise a
  timeout-flagged exception rather than decode errno.
- Implemented for POSIX (`share/net.c`) and Windows (`windows/net.c`).
- Threaded through `HttpClientConfiguration.getConnectTimeout()`,
  `HttpClient.connect()`, `WebSocketClient`, `QwpQueryClient`,
  `QwpWebSocketSender`, and the `Sender` builder, with a `connect_timeout`
  connect-string key (legacy http and ws/wss parsers) and a `ConfigSchema`
  COMMON entry.
- Default is unset (0 = OS fallback): behavior is unchanged unless
  `connect_timeout` is configured.
- The same client bump also bounds the **TLS handshake** and stops it
  busy-spinning. `JavaTlsClientSocket.startTlsSession` ran the handshake on the
  non-blocking socket with raw `recv`/`send` and no readiness wait, so a peer
  that completed TCP but stalled mid-handshake caused a 100% CPU spin with no
  deadline (`connect_timeout` covered only the TCP connect, and the upgrade's
  request/auth timeout never covered the handshake). The handshake now runs
  through the deadline-aware `ioWait`, bounded by `connect_timeout` and falling
  back to the request timeout when unset. Tradeoff: with `connect_timeout`
  unset, a `wss://` handshake that stalls now fails after the request timeout
  instead of spinning forever -- strictly better, but a behavior change for
  that edge.

## QuestDB facade configuration

- Single cluster config: collapses the dual ingest/query config surface into one
  `ws`/`wss` string for the whole cluster (one `addr` server list drives both the
  sender and query pools). This removes the `QuestDB.connect(ingest, query)`
  overload. The facade is the only caller of that overload; the single-arg
  `connect(config)` and `builder()` paths cover the same ground.
- `lazy_connect` for tolerant startup: a single connect-string flag that lets the
  handle start while the server is down (ingest connects async, the read pool
  defaults to `query_pool_min=0`), buffers writes meanwhile, and reads once the
  server is up (the read pool stays enabled and connects lazily on first query).
  Conflicting knobs that force a blocking/fail-fast startup
  (`initial_connect_retry` other than async, or an explicit `query_pool_min > 0`)
  are rejected up front with a clear remedy. This supersedes an earlier
  write-only mode explored on this branch, which permanently disabled reads.
- Ingest callbacks on the builder: `errorHandler(...)` and
  `connectionListener(...)` are now reachable from the pooled facade and threaded
  to every pooled `Sender` (internal recovery senders are deliberately excluded).
  Defaults are null, so behavior is unchanged unless a callback is set.

## Build, CI, and compatibility

- Native libraries (`.so`/`.dll`/`.dylib`) are no longer committed to the
  submodule; CI builds them on the test runners instead, including in the
  Coverage and JDK 8 legs.
- The Linux native build pins `clock_gettime` to `GLIBC_2.2.5` to keep the glibc
  floor low for broad Linux compatibility.

## Server-side QWP ingress: role changes are transient, not terminal (Invariant B)

Production incident: after `SWITCH ROLE TO REPLICA`, an existing QWP ingress
connection had its next batch rejected by the per-batch read-only re-check as
an authorization error → `SECURITY_ERROR` (wire 0x8). A store-and-forward /
durable-ack client classifies `SECURITY_ERROR` as a terminal HALT, so the
producer died on a routine failover. Fixes, in order:

- **Demote closes instead of NACKing** (`3abbb1cf76`): the read-only re-check
  now flags the connection and closes it with a reconnect-eligible
  `NORMAL_CLOSURE` (after flushing pending ACKs so the client learns what
  committed). The client reconnects, hits the 421 role gate on the
  now-replica, and retries from SF — the transient path it already handles. A
  genuine ACL denial on a writable node still maps to `SECURITY_ERROR`.
- **Mid-batch flip containment** (`86f8556c8e`): the read-only flag can flip
  BETWEEN the top-of-batch gate and the WAL-writer acquisition / commit;
  the deeper engine gates then throw `CairoException.authorization()`, which
  mapped back to `SECURITY_ERROR`. Batch-rejection errors are now routed
  through a handler that re-checks the live read-only state at catch time and
  converts the transient demote refusal into the role-change close.
- **Fuzzed across every refusal window** (`d5d2db9206`): an ingress fuzz
  drives the demote race across all refusal points and asserts
  `SECURITY_ERROR` never reaches a client during a role switch.
- **Durable-ack deferred close** (`3b6564603b`): the demote flips read-only
  FIRST and uploads AFTERWARDS, so the role-change close could sever a
  durable-ack connection before its final durable ack was deliverable — the
  client (whose replay watermark advances only on durable acks) then replayed
  a batch the promoted replica already had, landing it twice on tables
  without dedup. The close is now DEFERRED (bounded by a 10s stall guard)
  until the durable-upload registry covers the connection's pending seqTxns;
  the final durable ack is flushed and only then does the connection close —
  every in-flight batch lands exactly once. Non-durable-ack connections keep
  the immediate close.
- **`STATUS_NOT_WRITABLE` (0x0C) reserved** (`abbc04f204`): a wire status byte
  for "cannot accept writes" (read-only replica / demoting primary), so a
  future server can NACK mid-stream once deployed client fleets classify it
  as retriable-with-endpoint-rotation (client NACK policy v2
  `RETRIABLE_OTHER`).
- **E2e error-path migration to NACK policy v2** (`a8bec7b1ff`): the client
  bump removed the v1 drop policy, which the server-side e2e suites still
  encoded (32 failures). The suites now pin the v2 observables: retriable
  NACKs replay until the poison-frame detector escalates, `SCHEMA_MISMATCH`
  latches terminal on the first strike, a latched terminal surfaces on
  exactly one channel, and SF purge applies to clean producers only.

## Server-side QWP ingress: cumulative OK acks follow commit boundaries (deferred-commit ack hole)

The deferred-commit feature (#7144) let `FLAG_DEFER_COMMIT` frames fall into the pre-existing `STATUS_OK` ack branch: the cumulative-ack watermark advanced (and flushed — every `ACK_BATCH_SIZE` frames, before error responses, and on role-change closes) over rows that were appended but **not committed**. Any subsequent error, demote, or disconnect rolled those rows back after the store-and-forward client had already trimmed their replay slots: silent, unrecoverable loss of up to a full uncommitted transaction (`qwp.max.uncommitted.rows`, 1M rows/table default). This inverted #7144's own error-handling contract — "the client's store-and-forward replays all unacknowledged messages on reconnect" requires deferred frames to stay unacknowledged until their group commits.

Fix: the watermark advances only on frames whose processing ended in a successful `commitAll()`:

- OK deferred frames get no watermark advance and no ack; coverage arrives with the group-closing commit frame's cumulative ack, which also carries the group's real per-table seqTxns for durable-ack tracking (closing the "empty OK frames are trivially durable" hole for deferred groups).
- A group-closing commit flushes its ack eagerly (`hasPendingAck`) instead of waiting for the batch cadence — transaction confirmation must not stall behind unrelated future traffic.
- The per-table force-commit at the max-uncommitted-rows cap does **not** advance the watermark (it gives no full-coverage guarantee); those rows simply replay at-least-once after a reconnect.
- `QwpIngressProcessorState.setHighestProcessedSequence` grows a last-resort clamp mirroring the unresolved-sequence clamp: watermark motion over uncommitted deferred rows clamps and screams — an availability hit, never data loss.

Non-transactional traffic is byte-for-byte unchanged (every frame commits before its watermark advance; identical ack cadence). Client companion in java-questdb-client#60: `close()` drains to the last commit boundary, and a recovered orphaned deferred tail (producer killed mid-transaction) is aborted via skip-and-self-ack instead of replayed into the next session's commit.

Tests: state-level clamp tests; wire-level e2e regressions (`testDeferredFramesNotAckedUntilCommit` — 20 deferred frames, zero acks, group ack covers 0..20; `testGroupCommitAckFlushesEagerly` — sub-batch-threshold group acked on commit); demote fuzz/shape drivers re-synced to the new processor ordering.

## Design docs

Adds `design/qwp-client-startup-failover-behavior.md` and
`design/qwp-client-ergonomics-issues.md` documenting the client's
startup/failover behavior and the ergonomics rationale behind these changes.
The NACK policy v2 design lives in the client PR
(`design/qwp-nack-policy-v2.md`).

## Parent-repo change

- Bumps the `java-questdb-client` submodule pointer to the branch above — the
  bumps carry the connect-timeout/facade work, the Invariant B reconnect
  rework, NACK policy v2 + `max_frame_rejections`, and the M8–M11 hardening
  (drainer listener on the facade/builder, role-reject observability split,
  lock-free background connect walks, fast close under outage).
- Fixes `QuestDBFacadeE2ETest`, which still called the removed
  `connect(ingest, query)` overload. Both strings in that test were already
  identical, so they collapse to a single `connect(config)`; the test is renamed
  `testSingleConfigConnectRoundTrip` to reflect the one-cluster-config model.

## Parent-repo CI: build the client native library under local-client

Dropping the committed client binaries also broke the parent test pipeline.
With a `-SNAPSHOT` client, `detect-local-client.yml` activates the
`local-client` profile and the reactor builds the client from source, but
nothing compiled its native library, so every client-dependent test failed at
class load with `cannot find /io/questdb/client/bin/<platform>/libquestdb.so`
(self-hosted Linux `Other tests` and parts of `Cairo tests`, plus the
on-demand macOS/Windows legs that share the same steps).

- Adds `ci/templates/build-client-native.yml`, included by `steps.yml` right
  after JDK activation, which compiles the client's native library with CMake
  when `local-client` is active. It initializes the client's `zstd` submodule,
  builds with the agent toolchain, and copies the result into the client's
  `src/main/resources/io/questdb/client/bin/<platform>/` so it survives the
  test phase's `mvn clean` and resolves on the production resource path the
  loader checks (`io.questdb.client.std.Os`).
- The step is a no-op when the client comes from Maven Central (the release
  jar bundles the binaries), since `CLIENT_PROFILE` is then empty.
- Tradeoff: each test job that runs under `local-client` now spends the extra
  CMake build time (toolchain install is skipped when cmake/nasm are already
  present). The Linux and macOS paths are exercised directly; the Windows path
  only runs in the on-demand hosted pipeline and mirrors the client's
  mingw-based build.

## Tradeoffs and compatibility notes

- API change: the `QuestDB.connect(ingest, query)` overload is gone. Callers that
  used distinct ingest/egress endpoints must move to a single cluster config or
  the builder.
- The connect timeout adds a non-blocking-connect path in native code on both
  POSIX and Windows. The default keeps the prior OS-timeout behavior, so there is
  no change unless `connect_timeout` is set.
- Dropping committed native binaries shifts the cost of producing them onto CI;
  builds that previously relied on the checked-in artifacts now depend on the
  native build step succeeding.

## Test plan

- `mvn clean package -DskipTests -P local-client` builds cleanly (verified).
- `NetConnectTimeoutTest` covers loopback success, refused-vs-timeout
  disambiguation, and a black-hole timeout firing within budget.
- `QuestDBLazyConnectTest` covers start+write while the server is down, reads
  staying enabled, both startup conflicts (via connect string and builder), and
  flag parsing.
- `QuestDBServerRecoveryTest` dogfoods `lazy_connect=true` through the full
  down -> write -> up -> read lifecycle.
- `QuestDBFacadeCallbacksTest` asserts the facade-wired `errorHandler` and
  `connectionListener` receive ingest errors and connection events with no
  server.
- `QuestDBFacadeE2ETest.testSingleConfigConnectRoundTrip` exercises the
  single-config sender + query round trip end-to-end.

## Merge prerequisite

The corresponding `java-questdb-client` branch must merge first; the submodule
pointer here will be re-bumped to that merged commit before this PR merges.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved local-client builds so required native QuestDB libraries are compiled and packaged for Linux, macOS, and Windows, reducing startup failures from missing native dependencies.
  * Improved QWP demote/read-only race handling so client-visible refusals avoid `SECURITY_ERROR`, and role-change close is deferred safely with durable-ack grace to prevent ack/ordering issues.
* **Tests**
  * Refreshed and expanded end-to-end coverage for pooled query leasing, handler behavior, and concurrency isolation.
  * Added fuzz/invariant coverage for demote races and stronger “terminal error” escalation checks under NACK policy v2.

### DIFF
--- a/.claude/skills/review-pr/SKILL.md
+++ b/.claude/skills/review-pr/SKILL.md
@@ -146,7 +146,7 @@ Every agent receives:
 
 Launch the following agents in parallel.
 
-**Agent 1 — Correctness & bugs:** NULL handling, edge cases, logic errors, off-by-one, operator precedence, error paths. Cross-reference every changed symbol against its callsite inventory and verify the new behavior is correct at each callsite.
+**Agent 1 — Correctness & bugs:** NULL handling, edge cases, logic errors, off-by-one, operator precedence, error paths. Cross-reference every changed symbol against its callsite inventory and verify the new behavior is correct at each callsite. When the diff touches the QWP ingress / role-gating path, an in-place switch or failover, a `java-questdb-client` submodule bump, or store-and-forward test suites, also verify the "Store-and-forward & pool startup invariants" checklist — a change that lets a running SF drainer surface transport errors to the producer, imposes a reconnect time budget on it, or hard-fails it on a transient outage is a Critical (data-loss) finding.
 
 **Agent 2 — Concurrency:** Race conditions, shared mutable state, missing volatile, lock ordering, thread-safety of data structures. Use the implicit contract list (lock order, thread-affinity) and check every callsite from 2.5b for violations of the new contract.
 
@@ -371,6 +371,82 @@ For each new or changed allocation site, verify:
 - **Same tracker for malloc and its matching free.** A site that allocates with a tracker but frees with `null` (or vice versa) desyncs the counter and trips the live `recordPerQueryMemAlloc` balance assert. Trace every free / close path — error paths and `toTop()` / `clear()` / cursor-close reuse included — and confirm the identical tracker is used on both ends.
 - **Nested SQL inherits the outer tracker.** Subqueries, the mat-view refresh inner SELECT, and WAL apply inner SQL must inherit the tracker already bound on the context, not acquire their own. A new acquisition site that acquires unconditionally (instead of only when no outer tracker is present) double-counts — flag it.
 - **Coverage has a test.** A newly wired allocator needs a `*MemoryTrackerTest` proving (a) a breach throws the per-query out-of-memory message, (b) an under-limit run succeeds, and (c) a `getCursor()`-to-close leak loop stays balanced. Missing tracker tests for a newly wired site is a high-priority finding; so is a factory-class routing guard that no longer pins the test to the intended plan.
+
+### Store-and-forward & pool startup invariants (QWP client contract)
+Apply this whenever the diff touches the QWP ingress path (upgrade/role
+gating, in-place demote / lifecycle switch, connection handling on role
+change), replication failover, a `java-questdb-client` submodule bump, or
+tests that drive a producer through a failover/switch window. These are the
+CLIENT's store-and-forward guarantees (the client code lives in the
+`java-questdb-client` submodule); server-side changes and tests in this repo
+must be reviewed against them. A violation here is a **Critical** finding:
+the whole point of store-and-forward is that a running producer never loses
+data and never hard-fails on a transient outage.
+
+**Drainer (steady state — once the pool is running).**
+- Once the pool is running, an async drainer thread ships buffered SF data to
+  the server. It MUST NOT propagate server / transport errors back to the
+  client (`Sender` producer calls, `flush()`, the pooled handle). The ONLY
+  error a running drainer may surface to the caller is **SF out of space** (the
+  on-disk / backing buffer is full and can accept no more rows). Flag any other
+  failure class (connect-refused, DNS, unreachable/black-hole, TLS/cert, auth,
+  role-reject, upgrade/protocol timeout, reset) that can escape the drainer
+  onto a producer or borrow call.
+- Primary reconnect MUST be fully contained inside the drainer thread and MUST
+  have **no time limit** — no `reconnect_max_duration_millis`-style budget, no
+  deadline, no "give up and latch terminal after N ms". A budget that latches
+  the sender terminal on a long outage is a Critical violation: it drops a
+  producer that store-and-forward promised to keep alive. Flag any bounded
+  reconnect loop, `deadlineNanos` / `while (now < deadline)`, or terminal
+  `SenderError` reachable from the running drainer's reconnect path.
+- The drainer must retry with **exponential backoff** and handle every connect
+  failure class gracefully, without a hard fail — it keeps buffering and keeps
+  retrying until the wire is back. The per-attempt backoff may be capped (a max
+  delay between attempts), but the RETRY LOOP ITSELF must be unbounded. Flag a
+  capped total retry duration or an attempt-count cap on the steady-state
+  drainer.
+- **Sanctioned terminals (orphan-slot drainer only).** The orphan drainer
+  (`BackgroundDrainer`) MAY quarantine its slot (`.failed` sentinel,
+  human-in-the-loop) on conditions that are terminal by design: auth failure,
+  a non-421 upgrade reject, and a genuine cluster-wide durable-ack capability
+  gap that exhausted its documented settle budget (16 consecutive
+  capability-gap sweeps, or a wall-clock budget anchored at the FIRST
+  capability-gap error of the episode — whichever is hit first). These are
+  NOT violations of the no-budget rule above. The settle budget applies ONLY
+  to consecutive capability-gap attempts: transient classes (role reject,
+  transport error) must never increment it or burn its wall clock — a
+  transient state consuming the terminal budget (shared attempt counter,
+  entry-anchored deadline) IS a Critical violation of this checklist.
+
+**Pool startup — two modes; the mode decides who sees connectivity errors.**
+- `lazy_connect=true`: `build()` MUST succeed with **no server present**. The
+  producing `Sender` must work immediately (writes buffer via SF), and once the
+  server comes up the read side must also connect and read (reads are deferred,
+  not disabled).
+- `lazy_connect=false` (default): `build()` / the initial connect MUST expose
+  connectivity problems to the caller — DNS errors, connect-refused /
+  unreachable, TLS/cert, authentication/authorization, and connect/upgrade
+  timeouts must all surface as a thrown exception at startup, not be swallowed.
+- **In BOTH modes the boundary is the same:** connectivity errors are only
+  ever the caller's problem DURING initialization. Once the client has
+  connected and is past initialization, the running drainer reverts to the
+  steady-state contract above — it must NEVER expose transport problems, NEVER
+  impose a reconnect time budget, and NEVER hard-fail on a transient outage.
+
+**Server-side & test application (this repo).**
+- The server MUST NOT rely on producer-visible role errors: an in-place
+  demote CLOSES QWP ingress connections (no per-write SECURITY_ERROR to an SF
+  sender). A server change that reintroduces per-write role errors on the QWP
+  ingress path breaks the containment contract above.
+- Flag any test (unit, integration, or e2e) that uses QWP producer-visible
+  role errors as evidence of the REPLICA write gate — under the containment
+  contract the producer is silent by design. Write-gate evidence belongs on
+  pg-wire probes, frozen commit counts on the settled replica, and
+  post-promotion SF drain (durable-ack await barriers + dense oracles).
+- Dense/count oracles over rows produced through an SF sender must account
+  for at-least-once replay: durably ack (await) seed rows before the
+  disturbance, or use a DEDUP table — otherwise the oracle reports replay
+  duplicates as data corruption.
 
 ### SQL conventions (if tests or SQL involved)
 - Keywords in UPPERCASE
--- a/.pi/skills/review-pr/SKILL.md
+++ b/.pi/skills/review-pr/SKILL.md
@@ -192,7 +192,7 @@ Every agent receives:
 
 Launch the following agents in parallel (in pi, via `subagent(...)` fresh-context `reviewer` tasks — see "Spawning review agents in pi").
 
-**Agent 1 — Correctness & bugs:** NULL handling, edge cases, logic errors, off-by-one, operator precedence, error paths. Cross-reference every changed symbol against its callsite inventory and verify the new behavior is correct at each callsite.
+**Agent 1 — Correctness & bugs:** NULL handling, edge cases, logic errors, off-by-one, operator precedence, error paths. Cross-reference every changed symbol against its callsite inventory and verify the new behavior is correct at each callsite. When the diff touches the QWP ingress / role-gating path, an in-place switch or failover, a `java-questdb-client` submodule bump, or store-and-forward test suites, also verify the "Store-and-forward & pool startup invariants" checklist — a change that lets a running SF drainer surface transport errors to the producer, imposes a reconnect time budget on it, or hard-fails it on a transient outage is a Critical (data-loss) finding.
 
 **Agent 2 — Concurrency:** Race conditions, shared mutable state, missing volatile, lock ordering, thread-safety of data structures. Use the implicit contract list (lock order, thread-affinity) and check every callsite from 2.5b for violations of the new contract.
 
@@ -402,6 +402,82 @@ Every new loop, traversal, data structure choice, and computation must be justif
 - Resources properly closed in all code paths (especially error paths)
 - try-with-resources used where applicable
 - Native memory freed correctly
+
+### Store-and-forward & pool startup invariants (QWP client contract)
+Apply this whenever the diff touches the QWP ingress path (upgrade/role
+gating, in-place demote / lifecycle switch, connection handling on role
+change), replication failover, a `java-questdb-client` submodule bump, or
+tests that drive a producer through a failover/switch window. These are the
+CLIENT's store-and-forward guarantees (the client code lives in the
+`java-questdb-client` submodule); server-side changes and tests in this repo
+must be reviewed against them. A violation here is a **Critical** finding:
+the whole point of store-and-forward is that a running producer never loses
+data and never hard-fails on a transient outage.
+
+**Drainer (steady state — once the pool is running).**
+- Once the pool is running, an async drainer thread ships buffered SF data to
+  the server. It MUST NOT propagate server / transport errors back to the
+  client (`Sender` producer calls, `flush()`, the pooled handle). The ONLY
+  error a running drainer may surface to the caller is **SF out of space** (the
+  on-disk / backing buffer is full and can accept no more rows). Flag any other
+  failure class (connect-refused, DNS, unreachable/black-hole, TLS/cert, auth,
+  role-reject, upgrade/protocol timeout, reset) that can escape the drainer
+  onto a producer or borrow call.
+- Primary reconnect MUST be fully contained inside the drainer thread and MUST
+  have **no time limit** — no `reconnect_max_duration_millis`-style budget, no
+  deadline, no "give up and latch terminal after N ms". A budget that latches
+  the sender terminal on a long outage is a Critical violation: it drops a
+  producer that store-and-forward promised to keep alive. Flag any bounded
+  reconnect loop, `deadlineNanos` / `while (now < deadline)`, or terminal
+  `SenderError` reachable from the running drainer's reconnect path.
+- The drainer must retry with **exponential backoff** and handle every connect
+  failure class gracefully, without a hard fail — it keeps buffering and keeps
+  retrying until the wire is back. The per-attempt backoff may be capped (a max
+  delay between attempts), but the RETRY LOOP ITSELF must be unbounded. Flag a
+  capped total retry duration or an attempt-count cap on the steady-state
+  drainer.
+- **Sanctioned terminals (orphan-slot drainer only).** The orphan drainer
+  (`BackgroundDrainer`) MAY quarantine its slot (`.failed` sentinel,
+  human-in-the-loop) on conditions that are terminal by design: auth failure,
+  a non-421 upgrade reject, and a genuine cluster-wide durable-ack capability
+  gap that exhausted its documented settle budget (16 consecutive
+  capability-gap sweeps, or a wall-clock budget anchored at the FIRST
+  capability-gap error of the episode — whichever is hit first). These are
+  NOT violations of the no-budget rule above. The settle budget applies ONLY
+  to consecutive capability-gap attempts: transient classes (role reject,
+  transport error) must never increment it or burn its wall clock — a
+  transient state consuming the terminal budget (shared attempt counter,
+  entry-anchored deadline) IS a Critical violation of this checklist.
+
+**Pool startup — two modes; the mode decides who sees connectivity errors.**
+- `lazy_connect=true`: `build()` MUST succeed with **no server present**. The
+  producing `Sender` must work immediately (writes buffer via SF), and once the
+  server comes up the read side must also connect and read (reads are deferred,
+  not disabled).
+- `lazy_connect=false` (default): `build()` / the initial connect MUST expose
+  connectivity problems to the caller — DNS errors, connect-refused /
+  unreachable, TLS/cert, authentication/authorization, and connect/upgrade
+  timeouts must all surface as a thrown exception at startup, not be swallowed.
+- **In BOTH modes the boundary is the same:** connectivity errors are only
+  ever the caller's problem DURING initialization. Once the client has
+  connected and is past initialization, the running drainer reverts to the
+  steady-state contract above — it must NEVER expose transport problems, NEVER
+  impose a reconnect time budget, and NEVER hard-fail on a transient outage.
+
+**Server-side & test application (this repo).**
+- The server MUST NOT rely on producer-visible role errors: an in-place
+  demote CLOSES QWP ingress connections (no per-write SECURITY_ERROR to an SF
+  sender). A server change that reintroduces per-write role errors on the QWP
+  ingress path breaks the containment contract above.
+- Flag any test (unit, integration, or e2e) that uses QWP producer-visible
+  role errors as evidence of the REPLICA write gate — under the containment
+  contract the producer is silent by design. Write-gate evidence belongs on
+  pg-wire probes, frozen commit counts on the settled replica, and
+  post-promotion SF drain (durable-ack await barriers + dense oracles).
+- Dense/count oracles over rows produced through an SF sender must account
+  for at-least-once replay: durably ack (await) seed rows before the
+  disturbance, or use a DEDUP table — otherwise the oracle reports replay
+  duplicates as data corruption.
 
 ### SQL conventions (if tests or SQL involved)
 - Keywords in UPPERCASE
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -117,7 +117,7 @@
           -->
         <uberjar.name>benchmarks</uberjar.name>
 
-        <questdb.client.version>1.3.5-SNAPSHOT</questdb.client.version>
+        <questdb.client.version>1.3.6-SNAPSHOT</questdb.client.version>
     </properties>
 
     <build>
--- a/ci/templates/build-client-native.yml
+++ b/ci/templates/build-client-native.yml
@@ -1,0 +1,90 @@
+# Build the java-questdb-client native library from source when the local-client
+# profile is active.
+#
+# The client submodule no longer commits its compiled native libraries
+# (libquestdb.so / .dylib / .dll). When core/pom.xml references a -SNAPSHOT
+# client, detect-local-client.yml sets CLIENT_PROFILE to "-P local-client" and
+# the reactor builds the client from source, so CI must also compile its native
+# library -- otherwise every client-dependent test fails at class load with:
+#   FatalError: cannot find /io/questdb/client/bin/<platform>/libquestdb.so
+# A Maven-Central client (release version) ships the binaries inside its jar, so
+# CLIENT_PROFILE is empty in that case and every step below is skipped.
+#
+# CMake writes the library into core/target/classes/.../bin-local/, which the
+# subsequent `mvn clean` would wipe before the test phase runs. We copy it into
+# the client's source resources (src/main/resources/io/questdb/client/bin/
+# <platform>/) so it survives the clean and lands on the production resource
+# path the loader checks first (see io.questdb.client.std.Os).
+#
+# These steps run after the JDK activation steps so $JAVA_HOME points at the JDK
+# whose jni.h CMake compiles against.
+steps:
+  - bash: |
+      set -eux
+      # zstd is a submodule of the client and is required by its CMake build.
+      git submodule update --init --recursive java-questdb-client
+      # gcc is already present (the core and Rust builds need it); add cmake and
+      # nasm (the x86-64 asmlib) only when missing.
+      if ! command -v cmake >/dev/null 2>&1 || ! command -v nasm >/dev/null 2>&1; then
+        sudo apt-get update
+        sudo apt-get install -y cmake nasm build-essential
+      fi
+      export JAVA_HOME="${JAVA_HOME:-$(dirname "$(dirname "$(readlink -f "$(command -v java)")")")}"
+      cd java-questdb-client/core
+      cmake -DCMAKE_BUILD_TYPE=Release -B cmake-build-release -S.
+      cmake --build cmake-build-release --config Release --parallel
+      case "$(uname -m)" in
+        aarch64|arm64) plat=linux-aarch64 ;;
+        *)             plat=linux-x86-64 ;;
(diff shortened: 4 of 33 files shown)


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 11 / 19 (57.89%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpIngressProcessorState.java|8|14|57.14%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpIngressUpgradeProcessor.java|3|5|60.00%|

jerrinot:
enterprise failed due to a infra hiccup:
```
+ git fetch --depth 1 origin a8bec7b1fff310b9500ed2db6a84a7bc14591ddd
OSS-triggered build detected. Updating questdb submodule to commit: a8bec7b1fff310b9500ed2db6a84a7bc14591ddd
fatal: unable to access 'https://github.com/questdb/questdb.git/': Could not resolve host: github.com
```

jerrinot:
## Critical — data integrity (production code)

### 1. Cumulative ACK leapfrogs a silently rejected frame → silent data loss
**CONFIRMED** — `core/src/main/java/io/questdb/cutlass/qwp/server/QwpIngressUpgradeProcessor.java:883`

A frame gate-rejected during a demote gets no NACK and no sequence reservation. `roleChangeCloseWithUploadGrace` defers, sends nothing, and the `finally`-block `state.clear()` re-arms the connection (resets `currentStatus`, `roleChangeClosePending`). The rejected frame's sequence number is consumed (`messageSequence++`) but never acked or errored.

**Failure chain:** durable-ack SF client pipelines frames N and N+1 → demote gate-rejects N, deferral armed → node re-promotes within the 10s grace → N+1 passes the live `engine.isReadOnlyMode()` gate, commits, `setHighestProcessedSequence(N+1)` emits a **cumulative** ACK → client's `SegmentRing.acknowledge` ("the server has confirmed every FSN up to and including this value") trims frame N's SF slot → N's rows are never written and never replayed.

### 2. Role-change deferral survives re-promotion; never cancelled
**CONFIRMED** — `QwpIngressUpgradeProcessor.java:996`, `QwpIngressProcessorState.java:396`

`roleChangeCloseDeferredDeadline` has exactly three writes: init `-1`, set in `deferRoleChangeClose()`, reset **only** in `onDisconnected()`. No path cancels it when the node becomes writable again; `handlePing`'s close arm checks only `isRoleChangeCloseDeferred() && (fullyUploaded || graceExpired)` — no re-check of `engine.isReadOnlyMode()`.

- (a) After a demote→re-promote flap, the next keepalive PING closes a healthy, actively-committing connection with `NORMAL_CLOSURE` and the stale read-only reason.
- (b) In the grace-expired arm, it closes while fresh post-promotion commits are still un-acked in `pendingDurableSeqTxns` → client replays → duplicates on non-DEDUP tables.
- (c) `deferRoleChangeClose()` is a strict no-op while `deadline != -1`, so a **second** demote on the same connection inherits the long-expired deadline and gets **zero** upload grace.

**Fix shape:** reset deadline + reason when a frame passes the writability gate (or on successful commit), and/or re-validate `engine.isReadOnlyMode()` in `handlePing`'s close arm.

### 3. Enterprise demote swaps the DurableAckRegistry to a no-op sentinel while deferred connections still poll it
**CONFIRMED** — `QwpIngressUpgradeProcessor.java:1079` (+ `questdb-ent PrimaryRoleState.java:197-201`)

The deferral polls `engine.getDurableAckRegistry()` live on every check. `PrimaryRoleState.close()` swaps the registry to `DefaultDurableAckRegistry.INSTANCE` (`getDurablyUploadedSeqTxn()` always `-1`) as soon as the upload drain finishes — intended to gate *new* connections, but it starves *existing* deferred ones: `isDurableWorkFullyUploaded()` reads `-1 < pendingSeqTxn` forever, and `collectDurableProgress` skips negative watermarks so the pending set can never drain.

**Consequence:** the demote cascade (`prepareForRoleSwitch` → `old.drain()` → `old.close()`, back-to-back on one thread; `EntQwipEnvelope` only blocks new connections) leaves deferred connections a microseconds-wide window to observe coverage. In practice every demote-deferred durable-ack connection lingers the full 10s, closes "with un-acked durable work", and the client replays batches the object store already holds. The 10s "stall guard" is the **common path**, not the exception.

**Fix directions:** leave the upload registry readable after close (swap only `isEnabled`), snapshot final watermarks into the sentinel, or rendezvous with deferred QWP closes between `old.drain()` and `old.close()`.

### 4. Blocked-close resume path never sends the final durable ack
**CONFIRMED** — `QwpIngressUpgradeProcessor.java:595` (`SEND_STATE_RESUME_ACK_THEN_CLOSE`)

Happy path: `sendFatalClose` → `flushPendingAck` sends cumulative ack **then** `trySendDurableAck`, then CLOSE — matching the deferral's documented invariant. But if the cumulative ack blocks (`PeerIsSlowToRead`), state moves to `RESUME_ACK_THEN_CLOSE`, and on resume `sendDeferredFatalClose` (lines 1283-1310) writes **only the CLOSE frame** — no `trySendDurableAck` — unlike the plain `RESUME_ACK` branch which performs both.

Wider than first thought: if the connection is already in *any* `RESUME_*` state when the deferral exits, `flushPendingAck` no-ops entirely (`hasPendingAck()` requires `SEND_STATE_READY`), so the durable ack is never even attempted. Cumulative OK acks cannot substitute — the client advances its replay/trim watermark only on `STATUS_DURABLE_ACK` frames. Result: stale watermark → reconnect → replay → duplicates, precisely under send backpressure (a slow client at demote time — the non-exotic case).

---

## High — production correctness / observability

### 5. Demote-revert race maps a transient refusal to SECURITY_ERROR (terminal HALT)
**PLAUSIBLE** (mechanism confirmed, trigger narrow) — `QwpIngressProcessorState.java:936`

`rejectCairoError` classifies via `e.isAuthorizationError() && engine.isReadOnlyMode()` — a **live** re-read at catch time. If the demote reverts between the commit-path throw and the catch (real path: `drainWriterPool(restorePrimaryOnTimeout=true)` restores PRIMARY when the drain budget expires on busy non-WAL writers), the refusal falls through to `cairoExceptionStatus` → `Status.SECURITY_ERROR` → NACK-v2 client latches a permanent terminal HALT for a milliseconds-long condition.

The exception carries only the generic authorization flag; the message (`READ_ONLY_ACCESS_MESSAGE`) is never inspected. The branch's own fuzz only flips writable→read-only, never the revert. **Fix:** match the read-only refusal by message/marker instead of re-reading live engine state.

### 6. Grace-expiry diagnostics wrong on both paths
**CONFIRMED** — `QwpIngressUpgradeProcessor.java:1093` (and 996-1001)

- `handlePing` completes a deferred close via an inline copy of the completion predicate and **omits** the `role-change close upload grace expired; closing with un-acked durable work, client replay may duplicate` LOG.error. PING is the designated re-entry poll (per the code's own javadoc), so most real grace-expired closes — including every one caused by finding #3 — happen **silently**.
- Conversely, `roleChangeCloseWithUploadGrace` fires that LOG.error whenever the grace expired, even when `isDurableWorkFullyUploaded()` is also true — a false "un-acked durable work" alarm on a slow-but-clean close.

**Fix together:** route `handlePing` through `roleChangeCloseWithUploadGrace`, and gate the log on `!isDurableWorkFullyUploaded(...)`.

### 7. Grace is not wall-time enforced; deferral sends nothing for the rejected frame
**PLAUSIBLE** — `QwpIngressUpgradeProcessor.java:1090`

Deferral completion is evaluated only on inbound frames/PINGs; the IODispatcher HEARTBEAT never reaches the processor (`HttpConnectionContext` re-registers it without invoking the processor). The deferral path sends no NACK and no CLOSE for the rejected frame. A durable-ack client with keepalive disabled (`interval <= 0` is documented) or a third-party non-pinging client gets neither response nor close and sits until the ~5-minute idle sweep (abrupt TCP teardown, no final durable ack). Mitigations that bound the common case: shipped client pings every 200ms by default; non-durable connections close immediately.

---

## Test defects

### 8. `awaitTxn(name, 2)` flake via default 100ms interval auto-flush
**CONFIRMED** — `core/src/test/java/io/questdb/test/client/QuestDBFacadeE2ETest.java:732`

`testTwoSendersPerTableWhileReaderThreadStreamsTwoQueries` assumes exactly two flush txns per table, but the pooled WS senders keep `DEFAULT_WS_AUTO_FLUSH_INTERVAL_NANOS = 100ms` (config string sets no auto-flush keys). A ≥100ms stall (GC/JIT on a loaded CI agent) inserts an extra committing txn; `awaitTxn` returns at `writerTxn >= 2` before the close-flush applies; the reader asserts against a partial table → intermittent `assertEquals(rowsAlpha, alphaRows.get())` failure. **Fix:** await observed row counts, or disable interval auto-flush in the config.

### 9. Slot-purge assertion vacuous in 37.5% of runs; poisoned-slot retention never asserted
**CONFIRMED** — `core/src/test/java/io/questdb/test/cutlass/qwp/e2e/QwpIngressOracleFuzzTest.java:670`

`producerCount = 2 + nextInt(2)`, producer 0 always poisoned, others coin-flipped → P(cleanDirs empty) = 3/8; `assertSlotsPurged` iterates zero dirs and asserts nothing. Separately, the comment "poisoned producers' slots intentionally RETAIN the rejected bytes" is backed by **no assertion** — a client bug purging the poisoned frame's SF slot (destroying the only replayable copy) passes every check. **Fix:** assert a retention lower bound on poisoned dirs — producer 0 is always poisoned, so this also makes the check non-vacuous in every run.

### 10. Rows 4-6 left entirely unbounded by the weakened prefix-only assertion
**CONFIRMED** — `core/src/test/java/io/questdb/test/cutlass/qwp/e2e/QwpWebSocketSenderReceiverTest.java:1871`

Old: dense `v = 1..6` assert. New: `WHERE v <= 3` prefix only. The tail is genuinely timing-dependent under NACK v2 (a dense assert would flake), but no envelope bounds it: duplicates of 4-6 or phantom `v > 6` rows pass. **Fix:** add a deterministic envelope, e.g. `count() = count_distinct(v) AND min(v) = 1 AND max(v) <= 6 AND count() BETWEEN 3 AND 6`.

### 11. Demote fuzz's concurrent phase cannot exercise the writer-acquisition window its javadoc advertises
**CONFIRMED** — `core/src/test/java/io/questdb/test/cutlass/qwp/QwpIngressDemoteRaceFuzzTest.java:277`

The concurrent phase drives only zero-table frames; `RaceTudCache.getTableUpdateDetails` unconditionally throws for table-bearing frames (would map to INTERNAL_ERROR and fail the harness). The gate-to-WAL-writer-acquisition window is covered only by same-thread deterministic scenario 1. **Fix:** soften the javadoc, or add a concurrent-safe TUD path to `RaceTudCache`.

### 12. Mid-stream poison fuzz coverage removed
**CONFIRMED** — `QwpIngressOracleFuzzTest.java:457`

The poisoned chunk is now always each producer's last chunk, and every poison path pins `max_frame_rejections=1`, so NACK-triggered replay-with-successors (head-of-line, replay ordering, watermark bugs) has no e2e coverage; the one directed test that sends successors asserts nothing about them (finding #10). A weaker-observable mid-stream fuzz (clean-prefix oracle + poisoned-frame absence + terminal; successors excluded from the dense oracle) is feasible and absent.

bluestreak01:
## Review: PR #1105 (ENT) + questdb#7341 (OSS) — tandem

**Scope reviewed:** ER005 hot-promote ownership pre-flight (Rust + Java wiring), QWP ingress Invariant-B role-change close + deferred-commit ack-hole fix, durable-ack registry freeze-on-demote, promote-path sequencer reload, the `CairoException.readOnlyAccess()` marker refactor, CI client-native builds, submodule bump.

**Verdict: approve (both PRs), with two small changes requested.** This is an exceptionally well-fenced, heavily-tested change. The review found **zero confirmed Critical correctness / data-loss / concurrency / resource / Rust-panic defects**. The five store-and-forward loss-modes the PR targets are each closed by a traced path; Invariant B holds (an in-place demote closes ingress with `NORMAL_CLOSURE`, never a producer-visible `SECURITY_ERROR`).

### Critical
None.

### Moderate

**1. [OSS] Two new `.returnsOnce(...)` on deterministic queries** — `core/src/test/java/io/questdb/test/cutlass/qwp/e2e/QwpSenderE2ETest.java:2995` (`.returnsOnce("count\n21\n")`, `defer_no_ack`) and `:3028` (`.returnsOnce("count\n4\n")`, `defer_eager_ack`). Both are `SELECT count()` after `drainWalQueue()` with no `rnd_*`/`now()` — fully deterministic — via the `assertQuery(...)` builder. `.returnsOnce` on a deterministic query skips the second cursor pass, `calculateSize()` cross-check, and factory-property assertions. *Honest severity is lower:* this file has 195 `.returnsOnce` vs 4 `.returns(` — an entrenched class-wide convention these two merely follow, and count() oracles gain little from the second pass. **Fix:** convert these two new oracles to `.returns(...)`; the class-wide cleanup is out of scope.

**2. [ENT] Residual un-fenced acked-row-loss windows — documented, narrowed, not closed (product decision).** The data-parity fence has **no boot-time backstop** (unlike the ownership fence, which the uploader boot re-validates). On its fail-open paths it genuinely forgoes the check for that promote:
- store blip inside the clamped `PREFLIGHT_MAX_RETRY_ATTEMPTS=1` download → `Err` → fail-open;
- unreadable exclusion snapshot → `excluded=None` → the **entire** data fence is skipped (`preflight_store_ownership_with_client`);
- two-owner overlap after a clean pass (old primary not fully quiesced does one more sync bump) → ownership stays one-behind (tolerated), data fence never re-runs → the same seqTxn-collision loss (build 248267 signature).

`EntCairoEngine.reloadWalTableSequencersForPromote` narrows the remaining window but its own javadoc concedes a teardown-budget-overrun race on slow nodes. **None of this is a regression** — the PR *reduces* pre-existing loss modes and the tradeoffs are explicitly documented in-code (`WalUploader.validateStoreOwnership` "Backstop asymmetry"). Flagging so the owner can consciously accept "advisory data parity / fail-open" vs. failing *closed* on a reachable-but-lagging store, and add an operational runbook constraint (never promote until the old primary is provably stopped).

**3. [ENT] Pre-flight blocking store download runs inside `synchronized (teardownMonitor)`** — `EntServerMain.ReplicationEnvelope.switchRole`. The attempt *count* is clamped (≤4), but a connect-level black-hole is bounded only by the HTTP client's connect behavior, so a promote against a black-holed object store can hold `teardownMonitor` and defer a concurrent SIGTERM/`close()` by the connect budget. Not a deadlock. Pre-existing for the boot read, newly on the promote-cascade critical path. Bounded — worth an operator note.

### Minor

**4. [ENT] Test asserts a weaker property than it describes** — `questdb-ent/rust/qdb-ent/src/wal/uploader/mod.rs`, `data_parity_missing_local_table_refuses` (~:7095) uses `index_with_live_table(..., first_txn=1)`, so `is_baseline_stalled(None, 1)` is `true` → the refusal is the **baseline-stalled "re-baseline the replica"** message, but the test comment claims the *self-clearing* "a retry clears the refusal" intent, and it only asserts `local_txn=none` (which both flavors contain). The classification is **correct and consistent** with the downloader's shared `is_baseline_stalled` predicate (no drift) — no functional bug — but tighten the test to pin the non-stalled message for a genesis (`first_txn=0`) missing table and the stalled message for a purged one.

**5. [ENT] Eager `String` allocation in `preflight_check_local_data`** (`uploader/mod.rs:3936`) — `detail` is `format!`-allocated for every lagging table but only 5 are kept (`PREFLIGHT_MAX_NAMED_LAGGING_TABLES`). Cold refusal path; move the `format!` inside the two `push` branches (keep the excluded-table `warn!` separate).

**6. [ENT] Redundant recompute during the demote deferral window** — `QwpIngressProcessorState.isDurableWorkFullyUploaded` is called up to 3× per deferred-close re-entry and re-hashes the key it is already iterating (`pendingDurableSeqTxns.get(tableName)` on the key from `.keys()`). Warm only during a demote deferral with small `T`; compute once per inbound event and iterate the backing arrays. `frozenSnapshot` similarly uses `keySet()`+`get()` instead of `entrySet()` (cold).

**7. [OSS] Field ordering / reserved constant** — new fields in `QwpIngressProcessorState` deviate from the alphabetical member-ordering standard (IDE rearrange fixes all). `STATUS_NOT_WRITABLE = 0x0C` (`QwpConstants`) is reserved-but-unused — track it so it doesn't linger.

**8. [both] Metadata** — OSS #7341's title (`add a connect timeout and tolerant client startup`) under-represents its highest-risk content: it also carries the **silent-data-loss** deferred-commit ack fix and the Invariant-B ingress rework; surface that in the title. Neither PR has a `Fixes #NNN`. File a tracking issue for the acknowledged `BEGIN_TXN` sidecar + kill-mid-transaction e2e follow-up so it survives the squash-merge.

**9. [both] Coverage gaps (non-blocking)** — `reloadWalTableSequencersForPromote` has no targeted unit regression (its per-table `Throwable`-skip branch is untested; only indirect e2e coverage); the OSS `LifecycleOrchestrator` halted-logger `try/catch` guard is untested.

### Cleared / verified (not defects)
- **Marker completeness (top data-loss exposure): CLEARED.** All ~20 read-only role gates converted to `readOnlyAccess()` (marked); genuine ACL denials correctly stay unmarked → `SECURITY_ERROR`. The QWP write path (`deniedOnReplica` → `EntReplicaSecurityContextImpl`, and `EntCairoEngine.getWalWriter/getTableWriterAPI` backstops) is fully marked. `clear(errno)` resets the sticky flag.
- **Rust panic-via-JNI: NONE reachable.** The PR *removes* an abort risk (old `debug_assert` + unconditional `fetch_add` → CAS `try_advance_from`). All `unsafe` mmap/atomic invariants verified; JNI paths use `?`/`context`/`unwrap_or(false)`, fail-open; `peek` is genuinely read-only.
- **Resource leaks: NONE.** `CommonConfig` try-with-resources frees on all paths (incl. refusal throw outside the try); `peek` File+Mmap drop on the error path; `read_excluded_table_dirs` AutoLocal per element; `PrimaryRoleState.close()` null-then-free is double-free safe; `FrozenSnapshot` is GC'd on next promote.
- **Concurrency: CLEARED.** `peek` tear-free (MAP_SHARED page aliasing + SeqCst); `try_advance_from` CAS exclusive under contention; `segmentUploaded` re-route records through the live registry across the whole drain (`instanceof` guard makes post-close stragglers no-ops); QWP deferral fields are per-connection/event-loop confined.
- **`RoleSwitchOwnershipRefusedException` throw→catch: CLEARED.** No intermediate catch mis-classifies it as torn.
- **Deferred-close interleavings (demote-revert, re-promote in grace, table-drop, group-commit-throws, idle-reaper): CLEARED.** All terminate with a best-effort final durable ack; no cumulative-ack leapfrog constructible.

### Summary
**ENT #1105: approve. OSS #7341: approve.** Two changes requested: convert the two new deterministic `.returnsOnce` count oracles (finding 1), and make a conscious product decision on the fail-open data-parity fence (finding 2, not a regression). Cross-context walk of every changed symbol's out-of-diff callsites returned **0 BROKEN** — the widely-used `readOnlyAccess()` refactor and the registry-freeze both satisfy every external caller.

<sub>Reviewed at level 3 (tandem, all agents + per-finding source verification).</sub>

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 152 / 191 (79.58%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|0|2|00.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|1|7|14.29%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/InsertAsSelectOperationImpl.java|1|2|50.00%|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGPipelineEntry.java|5|8|62.50%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpIngressUpgradeProcessor.java|41|57|71.93%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/OperationDispatcher.java|3|4|75.00%|
|:large_blue_circle:|io/questdb/lifecycle/LifecycleOrchestrator.java|19|23|82.61%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpIngressProcessorState.java|69|75|92.00%|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/TableUpdateDetails.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/JsonQueryProcessor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/InsertOperationImpl.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoException.java|6|6|100.00%|

nwoolmer (dismissed):

ideoma (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 159 / 198 (80.30%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|0|2|00.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|1|7|14.29%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/InsertAsSelectOperationImpl.java|1|2|50.00%|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGPipelineEntry.java|5|8|62.50%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpIngressUpgradeProcessor.java|47|63|74.60%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/OperationDispatcher.java|3|4|75.00%|
|:large_blue_circle:|io/questdb/lifecycle/LifecycleOrchestrator.java|19|23|82.61%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpIngressProcessorState.java|70|76|92.11%|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/TableUpdateDetails.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/JsonQueryProcessor.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/InsertOperationImpl.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoException.java|6|6|100.00%|

PR #7343, opened 2026-06-30
fix(core): make DataID publication thread-safe

bluestreak01:
**Related PRs**
- Enterprise: questdb/questdb-enterprise#1100
- OSS: questdb/questdb#7343

---

## What

Make `DataID`'s in-memory UUID publication thread-safe by `synchronized`-ing every accessor that touches the `lo`/`hi` pair: `initialize()`, `isInitialized()`, `getLo()`, `getHi()`, and `toSink()` — symmetric with the already-`synchronized` `change()`.

## Why

`DataID` wraps a 128-bit UUID stored as two `long`s (`Uuid.lo`/`Uuid.hi`), which are **non-volatile**. The value is published by one thread and read by another:

- **Writer:** the replication WAL downloader learns the DataID from the object store and publishes it via a `DataID.initialize()` JNI up-call, which does two plain stores through `Uuid.of(lo, hi)`.
- **Reader:** `isInitialized()` reads `lo` and `hi` as two separate, unsynchronized loads (and the native backup/replication path plus the `current_data_id()` SQL function read via `getLo()`/`getHi()`).

Because the stores and loads are unsynchronized and non-volatile, a reader can:

- observe a **torn** half-written value (`lo` set, `hi` still unset) and act on it — e.g. exit an initialization wait early, or
- in theory **never observe** the publish at all (visibility).

`change()` was already `synchronized`, but `initialize()` and the readers were not, so the publication had no happens-before edge.

This is a pre-existing latent race in `DataID`/`Uuid` (these files are otherwise unchanged). It becomes newly reachable in Enterprise, where a booting backup scheduler makes a **control-flow decision** on `isInitialized()` while the downloader concurrently publishes the DataID — see questdb/questdb-enterprise#1100.

## Fix

Read/write the `lo`/`hi` pair as a single critical section by synchronizing the accessors on the `DataID` instance. Intrinsic locks are reentrant, so `initialize()`/`change()` calling `isInitialized()` is fine.

The fix is deliberately kept in `DataID` rather than making `Uuid.lo`/`Uuid.hi` `volatile`: `Uuid` is a hot, general-purpose value holder (record comparators, UUID columns, tight loops), and the unsafe publication is specific to `DataID`'s one-writer / other-reader pattern. Adding volatile/lock overhead to every UUID operation to fix a boot-time publish would be the wrong trade.

## Testing

- `DataIDTest` — green (5/5).
- `mvn compile -pl questdb/core` — clean.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/DataID.java
+++ b/core/src/main/java/io/questdb/cairo/DataID.java
@@ -42,6 +42,16 @@ import org.jetbrains.annotations.NotNull;
  * <p>
  * One shouldn't modify the data id in an unblank database as it may cause data loss.
  * </p>
+ * <p>
+ * Thread-safety: the in-memory UUID can be published by one thread (e.g. the replication
+ * downloader's {@link #initialize(long, long)} JNI up-call) and read by another (e.g. a booting
+ * backup scheduler gating on {@link #isInitialized()}). The 128-bit value is two longs, so a plain
+ * read could observe a torn half-written value or miss the publish entirely. All accessors that touch
+ * the {@code lo}/{@code hi} pair are therefore {@code synchronized} on this instance, reading/writing
+ * the pair as one critical section. The underlying {@link Uuid} fields are deliberately left
+ * non-volatile: that class is a hot, general-purpose value holder, so the synchronization is kept
+ * here where the one-writer/other-reader publication actually happens.
+ * </p>
  */
 public final class DataID implements Sinkable {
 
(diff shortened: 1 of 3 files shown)


mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/DataID.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/CurrentDataIdFunctionFactory.java|2|2|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=247262)
**Tested Commit:** `3aa54faee9b4801c6bd8c55fc47659ca90e856ce`

bluestreak01:
Superseded by #7344 — branch renamed to `fix/replica-backup-scheduler-dataid-race` to match the enterprise branch name (fixes the OSS CI). Same commit/content.

PR #7344, opened 2026-07-01
fix(core): make DataID publication thread-safe

bluestreak01:
**Related PRs**
- Enterprise: questdb/questdb-enterprise#1100
- OSS: questdb/questdb#7344

---

## What

Make `DataID`'s in-memory UUID publication thread-safe by `synchronized`-ing every accessor that touches the `lo`/`hi` pair: `initialize()`, `isInitialized()`, `getLo()`, `getHi()`, and `toSink()` — symmetric with the already-`synchronized` `change()`.

## Why

`DataID` wraps a 128-bit UUID stored as two `long`s (`Uuid.lo`/`Uuid.hi`), which are **non-volatile**. The value is published by one thread and read by another:

- **Writer:** the replication WAL downloader learns the DataID from the object store and publishes it via a `DataID.initialize()` JNI up-call, which does two plain stores through `Uuid.of(lo, hi)`.
- **Reader:** `isInitialized()` reads `lo` and `hi` as two separate, unsynchronized loads (and the native backup/replication path plus the `current_data_id()` SQL function read via `getLo()`/`getHi()`).

Because the stores and loads are unsynchronized and non-volatile, a reader can:

- observe a **torn** half-written value (`lo` set, `hi` still unset) and act on it — e.g. exit an initialization wait early, or
- in theory **never observe** the publish at all (visibility).

`change()` was already `synchronized`, but `initialize()` and the readers were not, so the publication had no happens-before edge.

This is a pre-existing latent race in `DataID`/`Uuid` (these files are otherwise unchanged). It becomes newly reachable in Enterprise, where a booting backup scheduler makes a **control-flow decision** on `isInitialized()` while the downloader concurrently publishes the DataID — see questdb/questdb-enterprise#1100.

## Fix

Read/write the `lo`/`hi` pair as a single critical section by synchronizing the accessors on the `DataID` instance. Intrinsic locks are reentrant, so `initialize()`/`change()` calling `isInitialized()` is fine.

The fix is deliberately kept in `DataID` rather than making `Uuid.lo`/`Uuid.hi` `volatile`: `Uuid` is a hot, general-purpose value holder (record comparators, UUID columns, tight loops), and the unsafe publication is specific to `DataID`'s one-writer / other-reader pattern. Adding volatile/lock overhead to every UUID operation to fix a boot-time publish would be the wrong trade.

## Testing

- `DataIDTest` — green (5/5).
- `mvn compile -pl questdb/core` — clean.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/DataID.java
+++ b/core/src/main/java/io/questdb/cairo/DataID.java
@@ -42,6 +42,16 @@ import org.jetbrains.annotations.NotNull;
  * <p>
  * One shouldn't modify the data id in an unblank database as it may cause data loss.
  * </p>
+ * <p>
+ * Thread-safety: the in-memory UUID can be published by one thread (e.g. the replication
+ * downloader's {@link #initialize(long, long)} JNI up-call) and read by another (e.g. a booting
+ * backup scheduler gating on {@link #isInitialized()}). The 128-bit value is two longs, so a plain
+ * read could observe a torn half-written value or miss the publish entirely. All accessors that touch
+ * the {@code lo}/{@code hi} pair are therefore {@code synchronized} on this instance, reading/writing
+ * the pair as one critical section. The underlying {@link Uuid} fields are deliberately left
+ * non-volatile: that class is a hot, general-purpose value holder, so the synchronization is kept
+ * here where the one-writer/other-reader publication actually happens.
+ * </p>
  */
 public final class DataID implements Sinkable {
 
@@ -95,11 +105,11 @@ public final class DataID implements Sinkable {
         return id;
     }
 
-    public long getHi() {
+    public synchronized long getHi() {
         return id.getHi();
     }
 
-    public long getLo() {
+    public synchronized long getLo() {
         return id.getLo();
     }
 
@@ -112,7 +122,7 @@ public final class DataID implements Sinkable {
      * @return true if the value was initialized, or false if it could not be initialized
      * because it was already set.
      */
-    public boolean initialize(long lo, long hi) {
+    public synchronized boolean initialize(long lo, long hi) {
         if (isInitialized()) {
             return false;
         }
@@ -126,12 +136,12 @@ public final class DataID implements Sinkable {
      *
      * @return true if the data id is initialized.
      */
-    public boolean isInitialized() {
+    public synchronized boolean isInitialized() {
         return !Uuid.isNull(id.getLo(), id.getHi());
     }
 
     @Override
-    public void toSink(@NotNull CharSink<?> sink) {
+    public synchronized void toSink(@NotNull CharSink<?> sink) {
         id.toSink(sink);
     }
 
--- a/core/src/main/java/io/questdb/cairo/DataID.java
+++ b/core/src/main/java/io/questdb/cairo/DataID.java
@@ -46,18 +46,31 @@ import org.jetbrains.annotations.NotNull;
  * Thread-safety: the in-memory UUID can be published by one thread (e.g. the replication
  * downloader's {@link #initialize(long, long)} JNI up-call) and read by another (e.g. a booting
  * backup scheduler gating on {@link #isInitialized()}). The 128-bit value is two longs, so a plain
- * read could observe a torn half-written value or miss the publish entirely. All accessors that touch
- * the {@code lo}/{@code hi} pair are therefore {@code synchronized} on this instance, reading/writing
- * the pair as one critical section. The underlying {@link Uuid} fields are deliberately left
- * non-volatile: that class is a hot, general-purpose value holder, so the synchronization is kept
- * here where the one-writer/other-reader publication actually happens.
+ * read could observe a torn half-written value or miss the publish entirely. The accessors that read or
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/questdb/cairo/DataID.java
```diff
@@ -95,14 +112,28 @@ public Uuid get() {
...
+     * within a single critical section, so the returned pair is never torn across a concurrent
+     * {@link #initialize(long, long)} / {@link #change(long, long)} publish -- unlike reading
+     * {@link #getLo()} and {@link #getHi()} as two separate calls, which drops and re-acquires the
+     * monitor in between. Callers that need both halves consistently must use this accessor.
+     *
+     * @return a new {@link Uuid} holding a consistent {@code (lo, hi)} snapshot; the NULL Uuid when not
+     * yet initialized.
+     */
+    public synchronized Uuid getSnapshot() {
+        return new Uuid(id.getLo(), id.getHi());
```
kafka1991, line 161: Can here avoid creating temp object? (thread-local or construct UuidConstant directly)

bluestreak01: this is relatively minor alloc, function creation during compilation, rare function to call at all. UuidConstant would be a good choice if it wasn't a cross-package reference. I would prefer to avoid circular dependency between cairo<->griffin

bluestreak01:
## Review — level 3, tandem (OSS #7344 + ENT questdb/questdb-enterprise#1100)

The production change is **correct**. `synchronized`-ing every accessor that touches the `lo`/`hi` pair supplies the happens-before publication edge; `getSnapshot()` gives a single-acquisition atomic pair; `CurrentDataIdFunctionFactory` correctly moves from the tearable two-call `new UuidConstant(getLo(), getHi())` to `getSnapshot()`. The `get()` escape hatch is unchanged and its one production consumer (the `ServerMain` boot advisory log, gated on `isInitialized()`) tolerates a stale/torn read as documented.

### CI note (investigated)
The red check `SelfHosted … cover on linux-griffin-root` is a **transient Maven infra flake, not a code failure**: the task log shows `BUILD FAILURE … Could not find artifact org.questdb:questdb-client:jar:1.3.5-SNAPSHOT` in the internal mirror, total time **1.6 s**, `No test result files … were found` — zero tests ran. Unrelated to this change; a re-run should clear it.

### Critical
None.

### Minor

**1. The two *concurrent* `getSnapshot` tests are weak** (`testConcurrentSnapshotStaysAtomicAcrossInitialize` / `…AcrossChange`): the reader reads `lo`/`hi` off the returned **immutable copy**, so their `torn` branch is structurally unreachable — they only guard against someone de-synchronizing `getSnapshot()` itself, not the real two-separate-call tear. This is honestly documented in the javadoc, and the actual bug **is** properly covered by the deterministic `testTwoSeparateCallsCanTearWhileSnapshotStaysAtomic` and the real-callsite `testCurrentDataIdStaysAtomicUnderConcurrentChange` (reverting the factory to `new UuidConstant(getLo(), getHi())` turns the latter RED). No change required — just don't over-trust the concurrent pair as race reproductions; their assertive power is low.

**2. Contract note: `toSink()` is now `synchronized`,** and `initialize()`/`change()` hold the monitor across `ff.write` + `ff.fsync`. Sinking/logging a DataID can now block on the monitor during a write. Negligible today (only boot/PITR log a DataID), but don't add DataID logging to a hot path — log-stall hazard.

### Downgraded (false positives, cleared after source verification)
- **`getSnapshot()` / the newly-`synchronized` getters sit on a hot path** — FALSE. All callers are cold: `current_data_id()` is a runtime constant built once per query-compile (not per row), and the remaining `getLo`/`getHi`/`isInitialized`/`toSink` callers are boot/advisory/~1 Hz. Monitor cost and the per-compile `Uuid` allocation are immaterial. Keeping the sync in `DataID` rather than making `Uuid` fields `volatile` is the right trade.
- **Rust `Uuid::from_u64_pair(hi, lo)` NULL misclassification** — FALSE. The NULL sentinel has both halves equal (`i64::MIN`), so classification is order-insensitive; it maps to SQL NULL identically.
- **The `get()` escape hatch can still tear** — accepted by design: it's the documented advisory-only path (boot log), gated on `isInitialized()`. Not a defect.

### Summary
- **Verdict: OSS #7344 — approve.** The synchronization is correct, minimal, and correctly scoped to `DataID` (not `Uuid`); the fixed SQL callsite is protected by a genuine concurrent regression test against the real factory.
- In-diff findings only; every out-of-diff `DataID` reader across OSS + enterprise was walked and verified SAFE (the one production two-call reader was fixed; remaining two-call reads are single-threaded test code against quiesced state).
- No regressions.

<sub>🤖 level-3 tandem review (11 agents, per-finding source verification)</sub>

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/DataID.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/CurrentDataIdFunctionFactory.java|2|2|100.00%|

kafka1991 (approved):

PR #7346, opened 2026-07-01
fix(sql): fix crash joining or grouping on empty VARCHAR values

sklarsa:
## Summary

A customer on 3.3.x reported that an inner (hash) join on a `VARCHAR` key
crashed with an `AssertionError`, while an ASOF join over the same data
worked. The failing query was a plain equi-join:

```
SELECT rfq.* FROM QDBK_CREDIT_RFQ rfq
INNER JOIN SALES_MAPPING ON rfq.SalesPersonId = InputSalesIdentifer
WHERE LoginId = 'zu1450' AND timestamp > '2026-06-28'
```

Stack trace:

```
java.lang.AssertionError
  at io.questdb.cairo.map.UnorderedVarcharMap$Key.putVarchar(UnorderedVarcharMap.java)
  at io.questdb.cairo.map.UnorderedVarcharMap$Key.put(...)
  at io.questdb.griffin.engine.join.HashJoinLightRecordCursorFactory$HashJoinRecordCursor.hasNext(...)
```

## Root cause

The crash triggers when a join/`GROUP BY`/`DISTINCT` `VARCHAR` key is an
**empty, non-ASCII** varchar. An empty varchar is ASCII by definition, but
the `Utf8Sequence` contract only guarantees that `isAscii() == true` means
ASCII, not the converse: a producer is free to report `isAscii() == false`
for an empty value ("Returning false does not guarantee anything").

`UnorderedVarcharMap` packs the key hash, size and flags into a single
`long`. An empty non-ASCII key packs to `0` (zero hash, since
`Hash.hashMem64(p, 0) == 0`; zero size; no flags), which is exactly the
map's empty-slot sentinel. The entry then silently corrupts the table.
An assertion existed to catch this, but the invariant was *asserted*
rather than *enforced*. ASOF joins do not use this map, which is why the
workaround worked.

## Fix

The same "empty ⇒ ASCII" invariant is violated at three layers, all
addressed here:

- **`UnorderedVarcharMap.putVarchar`** forces the ASCII flag for empty
  keys. This is the crash fix and it also covers values already at rest,
  which no read-side change alone could. `makePackComparable()` already
  ignores the ASCII bit when probing, so key identity is unchanged (an
  empty ASCII and empty non-ASCII key still resolve to the same slot).
- **`QwpEgressRequestDecoder`** built varchar bind views with the 2-arg
  `DirectUtf8String.of()`, which hardcodes `ascii=false`; an empty
  varchar bind therefore became an empty non-ASCII varchar. It now
  computes the flag with `Utf8s.isAscii()`, mirroring the pgwire decoder.
- **`VarcharTypeDriver.appendPlainValue`** (single-vector "plain" format,
  used by `RecordChain` and `BlockFileWriter`) wrote a `0` header for an
  empty non-ASCII value, colliding with the NULL/empty sentinel that
  `getPlainValue` rejects via `assert header != 0`. It now forces the
  flag for empty values, matching the Rust parquet writers, which already
  special-case `len == 0`.

## Scope and tradeoffs

- The column storage format (`VarcharTypeDriver.appendValue`, aux/data
  vectors) is intentionally **not** changed. There, an empty value still
  sets the INLINED flag, so the header is non-zero and harmless; the map
  fix handles such values on read. This keeps existing on-disk data valid
  without a rewrite and preserves the natural reproduction path in tests.
- `QwpEgressRequestDecoder` now scans the bind bytes to compute the ASCII
  flag (`Utf8s.isAscii`). This is a per-varchar-bind cost, equivalent to
  what pgwire already does, and negligible for typical bind sizes. A side
  effect is that ASCII varchar binds are now flagged ASCII rather than
  always non-ASCII, which is more correct and can enable faster
  downstream ASCII paths.
- `appendPlainValue` now reads empty non-ASCII values back as empty ASCII.
  This only affects the in-memory plain format (chains/block files), not
  persisted column data, so there is no migration concern.

## Test plan

- `UnorderedVarcharMapTest#testBlankNonAsciiKey` - the map accepts an
  empty non-ASCII key, keeps it distinct from a missing entry, dedups it,
  and collapses it with an empty ASCII key.
- `HashJoinTest#testHashJoinEmptyNonAsciiVarcharKey` - an inner join over
  a column holding an empty non-ASCII varchar returns the expected rows
  (reproduces the customer scenario via at-rest data).
- `QwpEgressBindRoundTripTest#testBindVarcharEmptyHashJoinKey` - an empty
  varchar bind used as a hash-join key succeeds end-to-end over QWP.
- `VarcharTypeDriverTest#testAppendPlainEmptyNonAsciiVarchar` - the plain
  format round-trips an empty non-ASCII value without a 0 header.
- Each test fails with an `AssertionError` before the fix and passes
  after it. The full `UnorderedVarcharMapTest`, `VarcharTypeDriverTest`,
  `HashJoinTest`, `QwpEgressBindRoundTripTest` and `OrderedMapTest`
  suites (108 tests) pass with the change.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/VarcharTypeDriver.java
+++ b/core/src/main/java/io/questdb/cairo/VarcharTypeDriver.java
@@ -89,7 +89,10 @@ public class VarcharTypeDriver implements ColumnTypeDriver {
         if (eraseAsciiFlag) {
             Unsafe.putInt(dataMemAddr, hi);
         } else {
-            final boolean ascii = value.isAscii();
+            // An empty varchar is ASCII by definition; force the flag so that an empty value
+            // does not write a 0 header, which collides with the "size == 0" NULL/empty-slot
+            // sentinels the readers rely on.
+            final boolean ascii = value.isAscii() || hi == 0;
             // ASCII flag is signaled with the highest bit
             Unsafe.putInt(dataMemAddr, ascii ? hi | Integer.MIN_VALUE : hi);
         }
@@ -108,7 +111,10 @@ public class VarcharTypeDriver implements ColumnTypeDriver {
             return;
         }
         final int size = value.size();
-        dataMem.putInt(value.isAscii() ? size | Integer.MIN_VALUE : size);
+        // An empty varchar is ASCII by definition; force the flag so that an empty value does
+        // not write a 0 header, which collides with the "size == 0" NULL/empty-slot sentinels.
+        final boolean ascii = value.isAscii() || size == 0;
+        dataMem.putInt(ascii ? size | Integer.MIN_VALUE : size);
         dataMem.putVarchar(value, 0, size);
     }
 
--- a/core/src/main/java/io/questdb/cairo/VarcharTypeDriver.java
+++ b/core/src/main/java/io/questdb/cairo/VarcharTypeDriver.java
@@ -90,8 +90,9 @@ public class VarcharTypeDriver implements ColumnTypeDriver {
             Unsafe.putInt(dataMemAddr, hi);
         } else {
             // An empty varchar is ASCII by definition; force the flag so that an empty value
-            // does not write a 0 header, which collides with the "size == 0" NULL/empty-slot
-            // sentinels the readers rely on.
+            // writes a non-zero header (size 0 | high bit) instead of 0. getPlainValue treats a
+            // 0 header as absent/uninitialized memory (assert header != 0), so a 0 header for an
+            // empty non-ASCII value would trip that assertion on read.
             final boolean ascii = value.isAscii() || hi == 0;
             // ASCII flag is signaled with the highest bit
             Unsafe.putInt(dataMemAddr, ascii ? hi | Integer.MIN_VALUE : hi);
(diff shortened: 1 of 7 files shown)


mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/map/UnorderedVarcharMap.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressRequestDecoder.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/VarcharTypeDriver.java|3|3|100.00%|

sklarsa:
```
  Deep review outcome (9 agents + per-finding verification)

  No blocking issues. Verdict: approve. 0 in-diff bugs, 0 out-of-diff bugs. Every correctness/concurrency/performance/resource/cross-context agent cleared the change, and two of them
  surfaced positive corroboration:

  - Correctness + Fresh-adversarial independently re-derived the packing proof (hashMem64(ptr,0)=0 → empty non-ASCII packs to the empty-slot sentinel 0; fix packs to 0x8000000000000000
  with identity/NULL preserved).
  - Concurrency found the fix removes a latent data-loss hazard: an empty non-ASCII key packed to 0 was skipped by parallel GROUP BY's merge sentinel guard, silently dropping that group.
  - Cross-context confirmed the fix incidentally closes the same latent assert header != 0 crash in the block-file and WAL-event readers.
  - Performance (both agents): perf-neutral; the QWP Utf8s.isAscii scan is per-query, word-vectorized, and perf-positive.
```

bluestreak01 (approved):
Approved after a level-3 mission-critical review.

Root cause and packing proof independently verified from source: an empty non-ASCII varchar packs to 0 (hashMem64(p,0)=0, xxh3Avalanche64(0)=0), colliding with UnorderedVarcharMap's empty-slot sentinel. The fix normalizes the empty=>ASCII invariant at the catch-all point (UnorderedVarcharMap.putVarchar) and additionally closes two real latent 'assert header != 0' crashes (BlockFileReader, WalEventCursor) and a silent parallel-GROUP-BY data-loss path in merge(). Leaving the on-disk column format and the eraseAsciiFlag=true dedup path untouched is correct.

0 Critical, 0 Moderate, 3 Minor. 5 draft findings verified and downgraded as false positives. Tests use the builder API correctly.

PR #7348, opened 2026-07-02
perf(sql): stream ordered UNION ALL without sorting

### LINKED ISSUE #7339: perf(sql): add order-preserving UNION ALL for timestamp streams
`UNION ALL` currently concatenates its input cursors. This is lazy and cheap, but it destroys global timestamp ordering even when every branch is already scanning in the same timestamp order.

That prevents downstream operators from using streaming/time-series paths. For example:

```sql
SELECT
  first_value(value) OVER (PARTITION BY symbol ORDER BY timestamp)
FROM (
  SELECT ... FROM t WHERE ...
  UNION ALL
  SELECT ... FROM t WHERE ...
);
```

Each branch may be ordered by `timestamp`, but concat produces:

```text
all branch A rows, then all branch B rows, ...
```

rather than one globally ordered timestamp stream. The window planner cannot prove that the input satisfies `ORDER BY timestamp`, so it falls back to `CachedWindow` and materializes the input.

We should add a merge-style physical operator:

```text
MergeUnionAll order=[timestamp ASC|DESC]
  branch A ordered by timestamp
  branch B ordered by timestamp
  ...
```

The existing concat `UnionAll` should remain the default. `MergeUnionAll` should only be selected when:

1. every branch can provide the requested timestamp order, and
2. a downstream operator benefits from preserving that order.

QuestDB already has ordering advice / scan-direction plumbing, which may be enough for explicit `ORDER BY timestamp` cases. Ordered windows need extra care: the useful order demand can come from `OVER (ORDER BY timestamp)`, not from a model-level `ORDER BY`, and today the window planner generates the base before deciding whether `CachedWindow` is needed.

`MergeUnionAll` must preserve `UNION ALL` multiplicity. Equal timestamp peer ordering does not need to match concat order; if the query only orders by `timestamp`, peers are under-specified. The operator can use a deterministic internal tie-breaker without advertising it as SQL-visible ordering.

Initial scope can be narrow: designated timestamp ASC/DESC, binary union first, then follow up with true N-way merge / nested `UNION ALL` flattening.

brunocalza:
## Summary
Closes https://github.com/questdb/questdb/issues/7339

Adds a new physical operator `MergeUnionAll` in certain cases that may benefit from it. Read the issue to understand the context better. 

## High-level Implementation
All changes have the aim of helping us pick the new `MergeUnionAll` operator over `UnionAll`, in the following case:

1. Union branches are in a mergeable shape
2. Downstream consumer is asking for the same order

We've broken the implementation into parts:

**Part I: Implement `MergeUnionAllRecordCursor` and `MergeUnionAllRecordCursorFactory`**
Straightforward 2-way merge implementation

**Part II: Add a check in `SqlCodeGenerator` at `SET_OPERATION_UNION_ALL`**

A helper private method `canMergeUnionAll` decides if we go with `MergeUnionAll` or `UnionAll`. It checks:
- whether union branches have designated timestamps of the same type
- whether union branches have the same scan direction
- whether a timestamp order was requested and whether it matches the scan direction

**Part III: Change the `SqlOptimiser` to advice on the window's OVER order-by**
We change the `optimiseOrderBy` method to

1. Capture the window's OVER order-by and inject it as model advice
2. Pushdown of order-by advice to UNION-ALLs

## Implementation Reasoning
There's a reason why we went with the above implementation, especially **Part III**. Theoretically, we do not need **Part III**. But some findings led us to that route. First, we need a pass to capture the window's OVER order-by information because that is not present in the model order-by advice. That place does not need to be in the optimizer, but it turned out to be a good place. 

Some odd behaviors on UNIONs scan orders were found:
```bash
EXPLAIN (SELECT * FROM core_price ORDER BY timestamp DESC)
UNION ALL (SELECT * FROM core_price ORDER by timestamp DESC)
ORDER BY timestamp DESC;

Encode sort
  keys: [timestamp desc]
    Union All
        PageFrame
            Row backward scan
            Frame backward scan on: core_price
        PageFrame
            Row forward scan
            Frame forward scan on: core_price
```

We have explicit inner and outer order-bys and still we get a forward scan on the second branch. Something inside the optimizer ignores those order-bys. That means, if we only check for scan directions to decide on the `MergeUnionAll`, we may never see the `MergeUnionAll`, because of that odd behavior. 

So we opted to push down order-by advice to UNION-ALLs branches, so they behave uniformly and consistently. Since we are already pushing down order-by advice, it becomes convenient to push down, as well, any window's OVER order-by that we may have. 

By doing those changes in the optimizer, we have a uniform way of dealing with non-windowed and windowed cases. 

## Cases covered

| Query                                                                 | Plan                              |
| --------------------------------------------------------------------- | --------------------------------- |
| `(a) UNION ALL (b) ORDER BY ts` / `ts DESC`                           | `Union All Merge`, no sort        |
| chained `a UNION ALL b UNION ALL c ORDER BY ts [DESC]`                | nested `Union All Merge`, no sort |
| inner/nested unions (left/right grouped, union in a subquery branch)  | nested `Union All Merge`          |
| window `OVER (ORDER BY ts [DESC])` over union                         | `Window` + `Union All Merge`      |
| window `OVER (PARTITION BY sym ORDER BY ts)` over union (symbol/cast) | `Window` + `Union All Merge`      |
| union with no ts order requested                                      | concat `Union All` (unchanged)    |
| ts-type/precision mismatch across branches                            | concat + `Sort` (declined)        |
## Review orientation
- Review each one of those parts separately and how they fit together to build the entire solution
- Review tests
- Review benchmark
## Testing
- `core/.../griffin/MergeUnionAllTest.java` tests the new code
- Regression clean 
- A bunch of cases manually tested. Check the appendix section.

### Union All Merge vs Union All + Sort
|rows/branch|`mergeFullScan`|`sortFullScan`|full-scan speedup|`mergeLimit10`|`sortLimit10`|limit-10 speedup|
|---|---|---|---|---|---|---|
|100k|3.33|28.09|8.4×|0.004|20.95|~5,200×|
|1M|33.47|261.37|7.8×|0.003|237.56|~79,000×|
|5M|143.40|1304.07|9.1×|0.004|1228.00|~307,000×|

### Streaming Window vs Materializing `CachedWindow` 
|rows/branch|`windowFullScan`|`cachedWindowFullScan`|full-scan speedup|`windowLimit10`|`cachedWindowLimit10`|limit-10 speedup|
|---|---|---|---|---|---|---|
|100k|5.28|25.56|4.8×|0.004|23.67|~5,900×|
|1M|54.48|201.74|3.7×|0.004|191.99|~48,000×|
|5M|233.30|1179.30|5.1×|0.005|1188.50|~238,000×|

## Potential Follow-Ups
I'd keep the scope of this work narrow. Here are some potential follow-ups:
### Timestamp required queries
Timestamp required queries can benefit from `MergeUnionAll`. Right now, the following queries fail because `UNION ALL` removes the timestamp, but if `MergeUnionAll` is leveraged on them, they can be run:

```sql
-- SAMPLE BY
SELECT ts, sum(px) FROM ((SELECT * FROM a) UNION ALL (SELECT * FROM b)) SAMPLE BY 2s;

-- ASOF JOIN, union as master                                          
SELECT t.sym, t.px, q.bid FROM ((SELECT * FROM a) UNION ALL (SELECT * FROM b)) t ASOF JOIN q ON (sym);
                     
-- LT JOIN
SELECT t.sym, t.px, q.bid FROM ((SELECT * FROM a) UNION ALL (SELECT * FROM b)) t LT JOIN q ON (sym);
                             
-- SPLICE JOIN
SELECT t.sym, t.px, q.bid
FROM ((SELECT * FROM a) UNION ALL (SELECT * FROM b)) t
SPLICE JOIN q ON (sym);             
```
### N-way merge
We currently nest multiple 2-way merges. A single N-way merge operator would give a better performance. 
### Timestamp precision normalization
If UNION branches disagree on timestamp precision, we don't pick `MergeUnionAll`. We could normalize the precision so they can be compared. 
## Appendix
Some queries for manual testing and investigation

```sql
CREATE TABLE a (sym SYMBOL, px DOUBLE, ts TIMESTAMP) TIMESTAMP(ts) PARTITION BY DAY;
CREATE TABLE b (sym SYMBOL, px DOUBLE, ts TIMESTAMP) TIMESTAMP(ts) PARTITION BY DAY;                                                

INSERT INTO a VALUES
  ('AAPL', 10.0, '2024-01-01T00:00:01.000000Z'),
  ('MSFT', 11.0, '2024-01-01T00:00:03.000000Z'),
  ('AAPL', 12.0, '2024-01-01T00:00:05.000000Z'),
  ('MSFT', 13.0, '2024-01-01T00:00:07.000000Z'),
  ('AAPL', 14.0, '2024-01-02T00:00:01.000000Z'),
  ('MSFT', 15.0, '2024-01-02T00:00:09.000000Z');                                               
INSERT INTO b VALUES
  ('AAPL', 20.0, '2024-01-01T00:00:02.000000Z'),
  ('MSFT', 21.0, '2024-01-01T00:00:04.000000Z'),
  ('AAPL', 22.0, '2024-01-01T00:00:05.000000Z'),
  ('MSFT', 23.0, '2024-01-01T00:00:06.000000Z'),
  ('AAPL', 24.0, '2024-01-02T00:00:03.000000Z'),
  ('MSFT', 25.0, '2024-01-02T00:00:09.000000Z');

-- 1. Explicit ORDER BY. The outer order-by wins over inner order-bys
EXPLAIN (SELECT * FROM a) UNION ALL (SELECT * FROM b) ORDER BY ts;

EXPLAIN (SELECT * FROM a) UNION ALL (SELECT * FROM b) ORDER BY ts DESC;

-- 2. No explicit outer ORDER BY, defaults to UNION ALL
EXPLAIN (a ORDER BY ts DESC) UNION ALL (b ORDER BY ts DESC);

-- 3. Implicit ORDER BY via windowed OVER
EXPLAIN SELECT sym, px, first_value(px) OVER (PARTITION BY sym ORDER BY ts) FROM ((SELECT * FROM a) UNION ALL (SELECT * FROM b));

EXPLAIN SELECT sym, px, first_value(px) OVER (PARTITION BY sym ORDER BY ts DESC) FROM ((SELECT * FROM a ORDER BY ts DESC) UNION ALL (SELECT * FROM b ORDER BY ts DESC));
       
--- 4. Nested
EXPLAIN a UNION ALL b UNION ALL A ORDER BY ts DESC;

```


bluestreak01:
@brunocalza one gap from review before merge: **no test exercises the cast-required merge path**.

All 16 tests in `MergeUnionAllTest` use identical-type columns, so `castIsRequired == false` in every case. That means the `UnionCastRecord` branch of `MergeUnionAllRecordCursor` and the `widenSetMetadata` / `generateCastFunctions` branch of `generateSetFactory` (`SqlCodeGenerator.java:10058-10066`) are never covered. This is the path most likely to hide a bug, because the merge reads the **raw** record timestamp (`recordA.getLong(timestampIndex)`) while the other columns flow through cast functions.

I verified manually that it currently works (asc and desc), but there is no regression guard. Could you add a test with a widened non-timestamp column, e.g.:

```sql
create table a (px int, ts timestamp) timestamp(ts) partition by day;
create table b (px double, ts timestamp) timestamp(ts) partition by day;
-- (select * from a) union all (select * from b) order by ts   -> Union All Merge, px widened to double
```

Assert both `order by ts` and `order by ts desc`, checking `.withPlanContaining("Union All Merge")` and the widened `px` values via `.returns(...)`.

brunocalza:
Thanks @bluestreak01, pushed another commit with the test

bluestreak01 (approved):

bluestreak01 (changes requested):
@brunocalza — deep re-review of this PR. The merge operator itself is correct and the two scariest hypotheses turned out to be false alarms (details at the bottom), but there's a hot-path inefficiency and a strictly-worse `calculateSize`, plus a few untested edges the change introduces. Since this is a `perf(sql)` PR, those cross the "best known algorithm + prove the new behavior" bar, so I'm requesting changes rather than nit-approving.

## Must fix (all small, all in this PR)

### 1. `MergeUnionAllRecordCursor.calculateSize()` is strictly worse than the default it overrides
`MergeUnionAllRecordCursor.java:63-72`. The override drains via `while (hasNext()) counter.inc()`, which (a) drops the per-row `circuitBreaker.statefulThrowExceptionIfTripped()` the `RecordCursor` default performs, and (b) pays the full per-row merge comparison cost to produce a count that is order-independent (`countA + countB`). The concat sibling delegates to the branches and keeps both properties. This path is reached on filtered unions via pgwire/REST `count`. Suggested fix restores the breaker *and* the O(frames) fast path:
```java
@Override
public void calculateSize(SqlExecutionCircuitBreaker circuitBreaker, Counter counter) {
    if (!isStarted) { // fresh cursor: nothing buffered, branch fast paths + breaker are exact
        cursorA.calculateSize(circuitBreaker, counter);
        cursorB.calculateSize(circuitBreaker, counter);
        return;
    }
    while (hasNext()) { counter.inc(); } // mid-iteration fallback (callers toTop() first, so unused today)
}
```

### 2. `hasNext()` re-reads BOTH branch timestamps every row
`MergeUnionAllRecordCursor.java:88-96`. Each `hasNext()` advances exactly one leg, so the other leg's buffered timestamp is unchanged — re-reading it is a wasted virtual `getLong` on the merge's innermost loop, in exactly the fully-interleaved case the benchmark is built around. Cache `tsA`/`tsB` in fields and refresh only the advanced leg (correctness holds through ties). Roughly halves the per-row timestamp reads.

### 3. Missing tests for behavior this PR introduces
- **Timestamp-preservation contract** — the headline change is that `MergeUnionAll` preserves the designated timestamp (concat stripped it), which is what newly enables SAMPLE BY / ASOF-LT-SPLICE-as-master over a union. It's asserted only indirectly in `SqlOptimiserTest`; a regression that dropped `setTimestampIndex` but kept row order would pass every `MergeUnionAllTest`. Add a `MergeUnionAllTest` case that projects `ts` and pins `.timestamp("ts")`.
- **Empty branch** — no test has a 0-row branch, so the `hasNext()` terminal transitions (`!isStarted` with one branch empty, final `return false`) are unexercised.
- **LIMIT-in-branch invariant** — safe today (verified below), but load-bearing on model wrapping; one `(a) union all (b limit 3) order by ts desc` test with `.returns(...)` locks it cheaply.
- Cancellation behavior of the `calculateSize` path from item 1.

## Minor
- `withPlanNotContaining("Sort")` at `MergeUnionAllTest.java:127,172,195` never matches the actual `Encode sort` node (`Chars.contains` is case-sensitive), so the token asserts nothing — use `"Encode sort"` like the sibling tests. Harmless in practice only because `.noRandomAccess()` independently catches a stacked encoded sort, but `testExplicitOrderByTsMergesAndElidesSort` should say what it means.
- `MergeUnionAllRecordCursorFactory.getOperation():87` is missing `@Override`.
- `MergeUnionAllBenchmark` `@Param` numbers could use `_` separators.

## Behavioral notes worth a line in the description
- Window-over-union with no explicit outer `ORDER BY` now emits ts-merged order instead of concat order (SQL-legal, values unchanged, but observable).
- `canMergeUnionAll` trusts `getScanDirection()` + ts-index to reflect emission order without requiring the optimiser push-down ran — same invariant the existing sort-elision relies on, but load-bearing across two files; an assert-build order check at branch boundaries would harden it.

## Verified safe (so you don't have to re-chase these)
I built the branch and probed a running engine:
- **LIMIT in a branch does not change its result set.** For `(a) union all (b limit 3) order by ts desc`, the injected `Encode sort` lands *above* the branch `Limit take-rows:3` (limit-then-sort), so `b` keeps its first-3 storage rows. `pushTimestampOrderIntoUnionBranches` mutates the branch wrapper model, so the order can't sink below the limit. A one-line code comment stating this invariant would help. The missing `getLimitLo()` check is not a bug.
- **DESC ROWS-frame window over a merged union is not wrong.** All-ties DESC rows-frame returns identical output to the equivalent single-table query; any intra-tie ordering difference is SQL-unspecified.
- Tie rule (A-before-B, both directions), cast-required path (raw ts read is safe because the ts column type must match), chained/nested merges, and resource cleanup on error paths are all correct.

Coverage map: 13 behavioral changes, 10 tested, 3 untested (empty branch, LIMIT invariant, calculateSize cancellation). No correctness/leak/concurrency blocker — the requested changes are the two hot-path fixes plus the missing tests.

brunocalza:
Thanks for reviewing. @bluestreak01, I pushed a new commit with
- re-implementation calculateSize for a fast path when the cursor hasn't started yet
- caching of `tsA` and `tsB` on `hasNext`
- new tests as proposed by review

bluestreak01 (approved):
Reviewing PR #7348 at level 3.

### Critical
None.

### Moderate
None.

### Minor
None.

### Adjacent findings
None.

### Coverage map
Test gate: **pass**. Admitted coverage gaps: **0**.

Validation:
- `MergeUnionAllTest` and `SqlOptimiserTest`: 213 tests passed.
- `UnionTest`: 50 tests passed.
- `git diff --check`: passed.
- PR CI checks: passing.
- Head/base probes confirmed outer-order barriers, staged N-way cast widening, and CTAS timestamp metadata.

### Summary
- **Verdict: approve**
- Correctness gate: pass.
- Severity distribution: Critical 0, Moderate 0, Minor 0.
- Attribution: in-diff 0; out-of-diff-breakage 0.
- Submodule provenance: no submodule pointers changed.
- Metadata: title and labels match the SQL performance scope.

Tradeoff: an ordered nested union may now preserve timestamp order through an otherwise unordered outer `UNION ALL`, changing unspecified output order while avoiding sorting. No SQL contract or user-visible regression was found.

PR #7349, opened 2026-07-03
fix(core): fix POSTING index covering scans failing when queries migrate across worker threads

nwoolmer:
## What

Fixes the assertion failure

```
java.lang.AssertionError: posting index cursor closed off the reader's owning thread
    at io.questdb.cairo.idx.PostingIndexFwdReader$Cursor.close(PostingIndexFwdReader.java:214)
```

and its downstream failures on assertions-enabled builds: a multi-worker HTTP parquet export (`/exp?fmt=parquet`) of a POSTING covering scan (`symbol IN (...)`) aborts mid-close and leaks the `TableReader`, and the pgwire covering-record-cursor path double-closes the reader (`CairoException: double close`), dropping the client connection. Builds without `-ea` are unaffected.

The branch carries the failing reproducer (first commit) and the fix (second commit): `ExpParquetExportTest#testParquetExportPageFramePostingIndexCoveringScanMultiWorker`.

## Root cause

Suspendable queries migrate the connection — and the `TableReader` it holds — across worker threads between fragments, with the event loop serializing the handoff. A posting index row cursor checked out on one worker can therefore legitimately close on another: on a multi-worker HTTP server the parquet export streams across workers, and `HttpConnectionContext.reset()` closes the `MultiKeyCoveringPageFrameCursor` (and its merge row cursors) on whichever worker last drove the connection.

The assertion-only tripwire that #7256 added to every posting cursor `close()` (`assertSameOperatingThread`) encoded the premise that a reader is driven by one OS thread between reader-pool acquire and release. That premise does not hold for suspendable queries, so the assert fires on sanctioned behavior. The existing export tests pin `workerCount=1`, which is why they never caught it.

## Fix

`getCursor()` now stamps the reader's operating thread unconditionally (previously the stamp was written only under `-ea`), and each of the four cursor `close()` paths (Fwd/Bwd × Cursor/NullCursor) re-pools into `freeCursors` only when the close runs on that thread. The unsynchronized `isOpen()`-then-`freeCursors.add(this)` check-then-act — the race the assert was guarding — thereby stays serialized with `getCursor()`. An off-thread close falls through to `releaseResources()`, which frees only cursor-local buffers and touches no reader-shared state, so it is safe from any thread.

This also changes the outcome of the stale-cursor hazard the assert was built to flag (a cursor outliving its reader's release to the reader pool, with another thread re-acquiring and re-stamping the reader): instead of an `AssertionError` that aborts the orderly close, the stale close degrades to a local release and cannot race the new owner's `getCursor()`. The `isOpen()` leak guard from #7256 is unchanged, and its four regression tests still pass.

## Tradeoffs

- An off-thread close skips cursor pooling, so the next `getCursor()` on that reader reallocates the decode block buffer instead of reusing a pooled one. Mid-query closes run on the driving thread and still pool; the release-instead-of-pool case occurs once per worker migration.
- The operating-thread stamp is now written on every `getCursor()` call in all builds (one thread-id read and one plain field write), not just under `-ea`.

## Test plan

The off-thread close hazard is not specific to the multi-key export path; the branch carries one regression per cursor holder, each asserting the execution plan and the query result on either side of the off-thread close, and each verified red on the pre-fix tree (reproducer commit + test changes only):

- `ExpParquetExportTest#testParquetExportPageFramePostingIndexCoveringScanMultiWorker` — the concurrent phase drives `MultiKeyCoveringPageFrameCursor` (red on master: 1–4 of 4 exports fail per run); a verification phase asserts the `CoveringIndex` EXPLAIN plan for both the `IN`-list and single-literal shapes, exports each once more, and compares the parquet bytes against the source query via `read_parquet()` — the single literal drives the previously untested `SingleKeyCoveringPageFrameCursor`.
- `PostingIndexCriticalIssuesTest#testCoveringRecordCursorSingleKeyClosedOffOperatingThread` / `#testCoveringRecordCursorMultiKeyClosedOffOperatingThread` / `#testPostingIndexRowCursorClosedOffOperatingThread` — cairo-level regressions for the record-cursor holders (`CoveringCursor` single/multi key, and the plain index scan retained by `PageFrameRecordCursorImpl`): check a cursor out, partially drain, close it on a second thread, then assert no close error, no busy reader left in the pool, the same plan and the same aggregates afterwards, under `assertMemoryLeak`. All three fail on the pre-fix tree with "posting index cursor closed off the reader's owning thread" as the cause. The tests use `getBaseFactory()` because the `QueryProgress` cursor wrapper catches close errors and recovers the leaked reader, masking the failure; the production teardown paths (`HttpConnectionContext.reset()`, pgwire covering cursor close) have no such guard.
- `mvn -pl core -Dtest=PostingIndexCriticalIssuesTest test` — 122 tests pass, including the four #7256 block-buffer leak regressions and the concurrent covering LATEST BY fuzz.
- `mvn -pl core -Dtest=ExpParquetExportTest test` — 75 of 78 pass; the 3 errors (`testParquetExportReadOnlyHttp`, `testParquetExportLimit`, `testParquetExportDisabledReadOnlyInstance`) are local-environment port collisions on 9009 (a separate server instance was running) and reproduce identically without this change.

Paths audited and found not exposed: `LatestByValueIndexedRowCursorFactory` (opens and closes the index cursor within one `getCursor()` call on one thread) and `PostingIndexDistinctRecordCursorFactory` (uses the reader-level `collectDistinctKeys` API, no cursor checkout).

## Follow-up: single-key covering scans exported all-null covered columns to parquet (review C1)

`testParquetExportPageFramePostingIndexCoveringScanMultiWorker`'s verify phase exports the single-literal (`symbol = 'S3'`) covering scan and reads the parquet back with `read_parquet()`; the covered columns came back all-null (epoch-0 timestamps, empty symbols) for every row on all CI environments. The root cause is orthogonal to the off-thread-close fix and lives in the covered-decode path added by #7284.

A single-key covering scan (`sym = 'x'`) produces **metadata-only** page frames: the covered columns are decoded lazily on the async reduce workers via `PageFrameMemoryPool#patchCoveredFrameMemory`, and the raw frame page addresses are placeholders. Every zero-copy parquet export route that reads those raw addresses (`writePageFrame` → `frame.getPageAddress`) ships all-null covered columns. Only the multi-key (IN-list) merge materializes eagerly, which is why the IN-list export was unaffected.

The fix adds `RecordCursorFactory.producesMaterializedPageFrames()` (default true, delegating to the base factory), overridden to false on the single-key `CoveringIndexRecordCursorFactory`, and applies it at every export route in `ParquetExportMode.determineExportMode()`:

- **DIRECT_PAGE_FRAME** (bare covering scan): kept only when the factory both supports a page-frame cursor and materializes it.
- **PAGE_FRAME_BACKED** (a projection with a computed column over the covering scan, e.g. `SELECT sym, price * qty FROM t WHERE sym = 'x'`): the hybrid path reads the base's raw addresses for pass-through columns, so it now demands the virtual factory's base materialize; otherwise it falls to `CURSOR_BASED`.
- **`StaleViewCheckFactory`** (any query that references a view): it delegates `supportsPageFrameCursor()`/`getPageFrameCursor()` to its base but its `getBaseFactory()` returns `base.getBaseFactory()` (it skips its own base for top-K peeling), so the default `getBaseFactory()`-delegating implementation stepped over the covering scan. It now delegates `producesMaterializedPageFrames()` to its immediate base like its sibling page-frame methods.

A metadata-only producer therefore always exports row-wise through `CURSOR_BASED`, which drives the same covered decode the query path uses. Multi-key covering export stays on the zero-copy fast path.

### Tradeoff

- Single-key covering-index parquet export now streams row-wise through the record cursor instead of the zero-copy page-frame path. It decodes covered columns exactly as the equivalent `SELECT` does; this trades the throughput of that one (niche) export shape for correctness.

### Test plan (C1)

- `testParquetExportPageFramePostingIndexCoveringScanMultiWorker` — red on every CI environment at the single-key verify assertion; green with the fix. Reverting the DIRECT_PAGE_FRAME hunk reproduces the all-null export (`1970-01-01T00:00:00.000000Z null null null null`) locally, so it is a red→green regression for that route.
- `testParquetExportCoveringScanShapes` (new) — exports four covering shapes and asserts the `CoveringIndex` plan plus that `read_parquet()` of each export matches the source query: a single-key projection with a computed column (PAGE_FRAME_BACKED), a single-key query through a view (StaleViewCheckFactory), a single-key var-size covered column (CURSOR_BASED), and a multi-key IN-list over a var-size covered column (DIRECT_PAGE_FRAME, eager merge). Reverting the PAGE_FRAME_BACKED guard turns the projection shape red; reverting the StaleViewCheckFactory delegation turns the view shape red (both reproduce the empty covered columns). The multi-key var-size case confirms the eager-materialization DIRECT path is sound for var-size covered columns.
- Also corrected the `RowCursor` javadoc so the off-thread close is described precisely: a POSTING cursor releases its native buffers directly, while BITMAP / all-null cursors hold no native memory and simply drop to GC.
## Follow-up: extend the gate to the bitmap and all-null index readers (review M1)

`BitmapIndexFwdReader`, `BitmapIndexBwdReader`, `IndexFwdNullReader`, and `IndexBwdNullReader` re-pool their row cursors with the same unsynchronized `size()`-then-`freeCursors.add()` on a plain `ObjList` that this PR gates for the posting readers, and the same suspend/resume worker-migration path reaches them (`PageFrameRecordCursorImpl` retains the row cursor from `DeferredSymbolIndexRowCursorFactory` for plain `SYMBOL INDEX` columns). These cursors hold no native memory and never carried the #7256 `-ea` assert, so the serialized dispatcher handoff fences the race in practice. `SYMBOL INDEX` defaults to `BITMAP`, so this is the more common reader family; the gate hardens it against a future change that unfences the handoff, and keeps the two families consistent.

`AbstractBitmapIndexReader` now carries the operating-thread stamp. `getCursor` on both readers and `BitmapIndexFwdReader.getFrameCursor` stamp on vend; each `Cursor`/`NullCursor` `close()` re-pools only on the stamping thread and otherwise drops the cursor to GC. `IndexFwd/BwdNullReader.NullCursor` becomes a non-static inner class so it can consult the reader's stamp. The gate is fail-safe: a stamp mismatch skips reuse, it cannot corrupt the list.

`ConcurrentBitmapIndexFwdReader` stays unchanged. Its pool is a thread-safe `ConcurrentPool` (it backs `SymbolMapReaderImpl` and is designed for concurrent cursor use), so it has no unsynchronized mutation to serialize.

### Tradeoff

- An off-thread close on a bitmap or null reader now skips pooling, so the next vend allocates a fresh cursor object instead of reusing a pooled one. These cursors hold no native memory, so the cost is a small short-lived allocation, and it occurs once per worker migration.

### Test plan (follow-up)

- `PostingIndexCriticalIssuesTest#testPostingIndexBwdRowCursorClosedOffOperatingThread` / `#testPostingIndexBwdNullCursorClosedOffOperatingThread` / `#testPostingIndexFwdNullCursorClosedOffOperatingThread` — raw-reader regressions that close a parked `PostingIndexBwdReader.Cursor` and the `PostingIndexFwd/BwdReader.NullCursor` on a second thread while the reader stays open. All three fail on the pre-fix readers with "posting index [null] cursor closed off the reader's owning thread" and pass with the gate. They cover the backward `Cursor` (which frees `efRankDirAddr` on the release path) and the `NullCursor` release branches the first-round record-cursor tests did not reach.
- `PostingIndexCriticalIssuesTest#testBitmapIndexRowCursorClosedOffOperatingThread` — the same off-thread close on a default `SYMBOL INDEX` (BITMAP) scan. It passes with or without the gate: bitmap cursors hold no native memory and never had the assert, so the off-thread close never threw and the skip-pool effect is not observable through the query API. It documents the sanctioned migration path rather than guarding a crash; the posting tests above exercise the gate's mechanism red-to-green.
- `mvn -pl core -Dtest=PostingIndexCriticalIssuesTest test` — 126 tests pass (the four first-round regressions, plus these four).
- `mvn -pl core -Dtest=BitmapIndexTest,BitmapIndexFuzzTest,BitmapIndexConcurrentFuzzTest,IndexFwdNullReaderTest,IndexBwdNullReaderTest,IndexFactoryTest,IndexReloadTest,LatestByTest test` — the bitmap/null reader and symbol-index cursor suites pass. `ExpParquetExportTest` did not run in this local environment (a separate server held port 9001); CI covers it.

### Review minors

- Added `.noLeakCheck()` to the three `assertsPlanContaining` calls in the new cairo tests, so they no longer nest a second `assertMemoryLeak` and `engine.clear()` inside the outer wrapper (m2).
- Corrected the `RowCursor` javadoc that described BITMAP cursors as having a no-op `close()`; pool-backed cursors (POSTING, BITMAP, and the all-null readers) re-pool on the operating thread and release directly off it.


nwoolmer:
## Self-review (multi-angle, adversarially verified)

Ran an 8-angle review of this diff (line-by-line, removed-behavior, cross-file tracing, reuse, simplification, efficiency, altitude, conventions); every candidate finding was independently verified against the code before being kept. 3 confirmed, 1 plausible, 3 refuted.

### Confirmed

**1. Bitmap index readers carry the same ungated re-pool this PR fixes for posting readers** — `BitmapIndexFwdReader.Cursor.close()` (`BitmapIndexFwdReader.java:135`), `BitmapIndexBwdReader.java:108`, `IndexFwdNullReader.java:147`, `IndexBwdNullReader.java:136` all do `freeCursors.add(this)` on a plain `ObjList` with no operating-thread gate and no `isOpen()` guard. The same migration path reaches them: `PageFrameRecordCursorImpl` retains `rowCursor` across suspensions and its `close()` (line 96) runs on whichever worker owns the connection at teardown, via `DeferredSymbolIndexRowCursorFactory` for plain `SYMBOL INDEX` columns. Unlike posting cursors these hold no native memory and the dispatcher handoff fences the race in practice, so this is latent, not a crash — but the rationale added in this PR applies to them verbatim. The `RowCursor.java:38` javadoc claiming bitmap cursors have "a no-op close()" is also stale.

**2. The "defuses the stale-cursor hazard" comment overclaims** (`AbstractPostingIndexReader.java:1140-1143`, duplicated in `PostingIndexBwdReader.Cursor.close()`). Two holes: a stale close running on the original stamping thread *before* the new owner's first `getCursor()` re-stamps passes the gate trivially; and `operatingThreadId` is a plain non-volatile long, so even after a re-stamp the stale closer has no happens-before edge and may legally read its own old stamp. Not a live bug — every reachable close path frees cursors before the `TableReader` returns to the pool, and behavior is no worse than the pre-PR assert (which also passed in these scenarios) — but the comment should say the gate narrows the window as defense-in-depth while correctness still relies on close-before-release ordering. `volatile` would fix visibility but cannot close the before-re-stamp timing hole, so an honest comment beats a stronger fence here.

**3. The four-term re-pool gate is copy-pasted character-identically into four `close()` methods**, and the deleted assert was the only tripwire that would catch a future copy dropping the `isOperatingThread()` term. Centralizing the condition is mechanical: `AbstractCoveringCursor` can host `protected final boolean canRepool(int poolSize)` (hoisting the per-subclass `isPooled` field; `MAX_CACHED_FREE_CURSORS` is already interface-visible). The branch bodies legitimately stay per-subclass (Bwd frees `efRankDirAddr`, each pool needs its concrete type). The long rationale comment is also written out twice and should live once, on `isOperatingThread()`.

### Plausible (take-it-or-leave-it)

**4.** The test's hand-rolled builder chain could be `getExportTester().withWorkerCount(4).withSendBufferSize(2048)` (setters are last-write-wins; precedent at line 521) — but `getExportTester()` injects randomized 1–1024-byte forced fragmentation this file documents as slow on Mac/Windows, and this is the heaviest test in the class, so bypassing it is defensible. Worth a one-line comment stating why.

### Refuted during verification

- *Per-fragment pooling churn*: merge cursors close per **partition** (`closeMergeCursors()` at `CoveringIndexRecordCursorFactory.java:2123`), not per fragment; the large covering buffers are freed on the re-pool path anyway; the stamp refreshes on the next `getCursor()`. Worst case per migration is a few small block-buffer frees.
- *Stale `isPooled` memory leak*: structurally impossible — the gate's failure path falls through unconditionally to `releaseResources()` in all four close methods.
- *Test can pass on broken code*: a post-stream assert failure strands `NATIVE_INDEX_READER` allocations that `HttpQueryTestBuilder.run()`'s built-in per-tag `assertMemoryLeak` catches; mid-stream failures fire at ~7 partition boundaries per export and produced client-visible errors on every observed master run. 300k rows is load-bearing (multiple monthly partitions) and the test body runs in ~1.4s.

Follow-ups for findings 2–4 will land on this branch; finding 1 is tracked separately.

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexBwdReader.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/AbstractPostingIndexReader.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexFwdReader.java|3|3|100.00%|

bluestreak01 (changes requested):
@nwoolmer — review at level 3 (full pass: change-surface map, callsite walk, per-finding source verification). The fix is sound; requesting changes only to get an explicit decision on **M1** (see below).

## What I verified (all CONFIRMED correct)

- All four `freeCursors.add`/`freeNullCursors.add` sites are gated by `canRepool`; all four `popLast` sites clear `isPooled`; the 3-arg `getCursor` delegates to the stamped 4-arg form. No ungated posting re-pool path remains.
- Off-thread release frees everything: `Cursor.releaseResources()` frees `blockBufferAddr` + `closeCoveringResources()`; the Bwd variant also frees `efRankDirAddr`. `closeCoveringResources()` touches only cursor-local native buffers — no reader-shared collection.
- The close-before-release ordering the fix relies on holds at every caller: `CoveringCursor.close()` (527-528), `CoveringPageFrameCursor.close()` (877-878), `MultiKeyCoveringPageFrameCursor.close()` (2104-2106), and `SingleKeyCoveringPageFrameCursor` (via `pendingRowCursor`). Every callsite is SAFE.
- No double-free / stale-`isPooled` leak: an off-thread-released cursor is never added to `freeCursors`, and the gate's fall-through is unconditional.
- The change also closes a real non-`-ea` race: pre-PR (verified against `f4510e0954`) the runtime re-pool decision had no thread term, so an off-thread close performed an unsynchronized `freeCursors.add` concurrent with `getCursor`'s `popLast`. The new gate serializes re-pooling on the stamping thread.

## Critical

None. No correctness, resource-leak, or concurrency defect is introduced.

## Moderate

**M1 — Bitmap/plain index readers carry the identical ungated re-pool (out-of-diff, latent).** `BitmapIndexFwdReader.Cursor.close()` (`BitmapIndexFwdReader.java:135`), `BitmapIndexBwdReader.java:108`, `IndexFwdNullReader.java:147`, `IndexBwdNullReader.java:136`, and `ConcurrentBitmapIndexFwdReader.java:124` all do `freeCursors.add(this)` on a plain `ObjList` with no operating-thread gate. The same migration path reaches them: `PageFrameRecordCursorImpl` retains `rowCursor` from `DeferredSymbolIndexRowCursorFactory` for plain `SYMBOL INDEX` columns and closes it during connection teardown. Unlike posting cursors these hold no native memory and never had the posting-specific `-ea` assert, so this is latent (fenced by the serialized dispatcher handoff), **not a crash and not a regression from this PR**.

Per the repo's "bundle related fixes into one PR" convention this is a candidate to fold in here, since it's the same pattern reachable by the same migration path. Please either fold the gate into the bitmap/plain readers on this branch, or leave a comment stating why it stays separate (e.g. non-crashing, different reader family) and link the tracking issue. Requesting changes on this point so the decision is explicit rather than implicit.

## Minor

- **m1 — Off-thread-close test matrix is forward/non-null only.** The regressions drive forward, non-null posting cursors; the `Bwd` `Cursor`/`NullCursor` and `Fwd` `NullCursor` off-thread *release* branches aren't directly exercised. A descending-order query and a NULL-symbol query would close the matrix. Completeness suggestion, not a gap in the fix.
- **m2 — Inconsistent leak-check style in the new cairo tests.** Within each `assertMemoryLeak(() -> { ... })` body the data assertions use `assertQuery(aggSql).noLeakCheck()....returns(...)`, but the plan assertions use `assertQuery(scanSql).assertsPlanContaining(...)` without `.noLeakCheck()`, which nests a second `assertMemoryLeak` (and an `engine.clear()`) inside the outer wrapper. Harmless but the two forms in the same lambda should agree — add `.noLeakCheck()` to the `assertsPlanContaining` calls.
- **m3 — No `Fixes #NNN`.** Fine if no public issue exists; add the reference if one does.

## Downgraded (false positives dismissed after source verification)

- "Off-thread close leaks native memory" — `releaseResources()` frees `blockBufferAddr`, `efRankDirAddr` (Bwd), and all covering caches.
- "Plain non-volatile `operatingThreadId` is a concurrency bug" — serialized handoff provides happens-before; `volatile` wouldn't close the before-re-stamp timing hole, so plain is the honest choice (as the comment says).
- "Closer thread needs `Path.clearThreadLocals()`" — matches the file's existing spawned-thread pattern; the closer does no `Path`-allocating work; suite passes under `assertMemoryLeak`.
- "Skipping the pool on off-thread close strands a cursor / breaks reuse" — nothing depends on a closed cursor being re-poolable.
- "Losing the `-ea` assert removes a safety net" — the assert fired on sanctioned migration; `canRepool` centralization preserves the operating-thread guarantee.

## Summary

Root cause is correctly diagnosed and the runtime gate is the right fix: it removes the false `-ea` crash and closes the latent production `freeCursors.add` race, degrading off-thread closes to a safe local buffer release. Every close-path/callsite the correctness argument depends on was verified in source. Comments and commit messages are precise and honest about the gate's limits.

Findings: 0 critical, 1 moderate (M1), 3 minor confirmed; 5 draft findings dropped as false positives. In-diff vs out-of-diff: 3 in-diff (minor/test), 1 out-of-diff (M1).

Requesting changes: please address **M1** (fold the gate into the bitmap/plain index readers, or record an explicit rationale + tracking link for keeping it separate).

nwoolmer:
**M1** — folded the gate into the bitmap/plain readers (`9017688`). `AbstractBitmapIndexReader` carries the operating-thread stamp; `BitmapIndexFwd/BwdReader` stamp on every free-pool vend (`getCursor` + `getFrameCursor`) and all four `Cursor`/`NullCursor` `close()` paths gate re-pool on `isOperatingThread()`; `IndexFwd/BwdNullReader.NullCursor` became a non-static inner class so it can consult the stamp. Off-thread closes skip pooling and drop the cursor to GC — bitmap/null cursors hold no native memory, so nothing strands. The gate is fail-safe: a stamp mismatch skips reuse, it cannot corrupt the list.

One scope note: I left `ConcurrentBitmapIndexFwdReader` (`:124`) alone — its `freeCursors` is a `ConcurrentPool`, not a plain `ObjList`, and it is built for concurrent cursor use (backs `SymbolMapReaderImpl`), so the check-then-act there is on a thread-safe pool with nothing to serialize. The real fold-in surface was the two plain bitmap readers plus the two all-null readers.

**m1** — added `testPostingIndexBwdRowCursorClosedOffOperatingThread`, `testPostingIndexBwdNullCursorClosedOffOperatingThread`, and `testPostingIndexFwdNullCursorClosedOffOperatingThread`: raw-reader regressions that drive the backward `Cursor` (frees `efRankDirAddr` on the release path) and the Fwd/Bwd `NullCursor`. All three fail on the pre-fix readers with `posting index [null] cursor closed off the reader's owning thread` and pass with the gate. Also added `testBitmapIndexRowCursorClosedOffOperatingThread` — it passes with or without the gate (bitmap never had the assert and holds no native memory, so the off-thread close never threw and the skip-pool effect is not observable through the query API), so it documents the sanctioned path; the posting tests exercise the mechanism red-to-green.

**m2** — added `.noLeakCheck()` to the three `assertsPlanContaining` calls. **m3** — no public issue (internal report). Also corrected the stale `RowCursor` javadoc that called BITMAP `close()` a no-op.

Local: `PostingIndexCriticalIssuesTest` 126, `BitmapIndexTest` 42, `IndexFwd/BwdNullReaderTest` 4+4, `Bitmap[Concurrent]FuzzTest`, `IndexFactoryTest` 22, `LatestByTest` 116 all green. `ExpParquetExportTest` did not run here (a server held `:9001`); leaving it to CI.

bluestreak01 (dismissed):
Approve — level 3 review (change-surface map, full callsite walk, per-finding source verification).

The runtime `canRepool()` gate is the correct fix: it removes the false `-ea` `assertSameOperatingThread` crash, closes the latent production `freeCursors.add`-vs-`popLast` race, and degrades off-thread cursor closes to a thread-local `releaseResources()` that frees `blockBufferAddr`/`efRankDirAddr`/covering caches — no native-memory leak. Verified all four posting close paths route pooling through the gate, every vend path stamps before popping, the close-before-release ordering holds at every production callsite (`CoveringCursor.close()`, `closeMergeCursors()`, `closePendingCursor()`), and the M1 fold-in into the bitmap/plain readers is complete. Documented tradeoffs (one skipped pooling per worker migration; one plain stamp write per `getCursor`) are accurate and negligible.

Non-blocking minors (fine to defer): `ConcurrentBitmapIndexFwdReader` now inherits the unused stamp members; the off-thread release path leaves `isPooled == false` (benign under the close-once contract).

One open item to confirm before merge, not a code defect: the enterprise CI red on `70700114` should be shown to be infra/flake (I can't reach the Azure build to prove it either way).

nwoolmer:
@bluestreak01 the requested changes from your review are addressed:

- **M1** (fold operating-thread re-pool gate into bitmap/plain index readers): `90176887` — folds it into `BitmapIndexFwd/BwdReader` + `IndexFwd/BwdNullReader`, deliberately excludes `ConcurrentBitmapIndexFwdReader` (rationale in the commit message)
- **m1** (thin off-thread-close test matrix): `70700114` — adds off-thread-close tests for posting Bwd/NullCursor + a bitmap gate test
- **m2** (inconsistent `.noLeakCheck()` style + stale RowCursor javadoc): `5f259495`

Whenever you get a chance, would appreciate a re-review. Thanks!

bluestreak01 (approved):
# Review: PR #7349 — `fix(core): fix POSTING index covering scans failing when queries migrate across worker threads`

**Reviewed at level 3** (full mission-critical pass: all 14 agents — Agent 8/Rust N/A, no `.rs` files — plus per-finding source verification).

## What this PR does

#7256 added an `-ea`-only tripwire `assert assertSameOperatingThread()` to every posting index cursor `close()`, encoding the premise that one OS thread drives a `TableReader` between reader-pool acquire/release. That premise is false for suspendable queries: HTTP `/exp` parquet export and pgwire fragment streaming migrate the connection — and the `TableReader` its open cursor holds — across worker threads between fragments, with the event loop serializing the handoff. So a posting cursor checked out on worker A legitimately closes on worker B; under `-ea` the assert fired on sanctioned behavior, aborting the orderly close and leaking the `TableReader` (HTTP) or double-closing it (pgwire).

The fix removes the assert and replaces the unconditional re-pool with a runtime gate: `getCursor()`/`getFrameCursor()` stamp `operatingThreadId` unconditionally; each cursor `close()` re-pools into the plain (unsynchronized) `freeCursors` `ObjList` only via `canRepool()` (`!isPooled && isOperatingThread() && isOpen() && poolSize < MAX`), otherwise falls through to `releaseResources()`. The same gate is folded into the bitmap and all-null index readers for consistency (bundled M1 follow-up, clearly sectioned in the body).

## Critical

None.

## Moderate

None that block. Two verified test-completeness gaps, both with a stated reason they are inert to shipped correctness:

- **`core/.../idx/` bitmap & all-null off-thread close paths are UNTESTED off-thread** (`BitmapIndexFwdReader.NullCursor.close`, `BitmapIndexFwdReader.getFrameCursor` stamp, `BitmapIndexBwdReader.getCursor`+`Cursor`/`NullCursor.close`, `IndexFwd/BwdNullReader.NullCursor.close`). Verified inert: these paths never carried the `-ea` assert and hold **no** native memory (grep for `Unsafe.malloc`/`MemoryTag.NATIVE*` across all four readers returns nothing), so a broken gate here degrades only to "GC the cursor instead of pooling it" on the rare migration — no leak, no crash, not black-box observable. This is Moderate (an internal defensive branch reachable but hard to observe in isolation), not Critical. A backward-bitmap query test and an all-null-column scan test closed off-thread would close the literal gap; optional.

- **`IndexFwd/BwdNullReader.NullCursor` `static`→non-static-inner refactor has no direct test.** A wrong field wire-up would surface only as lost pooling (benign). Justification for non-Critical: no native state, no correctness path.

## Minor

- **`RowCursor.java:38-42` javadoc is slightly loose.** "an off-thread close releases directly instead" is precise for POSTING cursors (they call `releaseResources()`) but imprecise for BITMAP/all-null cursors, which hold no native memory and are simply skipped-from-pool and left for GC. The abstract-reader comments state this accurately; the interface javadoc could say "releases directly (POSTING) or is left for GC (BITMAP/null)". Behaviorally inert.
- **`testBitmapIndexRowCursorClosedOffOperatingThread` provides zero regression protection** (confirmed by three independent agents and the PR's own javadoc): bitmap `close()` never had the assert, so the test is green with *and* without the gate. Acceptable as labeled documentation, but it must not be counted as coverage for the gate mechanism — the six posting tests carry that.
- **`testParquetExportPageFramePostingIndexCoveringScanMultiWorker` is a probabilistic reproduction** and its `errorCount == 0` guard has weak teeth (pre-fix the server-side `AssertionError` leaks the reader but the client still receives HTTP 200). Its real teeth come from the wrapping `assertMemoryLeak` in `HttpQueryTestBuilder.run()`. Deterministic red coverage lives in the posting unit tests, so this is fine; noted so the gate isn't over-credited to this test. Adding a `getBusyReaderCount()==0`-style assertion after the concurrent phase would harden it.
- **`testPostingIndexFwdNullCursorClosedOffOperatingThread` has no native-buffer-free teeth** (its `-ea`-assert teeth are intact, so it still fails on revert): `PostingIndexFwdReader.NullCursor.hasNext()` emits implicit nulls *first* (verified: `nullPos < nullCount` before `super.hasNext()`), so 128 `next()` calls over `columnTop=256` consume only nulls and never allocate `blockBufferAddr`. The buffer-free path is deterministically covered by the Bwd null sibling (`super.hasNext()` first), so overall coverage is intact.
- **Test-code maintainability** (all Minor, no assertion weakened): the four SQL-level off-thread-close tests share a ~30-line body plus a verbatim-repeated 6-line "Peel the QueryProgress wrapper" comment — extractable into one parameterized helper; the two `closeCursorOffOperatingThread` overloads share a copy-pasted spawn/join/rethrow core (defensible given `RecordCursor` vs `RowCursor` have no common `close()` supertype); one trailing javadoc clause on the single-key test merely narrates the body.
- **PR metadata**: consider adding the **`Postgres Wire`** label — the body documents that the pgwire covering-cursor path (`CairoException: double close`, dropped connection) is a second affected protocol distinct from the HTTP `/exp` path that `REST API` covers.

## Downgraded (candidate findings dismissed after source verification)

- *"Removing the assert reintroduces the #7256 NATIVE_INDEX_READER leak"* — **false.** The #7256 guard is `isOpen()`, which `canRepool()` preserves. The assert was never the leak mitigation (its own removed comment said so).
- *"Off-thread `releaseResources()` leaks the backward reader's `efRankDirAddr`"* — **false.** `PostingIndexBwdReader.releaseResources()` frees `blockBufferAddr` **and** `efRankDirAddr` and `closeCoveringResources()`; verified it frees exactly the union the old pool-retain-plus-drain path freed.
- *"Moving `isPooled` to `AbstractCoveringCursor` causes double-pool/stale flag"* — **false.** `NullCursor extends Cursor extends AbstractCoveringCursor` already shared one field; every pop clears it, every re-pool sets it, `getDetachedCursor` never touches the pool.
- *"Unconditional stamp introduces a new race in the parallel covered-decode pipeline (#7284)"* — **false.** Workers use `getDetachedCursor` (no stamp, no `freeCursors` touch); the pooled `getCursor` runs only single-threaded in `warmForKeys` before the freeze. The plain-long field is never concurrently read+written in practice.
- *"Torn/stale read of the plain `long operatingThreadId`"* — **not reachable as a defect.** Aligned 64-bit long reads/writes are atomic on QuestDB's target JVMs (Java 11+); the only failure mode is a benign gate mismatch that skips pooling (fail-safe, never list corruption). The suspend/resume handoff is serialized through the fenced IODispatcher queues (`HttpConnectionContext.reset()` runs on the single current fd owner), so `getCursor` on A *happens-before* `close` on B.
- *"`BitmapIndexFwdReader.getFrameCursor` stamp strands or double-pools a frame cursor"* — **false.** Its only caller, `SampleByFirstLastRecordCursorFactory`, never calls `close()` on the returned `IndexFrameCursor`; the stamp is inert there and the cursor is single-owner/GC-reclaimed.

## Coverage map (behavioral changes → tests)

| Changed path | Verdict | Test / evidence (verified red-on-revert under `-ea`) |
|---|---|---|
| `PostingIndexFwdReader.Cursor.close()` gate | **TESTED** | `testCoveringRecordCursorSingleKey…` (covering→`currentRowCursor`), `…MultiKey…` (per-key `mergeCursors`), `testPostingIndexRowCursor…` (plain scan via `DeferredSymbolIndexRowCursorFactory`) — all trip the removed assert on T2 → rethrown `AssertionError` → fail on revert |
| `PostingIndexFwdReader.NullCursor.close()` gate | **TESTED** | `testPostingIndexFwdNullCursorClosedOffOperatingThread` (asserts-teeth; no buffer-free teeth — see Minor) |
| `PostingIndexBwdReader.Cursor.close()` gate | **TESTED** | `testPostingIndexBwdRowCursorClosedOffOperatingThread` (raw reader, allocates block buffer) |
| `PostingIndexBwdReader.NullCursor.close()` gate | **TESTED** | `testPostingIndexBwdNullCursorClosedOffOperatingThread` (columnTop-null path, buffer-free teeth) |
| `getCursor`/`getFrameCursor` unconditional stamp | **TESTED** | Exercised transitively by every test above + the multi-worker `/exp` integration test |
| HTTP `/exp` multi-worker migration end-to-end | **TESTED (probabilistic)** | `testParquetExportPageFramePostingIndexCoveringScanMultiWorker` (workerCount=4) + wrapping `assertMemoryLeak` |
| `BitmapIndexFwd.Cursor.close()` gate | **GREEN-ONLY** | `testBitmapIndexRowCursor…` — passes with/without fix (no assert, no native memory); documentation only |
| `BitmapIndexFwd.NullCursor` / `getFrameCursor` / `BitmapIndexBwd.*` / `IndexFwd/BwdNullReader.NullCursor` gate | **UNTESTED (inert)** | No off-thread test; verified benign (no assert, no native memory) → Moderate note above |
| `RowCursor.java` javadoc | **EXEMPT** | Doc-only, no behavioral delta |

Totals: **9 behavioral-change groups, 6 TESTED (deterministic red), 1 tested-probabilistic, 1 green-only/documentation, 1 UNTESTED-inert group.** The full off-thread-close surface that carries the `-ea` assert **and** native memory — the only class that can crash or leak — is deterministically covered.

## Summary

- **Verdict: approve.** The fix is correct under the real concurrency model and is *strictly safer* than the pre-PR behavior: the non-`-ea` path used to re-pool off-thread into a plain `ObjList` guarded only by `isOpen()`; this PR removes that off-thread mutation entirely. The load-bearing invariant (row cursor freed before the frame cursor releases the `TableReader`) is verified intact at every teardown callsite (`CoveringCursor.close`, `CoveringPageFrameCursor.close`/`closePendingCursor`, multi-key `closeMergeCursors`, `PageFrameRecordCursorImpl.close`). Off-thread `releaseResources()` frees exactly the same native memory as the old pool-and-drain path (including the backward reader's `efRankDirAddr`), is idempotent, and touches no reader-shared state. No leak, no double-free, no BROKEN callsite.
- **Correctness & performance gate: PASS** — zero confirmed findings touching correctness, data integrity, resource safety, concurrency, or performance/IO on a reachable path. The added per-frame cost is one plain-long store + one `threadId()` compare (not per-row; verified `getCursor` fires per (key, partition-frame)). The documented release-then-reallocate tradeoff fires only at rare suspend/resume boundaries.
- **Test gate: PASS** — the six posting regression tests have verified failure links (revert re-adds the assert; off-thread close on T2 throws; `closeCursorOffOperatingThread` rethrows → test fails). No UNTESTED **Critical** row; the UNTESTED rows are the inert bitmap/null branches (no assert, no native memory).
- **Tradeoff worth surfacing (not a blocker):** the removed assert also flagged a genuinely distinct lifecycle bug (a cursor outliving its reader across a pool re-acquire), which now fail-safes to `releaseResources()` instead of failing loudly in CI. The plain-long gate does not fully close that race (the author documents this), so correctness for that class now rests entirely on the close-ordering invariant. This is inherent — the assert's premise was wrong and had to go; preserving a tripwire would require a logical-owner token rather than a thread id. Acceptable as a deliberate, documented decision.
- **Findings:** ~7 candidate concerns raised across agents, all verified → 0 Critical, 0 blocking Moderate, several Minor; ~6 draft false-positives dismissed against source. In-diff / out-of-diff / pre-existing split: all confirmed items in-diff or test-code; **0 out-of-diff breakage, 0 pre-existing bugs** surfaced in the visited callsite/exposure trail.
- **Empirical limitation:** the ~40-min build/test suite was not executed; the regression-test verdicts are from the conclusive static revert-chain analysis (the removed `-ea` assert is directly visible in the diff), and `core/pom.xml` confirms tests run under `-ea`.

Recommended (all optional, none blocking): add the `Postgres Wire` label; tighten the `RowCursor` javadoc for BITMAP/null cursors; if cheap, add a backward-bitmap and an all-null-column off-thread-close test to close the literal coverage gap.

bluestreak01 (changes requested):
@nwoolmer

**This review supersedes my earlier approval on this PR.** That approval was issued from a static level-3 analysis without the ~40-minute build/test run (I flagged the empirical gap explicitly). CI has now produced decisive new evidence that overturns it: the PR's own new test fails deterministically with a **data-correctness** mismatch. Under the review correctness gate, a confirmed data-integrity finding forces "request changes."

## Critical (blocking) — CI-confirmed

### C1. Single-key POSTING covering scan exported to parquet returns all-null covered columns

`ExpParquetExportTest.testParquetExportPageFramePostingIndexCoveringScanMultiWorker` — the new test this PR adds — **fails on every CI environment**: `SelfHosted Other tests (A)` on `linux-arm64`, `linux-x64-zfs`, `linux-x86-graal`, and the `Running tests with cover on linux-other` cover job (build 249481). All four fail at the identical assertion:

```
ExpParquetExportTest.testParquetExportPageFramePostingIndexCoveringScanMultiWorker
  :2006 -> lambda$...$0:2092 -> assertParquetMatchesQuery:2523
```

The failure is in the **verify phase**, not the concurrent phase (`errorCount == 0` / `successCount == 4` passed). Extracted from the surefire output on all three platform logs, the re-read of the exported parquet (`read_parquet(...)`) yields, for **all 37,500 rows**:

```
expected (source query):  2025-01-01T00:02:00.000000Z  0.72487...  ...  S3   (real data)
but was  (read_parquet):  1970-01-01T00:00:00.000000Z  null null null null   (empty symbol)
```

Row **count and schema are correct** (37,500 rows, 6 columns); every **value** is null/epoch-0/empty. This is not truncation and not row-ordering — it is a structurally valid parquet whose covered-column data was never materialized.

**Which path:** the failing verify query is the single-literal one (`WHERE symbol = 'S3'`; the entire expected block is S3), i.e. `SingleKeyCoveringPageFrameCursor`. The IN-list multi-key verify (`verifyQueries[0]`) passed. That split is the tell: `CoveringIndexRecordCursorFactory` documents single-key frame production as **metadata-only** ("decode happens on the workers"), while the multi-key merge **materializes eagerly**. So the parquet copy path drains single-key frames without ever running the worker-side covered decode, and exports null covered columns. The PR description itself notes the single literal "drives the previously untested `SingleKeyCoveringPageFrameCursor`" — this test is the first to exercise single-key covering + parquet export, and it exposes that the covered columns come out null.

**Determinism rules out a race:** identical failure on arm64, x64-zfs, x86-graal, and the cover job cannot be a timing/scheduling artifact. This is a logic bug in the single-key covering → parquet decode path, surfaced (correctly) by the new test, and **not fixed by this PR** — so the PR ships a red test.

**Required for this PR (do not defer):** make the single-key covering-scan parquet export materialize its covered columns (drive the same covered decode the query/aggregate path uses, or route the parquet copy through the eager-materialization path the multi-key case already uses), so `read_parquet` of the exported bytes equals the source query. Then the new test must go green. If, on investigation, the all-null export turns out to be a pre-existing defect in `SingleKeyCoveringPageFrameCursor`'s parquet path independent of the off-thread-close change, it is still in scope here — the fix and its now-passing regression test belong in this PR, since this branch is what surfaced it and shares its CI lineage.

## Note on the rest of the change

The operating-thread re-pool gate itself (the actual subject of the PR — replacing the `-ea` `assertSameOperatingThread()` tripwire with `canRepool()`/`isOperatingThread()`, and the bitmap/null fold-in) held up under the full level-3 static pass: no leak, no double-free, no BROKEN callsite, correct teardown ordering, and the six posting off-thread-close unit tests are genuine regression tests (all Cairo CI jobs are green, confirming `PostingIndexCriticalIssuesTest` passes). That work is sound. The blocker is strictly C1: the new end-to-end test the PR added does not pass, and it fails on a real data-correctness problem in the single-key covering parquet export.

Everything below is unchanged from the prior analysis and remains non-blocking relative to C1.

## Moderate / Minor (unchanged, non-blocking)

- Bitmap & all-null off-thread close paths are untested off-thread, but verified inert (no `-ea` assert, no native memory) — Moderate coverage note.
- `testBitmapIndexRowCursorClosedOffOperatingThread` provides zero regression protection (green with or without the gate); acceptable as labeled documentation, do not count as gate coverage.
- `testPostingIndexFwdNullCursorClosedOffOperatingThread` never allocates the block buffer (Fwd `NullCursor.hasNext()` emits nulls first), so it has assert-teeth but no buffer-free teeth; the Bwd null sibling covers the buffer-free path.
- `RowCursor.java` javadoc "releases directly instead" is loose for BITMAP/null cursors (they drop to GC, hold no native memory).
- Consider adding the `Postgres Wire` label (the body documents the pgwire double-close as a second affected protocol).
- Test-code maintainability: the four SQL-level off-thread-close tests share a ~30-line body + a verbatim-repeated 6-line comment (extractable helper); the two `closeCursorOffOperatingThread` overloads share a copy-pasted spawn/join/rethrow core.

## Summary

- **Verdict: request changes.** One Critical, CI-confirmed: single-key POSTING covering scan exported to parquet returns all-null covered columns (`SingleKeyCoveringPageFrameCursor` metadata-only frames never decoded on the parquet copy path), deterministic on all four CI environments. The PR's own new test catches it and it is not fixed.
- **Correctness & data-integrity gate: FAIL** — a confirmed data-correctness defect is open on a reachable path.
- **Test gate: FAIL** — the PR ships a test (`testParquetExportPageFramePostingIndexCoveringScanMultiWorker`) that is red in CI.
- The core re-pool gate change is correct and its posting unit tests are valid regressions (Cairo CI green); the blocker is orthogonal to the concurrency fix — it is the covered-column materialization on the single-key parquet export path.
- Fix C1 in this PR, get the new export test green in CI, then re-request review.

(Process note: I had approved and applied `QUEUED FOR MERGE`; I have cleared both `QUEUED FOR MERGE` and `READY` as part of this rejection so the red PR cannot merge.)

nwoolmer:
@bluestreak01 — **C1 is fixed**, and your root-cause read was exactly right.

A single-key POSTING covering scan (`WHERE symbol = 'S3'`) produces **metadata-only page frames** — the covered columns are decoded lazily on the async reduce workers via `PageFrameMemoryPool`, and the raw frame page addresses are placeholders. The parquet `DIRECT_PAGE_FRAME` export drained those frames without ever running the worker-side decode, so it shipped all-null covered columns, epoch-0 timestamps and empty symbols for every row — structurally valid parquet whose data was never materialized. The multi-key IN-list path materializes eagerly, which is why only the single-literal verify failed.

**Fix:** `1a9b59b6d2` adds `RecordCursorFactory.producesMaterializedPageFrames()` (default true) and overrides it to `false` for the single-key covering path; `determineExportMode()` now keeps `DIRECT_PAGE_FRAME` only when the factory both supports a page-frame cursor **and** materializes it, otherwise it falls through to the row-wise cursor path that drives the same covered decode as the query path. `805b6d77db` extends the same guard to the other two covering export routes. `testParquetExportPageFramePostingIndexCoveringScanMultiWorker` is now green, and reverting the production hunks reproduces the all-null export.

I also independently reproduced the failure on the CI merge tree before the fix (all-null read-back on the single-key verify, identical `assertParquetMatchesQuery` assertion) and confirmed it passes after — so this is the correct fix, in scope on this branch as you noted.

**On the Moderate coverage note** (bitmap / plain-null off-thread close untested): now addressed in `671d11aac8`. It adds genuine gate-regression tests for `BitmapIndexFwdReader`, `IndexFwdNullReader` and `IndexBwdNullReader` that make the re-pool gate observable through cursor object identity — an on-thread close re-pools (`assertSame` on the next `getCursor()`), an off-thread close must not (`assertNotSame`). Removing `isOperatingThread()` from the three exercised close sites makes every `assertNotSame` fail, so they genuinely guard the gate rather than documenting it (unlike the prior `testBitmapIndexRowCursorClosedOffOperatingThread`, which stays as labeled documentation — your Minor note stands).

bluestreak01 (approved):
# Review — PR #7349: `fix(core): fix POSTING index covering scans failing when queries migrate across worker threads`

**Reviewed at level 3** (full mission-critical pass: all applicable agents + per-finding verification).

## Verdict: approve

Three bundled fixes, all sound and conservative:
- **Part A** — replaces an `-ea`-only owning-thread *assert* (added by #7256) on posting cursor `close()` with an always-on operating-thread *gate* (`canRepool()`); an off-thread close now falls through to `releaseResources()` instead of aborting the close. Fixes the `AssertionError` that leaked a `TableReader` on multi-worker HTTP parquet export and double-closed the reader on the pgwire path.
- **Part B (M1)** — extends the same gate to the bitmap and all-null index readers (latent, not a crash — those cursors hold no native memory).
- **Part C (C1)** — adds `RecordCursorFactory.producesMaterializedPageFrames()` and routes metadata-only single-key covering scans through the row-wise export path, fixing all-null covered columns in parquet export (a pre-existing data-correctness bug from #7284).

## Critical
None.

## Moderate
Two coverage nuances — neither blocks, each carries the one-sentence inert justification required by the rubric:

1. **No pgwire end-to-end regression test for the cited `CairoException: double close` symptom** (`PostingIndexCriticalIssuesTest` / `ExpParquetExportTest`). *Inert:* the fix (`isOperatingThread()`/`canRepool()`) lives entirely in the protocol-agnostic reader and the identical off-thread-close code path is red-tested at the cairo level (`testCoveringRecordCursor*ClosedOffOperatingThread`) and over HTTP `/exp`; the pgwire suspend/resume plumbing only re-drives that same reader teardown, so no reader behavior is left unexercised. Adding a pgwire covering-scan case would still be worth it while in here.

2. **The four POSTING off-thread tests pin the `-ea` AssertionError-abort symptom and reader survival, but do not directly assert the re-pool *skip* for POSTING** (object identity), unlike the bitmap/null identity tests. *Inert:* `canRepool()` is shared code and its `isOperatingThread()` term is red-green–verified by `BitmapIndexTest`/`IndexFwd/BwdNullReaderTest.testCursorNotRepooledWhenClosedOffOperatingThread` (`assertNotSame` fails without the gate), so the mechanism is covered; the POSTING tests additionally prove the reported production symptom and `assertMemoryLeak` balance.

## Minor
1. **Member-ordering deviation.** `producesMaterializedPageFrames()` is placed in the `supports*` block (`RecordCursorFactory.java:410`, `CoveringIndexRecordCursorFactory.java:408`) rather than its alphabetical slot among `p` methods (before `recordCursorSupportsLongTopK`). `StaleViewCheckFactory.java:134` got the alphabetical slot right, so the three files are inconsistent. Defensibly a deliberate semantic grouping with `supportsPageFrameCursor`, and the interface already carries pre-existing inversions — low stakes, but per CLAUDE.md it should sit in alphabetical order.
2. **Re-pool condition duplicated across 6 bitmap/null `close()` methods.** The posting readers centralize it as `AbstractCoveringCursor.canRepool()`, but `BitmapIndexFwd/BwdReader` (x2 each) and `IndexFwd/BwdNullReader` inline `!isPooled && isOperatingThread() && freeCursors.size() < MAX` with near-duplicate comments. The four bitmap cursors share `AbstractBitmapIndexReader` and could host a `canRepool` helper there.
3. **Cross-file test-helper duplication.** `closeOffThread(RowCursor)` is copied verbatim into `BitmapIndexTest`, `IndexFwdNullReaderTest`, `IndexBwdNullReaderTest`, and `closeCursorOffOperatingThread` (2 overloads) lives in `PostingIndexCriticalIssuesTest`. A single `TestUtils.closeOnNewThread(QuietCloseable)` (Throwable-capturing variant canonical) would replace all five. Also the ~6-line "Peel the QueryProgress wrapper" comment is duplicated across four `PostingIndexCriticalIssuesTest` methods.
4. **Unenforced convention (doc).** The export router's correctness now depends on two conventions the type system does not enforce: any future metadata-only page-frame producer must override `producesMaterializedPageFrames()`, and any future wrapper whose `getBaseFactory()` skips a level must override it like `StaleViewCheckFactory`. A one-line reminder in the interface javadoc would harden this against a future forgetful wrapper silently reopening the all-null export.
5. **Test-comment mislabel.** In `testParquetExportCoveringScanShapes`, query 0 is labeled `-> PAGE_FRAME_BACKED`; it *enters* that branch but *exits* via `CURSOR_BASED` after the new guard. The comment names the branch under test, not the final route — mildly misleading; the assertion itself is correct.

## Downgraded (candidate concerns investigated and cleared)
- *"Off-thread close leaks `efRankDirAddr` (Bwd, freed only in the pool branch)."* **False positive** — `PostingIndexBwdReader.Cursor.releaseResources()` (lines 1034-1046) frees both `blockBufferAddr` and `efRankDirAddr`; the off-thread path routes there. Verified by direct read.
- *"C1 routing could downgrade a normal materialized DIRECT export to CURSOR_BASED (perf regression) via getBaseFactory() chain."* **False positive** — only the single-key covering factory and `StaleViewCheckFactory` ever originate `false`; every other wrapper returns its immediate base, and the default (`base == null -> true`) leaves plain scans and multi-key covering on DIRECT. The change is purely restrictive.
- *"Plain-long `operatingThreadId` is a data race."* **Cleared with nuance** — the field is defense-in-depth, not a concurrency primitive; the plain readers are never driven concurrently (single-owner `TableReader` handoff is serialized; the genuinely-concurrent covered-decode path uses `getDetachedCursor` which bypasses the gate, and `ConcurrentBitmapIndexFwdReader` uses a `ConcurrentPool`). The PR's own comments state these limits honestly. `threadId()` reuse after thread death is benign for the same reason.
- *"`canRepool(freeCursors.size())` reads reader-shared `freeCursors.size()` off-thread, contradicting the 'touches no reader-shared state' comment."* **Cleared** — an atomic `int` read whose value is discarded when `isOperatingThread()` is false; harmless, comment is very slightly imprecise for the posting path only.
- *"`testBitmapIndexRowCursorClosedOffOperatingThread` is vacuous."* **Confirmed but not a defect** — the PR openly documents it as documentation-only (green with/without the gate); the real M1 red-green lives in the identity tests. Not a regression test, correctly labeled.

## Coverage map
Every behavioral change is TESTED with a verified failure link (revert-the-hunk -> red):

| # | Change | Test | Failure link | Verdict |
|---|--------|------|--------------|---------|
| A1 | `PostingIndexFwdReader.Cursor.close()` off-thread -> release, not assert-abort | `testCoveringRecordCursorSingleKeyClosedOffOperatingThread`; single-key verify in `...MultiWorker` | pre-fix `-ea` assert throws -> rethrown as AssertionError; `busyReaderCount==0` + `assertMemoryLeak` | TESTED |
| A2 | `PostingIndexFwdReader.NullCursor.close()` off-thread | `testPostingIndexFwdNullCursorClosedOffOperatingThread` | off-thread close threw pre-fix; reader survives + leak check | TESTED |
| A3 | `PostingIndexBwdReader.Cursor.close()` off-thread (frees `efRankDirAddr`) | `testPostingIndexBwdRowCursorClosedOffOperatingThread` | same; `assertMemoryLeak` proves buffer freed | TESTED |
| A4 | `PostingIndexBwdReader.NullCursor.close()` off-thread | `testPostingIndexBwdNullCursorClosedOffOperatingThread` | same | TESTED |
| A5 | Multi-key covering cursor off-thread (1 cursor/key) | `testCoveringRecordCursorMultiKeyClosedOffOperatingThread` | pre-fix threw on first retained cursor, stranded rest | TESTED |
| A6 | Plain index scan (`PageFrameRecordCursorImpl` retains row cursor) off-thread | `testPostingIndexRowCursorClosedOffOperatingThread` | same, no covering machinery | TESTED |
| A7 | Unconditional `stampOperatingThread()` on every `getCursor()` | all A-tests (stamp is prerequisite) | without stamp the gate can't discriminate | TESTED (mechanism) |
| B1 | `BitmapIndexFwdReader.Cursor` off-thread skip re-pool | `BitmapIndexTest.testCursorNotRepooledWhenClosedOffOperatingThread` | `assertNotSame(c2,c3)` fails without gate (c2 re-pooled) | TESTED (red-green) |
| B2 | `IndexFwdNullReader.NullCursor` off-thread skip re-pool | `IndexFwdNullReaderTest.testCursorNotRepooled...` | same identity red-green | TESTED |
| B3 | `IndexBwdNullReader.NullCursor` off-thread skip re-pool | `IndexBwdNullReaderTest.testCursorNotRepooled...` | same | TESTED |
| B4 | `BitmapIndexBwdReader.Cursor`/bitmap NullCursor off-thread | — (identical shared gate; Fwd + null identity tests) | mechanism covered indirectly | Covered (indirect) |
| C1 | Single-key covering DIRECT -> CURSOR_BASED | `testParquetExportCoveringScanShapes` shape 3 (var-size) + `...MultiWorker` single-key (fixed-width) | revert `CoveringIndexRecordCursorFactory` override -> DIRECT -> all-null -> read_parquet vs source diverges | TESTED |
| C2 | Single-key projection PAGE_FRAME_BACKED -> CURSOR_BASED | shape 1 (`open+1`) | revert `ParquetExportMode` PAGE_FRAME_BACKED hunk -> all-null | TESTED |
| C3 | Single-key via view (`StaleViewCheckFactory`) | shape 2 (`deriv_v`) | revert `StaleViewCheckFactory` override -> default delegation steps over covering scan -> all-null | TESTED |
| C4 | Multi-key DIRECT stays (var-size) | shape 4 (`IN (...)`) | non-regression guard (green pre & post) | Guard |

## Summary
- **Verdict: approve.** Zero confirmed findings touch correctness, data integrity, resource safety, concurrency, or performance/IO on a reachable path. The off-thread gate is correct at every posting close path (Cursor/NullCursor x Fwd/Bwd, detached, reader-closed, off-thread, pool-full) with no leak or double-free; the export-mode change is purely restrictive and additionally fixes a pre-existing all-null data-correctness bug; the plain-long stamp is honestly scoped as defense-in-depth and the real safety rests on the (verified) serialized single-owner handoff and row-cursor-before-frame-cursor close ordering.
- **Correctness & performance gate:** passes — no open correctness/resource/concurrency/perf finding. The two verified performance tradeoffs (unconditional `stampOperatingThread()` ~1 ns per cursor checkout, not per row; off-thread re-pool skip reallocating the decode block buffer, bounded by migrating suspend points) are correctness-forced trades replacing a pre-existing cross-thread data race, not regressions against a legitimate baseline.
- **Test gate:** passes — the fix ships with regression tests carrying verified failure links; all three C1 export routes and the M1 gate are independently red-green; no UNTESTED Critical row.
- **Coverage map:** 14 behavioral changes + 1 non-regression guard; 13 TESTED with verified failure links, 1 covered indirectly (B4 Bwd-bitmap/bitmap-NullCursor identity), 0 UNTESTED.
- **Draft findings:** 7 review agents produced ~20 candidate observations; 0 verified as Critical, ~5 candidate concerns verified as false positives / cleared-with-nuance (Downgraded section), the remainder are Moderate coverage nuances and Minor style/maintainability/doc items.
- **In-diff / out-of-diff / pre-existing split:** the C1 covered-column bug is **pre-existing** (from #7284) and is fixed here with its own red-green tests, as required; everything else is in-diff. No out-of-diff callsite is broken by the change (cross-context pass over all `getCursor` callers and retained-cursor holders returned all-SAFE).
- **PR metadata:** Conventional Commits title with repeated verb; end-user-focused, level-headed description listing each bundled fix with its own section and tradeoffs; labels (Bug, Core, Postgres Wire, REST API) match scope. No `Fixes #NNN` — the PR references an internal assertion failure rather than a filed GitHub issue, which is acceptable.

The Minor items (member ordering on `producesMaterializedPageFrames()`, the bitmap/null re-pool helper, the test-helper/comment duplication, the javadoc convention reminder, and the one mislabeled test comment) are worth cleaning up while in here but do not block the approve.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 47 / 47 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexBwdReader.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/IndexBwdNullReader.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/BitmapIndexBwdReader.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/StaleViewCheckFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/AbstractBitmapIndexReader.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/RecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/AbstractPostingIndexReader.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cutlass/parquet/ParquetExportMode.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/IndexFwdNullReader.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexFwdReader.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/BitmapIndexFwdReader.java|4|4|100.00%|

PR #7350, opened 2026-07-03
fix(qwp): prevent stale egress plans from leaking to clients

jerrinot:
## fix(qwp): prevent stale egress plans from leaking to clients

QWP egress only retried once when a cached SELECT plan became stale. A second stale-plan hit during recompile surfaced as a query error to the client, unlike REST and PGWire which use the configured SQL recompile limit.

Retry QWP egress cursor acquisition up to `maxSqlRecompileAttempts` before returning an error, and keep the retry before any result bytes are sent.

## ci: split the mac-griffin test shard to avoid timeouts

The `mac-griffin` hosted job ran the entire `**/griffin/**` tree (~971 test classes, including all fuzz tests) in a single macOS leg. It was the pipeline's long pole and sat right at the 60-minute per-job cap, so macOS agent variance or steady test growth could tip it over. This PR's own run (Azure build 249084) timed out on `mac-griffin` at ~55 minutes with tests still progressing normally - no hung test and no crash, just wall-clock exhaustion.

Split it into three self-contained legs, mirroring the existing `windows-griffin` base/sub/fuzz layout:

- `mac-griffin-base` - `**/griffin/*` minus fuzz (212 classes)
- `mac-griffin-sub` - `**/griffin/**/*` minus base/fuzz (738 classes)
- `mac-griffin-fuzz` - the `griffin/fuzz` package plus every `*Fuzz*` class (21 classes)

The three legs form a disjoint, complete partition of the griffin tree; only griffin is repartitioned, so `mac-cairo`, `mac-cairo-fuzz`, `mac-pgwire`, and `mac-other` are untouched (`mac-other` still excludes `**/griffin/**`).

Tradeoff: this adds two more concurrent macOS agents (mac legs 5 -> 7), and macOS hosted agents are the most constrained pool. Balance is by class count rather than measured runtime, so `mac-griffin-sub` is the leg to watch; the base/sub boundary can be nudged later if it runs hot.

## Test plan

- Verified the three griffin globs form a disjoint, complete partition over all 971 griffin test classes: every class lands in exactly one leg, with no double-runs and no gaps.
- The full Azure macwin pipeline run on this branch is the validation that each new leg completes under the 60-minute cap.


jerrinot:
CI hiccup:
[ERROR] Failed to execute goal org.jacoco:jacoco-maven-plugin:0.8.14:prepare-agent (default-prepare-agent) on project questdb: Execution default-prepare-agent of goal org.jacoco:jacoco-maven-plugin:0.8.14:prepare-agent failed: Plugin org.jacoco:jacoco-maven-plugin:0.8.14 or one of its dependencies could not be resolved: Could not transfer artifact org.apache.maven.doxia:doxia-sink-api:jar:1.0 from/to hetzner-reposilite (http://maven.internal:8081/releases/): transfer failed for http://maven.internal:8081/releases/org/apache/maven/doxia/doxia-sink-api/1.0/doxia-sink-api-1.0.jar: Connect to maven.internal:8081 [maven.internal/10.0.0.12] failed: Connect timed out -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.

jerrinot:
Review — PR #7350 fix(qwp): prevent stale egress plans from leaking to clients (level 2)

Verdict: Approve. The core change is correct and is a near-verbatim port of the HTTP/PGWire bounded-recompile loop. 13 draft leads investigated, 11 confirmed clean (false positives), 2 Minor observations survived. All findings in-diff; the cross-context surface (callers of the changed private method + the
shared static hook) was walked and is clean.

Critical

None.

Moderate

None.

Minor

1. Orphaned section comment after alphabetical re-sort — in-diff
QwpEgressUpgradeProcessor.java:~862. The relocated comment // Egress message dispatch and query execution now sits directly above executeNonSelect, but dispatchEgressMessage was sorted away to ~line 802. The comment reads as a section heading for methods that alphabetization has already scattered — exactly
the "won't stay grouped by category" failure CLAUDE.md warns about. Not a banner-comment violation (it's plain prose), but it's now misleading; consider deleting it or narrowing it to describe only executeNonSelect.

2. DEBUG_FORCE_STALE_PLAN_RECOMPILES is a process-global static with a non-atomic read-modify-write — in-diff (test hook)
QwpEgressUpgradeProcessor.java:1224-1227. The counter is public static volatile, shared across all egress connections and all server instances in the JVM, and the decrement (read local; write local-1) is not atomic. Safe today: surefire runs test classes sequentially per fork and both new tests reset the
field in finally. But the exact-count assertions (assertEquals(0, DEBUG_FORCE_STALE_PLAN_RECOMPILES)) would flake if intra-fork test parallelism is ever enabled, two QWP servers run in one JVM, or a stray concurrent SELECT lands while the counter is armed. This matches the existing
DEBUG_FORCE_INTERNAL_ERROR_ON_RESUME shape, so it's consistent practice, not a regression — worth a one-line comment noting the single-connection assumption.

Downgraded (false positives)

- Resource leak / double-free on retry or exhaustion — Misc.free nulls cursor/pageFrameCursor/factory at 1244-1246; the outer catch re-frees three already-null refs (no-op). SqlCompiler is try-with-resources per iteration. Clean.
- SqlException.$(0, e.getFlyweightMessage()) aliases the exception's mutable StringSink — position(0).put(msg) copies char-by-char into SqlException's own sink; getMessage() then materializes a fresh String. Independent of e's buffer. Safest of the four processor impls.
- Thread-local SqlException corrupted by LOG.error().$(e) before sendQueryError reads it — nothing between the throw and e.getMessage() re-enters SqlException.position(); under -ea position() even allocates fresh. Clean.
- Off-by-one vs maxSqlRecompileAttempts — maxSqlRecompileAttempts + 1 total attempts, identical accounting to JsonQueryProcessor:474/513. Both tests confirm the counts.
- Metrics double-count across retries — markQueryStarted is once before the loop; error/cancel metrics only in the outer catch. No per-retry counting.
- applyCacheReset staging left inconsistent on exhaustion — staged mask is retained and OR-merged by the next query; the method's own comment covers "a SELECT that threw before emitPendingCacheReset ran." Unchanged path count.
- Moved methods changed behavior — git show 98b87526e3 confirms the relocation commit is byte-identical (per-method diff empty, MD5 match); the behavioral change lives entirely in dbd04d659b.
- Member ordering broken by the move — both peer groups (private-static, private-instance) are fully alphabetical after the move; stageReject, applyCacheResetForUpcomingQuery, and emitPendingCacheReset all land in their correct slots.
- Test API / config plumbing — startServerWithRetry, newPlainText, QwpConstants.STATUS_PARSE_ERROR, and CAIRO_SQL_MAX_RECOMPILE_ATTEMPTS (default 10) all exist and wire through as the tests assume.
- Test assertion style — the tests drive a real QWP WebSocket client via callback handlers (protocol-level), so the assertQuery(...) builder rule doesn't apply; assertMemoryLeak wrappers are legitimate multi-statement server+client scopes.
- Commit titles (order!, Conventional-Commits prefix) — CLAUDE.md: PR branch history is throwaway under squash-merge; not flagged.

Summary

- 11 draft findings dropped as false positives, 2 Minor confirmed. All in-diff.
- Regressions/tradeoffs: none. The change strictly widens the retry budget (retry-once → maxSqlRecompileAttempts) and converts the exhaustion error from a raw TableReferenceOutOfDateException (mapped to STATUS_INTERNAL_ERROR) to SqlException (STATUS_PARSE_ERROR) — the intended, better-classified
client-facing behavior, matching REST/PGWire.
- Test coverage is good: one regression test proving 2 consecutive stale plans now succeed (would fail on the old retry-once loop) and one proving exhaustion returns a clean parse error with zero rows streamed.

Nothing blocks merge; the two Minor items are optional cleanups.

bluestreak01 (approved):
Approve (level-3 review).

The QWP fix is a faithful port of the HTTP/PGWire bounded-recompile loop: retry accounting (`retries == maxSqlRecompileAttempts` -> maxSqlRecompileAttempts+1 total attempts) matches JsonQueryProcessor/ExportQueryProcessor exactly, and the exhaustion error is correctly reclassified from a raw TableReferenceOutOfDateException (STATUS_INTERNAL_ERROR) to SqlException (STATUS_PARSE_ERROR). No aliasing/double-free/leak on the retry or exhaustion paths (verified: fresh exception per throw, char-copy into the thread-local SqlException, null no-op frees).

The CI mac-griffin split is a provably disjoint and complete partition of the griffin tree: base=griffin/* minus fuzz, sub=griffin/**/* minus base/fuzz-pkg/fuzz-named, fuzz=griffin/fuzz/** plus **Fuzz**. Union = all griffin classes, pairwise intersections empty, and it's robust to the exact glob semantics of **Fuzz** because sub's exclude and fuzz's include use the identical pattern. mac-other still excludes **/griffin/**, so no cross-leg double-run.

No Critical or Moderate findings. Minor nits (all optional):
- ci/templates/hosted-jobs.yml:90 - the mac-other comment still says 'run in mac-griffin', a job this PR renamed away; the parallel windows-other comment was kept accurate, so worth syncing.
- The relocated // Egress message dispatch and query execution comment now sits above executeNonSelect as a misleading section heading (dispatchEgressMessage was sorted elsewhere).
- DEBUG_FORCE_STALE_PLAN_RECOMPILES exact-count assertions assume single-connection/sequential forks; consistent with the existing DEBUG_FORCE_* hooks.
- The added cursor/pageFrameCursor frees in the inner catch are dead (both always null at the throw site) but harmless.

Good regression coverage: one test proving two consecutive stale plans now succeed (fails on the old retry-once loop) and one proving exhaustion returns a clean parse error with zero rows streamed.

bluestreak01 (changes requested):
@jerrinot

# Review — PR #7350 `fix(qwp): prevent stale egress plans from leaking to clients` (level 3)

**Reviewing at level 3** — full mission-critical pass: change-surface + coverage map, all 13 applicable agents (Agent 8/Rust N-A — no `.rs`), fresh-context adversarial pass, per-finding source verification. Two bundled changes: a QWP egress bug fix and a CI shard split.

## Coverage map

| Change | Test | Failure link | Dims | Verdict |
|---|---|---|---|---|
| `handleQueryRequest`: retry budget widened `retry-once` → `maxSqlRecompileAttempts` (default 10); total attempts `= max+1` | `testRepeatedStalePlanRecompilesDoNotLeakToClient` (DEBUG=2, max=10) | Forces 2 consecutive stale opens then success; asserts row `42` returned + `DEBUG==0`. Under the reverted retry-once loop the 2nd throw (`attempts>=2`) surfaces to the client → `onError`→`Assert.fail`. Verified regression guard. | happy ✓, error ✓, boundary (partial), leak ✓ (`assertMemoryLeak`), concurrency N-A (single conn) | **TESTED** |
| Exhaustion reclassified `TableReferenceOutOfDateException`(INTERNAL) → `SqlException`(PARSE) | `testStalePlanRecompileExhaustionReturnsParseError` (max=2, DEBUG=3) | Asserts `STATUS_PARSE_ERROR` + message contains `"cached query plan cannot be used"` + `rows==0` + `onEnd` not seen + `DEBUG==0`. Fails if the reclassification or the byte-count regresses. | happy N-A, error ✓, NULL N-A, boundary ✓ (exact exhaustion), leak ✓ | **TESTED** |
| Real DDL-race stale recompile (non-debug path) | `testDropRecreateRecompilesStaleFactory` (pre-existing, 5 real DROP+CREATE cycles) | Exercises a genuine `TableReferenceOutOfDateException` from `getCursor`/`getPageFrameCursor`. | happy ✓, error ✓ | **TESTED** |
| CI: mac-griffin → base/sub/fuzz partition | none (CI config) | Validated by the macwin pipeline run, not a unit test. EXEMPT as CI-config; but see the Moderate finding — the partition is valid yet mis-balanced. | N-A | **EXEMPT (CI)** |

Coverage map: 3 behavioral code changes, 3 TESTED, 0 UNTESTED. The exhaustion-boundary edge is bracketed but not pinned exactly (`DEBUG==max` last-chance success is untested); low-severity, noted below.

## Critical

None. The QWP code change is correct on every path I traced:
- Retry accounting `for (int retries=0;;retries++)` with `if (retries == maxSqlRecompileAttempts) throw SqlException.$(0, e.getFlyweightMessage())` is byte-for-byte the `JsonQueryProcessor.java:512-514` reference (`max+1` total attempts). No off-by-one.
- Exhaustion reclassification INTERNAL→PARSE verified through `mapErrorStatus` (`SqlException`→`STATUS_PARSE_ERROR`; raw `TableReferenceOutOfDateException`, a bare `RuntimeException`, would fall through to `STATUS_INTERNAL_ERROR`).
- Thread-local `SqlException` aliasing is safe: `SqlException.$` char-copies from the exception's own `StringSink` into the `CarrierLocal` sink, and nothing between the throw and `e.getMessage()` re-enters `SqlException.position()`; the message is materialized to a `String` as a method argument before `sendQueryError` runs.
- Resource lifecycle (cache poll removes ownership, per-iteration `SqlCompiler` try-with-resources, `factory`/`cursor`/`pageFrameCursor` freed-and-nulled each retry, no double-free on exhaustion, no poisoned cache) is clean.
- The three relocated methods (`applyCacheResetForUpcomingQuery`, `emitPendingCacheReset`, `stageReject`) are **byte-identical** to base (verified by brace-matched extraction) — pure alphabetical reordering.
- Egress-only: no ingress/producer/store-and-forward contract is touched, so the QWP store-and-forward invariants do not apply here.

## Moderate

**1. CI fuzz-shard split does not isolate the fuzz long-pole it targets — `**/griffin/**Fuzz**` misses all subpackage fuzz classes.** `ci/templates/hosted-jobs.yml:18,33,46` — in-diff. **Verified empirically** by running the actual surefire matcher (`TestListResolver`, surefire-api 3.5.3) against representative class paths:

```
class                                          base   sub    fuzz
griffin/CreateTableFuzzTest (base pkg)         false  false  true    <- correct
griffin/fuzz/... (fuzz pkg)                     false  false  true    <- correct
griffin/engine/join/AsOfJoinFuzzTest (subpkg)  false  TRUE   false   <- WRONG leg
griffin/unnest/UnnestFuzzTest (subpkg)         false  TRUE   false   <- WRONG leg
griffin/engine/window/NamedWindowFuzzTest      false  TRUE   false   <- WRONG leg
```

`**Fuzz**` is not the special Ant `**` segment token; after `/griffin/` it collapses to a single-segment matcher that requires the `Fuzz` file to be an *immediate child of griffin*. So the fuzz leg captures only the 6 base-package fuzz classes plus the `griffin/fuzz` package — the **13 subpackage `*Fuzz*` classes run on `mac-griffin-sub`**:

```
griffin/engine/join/AsOfJoinFuzzTest (6 @Test, 515 lines)
griffin/unnest/UnnestFuzzTest (6 @Test)   griffin/unnest/JsonUnnestFuzzTest (4 @Test)
griffin/engine/window/NamedWindowFuzzTest (3 @Test)   griffin/engine/orderby/OrderByNegativeLimitFuzzTest
griffin/engine/groupby/{GroupByCharSequenceLongHashMap,GroupByIntHashSet,GroupByLongHashSet,GroupByLongLongHashMap,GroupByUtf8SequenceLongHashMap}FuzzTest
griffin/engine/functions/array/{BuildArray,DoubleArraySort}FuzzTest   griffin/engine/functions/groupby/ArrayAggDoubleFuzzTest
```

Consequences: (a) the partition is **still valid** — every griffin class runs in exactly one leg (`#legs==1` for all; no double-run, no gap), because sub's exclude uses the identical `**/griffin/**Fuzz**` as fuzz's include, so this is not a coverage bug; but (b) the split **fails its stated purpose** — several genuine long-running query fuzzers (AsOfJoin, unnest, window, order-by) stay on `mac-griffin-sub`, which the PR body itself calls "the leg to watch," so the very timeout the change fixes can recur there; (c) the `mac-griffin-fuzz` comment (`:37` "every `*Fuzz*` class under griffin") and the "21 classes" balance claim are factually wrong; (d) it diverges from the Windows shards, which use depth-agnostic `**/**Fuzz**`.

*Justification for Moderate (not Critical):* zero dimension on database correctness, data integrity, resource safety, or data-path performance/IO — every test still runs exactly once; the only effect is CI leg balance + a wrong comment.

*Fix (verified with the real matcher to keep the partition valid AND route all fuzz to the fuzz leg): replace `**/griffin/**Fuzz**` with `**/griffin/**/*Fuzz*` in all three places:*
- base exclude (`:18`): `**/griffin/**/*Fuzz*`
- sub exclude (`:33`): `**/griffin/*,**/griffin/fuzz/**,**/griffin/**/*Fuzz*`
- fuzz include (`:46`): `**/griffin/fuzz/**,**/griffin/**/*Fuzz*`

The macwin run then validates the rebalanced legs.

## Minor

1. **Stale comment** `ci/templates/hosted-jobs.yml:90` — `# Exclude griffin: run in mac-griffin`; the `mac-griffin` job no longer exists (now base/sub/fuzz). The parallel `windows-other-1:194` (`run in windows-griffin-base and windows-griffin-sub`) is the correct form. Sync it. (The exclude `**/griffin/**` itself is correct — verified mac-other still excludes all griffin.)
2. **Misleading section-heading comment** `QwpEgressUpgradeProcessor.java` — the relocated `// Egress message dispatch and query execution` now sits above `executeNonSelect` in an alphabetically-sorted file, describing a region alphabetization has scattered (the actual `dispatchEgressMessage` is elsewhere). Not a `// ===` banner, so not a hard CLAUDE.md violation, but misleading — delete it.
3. **Dead defensive frees** `QwpEgressUpgradeProcessor.java:1244-1245` — `cursor = Misc.free(cursor); pageFrameCursor = Misc.free(pageFrameCursor);` in the retry catch are provably always null at the throw site (Java assigns the LHS only after `getCursor`/`getPageFrameCursor` returns; the DEBUG hook throws before either). Harmless no-ops; optionally trim, or narrow the "partially-acquired resources" comment which overstates the case.
4. **Test hook non-atomicity** `QwpEgressUpgradeProcessor.java:1224-1226` — `DEBUG_FORCE_STALE_PLAN_RECOMPILES` is a JVM-global `volatile` with a non-atomic read-modify-write; the exact-count `assertEquals(0, …)` assertions assume single-connection/sequential forks. Consistent with the existing `DEBUG_FORCE_INTERNAL_ERROR_ON_RESUME` idiom and reset in `finally`; production leaves it 0 (one volatile read, no write). Worth a one-line comment noting the single-connection assumption.

## Pre-existing (out-of-diff, non-blocking, low confidence)

**Narrow double-free window in the handoff target** `QwpEgressProcessorState.beginStreaming:390-401` / `beginStreamingPageFrame:440-454` — both assign `streamingFactory`/`streamingCursor` **before** `streamingSqlText = Chars.toString(sqlText)` and `streamingActive = true`. If `Chars.toString` fails (OOM) in that window, `state` aliases the local `factory`/`cursor` while `streamingActive` is still false, so `handleQueryRequest`'s outer `else if (!streamingHandedOff)` branch frees them locally, leaving `state` holding freed refs that a later `clear()`/`close()` can free again. **Not introduced or touched by this PR** (it does not modify `beginStreaming*`), OOM-gated in a sub-microsecond window, and a true fault only if factory/cursor `close()` is non-idempotent (unverified). Cheap defensive fix if bundled: compute `String sql = sqlText != null ? Chars.toString(sqlText) : null;` first, then assign the `streaming*` fields, with `streamingActive = true` last after all allocation. Reported for completeness; does not block.

- **SqlException thread-local corruption between throw and read** — dismissed: the copy is between two distinct `StringSink`s; nothing re-enters `SqlException.position()` before `e.getMessage()` materializes a `String` argument.
- **Double-free / leak on retry or exhaustion in the changed loop** — dismissed: `Misc.free` nulls each ref; on exhaustion all three are null when the outer catch runs; `SqlCompiler` is per-iteration try-with-resources. Cache is not poisoned (`poll` removes; success re-caches fresh; exhaustion leaves no entry).
- **Fresh `SqlCompiler` per retry is a perf regression vs HTTP** — dismissed: `getSqlCompiler()` is a pooled acquire on the exceptional DDL-race path only; cache-hit path acquires none. Not a data-path cost.
- **CI partition has a gap/double-run** — dismissed: the real matcher shows `#legs==1` for every class; it is a valid partition (the defect is mis-routing, filed as Moderate above).
- **Moved `testEmptyAndNullArrayColumns` changed behavior** — dismissed: pure relocation, byte-identical body.

## Summary

- **Verdict: request changes.** The QWP bug fix itself is correct, faithfully ports the HTTP/PGWire bounded-recompile loop, correctly reclassifies the exhaustion error, is leak-free on every path, and ships strong regression coverage (verified across 7 fresh-context agents plus my own source trace and empirical checks). It is merge-ready on its own. What blocks approval is one confirmed **Moderate** finding on the CI half: the `mac-griffin-fuzz` glob misses all 13 subpackage fuzz classes, so the split does not isolate the fuzz long-pole it is designed to move — leaving heavy fuzzers on the "leg to watch" and shipping a factually wrong comment. The fix is a one-token glob change in three lines, empirically verified to preserve a valid partition while routing all fuzz to the fuzz leg.
- **Correctness & performance gate:** no open finding touches database correctness, data integrity, resource safety, concurrency, or data-path performance/IO. The Moderate finding is CI-config (leg balance + comment accuracy), so it does not trip the database gate — but per the severity rubric an open, non-cosmetic (Moderate) item cannot remain under an "approve", which is why the verdict is request changes rather than approve.
- **Test gate:** passes — every behavioral code change is tested with a verified failure link; the fix carries regression tests that fail against the reverted hunk.
- Findings: 1 Moderate, 4 Minor, 1 pre-existing (non-blocking). ~13 draft leads verified, 5 dropped as false positives.
- In-diff / out-of-diff / pre-existing split: 5 in-diff (1 Moderate + 4 Minor), 1 pre-existing (out-of-diff, non-blocking).
- Regressions/tradeoffs: the QWP change strictly widens the retry budget and improves error classification (no regression). The CI change adds 2 macOS agents (mac legs 5→7) on the most constrained pool and, as written, does not fully deliver its timeout-avoidance goal until the glob is fixed.

The one action to clear approval: fix the fuzz glob (`**/griffin/**Fuzz**` → `**/griffin/**/*Fuzz*` in the base exclude, sub exclude, and fuzz include) and refresh the two stale comments (`mac-other:90`, the relocated Java section comment). The two prior approvals verified the partition's validity but did not check where the fuzz classes actually land — which is exactly the gap the level-3 adversarial pass caught.

bluestreak01 (approved):
# Review — PR #7350 `fix(qwp): prevent stale egress plans from leaking to clients` (level 3)

Full mission-critical pass: change-surface map, 5 fresh-context review agents (correctness, resource+concurrency, performance/adversarial, tests/efficacy/regression, fresh-context adversarial), and per-finding verification against source. The PR bundles two changes: a QWP egress stale-plan fix (Java) and a `mac-griffin` CI shard split (YAML), plus a pure alphabetical member-reorder (`order!` commit, verified 95/95 pure move).

## Verdict: Approve

The core change is a faithful, verified port of the HTTP `JsonQueryProcessor` bounded-recompile loop. Both new behaviors ship with genuine regression tests that fail against the old code. The CI partition is empirically proven disjoint and complete. No correctness, resource, concurrency, or performance/IO finding stands.

## Critical

None.

## Moderate

**M1 — PR body's fuzz-leg class count (21) is stale and materially understates the leg.** `ci/templates/hosted-jobs.yml` fuzz glob `**/griffin/fuzz/**,**/griffin/**/*Fuzz*` captures the `griffin/fuzz` package (~50 source files) **plus 19 `*Fuzz*` classes living in griffin subpackages** (join, unnest, window, order-by, ...). Empirically the three legs split as base≈220 / sub≈741 / fuzz≈69, not the body's 212/738/21. The `21` predates the later `4e3d872830` commit that widened the fuzz glob to be depth-agnostic — the body's own YAML comment now contradicts its prose. This does not affect partition correctness (proven disjoint+complete below), but it undercuts the tradeoff analysis: the body names `mac-griffin-sub` as "the leg to watch", yet the 19 heaviest subpackage fuzzers were moved *out* of sub and *into* fuzz, so `mac-griffin-fuzz` is at least as likely to be the hot leg. Behaviorally and performance-wise inert with respect to the shipped code (globs are correct); it is a description-accuracy issue. Suggested fix: update the body's per-leg counts to match the shipped globs (~220 / ~741 / ~69) and revise "leg to watch" accordingly.

## Minor

- **m1** — Comment at `QwpEgressUpgradeProcessor.java:~1184` says the loop matches "HTTP/PGWire's bounded `maxSqlRecompileAttempts` behavior." It matches HTTP exactly (`for (retries=0;;retries++)`, `retries == max` ⇒ `max+1` attempts). PGWire uses `for (attempt=1;;attempt++)` with `attempt == maxRecompileAttempts` ⇒ `max` attempts, i.e. one fewer. Cosmetic wording only; the code correctly tracks the closest analog (HTTP).
- **m2** — `mac-other` comment still reads "Exclude griffin: run in mac-griffin" (singular); that leg no longer exists. Stale comment, glob is correct.
- **m3** — New `@TestOnly` was added to `DEBUG_FORCE_STALE_PLAN_RECOMPILES` but the sibling `DEBUG_FORCE_INTERNAL_ERROR_ON_RESUME` lacks it. Annotation inconsistency, no functional impact. (`@TestOnly` is already used in production source, so the import compiles cleanly.)

## Downgraded (false positives / non-issues)

- **"The added `cursor = Misc.free(cursor)` / `pageFrameCursor = Misc.free(pageFrameCursor)` in the catch fix a leak"** — Dismissed. `TableReferenceOutOfDateException` (and the debug throw) always fires *during* acquisition, before the assignment target receives a value, and a successful `getPageFrameCursor`/`getCursor` is immediately followed by `break`. Both locals are provably `null` at every reachable point the catch runs. The frees are inert defensive no-ops, not a leak fix — but harmless and reasonable future-proofing.
- **"Non-atomic RMW on `DEBUG_FORCE_STALE_PLAN_RECOMPILES` is a concurrency bug"** — Dismissed. Production leaves it at 0, so the write branch is never entered (one plain volatile read + not-taken branch). Tests set/reset it in `try/finally`, single connection, single query — no reachable contamination. Mirrors the accepted `DEBUG_FORCE_*` precedent.
- **"SqlException thrown in the loop could be re-caught by the inner `catch (TableReferenceOutOfDateException)`"** — Dismissed. `SqlException extends Exception`, `TableReferenceOutOfDateException extends RuntimeException`; unrelated hierarchies. It propagates to the outer `catch (Throwable)`, which legally handles the checked exception and maps it to `STATUS_PARSE_ERROR`.
- **"Metrics double-count across retries"** — Dismissed. `markQueryStarted()` runs once before the loop; `markQueryErrored()` runs once in the outer catch. Balanced regardless of retry count.
- **"Extra per-retry `SqlCompiler` checkout is a regression"** — Dismissed. The `try (SqlCompiler …)` inside the loop pre-dates this diff; the cost is a bounded pool op only on rare real stale-plan retries, compile-time, not a data path.

## Coverage map

| Change | Test | Failure link | Verdict |
|---|---|---|---|
| Bounded loop, success after N recompiles (`handleQueryRequest`) | `testRepeatedStalePlanRecompilesDoNotLeakToClient` (DEBUG=2, max=10) | Old retry-once throws raw exception on 2nd stale → `onError` → helper's `rows==1`/`sum==42` assertion fails; `counter==0` proves exactly 2 forced throws consumed | TESTED |
| Bounded loop, exhaustion → `SqlException` → `STATUS_PARSE_ERROR` | `testStalePlanRecompileExhaustionReturnsParseError` (DEBUG=3, max=2) | Old loop → raw exception → `STATUS_INTERNAL_ERROR`; test asserts `STATUS_PARSE_ERROR` + message-contains + `rows==0` + `!endSeen` (all fail against old behavior) | TESTED |
| Error remap: raw `TableReferenceOutOfDateException` (INTERNAL) → `SqlException` (PARSE) | same exhaustion test | `mapErrorStatus` arm verified: `SqlException`→PARSE, raw→INTERNAL | TESTED |
| `DEBUG_FORCE_STALE_PLAN_RECOMPILES` hook | both tests exercise + reset in `finally` | counter arithmetic asserted `== 0` | TESTED (test infra) |
| Real stale *cached* factory, ≥2 real-DDL retries | `testDropRecreateRecompilesStaleFactory` (single real retry) | Deterministic ≥2 real-DDL race is inherently flaky; debug hook is a fair proxy | Covered (single-retry real; ≥2 by proxy) — acceptable |
| Added `cursor`/`pageFrameCursor` frees in catch | — | Inert no-op on every reachable path (locals always null) | EXEMPT (verified no-behavioral-change) |
| CI `mac-griffin` → base/sub/fuzz split | Empirical partition proof (1030 classes → 220/741/69, 0 gaps, 0 double-runs) + Azure pipeline run | Config-only, no production behavior | EXEMPT |
| `order!` member reorder + `testEmptyAndNullArrayColumns` move | — | Pure move, verified byte-identical (95/95) | EXEMPT |

## Summary

- **Verdict: approve.** Correctness & performance gate: passed — zero open findings touch correctness, data integrity, resource safety, concurrency, or performance/IO. Test gate: passed — both new behaviors have genuine regression tests with verified failure links against the old code; the fix is not shipping untested.
- **Coverage map: 8 rows — 4 TESTED, 1 covered-by-proxy (acceptable), 3 EXEMPT (verified no-behavioral-change). 0 UNTESTED.**
- **CI partition independently proven:** all 1030 griffin test classes map to exactly one leg (base 220 / sub 741 / fuzz 69), 0 gaps and 0 double-runs, and the 6 `*Fuzz*` classes sitting directly in the `griffin` base package correctly route to the fuzz leg via the depth-agnostic `**/griffin/**/*Fuzz*` glob. `mac-other` still excludes `**/griffin/**`.
- **Findings:** 1 Moderate (PR-body count staleness), 3 Minor (cosmetic comment/annotation). All in-diff; cross-context surface (private-method callsites, the shared static hook, the CI leg-name rename) was walked and is clean — no out-of-diff or pre-existing bugs surfaced.
- **Verification:** ~13 draft leads investigated across 5 agents; 5 dismissed as false positives after source verification, the rest confirmed clean. No blocking item survived.
- The only open items are the Moderate description-accuracy fix and three cosmetic nits — none block merge under the gate, though updating the PR body's leg counts (M1) before merge is worth doing since the release-notes/reviewer-facing description currently understates the fuzz leg.

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressUpgradeProcessor.java|46|46|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=249494)
**Tested Commit:** `f57ab9d43393aee7376be93b1f8fc047ae0cf852`

PR #7353, opened 2026-07-04
fix(sql): fix PIVOT and expression column names that embed double quotes

### LINKED ISSUE #6471: Inconsistent Double Quotes Naming Behavior on Calculated Columns
### To reproduce

Test on [QdbDemo](https://demo.questdb.io/index.html)
Here output column naming convention is exactly identical to input column text `indicator2 / 2` on Query:  `SELECT indicator2 / 2 from core_price limit 0`

Here output column naming convention is not identical to input column text `"indicator2 / 2.314"` on Query: `SELECT indicator2/0.1289543 from core_price limit 0`

Another weird output bug `SELECT indicator2 / 2.314 as 'indicator2 / 2.314' from core_price limit 0`

Clearly something not good is happening in the output naming code handling insidq QDB... I would highly appreciate it if you would remove the double quotes. Thanks!

### QuestDB version:

9.2.1

Cloud Demo

Cloud Demo

### Full Name:

Super Richmann

### Affiliation:

None

- [x] Yes, I have

### Additional context

This situation renders my current use case impossible.

There is no possible way I could find to use calculated columns with QuestDB using .NET NpgSql 9.0.2 on ML.NET AutoML Experiment generation,

ML.NET AutoML CreateDatabaseLoader function which receives the SQL to query and the column names MUST have  the column names inside the SQL query 100% exactly identical to the output column names.

Adding the double quotes to a calculation makes it impossible to use ML.net CreateDatabaseLoader and successfully running an AutoML Experiment on it.

glasstiger:
Fixes #6471

### Problem

The SQL compiler wraps column aliases in double quotes when it cannot process the bare name: names containing dots (PIVOT IN values such as `FNCL 2.5 7/26`, auto-generated expression aliases such as `t.x+1`, user aliases such as `"a.b"`) and names colliding with operator tokens (`in`, `and`). The quotes protect the alias from dot-splitting and operator collisions inside the optimiser, but they leaked verbatim into result set metadata, so the web console, PG wire clients and CSV exports displayed column names with literal double quotes around them:

```sql
SELECT * FROM trades
PIVOT (sum(amount) FOR security IN (SELECT DISTINCT security FROM trades) GROUP BY sub);
-- columns: sub, "FNCL 2.5 7/26", "FNCL 3.0 8/26"   (quotes are part of the name)
```

### Change

The protective quotes stay compiler-internal and no longer reach result set metadata:

- `SqlUtil.isQuoteProtectedAlias` recognises compiler-added quotes by re-testing the quoting condition (unquoted dot or operator token inside the quotes). Aliases whose quotes are genuine content, e.g. the string alias `'"f,g"'`, keep them. A ranged overload tests slices of composed `alias.column` references.
- `SqlUtil.toColumnName` strips protected quotes wherever a projection alias becomes a metadata column name (`SqlCodeGenerator`, `GroupByUtils`, `RecordAsAFieldRecordCursorFactory`).
- `SqlUtil.getColumnIndexQuiet(RecordMetadata, CharSequence)` resolves compiler-side column references: it looks the token up verbatim first and, on a miss, retries with the protective quotes stripped. All compile-layer resolution sites (`SqlCodeGenerator`, `FunctionParser.createColumn`, `GroupByUtils`) call this helper, and the join-key lookups (`lookupColumnIndexes`, `collectColumnIndexes`) apply the same retry to the column slice of composed references such as `m."k.b"`.
- The `RecordMetadata.getColumnIndexQuiet(CharSequence)` interface default stays strictly verbatim. Ingestion and storage paths (ILP, QWP, CSV import, WAL) resolve wire-supplied names through it, and a name such as `"in"` (quote bytes included) must keep failing as an invalid column instead of being redirected into an existing column named `in`.
- Wildcard deduplication (`SqlUtil.createColumnAlias`) inserts its sequence suffix inside the closing quote of a quote-protected alias (`"a.b"` -> `"a.b1"`) rather than after it (`"a.b"1`), so a duplicated dotted alias stays protected and strips to a clean name.
- Dotted UNNEST column aliases (`UNNEST(t.arr) AS u("a.b")`) keep their quotes through the parser, matching the SELECT-alias convention, so their dots stay content instead of being split into a `table.column` reference. The column surfaces with a clean name and resolves through the qualified form `u."a.b"`.
- `JoinRecordMetadata` keeps its dot-splitting map convention (an unquoted dot in a stored name means a `table.column` prefix): both `add` overloads share one helper that restores the protective quotes locally for dotted plain columns of aliased sides. The verbatim-first lookup order means it needs no override; join internals behave as before and join outputs surface clean names through the projection on top.

Result set column names change as follows:

- PIVOT on dotted values: `FNCL 2.5 7/26` instead of `"FNCL 2.5 7/26"`
- PIVOT on operator-token values: `in` instead of `"in"`
- expression auto-aliases: `t.x+1` instead of `"t.x+1"`
- user aliases: `SELECT 1 AS "a.b"` returns column `a.b` instead of `"a.b"`, matching PostgreSQL
- duplicated dotted aliases: the second column is `a.b1` instead of `"a.b"1`
- joins over dotted-named projections: `foo.bar` instead of `"foo.bar"`
- dotted UNNEST aliases: `a.b` instead of a crash or a mis-split `b`

### Follow-up fixes on this branch

Review of the branch surfaced further issues the change above missed or introduced - a stale-quote leak, a leaked trailing space, three regressions (a compiler hang, a join-compile failure, and NULL data from a dotted JSON UNNEST key), two efficiency fixes, and a defensive hardening. Each lands as a separate commit on top of the change:

- **Stale protective quotes after alias mutation.** `SqlUtil.createExprColumnAlias` decided to quote an alias up front but then mutated it, and `isQuoteProtectedAlias`/`toColumnName` re-derive "is this protected?" from the final content, so after the mutation they no longer recognised the quotes and left them in the name. Two triggers: truncation to `column.alias.generated.max.size` (default 64) dropped the discriminating dot, and a dedup suffix turned an operator token into a plain identifier (`"in"` -> `"in_2"`). The dedup case is reachable through a PIVOT because the duplicate-value guard is case-sensitive while the alias map is case-insensitive, so `IN ('in', 'IN')` collides during aliasing. Both leaked quotes into result set names and broke `CREATE TABLE AS SELECT` over the affected query - the exact failure this PR set out to remove, still open for these cases. `createExprColumnAlias` now emits the wrapping quotes per candidate only when the final content still needs them, and checks uniqueness on that final form, so a truncated or deduped alias that no longer needs protection surfaces bare.
- **Window ORDER BY resolution (defensive).** `RankFunctionFactory`'s streaming fast path resolved the window ORDER BY token with a bare metadata lookup, unlike its own `toOrderIndices` branch and every peer window factory, which unquote a compiler-protected alias through `SqlUtil.getColumnIndexQuiet`. The path is not reachable today (it fires only on a dismissable timestamp order, which arrives clean), but a miss would have added `-1 + 1 = 0` and corrupted the sort key, so it now resolves through `SqlUtil` and throws `invalidColumn` on a miss, matching the sibling paths.
- **Empty PIVOT value hung the compiler.** `SqlUtil.createExprColumnAlias` spun forever on an empty base: an empty PIVOT value (`IN ('')`) unquotes to `""`, which pins the content length at zero, so the loop's return gate (re-keyed on content length by the stale-quote fix above) never fired and the loop grew the `CharacterStore` until it threw `NegativeArraySizeException`. The prior code gated on a length that still counted the quote budget, so it always terminated. The loop now emits a `column` placeholder when no content survives, so it always terminates and produces a valid, unquoted name.
- **Operator-token pivot columns broke through a join.** A `SELECT *` over a join builds the composed reference `t1."in"` from the model alias, but the join metadata stores an operator-token column bare as `t1.in` (only dotted names are re-quoted), so `generateSelectChoose` resolved `-1` and asserted at compile time - a query that compiled before this PR (with the ugly quoted name). `JoinRecordMetadata.getColumnIndexQuiet` now retries the qualified lookup with the protective quotes stripped off the column part, and honours the `lo`/`hi` bounds it had previously ignored in its dotted branch. The wildcard dedup (`createColumnAlias`) this unblocks now decides the quotes per candidate too, so a duplicated operator-token column surfaces as `in1` rather than a leaked `"in1"`. Referencing such a column explicitly (`SELECT t1."in" FROM ...`) still fails in the parser, unchanged (see limitations).
- **Store reuse in the alias dedup loops.** `createExprColumnAlias` and `createColumnAlias` allocated a fresh `CharacterStore` entry for every dedup candidate, so rejected candidates accumulated in the store (O(K) growth for K collisions, plus a doubled high-water baseline). Both loops now hoist one candidate entry and rewind it with `trimTo()` each iteration, keeping store use to a constant two regions regardless of collision count.
- **Single dot-scan at join-metadata build.** `JoinRecordMetadata.add` scanned each joined column name for its dot twice - once in `protectDottedColumnName` to decide re-quoting, once in `addAlias` to split `table.column`. Both `add` overloads now scan once with `Chars.indexOfLastUnquoted` and thread the result through both helpers, dropping the redundant pass per column.
- **Dotted JSON UNNEST keys returned NULL.** The parser keeps the protective quotes on a dotted `COLUMNS("a.b" ...)` field name so its dots stay content for the display name, but `JsonUnnestSource` wrote that name verbatim into the native column descriptor, which the C++ extractor matches against JSON object keys. It looked up a field literally spelled `"a.b"` and never matched the JSON key `a.b`, so every value came back NULL - a silent data regression from the UNNEST-alias change above. `JsonUnnestSource` now strips the protective quotes through `SqlUtil.toColumnName` when building the native extraction key (and in the overflow error message); the displayed column name is unaffected, since it comes from the outer projection metadata that already strips the quotes.
- **Trailing space leaked into truncated aliases.** `createExprColumnAlias` gated its trailing-space trim on `!quote`, but the per-candidate `emitQuote` decision turns false when truncation to `column.alias.generated.max.size` drops the discriminating dot from a quote-protected base. The old code always emitted the quotes for a quote-base, hiding the space inside them, so once they were dropped the bare, space-terminated content leaked into the column name and an all-space slice surfaced as a lone space. The loop now decides `emitQuote` before the trim and keys the trim on `!emitQuote`, so bare content is always trimmed and an all-space slice falls through to the `column` placeholder.

### Trade-offs and limitations

- This is a breaking change for clients that match the old quoted names; any consumer keying on `"t.x+1"`-style names must adjust.
- Referencing a dotted output column from an enclosing query (`SELECT "a.b" FROM (...)`, `ORDER BY "a.b"`) fails in the parser exactly as before; this PR does not attempt to fix referenceability, only the displayed names. Consumers still need `SELECT *`, a table-qualified reference (`t."a.b"`), or positional access for such columns.
- A pivot value whose data genuinely starts and ends with a double quote and contains a dot or matches an operator token (e.g. an alias literally named `"in"`) loses those quotes in the column name (the protected-alias test cannot distinguish them from compiler-added quotes). The previous behaviour rendered such names with both sets of quotes. This is now pinned by a test rather than left to drift; preserving the quotes would require tracking alias provenance out-of-band.

### Test plan

- New behaviour covered by updated expectations in `PivotTest` (dotted DOUBLE/FLOAT/DATE/TIMESTAMP/IPv4 pivot values), `ColumnAliasExpressionTest` (dotted auto-aliases) and `JoinTest#testJoiningSubqueryWithDotInColumnName` (dotted name through a join)
- New regression tests pin the lookup design and the alias handling, and fail without the fix: `HorizonJoinTest#testHorizonJoinKeyedOnQuoteProtectedAlias` (keyed HORIZON JOIN on a quote-protected dotted alias); `SqlUtilTest#testGetColumnIndexQuietInterfaceDefaultIsVerbatim` (the interface default must not redirect `"in"` into a column named `in`); `JoinTest#testJoiningSubqueriesWithDuplicateDottedColumnNames` (dedup keeps the second name clean); `UnnestTest#testDottedColumnAlias` (dotted UNNEST alias via `SELECT *` and a qualified reference); `PivotTest` static dotted / operator-token values and CREATE TABLE AS SELECT over an operator-token pivot; plus truth tables for `SqlUtil.getColumnIndexQuiet` and the ranged `isQuoteProtectedAlias`
- Follow-up regression tests for the stale-quote leak, each failing without its fix: `SqlUtilTest#testExprColumnAliasDedupOperatorTokenDropsStaleQuotes` and `#testExprColumnAliasTruncationDropsStaleQuotes` (unit level, both triggers); `PivotTest#testPivotStaticListWithCaseVariantOperatorTokenValues` (`in` / `IN` yield clean `in` / `IN_2` and a valid `CREATE TABLE AS SELECT`); `PivotTest#testPivotStaticListWithQuotedContentValue` pins the documented quoted-content-value limitation
- Regression tests for the later follow-up fixes, each failing without its fix: `PivotTest#testPivotEmptyValue` and `SqlUtilTest#testExprColumnAliasEmptyBase` (the empty-value hang, now a `column` placeholder); `PivotTest#testPivotOperatorTokenValueThroughJoin` (an operator-token pivot column through a join wildcard); `PivotTest#testPivotDuplicateOperatorTokenValuesThroughJoin` (two such columns dedup to `in` / `in1`); `SqlUtilTest#testExprColumnAliasDedupManyCollisions` (the dedup loop numbers correctly across many collisions under the store reuse)
- `IODispatcherTest#testJsonQueryQuotedColumnNames` passes unchanged, covering aliases whose quotes are content
- Lookup and alias sweep over the affected paths, all green: `SqlParserTest`, `JoinTest`, `AsOfJoinTest`, `HashJoinTest`, `JoinPrefixedTablesTest`, `JoinRecordMetadataTest`, `GroupByTest`, `SampleByTest`, `LatestByTest`, `OrderByAdviceTest`, `OrderByExpressionTest`, `WindowJoinTest`, `UnnestTest`, `UnnestFuzzTest`, `JsonUnnestFuzzTest`, `HorizonJoinTest`, `PGPivotTest`
- Window sweep after the `RankFunctionFactory` change, all green: `WindowFunctionTest` (635), `WindowFunctionUnitTest`, `PercentRankFunctionTest`
- Full `griffin` + JSON query sweep: 23,196 tests, 0 failures, 0 errors
- After the later follow-up fixes, the affected suites re-run green (~4,000 tests, 0 failures): `SqlUtilTest`, `PivotTest`, `JoinTest`, `HorizonJoinTest`, `AsOfJoinTest`, `HashJoinTest`, `WindowJoinTest`, `LateralJoinTest`, `GroupByTest`, `SampleByTest`, `LatestByTest`, `OrderByAdviceTest`, `SqlParserTest`, `WindowFunctionTest`, `UnnestTest`, `ColumnAliasExpressionTest`, `JoinRecordMetadataTest`, `PGPivotTest`
- Verified end-to-end against a live server over HTTP `/exec`: pivot with dotted/space/keyword values, nested and double-nested `SELECT *`, joins over pivot output, `EXPLAIN`, and the content-quotes alias case
- Later review round - the two fixes above plus coverage the earlier rounds lacked, each new test failing without its fix: `JsonUnnestTest#testColumnsDottedFieldNameExtractsValue` (a dotted JSON key extracts its value instead of returning NULL); `SqlUtilTest#testExprColumnAliasAllSpaceBaseYieldsColumn` and `#testExprColumnAliasTruncationDropsTrailingSpace` (no leaked trailing space); end-to-end column-name assertions across the wire and export layers the leak surfaced in - `PGPivotTest#testPivotOperatorTokenColumnNames` / `#testPivotDottedColumnNames` (PG `RowDescription`), `IODispatcherTest#testJsonQueryPivotProtectedColumnNames` (HTTP `/exec` JSON), `ExpParquetExportTest#testExpCsvExportPivotProtectedColumnNames` (CSV `/exp` header); direct `JoinRecordMetadataTest` unit tests `#testAliasedDottedColumnResolvesViaQualifiedRef`, `#testComposedOperatorTokenReferenceResolvesViaStrip`, `#testRangedLookupHonorsBounds`; `SqlUtilTest#testGetColumnIndexThrowsOnMiss`; and `@Test(timeout = 30000)` on `SqlUtilTest#testExprColumnAliasEmptyBase` and `PivotTest#testPivotEmptyValue`, so a revert of the empty-base hang fix fails in 30s rather than hanging a fork


glasstiger:
## Code review

Blocking-style review of the quoted-column-name change, run at high depth: the full diff plus every touched production file was read, the callsite blast radius was inventoried across the whole engine and the enterprise repo, the alias functions were fuzzed (~40k iterations), and the strip-retry was instrumented to confirm which paths it actually reaches. Every draft finding was verified against source before landing here; false positives are listed at the end so the reasoning can be checked.

### Critical
**None.** No correctness bug, resource leak, concurrency hazard, data-redirect, or quote leak. The core invariant holds everywhere: **compile-path resolution uses the strip-retry (`SqlUtil.getColumnIndexQuiet`); every ingestion/storage/wire path stays verbatim** — grep-confirmed that no `SqlUtil.*` resolver appears under `cutlass`/`cairo`/`jit`, and the enterprise repo calls none of them and defines no custom `RecordMetadata`.

### Moderate
**None.**

### Minor

1. **`SqlUtil.java:635,639-642` — empty-content placeholder can exceed `maxLength`.** At the minimum configurable `column.alias.generated.max.size=4`, ~100 colliding empty/whitespace aliases reach `sequence=100`; the `Math.max(1, …)` placeholder floor plus the uncapped `_<seq>` suffix produce `c_100` (5 chars). This is a **deliberate, documented tradeoff** (the comment prefers a 1-char placeholder over a bare `_100`), and the result is a valid, unique name ~1 char over a *soft* cap — no crash/corruption. Either accept it and soften the comment's "within the configured cap" wording, or fold the suffix width into the truncation. Lowest priority.

2. **`RecordMetadata.java:236` — `splitsOnDot()` breaks the boolean-naming convention.** Its interface neighbors are `isColumnIndexed()`, `isSymbolTableStatic()`, `isWalEnabled()`; per the `is`/`has` rule this reads better as e.g. `isDotSplitting()` (propagates to the two overrides). Public API, so worth settling now.

3. **`SqlUtilTest` (`testExprColumnAliasSeqSizeDigitBoundaryUnderTruncation`) — incorrect comment.** It claims `Math.log10` "rounds wrong at exact powers of ten." That is false — Java's `Math.log10` javadoc *guarantees* `10^n → n`, and the old vs new `seqSize` were verified identical for every value (10/100/1000/10000…). The integer digit-count is a fine micro-optimisation, but it is **behavior-neutral, not a correctness fix**; the comment should say so.

4. **`CursorDereferenceFunctionFactory.java:59` (out-of-diff) — the one un-switched compile-path resolver.** It resolves a `cursor."col"` dereference against post-projection metadata with a quote-retaining token via verbatim `metadata.getColumnIndexQuiet`. Unreachable today (dereference needs a RECORD source; protected aliases only come from subqueries, which are CURSOR-typed and rejected earlier) and it fails closed with a clean `invalidColumn`. A one-line comment or a switch to `SqlUtil.getColumnIndexQuiet` would keep it consistent, so a future dereferenceable table-function column named `in`/`a.b` can't silently fail.

5. **PR description accuracy.** The "Store reuse in the alias dedup loops" and "Single dot-scan at join-metadata build" follow-ups repair regressions introduced **earlier on this branch** — against the merge-base, master already used a single hoisted `trimTo` entry and a single dot scan, so the merged net effect is neutral, not an efficiency gain. Accurate about branch history, but since the squash-commit is the permanent record, consider trimming the "efficiency" framing. (The genuine wins — the quote-leak/hang/regression fixes and `JsonUnnestSource` — are described correctly.)

6. **`SqlParser.java:4492-4507` / `4582-4595` — DRY.** The dotted-name normalize-and-reject-embedded-quote block is duplicated for the COLUMNS field name and the column alias (differing only by error string); worth a small shared helper.

7. **`SqlUtil.java:953` — clarity.** Returns `index` where it is provably `-1`; `return -1` reads clearer. Trivial.

8. **Test coverage gaps (non-blocking).** All honestly scoped, but worth closing:
   - The CSV `/exp` test covers only operator-token values — the **dotted `FNCL 2.5` case (the exact #6471 symptom) is not asserted over CSV**, a distinct quoting path.
   - HORIZON-JOIN and SAMPLE-BY tests cover only a dotted key, not an operator-token key; `SampleByTest` also omits `.expectSize()`.
   - No operator-token UNNEST alias `u("in")`, no NULL-measure pivot, no empty-source-table pivot.
   - Several shipped changes are **defensive/dead with no fail-without-fix test** (the mat-view `toColumnName` keying — proven dead by reverting it with tests still green — and the four "not currently exercised" runtime `getColumnIndex` swaps). Correct and honestly labeled, but a future refactor could delete them undetected.

### Downgraded (draft findings dismissed after source verification)

- **CharacterStore pivot aliasing corruption** (`protectColumnAlias(characterStore, cse.toImmutable())`) — SAFE. The growable buffer copies forward on resize, the inner-class view reads the *current* array, `ObjectPool.next()` grows rather than recycling live entries, and the `seqKey`/candidate regions are disjoint.
- **`JoinRecordMetadata` thread-local-sink double-use** — SAFE. `protectDottedColumnName` returns a materialized `.toString()` and `getThreadLocalSink()` clears on each call, so no live view survives into the second use.
- **`seqSize` "correctness fix"** — refuted empirically; identical output for all inputs including powers of ten.
- **`createExprColumnAlias` dotted-bare candidate escaping both checks (base `"".x"`)** — unreachable; could not be constructed from any SQL (`nonLiteral` bases can't take that shape). Defensive-only.
- **`splitsOnDot()` coverage incomplete** — refuted; all 5 ranged `getColumnIndexQuiet(cs,lo,hi)` impls are accounted for and pinned by `testSplitsOnDotContract`.
- **`getColumnIndex` swaps change the exception type** — refuted; old `RecordMetadata.getColumnIndex` and new `SqlUtil.getColumnIndex` both throw `InvalidColumnException.INSTANCE`.
- **`JsonUnnestSource` native buffer under-allocation from the stripped length** — non-issue; the arg is a growable capacity hint (pre-existing pattern).
- **UNNEST operator-token alias `u("in")` stored bare** — pre-existing (only the *dotted* branch changed), unconfirmed to manifest; out of scope.

### Summary

**Approve.** An unusually careful, well-tested change to a subtle SQL-compiler surface. Across all review dimensions there are **zero blocking findings** — the only concrete code defect is a deliberate, documented, harmless soft-cap overshoot in a pathological min-config case. The strip-retry / verbatim split is correctly maintained at every callsite (compile vs. ingestion/storage/wire), including enterprise; the `splitsOnDot()` metadata hierarchy is complete and pinned; the two headline aliasing risks (CharacterStore, thread-local sink) are provably safe; and the reachability labels on the defensive paths were empirically verified.

- 8 minor/low items retained; 8 draft concerns dropped as false positives after source verification (2 refuted empirically).
- 7 in-diff findings (+ test coverage), 1 out-of-diff (`CursorDereferenceFunctionFactory:59`, unreachable). The blast-radius pass classified 89 switched sites and swept `cutlass`/`cairo`/`jit`/enterprise, so the low out-of-diff count reflects a well-contained change, not an underrun.
- Tradeoffs are documented and test-pinned: a **breaking change** for clients keying on the old `"t.x+1"`-style names, and dotted output columns remain non-referenceable from an enclosing query and unpersistable via CTAS/mat-view (both fail cleanly).

Optional before merge: correct the `Math.log10` comment (#3), decide the `splitsOnDot()` name (#2), and add the CSV-dotted assertion (#8). None blocks.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

bluestreak01 (approved):
## Code review — PR #7353 (level 3)

`fix(sql): fix PIVOT and expression column names that embed double quotes` — Fixes #6471.

Full mission-critical pass: the diff plus every touched production file was read, the change surface was mapped (semantic deltas, callsite inventory, implicit contracts, cross-context exposures), a per-branch test coverage map was built, and 13 fresh-context review agents ran in parallel (Agent 8/Rust N/A — no `.rs` files), followed by per-finding source verification. This is a subtle SQL-compiler change; the two highest-risk dimensions — the *verbatim ingestion/storage/wire contract* and the *completeness of `splitsOnDot()`* — were verified independently and hold. Findings below are all verified against source; false positives are listed at the end.

### Critical
**None.** No correctness bug, off-by-one, NULL/sentinel mishandling, resource leak, concurrency hazard, data-redirect, quote leak, IO amplification, or algorithmic/data-path inefficiency was found on any reachable path. Specifically verified:

- **`splitsOnDot()` hierarchy is complete.** Every ranged `getColumnIndexQuiet(cs,lo,hi)` implementation was enumerated: `AbstractRecordMetadata` (verbatim `keyIndex`), `AnyRecordMetadata` (returns 0), `EmptyRecordMetadata` (returns -1), `PriorityMetadata` (delegates + dotted-protected local retry), `JoinRecordMetadata` (splits on unquoted dot). Only `JoinRecordMetadata` splits → correctly returns `true`; `PriorityMetadata` is the only delegating wrapper → forwards the flag. The single-arg `getColumnIndexQuiet(CharSequence)` has only the interface default, so no class can split differently via that overload. No enterprise module exists in-tree; `java-questdb-client` (ILP client submodule) contains no `RecordMetadata`/`SqlUtil` reference.
- **Verbatim contract intact.** All `SqlUtil.getColumnIndex(Quiet)` usages live in `io.questdb.griffin` (compile path). The 39 raw `.getColumnIndexQuiet(...)` ingestion/storage/wire calls (ILP, QWP, PG wire, HTTP, CSV import, WAL, TableWriter, parquet) were **not** switched; spot-checked >5. A wire-supplied `"in"` (quote bytes included) still fails to match column `in`. No redirect path exists.
- **Slice arithmetic sound.** The `hi-lo<3` guard in `quoteProtectedInteriorDot` fires before any `charAt`, so empty/1/2-char slices return `MIN_VALUE` without OOB; the retry `(1,len-1)` / `(dot+2,hi-1)` / `(lo,dot)+(dot+1,hi)` bounds were traced and are correct. `JoinRecordMetadata.getColumnIndexQuiet` now honours `lo/hi` in its dotted branch — this fixes a **pre-existing** latent mis-key on a non-zero slice.
- **Dedup loop terminates.** Empty/all-space/numeric base → `emptyContent` `"column"` placeholder; `emitQuote` requires `contentLen>0` so it's mutually exclusive with `emptyContent`; the loop always advances `sequence`. The empty-PIVOT-value infinite loop / `NegativeArraySizeException` is fixed and pinned by two `@Test(timeout=30000)` guards.
- **No thread-local-sink clobber.** `JoinRecordMetadata.protectDottedColumnName` materializes via `.toString()` before the later `getThreadLocalSink()` (which `clear()`s) reuses the sink; verified by trace and by live-engine value-dumping (`m."in"` binds to the correct column, not a base `in`).
- **Resource safety.** `JsonUnnestSource` sizes/populates its native `DirectUtf8Sink` with the same stripped name used for the C++ extraction key (sink is growable, so a shorter stripped name is a valid capacity hint); `columnNameSinks` is assigned before the `try`, and the `catch(Throwable){close();throw;}` frees every already-allocated sink on any mid-loop failure. `CharacterStore`/metadata paths hold no native memory.

### Moderate
**None that block.** Two genuine but non-blocking items:

1. **PR description framing — `SqlUtil.java` / `JoinRecordMetadata.java` "efficiency" follow-ups.** The "Store reuse in the alias dedup loops" and "Single dot-scan at join-metadata build" bullets are described as efficiency wins, but verified against `origin/master`: master already uses a single hoisted `trimTo` entry and a single dot scan (`protectDottedColumnName` does not exist on master). Both follow-ups repair regressions introduced *earlier on this branch*, so the net effect vs. the merge target is **neutral, not a gain**. Because the squash-commit is the permanent record, a future reader will misread these as improvements over master. Recommend re-labelling as branch-internal regression repairs. *(Behaviorally and performance-wise inert vs. the merge target — description accuracy only.)*

### Minor
1. **`RecordMetadata.splitsOnDot()` breaks the `is`/`has` boolean-naming convention** (`RecordMetadata.java:236`; overrides in `JoinRecordMetadata`, `PriorityMetadata`). Every other boolean on the interface obeys it (`isColumnIndexed`, `isWalEnabled`, `hasColumn`). This is brand-new internal API with a handful of call sites — cheapest to settle now (e.g. `isDotSplitting()`).
2. **`SqlParser.parseUnnest0` DRY** (`SqlParser.java` ~4486–4507 and ~4579–4596). The dotted-name normalize-and-reject-embedded-quote block is duplicated for the COLUMNS field name and the column alias, differing only by the error noun. Extract a shared private helper. (`SqlUtil.protectColumnAlias` is not a drop-in — it doesn't reject an embedded `"`.)
3. **Compile-time constant-factor redundancies (both cold, bounded by alias length ≤ `column.alias.generated.max.size`, default 64; fired only on the uncommon quote-protected-alias path).** (a) `createColumnAlias` (`SqlUtil.java` ~293/351 vs ~378) computes the interior dot inside `isQuoteProtectedAlias(...)`, discards it, then `contentHasDot` re-scans the identical slice — capture and reuse the `quoteProtectedInteriorDot` int. (b) `createExprColumnAlias` computes `baseDisallowed` unconditionally though it is dead on the literal+`quote` path. Both are strictly eliminable; the dedicated performance agent independently judged the surrounding paths algorithmically optimal, and neither has any IO, per-row, or complexity-class dimension — hence non-blocking, but worth tidying while in there.
4. **Duplicate-column error message leaks the compiler-internal quoted form (pre-existing).** `SELECT x AS "a.b", y AS "a.b"` raises `Duplicate column [name="a.b"]` (quotes shown) while the operator-token equivalent raises the clean `Duplicate column [name=in]` (`SqlException.duplicateColumn`, verbatim `colName`). Thematically the exact class of leak this PR removes, in an error-message path rather than result-set metadata. Route the dotted branch through `SqlUtil.toColumnName` for consistency.
5. **Empty-content placeholder can exceed the *soft* cap at pathological min-config.** `createExprColumnAlias` (`SqlUtil.java` ~630): at `maxLength=4`, a ~100-collision empty/whitespace base yields `c_100` (5 chars). Deliberate, documented tradeoff (1-char placeholder floor over a bare `_100`); result is valid and unique. Either soften the "within the configured cap" comment or fold the suffix width into the truncation.
6. **`@Test(timeout = 30000)`** (`SqlUtilTest.java:237`, `PivotTest.java:457`) — per the 5+-digit underscore convention, `30_000`.
7. **`SqlUtilTest` reinvents alias scaffolding ~19×.** The `CharacterStore`/`aliasMap`/`seqMap` three-liner is repeated across the new `createExprColumnAlias` tests; the sibling `SqlUtilCreateColumnAliasTest` already solves this with a `Ctx` holder + `alias(...)` helper. Extract a shared helper/`@Before`.
8. **Two comments overstate what they prove.** `SqlUtilTest.testExprColumnAliasSeqSizeDigitBoundaryUnderTruncation` claims `Math.log10` "rounds wrong at exact powers of ten" — Java's `Math.log10` is contractually exact there, and old vs. new `seqSize` were verified identical for all inputs; the integer digit-count is a behavior-neutral micro-opt, not a correctness fix. Soften the comment.

### Coverage gaps (low, non-blocking — behavior is covered elsewhere)
- **Dotted `FNCL 2.5` header over CSV `/exp` is unasserted** (`ExpParquetExportTest.testExpCsvExportPivotProtectedColumnNames` covers only operator-token headers). The dotted display name — the literal #6471 symptom — is asserted over JSON (`IODispatcherTest`) and PG (`PGPivotTest`) but not CSV. Same `toColumnName` code path; add the dotted assertion to the CSV test.
- **Operator-token key is unasserted for HORIZON JOIN and SAMPLE BY** (`HorizonJoinTest`/`SampleByTest` cover only the dotted `"a.b"` key). `GroupByTest.testGroupByKeyOnQuoteProtectedAlias` covers both, and the strip-retry resolves against verbatim `GenericRecordMetadata` here, so risk is low.
- No NULL-measure or empty-source-table pivot with protected column names (orthogonal to naming; empty-*value* is covered).

### Downgraded (draft findings dismissed after source verification)
- **`createExprColumnAlias` embedded-quote leak for a non-literal compound base** (`t."a.b"+1`, `"a.b"*"c.d"`) — real in isolation but **unreachable via SQL**: physical column names cannot contain dots (`CREATE TABLE t ("a.b" INT)` is rejected), and expressions over a qualified protected subquery alias collapse to the `column` placeholder. Dead today; aligns with the documented quoted-content limitation. Brittle, not a finding.
- **~6 "defensive uniformity" resolver swaps** (`RankFunctionFactory`, `SampleByFirstLastRecordCursorFactory`, `assembleFunctionReference`, `getSampleBySymbolKeyIndex`, `prepareLatestByColumnIndexes`, `allGroupsFirstLastWithSingleSymbolFilter`) — **verified behavior-preserving**: each token names a physical page-frame/reader column that arrives bare, so the verbatim lookup hits first and the strip never engages. No behavioral change → EXEMPT from a coverage requirement. (`RankFunctionFactory` additionally converts a latent silent-wrong-column into a clean `invalidColumn` throw — strictly safer.)
- **`CreateMatViewOperationImpl.toColumnName` keying** — proven identity/defensive (dotted mat-view alias fails earlier as an invalid column; PIVOT is disallowed in a mat view); `CreateMatViewTest.testCreateMatViewCopySymbolCapacityProtectedAlias` is a reachability-boundary pin, honestly labelled, green even if reverted.
- **CharacterStore aliasing / thread-local-sink double-use / concurrent-compile hazards** — all SAFE: entries live in strictly increasing disjoint store regions, `trimTo` never rewinds below committed `seqKey`/sibling regions, growth `arraycopy`s and reassigns the shared `chars` field so committed views stay valid, `disallowedAliases` is `static final` read-only, and `CharacterStore`/`JoinRecordMetadata` are per-compiler single-threaded.
- **`seqSize` "correctness fix"** — refuted empirically; identical output for all inputs including exact powers of ten.
- **`splitsOnDot()` coverage incomplete** — refuted; all ranged impls accounted for and pinned by `testSplitsOnDotContract`.
- **`JsonUnnestSource` native under-allocation from the stripped length** — non-issue; `DirectUtf8Sink` is growable and the capacity is a hint.
- **Submodule / local compile failure** — environmental only; the `java-questdb-client` bump is **not** part of this PR's 30 files (it arrived via a merged-master commit); CI syncs submodules.

### Coverage map (per behavioral change → fail-without-fix guard)
| Behavioral change | Guard test (fails if reverted) | Verdict |
|---|---|---|
| `toColumnName` strip at result-set metadata (27 sites) | `PivotTest.testPivotStaticListWith{Dotted,OperatorToken}Values`, `IODispatcherTest.testJsonQueryPivotProtectedColumnNames`, `PGPivotTest.testPivot{Dotted,OperatorToken}ColumnNames`, `WindowFunctionTest.testProtectedOutputAliasSurfacesClean`, `ColumnAliasExpressionTest` (modified quoted→clean) | TESTED |
| `getColumnIndexQuiet` strip-retry | `SqlUtilTest.testGetColumnIndexQuietHelperStripsProtectedAliasOnMiss`, `GroupByTest.testGroupByKeyOnQuoteProtectedAlias` | TESTED |
| `splitsOnDot` mis-split skip | `JoinRecordMetadataTest.testQuoteProtectedDottedAliasDoesNotMisbind{,ThroughPriorityMetadata}`, `testSplitsOnDotContract` | TESTED |
| empty-PIVOT hang fix | `SqlUtilTest.testExprColumnAliasEmptyBase` + `PivotTest.testPivotEmptyValue` (both `timeout=30000`) | TESTED (via timeout) |
| trailing-space leak fix | `SqlUtilTest.testExprColumnAliasTruncationDropsTrailingSpace` / `...ProtectedTrailingSpaceTrimmed` / `...AllSpaceBaseYieldsColumn`, `PivotTest.testPivot*WithTrailingSpace` | TESTED |
| stale-quote after dedup/truncation | `SqlUtilTest.testExprColumnAliasTruncationDropsStaleQuotes` / `...DedupOperatorTokenDropsStaleQuotes`, `ColumnAliasExpressionTest.testMaxSizeTruncationDropsProtectiveQuotes` | TESTED |
| dotted UNNEST alias | `UnnestTest.testDottedColumnAlias{,NonStandardQuotes,WithEmbeddedQuoteRejected}` | TESTED |
| JSON UNNEST dotted key | `JsonUnnestTest.testColumnsDottedFieldNameExtractsValue` (+variants) | TESTED |
| operator-token pivot through join | `PivotTest.testPivotOperatorTokenValueThroughJoin` / `...DuplicateOperatorTokenValuesThroughJoin`, `JoinRecordMetadataTest.testComposedOperatorTokenReferenceResolvesViaStrip` / `testRangedLookupHonorsBounds` | TESTED |
| PIVOT case-variant dedup | `PivotTest.testPivotStaticListWithCaseVariantOperatorTokenValues`, `...QuotedContentValueCollidesWithOperatorToken` | TESTED |
| CreateMatView symbol-capacity keying | `CreateMatViewTest.testCreateMatViewCopySymbolCapacityProtectedAlias` (reachability-boundary pin) | EXEMPT (no-behavioral-change) |
| ~6 defensive resolver swaps | none (verified behavior-preserving on all reachable inputs) | EXEMPT |

Mutation spot-check: each of the 5 most dangerous changed lines (the `splitsOnDot` guard, the verbatim-contract `MIN_VALUE` short-circuit, `emitQuote`-after-trim, the `emptyContent` placeholder, the `JoinRecordMetadata` ranged split bounds) is pinned by a dedicated fail-without-fix test.

### Summary
**Verdict: approve.** An unusually careful, exhaustively-tested fix to a subtle SQL-compiler surface, with the two highest-risk invariants (verbatim-ingestion contract; `splitsOnDot()` completeness) verified airtight and every headline behavior pinned by an adversarially-verified fail-without-fix test.

- Correctness & performance gate: **passed** — zero confirmed correctness, data-integrity, resource, concurrency, IO, or data-path/algorithmic findings. The only performance items (Minor #3) are immeasurable, cold-compile-path, constant-factor redundancies bounded by a small constant (alias length ≤ 64), which the primary and adversarial performance agents both judged non-blocking; no IO/per-row/complexity dimension.
- Test gate: **passed** — no query-result `assertSql`/`TestUtils.assertSql`/`getPlan`/`assertPlanNoLeakCheck` and no `.returnsOnce(...)` introduced; ~85 new tests, all terminating in real `.returns(...)`/`.fails(...)`/`assertException` assertions; every UNTESTED-classifiable behavioral change is either covered or verified no-behavioral-change (EXEMPT).
- Coverage map: 12 behavioral-change rows — 10 TESTED, 2 EXEMPT (verified no-behavioral-change), 0 UNTESTED.
- Draft findings: ~40 raised across 13 agents; **verified retained**: 1 Moderate (description framing) + 8 Minor/low + 3 low coverage gaps; **dropped as false positives / EXEMPT**: 8 (embedded-quote path unreachable, defensive swaps behavior-preserving, CharacterStore/sink/concurrency safe, seqSize neutral, splitsOnDot complete, native-alloc hint, submodule environmental).
- In-diff / out-of-diff / pre-existing split: findings are in-diff except the pre-existing duplicate-column error-message leak (Minor #4) and the pre-existing `JoinRecordMetadata` non-zero-slice mis-key that this PR fixes. The low out-of-diff count reflects a well-contained, compile-path-only change (89 switched sites swept across `cutlass`/`cairo`/`jit`), not an underrun.
- Tradeoffs (documented and test-pinned): breaking change for clients keying on the old `"t.x+1"`-style names; dotted output columns remain non-referenceable from an enclosing query and unpersistable via CTAS/mat-view — both fail cleanly.

Optional before merge (none blocking): re-label the two branch-internal "efficiency" follow-ups (Moderate #1), reuse the computed interior dot / gate `baseDisallowed` (Minor #3), settle the `splitsOnDot` name (Minor #1), and add the CSV dotted-header assertion (coverage gap).

<sub>Reviewed at level 3 with 13 parallel fresh-context agents + per-finding source verification.</sub>

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 336 / 347 (96.83%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/RankFunctionFactory.java|2|3|66.67%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CreateMatViewOperationImpl.java|7|8|87.50%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|73|81|90.12%|
|:large_blue_circle:|io/questdb/griffin/SqlUtil.java|179|180|99.44%|
|:large_blue_circle:|io/questdb/griffin/SqlParser.java|17|17|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/RecordMetadata.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/JoinRecordMetadata.java|26|26|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFirstLastRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|16|16|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/RecordAsAFieldRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/FunctionParser.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/GroupByUtils.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/JsonUnnestSource.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/PriorityMetadata.java|5|5|100.00%|

PR #7355, opened 2026-07-05
fix(sql): preserve row order for order-sensitive aggregates

brunocalza:
## Summary

Make sure `string_agg`, `haversine_dist_deg`, `array_agg`, `sparkline`, and `string_distinct_agg` preserve the order

## Context

 [optimiseOrderBy](https://github.com/questdb/questdb/blob/9.4.3/core/src/main/java/io/questdb/griffin/SqlOptimiser.java#L6117) inside `SqlOptimiser` drops the ordering in sub-queries when not needed. 

Some GROUP BY functions require order though, and there's a check for that:

```java
if (model.getSelectModelType() == IQueryModel.SELECT_MODEL_GROUP_BY) {
	for (int i = 0; i < n && orderByMnemonic != OrderByMnemonic.ORDER_BY_REQUIRED; i++) {
		if (hasOrderedGroupByFunc(columns.getQuick(i).getAst())) {
			orderByMnemonic = OrderByMnemonic.ORDER_BY_REQUIRED;
			break;
		}
	}
}
```

`hasOrderedGroupByFunc` uses the following list to decide on that:
```java
static {
	orderedGroupByFunctions = new LowerCaseAsciiCharSequenceHashSet();
	orderedGroupByFunctions.add("first");
	orderedGroupByFunctions.add("first_not_null");
	orderedGroupByFunctions.add("last");
	orderedGroupByFunctions.add("last_not_null");
}
```
`string_agg`, `haversine_dist_deg`, `array_agg`, `sparkline`, and `string_distinct_agg` are missing from that list.

Example for `string_agg`:
```sql
-- Non-designated timestamp => table scans in INSERTION order
CREATE TABLE words (val STRING, seq INT, ts TIMESTAMP);                                                                         
-- Insertion order spells  A, D, B, C  (scrambled).
-- The `seq` column encodes the INTENDED order:  A, B, C, D.                                                                                           
INSERT INTO words VALUES                                                       
('A', 1, '2020-01-01T00:00:00.000000Z'),                                         
('D', 4, '2020-01-01T00:00:01.000000Z'),                                         
('B', 2, '2020-01-01T00:00:02.000000Z'),                                        
 ('C', 3, '2020-01-01T00:00:03.000000Z');

SELECT string_agg(val, ',') FROM (SELECT * FROM words ORDER BY seq);
--  returns A,D,B,C

SELECT string_agg(val, ',') FROM (SELECT * FROM words ORDER BY seq DESC);
--  also returns A,D,B,C
```

Also, `EXPLAIN` does not show any `SORT`. 
```bash
explain SELECT string_agg(val, ',') FROM (SELECT * FROM words ORDER BY seq DESC);
              QUERY PLAN              
--------------------------------------
 GroupBy vectorized: false
   values: [string_agg(val,,)]
     PageFrame
         Row forward scan
         Frame forward scan on: words
(5 rows)
```

## Changes

- Adds `string_agg` and `haversine_dist_deg` to `orderedGroupByFunctions` list

### DIFF
--- a/core/src/test/java/io/questdb/test/griffin/OrderByAdviceTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/OrderByAdviceTest.java
@@ -32,6 +32,28 @@ import org.junit.Test;
 public class OrderByAdviceTest extends AbstractCairoTest {
 
     @Test
+    public void testArrayAggPreservesOrderByAdvice() throws Exception {
+        execute("create table nums (val double, seq int)");
+        execute("""
+                insert into nums values
+                (1.0, 1),
+                (4.0, 4),
+                (2.0, 2),
+                (3.0, 3)
+                """);
+
+        assertQuery("select array_agg(val) arr from (select * from nums order by seq)")
+                .ddl(null)
+                .noRandomAccess()
+                .expectSize()
+                .withPlanContaining("GroupBy", "Encode sort light", "keys: [seq]")
+                .returns("""
+                        arr
+                        [1.0,2.0,3.0,4.0]
+                        """);
+    }
+
+    @Test
     public void testCreateDesignatedTimestampFromMultipleOrderBy() throws Exception {
         assertQuery("select * from x order by t, a")
                 .ddl("create table x as (" +
@@ -813,6 +835,28 @@ public class OrderByAdviceTest extends AbstractCairoTest {
                         1326447242\t1970-01-01T00:00:00.006000Z
                         1548800833\t1970-01-01T00:00:00.002000Z
                         1868723706\t1970-01-01T00:00:00.008000Z
+                        """);
+    }
+
+    @Test
+    public void testSparklinePreservesOrderByAdvice() throws Exception {
+        execute("create table series (val double, seq int)");
+        execute("""
+                insert into series values
+                (0.0, 1),
+                (3.0, 4),
+                (1.0, 2),
+                (2.0, 3)
+                """);
+
+        assertQuery("select sparkline(val) spark from (select * from series order by seq)")
+                .ddl(null)
+                .noRandomAccess()
+                .expectSize()
+                .withPlanContaining("GroupBy", "Encode sort light", "keys: [seq]")
+                .returns("""
+                        spark
+                        ▁▃▅█
                         """);
     }
 
--- a/core/src/test/java/io/questdb/test/griffin/OrderByAdviceTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/OrderByAdviceTest.java
@@ -183,6 +183,27 @@ public class OrderByAdviceTest extends AbstractCairoTest {
     }
 
     @Test
+    public void testHaversineDistDegPreservesOrderByAdvice() throws Exception {
+        execute("create table geo (lat double, lon double, seq int, ts timestamp)");
+        execute("""
+                insert into geo values
+                (0, 0,  1, '2020-01-01T00:00:00.000000Z'),
+                (0, 30, 4, '2020-01-01T00:00:03.000000Z'),
+                (0, 10, 2, '2020-01-01T00:00:01.000000Z'),
+                (0, 20, 3, '2020-01-01T00:00:02.000000Z')
+                """);
+
+        assertQuery("select haversine_dist_deg(lat, lon, ts) d from (select * from geo order by seq)")
+                .ddl(null)
+                .noRandomAccess()
+                .expectSize()
+                .returns("""
+                        d
+                        3335.893876029014
+                        """);
+    }
+
+    @Test
     public void testNoKeyGroupBy() throws Exception {
         assertQuery("select sum(price) / count() from x where price > 0")
                 .ddl("""
@@ -791,6 +812,27 @@ public class OrderByAdviceTest extends AbstractCairoTest {
                         1326447242\t1970-01-01T00:00:00.006000Z
                         1548800833\t1970-01-01T00:00:00.002000Z
                         1868723706\t1970-01-01T00:00:00.008000Z
+                        """);
+    }
+
+    @Test
+    public void testStringAggPreservesOrderByAdvice() throws Exception {
+        execute("create table words (val string, seq int)");
+        execute("""
+                insert into words values
+                ('A', 1),
+                ('D', 4),
+                ('B', 2),
+                ('C', 3)
+                """);
+
+        assertQuery("select string_agg(val, ',') s from (select * from words order by seq)")
+                .ddl(null)
+                .noRandomAccess()
+                .expectSize()
+                .returns("""
+                        s
+                        A,B,C,D
                         """);
     }
 
--- a/core/src/test/java/io/questdb/test/griffin/OrderByAdviceTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/OrderByAdviceTest.java
@@ -46,7 +46,7 @@ public class OrderByAdviceTest extends AbstractCairoTest {
                 .ddl(null)
                 .noRandomAccess()
                 .expectSize()
-                .withPlanContaining("GroupBy", "Encode sort light", "keys: [seq]")
+                .withPlanContaining("GroupBy", "keys: [seq]")
                 .returns("""
                         arr
                         [1.0,2.0,3.0,4.0]
@@ -219,7 +219,7 @@ public class OrderByAdviceTest extends AbstractCairoTest {
                 .ddl(null)
                 .noRandomAccess()
                 .expectSize()
-                .withPlanContaining("GroupBy", "Encode sort light", "keys: [seq]")
+                .withPlanContaining("GroupBy", "keys: [seq]")
                 .returns("""
                         d
                         3335.893876029014
@@ -853,7 +853,7 @@ public class OrderByAdviceTest extends AbstractCairoTest {
                 .ddl(null)
                 .noRandomAccess()
                 .expectSize()
-                .withPlanContaining("GroupBy", "Encode sort light", "keys: [seq]")
+                .withPlanContaining("GroupBy", "keys: [seq]")
                 .returns("""
                         spark
                         ▁▃▅█
@@ -875,10 +875,56 @@ public class OrderByAdviceTest extends AbstractCairoTest {
                 .ddl(null)
                 .noRandomAccess()
                 .expectSize()
-                .withPlanContaining("GroupBy", "Encode sort light", "keys: [seq]")
+                .withPlanContaining("GroupBy", "keys: [seq]")
                 .returns("""
                         s
                         A,B,C,D
+                        """);
+    }
+
+    @Test
+    public void testStringDistinctAggPreservesOrderByAdvice() throws Exception {
+        execute("create table words (val string, seq int)");
+        execute("""
+                insert into words values
+                ('A', 1),
+                ('D', 4),
+                ('B', 2),
+                ('C', 3),
+                ('B', 5)
+                """);
+
+        assertQuery("select string_distinct_agg(val, ',') s from (select * from words order by seq)")
+                .ddl(null)
+                .noRandomAccess()
+                .expectSize()
+                .withPlanContaining("GroupBy", "keys: [seq]")
+                .returns("""
+                        s
+                        A,B,C,D
+                        """);
+    }
+
+    @Test
+    public void testStringDistinctAggPreservesOrderByAdviceDesc() throws Exception {
+        execute("create table words (val string, seq int)");
+        execute("""
+                insert into words values
+                ('A', 1),
+                ('D', 4),
+                ('B', 2),
+                ('C', 3),
+                ('B', 5)
+                """);
+
+        assertQuery("select string_distinct_agg(val, ',') s from (select * from words order by seq desc)")
+                .ddl(null)
+                .noRandomAccess()
+                .expectSize()
+                .withPlanContaining("GroupBy", "keys: [seq desc]")
+                .returns("""
+                        s
+                        B,D,C,A
                         """);
     }
 
--- a/core/src/test/java/io/questdb/test/griffin/OrderByAdviceTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/OrderByAdviceTest.java
@@ -197,6 +197,7 @@ public class OrderByAdviceTest extends AbstractCairoTest {
                 .ddl(null)
                 .noRandomAccess()
                 .expectSize()
+                .withPlanContaining("GroupBy", "Encode sort light", "keys: [seq]")
                 .returns("""
                         d
                         3335.893876029014
@@ -830,6 +831,7 @@ public class OrderByAdviceTest extends AbstractCairoTest {
                 .ddl(null)
                 .noRandomAccess()
                 .expectSize()
+                .withPlanContaining("GroupBy", "Encode sort light", "keys: [seq]")
                 .returns("""
                         s
                         A,B,C,D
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/questdb/test/griffin/OrderByAdviceTest.java
```diff
@@ -182,6 +182,27 @@ from long_sequence(1000)) timestamp (ts)""")
...
+        execute("create table geo (lat double, lon double, seq int, ts timestamp)");
+        execute("""
+                insert into geo values
+                (0, 0,  1, '2020-01-01T00:00:00.000000Z'),
+                (0, 30, 4, '2020-01-01T00:00:03.000000Z'),
+                (0, 10, 2, '2020-01-01T00:00:01.000000Z'),
+                (0, 20, 3, '2020-01-01T00:00:02.000000Z')
+                """);
+
+        assertQuery("select haversine_dist_deg(lat, lon, ts) d from (select * from geo order by seq)")
```
puzpuzpuz, line 218: nit: we could also assert execution plan here.

puzpuzpuz (dismissed):
Thanks for the contribution!

core/src/test/java/io/questdb/test/griffin/OrderByAdviceTest.java
brunocalza: added that in a new commit

puzpuzpuz (dismissed):

bluestreak01 (changes requested):
@brunocalza — thanks for the fix. Reviewed at a deep level (built the branch and ran the new tests + empirical probes). The change as written is **correct**, but it's an **incomplete** fix for the bug class it addresses, so I'm requesting changes to fold in the remaining cases.

## What's correct (verified)
- Both `string_agg` (`StringAggGroupByFunction`) and `haversine_dist_deg` (`HaversineDistDegreeGroupByFunction`) are `supportsParallelism() == false` and aggregate in `computeFirst`/`computeNext` call order, so preserving the sub-query `ORDER BY` is exactly the right fix.
- Added tokens match the registered SQL names; `LowerCaseAsciiCharSequenceHashSet` handles case; both `string_agg` overloads share the `string_agg` token.
- Built and ran the two new tests: `Tests run: 2, Failures: 0`. They're genuine regression tests (fail without the fix on both the `.returns(...)` value and the `Encode sort light` plan fragment).

## Moderate

### M1 — Incomplete: `array_agg` and `sparkline` have the identical bug and are left broken
The root cause is a hand-maintained name list, and it's missing more than these two functions. Confirmed empirically (built a scratch probe, ran it, reverted):

**`array_agg`** — `SELECT array_agg(val) FROM (SELECT * FROM words ORDER BY seq)`:
- `ORDER BY seq` (correct = `[10,20,30,40]`) returned `[10.0,40.0,20.0,30.0]` (storage order)
- `ORDER BY seq DESC` returned the **identical** wrong `[10.0,40.0,20.0,30.0]`
- Plan shows no sort — inner `ORDER BY` dropped, exactly like `string_agg` before this fix.

I then temporarily added `orderedGroupByFunctions.add("array_agg")` and re-ran: `ORDER BY seq` → `[10,20,30,40]` and `ORDER BY seq DESC` → `[40,30,20,10]`. **Same root cause, same one-line fix.** Note `array_agg(D)` and `array_agg(D[])` both map to the `array_agg` token, so one entry covers both.

**`sparkline`** — `... ORDER BY seq` and `... ORDER BY seq DESC` both returned the identical `▁█▃▅` (storage order), i.e. order-insensitive output ⇒ same dropped-order bug.

Per our "bundle related fixes into one PR" convention, please add `array_agg` and `sparkline` (same token covers the `sparkline` with-params overload) to the list in this PR, each with a regression test mirroring the two you added. This is the same silent-wrong-results class you're fixing.

Checked and **not** flagging:
- `string_distinct_agg` — backed by a hash set; DISTINCT-aggregation order is implementation-defined, no clear contract. Your call whether to include it.
- `twap` — rejects any non-designated timestamp (`twap() requires the table's designated timestamp as the second argument`), so the reorder scenario can't arise. Moot.

### M2 — Design: the name-string list is fragile and keeps drifting out of sync
`orderedGroupByFunctions` couples optimizer correctness to string literals that must be manually kept in lockstep with every order-sensitive aggregate. The `array_agg`/`sparkline` gaps are direct symptoms of that drift. A more robust design is a self-declared `isOrderSensitive()` default on `GroupByFunction` (mirroring `supportsParallelism()`), queried during optimisation, so each function ships its own correctness metadata and the central list goes away. Pre-existing and larger than this PR — raising it because it's why the bug recurs.

## Minor
- **M3** — Both new tests only assert ascending order. Your own description shows the `DESC` case returning wrong results; a `DESC` assertion (different expected value/string) would lock in that direction is honored, not just that some sort was inserted.
- **M4** — `.withPlanContaining("Encode sort light")` binds to a specific sort operator name and will break if the planner later picks a different sort variant even when behavior is correct. The `.returns(...)` value is the real guarantee.
- Add **SQL** and **Bug** labels.

## Downgraded (dismissed after verification)
- "Parallel GROUP BY could still reorder these" — both are `supportsParallelism() == false`.
- "string_distinct_agg / twap have the same bug" — see M1 (implementation-defined order / designated-ts requirement).
- "array_agg can't be fixed via the list because it sorts by rowId + supports parallelism" — empirically, adding it to the list produces correct asc/desc results, same remedy.

**Verdict:** request changes — nothing in the diff is wrong, but please extend it to `array_agg` and `sparkline` (with tests) so the fix covers the whole bug class rather than two of the affected functions.

brunocalza:
Thanks for the review @bluestreak01 

I consciously left `array_agg` and `sparkline` out because adding them makes them not use parallelism (different from `string_agg` and `haversine_dist_deg` that do not support parallelism). And because their order semantics don't seem well specified, I thought it was okay to leave them out. Should we still add them? Just wanted to confirm that.

About M2. That is a much bigger change. I don't think it's worth addressing in this PR.

M3 and M4 seems really minor.

brunocalza:
I added `array_agg` and `sparkline`. 

Actually a comment I made earlier is not 100% correct: "because adding them makes them not use parallelism". The parallel group by is only disabled in the case of an ordering different than the designated timestamp. Adding `array_agg` and `sparkline` seems the right call.

bluestreak01 (changes requested):
@brunocalza

# Review: PR #7355 — `fix(sql): preserve row order for order-sensitive aggregates`

**Level 3** (full mission-critical pass). The production diff is 4 lines (`orderedGroupByFunctions.add(...)` for `string_agg`, `haversine_dist_deg`, `array_agg`, `sparkline`) plus 4 regression tests. Small diff, but the changed symbol (`orderedGroupByFunctions`) gates optimizer correctness for *every* order-sensitive aggregate, so I inventoried the whole aggregate surface and verified empirically (built the branch, ran the tests, reverted the fix in a scratch worktree, and probed sibling functions).

## What I verified (empirically)

- **The mechanism is correct for all four functions.** `GroupByNotKeyedRecordCursorFactory` / `GroupByRecordCursorFactory` feed a **synthetic `rowId++` counter in cursor order** to `computeFirst/Next`, not the storage rowId. So even the rowId-sorting collectors (`array_agg`, `sparkline`) reproduce cursor order, and forcing the sub-query `ORDER BY` to survive makes cursor order = the requested order. `string_agg`/`haversine_dist_deg` are `supportsParallelism()==false` and aggregate in call order. All four are correct.
- **This refutes the prior-review "array_agg can't be fixed via the list (sorts by rowId)" worry.** The rowId is the synthetic counter, so the list fix works. Empirically: base `[1.0,4.0,2.0,3.0]` → fixed `[1.0,2.0,3.0,4.0]`.
- **All 4 new tests are genuine regression tests.** In a scratch worktree at the PR base (fix absent, tests present) all four **fail** with exactly the storage-order wrong values (`string_agg` A,D,B,C; `haversine` 6671.79 vs 3335.89; `array_agg` [1,4,2,3]; `sparkline` ▁█▃▅). On the PR branch all four pass. 214 tests across the related aggregate suites pass — no regressions.
- **PR metadata is good:** Conventional-commit title with repeated verb; labels `Bug`, `SQL` present; no `returnsOnce`; tests use the strict `.returns(...)` battery with `.expectSize()`/`.noRandomAccess()`.

## Critical

### C1 — Incomplete fix: `string_distinct_agg` has the identical bug and is left broken (pre-existing, in-scope)

`string_distinct_agg` (all three overloads — `(Sa)` string, `(Øa)` varchar, `(Ka)` symbol; one token covers all, like `string_agg`) is **not** added to the list, yet it is order-sensitive in exactly the way the PR fixes:

- `StringDistinctAggGroupByFunction.getStr()` emits `set.get(i)` over a `CharSequenceHashSet`, and `CharSequenceHashSet.get(i)` returns `list.getQuick(i)` — the **insertion-ordered** backing list (`core/src/main/java/io/questdb/std/CharSequenceHashSet.java:161`). So the output is distinct values in **deterministic first-seen order**.
- All three variants are `supportsParallelism() == false` — structurally identical to `string_agg`.

**Empirically confirmed** the same dropped-`ORDER BY` bug (probe in a scratch worktree):

```sql
CREATE TABLE sda (val STRING, seq INT);
INSERT INTO sda VALUES ('B',2),('A',1),('C',3);   -- storage first-seen: B,A,C ; ORDER BY seq: A,B,C
SELECT string_distinct_agg(val, ',') FROM (SELECT * FROM sda ORDER BY seq);
```
Returns `B,A,C` (storage order — sub-query `ORDER BY` silently dropped, no `SORT` in the plan) instead of `A,B,C`. I then added `orderedGroupByFunctions.add("string_distinct_agg");` and re-ran: returns `A,B,C`. **Same root cause, same one-line fix.**

This is a pre-existing bug found in the exact list this PR modifies, and the PR's own title claims to cover "order-sensitive aggregates." The prior review dismissed it as "implementation-defined (hash set)", but the order is not implementation-defined — it is deterministic first-seen order, and an explicit user `ORDER BY` is being discarded. Per the bundle-fixes convention, close it here.

**Suggested fix (this PR):**
```java
orderedGroupByFunctions.add("sparkline");
orderedGroupByFunctions.add("string_distinct_agg");
```
plus a regression test mirroring the four added (choose data where storage first-seen order ≠ `ORDER BY` first-seen order, e.g. the probe above, asserting `A,B,C` and a `.withPlanContaining(...)` sort fragment).

I checked the remaining collectors and confirmed no further gaps: `twap` throws on a non-designated timestamp (`TwapGroupByFunction.java:466`) so the reorder can't arise; `vwap` is a commutative notional/volume sum (order-independent); `mode`/`arg_min`/`arg_max` are order-independent for a clear winner (tie-break genuinely implementation-defined). `string_distinct_agg` is the only omission.

## Moderate

### M1 — DESC direction is untested for all four functions
Each new test asserts only ascending order. The ascending tests *are* valid regression tests (their data distinguishes sorted order from storage order — verified they fail without the fix), so this is not Critical. But a `DESC` case (different expected value) would additionally lock in that the sort *direction* propagates, not merely that some sort was inserted. Cheap to add while editing these tests. Justification for Moderate (not Critical): the ascending assertions already observe the exact ordered output, so correctness of the ASC path is covered; the gap is only the DESC path's direction plumbing, which is shared with the already-tested `first`/`last`.

### M2 — The hand-maintained name list is the root cause of the recurring drift (design)
`orderedGroupByFunctions` couples optimizer correctness to string literals kept in manual lockstep with every order-sensitive aggregate. C1 is the *third* function silently missed by this list; the prior review already flagged the same fragility. A self-declared `isOrderSensitive()` default on `GroupByFunction` (mirroring `supportsParallelism()`), queried during `optimiseOrderBy`, would let each function ship its own metadata and delete the list. This is a larger refactor than this PR should carry — not blocking — but it is the durable fix and worth a follow-up. Justification for Moderate: purely structural; the immediate one-line additions (including C1) fully restore correctness without it.

## Minor

- **M3 — `.withPlanContaining("Encode sort light")` binds to a specific sort operator name.** If the planner later selects a different sort variant while behavior stays correct, the test breaks spuriously. Keeping a plan assertion is valuable (it proves a sort was actually inserted rather than the result matching storage order by luck); consider asserting a more stable fragment. The `.returns(...)` value is the real guarantee.
- **M4 — PR body "## Changes" is stale.** It lists only `string_agg` and `haversine_dist_deg`; `array_agg` and `sparkline` were added later but not listed. Update the section to enumerate all bundled functions (and C1 if added).

## Downgraded (false positives, dismissed after verification)

- *"array_agg sorts by rowId + supports parallelism, so the list fix can't work."* — Refuted empirically: the group-by feeds a synthetic `rowId++` in cursor order (`GroupByNotKeyedRecordCursorFactory.java:396-400`), so array_agg reproduces cursor order; adding it to the list yields correct ASC/DESC results.
- *"Parallel GROUP BY could still reorder these."* — With the sub-query `ORDER BY` preserved, the group-by runs single-threaded over the sort cursor (`string_agg`/`haversine`/`string_distinct_agg` are `supportsParallelism()==false`; `array_agg`/`sparkline` fall back off the page-frame path). Confirmed by the passing `.returns(...)` value tests.
- *"`twap`/`vwap`/`mode` have the same bug."* — `twap` rejects non-designated timestamps; `vwap` is commutative; `mode`'s winner is order-independent. No dropped-order defect.

## Coverage map

| Change (symbol → behavior) | Test | Failure link (verified) | Dimensions | Verdict |
|---|---|---|---|---|
| `string_agg` forces ORDER_BY_REQUIRED → sub-query ORDER BY kept | `testStringAggPreservesOrderByAdvice` | Asserts `A,B,C,D`; base returns `A,D,B,C` → fails (ran it) | happy ✓; DESC uncovered; concurrency N/A (parallel=false); leak ✓ (builder) | TESTED |
| `haversine_dist_deg` same | `testHaversineDistDegPreservesOrderByAdvice` | Asserts `3335.89…`; base returns `6671.79…` → fails | happy ✓; DESC uncovered; concurrency N/A; leak ✓ | TESTED |
| `array_agg` same | `testArrayAggPreservesOrderByAdvice` | Asserts `[1.0,2.0,3.0,4.0]`; base returns `[1.0,4.0,2.0,3.0]` → fails | happy ✓; DESC uncovered; NULL-elem uncovered; leak ✓ | TESTED |
| `sparkline` same | `testSparklinePreservesOrderByAdvice` | Asserts `▁▃▅█`; base returns `▁█▃▅` → fails | happy ✓; DESC uncovered; leak ✓ | TESTED |
| `string_distinct_agg` (same bug class, **not fixed**) | none | Probe returns `B,A,C` vs `A,B,C`; drops ORDER BY | happy UNTESTED; all uncovered | **UNTESTED / UNFIXED (C1)** |

## Summary

- **Verdict: request changes.** The four-function fix is correct, cleanly implemented, and backed by genuine regression tests (empirically verified fail-without-fix, pass-with-fix, zero collateral regressions). But the correctness gate is open: **C1 — `string_distinct_agg` carries the identical silent-wrong-results bug** (empirically confirmed) and belongs in this PR, since the diff modifies exactly the list that governs it and the PR title claims the whole class. Add the one token plus a regression test.
- **Coverage map: 5 behavioral changes in the bug class, 4 tested, 1 UNTESTED/UNFIXED (`string_distinct_agg`).**
- **Findings: 6 (1 Critical, 2 Moderate, 2 Minor + metadata); 3 draft false positives dropped after verification.**
- **In-diff / out-of-diff / pre-existing split:** 0 in-diff bugs, 0 out-of-diff, **1 pre-existing** (C1). The core diff itself is bug-free; the only blocking item is the omission.
- **Tradeoffs:** forcing `ORDER_BY_REQUIRED` reinstates the sort the user asked for (and may disable parallel GROUP BY when the sort is on a non-designated column) — this is the intended cost of correctness, not a regression, since the prior fast path returned wrong results. No performance concern on the optimizer path: the check is an O(1) hash lookup at compile time.

Once `string_distinct_agg` (+ test) lands, this is an approve. M1 (DESC tests) and M4 (body update) are worth folding in at the same time; M2 (self-declared `isOrderSensitive()`) is a reasonable follow-up but not required here.

brunocalza:
Added `string_distinct_agg` and a single test for DESC. Adding a DESC test for every function seems redundant.

PR #7359, opened 2026-07-07
fix(qwp): fail fast on deterministic ingest rejections instead of retrying them

kafka1991:
## The problem

When a QWP/WebSocket client sends data that a QuestDB table can **never** accept — a value that can't be converted to the column type, an array with the wrong shape, a column the schema forbids — the server rejected the frame during WAL append, but reported the rejection as a **transient** `WRITE_ERROR`.

`WRITE_ERROR` is a *retriable* policy on the client — it exists for genuinely transient conditions like disk pressure or a suspended table, and nothing in the status byte lets the client tell a permanent error apart from a transient one. So the store-and-forward client replayed the byte-identical frame up to `max_frame_rejections` times before giving up.

Net effect for a **permanent, deterministic** data error (e.g. `type coercion from VARCHAR to IPV4 is not supported`): several wasted reconnect/replay cycles for a frame that can never succeed, and the failure surfaces only after the retries instead of on the first NACK.

### Root cause

`QwpIngressProcessorState.cairoExceptionStatus` split errors coarsely: `authorization → SECURITY_ERROR`, `critical → INTERNAL_ERROR`, and **everything else → the retriable `WRITE_ERROR`**. That last bucket lumped together two very different things the client can't tell apart from the status byte:

- deterministic data/schema rejections that will fail identically on every replay, and
- genuinely transient write refusals that a retry can clear.

## The fix

Introduce a `CairoException.schemaMismatch()` marker (following the existing `authorization()` / `readOnlyAccess()` marker convention) and stamp it on the QWP-ingest error sites that are **deterministic under byte-identical replay**. `cairoExceptionStatus` maps the marker to a terminal `SCHEMA_MISMATCH`, so the client fails fast on the first NACK — zero wasted retries — and surfaces the server's real reason. String-parse failures reuse the existing `QwpParseException(SCHEMA_MISMATCH)` path (the same mechanism `numericParseError` already used).

### Now terminal (fail fast, reason preserved)

Type / structure mismatches — `QwpWalAppender`:
- unsupported type coercion (e.g. `VARCHAR → IPV4`), `cannot write X to column`, geohash precision mismatch
- long-array unsupported, array dimensionality mismatch
- designated-timestamp mismatch, new-columns-not-allowed, invalid-column-name

Value conversions — `WalColumnarRowAppender`:
- numeric precision loss (`DOUBLE 3.14 → INT`) and range overflow (`LONG → INT`)
- timestamp `micros → nanos` overflow
- decimal overflow / precision loss / value-cannot-be-converted
- string-parse failures: boolean, long256, timestamp, uuid, geohash, decimal (numeric already terminal)

Table target — `QwpTudCache`:
- write to a non-WAL table (ILP/QWP can only write WAL tables; a non-WAL table won't become writable on replay)
- batch array-dimensionality mismatch

### Deliberately still retriable `WRITE_ERROR`

The change is **surgical** — genuinely transient and forward-compat errors keep the retriable default, so at-least-once delivery on recoverable hiccups is unchanged:
- `metadata changed too many times` (concurrent ALTER exhausted its retry budget)
- `writer is null` (table busy / locked)
- `unknown QWP wire type` (a newer client against an older server — degrade to retry, not a dead sender)

The status byte also stays honest for the retriable/terminal split; the terminal data errors carry the server's message either way.

## Incidental: `CairoException` refactor

`CairoException`'s boolean marker fields are packed into a single `int flags` bitmask. Behavior-preserving (public getters/setters unchanged); its value is that `clear()` — load-bearing for the pooled flyweight subclasses — collapses to `flags = 0`, so adding a marker can no longer forget a per-field reset.

## Testing

- **Unit**: `CairoExceptionTest` (marker set/reset + flag independence) and `QwpIngressProcessorStateTest` (a `schemaMismatch()` exception → terminal `SCHEMA_MISMATCH` with the reason preserved; an *unmarked* non-critical exception stays retriable `NOT_ACCEPTING_WRITES`).
- **E2E**: `assertCoercionError` is tightened to strictly assert terminal `SCHEMA_MISMATCH` — turning the ~50 coercion/type/value-conversion cases into real regression guards — plus new cases for decimal overflow / precision loss and array-dimensionality mismatch, and `testNonWalTableRejected` updated for the non-WAL flip.
- Every new trigger was verified against a live server to produce a terminal rejection carrying the exact asserted message (e.g. `decimal value overflows DECIMAL(2,1) [column=v, decimal=100.0]`, `array dimensionality mismatch [column=v, expected=DOUBLE[], actual=DOUBLE[][]]`).

### DIFF
--- a/core/src/main/java/io/questdb/cairo/CairoException.java
+++ b/core/src/main/java/io/questdb/cairo/CairoException.java
@@ -69,43 +69,23 @@ public class CairoException extends RuntimeException implements Sinkable, Flywei
     // literal first lets any later reword land without scattering the change.
     public static final String READ_ONLY_ACCESS_MESSAGE = "replica access is read-only";
     private static final StackTraceElement[] EMPTY_STACK_TRACE = {};
+    private static final int FLAG_AUTHORIZATION_ERROR = 1;
+    private static final int FLAG_CACHEABLE = 1 << 1;
+    private static final int FLAG_CANCELLATION = 1 << 2; // query was explicitly cancelled by the user
+    private static final int FLAG_HOUSEKEEPING = 1 << 3;
+    private static final int FLAG_INTERRUPTION = 1 << 4; // query timed out
+    private static final int FLAG_OUT_OF_MEMORY = 1 << 5;
+    private static final int FLAG_PREFERENCES_OUT_OF_DATE_ERROR = 1 << 6;
+    private static final int FLAG_READ_ONLY_ACCESS_REFUSAL = 1 << 7;
+    private static final int FLAG_SCHEMA_MISMATCH = 1 << 8;
     protected final StringSink message = new StringSink();
     protected final StringSink nativeBacktrace = new StringSink();
     protected int errno;
-    private boolean authorizationError = false;
-    private boolean cacheable;
-    private boolean cancellation; // when query is explicitly cancelled by user
-    private boolean housekeeping;
-    private boolean interruption; // used when a query times out
+    private int flags;
     private int messagePosition;
-    private boolean outOfMemory;
-    private boolean preferencesOutOfDateError = false;
-    private boolean readOnlyAccessRefusal = false;
 
     public static CairoException authorization() {
         return nonCritical().setAuthorizationError();
-    }
-
-    /**
-     * A write refused BECAUSE the node is read-only -- statically
-     * ({@code readonly=true} in the server configuration) or dynamically (the
-     * role-derived read-only of an enterprise replica, including a demote in
-     * flight). Carries the authorization flag, the canonical
-     * {@link #READ_ONLY_ACCESS_MESSAGE} and the read-only-refusal marker
-     * ({@link #isReadOnlyAccessRefusal()}) in lockstep, so the refusal's CAUSE
-     * travels with the exception.
-     * <p>
-     * Protocol layers that must tell a transient role-demote refusal apart from
-     * a genuine ACL denial (e.g. the QWP NACK classification) key on the marker.
-     * They must NOT re-read live engine state at catch time -- that races with a
-     * demote revert (the drain-timeout PRIMARY restore can land between the
-     * throw and the catch; nothing fences the propagation) -- and must NOT match
-     * the message text, which is brittle. Every "node is read-only" refusal site
-     * uses this factory so the marker is correct by construction; sites must not
-     * hand-roll {@code authorization().put(READ_ONLY_ACCESS_MESSAGE)}.
-     */
-    public static CairoException readOnlyAccess() {
-        return authorization().setReadOnlyAccessRefusal().put(READ_ONLY_ACCESS_MESSAGE);
     }
 
     public static CairoException critical(int errno) {
@@ -216,6 +196,44 @@ public class CairoException extends RuntimeException implements Sinkable, Flywei
         return nonCritical().put("timeout, query aborted").setInterruption(true);
     }
 
+    /**
+     * A write refused BECAUSE the node is read-only -- statically
+     * ({@code readonly=true} in the server configuration) or dynamically (the
+     * role-derived read-only of an enterprise replica, including a demote in
+     * flight). Carries the authorization flag, the canonical
+     * {@link #READ_ONLY_ACCESS_MESSAGE} and the read-only-refusal marker
+     * ({@link #isReadOnlyAccessRefusal()}) in lockstep, so the refusal's CAUSE
+     * travels with the exception.
+     * <p>
+     * Protocol layers that must tell a transient role-demote refusal apart from
+     * a genuine ACL denial (e.g. the QWP NACK classification) key on the marker.
+     * They must NOT re-read live engine state at catch time -- that races with a
+     * demote revert (the drain-timeout PRIMARY restore can land between the
+     * throw and the catch; nothing fences the propagation) -- and must NOT match
+     * the message text, which is brittle. Every "node is read-only" refusal site
+     * uses this factory so the marker is correct by construction; sites must not
+     * hand-roll {@code authorization().put(READ_ONLY_ACCESS_MESSAGE)}.
+     */
+    public static CairoException readOnlyAccess() {
+        return authorization().setReadOnlyAccessRefusal().put(READ_ONLY_ACCESS_MESSAGE);
+    }
+
+    /**
+     * A non-critical error raised BECAUSE the wire value's type, shape or
+     * precision is fundamentally incompatible with the target column (e.g.
+     * an unsupported type coercion or a geohash precision mismatch). The
+     * refusal is DETERMINISTIC under byte-identical replay -- the same bytes
+     * produce the same rejection every time -- so protocol layers that must
+     * tell a permanent data error apart from a transient write refusal (the
+     * QWP NACK classification) key on {@link #isSchemaMismatch()} to reject
+     * it terminally instead of retrying. Every such site uses this factory so
+     * the marker is correct by construction; callers must not re-derive the
+     * distinction from message text.
+     */
+    public static CairoException schemaMismatch() {
+        return nonCritical().setSchemaMismatch();
+    }
+
     public static CairoException sequencerMetadataOpenFailed(TableToken tableToken, int causeErrno, CharSequence causeMessage) {
         return critical(SEQUENCER_METADATA_OPEN_FAILED)
                 .put("could not open sequencer metadata [table=").put(tableToken)
@@ -290,7 +308,7 @@ public class CairoException extends RuntimeException implements Sinkable, Flywei
     }
 
     public boolean isAuthorizationError() {
-        return authorizationError;
+        return (flags & FLAG_AUTHORIZATION_ERROR) != 0;
     }
 
     public boolean isBlockApplyError() {
@@ -298,11 +316,11 @@ public class CairoException extends RuntimeException implements Sinkable, Flywei
     }
 
     public boolean isCacheable() {
-        return cacheable;
+        return (flags & FLAG_CACHEABLE) != 0;
     }
 
     public boolean isCancellation() {
-        return cancellation;
+        return (flags & FLAG_CANCELLATION) != 0;
     }
 
     public boolean isCritical() {
@@ -325,11 +343,11 @@ public class CairoException extends RuntimeException implements Sinkable, Flywei
     }
 
     public boolean isHousekeeping() {
-        return housekeeping;
+        return (flags & FLAG_HOUSEKEEPING) != 0;
     }
 
     public boolean isInterruption() {
-        return interruption;
+        return (flags & FLAG_INTERRUPTION) != 0;
     }
 
     public boolean isMetadataValidation() {
@@ -343,7 +361,11 @@ public class CairoException extends RuntimeException implements Sinkable, Flywei
     }
 
     public boolean isOutOfMemory() {
-        return outOfMemory;
+        return (flags & FLAG_OUT_OF_MEMORY) != 0;
+    }
+
+    public boolean isPreferencesOutOfDateError() {
+        return (flags & FLAG_PREFERENCES_OUT_OF_DATE_ERROR) != 0;
     }
 
     /**
@@ -352,11 +374,16 @@ public class CairoException extends RuntimeException implements Sinkable, Flywei
      * for genuine ACL denials.
      */
     public boolean isReadOnlyAccessRefusal() {
-        return readOnlyAccessRefusal;
+        return (flags & FLAG_READ_ONLY_ACCESS_REFUSAL) != 0;
     }
 
-    public boolean isPreferencesOutOfDateError() {
-        return preferencesOutOfDateError;
+    /**
+     * Whether this refusal is a deterministic wire-value/column-type mismatch
+     * (set only by {@link #schemaMismatch()}). Implies the error is
+     * non-critical and permanent -- replay of the same bytes cannot succeed.
+     */
+    public boolean isSchemaMismatch() {
+        return (flags & FLAG_SCHEMA_MISMATCH) != 0;
     }
 
     public boolean isSequencerMetadataOpenFailed() {
@@ -426,26 +453,26 @@ public class CairoException extends RuntimeException implements Sinkable, Flywei
     }
 
     public CairoException setCacheable(boolean cacheable) {
-        this.cacheable = cacheable;
+        setFlag(FLAG_CACHEABLE, cacheable);
         return this;
     }
 
     public CairoException setCancellation(boolean cancellation) {
-        this.cancellation = cancellation;
+        setFlag(FLAG_CANCELLATION, cancellation);
         return this;
     }
 
     public void setHousekeeping(boolean housekeeping) {
-        this.housekeeping = housekeeping;
+        setFlag(FLAG_HOUSEKEEPING, housekeeping);
     }
 
     public CairoException setInterruption(boolean interruption) {
-        this.interruption = interruption;
+        setFlag(FLAG_INTERRUPTION, interruption);
         return this;
     }
 
     public CairoException setOutOfMemory(boolean outOfMemory) {
-        this.outOfMemory = outOfMemory;
+        setFlag(FLAG_OUT_OF_MEMORY, outOfMemory);
         return this;
     }
 
@@ -494,17 +521,30 @@ public class CairoException extends RuntimeException implements Sinkable, Flywei
     }
 
     private CairoException setAuthorizationError() {
-        this.authorizationError = true;
+        this.flags |= FLAG_AUTHORIZATION_ERROR;
+        return this;
+    }
+
+    private void setFlag(int flag, boolean value) {
+        if (value) {
+            this.flags |= flag;
+        } else {
+            this.flags &= ~flag;
+        }
+    }
+
+    private CairoException setPreferencesOutOfDateError() {
+        this.flags |= FLAG_PREFERENCES_OUT_OF_DATE_ERROR;
         return this;
     }
 
     private CairoException setReadOnlyAccessRefusal() {
-        this.readOnlyAccessRefusal = true;
+        this.flags |= FLAG_READ_ONLY_ACCESS_REFUSAL;
         return this;
     }
 
-    private CairoException setPreferencesOutOfDateError() {
-        this.preferencesOutOfDateError = true;
+    private CairoException setSchemaMismatch() {
+        this.flags |= FLAG_SCHEMA_MISMATCH;
         return this;
     }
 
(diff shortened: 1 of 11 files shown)


kafka1991:
# Review: PR #7359 — `fix(qwp): fail fast on deterministic ingest rejections instead of retrying them`

Reviewed at **level 3** (full pass: bitmask refactor verified line-by-line, marker propagation traced from throw site to classifier, blast radius grepped, client-side terminal contract confirmed in the submodule, plus two parallel adversarial agents — fresh-context and test-correctness).

## Verdict: **Approve** (2 minor test/metadata items, none blocking)

This is a clean, surgical, well-documented change. The core contract holds end to end.

## What I verified as correct

- **The `int flags` bitmask refactor is behavior-preserving.** All 8 old booleans + the new `schemaMismatch` map to distinct bits (`1<<0`..`1<<8`). The one real hazard — `setCacheable(false)`/`setOutOfMemory(false)` etc. must *clear* a bit — is handled by `setFlag(flag, value)`'s `flags &= ~flag` branch (`CairoException.java:528`). `clear()` collapses to `flags = 0` (line 560), equivalent to the 8 old resets and now covering the new marker. All six subclasses are safe: `LineProtocolException`/`CopyExportException` call the full `clear(errno)`; the other three only `message.clear()` but never set a flag, and none reference the (private) removed fields. JNI `paramInstance` signature is unchanged and Rust never touches the fields.
- **Marker propagation is intact.** `isSchemaMismatch()` has exactly one reader — `cairoExceptionStatus` (`QwpIngressProcessorState.java:1052`). The `catch (Throwable t)` at `QwpWalAppender.java:907` only downgrades to a retriable `MetadataChangedException` when metadata *actually* changed concurrently; otherwise `throw t` preserves the marker. Frame-drop atomicity holds (`cancelColumnarWrite()` on all error paths).
- **Every reclassified site is genuinely deterministic under byte-identical replay**, and the deliberately-retained retriable set (`metadata changed too many times`, `writer is null`, unknown wire type) is correct. This satisfies the store-and-forward contract: SCHEMA_MISMATCH is a permanent data error, not a transient outage, so failing the producer loudly is correct — not a data-loss violation.
- **The `QwpParseException` (checked) type change compiles and classifies correctly.** All string-parse methods already declared `throws QwpParseException`; the catch at `QwpIngressProcessorState.java:894-899` maps `ErrorCode.SCHEMA_MISMATCH → Status.SCHEMA_MISMATCH`. The `WalColumnarRowAppenderTest` catch-type change is mandatory (not cosmetic), since `QwpParseException` is not a `CairoException`.
- **Client contract already exists.** `SenderError.Category.SCHEMA_MISMATCH` + `defaultPolicyFor → TERMINAL` (`CursorWebSocketSendLoop.java:666`) are already in the submodule — no client bump needed.
- **`assertCoercionError` is not over-constrained.** All ~110 reachable callers now produce terminal SCHEMA_MISMATCH; the two surviving `nonCritical()` arms a coercion test could reach (`QwpWalAppender.java:129` unknown wire type, `:889` "should not happen" fallback) are unreachable by the typed client.

## Moderate

**1. The batch array-dimensionality-mismatch reclassification ships without a test** (in-diff coverage gap). The new `testArrayDimensionalityMismatchRejected` sends one 2-D array into a 1-D column, which hits `validateArrayColumnType` (`QwpWalAppender.java:367`) — **not** the within-batch `getArrayBatchDimensionality` paths reclassified at `QwpWalAppender.java:227` and `QwpTudCache.java:619` (which need a single flush with rows of *differing* dimensionality). This is the uncovered `QwpTudCache 1/2` line in the coverage report. A newly-reclassified terminal path with no regression test is worth closing with a multi-row mixed-dimensionality batch test.

## Minor

- **PR has no labels.** Per repo conventions add labels matching scope, e.g. `Bug` + `ILP` (QWP ingest).
- **Other uncovered reclassified overflow edges** (matching the coverage report): string→`TIMESTAMP_NS` overflow (`WalColumnarRowAppender.java:1558`), non-designated micros→nanos overflow (`:1827`), float→decimal `QwpParseException` catch (`:896`). Defensive edges; optional to cover.
- **`CairoExceptionTest.testMarkerFlagsAreIndependent` proves only partial independence.** It never clears OOM or schemaMismatch individually to observe the other surviving, so a hypothetical `FLAG_OUT_OF_MEMORY == FLAG_SCHEMA_MISMATCH` aliasing would still pass. The real constants are correct distinct powers of two, so this is theoretical — one extra clear-and-assert step would make it airtight.

## Downgraded (considered, dismissed)

- *"Concurrent ALTER could make a schemaMismatch succeed on replay, so terminal is wrong"* — refuted by the metadata-version guard at `QwpWalAppender.java:913`, which re-routes to `MetadataChangedException` (retry) when metadata actually changed.
- *"non-WAL table / new-columns-not-allowed terminal is too aggressive"* — deliberate, defensible design: `autoCreateNewColumns` is a final config field and WAL-ness is a fixed table property, so these are permanent from the ingest path.
- *`QwpWalAppender.java:889` staying `nonCritical()`* — unreachable "should not happen" fallback after all real types are handled; correct to leave.

## Summary

Draft findings verified: **1 Moderate, 3 Minor confirmed; 3 false positives dropped.** In-diff: all findings (the diff's own reclassified lines lacking tests). **Out-of-diff: zero** — legitimately so; I traced the single `isSchemaMismatch()` reader and confirmed the blast radius is one classifier, not an underrun. No regressions or tradeoffs beyond the intended, documented behavior change (faster, more accurate terminal on permanent data errors). Ship it; the batch-dimensionality test is the only item I'd genuinely recommend adding.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 51 / 57 (89.47%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/line/tcp/QwpWalAppender.java|8|10|80.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalColumnarRowAppender.java|14|17|82.35%|
|:large_blue_circle:|io/questdb/cairo/CairoException.java|25|26|96.15%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpTudCache.java|2|2|100.00%|

bluestreak01 (approved):
# Review: PR #7359 — `fix(qwp): fail fast on deterministic ingest rejections instead of retrying them`

Reviewed at **level 3** (full mission-critical pass): bitmask refactor verified line-by-line, marker/exception propagation traced from every throw site to the NACK classifier, blast radius grepped, the store-and-forward containment contract checked, plus three parallel fresh-context adversarial agents (correctness/cross-context, unstructured bug-hunt, test-efficacy). Every draft finding was verified against source before landing here.

## Verdict: **Approve** — 0 Critical, 2 Moderate coverage gaps, a few Minor notes

The change is surgical and correct end to end. The reclassification only ever moves errors in the safe direction (deterministic-under-replay → terminal), the `CairoException` bitmask refactor is provably behavior-preserving, and the core mechanism plus the bulk of reclassified sites carry real regression tests with verified failure links. No correctness, data-loss, resource, concurrency, or performance finding stands.

## What I verified as correct

- **The `int flags` bitmask refactor is 100% behavior-preserving.** The 8 pre-existing booleans + the new `schemaMismatch` map to 9 distinct bits (`1<<0`..`1<<8`, `CairoException.java:71-79`). Every `isX()` reads its own bit; the five mutable setters route through `setFlag(flag, value)` (`:528-534`), whose `flags &= ~flag` branch preserves the ability to *clear* a bit (`setCacheable(false)` etc.); the four write-once markers only OR. `clear()`'s `flags = 0` (`:561`) resets a strict superset of the old 8 per-field resets (and still zeroes `messagePosition`). The flag fields were `private`, so none of the 6 subclasses (`EntryUnavailableException`, `PoolClosedException`, `EntryLockedException`, `LimitOverflowException`, `CopyExportException`, `LineProtocolException`) could ever reference them; the two flyweight subclasses reset via the full `clear(errno)`. `testMarkerFlagsAreIndependent` clears OOM and cacheable *individually* and re-asserts the schema-mismatch bit survives — genuinely proving distinct bits, not vacuous.
- **The reclassification direction is safe (no data-loss risk).** Every site moved to terminal is deterministic under byte-identical replay against the same server state: unsupported coercion, type mismatch, geohash precision, long-array-unsupported, single-row and within-batch array-dimensionality mismatch, integer-range/double-precision/decimal/timestamp overflow, and all string-parse failures. The dangerous direction — marking a *transient* condition terminal — does not occur. `writer is null` (busy/locked), `metadata changed too many times`, and the unknown-wire-type default all correctly stay retriable. `MetadataChangedException` still intercepts the concurrent-ALTER case in `QwpWalAppender`'s `catch (Throwable t)` (`:907-914`) and converts a would-be terminal into a retry — a real safety net that survives the change untouched.
- **The reclassification is fundamentally a fail-*faster* on an already-terminal path.** Before this PR a deterministic reject surfaced as a retriable `WRITE_ERROR`, but the client's poison-frame detector still escalated it to a terminal after `max_frame_rejections` wasted replays. This PR latches the correct terminal (`SCHEMA_MISMATCH`, reason preserved) on the first NACK. So even a hypothetically regressed single site degrades only to "slower terminal via poison escalation," not data loss or a wrong result — which is why the tail coverage gaps below are Moderate, not Critical.
- **The checked-`QwpParseException` swap breaks nothing.** `QwpParseException` (a `CarrierLocal` thread-local flyweight, `extends Exception`, not `CairoException`) is returned by the 3 helpers now, but every throw site (`putStringToBooleanColumn`, `...Long256`, `...Timestamp`, `...Uuid`, `...GeoHash`, the decimal loops) *already* declared `throws QwpParseException` — this is the identical path the pre-existing `numericParseError` used, so no new checked-exception surface and no compile fallout. The unwind (throw → `QwpWalAppender` `catch (Throwable t)` cancels the columnar write and rethrows → `processMessage` `catch (QwpParseException e)` at `:894`) touches no code that calls `QwpParseException.instance()`, so the flyweight message survives intact to `getFlyweightMessage()`.
- **Frame-drop atomicity and no-distress hold on the new paths.** `appendToWalColumnar`'s `catch (Throwable t)` calls `cancelColumnarWrite()` for both new exception types before rethrowing. In `processMessage`, only `catch (CommitFailedException)` and the final `catch (Throwable)` call `tudCache.setDistressed()`; the `QwpParseException` and `CairoException` catches reject the frame without distressing the connection — correct for a per-frame deterministic reject. `cairoExceptionStatus` keys on `isAuthorizationError()` first, then `isSchemaMismatch()` → `SCHEMA_MISMATCH`, and since `schemaMismatch()` is `nonCritical()`, `isCritical()` stays false, so non-QWP consumers see identical behavior.
- **The blast radius is contained to QWP.** `WalColumnarRowAppender` is reachable only through `WalWriter.getColumnarRowAppender()`, whose sole caller is `QwpWalAppender`. `isSchemaMismatch()` has exactly one reader (`cairoExceptionStatus`). Nothing outside the QWP ingress path observes the marker or the exception-type change.
- **The test tightening is a real regression guard.** `assertCoercionError` now strictly asserts `SCHEMA_MISMATCH` + `TERMINAL` (`QwpSenderE2ETest.java:4907-4914`), turning its ~165 callers into guards that fail if any exercised site regresses to `WRITE_ERROR`. Reverting either core mapping (`isSchemaMismatch()→SCHEMA_MISMATCH`, or the `QwpParseException` `SCHEMA_MISMATCH` arm) fails these plus the two reflection-based `QwpIngressProcessorStateTest` cases. The reflection in those tests (injecting a throwing `tudCache`) is justified — there is no public seam to force a chosen exception from the classifier.

## Moderate

Both items below are UNTESTED reachable reclassifications. They are Moderate rather than Critical because the pre-PR outcome for these same inputs was *also* terminal (via poison escalation), so an unguarded regression degrades latency, not correctness — but the PR's own thesis is to turn these into regression guards, and closing them here is a few lines.

1. **`invalid column name` reclassification is untested** — `QwpWalAppender.java:1010`. A new column whose name is invalid per `TableUtils.isValidColumnName` (e.g. contains `.`) against an auto-create-enabled table now fails as terminal `SCHEMA_MISMATCH` instead of retriable `WRITE_ERROR`. The QWP parse layer does not apply QuestDB's column-name character rules, so this guard is reachable, but no test asserts it — reverting `:1010` to `nonCritical()` leaves CI green. Add one `assertCoercionError` case sending a column named e.g. `"a.b"` to an auto-create table, asserting `"invalid column name"`.

2. **`string → TIMESTAMP_NS` micros→nanos overflow is untested** — `WalColumnarRowAppender.java:1555`. A timestamp *string* that parses beyond `Long.MAX/1000` micros into a nanos column now throws terminal `SCHEMA_MISMATCH`; the only "timestamp overflow" assertion in the suite covers the `long`/designated path (`:1827`), not the string-parse site. Obscure (requires a timestamp near year 294247), so low-priority, but it is a distinct changed line with no guard.

## Minor

- **Dead reclassified catch.** `WalColumnarRowAppender.java:895-896`: `putFloatToDecimalColumn`'s `catch (QwpParseException e)` is unreachable — the float→decimal loops throw only the *unchecked* `CairoException` from `doubleToDecimalConversionError` (which the PR already routes to `schemaMismatch()`, tested at `QwpSenderE2ETest:1224`). Changing the dead rethrow from `nonCritical()` to `schemaMismatch()` is a cosmetic no-op. Consider removing the vestigial `catch` (and the loops' vestigial `throws QwpParseException`) rather than restyling dead code.
- **Contract-scope drift on three config/property refusals.** `schemaMismatch()`'s javadoc scopes the marker to "the wire value's type, shape or precision." Three now-terminal sites are table-property/config refusals, not wire-value mismatches: non-WAL table (`QwpTudCache:359`), `new columns not allowed` (`QwpWalAppender:1005`), and — defensively, though unreachable for WAL tables — designated-timestamp mismatch (`:967`). Terminal is a defensible product call (an operator action, not a retry, is what unblocks them), but the user-facing category name is slightly misleading for a non-WAL rejection; one sentence in the javadoc/PR body naming this "permanence under the current server config" tradeoff would make the contract honest.
- **Unknown-wire-type asymmetry.** `TYPE_LONG_ARRAY` is now terminal (`QwpWalAppender:125`) while the `default` unknown-type branch stays retriable (`:129`) — both are deterministic under replay. Leaving unknown-type retriable is a deliberate forward-compat hedge (newer client vs older server) and is safe (over-retry, never drop), so this is acceptable as-is; worth a one-line comment so the asymmetry reads as intentional.
- **Vestigial test config.** `QwpIngressOracleFuzzTest.java` still sets `max_frame_rejections=1`, which no longer drives escalation now that the decimal-overflow poison rows latch terminal on the first NACK. Harmless dead knob; drop it or note why it stays.

## Downgraded (verified false positives from the draft)

- *"designated-timestamp mismatch (`QwpWalAppender:967`) is an untested user-visible reclassification"* — **Exempt (unreachable).** `CairoEngine` asserts WAL⟹partitioned, and partitioning requires a designated timestamp; QWP rejects non-WAL tables upfront, so `timestampIndex` is always `>= 0` on this path. The guard is defensive and cannot fire in normal ingest, so its reclassification is not an observable behavioral change.
- *"float→decimal fallback catch (`:896`) is an untested reclassification"* — **Exempt (dead code).** As above, the `catch (QwpParseException)` is unreachable; the live float→decimal error path is `doubleToDecimalConversionError → schemaMismatch()`, which is tested.
- *"tightened `assertCoercionError` could false-fail on a legitimately transient error"* — **Not a production defect.** Every sampled caller exercises a deterministic reclassified site; a rare transient WRITE_ERROR (writer busy) surfacing first is a theoretical CI-flake risk only, caught by CI, not shipped behavior.

## Coverage map (reclassified site → regression test → verdict)

| Reclassified site | Test / failure link | Verdict |
|---|---|---|
| `CairoException` bitmask + `schemaMismatch()`/`isSchemaMismatch()` | `CairoExceptionTest.testMarkerFlagsAreIndependent` / `...IsNonCriticalAndNotAuthorization` / `...ClearResetsStickySchemaMismatchFlag` — assert distinct bits + clear reset | TESTED |
| `cairoExceptionStatus`: `isSchemaMismatch→SCHEMA_MISMATCH` | `QwpIngressProcessorStateTest.testCairoExceptionStatusReturnsSchemaMismatchFor...` (+ unmarked→`NOT_ACCEPTING_WRITES` negative) — revert → observes `NOT_ACCEPTING_WRITES` | TESTED |
| `QwpParseException(SCHEMA_MISMATCH)→Status.SCHEMA_MISMATCH` | string-parse E2E cases (boolean :901, geohash :1673, long256, timestamp, uuid) — revert to `PARSE_ERROR` fails category assert | TESTED |
| boolean / long256 / timestamp / uuid / geohash / decimal string parse | `assertCoercionError` E2E + `WalColumnarRowAppenderTest` unit (now catches `QwpParseException`) | TESTED |
| `checkIntegerRange`, `checkDoubleToInteger`, `doubleToDecimalConversionError` | `assertCoercionError` E2E :951/954/957/1224 | TESTED |
| `throwDecimalOverflow`, `throwDecimalPrecisionLoss` | new E2E :1245-1251 ("decimal value overflows" / "precision loss") | TESTED |
| micros→nanos overflow (`long`/designated) | E2E timestamp-overflow case :1827 path | TESTED |
| `LONG_ARRAY` unsupported, coercion-not-supported, geohash-precision, `typeMismatchException` | `assertCoercionError` E2E :614/863/3448/3601 | TESTED |
| single-row array-dim mismatch (`validateArrayColumnType` :364) | new E2E `testArrayDimensionalityMismatchRejected` | TESTED |
| within-batch array-dim mismatch — both sites (`QwpWalAppender:227`, `QwpTudCache:619`) | new E2E `testArrayBatchDimensionalityMismatchRejected` (existing-table + auto-create branches) | TESTED |
| non-WAL table (`QwpTudCache:359`) | `QwpWebSocketSenderReceiverTest.testNonWalTableRejected` — asserts `SCHEMA_MISMATCH`+`TERMINAL` on first error and close | TESTED |
| decimal-overflow poison rows (client fail-fast contract) | `QwpIngressOracleFuzzTest.testOraclePoisonRowsFailFastAsTerminal` | TESTED |
| **`invalid column name` (`QwpWalAppender:1010`)** | none | **UNTESTED (Moderate)** |
| **string→`TIMESTAMP_NS` overflow (`WalColumnarRowAppender:1555`)** | none | **UNTESTED (Moderate)** |
| designated-timestamp mismatch (`:967`) | n/a — unreachable for WAL tables | EXEMPT |
| `putFloatToDecimalColumn` catch (`:896`) | n/a — dead catch; live path via `doubleToDecimalConversionError` | EXEMPT |

## PR metadata

- Title `fix(qwp): fail fast on deterministic ingest rejections instead of retrying them` — Conventional Commits ✓, standalone-readable end-user description ✓.
- Labels `Bug`, `QWP` present and correct. (Consider adding `ILP` since QWP is the ingest path, but not required.)
- Body is level-headed, lists the bundled reclassifications, and explicitly names the deliberately-still-retriable set and the tradeoff — good. No `Fixes #NNN`; fine if there is no tracking issue.
- Commit titles use `fix(qwp):`/`test(qwp):` prefixes (CLAUDE.md prefers plain-English commit titles), but PRs squash-merge and the branch history is throwaway, so this does not matter.

## Summary

- **Verdict: approve.** Correctness & performance gate: satisfied — zero confirmed correctness, data-integrity, resource, concurrency, or performance/IO findings on any reachable path. Test gate: satisfied — no UNTESTED Critical row; the fix ships with strong regression tests carrying verified failure links (the tightened `assertCoercionError` suite + ingress classifier tests + oracle fuzz).
- Coverage map: **16 behavioral changes, 12 tested, 2 UNTESTED (Moderate), 2 EXEMPT.** The two UNTESTED sites (`invalid column name`, string→`TIMESTAMP_NS` overflow) are reachable but their worst-case unguarded regression is a slower terminal failure, not data loss; recommend adding the two `assertCoercionError` cases in this PR to match the guard standard the change itself establishes.
- Tradeoff to acknowledge (not a defect): three config/property refusals (non-WAL, new-columns-not-allowed, designated-ts) are now terminal, so an operator fix + byte-identical replay that would once have succeeded after buffering now drops on the client. This is the intended fail-fast-vs-infinite-silent-retry trade and is tested for the non-WAL case; a one-line javadoc/PR-body note would make the "permanence under current server config" scope explicit.
- Draft findings: 3 agents run; 6 candidate items verified, 2 dropped as false positives (unreachable designated-ts and dead float→decimal catch), leaving 2 Moderate + 4 Minor. Split: all findings in-diff; 0 out-of-diff; 0 pre-existing.

PR #7360, opened 2026-07-07
chore(core): fix catalogue queries omitting tables after a concurrent metadata cache clear

RaphDal:
## Summary

`tables()`, `all_tables()`, `information_schema.columns` and `pg_catalog.pg_attribute` read the metadata cache, which `hydrateAllTables()` reconciles against the table registry before each catalogue read. That reconcile has a give-up budget: after `MAX_INCOMPLETE_RECONCILE_PASSES` (8) consecutive passes that make no progress, it latches `cacheComplete=true` and stops retrying, so a table whose `_meta` is genuinely unreadable is not re-read (and CRITICAL-logged) on every catalogue query forever. That budget was added in #7259.

The budget miscounts a pass as "no progress" when a concurrent `clearCache()` wipes tables the pass had just hydrated. Those tables are perfectly hydratable; only the clear made them look missing. Under a burst of `clearCache()` racing a reconcile storm, several reconcilers can each observe this spurious "no progress" and drive the shared budget to exhaustion, latching `cacheComplete=true` over a cache that a clear merely emptied. Catalogue queries then short-circuit on the completeness flag and return an incomplete table list until the next `clearCache()` or until a writer touches the missing table.

## Root cause

The give-up path decides "no progress" by re-reading the map (`is any missing table currently present?`) rather than by whether the pass actually hydrated a table. A `clearCache()` between a reconciler's hydration and its give-up check makes a hydrated table read as missing. Because the budget counter is shared and a single clear resets it to zero, enough reconcilers arriving at the give-up check just after one clear can bump the counter from zero to the threshold in one burst, latching completeness over the emptied cache. The write lock the give-up path holds does not prevent this — the clear that causes the miscount happens before that lock is taken.

This only bites when `clearCache()` runs frequently enough to overlap reconciles, so it is rare and transient (the cache self-heals on the next clear-free reconcile or writer touch), but within that window the catalogue functions under-report.

## Fix

`clearCache()` bumps a `clearEpoch`. `hydrateAllTables()` snapshots the epoch at the start of a reconcile and, in the give-up path, skips spending the budget when the epoch has moved — a concurrent clear, not an unreadable `_meta`, produced the "no progress", so it must not count toward giving up. A reconcile with no overlapping clear is unchanged: the counter still climbs and the give-up still fires after 8 clear-free stuck passes.

So the regression test can force the exact interleaving deterministically (instead of relying on a probabilistic stress loop), the change extracts the give-up assessment out of `hydrateAllTables()` into a protected `assessReconcileGiveUp(missing, startEpoch)` method and routes every completeness publish through a protected `publishCacheComplete()`. The test subclasses `MetadataCache` on a test-local instance and wraps those methods (park at a barrier, then `super`); production carries no test-hook state. The same pattern replaces the `@TestOnly` `setLatchCompleteTestHook` API that #7259 added — the latch race test now overrides `publishCacheComplete()` — so both `Runnable` hook fields and both public setters are gone.

The PR also removes leftover `M2:`/`M3:` planning-label prefixes from six pre-existing `MetadataCacheTest` comments (comment-only cleanup, bundled here because the PR already touches the file).

## Tradeoffs

- Under sustained, high-frequency `clearCache()` (every reconcile overlaps a clear), the give-up for a genuinely unreadable `_meta` is delayed rather than firing at 8 passes, which weakens the CRITICAL-log / re-read bound #7259 added in that regime. `clearCache()` already resets the budget on every call, so this regime was already degraded; in normal operation clears are infrequent and the give-up fires as before.
- Two `MetadataCache` methods become protected (`assessReconcileGiveUp`, `publishCacheComplete`) so tests can subclass instead of production carrying runnable hook fields. This removes the `setLatchCompleteTestHook` public API from #7259; anything that adopted that hook needs the same subclass migration.
- No measurable runtime cost: the epoch snapshot is one volatile read on the reconcile slow path; the warm fast path (`if (cacheComplete) return`) is untouched, so catalogue queries against a warmed cache are unaffected.

## Test plan

- Adds `testGiveUpBudgetIgnoresRoundsThatOverlapAClearCache`: a `MetadataCache` subclass parks 8 reconcilers right before the give-up assessment, a single `clearCache()` fires, then all 8 proceed. Without the guard the 8 post-clear bumps exhaust the budget and the test's `assertFalse(isCacheComplete())` fails ("give-up latched cacheComplete over a clear-wiped cache"); with the guard every bump is skipped and the flag stays off. The test then asserts that a clear-free reconcile still warm-latches, so the guard does not suppress legitimate latching. Deterministic — verified failing on the unguarded build and passing 30/30 on the guarded build.
- `testCompletenessLatchIsMutuallyExclusiveWithClearCache` (from #7259) migrates from the removed hook API to a `publishCacheComplete()` override on a test-local instance and still passes; because the override wraps the actual publish, it keeps exposing a regression that moved the publish outside its lock.
- The existing give-up tests `testReconcileGivesUpAfterRepeatedHydrationFailures` and `testReconcileBudgetCountsOnlyZeroProgressRounds` still pass: with no overlapping clear the budget behaves exactly as before, so the give-up still fires after 8 stuck passes.
- The existing probabilistic test `testConcurrentReconcileAndClearCacheNeverMarksEmptyCacheComplete` (added in #7259, which surfaced this on arm64 CI) still passes and is no longer flaky.
- Full `MetadataCacheTest`: 40 tests, all green.


RaphDal:
# Review of PR #7360 — `chore(core): fix catalogue queries omitting tables after a concurrent metadata cache clear`

- **Review level:** 3 (full pass: change-surface map, 10 parallel review agents, per-finding verification)
- **Date:** 2026-07-07
- **Branch reviewed:** `rd/fix-metadatacache-clear-giveup` (`cef86d659b`, one commit ahead of master)
- **Verdict:** **Approve.** No Critical findings; two Moderate (documentation precision, test diagnosability), the rest Minor.

## Verified soundness of the core fix

The fix is correct, the race it claims existed is real (reproduced analytically against `HEAD~1`), and the
epoch guard provably closes it. Three independent adversarial passes (correctness, concurrency,
fresh-context) enumerated the clear-vs-pass interleaving space and none found a way to latch
`cacheComplete` over a wiped-but-hydratable cache after this diff. Both performance passes confirm zero
warm-path cost (the epoch read sits after the volatile fast-path return) and the publish path is strictly
cheaper than the removed hook. `MetadataCacheTest` ran locally: **40/40 green**, confirming the
test-plan claim.

Key facts, all grep-/trace-verified:

- `clearEpoch` is written only under the write lock and read under it in the assessment, so any clear that
  can manufacture "no progress" (it must wipe hydrations that happened after the snapshot at
  `MetadataCache.java:199`) necessarily bumps the epoch past `startEpoch` and is caught at `:329`. A clear
  before the snapshot is fully visible to the scan, so zero progress then means genuinely failed
  hydrations. The guard only ever skips a spend — it adds no publish path.
- `publishCacheComplete()` is the single `cacheComplete = true` site; all three callers (`:227` read lock,
  `:413` via read-lock callers `:154`/`:270`, `:336` write lock) hold a lock excluding `clearCache()`.
- The new test is deterministic in both directions: all 8 reconcilers provably reach the barrier (`stuck`
  keeps every pass incomplete; no latch site is reachable pre-barrier), and on the unguarded build the 8
  post-clear bumps deterministically latch and fail `assertFalse`.
- Cross-context: `hydrateAllTables()` is reachable only via `MetadataCache.snapshot()` (the four catalogue
  function factories); `clearCache()` has exactly one OSS production caller — `CairoEngine.clear()`,
  which is `@TestOnly` — so production give-up/latch behavior is bit-identical to master. The removed
  `setLatchCompleteTestHook` API has zero remaining references repo-wide. All 17 callsite groups
  (catalogue factories, startup hydrator, DDL/writer paths, SHOW/plan-generation hydrates, embedded API)
  returned SAFE.

## Critical

None.

## Moderate

### M1. New invariant documentation is imprecise in three places (in-diff)

- `MetadataCache.java:298-303` — the javadoc bullet says a clear-overlapped pass "is ignored". The code
  only epoch-guards the **spend**: an overlapped pass that observed progress still takes the `progressed`
  branch at `:324-327` and resets the counter — wiping bumps legitimately accumulated by newer
  epoch-stable stuck rounds (found independently by three agents). The direction is safe (delays give-up,
  never wrong-latches, and the reset evidence is genuinely current-epoch), so the code is fine — but the
  doc should say "does not spend the budget" rather than "is ignored", or state the reset precedence
  explicitly. No test pins the `progressed && epoch-moved` precedence either way.
- `MetadataCache.java:62-74` — the `MAX_INCOMPLETE_RECONCILE_PASSES` header comment still defines the
  counting rule as "a pass that still finds a table missing yet hydrates none" and lists only the progress
  exemption. The diff added a second exemption (clear overlap) that this authoritative comment doesn't
  know about. Add three words or point at `assessReconcileGiveUp`.
- `MetadataCache.java:428-436` — the `publishCacheComplete()` javadoc explains the lock scope but never
  warns overriders: an override runs inside a **non-reentrant** `SimpleReadWriteLock` scope, so acquiring
  the cache's write lock from the override self-deadlocks, and any blocking stalls all metadata writers.
  One sentence ("overrides must not acquire cache locks or block") closes the trap the two test overrides
  deliberately dance around.

### M2. The new regression test's failure modes for violated assumptions are a hang or a masked error, not a red assert (in-diff)

`MetadataCacheTest.java:1386-1394, 1413, 1431`:

- On barrier timeout/breakage the override records `error` and still falls through to
  `super.assessReconcileGiveUp()`. If a future change moved the assess call under a lock (the exact
  contract the test guards), the parked readers plus the barrier action's `writeLock()` produce a
  permanent spin-hang rescued only by the 20-minute class timeout. Throwing after a broken/timed-out
  barrier instead of calling `super` fails fast with the right cause.
- The final self-heal `hydrateAllTables()` at `:1431` runs after the only `error` re-check (`:1413`). The
  barrier auto-resets, so if that pass ever reached the assessment (environmental hydration failure), the
  main thread parks 30 s, stores a `TimeoutException` nobody reads, and the test fails later on
  `assertTrue` with the root cause hidden. Re-check `error` after `:1431` (or make the override one-shot).

## Minor

### Production style

1. `MetadataCache.java:100-102, 196-198, 298-303, 989-991` — the clear-epoch rationale is stated four
   times; only `:330` correctly defers with "see javadoc". Keep the javadoc bullet authoritative, trim the
   other three to one line each. Relatedly the `assessReconcileGiveUp` javadoc is ~32 lines of
   PR-body-depth rationale; ~12 lines covers the contract (deep rationale lives in the PR body).
2. `MetadataCache.java:314, 437` — the two new protected methods are not adjacent; members are supposed to
   group by visibility and sort alphabetically. The file was already disordered (pre-existing), so
   cosmetic.
3. `MetadataCache.java:315, 317` — `gaveUp`/`progressed` violate the `is`/`has` boolean-prefix rule.
   Carried verbatim from the old block, but the extraction re-created the lines — natural moment for
   `hasGivenUp`/`hasProgressed`.
4. `MetadataCache.java:314` — `missing` could take `@NotNull` since the method is now an overridable seam;
   note the file's other `ObjList` param (`latchCacheCompleteIfWarmed:402`) is also unannotated, so this
   is optional consistency, not a convention break.

### Tests

5. Coverage gap: no test pins "a clear **between** counted passes grants a fresh budget of 8" (the
   javadoc's own safety argument at `:302`). Existing tests only clear at counter=0. A sequential test
   (3 stuck passes, then `clearCache()`, then 7 passes assert-unlatched, then the 8th latches) needs no
   seam.
6. Coverage gap: nothing asserts the user-visible symptom — `tables()` output — around the give-up state
   (stuck table omitted after budget exhaustion, restored after clear). A sequential engine-cache drive
   plus `assertQuery(...).returns(...)` covers it.
7. `MetadataCacheTest.java:1367` — `reconcilers` names an `int` here but a `Thread[]` in the sibling
   hammer test (`:1310`); rename to `reconcilerCount`. And `:1358` — the local `8` mirrors the private
   production constant without the "Mirror of MetadataCache.MAX_INCOMPLETE_RECONCILE_PASSES" pointer that
   `:1081` carries; if the constant drifts, this test silently loses its bug-detection power (sibling
   tests fail loudly on drift, which mitigates).
8. `MetadataCacheTest.java:1372` — the `_meta` move sits three statements outside the `try` whose
   `finally` restores it; only a constructor OOME can strand the file (and the per-test db root recreation
   makes even that harmless), but starting the `try` right after the move costs nothing.

### PR body / metadata (edits go to `PR.md`; the PR itself has not been touched)

9. Factual error: "so **both** `Runnable` hook fields and **both** public setters are gone" — verified
   against `HEAD~1`: there was exactly one field and one setter repo-wide. Say "the `Runnable` hook field
   and its public setter are gone".
10. Active-voice rule: four "is/are + participle" constructions ("That budget was added in #7259",
    "before that lock is taken", "the give-up ... is delayed", "is not re-read"). Also "So the regression
    test can force..." garden-paths — "So **that** the regression test...". And "No measurable runtime
    cost" is an analytical estimate, not a measurement — "negligible expected runtime cost".
11. The test-plan claims ("verified failing on the unguarded build", "passing 30/30") are author-run
    results — attribute them as such. (The 40-test green claim was reproduced locally during this
    review.)
12. Worth one disclosure sentence: the only in-tree `clearCache()` caller is the `@TestOnly`
    `CairoEngine.clear()`, so the race surfaces via test harnesses and the public embedded
    `MetadataCacheWriter` API rather than a stock server path. The `regression` label is nonetheless
    justified — released tag 9.4.3 contains #7259. Consider adding the `SQL` label (the symptom is SQL
    catalogue functions; #7259 carries it).
13. Branch commit title uses a Conventional-Commits prefix, exceeds 50 chars, and has no body — throwaway
    under squash-merge, so informational only.

## Pre-existing, out-of-diff (context, not blockers — candidates for a follow-up issue)

- **Rename racing the reconcile can poison the cache with a stale old-name entry.** Scan collects old
  token Y in `missing`; `RENAME Y->Z` lands (cache `renameTable:1137` no-ops since Y is uncached); pass 2
  then hydrates the old token from the unchanged WAL dir and `tableMap` gains a Y the registry no longer
  knows — served by `snapshot()` until the next clear/drop; with the reused token snapshot the pass can
  even latch while Z is uncached. Hunk ranges confirm this code is untouched by the diff; the race exists
  identically on master and in `onStartupAsyncHydrator`.
- **`hydrateTableOnDemand` has no budget of its own**: for an unreadable `_meta` it re-reads disk and
  CRITICAL-logs on every SHOW/plan-generation invocation until some catalogue latch fires — pre-existing,
  unchanged, and worth remembering when reasoning about the give-up's log-bounding purpose.

## Downgraded (dismissed after verification)

- "Every other collection param is `@NotNull`-annotated" (premise of the annotation finding) — false:
  `latchCacheCompleteIfWarmed:402` is unannotated; reframed as optional (Minor 4).
- "Failed `_meta` restore can mask the in-flight failure" — verified unreachable including on Windows: all
  threads joined before the `finally`, no fd ever opened on the hidden file.
- "OOME during thread start-up races `close()`" — OOME-only trigger; not actionable.
- "Fast-path TOCTOU gives one transiently-empty catalogue read after a clear" — pre-existing inherent
  check-then-act, self-heals on the next query; not this PR.
- "Concurrent zero-progress passes exhaust the budget in one wall-clock burst" — pre-existing semantics;
  each bump is one genuine failed hydration attempt, and the "sequential rounds" comment wording is
  carried from master with the same progress-storm scoping.
- "Epoch check should precede the progressed loop for performance" — the saved lookups occur only on a
  production-unreachable branch (`clearCache` is `@TestOnly`-reachable), and the current order errs toward
  retrying, the safe direction.
- "Give-up starvation under sustained clears is a new regression" — pre-existing: every clear already
  reset the counter, so no caller loses a bound it actually had; the PR's Tradeoffs section discloses this
  accurately.

## Summary

Approve — no Critical findings; the two Moderates are documentation-precision and test-diagnosability
hardening, both cheap. The design tradeoffs (give-up deferral under clear-heavy regimes, one public
`@TestOnly` API removed, two methods opened to `protected`) are real but honestly disclosed in the PR body
and verified as strictly-better than the pre-fix behavior (production give-up timing is bit-identical
since `clearCache()` never runs in an OSS server).

From ~41 raw agent claims, 24 distinct draft findings survived dedup; **17 verified** (0 Critical,
2 Moderate, 15 Minor including PR-body items), **7 downgraded** as
false-positive/unreachable/pre-existing-mislabeled. In-diff vs out-of-diff: 15 in-diff, 2 out-of-diff
pre-existing races reported as context — the cross-context pass walked all 17 callsite groups and
returned SAFE on every one.

bluestreak01 (approved):
# Review of PR #7360 — `chore(core): fix catalogue queries omitting tables after a concurrent metadata cache clear`

**Review level:** 3 (full pass — change-surface map, all agent roles, per-finding verification, plus empirical test run)
**Branch:** `rd/fix-metadatacache-clear-giveup`

## What the change does

`hydrateAllTables()` reconciles the metadata cache against the table registry before catalogue reads and has a give-up budget (`MAX_INCOMPLETE_RECONCILE_PASSES = 8`) so a genuinely-unreadable `_meta` isn't re-read + CRITICAL-logged on every query forever. A regression from #7259: a `clearCache()` landing between a reconciler's hydration and its give-up check makes a just-hydrated table read as "missing"; because `clearCache()` resets the shared counter to 0, a burst of reconcilers each bump it 1..8 in one go and latch `cacheComplete` over an emptied-but-hydratable cache.

The fix adds a `volatile long clearEpoch`, bumped under the write lock in `clearCache()`. `hydrateAllTables()` snapshots it before scanning; `assessReconcileGiveUp()` skips spending the budget when the epoch moved (a clear, not an unreadable `_meta`, caused the "no progress"). It also funnels all `cacheComplete = true` writes through `publishCacheComplete()` and replaces the `@TestOnly setLatchCompleteTestHook` API with test subclassing of two now-`protected` methods.

## Critical

None.

## Moderate

None.

## Minor

1. **Title type `chore` vs `fix` — pre-merge judgment call (metadata only, non-blocking).** The title is `chore(core): ...` while the labels are `Bug` + `regression` and the body says "fix ... omitting tables". Per CLAUDE.md the Conventional-Commit type feeds release notes, and a bug fix conventionally lands under `fix`. **However**, I verified that `MetadataCacheWriter.clearCache()`'s only production caller is `CairoEngine.clear()` (CairoEngine.java:612), which is `@TestOnly` — so in this repo the raced scenario is not reachable by any production path; the change hardens an invariant and de-flakes an arm64 CI test. On that basis `chore` is defensible. This is a metadata reconciliation between title/labels, not a correctness issue — reconcile it either way (label to match `chore`, or retitle to `fix`), but it does not block.

2. **Test restore-on-exception window (`MetadataCacheTest.java:1399`, test-robustness, behaviorally inert).** `Files.move(stuckMeta → stuckHidden)` runs *before* the try-with-resources and the inner `try/finally` that restores it. If the anonymous `new MetadataCache(engine){...}` construction (line 1400) threw, `stuck/_meta.hidden` would be left renamed. The constructor is trivial (assigns `engine`, inits the barrier), so this is essentially unreachable, and `assertMemoryLeak` scopes each test to a fresh temp dir so there's no cross-test contamination. Optional tightening: move the rename inside the `try`.

3. **Hardcoded `maxIncompleteReconcilePasses = 8` in the test (`MetadataCacheTest.java:1384`) duplicates the private `MAX_INCOMPLETE_RECONCILE_PASSES`.** Unavoidable coupling — the production constant is `private`, so the test cannot reference it (the two existing give-up tests do the same). Note only.

## Downgraded (false positives / notes verified as non-issues)

- **"Other eviction paths (`dropTable`/`renameTable`/silent hydrate-failure) manufacture spurious no-progress not covered by the epoch guard"** — dismissed. Traced: those paths reflect a *genuinely* dropped/absent table (`getTableTokens()` and `hydrateTableStartup`'s `engine.isTableDropped` early-return agree), so a latch that omits them is correct, not the emptied-but-hydratable hazard the guard targets. No burst-reset like `clearCache()`. No change needed.
- **"Guard permanently suppresses legitimate give-up"** — dismissed. With no overlapping clear, `clearEpoch == startEpoch`, the budget is spent normally and give-up fires after 8 clear-free stuck passes (preserved by `testReconcileGivesUpAfterRepeatedHydrationFailures`). The sustained-overlapping-clear degradation is the documented tradeoff and already existed (clearCache resets the counter regardless).
- **"Redundant `if (cacheComplete) return;` before `assessReconcileGiveUp` is racy"** — dismissed. Idempotent: if another thread latches in the gap, `assessReconcileGiveUp` either resets on progress or re-sets an already-true flag; `publishCacheComplete()` is idempotent. Harmless.
- **`publishCacheComplete()` routing** — verified complete: `rg` confirms the only `cacheComplete = true` write is inside `publishCacheComplete()` (line 438); no bypass anywhere in `core/src/main/java`.
- **`startEpoch` snapshot outside the lock** — verified correct and necessarily `volatile` (non-torn 64-bit read + happens-before with the `clearCache()` write). Snapshotting before the read-lock scan is strictly more conservative (catches scan-racing clears), never misses one. `!=` ≡ `>` on a monotonic long; no ABA.

## Coverage map

| # | Behavioral change | Test | Failure link | Verdict |
|---|---|---|---|---|
| 1 | `clearCache()` bumps `clearEpoch++` (:990) | `testGiveUpBudgetIgnoresRoundsThatOverlapAClearCache` | Revert → epoch stays 0 → `1!=0` false → else-branch → 8 bumps latch → `assertFalse(isCacheComplete())` fails | TESTED |
| 2 | `hydrateAllTables()` snapshots `startEpoch` (:198) | same | same interleaving; without the snapshot the guard has no baseline | TESTED |
| 3 | `assessReconcileGiveUp()` skip branch `else if (clearEpoch != startEpoch)` (:326) | same | With guard all 8 skip (flag off); reverted, all 8 bump and the 8th latches → assert fails | TESTED |
| 4 | give-up budget bump + latch, progress reset (preserved) | `testReconcileBudgetCountsOnlyZeroProgressRounds`, `testReconcileGivesUpAfterRepeatedHydrationFailures` | 3 progress rounds don't spend budget; 8 clear-free stuck rounds latch (`assertTrue`) | TESTED |
| 5 | all `cacheComplete=true` routed via `publishCacheComplete()` (:437) | `testCompletenessLatchIsMutuallyExclusiveWithClearCache` (override), `testReconcileLatchesCompleteAfterHydratingMissingTables` | Override wraps the publish inside its lock; a publish moved outside the read lock → clear empties cache → `assertEquals(tableCount, 0)` fails | TESTED |
| 6 | guard does not suppress legitimate warm-latch | `testGiveUpBudgetIgnoresRoundsThatOverlapAClearCache` tail | After stuck restored, self-heal `assertEquals(2, tableCount)` + `assertTrue(isCacheComplete())` | TESTED |
| — | removal of `setLatchCompleteTestHook` / `latchCompleteTestHook` | migrated tests compile+pass | no-behavioral-change: test-only API replaced by subclassing (`rg` finds zero residual refs) | EXEMPT |
| — | M2/M3 comment-prefix cleanup (6 comments) | n/a | verified comment-only, no code change | EXEMPT |

**Empirical:** the affected suite (7 reconcile/clear tests incl. the new one) was run green on the guarded build. Revert efficacy confirmed by trace (both halves of the hunk are pinned).

## Summary

- **Verdict: approve.**
- **Correctness & performance gate:** passed — zero confirmed correctness, concurrency, resource, or performance/IO findings. The `clearEpoch` guard is sound on every interleaving I could construct; it can only ever set `gaveUp=false`, never latch; the sole latch branch runs under the write lock with `clearEpoch == startEpoch`, and no cache wipe is possible without moving the epoch. Warm fast path (`if (cacheComplete) return;`) is untouched — the one added `volatile` read sits after the fast-path return, on the reconcile slow path only; no new allocation, no boxing, complexity identical.
- **Test gate:** passed — this is a fix with a deterministic regression test whose assertion demonstrably fails if either half of the production hunk is reverted; coverage map is 6 behavioral changes, 6 tested, 0 UNTESTED (2 exempt: test-API removal + comment cleanup).
- **Tradeoffs/context:** the guarded scenario is only reachable via `@TestOnly` code in this repo, so the production catalogue-query impact is nil; the fix's real value is hardening the invariant and de-flaking the #7259 arm64 test. Documented tradeoff (delayed give-up under sustained overlapping clears) already existed pre-fix.
- **Findings:** 5 agent roles + verification; 0 correctness/concurrency/perf/test blockers verified, ~4 draft notes dismissed as non-issues, 3 Minor items (all metadata/test-robustness, none blocking).
- **Split:** 2 in-diff behavioral changes (epoch bump + guard), 1 in-diff refactor (publish choke point), 0 out-of-diff, 0 pre-existing bugs. Cross-context pass confirmed the only external caller is `@TestOnly`, so zero out-of-diff findings is expected here, not an under-run.

The three Minor items are optional; none blocks merge. This meets the bar for approval.

mtopolnik:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/MetadataCache.java|10|10|100.00%|

PR #1109, opened 2021-06-09
Iodispatcher limit

patrickSpaceSurfer:
Change IODispatcher such that back pressure is put on the listen queue when the maximum active connection limit is reached

PR #7361, opened 2026-07-07
fix(core): signal restore cancellation promptly on shutdown

jovfer:
**Tandem:** [questdb/questdb-enterprise#1109](https://github.com/questdb/questdb-enterprise/pull/1109)

Small OSS pieces the tandem Enterprise PR builds its close/switch-race and demote data-safety
fixes on: a pre-join lifecycle cancel hook, a private-ownership materialized-view hydrate contract,
a non-creating WAL-suspension tracker lookup, and a genuine `LogCapture.assertOnlyOnce` (which
exposed a real double-logging bug on the way).

## Pre-join lifecycle cancel hook

`LifecycleOrchestrator.close()` bounded-joins the boot thread (30s budget) before the reverse-topo
stop loop, but restore cancellation was signalled only from a component's `stop()`, which runs
*inside* that later loop. A SIGTERM landing during a long point-in-time restore therefore burned
the whole join budget before the restore was even asked to unwind — on a pod with a 30-second
termination grace, the node was SIGKILLed mid-restore.

`close()` now runs a `preJoinCancelHook` (installed via `setPreJoinCancelHook`, mirroring the
existing `preStopHook` seam) after the executor drain and before the boot join, so the join budget
is spent unwinding the restore instead of waiting it out. The hook contract is a pure,
non-blocking cancel signal; resource teardown stays in `stop()`. The enterprise overlay installs
the restore-cancel signaller through this seam; `stop()` keeps a second-line signal for the case
where a restore starts after boot. A throwing hook is absorbed — logged, not propagated — at both
the pre-join and pre-stop sites, so a misbehaving hook cannot abort the shutdown it was installed
to accelerate: the reverse-topo stop loop still runs either way.

## Private-ownership materialized-view hydration

`hydrateMatViewStateStore` now takes the target `MatViewStateStore` as a parameter (the no-arg
method stays as a delegator, hydrating the installed field). This is the OSS half of the
enterprise private-ownership promote: the caller hydrates a private, not-yet-installed store and
publishes it into the engine only after hydration completes and a closing check passes, so an
expired close rendezvous can never free the store a hydration loop is still writing into. Two
`isClosing()` aborts guard the load: one at entry, before the loop's first read of engine state
(`getTableTokens`), and one inside the per-token loop, so a close landing mid-walk aborts promptly
without waiting for the whole view set to load. Both are load-bearing rather than belt-and-braces:
the entry check is the only one that can fire before any per-token iteration exists, and the
per-token poll is the only thing that catches a close landing strictly between two views. The
target-taking overload throws an explicit `NullPointerException` on a null target, chosen over
letting the per-view `catch (Throwable)` further down silently swallow it.

`MatViewStateStoreImpl.close()` is now idempotent across sequential closes. It used to free every
`MatViewState` in `stateByTableDirName` and leave the map populated — the clearing lived only in
the `@TestOnly clear()` helper — so a second, later `close()` call (e.g. a promote unwind freeing a
never-installed private store, followed by engine teardown freeing whatever ended up installed)
walked the same entries again and double-freed them. `close()` now clears the map after the free
loop. This covers sequential closes only; the method is not synchronized, so a genuinely concurrent
double-close is still out of contract.

## Non-creating suspension-tracker lookup

`TableSequencerAPI.getTxnTrackerIfExists(TableToken)` returns an already-installed tracker without
allocating/installing one on a miss (unlike the existing `getTxnTracker`), for hot read paths — a
metrics scrape being the motivating one — where "no tracker yet" simply means "never suspended".
`CairoEngine.isWalApplySuspended(TableToken, SeqTxnTracker)` takes an already-resolved tracker so a
caller holding one from the accessor above doesn't pay for a second lookup; `null` means "no
tracker installed", which cannot be hard-suspended by construction, so only the config-list
suspension check still applies. The enterprise tandem PR uses this pair to hoist a tracker lookup
out of its own per-table demote-audit scrape loop.

## A genuine `LogCapture.assertOnlyOnce`, and the bug it had been hiding

`LogCapture.assertOnlyOnce(String regex)` was checking `matcher.groupCount() == 0` after a single
`find()` — `groupCount()` is the *regex's own capture-group count*, a static property of the
pattern, not a count of matches. For any regex with no capturing groups (the common case), this
assertion always passed regardless of how many times the message actually appeared in the log; the
method never detected a genuine duplicate. It's fixed to call `find()` twice and assert the second
call returns `false`. `LogCapture.waitFor` also gained a `(value, timeoutMs)` overload with an
injectable clock/sleeper for deterministic test control over the wait boundary.

`waitForRegex` carried a boundary defect of the same family: its loop exited once elapsed time
reached the budget, but it only threw once elapsed time *exceeded* it, so a wait that timed out
exactly on the deadline returned silently and read as a successful match. `SqlLoggingTest` — the
suite the completion-logging fix below targets — is one of its callers. It now shares `waitFor`'s
shape (deadline checked before the sleep, `>=` treated as timed out) and the injectable clock.

Making the assertion real immediately exposed a genuine production bug it had been silently
missing: every `CREATE TABLE` (and system-table creation at boot) was logging two `fin` completion
records — one premature, sentinel-id (`id=-1`) record from `SqlCompilerImpl.compileInner()`'s
generic post-executor log firing even though the `CREATE`-keyword executor had bailed out and left
`compiledQuery` at `CompiledQuery.NONE` (the existing signal that the statement still needed to
fall through to the model-compile-and-execute path), and one correct, real-registry-id record from
the actual execution site. The fix gates the generic log on
`compiledQuery.getType() != CompiledQuery.NONE` — the same condition the fall-through check 15
lines below already uses to decide whether to fall back to model compilation — so the premature
record is suppressed and the real one survives untouched, carrying a real registry id rather than
the old `-1` sentinel. (The Enterprise tandem PR's `COPY` handling has the identical fall-through
shape and picks up an equivalent fix automatically, since both share the same guarded call site.)

**One consequence is a deliberate behavior change, not just de-duplication:** completion logging
for these statements now belongs entirely to the execution site, so a `CREATE` that is compiled
but never executed — the compile hands back an `Operation` the caller then discards — emits no
completion record at all, where before it emitted the premature `id=-1` one. That record was
always wrong (it announced the completion of something that had not run), but anyone counting
`fin` records per compiled `CREATE` will see the count drop rather than merely deduplicate.

**A bug this PR itself introduced, found while root-causing an unrelated tandem-PR test failure, is
fixed here.** `CairoEngine.loadMatViewIntoStore`'s "no persisted state" branch (a materialized view
that has never refreshed) enqueued its initial kickstart refresh on the engine's own
`matViewStateStore` field instead of the method's `target` parameter — during a promote-hydration
rehydrate those are different objects, so the kickstart was silently scheduled on a store that
never gets installed and the view never converged.

It is worth being precise about where this came from, because it is not a `master` bug. On
`master`, `loadMatViewIntoStore` takes no `target` parameter at all: the engine's
`matViewStateStore` field *is* the hydration target there, so every call in the method is correct
by construction. This PR is what introduces the parameter and the private-ownership contract that
goes with it, converting every site that existed at the time. The kickstart in the "no persisted
state" branch was added independently and later, by the live-views work, written — correctly for
its own tree — against the field. Merging that upstream work into this branch is what put a
field-based call inside a method that now hydrates into a caller-supplied store. Neither change is
wrong on its own; the combination is.

**This makes the refactor fragile against future merges, and reviewers should treat it that way:**
any upstream commit that adds a `matViewStateStore` use inside `loadMatViewIntoStore` will silently
re-break the target-ownership contract in exactly this way, and the resulting bug is invisible
outside a promote-hydration rehydrate — the view simply never converges, with nothing logged.
`MatViewHydrateTargetTest#neverRefreshedViewKickstartsOnDistinctTarget` is the standing guard
against that recurrence. The fix itself points the call at `target`, mirroring the already-correct
sibling 31 lines below.

## Tests

One note on approach rather than a tally; CI is the evidence that the suite passes. The lifecycle
and hydration coverage here is deterministic rather than timing-based — waits park on an explicit
state seam (a cancel flag, a drain-boundary latch, a closing flag) and are released by the test,
replacing fixed sleeps — and the boundary fixes in `LogCapture` are pinned in both directions, so
an exact-deadline timeout throws while an already-matching regex still returns.

## Diff size

18 files changed, +1142/−90.


jovfer:
# Review: PR #1109 (ENT) + #7361 (OSS) — L3 tandem (r202)

**Level 3, tandem mode.** Heads: ENT `52f9dc662` / OSS `ec22eba8e6`.

This PR was reviewed from scratch earlier the same day; the only production change since is the fix wave `a809b0306` (the accepted fixes from that round) plus a test-trim. I ran a 5-agent adversarial fleet (Rust, Java+tandem, concurrency, test-trim, fresh-context) and verified every finding against source. **12 findings verified, 2 dropped as false positives.**

**Submodule provenance:** `questdb` (OSS `b06dd711 → ec22eba8e6`) = OFF-DEFAULT → in scope (the tandem half); `java-questdb-client` = no pointer move → out of scope.

## Headline result

**The close-protocol redesign is sound, and the earlier close-window Critical backlog is refuted.** Prior rounds found a cluster of `close()`-chokepoint Criticals plus Rust dirty-latch / lifetime-fence Criticals. This PR **added and then removed** that entire machinery (`555228e1b` / `fbbde860b`), so the Rust teardown surface is now byte-for-byte the merge base, and the Java `close()` uses a two-sided admission counter + 7 fence spans — all verified enter-outside-try / finally-exit, with `closing` published before the counter read and the bounded 5 s wait before every free. REPLICA heal tail, ACL spans, resync entry, and `quiesceMatViewRefresh` are all correctly wrapped. Two-sided admission is a correct Dekker pair; `preJoinCancelHook` is non-blocking + race-safe; the private-store hydrate has no UAF.

**CI at HEAD (fresh build): green on all test jobs** (lifecycle/acl/other, replication S3·GCS·NFS·Azure·S3Mock, e2e, Rust-lint). `windows-amd64-lifecycle` still building. The only fail is `check_oss_submodule (oss is mergeable)` = OSS master advanced past the PR's pin — merge mechanics, out of scope (resolves via master-merge + pin bump at merge).

## Critical

**None.**

## Moderate (worth addressing — none block)

**M1 — Phantom demote blocker on a transient `startupTable` `-1`** *(in-diff)*
`WalUploader.triggerUploadMissed` seeds `notified` unconditionally (`lastTxn`) but `uploaded` only when `startupTableNative >= 0`; the Rust `TxnLogReader::open` failure arm returns a silent `-1` → permanent gap → `RoleSwitchUploadDrainIncompleteException` → node read-only + UNKNOWN, undemotable until heal→PRIMARY+restart. This is the already-adjudicated deferral from the prior round (quick fix rejected — trades fail-closed safety / boot availability; needs a designed fix; B1 was shipped to make it diagnosable). It is the fail-*safe* direction (refuse, don't lose data) and recoverable. Reported so it stays tracked. Suggested designed fix: a distinct "seed-unavailable" sentinel (e.g. `i64::MIN`) so Java skips the `notified` seed rather than manufacturing a backlog.

**M2 — `BackupManager.close()` acquires its write lock unbounded, behind a promote's uploader `create()`** *(in-diff, incomplete-hardening)*
`BackupManager.java:442` `writeLock.lock()` is untimed while the WalUploader ctor holds the read side across the whole native `create()` (object-store retry loop). Confirmed by two agents. The hazard is honestly documented at `:405-424`. Base was **worse** for the same trigger (lock-free `getPtr()` = a UAF), so this PR improved safety; the residual is a documented, shutdown-only stall that leaks OS-reclaimable resources → possible SIGKILL mid-teardown against a slow store. Optional: `writeLock.tryLock(budget)` → the leak-don't-free fallback `close()` already uses, matching every other teardown wait in the PR.

**M3 — Three observability coverage gaps on fix-wave behaviors** *(in-diff)*
- **(a) ENT ACL-enabled CREATE LIVE VIEW fin logging is UNTESTED.** OSS `SqlLoggingTest` covers only the OSS path; `EntCreateLiveViewOperationImpl.getSqlText()` threading (the ACL-enabled production shape, ENT ACL default = true) has no ENT assertion — revert it and nothing fails. Fix: one `LogCapture` + `assertOnlyOnce` in an ENT ACL live-view test.
- **(b) B1's cause-log line (`.$(t)`) is UNTESTED.** The fail-closed refusal *contract* is tested (`PromotePreflightThrowableRefusalTest`), but deleting `.$(t)` fails nothing — the exact diagnostic B1 exists to restore.
- **(c) `begin_table_startup` hoist guard is only probabilistic.** The Rust regression test asserts the cleaner fires, but its fixture doesn't deterministically discriminate the ordering (revert the hoist → green on most runs). A deterministic witness would observe `pending_tables >= 1` at the first dispatch.

**M4 — `SwitchFuzzTest` refusal classification lacks a flush barrier + guaranteed logging** *(in-diff, trim-introduced, test-reliability)*
`loggedAny` → `assertLogged` → `sink.indexOf(...)` is an instantaneous one-shot read of the async log sink; the advisory can still be in the ring, or dropped under a saturated ring (no `LogFactory.enableGuaranteedLogging(EntLifecycleOrchestrator.class)`). Fail-**loud** flake on starved agents, on a class with flake history. Fix: bounded `waitFor(..., 30_000L)` + the guaranteed-logging `@BeforeClass/@AfterClass` pair. (The rest of that rewrite is a genuine strengthening.)

**M5 — Comment/code drift in the new close protocol** *(in-diff)*
`close()`'s invariant comment names only one span exclusion, while `drainWriterPool` / `quiesceLiveViewRefresh` / `ensure*StoresBuilt` sit in no span (behavior == base — never spanned). The OSS hydrate javadoc (`CairoEngine:1206-1215`, mirrored `EntCairoEngine:2918-2922`) still documents the **removed** "in-flight flag" design (now the admission counter). Fix: amend both comments; optionally wrap the unspanned readers.

## Minor

- `EntLifecycleOrchestrator.java:351` — the ownership-refusal arm uses `$()` where its two sibling arms use `$safe()`; a non-ASCII table name renders as mojibake in the one line that names the blocking table. In-diff. One-token fix: `$safe`.
- Rust `uploader/mod.rs:4315-4324` — the excluded-table preflight WARN lost its `local_txn`/`store_txn` orphan detail and now logs unconditionally per excluded table (diagnostic-only; verdict byte-identical).
- Rust `uploader/mod.rs:4949-4953` — `error!` after a failed `throw_new` upcalls into Java and clears the pending exception → on a double failure, a silent `-1` seed with no Java exception. Near-zero frequency.
- F4 — during the backup-teardown window, `CHECKPOINT RELEASE` can fail with a slightly-misleading message or block for the shutdown budget (the sibling status getters have a lock-free closing fast-path that `hasActiveBackupOrClosing()` skips for the false answer). Shutdown-window only.
- F5 — `PrimaryRoleState.close()` nulls `walUploader` inside the try (after `uploader.close()`), while the sibling `ReplicaRoleState.close()` was changed to null in `finally` with a comment claiming a symmetry that isn't there. The double-close it guards against is unreachable; consistency nit.
- F6 — `matViewRefreshSuspended` can latch `true` if `switchLiveViewMachinery(PRIMARY)` throws after `switchMatViewMachinery` set it; the new early-return comment reads as if the flag is a non-issue there. Adjacent; self-heals on the next demote/successful switch.

## Adjacent (pre-existing — do not block)

- 10 `throw_new(...).expect(...)` panic-abort sites remain in `wal/` + `backup/` JNI (7 in touched files); the fix wave fixed exactly one. An `extern "system"` unwind aborts the JVM. Delta vs base: zero; `throw_new` realistically fails only on class-not-found/OOME.
- 18 stale `panic=abort` comments contradict `Cargo.toml` (`panic = 'unwind'`); the PR corrected one. Doc-only.
- `Java_..._WalUploader_destroy` / `WalDownloader_dtor` are not `ffi_panic_guard`-wrapped — unchanged from base.

## Downgraded (false positives / overstated)

- **"Post-restore init falls through and builds a native BackupManager on a closing engine" (claimed Moderate-High): FALSE POSITIVE.** `initBackupInstanceNameRegistryAndManager()` (`EntCairoEngine:908-921`) has no `isClosing()` guard, but it is **byte-identical at base** and runs **unconditionally on every boot** (trigger==null path included). The new `shutdownRacedPublication` abort arm skips only the pointless restore; it does not change whether the init runs on a closing engine (base reached the same init in the same scenario). Not a regression; at most a pre-existing, shutdown-only, join-expiry-narrow missing guard. Optional hardening: have the abort arm `return`, and/or guard the init with `isClosing()`.
- **"Preflight can't reproduce boot's upgrade-path token tolerance": unreachable.** The downloader creates `_replication_sync_id.d` at construction and boot migration bumps `db_sys_version`.
- **"Deleted `UploaderTeardownBudgetBoundedTest` = coverage loss": downgraded to Minor.** The deleted test's dirty-latch machinery is gone; the surviving bounded reap keeps a Rust RED/GREEN unit pair (`join_teardown_budget_tests`) + a strengthened `MultiTableDeadStoreDemoteTest` + no-row-loss integration coverage. Only the TCP-proxy end-to-end *flavour* is lost (the deleted test's own javadoc conceded it couldn't isolate the reap shape).
- Prior-round close-window + Rust dirty-latch/lifetime-fence/`landed_but_unindexed` Criticals — **moot**: that machinery was added and removed within this PR.

## Coverage map (fix-wave delta)

| Change | Test | Verdict |
|---|---|---|
| Rust `begin_table_startup` hoist → cleaner one-shot fires | `startup_open_failure_still_completes_pending_tables_and_signals_cleaner` | TESTED (probabilistic — M3c) |
| B1 promote-preflight fail-closed refusal | `PromotePreflightThrowableRefusalTest:103` | TESTED |
| B1 cause-log line `.$(t)` | — | UNTESTED (M3b, observability) |
| B2 `reason=`+table on drainIncomplete advisory | `DemoteUploadDrainRefusalSwitchTest:266` (`assertLoggedRE reason=.*tables=[token]`) | TESTED |
| B2 `reason=` on uploadPaused advisory | `EntLifecycleOrchestratorTest:533` (role/state; msg via `publishInternal`) | TESTED (partial) |
| registry-status non-creating tracker (HIT) | `ReplicationTest:1138` + `PrimaryRoleStateDrainForDemoteTest:523` | TESTED |
| registry-status MISS-arm dead-block removal | dead code (verified ×3) | EXEMPT |
| CREATE LIVE VIEW fin (OSS) | `SqlLoggingTest:34` (`assertOnlyOnce`, now non-vacuous) | TESTED |
| CREATE LIVE VIEW fin (ENT, ACL, `getSqlText`) | — | UNTESTED (M3a, observability) |
| QueryProgress double-fin gate (OSS) | `SqlLoggingTest` (un-vacuumed `assertOnlyOnce`) + CI green | TESTED |
| hydrate `target` private store | `CloseDuringPromoteHydrateRaceTest` | TESTED |
| Heal text (suspended.tables removal) | `PromotePreflightThrowableRefusalTest:257` | TESTED |

~12 behavioral changes, 9 tested, 1 exempt, ~2–3 UNTESTED (all observability log lines → Moderate, none Critical-rubric).

## Summary

- **Verdict — ENT: approve with comments · OSS: approve.** Both gates pass. Comments worth addressing at/before merge: M3 (the three coverage gaps — cheapest wins), M2 (bound the `BackupManager.close()` acquire), M4 (the `SwitchFuzzTest` flake). None block.
- **Correctness gate:** 0 open Criticals. Severity audit run both ways; prior close-window Criticals re-audited and refuted with evidence (machinery removed).
- **Test gate:** passes. The fix PR's core behaviors have regression tests; the UNTESTED rows are diagnostic log lines (Moderate), not untested data-safety fixes, none Critical-rubric.
- **Findings:** 12 verified, 2 dropped as false positives. **0 Critical, 5 Moderate, 6 Minor, 3 Adjacent.**
- **Splits:** mostly in-diff; a few adjacent (pre-existing panic sites / stale comments). Out-of-diff-breakage: **0** — the diff is well-contained.
- **CI:** green at HEAD except the merge-freshness gate (out of scope).

jovfer:
# Review: PR #1109 (ENT) + #7361 (OSS) — L3 tandem, from scratch

**Level 3, tandem mode.** Heads: ENT `41b04940f` / OSS `674c669287`. Base: ENT `d0224bec7b` / OSS master-ancestor `6610ab113b`. Reviewed independently via 10 fresh-context agents (Rust, role-switch, engine-concurrency, OSS-tandem, perf, resource, tests, test-quality, fresh-adversarial, meta) plus my own source verification of the load-bearing paths. Prior r202/r203 comments were used only as one input; every finding below was re-derived from source.

**Submodule provenance (Step 2.4):**
- `questdb` OSS `b06dd711 -> 674c669287`: **OFF-DEFAULT -> in scope** (the tandem half).
- `java-questdb-client` nested pointer: **unchanged** (`2489b243c`) -> not a pointer move -> out of scope. The ENT `pom.xml` `1.3.8->1.3.9-SNAPSHOT` bump aligns with that unchanged client, which already builds 1.3.9 — the prior r203 client-desync blocker is **resolved** by commit `41b04940f`.

## Headline

**No Critical findings.** This is an unusually well-defended change; the mechanisms I probed hardest were all explicitly handled with efficacious tests. Independent high-confidence verifications:

- **The two-sided close-vs-switch admission is a correct Dekker protocol** — `enterSwitchCriticalSection` does `incrementAndGet()` (full-fence) then reads the `volatile closing`; `close()` does `signalClose()` (sets `volatile closing=true`) then reads the `AtomicInteger`. Both sides use only volatile/atomic accesses -> a single sequentially-consistent order -> no interleaving admits a span after `close()` starts freeing. No UAF. (Verified by me + engine-concurrency agent.)
- **The OSS QueryProgress double-fin gate drops no completion log** — `SqlCompilerImpl:3420` structurally enforces `compiledQuery==NONE <=> falls through to compileUsingModel`; a full enumeration of all 20 OSS + 9 ENT executors confirms the only NONE-on-success paths are side-effect-free bails that each log their own fin downstream. (Me + oss-tandem agent.)
- **Rust panic-safety is strictly improved** — the pre-existing `startupTable` `.expect()` is removed and replaced with an `is_err()`-guarded throw; no new production unwind; the `clean_up()` self-heal on both `run()` early-return arms is correct. (Rust agent.)
- **BackupManager `implLock` fully fences the freed-but-nonzero-handle UAF** the PR set out to close (+ a genuine new `unwatchConfigReload` leak fix); no new leak, double-free, or UAF anywhere in the teardown redesign. (Engine + resource agents.)
- **No data-path performance regression** — `recordNotified` on the commit path is O(1) alloc-free; the O(T) gauge scrape is bounded, off-path, and doesn't box.

## Moderate (worth addressing — none block)

**M1 — Phantom demote-refusal on a transient Rust txnlog-open failure** *(in-diff, incomplete-hardening)*
`WalUploader.triggerUploadMissed` seeds `notified` on `lastTxn>0` but `uploaded` only on the native seed `>=0`; the Rust `startupTable` `TxnLogReader::open` failure arm (`mod.rs:4925-4937`) returns `-1` silently for **both** "file missing" and a **transient I/O error**. So a caught-up, fully-uploaded, quiet table whose local txnlog transiently fails to open during the boot/promote scan seeds `notified` but not `uploaded` -> phantom gap -> `drainForDemote` refuses the demote (node torn read-only/UNKNOWN). This is the **fail-safe** direction (it can only over-refuse, never manufacture a false all-clear that strands data), it is **self-healing** (heal->PRIMARY re-runs the scan; the open succeeds and the gap clears), and I confirmed the **dropped/purged-table variant is unreachable** (`forAllWalTables` returns `lastTxn=-1` for those, and the `lastTxn>0` gate skips the seed). Net vs base: base silently completed the demote (correctly for a fully-uploaded table, but silently losing data on a *real* gap); the PR trades that for a rare, recoverable over-refusal. The clean fix (a distinct "seed-unavailable" sentinel so Java skips the `notified` seed too) reintroduces the data-loss the audit exists to prevent, which is why the deliberate choice is defensible. Optional: distinguish a transient open error from "no index entry" in the Rust arm so only genuinely-absent tables skip the uploaded seed.

**M2 — Promote pre-flight flipped from fail-open to fail-closed (a deliberate data-safety/availability tradeoff worth confirming for operators)** *(in-diff, theme 4 — documented)*
The parity pre-flight now refuses a promote on **any** indeterminate result (store unreachable, unreadable local txnlog/exclusion snapshot) after only `PREFLIGHT_MAX_RETRY_ATTEMPTS=1`. Base failed **open**, letting the promote proceed to the authoritative uploader-boot fence with a full retry budget. Concrete consequence: an emergency failover (dead primary) that coincides with a transient object-store blip now **hard-refuses the promote -> no writable primary** until the store recovers and the operator retries. This is the PR's stated purpose (theme 4: refuse rather than promote on an unproven parity claim) and is the safer direction — base risked promoting a *diverged* node (split-brain, acked rows lost on both sides). It is not a defect; I flag it because it changes emergency-failover availability semantics and the fail-closed behavior deserves an explicit line in operator-facing failover docs, not just the PR body. (The "missing local token -> permanently un-promotable" sub-case is **unreachable** — the downloader creates `_replication_sync_id.d` at construction; see Downgraded.)

**M3 — `BackupManager.close()` write-lock acquire is unbounded behind a promote's uploader `create()`** *(in-diff, incomplete-hardening)*
`BackupManager.java:441` `writeLock.lock()` is untimed while `WalUploader`'s ctor holds the read side across the whole native `create()` object-store retry loop. Base was **worse** for the same trigger (a lock-free `getPtr()` = a UAF), so the PR improved safety; the residual is a documented, shutdown-only stall (memory-safe -> at worst a SIGKILL-past-grace, never a UAF) that only goes truly unbounded under a zero/disabled backup-timeout config. Optional: `writeLock.tryLock(budget)` -> the leak-don't-free fallback the rest of the teardown already uses. Author documents the deferral.

**M4 — Close-quiesce budget (5s) is tighter than the orchestrator rendezvous (30s); a healthy-but-slow ACL reload or store-build can trip the leak-don't-free fallback** *(in-diff, incomplete-hardening)*
`EntCairoEngine.close()`'s `SWITCH_CRITICAL_SECTION_QUIESCE_ON_CLOSE_MS = 5_000` waits for spanned cascade work; on expiry it leaks the whole engine teardown to process exit (deliberate, memory-safe). But the ACL-reload span (`reloadAccessLists()` + 3x `loadSecrets()`) is a single uninterruptible disk sequence, and the cascade **store-builds run outside any span entirely** (covered only by the 30s orchestrator rendezvous). A slow-but-live reload/build under contended disk during shutdown can exceed 5s and trip the leak even though nothing is wedged — and a non-exiting `close()` (embedded/test reuse) accumulates the leak instead of relying on OS reclaim. Net-better-than-base (base had no quiesce at all), but the 5s/30s asymmetry is worth reconciling: either match the budget to `CLOSE_RENDEZVOUS_FLOOR_MS` or wrap the store-builds in a span. (engine-concurrency F1 + fresh-adversarial #3.)

**M5 — Rust non-moribund park-arm is a sibling silent-stall the self-heal fix doesn't cover** *(in-diff, incomplete-hardening)*
`process_events_task` (`mod.rs:3388-3395`): on a non-moribund `get_lastest_index_metadata` failure at task startup, the table parks forever consuming and discarding WAL events (no sender-drop, no re-spawn) while the primary keeps acking. The PR **improved** this vs base (raised `error!->critical!` paging + released the pending-startup slot so the cleaner isn't wedged) but deliberately kept the park (the moribund case genuinely needs it). Net-positive vs base; the author documents it as "the still-parked-startup failure case." Optional future improvement: in the non-moribund branch, drop the sender + `clean_up()` to self-heal, mirroring the two `run()` arms this PR fixed.

## Minor

- **Pending-upload count off-by-one** *(in-diff)* — `DurableUploadRegistry.getPendingUploadTxnCount`/`forEachPendingUpload` compute `notified - uploaded` with `uploaded` defaulting to `-1`, so a never-uploaded table with N committed txns reports **N+1** in the gauge/refusal message. Cosmetic: it never changes the refuse/proceed decision (gap>0 either way) and the gauge still reaches exactly 0 when caught up. A `max(0, notified - max(0, uploaded))`-style clamp fixes it. (role-switch F2.)
- **OSS PR title under-describes its contents** *(metadata)* — `fix(core): signal restore cancellation promptly on shutdown` names 1 of 4 changes; the all-users-facing QueryProgress double-fin fix is invisible in the squash-merge `git log` subject. Body is complete; consider a broader title.
- **`PrimaryRoleState.close()` uploader-handle/gauge removal sits in the `try` head** *(in-diff, incomplete-hardening)* — skipped only if an early teardown-tail statement throws (OOM-class); base skipped everything, so strictly better. (resource F1.)
- **Test javadoc duplication** — `CriticalSectionAdmissionRaceTest` maintains each test's rationale in both a class-level `<ul>` and per-method javadoc; collapse one. (+ 2 adjacent lower-degree cases.)
- **One leftover reflective probe** — `RoleSwitchObservabilityTest:243` still uses `getDeclaredMethod(...).setAccessible(true)` for `hadActiveBackupAtLastDemote` even though the same file was de-reflected to `engine.setBackupManagerForTest(...)`; add a `@TestOnly` accessor.
- **`$()` vs `$safe()` nits** — `DurableUploadRegistry:464` (DEBUG) and `EntCairoEngine:1542` (`TableToken`) use `$()`; trivial, ASCII-sanitized on disk, matches codebase convention.
- **Rust coverage-notes** *(Minor)* — the unmarshallable-refusal-throws-not-null arm (`mod.rs:4473-4495`, needs a JVM oracle; the *reachable* refusal path is tested), the `error!->critical!` level string, and the `reopened_txn < txn` regressed-txn guard are only transitively/indirectly exercised. (tests F1-F3.)

## Downgraded (false positives / non-issues)

- **"Role-switch drain default hard-wired to 5s = config regression"** — FALSE POSITIVE. Verified at base: `roleSwitchTimeoutMs` was an `AtomicLong` seeded to the 5s constant, written only by the dead consumer seam, and there is **no PropertyKey** for it. The default was always 5s in production; removal is a pure dead-code cleanup with zero behavior change.
- **"`assertOnlyOnce` fix turns unrelated suites red (CI risk)"** — the fix is correct; blast radius is essentially zero (3 OSS callers, 0 ENT callers -> only `SqlLoggingTest`, the intended target). Not a user-impact finding.
- **"Missing local token -> permanently un-promotable"** — unreachable: the downloader creates `_replication_sync_id.d` at construction, so a live replica always has the token.
- **Prior-round close-window + Rust dirty-latch/lifetime-fence Criticals** — moot: that machinery was added and removed within this PR (`555228e1b`/`fbbde860b`); the Rust teardown surface is byte-identical to base.

## Adjacent (pre-existing — file as issues, do not block)

- **~10 `throw_new(...).expect(...)` panic-abort sites** remain in `wal/`+`backup/` JNI; an `extern "system"` unwind aborts the JVM. Delta vs base: zero (the fix wave removed one). Realistically fires only on class-not-found/OOME. Not this PR's responsibility.
- **Stale `panic=abort` comments** contradict `Cargo.toml` (`panic='unwind'`); the PR corrected one. Doc-only.

## Coverage map (14 behavioral changes)

| Change | Test / assertion | Verdict |
|---|---|---|
| drainForDemote refuses on unuploaded gap | `PrimaryRoleStateDrainForDemoteTest` + `DemoteUploadDrainRefusalSwitchTest` (`waitForRole(UNKNOWN)`, `tables=[..]`) | TESTED |
| boot-seed gate `lastTxn>0` | `emptyWalTableDoesNotSeedPhantomDemoteBlocker` (incl. 8 `sys.acl_*`) + `uploadedSeedAtZeroBoundaryStillSeeds` | TESTED |
| Rust `clean_up` self-heal (both arms) | `..._removes_sender_so_next_event_respawns` x2 | TESTED |
| `RoleSwitchUploadPausedException` | `preflightDemoteRefusesPausedBacklogWithoutStoppingUploader` (+`cleanShutdownCount==0`) | TESTED |
| frozen-snapshot registry swap | `...ClosesOldStateAndRefusesRetry` (`!(getDurableAckRegistry() instanceof DurableUploadRegistry)`) | TESTED |
| promote parity fail-closed | Rust `data_parity_*`/`store_failure_is_indeterminate`; JNI `RefusedPromoteReplicaRecoveryTest`, `PromotePreflightThrowableRefusalTest` | TESTED |
| unmarshallable refusal throws-not-null | — (needs JVM oracle; reachable sibling tested) | UNTESTED (incomplete-hardening, Minor) |
| `replication.disabled.tables` parsed for NONE role | `EntPropServerConfigurationTest.testDisabledTablesParsedWithoutReplicationRole` | TESTED |
| demoted replica no longer re-arms backups | `BackupManagerScheduleArmGateTest` (truth table + role-vs-read-only) | TESTED |
| BackupManager implLock close-vs-dispatch | `BackupManagerSignalRestoreCancelCloseRaceTest` (9 tests, `dispatchCount==0`) | TESTED |
| two-sided admission | `CriticalSectionAdmissionRaceTest` (5 witnesses, state-based park) | TESTED |
| `ReplicaRoleState.openLoops` isClosing guard | `RoleStateOpenLoopsFaultInjectionTest` | TESTED |
| OSS QueryProgress double-fin gate | `SqlLoggingTest.testSimple` (`assertOnlyOnce fin.*create table x`) | TESTED |
| OSS executeCreateLiveView fin + sqlText | `SqlLoggingTest.testCreateLiveView` | TESTED |
| ENT ACL CREATE LIVE VIEW fin (r202 M3a) | `HttpAclLoggingTest.testCreateLiveViewStatementTextIsLogged` | TESTED (M3a closed) |
| OSS hydrate private-target + closing aborts + double-close | `MatViewHydrateTargetTest`, `MatViewHydrateClosingAbortTest`, `MatViewStateStoreImplDoubleCloseTest` | TESTED |
| OSS `preJoinCancelHook` | `LifecycleOrchestratorPreJoinCancelHookTest`, `...HookThrowsTest`, `RestoreCancelHookWiringTest` | TESTED |
| `triggerUploadMissed` per-table read-hold | `WalUploaderBootReplayCloseRaceTest` | TESTED |

## Summary

- **Verdict — ENT: approve with comments / OSS: approve.** Both gates pass. Comments worth addressing at/before merge: M2 (document the fail-closed failover behavior for operators — this is the most consequential semantic change), M1 (optional Rust sentinel to distinguish transient open-failure), M4 (reconcile the 5s/30s budget asymmetry), the pending-count off-by-one. None block.
- **Correctness gate:** 0 open Criticals. Severity audit run both directions: downward — no Critical to demote; upward — the two candidate escalations (M1 phantom-refusal, M2 promote fail-closed) are both **fail-safe** directions (they trade availability for data-safety), deliberately chosen, documented, recoverable, and net-positive vs base on the more-likely inputs, so neither completes a net-negative Critical symptom sentence.
- **Test gate:** passes. All 14 user-visible behavioral changes have efficacious, revert-sensitive regression tests; the only UNTESTED rows are defensive-hardening/log-level/internal-guard items (Moderate/Minor, none Critical-rubric). Prior r202 M3a and M4 are both resolved at this head.
- **Findings verified vs dropped:** ~28 raised across the fleet, ~4 dropped as false positives.
- **Split:** predominantly in-diff and incomplete-hardening (all net->=base); a few adjacent (pre-existing panic sites / stale comments); **out-of-diff-breakage: 0** — the change is genuinely well-contained (verified against the callsite/cross-context passes, not a level-shortcut).
- **Severity distribution:** 0 Critical, 5 Moderate, ~7 Minor, 2 Adjacent — proportional for a 90-file hardening PR.
- **CI note:** the OSS PR's "Enterprise CI Failed" comment predates the client-version fix (`41b04940f`); re-run should clear it.

*Generated by an automated multi-agent code review (Claude Opus). Prior review comments used as input only, not ground truth.*

jovfer:
# Review: PR #7361 — OSS half of the tandem, L3, from scratch

**Level 3, tandem mode** (paired with the enterprise PR). Head reviewed: **`674c669287`**; base `6610ab113b` (master ancestor). Reviewed via multiple fresh-context agents (correctness, cross-context caller-impact, tests, code-quality) plus source verification of the load-bearing paths; the changed files inside this PR are the review scope (the master-merge commits in the range are upstream and out of scope). Prior review comments were used as one input only.

## Headline

**No Critical findings; no out-of-diff breakage on the OSS surface.** The two behavior-changing pieces (the query-completion logging gate and the `LogCapture.assertOnlyOnce` fix) were confirmed with an actual green `SqlLoggingTest` run, and the branch-introduced mat-view hydrate bug is correctly fixed and guarded.

## Verified sound

- **`compileInner` double-`fin` gate drops no completion log.** `if (compiledQuery.getType() != CompiledQuery.NONE)` gates the generic `QueryProgress.logEnd(-1, …)` on exactly the negation of the `compileUsingModel` fall-through condition, so completion-log ownership moves cleanly: a statement either logs the generic `fin` (non-fallthrough) or falls through to `compileUsingModel` / a deferred `Operation` that logs its own. Enumerated all executors — the only `NONE`-on-success paths are side-effect-free bails that each log downstream. The only removed record is the always-wrong premature `id=-1` one. Verified: `mvn test -pl core -Dtest=SqlLoggingTest` → `Tests run: 2, Failures: 0` (`testSimple` single-logs create table; `testCreateLiveView` single-logs the new completion record).
- **`LogCapture.assertOnlyOnce` is now a real duplicate detector** (was checking `matcher.groupCount() == 0`, the regex's static capture-group count, so it always passed). Blast radius is fully contained: the only two callers in the whole tree are `LogCaptureTest` (new) and `SqlLoggingTest` (in-PR) — no external suite relied on the old always-pass behavior, so making it real reddens nothing (verified green). The `testSimple` assertions on create-table / insert / select / alter / update / rename / drop now genuinely verify single-logging.
- **`waitFor`/`waitForRegex` boundary fix is correct and pinned both ways.** The old `while (elapsed < max)` loop / `if (elapsed > max)` post-check returned silently on an exact-deadline timeout (a timed-out wait read as a successful match). Now the deadline is checked before the sleep and `>=` is treated as timed out. `LogCaptureTest` pins both directions with an injectable clock/sleeper: an exact-deadline wait throws, an already-matching regex still returns without sleeping.
- **`hydrateMatViewStateStore(MatViewStateStore target)` retarget is complete and fixes a branch-introduced bug.** Every enqueue in `loadMatViewIntoStore` (both `createViewState`, both `enqueueIncrementalRefresh`, all four `enqueueInvalidate`) now targets the `target` parameter; none is left on the engine field. The no-persisted-state kickstart that previously pointed at the field is fixed and guarded by `MatViewHydrateTargetTest#neverRefreshedViewKickstartsOnDistinctTarget` (a distinct recording target — the pre-existing `MatViewTest` no-arg path, where `target == field`, could never catch it). The entry + per-token `isClosing()` aborts are correctly placed in the loop body (not inside `loadMatViewIntoStore`, whose `catch (Throwable)` would swallow them), and `Objects.requireNonNull(target)` is checked up front. Witnessed by `MatViewHydrateClosingAbortTest` (each abort against a revert the other would mask).
- **`MatViewStateStoreImpl.close()` idempotency** (clear the map after the free loop; `clear()` drops its own map-clear) is sound for the sequential-close contract it documents; no concurrent double-close is reachable on the OSS side.
- **`TableSequencerAPI.getTxnTrackerIfExists` (non-creating) + `CairoEngine.isWalApplySuspended(TableToken, @Nullable SeqTxnTracker)`** — the 1-arg `isWalApplySuspended` still delegates through the creating `getTxnTracker`, so every existing 1-arg caller (OSS `ApplyWal2TableJob`, `MatViewRefreshJob`, `CheckWalTransactionsJob`, `TableSequencerImpl`, etc.) is byte-identical to base. Pinned by `WalWriterTest#testTxnTrackerIfExistsDoesNotCreateOrInstall` (both the null-tracker and resolved-tracker legs).
- **`LifecycleOrchestrator.preJoinCancelHook`** runs after the executor drain and before the bounded boot-thread join; a throwing hook is absorbed (logged, not propagated) so it cannot abort the stop loop. Tested by `LifecycleOrchestratorPreJoinCancelHookTest` (close doesn't burn the 30s join once the hook releases the in-flight start) and `LifecycleOrchestratorHookThrowsTest` (throwing hooks still run the full stop loop). `LifecycleTransitionTableTest` adds the `STARTING → STOPPING` row.
- **`CreateLiveViewOperation` `sqlText` field + `MetricsRegistryImpl.getTargetCount` (@TestOnly)** — additive; the single `new CreateLiveViewOperation(...)` callsite (the builder) passes `Chars.toString(sqlText)`; no null and no compile break.

## Minor

- **PR title under-describes the change** — `fix(core): signal restore cancellation promptly on shutdown` names one of four concerns. The all-users-facing `CREATE TABLE` double-`fin` fix (and the new `CREATE LIVE VIEW` completion logging) is invisible in the squash-merge subject; a reader scanning `master` history for the double-fin fix would never find it. The body is complete and accurate — only the title is narrow. Suggest e.g. `fix(core): harden lifecycle shutdown, mv hydration, and query-completion logging`.

## Coverage map (OSS behavioral changes — all TESTED)

| Change | Test | Verdict |
|---|---|---|
| `compileInner` double-fin gate | `SqlLoggingTest#testSimple` (green) | TESTED |
| `executeCreateLiveView` completion log + `sqlText` | `SqlLoggingTest#testCreateLiveView` (green) | TESTED |
| hydrate private-target + closing aborts + null reject | `MatViewHydrateTargetTest`, `MatViewHydrateClosingAbortTest` | TESTED |
| `MatViewStateStoreImpl.close()` idempotent + map-clear | `MatViewStateStoreImplDoubleCloseTest#closeIsIdempotent` | TESTED |
| `getTxnTrackerIfExists` + 2-arg `isWalApplySuspended` | `WalWriterTest#testTxnTrackerIfExistsDoesNotCreateOrInstall` | TESTED |
| `preJoinCancelHook` + throwing-hook absorption | `LifecycleOrchestratorPreJoinCancelHookTest`, `…HookThrowsTest`, `LifecycleTransitionTableTest` | TESTED |
| `LogCapture.assertOnlyOnce` real + boundary fixes | `LogCaptureTest` (6 methods) | TESTED |

## Summary

- **Verdict: approve.** 0 Critical, no out-of-diff breakage, all behavioral changes covered by efficacious revert-sensitive tests. One Minor (a broader title). No banned test-assertion patterns (no query-result `assertSql`, no `.returnsOnce` on deterministic queries, no vacuous assertions); the two behavior-changing test helpers were confirmed green.
- **Adjacent (pre-existing, non-blocking):** `LifecycleTransitionTableTest` javadoc carries pre-existing internal refs, but not in this PR's added lines.

*Automated multi-agent code review (Claude Opus, level 3, tandem). Prior review comments used as input only.*

ideoma:
#### [PR Coverage check]

:heart_eyes: **pass** : 58 / 61 (95.08%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/metrics/MetricsRegistryImpl.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/lifecycle/LifecycleOrchestrator.java|8|9|88.89%|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|18|19|94.74%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CreateLiveViewOperationBuilderImpl.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/seq/TableSequencerAPI.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|27|27|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewStateStoreImpl.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CreateLiveViewOperation.java|2|2|100.00%|

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=261961)
**Tested Commit:** `bbde297c206f8b36d468bfb99ddcaa82cdbcca66`

bluestreak01 (approved):
Reviewing PR #7361 at level 3 (full review).

### Critical

None.

### Moderate

None.

### Minor

#### Regex waits copy the captured log on each retry

Problem: Regex waits copy the entire captured log every retry.
Net impact: None for users; test-only helper, cost paid on already-failing runs.
Evidence: bbde297 benchmark, 1MB sink x 5,000 iterations: 3.16s vs 1.11s at base.

`core/src/test/java/io/questdb/test/tools/LogCapture.java:138-144`

Each unsuccessful 1 ms poll creates a full `String` snapshot and a new `Matcher`;
`StringSink.toString()` copies its buffer. The merge base reused a single matcher
against the live sink with no snapshot.

This is a deliberate trade, not a defect: matching against the live mutable
`SynchronizedSink` raced an in-flight `put()` and could throw
`StringIndexOutOfBoundsException` out of a waiting test. The snapshot removes that
race. `LogCapture` is test scaffolding under `core/src/test`, with eight test-file
callers and no `src/main` caller, so no database user or operator can observe the
cost. It also only accumulates when the awaited regex never appears — that is, on a
run that is already failing — since a present record matches within a poll or two.

Optional follow-up: match under the existing sink monitor, or refresh the snapshot
only when the sink has grown, keeping the race fix without the per-poll copy.

#### PR title omits bundled fixes

Problem: PR title omits three user-visible fixes.
Net impact: Release notes hide SQL logging and promotion safety changes.
Evidence: PR #7361 title at bbde297 mentions only restore cancellation.

The branch also changes materialized-view promotion hydration, WAL suspension
inspection, and SQL completion logging. Consider a broader title, e.g.
`fix(core): fix shutdown cancellation, role promotion, and SQL logging`.

#### Test method has excessive Javadoc

Problem: Regression test embeds multi-paragraph branch history in Javadoc.
Net impact: Maintainers must parse history before seeing test intent.
Evidence: `MatViewHydrateTargetTest.java:85-100` at bbde297.

Reduce to a precise test name plus a short comment; the durable background belongs
in class-level documentation.

### Coverage map

Test gate passes with **0 admitted coverage gaps**.

Local validation at `bbde297`:

- 18 targeted lifecycle, hydration, logging, and close tests passed.
- `WalWriterTest#testTxnTrackerIfExistsDoesNotCreateOrInstall` passed.
- OSS PR CI and macOS/Windows CI pass.

Runtime evidence for the completion-logging change: at head, `CREATE TABLE` emits
exactly one `fin` record carrying a real registry id, and `CREATE LIVE VIEW` emits
its own completion record; `ALTER`, `DROP`, and `INSERT` keep their existing
`id=-1` executor-path records unchanged.

Verified by source trace rather than by test: every keyword executor that returns
normally with `compiledQuery` at `NONE` falls through to `compileUsingModel`, so the
gated generic log cannot drop a completion record for a statement that actually ran.
The only removed record is the premature `id=-1` one. A `CREATE` compiled but never
executed now emits no completion record — a deliberate, documented behavior change.

### Summary

- **Verdict: approve.** Both gates pass.
- Correctness gate: passes; 0 Critical findings.
- Test gate: passes; 0 admitted Critical coverage gaps.
- Severity: 0 Critical, 0 Moderate, 3 Minor.
- Split: 2 in-diff findings, 1 metadata finding, 0 out-of-diff breakages.
- Submodules: no pointer changes, so no provenance question arises.
- Tradeoff worth naming: the `LogCapture` snapshot fixes a real race at the cost of
  data-sized allocation per poll on the failure path.
- Tandem Enterprise CI currently fails only `cargo fmt` at
  `questdb-ent/rust/qdb-ent/src/wal/uploader/mod.rs:4950`. That is outside this OSS
  diff and does not affect this verdict, but it should be formatted before the
  tandem lands.

PR #7184, opened 2026-05-29
feat(core): add per-query memory limits

puzpuzpuz:
Doc PR: https://github.com/questdb/documentation/pull/462

### What this adds

A runaway query, materialized view refresh, or WAL apply batch can today drive the whole server toward OOM. The only existing guard is the global RSS limit, which is process-wide: it kills whichever allocation happens to cross the line, not necessarily the workload responsible.

This PR adds an opt-in memory limit that caps how much native memory a *single* bounded workload may allocate, and throws at the offending allocation site when the cap is crossed -- so a runaway is stopped at its source while unrelated workloads keep running. There are three independent limits, one each for user SQL queries, materialized view refresh, and WAL apply (the latter two reuse the per-query mechanism with their own limits). A runaway in one does not draw against the budget of the others, nor against the global RSS limit.

All three default to `0` (unlimited), so behavior is identical to today's on upgrade. Operators opt in by setting a limit, and can raise, lower, or remove a limit at runtime through a config reload, without restarting the server.

On breach the engine throws a `CairoException` with `isOutOfMemory()` set -- the same type as a global-RSS breach, so existing handlers apply unchanged -- but with a deliberately distinct message so operators can tell the two apart in logs:

```
query memory limit exceeded [workload=QUERY, queryId=..., limit=..., used=..., size=..., memoryTag=...]
```

### How it works

- A new `MemoryTracker` wraps a 16-byte native `{used, limit}` block shared with Rust. A pluggable `MemoryTrackerProvider` hands one out per workload; the OSS provider pools them per workload invocation. The design leaves room for an enterprise provider that shares a tracker per principal without changing call sites.
- The tracker is acquired and released on the existing `QueryRegistry.register`/`unregister` pair (queries), in `MatViewRefreshJob` (refresh), and in `ApplyWal2TableJob` (WAL apply), and is bound on `SqlExecutionContext`. Acquisition is conditional on no outer tracker already being present, so subqueries and the two background workloads' inner SQL inherit the outer tracker rather than acquiring their own.
- New tracker-aware `Unsafe.malloc`/`realloc`/`free` overloads, and a per-query pointer on the Rust `QdbAllocator`, update and check both the existing global RSS counter and the per-query block. A null tracker degrades to today's global-only behavior.
[...]

PR #7362, opened 2026-07-07
test(griffin): speed up recently-added slow tests

puzpuzpuz:
This branch trims runtime from several recently-added test classes without changing what they verify. Each reduction was measured against the actual behavior it depends on (breach threshold, fault-injection extent, or fixture overhead) rather than guessed. No production code is touched.

## 1. Per-query memory limit tests (#7184)

The per-query memory limit tests fed their "fails on large input" cases far more rows than needed to breach the limit, and the OOM sweeps ran well past the last failing ceiling. Both wasted CI time without adding coverage.

- **`JoinMemoryTrackerTest`** — the six `FailsOnLargeInput` joins breach the unchanged 512 KiB limit at 10-15K distinct keys (measured). Inputs cut from `100_000` to `40_000` rows, a ~2.7-4x margin. Standalone: 21.1s -> 7.1s.
- **`AsOfJoinFastOomTest`** — the RSS-ceiling sweep emitted OOMs only up to ~21 KiB of slack, then ran the query to completion for ~5.5K more iterations. Cap the sweep at 24 KiB (from 64 KiB), keeping the 8-byte step across the entire OOM/leak-transition region. 5.6s -> 2.7s.
- **`GroupByVectorizedOomTest`** — same pattern; caps lowered to 80 KiB (from 128 KiB) and 48 KiB (from 96 KiB), just past the last observed OOM (~68 KiB / ~34 KiB). The parquet method ran two parquet drains per iteration, so trimming its tail is the bigger win. 3.35s -> 2.0s.
- **Horizon/window join memory tracker tests** (`ParallelHorizon`, `SyncHorizon`, `SyncWindow`, `ParallelWindow`) — the `array_agg` `FailsOnLargeSet` cases breach the unchanged 8 MiB limit at ~5K master rows (measured). Inputs cut from `200_000`/`2_000_000` to `40_000`/`400_000`, an 8-12x margin, preserving the 10x price density needed for ASOF matches and enough page frames for parallel fan-out.

Tradeoffs: the sweeps no longer exercise the post-recovery tail of all-succeed iterations (the recovery drain after each sweep already covers that), and the breach margins shrink from roughly 10-40x to 3-12x. All thresholds are
deterministic: the tracker counts requested bytes, not OS pages, and the first sweep step always trips (limit == current usage), so the "sweep never tripped" guard cannot silently pass. The parallel `*Horizon`/`*Window` classes see only a small wall-clock gain because they are dominated by per-test `CairoEngine` + `WorkerPool` setup, not table build; the input reduction still lowers their memory and I/O footprint.

The `*MemoryTrackerTest` + `*OomTest` warm-batch total drops from ~50s to 37s.

## 2. FirstLastNotNullParallelMergeTest consolidation (#7224)

This test spent ~11s almost entirely on per-`@Test` fixture overhead: each of its 28 type methods spun up a fresh `CairoEngine` + 4-worker `WorkerPool` and
ran the memory-leak scaffolding. Table size (100K rows) and iteration count (10) were proven irrelevant to runtime — shrinking them to 2K rows / 1 iteration left the total unchanged — so only the method count mattered.

Collapsed the 28 type methods into two (one per aggregate function) that loop over every value type inside a single shared engine and worker pool, so the fixture is paid twice instead of 28 times.

Coverage is unchanged: all 14 value types still run for both `first_not_null` and `last_not_null`, each repeated 10 times over a 100K-row table with the same sharding/small-frame configuration. Verified detection is intact by reverting the 25 pre-#7224 production files: `testLastNotNullAllTypes` then fails on the first type (`last_not_null` was the actual defect; `first_not_null` is a standing regression guard), and passes again once restored.

Runtime drops from ~11.1s to ~5.5s (measured back-to-back under identical conditions). Tradeoff: a per-type failure is reported with a `type=<T> func=<f>` prefix instead of a distinct method name, so per-type reporting is
coarser.

## Test plan

- `mvn -pl core -Dtest='*MemoryTrackerTest,*OomTest' test` passes; warm-batch total drops from ~50s to 37s.
- `mvn -pl core -Dtest=FirstLastNotNullParallelMergeTest test` passes in ~5.5s (was ~11.1s); fails as expected when the pre-#7224 production files are reverted.
- Confirmed every `FailsOnLargeInput` / `FailsOnLargeSet` case still throws `isOutOfMemory()` with "query memory limit exceeded" / "workload=QUERY".
- Confirmed the reduced sweeps still cross the OOM -> success boundary, so the in-loop recovery path is still exercised.

### DIFF
--- a/core/src/test/java/io/questdb/test/griffin/JoinMemoryTrackerTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/JoinMemoryTrackerTest.java
@@ -79,8 +79,8 @@ public class JoinMemoryTrackerTest extends AbstractCairoTest {
         // asof_dense + multi-key routes to AsOfJoinDenseRecordCursorFactory. Its scan maps memorize every
         // distinct join key (no eviction without TOLERANCE), so a high-cardinality join grows them past the limit.
         assertMemoryLeak(() -> {
-            execute("CREATE TABLE m AS (SELECT cast(x AS SYMBOL) k1, cast(x AS SYMBOL) k2, (x * 1_000_000L)::timestamp ts FROM long_sequence(100_000)) TIMESTAMP(ts) PARTITION BY DAY");
-            execute("CREATE TABLE s AS (SELECT cast(x AS SYMBOL) k1, cast(x AS SYMBOL) k2, (x * 1_000_000L)::timestamp ts FROM long_sequence(100_000)) TIMESTAMP(ts) PARTITION BY DAY");
+            execute("CREATE TABLE m AS (SELECT cast(x AS SYMBOL) k1, cast(x AS SYMBOL) k2, (x * 1_000_000L)::timestamp ts FROM long_sequence(40_000)) TIMESTAMP(ts) PARTITION BY DAY");
+            execute("CREATE TABLE s AS (SELECT cast(x AS SYMBOL) k1, cast(x AS SYMBOL) k2, (x * 1_000_000L)::timestamp ts FROM long_sequence(40_000)) TIMESTAMP(ts) PARTITION BY DAY");
             drainWalQueue();
             final String sql = "SELECT /*+ asof_dense(m s) */ m.k1 FROM m ASOF JOIN s ON (m.k1 = s.k1 AND m.k2 = s.k2)";
             assertUsesFactory(sql, AsOfJoinDenseRecordCursorFactory.class);
@@ -136,8 +136,8 @@ public class JoinMemoryTrackerTest extends AbstractCairoTest {
         // asof_dense + single SYMBOL routes to AsOfJoinDenseSingleSymbolRecordCursorFactory (no sinks; scan
         // maps bound via the base setMemoryTracker). A high-cardinality join grows the maps past the limit.
         assertMemoryLeak(() -> {
-            execute("CREATE TABLE m AS (SELECT cast(x AS SYMBOL) k, (x * 1_000_000L)::timestamp ts FROM long_sequence(100_000)) TIMESTAMP(ts) PARTITION BY DAY");
-            execute("CREATE TABLE s AS (SELECT cast(x AS SYMBOL) k, (x * 1_000_000L)::timestamp ts FROM long_sequence(100_000)) TIMESTAMP(ts) PARTITION BY DAY");
+            execute("CREATE TABLE m AS (SELECT cast(x AS SYMBOL) k, (x * 1_000_000L)::timestamp ts FROM long_sequence(40_000)) TIMESTAMP(ts) PARTITION BY DAY");
+            execute("CREATE TABLE s AS (SELECT cast(x AS SYMBOL) k, (x * 1_000_000L)::timestamp ts FROM long_sequence(40_000)) TIMESTAMP(ts) PARTITION BY DAY");
             drainWalQueue();
             final String sql = "SELECT /*+ asof_dense(m s) */ m.k FROM m ASOF JOIN s ON k";
             assertUsesFactory(sql, AsOfJoinDenseSingleSymbolRecordCursorFactory.class);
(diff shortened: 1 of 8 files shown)


puzpuzpuz:
## Code review

Test-only change (+101/−191, 8 files), reviewed in full: the two RSS-ceiling OOM sweeps, the per-query-tracker breach tests, the `FirstLastNotNull` consolidation, and the load-bearing primitives underneath (`Unsafe` RSS accounting, `TestUtils.execute` threading).

### Verdict: approve — no blocking issues

The design is sound where it matters most: **every breach assertion fails loudly, never silently passes.** `assertQueryBreaches` (Join + Horizon/Window variants) and the `sawOom` sweep guard all end in `Assert.fail(...)` if the expected OOM doesn't occur — so an over-aggressive input cut produces a red test, not a hollow green one. And the breach point is deterministic: `getRssMemUsed()` / the per-query tracker count **requested bytes** on a counter (`Unsafe.checkAllocLimit`, `usage + size > limit`), not OS pages, so the measured margins hold run-to-run.

Verified explicitly:

- **Breach still fires.** Join tests: 40K distinct SYMBOL keys breach 512 KiB deterministically (measured breach at 10–15K keys → 2.7–4× margin). Horizon/Window: 40K/400K preserves the 10× price density and 8-symbol cardinality; breach at ~5K trades → 8× margin. Success / leak-loop cases (inputs 4–5000) are untouched, so none can flip.
- **Consolidation preserves coverage & detection.** `ROW_COUNT=100_000`, `KEY_COUNT=1000`, `ITERATIONS=10` are all unchanged — 14 types × 2 funcs = 28 runs, same as before; only the fixture count drops 28→2. `TestUtils.execute` runs the lambda on the JUnit thread, so the `catch (AssertionError) → rethrow` cannot swallow a failure, and the busy reader/writer asserts still run once at the end.
- **RSS accounting is deterministic requested-bytes**, so the sweep-cap cuts can't spuriously fail: slack=0 → limit==usage → first alloc trips → `sawOom` always true; the recovery drain runs unlimited.

---

### Moderate

**1. `AsOfJoinFastOomTest` sweep cap (24 KiB) has the thinnest margin — consider a bit more headroom.** `assertNoLeakOnCursorOom`, the `slack <= 24 * 1024` loop.

The leak this guards is detected only when the swept ceiling lands in the narrow window where the first sink `reopen()` succeeds and the second trips — a transition at the top of the OOM region. Measured last-OOM is ~21 KiB, capped at 24 KiB (~14% headroom, down from ~205% at the old 64 KiB). The ceiling is checked against a **global** counter shared with any concurrent native allocation. In practice this test drives a single-threaded fast-path ASOF drain with idle background threads, so noise is minimal — but if the tracked cursor-open allocation drifts up by more than ~3 KiB on a loaded CI host, the transition ceiling shifts above the cap and a *future* re-introduction of the sink leak escapes **this** test (still caught by the tracker tests + malloc fuzzer). This cannot turn the test red; worst case it silently loses regression teeth.

Suggestion: bump this one cap to ~32 KiB — cheap insurance on the tightest margin, barely dents the 5.6s→2.7s win. The GroupBy caps (80/68 ≈ 18%, 48/34 ≈ 41%) are comfortable and fine as-is.

---

### Minor

**2. Consolidation is fail-fast across types.** `assertEveryTypeKeepsNonNull`. When an early type in `TYPES` fails, the loop throws and the remaining types don't run that pass, where the old 28 independent methods each reported independently. Irrelevant for an all-types regression (the merge bug); for N independently-broken types you fix-and-rerun iteratively instead of seeing all N at once. Same "coarser reporting" tradeoff already disclosed; the `type=<T> func=<f>` prefix keeps attribution on the first failure. Acceptable.

**3. `private static final String[][] TYPES`.** CLAUDE.md prefers `ObjList<T>` over `T[]` object arrays, but that targets resizable, cleanup-integrated collections on production paths — a compile-time-constant type→expr lookup table in a test is idiomatically `String[][]` across the test tree. Borderline; fine to leave.

**4. "Performance" label on a test-only PR.** The body states "No production code is touched," yet the PR carries `Performance` (defined as product perf). Defensible as CI performance, but slightly misleading for release-notes/label filtering; `CI` alone is the cleaner fit. Trivial.

---

### Adversarial hypotheses tested and dismissed

- "Reduced input might silently stop breaching" — no; guards `Assert.fail` on no-breach and breach is deterministic.
- "Consolidation loses coverage" — no; `ROW_COUNT`/`KEY_COUNT`/`ITERATIONS` unchanged, all 28 runs preserved.
- "`catch (AssertionError)` in a pooled lambda swallows failures" — no; `TestUtils.execute` runs the runnable on the calling thread.
- "`.noLeakCheck()` disables leak detection" — no (and pre-existing); the outer `assertMemoryLeak` legitimately wraps the WorkerPool + multi-statement scope.
- "Shared engine leaks state between types and masks the bug" — no; `DROP TABLE IF EXISTS` on a non-WAL table is synchronous, parallel-merge state is per-query, config unchanged; detection independently confirmed by the revert test.
- "Removed per-type methods / helpers referenced elsewhere" — no; grep shows zero external references to `assertEveryKeyKeepsNonNull` or the deleted per-type methods; `createTrades`/`createPrices`/`createMultiHorizonTables` are class-private.
- "Lowering the OOM cap could make `sawOom` false" — no; slack=0 trips on the first counted allocation of the net-positive compile+open.

---

### Summary

Clean, well-reasoned, low-risk test speedup; the body honestly discloses every tradeoff (shrunken margins, coarser per-type reporting, dropped all-succeed tails). Only substantive tradeoff: reduced *leak-regression* robustness on the thinnest RSS sweep (AsOf 24/21 KiB) — non-blocking, corroborated elsewhere, suggest ~32 KiB. Findings: 0 Critical, 1 Moderate, 3 Minor; 7 adversarial hypotheses verified and dismissed. All findings in-diff; 0 out-of-diff (the entire change surface is private test helpers / `@Test` methods with no external callers, so there is no cross-context blast radius).

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

puzpuzpuz:
Superseded by #7371, which is based on the same branch pushed to the main repo (questdb/questdb) instead of my fork so CI runs with the correct permissions. Closing this one in favor of #7371.

PR #7363, opened 2026-07-07
chore(core): resolve column files by writer index when converting a partition to Parquet

RaphDal:
## Summary

`TableUtils.produceParquetFromNative` resolves each column's `_cv` entries (name txn, column top, symbol-table name txn) by the dense metadata index. That is only correct when a partition's columns have never been re-keyed. After `ALTER COLUMN TYPE`, a column's dense index and writer index diverge, and the encoder opens the wrong file name — `x.d` instead of the real `x.d.<txn>` — and fails with `could not open, file does not exist`.

## Who hits this

The failure surfaces through the Enterprise storage-policy `TO PARQUET` conversion, which calls `produceParquetFromNative` with a `TableReader`'s metadata. When a policy-managed table has had a column type changed and one of its native partitions is converted to Parquet, the conversion fails and the storage-policy pipeline stalls on that partition.

The OSS `CONVERT PARTITION TO PARQUET` statement is **not** affected: it runs through `TableWriter`, whose metadata keeps deleted/re-keyed columns as tombstones in place, so the dense index still equals the writer index. There is therefore no OSS SQL reproduction; the regression test exercises the function directly with a reader's metadata (see Test plan).

## Root cause

`TableUtils.produceParquetFromNative` iterates a partition's columns by dense metadata index and reads the `_cv` with that index:

```java
final long columnNameTxn = columnVersionReader.getColumnNameTxn(partitionTimestamp, columnIndex);
final long columnTop     = columnVersionReader.getColumnTop(partitionTimestamp, columnIndex);
...
final long symbolTableNameTxn = columnVersionReader.getSymbolTableNameTxn(columnIndex);
```

The column-version file is keyed by writer index — the index `TableReader` uses for exactly these lookups (`getColumnNameTxn(ts, metadata.getWriterIndex(i))`). `ALTER COLUMN TYPE` removes the old column and appends the rewritten one at a new, higher writer index. In a `TableReader`'s compacted metadata the dense index no longer equals the writer index, so `getColumnNameTxn` misses and returns the default (-1); the encoder then builds `x.d` instead of `x.d.<txn>`. The same divergence feeds `getColumnTop` and `getSymbolTableNameTxn`, so those read another column's entry for a re-keyed column.

`TableWriter`'s metadata keeps the tombstone slot, so dense == writer there and the writer-driven `CONVERT` path is unaffected — which is why the defect only shows up through a reader-metadata caller.

## Fix

Resolve the three `_cv` lookups by writer index, mirroring `TableReader`, and
reuse the per-column metadata object while doing so. The name txn and column
top now share one `_cv` record lookup instead of searching the same
`(partitionTimestamp, writerIndex)` pair twice:

```java
final TableColumnMetadata tableColumnMetadata = metadata.getColumnMetadata(columnIndex);
final int writerIndex = tableColumnMetadata.getWriterIndex();
final int versionRecordIndex = columnVersionReader.getRecordIndex(partitionTimestamp, writerIndex);
long columnNameTxn = columnVersionReader.getColumnNameTxnByIndex(versionRecordIndex);
if (columnNameTxn == -1) {
    columnNameTxn = columnVersionReader.getDefaultColumnNameTxn(writerIndex);
}
final long columnTop = columnVersionReader.getColumnTopByIndexOrDefault(
        versionRecordIndex,
        partitionTimestamp,
        writerIndex,
        -1L
);
...
final long symbolTableNameTxn = columnVersionReader.getSymbolTableNameTxn(writerIndex);
```

`columnId` keeps `getOriginalWriterIndex()` unchanged — that is the Parquet field id (the replacing-chain head), a separate concern that is already correct. For callers whose dense and writer indices coincide (the writer-driven `CONVERT` path, and any table with no re-keyed columns) the resolved index is identical, so their output is unchanged.

The review follow-up also fixes the same dense-index lookup in
`DebugUtils.reconcileColumnTops`. That method is only referenced by a commented
debug assertion in `TableReader.reload()`, so the change has no live production
caller; it keeps the debug helper consistent with the production reader path if
the assertion is re-enabled. A small regression test now exercises that helper
with a reader whose dense column index and writer index resolve different
column tops.

## Test plan

Three `AlterTableChangeColumnTypeTest` cases exercise `TableUtils.produceParquetFromNative` directly with a `TableReader`'s compacted metadata — the shape the storage-policy `TO PARQUET` conversion uses — one per writer-index `_cv` lookup. Each opens a reader, asserts the dense/writer divergence as a precondition, produces the parquet, then decodes it back through `read_parquet(...)` and asserts the actual column values (not just a positive length).

- `testProduceParquetFromNativeResolvesReKeyedColumnByWriterIndex` covers the column name txn (`getColumnNameTxn`). `INT -> DOUBLE` re-key; the re-keyed column decodes to `{10.0, 20.0}`. Reverting only that lookup opens `x.d` instead of `x.d.<txn>` and fails this case (and, since the name txn resolves every column's `.d` file, the other two as well).
- `testProduceParquetFromNativeResolvesColumnTopByWriterIndex` covers the column top (`getColumnTop`). A column added after the partition existed has top 2; dropping an earlier column shifts its dense index below its writer index. It decodes to `{null, null, 100, 200}`. Reverting only that lookup resolves an earlier column's top (0), mis-aligns the data, and fails this case alone — the other two pass. A re-key alone keeps a column's display position and propagates its top to the new writer index, so the top does not diverge; a `DROP` is what drives the split, which is why this case does not reproduce through a pure `ALTER COLUMN TYPE`.
- `testProduceParquetFromNativeResolvesSymbolFilesByWriterIndex` covers the symbol-table name txn (`getSymbolTableNameTxn`). `INT -> SYMBOL` re-key; the symbol column decodes to `{1, 2}`. Reverting only that lookup opens the wrong (missing) `.o`/`.c` files and throws in this case alone — the other two pass.
- `testReconcileColumnTopsResolvesByWriterIndex` covers `DebugUtils.reconcileColumnTops`. It builds the same dense-index vs writer-index column-top split, verifies the two indices resolve different `_cv` tops, and asserts the debug helper reconciles against the writer-index top.


RaphDal:
# Review: PR #7363 — `chore(core): resolve column files by writer index when converting a partition to Parquet`

**Reviewed at level 3** (full mission-critical pass: 11-agent-equivalent parallel review, per-finding source verification). Diff: 2 files, +86/−3, Java only (no Rust).

**Verdict: approve.** The fix is correct, complete, and provably regression-free. There are **no Critical or blocking findings**. The one substantive theme is test coverage — 2 of the 3 changed lookups ship without a discriminating test. Everything below is a hardening suggestion, not a merge blocker.

The core mechanics were verified independently and by four parallel review passes (correctness/cross-context, test coverage, fresh-context adversarial, adversarial performance), which converged on the same conclusion. The test was *empirically* run with and without the fix.

---

## Why the fix is correct (verified, not assumed)

- **`_cv` is writer-keyed.** `ColumnVersionReader.getRecordIndex` keys records by writer index (`ColumnVersionReader.java:200-221`). `TableReader` always feeds these three lookups `metadata.getWriterIndex(columnIndex)` (`TableReader.java:371, 1287, 1752-1753, 1793`). The old `produceParquetFromNative` passed the dense `columnIndex` — wrong the moment dense ≠ writer. The fix (`TableUtils.java:1879`, then `:1881, :1882, :1907`) makes the function a faithful mirror of the reader. `getColumnMetadata(i).getWriterIndex()` is exactly what `AbstractRecordMetadata.getWriterIndex(i)` does (`:101-102`).
- **No writer/CONVERT-path regression — proven, not hand-waved.** `TableWriterMetadata.addColumn` stores `writerIndex = columnIndex` and appends at `columnMetadata.size()` (`:272-289`); `removeColumn` only `markDeleted()`s, never compacts (`:299-306`); reload re-assigns `writerIndex = i` for every slot including tombstones (`:187, :195`). So **dense == writer for every column on the writer path**, and the only OSS caller (`TableWriter.produceParquetFromNative:11295`) sees a pure no-op. The change bites only for a *reader's* compacted metadata (`TableReaderMetadata` drops deleted columns but keeps the true writerIndex) — i.e. the Enterprise storage-policy `TO PARQUET` path.
- **Completeness — nothing else was mis-keyed.** Every other `columnIndex` use in the loop is correctly dense (`getColumnType`, `getColumnName`, `getParquetEncodingConfig`, and the symbol-provider `containsNullValue`/`getSymbolCount`, which delegate to the dense-indexed `reader.getSymbolMapReader` — `TableUtils.java:2946-2952`). `columnId = getOriginalWriterIndex()` correctly stays put (it's the stable parquet field id). Bloom-filter indexes use descriptor ordinals (`deriveBloomFilterColumnIndexes:813-826`), unaffected. The three `_cv` calls were the only writer-keyed lookups.
- **Pre-fix impact.** For any partition with a re-keyed column, the Enterprise conversion resolves the name txn to the default `-1` and opens `x.d` instead of `x.d.<txn>` → hard `could not open, file does not exist` (empirically reproduced), stalling the storage-policy pipeline on that partition. In adjacent scenarios where the old column file / stale `_cv` entry survives *and* is mappable at the new type's width, the old code could instead map stale pre-conversion bytes and encode them as the new type — a plausible silent-corruption path, though not one that could be demonstrated; the confirmed behavior is the hard failure.

---

## Moderate (worth addressing; not blocking)

### 1. Two of the three changed lookups ship untested — in-diff (test)
The test uses `t(ts, x INT)` → `INT→DOUBLE`, which only exercises the **fixed-size branch**:
- **`getSymbolTableNameTxn(writerIndex)` (`TableUtils.java:1907`) has zero coverage.** It lives only in the symbol branch (`if (ColumnType.isSymbol(...))`, `:1887`), which `INT→DOUBLE` never enters. All three independent bug-hunting passes flagged this. In OSS the reader-metadata symbol path has *no* production caller — a unit test is the only thing that can cover it, and none does. The untested scenario: a re-keyed **SYMBOL** column (e.g. `ALTER COLUMN … TYPE SYMBOL`) converted with reader metadata, which resolves the root `.o`/`.c` files by the (now writer) index.
- **`getColumnTop(writerIndex)` (`:1882`) is not discriminated.** `x` is present from table creation, so its column top is 0 and `getColumnTop(dense)` and `getColumnTop(writer)` both return 0 (the pre-fix run still took the `columnRowCount > 0` branch). A regression reverting *only* this line would still pass. It bites when a re-keyed column has a non-zero top — a column **added after** the partition existed, then re-keyed.

Suggestion: add a case with (a) a re-keyed SYMBOL column and (b) a column added post-partition (top > 0) then re-keyed. The var-size branch is lower priority — it introduces no new `_cv` lookup beyond the two already at play.

### 2. Assertion only checks `parquetLen > 0` — in-diff (test)
`Assert.assertTrue(..., parquetLen > 0)` (`:1444-1445`) verifies only that *something* was produced, not *what*. It's an adequate discriminator for the shipped scenario (pre-fix throws before producing anything), but it cannot catch a wrong-column/stale-data encode. Decoding the output and asserting `x = {10.0, 20.0}` over 2 rows would be strictly stronger and would also cover the `getColumnTop` resolution that `> 0` leaves unverified.

---

## Minor

- **Test doc comment slightly over-reaches — in-diff (`:1382-1385`).** It lists all three `_cv` resolutions ("name txn, column top, symbol table name txn") as things the test guards, but the test only *forces* the name-txn path. Worth a clause noting the other two ride on analogy to `TableReader` (until finding #1 is addressed).
- **Optional readability tidy — in-diff.** The loop now calls `metadata.getColumnMetadata(columnIndex)` three times (`:1875, :1879, :1884`). Hoisting `final TableColumnMetadata cm = metadata.getColumnMetadata(columnIndex);` and reading the ids off `cm` reads cleaner. This is **not** a performance fix — the accessor is an array load and the JIT already CSEs the repeats; purely style.
- **PR metadata.** The title leans on an internal detail ("resolve column files by writer index"); the repo convention asks titles to speak to user-facing impact, so something like "fix Parquet conversion failing after a column type change" reads better in release notes — though for a `chore` hardening this is defensible. `chore` vs the `Bug` label is a judgment call (no OSS SQL repro, but a real Enterprise-path failure). The PR body itself is exemplary — level-headed, accurate root-cause, honest that it's a no-op for OSS callers, and empirically grounded ("verified in both states").

---

## Downgraded (candidate concerns verified and cleared)

- **Writer/CONVERT-path regression** → false positive. dense == writer invariant proven (`TableWriterMetadata` never compacts); the change is a no-op there.
- **Other mis-keyed lookups** → none. Symbol provider correctly stays dense; `columnId` correctly stays `getOriginalWriterIndex()`; bloom filters use descriptor ordinals.
- **Performance** → neutral. Cold path (once per column per conversion, dominated by mmap/encode syscalls); no per-row cost, no new allocation.
- **Test resource leak on the throw path** → none. `assertMemoryLeak` reported clean when the pre-fix throw fired; `bloomIndexes` is freed in `finally`, everything else via try-with-resources.
- **"Silent corruption pre-fix" as a stated fact** → downgraded to a plausible-but-undemonstrated secondary risk (the confirmed pre-fix behavior is a hard throw).
- **Out-of-diff `timestampIndex` vs `columnId` inconsistency** (`:1858-1860` uses `getWriterIndex()`, `:1875` uses `getOriginalWriterIndex()`) → **pre-existing, not this PR**. Safe today only because the designated timestamp cannot be type-converted (`TableWriter.java:1194-1197`); informational.

---

## Summary

- **Verdict: approve** — correct, complete, regression-free fix that mirrors the authoritative `TableReader` `_cv` access pattern. No blocking issues.
- **Tradeoff to note:** the change is a no-op for every current OSS caller and is exercised only by the new synthetic test; its real beneficiary is the out-of-repo Enterprise reader-metadata path. That's intended (shared-function hardening) and the PR states it plainly.
- **Actionable before merge (optional):** cover the untested symbol-branch and non-zero-column-top scenarios, and strengthen the assertion from `> 0` to a decoded-content check.
- **Draft findings:** ~6 verified (all Minor/Moderate, test-focused), ~6 candidate concerns verified and dropped as false positives/pre-existing.
- **In-diff vs out-of-diff: 6 in-diff, 0 out-of-diff.** This is a legitimate zero, not an underrun: the callsite inventory is exhaustive and tiny — one OSS caller (`TableWriter`, verified no-op), one test, and the out-of-repo Enterprise caller — each explicitly checked.

bluestreak01 (changes requested):
@RaphDal

# Review: PR #7363 — `chore(core): resolve column files by writer index when converting a partition to Parquet`

**Reviewed at level 3** (full mission-critical pass: 5 parallel fresh-context review agents covering correctness/cross-context, performance, tests/efficacy, code-quality/metadata, and a checklist-free adversarial pass; per-finding source verification including an **empirical revert-and-run**). Diff: 2 files, +86/−3, Java only (no Rust, no on-disk format change).

## What the change does

`TableUtils.produceParquetFromNative` iterates a partition's columns by **dense metadata index** and read three column-version (`_cv`) entries — name txn, column top, symbol-table name txn — with that same dense index. `_cv` is keyed by **writer index**. For a `TableReader`'s compacted metadata (tombstones removed), `ALTER COLUMN TYPE` makes dense ≠ writer, so those three lookups miss and the encoder opens `x.d` instead of `x.d.<txn>` and throws. The fix resolves the three lookups by `metadata.getColumnMetadata(columnIndex).getWriterIndex()`, exactly mirroring `TableReader` (`TableReader.java:371, 1287, 1793`). `columnId` stays `getOriginalWriterIndex()` (the Parquet field id / replacing-chain head) — correctly a separate concern.

**The production fix itself is correct and complete within the method** — confirmed independently by all five agents and by source trace:
- Writer-driven caller (`TableWriter.produceParquetFromNative`) passes writer metadata where deleted columns stay as tombstones, so `getWriterIndex(i) == i` and the change is a proven no-op there — no regression.
- The designated-timestamp `writerIndex`-vs-`columnId` question is a non-issue: the timestamp is unre-keyable (`TableWriter.java:1194`, `SqlCompilerImpl.java:1074` reject it), so `writerIndex == originalWriterIndex` for it always.
- The symbol provider legitimately stays on the dense index (`reader.getSymbolMapReader(columnIndex)`), independent of the writer-index `_cv` key — matching `TableReader.java:1752-1755`.
- The reverse parquet→native path and the bloom-filter index derivation use different index spaces and are not affected — not a partial fix.
- No performance concern: the added `getColumnMetadata(columnIndex).getWriterIndex()` is an O(1) `ObjList` array read + field load on a once-per-partition path.

## Critical

The fix touches **three** `_cv` lookups. The regression test proves only **one** of them; the other two changed lines are on reachable, data-integrity paths and ship **untested**. This is the blocking issue.

### C1 — `getColumnTop` change is untested (empirically proven unguarded) — in-diff
`TableUtils.java:1882` — `getColumnTop(partitionTimestamp, writerIndex)`.

The added regression test uses an `INT → DOUBLE` column whose column top is `0` in the partition. In that scenario **both** the correct writer index (2) and the buggy dense index (1) resolve column top to `0`, so this line's correctness is never observed. Verified empirically: reverting **only** the `getColumnTop` line back to the dense index (keeping the name-txn fix) leaves the test **green**.

Reachable regression path: a column **added after** a partition already has rows (non-zero column top) and then re-keyed by `ALTER COLUMN TYPE`, converted through a reader's metadata. A dense-index `getColumnTop` returns the wrong column top → wrong `columnRowCount` → wrong byte extent mapped for the column → converted-partition data corruption or a spurious throw. Worst reachable consequence is data integrity, so this is Critical, not Moderate.

**Fix (this PR):** extend the regression test with a column that has a non-zero column top in the converted partition and is re-keyed (dense ≠ writer), and assert the produced data — not just that it didn't throw (see C3).

### C2 — `getSymbolTableNameTxn` change is untested (symbol branch never runs) — in-diff
`TableUtils.java:1904` — `getSymbolTableNameTxn(writerIndex)`.

This line executes only inside the `ColumnType.isSymbol(...)` branch. The test's column is `INT → DOUBLE`, so the symbol branch never runs and this third lookup is never exercised.

Reachable regression path (does not require re-keying the symbol column itself — only shifting its dense index): `CREATE TABLE t (ts, a INT, s SYMBOL)`, then `ALTER COLUMN a TYPE DOUBLE` (re-keys `a` to a higher writer index, tombstones the old slot). In the reader's compacted metadata `s` is dense 1 / writer 2 — diverged while still a symbol. A dense-index `getSymbolTableNameTxn` then resolves the tombstoned `a`'s entry → opens the wrong `.o`/`.c` symbol offset/char files → throw, or reads another column's symbol data. Same defect class as the bug this PR fixes, on the same enterprise `TO PARQUET` path, untested.

**Fix (this PR):** add a test where a symbol column's dense index diverges from its writer index and the partition is converted, so the symbol branch runs against the corrected lookup.

### C3 — assertion only proves "did not throw," not correctness — in-diff
`AlterTableChangeColumnTypeTest.java:1444` — `assertTrue(..., parquetLen > 0)`.

The assertion catches the specific name-txn regression (which throws before returning), but a regression that resolves a *wrong-but-nonempty* column, or encodes stale bytes reinterpreted under the new type without a file-size error, would still return `parquetLen > 0` and pass. Reading the produced Parquet back (round-trip parquet→native, or decode and assert the `DOUBLE` values `10.0`/`20.0`) converts this into a genuine correctness guard and, done over the C1/C2 scenarios, closes all three gaps in one stroke.

## Moderate

### M1 — PR title type/wording vs. the change — metadata
The PR carries the `Bug` label and the body describes a real correctness defect (wrong file opened → conversion fails → storage-policy pipeline stalls), but the title uses `chore(core):`. Per CLAUDE.md, `Bug` pairs with `fix`. `chore` is *defensible* on the body's own argument (no OSS SQL path reaches it, so nothing lands in OSS release notes) — but then the `Bug` label is inconsistent. Also, the description "resolve column files by writer index" is internal mechanics; CLAUDE.md requires the description (copied to release notes) to state end-user impact. Suggested: `fix(core): fix Parquet partition conversion failing after a column type change`. Behaviorally inert (metadata only), hence Moderate.

## Minor

### m1 — new test method is out of alphabetical order
`AlterTableChangeColumnTypeTest.java:1379` — `testProduceParquetFromNativeResolvesReKeyedColumnByWriterIndex` (P) is placed right after `testConvertInvalidColumnFailsWriterIsOk` (C) and before `testFixedSizeColumnEquivalentToCast` (F). Per CLAUDE.md alphabetical member ordering it belongs between `testNewTypeMissing` (N, line 1591) and `testShouldTruncateConvertedColumns` (S, line 1596). No runtime effect.

*(Optional, not filed as a finding: `metadata.getColumnMetadata(columnIndex)` is now called three times in the loop body — hoisting to one local would read cleaner, but `getQuick` is trivially cheap and the duplication is pre-existing.)*

- **"Timestamp `writerIndex` vs `columnId` mismatch after re-key"** — dismissed: the designated timestamp cannot be re-keyed (`TableWriter.java:1194`, `SqlCompilerImpl.java:1074`), so `writerIndex == originalWriterIndex` for it; the encoder's `timestampIndex`/`columnId` match always holds.
- **"Bloom-filter helpers / reverse parquet→native have the same dense-vs-writer confusion"** — dismissed: both use a distinct descriptor-ordinal / writer-metadata index space (dense == writer on the writer path); neither touches `_cv` incorrectly.
- **"Writer path (`CONVERT PARTITION TO PARQUET`) could regress"** — dismissed: writer metadata keeps tombstones so `getWriterIndex(i) == i`; the substitution is a proven no-op there.

## Coverage map

| Behavioral change | Test | Failure link | Dimensions | Verdict |
|---|---|---|---|---|
| `getColumnNameTxn` keyed by `writerIndex` (name-txn) | `testProduceParquetFromNativeResolvesReKeyedColumnByWriterIndex` | Reverting the line → `could not open, file does not exist: .../x.d` (**empirically verified**: reverted, recompiled, test failed) | happy ✓; error ✓ (throws on wrong file); NULL N/A (name txn `-1` sentinel handled unchanged); boundary partial; cleanup ✓ (`assertMemoryLeak`) | **TESTED** |
| `getColumnTop` keyed by `writerIndex` (column-top) | same test — **does not guard it** (revert-only-this-line keeps test green; both indices resolve top=0) | none | error/data ✗ (untested) | **UNTESTED (Critical, C1)** |
| `getSymbolTableNameTxn` keyed by `writerIndex` (symbol-txn) | none — `isSymbol` branch never executes (INT→DOUBLE) | none | error/data ✗ (untested) | **UNTESTED (Critical, C2)** |

Totals: **3 behavioral changes, 1 tested, 2 UNTESTED.**

## Summary

- **Verdict: request changes.** The production code is correct, minimal, and faithfully mirrors `TableReader` (verified by all agents plus an empirical revert-and-run proving the name-txn regression). It is blocked purely on **test coverage**, not on a code defect: two of the three `_cv` lookups the fix changes (`getColumnTop`, `getSymbolTableNameTxn`) ship untested on reachable enterprise `TO PARQUET` data paths — one of them empirically proven to be unguarded by the current test. Under the test gate this cannot be approved as-is.
- **Correctness & performance gate:** passes — no confirmed correctness, resource, concurrency, or performance/IO finding remains open (the fix is a proven no-op on the writer path and correct on the reader path).
- **Test gate:** fails — 2 UNTESTED Critical rows (C1, C2) for changed, reachable, user-visible behavior. Blocking.
- **The ask is cheap and belongs in this PR:** extend the one regression test to (a) drive a re-keyed column with a non-zero column top and (b) reach the symbol branch with a dense≠writer symbol column, and strengthen `parquetLen > 0` into a parquet→native round-trip data assertion (C3) — this single stronger test closes C1, C2, and C3 together.
- Coverage map: **3 behavioral changes, 1 tested, 2 UNTESTED.**
- Findings: **6 verified** (3 Critical, 1 Moderate, 1 Minor, plus 1 empirically-confirmed test-efficacy positive), **3 false positives removed**.
- Split: **all findings in-diff**; 0 out-of-diff, 0 pre-existing. The cross-context pass cleared the sole in-repo callsite (`TableWriter.produceParquetFromNative`) as SAFE and found no broken external callers — expected here since the true beneficiary (enterprise storage-policy conversion) lives out of this repo.

RaphDal:
# Review: PR #7363 — `chore(core): resolve column files by writer index when converting a partition to Parquet`

**Reviewed at level 3** (full mission-critical pass). Diff: 2 files, +86/−3, Java only, no Rust, no on-disk format change. Verification included an **empirical per-line revert-and-run of all three changed lookups**, four independent review agents (two fresh-context adversarial, one test-efficacy, one correctness/cross-context), and direct source tracing.

## Context: this supersedes the prior "request changes"

An earlier review requested changes on three Critical test-coverage gaps (C1: `getColumnTop` untested; C2: `getSymbolTableNameTxn` untested; C3: assertion only checked `parquetLen > 0`). The current diff replaces the single test with **three separate tests** — one per `_cv` lookup — each decoding the parquet back via `read_parquet(...)` and asserting real values. **All three C-findings are confirmed resolved** (see Coverage map, empirically verified). There are **no Critical, no Moderate findings**. This is an approve.

## What the change does (verified correct, not assumed)

`produceParquetFromNative` iterated a partition's columns by dense metadata index and read three column-version (`_cv`) entries with that same dense index. `_cv` is keyed by **writer index** (`ColumnVersionReader.getRecordIndex` matches `COLUMN_INDEX_OFFSET`, `ColumnVersionReader.java:200-221`). For a `TableReader`'s compacted metadata, `ALTER COLUMN TYPE` / `DROP COLUMN` makes dense ≠ writer, so the lookups miss. The fix resolves all three by `metadata.getColumnMetadata(columnIndex).getWriterIndex()` — which is exactly `metadata.getWriterIndex(columnIndex)` (`AbstractRecordMetadata.java:102`), i.e. a faithful mirror of `TableReader` (`TableReader.java:371, 1287, 1753, 1793, 1991`).

Correctness confirmed four ways:

- **Right key.** `TableWriter.changeColumnType` writes the re-keyed `_cv` entry under the *new* writer index (`TableWriter.java:1219-1224`), so `getWriterIndex()` (current), not `getOriginalWriterIndex()` (chain head), is the correct lookup key. `columnId = getOriginalWriterIndex()` is a separate concern (parquet field id) and correctly left unchanged.
- **Complete, not partial.** The three `_cv` calls are the *only* writer-keyed lookups in the loop. Every other `columnIndex` use is legitimately dense — `getColumnType` / `getColumnName` / `getParquetEncodingConfig`, and the symbol provider (`getSymbolMapReader(columnIndex)` is dense-indexed, `TableUtils.java:2946-2952`). Nothing was missed or over-corrected.
- **No-op for the only OSS caller.** `TableWriter.produceParquetFromNative` (`TableWriter.java:11269`, the `CONVERT PARTITION TO PARQUET` path) passes writer metadata; `TableWriterMetadata.removeColumn` only `markDeleted()`s (line 304), never compacts, so dense == writer and the substitution is a proven no-op. The change bites only when a caller passes a reader's compacted metadata.
- **Matches the codebase convention.** Every other reader-metadata `_cv` caller already converts to writer index (`TableReader`, `MetadataCache.java:697`, `TableSnapshotRestore.java:1339`, `VacuumColumnVersions.java:177`). `produceParquetFromNative` was the lone outlier passing a dense index to a reader-keyed `_cv` — precisely the defect this corrects.

## Minor

1. **Test-fidelity: parquet artifacts land in the live native partition directory — in-diff (test).** `produceParquetForPartition` passes `partitionNameTxn` for *both* the native-source `partitionNameTxn` and the parquet-destination `parquetNameTxn` args (`AlterTableChangeColumnTypeTest.java:1791-1792`), whereas the real caller passes a fresh `getTxn()` for the destination (`TableWriter.java:1685`). It's harmless to what's under test — source `_cv` resolution is independent of the destination name-txn, and the file is read back from the same reconstructed path — but it writes `data.parquet` / `_pm` *inside* the live native partition dir instead of a fresh txn-suffixed dir, less faithful to production layout. Both fresh-context agents independently flagged this. Optional: pass a distinct value (e.g. `reader.getTxn()`) as `parquetNameTxn`.

2. **Comment length on the three tests — in-diff (test).** Each carries a 5-6 line doc comment and the "no OSS SQL statement reaches this" rationale is repeated across all three. Per the concise-comment guideline (deep rationale lives in the PR body, where it already is), trim to a one-line intent per test and keep the fullest explanation in one place.

3. **PR title states internal mechanics, not user impact — metadata.** "resolve column files by writer index when converting a partition to Parquet" is implementation detail; the release-note description should speak to end-user impact, e.g. `fix Parquet partition conversion failing after a column type change`. Low stakes for a `chore` that by design produces no OSS release-note entry. `chore` vs `fix` is *not* flagged — the body's "no OSS SQL repro" argument makes `chore` defensible and the type is the author's call.

**Optional (not filed):** `metadata.getColumnMetadata(columnIndex)` is called 3× in the loop (`:1875, :1879, :1884`); hoisting to one local reads cleaner but is a trivial array load the JIT already CSEs. The fix could also write `metadata.getWriterIndex(columnIndex)` directly for the same result — micro-style; the current form is fine.

## Downgraded (candidate concerns verified and cleared)

- **Writer/CONVERT-path regression** → false positive. `TableWriterMetadata` never compacts (`markDeleted` only) → dense == writer → no-op. Verified.
- **A test that doesn't discriminate its line** → disproven empirically. Each of the three reverts fails exactly its own test (below).
- **Other mis-keyed `_cv` lookups in the loop** → none; symbol provider correctly dense, `columnId` correctly `getOriginalWriterIndex()`.
- **Timestamp `writerIndex`-vs-`columnId` inconsistency (`:1858-1860` vs `:1875`)** → not a bug; the designated timestamp cannot be re-keyed (`SqlCompilerImpl.java:1074`), so `writerIndex == originalWriterIndex` for it. Pre-existing, unchanged, safe.
- **Reverse parquet→native path / bloom-filter derivation carrying the same confusion** → cleared; both use a distinct field-id / descriptor-ordinal index space, not the reader `_cv` key.
- **Performance / resource leak** → neutral. Cold path (once per column per conversion), no per-row cost, no new allocation; `assertMemoryLeak` reported **clean on both throw-path reverts** (the `MappedMemoryPartitionDescriptor` try-with-resources unmaps on throw; the helper frees `bloomIndexes` in `finally`).

## Coverage map

All failure links are **empirically verified** (reverted the one line, recompiled, ran the test):

| Behavioral change | Test | Failure link (revert-and-run) | Dimensions | Verdict |
|---|---|---|---|---|
| `getColumnNameTxn` by `writerIndex` (`:1881`) | `…ResolvesReKeyedColumnByWriterIndex` | Reverted → `CairoException: could not open … /x.d` (should be `x.d.<txn>`) | happy ✓; error ✓; NULL N/A (name-txn `-1` path unchanged); boundary ✓ (re-key); cleanup ✓ | **TESTED** |
| `getColumnTop` by `writerIndex` (`:1882`) | `…ResolvesColumnTopByWriterIndex` | Reverted → decodes `{100,200,0,0}` ≠ `{null,null,100,200}` (assertion fails) | happy ✓; data-integrity ✓; NULL ✓ (column-top nulls asserted); boundary ✓ (top>0 via ADD+DROP); cleanup ✓ | **TESTED** |
| `getSymbolTableNameTxn` by `writerIndex` (`:1907`) | `…ResolvesSymbolFilesByWriterIndex` | Reverted → `CairoException: offset file does not exist … /k.o` | happy ✓; error ✓; NULL N/A (symbol txn); cleanup ✓ | **TESTED** |

Baseline (all lines fixed): **3/3 pass**. After restoring: **3/3 pass**. Totals: **3 behavioral changes, 3 tested, 0 UNTESTED.**

## Summary

- **Verdict: approve.** The production fix is correct, minimal, complete, and a proven no-op on the only OSS caller; it faithfully mirrors the authoritative `TableReader` `_cv` access. The prior review's three blocking test gaps are resolved and **empirically proven** — each of the three changed lines is uniquely guarded by a regression test that fails when and only when that line is reverted.
- **Correctness / perf / resource gate:** passes — no open correctness, concurrency, performance, or leak finding.
- **Test gate:** passes — 3/3 behavioral changes tested with verified failure links; error and NULL/column-top paths covered.
- **Tradeoff to note:** the change is a no-op for every current OSS caller and is exercised only by the new synthetic test driving `produceParquetFromNative` with a reader's compacted metadata (the shape the storage-policy conversion uses). That's intended shared-function hardening and the PR body states it plainly.
- **Findings:** 3 verified (all Minor: 2 test, 1 metadata). ~6 candidate concerns verified and dropped as false-positive / pre-existing-safe.
- **Split:** all 3 findings **in-diff**; **0 out-of-diff, 0 pre-existing.** A legitimate zero — the callsite inventory is one OSS caller (verified no-op) plus the test, and a repo-wide sweep of every reader-metadata `_cv` caller found them all already correct, so there is no out-of-diff breakage or pulled-in pre-existing bug.

bluestreak01 (changes requested):
@RaphDal

Requesting changes to address the two **Moderate** items below (the `colMeta` hoist and the `DebugUtils.reconcileColumnTops` one-liner) in this PR before merge. Full review follows.

---

# Review: PR #7363 — `chore(core): resolve column files by writer index when converting a partition to Parquet`

**Reviewed at level 3** (full pass; all 13 agent roles ran inline with per-finding source verification and empirical regression-test runs in a scratch worktree, since removed).

Diff: 2 files, +86/−3, Java only (no Rust → Agent 8 skipped).

---

## Verified core claim

The fix is correct, and every claim in the PR body checked out under adversarial verification:

- **`_cv` is keyed by writer index** — confirmed against `TableReader.java:371` (`getColumnNameTxn(partitionTimestamp, metadata.getWriterIndex(columnIndex))`) and `:1287` (`getSymbolTableNameTxn(metadata.getWriterIndex(columnIndex))`).
- **Reader metadata compacts; writer metadata doesn't** — `TableReaderMetadata` skips deleted slots (`writerIndex < 0` / `columnType <= 0`), so dense ≠ writer after `DROP COLUMN` / `ALTER COLUMN TYPE`. `TableWriterMetadata.reload()` constructs every column with `writerIndex = i` (tombstones kept in place) and `addColumn()` appends at the matching slot, so dense == writer always → the change is a provable no-op on the writer-driven `CONVERT` path.
- **Callsite inventory** (`rg produceParquetFromNative`, repo-wide + enterprise checkout): `TableWriter.java:11269→1685` (SQL CONVERT, writer metadata) — **SAFE**; enterprise `StoragePolicyJob.java:166-203` (passes `reader.getMetadata()` + `reader.getColumnVersionReader()` + `SymbolTableProviderFromReader`) — the broken caller, now **FIXED**; new test helper — SAFE. No overrides (static), no reflection callers. `O3PartitionJob.java:4591` is a comment-only reference; its own encoder path uses writer metadata.
- **Adjacent index spaces stay correct**: `symbolTableProvider.containsNullValue/getSymbolCount(columnIndex)` remain dense — correct for `SymbolTableProviderFromReader` (`reader.getSymbolMapReader` is dense-keyed) and for `FromWriter` (writer space, dense==writer). Bloom-filter index derivation emits descriptor-space indexes (skips deleted columns) — unaffected. The unchanged `timestampIndex` line uses `getWriterIndex()` vs `columnId`'s `getOriginalWriterIndex()`, which cannot diverge for the timestamp column because `ALTER COLUMN TYPE` on the designated timestamp is forbidden (`SqlCompilerImpl.java:1074`, `TableWriter.java:1195`).
- The old dense-index code was worse than the reported file-not-found: a dense `getColumnTop` hit on another column's record silently mis-aligns data (proven empirically below) — the fix closes a **silent-corruption** path, not just an error path.

## Empirical regression-test verification (step 3b.10)

Run in a scratch worktree at PR head `4ad7c2cb`, production hunk reverted via reverse-patch:

| Configuration | ReKeyed test | ColumnTop test | Symbol test |
|---|---|---|---|
| With fix | pass | pass | pass |
| Full production revert | **ERROR** `x.d` not found | **ERROR** `c.d` not found | **ERROR** `k.d` not found |
| `getColumnTop` lookup only reverted | pass | **FAIL** — decodes `{100,200,0,0}` instead of `{null,null,100,200}` | pass |
| `getSymbolTableNameTxn` lookup only reverted | pass | pass | **ERROR** `k.o` not found |

Every per-lookup discrimination claim in the PR body is accurate. The columnTop failure mode confirms the tests assert **values**, not just success — a length-only assertion would have passed on silently corrupted output. Writer-path invariance claim also verified: `ParquetColumnTypeConversionTest` 98/98 pass locally with the fix.

---

## Critical

None.

## Moderate

1. **In-diff, redundant metadata lookups** — `TableUtils.java:1876/1880/1884`: the loop now calls `metadata.getColumnMetadata(columnIndex)` three times per column (`getOriginalWriterIndex()`, `getWriterIndex()`, `getParquetEncodingConfig()`). Hoist once: `final TableColumnMetadata colMeta = metadata.getColumnMetadata(columnIndex);` and derive all three from it. Inertness justification: two extra `ObjList.getQuick` reads per column, once per partition conversion — structurally bounded by column count and dominated by mmap+encode of the whole partition; measurably zero cost, so this is structure/readability, not a blocking perf defect. Apply in this PR.

2. **Pre-existing, same bug class in dead debug code** — `DebugUtils.reconcileColumnTops` (`DebugUtils.java:71-72`) passes the dense reader index `c` to `columnVersionReader.getColumnTop(partitionTimestamp, c)` while comparing against the reader's writer-index-derived cached tops — the exact dense-vs-writer confusion this PR fixes. Its only reference is a commented-out assert (`TableReader.java:731`), so it is unreachable today; but whoever re-enables that assert on a table with dropped/re-keyed columns gets false reconciliation failures. One-line fix for this PR: `columnVersionReader.getColumnTop(partitionTimestamp, reader.getMetadata().getWriterIndex(c))`. Inertness justification: zero live callsites — behavior cannot change on any reachable path.

## Minor

None. (Branch commits use Conventional-Commits prefixes and lack bodies, contrary to the commit conventions — but PR history is throwaway under squash-merge and the squashed message takes the PR title/body, so this is moot by the project's own rules.)

- **"Title should be `fix(core):` since the PR carries a Bug label"** — dismissed: repo precedent uses `chore(core)` for enterprise-only-reachable fixes with no OSS user-visible surface (e.g. `6c44674e91 chore(core): fix wrong symbol values on replicas…`, `88f051feef chore(core): allow ADD INDEX over a parquet partition…`), and the bug is verified unreachable from OSS SQL.
- **"`timestampIndex` (writer index) can mismatch `columnId` (original writer index) for a re-keyed timestamp column"** — dismissed: designated timestamp cannot be re-keyed (`SqlCompilerImpl.java:1074`, `TableWriter.java:1195` both forbid it).
- **"`inputRoot = root` leaks into subsequent tests (never reset)"** — dismissed: established pattern in 9+ test classes (`ExplainPlanTest`, `ParallelFilterTest`, …) and it points at the per-run test root; harmless.
- **"Lone `assertQuery` wrapped in `assertMemoryLeak` anti-pattern"** — dismissed: each lambda holds DDL, inserts, `drainWalQueue`, reader work and the encode call; the wrapper is the legitimate multi-statement case and `.noLeakCheck()` correctly avoids the nested leak check.

## Coverage map

| # | Change (symbol + path) | Test | Failure link | Dimensions | Verdict |
|---|---|---|---|---|---|
| 1 | `produceParquetFromNative`: `getColumnNameTxn` by writer index | `testProduceParquetFromNativeResolvesReKeyedColumnByWriterIndex` | Revert → opens `x.d` instead of `x.d.<txn>` → CairoException (**empirically confirmed**) | happy ✓, error path ✓ (revert = old error), NULL N/A (no null data in scenario; covered by row 2), boundary ✓ (sealed 2-row partition), concurrency N/A (static pure fn over reader snapshot), resources ✓ (`assertMemoryLeak`) | **TESTED** |
| 2 | `produceParquetFromNative`: `getColumnTop` by writer index | `testProduceParquetFromNativeResolvesColumnTopByWriterIndex` | Revert → dense index reads dropped column's top (0) → decodes `{100,200,0,0}` ≠ `{null,null,100,200}` (**empirically confirmed**) | happy ✓, NULL ✓ (asserts null rows above column top), error N/A (this lookup's failure mode is wrong data, asserted), concurrency N/A, resources ✓ | **TESTED** |
| 3 | `produceParquetFromNative`: `getSymbolTableNameTxn` by writer index | `testProduceParquetFromNativeResolvesSymbolFilesByWriterIndex` | Revert → wrong txn → missing `k.o` → CairoException (**empirically confirmed**) | happy ✓, error path ✓, NULL N/A (symbol values non-null; null-flag handled via dense provider, unchanged), concurrency N/A, resources ✓ | **TESTED** |
| 4 | Writer-driven CONVERT path | `ParquetColumnTypeConversionTest` (98/98, run locally) + rest of `AlterTableChangeColumnTypeTest` | n/a — verified **no behavioral change**: `TableWriterMetadata` sets `writerIndex == slot index` and keeps tombstones, so resolved index is identical | — | **EXEMPT** (verified delta: no-op) |

All preconditions inside the tests (`assertNotEquals(dense, writer)`, differing `_cv` reads per index) guard against the tests going vacuous if metadata compaction behavior ever changes — good hardening, verified non-vacuous.

## Summary

- **Verdict: request changes** — please address the two Moderate items (the `colMeta` hoist and the `DebugUtils.reconcileColumnTops` one-liner) in this PR; neither touches a reachable correctness or measurable performance path, per the stated justifications, and both are small, zero-risk edits.
- Coverage map: 3 behavioral changes, 3 tested (all with empirically verified failure links), 0 UNTESTED, 1 exempt row with verified no-behavioral-change delta. Test gate satisfied: this is a fix PR **with** regression tests proven to fail without the fix — 4 build/test cycles (with fix, full revert, two single-lookup reverts) all matched the PR body's predictions exactly.
- Findings: 6 draft findings → 2 confirmed, 4 dropped as false positives. Split: 1 in-diff (Moderate), 0 out-of-diff broken callsites (all 3 callsites walked with explicit verdicts: TableWriter SAFE, enterprise StoragePolicyJob FIXED, test SAFE), 1 pre-existing (Moderate, unreachable dead debug code, one-line fix proposed for this PR).
- Tradeoffs/regressions: none found — the writer path is analytically and empirically unchanged; the fix additionally closes a silent data-mis-alignment path (dense `getColumnTop` hit on the wrong column) that the PR body undersells as merely a file-not-found error.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 13 / 13 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/TableUtils.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/cairo/DebugUtils.java|3|3|100.00%|

bluestreak01 (approved):
# Review: PR #7363 — `chore(core): resolve column files by writer index when converting a partition to Parquet`

**Reviewed at level 3** (full mission-critical pass: change-surface map, coverage map, all agent roles inline with per-finding source verification, and **empirical revert-and-run of every changed line** with a warm Java-25 build). Diff: 3 files, +153/-4, Java only (no Rust -> Agent 8 skipped).

This PR has already been through two review rounds; the current branch state incorporates the prior fixes (the `tableColumnMetadata` hoist, the `DebugUtils.reconcileColumnTops` one-liner, and dedicated tests for all three lookups). I reviewed the **current** state from scratch and verified every claim against source and by running the tests.

## What the change does

`TableUtils.produceParquetFromNative` iterated a partition's columns by **dense metadata index** and read three column-version (`_cv`) entries — name txn, column top, symbol-table name txn — with that dense index. `_cv` is keyed by **writer index**. In a `TableReader`'s compacted metadata (tombstones dropped), `ALTER COLUMN TYPE`/`DROP COLUMN` make dense != writer, so those lookups miss and the encoder opens `x.d` instead of `x.d.<txn>` and throws. The fix resolves the three lookups via `tableColumnMetadata.getWriterIndex()`, sharing one `getRecordIndex` between the name-txn and column-top reads. `DebugUtils.reconcileColumnTops` gets the same dense->writer correction.

## Verified core mechanics (source-traced, not assumed)

- **Faithful mirror of `TableReader`.** The name-txn block (`TableUtils.java:1884-1887`) is byte-for-byte the reader's own inline resolution at `TableReader.java:1794-1801` — `getColumnNameTxnByIndex(versionRecordIndex)` then `if (== -1) getDefaultColumnNameTxn(writerIndex)`. The symbol-txn move (`:1907`) mirrors `TableReader.java:1287`. The column-top (`:1888`) keeps the original method's `-1L` default (`getColumnTopByIndexOrDefault(..., -1L)`), so `columnRowCount` semantics are unchanged; it just corrects the index and reuses the shared record.
- **Writer/CONVERT path is a proven no-op.** `TableWriterMetadata.reload()` assigns `writerIndex = i` for every slot including tombstones and never compacts (`:182,195`), so dense == writer for the OSS `TableWriter.produceParquetFromNative` caller (`:11271`). The change cannot alter its output.
- **Adjacent index spaces are correctly left dense.** The symbol provider (`containsNullValue`/`getSymbolCount`) delegates to `reader.getSymbolMapReader(columnIndex)` which is dense-keyed (`TableReader.java:559-561`); `columnId = getOriginalWriterIndex()` (the stable Parquet field id) correctly stays put; `deriveBloomFilterColumnIndexes` emits descriptor ordinals (`:813-826`), a separate self-consistent space. The three `_cv` calls were the only mis-keyed lookups.
- **Callsite inventory (exhaustive, tiny):** `produceParquetFromNative` — `TableWriter` (writer meta, SAFE/no-op), the new test helper (SAFE), and the out-of-repo Enterprise storage-policy path (the broken caller, now FIXED). `reconcileColumnTops` — one commented-out assert at `TableReader.java:731` (dead) + the new test.

## Empirical per-line discrimination (level-3 step 3b.10)

I reverted each changed production line individually against the current tests:

| Reverted line | ReKeyed | ColumnTop | Symbol | Reconcile |
|---|---|---|---|---|
| none (HEAD) | pass | pass | pass | pass |
| name-txn -> dense | **FAIL** `x.d` | **FAIL** `c.d` | **FAIL** `k.d` | — |
| column-top -> dense | pass | **FAIL** value `{100,200,0,0}` | pass | — |
| symbol-txn -> dense | pass | pass | **FAIL** `k.o` | — |
| DebugUtils -> dense | — | — | — | **FAIL** reconcile=false |

Every changed line is guarded by a discriminating test. The column-top test fails on a **decoded-value mismatch**, not a throw — so it catches silent data mis-alignment, not just file-not-found. Working tree restored clean afterward (verified: no diff vs HEAD, no leftover markers).

Note on the CI coverage bot reporting `DebugUtils.java 0/3 (00.00%)`: that metric is stale/incomplete. `testReconcileColumnTopsResolvesByWriterIndex` genuinely exercises the changed line — reverting it flips reconciliation to `false` and fails the test. Not a real coverage gap.

## Critical

None. The three prior-round blocking items (untested column-top, untested symbol branch, assertion-only `parquetLen > 0`) are all resolved: each lookup now has a discriminating test and each test decodes the parquet back via `read_parquet(...)` and asserts values.

## Moderate

None. The prior-round Moderates (the `getColumnMetadata` triple-lookup hoist and the `DebugUtils` dense-index one-liner) are both applied in the current diff.

## Minor

1. **PR title leans on an internal detail — metadata.** "resolve column files by writer index" describes the mechanism, not the user-facing effect; CLAUDE.md asks titles (which feed release notes) to speak to impact, e.g. "fix Parquet conversion failing after a column type change". Defensible for a `chore` with no OSS-visible surface, but worth a reword. `chore(core)` vs the `Bug` label is a judgment call — there's repo precedent for `chore` on enterprise-only-reachable fixes, and the PR body is exemplary (accurate root cause, honest that it's a no-op for OSS callers, per-lookup test plan).

2. **`testReconcileColumnTopsResolvesByWriterIndex` is mildly fragile — in-diff (test).** It passes a 1-element `LongList` with `partitionsSlotSize=1` while the table has two partitions; `reconcileColumnTops` iterates all partitions and would index past the list at `openPartitionInfo.getQuick(1)`. It's safe only because the test opens partition 0 alone, so partition 1's size stays `-1` (`TableReader.java:1239`) and the `if (partitionRowCount != -1)` guard skips it before the access. Behaviorally correct as written (whitebox is the only option for dead debug code), but adding the second partition's timestamp to the list — or a one-line comment noting the guard — would make the invariant explicit and refactor-proof. Zero behavioral/performance dimension -> Minor.

- **Name-txn sentinel divergence** — I suspected the inlined `if (columnNameTxn == -1) -> getDefaultColumnNameTxn` diverges from `ColumnVersionReader.getColumnNameTxn` (which returns a stored `-1` without falling back). It does diverge from that *method*, but it's an exact copy of what `TableReader` itself does when opening the column file (`:1796-1800`), and matching the reader is precisely the goal. Correct.
- **Other mis-keyed lookups in the loop** — none; symbol provider (dense), `columnId` (originalWriterIndex), bloom indexes (descriptor ordinals) are all correctly not writer-keyed.
- **`timestampIndex` (`getWriterIndex`) vs `columnId` (`getOriginalWriterIndex`) mismatch** — pre-existing, unchanged by this PR, and safe: the designated timestamp cannot be re-keyed (`SqlCompilerImpl.java:1074`, `TableWriter.java:1195`), so the two coincide for it.
- **Resource leak on the pre-fix throw path** — none; the revert runs threw the expected `CairoException` with `assertMemoryLeak` clean (`partitionDescriptor` try-with-resources frees mapped columns, `bloomIndexes` freed in `finally`).
- **DebugUtils test OOB** — not reachable; guarded as above.

## Coverage map

| # | Change (symbol + path) | Test | Failure link (empirically confirmed) | Dimensions | Verdict |
|---|---|---|---|---|---|
| 1 | `produceParquetFromNative`: name txn by writer index (`:1884`) | `testProduceParquetFromNativeResolvesReKeyedColumnByWriterIndex` | revert-> opens `x.d` -> CairoException file-not-found | happy Y, error Y, NULL N/A (covered by row 2), boundary Y (sealed 2-row partition), concurrency N/A (static fn over reader snapshot), resources Y (`assertMemoryLeak`) | **TESTED** |
| 2 | `produceParquetFromNative`: column top by writer index (`:1888`) | `testProduceParquetFromNativeResolvesColumnTopByWriterIndex` | revert-> reads dropped column's top(0) -> decodes `{100,200,0,0}` != `{null,null,100,200}` | happy Y, NULL Y (asserts null rows above top), error N/A (failure mode is wrong data, asserted), concurrency N/A, resources Y | **TESTED** |
| 3 | `produceParquetFromNative`: symbol-table name txn by writer index (`:1907`) | `testProduceParquetFromNativeResolvesSymbolFilesByWriterIndex` | revert-> wrong txn -> missing `k.o` -> CairoException | happy Y, error Y, NULL N/A (symbol values non-null; null-flag via dense provider, unchanged), concurrency N/A, resources Y | **TESTED** |
| 4 | `DebugUtils.reconcileColumnTops`: column top by writer index (`DebugUtils.java:72`) | `testReconcileColumnTopsResolvesByWriterIndex` | revert-> dense index reads dropped column's top -> reconcile returns false -> assertTrue fails | happy Y, concurrency N/A, resources Y | **TESTED** |
| 5 | Writer-driven CONVERT path | `ParquetColumnTypeConversionTest` + rest of class | verified **no behavioral change**: `TableWriterMetadata` sets `writerIndex == slot` and keeps tombstones -> resolved index identical | — | **EXEMPT** (verified no-op delta) |

## Summary

- **Verdict: approve.** Correct, complete, empirically regression-proof fix that mirrors the authoritative `TableReader` `_cv` access pattern. No Critical or Moderate findings remain; both prior-round Moderates and all three prior-round Criticals are resolved in the current branch.
- **Correctness & performance gate:** satisfied — zero open correctness/resource/concurrency findings; the change is a proven no-op on the writer path and a tiny improvement on the reader path (one shared `getRecordIndex` instead of two, one cached metadata lookup instead of three), all on a cold once-per-column-per-conversion path.
- **Test gate:** satisfied — fix PR **with** regression tests, and I empirically proved all four changed lines fail their dedicated test when reverted (5 build/run cycles: HEAD-green + 4 isolated reverts, every result matched the PR body's predictions).
- **Coverage map:** 5 behavioral rows — 4 TESTED (all with empirically verified failure links), 0 UNTESTED, 1 EXEMPT (verified no-op).
- **Findings:** ~7 candidate concerns investigated -> 2 verified (both Minor), 5 dropped as false positives/pre-existing. In-diff / out-of-diff / pre-existing split: **2 in-diff (Minor), 0 out-of-diff, 0 open pre-existing.** The zero out-of-diff is legitimate, not an underrun — the callsite inventory is exhaustive and tiny (one OSS caller verified no-op, one dead debug reference, the out-of-repo Enterprise caller now fixed), each explicitly walked.
- **Tradeoff to note:** the change is a no-op for every current OSS caller; its real beneficiary is the out-of-repo Enterprise storage-policy `TO PARQUET` path, exercised here only through the new synthetic reader-metadata tests. That's intended shared-function hardening and the PR states it plainly.

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=250018)
**Tested Commit:** `66b32d1e28085990543164504fd66f04a47d2166`

PR #7364, opened 2026-07-07
test(qwp): speed up sender coercion-error tests

RaphDal:
## Summary

`QwpSenderE2ETest`'s 20 `testCoercionTo*Errors` methods spend almost all of their time waiting rather than asserting. This change pins the poison-frame detector's minimum escalation window to `0` for the error-path sender helper, so a deterministic rejection escalates to its terminal on the first strike instead of after a fixed ~5s wall-time gate. The 20 tests drop from ~19 min to under a second combined, and the full `cutlass.qwp.**` package wall-clock drops from ~22.5 min to ~2.5 min locally, with the same 1333 tests passing.

## Background

`AbstractQwpWebSocketTest.connectWs(port, handler, maxFrameRejections)` — used only by `assertCoercionError` — builds the sender for the coercion error-path tests. Its javadoc states that pinning `max_frame_rejections=1` makes a deterministic rejection "escalate on the first strike." In practice it did not.

The poison-frame detector also enforces `poison_min_escalation_window_millis`, whose default (`CursorWebSocketSendLoop.DEFAULT_POISON_MIN_ESCALATION_WINDOW_MILLIS`) is 5000. Until that wall-time window elapses, the client reconnect-replays the doomed frame with capped exponential backoff (100 ms .. 5 s). So each rejected sub-case waits out ~5s plus reconnect overhead before the terminal the test asserts arrives. With ~8-13 sub-cases per method, one `testCoercionTo*Errors` method ran 40-100 s.

These 20 methods were the dominant cost of the mac "other" CI shard, which runs `cutlass.qwp.**`: they accounted for ~17-19 min on their own and were a large part of that shard reaching the 55-minute pipeline timeout.

## Change

One functional line in the error-path helper (plus an explanatory comment):

```java
.maxFrameRejections(maxFrameRejections)
.poisonMinEscalationWindowMillis(0)   // <-- added
.closeFlushTimeoutMillis(CLOSE_FLUSH_TIMEOUT_MS)
```

`poison_min_escalation_window_millis = 0` is the documented "legacy immediate escalation" setting. Combined with the existing `max_frame_rejections = 1`, the first strike escalates to the same terminal category and message the tests already assert.

## Impact

Full `cutlass.qwp.**` package, local (Linux, mvnd, parallel forks):

| metric                                 | before       | after        |
| -------------------------------------- | ------------ | ------------ |
| package wall-clock                     | 22:37 min    | 02:29 min    |
| 20 `testCoercionTo*Errors` (sum)       | 1130 s       | 0.4 s        |
| `QwpSenderE2ETest` class (130 methods) | 20.1 min     | 6.0 s        |
| tests run / failures / skipped         | 1333 / 0 / 1 | 1333 / 0 / 1 |

Per method, the coercion error tests fall from 40-100 s to ~0.02 s each. These are local numbers on parallel forks, not the single-fork mac CI runner, so the absolute times differ; the saving is the removal of a wall-time wait, which transfers to CI.

## Tradeoffs and scope

- These tests no longer traverse the reconnect-replay-under-backoff path on the way to escalation; they escalate on the first strike. That path stays covered by the dedicated poison/restart tests (`QwpIngressOracleFuzzTest.testOraclePoisonRowsEscalateToPoisonTerminal`, `QwpIngressServerRestartFuzzTest`). The coercion tests only ever asserted the terminal category and message, which are unchanged.
- Test-only change. The client's production default escalation window (5 s) is untouched.
- It does not address why `max_frame_rejections=1` alone does not already short-circuit the window. That may be worth a separate look at the client behaviour, but is out of scope here.

## Test plan

- Ran the full `cutlass.qwp.**` package before and after the change: 1333 tests, 0 failures, 1 skipped in both runs.
- Confirmed all 20 `testCoercionTo*Errors` methods still pass with the window pinned to 0.
- Confirmed the diff is a single test-support file with one functional line.


RaphDal:
# Review: PR #7364 — `test(qwp): speed up sender coercion-error tests`

**Reviewed at level 3** (full pass: change-surface map, two parallel review agents
including the mandatory fresh-context adversarial agent, per-finding verification,
plus an empirical test run).

## Overview

The diff is a single functional line — `.poisonMinEscalationWindowMillis(0)` added to
the error-path `connectWs(int, SenderErrorHandler, int)` helper — plus a 6-line
explanatory javadoc.

Verified end-to-end:

- The builder method exists: `Sender$LineSenderBuilder.poisonMinEscalationWindowMillis(long)`
  in the cached `questdb-client-1.3.6-SNAPSHOT.jar` (the version the build resolves).
- The documented default is real: `DEFAULT_POISON_MIN_ESCALATION_WINDOW_MILLIS = 5000L`.
- Escalation is an AND-gate of strike-count *and* wall-time window
  (`recordHeadRejectionStrike`: `poisonStrikes >= maxHeadFrameRejections`
  **and** `now - poisonFirstStrikeNanos >= poisonMinEscalationWindowNanos`).
- `haltOnPoisonedFrame` builds `SenderError(PROTOCOL_VIOLATION, TERMINAL, …)` from
  compile-time constants, so window=0 changes *when* escalation fires, never *which*
  terminal category/message it produces.
- Blast radius is exactly one callsite: `QwpSenderE2ETest.java:4852`
  (`connectWs(port, handler, 1)` inside `assertCoercionError`).

**Empirical check:** staged a client native lib onto the classpath and ran the tests —
all 20 `testCoercionTo*Errors` pass in **1.78 s** total (0 failures / 0 errors). They
otherwise cannot run in this worktree because the `java-questdb-client` submodule is
not checked out and the client `libquestdb.so` is absent, so `.build()` fails before
any poison logic; the staged lib was removed afterward. CI has the real client build.

## Critical

None.

## Moderate

### 1. The edited javadoc now contradicts itself (in-diff, documentation)

`core/src/test/java/io/questdb/test/cutlass/qwp/e2e/AbstractQwpWebSocketTest.java:104-114`

The pre-existing first paragraph (lines 104-108) says passing `max_frame_rejections=1`
makes a rejection escalate *"to its PROTOCOL_VIOLATION terminal **on the first strike**
instead of reconnect-replaying the frame."* The newly-added second paragraph
(lines 110-114) says that without the window pinned to 0 *"the detector waits out its
default 5s wall-time gate before escalating ... reconnect-replaying the doomed frame
the whole time."*

Both describe the same configuration this helper always uses. They cannot both be
true — and the new paragraph is the correct one (it is the entire reason this PR
exists: `max_frame_rejections=1` was already set, yet each sub-case still cost ~5 s).
The diff appended the correct description but left the now-false sentence standing.

**Fix:** reword lines 104-108 so the count gate no longer claims immediate / no-replay
escalation on its own — e.g. "passes `1` so the strike-count gate is met on the first
rejection; combined with the window pinned to `0` below, the terminal is reached
immediately instead of after the wall-time gate." Keep it short, per repo comment
conventions.

### 2. Two sibling comments repeat the same now-disproven claim (out-of-diff; fix in this PR)

- `core/src/test/java/io/questdb/test/cutlass/qwp/e2e/QwpSenderE2ETest.java:4834-4836`
  — the caller `assertCoercionError`: *"We pin max_frame_rejections=1 so the escalation
  is immediate."* Immediacy actually comes from the `window=0` this PR adds in
  `connectWs`; this comment does not mention it and is now misleading for the exact
  helper the PR changed.
- `core/src/test/java/io/questdb/test/cutlass/qwp/e2e/QwpIngressOracleFuzzTest.java:557-559`
  — *"max_frame_rejections=1 escalates the first rejection straight to the poison
  terminal instead of reconnect-replaying the frame."* Inaccurate: that test keeps the
  **default 5000 ms** window and *does* reconnect-replay for ~5 s before escalating (it
  waits up to 120 s for the terminal).

These are out-of-diff, but they share one root misconception that this PR definitively
corrects, and #1's caller is one of them. Bundling the reconciliation here (the repo
prefers bundling related fixes) avoids leaving three contradictory statements about the
same mechanism across the suite. Not blocking.

## Minor

- **PR label.** Labeled `Performance`; this is a test/CI wall-time reduction, not a
  product-performance change. `CI` (and/or `Flaky Test`) is more accurate and avoids a
  release-notes reader mistaking it for a runtime improvement. Optional.
- **`assertCoercionError`'s fsn-span sub-assertion runs slightly less often.** With
  window=0 the terminal latches sooner, so `flushAndGetSequence()` loses the race and
  returns `publishedFsn = -1` a bit more often, skipping the span check at
  `QwpSenderE2ETest.java:4880-4884`. The byte-fragmented loopback round-trip still
  dominates, so `publishedFsn >= 0` should hold nearly always; the category/message
  assertions are unaffected. Negligible, noted for completeness.

## Downgraded (false positives / considered-and-cleared)

- **Ordering race — `firstErrFut` could capture the TERMINAL instead of the
  informational RETRIABLE strike under window=0.** Dismissed: the client dispatches the
  informational strike (`dispatchError`) *before* `haltOnPoisonedFrame` (confirmed by
  reading the client source in a sibling checkout), and empirically all 20 tests pass,
  including WRITE_ERROR-branch cases that assert `firstErrFut` is `RETRIABLE WRITE_ERROR`
  (`QwpSenderE2ETest.java:4872-4876`) — those would fail deterministically if the
  terminal were dispatched first. This was the adversarial agent's flagged "crux"; it is
  verified, not open.
- **window=0 could yield a different terminal category.** Dismissed: `haltOnPoisonedFrame`
  builds `SenderError(PROTOCOL_VIOLATION, TERMINAL, …)` from constants; the window gates
  timing only (bytecode-confirmed).
- **Coverage silently reduced.** Dismissed:
  `QwpIngressOracleFuzzTest.testOraclePoisonRowsEscalateToPoisonTerminal` sets
  `max_frame_rejections=1` but leaves the window at the default 5000 ms (`:565`, `:706`)
  and asserts the `PROTOCOL_VIOLATION` poison terminal — it genuinely traverses the
  reconnect-replay-under-backoff path the coercion tests drop. Neither coverage test pins
  the window to 0.
- **Store-and-forward "never hard-fail on transient outage" invariant.** Dismissed: the
  poison-escalation window governs *deterministic-rejection* (poison-frame) handling,
  which is supposed to terminate — distinct from transient-outage reconnect (unbounded,
  non-failing). Production default (5000) untouched; test-only; the poison terminal is a
  sanctioned terminal.
- **Non-ASCII introduced.** Dismissed: added lines are pure ASCII (`~5s` is an ASCII
  tilde). The em-dashes at lines 88/129/132 are pre-existing, out-of-diff, and in javadoc
  (not log messages).
- **Reduced test power (skips the byte-identical-replay determinism check).** Real but
  explicitly acknowledged in the PR and mitigated: the windowed replay-then-escalate path
  stays covered by the Oracle fuzz test above. Not a defect in the change.

## Summary

**Verdict: approve.** The change is correct, minimal, and does exactly what it claims —
the 20 coercion-error tests drop from ~5 s/sub-case to immediate escalation with
identical assertions, verified empirically (20/20 in 1.78 s) and mechanistically (the
window gates timing, not the asserted category/message). The one real issue is
documentation: the PR leaves the method's own javadoc self-contradictory and does not
reconcile two sibling comments that state the same disproven claim. All findings are
non-blocking.

- **6 draft findings verified, ~7 dropped** as false positives after
  source / bytecode / empirical verification.
- **In-diff vs out-of-diff:** 1 in-diff (javadoc self-contradiction) + 2 minor in-diff
  behavioral notes; 2 out-of-diff (sibling comments). The change surface is genuinely
  tiny — one line, one callsite — so the small out-of-diff count is expected, not an
  under-run.
- **Tradeoff, stated plainly:** the coercion tests no longer exercise the
  reconnect-replay-under-backoff escalation path; this is covered by
  `QwpIngressOracleFuzzTest` / `QwpIngressServerRestartFuzzTest`, which were confirmed to
  use the default 5 s window.

bluestreak01 (approved):

PR #7366, opened 2026-07-07
fix(qwp): fix duplicate writes and store-and-forward recovery during failover

bluestreak01:
> [!IMPORTANT]
> **Tandem PR — merge all three together.**
> - Server: questdb/questdb#7366
> - Enterprise: questdb/questdb-enterprise#1113
> - Client: questdb/java-questdb-client#73
>
> questdb/questdb#7366 pins the client branch through its `java-questdb-client` submodule, and questdb/questdb-enterprise#1113 pins the server branch in turn. All three use the branch name `fix/qwp-deferred-close-echo-wait`. Merging the server PR against `java-questdb-client` `main` instead of questdb/java-questdb-client#73 reintroduces the ROLE_CHANGE (4001) misclassification.

## Problem

During a replica role change, QWP can flush the final durable ACK and send a WebSocket CLOSE, then tear down TCP before the store-and-forward client consumes that tail. Unread client data can turn the close into an RST. The client then reconnects with a stale replay watermark and can resend rows already present on the promoted replica.

Enterprise CI exposed this as 700 duplicate rows in `SqlFailoverQwpDeferredCloseExactlyOnceTest` (`expected:<2800> but was:<3500>`).

The review also found related defects in close-frame resumption, parser fairness, client shutdown, and store-and-forward recovery. This PR keeps those fixes together because they share the same QWP failover and CI path.

### Server-side role-change handoff

- Wait for the RFC 6455 CLOSE echo after the final durable ACK instead of immediately closing TCP when durable coverage is complete.
- Send the private-use `ROLE_CHANGE` close code (4001) and accept only the matching echo as delivery confirmation. A voluntary `NORMAL_CLOSURE` already queued by the client is treated as a crossing CLOSE and reports that replay may duplicate.
- Keep client-initiated CLOSE responses separate from server-initiated role-change CLOSE state so a parked response cannot incorrectly arm the echo wait.
- Preserve the final durable ACK when client CLOSE handling encounters send backpressure.
- Prevent PONG, error, ACK, and additional CLOSE frames after the server CLOSE.
- Discard inbound data while waiting, poll expiry on every receive re-entry, and bound work by frame count, byte count, and grace time. A known legal large frame can consume the byte budget without crossing its frame boundary, avoiding thousands of small dispatcher reads while preserving fairness.
- Reuse the durable-progress result during role-change close processing instead of repeating registry scans.

### Store-and-forward client lifecycle and recovery

- Keep slot ownership until the segment manager and worker become quiescent, including interrupted shutdown paths.
- Make close-lifecycle interrupt tests coordinate on acknowledged pool retries instead of transient condition-waiter membership, preserving deadline-restart coverage without scheduler races.
- Move delayed engine cleanup to one bounded scheduled daemon with deduplicated retries and a finite retry budget.
- Validate recovered segment contents with positional reads before mmap, retain corrupt files for diagnosis, and fail closed when existing SF segments or directory enumeration cannot be read.
- Map only the validated prefix of recovered sparse segments. The producer rotates to a newly allocated spare before appending, avoiding SIGBUS when disk space is exhausted.
- Preserve file-facade ownership across recovered segments and use logical file size for SF capacity accounting.
- Carry the maximum recovered file generation into registration to avoid a second directory scan.
- Remove fully acknowledged segment prefixes in one range operation instead of repeated front removal.
- Classify `ROLE_CHANGE` as an orderly, reconnect-eligible client close.
- Ensure deferred-close callers release their own buffers and dispatchers after delayed engine shutdown completes.

## Operational effects and tradeoffs

- A covered role-change close can keep the connection open for up to the five-second echo grace. A silent peer may remain until the transport idle reaper instead.
- Third-party clients that normalize or replace close code 4001 take the conservative path: teardown still completes, but QuestDB can report a duplicate-risk warning rather than delivery confirmation.
- Recovery performs additional bounded positional I/O before mapping persisted data. Unreadable existing SF state now prevents startup for that slot instead of silently starting from an empty store.
- A recovered segment is treated as full, so the first append may briefly backpressure while the manager installs a block-reserved spare.
- If a worker remains permanently stuck after the deferred-close retry budget, QuestDB deliberately retains the slot lock and related resources until process exit. This prevents a replacement engine from reusing storage that the stale worker could still mutate; the bounded daemon logs and counts the abandonment.

## Test plan

- `mvn -pl core -Dtest=QwpIngressProcessorStateTest,QwpIngressDeferredCloseDurableAckTest test -DskipITs`
- Run the java-client core tests covering `CursorSendEngineSlotReacquisitionTest`, `CursorDeferredCloseCallerTest`, `CursorWebSocketSendLoopPoisonFrameTest`, `MmapSegmentRecoveryFaultTest`, `SegmentRingEdgeRecoveryTest`, `SegmentRingRecoveryGenerationTest`, and `SegmentRingTest`.
- Run the full java-client core test suite, including `QuestDBImplCloseLifecycleTest`; repeat both acknowledged-interrupt cases 15 times on one CPU; verify a deliberately restarted deadline fails the regression assertion.
- Run `SqlFailoverQwpDeferredCloseExactlyOnceTest` from questdb/questdb-enterprise#1113 on the tandem branch.


bluestreak01:
## Review (tandem with questdb/questdb-enterprise#1113, level 3)

**Verdict: request changes.** The core fix is sound and verified end-to-end — the client (`WebSocketClient.tryParseFrame`) really does echo CLOSE before dispatching `onClose`, the echo proves in-order consumption of the `[durable ack][CLOSE]` tail, and the enterprise exactly-once witness passes 3× with the handshake observed in logs. But the echo-wait gate has holes that violate the PR's own "nothing after CLOSE" invariant on inputs a broken client or fragmenting intermediary can produce during the exact demote window this code serves.

### Moderate

**1. The echo-wait discard gate is per-opcode, and `sendFatalClose`'s changed contract breaks its un-audited callers — a second CLOSE frame can go out after ours.**
The diff changed `sendFatalClose`'s postcondition from "never returns normally" to "returns normally when the echo wait is armed" (the two updated `return;` comments show this was known). But `rejectFragmentedFrame` and `rejectTextFrame` (`QwpIngressUpgradeProcessor.java` ~1472–1496, untouched) still call it with no `isAwaitingCloseEcho()` gate: a TEXT or fragmented-BINARY/CONTINUATION frame arriving inside the echo window — and the code's own log message documents fragmenting proxies/LBs as a real occurrence — writes a **second CLOSE frame** onto the wire after the role-change CLOSE, violating RFC 6455 and the exact invariant `assertCloseIsFinalFrame` pins, and skips the expiry poll. `beginCloseEchoWaitIfEligible` then returns true (all three conditions still hold: `durableAckEnabled` and the deferral deadline reset only on disconnect; `pendingDurableSeqTxns` was pruned by the final ack so coverage is trivially true), so the connection silently stays in the wait.
**Fix:** hoist the gate to the top of `handleWebSocketFrame` — while awaiting the echo, discard every opcode except CLOSE and poll `checkCloseEchoWaitExpiry` on each frame; and/or make `sendFatalClose` expiry-poll-and-return while awaiting.

**2. The oversized-frame paths during the echo wait loop emitting CLOSE frames and are not bounded by the 5 s grace.**
`resumeRecv`'s full-buffer branch and `processWebSocketFrames`' `STATE_NEED_PAYLOAD` too-big branch now call `sendFatalClose`, which returns without disconnecting while the wait is armed — but the offending bytes are **never consumed**, so every subsequent inbound event re-enters the same branch and sends yet another CLOSE frame. `checkCloseEchoWaitExpiry` is unreachable from this cycle, so only peer death or the transport idle reaper ends it. Pre-fix these paths disconnected immediately. Trigger: a broken/hostile client exceeding the negotiated max frame size during a demote window — availability degradation, no data loss.
**Fix:** while awaiting the echo, these two branches should disconnect immediately (the frame cannot be drained, so the RST guard is unattainable anyway) or at minimum poll the expiry.

### Minor

3. PONG and unknown opcodes during the echo wait are consumed without polling the expiry budget (only BINARY/PING poll). Covered by the same centralization as #1.
4. The new `isAwaitingCloseEcho()` guard at the top of `roleChangeCloseWithUploadGrace` is unreachable — both callers (`handleBinaryMessage`, `handlePing`) return earlier when the wait is armed. Dead code today; fine as a defensive guard, but it is not covering anything.
5. The "no acks after CLOSE" invariant holds by data-state, not by construction: `processWebSocketFrames`' trailing `flushPendingAck` is not gated on the echo wait. Provably a no-op today (`highestProcessedSequence == lastAckedSequence` at wait entry; discarded frames advance nothing; durable maps pruned by `onDurableAckSent` + frozen registry), but any future change to those mechanics could emit an ack after CLOSE without tripping anything. Gate it or add an assert.
6. `beginCloseEchoWaitIfEligible` logs "role-change CLOSE sent…" even when entered from a non-role-change fatal close during a deferral window (e.g. a protocol-error close). Cosmetic.

### Verified / downgraded during review
- Echo-before-dispatch client claim: **confirmed** in `java-questdb-client` (`sendCloseFrameEcho` before `handler.onClose`, RFC 6455 §5.5.1).
- "Registry swap at demote breaks the CLOSE-time eligibility re-check": false — enterprise `PrimaryRoleState` installs a *frozen* snapshot precisely for deferred closes, and `onDurableAckSent` prunes `pendingDurableSeqTxns` so the re-check is trivially true after the final ack.
- "A fully silent peer holds the fd forever": false — `AbstractIODispatcher.idleConnectionTimeout` reaps it, as the code comment says.
- "`handleClose` echo path skips a needed ack flush": false — no committed-but-unacked work can exist at echo time.
- `closeEchoDeadline` lifecycle: correct — mirrors `roleChangeCloseDeferredDeadline` exactly (survives `clear()`/`clearMessageState()`, reset in `onDisconnected()` via `LocalValueMap`).
- Test suite: `QwpIngressDeferredCloseDurableAckTest` 8/8 green locally.

bluestreak01:
## Review (level 3, full mission-critical pass)

Tandem with questdb-enterprise#1113. Reviewed as currently merged on `fix/qwp-deferred-close-echo-wait` (HEAD `62ae3d2111`), i.e. *after* the earlier `request changes` review and the follow-up commits that answer it.

### What the PR does and whether it is correct

The core defect is real and the fix is sound. On a deferred role-change close the server previously flushed the final durable ack + `NORMAL_CLOSURE` and immediately closed the fd; with a store-and-forward writer pumping frames, the close turned abortive (RST) and the client kernel discarded the unread `[durable ack][CLOSE]` tail, so the replay watermark never advanced and the client replayed its whole corpus to the promoted replica — duplicates on non-DEDUP tables (enterprise CI witness: `expected:<2800> but was:<3500>`).

Verified end to end:

- **Client contract confirmed.** `java-questdb-client/.../WebSocketClient.java:1222` calls `sendCloseFrameEcho(closeCode)` *before* `handler.onClose(...)` at :1225 (RFC 6455 §5.5.1). The echo proves in-order TCP consumption of everything before the server's CLOSE, the final durable ack included.
- **The prior review's two blocking Moderates are resolved.** The per-opcode gate hole is closed by the hoisted gate at the top of `handleWebSocketFrame` (:1466, discards every non-CLOSE opcode + polls expiry) *and* the `sendFatalClose` echo-wait gate (:1783), so `rejectTextFrame`/`rejectFragmentedFrame` can no longer emit a second CLOSE. The oversized-frame busy-loop is closed by `hasLostCloseEchoSync` + `discardInboundBytes` + `PeerIsSlowToWriteException` re-arm (drains to EAGAIN, so wakeups are one-per-segment).
- **TOCTOU fix is genuinely race-free.** `beginCloseEchoWaitIfEligible` decides from `hasPendingDurableWork()` (local `pendingDurableSeqTxns`), never a fresh registry read; `trySendDurableAck`→`onDurableAckSent` prunes against the T1 snapshot only, so a concurrent (T1,T2] registry advance cannot flip the decision. `flushPendingAck` is additionally gated on `isAwaitingCloseEcho()`, making "no ack after CLOSE" structural.
- **`RESUME_CLOSE` vs `RESUME_CLOSE_RESPONSE` split** is correct: the former returns to READY (`onResumeCloseComplete`) so the fall-through `drainBufferedFrames` dispatches a buffered crossing echo; the latter disconnects unconditionally. The not-yet-arrived-echo case rides the same framework `reset()`+re-register-for-read path the existing `RESUME_ACK`/`RESUME_PONG` branches use (state survives `reset()` via `LocalValueMap.clear()`→`Mutable.clear()`, which does not touch `closeEchoDeadline`).
- **Submodule bump** (`24fbbd5a`→`e2d1eea1`) is a client-side shutdown resource fix (`SegmentManager` close race) + skill docs; it does **not** introduce a reconnect budget, drainer hard-fail, or any store-and-forward contract change (the only `deadlineNanos` is a shutdown worker-join timeout).

Callsite inventory: the ten new public state symbols are referenced only by the two production files and their two test files — no cross-context callers, no broken out-of-diff callsite.

No Critical or Moderate findings. Everything below is Minor.

### Minor

1. **`beginCloseEchoWaitIfEligible` / the "role-change CLOSE sent" log are close-code-agnostic** (`QwpIngressUpgradeProcessor.java:826`). The predicate is `isDurableAckEnabled() && isRoleChangeCloseDeferred() && !hasPendingDurableWork()` — it never inspects the close code. A non-`NORMAL_CLOSURE` fatal close landing during a role-change deferral window with the pending maps already empty (a client TEXT/fragmented frame → `PROTOCOL_ERROR`/`UNSUPPORTED_DATA`, or an oversized frame → `MESSAGE_TOO_BIG`, arriving between the registry catching up and the next normal poll) will arm the echo wait and log `role-change CLOSE sent, awaiting client close echo` for a close that is not the role-change close. Behaviorally benign — empty maps mean no replay window, the client-visible close code is written before the arming decision, the `MESSAGE_TOO_BIG` case self-heals into discard mode, availability is bounded by the same 5 s grace + idle reaper — so completing the handshake for any close is defensible; the concrete wart is the **misleading operator log** during a demote. Consider gating arming to the `NORMAL_CLOSURE` path or making the log accurate. (Expanded form of the earlier Minor #6.)

2. **Inaccurate "unreachable" comment** (`QwpIngressUpgradeProcessor.java:1567`, also :524): `sendFatalClose` returns normally rather than throwing when it arms a *fresh* echo wait, so the `return;` after it *is* reachable. Code is correct; only the comment overstates unreachability.

3. **Dead defensive guard**: `roleChangeCloseWithUploadGrace` :1414 `if (state.isAwaitingCloseEcho())` is unreachable — its callers (`handleBinaryMessage`, `handlePing`) sit behind the top-of-`handleWebSocketFrame` gate that discards all non-CLOSE opcodes during the wait. Harmless defense-in-depth (earlier Minor #4).

4. **Member ordering** (`QwpIngressProcessorState.java`): `isAwaitingCloseEcho()`/`isCloseEchoWaitExpired()` are inserted after `isOk()` (alphabetically precede it), and `beginCloseEchoWait()` sits next to `deferRoleChangeClose()` rather than in `b…` position. Follows the existing local grouping, but the convention wants alphabetical.

5. **Labels**: only `tandem`. A `Bug` label matches the scope. No `Fixes #NNN` expected (found in enterprise CI, not a public issue).

### Downgraded (false positives eliminated during verification)
- "reject paths still emit a second CLOSE during the echo wait" — closed by the hoisted gate + `sendFatalClose` stand-down.
- "oversized-frame paths spin the dispatcher" — closed by `onCloseEchoSyncLost` + drain-to-EAGAIN + edge-triggered re-arm.
- "TOCTOU can strand a durable ack behind the CLOSE" — `hasPendingDurableWork()` reads local maps; `flushPendingAck` also gated.
- "resumeSend arming with no buffered echo hangs to the idle reaper" — framework `reset()`+re-register-for-read receives the later echo; state survives `reset()`.
- "`handleClose` echo path skips a needed ack flush" — no committed-but-unacked work exists at echo time (T1 emptied the maps; discarded frames advance nothing).
- "non-ASCII in log messages" — all em-dashes/subscripts are in comments; every new `LOG.*` call is ASCII.

### Coverage map (15 behavioral changes, 15 tested, 0 UNTESTED)

| # | Behavioral change | Test | Verdict |
|---|---|---|---|
|1|Echo-wait deadline arm/poll/expiry/idempotency, survives `clear()`, reset on disconnect|`testCloseEchoWaitLifecycle`|TESTED|
|2|`hasPendingDurableWork()` tracks local maps, not registry|`testHasPendingDurableWork`|TESTED|
|3|Eligibility ignores racing (T1,T2] registry advance; CLOSE stays final|`testGraceExpiredEligibilityIgnoresRacingRegistryAdvance`|TESTED|
|4|Coverage-complete close arms wait; inbound discarded; echo completes; only seq0 commits|`testCloseEchoWaitDiscardsInboundAndCompletesOnEcho`|TESTED|
|5|Expiry tears down with duplicate-risk diagnostic|`testCloseEchoWaitExpiryTearsDownWithoutEcho`|TESTED|
|6|Discard gate covers TEXT/fragmented (no 2nd CLOSE), echo still completes|`testCloseEchoWaitDiscardsProtocolViolatingFramesWithoutSecondClose`|TESTED|
|7|Expiry polled on protocol-violating re-entry|`testCloseEchoWaitExpiryPolledOnProtocolViolatingFrame`|TESTED|
|8|Oversized frame → sync-lost, bounded by grace, no busy-loop, no 2nd CLOSE|`testCloseEchoWaitOversizedFrameBoundedByEchoGrace`|TESTED|
|9|Full-buffer jam enters discard mode (divergent geometry)|`testCloseEchoWaitFullBufferJamEntersDiscardMode`|TESTED|
|10|Peer FIN ends sync-lost wait as success|`testCloseEchoWaitSyncLostPeerFinEndsWait`|TESTED|
|11|`RESUME_CLOSE` resume dispatches buffered crossing echo|`testParkedRoleChangeCloseResumeMustDispatchBufferedCrossingEcho`|TESTED|
|12|`RESUME_CLOSE_RESPONSE` resume disconnects, no echo wait|`testParkedCloseResponseResumeMustDisconnectWithoutEchoWait`|TESTED|
|13|`onFatalCloseBlocked` routes `RESUME_CLOSE_RESPONSE` (table + fuzz)|`testOnFatalCloseBlockedTransitionTableCoversAllSendStates`, `…Fuzz`|TESTED|
|14|Final durable ack precedes CLOSE / no frame after CLOSE, all deferred-close exits|`testRoleChangeCloseMustFlushFinalDurableAckWhenAckParked`, `…WhenPongParked`, `testGraceExpiredCleanClose…`, `testDataFrameReEntry…` (+ `assertCloseIsFinalFrame`)|TESTED|
|15|End-to-end exactly-once demote with handshake|enterprise `SqlFailoverQwpDeferredCloseExactlyOnceTest` (tandem)|TESTED (tandem)|

### Summary

**Verdict: approve.** The data-loss bug (abortive RST destroying the final durable ack → full-corpus replay → duplicates) is correctly fixed via the RFC 6455 close handshake, verified against the client's confirmed echo-before-dispatch contract, and backed by an unusually thorough, blackbox-oriented test suite. The four follow-up commits resolve every blocking item from the earlier review; each closure was re-verified against source. Reflection in tests is used only to *install* a fixture (`getLV()`), not to assert on private state — the sanctioned use.

- Correctness & performance gate: **passes** — no open correctness, data-integrity, resource, concurrency, or performance/IO finding on any reachable path.
- Test gate: **passes** — every behavioral delta has a dedicated test with a real failure link; the fix ships with regression tests verified RED→GREEN.
- Findings: 5 verified (all Minor); ~6 draft findings dropped as false positives after source verification (the prior blocking Moderates among them, closed by the follow-ups).
- In-diff / out-of-diff / pre-existing split: 5 in-diff (all Minor), 0 out-of-diff (callsite inventory clean), 0 pre-existing.

The Minor items are worth a quick polish pass (chiefly the misleading "role-change CLOSE sent" log for a non-`NORMAL_CLOSURE` close), but none blocks merge.

jerrinot (changes requested):
### Critical

  1. In-diff: the echo handshake is skipped when uploads finish before the first demote rejection.

  The role-change path creates the deferral marker only while durable work is still uncovered at core/src/main/java/io/questdb/cutlass/qwp/server/QwpIngressUpgradeProcessor.java:1420. If the registry already covers prior commits, execution goes directly to sendFatalClose(). However, echo eligibility
  additionally requires isRoleChangeCloseDeferred() at core/src/main/java/io/questdb/cutlass/qwp/server/QwpIngressUpgradeProcessor.java:823, so it returns false and immediately disconnects at line 1837.

  Reachable sequence:

  1. A batch commits while its durable watermark is -1.
  2. The uploader advances the registry before the next client frame.
  3. The node demotes; that next frame is rejected.
  4. sendFatalClose() emits the first durable ACK followed by CLOSE, but skips the echo wait.
  5. In-flight client writes can still make teardown abortive, losing the ACK and causing full replay and duplicates—the original defect.

  A scratch regression probe moved durableWatermark.set(Long.MAX_VALUE) before the first demote-rejected frame. It failed with ServerDisconnectException from sendFatalClose:1837 instead of entering the echo wait.

  Fix by explicitly marking every durable role-change close before checking upload completeness, or by passing the role-change cause into the echo-eligibility decision. Add a regression test for “registry advances after the prior batch but before the first rejected frame.”

  2. In-diff: the supposedly bounded sync-loss drain can pin a worker indefinitely.

  discardInboundBytes() loops until recv() returns zero at core/src/main/java/io/questdb/cutlass/qwp/server/QwpIngressUpgradeProcessor.java:867. The five-second deadline is checked only before entering this loop. A peer continuously supplying bytes after an oversized frame can keep recv() positive
  indefinitely, preventing another deadline poll and monopolizing the HTTP worker.

  The expiry path also reaches the pre-existing unbounded loop in core/src/main/java/io/questdb/cutlass/http/HttpConnectionContext.java:241. The current test releases a finite body, drains to EAGAIN, then advances the clock on a separate invocation; it does not cover a continuously readable socket.

  Use a per-dispatch byte/read budget and poll the deadline within that budget. On expiry, perform only a bounded best-effort drain before closing. Test with a facade that remains readable past the deadline.

bluestreak01:
## Review — tandem level 3 (OSS half)

Reviewed together with questdb/questdb-enterprise#1113 and questdb/java-questdb-client#73. All three trees at PR heads; merge base `c4cfd3ba`.

**Verdict: request changes** — one blocking item, narrow and cheap to close.

Verification was empirical where it mattered: regression tests were checked by mutation testing in a scratch worktree, and the two candidate blockers were disproved by tracing the dispatcher contract rather than by reading.

---

### Critical

**C1 — The close handshake is a wire-format change, and it ships with zero tests against the pinned client, under the absolute rule this PR itself adds.**

This PR adds to `CLAUDE.md`:

> *"Never assert peer behaviour with a hand-built frame or a fake socket… **Any change to a wire constant or wire format requires a test that exercises the pinned submodule client.** A green OSS build does not otherwise prove the pinned client agrees."*

It then changes the wire contract — the server now holds the connection open awaiting a client CLOSE frame — and asserts the client's half with hand-forged frames: `closeEchoFrame()` (`QwpIngressDeferredCloseDurableAckTest.java:4182`) and `clientCloseFrame()` (`:4194`), used by six tests including the core regression test. `roleChangeCloseEchoFrame()` (`QwpIngressUpgradeProcessorResumeRecvTest.java:~1750`) reasons in its javadoc about how "deployed store-and-forward clients treat only NORMAL_CLOSURE and GOING_AWAY as orderly" — naming client behaviour in a comment to justify an assertion, which the rule explicitly forbids ("it must import that class instead"). The suite is 4858 lines with **zero** `io.questdb.client.*` imports.

The rule exists because the ROLE_CHANGE (4001) misclassification shipped exactly this way. This is the first change under it.

**Counterfactual symptom:** if the client's echo behaviour drifts — or echoes before consuming the durable-ack frame — the server's delivery proof is false, the replay watermark stays stale, and the client replays its pre-demote corpus to the promoted replica. Duplicate rows on tables without `DEDUP UPSERT KEYS`: the exact bug this PR fixes.

**No live defect.** The premise holds on the pinned client — `WebSocketClient.java:1215-1231` calls `sendCloseFrameEcho(closeCode)` on `WebSocketOpcode.CLOSE`, before `upgraded = false` and before `handler.onClose(...)`. Verified directly. What is unpinned is the *contract*, and specifically the two things fake sockets cannot cover: the **ordering** guarantee the fix depends on (durable-ack frame consumed and watermark advanced **before** the echo is emitted), and the half-close interaction (the server issues `shutdown(SHUT_WR)` before the client has read the CLOSE).

**Either remedy closes this:**
1. **Close the gap** — one test driving the real `java-questdb-client` through a role-change close. The client is already on the test classpath (53 of 91 QWP tests import it), so this is additive.
2. **Scope the promise** — narrow the `CLAUDE.md` rule to what it actually requires, and record why fake-socket coverage plus production-code inspection suffices here.

---

### Moderate

**M1 — Client CLOSE under ack backpressure parks with no QWP-level deadline** (`QwpIngressUpgradeProcessor.java:1611-1627`)
`handleClose` now parks an ack-then-close-response continuation and propagates the backpressure. Base tore the connection down immediately. The park waits for a WRITE event a non-reading client never generates, so it is held to the 5-minute HTTP idle reaper — client-drivable. Every sibling close path has a bounded QWP-level deadline (`CLOSE_DRAIN_TIMEOUT_MICROS`, `ROLE_CHANGE_CLOSE_UPLOAD_GRACE_MICROS`, `CLOSE_ECHO_WAIT_GRACE_MICROS`); this new one does not. The trade buys real data integrity — the parked frame is the client's last watermark — so not blocking, but the inconsistency is worth closing with a deadline.

**M2 — `testOnFatalCloseBlockedTransitionTableCoversAllSendStates` omits the two states this PR adds** (`QwpIngressProcessorStateTest.java:2981`)
The test declares 12 `SEND_STATE_*` constants; `SEND_STATE_RESUME_ACK_THEN_CLOSE_RESPONSE` (12) and `SEND_STATE_RESUME_DURABLE_ACK_THEN_CLOSE_RESPONSE` (13) appear **nowhere** in the file (`grep -c` → 0). It was extended with state 11 only. The name now claims an exhaustiveness it does not have, over a routing contract this PR extended.

**M3 — `drainPendingResponse`'s two new cases have zero test references.** Same close-response state machine as M2.

**M4 — `onDisconnected()` reset test asserts 3 of the 6 new fields.** Production is correct (all six verified reset); the gap is assertion coverage. A dropped reset would leave a reused connection inheriting `isAwaitingCloseEcho`, silently discarding every inbound frame.

**M5 — The `isCloseDraining()` disjunct of the ack gate is a base-behaviour change with no test.** Mutant M5b escapes. No user impact — at base the write side was already `shutdown(SHUT_WR)` before the drain armed, so the suppressed flush could never have sent — but the behaviour did change and nothing pins it.

**M6 — `collectDurableProgress` key-set change and the `dirName == null` branch have no direct coverage.** Verified safe (both maps written together at `:1409-1410`, pruned together in `onDurableAckSent`), so the branch is defensive. Coverage gap only.

**M7 — TLS deferred half-close is retried in the wrong method** (`:1363-1375`)
`halfCloseWriteSide` defers FIN on `wantsTlsWrite()`, and the only retry is in `resumeRecv` — but the dispatcher publishes **WRITE** precisely on `wantsTlsWrite()`, so the event signalling "the ciphertext can now drain" lands in `resumeSend`, where no retry exists. The flag is also never cleared after epoll's `tlsIO` drains it. A silent TLS peer therefore gets neither FIN nor deadline poll and is held to the idle reaper. Fix: mirror the `:589-599` retry block into `resumeSend` and clear `hasPendingCloseEchoHalfClose` on success.

**M8 — `roleChangeCloseInitiated` is sticky across a parked send** (`:~1930`) — latent
The escalation needs a second fatal close while the first is parked, which requires the recv path — unreachable while parked for WRITE (see Downgraded D1). Mechanism real, currently unreachable.

**M9 — `collectDurableProgress` adds a per-table hash lookup per recv on the steady-state ingest path** (`QwpIngressProcessorState.java:383`)
The new `pendingDurableSeqTxns.get(tableName)` costs O(nameLen) hash + compare per table per recv on every durable-ack ingest connection — and the coverage value it computes is read *only* on the demote path. Magnitude: recvs × tables, no fixed bound, on the data path — but amortized across a whole batch (~0.2% of a 64 KiB-batch recv, 1–2% of a tiny-batch recv), so a user does not feel it. Fix: a `collectDurableProgress(registry, boolean computeCoverage)` overload; the demote path passes `true`, ingest `false`.

Related: the fused-traversal win is **demote-path-only** — 3→1 for a deferred PING, 4→1 completing, 3→1 gate-refused BINARY, but **1→1 unchanged for a normal BINARY**, with each surviving traversal now ~50% more expensive.

**M10 — Durable-ack poll suppressed on partial-frame receives** (`:2147`)
Cumulative acks are untouched; the *durable* poll is dropped on frame-incomplete recvs. The base comment explicitly warned that a `COMPLETE`-only check "starved senders of ACKs… leaving the client's `drainOnClose` to time out". Recovery is bounded by the client keepalive (`DEFAULT_DURABLE_ACK_KEEPALIVE_INTERVAL_MILLIS = 200L`, operator-settable via `durable_ack_keepalive_interval_millis`). Not cosmetic: SF trim only advances on durable acks, so the producer backpressures at `sf_max_segment_bytes`. Invisible at the default; a large configured interval extends it proportionally.

**M11 — Boolean-parameter explosion.** `flushPendingAck` ×3 overloads with two boolean flags; `roleChangeCloseWithUploadGrace` 7 args including four booleans; `sendFatalClose` 6 args with two. The `(isDurableAckPollAllowed, isDurableProgressCollected)` pair encodes 3 states in 4 combinations with `(true,true)` meaningless, and its derivation is duplicated verbatim at `:1879` and `:2376` behind a double negative. A 3-valued enum is mechanical and clearly better — `Status` in the same class shows enums are idiomatic here.

**M12 — PR body accuracy.** Three items:
- The "Send the private-use `ROLE_CHANGE` close code (4001)" bullet describes code that does not exist — `:1919` sends `WebSocketCloseCode.NORMAL_CLOSURE`, and the literal `4001` appears nowhere in `qwp/server/`. Its second half ("accept only the matching echo… reports that replay may duplicate") describes a discrimination the code says is *impossible*: the comment at `:1577-1594` states a genuine echo is indistinguishable from a crossed CLOSE and both take one identical branch.
- The ten "Store-and-forward client lifecycle and recovery" bullets are in neither this diff nor the client diff — they already landed on client `main` via #64/#67/#72 and are inherited through the merge base. `git diff --name-only` on this PR shows zero SF files.
- The body omits its own 42-line `CLAUDE.md` rule change — the rule that produces C1.

---

### Minor

Member ordering: states 11/12/13 inserted before `SEND_STATE_RESUME_DRAIN_THEN_CLOSE = 10`; `closeEchoDeadline`/`hasLostCloseEchoSync`/`hasPendingCloseEchoHalfClose` between `roleChangeClosePending` and `roleChangeCloseDeferredDeadline`; `getPendingCloseResponseCode()` among the `on*` handlers; `sendCloseResponse` ahead of `handlePing`; `CLOSE_ECHO_DISCARD_READ_BUDGET` before `_BYTE_BUDGET`; `closeEchoDiscardedFrames:191` non-final inside the `private final` block (17 deviations total, against an already-unsorted base) · dead `long sendBuf = buffers.sendBuffer;` in **27** tests, never read · javadoc bloat — 566 javadoc lines over 36 `@Test` methods, 28% of the file, worst is 28 javadoc lines on a 3-line delegation · class javadoc no longer describes ~80% of the class · table name `tabg` used by two tests · `boolean disconnected`/`peerClosed` missing `is` prefix · three `assertTrue("test scaffolding", …)` messages · `256 * 1024` literal shadowing `CLOSE_DRAIN_BYTE_BUDGET` · two near-identical recording facades with divergent field names · hidden clock side effect in `AlwaysReadableNetworkFacade`'s constructor parameter · the new `catch (ServerDisconnectException)` comment claims every reachable teardown drains into the buffer — false for the non-echo `STATE_ERROR` throw and `gracefulCloseAndDrain`'s failed-shutdown path (behaviour correct, justification wrong) · label is `ILP` where a dedicated `QWP` label exists and was used by the two preceding `fix(qwp)` PRs.

---

### Verified correct (recorded so it is not re-litigated)

- **Regression tests are genuinely regressive — EXECUTED.** Baseline `38 run, 0 failures`. Whole-file revert does not compile (200 × `cannot find symbol`), so surgical behavioural mutations were used instead, attributing failures to single lines. **M1** (disable the echo wait = merge-base teardown): `38 run, 5 failures, 20 errors`; all four nominated regression tests fail with their exact assertion messages. **M2–M6** independently isolate the eligibility predicate, the arming half-close, the expiry predicate and the expiry teardown. Three tests still pass under M1 (`…LegalLargeFrameAdmittedInOneTurn`, `…ParserValidFloodBoundedByByteBudget`, `…CompleteFrameFloodBoundedByByteBudget`) — they pin worker-fairness budgets, not the duplication bug; legitimate coverage, correctly scoped.
- **The echo wait provably arms on both paths.** `onDurableAckSent` prunes exactly the tables in `durableProgressSnapshot`; `collectDurableProgress` inserts iff `uploaded > lastSent`; `lastSent < pending` holds for every surviving entry, so `uploaded >= pending ⇒ pruned`. The eligibility key cannot latch a false negative.
- **No registry read occurs during the echo wait.** All five `engine.getDurableAckRegistry()` sites (`:421, 1713, 1831, 1840, 2517`) are unreachable while awaiting, via the discard gate (`:1944`), the `flushPendingAck` gate (`:1247`), and the proof that `sendState` stays READY throughout. The ENT `frozenSnapshot()` window is **not** widened.
- **State-machine completeness:** all three new states have explicit arms in all six switch/if sites; none falls through to a default.
- **Cross-context:** 21 callsites walked, 20 SAFE, 1 latent-unreachable, **0 BROKEN**. Zero `sendState` consumers outside the two changed files. SECURITY_ERROR never used `sendFatalClose` (routes through `sendErrorResponse`), so the new gate cannot suppress a typed terminal NACK.
- **LOG chains: 89 scanned across both production files, zero unterminated, zero `.put()` mid-chain.** All four changed test files clean too.
- **Zero-GC clean**; no `Os.currentTimeMicros()` JNI transition added to any steady-state path; the `Math.min(recvBufferSize, BUDGET − drained)` cap provably cannot degrade into tiny reads.
- **Per-frame steady-state cost** is ~3 field loads + ~6 fully-predicted branches per frame against 10⁴–10⁶ cycles of WAL work — under 0.1%.

---

### Downgraded (false positives, recorded with their disproof)

- **D1 — "CLOSE arm's `SHUT_WR` precedes `onConnectionClosed`'s parked-ack flush → lost watermark → duplicates."** The load-bearing step was reachability of `handleClose` with `sendState != READY`. `HttpServer.java:378-383` registers **WRITE only** on `PeerIsSlowToReadException` (READ only on `PeerIsSlowToWrite`), so a parked send is never dispatched to `resumeRecv`; and `drainBufferedFrames` guards on `isSendReady()`. `handleClose` is only ever reached with `sendState == READY`. This also collapses the `roleChangeCloseInitiated` escalation (M8).
- **D2 — "5s × N resource hold at mass demote is new."** Base `sendFatalClose` already ended in `gracefulCloseAndDrain` with `CLOSE_DRAIN_TIMEOUT_MICROS = 5_000_000`, same recv-driven polling, same idle-reaper fallback, identical resource set. The two waits are mutually exclusive at `finishServerFatalClose:1220`. **Delta ≈ 0.**
- **D3 — "The wait widens the post-`close()` registry-read window."** No registry read happens during the wait at all; `FrozenSnapshot` is a pure heap object with no `close()` or native memory, so no lifetime bound exists that 5 s could exceed.
- **D4 — "`collectDurableProgress` key-set change can skip tables."** Both maps written together (`:1409-1410`), pruned together (`:878-885`).
- **D5 — "`closeEchoDiscardedFrames` on the shared processor is a race."** One thread at a time (`HttpServer.java:592/600`, per-worker selectors or exclusive `ConcurrentPool.acquire()`), and `processWebSocketFrames` is non-re-entrant. Same argument as the pre-existing `frameParser`.
- **D6 — "The hold can exceed the ENT demote drain budget."** `drainWriterPool` counts `getWriterPoolEntries()` = the non-WAL `writerPool`; QWP holds `walWriterPool` writers. Zero contribution. ENT never awaits QWP connections.

---

### Adjacent (pre-existing — file separately, not for this PR)

1. `QwpIngressUpgradeProcessorResumeRecvTest`: two sequential `Unsafe.malloc` outside try — a failed second strands the first. Pre-existing shape; this PR's new tests copied it instead of the `NativeSocketBuffers` RAII helper the same PR introduces.
2. `createMaskedFrame`/`oneRowMessage`/`writeLeLong`/`drive` duplicated across 4 test classes; a superset exists as `QwpWireTestFixtures.maskedFrame` but package-private in a different package.
3. `CLOSE_DRAIN_BYTE_BUDGET` is private while its three siblings are public, forcing a literal in tests.

---

### Summary

- **Verdict: request changes** — C1 only.
- **Test gate: passes.** The fix ships with regression tests proven to fail without it, by executed mutation. No UNTESTED row carries a user-visible behavioural delta; C1 is a wire-*contract* gap, not a missing regression test.
- **Coverage map:** 91 behavioural rows enumerated with recorded searches and explicit failure links; UNTESTED rows consolidate into 12 groups, all internal branches, defensive guards, or assertion-completeness gaps.
- **Split:** in-diff, 0 out-of-diff-breakage, several incomplete-hardening, 3 adjacent. The zero out-of-diff-breakage is a real result with a structural cause — ENT never names the changed classes, and every new send state has an explicit arm in all six `sendState` consumers.
- **Severity distribution:** 1 Critical, 12 Moderate, ~13 Minor. Both severity audits were run: downward (two candidate Criticals disproved — D1 by the dispatcher contract, D2 by base already parking 5 s) and upward (M1 is closest to promotion but is reaper-bounded and buys data integrity; M8 is structurally unreachable, cited).

The engineering is sound — the handshake is correct, the state machine complete, the eligibility predicate cannot latch a false negative, and the regression suite is genuinely regressive under executed mutation. The blocker is that the PR writes a rule about proving wire contracts against the pinned client and then does not follow it, on a wire-contract change, in the codebase where skipping that step already shipped one incident.

<sub>Review performed at level 3 with a 13-agent fan-out; findings verified against source, with the regression set and the two candidate blockers verified empirically.</sub>

bluestreak01:
## Independent validation of review finding C1 (wire-format change vs. pinned-client coverage rule)

I reproduced every load-bearing assertion of C1 against the working tree (`fix/qwp-deferred-close-echo-wait` + the enterprise branch that consumes it). Verdict: **validated as scoped, with one detail understated and one element overstated**. Details below so the remedy discussion can be concrete.

### Confirmed (OSS)

| Claim | Evidence |
|---|---|
| Rule added, +16 lines | `CLAUDE.md` diff: verbatim *"Never assert peer behaviour with a hand-built frame or a fake socket"* … *"Any change to a wire constant or wire format requires a test that exercises the pinned submodule client. A green OSS build does not otherwise prove the pinned client agrees."* |
| Wire-format change is real | `QwpIngressUpgradeProcessor` (+1081) adds the echo-wait machine: `isAwaitingCloseEcho()`, `checkCloseEchoWaitExpiry`, `halfCloseWriteSide`, `CLOSE_ECHO_*` budgets. Server now holds the fd awaiting the client CLOSE — new wire-visible surface. |
| Zero pinned-client coverage in the OSS diff | All four changed test files have **0** `io.questdb.client.*` imports — including the e2e `QwpServerCloseDrainTest`, which drives a raw `Socket` with `createMaskedFrame(WebSocketOpcode.CLOSE, ...)`. |
| Forged-frame helpers at cited lines | `closeEchoFrame()` at :4182, `clientCloseFrame()` at :4194 (`QwpIngressDeferredCloseDurableAckTest`, 4859 lines, 38 tests); `roleChangeCloseEchoFrame()` at `QwpIngressUpgradeProcessorResumeRecvTest:1764`, whose javadoc justifies the forged echo by describing deployed client behaviour (*"treat only NORMAL_CLOSURE and GOING_AWAY as orderly"*) without importing the client. |
| Pinned-client premise (the offset) | `WebSocketClient.java:1215–1231` verified: CLOSE arm → `sendCloseFrameEcho(closeCode)` **before** `upgraded = false` and `handler.onClose`. The submodule bump is version-only for this file, so the echo behaviour predates this PR and nothing in this PR pins it. No live defect — coverage-gap classification is correct. |

**Understated:** C1 says the forged frames back "six tests". Mapping call sites to enclosing methods gives **14 distinct test methods** in the DurableAck suite alone (both `testRoleChangeCloseMust*` regression tests and the whole `testCloseEchoWait*` family), plus one in ResumeRecv.

### Overstated: "wire drift on the demote path would ship unnoticed"

The enterprise repo has two tiers of real-pinned-client coverage that traverse this exact handshake and run in this PR's gate:

- **`SqlFailoverQwpDeferredCloseExactlyOnceTest`** (Java, real client facade; *modified by this PR*, +42/−17) — per its own javadoc a *"DETERMINISTIC end-to-end witness for the QWP role-change close deferral (INVARIANT B, `roleChangeCloseWithUploadGrace`)"*: durable-ack client through a SQL-triggered failover, **exactly once, on a table without `DEDUP UPSERT KEYS`, with committed work forced pending-upload at demote time**. That is C1's counterfactual, verbatim. Its `>=`-then-`==` oracle fails immediately on `count > total`.
- **Python e2e** — `test_demotion_mid_stream.py` drives the real `java-questdb-client` sender through an in-place demote mid-stream (dense-sequence oracle: no loss AND no duplicates), plus `test_failover_graceful` / `test_close_drain_failover` / `test_durable_ack_failover` / `test_switch_fuzz` covering adjacent teardown flavors.
- Additional pre-existing real-client demote coverage: `DurableAckThroughDemoteTest`, `SqlFailoverQwpClientLosslessTest`, `QwpRoleBounceChaosLosslessTest`, `QwpEgressDropDemoteRaceTest`.

If the client echoed before consuming the final durable ack (the ordering C1 flags as unpinned), its watermark would be stale, it would replay to the promoted node, and both witnesses above go red in this PR's own CI. So drift on the *semantic* contract would not ship unnoticed.

### What remains genuinely unpinned anywhere

Narrower than C1 states, but real:

1. **Echo-vs-expiry completion.** The e2e oracles pass even if the echo-wait silently degrades to the expiry fallback on every close (outcome is still exactly-once, just slower + fd held for the grace window). Only the OSS forged-frame tests distinguish a parsed echo from expiry — and they do it against an assumed client.
2. **Half-close/FIN byte-level interaction** (server `shutdown(SHUT_WR)` racing the client's read of the CLOSE).

### Suggested resolution

C1 stands as a rule-consistency / coverage-**locality** finding: the rule lives in this repo and demands the OSS build prove the contract (this repo syncs to a public tree where enterprise CI does not exist), and the OSS build does not. Given the enterprise evidence, the proportionate fix is C1's remedy 2 — scope the `CLAUDE.md` promise and record where the real-client proof lives — optionally plus one OSS test pinning the echo-vs-expiry mechanics that no tier currently distinguishes. Remedy 1 as originally stated (full real-client role-change close test in OSS) is stronger than the residual gap requires, though it remains the letter of the rule as written.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 280 / 315 (88.89%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpIngressUpgradeProcessor.java|229|259|88.42%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpIngressProcessorState.java|51|56|91.07%|

PR #2226, opened 2022-06-14
chore: fix flapping test leaking fd in http

ideoma:
Fixes #2222

PR #7368, opened 2026-07-07
chore(core): fix socket leaks on shutdown when queries are parked or retrying

RaphDal:
## Summary

When an I/O dispatcher (PGWire, HTTP, ...) shuts down, a connection whose context has been *checked out* of the dispatcher can have its server-side socket leaked. A checked-out context has been removed from the dispatcher's `pending` set and consumed from the I/O event queue; the dispatcher re-acquires it only when the worker re-arms it (`registerChannel`) or frees it (`disconnect`). While checked out, the context is owned by some other holder -- a parked `sleep(...)` worker continuation, a suspended query on the continuation wake interval, or an HTTP request waiting on a retry. `AbstractIODispatcher.close()` sweeps only the collections it still owns, so a checked-out context is invisible to it. If the holder that owns the context at shutdown does not free it, the context and its socket descriptor are freed by nobody and stay open for the lifetime of the JVM.

A single leaked descriptor at process exit is reclaimed by the OS. The leak matters for long-lived JVMs that start and stop the engine repeatedly -- embedded deployments and the test harness -- where it accumulates across restarts. It is a leak path that intermittently trips the `assertMemoryLeak` descriptor check in tests such as `ServerMainSleepTest.testSleepCancelledByConnectionDrop`.

## Fix 1 -- dispatcher: a terminal `disconnect()` / `registerChannel()` after `close()`

When a parked query unwinds *after* the dispatcher has already closed, its terminal `disconnect()` / `registerChannel()` call lands on a closed dispatcher. Those calls publish the context to the disconnect / interest ring queue through `bullyUntilClosed()`. Once the dispatcher is closed, `bullyUntilClosed()` still returns a valid cursor whenever the queue has room, so the call enqueues successfully -- but `close()` has already run its single `processDisconnects()` / `consumeAll()` drain and never re-drains, so a post-close enqueue is never processed. The pre-existing `cursor < 0` guard only caught the rarer full-queue case, so the leak survived it.

`disconnect()` and `registerChannel()` now check `closed` up front and, when it is set, free the context inline via `doDisconnect()` instead of enqueuing it. `doDisconnect()` already frees the context (`Misc.free`) and balances the connection counters on its closed path, so this reuses the existing teardown rather than adding a second one. The `cursor < 0` branch routes to the same `doDisconnect()` so the full-queue-during-close case is covered too.

`checkConnectionLimitAndRestartListener()` gains a `closed` guard: `doDisconnect()` calls into it, and without the guard the connection-count decrement could drop below the limit and re-create the listener socket that `close()` has just torn down. Normal operation is unchanged -- the new branches execute only when `closed` is true, which never happens on the hot path.

### Freeing the checked-out context exactly once

Routing `disconnect()` / `registerChannel()` into `doDisconnect()` when `closed` lets the same context be freed from two places during shutdown. A connection whose heartbeat is mid-handshake sits in `pendingHeartbeats` while it is simultaneously checked out to the worker that acks it with `registerChannel(HEARTBEAT)`. During teardown `close()` sweeps `pendingHeartbeats` and frees the context, and the worker's inline `doDisconnect()` frees it as well. `doDisconnect()`'s only prior guard against a repeat free was `context.invalid()` -- `socket.isClosed()` over a plain, non-volatile `fd` -- which is a check-then-act, not an interlock. When the two frees run concurrently, which they can once `close()` proceeds past the worker-pool halt timeout with a worker still running, both read `fd != -1`, both call `nf.close(fd)` (double-closing a descriptor the OS may already have reused), and both free the context's native buffers.

`doDisconnect()` now claims the context before freeing it: `IOContext.tryDisconnect()` CAS-es a per-context flag from 0 to 1, the single winning caller frees the context and adjusts the connection counters, and every other caller returns. `of()` resets the flag at each checkout, so each connection starts unclaimed. The claim also covers `close()`'s own sweeps, which previously depended on draining the queues before iterating `pendingHeartbeats` to avoid closing a context twice. The cost on the normal path is one CAS per connection teardown, on the cold disconnect path; the hot request path is untouched.

## Fix 2 -- HTTP wait processor: retries stranded in the out queue

An HTTP request that hits backpressure (`RetryOperationException`, e.g. a busy writer) is checked out of the dispatcher and parked in `WaitProcessor`: `HttpConnectionContext.scheduleRetry()` hands the context to `reschedule()` and returns without re-arming it on the dispatcher, so `WaitProcessor` becomes its sole owner. A retry then flows through three stages -- the inbound queue, the `nextRerun` priority queue, and the outbound queue -- before the HTTP worker's `runReruns()` picks it up and re-runs it.

`WaitProcessor.close()` drained the inbound queue (via `processInQueue()`) and freed `nextRerun`, but never drained the outbound queue. The outbound queue is consumed only by `runReruns()`, which runs on the HTTP worker job; shutdown halts the worker pool before `close()` runs, so nothing drains it. A retry that `sendToOutQueue()` promoted to the outbound queue but that no `runReruns()` consumed before the halt is therefore freed by nobody, and its `HttpConnectionContext` (a socket-owning `IOContext`) leaks its descriptor. This is not a precise-timing race: any retry mid cross-job handoff at shutdown strands. The `WaitProcessor.close()` teardown was originally added (#2226) to fix exactly this fd-leak-on-shutdown symptom; it covered the inbound and rerun stages and missed the outbound one.

`close()` now also drains the outbound queue via `getNextRerun()` and frees each retry, mirroring the existing `nextRerun` teardown. Both the main and min HTTP servers construct a `WaitProcessor`, so both are covered; in practice the min server's only handler never issues a retry, so it never parks a context.

## Fix 3 -- timer shards: a popped entry dropped when shutdown races the fire

The parked-`sleep(...)` / `wait_wal_table()` machinery binds a checked-out context into a `TimerCont` registered on `TimerShards`. Each shard's daemon thread runs `e = take(); if (e == POISON || !running) return; e.expire();`. `take()` removes the entry from the heap (`poll()`) before returning it. If `shutdown()` clears the volatile `running` flag in the window between `take()` returning a real, fired entry and the `!running` re-check, the daemon returns without calling either `expire()` or `shutdown()` on that entry. Because `take()` already removed it from the heap, `shutdown()`'s post-join `toArray()` snapshot never sees it either, so its `shutdown()` drain does not fire it. The `TimerCont` stays `PENDING` forever: the continuation bound to it is never resumed, so the context it carries -- and its socket descriptor plus native buffers -- leaks.

This is upstream of Fix 1: the body never unwinds, so it never reaches the dispatcher's terminal call at all. The window is narrow -- a timer fire must coincide with `shutdown()` within a few instructions -- but the fix is complete. When the daemon pops a real entry and finds `!running`, it now calls `e.shutdown()` instead of dropping it. The daemon is the sole owner of the entry after `take()` removed it from the heap, so this is exactly one terminal call, and the resumed body observes the shutdown flag and unwinds. Poison is still short-circuited before the `running` check, unchanged.

## Tradeoffs and limits

- Fix 1 closes the deterministic case: a terminal `disconnect()` / `registerChannel()` that runs after `close()` has completed. A much narrower race remains -- if a caller reads `closed` as false, then `close()` runs its full drain, and only then does the caller's enqueue land, the context can still strand. That window is nanoseconds wide and is bounded by the worker-pool halt timeout that already bounds shutdown. It is strictly smaller than the pre-fix window, in which every post-close disconnect stranded.
- The atomic claim in Fix 1 closes the concurrent double-free described above. What remains of the shutdown race is the stranding leak in the previous point: it loses at most one descriptor and never double-frees or double-closes a descriptor.
- `publishOperation()` shares the `bullyUntilClosed` shape, but its caller (`IODispatcherLinux.handleSocketOperation`) still holds the context in `pending` at the refused publish and races `close()`'s `pending` sweep there. That is a separate, pre-existing `run()`-versus-`close()` data race and is left untouched here.
- Fix 3 removes the pop-side drop completely; no residual remains on that path. The pre-existing, deliberately-unfixed `register()`-after-`shutdown()` race in `TimerShards` (documented in that method) is unrelated and untouched -- it is bounded to entries registered in the same instant as shutdown and is a separate tradeoff.
- An alternative to all three -- have the context factory track leased contexts so `close()` can free any that are in no queue or matrix -- would also cover a checked-out context whose terminal call is never issued at all. It carries per-lease bookkeeping and was not needed for the paths a parked-query unwind or an HTTP retry actually takes, so each fix stays local to its holder.

## Reproduction status

The dispatcher-level leak (Fix 1) is reproduced deterministically by `IODispatcherCloseParkedContextTest`. The out-queue leak (Fix 2) is reproduced deterministically by a new `WaitProcessorTest` method. The timer-shard drop (Fix 3) is a race; its new `TimerShardsTest` method reproduces it reliably by racing register-due-now against `shutdown()` and asserting every entry receives exactly one terminal transition. The end-to-end symptom in `testSleepCancelledByConnectionDrop` remains dependent on shutdown scheduling and, as noted on the earlier work in this area, is rare and not reliably reproducible on a fast local box; the new tests remove that dependency by exercising each leaking holder directly.

## Tests

- `IODispatcherCloseParkedContextTest` reproduces both dispatcher failures: it accepts a connection, hands the context to a processor that checks it out and does not re-register it (the parked-query state), closes the dispatcher, then issues the terminal call. Two methods issue it once -- `disconnect()` and `registerChannel()` -- and before the inline-free change each leaks one NON_CACHED socket descriptor and trips `assertMemoryLeak`. Two further methods issue it twice, standing in for a concurrent second free; they assert the connection count returns to zero. Without the atomic claim the second free double-frees the context's native buffer and crashes the fork.
- `WaitProcessorTest.testCloseFreesRetryStrandedInOutQueue` reschedules a retry, drives it into the out queue (reschedule, run once, advance the clock, run again), skips `runReruns()` so it stays parked there, then closes the processor and asserts the retry was freed. Before Fix 2 the retry's `close()` is never called and the assertion fails.
- `TimerShardsTest.testShutdownDoesNotDropPoppedEntry` races register-due-now entries against `shutdown()` across many iterations and asserts every entry receives exactly one terminal call. Before Fix 3 it fails on the first iteration that hits the drop window (an entry gets zero terminal calls).
- `IODispatcherHeartbeatTest` (dispatcher close with active connections), `IODispatcherTest` (full HTTP connection lifecycle), the existing `WaitProcessorTest` / `TimerShardsTest` methods, and the PG-based `ServerMainSleepTest` methods -- including `testSleepCancelledByConnectionDrop` -- pass with the change.

## Test plan

- Run `IODispatcherCloseParkedContextTest` and confirm all four methods pass. Revert the `closed` inline-free and confirm the two single-call methods fail with a leaked descriptor; disable the `tryDisconnect()` claim and confirm the two double-call methods fail (the second free crashes the fork).
- Run `WaitProcessorTest` and confirm all methods pass. Revert the out-queue drain and confirm `testCloseFreesRetryStrandedInOutQueue` fails (the stranded retry is never freed).
- Run `TimerShardsTest` and confirm all methods pass. Revert the timer-shard fix and confirm `testShutdownDoesNotDropPoppedEntry` fails (an entry is dropped).
- Run `IODispatcherHeartbeatTest` and confirm dispatcher close still frees all connections.
- Run `IODispatcherTest` and confirm the HTTP connection lifecycle still passes.
- Run the PG-based `ServerMainSleepTest` methods and confirm they pass.


RaphDal (commented):
# Code review — PR #7368

**`chore(core): fix socket leaks on shutdown when queries are parked or retrying`**

Level 3 review (all steps, 6 parallel agents including a fresh-context adversarial pass, an adversarial performance pass, and an empirical revert-and-fail verification in an isolated worktree; per-finding source verification).

**Verdict: Approve.** All three fixes are correct and thread-safe on the normal shutdown path, and each ships with a regression test empirically proven to fail (or crash the JVM) when its production hunk is reverted. The two Moderate items (M1, M2) have been **fixed in this branch** and re-verified — see "Fixes applied" below. No Critical issues remain.

---

## Fixes applied in this review

Both Moderate findings were addressed directly rather than deferred.

### M1 — `WaitProcessor.close()` out-queue drain no longer gives up on contention

The new out-queue drain went through `getNextRerun()`, whose `if (cursor < 0) return null` collapses an `MCSequence` `-2` (lost-the-slot-to-a-concurrent-consumer, *transient*) into the same "stop" as `-1` (empty). If an HTTP worker was still draining the out-queue past `WorkerPool.halt()`'s bounded timeout, `close()`'s drain could end early and strand the remaining retries' fds — the exact leak class this PR fixes. The sibling `processInQueue()` already retries `-2`.

The drain now runs inline in `close()` with the same `-2` retry idiom as `processInQueue()`, instead of via `getNextRerun()` (whose `null`-on-`-2` behavior is correct for `runReruns()` and is left unchanged):

```java
processInQueue();
// Drain the out queue directly, not via getNextRerun(): that helper returns null on a
// -2 (lost the slot to a worker still draining past the halt timeout), which would end
// this drain early and strand the rest. Mirror processInQueue()'s -2 retry instead.
while (true) {
    long cursor = outSubSequence.next();
    if (cursor < -1) {          // -2: lost the slot, retry
        Os.pause();
        continue;
    }
    if (cursor < 0) {           // -1: empty, done
        break;
    }
    RetryHolder retryHolder = outQueue.get(cursor);
    Misc.free(retryHolder.retry);
    retryHolder.retry = null;
    outSubSequence.done(cursor);
}
for (int i = 0, n = nextRerun.size(); i < n; i++) {
    Misc.free(nextRerun.poll());
}
```

### M2 — the two untested `cursor < 0 → doDisconnect()` branches removed (coverage gap eliminated by construction)

`disconnect()` and `registerChannel()` each had two separate closed-paths — an up-front `if (closed)` block and a post-bully `cursor < 0` block — both routing to the identical `doDisconnect(context, DISCONNECT_SRC_SHUTDOWN)`. The post-bully pair were the two lines the coverage bot flagged uncovered (`10/12`), and they cannot be hit deterministically (they require `closed` to flip *during* the bully loop against a full queue — a race that a test can only reach flakily, and that a deterministic test would need a banned production seam to force).

Rather than accept an untested branch or add a flaky test, the two branches were collapsed into a single guard, so the up-front-closed case and the mid-bully-closed case flow through **one** `doDisconnect` sink that the existing `IODispatcherCloseParkedContextTest` already exercises:

```java
// A negative cursor means the dispatcher is closed -- either up front, or close() flipped
// it during the bully loop (bullyUntilClosed only returns < 0 when closed) ...
final long cursor = closed ? -1 : bullyUntilClosed(disconnectPubSeq);
if (cursor < 0) {
    assert closed;
    doDisconnect(context, DISCONNECT_SRC_SHUTDOWN);
    return;
}
```

This is behavior-preserving (the ternary short-circuits `bullyUntilClosed` when `closed`, exactly as the old up-front `if (closed)` did; the normal `closed == false` hot path calls `bullyUntilClosed` unchanged), removes the duplicated `doDisconnect`+`return`, and makes the previously-uncovered lines reachable from the tested closed-path. Net: `AbstractIODispatcher` loses code and gains coverage.

### Verification of the fixes

Re-ran the affected suites after applying both fixes — **18 tests, 0 failures**:

- `WaitProcessorTest` — 4/4 (incl. `testCloseFreesRetryStrandedInOutQueue`; the single-consumer drain path still frees the stranded retry).
- `IODispatcherCloseParkedContextTest` — 4/4 (all four now flow through the collapsed M2 guard).
- `IODispatcherHeartbeatTest` — 2/2 and `TimerShardsTest` — 8/8 (no regression on the normal dispatcher close / connection-teardown paths).

Files touched by the fixes: `core/src/main/java/io/questdb/cutlass/http/WaitProcessor.java`, `core/src/main/java/io/questdb/network/AbstractIODispatcher.java`.

---

## Critical

None. The three high-risk mechanisms were stress-verified:

- **`tryDisconnect()` CAS is necessary and sufficient.** Reverting *only* the CAS (keeping inline-free) makes the double-call tests abort the JVM with glibc `free(): double free detected` (exit 134). With it, exactly-once holds. `of()` is `final` (`IOContext.java:92`) and `newInstance()→of()` is the sole checkout path, so the claim flag can never be stale-`1` on a live lease.
- **Concurrent worker `doDisconnect()` during `close()` touches only thread-safe state** — `connectionCount` (AtomicInteger), the gauge, a per-context CAS free, and `checkConnectionLimitAndRestartListener()` (now early-returns under `closed`). That guard is load-bearing, not cosmetic: without it a post-close drop below the limit would `createListenerFd()` and resurrect a listener socket after teardown.
- **Fix 3 exactly-once** — `DelayHeap.take()` uses `poll()` (removes from heap) and `toArray()` reflects post-poll state, so a popped entry is never also in the drain snapshot. Empirically `64==64` across 200×64 iterations.

## Moderate

Both fixed in this branch (see "Fixes applied" above).

- **M1 — [in-diff] `WaitProcessor.close()` out-queue drain gave up on `MCSequence` `-2`.** FIXED — inline drain with `-2` retry.
- **M2 — [in-diff, coverage] the two `cursor < 0 → doDisconnect()` branches were untested.** FIXED — branches collapsed into one tested guard.

## Minor (unaddressed — optional polish)

- **m1 — [test] boolean naming.** `IODispatcherCloseParkedContextTest.java:263` `reRegisterInsteadOfDisconnect` and `:295` `dispatcherClosed` lack the `is`/`has` prefix CLAUDE.md requires (applies to tests): `isReRegister` / `isDispatcherClosed`.
- **m2 — [log accuracy] `TimerShards.java:232`.** If the new `e.shutdown()` throws, the outer `catch (Throwable t)` at `:241` logs `"timer shard expire failed"` — it was a shutdown, not an expire. The drain path (`:164-167`) has an accurate message; mirror it.
- **m3 — [comments] the same race is documented three times** — `IOContext.tryDisconnect()` javadoc (`:119-126`), `doDisconnect()`'s inline comment (`AbstractIODispatcher.java`), and the `disconnecting` field comment (`:37-38`). Per the house "few/short comments" rule, let the javadoc own it and shorten the others to a one-line pointer. (Each sits at a real invariant, so this is polish, not a defect.)
- **m4 — [hardening, optional] tighten the CAS contract.** `tryDisconnect()`'s javadoc says `doDisconnect` is "idempotent when two threads reach it concurrently" — true today only because the pooling/reuse path is single-threaded (`SynchronizedJob`) and the closed path never pools. Resetting `disconnecting` in `_clear()` as well as `of()`, and narrowing the comment, would harden against a future off-dispatcher-thread or pool-in-closed-path change.
- **m5 — [semantics, ~unreachable] `halt()` can now fire one `shutdown()`.** Fix 3's `!running` arm lets `halt()` fire `shutdown()` on the single entry a daemon pops in the race window — a slight deviation from halt's "without draining." Harmless/idempotent and unreachable in production (`CairoEngine` calls `halt()` after `shutdown()`, so `running` is already false).
- **m6 — [metadata, advisory] label.** Fix 2 is entirely HTTP-specific (`WaitProcessor`/`HttpConnectionContext`); consider adding **`REST API`** alongside `Bug`/`Core`. Do *not* add `Postgres Wire`/`ILP` — those only share the dispatcher.
- **m7 — [test] `TimerShardsTest.testShutdownDoesNotDropPoppedEntry` is a probabilistic reproducer.** It only detects the reverted bug on iterations that hit the take()/running-flip window (200×64 makes a hit likely — it fired on iteration 0 empirically). No false-*positive* risk. Acceptable, consistent with the file's existing race tests.
- **m8 — [wording] PR body.** "The hot request path is untouched" is marginally imprecise: `registerChannel()` (the per-request re-arm) gains one predicted, always-false volatile `closed` load — immeasurable next to the `MPSequence` CAS it already runs, so the "performance-neutral" conclusion stands, but "adds one predicted branch to the re-arm" is the accurate phrasing.

## Pre-existing (visited, scoped out — not introduced by this PR)

- **P1 — `IODispatcherLinux.handleSocketOperation()` publish-refused stranding + `pending` race.** `:98-99` calls `publishOperation()` then `pending.deleteRow(row)`; if `publishOperation` is refused as `closed` flips (cursor<0 → returns without enqueuing), the context is removed from `pending` but never enqueued → leaked, and the `pending` mutation races `close()`'s non-synchronized sweep. The PR explicitly scopes this out, and I concur it should **not** be fixed here — a correct fix needs a quiesce-before-close handshake (or synchronized `pending`) across PG/HTTP/ILP, i.e. a change to the dispatcher's single-threaded-run model that would destabilize this focused leak fix. Named per policy; a genuine latent leak/data-race, narrower than the leaks this PR closes.
- **P2 — `WaitProcessor.close()` `nextRerun`/`inSubSequence` races in the degraded halt-timeout window.** In the same window M1 targeted, `close()` drains the non-thread-safe `nextRerun` `PriorityQueue` and consumes the single-consumer `inSubSequence` concurrently with a still-live worker → possible heap corruption/CME. Both predate this PR. Same root cause and same broad fix as P1; scoped out for the same reason. (M1's fix hardened only the new out-queue drain, which is multi-consumer-safe via `MCSequence`.)

## Downgraded (false positives — verified against source)

- **LineTCP `busyContext.getFd()` after inline free (`LineTcpNetworkIOJob.java:125`) is a UAF** — No. `PlainSocket.close()` sets `fd=-1`; `getFd()` reads a Java field on the non-null `final` `Socket`, returning `-1`. Benign debug-log value.
- **Stale `disconnecting=1` via a subclass `of()` override** — Impossible; `of()` is `final` and `newInstance()` is the only lease path (no factory overrides it).
- **TimerShards double terminal call from the new branch** — Impossible; `take()`=`poll()` removes from the heap, `toArray()` is post-poll. Empirically `64==64`.
- **WaitProcessor double-free (out-queue + nextRerun, or vs `dispatcher.close()`)** — The three stages are disjoint (retries are *moved*), a parked retry is invisible to `dispatcher.close()` (checked-out; `invalid()` true while `pendingRetry`), and the MCSequence consumes each item once.
- **Negative `connectionCount`/gauge under the new concurrency** — One `accept()`++ ↔ one CAS-win `--`; the double-call test asserts `count==0`.

## Coverage map

| # | Behavioral change | Test (verified) | Failure link | Verdict |
|---|---|---|---|---|
| 1 | `disconnect()` inline-free when `closed` | `IODispatcherCloseParkedContextTest.testCloseFreesParkedContextOnDisconnect` | Revert → count stuck at 1 + fd/native leak; empirically `expected:<0> but was:<1>` | **TESTED** |
| 2 | `registerChannel()` inline-free when `closed` | `…testCloseFreesParkedContextOnRegisterChannel` | Same; empirically identical | **TESTED** |
| 3 | `doDisconnect()` `tryDisconnect()` idempotency | `…testDoubleDisconnect/RegisterChannel…FreesContextOnce` | Revert CAS → glibc double-free abort (exit 134); with fix, count==0 | **TESTED** (concurrent *trigger* only sequentially exercised; defense proven by crash-on-revert) |
| 4 | closed-path guard in `disconnect()`/`registerChannel()` (was two branches, now one) | `IODispatcherCloseParkedContextTest` (all 4) | After the M2 collapse the closed-path test flows through the single `if (cursor < 0)` sink; revert → stranded context, `count != 0` | **TESTED** (was UNTESTED before the M2 fix) |
| 5 | `checkConnectionLimitAndRestartListener()` closed-guard | Implicit in tests 1–2 | Without it, post-close `doDisconnect`→`createListenerFd()` leaks a listener fd → `assertMemoryLeak` fails | **TESTED** (implicit) |
| 6 | `of()` resets claim flag per lease | Every test (accept path) | Stale flag → silent-skip leak on next lease | **TESTED** |
| 7 | `WaitProcessor.close()` out-queue drain | `WaitProcessorTest.testCloseFreesRetryStrandedInOutQueue` | Revert → `closed[0]==0` (never freed); empirically `expected:<1> but was:<0>` | **TESTED** (single-consumer path; the `-2` contention edge added by the M1 fix is defensive, exercised only under a concurrent consumer) |
| 8 | `TimerShards` popped-entry `!running` → `shutdown()` | `TimerShardsTest.testShutdownDoesNotDropPoppedEntry` | Revert → dropped entry; empirically `expected:<64> but was:<63>` | **TESTED** |

## Summary

- **Verdict: approve.** The fixes are correct and thread-safe on the normal shutdown path; each residual window is acknowledged in the PR and is strictly narrower than the leaks it closes. The two Moderate items are fixed in this branch.
- **Test gate: passed.** Bug-fix PR; every fix ships a regression test whose failure link is empirically verified by reverting the production hunk in an isolated worktree (leak/assert failure for Fixes 1a/2/3; JVM double-free crash for Fix 1b). No UNTESTED Critical rows; no user-visible SQL/API/NULL surface.
- **Coverage map: 8 behavioral changes, 8 tested** (row 4 moved UNTESTED→TESTED via the M2 collapse; the M1 `-2` edge is defensive and shares the tested drain's sink).
- **Findings: 2 Moderate (both FIXED), 8 Minor (optional), 2 pre-existing (scoped out), 5 false positives dropped.** In-diff/out-of-diff/pre-existing split: 4 in-diff (M1, M2, m1, m2), 4 cross-cutting minor (m3–m6), 2 pre-existing (P1, P2). Cross-context callers (PG/HTTP/ILP dispatch, WaitProcessor, close()-sweeps) were all walked; no out-of-diff breakage — the callers never deref the context after `registerChannel`/`disconnect`.
- **PR metadata:** compliant — `chore(core):` is correct for an internally-found (non-customer) leak, the verb is repeated, the "Tradeoffs and limits" / "Reproduction status" sections are exemplary and analytical, `Bug`/`Core` labels fit; no issue reference needed (no linked GitHub issue). Consider adding `REST API` (m6).

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 33 / 35 (94.29%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/http/WaitProcessor.java|10|12|83.33%|
|:large_blue_circle:|io/questdb/mp/continuation/TimerShards.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/network/AbstractIODispatcher.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/cairo/MetadataCache.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/network/IOContext.java|4|4|100.00%|

bluestreak01 (approved):
# Code review — PR #7368: `chore(core): fix socket leaks on shutdown when queries are parked or retrying`

**Level 3** (full pass). All agent roles (1–7, 9–14; no `.rs` files so Agent 8 skipped) were executed with the same scope, including the fresh-context adversarial and adversarial-performance passes, plus **empirical revert-and-fail verification with real test runs**.

This PR was previously reviewed and carries two fixes (M1 out-queue `-2` retry, M2 collapsed closed-guard) already applied to the branch; this review covers the current head `e9fa27a162`.

## Empirical verification

- **Branch as-is:** `IODispatcherCloseParkedContextTest` 4/4, `WaitProcessorTest` 4/4, `TimerShardsTest` 8/8, `IODispatcherHeartbeatTest` 2/2 — 18/18 pass.
- **All 4 production files reverted to `origin/master` (tests kept):** all 6 new tests fail with exactly the claimed modes — 4× `expected:<0> but was:<1>` (leaked checked-out context, connection count never balanced), `close() must free the retry stranded in the out queue expected:<1> but was:<0>`, and `iteration 0: ... expected:<64> but was:<63>` (timer daemon drops a popped entry on the very first iteration — the race window is wide open, not theoretical).
- **Surgical revert of only the `tryDisconnect()` CAS (inline free kept):** forked JVM aborts with `free(): double free detected in tcache 2` — the CAS is load-bearing and the double-call tests catch its absence with a crash, not a soft failure.
- Working tree restored clean after each revert.

## Cross-context pass (Agent 9)

Callsite inventory (recorded `rg` searches) for `disconnect()` / `registerChannel()` / `doDisconnect()` / `tryDisconnect()` / `IOContext.of()`:

| Callsite | Verdict |
|---|---|
| `PGServer` processor (7 terminal calls, PGServer.java:127–151) | SAFE — context never touched after the terminal call |
| `HttpServer.handleClientOperation` (:377–383) | SAFE — same |
| `LineTcpNetworkIOJob.handleIO`/`onRequest`/`close()` (:87, :149–167) | SAFE — `busyContext` nulled after `disconnect()`; the ILP `busyContext`-at-shutdown path previously **leaked** and is now fixed by the same mechanism |
| `WaitProcessor.run()` catch blocks (:238–244) | SAFE — `getDisconnectReason()` evaluated before the call |
| `IODispatcherLinux/Osx/Windows` `doDisconnect` (IDLE/TLS_ERROR) | SAFE — dispatcher thread, `closed == false`, CAS wins trivially, behavior unchanged |
| `AbstractIODispatcher.close()` sweeps (pending, queues, pendingHeartbeats) | SAFE — CAS interlock makes the heartbeat-window double free impossible (empirically proven) |
| `IOContextFactoryImpl.newInstance()` → `of(fd)` (:68) | SAFE — per-lease flag reset; `of()` is `final`; pooling is per-carrier-thread (`CarrierLocal`), so no reset/claim race |
| `HttpServer.close()` ordering (:346–347) | SAFE — dispatcher closes before `WaitProcessor`, so drain-freed retries can never overlap a dispatcher sweep; queue-slot ownership (MCSequence) excludes worker overlap |

Zero out-of-diff breakage is a verified outcome of this pass, not an underrun.

## Critical

None.

## Moderate

- **M-1 — [in-diff, coverage] `TimerShards.runShard()`: the new `e.shutdown()` call site's throw path is untested.** If the shutdown hook throws inside the new `!running` branch, the pre-existing outer `catch (Throwable)` (TimerShards.java:241) handles it — logged, loop exits via `while (running)` — identical handling to an `expire()` throw, so no entry beyond the throwing one is affected (remaining heap entries reach the drain snapshot). A test would need a shutdown-hook-throwing entry to land in the take()/flip window, which is inherently probabilistic. Reachable but hard to trigger in isolation; handling reuses a pre-existing, drain-equivalent path. Inertness: worst case is one log line and a daemon exit that was about to happen anyway. See also m-3 for the misleading message on that path.

## Minor

- **m-1 — [metadata] PR title type: `chore(core):` should be `fix(core):`.** This is a Bug-labeled fix for real production socket leaks (embedded deployments, engine restarts); the squash commit takes the PR title, and `chore` misclassifies it in release notes. The description itself ("fix socket leaks on shutdown...") reads well — only the type token is wrong.
- **m-2 — [metadata] missing issue reference.** Issue #7296 ("sleep() over PGWire: a dropped connection's socket can leak during server shutdown") is this exact symptom. It is already closed, so `Fixes #7296` is moot, but a "Relates to #7296" line at the top of the body is warranted per convention.
- **m-3 — [log accuracy] `TimerShards.java:241`:** a throw from the new `e.shutdown()` is logged as `"timer shard expire failed"`. The drain path (:164–167) has an accurate message; mirror it (e.g. wrap the new call in its own try/catch or generalize the message).
- **m-4 — [test naming] `IODispatcherCloseParkedContextTest`:** `reRegisterInsteadOfDisconnect` and `dispatcherClosed` lack the `is`/`has` prefix required by CLAUDE.md (applies to test code): `isReRegister` / `isDispatcherClosed`.
- **m-5 — [comments] the close-race is documented at length three times** — `tryDisconnect()` javadoc, `doDisconnect()` inline comment, `disconnecting` field comment. Let the javadoc own it; shorten the other two to one-line pointers.
- **m-6 — [labels, advisory] Fix 2 is HTTP-specific** (`WaitProcessor`/`HttpConnectionContext`); consider adding `REST API` alongside `Bug`/`Core`.

- **Inline free in `disconnect()`/`registerChannel()` causes use-after-free at callers** — dismissed: every callsite (PGServer, HttpServer, LineTcpNetworkIOJob, WaitProcessor.run) verified to never touch the context after the terminal call; `context.getDisconnectReason()` is an argument expression evaluated before the call.
- **Stale reference to a pooled, re-leased context lets a second `doDisconnect` free a live connection** — dismissed: requires two terminal calls for one checkout, violating the pre-existing single-owner protocol; no current callsite can produce it, and the pre-PR `invalid()`-only code had the identical exposure — the CAS neither adds nor removes it.
- **`WaitProcessor.close()` drains bypass dispatcher connection-count/gauge accounting (retries freed via `Misc.free`)** — dismissed as inert: the `AtomicInteger` dies with the dispatcher, and the gauge belongs to the per-`ServerMain` `Metrics` instance which is discarded with the server — nothing reads either after `HttpServer.close()`; pre-existing pattern from #2226 in any case.
- **`WaitProcessor.close()` drain could double-free against a degraded still-running worker** — dismissed: MCSequence slot ownership makes out-queue consumption mutually exclusive; a retry held by a worker is in no queue.
- **Fix 3 could double-fire an entry (expire + drain shutdown)** — dismissed: `DelayHeap.take()` `poll()`s the entry out before returning (DelayHeap.java:84), so the drain's `toArray()` snapshot cannot contain it; empirically 64==64 across 200×64 iterations.
- **Volatile `closed` read per `registerChannel()` / volatile `disconnecting` write per `of()` as hot-path cost** — dismissed: per-request-re-arm / per-connection-checkout frequency (not per-row), required for correctness, no cheaper correct alternative; the CAS runs once per connection teardown on the cold disconnect path.

## Coverage map

| # | Change | Test | Failure link | Dimensions | Verdict |
|---|---|---|---|---|---|
| 1 | `disconnect()` closed → inline `doDisconnect` | `IODispatcherCloseParkedContextTest.testCloseFreesParkedContextOnDisconnect` | conn count `1≠0` + `assertMemoryLeak` fd/native check; **empirically failed on revert** | happy ✓, error path ✓ (closed = the error state), boundary ✓ (single parked context), concurrency → row 3, leak ✓ (`assertMemoryLeak`) | TESTED |
| 2 | `registerChannel()` closed → inline `doDisconnect` | `...testCloseFreesParkedContextOnRegisterChannel` | same; **empirically failed on revert** | as row 1 | TESTED |
| 3 | `doDisconnect()` CAS claim, exactly-once | `...testDoubleDisconnectAfterCloseFreesContextOnce`, `...testDoubleRegisterChannelAfterCloseFreesContextOnce` | **CAS-only revert → glibc `double free` abort in the fork**; count would go negative otherwise | concurrency ✓ (deterministic two-call stand-in for the racing sweep) | TESTED |
| 4 | `checkConnectionLimitAndRestartListener()` closed guard | rows 1–3 exercise it (inline `doDisconnect` → guard) | without the guard, the post-close decrement re-runs `createListenerFd()` → re-opened listener fd trips `assertMemoryLeak` (reasoned; guard is on the empirically-tested path) | error path ✓ | TESTED (reasoned link) |
| 5 | `IOContext.of()` resets `disconnecting` per lease | indirect: `IODispatcherHeartbeatTest` (2/2 run here), `IODispatcherTest` pooled-context reuse | without the reset, the second lease of a pooled context can never be claimed → its fd never freed, count never decremented → those suites' `assertMemoryLeak` trips | happy ✓ | TESTED (indirect) |
| 6 | `IOContext.tryDisconnect()` new public API | row 3 | — | — | TESTED |
| 7 | `WaitProcessor.close()` out-queue drain (with `-2` retry) | `WaitProcessorTest.testCloseFreesRetryStrandedInOutQueue` | `closed[0]` `0≠1`; **empirically failed on revert**. The `-2` contention arm is deterministic-untestable without a banned production seam; correct by the MCSequence contract and mirrors `processInQueue()` | happy ✓, error ✓, boundary ✓ (single stranded retry) | TESTED (main arm) |
| 8 | `TimerShards.runShard()` fires `shutdown()` on popped entry | `TimerShardsTest.testShutdownDoesNotDropPoppedEntry` | `63≠64` on iteration 0 — **empirically failed on revert** | concurrency ✓ (real race, 200 iterations), exactly-once ✓ | TESTED |
| 9 | `e.shutdown()` throw path inside the new branch | none | outer catch logs + daemon exits; same handler as pre-existing expire-throw | error path ✗ | **UNTESTED — Moderate (M-1)** |

## Summary

- **Verdict: approve** — after the Minor metadata items are addressed, above all m-1 (retitle to `fix(core):` before the squash-merge).
- **Correctness & performance gate:** zero confirmed correctness, resource, concurrency, or performance/IO findings remain. All three fixes verified correct: callsite inventory all-SAFE, exactly-once free proven by CAS-revert crash, drop window proven closed by the race test, all changes on cold shutdown/teardown paths (one predicted volatile load added to the re-arm path; required for correctness).
- **Test gate:** coverage map: 9 behavioral changes — 8 TESTED (6 with empirical revert-failure evidence, 1 reasoned, 1 indirect), 1 UNTESTED-Moderate (the probabilistic shutdown-hook-throw branch, M-1). No UNTESTED Critical rows; every claimed fix has a regression test with an empirically verified failure link.
- **Findings:** ~12 draft findings verified → 1 Moderate + 6 Minor confirmed, 6 dropped as false positives (listed above). Split: all confirmed findings in-diff (metadata/cosmetic/coverage); out-of-diff: 0 (verified outcome of the 8-callsite cross-context pass, not an underrun); pre-existing: the previously-scoped-out P1 (`IODispatcherLinux.handleSocketOperation` publish-refused stranding) and P2 (`WaitProcessor` in-queue/`nextRerun` degraded-window races) were re-confirmed as untouched and I concur with keeping them out — a correct fix needs a quiesce-before-close handshake across all three protocol servers and would genuinely destabilize this focused leak fix.
- **Tradeoffs (as the PR itself states, verified accurate):** a nanoseconds-wide enqueue-after-drain stranding window remains in Fix 1 (single-descriptor loss, no double-free — strictly smaller than the pre-fix window); `halt()` can now fire one `shutdown()` hook in the race window (unreachable in production: `CairoEngine` calls `halt()` only after `shutdown()`, CairoEngine.java:2144, :684).

Issue #18, opened 2014-09-03
Lazy / Cached Spill Over to Disk


sirinath:
When reliability can be traded for speed it might be good to have an option to have a delayed spill over to disk and use in memory caching to achieve more speed. 

For the example given, I about 15m to 30m objects per second is achievable for the given example, if not more.

PR #7369, opened 2026-07-07
fix(sql): fix explicit timestamp redesignation in subqueries

nwoolmer:
## Summary

An explicit `(subquery) timestamp(col)` redesignation was silently dropped on several code paths, so downstream consumers kept reading the original designated timestamp instead of the one the user asked for. This PR makes the redesignation authoritative across all of them.

The originally reported symptom: a SAMPLE BY subquery re-timestamped to a `first()` aggregate column and used as the master (left) side of an ASOF/LT JOIN matched against the wrong column — a silent wrong-rows correctness bug:

```sql
SELECT balances.ts, s_ts, s_f_ts
FROM (
  SELECT s_ts, first(s_ts) AS s_f_ts
  FROM (
      SELECT ts AS s_ts, count_distinct(account_id) AS matched
      FROM snapshots
      WHERE ...
      ORDER BY s_ts ASC
  ) timestamp(s_ts)
  WHERE matched = 1
  SAMPLE BY 30m FILL(NONE) ALIGN TO CALENDAR
) timestamp(s_f_ts) ASOF JOIN balances
ORDER BY balances.ts ASC
```

`s_f_ts` is what the user designated for the master side, but the ASOF JOIN matched against `s_ts` (the bucket boundary) instead.

## Root cause

The redesignation was dropped on three distinct paths:

- `SqlOptimiser#moveTimestampToChooseModel` hoists a nested model's explicit `timestamp()` up to its parent wrapper and clears it from the nested model, assuming the parent applies it. That breaks when the nested model is itself a **join root**: `generateJoins()` code-generates that model's own factory directly (to feed the join), before/independently of the outer wrapper.
- `SqlCodeGenerator#generateNoSelect` (bare `(subquery) timestamp(col)` FROM-item) returned the nested factory unchanged, never consulting the model's own explicit timestamp — unlike `generateSelectChoose`, which already does.
- `SqlCodeGenerator#generateSelectGroupBy` (the production-default `SELECT DISTINCT ...` → GROUP BY rewrite, and keyed GROUP BY generally) never applied the designation either, so e.g. `SELECT DISTINCT * FROM (SELECT * FROM tab) timestamp(ts2)` reported no designated timestamp at all (index `-1`).

## Fix

- `SqlOptimiser`: don't strip the explicit timestamp from a nested model that has its own join partners (`getJoinModels().size() > 1`); also carry `isExplicitTimestamp` onto the DISTINCT wrapper models.
- New `RetimestampedRecordCursorFactory` — a **transparent, O(1)** wrapper that re-labels ONLY the designated timestamp. Records are passed through byte-identically (`getCursor`/`getPageFrameCursor`/`getSharedCursor` return the base cursor unchanged); only the metadata timestamp index and the time-frame cursor's reported index change. It is deliberately non-peelable (`isProjection()`/`canPeelForTopK()`/`supportsFilterStealing()` all `false`) so no optimizer strips the relabel, and reports `SCAN_DIRECTION_OTHER` when the base has no designated timestamp (a keyed GROUP BY is unordered) so an enclosing ORDER BY / join master inserts the sort it needs.
- `SqlCodeGenerator`: a single `applyExplicitTimestamp(model, factory)` helper builds that wrapper (a no-op when there is no explicit timestamp or it already matches the base), with the whole ownership transfer guarded so the base factory is freed exactly once on any construction failure. It is applied at `generateNoSelect`, `generateSelectDistinct`, and — via a split of the non-SAMPLE-BY body into `generateSelectGroupBy0` — at every non-SAMPLE-BY `generateSelectGroupBy` return path. (SAMPLE BY manages its own timestamp and is intentionally left untouched.)

## Scope note

This PR bundles the join-root fix, the bare-subquery (`generateNoSelect`) fix, the `generateSelectDistinct` fix, and the default DISTINCT-to-GROUP-BY (`generateSelectGroupBy`) fix. They share the one `applyExplicitTimestamp` / `RetimestampedRecordCursorFactory` mechanism; each is independently reproducible and independently regression-tested below.

## Test plan

- `AsOfJoinTest#testAsOfJoinMasterSubqueryExplicitTimestampAfterSampleBy` — the reported scenario for both ASOF and LT JOIN, with an RHS row added between the two candidate timestamps so the ASOF-vs-LT and s_ts-vs-s_f_ts divergence is observable at row level; asserts `.timestamp(...)` and (for LT) `.noRandomAccess()`. Reverting either the `SqlOptimiser` or `SqlCodeGenerator` fix makes it fail.
- `AsOfJoinTest#testAsOfJoinMasterSubqueryExplicitTimestampMatchesNaturalTimestampIsNoop` — no `Retimestamp` wrapper when the explicit timestamp already matches the natural one.
- `DistinctTest#testDistinctExplicitTimestampSurvivesGroupByRewrite` — default-configuration DISTINCT-to-GROUP-BY now designates the redesignated column (was `-1`).
- `DistinctTimeSeriesMemoryTrackerTest#testExplicitTimestampRedesignationDrivesDistinctAdjacency` — dedup adjacency/grouping keyed off the redesignated column.
- `AsOfJoinTest#testSubqueryExplicitTimestampWithNullValues` — redesignation to a column containing NULL timestamps, consumed via LATEST ON.
- `AsOfJoinTest#testSubqueryExplicitTimestampDrivesLatestOn` — LATEST ON over a bare redesignated subquery (`generateNoSelect`); reverting the `generateNoSelect` fix makes it fail.
- Non-TIMESTAMP / missing-column redesignation targets are rejected with the correct position/message and no leak.
- Regression sweeps (all green): `AsOfJoinTest` 121, `DistinctTest` 22, `DistinctTimeSeriesMemoryTrackerTest` 5, plus `SampleByTest` 314, `GroupByTest` 115, `JoinTest` 225, `OrderByEncodeSortTest` 146, and the SqlParser/SqlOptimiser timestamp tests. The only non-passing results are `could not bind socket 0.0.0.0:8812` errors from `ServerMain`-based tests when the pgwire port is already in use on the build host — an environment conflict, not a code regression.


nwoolmer:
## lvl-3 review

8-angle review (3 correctness + reuse/simplification/efficiency + altitude + conventions), each independently verified.

**Findings, most severe first:**

1. **`generateSelectDistinct` (SqlCodeGenerator.java) has the same defect this PR fixes, uncovered by it.** Reproduced directly: `SELECT DISTINCT * FROM (subquery) timestamp(ts2)` compiles fine but the resulting factory's designated-timestamp index silently becomes `-1` instead of `ts2`'s position, when the DISTINCT→GROUP BY rewrite is disabled (reachable via `DistinctTimeSeriesRecordCursorFactory`). Root cause: `generateSelectDistinct` builds its factory straight off `factory.getMetadata().getTimestampIndex()` without ever consulting `model.hasExplicitTimestamp()`, exactly like `generateNoSelect` did before this PR's fix. Related architectural note: every join participant's factory funnels through one line in `generateJoins` (`slave = generateQuery(slaveModel, executionContext, index > 0)`) regardless of which codegen arm built it — a central enforcement point there would close this bug class in one place rather than requiring each new arm to remember it individually (I tried this centralization and it's real work to get right safely — see the fix-tracking comment below).

2. Two new tests used `.inferTimestamp()`/`.inferRandomAccess()` where the expected values were actually knowable, weakening the regression guard at the metadata-designation level (`QueryAssertion`'s own doc recommends the explicit form whenever the timestamp is known, specifically because the weak form can't catch a query that silently relabels its designated timestamp).

3. ~11 lines of near-identical master-subquery SQL text duplicated between the ASOF and LT JOIN assertions in the new test — risk of the two queries silently desyncing on a future edit.

4. `applyExplicitTimestamp` inserted out of alphabetical order among private methods, per this repo's own `CLAUDE.md` convention.

5. (plausible, unconfirmed) the async top-K "peel" optimization path reaches through a retimestamped factory to pre-relabel base metadata via `getBaseFactory()`; by inspection it recomputes the correct column positionally so is believed safe, but isn't exercised by a test.

One candidate — "uncommitted debug print in SqlOptimiser.java" raised independently by two finder agents — turned out to be a review agent's own transient instrumentation on the shared review worktree, not a real issue; discarded after verifying `git status` was clean.

Fix for finding #1 in a follow-up commit — see next comment.

nwoolmer:
## Fix for finding #1 (generateSelectDistinct gap)

Pushed in c8adb373b8.

**What's fixed:**
- `generateSelectDistinct` now applies the explicit timestamp redesignation *before* deciding which DISTINCT strategy to use (`DistinctTimeSeriesRecordCursorFactory` vs `DistinctRecordCursorFactory`) — that ordering matters: `DistinctTimeSeriesRecordCursor`'s dedup fast path reads its row-adjacency decision directly off the factory's own `timestampIndex`, so relabeling metadata only *after* the method returns would be too late to affect that choice.
- `SqlOptimiser#rewriteDistinct` (the DISTINCT→GROUP BY rewrite) now propagates the explicit-timestamp *flag* to its wrapper models, not just the timestamp *expression* it was already copying — a real, isolated correctness fix, verified inert for non-DISTINCT queries.

**What's still open:** the general case — DISTINCT over a re-timestamped subquery *with* the (default-on) DISTINCT→GROUP BY rewrite enabled — isn't closed yet. I attempted a fix in `generateSelectGroupBy` / a central post-`generateSelect()` wrap and reverted it: it corrupted row-by-row random access on an unrelated query (an ASOF JOIN whose master was a SAMPLE BY subquery) by double-relabeling `generateSelectChoose`'s already-correct, join-backed output. Left a detailed comment on `applyExplicitTimestamp` explaining what was tried and why, so a future fix doesn't repeat the same unsafe shape. Tracking this as a known gap rather than shipping something regression-risky under time pressure.

**Testing:** new regression test in `DistinctTimeSeriesMemoryTrackerTest` (the file that already exists specifically to exercise `generateSelectDistinct`'s `DistinctTimeSeriesRecordCursorFactory` path with the rewrite disabled). Full clean rebuild + full `AsOfJoinTest` class (118 tests) + all `Distinct*Test` classes + the original curated 67-class/4709-test regression sweep — all 0 failures (repeated 3x for the directly-affected tests to rule out flakiness). One meta-note: while chasing an apparent regression during this work, I found and reverted an earlier bad `replace_all` edit of my own that had accidentally stripped `.noRandomAccess()` from ~50 unrelated pre-existing tests in `AsOfJoinTest.java` — never made it into a pushed commit, but flagging in case it's useful context for why this took a few iterations.

bluestreak01 (changes requested):
@nwoolmer

### Critical

1. **[pre-existing] Default `DISTINCT` still drops the explicit timestamp**  
   `core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java:7407-7419`  
   `core/src/main/java/io/questdb/griffin/SqlOptimiser.java:8081,8084`

   The production-default DISTINCT-to-GROUP-BY rewrite reaches `generateSelectGroupBy()`, whose general keyed path never applies the explicit designation. I reproduced:

   ```sql
   SELECT DISTINCT * FROM (SELECT * FROM tab) timestamp(ts2)
   ```

   The factory reported timestamp index `-1` instead of `ts2` at index `1`. The new test disables the rewrite, so it cannot catch this production path.

   The merge-base source confirms this bug predates the PR, but this PR changes `rewriteDistinct()` and explicitly visits the defect. Fix the non-SAMPLE-BY GROUP BY return paths here and add a default-configuration regression test asserting `.timestamp("ts2")`, rows, plan, and a downstream temporal consumer.

2. **[in-diff] `applyExplicitTimestamp()` leaks its generated factory if wrapper construction fails**  
   `core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java:7427-7445`

   Only `getTimestampIndex()` executes inside the cleanup catch. Afterward, metadata copying, `IntList` allocation/growth, `SelectedRecordCursorFactory` allocation, or its eager cursor allocation can throw `OutOfMemoryError`. Neither caller can recover the hidden factory: `generateNoSelect()` returns directly, while `generateSelectDistinct()` enters its cleanup `try` only after the helper returns.

   Wrap the complete ownership-transfer sequence in `catch (Throwable)` and free `factory` unless construction succeeds.

3. **[in-diff] Retimestamping performs avoidable Θ(C) compilation work where Θ(1) is possible**  
   `core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java:7438-7445`  
   `core/src/main/java/io/questdb/cairo/GenericRecordMetadata.java:125-130`  
   `core/src/main/java/io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java:66-84`

   A differing timestamp causes three column-width passes: metadata copying, identity-index construction, and identity-index validation. It also retains Θ(C) metadata/map/list state. Result width is not structurally bounded by a small constant, and this runs during query compilation.

   Use a delegating metadata view plus a dedicated transparent retimestamping factory. Every metadata operation can delegate except `getTimestampIndex()`, making construction Θ(1) in column count.

4. **[in-diff] The new LT JOIN regression assertion does not detect the bug**  
   `core/src/test/java/io/questdb/test/griffin/engine/join/AsOfJoinTest.java:6376-6404`

   Correct `s_f_ts=18:24` and broken `s_ts=18:00` both select the same strict predecessor, `17:56/200`. Reverting either production fix leaves the entire LT result unchanged.

   Add an RHS row between the two timestamps, for example `18:10/250`, and expect it from LT. The broken path would continue returning `17:56/200`.

5. **[in-diff] The join tests infer the metadata that the fix must explicitly prove**  
   `core/src/test/java/io/questdb/test/griffin/engine/join/AsOfJoinTest.java:6366-6367,6396-6397`

   `.inferTimestamp()` accepts a lost or incorrectly relabeled designated timestamp, while `.inferRandomAccess()` accepts either cursor capability. Identical rows can therefore pass with broken final metadata.

   Replace them with:

   - ASOF: `.timestamp("s_f_ts")`; retain the default random-access expectation.
   - LT: `.timestamp("s_f_ts").noRandomAccess()`.

6. **[in-diff] Several changed user-visible paths remain untested**

   - The rewrite-disabled DISTINCT test uses globally unique rows with both timestamps advancing together, so it does not verify timestamp-dependent adjacency/deduplication or the “apply before strategy selection” requirement.
   - Nullable alternate TIMESTAMP values remain uncovered.
   - LATEST, ORDER BY/top-K, and LIMIT consumers after bare-subquery redesignation remain uncovered.

   Add repeated projected rows whose natural and redesigned timestamp groups differ, NULL timestamp cases, and focused downstream-consumer tests.

### Moderate

1. **[in-diff] PR metadata understates scope and limitations**

   The title/body describe only the join-root fix, omitting the bundled DISTINCT changes and the admitted default-rewrite defect. The test plan also:

   - uses checkboxes instead of plain bullets;
   - calls two errors “unrelated” and “pre-existing” without evidence;
   - claims “150+” `AsOfJoinTest` tests, while the current targeted run reported 119.

   Update the title/body to describe all bundled behavior and verified tradeoffs. This is runtime-inert because it changes documentation only.

2. **[in-diff] The ASOF and LT tests duplicate a long master-subquery block**  
   `core/src/test/java/io/questdb/test/griffin/engine/join/AsOfJoinTest.java:6349-6404`

   Extract the shared SQL fragment or helper so the two regression variants cannot silently diverge. This is currently behaviorally and performance-wise inert.

### Minor

1. **[in-diff] Member placement and test setup violate repository conventions**

   - `applyExplicitTimestamp()` is not alphabetically placed among private methods.
   - New public `AsOfJoinTest` methods appear after private helpers.
   - `testExplicitTimestampRedesignationSurvivesDistinct()` is not alphabetically placed.
   - `DistinctTimeSeriesMemoryTrackerTest.java:147-148` uses two single-row INSERTs instead of one multi-row INSERT.

- **Time-frame cursors use stale timestamps:** dismissed. For page-frame/table paths, timestamp propagation reaches table generation first, and the helper becomes a no-op or uses a crossed projection. An empirical factory chain reported timestamp index `1` consistently through the factory and cursor. Differing-index helper paths such as SAMPLE BY do not expose time-frame cursors.
- **The extra wrapper disables parallel top-K:** dismissed. No concrete adjacent peelable-wrapper chain attributable to this PR was found; the helper is either separated by JOIN/DISTINCT or is a no-op on table projections.
- **Missing timestamp-column error is untested:** dismissed. `SqlParserTest.testSelectWildcardAndNotTimestamp()` already asserts the missing-column error, before a factory exists; the new non-TIMESTAMP test covers the helper’s cleanup catch.

### Coverage map

Dimensions: **H** happy path, **E** error path, **N** NULL, **B** boundary, **Con** concurrency, **R** cleanup.

|---|---|---|---|
| Join-root redesignation drives ASOF master matching | `testAsOfJoinMasterSubqueryExplicitTimestampAfterSampleBy`; broken path returns `17:56/200` instead of `18:24/300` | H ✓, E N/A, N U, B ✓, Con N/A, R ✓ | TESTED |
| Same behavior through LT JOIN | Current expected result is identical before/after the fix | H U, E N/A, N U, B U, Con N/A, R ✓ | **UNTESTED** |
| Final joined factory designates `s_f_ts` and reports expected random access | No test: inference adopts the factory’s answer | H U, E N/A, N N/A, B U, Con N/A, R ✓ | **UNTESTED** |
| Explicit timestamp already equals natural timestamp | `test...MatchesNaturalTimestampIsNoop`; timestamp pinned and plan must omit `SelectedRecord` | H ✓, E N/A, N N/A, B ✓, Con N/A, R ✓ | TESTED |
| Explicit target has non-TIMESTAMP type | `testSubqueryExplicitTimestampToNonTimestampColumnRejected`; asserts position/message and leak scope | H N/A, E ✓, N N/A, B ✓, Con N/A, R ✓ | TESTED |
| Explicit target column is absent | `SqlParserTest.testSelectWildcardAndNotTimestamp`; asserts `Invalid column` before factory creation | H N/A, E ✓, N N/A, B ✓, Con N/A, R N/A | TESTED |
| Wrapper construction throws after validation | No ownership-failure test | H N/A, E U, N N/A, B U, Con N/A, R U | **UNTESTED** |
| Direct rewrite-disabled DISTINCT retains `ts2` and routes to time-series factory | `testExplicitTimestampRedesignationSurvivesDistinct`; timestamp and plan fail without helper | H ✓, E N/A, N U, B ✓, Con N/A, R ✓ | TESTED |
| DISTINCT uses redesigned timestamp for adjacency/strategy choice | Existing rows are unique and both timestamps advance together | H U, E N/A, N U, B U, Con N/A, R ✓ | **UNTESTED** |
| Default DISTINCT-to-GROUP-BY preserves explicit timestamp | No test; empirical probe reports timestamp index `-1` | H U, E N/A, N U, B U, Con N/A, R U | **UNTESTED** |
| Redesigned TIMESTAMP contains sentinel NULL | No test | H U, E N/A, N U, B U, Con N/A, R U | **UNTESTED** |
| LATEST/ORDER BY/top-K/LIMIT consume redesignated metadata | No focused test with distinguishing timestamp values | H U, E N/A, N U, B U, Con N/A, R U | **UNTESTED** |
| No explicit timestamp | Verified semantic delta: original factory is returned unchanged | All dimensions N/A: no behavioral change | EXEMPT |

### Summary

**Verdict: request changes.**

Coverage map: **12 behavioral changes, 5 tested, 7 UNTESTED; plus 1 EXEMPT no-change row**.

The targeted core tests completed with **123 tests, 0 failures/errors**. The enclosing reactor later failed because the benchmarks module had no matching tests. A separate default-configuration DISTINCT probe failed with timestamp index `-1`, confirming the open production bug.

The PR adds Θ(C) compilation work and retains a known default-DISTINCT correctness gap. Nine draft findings were verified; three false positives were removed. Split: **8 in-diff, 0 out-of-diff, 1 pre-existing**. A wider cross-context pass ran; its time-frame and top-K candidates were dismissed after source and empirical verification.

bluestreak01 (changes requested):
@nwoolmer

Reviewing PR #7369 at level 3.

## Critical

1. **[in-diff] Retimestamped time-frame cursors mix timestamp domains and can return wrong temporal-join results**

   - `core/src/main/java/io/questdb/griffin/engine/table/RetimestampedRecordCursorFactory.java:157-190,221-224,268-355`
   - Sequential `getTimestampIndex()` reports the redesigned column, but `TimeFrameCursorImpl.open()` reads exact bounds using the base metadata timestamp at `TimeFrameCursorImpl.java:238-269`. Estimates and `seekEstimate()` also use physical partition timestamps at `:323-332`.
   - Concurrent exact bounds use the redesigned index, but estimates and seeking still use physical partition bounds in `ConcurrentTimeFrameCursorImpl.java:122-152,230-240,273-281`.
   - ASOF/LT/WINDOW/HORIZON consumers treat these bounds as authoritative, e.g. `AbstractAsOfJoinFastRecordCursor.java:302-320`. A retimestamped slave can therefore skip the correct frame and silently return the wrong match.
   - Fix this PR by either implementing bounds/seeking in the redesigned timestamp domain or disabling time-frame support for this wrapper. Add slave-side ASOF/LT and initialized concurrent WINDOW/HORIZON tests with timestamp ranges crossing physical partitions.

2. **[in-diff] `Retimestamp -> Filter` disables existing fast execution plans**

   - `RetimestampedRecordCursorFactory.java:203-214`
   - The wrapper reports neither page-frame support nor filter stealing when its base is an `AsyncFilteredRecordCursorFactory`.
   - This hides the page-frame leaf from:
     - parallel GROUP BY: `SqlCodeGenerator.java:9006-9027`;
     - filtered ASOF/LT: `:5181-5234,5314-5360`;
     - filtered parallel top-K: `:1058-1065,7564-7634`.
   - Consequences include parallel-to-serial execution and, for encoded top-K fallbacks, O(N log K)/O(K) becoming O(N log N)/O(N).
   - Add a retimestamp-aware peel/rewrap path that preserves metadata while transferring the nested filter. Cover keyed/non-keyed ASOF/LT, parallel GROUP BY, and encoded top-K plans.

3. **[in-diff] Wrapper construction is O(columns), despite the O(1) claim**

   - `SqlCodeGenerator.java:1465`
   - `GenericRecordMetadata.copyOfNew()` traverses every column in `GenericRecordMetadata.java:36-39`.
   - This adds O(C) compilation time and storage for every differing redesignation. An O(1) read-only metadata decorator can delegate every property except `getTimestampIndex()`.
   - Either use that approach or accurately document and justify the O(C) cost. Under the review’s compile-path algorithm gate, the current avoidable traversal is blocking.

4. **[in-diff] Sequential time-frame acquisition allocates on every execution**

   - `RetimestampedRecordCursorFactory.java:157-162`
   - Every reused compiled factory creates another Java wrapper. Temporal joins call this from `getCursor()`, making it execution-path GC pressure.
   - Cache and rebind the sequential adapter as `SelectedRecordCursorFactory.java:181-190` already does.

5. **[in-diff] Concurrent wrapper allocation failure leaks native resources**

   - `RetimestampedRecordCursorFactory.java:183-190`
   - `base.newTimeFrameCursor()` returns a fresh externally-owned `ConcurrentTimeFrameCursorImpl`. If allocation of `SelectedConcurrentTimeFrameCursor` throws, nothing closes the base cursor.
   - That cursor owns a `PageFrameMemoryPool` and native `DirectLongList` resources (`ConcurrentTimeFrameCursorImpl.java:81-99`).
   - Guard the ownership transfer with `catch (Throwable)` and `Misc.free(baseCursor)`. Add close-counting failure injection.

6. **[pre-existing] `SelectedRecordCursorFactory` has the same concurrent-cursor leak**

   - `core/src/main/java/io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java:208-213`
   - This visited implementation performs the same acquire-then-wrap sequence without cleanup.
   - Apply the same guarded transfer and test in this PR.

7. **[pre-existing] `generateSelectDistinct()` leaks and double-closes resources on compilation failures**

   - `SqlCodeGenerator.java:8566-8597`
   - If `getLoFunction()` succeeds and `getHiFunction()` throws, the low function is lost; the catch frees only `factory`.
   - Both DISTINCT factory constructors close their transferred base on constructor failure:
     - `DistinctRecordCursorFactory.java:79-91,156-160`
     - `DistinctTimeSeriesRecordCursorFactory.java:60-87,126-129`
   - The outer catch then closes the same factory again. Neither `Misc.free()` nor `AbstractRecordCursorFactory.close()` provides a close-once guard.
   - Refactor around explicit ownership flags/locals and add failure-injection tests for both constructors and two-sided LIMIT compilation.

8. **[in-diff] The `generateSelectDistinct()` fix has no regression test**

   - Production hunk: `SqlCodeGenerator.java:8566`
   - Tests: `DistinctTimeSeriesMemoryTrackerTest.java:138-191`
   - Both tests obtain `ts2` from optimizer/table generation before this call. Reverting only line 8566 leaves their metadata, plan, adjacency, and row assertions unchanged.
   - Add a rewrite-disabled fixture where the generated random-access subquery still reports the old/no timestamp before `applyExplicitTimestamp()`. If no reachable model has that shape, remove the redundant production call instead.

9. **[in-diff] The new `SCAN_DIRECTION_OTHER` behavior is untested**

   - Production: `RetimestampedRecordCursorFactory.java:135-148`
   - Test: `DistinctTest.java:99-110`
   - `.inferTimestamp()` accepts `OTHER`; it does not pin the direction. The test data are already ordered by `ts2`, so an incorrect `FORWARD` result would also pass.
   - Add a direct direction assertion plus an outer `ORDER BY ts2` case with unordered GROUP BY output, asserting that a real sort occurs.

10. **[in-diff] New page-frame, shared-cursor, and optimizer-guard contexts lack execution coverage**

    - `RetimestampedRecordCursorFactory.java:82-86,129-132,151-154,175-180,203-219`
    - The white-box test compares capability booleans but does not execute page-frame/shared cursors, nor exercise filtered retimestamping through top-K, projection peeling, or filter-stealing gates.
    - These paths are concurrency-, ownership-, and result-sensitive. Add end-to-end page-frame, shared-cursor, filtered temporal-slave, and top-K tests.

11. **[in-diff] The helper’s failure-cleanup branch is not exercised**

    - Production: `SqlCodeGenerator.java:1458-1470`
    - Claimed test: `AsOfJoinTest.java:6404-6415`
    - The non-TIMESTAMP query fails in optimizer validation before `applyExplicitTimestamp()`. `assertMemoryLeak()` cannot cover code that never runs.
    - Add a close-counting metadata/factory seam that throws during metadata copying and verifies exactly one close, or remove the incorrect coverage claim.

12. **[PR metadata] Enterprise CI is failing at the current head**

    - Head: `ec800fc82e1862c61bdbc0bcc3d542d6644f4c45`
    - `enterprise-ci` reports failure for Azure build `250441`; no diagnosis was publicly available.
    - The body nevertheless says regression sweeps are “all green” and socket-binding errors are the only non-passing results.
    - Investigate the live failure, rerun it, and update the body with the actual status. It cannot be treated as environmental or unrelated without evidence.

## Moderate

1. **[in-diff] The wrapper resets `fragmentedSymbolTables()` to false**

   `RetimestampedRecordCursorFactory` does not delegate the property, while UNION factories return true. Current production code has no consumer beyond test infrastructure, so this cannot presently alter query results or measurable execution cost. Delegate the capability and add a UNION retimestamp test.

2. **[in-diff] The null time-frame-cursor branches are untested**

   `RetimestampedRecordCursorFactory.java:159-161,184-187` correctly returns null, but no focused test pins either branch. This is an internal bounded branch and the implementation is behaviorally/performance-wise inert as written.

3. **[in-diff] Test code contains excessive implementation narration and an oversized misplaced white-box test**

   `AsOfJoinTest.java:6283-6387` embeds roughly 100 lines of low-level factory testing in a join suite. Comments at `DistinctTest.java:80-98`, `DistinctTimeSeriesMemoryTrackerTest.java:139-147`, and `AsOfJoinTest.java:6253-6259,6406-6411` preserve review history rather than durable test intent. Moving this to a focused factory/time-frame test is behaviorally and performance-wise inert in production.

## Minor

1. **Test-style violations**
   - Multiple one-row INSERTs instead of one multi-row INSERT:
     - `DistinctTest.java:87-89`
     - `AsOfJoinTest.java:1378-1389,1434-1435`
   - Missing numeric separators at `AsOfJoinTest.java:6296-6297`.
   - `findRetimestamped()` interrupts the public test block and uses legacy `instanceof` plus cast at `AsOfJoinTest.java:6389-6402`; move it to the private helper section and use a pattern variable.

2. **Metadata conventions**
   - Suggested title: `fix(sql): fix explicit timestamp redesignation in subqueries`.
   - Branch commit subjects exceed 50 characters and authored commits use forbidden Conventional prefixes. Because this PR is squash-merged, do not rewrite branch history solely for this; ensure the final metadata is compliant.

- **“Forward scan is invalid whenever the timestamp index changes.”** Dismissed: QuestDB’s `timestamp(col)` clause asserts ordered input; forwarding direction for a timestamp-bearing base follows the existing contract.
- **“Missing `changePageFrameSizes()` delegation causes a WINDOW/HORIZON regression.”** Dismissed: current forwarding chains terminate in an empty default, and WINDOW factories set frame sizes on `SqlExecutionContext` at execution time.
- **“All retimestamped top-K queries lose AsyncTopK.”** Narrowed: a plain page-frame base remains eligible. Only the filtered-wrapper shape is broken.
- **“Sequential wrapper allocation failure permanently leaks its base cursor.”** Dismissed: the concrete sequential base caches that cursor and frees it when the factory closes.
- **“Concurrent per-worker wrapper allocation is inherently avoidable.”** Dismissed: the API requires independent worker-usable cursor instances; only the unguarded failure path is defective.
- **“Missing `getTableToken()` delegation produces wrong self-join results.”** Dismissed: the search established only conservative self-join detection loss, not a reachable wrong-result path.
- **“The wrapper loses small-frame tuning through `changePageFrameSizes()`.”** Dismissed after tracing actual execution-context sizing.
- **“Current tests cover helper cleanup because they use `assertMemoryLeak()`.”** Dismissed: the exception occurs before the helper.

## Coverage map

Legend: H happy path, E error, N NULL, B boundary, C concurrency, R cleanup. `—` means inapplicable with the reason stated.

| # | Change/path | Test and failure link | Dimensions | Verdict |
|---|---|---|---|---|
| 1 | Join-root timestamp retention | `AsOfJoinTest#testAsOfJoinMasterSubqueryExplicitTimestampAfterSampleBy`; reverting changes ASOF/LT balances | H✓ E— N✗ B✓ C— R✓ | TESTED |
| 2 | DISTINCT rewrite flag propagation | `DistinctTest#testDistinctExplicitTimestampSurvivesGroupByRewrite`; index becomes `-1` | H✓ E— N✗ B✓ C— R✓ | TESTED |
| 3 | Bare subquery, differing timestamp | `testSubqueryExplicitTimestampDrivesLatestOn` and NULL variant; compilation/plan/results fail without wrapper | H✓ E✗ N✓ B✓ C— R✓ | TESTED |
| 4 | Same-index no-op | `test...MatchesNaturalTimestampIsNoop`; unexpected wrapper fails plan assertion | H✓ E— N— B✓ C— R✓ | TESTED |
| 5 | No-explicit pass-through | Literal identity return; verified no behavioral delta | all — | EXEMPT |
| 6 | Helper failure cleanup | Non-TIMESTAMP test fails upstream | H— E✗ N— B— C— R✗ | **UNTESTED Critical** |
| 7 | Direct DISTINCT pre-strategy helper | New tests pass if hunk is reverted | H✗ E— N✗ B✗ C— R✓ | **UNTESTED Critical** |
| 8 | SAMPLE BY direct return | Structural extraction only; previous direct return preserved | all — | EXEMPT |
| 9 | Non-SAMPLE GROUP BY retimestamp | Direct metadata and plan assertions in `DistinctTest` | H✓ E✗ N✗ B✓ C— R✓ | TESTED |
| 10 | Base has no timestamp → `OTHER` | `.inferTimestamp()` does not pin direction | H✗ E— N— B✗ C— R— | **UNTESTED Critical** |
| 11 | Base has timestamp → delegate direction | White-box equality and ASOF/LT result | H✓ E— N— B✓ C— R— | TESTED |
| 12 | Record/metadata/NULL pass-through | LATEST, DISTINCT and join result assertions | H✓ E— N✓ B✓ C— R✓ | TESTED |
| 13 | Sequential base cursor returns null | No exact test | H✗ E— N— B✗ C— R— | **UNTESTED Moderate** |
| 14 | Sequential index/record delegation | White-box index and 72-row assertions | H✓ E— N✗ B✓ C— R✓ | TESTED |
| 15 | Sequential frame bounds and seek | Calls are made but results are not asserted; implementation is wrong | H✗ E— N✗ B✗ C— R✓ | **UNTESTED Critical** |
| 16 | Concurrent `of/open/seek` | Test reads index before initialization only | H✗ E✗ N✗ B✗ C✗ R✓ | **UNTESTED Critical** |
| 17 | Anti-peel/filter/projection guards | No filtered slave/top-K/parallel context | H✗ E✗ N— B✗ C✗ R✗ | **UNTESTED Critical** |
| 18 | Page-frame cursor execution | Only capability boolean compared | H✗ E✗ N— B✗ C✗ R✗ | **UNTESTED Critical** |
| 19 | Shared-cursor execution | Only capability boolean compared | H✗ E✗ N— B✗ C✗ R✗ | **UNTESTED Critical** |
| 20 | Scalar capability delegation | Direct equality assertions cover listed methods | H✓ E— N— B✓ C— R— | TESTED |
| 21 | UPDATE support/table identity | No reachable changed UPDATE behavior established | all — | EXEMPT |
| 22 | Normal wrapper/base close | Try-with-resources plus leak harness observes leaks | H✓ E— N— B✓ C— R✓ | TESTED |
| 23 | O(C) metadata copying | No wide-schema or allocation test; avoidable traversal confirmed | H✗ E— N— B✗ C— R— | **UNTESTED Critical** |
| 24 | Sequential per-execution allocation | No allocation/reuse assertion | H✗ E— N— B✗ C— R— | **UNTESTED Critical** |
| 25 | Temporal-slave fast paths | No retimestamped ASOF/LT/WINDOW/HORIZON slave test | H✗ E✗ N✗ B✗ C✗ R✗ | **UNTESTED Critical** |
| 26 | UNION fragmented-symbol capability | No wrapper-specific UNION test | H✗ E— N— B✗ C— R— | **UNTESTED Moderate** |

Totals: **26 behavioral rows: 10 TESTED, 13 UNTESTED, 3 EXEMPT**. The UNTESTED set contains 11 Critical and 2 Moderate rows.

## Summary

**Verdict: request changes.**

The wrapper introduces a reachable silent wrong-result bug in temporal-join slave paths, blocks several optimized filtered plans, adds avoidable compile/execution allocations, and contains a native-resource leak on concurrent cursor construction failure. Two visited pre-existing resource defects also require correction in this PR.

The sequential targeted validation passed: **149 tests, 0 failures/errors** across `DistinctTest`, `DistinctTimeSeriesMemoryTrackerTest`, and `AsOfJoinTest`. That does not resolve the missing failure links or the undiagnosed enterprise CI failure.

Coverage map: **26 changes, 10 tested, 13 UNTESTED, 3 EXEMPT**.

Review accounting: **17 findings verified** — 13 in-diff, 0 at out-of-diff callsites, 2 pre-existing, and 2 PR-metadata findings. **Five false-positive finding groups were removed or materially narrowed.**

nwoolmer:
Addressed the 2026-07-11 level-3 review at the current head.

**C1 (approach):** the redesignation wrapper (`RetimestampedRecordCursorFactory`) deliberately disables time-frame support (`supportsTimeFrameCursor()`/`supportsFilterStealing()` → false). Implementing seek/estimate in the redesignated-timestamp domain is not feasible: the base frame bounds derive from *physical partition* floors/ceilings, and an arbitrary non-partition-key `ts2` has no partition-metadata correspondence — correct estimates would require eagerly scanning every frame`s `ts2` min/max, defeating estimation and requiring invasive changes to shared `TimeFrameCursorImpl`/`ConcurrentTimeFrameCursorImpl`. Temporal consumers fall back safely: ASOF/LT use the record-cursor path (reads the `ts2` index), WINDOW/HORIZON reject cleanly with a `SqlException`. Covered by a crossing-partition ASOF/LT slave test that regresses if reverted.

**C2–C11:** resolved (O(1) metadata decorator, concurrent/DISTINCT leak guards + failure-injection tests, SCAN_DIRECTION_OTHER, page-frame/shared-cursor/top-K execution coverage). Also fixed a silent master-merge break (parallel GROUP BY declarations misplaced across `generateSelectGroupBy`/`generateSelectGroupBy0`). 270 targeted tests green.

**C12:** enterprise-ci needs a re-run on the merged head (handled separately).

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 105 / 129 (81.40%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/RetimestampedRecordCursorFactory.java|48|68|70.59%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|26|30|86.67%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/DistinctTimeSeriesRecordCursorFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/DistinctRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/test/TestAllocatingFunctionFactory.java|10|10|100.00%|

bluestreak01 (changes requested):
@nwoolmer

### Critical

1. **[in-diff] GROUP BY rejects valid projections that omit the input timestamp**  
   `core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java:8860-8870`

   `generateSelectGroupBy()` applies `timestamp(ts2)` after building the aggregate projection. If that projection omits `ts2`, `getTimestampIndex()` resolves against output metadata and throws:

   ```sql
   SELECT k, count()
   FROM (SELECT * FROM tab WHERE v > 0) timestamp(ts2)
   ```

   Current head fails with `Invalid column: ts2`; before this hunk, the query returned an undesignated `k,count` result.

   **Fix:** reapply the designation only when the source timestamp survives the projection. Track source-column identity rather than matching an output alias by name.

2. **[in-diff] Retimestamping reports ordering from the wrong timestamp domain**  
   `core/src/main/java/io/questdb/griffin/engine/table/RetimestampedRecordCursorFactory.java:131-135`  
   `core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java:7599-7606`

   `getScanDirection()` delegates the base direction even though that direction describes the old timestamp. A contract-valid reproduction orders the base by natural `ts DESC`, which yields redesignated `ts2 ASC`, then applies `timestamp(ts2)`. An outer `ORDER BY ts2 DESC` is incorrectly elided because the wrapper reports the old `BACKWARD` direction. The query returns `01:00, 02:00` instead of `02:00, 01:00`.

   **Fix:** store direction explicitly in the wrapper. Bare dynamic-timestamp passthrough should report `FORWARD`; unordered GROUP BY output should report `OTHER`. Do not infer direction from the old timestamp.

3. **[out-of-diff callsite broken by this change] `alloc()` now leaks native memory in successful LIMIT queries**  
   `core/src/main/java/io/questdb/griffin/engine/functions/test/TestAllocatingFunctionFactory.java:112-114`  
   `core/src/main/java/io/questdb/griffin/engine/LimitRecordCursorFactory.java:150-152,192-194`

   Marking `alloc()` runtime-constant makes it legal in `LIMIT`, but `LimitRecordCursorFactory` never closes its limit functions. The following completes successfully, closes its factory, and leaves `TestAllocatingFunctionFactory.getCloseCount()` at zero:

   ```sql
   SELECT * FROM tab LIMIT alloc(64)
   ```

   Other limited-sort and async-filter owners have the same omission.

   **Fix:** preferably revert the broad `alloc()` semantic change and use the existing `test_close_counter('64')::LONG` runtime-constant helper in the DISTINCT rollback test. Otherwise audit and fix every LIMIT-function owner.

4. **[in-diff] New rollback paths can mask the primary error and skip later cleanup**  
   `core/src/main/java/io/questdb/griffin/SqlCodeGenerator.java:1441-1443,8852-8856`  
   `core/src/main/java/io/questdb/griffin/engine/table/RetimestampedRecordCursorFactory.java:72-78`  
   `core/src/main/java/io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java:96-102`

   These paths call plain `Misc.free()` while another exception is in flight. If `close()` throws, it replaces the compilation/construction error. In `generateSelectDistinct()`, it also prevents remaining limit functions from being closed.

   **Fix:** use `Misc.free(resource, primary)` for every rollback resource and rethrow the original exception with close failures suppressed. Apply the same correction to the new DISTINCT-time-series cleanup.

5. **[in-diff, UNTESTED bug-fix hunk] DISTINCT wrapper flag propagation has no failure-linked regression**  
   `core/src/main/java/io/questdb/griffin/SqlOptimiser.java:8108,8111`

   Removing both new `setExplicitTimestamp()` calls still leaves `DistinctTest#testDistinctExplicitTimestampSurvivesGroupByRewrite` green. No changed test proves these setters affect observable behavior.

   **Fix:** add a query shape that fails when these exact setters are removed, or remove the redundant change and correct the PR description.

6. **[in-diff, UNTESTED resource lifecycle] Failure tests bypass production wiring and post-allocation cleanup**  
   `core/src/main/java/io/questdb/griffin/engine/table/SelectedRecordCursorFactory.java:225-234`  
   `core/src/main/java/io/questdb/griffin/engine/groupby/DistinctRecordCursorFactory.java:84-94`  
   `core/src/main/java/io/questdb/griffin/engine/groupby/DistinctTimeSeriesRecordCursorFactory.java:65-93`

   `testConcurrentTimeFrameCursorClosedWhenWrapperConstructionFails()` invokes the helper directly. Reverting the production `newTimeFrameCursor()` callsite leaves it green. The DISTINCT proxy test throws on the second metadata lookup, before either local cursor/map allocation, so it does not cover the newly added local cleanup.

   **Fix:** inject failure through the production entrypoint and after local resource creation, then assert exact-once cleanup.

7. **[in-diff, UNTESTED user-visible behavior] The DISTINCT adjacency regression test is ineffective**  
   `core/src/test/java/io/questdb/test/griffin/DistinctTimeSeriesMemoryTrackerTest.java:181-206`

   The duplicate pair has both equal natural `ts` and equal redesignated `ts2`. Using either timestamp index therefore removes the same row, so the assertion cannot prove that `ts2` drives adjacency.

   **Fix:** select only `(ts2, v)` from valid ascending `ts2` data where natural `ts` changes within one `ts2` group, e.g. `v=1,2,1`. Correct `ts2` adjacency removes the final duplicate; natural-`ts` adjacency emits it.

### Moderate

1. **PR description contains inaccurate and stale claims**  
   It says `applyExplicitTimestamp()` is used by `generateSelectDistinct()`, but its only production callsites are `generateNoSelect()` and `generateSelectGroupBy()`. It also says the wrapper changes a time-frame cursor’s index, while the implementation disables time-frame cursors. Reported test counts are stale (`DistinctTimeSeriesMemoryTrackerTest` now has six tests).

   This is behaviorally and performance-wise inert because it affects PR documentation only.

2. **The new factory test class is organized around implementation seams rather than behavior**  
   `core/src/test/java/io/questdb/test/griffin/engine/table/SelectedRecordCursorFactoryTest.java:43-181`

   Three of four tests primarily cover DISTINCT or retimestamped factories. The test name `testDistinctConstructorsCloseTransferredResourcesOnFailure` contradicts its assertion that incoming resources remain caller-owned. Public production constructor interfaces exist solely for these white-box tests.

   This is behaviorally and performance-wise inert because it affects test maintainability, not runtime execution.

### Minor

- `AsOfJoinTest.java:1406-1408` says LT JOIN selects `17:56/200`; the inserted and asserted latest preceding row is `18:10/250`.
- `RetimestampedMetadata` methods are not alphabetized (`getColumnIndexType`, `getColumnCount`, `getColumnIndexQuiet`).
- New test methods in `SelectedRecordCursorFactoryTest` are not alphabetized, and `AsOfJoinTest#findRetimestamped` is a static helper placed among instance helpers with an extra blank line.
- Ensure the eventual squash commit uses a plain-English subject and body; do not rewrite the throwaway branch history.

### Downgraded

- **Unsorted `timestamp(ts2)` reproductions:** dismissed because QuestDB explicitly requires dynamic timestamp input to already be ascending by the elected column. The scan-direction finding remains confirmed through the separate, valid opposing-direction reproduction.
- **LATEST ON over intentionally unordered dynamic timestamps:** dismissed for the same documented contract violation.
- **DISTINCT `(10,20,10)` reproduction:** dismissed as invalid dynamic-timestamp input; only the ineffective regression test remains.
- **Right-side join timestamp breaks first-input generation:** dismissed. Parser scoping keeps post-join `timestamp(rts)` on a wrapper over the completed join; qualified `timestamp(r.rts)` is invalid syntax.
- **Missing join-root regression:** dismissed. Empirically reverting only the join-count guard makes `testAsOfJoinMasterSubqueryExplicitTimestampAfterSampleBy` fail with balance `200` instead of `300`.
- **Missing Parquet-pruning delegation:** dismissed. WHERE pushdown reaches `read_parquet()` before the retimestamp wrapper is generated.
- **Blanket parallel Top-K/GROUP BY regression:** dismissed. Common filtered queries still generate Async Top-K/Group-By plans; temporal time-frame disabling is required for correctness.

### Coverage map

Legend: H=happy path, E=error path, N=NULL, B=boundary, C=concurrency, R=resource cleanup.

| # | Change | Test and failure link | Dimensions | Verdict |
|---|---|---|---|---|
| 1 | Join-root retains explicit timestamp | `AsOfJoinTest#testAsOfJoinMasterSubqueryExplicitTimestampAfterSampleBy`; guard reversion changes `300` to `200` | H✓, B✓; E/N/C N/A; R✓ | TESTED |
| 2 | DISTINCT wrappers copy explicit flag | No assertion fails when both setters are removed | H✗, B✗; E/N/C/R N/A | **UNTESTED** |
| 3 | Bare subquery applies differing timestamp | `testSubqueryExplicitTimestampDrivesLatestOn`; reversion loses required designation | H✓, B✓, R✓; E/C N/A; N covered by #18 | TESTED |
| 4 | Matching explicit timestamp is a no-op | `testAsOfJoinMasterSubqueryExplicitTimestampMatchesNaturalTimestampIsNoop`; asserts no wrapper | H✓, B✓, R✓; E/N/C N/A | TESTED |
| 5 | Wrapper creation failure closes base | `testBaseFactoryClosedWhenRetimestampedWrapperConstructionFails`; asserts original OOME and one close | E✓, B✓, R✓; H/N/C N/A | TESTED |
| 6 | Rollback preserves primary when close throws | No throwing-close test | E✗, B✗, R✗; H/N/C N/A | **UNTESTED** |
| 7 | Record/page/shared delegation; time frames disabled | `testSubqueryExplicitTimestampDoesNotExposePhysicalTimeFrames`, `testRetimestampedSharedCursorDelegates` | H✓, B✓; E/C N/A; N covered by #18; R✓ | TESTED |
| 8 | Unordered GROUP BY wrapper reports `OTHER` | `DistinctTest#testDistinctExplicitTimestampSurvivesGroupByRewrite` directly asserts direction | H✓, B✓, R✓; E/N/C N/A | TESTED |
| 9 | Bare wrapper direction differs from old timestamp domain | No committed opposing-direction test; temporary valid probe returns wrong order | H✗, B✗; E/N/C/R N/A | **UNTESTED** |
| 10 | Default DISTINCT rewrite retains `ts2` | `DistinctTest#testDistinctExplicitTimestampSurvivesGroupByRewrite`; asserts index, plan, rows | H✓, B✓, R✓; E/N/C N/A | TESTED |
| 11 | Regular GROUP BY omits the input timestamp | No test; temporary valid query throws `Invalid column: ts2` | H✗, B✗; E/N/C/R N/A | **UNTESTED** |
| 12 | LIMIT-high compile failure closes LIMIT-low | `testDistinctLimitHiCompilationFailureReleasesLimitLoFunction`; close count must be one | E✓, R✓, B✓; H/N/C N/A | TESTED |
| 13 | DISTINCT constructor failure leaves incoming owners to caller | `testDistinctConstructorsCloseTransferredResourcesOnFailure`; zero closes before caller cleanup | E✓, R✓, B✓; H/N/C N/A | TESTED |
| 14 | Post-allocation constructor cleanup and Selected production wiring | Current tests fail before allocation or call helper directly | E✗, R✗, B✗; H/N/C N/A | **UNTESTED** |
| 15 | Selected helper closes cursor on construction failure | `testConcurrentTimeFrameCursorClosedWhenWrapperConstructionFails`; close count one | E✓, R✓, B✓; H/N/C N/A | TESTED |
| 16 | Runtime-constant `alloc()` closes after successful LIMIT | No committed test; temporary probe observes close count zero | H✗, R✗, B✗; E/N/C N/A | **UNTESTED** |
| 17 | DISTINCT adjacency uses redesignated timestamp | Added test produces identical result using either index | H✗, B✗, R✓; E/N/C N/A | **UNTESTED** |
| 18 | NULL redesignated timestamp survives | `testSubqueryExplicitTimestampWithNullValues`; asserts NULL row and latest non-NULL row | H✓, N✓, B✓, R✓; E/C N/A | TESTED |
| 19 | Non-TIMESTAMP target is rejected | `testSubqueryExplicitTimestampToNonTimestampColumnRejected`; asserts position/message under leak check | E✓, B✓, R✓; H/N/C N/A | TESTED |

### Summary

**Verdict: request changes.**

- Coverage map: **19 behavioral changes, 12 tested, 7 UNTESTED**.
- Verified findings: **10**; false positives removed: **7**.
- Split: **8 in-diff, 1 out-of-diff callsite, 0 pre-existing**, plus one PR-metadata finding.
- Validation: clean test compilation and all newly affected committed tests passed. Independent temporary probes reproduced the GROUP BY rejection, valid retimestamp ordering failure, and LIMIT native-resource leak. A join-root mutation test failed as expected; removing the DISTINCT wrapper-flag setters left its claimed regression test green.

PR #7370, opened 2026-07-08
chore(core): fix adding a covering index over parquet after a column type change

kafka1991:
## Problem

Adding a POSTING covering index (`ALTER TABLE ... ALTER COLUMN ... ADD INDEX TYPE POSTING INCLUDE (...)`) over a table with parquet partitions suspended the WAL table when one of the INCLUDE'd columns had a pending `ALTER COLUMN TYPE`.

Column type changes on parquet partitions are lazy: the partition keeps the source type on disk and materialises the new type only when the data is read or rewritten. The covering-index build decoded each covered column from parquet using its current (post-ALTER) type. For a conversion that crosses the fixed&lt;-&gt;var or symbol-&gt;fixed boundary the parquet decoder cannot produce the target type directly and rejects the request (`requested column type long does not match file column type symbol`), which suspended the table during WAL apply.

## Fix

The covering-index build now handles a pending lazy conversion the same way the O3 write path already does: decode each covered column in a type the decoder can produce, then finish the cast in Java before writing it into the covering sidecar.

- Columns whose parquet-stored type matches their current type are unchanged - decoded and copied directly. The common case has no behavior or performance change.
- fixed-&gt;var, var-&gt;fixed and symbol-&gt;fixed covered columns are decoded in a Rust-producible type and converted with the same per-row converters the O3 merge/rewrite path uses. This adds a per-row-group buffer allocation and a conversion pass, but only for covered columns that actually carry a pending conversion.
- fixed-&gt;fixed, var-&gt;var and symbol-&gt;var stay byte-identical - the decoder converts them during decode.
- All-null row groups are now detected by decoded size rather than by pointer, which is robust against the decoder leaving a dangling aux pointer. This path is otherwise unchanged.

## Tradeoffs and notes

- No storage-format change: parquet partitions stay parquet; ADD INDEX does not rewrite or convert them.
- The conversion reuses the existing shared converters (`chooseParquetDecodeType`, `convertVarColumnToFixed`, `convertFixedColumnToVarchar`/`String`) rather than adding a second implementation, so covering and the O3 write path stay in sync.
- The `O3PartitionJob.java` diff is dominated by unrelated method reordering; the only semantic change there is making `chooseParquetDecodeType` package-visible and adding an `(int srcType, int columnType)` overload so `TableWriter` can reuse it.

## Test plan

- `CoveringIndexTest.testAddPostingCoveringIndexWithPendingLazyConversionParquetWal` - every crossing conversion arm (SYMBOL-&gt;LONG, VARCHAR-&gt;LONG, INT-&gt;VARCHAR, INT-&gt;STRING) plus a pass-through DOUBLE, single partition.
- `CoveringIndexTest.testAddPostingCoveringIndexWithPendingLazyConversionAcrossParquetRowGroupsWal` - the same columns under a tiny parquet row group size, exercising cross-row-group aux rebasing of the converted columns.
- `CoveringIndexTest.testAddPostingCoveringIndexWithPendingLazyConversionNullsAndVarcharSpillWal` - NULLs in every converting column and a spilling (&gt;9 byte) varchar, under tiny row groups.
- Each test asserts the WAL table is not suspended and compares the covering read against the `no_covering` base read for equality.


kafka1991:
## Code review (level 3)

Reviewed across correctness, concurrency, resource management, performance, cross-context caller impact, and tests. Verdict: **approve.** Correct, well-scoped fix; the covering-index build now handles a covered column's pending lazy `ALTER COLUMN TYPE` the same way the O3 write path does. No Critical or Moderate findings.

### Critical
None.

### Moderate
None.

### Minor

**M1 — Stale doc comments (fixed).** The two `covSlotMeta` method-header Javadocs (`accumulateCoveredColumnsFromRowGroup`, `prepareCoveredColumnMmaps`) still described the per-slot layout as "3 longs"; the layout is now 4 longs (added `parquetColType`). Updated both headers to match.

**M2 — Avoidable intermediate buffer + memcpy in the var->fixed covering branch.** `accumulateCoveredColumnsFromRowGroup` mallocs `fixBuf`, converts into it, then `dataMem.putBlockOfBytes(fixBuf, fixSize)` and frees. Since the destination size is known exactly upfront and `convertVarColumnToFixed` writes positionally by row index, the converter could target `dataMem.appendAddressFor(fixSize)` directly, saving one malloc/free and one full memcpy (~2 MB per row group for VARCHAR->LONG at the default row-group size). Low priority: one-time `ALTER TABLE ... ADD INDEX` DDL path, and it mirrors the existing `accumulateAllNullFixedChunk` convention. The fixed->var branch genuinely needs the intermediate buffer (exact size only known post-conversion), so it is correctly left as-is.

**M3 — Optional coverage extensions (not blocking).** The three tested crossing arms (symbol->fixed, var->fixed, fixed->var) are exactly the arms the fix changes, each covered with a strict row-by-row oracle. Two untested combinations remain, both low risk: (a) a DEDUP table with a covering index over a converting covered column (dedup affects the O3 merge, not the index build); (b) a var->var covered column with a pending conversion (VARCHAR->STRING transcode). A symbol->var covered column is not a gap: `chooseParquetDecodeType(SYMBOL, VARCHAR/STRING)` returns the target var type, i.e. exactly what the previous code passed, so that arm is behaviorally unchanged.

**M4 — Non-load-bearing test assertion.** The second, unfiltered `assertSqlCursors(... ORDER BY ts ...)` in `...ParquetWal` and `...NullsAndVarcharSpillWal` does not route through the covering index (the covering factory requires a key predicate), so it compares two identical non-covering plans. Harmless (validates read-path stability) but adds no regression coverage for the fix — the load-bearing assertions are the `WHERE sym='A0'` comparison plus `suspended=false`.

### Verified and dismissed (false positives)

- `numChunks = 1;` -> `int numChunks;` — definite assignment holds (assigned in both if/else branches before first read).
- `parquetFileSize = 0;` -> uninitialized — unconditionally assigned inside the `try` before both reads; no read in catch/finally.
- `estimateStringDataSize` int->long cast hoist — `maxCharsPerRow <= ~79`, so the old int intermediate never overflowed; identical results. Cosmetic/defensive only.
- `dataBuf == 0` dereference — 0 only for BOOLEAN/BYTE/SHORT/CHAR sources (inline, never spill); data write guarded by `if (actualDataSize > 0)`, free guarded by `if (dataBuf != 0)`.
- Size-based all-null detection could misclassify — verified an improvement over the pointer-based check (robust against the Rust `refresh_ptrs` dangling aux pointer); no conversion category misclassified.
- var->fixed / UUID-target uncaught exception leaks `fixBuf` — `Unsafe.free` runs in `finally` on every throw path, including the documented uncaught `IndexOutOfBoundsException` for UUID targets.

### Test gate

PASS. The PR is a fix and ships an effective regression test: `assertSqlCursors` does a strict record-by-record `assertEquals`, and reverting the production hunk goes red (symbol->long suspends the table; the var<->fixed arms produce garbage caught by the filtered comparison). No UNTESTED Critical rows. Coverage map: 8 behavioral changes, 6 tested, 2 exempt (verified no behavioral change), 0 untested.

### Notes verified during review

- `chooseParquetDecodeType` refactor (dropped `decoder`/`parquetIdx`, made package-visible): all 3 callsites pass `metadata().getColumnType(idx)`; the body is byte-identical to the pre-move version. Behavior-preserving at every callsite; the new package visibility is minimal (only `TableWriter`, same package, calls it).
- All five relocated O3PartitionJob methods differ from their originals only by the intended edits — nothing smuggled into the reordering.
- Native-buffer lifecycle is sound on all paths (happy, exception, early-return); MemoryTag consistent; buffers correctly left on the global counter (WAL-apply / index-build, not query execution).
- Read-path parity confirmed against `PageFrameMemoryRecord`: no-sentinel fixed passthrough and fixed->var column-top nulling both match the no_covering read.

bluestreak01 (approved):
# Code Review: PR #7370 — `chore(core): fix adding a covering index over parquet after a column type change`

Reviewing at **level 3**. Every finding below was verified against source.

## Summary of what the PR does

Building a POSTING covering index over a Parquet partition suspended the WAL table when an `INCLUDE`d column had a pending lazy `ALTER COLUMN TYPE` that crosses the fixed↔var / symbol→fixed boundary: `prepareCoveredColumnMmaps` asked the Rust decoder to produce the *post-ALTER* type, which the decoder cannot synthesize from the on-disk source type, so it rejected the request and WAL apply failed. The fix decodes each covered column in a Rust-producible type (`chooseParquetDecodeType`) and finishes the cast in Java with the same shared converters the O3 write path uses.

The change is essentially a **faithful port of an already-shipped, proven pattern**: `TableWriter.java:11150-11330` (the parquet→native rewrite path) already implements the identical size-based all-null detection, fixed→var `accumulateFixedToVarChunk`-equivalent, and var/symbol→fixed `convertVarColumnToFixed`-into-buffer logic. This PR brings `accumulateCoveredColumnsFromRowGroup` in line with it. That parallel is strong corroboration of correctness.

## Change surface verification

- **O3PartitionJob.java diff is 100% method reordering plus a tiny, fully-enumerated semantic delta.** A sort-and-diff of removed vs added lines confirms the only semantic changes are: `chooseParquetDecodeType` signature `(decoder, parquetIdx, columnType)` → `(srcType, columnType)` + package visibility (body byte-identical; 3 callsites all pass `metadata().getColumnType(idx)`); `int numChunks = 1` → `int numChunks`; `long parquetFileSize = 0` → `long parquetFileSize`; an `estimateStringDataSize` cast hoist; and a `strSink`→`utf16Sink` rename. Nothing smuggled into the reordering.
- **covSlotMeta stride 3→4 is fully consistent**: all 16 `add` sites in `prepareCoveredColumnMmaps` write 4 longs, and every reader uses `4L*slot+{0..3}`. No stale stride-3 access remains.

### Critical
None.

### Moderate
None.

### Minor

**M1 — Dropped method-level javadoc on `chooseParquetDecodeType` (O3PartitionJob.java:4435).** When the method was relocated and made package-visible, its header javadoc (the contract description + `@see mergeRowGroup/rewriteParquetRowGroupWithConversions` cross-refs) was dropped; only the per-case inline comments survive. Behaviorally inert — pure documentation loss. Suggest restoring a one-line header describing the decode-type-selection contract.

**M2 — Title uses `chore(core):` for a `Bug`-labeled fix.** CLAUDE.md maps a bug fix to `fix(...)`; the label is `Bug`. This is borderline-acceptable because the covering-index-over-Parquet feature is unreleased (it lives on the `nw_covering_parallel_decode` line) so no release note is warranted, and repo precedent uses `chore(core): fix ...` for exactly this case (e.g. `70ab5b115b`). Behaviorally inert. Left to author's discretion; flagging only so it's a conscious choice.

**M3 — Minor coverage gaps (low risk, not blocking).** The tests cover every arm the fix actually changes (SYMBOL→LONG, VARCHAR→LONG, INT→VARCHAR, INT→STRING, pass-through DOUBLE) across single/multi-row-group, all-null, null-mix and varchar-spill. Three combinations remain untested but each is low-risk because the underlying machinery is exercised elsewhere:
- **STRING→fixed via covering** routes through the STRING branch of `convertVarColumnToFixed` (reads `srcDataPtr`), which the covering tests never hit (they hit the VARCHAR_SLICE/aux-pointer branch). That branch is covered by the O3 `AlterTableChangeColumnType` tests, and the covering *routing* is covered by VARCHAR→LONG.
- **DECIMAL→VARCHAR/STRING via covering** — converter + estimate exist and are O3-tested; covering wiring untested.
- **var→var (VARCHAR→STRING) with pending conversion** — unchanged by this PR (`chooseParquetDecodeType` returns `columnType`, same as before; the decoder transcodes at decode time), so neither introduced nor fixed here.

**M4 — Non-load-bearing test assertion.** The second, unfiltered `assertSqlCursors(... ORDER BY ts ...)` (no `WHERE sym=`) in `...ParquetWal`, `...AllNullColumnWal`, and `...NullsAndVarcharSpillWal` compares two identical non-covering plans (the covering factory needs a key predicate), so it validates read-path stability but adds no covering regression coverage. Harmless; the load-bearing signals are the `WHERE sym='A0'` comparison and `suspended=false`. Cosmetic.

### Downgraded (false positives — verified and dismissed)

- **`default -> 40` in `estimateVarcharDataSize`/`estimateStringDataSize` under-estimates → buffer overflow** — unreachable. `getFixedToVarConverter` throws for any type outside {INT, UUID, IPv4, SHORT, BYTE, CHAR, LONG, DOUBLE, FLOAT, DATE, TIMESTAMP, BOOLEAN, DECIMAL8-256}, and the ALTER conversion matrix forbids LONG256/GEOHASH → VARCHAR/STRING. Every reachable source type has an adequate estimate; `-ea` `dataOffset <= dataSize` asserts would catch any residual gap.
- **`getDataVectorSizeAt(auxBuf, rowCount-1)` truncates the last spilled varchar** — false. `VarcharTypeDriver.getDataVectorSize` returns `dataOffset + size` for spilled entries, so the returned size includes the final value; `...NullsAndVarcharSpillWal` would fail otherwise.
- **Size-based all-null detection misroutes a non-null chunk** — false. Verified against `refresh_ptrs` in `decode.rs`: an empty vec still yields a dangling non-null ptr, so `size==0 && aux==0` is strictly more robust than the old pointer check; empty-string (aux>0) and all-inlined (data==0, aux>0) chunks are correctly NOT classified all-null.
- **`appendAddressFor(fixSize)` doesn't advance / next row group overwrites** — false. `MemoryCMARWImpl.appendAddressFor(bytes)` calls `checkAndExtend` then advances `appendAddress += bytes`; the direct-into-mmap write is contiguous and remap-safe (extend precedes the returned pointer, no intervening append).
- **`convertVarColumnToFixed` throw leaks the mmap-reserved region / malloc buffers** — false. The direct-into-mmap fixed arm allocates nothing; `accumulateFixedToVarChunk` frees `auxBuf`/`dataBuf` in nested `finally`s on every path; on any throw `indexParquetColumn`'s `finally` frees the covMmaps and cleans temp files.
- **New `coveringDecimal64/128/256` fields race** — false. The covering-index *build* runs single-threaded on the WAL-apply/writer thread; the `#7284` parallelism is query-time decode, not build.
- **`estimateStringDataSize` int-arithmetic overflow motivated the cast hoist** — false. `maxCharsPerRow` ≤ `MAX_PRECISION+3` (~79), so `4 + maxChars*2` never overflowed int; the hoist is defensive and produces identical results (EXEMPT).
- **`int numChunks;` / `long parquetFileSize;` uninitialized read** — false. Definite assignment is compiler-enforced; both are assigned before every read and not read in catch/finally.

### Coverage map

| Change (symbol + branch) | Test | Failure link | Dimensions | Verdict |
|---|---|---|---|---|
| `prepareCoveredColumnMmaps`: decode in `decodeType` (Rust-producible) not `columnType`; store `parquetColType` | `testAddPostingCoveringIndexWithPendingLazyConversionParquetWal` | Reverting → decoder rejects SYMBOL-as-LONG → WAL apply fails → `suspended=false` assertion goes red | happy ✓, error(suspend) ✓, NULL ✓(other tests), boundary ✓, concurrency N/A (single-threaded build), cleanup ✓ (assertMemoryLeak) | **TESTED** |
| `accumulateCoveredColumnsFromRowGroup`: var/symbol→fixed arm (`convertVarColumnToFixed` into mmap) | same + `...AcrossParquetRowGroupsWal` | `assertSqlCursors` record-by-record vs `no_covering` mismatches on garbage output | happy ✓, NULL ✓, multi-row-group ✓, cleanup ✓ | **TESTED** |
| `accumulateFixedToVarChunk`: fixed→var (INT→VARCHAR/STRING), spill + null branches | `...NullsAndVarcharSpillWal`, `...ParquetWal` | spilled (>9B) LONG→VARCHAR + null rows compared row-by-row | happy ✓, NULL ✓, spill boundary ✓, cross-RG aux rebase ✓, cleanup ✓ | **TESTED** |
| Size-based all-null routing (var + fixed) | `...AllNullColumnWal` | all-null converting columns compared vs `no_covering` | happy ✓, all-null boundary ✓, multi-RG ✓ | **TESTED** |
| `chooseParquetDecodeType` signature + visibility | — | verified byte-identical body; 3 callsites pass `metadata().getColumnType(idx)` | — | **EXEMPT** (no behavioral change) |
| `estimateStringDataSize` cast hoist | — | identical result, no int overflow (maxChars≤79) | — | **EXEMPT** |
| `numChunks`/`parquetFileSize` definite-assignment; `strSink`→`utf16Sink` rename | — | compiler-enforced / pure rename | — | **EXEMPT** |

### Verdict

- **Verdict: approve.** Correct, well-scoped, thoroughly tested fix that faithfully mirrors the already-shipped parquet→native conversion path. No Critical or Moderate findings; the four Minor items are documentation/metadata/cosmetic and behaviorally inert.
- **Correctness & performance gate: PASS.** No open finding touches correctness, data integrity, resource safety, concurrency, or performance/IO. The var→fixed arm writes directly into the destination mmap (optimal, zero intermediate copy); the fixed→var arm's intermediate buffer is structurally required (actual size known only post-conversion); converters resolve the per-type formatter once; `parquetColType` is cached in `covSlotMeta`. All allocations are on the one-time ADD-INDEX/reseal DDL path and correctly tagged `NATIVE_TABLE_WRITER` (writer-lived, deliberately not on a per-query `MemoryTracker`).
- **Test gate: PASS.** The PR is a fix and ships an effective regression test: reverting the production hunk makes `ALTER ... ADD INDEX` suspend the table, failing `suspended=false`; `assertSqlCursors` is a strict record-by-record oracle. No UNTESTED Critical rows.
- **Coverage map: 4 behavioral changes tested, 3 exempt (verified no-behavioral-change), 0 UNTESTED.**
- **Findings: 4 verified (all Minor), 9 draft false-positives dropped after source verification.**
- **In-diff / out-of-diff / pre-existing split: 4 in-diff (all Minor), 0 out-of-diff, 0 pre-existing.** The cross-context pass (single caller `indexParquetColumn`, reached by both ADD INDEX and O3 reseal; 3 `chooseParquetDecodeType` callsites) surfaced no broken callers.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 295 / 303 (97.36%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|231|238|97.06%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|64|65|98.46%|

PR #7371, opened 2026-07-08
fix(sql): fix NULL array handling, IPv4 aggregates, and parallel OOM handling

puzpuzpuz:
## Summary

This PR fixes correctness and resource-safety issues found while reducing the runtime of parallel SQL tests:

- fixes `first_not_null()` and `last_not_null()` for IPv4 columns on the non-sharded keyed-batch path;
- makes `dim_length()` return NULL for NULL arrays, NULL dimensions, and NULL PGWire array binds;
- prevents parallel reducers from leaking worker slots when frame navigation, temporary allocation, or cleanup fails;
- fixes a data race on the selectivity statistics that a thread-safe parallel filter shares across threads;
- reduces array-access overhead by reading dimension and element data directly through record accessors;
- removes a dead VARCHAR entry from `CaseCommon`'s dispatch table;
- keeps the slot-leak test instrumentation out of production code;
- strengthens OOM, array-forwarding, keyed-batch, and reducer-routing coverage while trimming redundant test work.

## SQL correctness

The IPv4 not-null aggregators now implement the same keyed-batch semantics as the other primitive types. Batched `last_not_null()` implementations also replace stored NULL values consistently with their row-by-row and merge paths.

`dim_length()` now handles NULL arrays and dimensions before validating bounds. Weakly typed NULL PGWire array binds defer dimensionality validation until a concrete array exists. Non-NULL out-of-range dimensions still report an error.

Record-based keyed batches evaluate aggregate arguments for every row, including rows whose value loses, so a throwing argument fails the query wherever it sits. That behaviour predates this branch and the branch preserves it; what the branch adds is a fixture that pins it. `testBatchedPathEvaluatesLosingRows` now puts the failing expression on a losing row, where the earlier fixture put it on the winning row and passed whether or not the batched path evaluated losers at all. Direct primitive-column batches retain their reduced-write fast path.

`CaseCommon`'s static dispatch table registered VARCHAR twice; the first entry never took effect because a later `extendAndSet()` in the same block overwrote it. Removing it changes no behaviour.

## Array access and ownership

`Record.getArrayDimLen()` and `getArrayDouble1d2d()` provide direct page-frame access without materializing an `ArrayView` per row. Storage-delegating records forward both methods, including union, order-by materialization, window, join, and base columns under `UNNEST`. Bounds assertions guard the direct shape read in assertion-enabled builds.

NULL-producing records consistently return a NULL `ArrayView`. The shared `ArrayConstant.NULL` uses a borrowed array, so an accidental `close()` cannot mutate global NULL state.

The array type-driver documentation now describes the on-disk shape header used by the direct readers. Constant array literals are released when a NULL index folds an access expression to NULL.

`ArrayFunctionMemoizer.getType()` delegates to the wrapped function rather than reporting `ArrayFunction`'s unset `type` field. No query is known to reach the wrapper's type: the code generator only swaps a memoizer in for a top-level projection function, whose arguments were bound before the swap, and callers that decode dimensionality reach the column through `ColumnFunction.unwrap()`. The override is hardening, and its javadoc says so rather than claiming a live caller.

## Parallel resource safety

Every reducer releases its worker slot from a `finally` entered immediately after acquisition. Cleanup that must precede slot publication runs inside a nested `try/finally`, so a cleanup failure cannot strand the slot or expose a worker-local memory pool early. This covers parallel filters, group by, top-K, horizon joins, multi-horizon joins, and window joins.

A leaked slot is permanent: `PerWorkerLocks` has no reset and the atom outlives the query that borrowed it, so once every slot has leaked, each later execution of the cached factory spins for a slot nobody will release.

`PerWorkerLocks.acquireSlot()` folds an incoming worker id into the slot range once, before the probe loop, instead of taking a modulo on every iteration; the loop then needs only a conditional subtraction. This changes no behaviour for any id a caller can currently pass — the previous in-loop modulo already bounded every non-negative id. It does remove a latent overflow: at ids near `Integer.MAX_VALUE` the in-loop `i + workerId` turns negative, and Java's `%` keeps the dividend's sign, which would index the lock array below zero. No caller produces such an id, so this is hardening rather than a fix for a reachable bug.

## Selectivity statistics

`AsyncFilterAtom` keys its selectivity statistics on the per-worker slot id, but a thread-safe filter is cloned for no worker, so it builds no per-worker filters, no locks and no per-slot statistics. `maybeAcquireFilter()` then returns `-1` to every caller, and the query owner, every reducing worker and every foreign work-stealing thread share the single owner instance.

On master those threads all called `update()` on a plain `double`/`int` pair with no synchronization. `SelectivityStats` now keeps the sample count and the average in one `volatile long` and updates it with a CAS loop, so the pair stays consistent for a reader without a second barrier. `update()` runs once per page frame rather than per row, so the CAS is not measurable against a frame decode.

The consequence of the race on master was bounded: the value only feeds the late-materialization heuristic, both branches return identical rows, and each caller reads it once into a local. A lost update or a stale read cost at most one suboptimal decode decision per frame. The visibility half was the more durable problem — with no barrier a worker could fail to observe another's updates indefinitely, pinning the heuristic at its warm-up default.

This change is not observable through the class's public surface, and that is a real gap rather than an oversight. Every observable path reduces to one boolean compared against a 20% threshold, and an EMA step is a convex combination, so a lost update cannot move the average outside the range of the samples that produced it. The accompanying tests therefore pin the packed encoding and the sample-count saturation directly, and do not claim to detect the race.

## Test instrumentation

The slot-leak assertions need a worker, rather than the query owner, to take a per-worker slot. The owner reduces with its own state and takes no slot, so an owner that steals every frame leaves the acquire/release path unexercised and the leak count reads zero for the wrong reason. `AdaptiveWorkStealingStrategy` does not give that on its own: it spins for `cairo.sql.parallel.work.stealing.spin.timeout`, 50us by default, which a worker routinely fails to wake up inside. Measured over repeated runs of `ParallelWindowJoinMemoryTrackerTest`, the owner won often enough to fail roughly one run in three.

`FactoryProvider` now supplies the work stealing strategy a page frame sequence binds, so a test can wrap the owner's stealing decision from its own sources. `SlotGatedWorkStealingStrategy` lives in the test tree and waits on a latch that `PerWorkerLocks` counts down as a worker takes a slot. `PerWorkerLocks` holds that latch in a volatile field, so each atom keeps a `final` reference to its locks and a reducer on any thread observes an installed latch; production reads the field once per frame and finds it null.

`StatefulAtom`, `WorkStealingStrategy`, `AdaptiveWorkStealingStrategy` and `AlwaysWorkStealingStrategy` are therefore unchanged against master. The seven atoms that own locks expose one accessor through the new `PerWorkerLockOwner` interface, and the separate acquire tally is gone because the latch already answers what it answered.

## Test changes

- keyed-batch fixtures compare batch output with the row-by-row reference and cover stored NULLs, repeated keys, ordering, and throwing expressions;
- array tests cover parquet, late-materialized parquet, forwarding through wrapper records, ragged `UNNEST`, and PGWire NULL binds;
- parallel memory-limit tests pin reducer selection, require a worker to acquire a slot, assert that no slot remains held, and verify recovery after a breach;
- the parquet memory-limit tests cap both reduce queues below the scan's frame count. A parquet scan cuts on row-group boundaries, so the previous fixture produced about five frames against a 32-deep ordered queue and a 4096-deep unordered one; neither filled, the owner never reached the latch-gated steal branch, and the acquire-count precondition rode on a race. `CAIRO_SQL_PAGE_FRAME_MAX_ROWS` does not help here because it does not subdivide row groups;
- OOM sweeps arm the limit around the target operation and require both a positive-slack OOM and a successful armed run;
- redundant mandatory breach iterations and oversized deterministic fixtures are removed.

## Tradeoffs

The direct array readers intentionally skip the `dataLim` and `auxLim` assertions performed while constructing a full `BorrowedArray`; this matches the existing direct element-access path. Dimension bounds remain assertion-checked at the public record accessor.

The slot-release hardening adds nested `finally` blocks on reducer cleanup paths. These paths do not allocate, and cleanup ordering remains unchanged.

`SelectivityStats` now holds its average as a float rather than a double, to fit the sample count into the same word. A 24-bit mantissa is far more precision than a ratio compared against a 20% threshold needs, but it is a precision reduction, and it is not pinned by any test that would notice a threshold decision flipping at the rounding boundary.

Production still pays something for the slot-leak coverage, and it is worth naming. `FactoryProvider` gains a method whose only override lives in the test tree. `PerWorkerLocks` carries a volatile `CountDownLatch` field that production never sets, and reads it once per successful acquire - a predictable never-taken branch, once per page frame rather than per row. `PerWorkerLockOwner` is a production interface with no production caller. The strategy now resolves through the provider on every page frame sequence construction rather than through a static factory.

In the same vein, the two WINDOW JOIN factories gain a `@TestOnly getReducerName()` accessor - test-support code on a production class, alongside the existing `getAtom()` seam - so `ParallelWindowJoinMemoryTrackerTest` pins the routed reducer by name instead of reflecting into a private field.

`AsyncFilterAtom` carries no functional change. Its diff against master is the `PerWorkerLockOwner` accessor and a member reorder.

`ParallelWindowJoinMemoryTrackerTest.testWindowJoinReleasesWorkerSlotsOnBreach` remains a single `@Test` running 37 compile-execute-breach cycles. On failure it reports one red method rather than the failing case. Parameterizing it is left out of this PR.

## Test plan

- `mvn -pl core -Dtest=AsyncFilterAtomTest,PerWorkerLocksTest,SelectivityStatsTest test`
- `mvn -pl core -Dtest=ParallelParquetMemoryTrackerTest,ParallelWindowJoinMemoryTrackerTest,ParallelHorizonJoinMemoryTrackerTest,SyncWindowJoinMemoryTrackerTest test`
- `mvn -pl core -Dtest='*GroupByFunctionKeyedBatchTest,FirstLastNotNullBatchedGroupByTest,FirstLastNotNullParallelMergeTest,IPv4GroupByFunctionKeyedBatchTest' test`
- `mvn -pl core -Dtest=ParallelGroupByFuzzTest,ParquetLateMaterializationFuzzTest,ParquetTest test`
- `mvn -pl core -Dtest=ArrayTest,ArrayAccessorForwardingTest,UnnestTest,PGArraysTest test`
- broader array, group-by batch, fuzz, memory-tracker, OOM, join, and PGWire suites listed in the branch commits


bluestreak01 (approved):
# Review: PR #7371 — `test(sql): speed up recently-added slow tests`

**Reviewed at level 3** (full mission-critical pass: 8-file change surface, 4 fresh-context reviewer agents + parent verification, per-finding source verification). This is a **test-only** PR (+101/-191, 8 test files, no production code).

## Step 2 — PR metadata

Sound. Title `test(sql): speed up recently-added slow tests` is valid Conventional Commits (`test` type is used elsewhere in the repo, e.g. `test(qwp): speed up sender coercion-error tests (#7364)`), reads well standalone. Labels **CI** + **Performance** are appropriate. The body is level-headed and analytical, lists every reduced file/case, and states tradeoffs (shrunk breach margins, coarser per-type failure reporting, dropped all-succeed sweep tail) with equal weight. No `Fixes #NNN` needed — it references source PRs #7184/#7224, doesn't close an issue.

## Step 2.5 — Change surface (test-only)

No production symbol changed. All 8 files are leaf test classes with **no subclasses** and only **private/private-static helpers** (`createTrades`/`createPrices`/`createMultiHorizonTables`, `TYPES`) — zero cross-file callsite exposure (verified with `rg`). The three mechanisms the reductions lean on were verified against source:

- **Per-query breach** (`JoinMemoryTracker`, horizon/window): `Unsafe.checkPerQueryAllocLimit` trips on `used + size > limit` where both are **requested-byte** tracker counters (`Unsafe.java:615-632`) → deterministic, environment-independent. Message `"query memory limit exceeded"/"workload=QUERY"` matches the assertions.
- **RSS-sweep OOM** (`AsOfJoinFastOomTest`, `GroupByVectorizedOomTest`): `getRssMemUsed()` reads a QuestDB-maintained counter incremented by the **requested** size in `recordMemAlloc` (`Unsafe.java:343,571-580`), **not OS RSS**. So the leak-transition slack is a fixed, config-controlled value, portable across machines.
- **Parallel merge bug** (`FirstLast`): detection parameters (`KEY_COUNT=1000`, `ROW_COUNT=100_000`, `ITERATIONS=10`, sharding/frame config) are **byte-identical** to the pre-consolidation regression test for #7224.

## Step 2.6 — Coverage map (guarded production behavior × still covered)

| Guarded behavior | Test (reduced) | Failure link | Verdict |
|---|---|---|---|
| Per-query tracker breaches on 6 secondary join cursors | `*FailsOnLargeInput` @ 40_000 keys / 512 KiB | `assertQueryBreaches` → `Assert.fail` if no OOM; 40K vs ~10-15K breach = 2.7-4x margin; deterministic tracker | **TESTED** (6/6 ran green in agents) |
| Per-query tracker breaches on horizon/window `array_agg` | `*FailsOnLargeSet` @ 40_000/400_000 / 8 MiB | inline drain→`Assert.fail`; 40K vs ~5K breach = 8x margin; 10x price density + page-frame fan-out preserved | **TESTED** (agents ran green) |
| AsOf cursor-open double-sink leak freed on OOM | `AsOfJoinFastOomTest` sweep 0-24 KiB | `assertMemoryLeak`; leak region `[8 B, ~21 KiB]` within `[0, 24 KiB]` | **TESTED** |
| Rosti cursor-open buffer leak + parquet work-stolen-survivor NPE | `GroupByVectorizedOomTest` sweeps 0-80 / 0-48 KiB | `assertMemoryLeak` + recovery drain; producing region bounded by last-OOM 68/34 KiB below caps | **TESTED** |
| `first/last_not_null` parallel-merge dest-NULL guard, 14 types × 2 aggregates | `FirstLast{First,Last}NotNullAllTypes` | `.returns(count==1000)` per type×10 iterations; any dropped value → fail tagged `type=`/`func=` | **TESTED** (14 types verified identical to deleted set) |
| `first_not_null` ordering defect | `FirstLastParallelOrderingTest` (untouched) | separate deterministic guard | **EXEMPT** (not modified) |

No UNTESTED rows. No production behavior lost coverage.

## Critical
None.

## Moderate

**M1 — OOM-sweep cap margins are thinner (verified inert, robustness watch-item).**
`AsOfJoinFastOomTest.java:102` (cap 64→24 KiB, last-OOM ~21 KiB), `GroupByVectorizedOomTest.java:78` (128→80, last-OOM ~68) and `:124` (96→48, last-OOM ~34). Three reviewers flagged these as a possible silent-coverage-loss/portability risk. **Verified they do not weaken detection:** the leak/NPE-producing region is structurally bounded *above* by the last-OOM slack point (you cannot strand a buffer or publish-then-orphan a work-stolen survivor at a slack where the allocation sequence never OOMs), and every cap exceeds that point. Because `getRssMemUsed()` is a deterministic **requested-byte** counter (not OS RSS), the measured last-OOM points are platform-portable; the AsOf case is single-threaded (zero concurrency variance, `select`→`getCursor` on the calling thread), and the two parallel Rosti caps keep 12-14 KiB of headroom to absorb peak-allocation interleaving. `slack=0` always trips, so `sawOom` cannot regress. *One-sentence inertness justification:* the detection region is provably contained below each cap and the counter is deterministic, so the reduced caps preserve the assertion — the only lost coverage is the redundant all-success tail that the recovery drain already exercises. Left as Moderate purely because the AsOf 3 KiB margin is the smallest in the batch and is worth an author sanity-check, not because any failure mode was found.

## Minor

**m1 — Non-underscored numeric literals on touched lines.** `FirstLastNotNullParallelMergeTest.java:88` `{"Date", "100000::date"}` and `:98` `{"Timestamp", "100000::timestamp"}` — QuestDB convention wants `100_000` for >=5-digit numbers. Moved verbatim from the deleted per-type methods, but the PR relocated these exact lines into the new `TYPES` array, so it can fix them in passing. Purely cosmetic, zero behavioral/performance dimension. (The UUID hex groups `00000000`/`000000000001` are part of a string literal — convention N/A.)

- *"Parquet cap 48 KiB could silently drop the survivor-NPE band if it lies in 48-96 KiB"* — **refuted**: a work-stolen survivor requires an OOM *inside* `buildRosti`; above the ~34 KiB total-allocation slack `buildRosti` completes with no OOM, so no survivor is ever published. The producing region is within `[0, 34 KiB]` which is within `[0, 48 KiB]`.
- *"Different platform malloc granularity could push the trip past the cap and hard-fail `sawOom`"* — **refuted**: the sweep is driven by `getRssMemUsed()`, a sum of QuestDB **requested** sizes, with no malloc-granularity/page/arena/OS-RSS component.
- *"Shared warm engine could mask the merge bug for types 2-14"* — **refuted**: the trigger is per-query dynamic frame dispatch under forced sharding (`~1562` frames / 4 workers), not engine warmth; type 1 runs in the fresh-engine state identical to the original first test, and the aggregate assertion fails on any type's dropped value.
- *"`sawOom` is vacuous"* — true but **pre-existing** and explicitly acknowledged in the PR body; the real detector is `assertMemoryLeak`. Not a regression.
- *Coarser per-type failure reporting / non-`AssertionError` escapes without `type=` label* — acknowledged tradeoff in the PR body; acceptable.

## Residual item (could not verify — not a diff finding)

**Enterprise CI failure** (`questdb-butler` bot): ran on commit `26491d1b22`, which is a **stale ancestor** — HEAD (`4e12b93377`) is **5 commits ahead**, including two later fixes and a fresh master merge. The 8 changed files are OSS test-only with no production change, and the enterprise suite exercises private enterprise code, so there is no plausible mechanism for these reductions to cause it — most likely a master-inherited flake on a superseded commit. I could not access the Azure DevOps build to prove root cause. **Recommend re-running enterprise CI on the current HEAD to confirm green before merge.**

## Summary

- **Verdict: approve.** Clean, well-reasoned test-runtime reduction. Every reduction was verified against the mechanism it depends on: the breach tests use a deterministic requested-byte tracker with 2.7-8x margins and fail *hard* (never silently) on over-reduction; the OOM sweeps keep the full leak-detection region below the reduced caps on a deterministic counter; the `FirstLast` consolidation preserves all 14 types × 2 aggregates × 10 iterations × 100K rows with byte-identical detection parameters.
- **Correctness & performance gate:** passes — zero confirmed correctness/resource/coverage findings. M1 was verified inert (detection region structurally preserved); m1 is cosmetic.
- **Test gate:** passes — test-only PR, no production behavior changed, no fix claimed; every guarded behavior retains its test (coverage map: 6 behavioral guards, 6 TESTED, 0 UNTESTED, 1 EXEMPT untouched sibling).
- **Findings tally:** 4 review agents ran; ~5 substantive draft notes verified, **2 dropped as false positives** (parquet-band silent loss, platform hard-fail), 1 kept as Moderate-informational, 1 kept as Minor. Two agents independently compiled and ran the reduced tests (12/12 and 16/16 green).
- **In-diff / out-of-diff / pre-existing:** all findings in-diff; the `100000` literals are pre-existing-moved-verbatim onto touched lines. Out-of-diff blast radius is zero (leaf test classes, private helpers, no subclasses).
- **Optional (non-blocking) polish for this PR:** change `100000::date`/`100000::timestamp` → `100_000::…` (m1); optionally nudge the `AsOfJoinFastOomTest` cap to ~32 KiB to restore a bit more headroom on the tightest margin (M1) — neither blocks merge. Re-run enterprise CI on HEAD to clear the stale-commit failure.

bluestreak01 (changes requested):
@puzpuzpuz

## Critical

### 1. OOM sweeps do not prove they reach the regression-specific failure point — **in-diff**

- `core/src/test/java/io/questdb/test/griffin/engine/join/AsOfJoinFastOomTest.java:95-114`
- `core/src/test/java/io/questdb/test/griffin/engine/groupby/vect/GroupByVectorizedOomTest.java:71-89`
- `core/src/test/java/io/questdb/test/griffin/engine/groupby/vect/GroupByVectorizedOomTest.java:115-136`

All three sweeps set `sawOom` for any OOM. At `slack=0`, `Unsafe.checkAllocLimit()` rejects the next positive allocation, so this flag succeeds without proving the intended transition occurred:

- ASOF requires the first sink reopen to succeed and the second to fail.
- Vector GROUP BY requires an earlier `PageFrameAddressCache` list reopen to succeed before a later reopen fails.
- The parquet case requires publication of work before the injected failure.

After the reduced 24/80/48 KiB caps, an allocation-path shift can move these transitions beyond the cap while every assertion remains green. The weakness predates the PR, but reducing the endpoints makes it a reachable coverage loss introduced by this diff.

Fix this PR by:

- Arming the limit immediately before the operation under test.
- Requiring at least one successful armed iteration in the deterministic cursor-open sweeps.
- Adding a deterministic publication/fault witness for the concurrent parquet case.
- Retaining `assertMemoryLeak()` and the recovery drain.

### 2. The OOM sweeps perform thousands of irrelevant compilations and result conversions — **pre-existing**

- `AsOfJoinFastOomTest.java:103-106,122-136`
- `GroupByVectorizedOomTest.java:78-81,124-174`

The reduced loops still:

- Compile under the RSS ceiling for every ASOF/vector cursor-open point.
- Allocate `StringSink`.
- Walk and format result columns that no assertion examines.
- Compile twice per parquet sweep point.

This both wastes CI time and allows compiler allocations to satisfy the supposedly cursor-specific fault injection.

Safe fix:

- Cursor-open tests: retain a separate factory per point, but compile before arming the ceiling and only open/close the cursor. Do not reuse one factory across points because later successful reuse can clean a pre-fix partial allocation and mask the leak.
- Parquet test: compile/reuse one factory, open before arming, set the limit immediately before the `hasNext()` drain, and drain without formatting.

### 3. The consolidated first/last test still repeats avoidable table builds and scans — **in-diff**

`core/src/test/java/io/questdb/test/griffin/engine/functions/groupby/FirstLastNotNullParallelMergeTest.java:134-160`

For each function, the test creates 14 separate 100,000-row tables and executes 140 fluent assertions. A wide table containing all 14 typed values can run all 14 aggregates in one grouped query per iteration.

That preserves the documented destination-NULL merge invariant while removing:

- 13 table lifecycle operations per function.
- Repeated key hashing and page-frame dispatch.
- Approximately 14× redundant input scans.

Keep 14 independently asserted counts, ten iterations, forced sharding, `.returns(...)`, and a parallel-factory/plan guard. The fused shape changes incidental narrow-map coverage, but not the invariant this class documents.

## Moderate

### 1. Consolidation loses type diagnostics for non-assertion failures — **in-diff**

`FirstLastNotNullParallelMergeTest.java:143-159`

The new loop wraps only `AssertionError`. `SqlException`, `CairoException`, and other exceptions now reach JUnit under `testFirstNotNullAllTypes` or `testLastNotNullAllTypes` without identifying the active type. Previously, the per-type method name supplied that information.

This is behaviorally and performance-wise inert because it changes only failure reporting after the test has already failed.

Catch `Exception | AssertionError`, add the existing type/function prefix, and preserve the cause. Do not catch `Throwable`.

## Minor

1. **Object-array rule — in-diff:**  
   `FirstLastNotNullParallelMergeTest.java:86-101,134-136` adds `String[][]`/`String[]`, contrary to the mandatory `ObjList<T>` rule. Use a flat `ObjList<String>` or `ObjList<TypeCase>`.

2. **Numeric separators — in-diff:**  
   `FirstLastNotNullParallelMergeTest.java:88,98` should use `100_000::date` and `100_000::timestamp`.

3. **Cast syntax — pre-existing on replaced lines:**  
   `JoinMemoryTrackerTest.java:82-83,139-140,241-242,272-273,484-485,739-740` should use `x::SYMBOL` rather than `cast(x AS SYMBOL)`.

4. **Missing label — in-diff metadata:**  
   The `test(sql)` title requires the `SQL` label in addition to `CI` and `Performance`.

5. **PR-body conventions — in-diff metadata:**  
   “Coverage is unchanged” omits the reduced per-type leak-isolation granularity. Qualify it as unchanged SQL type/function coverage and disclose the shared leak scope. The body also contains passive constructions contrary to the active-voice rule.

## Downgraded

- **“The first/last tests currently execute serially.”** Dismissed: current planner gates route all tested signatures through `AsyncGroupByRecordCursorFactory` with four workers. The absence of an explicit plan guard is a resilience gap, not a demonstrated current bypass.
- **“Compile one factory and reuse it for every cursor-open sweep point.”** Dismissed: later reuse can clean a partial allocation left by the pre-fix implementation and mask the leak.
- **“OOM thresholds vary with OS page or allocator granularity.”** Dismissed: `Unsafe` tracks requested native bytes, not operating-system RSS pages.
- **“The shared WorkerPool races table drops or outlives the engine.”** Dismissed: each assertion closes its factory before the next DDL, and `TestUtils.execute()` halts the pool before closing compiler/context/engine.
- **Merge-commit headline/body violations.** Dismissed: branch history is squash-merged and the project explicitly prohibits spending effort tidying it.
- **`sawOom` boolean naming.** Dismissed under touched-line policy: the declarations are unchanged from the base.
- **Member ordering violations.** Dismissed under touched-line policy: the ordering predates the PR; the newly added `TYPES` field is correctly placed relative to adjacent fields.

## Coverage map

This is a test-only PR, so the table audits preservation of guarded production behavior.

| Change/guard | Test and failure link | Dimensions | Verdict |
|---|---|---|---|
| Six join per-query tracker breaches | `JoinMemoryTrackerTest` six `*FailsOnLargeInput` methods; `assertQueryBreaches()` fails if draining completes and checks OOM/message | Happy: covered by sibling under-limit tests; error: covered; NULL: N/A; boundary: covered; concurrency: N/A; cleanup: covered | **TESTED** |
| Twelve horizon/window tracker breaches | Four sync/parallel classes; factory guards plus explicit failure if the query completes | Happy/error/boundary: covered; NULL: N/A; concurrency: sync+parallel covered; cleanup: covered | **TESTED** |
| ASOF partial sink-open cleanup | Both `AsOfJoinFastOomTest` methods | Happy recovery: covered; general OOM: covered; target boundary: uncovered; concurrency: N/A; cleanup: target not proven | **UNTESTED — Critical** |
| Vector GROUP BY partial cache reopen | `testVectorizedGroupByCleansUpWhenCursorRunsOutOfMemory` | Happy recovery: covered; general OOM: covered; target boundary: uncovered; cleanup: target not proven | **UNTESTED — Critical** |
| Parquet published-work cleanup | `testWorkStolenEntryDoesNotOutliveFreedPoolsOverParquet` | Happy recovery: covered; general OOM: covered; publication/failure ordering: uncovered; resource cleanup: target not proven | **UNTESTED — Critical** |
| `first_not_null` parallel merge, 14 types | `testFirstNotNullAllTypes`; count drops below 1000 if a value is lost | Happy/NULL/boundary/concurrency/resource: covered; error: N/A | **TESTED** |
| `last_not_null` parallel merge, 14 types | `testLastNotNullAllTypes`; same failure link | Happy/NULL/boundary/concurrency/resource: covered; error: N/A | **TESTED** |

## Summary

**Verdict: request changes.**

- Coverage map: **7 guard groups, 4 TESTED, 3 UNTESTED**.
- Focused validation passed: **82 tests, 0 failures/errors/skips**.
- GitHub reports **35/35 checks passing**, but green execution does not resolve the vacuous OOM-transition assertions.
- Verified findings: **9** — 7 in-diff, 0 out-of-diff, 2 pre-existing.
- **7 draft findings were removed as false positives.**
- Out-of-diff blast radius is zero because these are leaf test classes with private helpers and no external repository selectors or subclasses.

puzpuzpuz (commented):
Thanks — the three Critical items hold up, and C1 is the one that mattered. All addressed in 350725efc9.

## C1 — sweeps did not prove they reached the failure point: **valid, fixed**

Confirmed, and the root cause was worse than "the endpoints got trimmed". The sweeps armed the ceiling around the whole drain helper, **compilation included**, so a low ceiling tripped on a compiler allocation and the cursor open under test never ran. Combined with `sawOom` — which `slack=0` satisfies unconditionally — the assertion held whether or not the sweep ever crossed the transition.

Each sweep now arms the ceiling on the operation it targets and nothing else:

- cursor-open sweeps compile above the ceiling and open the cursor without draining it;
- the parquet sweep compiles **and opens** above the ceiling, then arms for the `hasNext()` drain, where `buildRosti` publishes work before the fault lands.

Each sweep additionally asserts one **armed iteration succeeded**. Paired with the guaranteed OOM at `slack=0`, that brackets the operation's entire allocation span, so the sweep provably crosses the OOM→success transition, and an over-trimmed range now fails loudly instead of passing vacuously.

I did not take the "compile/reuse one factory" suggestion for the parquet case. It would mask the bug for a stronger reason than the one you already applied to the cursor-open sweeps: the survivor must dereference *freed* pools, and reusing the factory hands it live ones. Fresh factory per point, in all three.

Measured, with the ceiling armed on the operation alone, the spans are sharp and deterministic (last OOM → first success): ASOF open **4104 → 4112 B**, vectorized GROUP BY open **2040 → 2048 B**, parquet drain **32320 → 32384 B**. So the ranges drop to 8 KiB / 8 KiB (parquet stays 48 KiB) — not by guessing a margin, but because the armed-success assertion now fails if a range is short.

**Detection re-proven by mutating each guarded fix:**

| Mutation | Result |
|---|---|
| drop `Misc.free(cursor)` from `AsOfJoinFastRecordCursorFactory.getCursor()` | both ASOF tests fail: `tracker recycled with used=8` |
| drop the atomic reopen in `PageFrameAddressCache.of()` | cursor-open sweep fails: 192 KiB `NATIVE_DEFAULT` leak |
| drop the drain-past-throw in `runWhatsLeft()` | parquet sweep reproduces the original NPE: `getFrameFormat(int)` on a null `addressCache` |

## C2 — irrelevant compilations and result conversions: **valid, fixed**

Fixed as a consequence of C1 — compilation moved above the ceiling — plus the sweeps no longer format rows no assertion reads, and the parquet recovery drain now runs only after an iteration that actually OOM'd (only an aborted drain can strand a survivor).

## C3 — redundant table builds and scans: **premise valid, fixed; one correction**

The premise checks out and the PR body was stale on this: at 2K rows / 1 iteration the class drops from 4222 ms to ~1700 ms, so ~2.5 s really was table-build and scan work. (That claim was measured in the 28-method shape, where the fixture dominated; it stopped being true once the fixture was paid twice.)

One correction: "approximately 14× redundant input scans" isn't right — the 14 value columns hold *distinct* data, so total bytes scanned is the same either way. What's actually redundant is the **key column**, re-scanned 13 extra times, and — the real cost — 130 redundant query compilations, frame dispatches and shard merges. Fusing removes those.

Done as suggested: one table with a value column per type, one query aggregating all 14, 14 independently asserted counts, 10 iterations, forced sharding, `.returns(...)`. I also added the **plan guard** you flagged as a resilience gap in your Downgraded section — fusing changes the query shape, so pinning it to `Async Group By` is now load-bearing: without the parallel factory there'd be no shard merge to test.

Verified against the pre-#7224 production revert: `testLastNotNullAllTypes` fails on **every** run (3/3), with all 14 counts dropping 1000 → 390 and each column naming its type. 4222 ms → 1540 ms.

## M1 — type diagnostics for non-assertion failures

Moot in the fused shape: there's no per-type loop to wrap. The type is carried by the column alias, so it survives into both the assertion diff and any SQL error message.

## Minors

- **m1 `ObjList`** — done, `ObjList<TypeCase>`.
- **m2 numeric separators** — done, `100_000::date` / `100_000::timestamp`.
- **m3 cast syntax** — done, all 38 `cast(x AS SYMBOL)` → `x::SYMBOL` in `JoinMemoryTrackerTest` (converted the whole file rather than leaving it half-and-half).
- **m4 `SQL` label** — added.
- **m5 PR body** — rewritten. "Coverage is unchanged" is now scoped to the merge invariant and SQL type/function coverage, and the shared leak scope plus the dropped narrow-map shape are called out as tradeoffs. Passive constructions cleaned up.

## Net

In-test time: `FirstLastNotNullParallelMergeTest` 4222→1540 ms, `AsOfJoinFastOomTest` 1754→917 ms, `GroupByVectorizedOomTest` 1888→1454 ms. `JoinMemoryTrackerTest` unchanged (cast syntax only).

Re: the stale enterprise CI failure — agreed, it ran on a superseded commit; this push re-runs CI on current HEAD.

puzpuzpuz:
Thanks for the review. It landed against `e9a168bf81`; the branch head is `350725efc9` ("Target OOM sweeps at the code under test"), which went in before the review was posted and already implements most of what you asked for. The review quotes `String[][] TYPES`, `100000::date`, 14 separate tables and the per-type `catch (AssertionError)` — none of which exist at head — so the findings below are answered by the current code rather than by new commits.

**Critical 1 (OOM sweeps don't prove they reach the failure point)** — done at head, and along the lines you prescribed. Each sweep now arms the ceiling on the operation it targets and nothing else: the cursor-open sweeps compile above the ceiling and open the cursor without draining; the parquet sweep compiles and opens above the ceiling, then arms it for the `hasNext()` drain, where `buildRosti` publishes work before the fault lands. Every sweep additionally asserts that at least one *armed* iteration succeeded, which — paired with the guaranteed OOM at `slack=0` — brackets the operation's full allocation span, so an over-trimmed range fails loudly instead of passing vacuously. `assertMemoryLeak()` and the recovery drain are retained. Detection is confirmed by mutating each of the three guarded production fixes (details in the PR body).

**Critical 2 (irrelevant compilations and result formatting)** — done at head. Compilation moved above the ceiling, the sweeps no longer format rows no assertion reads, and, per your own caveat, both cursor-open sweeps keep a fresh factory per point rather than reusing one.

**Critical 3 (14 tables per function)** — done at head. One table now carries a value column per type and one query aggregates all 14. The 14 independently asserted counts, the ten iterations, the forced sharding, `.returns(...)` and the parallel-factory plan guard are all kept.

**Moderate 1 (lost type diagnostics)** — superseded rather than fixed, and I don't think a `catch` is warranted now. There is no per-type loop left to wrap: a single query covers all 14 types, each count is aliased with its type name, so an assertion failure still names the type in the diff, and a `SqlException`/`CairoException` now comes from one query whose text names the offending column. Adding `catch (Exception | AssertionError)` would be dead code. Happy to reconsider if you see a path where the type stays ambiguous.

**Minor 1 (`ObjList`)** — done at head: `ObjList<TypeCase>`.
**Minor 2 (numeric separators)** — done at head: `100_000::date`, `100_000::timestamp`.
**Minor 3 (cast syntax)** — done at head: `JoinMemoryTrackerTest` has no `cast(... AS ...)` left; the rewritten lines use `x::SYMBOL`.
**Minor 4 (missing label)** — done: the PR carries `SQL`, `CI` and `Performance`.
**Minor 5 (PR body)** — the coverage half was already addressed (the body qualifies the claim as unchanged *merge-invariant* coverage and discloses the shared leak scope under Tradeoffs). The passive-voice half was a fair hit, and I have just rewritten those sentences in active voice.

Re-review against `350725efc9` when you get a chance.

puzpuzpuz:
Level 3 review verdict: **approve with non-blocking comments**.

Moderate:

1. `PerWorkerLocks` now updates a test-only acquisition tally on every successful production slot acquisition. This adds `AtomicIntegerArray` get/lazySet work to the per-frame parallel reducer hot path. Please consider making acquisition observation test-configurable or using an injected test hook so normal instances retain the CAS-and-return path.

2. `TestUtils.assertNoSlotLeakOnBreach` relies on bounded retries, polling, and `Os.sleep(1)` until a worker happens to acquire a slot. A correct implementation can fail on a constrained CI host if the owner thread wins every frame. A deterministic callback/latch at reducer entry or slot acquisition would make the test reliable.

3. The regression tests expose factory atoms and lock counters through production `@TestOnly` APIs. This couples them to factory nesting, atom ownership, and the lock implementation. A deterministic failure hook plus observable post-failure recovery assertions would reduce that coupling.

Minor:

- The multi-horizon join factories do not override `getAtom()`, although their atoms implement the new telemetry, leaving that telemetry unreachable through the generic helper.
- The new test-helper Javadocs are substantially longer than needed.
- The PR title should describe the user-visible effect of the slot leak, such as preventing parallel-query hangs after allocation failures.
- Several merge commits lack the required long-form body; two also exceed the 50-character title limit.

Test gate: passed. Coverage map reviewed 14 behavioral rows: 13 tested, 0 untested, 1 verified no-behavior-change exemption. The regression tests have credible failure links. Focused local verification passed `ArrayAccessorForwardingTest`, `IPv4GroupByFunctionKeyedBatchTest`, and `PerWorkerLocksTest`: 29 tests, 0 failures.

No Critical findings were verified, so these comments are non-blocking.

GitHub does not allow the PR author to submit an approving review, so this verdict is recorded as a comment.

puzpuzpuz:
Thanks — went through all seven. Five are addressed in 1e6c16f (section 24 of the PR body); two I'm pushing back on, with reasons.

## Moderate 1 — acquisition tally on the hot path: fixed

Valid, and it did not need a test-only knob. The tally now lives in the lock int itself rather than in a second int in the slot's padding. A slot's int is a counter, not a flag: acquire increments it to an odd value, release increments it again to an even one, so parity answers "is it held" and the value answers "how many times was it taken". The counter only counts up and never returns to a value an acquirer has already read, so it stays ABA-free with no version field.

Production is back to exactly one CAS to acquire and one store to release, the tally free. The read before the CAS turns the acquire into a test-and-test-and-set, which reduces line traffic under contention rather than adding to it. `releaseSlot()` pays one relaxed load of a line the caller already owns exclusively, and asserts the slot was held. `ParallelGroupByFuzzTest` (201 cases), `ParallelFilterTest`, the vectorized group-by suites and the four memory-tracker suites all pass with `-ea`; the new assert never fires.

## Moderate 2 — deterministic reducer/slot hook: declining, with evidence

The concern is real in principle, but the proposed remedy contradicts Moderate 1: a callback or latch at reducer entry or at slot acquisition is production code on the same per-frame path you just asked me to keep free of test instrumentation, and it would be the *only* such hook there.

It is also not a race that can be closed from the outside. Only a *non-owner* reduce takes a slot — `maybeAcquire(workerId, owner, ...)` returns -1 for the owner, which reduces with its own state — and owner-vs-worker is settled by whoever wins the shared reduce queue. There is no observation point that does not add a hook.

What is there instead: the test forces many small frames (`cairo.sql.page.frame.max.rows=1000`, and the same for the small-frame size, since the window join resizes the master scan), so each execution dispatches ~50 frames against a 4-worker pool. Measured, with the leak reintroduced, all four slots are taken and leaked on **iteration 0** — the owner winning all ~50 frames for 200 consecutive executions is what the bound has to survive, not one frame. And the loop was already validated against exactly the host you describe: pinned to two cores (`taskset -c 0,1`), it was red 3 of 3 with a bare `acquires > 0` assertion and green 6 of 6 with the retry loop (test plan, PR body). The failure mode is also a loud, specific assertion, never a vacuous pass.

## Moderate 3 — coupling to @TestOnly APIs: partly fixed

The observation surface is three `@TestOnly` methods with defaults (`getAtom()`, `getAcquiredSlotCount()`, `getSlotAcquireCount()`), and after Moderate 1 they are pure reads of state the lock already keeps — no extra production code path, no extra field. Factory nesting is handled by one helper walking `getBaseFactory()`, not by the tests. I'd rather keep that than trade it for a behavioural-only oracle, which cannot tell "no leak" from "no worker ran" — the exact hole this suite exists to close.

But the recovery assertion you asked for is worth having on top, so it is added: `testParallelQueryOverParquetRecoversAfterBreach` breaches the per-query limit eight times over one cached factory, then raises the limit and asserts the same factory still returns its 50,000 rows. No atom, no lock counter, just the symptom a user sees. It is bounded by a 30-second query timeout so a leaking build fails on an assertion rather than hanging: with `navigateTo()` moved back outside the `try`, the recovered query dies with `timeout, query aborted` — the workers spin in `acquireSlot` and the breaker stops them. Verified red on the reverted code, green on this branch.

## Minor — multi-horizon `getAtom()`: fixed

Correct. `AsyncMultiHorizonJoinRecordCursorFactory` and `AsyncMultiHorizonJoinNotKeyedRecordCursorFactory` now override it. Their reducers were already correct (acquire inside the `try`) and stay untouched, but the telemetry was unreachable through `findAtom()`. Both multi-horizon breach tests now carry the same no-leak guard the single-slave joins carry.

## Minor — Javadoc length: fixed

Trimmed on the two `TestUtils` helpers, on `StatefulAtom`'s two telemetry methods and on `PerWorkerLocks`.

## Minor — PR title: fixed

Retitled to `... and stop parallel queries hanging after a memory-limit failure`, which is the user-visible effect.

## Minor — merge-commit messages: declining

Branch history here is throwaway: PRs are squash-merged, so only the squashed message that lands on `master` survives, and these are `git`'s own auto-generated merge subjects. Rewriting them would cost a force-push and buy nothing that outlives the merge.

puzpuzpuz:
## Code Review — level 3 (full mission-critical pass)

114 files, all Java (~72 production, ~42 test). Reviewed with 13 parallel agents (correctness, concurrency, performance, resource, cross-context, two independent fresh-context adversarial passes, tests) plus independent lead verification. **No Critical findings.** Consensus is that the production changes are correct: IPv4 `first/last_not_null` keyed batch, `dim_length` NULL handling, parallel-reducer slot leaks, and the `SelectivityStats` data race are all fixed correctly, and the direct shape-header array reads are a genuine optimization.

### Critical

None.

### Moderate

**M1 — `UnorderedPageFrameSequence.getWorkStealingStrategy()` javadoc is false and hides a latent test-hang trap** · `core/src/main/java/io/questdb/cairo/sql/async/UnorderedPageFrameSequence.java:343-350`
The added javadoc claims the getter "returns the unwrapped base, never the reduce-phase instance … a later phase binding it would park on a gate nothing can open." The body is `return workStealingStrategy;`, and `SlotGatedWorkStealingStrategy` is a *subclass* of `AdaptiveWorkStealingStrategy`, not a wrapper — nothing is unwrapped; the gated instance is returned as-is. Production is unaffected (the default `FactoryProvider` installs no gate, so `testAcquireLatch` is always null). But post-aggregation callers (`AsyncGroupByRecordCursor:247`, `AsyncHorizonJoinRecordCursor:202`, `AsyncMultiHorizonJoinRecordCursor:218`) call `getWorkStealingStrategy().of(postAggregationStartedCounter)` and then `shouldSteal` on that same gated instance. It doesn't deadlock today only because the breach tests fail during reduce before post-aggregation, and for a passing query the count-1 latch is already spent by then. Residual footgun: a future latch-gated test whose query succeeds through reduce → shard → post-aggregation while the *owner* drains every frame via the tail-loop (which bypasses `shouldSteal`) would leave the latch un-decremented, and post-aggregation would block 30 s then throw `AssertionError`.
*Fix:* make the code match the javadoc — have `SlotGatedWorkStealingStrategy.shouldSteal` ignore the latch once rebound off the reduce-started counter, or hand back an unwrapped base for post-aggregation. At minimum, correct the javadoc to state it returns the same (possibly gated) instance.

**M2 — Cleanup-nesting-only slot hunks have no regression test** · `AsyncTopKRecordCursorFactory.findTopK`, `AsyncMultiHorizonJoinRecordCursorFactory.reduce/filterAndReduce`, `AsyncMultiHorizonJoinNotKeyedRecordCursorFactory.reduce/filterAndReduce`
In these reducers `navigateTo` was *already* inside the slot-releasing try on master; the only change is wrapping `releaseParquetBuffers()`/`recordB.clear()` in a nested `finally`. Reverting just these hunks leaves every test green — the nesting only matters if the cleanup call itself throws (e.g. `ParquetBuffers.close()` under `-ea`), which no test induces and which is unproven-reachable. The genuine slot-leak bug (`navigateTo` moved from outside to inside the try) is fully covered for every reducer that had it, including both plain `AsyncHorizonJoin(+NotKeyed)` factories. Defensive hardening, not a proven-reachable live bug.
*Fix:* add a fault-injection test that makes a cleanup call throw and asserts the slot is still released, or add a one-line comment on these `finally` blocks stating they defend an unproven cleanup-throw path so the intent survives future edits.

**M3 — `ParallelWindowJoinMemoryTrackerTest.reducerName()` uses reflection on a private field** · `core/src/test/java/io/questdb/test/griffin/ParallelWindowJoinMemoryTrackerTest.java:672-696`
Does `getDeclaredField("frameSequence")` + `setAccessible(true)`, then identity-matches a `PageFrameReducer` static constant to recover the reducer's field name. Renaming the private field or inlining the reducer constants silently turns the assertion into an unrelated `AssertionError`. Conspicuous because the PR elsewhere establishes clean `@TestOnly` seams for exactly this (`getAtom()`, `getPerWorkerLocks()`).
*Fix:* add a `@TestOnly` accessor returning the selected reducer (or its name) on the window-join factories, mirroring `getAtom()`, and drop the reflection.

**M4 — `ArrayAccessorForwardingTest` reaches package-private constructors/`of()` via reflection** · `core/src/test/java/io/questdb/test/cairo/sql/ArrayAccessorForwardingTest.java:113-124, 171-187`
Three of ten tests use `Class.forName`/`getDeclaredConstructor`/`getDeclaredMethod` + `setAccessible(true)` for `MaterializedRecord`, `UnnestRecord`, `WindowLightRecord`. The whitebox *approach* is justified (no SQL result can distinguish forwarding from the `getArray()` fallback; the throwing `DirectOnlyArrayRecord` fake is clean), but the reflection is the smell — the other seven records are constructed directly.
*Fix:* give those three types `@TestOnly` (or widened) construction/`of()` seams and call them directly.

### Minor

- **Per-row loop-invariant re-decode** · `PageFrameMemoryRecord.getArrayDimLen0`/`getArrayDouble1d2d0` re-run `ColumnType.decodeArrayDimensionality`/`decodeArrayElementType` on the constant `columnType` every row. Real but tiny (memory-bound path); could be hoisted into the `dim_length` `ConstFunc`/`Func` where `arrayColumnType` is already held.
- **Test-code nits** · `TestUtils.java:69-70` import ordering; `SlotGatedWorkStealingStrategy.java:77` static method placed among instance members; `TestUtils.assertNoSlotLeakOnBreach` javadoc run-on; duplicated ~5-line rationale comment + `createTrades`/`createPrices` fixtures across the four tracker tests (a shared base would hold them once).
- **Unused import** · a few converted per-type `*GroupByFunctionBatchTest` keep `import org.junit.After;` after moving teardown to `AbstractGroupByFunctionBatchTest`.
- **`PerWorkerLocksTest` comment over-claim** · says a double-count would show as a latch going "past" 0, but `CountDownLatch` saturates at 0; cosmetic.
- **Coverage micro-gaps** · `SelectivityStats` float-vs-double threshold-flip has no boundary test (heuristic-only); no dedicated SQL-level all-NULL IPv4 group assertion (covered by the byte-exact unit test); array/IPv4 changes not exercised on WAL tables or with JIT explicitly toggled (orthogonal paths).
- **PR metadata** · the body's "pins the IntelliJ version (`ci/templates/java-lint.yml`)" bullet and the `CI` label do not correspond to anything in the 114-file diff GitHub computes (that CI work appears to have merged to master separately). Either drop the bullet or confirm it landed elsewhere. Otherwise the title and level-headed, tradeoff-listing body are exemplary.

### Downgraded (false positives from the draft)

- **`ArrayConstant.NULL` / Horizon+MultiHorizon records return the shared NULL without `ofNull()`** — a fresh `BorrowedArray` has `type = UNDEFINED` (so `isNull()` would be false). **Dismissed:** `ArrayConstant` has a `static { NULL.ofNull(); }` initializer that sets `type = ColumnType.NULL` at class load, so `ArrayConstant.NULL.isNull()` is always true; `close()` on `BorrowedArray` is a no-op and nothing mutates the shared NULL. Safe hardening.
- **Missing `getArrayDimLen` forwards on non-page-frame records** (RecordChain, VirtualRecord, WAL, map cursors) — **Dismissed:** the `Record` default delegates to `this.getArray()`, so those records stay correct (they only forgo a fast path that wouldn't apply to a non-page-frame base anyway). No record forwards only one of the two accessors.

### Coverage map

| # | Behavioral change | Regression test (verified) | Failure link | Verdict |
|---|---|---|---|---|
| 1 | IPv4 `first_not_null`/`last_not_null` `computeKeyedBatch` | `IPv4GroupByFunctionKeyedBatchTest` (byte-exact oracle) + `FirstLastNotNullBatchedGroupByTest.testNotNullKeyedBatchKeepsNonNull` | revert → inherits `First/LastIPv4` rowId-only compare → trailing-NULL clobbers value → byte-diff / count RED | **TESTED** |
| 2 | `Last*NotNull.computeBatch` stored-NULL clause (11 types) | `Last*NotNull*BatchTest.testLastNotNull*BatchReplacesStoredNull` + `KeepsHigherRowIdNonNull` | store NULL@100, apply non-null@10; revert → `10>100` false → NULL kept → RED | **TESTED** |
| 3 | `Last*` `computeKeyedBatch` backward walk | equivalence test (perf-only) | order-independent by guard | **TESTED (perf)** |
| 4 | `dim_length` NULL array / NULL dim / OOB error / weak-dims PGWire bind / fast path | `ArrayTest.testLength{Null,Invalid,ConstantArrayInvalidDim,ColumnTop,OverArrayExpression,OverParquet,…}`, `PGArraysTest.testArrayBindDimLength{,WithNullArray}`, `UnnestTest` | revert → error/NPE/wrong value at asserted position | **TESTED (exhaustive)** |
| 5 | Direct array reads (`getArrayDimLen0` 1D fast path + general shape; `getArrayDouble1d2d0` refactor) | `ArrayTest.testLengthOverParquetOneDimension/ColumnTop/…`, `ArrayAccessorForwardingTest` | empty→0, NULL→INT_NULL, column-top→INT_NULL | **TESTED** |
| 6 | 12 forwarding records (`getArrayDimLen`/`getArrayDouble1d2d`) | `ArrayAccessorForwardingTest` (`DirectOnlyArrayRecord.getArray()` throws; mapped-index checked) | wrapper failing to forward → AssertionError | **TESTED** |
| 7 | Slot-leak fix — throwing statements moved inside slot-releasing try | `ParallelParquetMemoryTrackerTest`, `ParallelWindowJoinMemoryTrackerTest` (`assertNoSlotLeakOnBreach`: `getAcquiredSlotCount()==0` + latch + OOM msg) | revert → OOM-in-`navigateTo` leaks slot → RED; every reducer type named | **TESTED (all reducers that had the bug)** |
| 8 | Slot cleanup-nesting only (findTopK, Multi/MultiNotKeyed horizon) | — | revert leaves tests green (breach was already inside try) | **UNTESTED → M2 (Moderate, hardening)** |
| 9 | `SelectivityStats` packing/saturation/warm-up/clear | `SelectivityStatsTest` (7 tests) + `AsyncFilterAtomTest.testThreadSafeFilterSharesOneSelectivityEma` | revert encoding/shared-instance → RED | **TESTED** |
| 9b | `SelectivityStats` data race | — (structural: single `volatile long` CAS) | test infeasible; PR states this | **UNTESTED by design (non-blocking)** |
| 10 | `PerWorkerLocks.acquireSlot` overflow fold | `PerWorkerLocksTest.testAcquireSlotFoldsExtremeIdsWithoutOverflow` (both overloads) | revert → `INT_MAX` fold → negative index → AIOOBE RED | **TESTED** |
| 11 | `ArrayFunctionMemoizer.getType` delegates | `ArrayTest` `dim_length` over a >1×-referenced projection | revert → `UNDEFINED`/0-dims → `-ea` assert fires | **TESTED (indirect)** |
| 12 | `CaseCommon` VARCHAR entry removal | — | dead code: line 354 `VARCHAR→VarcharCaseFunction` overwrites it | **EXEMPT (verified inert)** |
| 13 | `ArrayConstant.NULL` `DirectArray→BorrowedArray` | — | no behavioral change (static `ofNull()`; `close()` no-op) | **EXEMPT** |
| 14 | `FactoryProvider.getWorkStealingStrategy` (prod) | — | default preserves `WorkStealingStrategyFactory.getInstance` | **EXEMPT (prod)** |

### Summary

**Verdict: approve.** No Critical or blocking findings survived verification. Well-constructed, well-tested consolidation of correct fixes plus deliberately production-neutral test scaffolding.

- **Test gate: SATISFIED.** Every fix ships with a regression test whose assertion flips red on revert (verified by assertion-tracing; empirical revert-in-worktree was impractical because the branch is far ahead of a stale local `master`). No UNTESTED Critical row. The one real UNTESTED item (M2) is defensive hardening of an unproven-reachable cleanup-throw; the `SelectivityStats` race is structurally untestable and the PR says so.
- **Coverage map: 14 behavioral-change rows — 11 tested, 3 exempt (verified inert), 1 untested-Moderate (M2), plus the by-design untested race.**
- **In-diff / out-of-diff / pre-existing: all findings in-diff; 0 out-of-diff; 0 pre-existing.** The zero out-of-diff is genuine, not an underrun — the cross-context pass covered every `Record` impl, every `acquireSlot` caller, all `FactoryProvider` impls, and all `SelectivityStats` consumers; the changes are contract-preserving by design.
- **Regressions/tradeoffs (as the PR body honestly states):** `SelectivityStats` EMA demoted `double→float` (immaterial vs a 20% threshold, unpinned at the rounding boundary); production now carries a never-null `volatile CountDownLatch` read and a never-taken branch once per frame, a `PerWorkerLockOwner` interface with no production caller, and strategy resolution through the `FactoryProvider` per sequence construction — all acknowledged, per-frame (not per-row) costs.

Recommend addressing **M1** (cheap; prevents a future flaky test and corrects a false comment) and the two test-reflection items (**M3, M4**) in this PR since CI round-trips are expensive — none blocks merge.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 489 / 506 (96.64%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncJitFilteredRecordCursorFactory.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/ArrayFunctionMemoizer.java|2|3|66.67%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFillRecordCursorFactory.java|3|4|75.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ExtraNullColumnRecord.java|4|5|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/HorizonJoinRecord.java|4|5|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/MultiHorizonJoinRecord.java|4|5|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinRecordCursorFactory.java|104|111|93.69%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/ArrayDimLengthFunctionFactory.java|37|39|94.87%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryRecord.java|23|24|95.83%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinFastRecordCursorFactory.java|31|32|96.88%|
|:large_blue_circle:|io/questdb/cairo/sql/async/PageFrameSequence.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastDoubleGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/UnnestRecord.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKRecordCursorFactory.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/DelegatingRecord.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastCharGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/PerWorkerLocks.java|27|27|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/FirstNotNullIPv4GroupByFunctionFactory.java|28|28|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastFloatGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastLongGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullDoubleGroupByFunction.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/union/UnionRecord.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinAtom.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/Record.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastDateGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/MaterializedRecord.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullLongGroupByFunction.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/constants/ArrayConstant.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastIPv4GroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKAtom.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullDateGroupByFunction.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/window/WindowLightRecord.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/async/UnorderedPageFrameSequence.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastBooleanGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastByteGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncMultiHorizonJoinRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/CoveringIndexRecordCursorFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncFilteredRecordCursorFactory.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastShortGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/RecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncGroupByNotKeyedRecordCursorFactory.java|14|14|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/JoinRecord.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/FactoryProvider.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/array/DoubleArrayAccessFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastIntGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncHorizonJoinRecordCursorFactory.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncFilterAtom.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/BaseAsyncMultiHorizonJoinAtom.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullTimestampGroupByFunction.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncGroupByAtom.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecord.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastTimestampGroupByFunction.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncMultiHorizonJoinNotKeyedRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullIntGroupByFunction.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/ArrayUnnestSource.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectivityStats.java|17|17|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncGroupByRecordCursorFactory.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullIPv4GroupByFunctionFactory.java|29|29|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/BaseAsyncHorizonJoinAtom.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullGeoHashGroupByFunctionFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncHorizonJoinNotKeyedRecordCursorFactory.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameFilteredMemoryRecord.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullCharGroupByFunction.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/groupby/LastNotNullFloatGroupByFunction.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncGroupByNotKeyedAtom.java|2|2|100.00%|

bluestreak01 (approved):

PR #7373, opened 2026-07-08
feat(core): apply ALTER COLUMN TYPE to parquet partition files immediately

RaphDal:
## Summary

`ALTER TABLE ... ALTER COLUMN ... TYPE` now converts parquet partitions in place at ALTER time, writing the new column type into the parquet files, instead of deferring the conversion to read time. Once the statement returns, a parquet partition's on-disk bytes carry the current column type. This brings parquet in line with native partitions, which already convert eagerly and block until done.

Because a parquet file now always stores the current column type, the read-time lazy-conversion path that the previous behaviour relied on is unreachable, and this PR also removes it (see Read-path simplification below).

## Background

`ALTER COLUMN TYPE` on native partitions rewrites the column files eagerly. On parquet partitions the conversion was lazy: the ALTER recorded the new type in table metadata but left the parquet file storing the old type, and each read (or the next out-of-order write) applied a per-row cast on the fly. That had two consequences:

- The parquet file on disk did not match the column's current type, so any consumer that reads the file by its own schema saw the old type.
- Reads of a partition that was never rewritten paid a decode-and-cast cost on every scan.

Behaviour was therefore format-dependent — native eager, parquet lazy. This change makes the two consistent.

## What the change does

- At ALTER time, for each parquet partition that stores the column, the writer decodes the partition and re-encodes it into a new partition directory with the new type in the footer, reusing the same row-group rewrite path the out-of-order write already uses for pending conversions.
- The re-encode commits together with the metadata change, so the parquet file and the table schema advance in one step. On failure the ALTER rolls back and the original partition is left untouched; the superseded directory is removed after commit by the standard scoreboard-gated partition purge.
- Partitions that do not store the column (it was added after the partition was converted to parquet) are skipped — their rows are already all NULL.
- The parquet index rebuild now matches columns by `originalWriterIndex`, the column id stored in parquet metadata, rather than by the current writer index. This fixes a pre-existing helper bug that the eager re-encode path now depends on for indexed SYMBOL columns whose writer index changed after an earlier `ALTER COLUMN TYPE`.
- There is no configuration flag; the eager behaviour is unconditional.

## Read-path simplification

Because a parquet partition's stored column type now always matches table metadata within any read snapshot, the lazy read-time conversion path that #7024 added is unreachable, and this PR removes it:

- `PageFrameMemoryPool.resolveParquetColumn` no longer inspects source-vs-target types or sets up a conversion. A type-converted column is still stored under its original writer index (the `replacingIndex` chain head), so the writer-index remap stays; the method now checks that the stored type tag equals the current tag and throws if they ever differ, instead of silently reinterpreting the bytes.
- `PageFrameMemoryRecord` loses its per-row cast machinery (`hasTypeCasts` / `sourceColumnTypes` and every `convertVarTo*` / `convertFixedTo*` accessor), and `PageFrameFilteredMemoryRecord` loses the matching `needsLazyConversion` fast-path guard.
- The `PageFrameMemory.getSourceColumnType` / `hasColumnTypeCasts` signals are removed, along with the callers that consulted them to opt out of a fast path: the JIT filter gate, the top-K sort-key batch encoder, and the group-by raw-address paths. With lazy casts gone these opt-outs never fired, so removing them changes no result. One consequence is that compiled (JIT) filters now run on parquet frames that a lazy-cast column would previously have excluded from JIT.

This is an internal simplification with no user-visible behaviour change for correct inputs; it removes roughly 900 lines. The write-side conversion machinery is untouched — the eager re-encode itself still uses it (the same row-group rewrite and the Rust decode-with-conversion), so nothing on the conversion path is lost. It is bundled with the eager change rather than split out because it is only correct as a consequence of it: under the lazy model the read path was load-bearing.

## Tradeoffs

- `ALTER COLUMN TYPE` over parquet partitions now performs the conversion up front and blocks until it finishes, so the statement takes longer and does more I/O than before, in proportion to the size of the parquet data storing the column. The lazy model deferred that work, and for partitions that were never read or rewritten it avoided the work entirely.
- The re-encode rewrites whole partitions into new directories rather than editing in place, so it also rewrites the columns that did not change in each affected partition.
- Reads of a converted parquet partition no longer perform per-row casts, so repeated scans after an ALTER are cheaper.

The net effect depends on the workload: eager is more total work when a partition is altered but rarely read, and less when a converted partition is scanned repeatedly.

## Behaviour notes

- Chained conversions (several `ALTER COLUMN TYPE` on the same column) now stay in parquet format across steps, because each ALTER keeps the file in sync with the metadata. Previously a chained conversion with a type mismatch fell back to native.
- A target type of SYMBOL still converts the partition to native first, unchanged, as symbol maps cannot be built from parquet.
- The read-time lazy-conversion path is removed (see Read-path simplification): a converted parquet partition is read directly at its stored type, which the eager re-encode keeps equal to the current type.

## Test plan

- Added tests asserting the parquet footer physically carries the new type immediately after ALTER, before any subsequent write, for fixed->fixed, fixed->var, var->fixed, var->var and symbol->non-symbol conversions.
- Added a multi-partition test that alters a table with several parquet partitions and checks every one is re-encoded, alongside a native active partition.
- Added a test for a column absent from an older parquet partition: that partition is skipped and reads NULL, while a newer partition storing the column is re-encoded.
- Added a test that a target type of SYMBOL converts the partition to native.
- Added a test that converting an *indexed* SYMBOL column to a fixed type re-encodes the parquet partition and drops the column's now-stale symbol index, rather than rebuilding it against the re-encoded file (which already stores the fixed type). The re-encode rebuilds `.k/.pk` indexes for symbol columns that remain symbol, so the index rebuild consults the column's post-ALTER type to skip the converted column.
- Added a regression test that re-keys a column with `STRING -> SYMBOL INDEX`, converts the partition to parquet, then eagerly rewrites a different column. The test verifies the re-encode rebuilds the surviving SYMBOL index by parquet column id and that indexed and non-indexed queries return the same rows.
- Added a test that a chained conversion stays in parquet format across steps.
- Added an error-path test that injects a file-open failure during the re-encode and verifies the ALTER rolls back with no native-memory or file-descriptor leak, leaves the committed parquet at the old type, and recovers on `RESUME WAL`.
- Updated the existing lazy-behaviour tests to expect eager re-encode, and confirmed a reader opened before the ALTER still sees the old type and values for its lifetime (MVCC isolation across the in-place re-encode).
- Existing parquet type-conversion parity, out-of-order rewrite, native `ALTER COLUMN TYPE`, and convert-partition suites pass unchanged.
- All tests run under `assertMemoryLeak`, covering the re-encode's native scratch allocation and its cleanup on both the success and error paths.

For the read-path simplification, the suites that exercise every removed surface pass unchanged (the value-parity tests read the re-encoded parquet directly and match a native control):

- `ParquetColumnTypeConversionTest` (reads converted parquet through the collapsed `resolveParquetColumn` and the group-by / sample-by / join fast paths over altered columns), `ParquetTest`, and `ParquetRowGroupPruningTest`.
- `ParquetLateMaterializationFuzzTest`, covering the filtered-record path (`PageFrameFilteredMemoryRecord`) whose `needsLazyConversion` guard was removed.
- `OrderByEncodeSortTest` / `OrderByEncodeSortFuzzTest` for the top-K sort-key batch encoder, and `CompiledFilterTest` / `CompiledFilterRegressionTest` / `AsyncFilteredRecordCursorFactoryTest` for the JIT filter gate.
- Across these runs the new `resolveParquetColumn` stored-tag-vs-current-tag assertion never fired, consistent with the invariant that the eager re-encode keeps the two equal.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/CLAUDE.md
+++ b/core/src/main/java/io/questdb/cairo/CLAUDE.md
@@ -1,29 +1,31 @@
 # Writing Parquet Partitions with Pending Column Conversions
 
-How O3 (out-of-order) writes land on a **parquet** partition that carries a *lazy*
-`ALTER COLUMN TYPE`, and the column conversions and native-memory allocations involved —
-with and without deduplication.
+How the parquet row-group rewrite machinery applies `ALTER COLUMN TYPE` column conversions,
+and the native-memory allocations involved — with and without deduplication.
+
+`ALTER COLUMN TYPE` re-encodes parquet **eagerly** at ALTER time (see `griffin/CLAUDE.md`);
+the eager re-encode (`O3PartitionJob.rewriteParquetPartition`) drives this machinery to decode
+the old parquet and write the new type. The same helpers back O3 (out-of-order) merges, but
+because ALTER is eager, an O3 merge onto a converted partition sees already-target-typed data
+and performs no conversion.
 
 This is the **write-path** counterpart to `griffin/CLAUDE.md`, which owns the conversion
-*semantics* (per-type cast rules, null sentinels, the `replacingIndex` chain) and the
-*read* path (`PageFrameMemoryPool` / `PageFrameMemoryRecord`). Read that first for the
-"what each cast does" rules; this file covers "how the writer materialises them during O3".
+*semantics* (per-type cast rules, null sentinels, the `replacingIndex` chain) and the direct
+read path. Read that first for the "what each cast does" rules; this file covers "how the
+writer materialises them".
 
 ## When this happens
 
-A parquet partition stores each column in the type it had when the partition was converted
-to parquet. A later `ALTER COLUMN TYPE` is **lazy** — `ConvertOperatorImpl` does *not*
-re-encode parquet (it only re-encodes native partitions, and runs a pre-pass for the two
-cases lazy decode cannot handle: target SYMBOL, or a chained conversion with a type
-mismatch — see `griffin/CLAUDE.md`). So the parquet file keeps the *source* type, and the
-conversion is deferred until something reads or rewrites the partition.
+`ALTER COLUMN TYPE` re-encodes each parquet partition eagerly (`ConvertOperatorImpl` calls
+`TableWriter.rewriteParquetPartitionWithConversions`, except for the pre-pass cases that go
+parquet→native first: target SYMBOL, or a chained conversion whose parquet no longer matches
+metadata — see `griffin/CLAUDE.md`). The re-encode decodes the old parquet and writes the new
+type using the machinery below, so the on-disk bytes always match the metadata.
 
-When O3 rows fall inside such a partition's time range, `O3PartitionJob` must merge them
-into the parquet data, which forces the lazy conversion to materialise **at write time**.
-
-There is **no conversion when O3 lands on a NATIVE partition**: `ALTER COLUMN TYPE`
-materialises eagerly into native column files, so a native O3 merge always sees
-already-target-typed columns and allocates zero conversion buffers.
+Because both native and parquet `ALTER COLUMN TYPE` are now eager, an O3 merge always sees
+already-target-typed columns and allocates zero conversion buffers — whether the partition is
+native or parquet. The merge/dedup mechanics below still describe how O3 rows land on a
+parquet partition; only the conversion step is a no-op there, having run eagerly at ALTER.
 
 ## The write paths (merge actions)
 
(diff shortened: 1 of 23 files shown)


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 218 / 236 (92.37%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/sql/PageFrameMemoryPool.java|4|9|44.44%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinRecordCursorFactory.java|6|8|75.00%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|31|34|91.18%|
|:large_blue_circle:|io/questdb/griffin/ConvertOperatorImpl.java|15|16|93.75%|
|:large_blue_circle:|io/questdb/cairo/O3PartitionJob.java|152|159|95.60%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinFastRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncJitFilteredRecordCursorFactory.java|1|1|100.00%|

PR #7374, opened 2026-07-09
feat(qwp): stop resending the full symbol dictionary on every message

glasstiger:
## Summary

Tandem OSS half of questdb/java-questdb-client#66. Bumps the `java-questdb-client` submodule to `3383f29b`, delivering the QWP delta symbol-dictionary change into the bundled client, and hardens the ingress decoder against the frame shape that change makes reachable.

Previously every QWP ingress message carried the entire symbol dictionary; the client now registers each symbol id with the server only once per connection and sends only new ids thereafter, re-registering the full dictionary when a connection is replaced.

The full design, durability contract, and an even-handed list of tradeoffs (per-reconnect full-dictionary replay, a per-slot dictionary side-file, and a bounded "resend required" failure on host crash rather than fsync) are in the client PR.

## Server-side change: reject a gapped delta symbol dictionary, atomically

The server has always been able to parse delta frames, but until this change no client ever sent one with a non-zero start id -- the old client hardcoded a full dictionary from id 0 on every message. Delta framing makes that path live, so its handling of a gap now matters.

`QwpMessageCursor.parseDeltaSymbolDict` grew the connection dictionary with nulls up to `deltaStartId + deltaCount`. That inflated `size()`, which is exactly the bound `QwpSymbolColumnCursor.of()` checks an incoming symbol index against, so a row referencing one of the padded ids passed the bounds check, read back null, and landed a NULL symbol value. The frame was accepted and nothing was logged -- the padding silently disarmed a guard the decoder already had.

The decoder now rejects `deltaStartId > size()` with its own error code, `DELTA_DICT_GAP`, which reaches the sender as a new wire status byte, `STATUS_DICTIONARY_GAP` (0x0D). The gap gets a dedicated code rather than reusing `INVALID_DICTIONARY_INDEX`/`PARSE_ERROR` because the verdict depends on this connection's dictionary coverage -- server state, not the frame's bytes -- so the identical frame succeeds once the sender has re-registered from an id the server actually holds. The server therefore sends the NACK and keeps the connection open; the bundled client treats the status as retriable, recycles the wire and replays its catch-up. A contiguous append (`deltaStartId == size()`) and a lower start that re-registers or remaps existing ids both remain allowed.

The parse is also atomic on failure. The entry loop snapshots each pre-existing slot into a per-cursor scratch right before overwriting it; a rejected delta restores exactly those slots, nulls the slots it grew into, and resets the size, so a failed frame leaves the connection dictionary byte-for-byte what it was -- and a parsed dictionary can no longer hold a null. The scratch clears after every commit and rollback and in `clear()`, so it pins nothing across messages or connection reuse.

Wire-compat note: the server now rejects a frame shape it previously (wrongly) accepted, and 0x0D is a status byte no earlier server emitted, under an unchanged protocol version. QWP is experimental and unreleased and the bundled client moves in lockstep, which is what the `non-breaking-tandem` label asserts; the client maps unknown status bytes to a retriable category, so an older bundled client degrades to retry against a newer server. Only a client bug, a torn store-and-forward dictionary, or a third-party implementation can produce a gap -- the bundled client's send loop refuses to transmit one.

## Test plan

- `QwpSymbolDecoderTest` -- a gapped delta is rejected with `DELTA_DICT_GAP` and routed to the `DICTIONARY_GAP` status; the `deltaStartId == size()` boundary that a too-strict check would break is still accepted; a rejected delta restores every entry it overwrote -- including at a non-zero `deltaStartId`, and including when the rejected frame was the connection's first -- and a parsed dictionary holds no null entries. Verified both ways: with the new guard removed, the gap test fails.
- `QwpIngressServerRestartFuzzTest` -- the producer now cycles a bounded set of SYMBOL values, so rows past the first `TAG_CARDINALITY` back-reference ids an earlier frame registered. That is what puts the reconnect catch-up on the critical path while the server bounces; previously every value was distinct and each frame carried its own registration.
- OSS QWP server and end-to-end ingest tests build and run against the bumped client submodule in CI.
- Client-side coverage lives in the tandem PR: `DeltaDictCatchUpTest`, `DeltaDictRecoveryTest`, `PersistedSymbolDictTest`, `CursorWebSocketSendLoopCatchUpAlignmentTest`, `MmapFaultDegradesTest`, `ReconnectTest`, `SelfSufficientFramesTest`.

## CI follow-up

- `QwpSenderOversizeRowInBatchTest` runs against the new client behaviour for a batch the server's advertised cap can never accept: `flush()` throws and retains the batch, and `close()` completes teardown -- committing and draining everything already accepted -- instead of aborting on a retry of the permanently invalid batch. An earlier workaround that reset buffered rows inside the test is gone; the test now asserts the real contract.
- Verified both tests in the class pass with JDK 25 and the `local-client` profile.

## Tandem

Required by https://github.com/questdb/java-questdb-client/pull/66


ideoma:
#### [PR Coverage check]

:heart_eyes: **pass** : 53 / 55 (96.36%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/qwp/protocol/QwpMessageCursor.java|43|45|95.56%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpIngressProcessorState.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/protocol/QwpParseException.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpIngressUpgradeProcessor.java|1|1|100.00%|

ideoma (approved):

PR #7375, opened 2026-07-09
test(core): stop fuzz setup reader-opens flaking on transient metadata-read timeouts

RaphDal:
## Summary

Several table-open paths in the shared fuzz harness open a `TableReader` (or table `TableMetadata`) while a table's structural (`ADD COLUMN`) WAL changes are still being applied, and do not tolerate the transient `read timeout` that the reload path can throw in that window. The harness already treats that timeout as an expected transient on its *concurrent-read* paths (via `getReaderHandleTableDropped`, which retries it). This PR routes the remaining setup/verify opens through the same retry, factors the retry into one shared helper, and broadens it to the whole read-timeout family:

- `FuzzRunner.generateSymbols` and `FuzzRunner.assertStringColDensity` — reader opens, and
- `FuzzRunner.generateTransactions` — the `engine.getTableMetadata(...)` open, which sits on the exact stack from the observed failure (`generateTransactions → generateSet → generateSymbols`) and, being evaluated first in the try-with-resources header, is actually the *first* open in that method that can trip the transient.

Test harness only — no product code changes.

## How it surfaced

A one-off failure of the enterprise fuzz test `BackupTest.testBackupFuzzyIncrementalPIT` (which drives this OSS harness) on a heavily-loaded macOS CI runner: 1 failure in 1580 tests, and 0 on the Linux and Windows runs of the same suite.

```
io.questdb.cairo.CairoException: [0] Metadata read timeout [src=reader, timeout=1000ms]
    at io.questdb.cairo.TableReader.reloadMetadata(TableReader.java)
    at io.questdb.cairo.TableReader.reloadSlow(TableReader.java)
    at io.questdb.cairo.TableReader.<init>(TableReader.java)
    ...
    at io.questdb.test.cairo.fuzz.FuzzRunner.getReader(FuzzRunner.java)
    at io.questdb.test.cairo.fuzz.FuzzRunner.generateSymbols(FuzzRunner.java)
    at io.questdb.test.cairo.fuzz.FuzzRunner.generateSet(FuzzRunner.java)
    at io.questdb.test.cairo.fuzz.FuzzRunner.generateTransactions(FuzzRunner.java)
```

The failure is in fuzz *setup*, not in the code under test. `createInitialTableWal` creates a table and fires a burst of `ALTER TABLE ADD COLUMN`s that apply asynchronously via the WAL-apply job; `generateTransactions → generateSet → generateSymbols` then opens table resources to sample existing symbols while those structural changes are still landing.

## Root cause

`TableReader.reloadSlow` reads `_txn` (which carries the committed metadata version) and then loads the `_meta` matching that version, spinning until the two agree or `spinLockTimeout` (default 1000 ms) elapses. During an `ADD COLUMN`, `TableWriter.rewriteAndSwapMetadata` publishes the new `_meta` (version V) first, then `bumpMetadataVersion()` commits `_txn` to V last — so there is a window in which `_meta` = V but `_txn` = V−1 and any reader that samples it sees a mismatch.

The read itself is cheap: map `_meta`, compare one 8-byte version field, `return false` on mismatch, with no sleep on that path. So a full-second timeout does **not** mean the reads were slow — it means the writer held `_txn` and `_meta` inconsistent for ~1 s. With the default `NOSYNC` commit mode there is no `fsync` in that window; the WAL-apply worker simply made no progress for ~1 s (descheduled / a VFS syscall stalled) on a saturated CI box, while the zero-backoff reader spin pinned a core. This is an environmental transient inherent to opening a table resource concurrently with a structural WAL apply.

The reader open is not the only door onto that reconciliation. `engine.getTableMetadata(...)` hands out a pooled `TableReaderMetadataTenantImpl`, and refreshing an already-pooled tenant runs the identical `readTxnSlow` + `reloadMetadata` spin (`TableReaderMetadataTenantImpl` throws `Metadata read timeout [src=metadata, ...]`, byte-for-byte the reader logic with a different `src=` label). The WAL-apply worker itself warms that pool during the initial apply (`WalApplySqlExecutionContext.getMetadataForWrite → getTableMetadata`), so the metadata open in `generateTransactions` takes the same warm-tenant reload path with the same probability as the reader. Leaving it bare only relocates the flake one frame up the stack. The sibling `engine.getLegacyMetadata(...)` open, by contrast, reads the sequencer from memory under a read lock — no `_txn`/`_meta` reconciliation, no timeout — so it is deliberately left bare.

## Fix

- Extract one shared helper, `openWithRetries(Supplier<T>, boolean tolerateTableRecreate)`, that retries a table-resource open past the read timeouts thrown while a structural change is mid-apply (up to 10×, 100 ms backoff).
- Broaden the retried transient from just `Metadata read timeout` to the whole read-timeout family the reload path can throw — `Metadata` / `Transaction` / `Column Version read timeout` — via a `read timeout` substring match. The three are the same "sampled a file mid-writer-commit" transient; the observed failure was one of them, and the reader path can throw the transaction variant too (`TableReader` line ~1592). `block file read timeout` is a different subsystem and is never thrown on these open paths.
- `getReaderWithRetries` now delegates to the shared helper with `tolerateTableRecreate = true` — unchanged behavior for the five concurrent-read call sites, where the table is legitimately dropped and recreated (dropped / name reserved / entry locked are waited out).
- The three setup/verify opens — `generateSymbols`, `assertStringColDensity`, and the `getTableMetadata` open in `generateTransactions` — use `tolerateTableRecreate = false`: they retry only the read-timeout transient. At these sites the table is never concurrently dropped (setup runs before the apply threads start; the density check runs after the WAL is fully drained), so a dropped / reserved / locked table there is a genuine failure and now surfaces immediately instead of spinning forever. Relative to `master` (which used a bare `getReader` at these sites) this only *adds* the read-timeout retry — the fail-fast-on-real- error behavior is preserved, not weakened.

## Why not fix the writer instead

The stall is the OS not scheduling the WAL-apply worker (or a slow VFS syscall) under CI load — not something QuestDB can speed up, and not an `fsync` (commit mode is `NOSYNC`). The inconsistency window is a deliberate two-phase commit (`_meta` then `_txn`) that exists for crash-safety; the real elimination — coexisting versioned `_meta` files so a reader never observes a mismatch — is already flagged as future work in `TableUtils.handleMetadataLoadException`. Tolerating the transient at the call site is the proportionate fix and is consistent with how the harness already handles it elsewhere.

## Deliberately left alone

- `AbstractFuzzTest.generateSymbols` is a near-copy of `FuzzRunner.generateSymbols` used by `DedupInsertFuzzTest`, still opening via a bare `getReader`. It is **not** folded into the fixed helper for two reasons: its three call sites do not create a concurrent-ADD-COLUMN apply window (the tables are freshly created with no async structural churn in flight, and Dedup applies WAL synchronously), so it is not exposed to this transient; and the two copies intentionally differ — `AbstractFuzzTest` generates **unicode** symbols (`rnd.nextUtf8Str`), `FuzzRunner` generates ASCII (`rnd.nextChars`) — so delegating one to the other would silently drop unicode-symbol coverage in the Dedup fuzz suite. Deduplicating it would be a coverage regression, not a cleanup.

## Testing & risk

- Test-only change: `FuzzRunner` (the shared retry helper plus its call sites) and a new `FuzzRunnerTest`. `openWithRetries` is now package-private (it holds no instance state) so the test drives it directly, without reflection.
- `FuzzRunnerTest` covers the helper's branches deterministically: a read timeout retries then rethrows once the 10-retry budget is spent (11 attempts total), a read timeout that clears mid-retry succeeds, `tolerateTableRecreate = false` fails fast on a dropped / reserved / locked table (a single attempt, no spin), `tolerateTableRecreate = true` waits out a recreate (including `EntryLockedException`), and an unrelated `CairoException` always surfaces immediately. It also injects one transient timeout through each of the five real caller routes—transaction metadata, symbol generation, density verification, WAL drain, and purge—and verifies the selected retry/recreate policy. Runs green via surefire: `Tests run: 10, Failures: 0, Errors: 0`. (The full `mvn -pl core test-compile` needs a fresh `questdb-client` build — a stale worktree client otherwise fails an unrelated QWP test; a clean checkout with the client built compiles and runs it.)
- The test guards the retry logic and caller wiring, not the original flake. The CI failure is load-dependent (1-in-1580, only on the slowest runner) and does not reproduce deterministically, so no test recreates it; the unit test instead catches a future regression of the helper itself — narrowing the substring back to `Metadata read timeout`, flipping the gate, or breaking the retry count.
- Tradeoff: broadening the match to the whole read-timeout family means the five concurrent-read sites now also retry the transaction / column-version variants (previously rethrown). These are the same transient class, so this is strictly more tolerant with no behavioral downside; it does slightly widen what the helper swallows before surfacing an error.


RaphDal:
# Review: PR #7375

**`test(core): stop fuzz setup reader-opens flaking on transient metadata-read timeouts`**

- PR: https://github.com/questdb/questdb/pull/7375
- Review level: 3 (full pass — correctness, cross-context, resource, concurrency, test-coverage; verified against source with 2 independent adversarial agents + direct trace)
- Scope: ~40 lines, test-harness only (`core/src/test/java/io/questdb/test/cairo/fuzz/FuzzRunner.java`)
- **Verdict: Approve** — test-only, provably correct at every callsite, strictly safer than `master`.

> Level 3 is normally reserved for replication/JNI/on-disk-format/API/ACL PRs. This is a single test file, so the Rust-safety, enterprise-ACL, and SQL-convention passes were no-ops. Every technical claim in the PR description verified true against source.

---

## Critical

None.

## Moderate

None.

## Minor / Optional

### O1 (resolved) — deterministic test added for the `openWithRetries` retry/gate branches (`FuzzRunner.java:1244-1268`)

**Resolved on this branch.** `FuzzRunnerTest` now drives `openWithRetries` directly — 5 cases: read-timeout retry-then-exhaust (asserts exactly 11 attempts), retry-then-recover, gate-off fail-fast (1 attempt, no spin), gate-on wait-out (incl. `EntryLockedException`), and unrelated-error immediate rethrow. `openWithRetries` was made package-private (it holds no instance state) so the test reaches it without reflection. Verified green via surefire: `Tests run: 5, Failures: 0, Errors: 0` (1.44s). The rationale below is retained as the record of why the test exists and how it is shaped.

`openWithRetries` is the only genuinely new logic in the PR. Its happy path (open succeeds first try) runs on every fuzz invocation, but its branches only fire non-deterministically under CI load:

- the `"read timeout"` substring match retries, and after `--readTimeoutRetries < 0` rethrows (the 11th attempt);
- `tolerateTableRecreate=true` waits out `table does not exist` / `reserved` / `EntryLockedException`, while `=false` rethrows them immediately.

A silent regression here — e.g. someone narrows the substring back to `"Metadata read timeout"`, flips the gate, or breaks the retry count — would **not** fail the suite. It would quietly bring back the rare (1-in-1580) flake this PR exists to kill. That is the one part of the harness that is *not* self-validating: most harness code fails loudly and deterministically when broken; rare-error-path control flow does not.

**This is optional, not blocking.** Weighing it honestly:

- The blast radius of a regression is "an annoying flake returns," not data corruption or a silently-green suite. `openWithRetries` is not an assertion/oracle helper, so a broken version flakes or hangs (visible) rather than making callers vacuously pass (invisible).
- The helper is 24 obvious lines and unlikely to be edited.
- A unit test here risks the change-detector trap: pinning `== 11` and the exact message strings couples to the implementation, so a future legitimate retry-budget tune breaks a test that never caught a real bug.
- The rest of `FuzzRunner` (`generateSet`, `generateSymbols`, `createInitialTableWal`, …) is deliberately not unit-tested; singling out this helper is inconsistent.

If you do want the regression guard, it needs no flakiness or concurrency (would require making `openWithRetries` package-private / `@TestOnly`):

```java
@Test
public void testOpenWithRetriesRethrowsAfterExhaustingReadTimeoutRetries() {
    AtomicInteger calls = new AtomicInteger();
    try {
        fuzzer.openWithRetries(() -> { calls.incrementAndGet();
            throw CairoException.critical(0).put("Transaction read timeout [src=reader]"); }, false);
        Assert.fail();
    } catch (CairoException e) {
        Assert.assertTrue(Chars.contains(e.getFlyweightMessage(), "read timeout"));
        Assert.assertEquals(11, calls.get()); // 1 + 10 retries
    }
    // + throws N<11 times then returns (retries then succeeds)
    // + tolerateTableRecreate=false rethrows "table does not exist" on the first attempt (no spin)
}
```

### m1 — Comment overstates `getLegacyMetadata` as unconditionally timeout-free (`FuzzRunner.java:540-542`)

The comment says `getLegacyMetadata` "reads the sequencer from memory and cannot [time out]." That holds **only for WAL tables**: for a non-WAL token it delegates to `getTableMetadata` (`CairoEngine.java:1016-1017`), which *can* throw a read timeout. Every current caller of `generateTransactions` passes a WAL table (verified — the harness's non-WAL table reuses the WAL table's transactions via `createInitialTableNonWal`, it doesn't call `generateTransactions`), so this is correct today. But since that un-wrapped open is evaluated *first* in the try-with-resources, a future non-WAL reuse would silently reintroduce the flake one frame up. A one-clause tweak makes it airtight: *"for a WAL table, getLegacyMetadata reads the sequencer from memory…"*.

### m2 — Backoff is small relative to each attempt's internal spin (cosmetic, `FuzzRunner.java:1256`)

Each failing open internally spins up to `spinLockTimeout` (~1000 ms, zero-backoff `Os.pause`) before throwing; the helper then sleeps only 100 ms. Exhausting 11 attempts burns ~11 core-seconds of spinning on the already-saturated CI box the fix targets. Inherited from the old reader path (now also applied to the metadata open), test-only — informational, not worth changing.

---

## Downgraded (false positives ruled out during verification)

- **NPE on a null flyweight message** in `Chars.contains(message, "read timeout")` → **Dismissed.** `CairoException.message` is `protected final StringSink message = new StringSink()` (`CairoException.java:81`) — never null. `EntryLockedException` sets it to `"table busy [reason=...]"`. Also matched-first in the old code, so no new exposure.
- **Broadened `"read timeout"` swallows `block file read timeout`** → **Dismissed.** `BlockFileReader` (`BlockFileReader.java:93`) is constructed only by checkpoint / mat-view / settings / view-definition code; grep shows zero references from `TableReader`, `TableReaderMetadataTenantImpl`, or `TableReaderMetadata`. Unreachable from all 8 open sites, so the substring can only match the three real spin-reconciliation transients.
- **Resource leak across the 10 retries** → **Dismissed.** `openWithRetries` holds no reference between attempts. All opens are exception-safe: pool `refresh`-throw closes the tenant (`AbstractMultiTenantPool.get0:319-322`), `newTenant`-throw frees the slot (`:305`), both reader/metadata constructors `close()` on partial init (`TableReader.java:173-175, 229-231`; `TableReaderMetadataTenantImpl.initialize:180-182`), and the `generateTransactions` try-with-resources closes `sequencerMetadata` if the second initializer throws (JLS 14.20.3).
- **Fix is incomplete — the un-wrapped first open (`getLegacyMetadata`) could still flake** → **Dismissed.** For WAL tables it reaches `sequencerMetadataPool.get` → in-memory sequencer copy (`TableSequencerImpl.getTableMetadata`), no spin-lock deadline, no file read on the warm path. All callers pass WAL tables. (Captured as the m1 latent note.)
- **Branch reorder / broadening breaks the 5 concurrent-read sites** → **Dismissed.** No exception message satisfies both branches (the recreate strings and `EntryLockedException`'s message contain no `"read timeout"`; `EntryLockedException` is matched by `instanceof`), so ordering is irrelevant. The broadening only *adds* retrying of the Transaction/Column-Version variants that the old code rethrew — same transient class, strictly more tolerant, still bounded at 10.

---

## Coverage map

PR touches **test-harness code only** — no product behavioral change, so there are no product-behavior rows. Rows below are the changed test-harness symbols.

| Change | Test | Failure link | Dimensions | Verdict |
|---|---|---|---|---|
| `openWithRetries` — read-timeout branch: retry ≤10×, then rethrow | `FuzzRunnerTest.testReadTimeoutRethrownAfterRetriesExhausted` / `testReadTimeoutRetriedThenSucceeds` | asserts exactly 11 attempts on exhaustion (fails if the count or substring match regresses) and recovery after N<11 | happy: covered; retry/exhaust: **covered**; NULL: N/A (message never null); boundary (exhaustion): **covered** | Tested |
| `openWithRetries` — `tolerateTableRecreate` gating (true spins on recreate, false rethrows) | `FuzzRunnerTest.testTableRecreateNotToleratedFailsFast` / `testTableRecreateToleratedIsWaitedOut` / `testUnrelatedCairoExceptionAlwaysRethrown` | gate=false rethrows a dropped table in 1 attempt; gate=true waits out `reserved` + `EntryLockedException`; unrelated error rethrown even with gate=true | gate=true/false branches: **covered** | Tested |
| `getReaderWithRetries` (renamed, delegates, broadened family) | 5 concurrent-read sites, every fuzz run | fuzz invariants (`assertSqlCursors`) | happy: covered; retry: non-deterministic | Partially tested (happy) |
| `generateTransactions` — `getTableMetadata` now retried | every fuzz run | fuzz invariants | happy: covered; retry = the flake fix: non-deterministic | Partially tested (happy) |
| `assertStringColDensity` — `getReader` now retried | every fuzz run | density assertion | happy: covered; retry: runs post-drain, essentially never fires | Partially tested (happy) |
| `generateSymbols` — `getReader` now retried | every fuzz run | symbols feed fuzz txns | happy: covered; retry: non-deterministic | Partially tested (happy) |

---

## Summary

**Approve.** Test-only, provably correct at every callsite, and strictly safer than `master`: it adds a read-timeout retry to the 3 setup/verify sites while preserving fail-fast on real errors, and broadens the 5 concurrent-read sites to the same transient family — which incidentally closes a latent gap where the old `"Metadata read timeout"`-only match would have rethrown a `Transaction read timeout` from a reader open (`TableReader.java:1592`). The PR description is level-headed and every technical claim in it verified true.

- **Test gate:** passes cleanly. No UNTESTED Critical rows (no product behavior changed); the PR is correctly typed `test(core)`, not a customer `fix`. The one non-self-validating part — `openWithRetries`'s retry/exhaust/gate branches — now has a deterministic `FuzzRunnerTest` (5 cases, ran green), so even a returning-flake regression would be caught at author time.
- **Coverage-map totals:** 6 behavioral changes (all test-harness), all covered — 4 callsite happy-paths via the fuzz suite, 2 helper-logic rows now directly tested by `FuzzRunnerTest`. 0 untested.
- **Findings:** 3 verified (O1, m1, m2). O1 resolved (deterministic test added, `Tests run: 5, Failures: 0`); m1 fixed (comment scoped to WAL tables); m2 left as-is (cosmetic). 5 dropped as false positives after source verification.
- **In-diff / out-of-diff / pre-existing split:** 3 in-diff, 0 out-of-diff, 0 pre-existing. The zero out-of-diff count is justified: the cross-context pass traced all 8 callsites plus callers two levels up plus the pool/reader/metadata internals; the changed symbols are `private` to `FuzzRunner` with no external callers (the old `getReaderHandleTableDropped` name is fully gone).
- **Regressions/tradeoffs:** the only tradeoff is the one the PR names — the 5 concurrent-read sites now also retry the Transaction/Column-Version timeout variants before surfacing (bounded at 10×), slightly widening what the helper swallows. Same transient class, strictly more tolerant, no correctness downside.

**Branch note** (not blocking, per squash-merge conventions): the intermediate commit `2ef637c679` wired the setup sites through `getReaderWithRetries` (which would spin forever on a genuinely missing table); the final commit `ee9c245788` correctly switched them to `tolerateTableRecreate=false`. The squash collapses this — no action needed.

bluestreak01 (changes requested):
@RaphDal

Reviewing PR #7375 at level 3.

### Critical

1. **[in-diff] The retry policy multiplies the underlying spin timeout by 11**  
   `core/src/test/java/io/questdb/test/cairo/fuzz/FuzzRunner.java:1244-1256`  
   `core/src/test/java/io/questdb/test/cairo/mv/MatViewFuzzTest.java:55-64`

   Every reader/metadata attempt already spins for `CAIRO_SPIN_LOCK_TIMEOUT`. `openWithRetries()` restarts that full deadline ten times:

   - Server default: approximately `11 × 1s + 1s` backoff = 12 seconds per open.
   - `MatViewFuzzTest`: `SPIN_LOCK_TIMEOUT = 100_000_000ms`, producing approximately 12.73 days per open.
   - `generateTransactions()` can execute two newly wrapped opens sequentially at lines 544 and 1430.

   Use one monotonic end-to-end deadline rather than an attempt count that resets the callee’s full timeout. The PR body must also remove the claim that this has “no behavioral downside.”

2. **[in-diff] The regression tests do not exercise the repaired callsites or the complete broadened timeout family**  
   `core/src/test/java/io/questdb/test/cairo/fuzz/FuzzRunnerTest.java:40-115`  
   `core/src/test/java/io/questdb/test/cairo/fuzz/FuzzRunner.java:544,1199-1200,1236,1332-1342,1400,1430`

   All five tests call `openWithRetries()` directly. Removing every real wrapper—including the exact observed `generateSymbols()` reader fix at line 1430—leaves them green. Existing fuzz tests execute successful opens but do not deterministically force the timeout.

   The new `Column Version read timeout` behavior is also untested. A matcher that excludes only that variant passes all five tests.

   Add deterministic seams at the routing methods and inject one timeout before a real successful open. Add an exact Column Version case. A helper-only test is not a regression test for the reported flake.

3. **[pre-existing] Recreate retries cannot observe cancellation and can strand the purge thread permanently**  
   `core/src/test/java/io/questdb/test/cairo/fuzz/FuzzRunner.java:1244-1267,1325-1355`  
   `core/src/test/java/io/questdb/test/fuzz/FuzzTransactionGenerator.java:615-620`

   A generated DROP/CREATE operation can complete DROP and then fail CREATE. The writer records the error, but `runPurgePartitionJob()` can remain forever inside the missing/reserved/locked-table branch. It never observes `done` or `errors`, closes its readers, or clears thread-local paths.

   Pass a cancellation predicate into recreate-tolerant opens and rethrow when `done != 0` or `errors` becomes non-empty. Make the synchronous drain path fail fast unless an actual concurrent recreator exists. Add a deterministic cancellation test.

4. **[pre-existing] `applyManyWalParallel()` scans errors before joining their producers**  
   `core/src/test/java/io/questdb/test/cairo/fuzz/FuzzRunner.java:196-241,1300-1370`

   Two failures result:

   - An error queued before lines 231-233 throws immediately and skips all apply/purge joins.
   - An error queued after that scan but before worker termination is silently lost because there is no post-join scan.

   Join every started worker in cleanup, then inspect and propagate `errors`. Test both a prequeued failure and a latch-controlled late publication.

5. **[pre-existing] View fuzz workers leak on exceptional exits**  
   `core/src/test/java/io/questdb/test/cairo/view/ViewFuzzTest.java:115-143,212-235,259-276`

   Both paths start non-daemon compiler threads before reaching changed `generateTransactions()` calls, but set `stop` and join only on success. Retry exhaustion or any later exception leaves the threads running against teardown. `assertMemoryLeak()` and JUnit teardown do not own these method-local threads.

   Put stop-and-join cleanup in `finally` for every successfully started worker and add failure-path tests.

6. **[pre-existing] Materialized-view fuzz workers have the same lifecycle defect**  
   `core/src/test/java/io/questdb/test/cairo/mv/MatViewFuzzTest.java:131-151,616-644,893-917,974-1003`

   Three paths can leave non-daemon refresh threads and `MatViewRefreshJob` resources alive when transaction generation or later work fails. Superclass teardown only halts a different shared worker pool.

   Centralize refresh-thread cleanup in a `finally`, join all started workers, preserve the original exception, and test partial startup plus failure after startup.

### Moderate

1. **[in-diff] The tests duplicate expected-exception boilerplate**  
   `core/src/test/java/io/questdb/test/cairo/fuzz/FuzzRunnerTest.java:43-53,73-81,105-114`

   Use JUnit 4.13.2 `Assert.assertThrows()` and `TestUtils.assertContains()`. This is behaviorally and performance-wise inert; it only improves maintainability and diagnostics.

2. **[in-diff] The unit tests spend 1.22 seconds sleeping and emit excessive expected-error stacks**  
   `core/src/test/java/io/questdb/test/cairo/fuzz/FuzzRunner.java:1255,1261`  
   `core/src/test/java/io/questdb/test/cairo/fuzz/FuzzRunnerTest.java:41-99`

   A measured run emitted 14 expected exception records and 536 stack frames. Inject a no-op sleeper/backoff callback for unit tests and rate-limit or summarize expected retry logging.

   Downgraded to Moderate because this runs only in the test harness and has a fixed bound of 14 failures and 1.22 seconds per class run. The unbounded recreate logging is covered by Critical finding 3.

### Minor

1. **[in-diff] Boolean naming violates project rules**  
   `FuzzRunner.java:1241,1244,1257`  
   Rename `tolerateTableRecreate` to `isTableRecreateTolerated`.

2. **[in-diff] Package-private method is inside the private-method group**  
   `FuzzRunner.java:1244-1267`  
   Move `openWithRetries()` after package-private `assertCounts()` and before the protected methods.

3. **[PR metadata] Metadata does not meet repository conventions**
   - Rewrite the title around test reliability rather than `reader-opens` internals.
   - Add the `Core` label.
   - Replace the remaining passive constructions with active voice.

- **Claim:** the unwrapped first metadata open breaks the non-WAL `FrameAppendFuzzTest` caller.  
  **Dismissed:** `FrameAppendFuzzTest.java:167-174` performs no concurrent structural DDL before generation. Although non-WAL `getLegacyMetadata()` delegates to `getTableMetadata()`, the conditions required for the transient mismatch are unreachable on this path.

### Coverage map

Legend: H = happy path, E = error path, B = boundary, C = concurrency, R = resource cleanup.

|---|---|---|---|
| Metadata timeout retries then succeeds | `testReadTimeoutRetriedThenSucceeds`; return and 3-attempt assertions fail if retry stops | H covered; E covered; NULL N/A—message sink is non-null; B covered; C N/A—local state; R N/A—synthetic value | TESTED |
| Transaction timeout broadening and exhaustion | `testReadTimeoutRethrownAfterRetriesExhausted`; exact 11 attempts fail if old Metadata-only match returns | H N/A; E covered; B covered; C N/A; R N/A | TESTED |
| Column Version timeout broadening | None; excluding this exact variant leaves all tests green | E uncovered; B uncovered; C N/A; R N/A | **UNTESTED** |
| False recreate gate fails fast | `testTableRecreateNotToleratedFailsFast`; asserts one attempt and rethrow | E covered; B covered; C N/A; R N/A | TESTED |
| True gate retries reserved name | `testTableRecreateToleratedIsWaitedOut`; deleting this term fails on attempt 1 | E covered; B covered; C N/A; R N/A | TESTED |
| True gate retries `EntryLockedException` | Same test; deleting this term fails on attempt 2 | E covered; B covered; C N/A; R N/A | TESTED |
| True gate retries missing table | No direct test, but behavior is unchanged from the old helper | No behavioral delta verified | EXEMPT |
| Unrelated `CairoException` rethrows | `testUnrelatedCairoExceptionAlwaysRethrown`; message and one-attempt assertions observe behavior | E covered; B covered; C N/A; R N/A | TESTED |
| `generateTransactions()` metadata routing | No deterministic test; removing line 544 leaves all new tests green | H nominally executed; E uncovered; C uncovered where applicable; R uncovered | **UNTESTED** |
| `drainWalQueue()` broadened reader routing | No deterministic timeout test | H nominally executed; E uncovered; C uncovered; R uncovered | **UNTESTED** |
| `runPurgePartitionJob()` broadened routing | No deterministic timeout/cancellation test | H nominally executed; E uncovered; C uncovered; R uncovered | **UNTESTED** |
| `assertStringColDensity()` reader routing | No deterministic timeout test | H nominally executed; E uncovered; C N/A—post-drain; R uncovered | **UNTESTED** |
| `generateSymbols()` reader routing—the observed flake | No deterministic test; reverting line 1430 leaves all new tests green | H nominally executed; E uncovered; C uncovered during WAL apply; R uncovered | **UNTESTED** |

Coverage map: **13 behavioral rows: 6 tested, 6 UNTESTED, 1 exempt**.

### Summary

**Verdict: request changes.**

- The retry policy introduces a severe timeout-amplification regression.
- The PR has no deterministic regression test for the actual repaired callsites.
- Four confirmed pre-existing lifecycle defects on visited paths must be fixed and tested in this PR.
- Validation: `mvn -pl core -Dtest=FuzzRunnerTest -DfailIfNoTests=false test` passed with 5 tests; `git diff --check` passed.
- Findings: 11 actionable finding groups verified; 1 false positive removed.
- Split: 7 in-diff/metadata, 0 out-of-diff contract breaks, 4 pre-existing.
- Tradeoffs: persistent timeout failures surface up to 11 times later; the new unit tests add fixed sleep and log amplification.

RaphDal:
# Review: PR #7375 — `test(core): stop fuzz setup reader-opens flaking on transient metadata-read timeouts`

- PR: https://github.com/questdb/questdb/pull/7375
- Review level: **3** (full pass — all agents, per-finding source verification). Normally reserved for replication/JNI/on-disk-format/API/ACL PRs; here the change is **test-harness only**, so the Rust-safety, enterprise-ACL, and SQL-dialect passes were no-ops. The scope that mattered — retry control flow, thread lifecycle, resource cleanup, and test efficacy — got 6 parallel review agents, per-finding source tracing, and an empirical local run.
- Base: `496b24d996` · Head reviewed: `30b93faea4` (+ the M1 fix applied to this working tree)
- Scope: 5 test files, ~549 insertions / 160 deletions.

**Verdict: approve with comments.** No Critical findings. The code is correct at every callsite traced; `FuzzRunnerTest` runs green locally. The open items were a low-probability new flake risk in one test (**M1 — now fixed in this working tree**), regression-coverage gaps on the bundled lifecycle fixes, and a stale GitHub PR body — all non-blocking.

This PR evolved substantially since the earlier maintainer review (`request changes`). The current head addresses every one of those Critical items: the 11×-spin amplification is gone (single monotonic deadline), the setup-open callsites are now genuinely regression-tested (routing counters), `applyManyWalParallel` error/thread handling is fixed, and the mat-view/view worker leaks are closed. Those fixes are real and correct.

**Submodule provenance:** no submodule pointer moved in the PR's committed range (`496b24d9...30b93fae` touches 5 test files only). The `M java-questdb-client` working-tree entry is local drift, not part of this PR — no provenance question.

---

## Critical

None.

---

## Moderate (non-blocking)

### M1 — Timing-flake risk in `testReadTimeoutRetriedThenSucceeds` — **FIXED in this working tree**
**Problem:** Test needed 2 retries to fit a 300ms budget; one slow sleep failed it.
**Net impact:** Rare spurious CI red on a saturated runner — the exact box this PR targets.

`FuzzRunnerTest.java` (`testReadTimeoutRetriedThenSucceeds`) drove `openWithRetries` with `RETRY_TIMEOUT_MILLIS = 300`. The supplier threw on attempts 1 and 2 and succeeded on 3. Trace of `openWithRetries` (`FuzzRunner.java:1506-1537`):
- attempt 1 @ T0 → throw → `remaining ≈ 300ms` → `backoff = min(100, 300) = 100` → `Os.sleep(100)`
- attempt 2 @ ~T0+`sleep1` → throw → deadline check: `remaining = 300 − sleep1`; if `sleep1 ≥ ~300ms` → `remaining ≤ 0` → **rethrow** → `assertEquals("reader", reader)` fails with `CairoException`.
- attempt 3 returns `"reader"` with no deadline check.

`Os.sleep` is `Thread.sleep` in a loop (`Os.java:231`) with no overshoot cap, so the test failed iff the first 100ms backoff actually took ≥300ms — a 3× overshoot. Low probability, but it is the same pathological-load condition ("heavily-loaded macOS CI runner") that produced the original flake, and the repo's no-flaky policy is strict.

**Fix applied:** decoupled the recover-then-succeed budget from the deadline test's small budget. Added `AMPLE_RETRY_TIMEOUT_MILLIS = 10_000` and switched `testReadTimeoutRetriedThenSucceeds` to it; `RETRY_TIMEOUT_MILLIS = 300` stays for `testReadTimeoutRethrownAfterDeadline`, which must spend its whole budget sub-second. The two ~100ms backoffs now recover at ~200ms with ~50× headroom, so no realistic sleep overshoot can blow the deadline. Re-ran locally: `Tests run: 11, Failures: 0, Errors: 0`.

### M2 — The single-deadline mechanism (the headline fix) is not regression-guarded
**Problem:** Deadline test can't distinguish a per-attempt count from a single budget.
**Net impact:** A future revert to spin-multiplying retries would ship un-caught.

`testReadTimeoutRethrownAfterDeadline` asserts only a *floor* (`elapsed ≥ RETRY_TIMEOUT_MILLIS`) and `attempts ∈ (1,20]`, with a supplier that throws instantly (no spin to multiply). Reverting the deadline logic to the old fixed-count loop (`10 retries / Os.sleep(100)` → ~1000ms / 11 attempts) sits inside `[≥300ms, ≤20]` and **passes**. So the specific amplification the earlier review flagged as Critical is not locked in. In-diff coverage gap. **Suggested fix:** assert an elapsed *ceiling* (e.g. `< ~2× budget`) or drive it with a supplier that delays before throwing and assert total elapsed ≈ budget, not `N × per-attempt`.

### M3 — The bundled worker-lifecycle fixes have no failing-assertion regression test
**Problem:** Three lifecycle fixes on this branch regress silently.
**Net impact:** A revert of any of them keeps the suite green; the thread-leak / lost-error behavior returns unnoticed.

All verified *correct* by the correctness, concurrency, and resource agents — the gap is purely missing tests:
- **`applyManyWalParallel` finally-restructure** (`FuzzRunner.java:226-248`): the two new behaviors — surfacing a worker error queued *after* the writers finish, and not leaking apply/purge threads when a write-thread error is scanned first — are never triggered. `testPurgePartitionReadersUseRetries` only runs the happy path (empty transactions, no error). Reverting to the master structure passes all tests.
- **Purge cancellation early-return** (`runPurgePartitionJob`, `FuzzRunner.java:1410-1412, 1424-1426`): the `if (reader == null) return;` path plus its `finally`-free is never exercised; only the helper's null-return is tested in isolation (`testCancelledDuringRecreateWaitReturnsNull`).
- **`stopAndJoinJobs` failure path** (`AbstractFuzzTest.java:428`, mv/view try/finally): its whole reason to exist — joining a partially-started batch when a throw fires between start and join — is never forced; only the happy path runs, where the `finally` is redundant.

These are internal test-harness branches (no product-user impact), each a strict improvement over master, so a regression only returns to master's shipped-but-buggy behavior — Moderate, not Critical. **Suggested fix:** one test enqueueing a late worker error and asserting `applyManyWalParallel` throws + no thread outlives the call; and a mat-view/view test forcing a throw mid-batch and asserting the workers are joined.

---

## Minor

- **m1 — `getLegacyMetadata` comment is false for non-WAL tables.** `FuzzRunner.java:544-546` says it "reads the sequencer from memory and cannot [time out]." For a non-WAL token `CairoEngine.getLegacyMetadata` delegates to `getTableMetadata` (`CairoEngine.java:2133-2135`), which *can* throw a read timeout. The one non-WAL caller (`FrameAppendFuzzTest` → `createInitialTableNonWal` → `generateTransactions`) generates with no concurrent structural apply and non-WAL DDL is synchronous, so it is safe today — and both metadata opens were bare on master, so no regression (strictly ≥ base). But the unqualified "cannot" will mislead a maintainer. **Fix:** scope it, e.g. "for a WAL table, getLegacyMetadata reads the sequencer from memory (a non-WAL token would delegate to getTableMetadata)." *(Corroborated by 3 independent agents + direct trace.)*
- **m2 — "none of them may outlive this call" is not guaranteed.** `FuzzRunner.java:235-240`: the apply-thread join is sequenced *after* `Misc.freeObjList(writers)` in the same `finally`. If a `WalWriter.close()` threw, the join is skipped and the workers leak — contradicting the comment. Not a regression (master also threw before joining) and `close()` throwing is unrealistic here. **Optional:** put the apply-join in its own nested `finally`, or soften the comment.
- **m3 — GitHub PR body is stale relative to the current head.** It still describes the removed `getReaderHandleTableDropped`, the old param name `tolerateTableRecreate`, and "up to 10x, 100 ms backoff" (the current code is a single 10s deadline), and omits the bundled thread-lifecycle fixes. A re-reviewer would think the amplification bug is still present. (The local `PR.md` draft is already updated with the correct mechanism, a worker-lifecycle section, and `Core`/`Flaky Test` label suggestions — just push it to the PR.)
- **m4 — Labels.** Only `Flaky Test` is set on the PR; add `Core` (test-harness in the core module) and arguably `CI`.
- **m5 — Misleading test comment.** `FuzzRunnerTest` (`testPurgePartitionReadersUseRetries`) says the cancellation predicate "can legitimately end the wait" — but with `injectedCall=0` and a valid table, `getReader` never throws, so `isRunOver` is never consulted; the test verifies route *counts*, not cancellation. Reword.
- **m6 — A deadline regression would hang, not fail cleanly.** The direct-helper tests have no `@Test(timeout=...)`; if the deadline check were moved inside the loop, several tests loop forever and fail only via the class-level 20-minute `Timeout` rule. Caught, but slowly. Minor robustness.

---

## Downgraded (false positives ruled out during verification)

- **Broadened `"read timeout"` swallows `"block file read timeout"`** → **Dismissed.** `Chars.contains(msg, "read timeout")` does substring-match it, but `BlockFileReader` is constructed only by settings / SHOW CREATE / mat-view-hydration / checkpoint code — never by `TableReader`, `TableReaderMetadataTenantImpl`, or `TableReaderMetadata` (grep-confirmed). Unreachable from all 8 open sites; the match can only catch the three real spin transients.
- **Resource leak multiplied across 10 retries** → **Dismissed.** Every throwing `open.get()` fully releases its partial reader/metadata on both pool paths (`AbstractMultiTenantPool.get0` refresh-throw and newTenant-throw; `TableReader`/`TableReaderMetadataTenantImpl` ctor `close()`-on-throw). The helper holds no reference between attempts.
- **Branch reorder / broadening breaks the 5 concurrent-read sites** → **Dismissed.** The read-timeout and recreate message literals are disjoint (verified), and `EntryLockedException` is matched by `instanceof`, so no message satisfies both branches; the broadening only *adds* retrying of the Transaction/Column-Version variants the old code rethrew — strictly more tolerant, still bounded.
- **New join hang in `applyManyWalParallel`** → **Dismissed.** `done.incrementAndGet()` precedes the apply-join, all workers loop on `done==0 && errors.isEmpty()`, write threads are joined (WAL production stopped) before the apply-join, and `runPurgePartitionJob`'s in-flight open bails via `isRunOver` once `done != 0`. Joins are bounded.
- **Fix is incomplete — the bare first metadata open still flakes** → **Dismissed.** For WAL tables `getLegacyMetadata` → in-memory `getSequencerMetadata` (no spin); all fuzz callers pass WAL base tables. Captured as the latent m1 note.

---

## Adjacent findings (pre-existing; not blocking, file separately if desired)

- **A1 — `drainWalQueue`'s recreate-tolerant open has no deadline and no cancellation.** `FuzzRunner.java:1302-1303` calls `getReaderWithRetries(tableName, null)`; the recreate branch (`FuzzRunner.java:1532-1534`) only `Os.sleep(10)` and loops, never consulting `deadlineNanos`, and `isCancelled == null`. A permanently-missing table would spin forever. **Reachability:** none today — drain runs after applies complete and drop ops are paired with creates. **Not introduced by this PR:** identical to the removed `getReaderHandleTableDropped`. Severity if filed standalone: Minor. Note the asymmetry — `runPurgePartitionJob` was upgraded to pass `isRunOver` and can bail, while `drainWalQueue` retained the unbounded wait; passing a deadline or cancellation there would close it.

---

## Coverage map (test-harness symbols only — no product behavior changed)

| Change | Test | Failure link | Verdict |
|---|---|---|---|
| `openWithRetries` read-timeout retry→succeed | `testReadTimeoutRetriedThenSucceeds` | asserts "reader" after 3 attempts (M1 flake risk — now fixed) | TESTED |
| `openWithRetries` read-timeout rethrow at deadline | `testReadTimeoutRethrownAfterDeadline` | floor+attempts bound; does NOT distinguish deadline vs count (M2) | TESTED (partial) |
| Broadened `"read timeout"` matches Metadata/Transaction/Column Version | routing tests inject each variant through the real substring match; `readTimeoutAttempts==2` | narrowing the match → variant rethrown → count/attempt mismatch | TESTED |
| `isTableRecreateTolerated=false` fail-fast | `testTableRecreateNotToleratedFailsFast` | 1 attempt, rethrow | TESTED |
| `isTableRecreateTolerated=true` waits recreate (reserved + EntryLocked) | `testTableRecreateToleratedIsWaitedOut` | 3 attempts, success | TESTED |
| `isCancelled` → return null | `testCancelledDuringRecreateWaitReturnsNull` | null + 3 attempts (removing check → 20-min hang) | TESTED |
| Unrelated `CairoException` rethrown | `testUnrelatedCairoExceptionAlwaysRethrown` | 1 attempt, rethrow even w/ recreate tolerated | TESTED |
| `generateTransactions` getTableMetadata retried (false) | `testGenerateTransactionsUsesRetries` | route 1/0/0; revert wrapper → openCalls 0 → fails | TESTED |
| `generateSymbols` reader retried (false) | `testGenerateSymbolsUsesRetries` | route 1/0/0; revert → fails | TESTED |
| `assertStringColDensity` reader retried (false) | `testAssertStringColDensityUsesRetries` | route 1/0/0; revert → fails | TESTED |
| `drainWalQueue` readers via getReaderWithRetries (true) | `testDrainWalQueueUsesRetries` | route 4/4/0; revert → fails | TESTED |
| `runPurgePartitionJob` initial opens (true, cancellable) | `testPurgePartitionReadersUseRetries` | route 2/2/2; revert → fails | TESTED (routing) |
| `runPurgePartitionJob` cancellation early-return + reload free | — | — | UNTESTED (M3; internal branch) |
| `applyManyWalParallel` finally restructure (always-join, error-after-join) | — happy path only | — | UNTESTED (M3; internal branch) |
| `stopAndJoinJobs` + mv/view try/finally (leak-prevent on throw) | — happy path only | — | UNTESTED (M3; internal branch) |

The routing counters are a genuine regression guard, not a change-detector: `openCalls`/`recreateTolerantCalls`/`cancellableCalls` increment **only** inside the overridden `openWithRetries` (`FuzzRunnerTest`), so any callsite reverted to a bare open bypasses the counter and the `assertEquals` fails — established by construction and confirmed by the green run. (A separate revert-in-worktree wasn't run because a full test-compile needs the client native lib rebuilt; the structural proof plus the passing suite is conclusive here.)

---

## Summary

- **Verdict: approve with comments.** No Critical findings; both gates pass. M1 (the flake risk) is fixed in this working tree. M2/M3 (coverage) and m3 (stale GitHub PR body) are worth addressing but do not block.
- **Correctness gate:** passes. No open Critical, in-diff or out-of-diff. Severity audit run both ways: nothing downgraded from Critical (there were none); nothing in Moderate completes a product-user symptom sentence (all are developer/CI-visible, correctly capped at Moderate).
- **Test gate:** passes. No product behavior changed, so no product-user-visible UNTESTED rows; the PR is correctly typed `test(core)`, not a customer `fix`. The headline behavior (routing setup opens through the retry helper) *is* regression-tested. The UNTESTED rows are internal test-harness branches (Moderate), not blocking.
- **Empirical:** `FuzzRunnerTest` — `Tests run: 11, Failures: 0, Errors: 0` (ran locally via mvnd surefire against the current head, and again after the M1 fix).
- **Coverage-map totals:** 15 behavioral changes (all test-harness), 12 tested, 3 UNTESTED (internal branches, M3).
- **Findings:** 9 verified (3 Moderate, 6 Minor), 5 dropped as false positives, 1 adjacent. Severity distribution: **0 Critical, 3 Moderate, 6 Minor**.
- **Split:** all 9 in-diff; 0 out-of-diff-breakage (the changed retry helpers are `FuzzRunner`-internal with no external callers, and the public methods' signatures are unchanged, so no caller breaks — confirmed by the callsite inventory + the cross-context and adversarial agents); 1 adjacent (pre-existing `drainWalQueue` unbounded recreate wait).
- **Regressions/tradeoffs:** the only behavioral tradeoff is the intended one — the 5 concurrent-read sites now also retry the Transaction/Column-Version timeout variants (bounded by the 10s deadline), strictly more tolerant of the same transient class with no correctness downside. The restructures are net-positive: they fix real master bugs (leaked apply/purge threads on error, lost apply-thread errors, leaked mat-view/view workers on a mid-run throw, a latent double-close in the purge reload loop).

---

### Review method
6 parallel fresh-context agents (correctness + cross-context, concurrency, resource management, test efficacy + regression value, test-code quality + PR metadata, fresh-context adversarial) plus direct source tracing and a local `FuzzRunnerTest` run. Every finding was verified against source before inclusion; false positives were moved to the Downgraded section.

bluestreak01 (approved):

PR #7376, opened 2026-07-09
fix(sql): clarify materialized view error when timestamp lacks a sampling interval

### LINKED ISSUE #7356: Time-based GROUP BY materialized views ignore designated TIMESTAMP
### To reproduce

I don't know whether to file this as a bug or as a feature/documentation request. From a user perspective, creating materialized views with time-based GROUP BY can be quite finicky, and it would be helpful to have more documentation, along with examples and limitations on usage.

For example, the following statement fails with the error `TIMESTAMP column is not present in select list`.

```sql
-- Mock table with 1-hour timestamps
CREATE TABLE sensor_1h AS (
    SELECT
        generate_series AS ts,
        (rnd_float() * 150) AS temp_f,
        (rnd_float() * 150) AS wind_speed_mph
    FROM generate_series('2025-01-12', '2025-12-31', '1h')
) TIMESTAMP(ts)
PARTITION BY DAY;

-- Attempt to create a materialized view with time-based GROUP BY
CREATE MATERIALIZED VIEW wind_chill_1h
WITH BASE sensor_1h REFRESH IMMEDIATE AS (
    (
        SELECT
            ts,
            last(35.74 + 0.6215 * temp_f - 35.75 * power(wind_speed_mph, 0.16) + 0.4275 * temp_f * power(wind_speed_mph, 0.16)) AS wind_chill_f
        FROM sensor_1h
        ORDER BY ts
    ) TIMESTAMP(ts)
)
PARTITION BY DAY;
```

As shown above, `ts` is the designated timestamp and is listed in the `SELECT` clause.

The only workaround seems to be either using the `SAMPLE BY` syntax or using time-based GROUP BY with the `timestamp_floor()` function. And, upon exploring further, the `timestamp_ceil()` function does not work as a suitable replacement; there's no documentation as to why this is the case, and the intuition behind for this limitation.

Below is a workaround using the `timestamp_floor()` function:

```sql
CREATE MATERIALIZED VIEW IF NOT EXISTS 'wind_chill_1h'
WITH BASE sensor_1h REFRESH IMMEDIATE AS (
    SELECT
        timestamp_floor('1h', ts) AS ts,
        last(35.74 + 0.6215 * temp_f - 35.75 * power(wind_speed_mph, 0.16) + 0.4275 * temp_f * power(wind_speed_mph, 0.16)) AS wind_chill_f
    FROM sensor_1h
)
PARTITION BY DAY;
```

I understand that, in theory, you should be able to achieve all GROUP BY–related functionality using the `SAMPLE BY` syntax; however, there are real performance concerns that may necessitate avoiding it. This is especially true when trying to materialize billions of rows of data, where a `SAMPLE BY` operation can take hours to complete. From my estimates, running the underlying materialized queries directly shows that GROUP BY (without the `timestamp_floor()` function) can be 30x faster than `SAMPLE BY` (cold start, no caching).

In my use case, I already have a table with timestamps on a fixed interval, so there's no benefit to recomputing the intervals. The only expensive operation would be the binning and any ancillary transformations performed in the materialized view.

### QuestDB version:

9.4.3

Docker / macOS Tahoe 26.5.1

APFS

### Full Name:

Giovanni Thenstead

### Affiliation:

Beta Bytes LLC

- [x] Yes, I have

### Additional context

_No response_

Socialpranker:
Fixes #7356

## Summary

`CreateMatViewOperationImpl.validateAndUpdateMetadataFromModel()` throws "TIMESTAMP column is not present in select list" for every `CREATE MATERIALIZED VIEW` query that has neither a `SAMPLE BY` clause nor a `GROUP BY timestamp_floor(...)` column, regardless of whether the designated timestamp column is actually present in the select list or not.

As reported in #7356, a query such as:

```sql
CREATE MATERIALIZED VIEW wind_chill_1h
WITH BASE sensor_1h REFRESH IMMEDIATE AS (
    SELECT ts, last(...) AS wind_chill_f
    FROM sensor_1h
    ORDER BY ts
) TIMESTAMP(ts)
PARTITION BY DAY;
```

fails with that message even though `ts` is clearly present in the select list. The real problem, as the reporting maintainer clarified in the issue thread, is that no sampling interval could be inferred — a materialized view needs either `SAMPLE BY <interval>` or `GROUP BY timestamp_floor(<interval>, ts)` to define the fixed bucket the incremental refresh walks. A plain `GROUP BY` on the designated timestamp doesn't provide that, so it's rejected, but the existing error message names the wrong problem.

## Change

By the point the code decides no sampling interval was found, it has already confirmed (earlier in the same method) that the designated timestamp column, when present in the DDL's `TIMESTAMP(...)` clause, exists in the select list — otherwise it would have thrown a different, more specific error already. This PR uses that existing invariant to distinguish the two cases:

- Timestamp column present but no interval inferred: new message names the column and points at `SAMPLE BY` / `GROUP BY timestamp_floor(...)` as the two supported forms.
- No designated timestamp column at all: unchanged generic message, since that case really does mean the column is missing.

## Test plan

- Added `testCreateMatViewGroupByPlainTimestampNoSamplingInterval` to `CreateMatViewTest`, covering the exact query shape from the issue (`SELECT ts, avg(v) ... GROUP BY` implied by aggregation, with an explicit `TIMESTAMP(ts)` clause, no `SAMPLE BY`/`timestamp_floor`).
- Ran the new test alongside the three existing tests that exercise the same code path (`testCreateMatViewGroupByNoTimestamp`, `testCreateMatViewNoSampleBy`, `testCreateMatViewGroupByTimestamp3`) to confirm no regression: all 4 pass.
- Did not run the full test suite; the change is confined to a single error-message branch with no other callers.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/ops/CreateMatViewOperationImpl.java
+++ b/core/src/main/java/io/questdb/griffin/engine/ops/CreateMatViewOperationImpl.java
@@ -429,6 +429,15 @@ public class CreateMatViewOperationImpl implements CreateMatViewOperation {
 
         // We haven't found timestamp_floor() in SELECT.
         if (intervalExpr == null) {
+            if (timestamp != null) {
+                // The designated timestamp column was already confirmed present in the select
+                // list above, but the query has neither a SAMPLE BY nor a GROUP BY
+                // timestamp_floor(...), so no sampling interval could be inferred. Point the
+                // user at the two supported forms instead of claiming the column is missing.
+                throw SqlException.position(selectTextPosition)
+                        .put("materialized view query requires a sampling interval, use SAMPLE BY or GROUP BY timestamp_floor() [name=")
+                        .put(timestamp).put(']');
+            }
             throw SqlException.$(selectTextPosition, "TIMESTAMP column is not present in select list");
         }
 
--- a/core/src/test/java/io/questdb/test/cairo/mv/CreateMatViewTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/mv/CreateMatViewTest.java
@@ -549,6 +549,21 @@ public class CreateMatViewTest extends AbstractCairoTest {
     }
 
     @Test
+    public void testCreateMatViewGroupByPlainTimestampNoSamplingInterval() throws Exception {
+        assertMemoryLeak(() -> {
+            createTable(TABLE1);
+            // ts is the designated timestamp and is present in the select list, but there is
+            // neither a SAMPLE BY nor a GROUP BY timestamp_floor(...), so no sampling interval
+            // can be inferred. The error must not claim the timestamp column is missing.
+            assertQuery("create materialized view test as (select ts, avg(v) from " + TABLE1 +
+                    ") timestamp(ts) partition by day")
+                    .noLeakCheck()
+                    .fails(34, "materialized view query requires a sampling interval, use SAMPLE BY or GROUP BY timestamp_floor() [name=ts]");
+            assertNull(getMatViewDefinition("test"));
+        });
+    }
+
+    @Test
     public void testCreateMatViewGroupByTimestamp1() throws Exception {
         assertMemoryLeak(() -> {
             createTable(TABLE1);


bluestreak01 (approved):
# Review: PR #7376 — `fix(sql): clarify materialized view error when timestamp lacks a sampling interval`

**Level 3** (full mission-critical pass: change-surface map, coverage map, 4 parallel fresh-context review agents covering correctness/cross-context, adversarial, test-efficacy/regression, and code-quality/metadata; per-finding verification; empirical test run). Fixes #7356.

## What the PR does

In `CreateMatViewOperationImpl.validateAndUpdateMetadataFromModel(...)`, the `intervalExpr == null` branch previously always threw `"TIMESTAMP column is not present in select list"`. That message fires for any `CREATE MATERIALIZED VIEW` lacking a `SAMPLE BY` or `GROUP BY timestamp_floor(...)`, even when the designated timestamp column *is* in the select list (the issue's `SELECT ts, avg(v) ... TIMESTAMP(ts)` shape). The PR splits the branch on `timestamp != null` and, in that case, emits a message that names the column and points at the two supported forms. One regression test is added.

The PR is the single remote commit `f54338dd1b` (2 files, +24 lines).

## Correctness verification (the load-bearing claim)

The code comment asserts the designated timestamp is *guaranteed present in the select list* when this new branch is reached with `timestamp != null`. Verified true:
- `timestamp` is `final = createTableOperation.getTimestampColumnName()` and is never reassigned in the method.
- The earlier unconditional block (`:371-387`) does `createColumnModelMap.get(timestamp)` — the **same** case-insensitive map (`LowerCaseCharSequenceObjHashMap` of the top-level select output names) keyed the **same** way — and throws `"TIMESTAMP column does not exist [name=...]"` if absent.
- The `timestamp_floor` block only calls `setTimestampColumnName(...)` when `timestamp == null`, and whenever it finds a floor column it sets `intervalExpr` non-null, so the changed branch is not reached from there.

Behavior beyond the message is unchanged: before `SqlException.$(selectTextPosition, msg)`, after `SqlException.position(selectTextPosition).put(...)` — both throw at the same absolute position (`selectTextPosition`, empirically 34 for the test SQL) and abort the create. No allocations, native handles, locks, or shared state on this error path — concurrency/resource/performance/Rust agents are N-A by construction and produced nothing.

## Critical
None.

## Moderate
- **Optional coverage nicety — no positive test that `TIMESTAMP(ts)` + `SAMPLE BY`/`timestamp_floor` still *succeeds*** (the "new branch doesn't over-fire" case). This is behaviorally inert today: the new block is guarded by `if (intervalExpr == null)`, so it is structurally unreachable whenever a sampling interval exists, and existing success tests (`testCreateMatViewGroupByTimestamp1/2`) already exercise the `intervalExpr != null` path. A one-liner — `create materialized view test as (select ts, avg(v) from table1 sample by 30s) timestamp(ts) partition by day` succeeding — would pin the exact `timestamp != null && intervalExpr != null` combination against a future refactor that moves the guard. Non-blocking.

## Minor
- **PR has no labels.** Per CLAUDE.md, add `SQL` and `Materialized View` (and optionally `Bug`).
- Message begins lowercase (`materialized view query requires...`) while the sibling `TIMESTAMP column ...` messages are capitalized — **not a real inconsistency**: the uppercase there is the `TIMESTAMP` keyword, and lowercase sentence-style is the norm in this file (`"base table has to be WAL enabled"`, `"materialized view query is required to have designated timestamp"`). No change needed.

- **"A user who wrote `SAMPLE BY` could still get the new 'use SAMPLE BY' message."** Two agents raised this: `findSampleByNode` only returns a node when `type == CONSTANT` and stops at non-plain nested models. Dropped — this detection logic is **entirely unchanged by the PR**, neither agent could construct a reachable query for it (SAMPLE BY is normally rewritten to a `timestamp_floor(...)` projection that `findTimestampFloorColumn` then catches), and the pre-PR message was *strictly more* misleading for the identical input. Pre-existing at worst, not worsened, not reachable.
- **Title uses "clarify" instead of the literal "fix(sql): fix …" example.** Dropped — the verb-repeat rule's intent is that the description lead with a verb and read standalone; "clarify materialized view error…" satisfies it and is end-user-focused.
- **assertMemoryLeak wrapping a builder call (usual anti-pattern).** Dropped — the lambda holds three statements (`createTable` + `assertQuery(...).fails(...)` + `assertNull`), so the outer leak-check legitimately scopes the whole body and `.noLeakCheck()` correctly suppresses the builder's own wrapper. This is the sanctioned multi-statement form.

## Coverage map
| Behavioral change | Test | Failure link | Dimensions | Verdict |
|---|---|---|---|---|
| `intervalExpr==null && timestamp!=null` → new message at `selectTextPosition` | `testCreateMatViewGroupByPlainTimestampNoSamplingInterval` (`:551`) | `.fails(34, <new msg>)`; on revert the message reverts to the old string at the same pos 34 → `assertContains` fails. **Empirically passes on PR branch.** | error-path ✓; NULL N-A (guard ensures `timestamp` non-null); boundary N-A; concurrency N-A (no shared state); resource ✓ (`assertMemoryLeak`) | **TESTED** |
| `intervalExpr==null && timestamp==null` → old message retained | `testCreateMatViewGroupByNoTimestamp` (`:545`), `testCreateMatViewNoSampleBy` (`:803`) — both verified to declare no `TIMESTAMP(...)` clause | Assert old message at 34; still pass, proving the old branch still fires. **Empirically pass.** | error-path ✓ | **TESTED** |
| Over-fire guard: `timestamp!=null && intervalExpr!=null` still succeeds | (indirect) `testCreateMatViewGroupByTimestamp1/2` cover `intervalExpr!=null`; none pairs it with an explicit `TIMESTAMP(ts)` clause | New block guarded unreachable when interval exists → inert | happy-path ✓ (indirect) | **UNTESTED (Moderate, inert)** |

## Summary
- **Verdict: approve.** No confirmed correctness, data-integrity, resource, concurrency, or performance/IO finding — this is a self-contained error-message split on a compile-time validation path, and every claim it relies on (timestamp-in-select-list guaranteed, position unchanged, message ASCII, old branch still reachable) is verified.
- **Correctness & performance gate:** satisfied — zero open correctness/perf findings.
- **Test gate:** satisfied — the fix ships with a regression test whose failure link is verified (position identical, message substring differs on revert) and which **empirically passes** (`Tests run: 4, Failures: 0`).
- Coverage map: 3 behavioral rows, 2 TESTED, 1 UNTESTED (inert over-fire combination, non-blocking Moderate).
- 4 agent reviews synthesized; ~0 blocking draft findings, 3 notes dropped as false positives/non-findings on verification.
- In-diff / out-of-diff / pre-existing split: **0 blocking in-diff, 0 out-of-diff, 1 pre-existing observation** (the `SAMPLE BY`-detection gap — not reachable, not worsened, not a blocker).
- Before merge (non-blocking): add `SQL` + `Materialized View` labels; optionally add the one-line success test.

PR #7377, opened 2026-07-09
feat(sql): compare a numeric column with a scalar sub-query in predicates

bluestreak01:
## Summary

Adds `<`, `>`, and `=` comparisons between numeric expressions and single-column scalar sub-queries. The implementation supports `DOUBLE`, `FLOAT`, `LONG`, `INT`, `SHORT`, and `BYTE` left operands while preserving integer precision and evaluating the sub-query once per query execution. The `<=`, `>=`, `!=`, and `<>` forms, and the argument-swapped `(sub-query) <op> col` forms, are derived automatically from these factories.

```sql
SELECT * FROM trades
WHERE price > (
    SELECT avg(price) FROM trades WHERE timestamp IN today()
);

-- equality against a scalar sub-query; != and <> are derived automatically
SELECT * FROM orders
WHERE quantity = (SELECT max(quantity) FROM orders);
```

The PR also updates compiler state management, parallel worker state donation, runtime timestamp intervals, plan rendering, and resource rollback so scalar cursor functions work in nested and parallel query plans.

## SQL behavior

Nine new factories provide the numeric comparison overloads:

| Comparison width | Factories | Signatures |
|---|---|---|
| `double` | `LtDoubleCursorFunctionFactory`, `GtDoubleCursorFunctionFactory`, `EqDoubleCursorFunctionFactory` | `<(DC)`, `>(DC)`, `=(DC)` |
| `int`/`byte`/`short` | `LtIntCursorFunctionFactory`, `GtIntCursorFunctionFactory`, `EqIntCursorFunctionFactory` | `<(IC)`, `>(IC)`, `=(IC)` |
| `long` | `LtLongCursorFunctionFactory`, `GtLongCursorFunctionFactory`, `EqLongCursorFunctionFactory` | `<(LC)`, `>(LC)`, `=(LC)` |

Only `<`, `>`, and `=` are registered. `FunctionFactoryCache` derives the remaining operators from them: it negates the result to produce `>=`/`<=` (from `<`/`>`) and `!=`/`<>` (from `=`), and—because the two operand types differ—it swaps the arguments so the cursor may sit on either side of the operator. The three families share one width-routing base per left type and one cold-lifecycle base (`AbstractScalarCursorFunction`), so `=` reuses the sub-query execution, worker state donation, cardinality enforcement, and plan rendering introduced for `<`/`>`.

The comparison selects a width that does not narrow the scalar value:

- Integer operands and integer cursor values use `long` comparison, including values beyond the exact integer range of `double`.
- `FLOAT` and `DOUBLE` values use `double` comparison.
- The overload matrix routes each numeric left type to the matching factory without implicitly converting integers to timestamps.

The scalar cursor follows these rules:

- Zero rows produce no match.
- One row supplies the cached scalar value.
- More than one row raises `scalar sub-query returned more than one row`.
- Strict comparisons (`<`, `>`) with a null scalar produce no match.
- Equality and the negated inclusive forms (`=`, `<=`, `>=`) preserve QuestDB's sentinel-null convention: a null (or empty) scalar matches a null left operand (`null = null`), while the strict forms and `!=` exclude it.

The multi-row check also tightens existing designated-timestamp `=`, `<`, and `>` scalar sub-query predicates. Runtime interval evaluation now rejects extra rows instead of silently using the first row.

A bare `null` literal no longer binds to a cursor overload. `col <= null`, `col >= null`, `col = null`, and `col != null` continue through scalar null-comparison resolution instead of failing while requesting a cursor factory from the null constant.

Adding the numeric overloads changes some invalid-query diagnostics: an `INT` left operand on an `=` sub-query now routes to the numeric `=(IC)` overload instead of failing against the timestamp candidate, and unsupported `STRING` and `VARCHAR` operands now report the numeric cursor candidate (`DOUBLE or FLOAT`) rather than the timestamp candidate.

## Compiler and execution changes

- Scalar cursor functions open, read, and close their cursor during `Function.init()` and cache the result for row evaluation.
- `offerStateTo()` donates cached state to parallel worker clones, so async filters, parallel GROUP BY, window joins, horizon joins, and multi-horizon joins evaluate the scalar sub-query once per execution rather than once per worker.
- The SQL compiler preserves invocation-local scratch state during recursive compilation and transfers parsed worker projection functions through explicit ownership paths.
- GROUP BY clone assembly and SAMPLE BY factory construction close adopted functions and factories when compilation or constructor setup fails.
- Runtime interval models carry scalar cursor error positions and transfer function ownership only after model construction succeeds.
- Nested plan attributes restore the caller's base-metadata mode after rendering child plans.
- Record cursor factories make `close()` idempotent, which protects shared cleanup paths from duplicate teardown.
- Materialized-view refresh contexts forward intrinsic-model overrides while retaining the refresh token checks.

## Tradeoffs and limitations

- This PR adds numeric `<`, `>`, and `=` factories. `FunctionFactoryCache` derives the `>=`, `<=`, `!=`, and `<>` forms—and the argument-swapped cursor-left forms—from them automatically, so no separate negated or swapped factories are registered.
- A `long` left operand compared against a `float`/`double` cursor scalar compares in `double`, so equality cannot distinguish values within one ULP beyond 2^53; a `long`-vs-`long` cursor comparison stays exact. This matches the width-routing already used by `<`/`>`.
- The comparison factories require the cursor on the right. The optimizer's existing operand swap supports cursor-left expressions when it can rewrite them to a registered column-to-cursor overload; cursor-to-cursor comparisons and arbitrary cursor-left rewrites remain unsupported.
- The scalar sub-query must return one column and at most one row. Queries that previously relied on the first row of a multi-row designated-timestamp sub-query now fail explicitly.
- The cached scalar is execution-scoped and does not implement correlated scalar sub-queries.
- The compiler and ownership changes widen the PR beyond the new SQL overloads because nested and parallel plans require the same scalar cursor state and cleanup contracts.

## Developer tooling

- The `fix-pr` skill now delegates each finding to fresh-context worker and reviewer agents while storing durable workflow state outside the repository.
- The `review-pr` skill no longer imposes artificial turn or tool-call budgets on review and verification agents.

These changes affect repository agent workflows rather than database runtime behavior. They increase the PR's review surface independently of the scalar sub-query changes.

## Testing

The added and updated tests cover:

- Numeric `<`, `>`, `<=`, and `>=` results, null and empty cursors, mixed numeric widths, and precision around 2^53.
- Numeric `=`, `!=`, and `<>` results across every supported left type (`DOUBLE`, `FLOAT`, `LONG`, `INT`, `SHORT`, `BYTE`), argument-swapped cursor-left equality, null and empty cursors following the `null = null` convention, no-narrowing `long` cursor scalars, exact `long` equality beyond 2^53, multi-row/multi-column/unsupported-operand errors, plan rendering, worker state donation, and continued timestamp `=(NC)` routing.
- Multi-column, unsupported-type, and multi-row cursor errors, including designated `TIMESTAMP` and `TIMESTAMP_NS` interval predicates.
- Async filter, parallel GROUP BY, window join, horizon join, and multi-horizon join state donation with worker execution counters.
- Recursive compiler use, plan metadata restoration, clone-construction rollback, GROUP BY cleanup, SAMPLE BY construction failure, and interval-model ownership.
- Cleanup continuation, exception suppression, retry safety, and close idempotence when adopted resources throw from `close()`.
- The ordinary-`ObjList` lifecycle fallback and allocator isolation between borrowed owner functions and owned worker clones.
- `FROM` and `TO` function ownership on the `SAMPLE BY ... FILL(LINEAR)` interpolation path, including cached-factory reuse.
- Materialized-view intrinsic override forwarding and idempotent cursor-factory closure.

The `=` overloads add a focused `EqCursorFunctionFactoryTest` (21 tests) and update the two `EqTimestampCursorFunctionFactoryTest` diagnostics that shift with the new routing. The affected cursor-comparison, function-parser, where-clause, lateral-join, and registry suites pass locally.

The full Maven test suite was not run locally as part of this validation. CI provides the broader platform and suite coverage.


bluestreak01:
# Review — `feat(sql): compare a numeric column with a scalar sub-query in predicates`

Mission-critical review (adversarial, blocking). 11 fresh-context review agents across correctness, concurrency, performance (×2 anti-anchoring), resources, tests, cross-context caller impact, and PR metadata; every substantive finding verified against source. No Rust agent — zero `.rs` files.

Scope: net diff vs merge-base `0727e7a3fe` — **126 files, +17,305 / −1,855**. (An earlier `master...` comparison showed a phantom `java-questdb-client` submodule bump; against the true merge-base there is **no submodule change**.)

## Title & description
Compliant. Conventional-Commits, verb-leading, reads as a release note. Body is level-headed with a dedicated **Tradeoffs and limitations** section (no `=`/`!=` variants, cursor-must-be-on-right, >2^53 precision loss, execution-scoped cache, honest admission the ownership work widens the PR), lists each bundled change, and discloses the bundled `.pi/skills/*` tooling changes. Labels `Bug / New feature / SQL / Performance` are each justified.

## Critical
**None.** No confirmed correctness, concurrency, resource, or performance/IO defect on any reachable path.

Verified end-to-end:
- **Null/sentinel semantics** correct for every type×width×direction. `Numbers.lessThan(int/long,·,negated)` and the tolerance double form `negated ? (l>v || equals) : (l<v && !equals)` are byte-equivalent to the existing `<(DD)` `LtDoubleVVFunctionFactory`. INT_NULL/LONG_NULL left operands widened to the double path read as `NaN`, never leaking `MIN_VALUE`.
- **Width routing never narrows a value** (the only widen-to-double path is the documented >2^53 tradeoff, consistent with existing double-vs-long behavior).
- **`offerStateTo` state donation** publishes the cached scalar safely — owner writes in `init()` before the page-frame reduce queue dispatches; workers read after the volatile sequence consume (same happens-before the code already relies on for `AsyncFilterAtom.memoryTracker` and the timestamp-cursor precedent). Sub-query executes exactly once per execution, re-donated fresh on every re-execution.
- **`AbstractRecordCursorFactory.close()`** idempotence is unbypassable (`final`), covers all 105 subclasses, flag-before-`_close()` is correct, and every touched `_close()` is best-effort.
- **Multi-row rejection** copies the first-row value to a stable primitive *before* `assertNoMoreRows` advances the cursor — no flyweight hazard.
- **Timestamp wrap-around fixes** (`ts > MAX`, `lo==MIN`/`MAX`, `adjustment>0 && dynamicValue==MAX`) correct for negated/non-negated, micros and nanos; the `<` side needs no symmetric guard because `MIN==LONG_NULL` is caught earlier and negative literals tokenize as OPERATION nodes.
- **`ColumnType` NULL+CURSOR = OVERLOAD_NONE** is surgical (only a bare `null` literal loses cursor binding) — no regression to `in(cursor)`, timestamp cursor eq/lt/gt, or EXISTS.

## Moderate
**None confirmed.** One pre-existing, off-trail latent risk recorded below; it does not block this PR.

## Minor (all cosmetic or unreachable; none touches behavior/resources on a reachable path)
1. **`SampleByFirstLastRecordCursorFactory._close`** — the two `DirectLongList` frees use plain `Misc.free(...)` mid-chain instead of `freeBestEffort`; if either threw it would skip the temporal-function frees below. Verified **unreachable** (`DirectLongList.close()` is a non-throwing, idempotent `Unsafe.free`). Fold into the `failure` chain for consistency.
2. **`AbstractScalarCursorFunction`** boolean fields `stateInherited`/`stateShared` miss the `is`/`has` prefix → `isStateInherited`/`isStateShared`.
3. **`PerWorkerFunctionList`** places static helpers after instance methods (house style is static-first); license banner is `/*+****` (stray `+`) unlike sibling files.
4. **`Misc` new overloads** have inconsistent primary-arg position — `free(obj, primary)` vs `freeBestEffort(primary, obj)` — a transposition footgun worth aligning. Also a latent footgun: `freeObjList(list, primary)` calls `setQuick`, which `PerWorkerFunctionList` rejects (would mask the primary); no current caller passes one, but an `assert !(list instanceof PerWorkerFunctionList)` would fail fast under `-ea`.
5. **Horizon/multi-horizon catch blocks (pre-existing, byte-equivalent to master)** — free `groupByFunctions`+`keyFunctions` but not the outer/inner COLUMN projection slots. Verified those are **resourceless column-reader flyweights** (no native memory, no-op `close()`) — no meaningful leak. For uniformity, align with the group-by path's `GroupByUtils.freeAssembledProjectionFunctions(outer, inner, th)`.
6. **Missing `Core` label** — the PR reworks `Misc`, `CairoException`, and every-factory `close()` idempotency.

## Needs discussion (pre-existing, NOT introduced by this PR)
- **`RuntimeIntervalModelBuilder.mergeWithAddMethod` drops the dynamic-interval suffix** (merges only the static `[lo,hi]` prefix, ignoring `dynamicRangeList`), so a dynamic timestamp bound (`$1` or `ts =/</> (select …)`) reaching the `and_offset` calendar-offset pushdown could produce an over-broad interval. **Verified byte-identical to master** and outside the PR's modification trail — `analyzeAndOffset`, `mergeWithAddMethod`, and the `SqlOptimiser` pushdown are untouched, and this PR does not change whether `ts < (subquery)` produces a dynamic interval. Reachability is narrow (`hasTimestampOffset()` virtual-timestamp column + dynamic bound) and unproven. Recommend a follow-up confirmatory test; it does not block this PR.

## Downgraded (false positives after source verification)
- **"Moderate resource leak of outer projection functions"** → the unfreed entries are resourceless flyweights with no-op `close()`; byte-equivalent to master (pre-PR shared field was cleared-not-freed on error).
- **Non-volatile `closed` flag / de-finalized `PageFrameSequence.atom`,`frameAddressCache`** → not defects (single-threaded factory close; async fields published/nulled under the reduce-queue fence only after the sequence is awaited).
- All cross-context "NEEDS-VERIFICATION" callsites (idempotent close on 105 factories, the 7 position-gaining `IntrinsicModel` methods, the `TimestampCursorFunc` ctor `rightPos`, `Misc` overload callers, the overload matrix, de-finalized fields) → resolved **SAFE**.

## Test coverage
Every behavioral change maps to a deterministic test whose assertion flips if the change is reverted: numeric `<`/`>`/`<=`/`>=` across all widths, LONG-not-narrowed-to-INT, 2^53 precision, null/empty cursor + null==null convention, multi-row rejection (numeric + designated-ts + interval/`!=`/OR/NS), `col <= null` (matrix + e2e), timestamp wrap-around, once-per-query state donation across async filter / keyed+notkeyed parallel GBY / window+horizon+multi-horizon joins / lateral / serial (deterministic counter ==1, refresh ==2), idempotent close, best-effort cleanup with primary/suppressed, ownership rollback, plan base-metadata restore, tolerance/special-value double semantics (JIT+interpreted). Modes covered where relevant: WAL/non-WAL, JIT/interpreted, parallel/single, micros/nanos. Strict builder-API enforcement is **clean** (no `assertSql`/`.returnsOnce`/lazy `assertQuery`/vacuous asserts/`Thread.sleep` coordination).

## Verdict: **approve**
Large but disciplined PR; the feature is correct and the supporting refactor (idempotent close, best-effort cleanup, worker state donation, interval ownership) is internally consistent and well-tested.
- **Correctness & performance gate: PASS** — zero confirmed reachable defects.
- **Test gate: PASS** — no UNTESTED Critical rows; every fix ships a regression test with a verified failure-on-revert link.
- **Verification:** ~16 draft findings/notes → after source verification **0 Critical, 0 confirmed Moderate, ~6 Minor**, 1 false positive downgraded, 1 pre-existing off-trail latent risk noted. Cross-context caller impact of every changed signature verified SAFE (no out-of-diff breakage).

Suggested non-blocking tidy-ups: fold the `SampleByFirstLast._close` native frees into the best-effort chain, rename `isStateInherited`/`isStateShared`, fix the `/*+` banner + static-first ordering in `PerWorkerFunctionList`, and add the `Core` label.

<sub>🤖 Adversarial mission-critical review (level 3).</sub>

bluestreak01:
# Review — `feat(sql): compare a numeric column with a scalar sub-query in predicates`

**Level 3** mission-critical review (adversarial, blocking). Independent pass: deep read of the core feature, 8 fresh-context reviewer agents (correctness/interval, concurrency, resources, compiler, fresh-adversarial, adversarial-performance, cross-context, tests), self-verification of the trickiest logic, and a local compile + targeted test run.

**Scope:** net diff vs true merge-base `origin/master` (`b0b8944bf2`) = **133 files, +18,192 / −1,858** (100 main, 31 test, `function_list.txt`, 2 `.pi/skills/*`). No `java-questdb-client` submodule change (the delta seen against a stale local `master` is not in the PR — confirmed via `gh pr diff`).

## Title & description
Compliant. Conventional-Commits, verb-leading, reads as a release note. Body is level-headed, with a genuine **Tradeoffs and limitations** section (>2^53 precision loss, cursor-must-be-on-right, execution-scoped cache, no correlated sub-queries), lists each bundled change, and discloses the bundled `.pi/skills/*` tooling edits. Labels `Bug / New feature / SQL / Performance` are justified. Minor: no `Core` label despite reworking `Misc`, `CairoException`, and every-factory `close()`.

## Critical
**None.** No confirmed correctness, concurrency, resource, or performance/IO defect on any reachable path.

Verified end-to-end:
- **Comparison logic & null/sentinel semantics** correct for every type × width × direction. Int/long paths delegate to `Numbers.lessThan(a,b,negated)` and eq to `negated != (a==value)` / `negated != Numbers.equals(...)` — byte-identical to the existing column-to-column `LtDoubleVVFunctionFactory` / `EqLongFunctionFactory`. `INT_NULL`/`LONG_NULL` left operands widen null-safely (`IntFunction.getLong`→`intToLong`, `LongFunction.getDouble`→NaN, `intToDouble`→NaN); a sentinel never leaks as `MIN_VALUE`.
- **Width routing never narrows** (only widen-to-double is the documented >2^53 tradeoff).
- **Swap + negate derivation** (re-derived all 8 signatures): `<(DC)`, `>=(DC)`, `>(CD)`, `<=(CD)` from Lt and the symmetric four from Gt map to the algebraically correct comparison; Eq covers `=`/`!=`/`<>` + symmetric swap. No duplicate registrations.
- **`offerStateTo` state donation** publication-safe: owner writes the cached scalar in `init()` on the main thread as the last step of `PageFrameSequence.of()`; workers read only after the reduce-queue `Sequence` publish/consume fence. Sub-query executes exactly once per execution, re-donated fresh each re-execution.
- **`AbstractRecordCursorFactory.close()`** idempotence is unbypassable (`final`, flag set before `_close()`); the `factory` is closed once via `rightFunc`, not double-freed.
- **Multi-row rejection** copies the first-row scalar to a stable primitive **before** `assertNoMoreRows` advances the cursor — no flyweight hazard — on micros/nanos timestamp and numeric paths.
- **`NULL + CURSOR = OVERLOAD_NONE`** is surgical: a genuine cursor arg has type `CURSOR`, not `NULL`, so `in(cursor)`, EXISTS, timestamp cursor eq/lt/gt, and `col <op> null` scalar resolution are unaffected.
- **Resource cleanup** best-effort and single-owner-exact across the GROUP BY / SAMPLE BY / async-atom rework; construction rollback frees adopted resources exactly once.

## Moderate
1. **`AbstractScalarCursorFunction.stateInherited` is never reset to `false`.** Correctness depends on the implicit invariant that every re-execution re-donates the owner's fresh value before the clone's early-returning `init()`. *Behaviorally inert on every reachable path today — no current atom initialises a per-worker scalar clone without first calling `offerStateTo(owner→clone)`, so no stale read is possible.* Latent footgun for a future atom; a one-line comment on the field or a guard would harden it. Non-blocking.

## Minor (cosmetic / defense-in-depth — none touches behavior on a reachable path)
1. **`stateInherited` / `stateShared`** miss the `is`/`has` prefix → `isStateInherited` / `isStateShared`.
2. **`SampleByFirstLastRecordCursorFactory._close`** frees the two `DirectLongList`s with plain `Misc.free(...)` mid-chain instead of `freeBestEffort`. Verified **unreachable** (`DirectLongList.close()` is a non-throwing `Unsafe.free`) — fold into the best-effort chain for uniformity.
3. **`Misc` new overloads** transpose the primary-arg position between families (`free(obj, primary)` vs `freeBestEffort(primary, obj)`). A readability footgun, but **compile-enforced safe** (no type is both `Closeable` and `Throwable`).
4. **`offerStateTo`** uses `getClass()==that.getClass()` (new code) vs `instanceof` (retrofitted timestamp code) — both safe; could converge.
5. **Interval-position defense** — `intersectRuntimeIntervals`/`subtractRuntimeIntervals` would append a `cursorFunctionPositions` entry for a CURSOR-typed function, but their evaluator branch never increments `cursorFunctionIndex`. Unreachable today; a guarding assert/comment would prevent a future index-shift.
6. **Missing `Core` label.**

## Downgraded (candidate findings dismissed after source verification)
- **"double-free via idempotent close in Virtual/Filtered/GROUP BY catch paths"** → safety depends on the conjunction of idempotent `close()` + slot-nulling helpers + generator nulling locals before adoption; all three hold (verified + `FactoryCloseContractTest`, `SampleBy*OwnershipTest`).
- **"`freeObjList(list, primary)` masks the primary when the list rejects `setQuick`"** → no caller passes a `PerWorkerFunctionList` there; per-worker lists always route through `PerWorkerFunctionList.close()`.
- **"`<` const-string path lacks the `>` sibling's wrap guard"** → the trigger value `Long.MIN_VALUE` cannot arrive as a `CONSTANT` (it is `LONG_NULL`, caught earlier). Defense-in-depth only.
- **"bare-`null` left operand overload is scan-order dependent"** → observationally harmless; a null left yields the same result on every typed path, and tests pin it.
- **de-finalized `PageFrameSequence.atom`/`frameAddressCache`, `rethrowCleanupFailure` checked branch, `MatView` override forwarding** → all verified SAFE.

## Coverage map (behavioral change → test → revert-sensitivity)
| Behavioral change | Test | Fails on revert | Verdict |
|---|---|---|---|
| `<`/`>`/`<=`/`>=`/`=`/`!=`/`<>` across DOUBLE/FLOAT/LONG/INT/SHORT/BYTE | `EqCursorFunctionFactoryTest`, `lt/{Double,Int,Long}CursorFunctionFactoryTest` | `.returns(...)` pins matched rows | TESTED |
| LONG not narrowed to INT | `EqCursor#testEqIntWithLongCursorDoesNotNarrow` | 5e9 narrowed would spuriously match | TESTED |
| exact equality beyond 2^53 | `EqCursor#testEqLongExactBeyondDoublePrecision` | 2^53 vs 2^53+1 distinct only in long | TESTED |
| null & empty cursor + null==null | `EqCursor` + `AbstractCursor` null-left assertions | strict excludes null, `>=`/`<=`/`=` match null-left | TESTED |
| multi-row rejection (numeric + designated-ts + interval/NS) | `EqCursor#testMultiRowCursorFails`, `Gt/LtTimestamp…`, `RuntimeIntervalModelTest#testMultiRowSubQueryError*` | `.fails(pos,"…more than one row")` | TESTED |
| multi-column / unsupported-type errors | `EqCursor#testError*` | exact `.fails(...)` messages | TESTED |
| `col <= null` matrix + e2e | `AbstractCursor#assertBareNullBehavior`, `EqCursor#testBareNullLiteralDoesNotBindCursor` | empty result, no compile throw | TESTED |
| timestamp wrap-around bounds | `GtTimestamp#testStrictBoundAtTypeExtremeMatchesNothing`, inclusive/Lt mirror | pre-fix wrapped → all rows | TESTED |
| once-per-query donation (async filter, keyed/notkeyed parallel GBY) | `AbstractCursor#assertWorkerStateSharedBehavior` (`assertEquals(1, COUNTER)`) | per-worker execution bumps counter | TESTED |
| worker clone/donation (join/window/lateral/serial) | `TestWorkerCloneFunctionFactoryTest`, `SqlCodeGeneratorWorkerFunctionExtractionTest` | clone/borrow/close bookkeeping | TESTED (unit-level; e2e counter only for filter/GBY) |
| idempotent close | `PerWorkerFunctionListTest`, `AbstractRecordCursorFactoryTest`, `MiscTest`, `RuntimeIntervalModelTest#testClose*` | double-close bumps count | TESTED |
| best-effort cleanup + suppressed exceptions | `MiscTest`, `CairoExceptionTest#testRethrowCleanupFailure*`, `AsyncFilterContextTest` | asserts primary identity + `getSuppressed()` | TESTED |
| ownership rollback | `SampleByTemporalFunctionOwnershipTest`, `SampleByFillFactoryConstructionFailureTest` | close-once + continued rollback | TESTED |
| plan base-metadata restore | `TextPlanSinkTest#test*MetadataModeRestoredAfterException` | `getUseBaseMetadata()` after throw | TESTED |
| MatView WHERE intrinsic override | `MatViewRefreshSqlExecutionContextTest` | base endpoints forwarded, foreign no-op | TESTED |

**Totals: 15 behavioral-change groups, 15 tested, 0 UNTESTED.** No banned builder-API usage (`assertSql`/`.returnsOnce`/lazy-`assertQuery`), no vacuous assertions, deterministic thread coordination (latches/barriers, worker `AssertionError` captured to `AtomicReference` and asserted on the main thread).

## Local validation
Compiled `core` and ran the focused `=` suite — **`EqCursorFunctionFactoryTest`: 21/21 green** — plus confirmed green surefire reports across the supporting suites: `DoubleCursorFunctionFactoryTest` 64, `Int/LongCursorFunctionFactoryTest` 18/18, `Eq/Gt/LtTimestampCursorFunctionFactoryTest` 24/19/18, `WhereClauseParserTest` 880, `RuntimeIntervalModelBuilderTest` 31, `RuntimeIntervalModelTest` 12, `FactoryCloseContractTest` 10, `PerWorkerFunctionListTest` 14, `SampleByTemporalFunctionOwnershipTest` 9, `SampleByFillFactoryConstructionFailureTest` 16, `AsyncFilteredRecordCursorFactoryTest` 29, `SqlCodeGeneratorWorkerFunctionExtractionTest` 11, `MiscTest` 4, `CairoExceptionTest` 10, `MatViewRefreshSqlExecutionContextTest` 1, `InSymbolCursorFunctionFactoryTest` 21 (regression). Full Maven suite not run locally — CI covers the platform matrix.

## Verdict: **approve**
Large but disciplined PR; the feature is correct and the supporting refactor (idempotent close, best-effort cleanup, worker state donation, interval ownership, recursive-compile scratch preservation) is internally consistent and thoroughly tested.
- **Correctness & performance gate: PASS** — zero confirmed reachable defects; the per-row hot path is actually *faster* than the generic VV baseline (one virtual `getDouble` vs two; `Numbers.equals` short-circuited on ordering-settled rows).
- **Test gate: PASS** — no UNTESTED Critical rows; every bundled fix ships a regression test with a verified failure-on-revert link.
- **Verification tally:** ~14 candidate findings/notes → after source verification **0 Critical, 1 Moderate (defense-in-depth), ~6 Minor**, remainder downgraded. Cross-context caller impact of every changed signature (`Misc` overloads, `IntrinsicModel` position params, `close()` idempotency, de-finalized async fields, `ColumnType` matrix) verified SAFE — no out-of-diff breakage.

Suggested non-blocking tidy-ups: rename `isStateInherited`/`isStateShared`, comment the `stateInherited` re-donation invariant, fold the `SampleByFirstLast._close` native frees into the best-effort chain, and add the `Core` label.

<sub>🤖 Adversarial mission-critical review (level 3).</sub>

bluestreak01:
## Review — level 3 (full mission-critical pass)

### Scope note
The GitHub/3-dot diff for this PR is misleading: the branch merged a `master` whose features (parquet2, covering index, QWP demote, monotonic-timestamp) later landed on real `master` as squash commits, so the merge-base diff shows ~518 files, most byte-identical to `master`. The authoritative PR scope is **227 files** (189 production Java, 35 test, 1 resource, 2 skill docs; 28 new production files), obtained via `git diff origin/master <branch>` restricted to files this PR's own 41 commits touched. All findings use that authoritative diff.

Every finding below was verified against source.

---

### Critical
**None.** No confirmed correctness, data-integrity, resource-safety, concurrency, or performance/IO defect on any reachable path was found — in-diff, out-of-diff, or pre-existing. This PR has clearly been through multiple review/fix rounds already.

### Moderate

1. **`FactoryCloseContractTest` is a whitebox/reflection liability** — `core/src/test/java/io/questdb/test/griffin/engine/FactoryCloseContractTest.java` (22 reflective sites). It bypasses constructors via `Unsafe.allocateInstance`, reflectively injects private fields, and references package-private cursor classes by stringly-typed `Class.forName("io.questdb.griffin.engine.table.AsyncFilteredRecordCursor")` (lines 118, 130, 282, 294, 562, 583, 597…). A rename/move of any referenced class fails at runtime (`ClassNotFoundException`), not compile time, and the test couples to each factory's private field layout so a correct refactor can break it. It is test-only and a deliberate way to unit-test the safety-critical double-free-prevention contract in isolation, hence Moderate not Critical. Suggested improvement for this PR: replace `Class.forName(...)` with direct `.class` references where the test can reach the package, so renames are compile-checked; keep reflective allocation only where a package boundary truly forbids direct reference.

### Minor

2. **Missing Apache license header** — `core/src/main/java/io/questdb/griffin/engine/table/AsyncHorizonJoinResources.java` starts directly with `package …`; every other new file in the PR carries the header, and there is no RAT/checkstyle in the poms to catch it. Add the header.

3. **Boolean names without `is/has` prefix** (CLAUDE.md) — `stateInherited`/`stateShared` (`AbstractScalarCursorFunction.java:59-60`, also in `EqTimestampCursorFunctionFactory` inner classes), `areFunctionsInitialized` (`GroupByNotKeyedRecordCursorFactory.java:152`, uses `are`), local `noDynamicIntervals`/`noIndex`. Consistent internal choices, no behavioral dimension; rename optional.

4. **Misleading error for a degenerate query** — `NULL = (SELECT <string/varchar/…>)` routes through `EqNullCursorFunctionFactory` → `TIMESTAMP_FACTORY` and reports `"left operand must be a TIMESTAMP, found: NULL"` (`EqTimestampCursorFunctionFactory.java:88`) rather than naming the unsupported cursor column type. Clean `SqlException` at the correct position, no crash — just confusing for a nonsensical query.

5. **PR metadata nits** — no `Fixes #NNN` (legitimate only if no tracked issue maps to the bundled NPE/leak/multi-row fixes — please verify); 23/41 commit titles carry Conventional-Commits prefixes and exceed 50 chars, which CLAUDE.md reserves for the squashed PR title (low impact — branch history is discarded at squash-merge); consider adding **Core**/**Documentation** labels given the cross-cutting cairo/griffin cleanup and skill-doc changes. Description tone, tradeoffs, and bundling disclosure are otherwise exemplary.

### Notes (verified-correct today; recorded for durability, not blocking)

- **`AbstractRecordCursorFactory.close()` sets `closed=true` before `_close()`** (`:59-63`). Deliberate and documented: prevents a double-free when a factory ownership chain closes the same instance from two owners on an error path (a real bug this PR fixes). The tradeoff — a `_close()` that throws partway leaks the not-yet-released resource because the retry is a no-op — is mitigated by the PR's detach-all-then-best-effort-close conversions. Single-close behavior is unchanged, so not a regression.
- **`GroupByUtils.freeAssembledProjectionFunctions`** relies on an assert-guarded positional invariant (`j == innerSize`); with assertions off, a future producer making `inner` longer than the non-null outer count would silently leak the tail. Correct today.

---

### Downgraded (investigated, dismissed after source verification)

- **Donation staleness / owner-init-before-donation** → FALSE POSITIVE. `Function.init(clones,…,prototype)` donates before each clone's `init()`, and every async atom (`AsyncFilterAtom:140→145`, `AsyncGroupByAtom`, Base/Multi horizon, window-join) inits the owner (sub-query executed) first.
- **`stateInherited` never reset** → FALSE POSITIVE. Owner re-executes every run; clones are re-donated a fresh value each execution before their `init()` early-returns. Pinned by `DoubleCursorFunctionFactoryTest` owner-init-failure/reuse tests.
- **`isThreadSafe()=leftFunc.isThreadSafe()` shared-value race** → FALSE POSITIVE. Cached `value`/`epoch` written only in single-threaded `init`, read-only in parallel `getBool`; safe-published by the frame-queue barrier.
- **`allIntervalsHitOnePartition` first/last shortcut** → FALSE POSITIVE. Guarded by `isStatic()`; static intervals are sorted/merged so `floor(first)==floor(last)` ⇒ all equal.
- **Interval-union encode/decode off-by-one** → FALSE POSITIVE. `appendStaticIntervalsIntersection` encodes the period-count in slot 0 + N null functions; the `case null`/`INTERSECT_INTERVALS` decode reads exactly N slots, advances `i`/`dynamicIndex` by N-1, and `intersectInPlace(divider)` atomically intersects the appended union with the preceding expression.
- **`=(oC)` overload ambiguity with typed `=(IC/LC/DC)`** → FALSE POSITIVE. `'o'`→`ColumnType.NULL` (`FunctionFactoryDescriptor.java:110`); matches only a bare NULL literal (exact, distance 0) and routes on the cursor column type.
- **`PerWorkerFunctionList` mutator misuse** → FALSE POSITIVE. Populated only via two-arg `add(fn,isOwned)` (`SqlCodeGenerator:1794-1830`); consumed via `getQuick` + ownership-aware static helpers; the `setQuick(i,null)` calls are on the outer plain `ObjList`.
- **Interval methods' new position param broke a caller** → FALSE POSITIVE. All callers (WhereClauseParser + mat-view bind-var path with position 0) pass it correctly.
- **Tolerance-based `Numbers.equals` in `<`/`>`** → FALSE POSITIVE. Identical formula/semantics to the existing `LtDoubleVVFunctionFactory` column-to-column operator; consistent, pre-existing design.
- **`SqlCodeGeneratorWorkerFunctionExtractionTest` relaxation made assertions vacuous** → DOWNGRADED to Note. Only the flag-read count became an upper bound; behavioral contracts stay hard `==`, and exactly-once is independently pinned by `DoubleCursorFunctionFactoryTest` `assertEquals(1, COUNTER.get())` after a deterministic drain barrier.

---

### Coverage map
A 24-row behavioral-change map was built and each claimed test verified by reading its assertion (not names); the highest-risk regression tests were verified directly. **All 24 behavioral changes are TESTED** with a real failure link — e.g. multi-row `.fails(28,"scalar sub-query returned more than one row")`, strict-bound `ts > Long.MAX_VALUE → count 0` (const/ns/cursor), re-entrancy `.withPlanContaining("Index forward scan").withPlanNotContaining("Async Filter")`, and worker-donation counter `==1` after a deterministic drain. Execution modes covered: JIT vs interpreted, parallel vs serial, partitioned + non-partitioned, TIMESTAMP + TIMESTAMP_NS. No SQL builder-API violations (zero new `assertSql`/`getPlan`/`.returnsOnce`; `assertMemoryLeak` wrappers all hold multiple statements).

Coverage-map totals: **24 behavioral changes, 24 tested, 0 UNTESTED.**

---

### Summary

**Verdict: approve** — with the Minor nits worth a quick follow-up commit on this branch (license header; swap `FactoryCloseContractTest`'s `Class.forName` strings for compile-checked references).

- **Correctness & performance gate: PASS.** Zero correctness/data-integrity/resource/concurrency/performance findings. The per-row hot path is allocation-free and optimal (primitive short-circuit before `Numbers.equals`, `Numbers.lessThan` for integers), the sub-query executes exactly once per query (not per worker/frame/row), and one pre-existing O(n) partition check became O(1). NULL/sentinel handling is correct and consistent with the existing column-to-column operators.
- **Test gate: PASS.** No UNTESTED Critical rows; every bundled bug fix has a verified regression test whose assertion fails if the fix is reverted (designated-timestamp multi-row, strict-bound-past-domain, and null-literal-binding verified directly, each documenting the pre-fix bug).
- **Findings:** 0 Critical, 1 Moderate (test brittleness), 4 Minor, 2 Notes. ~9 findings verified, ~11 candidate concerns dropped as false positives after source verification.
- **In-diff / out-of-diff / pre-existing:** all standing findings are in-diff or in changed code; **0 out-of-diff BROKEN callsites** — justified (changed interval-method signatures were propagated to all callers within the PR; `close()` idempotency is a backward-compatible strengthening; `PerWorkerFunctionList` is consumed only via ownership-aware helpers).
- **Tradeoffs to keep visible:** `long`-vs-`float/double` cursor equality loses precision beyond 2^53 (documented; `long`-vs-`long` stays exact); the `close()`-idempotency design trades a possible leak-on-throwing-`_close()` for double-free prevention (net win, mitigated). Both disclosed in the PR body.

Unusually clean, well-tested, and well-documented for its size.

<sub>Level-3 review. 8 review agents completed (correctness, concurrency, performance, resources, tests/coverage, code-quality, metadata, test-efficacy); the cross-context, adversarial, test-quality, and regression-efficacy passes were completed manually with source-level verification after account rate limits interrupted those agents. Their manual coverage found nothing beyond the above.</sub>

nwoolmer (approved):
LGTM

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 3693 / 4010 (92.09%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncMultiHorizonJoinNotKeyedAtom.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncHorizonJoinNotKeyedAtom.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/eq/EqDoubleCursorFunction.java|6|12|50.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AbstractTreeSetRecordCursorFactory.java|9|16|56.25%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/LtIntCursorFunction.java|7|12|58.33%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/LtLongCursorFunction.java|7|12|58.33%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/PgClassFunctionFactory.java|8|13|61.54%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortKeyMaterializingRecordCursorFactory.java|8|13|61.54%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/PgAttrDefFunctionFactory.java|8|13|61.54%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncMultiHorizonJoinAtom.java|11|17|64.71%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncHorizonJoinAtom.java|11|17|64.71%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/CopyCancelFactory.java|10|15|66.67%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/eq/EqLongCursorFunction.java|8|12|66.67%|
|:large_blue_circle:|io/questdb/cairo/sql/async/PageFrameSequence.java|45|60|75.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/GtIntCursorFunction.java|9|12|75.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/eq/EqLongCursorFunctionFactory.java|3|4|75.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/GtLongCursorFunction.java|9|12|75.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/eq/EqIntCursorFunction.java|9|12|75.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/LtDoubleCursorFunction.java|10|13|76.92%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncGroupByNotKeyedAtom.java|16|21|76.19%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SortedSymbolIndexRecordCursorFactory.java|7|9|77.78%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinDenseRecordCursorFactory.java|7|9|77.78%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/AbstractSampleByFillRecordCursorFactory.java|21|27|77.78%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByAllSymbolsFilteredRecordCursorFactory.java|7|9|77.78%|
|:large_blue_circle:|io/questdb/griffin/engine/table/DeferredSingleSymbolFilterPageFrameRecordCursorFactory.java|7|9|77.78%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AbstractDeferredTreeSetRecordCursorFactory.java|7|9|77.78%|
|:large_blue_circle:|io/questdb/griffin/engine/table/BaseAsyncHorizonJoinAtom.java|34|44|77.27%|
|:large_blue_circle:|io/questdb/griffin/engine/table/BaseAsyncMultiHorizonJoinAtom.java|36|46|78.26%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/test/TestTableReferenceOutOfDateFunctionFactory.java|8|10|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/vect/GroupByRecordCursorFactory.java|20|25|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowCreateTableRecordCursorFactory.java|8|10|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowCreateDatabaseRecordCursorFactory.java|8|10|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowCreateViewRecordCursorFactory.java|8|10|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowCreateMatViewRecordCursorFactory.java|8|10|80.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinAtom.java|27|33|81.82%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByValuesIndexedFilteredRecordCursorFactory.java|10|12|83.33%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByAllIndexedRecordCursorFactory.java|10|12|83.33%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByValueFilteredRecordCursorFactory.java|10|12|83.33%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByAllFilteredRecordCursorFactory.java|10|12|83.33%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByValueIndexedFilteredRecordCursorFactory.java|10|12|83.33%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncGroupByAtom.java|28|33|84.85%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/AbstractLongCursorFunctionFactory.java|12|14|85.71%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestBySubQueryRecordCursorFactory.java|13|15|86.67%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByInterpolateRecordCursorFactory.java|40|46|86.96%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/AbstractIntCursorFunctionFactory.java|13|15|86.67%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AbstractDeferredValueRecordCursorFactory.java|14|16|87.50%|
|:large_blue_circle:|io/questdb/cairo/sql/async/UnorderedPageFrameSequence.java|34|39|87.18%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncJitFilteredRecordCursorFactory.java|36|41|87.80%|
|:large_blue_circle:|io/questdb/griffin/model/RuntimeIntervalModelBuilder.java|301|344|87.50%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FilterOnExcludedValuesRecordCursorFactory.java|16|18|88.89%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FilterOnSubQueryRecordCursorFactory.java|16|18|88.89%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FilterOnValuesRecordCursorFactory.java|16|18|88.89%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByDeferredListValuesFilteredRecordCursorFactory.java|16|18|88.89%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/test/TestWorkerCloneFunctionFactory.java|67|76|88.16%|
|:large_blue_circle:|io/questdb/griffin/model/IntrinsicModel.java|10|11|90.91%|
|:large_blue_circle:|io/questdb/griffin/model/RuntimeIntervalModel.java|99|110|90.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/AbstractSampleByRecordCursorFactory.java|27|30|90.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/GroupByUtils.java|44|48|91.67%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncHorizonJoinResources.java|61|66|92.42%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/GtDoubleCursorFunction.java|12|13|92.31%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/GroupByRecordCursorFactory.java|24|26|92.31%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFillRecordCursorFactory.java|34|36|94.44%|
|:large_blue_circle:|io/questdb/griffin/engine/table/MultiHorizonJoinNotKeyedRecordCursorFactory.java|19|20|95.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/HorizonJoinNotKeyedRecordCursorFactory.java|20|21|95.24%|
|:large_blue_circle:|io/questdb/griffin/engine/table/MultiHorizonJoinRecordCursorFactory.java|19|20|95.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|453|472|95.97%|
|:large_blue_circle:|io/questdb/griffin/engine/table/HorizonJoinRecordCursorFactory.java|20|21|95.24%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/test/TestCloseCounterFunctionFactory.java|55|57|96.49%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncHorizonJoinRecordCursorFactory.java|27|28|96.43%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/GroupByNotKeyedRecordCursorFactory.java|32|33|96.97%|
|:large_blue_circle:|io/questdb/griffin/WhereClauseParser.java|311|320|97.19%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncMultiHorizonJoinRecordCursorFactory.java|36|37|97.30%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/AbstractScalarCursorFunction.java|36|37|97.30%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinLightNoKeyRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinDenseRecordCursorFactoryBase.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncFilterContext.java|26|26|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/WindowJoinRecordCursorFactory.java|31|31|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinFastRecordCursorFactory.java|17|17|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/TableStorageRecordCursorFactory.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFillNoneRecordCursorFactory.java|22|22|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/HashOuterJoinFilteredRecordCursorFactory.java|21|21|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/NestedLoopLeftJoinRecordCursorFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/AbstractSampleByNotKeyedRecordCursorFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/std/Misc.java|38|38|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFillValueNotKeyedRecordCursorFactory.java|17|17|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/union/ExceptAllRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByRecordCursorFactory.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/NestedLoopRightJoinRecordCursorFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncTopKRecordCursorFactory.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/GtDoubleCursorFunctionFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/ScalarSubQueryUtils.java|28|28|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortedLightRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/MarkoutHorizonRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/HashOuterJoinLightRecordCursorFactory.java|14|14|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/QueryProgress.java|14|14|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/BasePlanSink.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/LtJoinNoKeyFastRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FilteredRecordCursorFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/ShowPartitionsRecordCursorFactory.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFirstLastRecordCursorFactory.java|33|33|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/GtIntCursorFunctionFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/union/IntersectRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/ReadParquetRecordCursorFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/GtTimestampCursorFunctionFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/FilteredAsOfJoinNoKeyFastRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/SpliceJoinLightRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/eq/EqIntCursorFunctionFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/PageFrameRecordCursorFactory.java|25|25|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFillPrevNotKeyedRecordCursorFactory.java|13|13|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncMultiHorizonJoinRecordCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinIndexedRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/AbstractDoubleCursorFunctionFactory.java|13|13|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/HorizonJoinSlaveState.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/FilesFunctionFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/union/AbstractSetRecordCursorFactory.java|20|20|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AbstractJoinRecordCursorFactory.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlOptimiser.java|16|16|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncFilteredRecordCursorFactory.java|38|38|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/ExportFilesFunctionFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/NestedLoopFullJoinRecordCursorFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/window/CachedWindowLightRecordCursorFactory.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/LtJoinLightRecordCursorFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFillNullRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/cairo/AbstractRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/DistinctTimeSeriesRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsyncWindowJoinRecordCursorFactory.java|17|17|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncHorizonJoinRecordCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/WindowJoinFastRecordCursorFactory.java|23|23|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncGroupByNotKeyedRecordCursorFactory.java|27|27|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/LtLongCursorFunctionFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/eq/EqDoubleCursorFunctionFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/eq/EqNullCursorFunctionFactory.java|16|16|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/union/ExceptRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFillNullNotKeyedRecordCursorFactory.java|17|17|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/VirtualRecordCursorFactory.java|27|27|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortLightRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LongTopKRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFillPrevRecordCursorFactory.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/EncodedSortLimitedLightRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/eq/EqTimestampCursorFunctionFactory.java|13|13|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/table/ReadParquetPageFrameRecordCursorFactory.java|15|15|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinNoKeyFastRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncFilterAtom.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/ColumnType.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/DistinctRecordCursorFactory.java|14|14|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/union/IntersectAllRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/LimitedSizeSortedLightRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/union/UnionRecordCursorFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/CrossJoinRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/window/WindowRecordCursorFactory.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/LatestByLightRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/LtTimestampCursorFunctionFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/UnnestRecordCursorFactory.java|15|15|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncMultiHorizonJoinNotKeyedRecordCursorFactory.java|37|37|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/LtJoinNoKeyRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFillNoneNotKeyedRecordCursorFactory.java|13|13|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinFastRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncGroupByRecordCursorFactory.java|19|19|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AbstractPageFrameRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/PerWorkerFunctionList.java|71|71|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/test/TestFaultFunctionFactory.java|64|64|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/LtJoinRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/orderby/SortedRecordCursorFactory.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncGroupByRecordCursor.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/HashJoinRecordCursorFactory.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/LtIntCursorFunctionFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/FilteredAsOfJoinFastRecordCursorFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/window/CachedWindowRecordCursorFactory.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/AsyncHorizonJoinNotKeyedRecordCursorFactory.java|28|28|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/catalogue/ImportFilesFunctionFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/union/UnionAllRecordCursor.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/GtLongCursorFunctionFactory.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshSqlExecutionContext.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/HashOuterJoinFilteredLightRecordCursorFactory.java|21|21|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/HashOuterJoinRecordCursorFactory.java|14|14|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinMemoizedRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/HashJoinLightRecordCursorFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/groupby/SampleByFillValueRecordCursorFactory.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/lt/LtDoubleCursorFunctionFactory.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoException.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/join/AsOfJoinLightRecordCursorFactory.java|8|8|100.00%|

PR #7380, opened 2026-07-10
fix(core): fix table suspension after a range replace removes its last partition

puzpuzpuz:
## Problem

A `REPLACE_RANGE` WAL commit — used by materialized-view refreshes and range-replace inserts — that removes the table's last partition moves the writer's `maxTimestamp` back to the new last partition, but leaves `partitionTimestampHi` pointing at the removed partition. In the replace path `partitionTimestampHi` is only ever advanced via `Math.max`, never lowered.

`finishO3Commit` normally re-syncs `partitionTimestampHi` by re-opening the new last partition, but that re-open is skipped when the new last partition is **parquet**. The next non-replace commit applied by the same `TableWriter`
instance then trips the partition-timestamp consistency assert at the top of `processWalCommit` (`TableWriter.java:9838`), which suspends the table under `-ea`.

Because the trigger is WAL content rather than timing, it reproduces deterministically once the new last partition is parquet.

## Fix

In the partition-removal branch of `o3ConsumePartitionUpdateSink`, re-derive `partitionTimestampHi` from the final `maxTimestamp` after the partition sizes are updated. This covers the parquet last-partition case that the `finishO3Commit` re-open path misses. A native last partition already received the same value via the re-open, so its behaviour is unchanged, and the fix runs only for replace commits that removed or trimmed partitions.

## Test

`WalWriterReplaceRangeTest.testRemovesLastPartitionThenNormalCommitDoesNotSuspend` builds the scenario deterministically: three day partitions, the soon-to-be-last one converted to parquet, a no-rows `REPLACE_RANGE` that drops the last partition, then a plain insert — both applied by the same writer. It is red without the fix (table suspended at the assert) and green with it, and also checks the resulting row count / min / max.

## Test plan

- New regression test is red without the fix, green with it
- `WalWriterReplaceRangeTest` (59), `MatViewIdenticalReplaceTest` (20), `MatViewFuzzTest` (22), `CreateMatViewTest` (102) and `MatViewTest` (202) pass

Found via a flaky WAL replication fuzz test (questdb/questdb-enterprise#1123).


puzpuzpuz:
## Deep review (level 3)

No submodule pointer moved in this diff, so scope is the two changed files. PR title/body/labels follow the repo conventions.

**0 Critical, 0 Moderate, 1 Minor.**

> Edited: an earlier revision of this comment listed an adjacent pre-existing finding at `TableWriter.java:8944` and suggested a fix for it. That reading was wrong and the suggested fix would have introduced a bug. See [Retracted](#retracted) at the bottom.

---

### Minor

**Problem:** the test's parquet precondition is only guarded indirectly.
**Net impact:** none today; risks silent coverage loss on a future change.

`core/src/test/java/io/questdb/test/cairo/wal/WalWriterReplaceRangeTest.java:200` converts `2022-02-25` to parquet, and the whole scenario depends on that partition being parquet when commit A lands — that is what makes `finishO3Commit` skip the re-open. Nothing observes it.

Today a failed conversion cannot slip through silently: `convertPartitionNativeToParquet` throws on a WAL table for a missing/active partition, `AlterOperation` turns a `false` return into a `CairoException`, and the apply job then suspends the table, which `Assert.assertFalse("table is suspended", …)` at line 218 catches. The residual risk is a future change that makes the conversion a silent no-op — the test would quietly degrade into the native case already covered by `testRemovesLastPartitionNoRowsAdded` and stop exercising the fix. One line using the file-adjacent idiom closes it:

```java
assertQuery("select name, isParquet from table_partitions('rg')")
        .noLeakCheck()
        .returns("name\tisParquet\n2022-02-24\tfalse\n2022-02-25\ttrue\n2022-02-26\tfalse\n");
```

placed before commit A. It also pins the removal claim the comment at line 220 makes but nothing asserts. Non-blocking.

---

### Coverage

**Test gate passes; 0 admitted coverage gaps.**

The regression test was verified by execution, not reasoning:

- HEAD `39cada6d35`: `WalWriterReplaceRangeTest` 61/61 green.
- HEAD tests + `git checkout c47ec377d4 -- core/src/main/java/io/questdb/cairo/TableWriter.java` in a scratch worktree: 61 run, **1 failure** — `testRemovesLastPartitionThenNormalCommitDoesNotSuspend`, `AssertionError: table is suspended` at line 218.

No other test in the class changes verdict, which is the executed form of the "native behaviour unchanged" claim in the PR body.

One reachability narrowing is worth adding to the PR description, because it is why a single test row is sufficient: **the parquet-survivor case is necessarily a pure trailing delete, so the shape the new test builds is the only shape in which the fix changes anything.** A replace commit that touches a parquet partition throws at `TableWriter.java:10271-10280`, so a parquet surviving last partition forces the replace range entirely above it; any row the replace added would land in that range and keep a partition alive above the survivor, so zero rows are added; zero rows leaves `replaceMaxTimestamp == Long.MIN_VALUE`, which is exactly why the `:10332` branch never runs and base leaves the field at the removed partition's ceiling.

In every other shape the new value is byte-identical to base:

- `lastPartitionRemoved == false` (middle/first-only removal): the max is either untouched or `replaceMaxTimestamp`, which `:10160-10167` only sets while processing a partition `>= lastPartitionTimestamp` carrying O3 rows — the highest post-commit partition. Same ceiling.
- last partition **trimmed**: `:8828-8839` sets the flag, `:9050-9057` re-reads the same partition's max, ceiling unchanged. A trimmed partition is itself the survivor, so it cannot be parquet.
- native survivor: `closeActivePartition(true)` at `:8845` has run, so `isLastPartitionClosed()` satisfies `:7368` and `openPartition` (`:9584`) installs the identical value in both revisions.
- split partitions: `getCurrentPartitionMaxTimestamp` **is** split-aware — `getNextPartitionTimestamp` (`TxReader.java:379-392`) returns the next *attached* partition's start when it shares the calendar floor — so the value equals what `openPartition` computes for the same partition.
- non-WAL: unreachable. `dedupMode = WAL_DEDUP_MODE_REPLACE_RANGE` is assigned only at `:11513` inside `processWalCommitDedupReplace`.

The fix also mirrors the established idiom: `TableWriter.java:7161-7167` already does exactly this in the DROP PARTITION path when the new last partition is parquet and `openPartition` is skipped.

Accepted, non-blocking, outside the gate: no HOUR/MONTH/YEAR/WEEK variant — the expression is unit-generic via `partitionCeilMethod` and is the codebase idiom at four other sites; no producer-level mat-view/live-view test for this shape — same storage-level mechanism, and reaching the parquet variant additionally needs an operator `CONVERT PARTITION TO PARQUET`.

Suites run green at head, sequentially, after a clean `mvn -pl core clean test-compile`: `WalWriterReplaceRangeTest` 61, `WalWriterTest` 101, `MatViewIdenticalReplaceTest` 20, `LiveViewBaseReplaceRangeTest` 7, `LiveViewStartFromReplayTest` 8, `LiveViewCheckpointBoundaryFixtureTest` 22, `AlterTableConvertPartitionTest` 40, `O3SplitPartitionTest` 13, `O3ParquetPartitionFuzzTest` 1.

---

<a name="retracted"></a>
### Retracted: the `removedIsLast` observation at `TableWriter.java:8944`

An earlier revision of this comment reported `boolean removedIsLast = partIndex == txWriter.getPartitionCount()` as a dead comparison on the `isSplitPartition` arm and suggested comparing against `getPartitionCount() - 1` instead. **Do not apply that.** The comparison is structurally false on that arm, but that is load-bearing rather than accidental, and the suggested change would have been a real bug.

On the split arm the partition is not removed — `:8939` only zeroes its size, and `o3ConsumePartitionUpdateSink_processSplitPartitionRemoval()` handles it later. That later pass can shrink the parent (`:9195`) and insert a *new* split off it (`:9198`), so the post-commit last partition is the new split, not `partIndex - 1`. Making `removedIsLast` true on the split arm would therefore run `columnVersionWriter.replaceInitialPartitionRecords` and `o3ConsumePartitionUpdateSink_trimPartitionColumnTops` (`:8952`, `:8957`) against the wrong partition, with a size that is about to change.

The max-timestamp recompute the split case does need still happens, via the other flag path: `:8828-8839` fires for the emptied split, because `partitionTimestamp` equals the split's own timestamp and `lastPartitionTimestamp` at `:11506` is split-aware. `:9050-9057` then reads `partitionCount - 1`, which is the newly inserted split. The current code is correct.

What remains is readability only — the condition reads as though it covers both arms. Nothing to fix here, and no issue filed.

---

### Verdict: approve

- **Correctness gate passes.** The correctness, cross-context caller, and fresh-context adversarial passes each returned zero candidates. Every hypothesis raised was falsified against source: lowered value defeating `finishO3Commit`'s re-open, stale `maxTimestamp` on the non-`lastPartitionRemoved` paths, split-partition ceiling divergence, `Long.MIN_VALUE` / open-range overflow, non-WAL reachability, and stale `lastPartitionTimestamp` in the window before the next commit. Callsite analysis covered every read of `partitionTimestampHi` (`:2971`, `:4474`, `:4698`, `:4745`, `:7368`, `:8521`/`:8524`, `:10480`/`:10482`, `:11533`/`:11567`/`:11579`, `:13363-13391`) and both cached locals (`initialPartitionTimestampHi`, `partitionTimestampHiLimit`).
- **Test gate passes.**
- No in-diff findings, no out-of-diff breakage.

**Tradeoff worth naming:** the fix makes the `:10479-10480` assert self-consistent by construction for states the replace path produces, since `partitionTimestampHi` is now *derived* from `txWriter.maxTimestamp`. That assert can therefore no longer act as a tripwire for a future stale-`maxTimestamp` regression in `o3ConsumePartitionUpdateSink`. This could not be turned into a behaviour delta — in the one stale-max scenario constructed, both revisions compute the same value and both asserts pass — so it is not a finding, but it is a real loss of an existing tripwire.

ideoma:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|

bluestreak01 (approved):
Reviewing PR #7380 at level 3.

### Critical

None.

### Moderate

None.

### Minor

Problem: Added test uses discouraged CAST syntax.  
Net impact: No runtime impact; test-source consistency only.  
Evidence: Line 194 violates CLAUDE.md at head `39cada6d`.

`core/src/test/java/io/questdb/test/cairo/wal/WalWriterReplaceRangeTest.java:194` adds `cast(x as string)`, while `CLAUDE.md:105-106` requires preferring `expr::TYPE`. Existing neighboring uses are unchanged and do not exempt the new line. Use `x::STRING`.

### Coverage map

Test gate passes; **0 admitted coverage gaps**.

- Focused regression: green at head; red with only `TableWriter.java` restored to the base blob, suspending the table at seqTxn 4.
- Full `WalWriterReplaceRangeTest`: 61 tests passed.
- Required CI, enterprise, coverage, macOS, and Windows checks pass.

### Summary

- **Verdict: approve**
- Correctness gate: passes.
- Test gate: passes.
- Severity: 0 Critical, 0 Moderate, 1 Minor.
- Admitted split: 1 in-diff cosmetic item, 0 out-of-diff breakages.
- Submodule provenance: no submodule pointers changed.
- Metadata: title, body, and labels follow repository conventions.
- Tradeoff: one allocation-free O(log partition-count) lookup per qualifying range-replace commit; no row-scaled cost.

PR #7382, opened 2026-07-10
fix(core): fix a POSTING index crash on rollback after a value-file I/O error

puzpuzpuz:
## What

A POSTING index writer can retain live key or pending state after an I/O failure
closes its mapped value file (`.pv`). A later rollback can then enter a flush or
reencode path that dereferences the closed mapping. Depending on the path and
whether assertions are enabled, this produces an assertion failure, a null
facade access, or a native-memory crash.

This PR rejects those invalid operations with a `CairoException` and clears the
mapped-memory append bounds when `MemoryCMARWImpl` closes. The table writer
continues to distress and discard the affected writer; recovery still occurs on
the next open.

## Mechanism

Two fault states are covered:

- During `PostingIndexWriter.flushAllPending()`, growing `.pv` calls
  `MemoryCMARWImpl.extend0()`. If `mremap` fails, `extend0()` calls
  `close(false)` and rethrows. The value memory is now unmapped, while the
  pending batch and old `valueMemSize` remain. A subsequent
  `rollbackValues()` re-enters `flushAllPending()` and previously attempted to
  append through the closed memory.
- `PostingIndexWriter.truncate()` closes the current `.pv` before opening its
  replacement. If that open fails, the writer retains existing generation
  metadata while `valueMem` remains closed. A later `rollbackValues()` can
  reach `reencodeAllGenerations()`, where generation zero previously resolved
  to native address zero and could terminate the JVM with `SIGSEGV`.

`MemoryCMARWImpl.close()` previously left `lim` and `appendAddress` pointing at
the released mapping. Those stale bounds could let `checkAndExtend()` accept an
address without remapping, turning post-close use into an access through
unmapped memory.

## Fix

- `flushAllPending()` rejects non-empty work when `valueMem` is closed.
- `flushAllPendingDense()` applies the same check after its existing
  `hasPendingData` no-op return. This is defensive coverage for the dense
  sibling; current production callers rebuild or discard the writer instead of
  retrying a failed dense commit on the same instance.
- `reencodeAllGenerations()` rejects a closed `valueMem` when `genCount > 0`.
  The `genCount == 0` no-op path remains unchanged.
- `MemoryCMARWImpl.close(boolean, byte)` clears `lim` and `appendAddress` after
  releasing the mapping. A later `of()` reconstructs both bounds from the new
  mapping.

The guards run before scratch allocation or posting-index mutation. They do not
attempt to make the degraded writer reusable.

## Tradeoffs / scope

- Table-level recovery is unchanged: the caller still distresses and discards
  the writer, then recovers the on-disk chain on reopen. The change replaces
  the invalid memory access with a controlled exception on the way there.
- Closed `MemoryCMARWImpl` instances now report zero append bounds instead of
  exposing stale addresses from a released mapping. Reopening the instance
  restores valid bounds.
- The sparse flush and reencode fault states have dedicated reproductions. The
  dense guard remains defense-in-depth because no supported production path
  was found that retries the same degraded dense writer.
- The change does not alter `extend0()`'s close-on-remap-failure policy or the
  ordering of `truncate()`'s close and replacement open.

## Discovery

Found while investigating a rare CI failure in
`WalWriterFuzzTest.testConvertPartitionToParquet`, which injects I/O faults.
The original failure reproduced locally at roughly 1 run in 100. Under `-ea`,
the rollback path could catch the resulting `AssertionError`, distress the
writer, and let the fuzz test continue, while assertion-free builds remained
exposed to the invalid native-memory access.

## Test plan

- Added
  `PostingIndexCriticalIssuesTest#testRollbackValuesRejectsFlushOverClosedValueMemAfterValueFileRemapFailure`.
  It fails a `.pv` remap with pending data and verifies that rollback reports
  the sparse-flush `CairoException`.
- Added
  `PostingIndexCriticalIssuesTest#testRollbackValuesRejectsReencodeOverClosedValueMemAfterTruncateOpenFailure`.
  It fails the replacement `.pv` open after `truncate()` and verifies that
  rollback reports the reencode `CairoException` instead of reading native
  address zero.
- Added `MemoryCMARWImplTest` coverage for `close(true)`, `close(false)`,
  repeated close without an open, and close followed by reopen.
- The six targeted regression tests pass on the PR head. Restoring the changed
  production files to their base versions makes the sparse-flush and mapped
  close regression assertions fail; the reencode reproduction terminates its
  isolated test JVM with `SIGSEGV` (exit 134).
- `PostingIndexCriticalIssuesTest` full class: 127 run, 0 failures.
- Ran `WalWriterFuzzTest.testConvertPartitionToParquet` approximately 135 times
  on the fixed build: no crashes; one injected fault reached the new guard and
  the test recovered.


puzpuzpuz:
## Code review (level 3)

**Verdict: approve.** The fix is correct, minimal, and backed by a regression test I verified empirically (red without the guard, green with it). No blocking findings. One non-blocking Moderate robustness recommendation plus a few Minor nits below.

The bug and mechanism check out end to end: `MemoryCMARWImpl.extend0` closes `valueMem` (fd=-1, pageAddress=0, size=0) on `mremap` failure but leaves `lim`/`appendAddress` stale; a follow-up `rollback → rollbackValues → flushAllPending` then reaches `valueMem.addressOf` (line 3160), which asserts under `-ea` and **wild-writes near address 0 in production**. `TableWriter.rollback` catches `Throwable`→`distressed`, so `-ea` hid it in CI while production could SIGSEGV/corrupt. The guard converts this into a clean `CairoException` with the same recovery outcome.

### Critical
None.

### Moderate

**1. The memory-safety guard is asymmetric — the structurally identical sibling `flushAllPendingDense()` is unguarded (pre-existing; defense-in-depth).**

`flushAllPendingDense()` has the same early-return shape and dereferences a closed `valueMem` the same fatal way: `getAppendOffset()` (returns a *stale high* pointer since `close` never resets `appendAddress`), `putLong(0L)`, `addressOf(genOffset)`. `reencodeAllGenerations()` (reached from `seal()`/`rollbackToMaxValue()`) has the same unguarded derefs. The new guard only fires when `hasPendingData == true`, so it does not cover these.

This is **not a confirmed live crash** — I could not construct a reachable state where they run with a closed `valueMem`: every `commitDense()` caller reopens `valueMem` via `discardForRebuild()`/`of()` or diverts to the guarded `seal()` branch when `genCount > 0`; `discardForRebuild` resets `genCount = 0` so `rollbackValues`/`seal` early-return before the reencode; and `PostingIndexWriter.truncate()` (the only `of()`-failure path that would leave `genCount > 0` + closed `valueMem`) has no production caller (`SymbolMapWriter`/`SymbolMapUtil` hold a `BitmapIndexWriter`, and `TableWriter.truncate` runs `rollback()` before any symbol truncate anyway).

So it's latent, not a blocker. But the PR's stated purpose is a memory-safety *backstop*, and the guard comment overclaims that it "also protects the `seal()` on the writer's distressed-close path" — it protects only that path's pending sub-case. A backstop covering one of three identical deref sites will silently reintroduce the wild-write if a future change to the `commitDense`/`discardForRebuild` invariants breaks the unstated reopen/poison assumptions. **Suggested for this PR** (cheap — a field read): add the same `!valueMem.isOpen()` guard to `flushAllPendingDense()`, and either guard the reencode entry or tighten the guard comment so it doesn't claim protection it doesn't provide. Note: don't hoist the check above `flushAllPending`'s `hasPendingData` early-return — that would turn today's benign no-op early-returns into throws.

### Minor

- **Dead `catch` + inaccurate comment in the test** (`PostingIndexCriticalIssuesTest`, the `finally`). `PostingIndexWriter.close()` never flushes or seals, so the `catch (CairoException ignore)` around `writer.close()` is unreachable, and its comment ("A close()-path flush over the still-closed valueMem fails through the same clean guard") describes a flush that does not exist. Fix: drop the `try/catch` and call `writer.close()` directly, or reword to "close() frees resources without flushing; catch is defensive only."
- **Boolean naming** (test): `armRemapFailure` violates the `is/has` rule; the file's own precedent is `failArmed`/`armed`. Rename to `isRemapFailureArmed`.
- **Duplicated explanation** (test): the mechanism narrative in the method javadoc is largely restated in the inline comment. Keep it in one place.
- **Pre-existing double-pass in `flushAllPending()`**: the `totalValues` loop and the `maxDataSize` loop both iterate `activeKeyIds` and recompute `count`; the dense sibling already fuses this via a scratch cache. O(1)/key, non-hot — optional cleanup while here.
- **`isPoisoned` not set by the guard** (by design): the guard throws without poisoning, relying on the caller to distress. No current caller retries `flushAllPending`, so the theoretical infinite-retry loop is unreachable — but setting `isPoisoned = true` would match the existing unrecoverable-state pattern if belt-and-suspenders is wanted.
- **PR title length**: 86 chars is on the long side (not a hard rule for PR titles, but could be tightened).

### Downgraded (checked, dropped)

- "`rollbackToMaxValue`/reencode is a *reachable* crash via `truncate`" — dropped. `PostingIndexWriter.truncate()` has no production caller; `truncate()` runs `rollback()` first; and `reencodeAllGenerations` sets `maxValue = target`, making a retry idempotent.
- "Guard placed after the `totalValues == 0` reset is a regression" — dropped. `activeKeyCount > 0 ⇒ totalValues > 0`, and `valueMem` only closes at the extend *past* that block, so a closed-`valueMem` re-entry always has `totalValues > 0`; the two never co-occur.

### Test coverage

The one behavioral change — `flushAllPending()` throwing a clean `CairoException` on a closed `valueMem` — is **tested and empirically verified**: in a scratch worktree on the PR branch the new test passes with the guard, and with the guard removed it fails with `AssertionError` at `PostingIndexWriter.java:3160`, the exact production crash site. Dimensions covered: error path, boundary (closed memory is the boundary), resource cleanup (`assertMemoryLeak`). Concurrency N/A (single-threaded writer), NULL N/A. Coverage map: 1 behavioral change, 1 tested, 0 untested.

### Summary

Correct, minimal, safe across all callers (`commit`/`sync`/`seal`/`rollbackValues`/`compactIfOverBudget`/`closeNoTruncate`), no leak/double-free/chain-corruption on the throw path. The main thing worth folding into this PR is the symmetric guard on `flushAllPendingDense()` (and the comment fix), so the backstop covers all three identical deref sites rather than one — cheap, and aligned with the fix's own goal.

<sub>🤖 Automated review by Claude Code (level 3: 6 parallel review passes + per-finding source verification).</sub>

puzpuzpuz:
## Code review (level 3)

Reviewed `a99a54a1d5` → `a6f6b2e415`. No submodule pointer moved in this range.

**Verdict: approve with comments.** No blocking findings. One Moderate coverage gap, four Minors, plus one pre-existing adjacent issue worth filing separately.

---

### Verification performed

The central claim is confirmed by execution rather than by reading:

- New test at head: `Tests run: 1, Failures: 0`.
- Full `PostingIndexCriticalIssuesTest` at head: `Tests run: 131, Failures: 0` — the three guards turn no exercised valid flow into a throw.
- Production hunk reverted to base in an isolated worktree, same test: **red**, at exactly the claimed crash site.

```
java.lang.AssertionError
  at io.questdb.cairo.vm.AbstractMemoryCR.addressOf(AbstractMemoryCR.java:63)
  at io.questdb.cairo.idx.PostingIndexWriter.flushAllPending(PostingIndexWriter.java:3466)
  at io.questdb.cairo.idx.PostingIndexWriter.rollbackValues(PostingIndexWriter.java:1617)
```

Mutation testing over the three guards (out-of-tree harness; repo untouched), run against 732 posting/covering tests:

| Mutation | Result |
|---|---|
| delete guard 1 (`flushAllPending`) | new test fails — `AssertionError` at `addressOf` |
| invert guard 1 to `if (valueMem.isOpen())` | 117 failures |
| delete guard 2 (`flushAllPendingDense`) | **0 of 732 fail** |
| delete guard 3 (`reencodeAllGenerations`) | **JVM SIGSEGV, exit 134**; 0 of 732 fail |
| move guard 1 above the `hasPendingData` early-return | 0 of 732 fail |
| drop `genCount > 0` from guard 3 | 0 of 732 fail |

Two design choices check out as deliberate rather than incidental:

- **`genCount > 0` in the reencode guard is load-bearing.** With `genCount == 0` the Phase-1 loop never runs, `totalValueCount` is forced to 0, and control routes to `truncate()`, which *reopens* `valueMem`. An unconditional guard would break that self-healing path.
- **All three guards precede every state mutation**, so a throw leaves no half-updated writer. `PostingIndexWriter.close()` never flushes or seals, so nothing leaks on the way out; every reachable handler distresses the writer or frees it in a `finally`, and no caller retries in a loop.

I also tried and failed to construct the guards' most plausible false positive — that placing them above the `totalValues == 0` early-return converts a clean base reset into a throw. `add()` appends to `activeKeyIds` only when the key's count is 0 and immediately writes count 1 (`:420`, `:431`), and every site that zeroes pending counts zeroes `activeKeyCount` in the same breath. The state cannot be reached.

---

### Moderate

**The `reencodeAllGenerations` guard has no test, and nothing else catches its removal**

`rg -n "cannot reencode posting index"` returns exactly one hit, in production (`PostingIndexWriter.java:4965`) — zero test references. Deleting the guard aborts the JVM while all 732 posting/covering tests stay green.

The failure mode there is worse than the PR describes. `MemoryCR.checkOffsetMapped(offset)` is `offset <= size()` (`MemoryCR.java:48-50`), and `size()` is 0 on a closed mapping. Gen 0 always sits at file offset 0. So `valueMem.addressOf(0)` **passes the `-ea` assert** and returns `pageAddress + 0 == 0`; Phase 1's `Unsafe.getInt(keyIdsBase + i*4)` (`:4980`) then reads from address 0. The javadoc's "under `-ea` it asserts in `MemoryCR.addressOf`" holds at the `flushAllPending` site (offsets > 0) but not here, where it is a hard SIGSEGV with no assert to catch it.

Producer: `truncate()` (`:2069-2075`) frees the pending buffers, closes `valueMem`, then reopens the new `.pv`. On an open failure (ENOSPC/EMFILE) the resets at `:2093-2100` never run, leaving `valueMem` closed with `genCount > 0` and `pendingCountsAddr == 0` — the degraded state `truncate()`'s own comment at `:2060-2062` already acknowledges. A later `seal()`/`rollbackValues()` sails past guard 1 (early-return on `pendingCountsAddr == 0`) straight into the reencode.

A regression test is cheap and non-invasive: the same `FilesFacade`-fault shape the new test already uses, failing the `.pv` open inside `truncate()` and then calling `seal()` or `rollbackValues()`. The file already carries ten such fault facades.

Classified Moderate rather than Critical deliberately: the executed reachability is at the `PostingIndexWriter` public-API level, and the `TableWriter` → `rollback()` → `rollbackIndexes()` → `rollbackValues()` continuation that would make it user-visible is static reading, not an executed artifact. The gap's regression consequence is also "revert to what master does today" — this PR strictly improves on base.

Also untested, lower value: guard 1's placement above the `hasPendingData` early-return, and guard 3's `genCount > 0` gating.

---

### Minor

**1. Comments overstate which paths the guards cover.** `:4957` calls the reencode site "the third deref site", and the description concludes "With all three flush/reencode deref sites guarded, the `seal()` ... is fully covered". `seal()` has two limbs: `sealIncremental` (`:1832`) and `sealFull` (`:1843`). Only `sealFull` reaches the new guard; `sealIncremental` dereferences `valueMem` unguarded at `:5923-5924`, `:6021-6022`, `:6043`. That code is byte-identical to base and I could not prove it reachable with a closed `valueMem`, so it is not a defect here — only the completeness wording.

Separately, `:3380` and `:4960` attribute the entry to "the `seal()` on the writer's distressed-close path". The only distressed close in this class is `closeNoTruncate()` (`:579`), which has no production caller — its sole call site is `SymbolMapWriter.java:441`, whose `indexWriter` field is declared `private final BitmapIndexWriter` (`SymbolMapWriter.java:61`, constructed at `:116`). `seal()` *is* production-reachable (ADD INDEX `TableWriter.java:7857,7898`; parquet index `:8068`; `rebuildColumnIndex` `:12681`; `sealIfMultiGen` `:14624`; `IndexBuilder.java:211`; `TableSnapshotRestore.java:1120,1832`) — just never from a distressed close.

**2. The mechanism claims are wrong at two sites, in opposite directions.** `:3641-3644` says the dense flush "would then dereference it (getAppendOffset / putLong / addressOf)". On the stated producer `extend0`'s catch leaves `appendAddress == lim` (`MemoryCMARWImpl.java:402-404`), so `getAppendOffset()` is pure arithmetic; `valueMem.putLong(0L)` (`:3735`) → `appendAddressFor(8)` → `checkAndExtend` does *not* short-circuit → `extend0` → `assert size > 0` under `-ea`, or NPE on the null `ff` in `TableUtils.allocateDiskSpace` in production; `addressOf` (`:3790`) is never reached. That site fails *more* loudly than claimed. The reencode site (above) fails *far more severely* than claimed. The guards are still worth having at both — a domain error beats an NPE on a null `FilesFacade` — but the descriptions should match.

**3. Test name states the opposite of what it asserts.** `testRollbackValuesToleratesClosedValueMem...` (`PostingIndexCriticalIssuesTest.java:944`) — `rollbackValues` does not tolerate the closed value memory, it rejects it (`:1010-1012`). Neighbours in this file name the asserted behaviour (`testCloseNoTruncateSkipsSealOnPoisonedWriter`, `testSquashPartitionsResealFailureMarksWriterDistressed`).

**4. Dead disarm line in the test's `finally`.** `armRemapFailure.set(false)` (`:1015-1017`) cannot affect anything: the only `.pv` fd was already removed from `valueFds` because `MemoryCMARWImpl.close(boolean,byte)` (`:118-123`) calls `Vm.bestEffortClose`, whose `finally` calls `ff.close(fd)` (`Vm.java:52-54`), which the test's own `close(long)` override (`:949-952`) hooks — and `PostingIndexWriter.close()` opens no file. Related: the `catch (CairoException ignore)` at `:1019-1023` cites "keyMem trim I/O", but `close()` already wraps `peekRegionLimit` + `setSize` in its own `catch (Throwable)` (`PostingIndexWriter.java:487-508`). Keeping the catch is defensible — a throw from a `finally` would mask a real assertion failure — but the rationale should say that.

Two things checked and found fine: guard 2 really is defence-in-depth (all three `commitDense` callers — `TableWriter.java:13924,13996`, `O3PartitionJob.java:3704` — re-`of()` the writer first, so the claim in the description holds there), and `CairoException.critical(0)` is the right flavour (`isCritical()` true, `isWALTolerable()` false, so `TableWriter.apply()` takes the `setSeqTxn(seqTxn - 1)`-and-rethrow branch).

---

### Adjacent — pre-existing, worth a separate issue

**`MemoryCMARWImpl.close()` leaves `lim` and `appendAddress` stale** (`MemoryCMARWImpl.java:89-137`). It zeroes `pageAddress`, `size`, `fd` and nulls `ff`, but never resets `lim`/`appendAddress`, so `checkAndExtend` (`:376-381`) short-circuits for any offset under the stale bound. Use-after-close therefore degrades from a loud failure into a wild pointer. This is exactly why the reverted-hunk run reached `addressOf` instead of failing at `extend0` — the three preceding `jumpTo` calls silently succeeded. It affects every mapped writer, not just posting indexes. Suggested fix: reset `lim = 0` and `appendAddress = 0` alongside `pageAddress`. Not attributable to this PR; happy to file it.

---

### Tradeoff worth stating

On a production (assertions-off) build the change converts a wild store that *might* have let `rollback()` run to completion into a clean abort that skips `purgeUnusedPartitions()` and `processCommandQueue()`. That is the right trade — the writer is distressed and discarded either way, and every skipped step is redone on reopen — but it is a behaviour change, not a pure no-op.

ideoma:
#### [PR Coverage check]

:heart_eyes: **pass** : 7 / 14 (50.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|7|14|50.00%|

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=262367)
**Tested Commit:** `a6f6b2e415838a38db3a42c242e1fa511376c7f2`

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=262427)
**Tested Commit:** `507bd97e5446f6df26ba6b3e5d13d282ca81505c`

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=262424)
**Tested Commit:** `507bd97e5446f6df26ba6b3e5d13d282ca81505c`

ideoma:
#### [PR Coverage check]

:heart_eyes: **pass** : 30 / 38 (78.95%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|10|14|71.43%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpTudCache.java|16|20|80.00%|
|:large_blue_circle:|io/questdb/cairo/vm/MemoryCMARWImpl.java|2|2|100.00%|

bluestreak01 (approved):
Reviewing PR #7382 at level 3.

### Critical

None.

### Moderate

#### 1. PR description no longer matches the implemented change

Problem: PR description no longer matches the implemented change.  
Net impact: Reviewers receive incorrect scope and regression-test coverage.  
Evidence: PR body lines 66 and 79-89 at `6f42031`; current tests contradict them.

The body:

- Names the obsolete `testRollbackValuesToleratesClosedValueMemAfterValueFileRemapFailure`.
- Says both sibling guards lack dedicated reproductions, but `testRollbackValuesRejectsReencodeOverClosedValueMemAfterTruncateOpenFailure()` now covers reencoding.
- Says only the failed-remap path is addressed, despite the added failed-open path and general `MemoryCMARWImpl.close()` fix.

Update the scope and test-plan sections to reflect the current head.

### Minor

#### 1. Test Javadocs duplicate private implementation details

Problem: Test Javadocs duplicate private implementation details.  
Net impact: Internal refactors require synchronizing redundant explanations.  
Evidence: Changed Javadocs at `6f42031`, including `PostingIndexCriticalIssuesTest.java:1299-1312,1410-1429`.

Several test Javadocs restate private control flow already documented in production. Prefer precise test names and short comments near safety-critical assertion ordering.

### Adjacent findings

None.

### Coverage map

**Test gate: pass — 0 admitted coverage gaps.**

Validation performed:

- Six targeted tests passed at head:
  - 4 `MemoryCMARWImplTest` tests
  - 2 posting-index fault tests
- Restoring both production files to base in an isolated clone caused:
  - The sparse rollback test and three mapped-close tests to fail.
  - The reencode test to terminate its isolated JVM with `SIGSEGV`, exit 134.
- `git diff --check` passed.
- Current PR CI, including Linux, macOS, Windows, fuzz, coverage, and enterprise checks, is green.

### Summary

- **Verdict: approve with comments** — update the stale PR description.
- **Correctness gate:** pass; no admitted runtime defects.
- **Test gate:** pass; 0 admitted coverage gaps.
- **Severity:** 0 Critical, 1 Moderate, 1 Minor.
- **Attribution:** 2 in-PR findings, 0 out-of-diff breakages.
- **Submodules:** no pointer changes.
- **Tradeoff:** the fault still distresses and discards the writer; the PR replaces invalid memory access with a controlled error.

PR #7385, opened 2026-07-10
chore(core): prevent WAL table suspension after posting-index partition squash

RaphDal:
## What this fixes

A WAL table with a `POSTING`-indexed column could suspend during an immediate partition squash after an out-of-order split. The failure was first observed in `WalWriterFuzzTest.testConvertPartitionToParquet` on Linux ARM64 with seeds:

```text
6424878685474624049L, 7598331127391241803L
```

The final fuzz assertion was misleading:

```text
Column count must be same expected:<13> but was:<14>
```

The WAL and non-WAL tables did not diverge because of a schema bug. WAL application had already suspended while reopening a posting index after the squash, so the WAL table stopped receiving later structural transactions. The first failure was an out-of-bounds read of the posting-chain head:

```text
PostingIndexWriter.of
  -> PostingIndexChainWriter.openExisting
    -> PostingIndexChainEntry.read
      -> MemoryCR.getInt
        -> AbstractMemoryCR.addressOf
```

The `.pk` header described a valid, non-empty chain, but its `headEntryOffset` was equal to or greater than the physical file length. On assertions-enabled builds this trips the mapped-range assertion; without that guard it is an invalid native-memory read.

## Confirmed failure sequence

The captured fuzz run exercised this ordering:

1. The WAL table added `new_col_4` as an indexed `SYMBOL` column. Its column top was above the rows in the eventual squash target, so the column was row-less in that target partition.
2. An out-of-order WAL write split partition `2022-02-24.84`, leaving a 680-row target and a 56-row split tail.
3. Housekeeping immediately squashed the tail back into the target through `FrameAlgebra.append`.
4. `FrameImpl` saw `columnTop >= target row count` and opened a fresh `ContiguousFileIndexedFrameColumn`. Its short-lived `PostingIndexWriter` mapped the same `.pk` file that the table's long-lived posting indexer already had open.
5. The frame writer appended the tail rows, committed them, and published the first v2 chain entry. Its close preserved the new entry correctly.
6. The table writer then forced a posting-index reseal through `sealPostingIndexForPartition`. `SymbolColumnIndexer.configureFollowerAndWriter` reopened the table indexer, and `PostingIndexWriter.of()` first closed its old mapping.
7. That old writer still cached the empty-chain `regionLimit` of 8192 bytes. Its close reset the append offset to 8192, then zeroed and truncated the shared `.pk` at the page-rounded 8192 boundary even though the published header now pointed to the frame writer's entry starting at offset 8192.
8. The reopen read the published head pointer and attempted to read the now-truncated entry. The table writer became distressed and `ApplyWal2TableJob` suspended WAL application.

For a non-covering entry with one generation, the state transition is:

| Point | Long-lived writer cache | Published v2 header | Physical `.pk` length |
| --- | ---: | --- | ---: |
| Both writers start from an empty chain | `regionLimit = 8192` | no head, `regionLimit = 8192` | `ceilPageSize(8192)` |
| Frame writer commits and closes | still `8192` | `headEntryOffset = 8192`, `regionLimit = 8296` | `ceilPageSize(8296)` |
| Long-lived writer closes before this fix | uses cached `8192` | still advertises the entry ending at 8296 | `ceilPageSize(8192)` |
| Reseal reopens the index | n/a | asks for the head at 8192 | head entry is missing or zeroed |

The source partition's seal purge happened near the failure in the fuzz log, but it was not the cause. The focused reproducer needs no purge job, O3 worker, parquet conversion, or WAL machinery: two overlapping writer mappings and the stale close are sufficient.

## Root cause

The v2 `.pk` layout reserves two 4 KiB seqlock-protected header pages and stores chain entries after them. The active header's `regionLimit` is the authoritative high-water mark past the last published entry.

`PostingIndexWriter.close()` trims the key file before unmapping it. It selected the trim point only from `PostingIndexChainWriter.getRegionLimit()`, an instance-local mirror populated when that writer opened the chain and advanced only by publishes made through that same instance.

That assumption is valid when one writer mapping owns the file for its entire lifetime. It is false during the frame/table-writer hand-off above: the frame writer publishes through a second mapping, so the shared header advances while the table writer's in-memory mirror remains at the empty-chain boundary. Closing the stale instance then turns harmless stale metadata into destructive truncation.

This is not an ARM64-specific publication or memory-ordering failure. The frame writer completed a valid seqlock publication and extended the file before the table writer closed. The stale writer simply ignored that published extent.

## Fix

`PostingIndexWriter.close()` now snapshots the currently published header with `PostingIndexChainHeader.readUnderSeqlock()` before choosing the trim size. For a v2 header it retains:

```text
max(writer-local regionLimit, published header regionLimit)
```

The existing 8192-byte minimum remains in place.

Using the maximum preserves both relevant invariants:

- A writer never truncates entries already published through another mapping.
- A stale or older header snapshot cannot make the writer discard an extent represented by its own newer in-memory chain state.

When the authoritative extent lies beyond the stale writer's mapping, close now releases that mapping without truncation. The completed frame writer already extended and closed the shared file, so growing the stale mapping would add no live data. Avoiding that growth also prevents an allocation failure from leaving the stale append offset armed for the normal truncating cleanup path.

The snapshot object is a reusable `PostingIndexChainHeader.Snapshot` field, so close does not allocate. The read touches only the two already-mapped header pages and uses the same bounded A/B seqlock reader as normal chain open; it does not scan entries or add filesystem I/O. A normal close still selects its local extent and truncates as before, while a stale close avoids an otherwise unnecessary allocation, remap, and immediate unmap.

The fix belongs at the close/truncate boundary rather than in the squash caller. The invalid operation is deciding a shared file's final extent from one mapping's private cache. Refreshing only one TableWriter call site would leave the same invariant exposed to any other lifecycle that temporarily overlaps posting-index mappings.

## Effects, costs, and limits

- The squash reseal continues to run. This does not mask the failure by skipping reseal or weakening chain validation; the index still has to reopen and rebuild from the partition data.
- There is no storage-format, metadata, SQL, or query-plan change. The fix consumes the existing v2 `regionLimit` field as its documented high-water mark.
- Normal single-writer closes retain the same file size because the local and published limits are equal.
- A close with an overlapping, already-published writer may retain more bytes than its local cache knows about. The header identifies the live part of that extent. On platforms that cannot trim while another mapping remains open, the non-truncating stale close can temporarily retain the publishing writer's page or append-segment tail; the next exclusive writer close trims it normally.
- The added cost is one bounded seqlock header snapshot on writer close. Posting-index add, commit, seal, and query hot paths are unchanged.
- A stale close no longer extends or remaps its shorter mapping just to establish a truncate offset.
- This does not make independent `PostingIndexWriter` instances general-purpose concurrent writers. It protects a completed publication made by the frame writer before the stale table writer closes, which is the serialized production ordering in `FrameAlgebra.append` followed by reseal.

This failure is distinct from the spill/reseal corruption fixed in #7203. In this case the frame writer publishes a structurally valid first entry and closes cleanly; the entry is lost afterward when the older mapping truncates the file.

## Regression test

`PostingIndexCriticalIssuesTest.testCloseDoesNotTruncateChainAppendedByAnotherWriter` isolates the file lifecycle directly:

1. Open writer A on a new empty posting index, leaving its cached chain limit at 8192.
2. Open writer B on the same path, add one row, set `maxValue`, commit the first generation, and close it.
3. Close stale writer A.
4. Reopen the posting index and assert `maxValue == 0` and `genCount == 1` under `assertMemoryLeak`.

Before the production change, step 4 fails in `PostingIndexChainEntry.read()` through the same `AbstractMemoryCR.addressOf()` stack as the fuzz run. After the change, the reopen succeeds and reads the generation writer B published.

The direct test replaces a scheduler-dependent end-to-end fuzz reproducer with the smallest deterministic state transition. It also proves that the nearby seal purge is not required to corrupt the file.

`PostingIndexCriticalIssuesTest.testCloseExtendFailureDoesNotTruncateExternalChain` covers the resource-failure boundary:

1. Create and close an empty index, then reopen writer A so it maps the page-rounded empty-file length.
2. Let writer B publish enough generations to extend the authoritative chain past A's mapping on the current OS page size, then close B.
3. Arm a `FilesFacade` that advertises A's stale length and rejects any `.pk` allocation attempted by A's close.
4. Close A and assert that it does not attempt the allocation, does not throw, and does not change the file length.
5. Reopen the index and assert that every generation B published remains visible.

Before the follow-up close change, step 4 entered `MemoryCMARWImpl.extend0()`. Its allocation runs before the remap failure guard, so the throw left A's old append offset active and the `finally` cleanup truncated B's chain. Closing the stale mapping without truncation removes that failure path entirely.

## Test plan

- Validated the regression red before the production change and green afterward with `mvnd -pl core -Dtest=PostingIndexCriticalIssuesTest#testCloseDoesNotTruncateChainAppendedByAnotherWriter test`.
- Ran `mvnd -pl core '-Dtest=PostingIndexCriticalIssuesTest#testCloseDoesNotTruncateChainAppendedByAnotherWriter+testCloseExtendFailureDoesNotTruncateExternalChain' test`; both close-path regressions passed.
- Ran `mvnd -pl core '-Dtest=PostingIndexCriticalIssuesTest#testCloseDoesNotTruncateChainAppendedByAnotherWriter+testReviewFinding3_PostingReaderHandlesStaleKeyFileLength+testSquashNonCoveringPostingWithMidStreamSpillFlush+testWalApplyAutoSquashCoveringPostingWithMidStreamSpillFlush' test`; all 4 selected tests passed.
- Ran `mvnd -pl core -Dtest=PostingIndexCriticalIssuesTest test`; all 131 tests passed.
- Mutation-checked `testCloseExtendFailureDoesNotTruncateExternalChain` by restoring the unconditional `setSize()` close path; it failed with an expected file length of 12,288 bytes and an actual length of 8,192 bytes, then passed with the fix restored.
- Ran `mvnd -pl core -Dtest=PostingIndexChainWriterTest test`; all 26 tests passed.
- Ran `git diff --check`; it reported no whitespace errors.
- Did not run the full `core` or repository test suite.


RaphDal:
# PR #7385 review

Reviewing PR #7385 at level 0 (default). Level 3 is available for the full multi-pass review.

## Critical

None.

## Moderate

None.

## Minor

1. **In-diff:** `core/src/test/java/io/questdb/test/cairo/fuzz/WalWriterFuzzTest.java:307` adds two large numeric literals without separators, contrary to the repository convention. Use `6_424_878_685_474_624_049L` and `7_598_331_127_391_241_803L`.

2. **PR metadata:** The title misclassifies a user-visible bug fix as `chore` and does not repeat the Conventional Commit action. Use `fix(core): fix WAL table suspension after posting-index partition squash`.

## Downgraded

- **Claim:** `keyMem.close(false)` followed by `Misc.free(keyMem)` could truncate the file anyway.  
  **Dismissed:** `core/src/main/java/io/questdb/cairo/vm/MemoryCMARWImpl.java:89` clears the mapping, descriptor, size, and facade state. The subsequent generic close has nothing left to truncate.

- **Claim:** Reading the shared header during close races the frame writer.  
  **Dismissed:** The production sequence closes the frame at `core/src/main/java/io/questdb/cairo/TableWriter.java:13784` before resealing at `core/src/main/java/io/questdb/cairo/TableWriter.java:13823`. The supported handoff is serialized, not simultaneous multi-writer publication.

## Coverage map

|---|---|---|---|
| `PostingIndexWriter.close()` retains a newer published v2 extent that still fits inside the stale mapping. | `PostingIndexCriticalIssuesTest.testCloseDoesNotTruncateChainAppendedByAnotherWriter` reopens the file and asserts `maxValue == 0` and `genCount == 1`; the old close truncates/zeros the published head, so reopening fails. `WalWriterFuzzTest.testPostingIndexCloseAfterPartitionSplitSquash` also reaches the production WAL squash/reseal sequence. | Happy: covered. Error: N/A—no new error branch. NULL: N/A—file offsets are not nullable. Boundary: covered with empty chain and first single-row generation. Concurrency: covered for the supported serialized overlapping-mapping handoff. Cleanup: covered by `assertMemoryLeak`. | **TESTED** |
| When the published extent exceeds the stale mapping, `close()` releases without growing, allocating, or truncating. | `PostingIndexCriticalIssuesTest.testCloseExtendFailureDoesNotTruncateExternalChain` asserts unchanged file length, no allocation attempt, no close exception, and successful reopening of every generation. | Happy: covered. Error: covered with an armed allocation-failing facade. NULL: N/A. Boundary: covered at the first extent beyond the OS-rounded mapping size. Concurrency: covered for overlapping mappings. Cleanup: covered by `assertMemoryLeak`. | **TESTED** |
| Normal single-writer close where local and published limits match. | Verified delta: `Math.max(local, published)` returns the original local limit and the existing `setSize(liveSize)` path remains unchanged. | All dimensions N/A because behavior is unchanged. | **EXEMPT** |

## Summary

**Verdict: approve.**

Coverage map: 2 behavioral changes, 2 tested, 0 UNTESTED, plus 1 verified no-behavior exemption.

The cost is one bounded O(1) seqlock header snapshot per close. A stale mapping shorter than the published extent may retain page-rounded tail bytes until a later exclusive close; it does not lose live data.

Local validation passed:

- 2 focused `PostingIndexCriticalIssuesTest` regressions
- 1 fixed-seed `WalWriterFuzzTest` regression
- `git diff --check`

CI was still running at review time. Two findings were verified, and two false positives were removed: 1 in-diff, 0 out-of-diff, 0 pre-existing, and 1 metadata finding.

bluestreak01 (approved):

bluestreak01 (changes requested):
tests are failing, needs re-review

bluestreak01 (changes requested):
@RaphDal

# PR #7385 review — level 3

Reviewing PR #7385 at level 3 (full mission-critical pass: all 14 agents, per-finding verification, executed evidence).

**Step 2.4 — Submodule provenance:** no submodule pointer moved. `git diff 496b24d9...46645ae1 --submodule=short` lists three ordinary Java files. Nothing out of scope on submodule grounds.

**Executed evidence gathered (scratch worktree, discarded):** the two unit regression tests and the fuzz test were run on HEAD (pass) and with the production hunk reverted to BASE (all three fail with their intended assertions). Two additional probes were written, run, and deleted. Commit SHAs, outputs and reasoning are attached per finding.

---

## Critical

### C1 — The fix is inert on every covering POSTING index

**Problem:** Version gate excludes V3, disabling the fix for covering indexes.
**Net impact:** Covering-index tables keep the full pre-fix truncation bug.

- **Location:** `core/src/main/java/io/questdb/cairo/idx/PostingIndexWriter.java:459-462`
- **Scope:** in-diff (**incomplete-hardening escalated** — see attribution below)
- **Symptom + trigger:** A user with a COVERING POSTING index ingests via WAL until a partition squash fires → the stale writer's `close()` skips the published-extent rescue, truncates the `.pk` back to its own cached `regionLimit`, and the reopened index silently comes back **empty** → indexed predicates return fewer rows than the table holds.

```java
if (PostingIndexChainHeader.readUnderSeqlock(keyMem, closeHeaderScratch)
        && closeHeaderScratch.formatVersion == V2_FORMAT_VERSION) {   // <-- V3 falls through
    liveSize = Math.max(liveSize, closeHeaderScratch.regionLimit);
}
```

- **Load-bearing step:** *a live production `.pk` can carry `FORMAT_VERSION == 3`, and the guard is then inert.* Verified two ways:
  1. **Static:** `PostingIndexChainWriter.raiseFormatVersionFor` (`:978-981`) sets `formatVersion = V3_FORMAT_VERSION` on the first de-aliased covering entry with `coverCount > 0`; `newEntryCoveringFormat()` returns `COVERING_FORMAT_DEALIASED` for any `coverCount > 0` and its only escape hatch, `FORCE_LEGACY_COVERING_FORMAT`, is `@TestOnly` / *"Never set in production"* (`PostingIndexWriter.java:909-913`). `openExisting` keeps the version already on the file, so later non-covering publishes republish `3`. The canonical gate is `PostingIndexUtils.isSupportedFormatVersion` (`:1142-1144`), whose javadoc states *"Every version gate goes through here"* — and the two sibling readers, `openExisting` (`PostingIndexChainWriter.java:557`) and `peekRegionLimit` (`:623`), both use it. This new line is the only one that hard-codes `== V2`.
  2. **Executed (this is the decisive one):** I ran `testCloseDoesNotTruncateChainAppendedByAnotherWriter`'s exact sequence **on HEAD with the fix applied**, stamping only `FORMAT_VERSION = 3` into the live active header page (seqlock left even and stable — the shape a real covering `.pk` has). Observed on commit `46645ae1df`:
     ```
     PROBEV3: seqA=2 seqB=4 active=4096 oldFormat=2 regionLimit=8296
     PROBEV3: genCount=0 maxValue=0
     AssertionError: V3 header: published generation must survive the stale close expected:<1> but was:<0>
     ```
     The identical sequence with `FORMAT_VERSION = 2` passes. The guard is inert on V3.
- **Reachability of the covering case:** `TableWriter.squashSplitPartitions` states the reseal exists precisely for this population — *"For COVERING POSTING indexes this is what drops rows from indexed predicates: the reader can resolve the index key but cannot resolve the covering columns for any rowid the squash merged in, so the rows fall out of the result"* (`TableWriter.java:14482-14490`). Covering indexes are the primary population of the path this PR fixes. Residual uncertainty I did **not** eliminate: I did not execute an end-to-end covering squash that lands on the stale-close overlap. The one-line fix removes the question entirely, which is why I am not spending CI on establishing it.
- **Net determination:** population = tables with covering POSTING indexes undergoing partition squash · delta vs base = **none for V3 files** (identical to merge base) · magnitude = full loss of the newest chain generation, same frequency as the V2 case the PR fixes · offsets = none found; searched every `formatVersion` comparison under `cairo/idx`, and the new fuzz detector `WalWriterFuzzTest.readPublishedRegionLimit` **also** requires `formatVersion == V2_FORMAT_VERSION` (`:1088`), so it is blind to the same input · net = net-neutral for V3, net-positive for V2.
- **Base behavior at BASE (`496b24d9`):** identical — `PostingIndexWriter.BASE.java:457-461` never reads the header. *Static citation; nothing regressed.*
- **Why this is Critical rather than Moderate.** By attribution this is incomplete-hardening and caps at Moderate. It escalates under the second escalation — realistic trigger **and** a guarantee the PR presents as absolute. The body claims generality in terms that cover exactly this case: *"The invalid operation is deciding a shared file's final extent from one mapping's private cache. Refreshing only one TableWriter call site would leave the same invariant exposed to any other lifecycle that temporarily overlaps posting-index mappings."* The new fuzz test also runs with `addCoveringIndexProb = 0.01`, so it *appears* to cover covering indexes while its detector structurally cannot. Per the rubric, both remediation options must be offered:
  - **Option A (close the gap — recommended, one line + one test):**
    ```java
    if (PostingIndexChainHeader.readUnderSeqlock(keyMem, closeHeaderScratch)
            && PostingIndexUtils.isSupportedFormatVersion(closeHeaderScratch.formatVersion)) {
        liveSize = Math.max(liveSize, closeHeaderScratch.regionLimit);
    }
    ```
    plus a `PostingIndexCriticalIssuesTest` case that reaches `V3_FORMAT_VERSION` before the stale close, and the same `isSupportedFormatVersion` change in the fuzz facade's `readPublishedRegionLimit` so the detector is not blind.
  - **Option B (scope the promise):** state in the body and in a code comment that the rescue deliberately covers V2 `.pk` files only, that covering (V3) indexes retain the pre-fix close, and drop `addCoveringIndexProb` from the new fuzz case so the test does not imply coverage it does not have. This leaves a known live defect on master and I do not recommend it.

---

## Moderate

### M1 — Header read has no mapping-size precondition; a clean `CairoException` becomes an unmapped read

**Problem:** Close reads two header pages without checking the mapping covers them.
**Net impact:** Operators on a sub-8 KiB index page size get a crash, not an I/O error.

- **Location:** `core/src/main/java/io/questdb/cairo/idx/PostingIndexWriter.java:459`
- **Scope:** in-diff
- **Symptom + trigger:** An operator sets `cairo.writer.data.index.key.append.page.size` to one OS page (4096) and hits ENOSPC during a posting-index rollback-truncate → instead of the previous `CairoException: No space left`, `close()` reads 8 bytes past the end of a 4096-byte mapping: `AssertionError` under `-ea`, an unmapped read (SIGSEGV) in a `-da` production build.
- **Load-bearing step:** *`close()` can run with `keyMem.isOpen() == true` and `keyMem.size() < 8192`.* **Verified by execution** on commit `46645ae1df`, with `cairo.writer.data.index.key.append.page.size = 4096`:
  ```
  PROBE: mremap fd=... prev=12288 new=4096      <- keyMem.truncate() inside initKeyMemory
  PROBE: truncate fd=... size=4096
  PROBE: rejecting allocate fd=... size=8192    <- extend0 -> allocateDiskSpace fails
  PROBE: truncate threw = CairoException: [2] No space left [size=8192, ...]
  PROBE: close threw = java.lang.AssertionError  <- HEAD
  ```
  and on the same probe with `PostingIndexWriter.java` reverted to `496b24d9`:
  ```
  PROBE: close threw = CairoException: [2] No space left [size=12288, ...]  <- BASE
  ```
  Mechanism: `MemoryCMARWImpl.extend0` calls `TableUtils.allocateDiskSpace` **before** its `try`, so an allocation failure does not run `close(false)` and leaves `keyMem` open at `minMappedMemorySize`. `readUnderSeqlock` touches `PAGE_B_OFFSET + V2_HEADER_OFFSET_SEQUENCE_END` = offset 8184..8192, and `MemoryCR.getLong` asserts `offset + 8 <= size()`. Production entry point is real: `rollbackConditionally(0)` → `truncate()`, reached from `SymbolColumnIndexer.java:190` and `O3CopyJob.java:865`.
- **Net determination:** population = deployments with `cairo.writer.data.index.key.append.page.size < 8192` (default 512 KiB; test harness 16384) on an allocation-failure path · delta vs base = new (BASE never dereferenced the mapping in `close()`) · magnitude = one process crash or one masked root cause · frequency = very low · offsets = none found; searched every `close()` caller · net = **net-negative**.
- **Base behavior at BASE (`496b24d9`, executed above):** a typed `CairoException`; no assertion failure, no unmapped read.
- **Not Critical because** the population requires a non-default configuration below 8192 conjoined with an allocation failure, and the on-disk damage is identical to base in both cases.
- **Suggested fix (this PR, one line):**
  ```java
  if (keyMem.size() >= KEY_FILE_RESERVED
          && PostingIndexChainHeader.readUnderSeqlock(keyMem, closeHeaderScratch)
          && PostingIndexUtils.isSupportedFormatVersion(closeHeaderScratch.formatVersion)) {
  ```
  This also matches the file's own established pattern — every other header reader bounds-checks first (`PostingIndexWriter.java:1242`, `PostingIndexUtils.java:1114`, `:1236`).

### M2 — An unreadable header silently selects the truncating path

**Problem:** `readUnderSeqlock == false` falls through to the destructive branch.
**Net impact:** Corrupt-header closes still destroy the chain, with no diagnostic.

- **Location:** `core/src/main/java/io/questdb/cairo/idx/PostingIndexWriter.java:459-462`
- **Scope:** incomplete-hardening
- **Why the user cannot observe a change today:** behaviour for this input is identical to the merge base (`PostingIndexWriter.BASE.java:457-461` never read the header at all), so nothing regressed. *Base-delta citation: `496b24d9`, same trigger, same truncation to `ceilPageSize(local regionLimit)`.*
- **Load-bearing step:** *every other caller treats `false` as an error.* Verified: `PostingIndexChainWriter.peekRegionLimit` throws `"posting index header unreadable"` (`:620-622`), and `POSTING_INDEX_CHAIN_DESIGN.md` §5.2 specifies *"`RESULT_HEADER_UNREADABLE`: propagate as an I/O error"*. This is the only site that swallows it, and the swallow selects precisely the outcome the PR exists to prevent.
- **Suggested fix (optional):** a teardown must not throw, so take the conservative branch instead of the destructive one — `keyMem.close(false)` plus one ASCII `LOG.error()` naming the index path. That makes "I could not establish the live extent" mean "do not truncate", which is the policy `of()`'s catch already applies at `:1310`. Same treatment fits a range-invalid header (`regionLimit < regionBase`), which `peekRegionLimit` rejects and this path trusts.

### M3 — The fuzz test's "we exercised the scenario" guard is satisfied by an ordinary single-writer close

**Problem:** Guard counter fires for the normal writer lifecycle.
**Net impact:** The end-to-end test can silently stop exercising the race.

- **Location:** `core/src/test/java/io/questdb/test/cairo/fuzz/WalWriterFuzzTest.java:318-322`, counter at `:961-974`
- **Scope:** in-diff
- **Symptom + trigger:** The seed or the surrounding fuzz drifts so the two-mapping overlap no longer occurs → `assertTrue("fixed seed did not exercise an empty-to-nonempty posting-index mapping close", … > 0)` still passes, and the test degrades to a no-op without anyone noticing.
- **Load-bearing step:** *`initialPostingKeyLimits` records 8192 for a writer's own fd.* `recordInitialPublishedLimit` is invoked from `openRW`, `trackMapping`, `fsync` and `msync` and folds with `Math::min`, so **any** writer that opens an empty `.pk` (published `regionLimit == KEY_FILE_RESERVED`), seals once, and unmaps satisfies `initialLimit == 8192 && current > 8192` at `munmap`. No second mapping is required. Found independently by three agents; confirmed by reading the counter.
- **Net determination:** population = CI and reviewers · delta vs base = new test · magnitude = false confidence in coverage · offsets = the **other** assertion is sound and load-bearing: I ran the fuzz test at BASE and it fails with `posting-index close truncated below a concurrently published chain limit expected:<0> but was:<1>`, and passes on HEAD. So the test *is* a genuine regression test; only its guard is vacuous · net = net-neutral.
- **Suggested fix:** count only closes where the fd had ≥ 2 concurrent mappings (`postingKeyMappings` already keys by address, so a per-fd mapping count is a two-line addition), or assert on a counter incremented when `close()` actually takes the new branch.

### M4 — PR body attributes an already-fixed failure to this PR and never mentions #7386

**Problem:** Body describes a pre-#7386 world and omits the prior fix.
**Net impact:** Reviewers and release notes cannot tell what this PR adds.

- **Location:** PR body, sections "What this fixes", "Confirmed failure sequence", "Regression test"
- **Scope:** in-diff (PR metadata)
- **Load-bearing step:** *the narrated failure is the one #7386 already fixed, and #7386 is in BASE.* `git merge-base --is-ancestor e495eb272f 496b24d9` → YES. #7386's body cites the same test, the same seeds `6424878685474624049L, 7598331127391241803L`, the same `AbstractMemoryCR.addressOf` stack and the same `"Column count must be same expected:<13> but was:<14>"` string. #7386's Fix 1 is physically on this branch at `PostingIndexWriter.java:1247-1251`. The body cross-references #7203 but never #7386 — so the omission is not stylistic. The branch's first commit predates the merge of #7386 and the body was never refreshed across three subsequent master merges.
- **The description is now factually wrong in a checkable way, and I measured it.** With #7386 present, the stale close no longer produces the cited crash: the reopen peeks `regionLimit = 8296 > keyFileSize = 8192`, extends with zeros, and reads a **zeroed** entry. Running the new regression test at BASE gives:
  ```
  AssertionError: expected:<1> but was:<0>       (getGenCount)
  ```
  — not the `AbstractMemoryCR.addressOf` stack the body claims twice. On today's master the failure mode is **silent index emptying**, not a loud suspension. That is a *more* serious user consequence than the title advertises, and the body understates it.
- **Net determination:** population = reviewers, release-note readers, future bisectors · magnitude = the PR's entire justification is unverifiable as written · offsets = none found · net = net-negative. Capped at Moderate because the affected population is the team, not database users.
- **Suggested fix:** add a `## Relationship to #7386` section stating that #7386 fixed the *reopen* path and this PR fixes the *close* path (distinct root causes, distinct code sites, both needed); replace the "Confirmed failure sequence" narrative with the signature observed at `496b24d9` (silent `genCount == 0`, no `addressOf` stack); and correct the regression-test paragraph the same way.

### M5 — `chore(core):` contradicts the `Bug` label and this PR's actual user impact

**Problem:** A data-loss fix is typed as a chore.
**Net impact:** Release notes omit a user-facing storage fix.

- **Scope:** in-diff (PR metadata)
- **Evidence:** labels are `Bug, Core, WAL, READY`; the title type is `chore`. CLAUDE.md: *"Always add GitHub labels consistent with the PR title"* and PR titles must follow Conventional Commits with the description repeating the verb. The sibling PR for the same defect family used `fix(core):`. My executed BASE runs show a real, reachable defect (a published chain generation is lost), so `chore` is the wrong type — not the label.
- **Suggested fix:** retitle so it does not duplicate #7386's release-note line, e.g. `fix(core): fix POSTING index dropping generations published by an overlapping writer`.

---

## Minor

1. **Seed literals lack thousands separators.** `WalWriterFuzzTest.java:307` — `6424878685474624049L, 7598331127391241803L`. CLAUDE.md requires `_` for 5+ digits. No counter-norm in the file (the only other `new Rnd(` is a javadoc placeholder). This was raised in the author's own self-review comment on the PR and not applied.
2. **`getTestName()` override hard-codes another test's method name.** `WalWriterFuzzTest.java:881-884` returns `"testConvertPartitionToParquet"`. Renaming or deleting that test silently changes the fuzz corpus this repro depends on, diagnostics misattribute the table to the wrong test, and the two methods now share a table-name namespace. Combined with M3 the drift would go undetected. Consider naming the shared corpus explicitly (a constant used by both tests) rather than borrowing a method name.
3. **22-argument probability block is a byte-identical duplicate.** `runPostingIndexCloseAfterPartitionSplitSquashFuzz` (`:837-862`) duplicates `testConvertPartitionToParquet` (`:197-221`), including both trailing comments and the `setFuzzCounts(...)` line. If the original is retuned, the repro silently diverges from the configuration it was captured under. Extract one shared private method.
4. **`assertFalse(wasAllocationAttempted())` and its fault-injection scaffolding are inert against BASE.** In `testCloseExtendFailureDoesNotTruncateExternalChain`, BASE's `close()` computes `liveSize = 8192 <= keyMem.size()`, so it never extends and never calls `allocate` — the assertion, the `arm(staleMapSize)` and the `length()` override all pass vacuously against the merge base. They only discriminate against the intermediate branch commit `b590ee9379`. The test is still a valid regression test: its *file-length* assertion fails at BASE (`expected:<12288> but was:<8192>`, executed). Worth a comment saying which state the allocate machinery guards.
5. **Comment overstates concurrency.** `PostingIndexWriter.java:452-457` says the frame writer "can publish through a second mapping **while** this table writer still holds the previously empty index open", which reads as concurrent publication. The real mechanism is sequential mapping aliasing on one thread: `Misc.free(targetFrame)` (`TableWriter.java:14471`) precedes `sealPostingIndexForPartition` (`:14508`), which asserts `o3PartitionUpdRemaining.get() == 0`. The design doc says single-writer-per-`.pk` with no CAS protocol; the comment contradicts it and could license a future change that breaks the ordering this code depends on.
6. **`wasAllocationAttempted()` deviates from the `is…`/`has…` boolean rule** (CLAUDE.md). `hasAttemptedAllocation()` fits.

---

- **"New fuzz test fails ~25% of runs because `CAIRO_DEFAULT_SYMBOL_INDEX_TYPE` is drawn from a time-seeded RNG."** Disproven by 30 executed runs across varied `-Dfuzz.s0`/`-Dfuzz.s1` (verified the seeds propagate: `random seeds: 3L, 999L` in the log), all green. The index types that matter come from `FuzzRunner.randomIndexTypeClause()` (`:1358-1374`), which reads `SharedRandom.RANDOM` — pinned by the test's fixed seed. `randomSymbolIndexTypeName` is genuinely uniform (I measured 25%/25%/25%/25% over 20 000 seeds), but it only affects `ALTER ADD COLUMN` without an explicit `TYPE`, and the fuzz always supplies one.
- **"`Math.max(local, published)` is wrong on shrink paths (`truncate`, rollback, recovery drop)."** `truncate()` calls `initKeyMemory`, which rewrites both header pages, so published equals local (8192) and `Math.max` is a no-op; the recovery/rollback paths republish the header before close. The trim point is a lower bound on bytes that must survive — over-preserving is never read back.
- **"`keyMem.close(false)` followed by `Misc.free(keyMem)` double-closes or truncates anyway."** `MemoryCMARWImpl.close(false, …)` sets `pageAddress = 0`, `fd = -1`, `ff = null`; the subsequent `close()` short-circuits on both `pageAddress != 0` and `ff != null`. `Vm.bestEffortClose` with `size = -1` skips `ff.truncate` entirely (`Vm.java:45-55`).
- **"The header read can go out of bounds on the ordinary `of()` paths."** `isInit=false` rejects `keyFileSize < KEY_FILE_RESERVED` (`:1242`) and maps that size; `isInit=true` has `initKeyMemory` auto-extend past 8192 before `of()` returns, and every intermediate throw routes to `keyMem.close(false)`. Only the narrow failure window in M1 escapes, which is why M1 is Moderate and not Critical.
- **"Close-time read can SIGBUS on a file another writer shrank."** The `KEY_FILE_RESERVED` floor means no close path trims below 8192, so the two header pages are always backed. The one exception is the same M1 window.
- **"The 12 mapped loads per close are a meaningful performance cost."** Magnitude: 12 8-byte loads over two already-resident pages per close, versus the 5 syscalls (`fstat`, `memset`, `munmap`, `ftruncate`, `close`) the same method already performs — under 0.5% of existing per-close cost. The stale branch also *removes* `allocateDiskSpace + mremap + madvise + ftruncate`. Net-neutral; not reportable. The residual is a ≤2-page fault tail on a cold cache, bounded and off any query path.
- **"Not truncating leaves `.pk` files growing unboundedly."** Self-healing: the next `of()` maps `ff.length(keyFile)`, so the following close takes the `setSize` branch and trims. Bound at any instant ≤ one append page (512 KiB default) per affected `.pk`.

---

## Adjacent findings (not blocking — file as GitHub issues)

**A1 — `close()` can truncate a `.pk` below the 8192-byte reserved header window**
- **Net impact:** Index unopenable after an ENOSPC during rollback-truncate.
- **Location:** `PostingIndexWriter.java:477` (`Misc.free(keyMem)` → `close(true)`)
- **Symptom:** After the M1 failure window, the `finally` truncates the `.pk` to `ceilPageSize(appendOffset)` = 4096. Every subsequent open then fails with `"Index file too short"` (`:1242`, `PostingIndexUtils.java:1114`, `:1236`) — the index requires a REINDEX.
- **Reachability:** Observed in my M1 probe on **both** BASE and HEAD (`PROBE: truncate fd=… size=4096` on both), so this PR neither introduces nor worsens it.
- **Suggested fix:** floor the close-time truncate at `KEY_FILE_RESERVED` regardless of `appendOffset`.
- **Severity if filed standalone:** Moderate.

**A2 — `close()` may shrink a `.pk` below its on-disk length while a reader has it mapped**
- **Net impact:** Possible SIGBUS in a query worker.
- **Location:** `PostingIndexWriter.java:474` (`setSize` branch) → `Vm.bestEffortTruncate`
- **Symptom:** `AbstractPostingIndexReader` maps grow-only (*"File can only have grown when the chain advanced"*, `:701`) and bounds reads against its stale mmap size; a close that shrinks the file below that faults those pages past EOF.
- **Reachability:** `TableWriter.java:7074-7100` performs the same trim and explicitly floors at `keyFileSize` for exactly this reason, citing SIGBUS — `close()` does not. Pre-existing at BASE; this PR narrows the window by raising `liveSize`.
- **Suggested fix:** adopt `TableWriter`'s floor — `keyMem.setSize(Math.max(liveSize, keyMem.size()))`. Note this would also subsume C1 and M2.
- **Severity if filed standalone:** Moderate.

---

## Coverage map

| # | Change (symbol + branch) | Test + failure link | Dimensions | Verdict |
|---|---|---|---|---|
| a | `close()` retains a newer published V2 extent that fits inside the mapping | `PostingIndexCriticalIssuesTest.testCloseDoesNotTruncateChainAppendedByAnotherWriter` asserts `genCount == 1` after reopen. **Executed at BASE:** `expected:<1> but was:<0>` | happy ✓ · error N/A (no new error branch) · NULL N/A (byte offsets, no sentinel) · boundary ✓ (empty chain, first generation) · concurrency ✓ (serialized overlap) · cleanup ✓ (`assertMemoryLeak`) | **TESTED** |
| b | Published extent exceeds the mapping → release without extend/truncate | `testCloseExtendFailureDoesNotTruncateExternalChain` asserts file length unchanged. **Executed at BASE:** `expected:<12288> but was:<8192>` | happy ✓ · error ✓ (armed facade, though see Minor 4) · NULL N/A · boundary ✓ · concurrency ✓ · cleanup ✓ | **TESTED** |
| c | End-to-end WAL squash + reseal does not truncate below the published limit | `WalWriterFuzzTest.testPostingIndexCloseAfterPartitionSplitSquash` asserts `truncateBelowPublishedLimitCount == 0`. **Executed at BASE:** `expected:<0> but was:<1>` | happy ✓ · concurrency ✓ · cleanup ✓ · guard assertion vacuous (M3) | **TESTED** |
| d | **V3 (covering) header → guard inert** | No test. Neither unit test configures covering, so `raiseFormatVersionFor` never fires; the fuzz detector hard-codes `== V2_FORMAT_VERSION` (`:1088`) so it is blind too. Searched: `rg "configureCovering\|V3_FORMAT_VERSION\|isSupportedFormatVersion"` over both test files → no hits around close | all uncovered | **UNTESTED — Critical (C1)** |
| e | `readUnderSeqlock` returns `false` → fall back to local limit | No test. `rg` over the test tree for a zeroed/torn header at close → no hits | error path uncovered | **UNTESTED — Moderate (M2)** |
| f | Mapping smaller than the 8192-byte header window | No test. Proven reachable by my probe; no repo test covers it | error path uncovered | **UNTESTED — Moderate (M1)** |
| g | `setSize(published)` branch where `published > local` **and** fits the mapping | No test on 4K/16K-page hosts: in both unit tests the stale mapping is exactly `KEY_FILE_RESERVED`, so any published entry exceeds it and both land in `close(false)`. Which branch runs is OS-page-size dependent | internal branch uncovered | **UNTESTED — Moderate** |
| h | Normal single-writer close (`local == published`) leaves file size unchanged | Verified delta: `Math.max` is a no-op and `liveSize <= keyMem.size()` always holds, so the `setSize` path is byte-identical to BASE. Every posting/covering suite exercises it implicitly (all CI jobs green) | all N/A — no behavioral change | **EXEMPT** |

---

## Summary

- **Verdict: request changes.**
- **Correctness gate:** one Critical remains open (C1). Both audit directions were run and are recorded. **Downward:** each Critical was re-tested against the symptom rule — the performance, disk-space, TOCTOU and double-close claims failed it and are in Downgraded; only C1 survives, with executed evidence. **Upward:** each Moderate was re-checked for a completable symptom sentence. M1's sentence *is* completable (crash), so I promoted it as far as the evidence allows and stopped at Moderate on the stated basis that its population requires a non-default `cairo.writer.data.index.key.append.page.size` below 8192 — a supported production key that is nonetheless never below 8192 by default (512 KiB) or in the harness (16384) — conjoined with an allocation failure. M2 is incomplete-hardening: its sentence is completable by design because it was already completable at the merge base, so it does not promote. M4/M5 cap at Moderate because their only population is the team.
- **Test gate: fails.** Row (d) is an UNTESTED user-visible behavior marked Critical by the 2.6 gate. Rows (e), (f) and (g) are UNTESTED internal branches (Moderate, non-blocking). The PR does ship real regression tests with verified failure links — I executed all three at BASE and all three go red — so the gate fails only on the V3 gap.
- **Coverage map: 8 behavioral changes, 3 tested, 4 UNTESTED, 1 exempt.**
- **Findings verified vs dropped:** 47 draft findings from 13 agents; 13 verified and reported, 34 dropped or merged as false positives / duplicates.
- **Submodule provenance:** no submodule pointer moved — nothing out of scope.
- **Split:** 6 in-diff (C1, M1, M3, M4, M5, Minors), 1 incomplete-hardening (M2), 0 out-of-diff-breakage, 2 adjacent. Out-of-diff-breakage is genuinely zero here and the change is well-contained: Agent 9 walked all 10 `TableWriter` callsites, the sole `ColumnIndexer` implementor, `IndexBuilder`, `RebuildColumnBase`, `TableSnapshotRestore` and `IndexFactory`, and returned SAFE for every one except the `truncate()` partial-failure window, which is M1 — the `Math.max` is a no-op whenever local and published agree, which is every single-writer close.
- **Severity distribution: 1 Critical, 5 Moderate, 6 Minor.**

**Regressions and tradeoffs, stated plainly.** The change is real and does what it says for V2 files — I confirmed the defect survives #7386 and that all three new tests go red at the merge base. Its costs are honest and small: 12 extra mapped loads per close on a path that already makes five syscalls, a bounded ≤2-page fault tail, and up to one append page of untrimmed `.pk` slack that the next reopen reclaims. Against that, three things are worse than the body claims: the fix does not reach covering indexes at all (C1), the close path can now dereference an undersized mapping where it previously threw cleanly (M1), and the description narrates a failure signature that #7386 already eliminated — the live failure mode on master is silent index emptying, which is a stronger argument for this PR than the one the body makes.

sklarsa:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=262409)
**Tested Commit:** `37acc879b0619de57d825d109531d7c1c22a7f85`

ideoma:
#### [PR Coverage check]

:heart_eyes: **pass** : 907 / 958 (94.68%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/FirstValueDoubleWindowFunctionFactory.java|6|12|50.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/MaxMinWindowFunctionFactoryHelper.java|4|6|66.67%|
|:large_blue_circle:|io/questdb/griffin/engine/window/WindowRecordCursorFactory.java|2|3|66.67%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/FirstValueWindowFunctionFactoryHelper.java|8|12|66.67%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/NthValueDecimalWindowFunctionFactory.java|30|42|71.43%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/LastValueWindowFunctionFactoryHelper.java|6|8|75.00%|
|:large_blue_circle:|io/questdb/cairo/lv/LiveViewCheckpointTimelineStoreWriter.java|11|13|84.62%|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|25|29|86.21%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/FirstValueDecimalWindowFunctionFactory.java|78|90|86.67%|
|:large_blue_circle:|io/questdb/cairo/lv/LiveViewCheckpointOutputKeyDomain.java|7|8|87.50%|
|:large_blue_circle:|io/questdb/cairo/lv/LiveViewWindow.java|31|34|91.18%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/GenerateSeriesTimestampStringRecordCursorFactory.java|18|19|94.74%|
|:large_blue_circle:|io/questdb/cairo/lv/LiveViewRefreshJob.java|78|79|98.73%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/BooleanFunctionMemoizer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/GenerateSeriesTimestampRecordCursorFactory.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/IPv4FunctionMemoizer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/MaxDateWindowFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/LastValueDecimalWindowFunctionFactory.java|60|60|100.00%|
|:large_blue_circle:|io/questdb/std/Hash.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|38|38|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/PartitionStateEvictor.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/AvgDoubleWindowFunctionFactory.java|16|16|100.00%|
|:large_blue_circle:|io/questdb/cairo/SymbolMapReaderImpl.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/ByteFunctionMemoizer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/StrFunctionMemoizer.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/AbstractStdDevDoubleWindowFunctionFactory.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/FloatFunctionMemoizer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/lv/LiveViewCheckpointTimelineStoreReader.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/TimestampFunctionMemoizer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/VarcharFunctionMemoizer.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/cairo/SymbolMapWriter.java|15|15|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/Long256FunctionMemoizer.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/NthValueDateWindowFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/BasePartitionedBivariateWindowFunction.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/NthValueWindowFunctionFactoryHelper.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/DoubleFunctionMemoizer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/FilterOnExcludedValuesRecordCursorFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlCodeGenerator.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/CharFunctionMemoizer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/AvgDecimalRescaleWindowFunctionFactory.java|42|42|100.00%|
|:large_blue_circle:|io/questdb/cairo/frm/file/ContiguousFileIndexedFrameColumn.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/RebuildColumnBase.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/map/MapFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/IntFunctionMemoizer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/TableReader.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/NthValueLongWindowFunctionFactory.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/DateFunctionMemoizer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/AbstractPostingIndexReader.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingGenLookup.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/SymbolFunctionMemoizer.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/MaxDoubleWindowFunctionFactory.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/BasePartitionedWindowFunction.java|25|25|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/ArrayFunctionMemoizer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/RowNumberFunctionFactory.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/lv/LiveViewCheckpointPartitionMapReader.java|18|18|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/LongFunctionMemoizer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/NthValueTimestampWindowFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/CountFunctionFactoryHelper.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/FirstValueLongWindowFunctionFactory.java|12|12|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/VirtualFunctionRecordCursor.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/SumDecimalWindowFunctionFactory.java|42|42|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/table/SelectedRecord.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/DecimalFunctionMemoizer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/NthValueDoubleWindowFunctionFactory.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/MaxTimestampWindowFunctionFactory.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/GenerateSeriesLongRecordCursorFactory.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/MaxDecimalWindowFunctionFactory.java|54|54|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/ShortFunctionMemoizer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/memoization/UuidFunctionMemoizer.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/KSumDoubleWindowFunctionFactory.java|17|17|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/AvgDecimalWindowFunctionFactory.java|36|36|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalReader.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/LastValueDoubleWindowFunctionFactory.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/MaxLongWindowFunctionFactory.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/AbstractBivariateStatWindowFunctionFactory.java|6|6|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/DirectCharSequenceIntHashMap.java|24|24|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/window/LastValueLongWindowFunctionFactory.java|8|8|100.00%|

ideoma:
#### [PR Coverage check]

#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|4|4|100.00%|

RaphDal:
Superseeded

PR #7386, opened 2026-07-10
fix(core): fix POSTING-indexed WAL table suspending after a partition squash

ideoma:
## Problem

A WAL table with a POSTING index could be suspended mid-apply during a partition squash. Under parallel WAL apply, `PostingIndexWriter.of` reopens a column's `.pk` key file and sized the mmap from `ff.length(keyFile)`. That reported length can lag the writer that just extended the file, so the chain's head entry fell outside a short mapping and `openExisting` read past it -- an `AssertionError` in `AbstractMemoryCR.addressOf` under `-ea` (a SIGSEGV in a `-da` build). The reseal failed, the writer distressed, and the table suspended. It surfaced as a `WalWriterFuzzTest.testConvertPartitionToParquet` failure ("Column count must be same expected:<13> but was:<14>", the downstream symptom of the halted apply).

This branch carries two related reopen-path fixes plus a small cleanup. Both fixes are reopen-only: neither alters how the `.pk` is written, nor the on-disk format.

## Fix 1 -- size the reopen mapping to cover the header's live region, not the file length

`PostingIndexWriter.of(isInit=false)` still seeds the mapping from `ff.length(keyFile)`, but no longer trusts that length as the live-data high-water. It reads the header's `regionLimit` (the entry-region high-water) through a new `PostingIndexChainWriter.peekRegionLimit`, and extends the mapping to it -- `keyMem.jumpTo(regionLimit)` -- only when the reported length genuinely lags (`regionLimit > keyFileSize`). `regionLimit` and `headEntryOffset` come from the same seqlock-consistent header read, so the head entry `[headEntryOffset, regionLimit)` is covered by construction, independent of the (laggable) file length.

On the dominant path a cleanly-closed `.pk` has length `ceilPageSize(regionLimit) >= regionLimit`, so the mapping already covers the entry region and nothing is remapped -- no extra `mremap`/`fstat`/`madvise` per reopen. The mapping is only ever grown, never shrunk.

`peekRegionLimit` range-checks the header before its `regionLimit` is trusted to size a mapping: format version, region descriptors (`regionBase >= KEY_FILE_RESERVED`, `regionLimit >= regionBase`), and an in-region head pointer (`regionBase <= headEntryOffset < regionLimit`). These are the same checks the positional-pread reader `PostingIndexUtils.readSealTxnFromKeyFdTriState` already enforces; a corrupt-but-version-consistent header is rejected with a `CairoException` instead of being read past the mapping.

## Fix 2 -- do not truncate the `.pk` when a reopen fails before the region is republished

On the reopen path `openExisting` is what publishes the real `regionLimit` into the chain; until it runs the chain still holds the reset sentinel (`KEY_FILE_RESERVED` = 8192). The old `catch` fell into the normal truncating `close()`, which sizes its `.pk` trim from that sentinel -- so a throw from `keyMem.of` / `peekRegionLimit` / `jumpTo` / `openExisting` shrank the file back to 8192 and discarded the live chain region `[8192, regionLimit)` that is physically on disk. A merely transient reopen failure (e.g. a disk-space error while extending the mapping) would corrupt the on-disk index.

The `catch` now releases `keyMem` without truncation (`keyMem.close(false)`), leaving the on-disk index intact for the next open; the follow-up `close()` skips the already-closed `keyMem` and still frees `valueMem` / native buffers and resets state. As a side effect this also removes a latent SIGBUS hazard, since the old path could `ftruncate` a `.pk` still mmap'd by concurrent readers.

## Fix 3 -- remove a redundant cast (cleanup)

`sealIncremental` had `Math.max((long) maxDirtyStrideTotal, 1024L)`; `maxDirtyStrideTotal` is an `int` that auto-widens to `long` for `Math.max(long, long)`, so the cast was redundant. No behavioral change.

## Scope / what was checked

- Audited every posting/bitmap index mapping that fetches a file length. Only this reopen was affected. The reader (`AbstractPostingIndexReader` + `PostingIndexChainPicker`) already bounds-checks each read and grows-and-retries; the `PostingIndexUtils` seal-txn/header helpers use positional preads (safe past EOF); `.pv`/`.pc`/`.pci` mappings are sized from header-published sizes; the rename / convert-to-parquet hard-link path reads a committed, non-active source under the exclusive write lock (no concurrent extender).
- The reopen's two-read design (peek `regionLimit`, then `openExisting`) is sound only under the single-writer-per-`.pk` invariant: the reseal reopen runs on the exclusive TableWriter thread after every O3 worker has joined (`o3PartitionUpdRemaining == 0`), so no live writer mutates the header between the two reads and concurrent access is readers-only.
- `PostingIndexChainWriter.openExisting` still has no internal bounds check -- its safety is a caller contract (the caller must size the mapping to cover `regionLimit`). All three callers honor it. A future caller that maps by a raw file length would reintroduce the issue; that is not addressed here.

## Test plan

- Added `WalWriterTest#testPostingIndexResealAfterSplitSquashReadsShortKeyFile` (Fix 1): drives the real split -> squash -> native reseal and stages the lagging file length with a `FilesFacade`. It reproduces the original crash (identical stack) without the fix and passes with it; an indexed lookup on the rebuilt index (pinned to the index scan via a plan assertion) returns the expected count.
- Added `PostingIndexCriticalIssuesTest#testReopenExtendFailureDoesNotTruncateKeyFile` (Fix 2): arms `allocate` to fail during the `jumpTo` extend so the reopen throws with `keyMem` still open and the chain at the sentinel; asserts the `.pk` is not shrunk and a clean reopen recovers the chain intact. Reverting Fix 2 truncates the file and fails the assertion.
- Added `PostingIndexCriticalIssuesTest#testReopenDoesNotRemapWhenLengthCoversRegion` (dominant path: zero `mremap` on a clean reopen) and `#testReopenRejectsShortKeyFile` (a `.pk` shorter than the reserved header window is rejected with "Index file too short").
- Added `PostingIndexChainWriterTest#testPeekRegionLimit*` (7 unit tests) covering `peekRegionLimit`'s accept/reject branches: bad format version, inverted region, region base below the reserved window, out-of-region head, unreadable header, and the empty / non-empty chain shapes.
- Posting/covering index suites pass locally: `PostingIndexCriticalIssuesTest`, `CoveringIndexTest`, `PostingIndexStressTest`, `PostingIndexOracleTest`, `PostingSealPurgeTest`, `PostingIndexConcurrencyTest`, `PostingIndexNativeTest`.


ideoma:
## Code review (level 2) -- approve

Ran a multi-agent level-2 pass (correctness, concurrency, performance, resource management, tests, code quality). The core fix is correct, leak-free, race-free, and performance-neutral, and both regression tests genuinely fail without the fix. No Critical or blocking findings.

The PR description was also updated to match the shipped code -- it previously described the earlier header-window-map design rather than the file-length-seed plus conditional `jumpTo` extend that actually ships, and it now documents each bundled fix (mapping fix, no-truncate guard, cast cleanup) separately.

### Why it's sound

- **Correctness:** the head entry always occupies `[headEntryOffset, regionLimit)`, so a mapping sized to `regionLimit` covers it by construction. `peekRegionLimit`'s region/head checks mirror the existing `PostingIndexUtils.readSealTxnFromKeyFdTriState` pread reader.
- **Concurrency:** the peek-then-`openExisting` two-read design is safe under the single-writer-per-`.pk` invariant -- the reseal reopen runs on the exclusive TableWriter thread after O3 workers join (`o3PartitionUpdRemaining == 0`), so the header is immutable between the two reads. Fix 2 additionally removes a latent reader-SIGBUS: the old path could `ftruncate` a `.pk` still mmap'd by concurrent readers.
- **Resource management:** no leak or double-free on any throw path. `keyMem.close(false)` releases the mapping without truncation and the follow-up `close()` skips the already-closed `keyMem`, then frees `valueMem` / native buffers and resets state.
- **Performance:** zero extra syscalls on the dominant path (`regionLimit <= keyFileSize`, no remap); strictly cheaper than a header-window-then-`jumpTo` design.

### Minor / optional (none blocking)

1. **`peekRegionLimit` bounds the head pointer, not the head entry** (`PostingIndexChainWriter.java:488-490`). It rejects `headEntryOffset` outside `[regionBase, regionLimit)` but not `headEntryOffset + entrySize > regionLimit`. A corrupt-but-version-consistent header could read up to `entrySize-1` bytes past the mapping -- as little as 1 byte of page slack above `regionLimit` on the lag path -- a SIGSEGV instead of a clean `CairoException`. Corruption-only, unreachable for well-formed indexes, and the pread reader has the same gap, so not a regression. Optional hardening: also check `headEntryOffset + V2_ENTRY_HEADER_SIZE <= regionLimit`.
2. **Untested branch:** the `headEntryOffset < regionBase` disjunct of the head check has no test -- dropping it survives all 7 `peekRegionLimit` unit tests (not a crash; `head < regionBase < regionLimit` stays inside the mapping). Add a case with `regionBase=16_384, regionLimit=24_576, head=12_288`.
3. **Lag-path `jumpTo` success is integration-tested only** (`WalWriterTest`); a focused `PostingIndexWriter` unit test would give unit-level protection for the successful-extend path.
4. **Two tests do not discriminate this change:** `testReopenRejectsShortKeyFile` guards the pre-existing `< KEY_FILE_RESERVED` check, and `testReopenDoesNotRemapWhenLengthCoversRegion` also passes on master (master maps `ff.length` and never remaps). Fine to keep, but their protective value is narrower than the comments imply.
5. **Test duplication:** the `.pk`-fd-tracking `FilesFacade` scaffold is copy-pasted between the two new reopen tests; the file already has named `Pk*Facade` helpers -- extract a shared `PkFdTrackingFacade` base.
6. **Style:** `PostingIndexChainWriterTest.java:333,334,350,351` use `16384L`/`12288L` -- should be `16_384L`/`12_288L` per the 5-digit-underscore rule.
7. **Comments:** `testReopenExtendFailureDoesNotTruncateKeyFile`'s lead comment plus per-override restatements can be trimmed; `PostingIndexChainWriter.java:466-469` slightly overstates why an oversized `regionLimit` is safe, and `:461-462` is passive voice.
8. **Error-path exception masking** (`PostingIndexWriter.java:1175-1181`): if `keyMem.close(false)` itself throws, the original open failure is lost. Very low severity (munmap/close rarely throw). Optional: log/swallow so the original exception surfaces.

**Verdict: approve.** Coverage map: 8 behavioral rows tested, 1 untested Minor branch, 1 exempt (the `Math.max` cast removal is behavior-identical). All 8 items above are optional and cheap to fold into this branch.

ideoma:
## Code review (level 3 — full mission-critical pass)

A well-constructed, correct fix. Three bundled changes (reopen-mapping sizing, no-truncate-on-failed-reopen, a cast cleanup), all reopen-only with no on-disk-format change. Multi-agent pass + per-finding source verification found **no Critical issues, no data-loss or leak defects, and both bug fixes carry genuine regression tests** (verified to fail on revert). The items below are hardening and polish worth folding into this PR while it's open.

### Critical
None.

### Moderate

**M1 — `peekRegionLimit` + `openExisting` take two independent seqlock snapshots; head-coverage is guaranteed by an external invariant, not structurally.** *(in-diff; hardening, not a live bug)*
`PostingIndexWriter.of:1114` reads the header (snapshot A) and sizes the map to `regionLimit_A`; `openExisting` at `:1128` (`PostingIndexChainWriter.java:409`) reads the header **again** (snapshot B) and reads the head entry at `headEntryOffset_B`. The doc comment and PR body sell this as "covered by construction," but the construction actually depends on *no concurrent publisher of this `.pk` between the two reads*. That invariant is airtight at every current callsite (O3 workers use basket-private instances gated by `partCounter`; reseal runs after the `o3DoneLatch` join), so **this is not a live bug** — and the PR body itself notes "a future caller that maps by a raw file length would reintroduce the issue." Cheap, strictly-better fix: have `peekRegionLimit` hand its already-populated `headerScratch` to a snapshot-consuming `openExisting` overload (it already fills every field `openExisting` needs), collapsing to one seqlock read — makes coverage structural *and* removes the redundant second header parse.

**M2 — `testReopenDoesNotRemapWhenLengthCoversRegion` does not discriminate this PR from master, and its comment states a false history.** *(in-diff, test-quality)*
`PostingIndexCriticalIssuesTest.java:153` asserts zero `.pk` mremaps on a clean reopen. But master already mapped the **whole file** (`keyMem.of(..., keyFileSize, ...)`) and never called `jumpTo`, so master also does zero mremaps on a clean reopen — the assertion passes identically on both trees. The comment ("Before M2 the reopen mapped a fixed 8192 window and then `jumpTo(regionLimit)` forced an mremap on every non-empty reopen") describes a transient mid-branch iteration, never a shipped state on master. The test still has value as a forward guard and its `getKeyCount() > 0` assertion covers dominant-path correctness — reframe the comment as a forward guard and drop the false master-regression narrative.

**M3 — S3 post-`openExisting` throw path is untested.** *(coverage gap)*
The C1 test covers a *pre-*`openExisting` throw (allocate fails during `jumpTo`). No test covers a throw *after* `openExisting` has published the real `regionLimit` (e.g. `valueMem.of`/`allocateNativeBuffers` failing at `PostingIndexWriter.java:1148/1160`). That path has a real behavioral delta: new code leaves the `.pk` untrimmed and larger than `regionLimit`, where old code trimmed to the now-real `regionLimit`. It is self-healing (next clean open trims via `close()`) and cleanup is leak-free, so **Moderate not Critical** — but it's an error path with an observable change and zero coverage. Add a test that fails `valueMem.of` after `openExisting`, asserts the `.pk` is intact, and that a clean reopen both recovers the chain and trims to `ceilPageSize(regionLimit)`.

### Minor

**m1 — Misleading `"could not open posting index"` log on `peekRegionLimit`/`jumpTo` failure.** *(in-diff)* `kFdUnassigned = false` now sits at `PostingIndexWriter.java:1119`, after `peekRegionLimit` (1114) and `jumpTo` (1116). A throw from either fires the `LOG.error("could not open posting index")` at 1183 even though the fd was assigned and the file mapped. Diagnostic-only. Fix: move `kFdUnassigned = false;` to immediately after `keyMem.of(...)` at line 1113.

**m2 — 5-digit literals without underscore separators.** *(test style)* `PostingIndexChainWriterTest.java` uses `16384L` and `12288L` (lines ~472, 488–489). These need separators per project convention: `16_384L` / `12_288L` — better, express as `2 * KEY_FILE_RESERVED` / `1.5 * KEY_FILE_RESERVED`, matching how the neighboring tests already reference the named constant.

**m3 — `peekRegionLimit` Javadoc overclaims parity with the pread reader.** *(in-diff, doc accuracy; corruption-only)* Three imprecise claims, all reachable only under header corruption and all parity-or-better than pre-PR: (a) "the same checks the pread reader enforces" — the pread `readSealTxnFromKeyFdTriState` reads only `sealTxn` (16 bytes, EOF-safe), whereas `openExisting` reads the full variable-length entry from the mmap, so `peekRegionLimit` bounds the head's *start* (`< regionLimit`) but not its *end* (`headEntryOffset + entrySize <= regionLimit`); (b) `!= V2_NO_HEAD` (exactly `-1`) diverges from the pread reader's `< 0` (any negative → empty); (c) "consistent-but-oversized value is caught downstream by the allocation failing in `jumpTo`" is unreliable — on a sparsely-growing FS the allocate succeeds, zero-extends the file, and `openExisting` restores a zero-filled head. Fix: correct the Javadoc; optionally add a cheap `headEntryOffset + V2_ENTRY_HEADER_SIZE <= regionLimit` bound.

**m4 — Two `peekRegionLimit` mutation-survivors untested.** *(coverage)* The `head < regionBase` disjunct (`PostingIndexChainWriter.java:489`) and the `head == regionBase` lower-boundary *acceptance* are unpinned — a mutation of `<` to `<=` on either bound goes uncaught. Add two tiny unit tests (head below regionBase → reject; single-entry chain with head exactly at regionBase → accept).

**m5 — WAL repro's anti-vacuity guard is a proxy.** *(test-quality)* `sabotageClampCount > 0` proves the length clamp fired *somewhere*, not that it fed the reseal reopen's `ff.length(keyFile)` specifically. The load-bearing discriminators (`assertFalse(isSuspended)` + the two counts) do fail on the bug under `-ea`, so the test is sound; the guard is just softer than its comment claims. Optional hardening.

**m6 — Minor test-comment bloat / duplication.** *(test-quality)* The `testReopenExtendFailureDoesNotTruncateKeyFile` header is a ~14-line two-paragraph essay whose second paragraph restates what the per-phase comments already cover; trim to one paragraph. The `.pk`-fd-tracking `TestFilesFacadeImpl` subclass is duplicated between the two `PostingIndexCriticalIssuesTest` tests — an optional shared nested helper would remove the copy. Both low priority.

### Considered and dismissed (false positives / overstatements)

- **"The two-read design reintroduces the exact SIGSEGV this PR fixes" (as a live race)** — no concurrent publisher of the same `.pk` exists at any callsite (single-writer-per-(column,partition) is proven), so the race is unreachable in production. Retained only as the M1 hardening suggestion.
- **"Oversized `regionLimit` causes silent index corruption" (as a live concern)** — reachable only via a corrupt-but-consistent header; parity with the already-shipped pread reader, and the PR strictly *tightens* validation vs master (which had none before `openExisting`). Folded into m3.
- **`peekRegionLimit` spuriously rejecting legitimate headers (regression risk)** — `regionBase` is always `V2_ENTRY_REGION_BASE == KEY_FILE_RESERVED == 8192` and published heads always satisfy `regionBase <= head < regionLimit`, so the new checks reject only genuine corruption; byte-identical to the production pread reader.
- **Resource leak / double-free on the new catch path** — `keyMem.close(false)` sets `fd=-1`, so `close()`'s `isOpen()` guard skips its keyMem branch; every throw site releases keyMem exactly once and `close()`'s finally frees valueMem/native buffers on all paths.

### Coverage map

| Behavioral change | Test | Verdict |
|---|---|---|
| **S1** `peekRegionLimit` validation (5 reject + 2 accept branches) | `PostingIndexChainWriterTest.testPeekRegionLimit*` (7) + both integration reopens | TESTED (2 boundary gaps → m4) |
| **S2** reopen dominant path (skip `jumpTo`) | `testReopenDoesNotRemapWhenLengthCoversRegion` (`getKeyCount() > 0`) | TESTED (mremap assertion non-discriminating vs master → M2) |
| **S2** reopen lagging path (`jumpTo(regionLimit)`) | `WalWriterTest.testPostingIndexResealAfterSplitSquashReadsShortKeyFile`, C1 | TESTED (fails on revert) |
| **S3** catch `close(false)` — pre-`openExisting` throw | `testReopenExtendFailureDoesNotTruncateKeyFile` | TESTED (fails on revert) |
| **S3** catch `close(false)` — post-`openExisting` throw (untrimmed file) | — | **UNTESTED** (M3, Moderate — benign/self-healing) |
| Short-file rejection (defense-in-depth) | `testReopenRejectsShortKeyFile` | TESTED |
| **S4** `sealIncremental` cast removal | — | EXEMPT — verified no behavioral change (`int`→`long` widening) |

### Summary

**Verdict: approve.** No Critical issues; the change is correct, resource-safe, and concurrency-safe under the (verified airtight) single-writer invariant, and both bug fixes ship with regression tests confirmed to fail on revert. Test gate passes (no UNTESTED Critical row; the one UNTESTED row is a benign self-healing error-path delta; S4 is a verified no-behavioral-change exemption). Metadata is clean. Suggest folding the low-cost items into this PR rather than a follow-up: the M1 peek/open fusion, the M2/m1/m2/m3 comment-log-style fixes, and the M3/m4 test gaps.

ideoma:
I'm going to sort ent CI failure in a separate PR, they are independent of the changes here
https://github.com/questdb/questdb-enterprise/pull/1131

bluestreak01 (changes requested):
@ideoma

## Critical (blocking): reopen silently manufactures a zero-filled chain entry from a physically truncated `.pk`

**Status: validated — empirically reproduced against this PR's head with assertions enabled.**

### Summary

The new lag branch in `openKeyAndValueMem` trusts the header's `regionLimit` over the file's physical extent. A `.pk` truncated to exactly `KEY_FILE_RESERVED` (8192) retains its two valid seqlock header pages (A@0, B@4096) and passes the `keyFileSize < KEY_FILE_RESERVED` check (strict `<`). `peekRegionLimit()` then returns the old, larger `regionLimit`, and `keyMem.jumpTo(regionLimit)` **physically zero-extends the missing entry region on disk before any validation runs**. `openExisting()` then reads the header (valid), follows `headEntryOffset` into the freshly manufactured zeroes, and `PostingIndexChainEntry.read()` ingests them with no structural validation. The writer opens "successfully" with `keyCount=0, genCount=0, maxValue=0, sealTxn=0` instead of rejecting corruption — while the partition's `.pv`/`.pc` sidecars still exist. The index silently reports empty; nothing signals that a REINDEX is needed.

### Empirical reproduction (PR head, `-ea`)

Built a 300-key index (`keyCount=300, maxValue=299`), commit + seal + close (`.pk` = 16384 B), truncated the `.pk` to exactly 8192, reopened via `of(path, name, COLUMN_NAME_TXN_NONE, 0L, 0L)`:

```text
before=16384 truncated=8192 after=16384
outcome=OPENED keyCount=0 genCount=0 maxValue=0 (pre-truncation keyCount=300 maxValue=299)
```

- `after=16384` proves the reopen itself re-extended the file with zeroes (`jumpTo` → `extend0` → `TableUtils.allocateDiskSpace`).
- A "truncated `.pk` must be rejected on reopen" assertion fails on this branch.

Repro test (drop into `PostingIndexCriticalIssuesTest` or a sibling class):

```java
@Test
public void testReopenAfterTruncationToExactlyKeyFileReserved() throws Exception {
    assertMemoryLeak(() -> {
        final FilesFacade rawFf = configuration.getFilesFacade();
        try (Path path = new Path().of(configuration.getDbRoot())) {
            final int plen = path.size();
            final String name = "trunc_8192_repro";
            try (PostingIndexWriter writer = new PostingIndexWriter(
                    configuration, path.trimTo(plen), name, COLUMN_NAME_TXN_NONE)) {
                long row = 0;
                for (int k = 0; k < 300; k++) {
                    writer.add(k, row++);
                }
                writer.setMaxValue(row - 1);
                writer.commit();
                writer.seal();
            }
            final long fd = rawFf.openRW(
                    PostingIndexUtils.keyFileName(path.trimTo(plen), name, COLUMN_NAME_TXN_NONE),
                    configuration.getWriterFileOpenOpts());
            Assert.assertTrue(fd > -1);
            try {
                Assert.assertTrue(rawFf.truncate(fd, PostingIndexUtils.KEY_FILE_RESERVED));
            } finally {
                rawFf.close(fd);
            }
            try (PostingIndexWriter reopen = new PostingIndexWriter(configuration)) {
                reopen.of(path.trimTo(plen), name, COLUMN_NAME_TXN_NONE, 0L, 0L);
                // BUG: opens with keyCount=0 genCount=0 maxValue=0 and the file
                // is silently re-extended with zeroes to the header's regionLimit.
                Assert.fail("reopen of a truncated .pk must be rejected, opened with keyCount="
                        + reopen.getKeyCount());
            } catch (CairoException expected) {
                // desired behavior
            }
        }
    });
}
```

### Mechanism (file/line, PR head)

1. `PostingIndexWriter.java:1107-1117` — lag branch: `keyFileSize = ff.length(keyFile)` = 8192 passes the strict `<` short-file check; mapping seeded at 8192; `peekRegionLimit` consulted; `regionLimit > keyFileSize` → `keyMem.jumpTo(regionLimit)`.
2. `PostingIndexChainWriter.java:471-499` (`peekRegionLimit`) — validates only header-internal consistency (format version, `regionBase >= 8192`, `regionLimit >= regionBase`, head within `[regionBase, regionLimit)`). All pass on a truncated-but-header-intact file. It never compares `regionLimit` against the file's backed extent.
3. `MemoryCMARWImpl.java:376-408` — `jumpTo` → `checkAndExtend` → `extend0` → `TableUtils.allocateDiskSpace(ff, fd, newSize)`: the missing entry region is physically allocated and zero-filled **before** `openExisting` validates anything.
4. `PostingIndexChainWriter.java:408-431` (`openExisting`) — header still valid, `headEntryOffset != V2_NO_HEAD`, so it reads the head entry from the zeroed region.
5. `PostingIndexChainEntry.java:110-160` (`read`) — no validation of `len` / `sealTxn` / structure: yields `len=0, sealTxn=0, genCount=0, keyCount=0, maxValue=0, txnAtSeal=0`.
6. The recovery walk cannot save it: `runRecoveryWalkIfRequested` no-ops on a plain reopen (`currentTableTxn == -1`), and the zero entry's `txnAtSeal=0` would never satisfy `txnAtSeal > committedTxn` anyway.

### Reachability — not hypothetical

The shape (physical extent < `regionLimit`, valid live-chain header) is minted by real channels; once on disk, consumption is automatic on the next routine reopen (writer pool cycling, WAL apply, O3):

- **This PR's own pre-Fix-2 code, proven by this PR's own test**: the `reopen_no_truncate` test demonstrates stale `ff.length` + ENOSPC in `jumpTo` → throw before `openExisting` publishes the real `regionLimit` → pre-fix truncating `close()` trims to the 8192 sentinel with header pages untouched. The non-truncating catch added in this PR stops *creating* the shape going forward but does nothing for shapes already on disk — which this same PR's lag branch then silently zero-fills.
- **Released 9.4.0/9.4.2/9.4.3 with `-ea`** (CI/dev/staging): the length-lag race this branch is named for makes master's `openExisting` read the head entry past the `ff.length`-sized mapping; `AbstractMemoryCR.addressOf`'s `assert checkOffsetMapped(...)` throws AssertionError → caught by master's plain `catch (Throwable) { close(); }` → truncating close mints the shape on a healthy index. (`-da` production instead crashes or reads page-slack by accident — loud, file intact.)
- **Power-loss crash reordering, any released build**: seal extends the `.pk` (fallocate, journaled i_size), writes the entry, flips the header via mmap; the durability barrier (`keyMem.sync(false)`) runs only after publish. Header-page writeback beating the journal commit of the size extension persists new-live-header + old-smaller-size. First seal of a partition gives exactly 8192; later seals give intermediate sizes — which the lag branch **also** zero-extends. The current check only rejects `< 8192`.
- **External** truncation (restore/copy/fs-repair) — and the suite already treats "truncated `.pk` must be rejected cleanly on reopen" as an explicit contract (`testReopenRejectsShortKeyFile`), which only covers `< 8192`.

Aggravating: the silent acceptance is **new in this diff**. On the same on-disk shape, master fails loudly (assert/SIGSEGV). This branch is the first code that manufactures the missing bytes and reports success — converting a detectable corruption into silent index emptiness, including for shapes carried into an upgrade.

### Required changes

1. In the lag branch (`PostingIndexWriter.java:1107-1117`), before extending the mapping to `regionLimit`, compare `regionLimit` with the **opened fd's actual extent** (`ff.length(fd)`), not the possibly-stale path-based length. If the backed extent genuinely covers `regionLimit` (the legitimate stat-lag case this PR targets), remap up to it — this allocates nothing. If the backed extent is short of `regionLimit`, throw `CairoException.critical` (e.g. "posting index key file truncated [expected>=regionLimit, actual=extent]") — never call `jumpTo`/`allocateDiskSpace` on the open path merely because the header advertises the bytes.
2. Validate the head entry's structure after `openExisting` reads it (defense-in-depth for the intermediate-size crash shapes): a live head entry must have `len >= V2_ENTRY_HEADER_SIZE` (56), `sealTxn >= 1`, and `headEntryOffset + len <= regionLimit`. An all-zero entry fails all three.
3. Add the exact-8192 truncation regression (test above): reopen must throw, must not grow the file, and must leave the on-disk `.pk` byte-identical (no allocation side effects on the reject path — mirror the existing non-truncating failure-path guarantees).
4. Optionally extend the regression to an intermediate truncation (e.g. truncate to a mid-region page boundary) to cover the partial-entry variant of the same acceptance bug.

Everything else about the lag-handling approach and the non-truncating failure path (Fix 2) checks out in my validation — this acceptance gap is the one blocking issue.

ideoma:
Point that the whole point of this PR is *NOT* to rely on directory entry as *MOST* of the mapping does (and it's an unwritten QuestDB rule generally). And now the review finds that it's *CRITICAL* bug not not rely on the dir entry length. Absolutely *bonkers*. Also, probability-wise, stale dir entry has *REAL* chances to occur in a running system without power loss and the suggested fix will be a *FOOTGUN* in normal *BAU*. Power loss is not handled well by QuestDB and not coded for in *MANY* places of the DB and matched file size does *NOT* protect from data corruption

bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 29 / 29 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexWriter.java|10|10|100.00%|
|:large_blue_circle:|io/questdb/cairo/idx/PostingIndexChainWriter.java|19|19|100.00%|

questdb-butler:
## ⚠️ Enterprise CI Failed

**Build:** [View Details](https://dev.azure.com/questdb/questdb-enterprise/_build/results?buildId=252208)
**Tested Commit:** `f3c26447117506e00fb87d3b8fd2a8cb0a4ffab3`

PR #1127, opened 2021-06-16
fix(cutlass): http and pg services should be default use shared pool rather than restricting thread count to small values

bluestreak01:
SQL queries executed via HTTP will use http worker pool. This pool was set to 2 workers on CPUs with high core count  (>16). This leads to incorrect load distributions between cores

PR #7387, opened 2026-07-10
chore(qwp): carry crash-safe SF cleanup and pool-capacity recovery

bluestreak01:
> **Tandem PRs:** Client — questdb/java-questdb-client#67 · Enterprise e2e — questdb/questdb-enterprise#1127

## Purpose

Land the server-side half of the QWP store-and-forward shutdown work and carry the matching client changes into the server build. The headline fix makes a server-initiated fatal WebSocket CLOSE **linger and drain** so the peer receives the final durable ACK instead of a connection reset. Validated across the QuestDB server and Enterprise CI matrices.

### Server fix — linger after a fatal CLOSE so the final ack survives (`fix(qwp)`)

A server-initiated fatal WebSocket CLOSE (role-change deferred close, frame reject, oversized frame) used to write the final cumulative/durable ACK + CLOSE, `shutdown(WR)`, drain the recv buffer once, and immediately close the fd. Against an actively-streaming client — the demote-time norm — the client's in-flight frames hit the closed fd and force a TCP **RST**, which destroys whatever the client has not yet read from its receive queue, including the final durable ack the role-change close deferral (Invariant B) exists to deliver. A store-and-forward client that never sees that ack replays every committed-but-unacked batch on the promoted replica → duplicates on tables without `DEDUP UPSERT KEYS`. (CI signature: `SqlFailoverQwpDeferredCloseExactlyOnceTest` red with `count == total + committed-at-demote`; Windows-leaning because WinSock purges the rx queue on RST.)

- `QwpIngressUpgradeProcessor.gracefulCloseAndDrain` + a `QwpIngressProcessorState` close-drain state machine (`beginCloseDrain` / `isCloseDraining` / `isCloseDrainExpired`) now keep reading and discarding the peer's pipelined frames after the fatal CLOSE until the peer closes (FIN) or a bounded deadline `CLOSE_DRAIN_TIMEOUT_MICROS` elapses — never an early fd close under in-flight bytes.
- Tests: `QwpServerCloseDrainTest` (lingers while a streaming peer drains its receive queue; bounds the linger against a live writer that never reads) and `QwpIngressProcessorStateTest.testCloseDrainLifecycle` (drain begin/expiry/reset invariants across per-message resets and connection recycle).

### Client gitlink

- Bumps `java-questdb-client` `5cc68a91e2d6` → `fee9903ca6fc` — the crash-safe SF cleanup + pool-capacity recovery client changes (see "Client behavior under test" below), plus a `closeStarted`-race fix in the closed-mid-creation lock-probe test. Client PR: questdb/java-questdb-client#67.

### Housekeeping

- Applies an IntelliJ code-formatting pass across core (nested-ternary reindentation in `Micros`/`Dates`/`Nanos`/`CharSink` and several test files; no behavioral change).
- Merges current `master` (brings in #7386, #7406).
- The PR now carries real source changes, so the earlier temporary `ServerMain` CI-trigger comment is no longer needed and has been removed — `CheckChanges` classifies this as source-changing on its own.

## Client behavior under test

The bumped client revision:

- prevents `CursorSendEngine.close()` from releasing a slot while a manager or I/O worker can still touch its ring, watermark, transport, or segment files;
- transfers safe cleanup to worker exit and wakes pool borrowers when deferred cleanup releases the slot;
- retries transient managed-slot startup recovery without requiring pool recreation;
- explicitly confirms native flock release before publishing slot reuse;
- uses a shared retry driver with capped exponential backoff for failed flock release;
- persists the final ACK watermark before close-time unlink so interrupted or partial cleanup cannot replay acknowledged frames;
- removes the watermark only after all acknowledged segment files are confirmed gone; and
- shuts down socket traffic before joining a worker blocked in native send/receive, on POSIX and Windows.

## Tandem chain

- OSS PR head: `cfece81562f0ee50a870108b1f4420d1f9b6b1bc`
- client gitlink: `fee9903ca6fc2b88d21ac63690d17debea7b6aa8`
- Enterprise PR #1127 pins `questdb@cfece81562` (this PR's head), so the ENT e2e suite runs against exactly this server + client — consistent tandem.
- Client PR: questdb/java-questdb-client#67 · Enterprise coverage: questdb/questdb-enterprise#1127

Current validation status is reported by this PR's server checks and the linked client and Enterprise checks.


bluestreak01:
## Senior-engineer review (tandem, level 3) — OSS #7387 ↔ ENT questdb/questdb-enterprise#1127

**Verdict: approve the gitlink; correct the stale PR body before merge.**

This PR is a 1-line `java-questdb-client` gitlink bump (`5cc68a91e2d6` → `a167e6503b9b`). The full review of the changeset lives on the Enterprise tandem PR (questdb/questdb-enterprise#1127); this comment covers the OSS-specific items.

### ✅ Gitlink verified consistent
- ENT #1127 pins `questdb@d078ef13e`, a merge commit that **contains** this PR's head `46e2f0e5` (`git merge-base --is-ancestor 46e2f0e5 d078ef13e` = true).
- `questdb@d078ef13e` points `java-questdb-client@a167e6503b9b` — matching this PR's bumped gitlink and the ENT PR body.
- The harness client seams the Enterprise e2e tests depend on (`setCloseStartedHook`, `getTotalFramesReplayed/Sent`, `getAckedFsn`, `setDeferCommit`, `PooledSender.slot()`→`SenderSlot.delegate()`) all exist at `a167e6503b9b`; the Enterprise sidecar compiles against it.

### 🟠 Moderate (M4) — PR body is stale/misleading (metadata)
The body pins client `cad1bb8d…` and OSS head `46e2f0e5`, and warns that a temporary `ServerMain` CI-trigger comment **"must be removed before merge."** Reality on the merged gitlink:
- client is **`a167e6503b9b`**, not `cad1bb8d…`; and
- the `ServerMain` comment was **already reverted** at `ad4b46d991`.

The gitlink content that actually merges is correct, so merge safety is fine — but a reviewer trusting the stale SHAs would diff the wrong client tree, and a merger honoring the obsolete "remove the comment" gate wastes time hunting a comment that no longer exists.

**Fix (before merge):** update the SHAs, drop the obsolete "remove before merge" warning, and link the transitive `java-questdb-client` PR (#67) so the real behavioral diff is reachable from here.

### Client behavior under test (context)
The bumped client carries the QWP store-and-forward shutdown/recovery changes (worker-quiescence close ordering, retryable managed-slot startup recovery, flock-release confirmation before slot reuse, crash-safe close-time segment cleanup, ACK-watermark persistence before unlink). These were reviewed **against the SF store-and-forward invariants** via the Enterprise e2e tests and found compliant: the close-time drain reconnect is unbounded regardless of `reconnect_max_duration_millis` (that config bounds only the initial blocking connect); a wedged/timed-out drain throws rather than silently completing; dense oracles are strict and handle at-least-once replay (durable-ack await / DEDUP); and write-gate evidence uses pg-wire frozen-count probes + post-promotion drain, never producer-visible role errors.

<sub>🤖 Automated senior-engineer review (`/review-pr --tandem --level 3`). Full changeset review on questdb/questdb-enterprise#1127.</sub>

bluestreak01:
## Review — QWP multi-slot recovery & close-drain failover (tandem)

**Tandem review** of Enterprise **#1127** + OSS **questdb/questdb#7387** (client **questdb/java-questdb-client#67** SF-contract-checked). Full mission-critical pass: 14 review dimensions across independent fresh-context reviewers (Rust N/A — no `.rs`); every load-bearing finding re-verified against source.

### What this PR is
An **e2e test PR** (ENT, ~3.5k lines of harness/tests) plus a submodule bump carrying **one OSS server behavioral change**: `gracefulCloseAndDisconnect` → **`gracefulCloseAndDrain`**, a bounded post-CLOSE read-drain so a server-initiated fatal WebSocket CLOSE lingers (FIN + keep reading/discarding) instead of closing the fd under in-flight client bytes. An fd close under in-flight bytes forces a TCP RST that destroys the final durable ACK in the peer's recv queue → SF client replays committed-but-unacked batches → duplicates on non-`DEDUP` tables.

**Tandem chain verified consistent** (despite stale descriptions — see Minor): OSS #7387 head `1898c786a6` == ENT-pinned `questdb@1898c786a6` → `java-questdb-client@9c9e1ddbda9a` == OSS transitive bump. **Client SF-contract sanity:** the steady-state drainer reconnect is **unbounded** (`reconnectMaxDurationMillis` is "NOT consulted by the background loop"; it only bounds the non-lazy initial connect — permitted); per-attempt backoff capped, retry loop unbounded — compliant with the store-and-forward invariants.

---

### Critical
**None.** The production change was deep-traced end-to-end and confirmed correct; its fix ships with an effective regression test. No confirmed correctness, data-integrity, resource, concurrency, or performance/IO defect on a reachable production path remains open. (One "Blocker" candidate was raised and dropped after verification — see **Downgraded**.)

---

### Moderate
Test-efficacy / coverage-clarity items (they do not touch production correctness/perf). All are cheap to fix in this PR.

**M1 — [ENT] `test_close_drain_failover.py` is not a regression guard for the RST-destroys-ACK bug (misattributed coverage).** The test withholds the 40-row backlog from node A via paused forward gates before demote (its own docstring: "forward gates retain the backlog before both endpoints, then A demotes while those bytes cannot reach it") and durably-acks the 30-row baseline beforehand. So **A holds no committed-but-unacked data at demote** — there is no final durable ACK for an RST to destroy. Reverting the production hunk (immediate RST) yields the identical dense `[0,70)` on B (client reconnects, replays from `ackedFsn+1`); the `close_pending`/ordering assertions are also insensitive to RST-vs-clean-CLOSE. It is a valid, well-built test of the *close-rides-failover choreography* and should be kept. Fix: either (a) add a deterministic variant where the backlog is durably **committed on A before demote** with the final durable-ACK withheld by a paused reverse gate — a reverted RST then replays committed rows and breaks the dense oracle on the non-DEDUP table; or (b) adjust the docstring so it isn't represented as the RST-fix witness and cite the actual guards (OSS `QwpServerCloseDrainTest` + pre-existing ENT `SqlFailoverQwpDeferredCloseExactlyOnceTest`). *The RST fix itself IS covered (see Coverage map) — this is an attribution gap, not an uncovered bug.*

**M2 — [OSS] Peer-close fast-exit (`recv < 0`) is executed but never asserted.** In `QwpIngressUpgradeProcessor.resumeRecv` drain branch, the `drained < 0 → ServerDisconnectException` path (the *common* conformant-client exit) is only implicitly exercised at `QwpServerCloseDrainTest.testFatalCloseLingersWhileStreamingPeerDrainsItsReceiveQueue` socket-close teardown; nothing asserts the server disconnects on the peer FIN. A regression (fall-through to would-block/park) would surface only non-deterministically via `assertMemoryLeak`/pool halt. Fix: after closing the client socket, assert the server tears the connection down promptly (poll connection count → 0, or that a fresh connect succeeds, within a bounded window).

**M3 — [OSS] The `SEND_STATE_RESUME_CLOSE` send-drain path is UNTESTED.** Arming the drain from a *partially-flushed* CLOSE frame (`resumeSend` RESUME_CLOSE → `gracefulCloseAndDrain`) needs a PeerIsSlow under a small send-fragmentation cap; `QwpServerCloseDrainTest` uses default large send buffers, so the CLOSE always leaves via the READY path. This is also the only path where the `drainBufferedFrames` `!isCloseDraining()` guard matters. Fix: add an e2e case forcing `getForceSendFragmentationChunkSize()` below the CLOSE-frame size (as `QwpUpgradeRejectFragmentationTest` already does for handshake rejects) to exercise resume-close arming + buffered-frame discard.

---

### Minor
- **[OSS] `resumeRecv` drain loop does not re-check the deadline in-loop** (optional hardening). The `while ((drained = socket.recv(...)) > 0)` discards to would-block and checks `isCloseDrainExpired()` only on re-entry. This is the **same loop shape as the pre-existing `HttpConnectionContext.drainRecvBuffer()`** and the new code yields to the dispatcher between read events, so it is self-limiting — but a per-N-iteration `isCloseDrainExpired()` recheck + bounded chunk would make the 5s budget a hard guarantee even against a continuous within-invocation flood. Not a regression.
- **[OSS] Stale javadoc on `sendDeferredFatalClose`:** still says "half-closes the write side and raises ServerDisconnect" — it now calls `gracefulCloseAndDrain` and returns into the read-drain. Update the one-liner.
- **[ENT] `QwpSidecarMain.connectPool` cleanup NPEs on null `senders`:** the `catch` does `for (Sender s : senders)` before `db.close()`; if `new Sender[count]` itself throws (pathological `CONNECT_POOL <n>` / OOM), `senders` is null → NPE masks the real error and leaks the connected `QuestDB` facade, defeating the method's own comment. Test-only + unreachable in practice (`count` validated `>=1`, test-controlled). Guard with `if (senders != null) { … }` so `db.close()` always runs.
- **[OSS] `closeDrainDeadline` defensive reset:** reset only in `onDisconnected()` (safe under current teardown wiring; mirrors the shipped `roleChangeCloseDeferredDeadline` pattern). A defensive reset in `of()` would harden against future teardown-path changes that could otherwise start a reused pooled context with `isCloseDraining()==true`.
- **[BOTH] Stale submodule commit hashes in both PR descriptions** — ENT body cites `questdb→d078ef13`/`client→a167e650`; OSS body cites `ENT pins cfece81562`/`client→fee9903ca6fc`; the actual final pins are `1898c786a6`/`9c9e1ddbda9a`. Descriptions-only; refresh before merge so provenance verification isn't misled.
- **[ENT] Scope creep:** the PR adds `.pi/skills/fix-pr/SKILL.md` (430 lines of agent tooling) + `.gitignore` `.pi-subagents/`, unrelated to the QWP change. Flag for author decision (separate tooling PR / personal config) unless the team has decided to vendor `.pi/skills/`.
- **[OSS] Duplicated test helpers:** `QwpServerCloseDrainTest.createMaskedFrame`/`performWebSocketHandshake` are byte-for-byte copies of the same helpers in `QwpWebSocketProtocolTest` (both extend `AbstractQwpWebSocketTest`). Hoist into the shared base to avoid framing-fix drift.
- **[ENT] `ReverseTrafficGate._pause()` has no timeout** — waits unbounded for an in-flight `sendall` on a backpressured peer; a misuse surfaces as a whole-suite hang rather than a clean assertion. **[ENT] `QwpSidecarMain.Client` methods not alphabetical within kind** (`close`/`closeWithWitness`/`closeQuietly`) — auto-sort reconciles.

---

### Downgraded (false positives / refuted after source verification)
- **"Blocker: post-CLOSE drain loop pins an IO worker indefinitely / DoS / regression."** *Refuted.* (1) The unbounded `while (socket.recv(recvBuffer, recvBufferSize) > 0)` is **not new** — the pre-PR `gracefulCloseAndDisconnect` ran the identical loop via `context.drainRecvBuffer()`, so no regression in loop shape. (2) The new code **yields between read events**: on would-block it `return`s → `handleProtocolSwitchedRecv` throws `registerDispatcherRead()`, so the worker services other connections between drain reads — it does not monopolize across the 5s. (3) Non-blocking `recv` reaches would-block far faster than any real network delivers (drain = memcpy ≫ link rate); the legit demote-time streamer stops and closes within ms of observing the CLOSE (→ `recv<0` exit). The "re-check deadline in-loop" idea survives as a Minor hardening, not a blocker.
- **"CLOSE/CLOSE_BEGIN on null client now replies ERR (was OK)"** — intentional test-harness behavior change, no production impact.
- **OSS "formatting" files carry hidden behavior** — refuted: `ColumnType`, `Micros`/`Dates`/`Nanos`, `CharSink`, `FunctionParser`, `SqlCodeGenerator`, `Concurrent*HashMap`, etc. (16 files) are **pure nested-ternary reindentation** — every changed line differs from its pair only in leading whitespace. EXEMPT.
- **Client bump imposes a reconnect budget on the SF drainer** — refuted: `reconnectMaxDurationMillis` is explicitly "NOT consulted by the background loop" (bounds only the non-lazy initial connect, which the SF contract permits).

---

### Coverage map (OSS QWP close-drain — the only production behavioral change)
| # | Branch / path | Test | Fails on regression? | Verdict |
|---|---|---|---|---|
| 1 | drain deadline expiry → disconnect | e2e `testCloseDrainDeadlineBoundsLingerAgainstLiveWriter` (`cutOffMillis>=3000 && <=20000`) + unit `testCloseDrainLifecycle` | Yes — revert (immediate RST) → cutoff ≈0–300ms fails `>=3000` | **TESTED** |
| 2 | `recv>0` discard loop (linger, no RST) | e2e `testFatalCloseLingersWhileStreamingPeerDrainsItsReceiveQueue` | Yes — revert → RST → follow-up write throws | **TESTED** |
| 3 | `recv==0` park → re-register read | e2e tests 1 & 2 (idle gaps) | implicit only | weak |
| 4 | `recv<0` peer-close → disconnect | e2e test 2 teardown | **not asserted** | **UNTESTED (M2)** |
| 5 | send-path RESUME_CLOSE → drain arm | — | — | **UNTESTED (M3)** |
| 6 | `sendFatalClose` arming | e2e tests 1 & 2 (TEXT reject → CLOSE 1003) | Yes | **TESTED** |
| 7 | deferred/durable-ACK-then-close arming (demote) | pre-existing ENT `SqlFailoverQwpDeferredCloseExactlyOnceTest` | Yes | **TESTED** (out-of-diff) |
| 8 | `drainBufferedFrames` skip guard | — | implicit | untested directly |
| 9 | `processWebSocketFrames` discard-rest | — | implicit | untested directly |
| 10 | `beginCloseDrain` idempotency | unit `testCloseDrainLifecycle` (deadline anchored at origin → expires at `now==TIMEOUT`) | Yes | **TESTED** |
| 11 | `onDisconnected` reset | unit `testCloseDrainLifecycle` | Yes | **TESTED** |
| 12 | `closeDrainDeadline` survives `clear()`/`clearMessageState()` | unit `testCloseDrainLifecycle` | Yes | **TESTED** |
| — | 16 OSS non-QWP files | — | pure reindentation, verified no behavior | **EXEMPT** |

---

### Summary
**Verdict — ENT #1127: approve · OSS #7387: approve**, with the Moderate coverage clarifications (M1–M3) and Minors recommended for THIS PR (all cheap; not deferred).

- **Correctness & performance gate: PASS.** The OSS close-drain change was traced end-to-end (recv semantics; recv- and send-path re-registration; state survival across `reset()`/`clear()`; pooled-reuse reset via `onConnectionClosed`→`onDisconnected`; idempotent arming; all fatal-close paths arm the drain incl. the demote durable-ACK path; FIN-after-CLOSE ordering; ~zero added steady-state cost). No open production defect. The single "Blocker" was refuted (pre-existing loop shape + yields between events + self-limiting).
- **Test gate: PASS.** The fix ships with a regression test whose failure link is verified by revert-reasoning: OSS `QwpServerCloseDrainTest` fails pre-fix (immediate RST → cutoff `<3000` / IOException). The state machine has a strong non-vacuous unit test. The ENT e2e suite complies with the SF store-and-forward test-application checklist (no producer-visible role errors as write-gate evidence; replay-aware oracles via withhold-from-A construction and a `DEDUP` table; acknowledged-boundary ordering, not timing).
- **Findings:** ~14 draft findings; **1 major false-positive dropped** (drain-loop Blocker) + 3 other refutations. Split: **0 Critical · 3 Moderate · ~8 Minor**. Essentially all **in-diff**; the drain-loop shape is **pre-existing** (`drainRecvBuffer`) and the demote regression witness (row 7) is a pre-existing out-of-diff test. Cross-context (egress processor, framework re-registration, pooled-context reuse, client SF drainer) all cleared.
- **Highest-value residual:** M1 — attribute the RST-fix regression coverage correctly (OSS `QwpServerCloseDrainTest` + pre-existing `SqlFailoverQwpDeferredCloseExactlyOnceTest`), and consider the committed-on-A-before-demote variant so the ENT deterministic test also discriminates the fix.

<sub>🤖 Generated with tandem PR review (level 3).</sub>

bluestreak01:
## Review — tandem: ENT #1127 + OSS questdb#7387 + client java-questdb-client#67 (level 3)

**QWP store-and-forward: multi-slot recovery & close-drain failover**

This is a **three-repo tandem change** reviewed at level 3 (all agents, per-finding source verification). The ENT PR authors **only test/harness code + a submodule bump**; the real production change is the OSS server *close-drain linger* fix (#7387) plus the nested `java-questdb-client` store-and-forward machinery (#67), pulled in via the `questdb` submodule pointer.

**Verdict: approve (ENT) / approve (OSS).** No Critical or reachable-path correctness, data-loss, resource, or performance findings. 2 non-blocking robustness nits + metadata. 8 draft findings verified, 6 dropped as false positives / non-reachable.

---

### Critical
**None.**

The headline data-loss fix is correct and well-tested. Verified end-to-end:

- **OSS close-drain state machine** — `closeDrainDeadline` is reset **only** in `onDisconnected()`, *not* in per-message `clear()`/`clearMessageState()`, so the linger survives per-message resets (matches the field's documented contract). `beginCloseDrain()` is idempotent (a late re-entry cannot extend the deadline). The drain discards pipelined pre-CLOSE frames without touching the engine (`drainBufferedFrames` + `processWebSocketFrames` gated on `!isCloseDraining()`), so no double-commit/mis-order. Peer close (`recv < 0`) → disconnect; would-block → parked; the 32-recv quantum yields via `PeerIsSlowToWriteException` so a continuously-readable peer can't monopolize the worker or outlive the deadline.
- **Final durable ACK is ordered before FIN** — `sendFatalClose`/`finishDeferredFatalClose` flush the cumulative + durable ACK before `shutdown(SHUT_WR)`, so the drain genuinely protects the ack the whole teardown exists to deliver.
- **Client reconnect containment (the core SF contract)** — the steady-state `CursorWebSocketSendLoop` retries transient failover (role reject, transport, version mismatch) **indefinitely** with no wall-clock cap (Invariant B); only auth / non-421 upgrade / capability-gap exit, with un-acked rows preserved in SF. The orphan `BackgroundDrainer` repurposes `reconnectMaxDurationMillis` as a **capability-gap-only settle budget** (gap-anchored; role reject *resets* the episode; transport only *pauses* the wall clock — anti-evasion-correct). So `reconnect_max_duration_millis=300000/60000` in the test connect strings is **not** a banned reconnect budget.
- **Client close/pool/watermark ordering** — no slot release while a worker can still touch its ring/watermark/transport/segment files; watermark removed only after acked segments are unlinked; flock published for reuse only after confirmed native release; unacked rows never silently dropped (left on disk for the next run / orphan drainer).
- **Test efficacy is real** — the e2e tests use *correct* write-gate evidence (pg-wire frozen-count probes, `close_pending()` verb-free selects, dense-sequence oracles, successor `startupSent==0` / no-residual-`.sfa` witnesses) — **no banned producer-visible role errors**. The dense oracle does not `DISTINCT`, so it catches replay duplicates via length mismatch. Replay accounting is sound: `test_close_drain_failover` durably-acks the seed before the disturbance and gates the backlog to exactly-once; `test_pool_multislot_recovery` and `test_sf_producer_kill9_cycles` use a `DEDUP UPSERT KEYS` table. `_wait_successor_replayed` (exact `startup_sent - replayed == published - acknowledged`) is the load-bearing recovery witness and is itself unit-tested with synthetic stats.
- **Regression link verified** — `QwpServerCloseDrainTest.testFatalCloseLingersWhileStreamingPeerDrainsItsReceiveQueue` fails pre-fix (pipelined writes hit the RST'd fd → `IOException`); passes post-fix.
- All three new close-drain `LOG` chains close with `.I$()` — no leaked ring-buffer slot.

---

### Moderate

**M1 — `[ENT] QwpSidecarMain.java` `connectPool`: `db` handle leaks if the sender-array allocation itself throws (test harness; not reachable in practice).**
`Sender[] senders = null;` then `senders = new Sender[count];` is the first statement in the `try`. The `catch (Throwable t)` does `for (Sender s : senders)` — if `new Sender[count]` throws (only possible via `OutOfMemoryError` at an absurd `count`), `senders` is still `null`, the loop NPEs, masks the original throwable, and skips `db.close()`, leaking the connected `QuestDB` facade. The method's own comment claims "an allocation failure cannot leak the connected facade handle" — a guarantee it doesn't fully deliver.
*Moderate, not Critical:* the only trigger is an OOM from `new Sender[count]`; every real caller passes `count` = 1–3, for which the allocation never throws, so the leak is not reachable on any real path. *Fix (cheap, aligns code with its stated intent):* move `senders = new Sender[count];` above the `try`, or guard the cleanup loop with `if (senders != null)`.

**M2 — `[OSS] QwpIngressUpgradeProcessor.gracefulCloseAndDrain`: dropped `catch(Throwable)` around `shutdown()` breaks parity with the recv path (defense-in-depth; not reachable today).**
The old `gracefulCloseAndDisconnect` wrapped `socket.shutdown()` in `catch(Throwable)`; the new code does `if (socket == null || socket.shutdown(Net.SHUT_WR) != 0) throw ServerDisconnectException`. On the resume/send path (`resumeSend` → `SEND_STATE_RESUME_CLOSE` → `finishDeferredFatalClose`), `handleClientSend` catches only `PeerIsSlowToRead/PeerDisconnected/ServerDisconnect` — *not* `Throwable` (unlike `handleClientRecv`), so a throw from `shutdown()` would escape and skip `onConnectionClosed`/`onDisconnected`.
*Moderate, not Critical:* verified `Socket.shutdown(int)` returns `int` with no `throws`, and both `PlainSocket`/TLS impls delegate to the native `nf.shutdown(fd, how)` returning an int — a real failure surfaces as a non-zero return (handled by the `!= 0` branch), never a throw. No reachable escape today. *Optional fix:* restore the `catch(Throwable)` (or add one to `handleClientSend`) for parity.

---

### Minor
- **[ENT] `QwpSidecarMain.java` member ordering** — `closeWithWitness()` precedes `closeQuietly()` (alphabetical is `close`, `closeQuietly`, `closeWithWitness`); in `QwpSidecarMainTest.java`, `...PartialBorrow...` precedes `...EveryLease...`.
- **[ENT] CLOSE/CLOSE_BEGIN null-client path** throws `IllegalStateException("no sender")` instead of the `out.println("ERR no sender")` pattern every other verb uses. Output is identical; cosmetic inconsistency.
- **[ENT] Coverage nit** — the sidecar's `CLOSE_BEGIN` witness fan-out (`closeWithWitness` + the `if (!witnessed.get()) throw` guard) and `FLUSH_DEFER`'s `setDeferCommit` toggle have no in-repo Java unit test (the fakes pass `ignored -> null` as the delegate resolver). Exercised by the client-side `CloseDrainTest` + the ENT e2e suite, so indirect-only, not a gap.
- **[ENT] `egress_sidecar.py` `query()`** returns `(0, 0.0)` on a `<2`-token reply — inert against today's `OK <rows> <latencyMs>` contract, but would silently read a regressed one-token reply as "0 rows".
- **[BOTH] Stale submodule SHAs in both PR descriptions** — ENT body cites `questdb→d078ef13e`, OSS body cites head `cfece81562` / client `fee9903ca6`; the *actual* pins are `questdb→dc6fb12db`, client `→5bbdfb8678`. Verified the actual pins are legitimate later tips of `fix/qwp-worker-quiescence` and correspond to the described #67 behavior — stale description text, not a wrong-commit pin. Worth refreshing the bodies.

---

### Downgraded (false positives / non-findings)
- **`throw (Error) first` in `Client.close()` casts a checked exception** → FP: `Sender.close()`/`QuestDB.close()` narrow `Closeable.close()` to declare no checked exceptions, so a caught `Throwable` is only `RuntimeException` or `Error`.
- **`closeDrainDeadline` wiped by per-message `clear()`** → FP: reset lives in `onDisconnected()` only (verified by source + `testCloseDrainLifecycle`).
- **`reconnect_max_duration_millis` is a banned drainer reconnect budget** → FP: it is the sanctioned capability-gap settle budget / startup connect budget; transient reconnect is uncapped (Invariant B), verified in both drainers.
- **Dense oracle misses duplicates** → FP: `fetch_column_sorted` does not `DISTINCT`.
- **Stale SHAs = unintended pin** → FP: actual pins are legitimate branch tips.
- **CI choco-retry loop can loop forever / mask failure** → FP: bounded `-le 3`, `throw` on the final non-success exit; `-in 0,1641,3010` is valid.

---

### Coverage map (production behavioral changes)
| Change | Test(s) | Verdict |
|---|---|---|
| OSS `gracefulCloseAndDrain` linger (fatal CLOSE no longer RSTs the final ack) | `QwpServerCloseDrainTest` (linger + deadline-bound), `QwpIngressUpgradeProcessorResumeRecvTest.testCloseDrainYields…` (32-recv quantum), `testCloseDrainLifecycle` (state machine), ENT `test_close_drain_failover` (end-to-end exactly-once through real failover) | **TESTED** (regression link verified) |
| OSS drain state (`beginCloseDrain`/`isCloseDraining`/`isCloseDrainExpired`, survives `clear`, reset on disconnect) | `QwpIngressProcessorStateTest.testCloseDrainLifecycle` | **TESTED** |
| Client SF close/pool/drainer/watermark machinery (#67) | 52 client test files in #67 + ENT e2e (multislot recovery, kill-9 cycles, close-drain) | **TESTED** (transitive; contract paths independently verified) |
| ENT sidecar harness: `CONNECT_POOL`, `*_ON`, partial-borrow cleanup, `ClientState` ownership | `QwpSidecarMainTest` (486L) | **TESTED** |
| ENT sidecar `CLOSE_BEGIN` witness fan-out + `FLUSH_DEFER` | client `CloseDrainTest` + ENT e2e | **TESTED (indirect)** |
| `ReverseTrafficGate` (pause/resume boundaries, concurrent close, thread accounting) | `test_reverse_traffic_gate.py` (449L) | **TESTED** |
| `sidecar.py` protocol (`_parse_stats` order-independent, coalesced close replies) | `test_sidecar.py` | **TESTED** |
| CI chocolatey retry, `.gitignore`, `fix-pr` skill doc | — | **EXEMPT** (infra/doc, no behavioral DB change) |

---

### Summary
- **Verdict — ENT #1127: approve. OSS #7387: approve.** Client #67 is transitively pinned; its SF-contract-critical paths (drainer reconnect containment, close/pool/watermark ordering) were independently verified compliant here, but its full line-by-line review is #67's own scope.
- **Correctness & performance gate: PASS** — zero confirmed correctness/data-loss/resource/performance findings on any reachable path. The two Moderate items are OOM-only (M1) and non-throwing-native (M2) — neither reachable; both are cheap defense-in-depth fixes, not blockers.
- **Test gate: PASS** — this is a test PR that ships the production behavioral change (OSS close-drain) *with* a regression test that fails pre-fix; the e2e oracles are non-vacuous and use contract-correct evidence.
- **Findings: 8 draft verified, 6 dropped.** Split: 2 Moderate (1 in-diff ENT, 1 in-diff OSS), 0 out-of-diff, 0 pre-existing. Cross-context pass ran clean.
- No tandem ACL gap: the OSS change is a WebSocket transport teardown in the QWP ingress path, not a new SQL/ALTER/state-mutating operation, so no enterprise `SecurityContext`/`Permission` wiring is required.

Optional before merge (none blocking): apply M1 + M2 (one-line each), refresh the two stale submodule SHAs in the PR bodies, and fix the two member-ordering nits.

<sub>🤖 Tandem review at level 3 (all agents, per-finding source verification).</sub>

RaphDal (dismissed):

bluestreak01:
Root-caused the two mac test failures in [build 254645](https://dev.azure.com/questdb/questdb/_build/results?buildId=254645): both are scheduling-starvation flakes on hosted 3-core mac agents — no product issue in either.

- `PerWorkerLocksTest.testConcurrentAcquireReleaseHoldsMutualExclusion` (arrived via the master merge, from #7371): 8 pure-spin threads oversubscribe the ~3 cores, so preempted slot holders convoy behind spinners — the test took 12.8s / 18.8s / 52.4s on the three green macwin runs since it landed, then hit its 60s deadline here. Fixed on this branch in a60b8e0f78: threads scale to the core count, rounds pool in a shared quota, and the backstop deadline reports progress diagnostics.
- `QuestDBImplCloseLifecycleTest.facadeCloseIsBoundedUnderRepeatedInterruptsDuringQueryCreation` (new in the carried client change): the post-join `interruptCount > 1` assert races the 100ms close budget against thread scheduling; the failing run's log shows `close()` honoring its bounded deadline exactly, so only the test's liveness precondition broke. Test-only fix in questdb/java-questdb-client#72; e2bdcd4272 bumps the submodule pin to that branch head (`b67d4f42`) for CI validation.

Follow-up before merge: once questdb/java-questdb-client#72 lands, move the pin to the client merge commit.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 34 / 39 (87.18%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpIngressUpgradeProcessor.java|26|31|83.87%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/QwpIngressProcessorState.java|8|8|100.00%|

PR #7391, opened 2026-07-12
fix(SQL): fix SAMPLE BY w to accept the week unit without a leading digit

### LINKED ISSUE #7293: Sample by week is the only sample by period not working without a digit
### To reproduce

1. Go to https://demo.questdb.io/index.html
2. Run the query
```
SELECT to_timezone(pickup_datetime, 'Europe/Rome') AS time, SUM(passenger_count) AS pass  FROM trips
SAMPLE BY w FROM '2009-01-01T01:02:08.000000Z' TO '2009-01-01T01:02:30.000000Z'
FILL(NULL)
ALIGN TO CALENDAR

```
3. see the error 'one letter sample by period unit expected'

Image

### QuestDB version:

9.4.3

Questdb demo

Questdb demo

### Full Name:

Pier Francesco Costa

### Affiliation:

Trailslight

- [x] Yes, I have

### Additional context

Instead running the same query with 'h' as a sample period 'hour' works,
```
SELECT to_timezone(pickup_datetime, 'Europe/Rome') AS time, SUM(passenger_count) AS pass  FROM trips
SAMPLE BY h FROM '2009-01-01T01:02:08.000000Z' TO '2009-01-01T01:02:30.000000Z'
FILL(NULL)
ALIGN TO CALENDAR
```

Image

nabeel001:
Fixes #7293

`SAMPLE BY w` (week) failed to parse with the error `one letter sample by period unit expected`, even though every other single-letter sampling unit works without a leading digit: `SAMPLE BY h`, `SAMPLE BY d`, `SAMPLE BY M`, and so on. Users had to write `SAMPLE BY 1w` as a workaround.

## Root cause

The parser validates the bare single-letter form against an allow-list in `SqlParser.isValidSampleByPeriodLetter()`. That list was missing `'w'`. The method backs both parser paths that accept a unit with no interval (`isFullSampleByPeriod()` and `expectSample()`), so `w` alone was rejected before it ever reached the sampler.

The week unit itself was already fully supported everywhere downstream: both the micros and nanos timestamp drivers map `'w'` to a `WeekTimestamp*Sampler`, and the no-digit form resolves to "1 week" via `TimestampSamplerFactory.parsePositiveInterval()` (which returns 1 when no number precedes the unit, exactly as `SAMPLE BY m` == `SAMPLE BY 1m`). This was purely a validation-list oversight, not a missing feature.

## Fix

- Add `'w'` to the allow-list in `isValidSampleByPeriodLetter()`.
- Complete the accompanying comment, which was also missing `weeks` and `years`.

## Impact and tradeoffs

The change only widens what the parser accepts; it does not alter the behavior of any query that already parsed. `SAMPLE BY w` now behaves identically to `SAMPLE BY 1w`. There is no performance impact and no change to existing valid or invalid inputs other than `w` moving from rejected to accepted.

## Test plan

- Added `SampleByTest#testSampleFillWithWeekStrideNoDigit`, which runs the issue's query shape (`SAMPLE BY w ... FILL(NULL) ALIGN TO CALENDAR`) and asserts it returns the same result as the explicit `1w` form.
- Ran the new test together with the existing `testSampleFillWithWeekStride`.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/SqlParser.java
+++ b/core/src/main/java/io/questdb/griffin/SqlParser.java
@@ -477,8 +477,10 @@ public class SqlParser {
             // minutes
             // hours
             // days
+            // weeks
             // months
-            case 'n', 'U', 'T', 's', 'm', 'h', 'd', 'M', 'y' -> true;
+            // years
+            case 'n', 'U', 'T', 's', 'm', 'h', 'd', 'w', 'M', 'y' -> true;
             default -> false;
         };
     }
--- a/core/src/test/java/io/questdb/test/griffin/engine/groupby/SampleByTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/engine/groupby/SampleByTest.java
@@ -16920,6 +16920,37 @@ public class SampleByTest extends AbstractCairoTest {
     }
 
     @Test
+    public void testSampleFillWithWeekStrideNoDigit() throws Exception {
+        assertMemoryLeak(() -> {
+            execute(FROM_TO_DDL);
+
+            String expected = """
+                    ts\tavg
+                    2017-12-20T00:00:00.000000Z\tnull
+                    2017-12-27T00:00:00.000000Z\t48.5
+                    2018-01-03T00:00:00.000000Z\t264.5
+                    2018-01-10T00:00:00.000000Z\t456.5
+                    2018-01-17T00:00:00.000000Z\tnull
+                    2018-01-24T00:00:00.000000Z\tnull
+                    """;
+
+            String queryNoDigit = "select ts, avg(x) from fromto\n" +
+                    "sample by w from '2017-12-20' to '2018-01-31' fill(null) align to calendar";
+            assertQuery(queryNoDigit)
+                    .timestamp("ts")
+                    .noRandomAccess()
+                    .noLeakCheck()
+                    .returns(expected);
+
+            assertQuery(queryNoDigit.replace("sample by w", "sample by 1w"))
+                    .timestamp("ts")
+                    .noRandomAccess()
+                    .noLeakCheck()
+                    .returns(expected);
+        });
+    }
+
+    @Test
     public void testSamplePeriodInvalidWithNoUnits() throws Exception {
         testSampleByPeriodFails(
                 "select sum(a), k from x sample by 300/10 align to calendar",


kafka1991:
## Code review — level 3 (full mission-critical pass)

Reviewed the one-line allow-list change plus the regression test. Ran the structured correctness/perf/test/cross-context analysis together with a dedicated fresh-context adversarial pass; both converged on the same conclusion.

### What the change does
- `SqlParser.isValidSampleByPeriodLetter()`: adds `'w'` to the bare-single-letter unit allow-list and completes the comment block (`// weeks`, and the previously-missing `// years`).
- Adds regression test `SampleByTest#testSampleFillWithWeekStrideNoDigit`.

### Verification performed
- **Root cause is real and correctly diagnosed.** Both reachable SAMPLE BY drivers already had a `'w'` arm (`MicrosTimestampDriver` -> `WeekTimestampMicrosSampler`, `NanosTimestampDriver` -> `WeekTimestampNanosSampler`). The bare form resolves to interval `1` via `TimestampSamplerFactory.parsePositiveInterval()`'s `intervalEnd == 0` early return. Purely a validation-list omission; the canonical unit set in `TimestampDriver.java:464`'s javadoc already listed `'w'`.
- **Post-fix allow-list `{n,U,T,s,m,h,d,w,M,y}` now exactly matches both driver switch statements.** No unit is accepted that a driver cannot handle.
- **Callsite inventory:** `isValidSampleByPeriodLetter` has 2 callers, both in `SqlParser` (bare-literal path via `isFullSampleByPeriod`, and the complex-expression unit path). `isFullSampleByPeriod` (public static) has no external callers. Other unit-letter lists checked and found consistent: `CommonUtils.getStrideUnit` (timestamp_floor rewrite) already has `w`; mat-view PERIOD-length and `SqlUtil.expectSeconds` intentionally exclude weeks (different grammar).
- **No grammar collision / no previously-valid query changes meaning.** Bare `w` always failed before, so nothing that parsed before parses differently now.
- **No existing negative test breaks.** The rejection tests use `x`, `L`, and `milli` - none use `w`.
- **MILLIS driver's `UnsupportedOperationException` is unreachable from SAMPLE BY** (`ColumnType.getTimestampDriver` only returns Micros/Nanos for a designated timestamp) and would fail identically for every unit, so not new to `w`.

### Critical
None.

### Moderate
None.

### Minor
- **M1 — Missing PR labels.** (Now fixed: `Bug` and `SQL` added.)
- **M2 — Title scope could be `sql`.** The change lives in `griffin` (the SQL engine); `fix(sql):` is marginally more precise than `fix(core):`. Non-blocking.
- **M3 — Nanos driver only covered indirectly.** The new test exercises the micros (default `TIMESTAMP`) driver. The changed production line is driver-agnostic (shared parser) and the nanos `'w'` arm is verified by code, so this is a nice-to-have: a bare-`w` case in `SampleByNanoTimestampTest` would broaden coverage but exercises the identical parser line.

### Investigated and cleared (false positives)
- **Mat-view cross-context exposure** — cleared. `CreateMatViewOperationImpl.findSampleByNode` requires `type == CONSTANT`, and a bare unit is a `LITERAL`, so bare units (any of them, pre-existing) never reach mat-view sampling extraction. The PR does not change mat-view behavior.
- **Second allow-list out of sync** (`CommonUtils.getStrideUnit`) — cleared, already contains `'w'`.
- **`assertMemoryLeak` + `.noLeakCheck()` anti-pattern** — cleared. The lambda holds multiple statements (DDL + two `assertQuery` chains) sharing one leak-check scope; sanctioned use, and matches the adjacent `testSampleFillWithWeekStride`.
- **MILLIS driver `UnsupportedOperationException`** — cleared as unreachable from SAMPLE BY.

### Coverage map

| Change (symbol + branch) | Test | Failure link | Verdict |
|---|---|---|---|
| `isValidSampleByPeriodLetter` returns `true` for `'w'` - bare `SAMPLE BY w` (micros) | `SampleByTest#testSampleFillWithWeekStrideNoDigit` | Without the fix, `isFullSampleByPeriod("w")`=false -> parser reads next token `from` -> `isValidSampleByPeriodLetter("from")`=false -> throws `"one letter sample by period unit expected"`; `assertQuery(...).returns(expected)` gets an exception instead of rows -> test fails | TESTED |
| `w` == `1w` equivalence | same test, `.replace("sample by w","sample by 1w")` | Both forms asserted against the identical `expected` block; diverging sampler behavior would fail one | TESTED |
| bare `w` on nanos driver | (indirect) | driver-agnostic parser line; nanos `'w'` arm verified by code | Minor gap (M3) |
| Comment additions `// weeks`, `// years` | — | comment-only, no behavioral change | EXEMPT |

Dimensions for the main change: happy path covered; error path N/A (change only widens acceptance, existing rejection tests unchanged); NULL covered (FILL(NULL) boundary rows asserted null); boundaries covered (leading/trailing empty weeks null); concurrency N/A (pure static fn, no shared state); resource cleanup covered (`assertMemoryLeak`).

### Verdict: approve
Clean, minimal, correctly diagnosed fix. The regression test has a verified failure link (reverting the one-line hunk makes it throw at parse), so the test gate passes. Coverage map: 2 behavioral changes, both TESTED; 1 comment change EXEMPT; no UNTESTED rows. No regressions or tradeoffs - the change is strictly additive to what the parser accepts and matches the existing `1w` semantics exactly. Split: 0 in-diff bugs, 0 out-of-diff bugs, 0 pre-existing bugs.

kafka1991 (approved):

nabeel001:
@kafka1991 
I believe the check failed due to some intermittent issue. Could you please check and reinitiate a run again.

PR #5771, opened 2025-06-20
feat(sql): introduce indexOf(array), `indexOfAssumeSorted(array)` and `arraySum(array)` function.

kafka1991:
Introduce indexOf(array), `indexOfAssumeSorted(array)` and `arraySum(array)` function.

PR #7392, opened 2026-07-12
fix(sql): fix dot_product returning wrong result for transposed arrays

raphaelroshan:
## What

`dot_product()` over two `DOUBLE[]`/`DOUBLE[][]` arrays could return a silently incorrect result when one of the operands is not stored in vanilla (contiguous, row-major) layout.

In the recursive code path, the deepest-dimension loop read the right-hand array with the left array's flat index (`flatIndexLeft`) rather than its own cursor (`flatIndexRight`). `flatIndexRight` was computed and advanced but never used to read a value.

When both operands share an identical layout the two indices coincide, so the existing tests (both operands vanilla, or both transposed the same way) passed. When the operands have different strides — e.g. one is `transpose(...)` or a slice — reading the right array with the left index selects the wrong element, so the sum is wrong. Example:

```
transpose([[1,3],[2,5]]) = [[1,2],[3,5]],  right = [[1,5],[7,2]]
dot_product = 1*1 + 2*5 + 3*7 + 5*2 = 42   (previously returned 40)
```

## Fix

Read the right array with `flatIndexRight`.

## Impact

- Corrects results for `dot_product` when an operand has non-vanilla layout (transposed/sliced arrays).
- No behavior change for arrays that were already stored in identical vanilla layout — the common case — so existing correct results are unaffected.
- No performance impact: the change swaps one index variable for another that was already being maintained.

## Test plan

- Added `ArrayTest.testArrayDotProductTransposedOperand`, which mixes a transposed and a vanilla operand and asserts the corrected value (42.0). This case fails before the fix (returns 40.0).
- The existing `testArrayDotProduct` and `testArrayDotProductScalarValue` continue to pass.

### DIFF
--- a/core/src/main/java/io/questdb/griffin/engine/functions/array/DoubleArrayDotProductFunctionFactory.java
+++ b/core/src/main/java/io/questdb/griffin/engine/functions/array/DoubleArrayDotProductFunctionFactory.java
@@ -156,7 +156,7 @@ public class DoubleArrayDotProductFunctionFactory implements FunctionFactory {
             if (atDeepestDim) {
                 for (int i = 0; i < count; i++) {
                     double leftVal = left.getDouble(flatIndexLeft);
-                    double rightVal = right.getDouble(flatIndexLeft);
+                    double rightVal = right.getDouble(flatIndexRight);
                     if (Numbers.isFinite(leftVal) && Numbers.isFinite(rightVal)) {
                         sum += leftVal * rightVal;
                     }
--- a/core/src/test/java/io/questdb/test/cairo/ArrayTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/ArrayTest.java
@@ -1017,6 +1017,23 @@ public class ArrayTest extends AbstractCairoTest {
     }
 
     @Test
+    public void testArrayDotProductTransposedOperand() throws Exception {
+        // transpose([[1,3],[2,5]]) = [[1,2],[3,5]], right = [[1,5],[7,2]]:
+        // 1*1 + 2*5 + 3*7 + 5*2 = 42.
+        assertMemoryLeak(() -> {
+            execute("CREATE TABLE tango (left DOUBLE[][], right DOUBLE[][])");
+            execute("INSERT INTO tango VALUES (ARRAY[[1.0, 3.0], [2.0, 5.0]], ARRAY[[1.0, 5.0], [7.0, 2.0]])");
+            assertQuery("SELECT dot_product(transpose(left), right) AS product FROM tango")
+                    .noLeakCheck()
+                    .expectSize()
+                    .returns("""
+                            product
+                            42.0
+                            """);
+        });
+    }
+
+    @Test
     public void testArrayFirstFunction() throws Exception {
         assertMemoryLeak(() -> {
             execute("create table test (ts timestamp, x int, v double[]) timestamp(ts) partition by DAY");


kafka1991:
## Code review (level 3)

One-line correctness fix with a regression test. Verified by hand-trace, sibling-function comparison, the `ArrayView` contract, and a fresh-context adversarial pass.

### Critical
None.

### Moderate
None.

### Minor

1. **Missing symmetric-direction test.** The new test covers only *left non-vanilla, right vanilla* (`dot_product(transpose(left), right)`). The reverse (`dot_product(left, transpose(right))`) exercises the same corrected line via a symmetric path; a second assertion would nail down that the fix reads *both* operands by their own index. Cheap to add:
   ```java
   assertQuery("SELECT dot_product(left, transpose(right)) AS product FROM tango")
           .noLeakCheck().expectSize()
           .returns("product\n38.0\n");   // 1*1 + 3*7 + 2*5 + 5*2 = 38
   ```
   Not blocking — the fixed line is already directly exercised.

2. **Pre-existing, orthogonal: NULL-array return inconsistency.** `DoubleArrayDotProductFunctionFactory.getDouble` returns `0d` when either array operand is NULL, whereas the scalar variant `DoubleArrayAndScalarDotProductFunctionFactory` returns `Double.NaN` for a NULL scalar (see `testArrayDotProductScalarValue`: `dot_product(left, null::double)` -> `null`). This predates the fix and is genuinely orthogonal to transposition; changing it would alter documented/tested behavior and needs its own decision on the intended NULL semantics. Flagged as an observation, not folded into this PR.

### Verification notes (non-issues confirmed)
- **Not a crash / OOB pre-fix.** For same-shape operands the left index range stays within the right's valid range, so the old bug was silent-wrong-value only, never a SIGSEGV. Traced a 2x3 case: max left index 5 < right length 6.
- **No copy-paste of the bug elsewhere.** Add / Subtract / Div / Multiply and `ArrayView.arrayEqualsRecursive` all correctly use `right.getDouble(flatIndexRight)`. `dot_product` was the lone outlier (bug present since #5771, 2025-07-02).
- **Fix complete for sliced views.** `getDouble` applies each view's own `flatViewOffset` internally, so relative index 0 is the correct recursion start; `shapeDiffers` compares shapes only, correctly admitting same-shape / different-stride operands into the recursive path.

### Coverage map

| Change (symbol + branch) | Test | Failure link | Verdict |
|---|---|---|---|
| `Func.applyRecursive` deepest-dim loop: read right operand at its own `flatIndexRight` (differing-stride operands) | `ArrayTest.testArrayDotProductTransposedOperand` | Asserts `42.0`; buggy line yields `40.0` (hand-traced two ways). Assertion observes the corrected element selection, so it fails if the line is reverted. | **TESTED** |

Dimensions: happy path covered; both-vanilla covered (`testArrayDotProduct`, separate fast path); both-transposed-identical covered (`testArrayDotProduct`); NULL element (NaN) via shared `Numbers.isFinite` guard covered in the vanilla path; reverse direction uncovered (Minor #1); concurrency N/A (static, no shared state); resource cleanup N/A (no native alloc, `assertMemoryLeak` used anyway).

One behavioral change, 1 tested, 0 UNTESTED.

### Summary
- **Verdict: approve.** The fix is correct, complete, and minimal — it aligns dot-product with the reference pattern used by every sibling binary array function, and it was the lone outlier. The regression test genuinely fails without the fix (42.0 vs 40.0, verified two independent ways).
- **Test gate: satisfied.** The single behavioral delta is TESTED with a verified failure link.
- **No regressions/tradeoffs.** Zero performance impact (one already-maintained variable substituted for another); the both-vanilla fast path and identical-layout cases are untouched.
- **Findings:** 0 Critical, 0 Moderate, 2 Minor; 0 false positives. Split: 1 in-diff (the fix), 0 out-of-diff, 1 pre-existing observation.

kafka1991 (approved):

PR #7393, opened 2026-07-12
Fix #7389: nanos timestamp validateBounds skips positive overflow

### LINKED ISSUE #7389: TIMESTAMP_NS: out-of-range value (beyond year 2261) is accepted at insert then silently suspends the table at WAL apply
## Summary

A `TIMESTAMP_NS` designated-timestamp value beyond `MAX_NANO_YEAR` (2261) — up to `Long.MAX_VALUE`, i.e. `2262-04-11T23:47:16.854775807Z` — is **not rejected at insert**. The row is accepted, and the failure only surfaces later at WAL apply, where creating the out-of-range partition fails and the table is **silently suspended**.

The same out-of-range magnitude on a micros `TIMESTAMP` column is correctly rejected at insert. So `TIMESTAMP_NS` is inconsistent with `TIMESTAMP`, and its failure mode (accept → suspend) is worse than a clean rejection: the client sees a successful insert, then the table is unusable.

Reproduced on **master (`cdb0ea073b`)**.

## Steps to reproduce

```sql
CREATE TABLE tn (ts TIMESTAMP_NS, x INT) TIMESTAMP(ts) PARTITION BY DAY WAL;

-- Accepted WITHOUT error (this is year 2262, beyond MAX_NANO_YEAR = 2261):
INSERT INTO tn VALUES (CAST(9223372036854775807 AS TIMESTAMP_NS), 1);

-- After WAL apply, the table is suspended:
SELECT name, suspended, errorMessage FROM wal_tables() WHERE name = 'tn';
```

## Actual result

The insert succeeds, then `wal_tables()` reports the table suspended:

```
suspended | errorMessage
true      | could not mmap file, invalid len [len=0, offset=0, fd=..., memoryTag=6]
```

Contrast — the same value on a micros `TIMESTAMP` column is rejected up front:

```sql
CREATE TABLE tm (ts TIMESTAMP, x INT) TIMESTAMP(ts) PARTITION BY DAY WAL;
INSERT INTO tm VALUES (CAST(9223372036854775807 AS TIMESTAMP), 1);
-- ERROR: designated timestamp beyond 9999-12-31 is not allowed
```

## Expected result

An out-of-range `TIMESTAMP_NS` value should be rejected at insert with a clear error (mirroring the micros message, e.g. `... beyond 2261-12-31 ... is not allowed`), not accepted and then left to suspend the table at apply.

## Root cause

`NanosTimestampDriver.validateBounds(long)` only runs the range check for **negative** timestamps, so positive beyond-max values are never validated:

```java
// core/.../io/questdb/cairo/NanosTimestampDriver.java:1629
public void validateBounds(long timestamp) {
    if (timestamp < 0) {                 // only negatives reach the check
        validateBounds0(timestamp);
    }
}
```

Two issues compound:
1. The outer guard `timestamp < 0` means a positive out-of-range value never reaches `validateBounds0`.
2. Even `validateBounds0`'s upper-bound clause is wrong — it compares against `CommonUtils.TIMESTAMP_UNIT_NANOS`, which is the unit enum constant `1`, not a max-nanos value (harmless today only because the clause is reached exclusively for negatives):

```java
// NanosTimestampDriver.java:1747
if (timestamp < TableWriter.TIMESTAMP_EPOCH || timestamp > CommonUtils.TIMESTAMP_UNIT_NANOS) { ... }
```

Compare `MicrosTimestampDriver.validateBounds`, which is correct — the unsigned compare catches both negatives and beyond-`YEAR_10000`:

```java
// MicrosTimestampDriver.java:1659
public void validateBounds(long timestamp) {
    if (Long.compareUnsigned(timestamp, Micros.YEAR_10000) >= 0) {
        validateBounds0(timestamp);
    }
}
```

## Suggested fix

Make `NanosTimestampDriver.validateBounds` mirror the micros pattern — an unsigned compare against the maximum valid nanos timestamp — so both negatives and beyond-`MAX_NANO_YEAR` values reach `validateBounds0`; and fix `validateBounds0`'s upper bound to compare against the real max-nanos value rather than `TIMESTAMP_UNIT_NANOS` (= 1).

## Reproducing test (passes on `cdb0ea073b`: `Tests run: 2, Failures: 0`)

<details><summary>NanosTimestampBoundsReproTest.java</summary>

```java
package io.questdb.test.griffin;

import io.questdb.cairo.TableToken;
import org.junit.Assert;
import org.junit.Test;

public class NanosTimestampBoundsReproTest extends AbstractCairoTest {

    // Long.MAX_VALUE nanos-since-epoch == 2262-04-11T23:47:16.854775807Z, beyond MAX_NANO_YEAR (2261).
    private static final long OUT_OF_RANGE = Long.MAX_VALUE;

    @Test
    public void testMicrosOutOfRangeIsRejectedAtInsert() throws Exception {
        assertMemoryLeak(() -> {
            execute("create table tm (ts timestamp, x int) timestamp(ts) partition by DAY WAL");
            try {
                execute("insert into tm values (cast(" + OUT_OF_RANGE + " as timestamp), 1)");
                Assert.fail("expected out-of-range micros timestamp to be rejected at insert");
            } catch (Exception e) {
                Assert.assertTrue("unexpected message: " + e.getMessage(),
                        e.getMessage() != null && e.getMessage().contains("designated timestamp"));
            }
        });
    }

    @Test
    public void testNanosBeyondMaxYearSuspendsTableInsteadOfRejecting() throws Exception {
        assertMemoryLeak(() -> {
            execute("create table tn (ts timestamp_ns, x int) timestamp(ts) partition by DAY WAL");
            // Accepted at insert (validateBounds only range-checks negatives):
            execute("insert into tn values (cast(" + OUT_OF_RANGE + " as timestamp_ns), 1)");
            // Fails at WAL apply -> table silently suspended:
            drainWalQueue();
            final TableToken tt = engine.verifyTableName("tn");
            Assert.assertTrue(
                    "out-of-range TIMESTAMP_NS should be rejected up front, but the table was suspended",
                    engine.getTableSequencerAPI().isSuspended(tt));
        });
    }
}
```
</details>

ayanasarkar:
What does this PR do?

Resolves #7389. This fixes an issue where the NanosTimestampDriver allowed positive, overflowing timestamps (beyond the year 2261 cap) to bypass validation during insert. This previously caused the table to silently suspend during WAL apply.

Root Cause:

validateBounds() only guarded against timestamp < 0, allowing any positive nanos value to pass.

validateBounds0() incorrectly compared the timestamp against CommonUtils.TIMESTAMP_UNIT_NANOS (a unit-tag constant with a value of 1) instead of the actual boundary CommonUtils.MAX_TIMESTAMP.

Changes made:

Fixed both validation methods in NanosTimestampDriver to properly compare against CommonUtils.MAX_TIMESTAMP (mirroring the safe logic already used in MicrosTimestampDriver).

Added bound validation unit tests in NanoTimestampDriverTest.

Added end-to-end WAL regression tests in TimestampBoundsTest to verify that inserting Long.MAX_VALUE into a TIMESTAMP_NS column correctly throws a CairoException on insert instead of suspending the table later.  

Note for Reviewers:

NanosTimestampDriver.getMaxDesignatedTimestamp() currently still returns Long.MAX_VALUE with a comment stating nanos are uncapped. I left this untouched to keep the scope of this fix tight, but flagging it here in case it needs a follow-up adjustment to match the enforced 2261-12-31 limit.

### DIFF
--- a/core/src/main/java/io/questdb/cairo/NanosTimestampDriver.java
+++ b/core/src/main/java/io/questdb/cairo/NanosTimestampDriver.java
@@ -1627,7 +1627,7 @@ public class NanosTimestampDriver implements TimestampDriver {
 
     @Override
     public void validateBounds(long timestamp) {
-        if (timestamp < 0) {
+        if (Long.compareUnsigned(timestamp, CommonUtils.MAX_TIMESTAMP) > 0) {
             validateBounds0(timestamp);
         }
     }
@@ -1748,7 +1748,7 @@ public class NanosTimestampDriver implements TimestampDriver {
         if (timestamp == Long.MIN_VALUE) {
             throw CairoException.nonCritical().put("designated timestamp column cannot be NULL");
         }
-        if (timestamp < TableWriter.TIMESTAMP_EPOCH || timestamp > CommonUtils.TIMESTAMP_UNIT_NANOS) {
+        if (timestamp < TableWriter.TIMESTAMP_EPOCH || timestamp > CommonUtils.MAX_TIMESTAMP) {
             throw CairoException.nonCritical().put("designated timestamp_ns before 1970-01-01 and beyond ").put(MAX_NANO_TIMESTAMP_STR).put(" is not allowed");
         }
     }
--- a/core/src/test/java/io/questdb/test/cairo/NanoTimestampDriverTest.java
+++ b/core/src/test/java/io/questdb/test/cairo/NanoTimestampDriverTest.java
@@ -24,6 +24,7 @@
 
 package io.questdb.test.cairo;
 
+import io.questdb.cairo.CairoException;
 import io.questdb.cairo.ColumnType;
 import io.questdb.cairo.ImplicitCastException;
 import io.questdb.cairo.TimestampDriver;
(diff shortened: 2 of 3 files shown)


ayanasarkar:
@nwoolmer could you please review this

ayanasarkar:
@nwoolmer pls could you review the changes once

PR #7395, opened 2026-07-13
test(core): stop query_activity churn test hanging and cascading on constrained CI

kafka1991:
## Root cause

`QueryActivityFunctionFactoryTest.testQueryActivitySnapshotUnderChurn` is a concurrency stress test: a reader thread polls `query_activity()` in a loop while four producers register and unregister 20,000 queries each, asserting the optimistic snapshot never exposes a torn query string.

On a CPU-constrained CI runner it hangs. The reader loop never yields, so on a small runner it can starve a producer out of CPU. The starved producer never finishes its iterations, so `runningProducers` never reaches 0 and the reader loops indefinitely too; both burn the full 120s `join` timeout. On timeout the test asserted and returned without stopping the still-live threads. Those leaked threads kept mutating the shared static engine and cascaded into 9 unrelated failures in the same Surefire fork: a leaked compiler-pool table, file-descriptor and native-memory leak-check failures, and `EqStrCharFunctionTest`'s teardown.

The registry under test is a lock-free `ConcurrentLongHashMap` read without synchronisation, so this is CPU starvation in the test harness, not a product deadlock.

## Fix

- A shared 90s wall-clock deadline caps both the producer and reader loops, so a slow run winds down cooperatively instead of tripping the join timeout. A healthy run finishes in ~1-2s (producers hit their 20,000-iteration cap) and never reaches the deadline.
- The reader yields between `query_activity()` passes so it can no longer starve the producers. The torn-read race window lives inside a single pass, so test efficacy is unchanged.
- A `finally` block interrupts any survivor after the join and joins again briefly. The interrupt also rescues threads stuck on `startBarrier.await()` when a peer dies before reaching the barrier, which was a second latent hang mode.
- The hung/fault assertions run after that cleanup, so a failure in this test can no longer leak live threads into the rest of the fork.

Thread termination also fixes the test's own native-memory leak: the reader's execution context, compiler and factory are closed by try-with-resources only when the thread exits.

## Impact and tradeoffs

- No production code changes; only the test is touched, and the behaviour under test is unchanged.
- The 90s deadline only bites on a starved run. On such a run the test does less churn than it otherwise would, which slightly weakens regression detection in that specific case; a healthy run is unaffected. 90s was chosen over a tighter cap to preserve churn coverage on slow runners, at the cost of a higher worst-case wall time.
- Worst case, a genuinely wedged run now spends ~100s on the join plus a 5s shutdown join before failing, versus 120s before. The difference is that it now fails as this test alone and cleans up, instead of cascading into unrelated tests. A real product deadlock still surfaces as a loud `worker thread hung` failure rather than being masked.

## Test plan

- Confirmed the changed file is the only file touched; no production code changes.
- Traced the revert: without the deadline and the reader yield, the original starvation hang reproduces under CPU contention, and the leaked threads reproduce the cascade.
- Suggested confirmation before merge: run `testQueryActivitySnapshotUnderChurn` in a loop pinned to 1-2 cores to confirm it no longer hangs and the surrounding tests in the class stay green.

### DIFF
--- a/core/src/test/java/io/questdb/test/griffin/engine/functions/activity/QueryActivityFunctionFactoryTest.java
+++ b/core/src/test/java/io/questdb/test/griffin/engine/functions/activity/QueryActivityFunctionFactoryTest.java
@@ -250,6 +250,7 @@ public class QueryActivityFunctionFactoryTest extends AbstractCairoTest {
             final QueryRegistry registry = engine.getQueryRegistry();
             final int producerCount = 4;
             final int iterations = 20_000;
+            final long runDeadlineNanos = System.nanoTime() + TimeUnit.SECONDS.toNanos(90);
 
             // Producer texts alternate between a short and a long, distinctly-keyed
             // string so recycling an entry changes the StringSink length and
@@ -280,7 +281,7 @@ public class QueryActivityFunctionFactoryTest extends AbstractCairoTest {
                 final Thread thread = new Thread(() -> {
                     try (SqlExecutionContextImpl context = new SqlExecutionContextImpl(engine, 1).with(AllowAllSecurityContext.INSTANCE)) {
                         startBarrier.await();
-                        for (int i = 0; i < iterations && fault.get() == null; i++) {
+                        for (int i = 0; i < iterations && fault.get() == null && System.nanoTime() < runDeadlineNanos; i++) {
                             final long queryId = registry.register(producerTexts[slot][i & 1], context);
                             // brief window for the reader to snapshot the entry
                             for (int j = 0; j < 8; j++) {
@@ -304,7 +305,7 @@ public class QueryActivityFunctionFactoryTest extends AbstractCairoTest {
                         RecordCursorFactory factory = CairoEngine.select(compiler, "SELECT query_id, worker_pool, username, state, query FROM query_activity()", context)
                 ) {
                     startBarrier.await();
-                    while (runningProducers.get() > 0 && fault.get() == null) {
+                    while (runningProducers.get() > 0 && fault.get() == null && System.nanoTime() < runDeadlineNanos) {
                         try (RecordCursor cursor = factory.getCursor(context)) {
                             final Record record = cursor.getRecord();
                             while (cursor.hasNext()) {
@@ -331,6 +332,8 @@ public class QueryActivityFunctionFactoryTest extends AbstractCairoTest {
                                 }
                             }
                         }
+                        // yield between passes so the reader cannot starve the producers on a busy CI box
+                        Thread.yield();
                     }
                 } catch (Throwable t) {
                     fault.compareAndSet(null, t);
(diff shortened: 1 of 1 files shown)


kafka1991:
## Review — test-only deflake

This is a test-only diff (one method, zero production code). The core deflake mechanism is sound and strictly better than master (it stops the thread-leak cascade). No Critical findings. **1 Moderate, 3 Minor** — all cheap robustness/diagnostic improvements worth folding into this branch while it's open; none block merge.

**Verdict: approve.**

---

### Moderate

**M1 — Sequential `join(joinTimeoutMs)` makes the pathological wind-down cost `(N−1)×100s`, not ~100s** (`QueryActivityFunctionFactoryTest.java:350-352`)

The join loop waits on each of the 5 threads sequentially, each with its own 100s timeout. `Thread.join(t)` blocks up to `t` **per call**, so the pre-interrupt window is additive.

Reachable corner case: the reader's `try (...CairoEngine.select(compiler, "SELECT ... query_activity()", context))` (line 305) compiles a query and acquires a pooled compiler **before** `startBarrier.await()` (line 307). If that construction throws (compiler-pool exhaustion, OOM-during-compile, etc.), the reader sets `fault`, never arms the barrier, and the 4 producers block forever on the 5-party `CyclicBarrier` — the `runDeadlineNanos` check is *inside* the loop body, downstream of the barrier, so the deadline cannot free them. Only the `finally` interrupt can (barrier `await()` is interruptible), but it does not fire until the join loop drains: up to **~400-500s** before any survivor is interrupted. A single early-construction failure inflates the test from ~2s to ~500s before it reports. It still recovers and reports the real fault, so this is a worst-case wall-time regression, not a hang — and against master it's strictly better (master leaked the stranded threads).

Bound the whole join phase by one shared deadline instead of N independent timeouts:

```java
final long joinDeadlineNanos = System.nanoTime() + TimeUnit.SECONDS.toNanos(100);
for (int i = 0, n = threads.size(); i < n; i++) {
    final long remainingMs = TimeUnit.NANOSECONDS.toMillis(joinDeadlineNanos - System.nanoTime());
    if (remainingMs > 0) {
        threads.getQuick(i).join(remainingMs);
    }
}
```

(Or `startBarrier.reset()` when `fault` is first set, so barrier waiters release without waiting out any timeout.)

---

### Minor

**m1 — `producerRowsObserved > 0` can now fail a deadline-truncated (starved) run** (`QueryActivityFunctionFactoryTest.java:375`)

Before this PR the loops ran to completion, so the reader always ran long enough to observe live producer rows. The new deadline adds an exit path where the reader can stop at 90s. On a pathologically starved box where the reader observed zero producer rows, the pre-existing `Assert.assertTrue("reader never observed a live query", producerRowsObserved.get() > 0)` fails — ironically on exactly the constrained CI box this PR exists to keep green. Probability is very low, but it's a genuinely new vector. Make the assert deadline-aware:

```java
if (System.nanoTime() < runDeadlineNanos) {
    Assert.assertTrue("reader never observed a live query", producerRowsObserved.get() > 0);
}
```

This also cleanly encodes the accepted tradeoff already stated in the PR body (a deadline-truncated run does less churn and passes with weaker coverage) — that reduced coverage on starved boxes is intended, and making the assert deadline-aware just stops it firing on the truncated path.

**m2 — An interruptible genuine hang reports `"worker thread failed: InterruptedException"` instead of `"worker thread hung"`** (`QueryActivityFunctionFactoryTest.java:358-359`, `370-373`)

The ordering is already correct — `assertFalse("worker thread hung", ...)` runs before the `fault` throw — so a thread still wedged in a non-interruptible loop is reported correctly as "hung". But if the hang is at an interruptible point, the `finally` interrupt injects `InterruptedException`, the worker's `catch` records it into `fault`, the thread then dies (so the "hung" assert passes), and the test throws the generic `"worker thread failed"` with an `InterruptedException` cause — muddier diagnostics for a real hang. Optional: skip recording `InterruptedException` into `fault` once `hasSurvivors` is set.

**m3 — `runDeadlineNanos` is captured at method entry, before setup and the barrier trip** (`QueryActivityFunctionFactoryTest.java:253`)

The 90s budget starts before building `producerTexts`/`validQueries`, constructing 5 threads, and the barrier rendezvous, so slow setup silently shrinks the churn window. Negligible at 90s; mention-only.

---

### Considered and dismissed

- **`Thread.yield()` reduces torn-read detection** — dismissed. The torn-read window is *within* a single reader pass (`hasNext()` → `QueryActivityRecord.of()` → `copy()`); the yield sits *between* passes (after the inner `try (RecordCursor ...)` closes). Per-pass detection is unchanged; the yield only reschedules between passes and, if anything, lets producers generate more churn. Efficacy-neutral.
- **Deadline masks starvation → green pass** — accepted, documented tradeoff (PR body); asserting producers completed all 20,000 iterations would reintroduce the flake. Folded into m1.
- **`interrupt()` is inert against the churn loops** — by design; the deadline is the stop signal for the loops, the interrupt's job is barrier rescue.
- **`System.nanoTime() < runDeadlineNanos` overflow-unsafe** — dismissed; `nanoTime()` is monotonic with no realistic wraparound over a 90s window.

---

### Coverage / metadata

No production code changes — all behavioral deltas are test-harness control flow, so the test gate is satisfied (flaky-test fix; revert-trace reasoned in the PR's own test plan; the `query_activity()` torn-read assertion at line 329 still propagates). Title/labels/tone all conform. `hasSurvivors` uses the `has` prefix; `20_000`/`100_000`/`5_000` use `_` separators.

M1 + m1 together close the two ways this deflake can still misbehave on a genuinely starved box — the exact environment it targets.

bluestreak01 (approved):

Issue #3461, opened 2023-06-09
Support deleting a record


joyswing:
### Is your feature request related to a problem?

For example, the sensor table has a historical data table SensorHis, but in practical applications, the sensor may add or delete records as business changes, but the sensor table cannot delete records. This is a very inconvenient place, so I have to place basic information in other databases such as SqlServer. However, in actual business queries, it is necessary to connect the sensor and SensorHis to query data, Is this causing trouble?

Utpal3:
@joyswing please assign this issue to me i can solve it

Utpal3:
**Option** 1: Enable Mutable Tables for Metadata
Introduce a new table type in QuestDB that supports UPDATE and DELETE, limited to small tables for metadata (e.g., sensor, user, device_type).
These tables would:

Be row-oriented or hybrid (not fully columnar)

Be designed for low-write, high-read use cases

Have optional foreign key relationships (soft enforced)

This allows:

_DELETE FROM sensor WHERE sensor_id = 'sensor_42';_
_UPDATE sensor SET location = 'Zone B' WHERE sensor_id = 'sensor_5';_

**Option** 2: Sync External Metadata Table in QuestDB
If full mutability isn't feasible, allow:

An external table source connector to sync data from mutable systems (e.g., SQL Server)

A materialized view that merges external metadata and SensorHis in QuestDB

PR #7403, opened 2026-07-15
feat(sql): DELETE FROM <table> WHERE ... on WAL tables

nwoolmer:
## What

Implements `DELETE FROM <table> WHERE <predicate>` for WAL tables. A `WHERE` clause is mandatory; non-WAL tables, views, and materialized views are rejected.

## How

DELETE is recorded as a WAL SQL transaction and recompiled by WAL apply. Apply selects one of these strategies:

1. Pure designated-timestamp ranges use one empty `replaceRange` over the deleted interval.
2. Other predicates rewrite survivor rows in timestamp windows under one atomic replace-range transaction.
3. The opt-in `cairo.wal.delete.disk.bounded` mode commits eligible Parquet-backed deletes one window at a time to reduce transient native disk usage. This mode is not reader-atomic.

The replay classifier keeps random, clock-dependent, and otherwise unstable predicates on the atomic route. User bind values are captured in WAL, and apply-owned indexed binds cannot collide with user variables.

Additional behavior:

- Parquet partitions are converted to native before rewrite. Native data, indexes, and the partition directory follow the configured durability mode before `_txn` publication.
- Dependent materialized views force the atomic route. DELETE also invalidates affected views, including deferred lock-contention retries and WAL-gap recovery.
- Subqueries in DELETE predicates are rejected because their results could change during windowed apply.
- pgwire, HTTP, and QWP acknowledge deferred DELETE submission. Because the exact row count is computed later by WAL apply, they return `0` rather than exposing a sequencer transaction as an affected-row count.
- Apply-time table-rename recompilation is bounded and yields for later WAL retry when rename churn prevents progress.

## Configuration and tradeoffs

- `cairo.wal.delete.rows.per.step` controls survivor window sizing for arbitrary predicates. Smaller values reduce staging memory but add window overhead.
- `cairo.wal.delete.disk.bounded=false` preserves reader-atomic apply by default.
- Enabling disk-bounded mode reduces peak Parquet conversion footprint but increases commits and write amplification, and concurrent readers may observe partial progress.

## Enterprise companion

The abstract `SecurityContext.authorizeTableDelete` hook requires the Enterprise companion PR questdb/questdb-enterprise#1132, which adds the DELETE permission and upgrade migration.

## Test plan

- DELETE parser, compiler, bind capture, timestamp precision, and predicate semantics
- Atomic and disk-bounded WAL apply, injected failures, suspension, resume, and crash-equivalent replay
- TableWriter replace-range, Parquet conversion durability, indexes, symbols, variable-size columns, and rollback cleanup
- Materialized-view invalidation, lock contention, restart, and checkpoint recovery
- pgwire authorization and affected-count behavior, including parked writers
- HTTP and QWP affected-count behavior
- Focused local suite: 534 tests covered; all non-PG classes passed, then the complete 72-test PG multi-statement class passed after updating obsolete pre-DELETE parser expectations


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 879 / 1010 (87.03%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/griffin/engine/functions/rnd/RndIntervalFunctionFactory.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/SqlValidationProcessor.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/cairo/DefaultCairoConfiguration.java|0|2|00.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalWriter.java|3|5|60.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/DeleteOperation.java|28|43|65.12%|
|:large_blue_circle:|io/questdb/cutlass/http/processors/JsonQueryProcessor.java|18|25|72.00%|
|:large_blue_circle:|io/questdb/cairo/wal/ApplyWal2TableJob.java|24|31|77.42%|
|:large_blue_circle:|io/questdb/cutlass/pgwire/PGPipelineEntry.java|36|46|78.26%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewRefreshJob.java|18|22|81.82%|
|:large_blue_circle:|io/questdb/cairo/TableWriter.java|134|158|84.81%|
|:large_blue_circle:|io/questdb/cairo/wal/WalEventCursor.java|12|14|85.71%|
|:large_blue_circle:|io/questdb/griffin/CompiledQueryImpl.java|12|14|85.71%|
|:large_blue_circle:|io/questdb/griffin/SqlCompilerImpl.java|167|189|88.36%|
|:large_blue_circle:|io/questdb/griffin/SqlParser.java|42|47|89.36%|
|:large_blue_circle:|io/questdb/cairo/wal/OperationExecutor.java|208|232|89.66%|
|:large_blue_circle:|io/questdb/cairo/CairoEngine.java|13|14|92.86%|
|:large_blue_circle:|io/questdb/cairo/DatabaseCheckpointAgent.java|18|19|94.74%|
|:large_blue_circle:|io/questdb/cairo/mv/WalTxnRangeLoader.java|47|48|97.92%|
|:large_blue_circle:|io/questdb/cutlass/qwp/server/egress/QwpEgressUpgradeProcessor.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewState.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/griffin/WhereClauseParser.java|28|28|100.00%|
|:large_blue_circle:|io/questdb/PropertyKey.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/CairoConfigurationWrapper.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/cairo/mv/MatViewGraph.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/griffin/SqlKeywords.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/AbstractOperation.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/functions/date/TimestampShuffleFunctionFactory.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/griffin/engine/ops/OperationDispatcher.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/cairo/wal/WalUtils.java|28|28|100.00%|
|:large_blue_circle:|io/questdb/griffin/ReadOnlyStatementGate.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/griffin/model/ExecutionModel.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/tasks/TableWriterTask.java|1|1|100.00%|

bluestreak01 (changes requested):
@nwoolmer

# Review: PR #7403 — `feat(sql): DELETE FROM <table> WHERE ... on WAL tables`

**Reviewed at level 3** (full mission-critical pass: change-surface map, 13 parallel agents, per-finding source verification).

Scale: 68 files, ~9.6k diff lines, 98 commits. `DeleteTest` (1621 l), `DeleteWindowedApplyTest` (780 l), `DeleteDiskBoundedApplyTest` (907 l), `TableWriterReplaceRangeDirectTest` (514 l), `OperationExecutorWindowStepTest` (141 l) are new.

---

## Critical

### C1. An arbitrary-predicate DELETE rewrites every row of the table and un-tiers every Parquet partition — even when it matches zero rows — **in-diff**

`core/src/main/java/io/questdb/cairo/wal/OperationExecutor.java:517-543` (`replaceWithSurvivors`), `:463-473` (`convertParquetPartitionsForDelete`, else-branch), `core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java:5100-5140` (`classifyDeleteTimeRange`).

```java
// replaceWithSurvivors — unconditional whole-table range
final long minTs = tableWriter.getMinTimestamp();
final long maxTs = tableWriter.getMaxTimestamp();
...
long wLo = minTs;
while (wLo <= maxTs) { ... }
```
```java
// convertParquetPartitionsForDelete, arbitrary branch
for (int i = 0; i < partitionCount; i++) {
    if (tableWriter.getPartitionFormat(i) == PartitionFormat.PARQUET) {
        tableWriter.convertPartitionParquetToNative(tableWriter.getPartitionTimestamp(i), false);
```

The fast path is only taken when `classifyDeleteTimeRange` proves the predicate reduces to a **single static interval with `intrinsicModel.filter == null`**. Every other shape falls to the whole-range survivor replace. That includes the shape users will actually write:

```sql
DELETE FROM trades WHERE ts IN '2025-01-01' AND symbol = 'BTC-USD';
```

`WhereClauseParser` extracts the interval *and* leaves `symbol = 'BTC-USD'` as a residual filter → `pureInterval == false` → the arbitrary route → the engine reads, decodes, sorts and rewrites **every partition of the table**, and converts **every Parquet partition to native**, to delete rows confined to one day. On a 10-billion-row table that is a full-table read + full-table write + full de-tiering, per DELETE. `DELETE FROM t WHERE id = 42` behaves the same.

The PR's own comment calls this "correct but pessimistic … bounded once a per-partition survivor fast-path lands". That is a shipped performance characteristic of the default route, not a plan, and it is the dominant cost of the feature.

**Fix in this PR:** the machinery to bound it is already present. `classifyDeleteTimeRange` already runs `WhereClauseParser.extract`; when it yields intervals *plus* a residual filter, keep the intervals and pass them to `executeDelete` as the outer sweep bound instead of `[minTs, maxTs+1)`, with the residual filter left on the survivor factory. Windows outside the intervals need no rewrite at all, and `convertParquetPartitionsForDelete`'s arbitrary branch must then convert only partitions overlapping those intervals (the `findPartitionFloorIndex` binary search at `:713` already exists for exactly this). Add a coverage-map row and a test asserting untouched partitions keep their `nameTxn` and their Parquet format.

### C2. Making `timestamp_shuffle()` / `rnd_interval()` non-deterministic silently breaks existing materialized views on upgrade — **out-of-diff**

`core/src/main/java/io/questdb/griffin/engine/functions/date/TimestampShuffleFunctionFactory.java:106-109`, `core/src/main/java/io/questdb/griffin/engine/functions/rnd/RndIntervalFunctionFactory.java:117-120`; consumers `core/src/main/java/io/questdb/griffin/FunctionParser.java:647-650`, `core/src/main/java/io/questdb/cairo/mv/MatViewRefreshSqlExecutionContext.java:73-75`, `core/src/main/java/io/questdb/griffin/SqlCompilerImpl.java:3596-3597`.

Verified:
```java
// FunctionParser.java:647
} else if (!sqlExecutionContext.allowNonDeterministicFunctions() && function.isNonDeterministic()) {
    Misc.freeObjList(args);
    throw SqlException.nonDeterministicColumn(node.position, node.token);
}
```
```java
// MatViewRefreshSqlExecutionContext.java:73
public boolean allowNonDeterministicFunctions() { return false; }
```
`Function.isNonDeterministic()` defaults to `false` (`Function.java:346`), and `isRandom()` is **not** consulted by that gate. So before this PR both functions compiled inside a materialized view. After it:

- `CREATE MATERIALIZED VIEW ... timestamp_shuffle(a, b) ...` is rejected;
- a view **already persisted on master** using either function fails its next refresh recompile with `nonDeterministicColumn` and is driven to invalid. No DDL is required — a restart plus the next refresh is enough.

`SAMPLE BY ... FILL(timestamp_shuffle(...))` is rejected at `SqlCodeGenerator.java:4009` for the same reason.

Neither override is needed by DELETE: `SqlCompilerImpl.isDeletePredicateReplayStable` short-circuits on `function.isRandom()` first (`SqlCompilerImpl.java:798-800`), and `RndFunction.isRandom()` is pre-existing.

**Fix in this PR:** revert both `isNonDeterministic()` overrides and make the DELETE gate check what it actually needs — `return !(function.isNonDeterministic() || function.isRandom());` in `isDeletePredicateReplayStable(Function)`. `TimestampShuffleFunction` still needs its new `isRandom()` override. If the mat-view rejection is intentional, it is a user-visible compatibility break that must be in the PR body and covered by a test.

### C3. The named-bind arm still replays an IPv4 bind as INT while the indexed arm was fixed — **in-diff**

`core/src/main/java/io/questdb/cairo/wal/WalEventCursor.java:677-680`:
```java
case ColumnType.INT:
case ColumnType.IPv4:
    bindVariableService.setInt(name, readInt());
    break;
```
versus the fixed indexed arm at `:575-577` (`setIPv4(i, readInt())`). The same hunk split `DATE` and `TIMESTAMP` in **both** arms but split `IPv4` in only one. A WAL `UPDATE t SET x = 1 WHERE ip = :addr` therefore replays with an INT-typed variable while `$1` replays with an IPv4-typed one — different `typeOf()`, different operator selection, different formatting.

**Fix:** split the named arm to mirror the indexed one, and add the round-trip test (see C11).

### C4. The disk-bounded route holds a read lock across the whole delete, and the matching write lock is a non-yielding busy spin — **in-diff (lock use) / pre-existing (lock impl)**

`core/src/main/java/io/questdb/cairo/mv/MatViewGraph.java:98-110` runs the **entire** `replaceWithSurvivorsDiskBounded` loop — N windows, Parquet conversion, N commits, unbounded IO — inside `dependentViews.lockForRead()` / `unlockAfterRead()`.

`core/src/main/java/io/questdb/std/SimpleReadWriteLock.java` (verified):
```java
public void lock() {
    while (!lock.compareAndSet(false, true)) { Thread.yield(); }
    int n = nReaders.addAndGet(MAX_READERS);
    while (n != MAX_READERS) { n = nReaders.get(); }   // no yield, no pause
}
```
The second loop is a **pure spin with no `Thread.yield()` and no `Os.pause()`**. Any concurrent writer — `CREATE MATERIALIZED VIEW` on that base table, view registration, startup hydration, role-promote rehydration — burns a full CPU core for the entire duration of the delete. The class javadoc also states the lock is **not reentrant**, so any path that re-enters it from inside the action deadlocks.

**Fix in this PR:** do not hold the dependency lock across the loop. Read the dependent-view count under the lock, release it, and re-check it once per window (a mat view added mid-delete already forces re-invalidation through the normal notify path). Separately, `SimpleReadWriteLock.WriteLock.lock()`'s drain loop should `Os.pause()` / `Thread.onSpinWait()` — that is a pre-existing defect reached by this change and cheap to fix here.

### C5. DELETE reports `0` affected rows to PG-wire and HTTP; the pre-existing UPDATE twin reports the sequencer transaction number — **in-diff + pre-existing**

`core/src/main/java/io/questdb/griffin/CompiledQueryImpl.java:93-98` and `core/src/main/java/io/questdb/cutlass/pgwire/PGPipelineEntry.java:1755` both hard-code `0`.

PostgreSQL clients branch on the `DELETE n` tag. Hibernate optimistic locking, `EntityManager.remove`, and every driver's `rowcount` treat `DELETE 0` as "no row matched" — a conflict/failure signal. Returning a deliberate lie is not obviously better than the alternative; the two honest options are to await apply, or to document `0` as "count unavailable" and surface it consistently.

The same investigation confirms a **pre-existing bug in visited code**: `PGPipelineEntry.java:1749`
```java
sqlAffectedRowCount = tableWriterAPI.apply(updateOperation);
```
and `WalWriter.java:829` `applyNonStructural` → `return getSequencerTxn();`. So a WAL `UPDATE` through a parked pg-wire writer reports the **sequencer transaction number** as the affected-row count. The PR's own comment ("Do not expose the submission seqTxn as rows") identifies exactly this hazard and fixes it only in its own copy.

**Fix in this PR:** apply the same `0` (or the same chosen semantics) to the UPDATE parked-writer path, and state the client-visible contract in the PR body and docs.

### C6. `executeDelete`'s `catch (CairoException)` rollback is unguarded, bypassing the disk-bounded suspend conversion — **in-diff**

`core/src/main/java/io/questdb/cairo/wal/OperationExecutor.java:262-300`. The `catch (Throwable)` branch wraps `rollback()` in try/catch and calls the comment-declared "LOAD-BEARING" `setSeqTxn(seqTxn - 1)` afterwards. The `catch (CairoException)` branch does not:

```java
tableWriter.rollback();          // unguarded
if (ex.isWALTolerable() && !isDiskBounded) { ... }
tableWriter.setSeqTxn(seqTxn - 1);
if (isDiskBounded && ex.isWALTolerable()) { throw CairoException.critical(0) ... }
```

If `rollback()` throws a WAL-tolerable `CairoException`, the original exception is replaced, `setSeqTxn(seqTxn - 1)` never runs, and the disk-bounded critical-conversion never runs — so `ApplyWal2TableJob.processWalSql`'s catch finalizes a **partially applied** delete at `S`. That is the exact silent-partial-delete the guard exists to prevent.

**Fix:** mirror the `catch (Throwable)` shape — guard the rollback, log the secondary failure, and always reach the `setSeqTxn` / conversion logic.

### C7. Pre-existing twin of the bug this PR fixed, 6000 lines away in the same file — **pre-existing**

`core/src/main/java/io/questdb/cairo/TableWriter.java:5686` `copyOrRebuildColumnIndexes`, reached from `convertPartitionParquetToNative:1709` — the very method this PR modified:

```java
final long dataAddr = TableUtils.mapRO(ff, dFile(...), LOG, dataSize, MemoryTag.MMAP_TABLE_WRITER);
IndexWriter iw = IndexFactory.createWriter(indexType, configuration);   // OUTSIDE the try
try {
    ...
} finally {
    Misc.free(iw);
    ff.munmap(dataAddr, dataSize, MemoryTag.MMAP_TABLE_WRITER);         // skipped if free() throws
}
```

Both defects — `createWriter` outside the mmap's ownership scope, and `munmap` reachable only if `Misc.free` returns normally — are exactly what the PR corrected in `rebuildPartitionIndexFiles:12447-12473`. It also lacks the `indexWriter.commit()` the PR added there.

**Fix in this PR:** apply the identical three changes to `copyOrRebuildColumnIndexes`, with one leak-forcing test covering both.

### C8. Pre-existing: `TableWriter.apply(AbstractOperation, seqTxn)`'s `catch (Throwable)` omits the load-bearing `setSeqTxn(seqTxn - 1)` — **pre-existing**

`core/src/main/java/io/questdb/cairo/TableWriter.java:940-948`. `rollback()` (`:3511`) gates its entire body — including the `txWriter.unsafeLoadAll()` that restores the durable seqTxn — on `(o3InError || inTransaction())`. An ALTER/UPDATE that throws before entering a transaction leaves both flags false, so `rollback()` is a no-op, the in-memory seqTxn stays at `S`, and `ApplyWal2TableJob`'s catch records `writer.getSeqTxn()` into the sequencer — marking a transaction applied that never was.

The PR's `executeDelete` documents this precisely and fixes it for its own copy only.

**Fix in this PR:** add `setSeqTxn(seqTxn - 1)` to `apply(AbstractOperation, seqTxn)`'s `Throwable` branch, with a test.

### C9. The "one window of rows" memory bound does not hold, and the test that certifies it is fixture-tuned — **in-diff**

`WalUtils.deleteAtomicWindowHiExcl` rounds every window **up to the logical partition ceiling**, so the real bound is *one logical partition*, not `cairo.wal.delete.rows.per.step`. On a 1-second-resolution `PARTITION BY DAY` table, `rows.per.step=1000` still stages ~86,400 rows per window.

`DeleteWindowedApplyTest.testWindowArithmeticAndActualApplyUseManyWindows:452` asserts `maxWindowStaged <= rowsPerStep` — but only because its fixture is one row per **minute** (1440/day, 720 survivors) so the day ceiling happens to fall under the 1000 target. Change the fixture density and the assertion fails while the code is unchanged.

**Fix:** either bound the window by rows (per-partition row counts are available in O(1) from `txWriter`) instead of by a uniform time-density estimate, or rename the knob and rewrite the test to assert the bound that actually holds (`maxWindowStaged <= rowsPerStep + maxRowsInOneLogicalPartition`) using a dense fixture that would fail under the current claim.

### C10. Two tautological assertions inside the test that gates the memory bound — **in-diff**

`core/src/test/java/io/questdb/test/cairo/wal/DeleteWindowedApplyTest.java:518-526`:
```java
final long singleWindowStaged = count("select count(*) from t where not (x % 2 = 0) and ts >= " + minTs + " and ts < " + (maxTs + 1));
Assert.assertEquals("the single (unwindowed) window stages the ENTIRE survivor set", totalSurvivors, singleWindowStaged);
```
`totalSurvivors` is `count("select count(*) from t where not (x % 2 = 0)")` and `[minTs, maxTs+1)` covers the whole table by construction — the assertion cannot fail. The next assertion, `maxWindowStaged * 10 <= singleWindowStaged`, is then a verbatim duplicate of `maxWindowStaged * 10 <= totalSurvivors` five lines above.

The test also re-implements the production tiling loop and asserts properties of **its own** re-implementation; only the final `actualWindowCount.get() >= 20` observes the production loop.

**Fix:** delete both dead assertions; assert the tiling through the `setDeleteWindowObserver` hook against the real apply, not against a copy of the formula.

### C11. Untested behavioral changes (Step 2.6 gate)

Each row below was located by recorded search and has no test with a plausible failure link.

| # | Untested change | Why it is Critical |
|---|---|---|
| a | `processO3Block` inline Parquet drop: the **empty-batch + partially-covered** case (must throw, must not drop) | wrong branch here deletes surviving rows |
| b | Inline drop of a partition that is simultaneously first **and** last (single-partition table), and of a **split** partition | the `partitionDataMin/Max` bounds differ in exactly these cases |
| c | `srcNameTxn != txWriter.txn` guards on **ordinary (non-windowed) O3 commits** — the guards sit on the hottest commit path in the engine | a skipped candidate leaks a partition directory; nothing asserts on-disk duplicate versions outside the new DELETE tests |
| d | `WalTxnRangeLoader.sortAndMergeNewIntervals` with **≥2 intervals**, unsorted/duplicate/overlapping input, odd counts | the only existing test (`MatViewTest:5104`) loads 1 interval and returns at `:111` before any merge runs. This code changes **every** mat-view incremental refresh |
| e | Indexed `IPv4` and indexed `TIMESTAMP_NANO` WAL bind replay; **no on-disk compatibility test** (`_event` written by master, replayed by this code) | `WalTableWriterFuzzTest` DATE cases were green on master *with* `setLong`, so they cannot detect a revert |
| f | `ApplyWal2TableJob` recompile exhaustion → `EjectApplyWalException` | reverting to the unbounded `while(true)` livelocks an apply thread and no test notices |
| g | The `DatabaseCheckpointAgent` **TRUNCATE** arm — the case that carried the pre-existing KNOWN-BUG comment | only the DELETE arm is covered (`CheckpointTest:4040`); the `hasTruncate() ? "truncate operation"` branch never executes |
| h | End-to-end WAL DELETE over a **DECIMAL** column | `applyReplaceRangeWindow`'s own comment says a null `executionContext` NPEs for DECIMAL destinations; only the direct `replaceRange` unit test covers decimals |
| i | HTTP `/validate` DELETE arm (`SqlValidationProcessor.java:310`) | zero test references |
| j | `WalWriter.apply(DeleteOperation)`'s `inTransaction()` rejection; `DeleteOperation.apply()` non-WAL throw; `DeleteOperation.authorize()` null-context throw | three new error paths, no test matches their messages |
| k | No `DeleteFuzzTest` / `FuzzDeleteOperation`, though `FuzzRunner` already drives replace-range (`FuzzRunner:248/392/873/1151`) and `UpdateFuzzTest` is a direct template | this is the highest-value coverage available for a whole-table rewrite path |
| l | JIT vs interpreted filters, and parallel vs single-threaded apply, for the survivor factory | `cairo.sql.jit.mode` defaults to `on`; no DELETE test pins which path runs |
| m | Nanos designated timestamp on the **pure-time-range** fast path and on Parquet | only the survivor route is covered for nanos |

### C12. `testCrossZeroRangeDoesNotOverflow` passes with the fix reverted — **in-diff**

`core/src/test/java/io/questdb/test/griffin/engine/functions/date/TimestampShuffleFunctionFactoryTest.java:56-70`. With `start = Long.MIN_VALUE + 1`, `end = Long.MAX_VALUE`, the old expression `start + rnd.nextPositiveLong() % (end - start)` is `% -2`, which for a positive dividend yields `0` or `1` — every sample is `MIN_VALUE+1` or `MIN_VALUE+2`, and both bounds-only assertions still hold. The real defect (a 2⁶⁴-wide range collapsing to two values) is invisible.

**Fix:** assert distribution, not bounds — track that values above and below zero are both produced and that the distinct count exceeds a threshold. Those hold under `Long.remainderUnsigned` and cannot hold under `%`.

### C13. `TimestampShuffleFunctionFactoryTest.setUp()` disables `AbstractCairoTest.setUp()` for the whole class — **in-diff**

```java
public class TimestampShuffleFunctionFactoryTest extends AbstractFunctionFactoryTest {
    @Before
    public void setUp() {                    // no super.setUp()
        sqlExecutionContext.setRandom(...);
    }
```
Chain verified: `AbstractFunctionFactoryTest` → `BaseFunctionFactoryTest` (`BaseFunctionFactoryTest.java:38`) → `AbstractCairoTest`, which declares `@Before public void setUp()` (`AbstractCairoTest.java:424-425`). Java overriding means the parent `@Before` is **not** invoked: no `engine.getViewStateStore().clear()`, no `resetNameRegistryMemory()`, no metadata-cache clear, no `SharedRandom` reset — for every test in the class, and the state leaks into whatever runs next in the JVM.

**Fix:** rename to `setUpTimestampShuffleRandom()` (or call `super.setUp()` first).

### C14. The two "crash" tests do not simulate a crash — **in-diff**

`DeleteDiskBoundedApplyTest.testPerWindowCommitReappliesIdempotentlyAfterMidLoopCrash:421` and `testAtomicRouteConvertCommitThenReplaceCrashSuspendsAtSMinus1:616` inject an `openRW` failure and let the in-process retry run. Neither discards the writer or re-opens the table from disk before the retry (`releaseAllWriters()` appears only at `:132`/`:230`, in other tests). The PR's central crash-safety argument is "durable state at `S-1` is re-read after process restart"; these tests validate the in-process error path instead.

**Fix:** add `engine.releaseAllWriters(); engine.releaseInactive();` between the fault and the re-drain so the retry genuinely re-reads `_txn` from disk, and assert `writerTxn`/`seqTxn` read back from a fresh reader.

### C15. `testConcurrentReaderPinsDiskBoundedWindowVersions` pins a magic window index and falls back to wall-clock timeouts — **in-diff**

`DeleteDiskBoundedApplyTest.java:93-124`. The pause hook fires only when `windowIndex == 24`, a value derived from the current tiling formula; any change to `deleteWindowStep` silently makes the hook never fire, after which `paused.await(10, TimeUnit.SECONDS)` fails as a timeout rather than as a behavioral assertion. `applyThread.join(10_000)` then proceeds regardless of whether the thread finished, and an `AssertionError` raised on `applyThread` is not re-thrown into the test thread.

**Fix:** trip on the first window index `>= 1` that leaves rows on both sides (or have the observer count and pause at `n/2` computed from the observed total), rejoin with `Assert.assertTrue(applyThread.join(...))` semantics, and propagate the worker's throwable.

---

## Moderate

### M1. `buildSurvivorCopier` generates a new bytecode class on every DELETE apply; the pattern it cites caches
`OperationExecutor.java:481-490`. Its javadoc says it "mirrors `MatViewRefreshJob.getRecordToRowCopier`", but `MatViewRefreshJob` *caches* the copier on `MatViewState` (`MatViewRefreshJob.java:1063`) and regenerates only when metadata changes. Each `generateCopier` defines a JVM class that is not unloaded. Inert for correctness and for a single delete; it grows Metaspace across many deletes.

### M2. New fsyncs are unconditionally blocking, unlike `syncColumns`
`TableWriter.java:11752-11764` fsyncs `2 × columnCount` fds per converted partition on the pre-existing ALTER CONVERT PARTITION path. `syncColumns` (`:14397-14398`) honours `CommitMode.ASYNC` via `msync(..., async)`; the new loop does not. The partition-dir fsync at `:1822` does follow the existing precedent at `:1107`, so only the column loop diverges. Inert for correctness — it strictly adds durability — but it changes ALTER latency for operators who chose ASYNC.

### M3. `OperationExecutor.tableWriterHasParquet` and the time-range Parquet pre-pass are O(partitionCount) linear scans
`OperationExecutor.java:404-460` and `:735-744`, next to `findPartitionFloorIndex` at `:713` written for exactly this job, and `TxReader.hasParquetPartitions()` which already answers the second question. Inert on correctness; both run once per DELETE.

### M4. `SqlCompilerImpl.isDeletePredicateReplayStable(ExpressionNode, ...)` re-parses and re-clones every FUNCTION subtree
`SqlCompilerImpl.java:5010-5045`. The whole tree is parsed once into a `Function` tree, then the AST probe calls `functionParser.parseFunction(ExpressionNode.deepClone(...))` again **per FUNCTION node** — O(nodes²) parse-and-clone with allocation, on both compiles of every DELETE. Inert on correctness and confined to the compile path; still avoidable by threading the already-parsed children through one walk.

### M5. `SqlCompilerImpl` allocates a dedicated `WhereClauseParser` per compiler instance
`SqlCompilerImpl.java:209`. Compilers are pooled per connection/worker; almost none ever compiles a DELETE. Inert; lazily create it.

### M6. `notifyMatViewBaseTableCommit` can fire twice for one batch
`ApplyWal2TableJob.java:584` (try path, sets `matViewCommitNotified`) and `:654` (`catch (Throwable)`, **not** guarded by that flag). The eject path at `:621` *is* guarded. Inert in effect — the refresh task is idempotent — but the asymmetry is clearly unintended.

### M7. `MatViewGraph.DEPENDENT_VIEWS_PRESENT = -1` shares the domain of the returned row count
`MatViewGraph.java:51`. `replaceWithSurvivorsDiskBounded` returns a row count that is always `>= 0`, so the sentinel is unambiguous today. Inert; a `boolean`-returning overload or an out-parameter removes the latent aliasing.

### M8. `AbstractDeleteApplyTest` (68 lines) does not carry the shared fixtures
`DeleteTest` does not extend it at all, and eight hand-copied fault-injecting `FilesFacade` subclasses plus repeated "create table / insert / drain / assert" blocks are duplicated across the three Delete test classes. Inert; it is duplication, not a weakened assertion.

### M9. Javadoc on `DeleteWindowedApplyTest.testDeleteCommitInvalidatesMatViewWhenHousekeepingFails:381-393` describes a different test
It documents a "50k-row table … roughly 35 day-aligned windows … memory-shape proxy", which is `testWindowArithmeticAndActualApplyUseManyWindows`. Copy-paste leftover; inert.

### M10. Config knob name and documentation overstate the guarantee
`cairo.wal.delete.rows.per.step` does not bound rows per step (C9). The `CairoConfiguration` javadoc does disclose the partition-ceiling expansion, but `server.conf`'s entry does not lead with it. Inert; documentation only.

---

## Minor

- **Banner comments — explicit house-rule violation.** 9 occurrences: `TableWriterReplaceRangeDirectTest.java:343,348`; `DeleteWindowedApplyTest.java:245,250`; `DeleteTest.java:190,406,632,1096,1411`. Several also embed internal process artifacts (`Task 1.10`, `Task 2.1`, `Task 3.2`, `SPIKE`, `spec 2026-07-11`) that mean nothing in this repo.
- **Alphabetical member ordering** — `SecurityContext.authorizeTableDelete` placed after `authorizeTableTruncate` (3 files); `CairoConfiguration.getWalDeleteRowsPerStep` / `getWalDeleteDiskBounded` (5 files) and `PropertyKey.CAIRO_WAL_DELETE_*`; `WalUtils`' new `@TestOnly` statics above the public constant block and `deleteWindowHiExcl`/`deleteWindowStep` after the `set*` methods; `TableWriter`'s five new public replace-range methods and the `public static setReplaceRangePostCommitObserver` dropped into the instance section; `OperationExecutor`'s seven private methods interleaved with public ones; `SqlCompilerImpl`'s new DELETE methods; `MatViewState.clearPendingInvalidation` (`:552`) after `isPendingInvalidation` (`:525`); `MatViewGraph`'s new public constant and nested interface above the private statics.
- **Unused import** — `OperationExecutor.java:33` imports `MatViewRefreshJob`; the only reference is prose in a comment.
- **Comment-to-code ratio** — `OperationExecutor.executeDelete` is ~140 lines of which the majority is commentary, including multi-paragraph essays inside catch blocks; `convertParquetPartitionsForDelete` carries a ~35-line inline comment justifying one boolean. Much of it duplicates the method javadoc directly above.
- **`.gitignore`** gains `.superpowers/` — agent-scaffolding noise in a public repo. No such files remain tracked.
- **Test naming and SQL style** — `testDeleteByArbitraryCondition` says nothing about what is asserted; SQL keyword case flips within files; several 5-digit literals lack `_` separators.
- **`TimestampShuffleFunctionFactoryTest.testEndBeforeStart:78-79`** asserts the identical call twice, and pins the magic value `643856` for an inverted range whose behavior is undefined (the code cannot distinguish `start > end` from a span wider than `Long.MAX_VALUE`).

---

| Original claim | Why dismissed |
|---|---|
| `CompiledQuery.DELETE = 39` overflows `ReadOnlyStatementGate`'s bitmask | The mask is a `long`, `bit()` returns `1L << n`, and `isRefusedOnReadOnly` guards `sqlType >= Long.SIZE` (`ReadOnlyStatementGate.java:75, 110`). |
| `JsonQueryProcessor`'s `assert queryExecutors.size() == TYPES_COUNT + 1` breaks on the renumbering | DELETE is registered at `:124` and EMPTY at `:141`; size 41 = 40 + 1. |
| `classifyDeleteTimeRange`'s `finally` double-frees the interval functions | `IntrinsicModel.clear()` calls `runtimeIntervalBuilder.clear()`, which branches on `isOwnershipTransferred` and only calls `freeAndClear()` when `build()` never ran (`RuntimeIntervalModelBuilder.java`, `IntrinsicModel.java:61-70`). The PR's comment misnames the method but the behavior is correct. |
| `WalTxnRangeLoader.sortAndMergeNewIntervals` can reorder pre-existing entries below `intervalStart` | All three callers clear or freshly allocate the list first (`MatViewRefreshJob:2281`, `CairoEngine:2528`, `DatabaseCheckpointAgent:667`), so `intervalStart == 0` always. |
| A test mock silently allows `authorizeTableDelete` | The only repo impls are `AllowAllSecurityContext` (no-op by design), `ReadOnlySecurityContext` (throws) and `DenyAllSecurityContext` (inherits the throw). `PGSecurityTest#testDisallowDelete:127` covers the deny path. |
| `replaceRange`'s survivor-in-range contract is "only an `assert`" | QuestDB runs with `-ea`; per house rules an `assert` is the preferred guard for an invariant no user operation can trip. |
| `EjectApplyWalException.INSTANCE` carries a stale stack trace | Established QuestDB singleton-exception pattern, unchanged by this PR. |
| `SqlParser.parseDelete` allocates `ExpressionNode`s outside the pools | `parseUpdate` does the same (`SqlParser.java:4742`); pre-existing convention, parse-time only. |
| `assertQuery` nested inside `assertMemoryLeak` is a defect (85 instances) | It is the repo-wide idiom: 485 of 907 test files do both. `QueryAssertion.assertMemoryLeak` intentionally brackets with `engine.clear()` / `releaseInactive(engine)`. No lambda in the new tests contains a *lone* `assertQuery`, so the house rule is not violated. |
| New `assertSql(...)` / `.returnsOnce(...)` in the diff | Zero occurrences. One real violation remains: `DeleteWindowedApplyTest.java:194-201` uses `getPlanSink(...)` + `Chars.contains` instead of `assertQuery(sql).assertsPlanContaining(...)`. |
| Inline Parquet drop uses an uninitialised sink slot | `TableWriter.java:9971-9974` already memsets the col-top region and writes slots 0 and 6; the new code writes 1,2,3,4,5,7 — exactly matching `O3PartitionJob.updatePartitionSink:4094-4100`. |
| `deleteAtomicWindowHiExcl`'s `PartitionBy.NONE` branch is a correctness hole | WAL tables must be partitioned (`CairoEngine.createTable` asserts it), so the branch is unreachable from the only caller. |
| `srcNameTxn != txWriter.txn` guards break ordinary O3 commits | Both sites are in `o3ConsumePartitionUpdateSink`'s `partitionMutates && newPartitionTimestamp == partitionTimestamp` branch, where a partition's name txn was stamped by an earlier *committed* txn, so `srcNameTxn < txWriter.txn` always holds outside a windowed bracket. Equality is reachable only inside the new multi-window bracket. (Coverage for this is still missing — see C11c.) |

---

## Coverage map

Behavioral changes are grouped by symbol; new error paths and NULL/boundary cases get their own rows. `H`=happy, `E`=error, `N`=NULL, `B`=boundary, `C`=concurrency, `R`=`assertMemoryLeak`.

| # | Change | Test (found by recorded `rg`) | Failure link | H | E | N | B | C | R | Verdict |
|---|---|---|---|---|---|---|---|---|---|---|
| 1 | `processO3Block` inline Parquet drop, fully-covered interior | `DeleteTest#testDeleteInteriorParquetPartitionsByTimeRange` | asserts count + min/max ts + not suspended; a wrong throw suspends | + | NA | NA | + | NA | + | TESTED |
| 2 | …fully-covered **last** partition | `DeleteTest#testDeleteWholeParquetLastPartitionDropsInline` | asserts the partition dir is gone via `table_partitions('t')` | + | NA | NA | + | NA | + | TESTED |
| 3 | …fully-covered **first** partition | `DeleteTest#testDeleteWholeParquetPartitionByTimeRange` | `assertSqlCursors` vs oracle | + | NA | NA | + | NA | + | TESTED |
| 4 | …first-and-last (single-partition table) | none | — | − | − | NA | − | NA | − | **UNTESTED (C11b)** |
| 5 | …**split** Parquet partition | none | — | − | − | NA | − | NA | − | **UNTESTED (C11b)** |
| 6 | …`!fullyCovered` throw, **non-empty** batch | `WalWriterReplaceRangeTest#testReplaceRangeNotSupportedParquetPartition:1046` | asserts suspension | NA | + | NA | + | NA | + | TESTED |
| 7 | …`!fullyCovered` throw, **empty** batch + partial cover | none | — | − | **−** | NA | − | NA | − | **UNTESTED (C11a)** |
| 8 | …non-DELETE caller (mat-view REPLACE_RANGE over Parquet) | `WalWriterReplaceRangeTest#testReplaceRangeMatViewEmptyDropsFullyCoveredParquetPartition:951` | asserts view not suspended + day-1 dropped | + | NA | NA | + | NA | + | TESTED |
| 9 | `srcNameTxn != txn` guard, windowed, fully-removed branch | `DeleteWindowedApplyTest#testMultiWindowReplaceReclaimsEveryWindowsSupersededPartitionDirs:596` | on-disk `countDuplicatePartitionVersionDirs == 0`; regression leaves 12 vs 6 | + | NA | NA | + | NA | + | TESTED |
| 10 | …split-partition sibling emptied | `#testSplitPartitionSiblingEmptiedDrainKeepsLivePartitionsIntact:691` | self-verifying 2-version fixture + oracle + not suspended | + | NA | NA | + | NA | + | TESTED |
| 11 | …**ordinary non-windowed O3 commit** | none | — | − | − | NA | − | − | − | **UNTESTED (C11c)** |
| 12 | `WalEventCursor` named TIMESTAMP_NANO | `DeleteTest#testDeleteFormerInternalNamedTimestampBindsPreserveValuesAndTypes:905` | predicate asserts `typeOf(...)='TIMESTAMP_NS'`; micros replay flips it false | + | NA | NA | + | NA | + | TESTED |
| 13 | …named/indexed DATE | `WalTableWriterFuzzTest:517,:667` | weak — green on master with `setLong` | ± | NA | − | − | NA | + | **UNTESTED (C11e)** |
| 14 | …**indexed TIMESTAMP_NANO**, **indexed IPv4** | none | — | − | − | − | − | NA | − | **UNTESTED (C11e)** |
| 15 | …**named IPv4 (not fixed)** | none | — | − | − | − | − | NA | − | **UNTESTED — and defective (C3)** |
| 16 | …on-disk `_event` compat (master writer → this reader) | none | — | − | − | − | − | NA | − | **UNTESTED (C11e)** |
| 17 | `sortAndMergeNewIntervals`, 1 interval | `MatViewTest:5104` | asserts exact lo/hi | + | NA | NA | + | NA | + | TESTED |
| 18 | …**≥2 / unsorted / duplicate / odd-count** | none — returns at `:111` before any merge | — | − | − | NA | − | NA | − | **UNTESTED (C11d)** |
| 19 | `hasDelete()` barrier, refresh + checkpoint | `MatViewTest:8507`, `CheckpointTest#…OnBaseTableDeleteInGap:4040` | persisted state INVALID + reason + non-advanced base txn | + | NA | NA | + | NA | + | TESTED |
| 20 | …**TRUNCATE arm** of the same new block | none | — | − | − | NA | − | NA | − | **UNTESTED (C11g)** |
| 21 | `invalidateView` deferred-lock exit + `clearPendingInvalidation` | `MatViewTest#testInvalidateRetriesAfterConcurrentRefresh:8310` | asserts pending true under lock, then false + invalid | + | NA | NA | + | + | + | TESTED |
| 22 | …suspended-skip, read-only, auth-error exits; `runOnce` yield | none | — | − | − | NA | − | − | − | **UNTESTED** |
| 23 | `ApplyWal2TableJob` recompile-exhaustion eject | none (`rg EjectApplyWalException core/src/test` → 0) | — | − | − | NA | − | NA | − | **UNTESTED (C11f)** |
| 24 | …mat-view notify reorder | `DeleteWindowedApplyTest#testDeleteCommitInvalidatesMatViewWhenHousekeepingFails:394` | pins the bundle; reverting only the reorder still passes | ± | + | NA | + | NA | + | **UNTESTED (attribution)** |
| 25 | `OperationDispatcher` `!isAsyncWriterCommandSupported()` | `WriteFenceEntryPointMatrixTest#testDeleteDoesNotUseAsyncWriterQueueWhenWalWriterBusy:144` | asserts `EntryUnavailableException` **and** `asyncBranchEntered == false` | + | + | NA | + | NA | + | TESTED |
| 26 | `timestamp_shuffle` `range == 0` | `TimestampShuffleFunctionFactoryTest#testEqualEndpointsReturnEndpoint` | revert → `/ by zero` | + | + | NA | + | NA | NA | TESTED |
| 27 | …`range < 0` unsigned remainder | `#testCrossZeroRangeDoesNotOverflow` | **passes with the fix reverted** | − | − | NA | − | NA | NA | **UNTESTED (C12)** |
| 28 | …/`rnd_interval` new `isNonDeterministic()` engine-wide effect | none | — | − | − | NA | − | NA | − | **UNTESTED (C2)** |
| 29 | `rows.per.step < 1` rejected at startup | `PropServerConfigurationTest:1407-1449` | asserts the exact message for `-1`, env `0`, prop `0` | + | + | NA | + | NA | NA | TESTED |
| 30 | `MatViewGraph.applyIfNoDependentViews` all 3 outcomes | `MatViewGraphAndStateStoreTest:86,93,100,121,161` | asserts value vs `DEPENDENT_VIEWS_PRESENT` | + | NA | NA | + | − | + | TESTED |
| 31 | `WalWriter.apply(DeleteOperation)` `inTransaction()` reject | none | — | − | − | NA | − | NA | − | **UNTESTED (C11j)** |
| 32 | `DeleteOperation.apply()` non-WAL throw; `authorize()` null-context throw | none (`testDeleteRejectsNonWal` hits the compile-time throw with the same text) | — | − | − | NA | − | NA | − | **UNTESTED (C11j)** |
| 33 | `authorizeTableDelete` deny path | `PGSecurityTest#testDisallowDelete:127` | `assertQueryDisallowed` | NA | + | NA | NA | NA | + | TESTED |
| 34 | `WhereClauseParser.extract` free-on-throw, **interval** funcs | `WhereClauseParserTest#testExtractFailureClosesRetainedRuntimeIntervalFunction:1500` | counts closes; without the catch it stays 0 | NA | + | NA | + | NA | − | TESTED |
| 35 | …**key-value / excluded-value** funcs | none — `IntrinsicModel.clear()` does not free them | — | − | − | NA | − | NA | − | **UNTESTED** |
| 36 | `rebuildPartitionIndexFiles` mmap-leak fix | none (`IndexFactoryTest` never goes through `TableWriter`) | — | − | − | NA | − | NA | − | **UNTESTED (C7)** |
| 37 | …new `indexWriter.commit()` (a durability, not corruption, fix) | `AlterTableConvertPartitionTest#testConvertPartitionFsyncsNativeFilesInSyncMode:1076` intercepts only `.d` fds, not index `.k`/`.v` msync | — | − | − | NA | − | NA | + | **UNTESTED** |
| 38 | `rollback()` clears `pendingParquetToNativeConversions` | `DeleteDiskBoundedApplyTest#testRollbackClearsPendingParquetConversionCleanupBeforeUnrelatedBatch:195` | asserts the unrelated batch is unaffected | + | + | NA | + | NA | + | TESTED |
| 39 | pure-time-range route (micros) | `DeleteTest#testDeleteOpenEndedTimeRangeDropsOldPartitions`, `…BoundedTimeRangeTrimsAndDrops`, `…TimeRangeBoundarySplitsPartition` | oracle comparisons | + | NA | NA | + | NA | + | TESTED |
| 40 | …**nanos** designated timestamp on the pure route / on Parquet | none | — | − | − | − | − | NA | − | **UNTESTED (C11m)** |
| 41 | arbitrary survivor route, dedup / O3 / symbol-index / var-size | `DeleteTest#testDeleteOnDedupTableKeepsKeysUnique`, `…MiddleBandWithOutOfOrderSurvivors`, `…BySymbolKeepsIndexConsistent`, `DeleteWindowedApplyTest#testVarSizeColumnsThroughSurvivorReplace` | oracle comparisons | + | NA | + | + | NA | + | TESTED |
| 42 | …over a **DECIMAL** column end-to-end | none (`rg -i decimal` over the three Delete tests → 0) | — | − | **−** | − | − | NA | − | **UNTESTED (C11h)** |
| 43 | disk-bounded route: per-window commit, replay, volatile-predicate fallback | `DeleteDiskBoundedApplyTest` ×16 | oracles + suspension assertions | + | + | NA | + | ± | + | TESTED |
| 44 | …crash recovery (writer discarded, reopened from disk) | `#testPerWindowCommitReappliesIdempotentlyAfterMidLoopCrash:421` — IO fault only, no reopen | — | ± | + | NA | ± | NA | + | **UNTESTED (C14)** |
| 45 | window tiling / `deleteWindowStep` / `deleteAtomicWindowHiExcl` arithmetic | `OperationExecutorWindowStepTest` ×11 | exact-value unit assertions | + | NA | NA | + | NA | NA | TESTED |
| 46 | …the "≤ rows.per.step staged" bound it claims to prove | `DeleteWindowedApplyTest#testWindowArithmeticAndActualApplyUseManyWindows:452` | fixture-tuned; two assertions tautological | ± | NA | NA | − | NA | + | **UNTESTED (C9, C10)** |
| 47 | pg-wire / HTTP / QWP / `ReadOnlyStatementGate` DELETE arms | `PGMultiStatementMessageTest:614,632`, `ScoreboardHandlingTest:65`, `QwpEgressDdlExecTest:367`, `ReadOnlyStatementGateMatrixTest:72` | exact affected-count + survivor set + refusal | + | ± | NA | + | NA | + | TESTED |
| 48 | HTTP `/validate` DELETE arm | none | — | − | − | NA | − | NA | − | **UNTESTED (C11i)** |
| 49 | JIT vs interpreted, parallel vs single-threaded apply | none | — | − | NA | NA | − | − | NA | **UNTESTED (C11l)** |
| 50 | fuzz coverage for the whole-table rewrite path | none — no `DeleteFuzzTest`, no `FuzzDeleteOperation`, though `FuzzRunner` already drives replace-range | — | − | − | − | − | − | − | **UNTESTED (C11k)** |
| 51 | parser: WHERE required, alias forms, subquery rejection, non-WAL/view/mat-view rejection | `DeleteTest#testDeleteRequiresWhere`, `…RejectsSubQueryPredicate`, `…RejectsNonWal`, `…RejectsPlainView`, `…RejectsMaterializedView`, `SqlKeywordsTest` | exact message + position | + | + | NA | + | NA | + | TESTED (**alias form `DELETE FROM t AS a WHERE …` untested**) |
| 52 | `$(sql)` → `$safe(sql)` logging | none | — | − | − | NA | − | NA | NA | **UNTESTED (Minor)** |

**Totals: 52 behavioral changes — 28 tested, 24 UNTESTED.**

---

## Summary

**Verdict: request changes.**

**Correctness & performance gate:** not met. C1 (whole-table rewrite and full Parquet de-tiering for the most common scoped-DELETE shape), C2 (existing materialized views stop refreshing after upgrade), C3 (named IPv4 bind replays with the wrong type), C4 (a non-yielding write-lock spin held for the duration of a delete), C5 (client-visible affected-row counts, plus the pre-existing UPDATE twin), C6 (a rollback path that can finalize a partial delete), C7 and C8 (pre-existing defects in code this review visited, each a direct twin of something the PR fixed elsewhere) all affect correctness, data integrity, concurrency, or IO on reachable paths.

**Test gate:** not met. 24 of 52 behavioral changes are UNTESTED, including six new error paths, the inline-Parquet-drop throw branch, the interval merge that runs on every materialized-view refresh, indexed IPv4/nanos WAL bind replay, on-disk WAL compatibility, and DECIMAL columns end to end. Three tests that *appear* to gate central claims do not: C10 (two assertions that cannot fail inside the memory-bound gate), C12 (a regression test that passes with its fix reverted), C14 (two "crash" tests that never discard the writer). C13 disables `AbstractCairoTest.setUp()` for a whole class. There is no fuzz coverage for a path that rewrites entire tables, though `FuzzRunner` already drives replace-range and `UpdateFuzzTest` is a direct template.

**Regressions and tradeoffs the PR body does not disclose.** The description is accurate about the DELETE feature but omits every behavior change to non-DELETE workloads, and omits most of the bundled fixes:

- `processO3Block`'s inline Parquet drop changes **every** replace-range commit, including materialized-view `REPLACE_RANGE` refresh — a refresh that used to fail loudly over a Parquet partition now silently drops it;
- the two `srcNameTxn != txWriter.txn` guards sit on the hottest commit path in the engine;
- `WalTxnRangeLoader`'s interval sort/merge rewrite changes every incremental refresh;
- `MatViewRefreshJob`'s `pendingInvalidation` rework changes every invalidation source;
- `WalEventCursor`'s bind-variable retyping changes replay of every existing WAL UPDATE;
- the `isNonDeterministic()` overrides break existing materialized views (C2);
- unlisted bundled fixes: `TimestampShuffleFunctionFactory` overflow/divide-by-zero, `WhereClauseParser.extract` leak-on-throw, `rebuildPartitionIndexFiles` mmap leak + missing `commit()`, new fsyncs in `convertPartitionParquetToNative`/`produceNativeFromParquet`, `TableWriter.rollback()` conversion cleanup, `ApplyWal2TableJob` notification ordering + rename-retry bound, `DatabaseCheckpointAgent` truncate-gap fix, `$(sql)` → `$safe(sql)`.

Nor does it state that an arbitrary-predicate DELETE rewrites the entire table and un-tiers all Parquet partitions even when it matches zero rows — the single most important operational fact about the feature.

**Other metadata items.** Title: `feat(sql): DELETE FROM <table> WHERE ... on WAL tables` does not repeat a verb and reads as a fragment in release notes — prefer `feat(sql): add DELETE FROM ... WHERE on WAL tables`. Issue **#3461 "Support deleting a record"** (OPEN, New feature) is almost certainly what this implements and is not referenced with `Fixes #3461`. Labels should add `Bug` and `Performance` given the bundled fixes. A second undisclosed enterprise merge-order hazard exists alongside `authorizeTableDelete`: `CairoConfiguration.getWalDeleteRowsPerStep()` / `getWalDeleteDiskBounded()` are abstract with no `default`, so any downstream implementer breaks identically. `PGMultiStatementMessageTest` **deletes** exact error-position assertions rather than updating them — worth confirming no user-visible error message for non-DELETE SQL changed silently. Commit titles use Conventional Commits prefixes and exceed 50 chars, which CLAUDE.md forbids — but the branch is squash-merged, so this is noise, not a finding.

None of the above is a reason to split this PR. All of it — including the two pre-existing defects (C7, C8) and the pre-existing UPDATE affected-count bug in C5 — should land here, each with its own regression test and its own coverage-map row.

**Statistics.** 13 review agents produced 132 draft findings; 41 verified and consolidated, 13 dropped as false positives after source verification (listed above), the remainder deduplicated. Split: 28 in-diff, 7 out-of-diff, 6 pre-existing.

PR #7404, opened 2026-07-15
chore(core): de-flake the SF recovery header-fault test

glasstiger:
Tandem of questdb/java-questdb-client#68 — this PR only bumps the
`java-questdb-client` submodule pointer to that fix (`5cc68a9` ->
`8c53638`). Merge the two together.

## What

The SF store-and-forward recovery header-fault regression test
(`MmapSegmentRecoveryFaultTest.testHeaderFaultOnMapPastEofIsSkippableAnyFilesystem`)
flaked the JDK 8 CI:

```
java.lang.InternalError: a fault occurred in a recent unsafe memory
access operation in compiled Java code
```

The test called `MmapSegment.openExisting(ff, path)` directly from the test
lambda. Under C2, HotSpot inlines `openExisting` into the handler-less test
frame and delivers the async unsafe-access `InternalError` past
`openExisting`'s own `catch (Throwable)`, so the conversion to a
`MmapSegmentException` never runs and a raw `InternalError` escapes. This is a
property of HotSpot's async "recent unsafe memory access" fault delivery for
JIT-compiled code, not a defect in the recovery code: production recovery runs
once at startup (interpreted/C1), where the fault is delivered precisely and
caught.

## Change (in the client PR)

- Route the two header-skippable tests through `SegmentRing.openExisting(...)`,
  production's real recovery caller, so the faulting read stays behind a
  non-inlined frame whose `catch (Throwable)` converts it to a
  `MmapSegmentException` that `SegmentRing`'s per-file catch skips.
- Add a facade-aware `SegmentRing.openExisting(FilesFacade, String, long)`
  overload so the length-injecting test facade reaches
  `MmapSegment.openExisting`; the `FilesFacade.INSTANCE` overload production
  uses is unchanged.

Fail-on-revert is preserved. Full detail and diff:
questdb/java-questdb-client#68.

## Verification

The routed header tests pass under full-suite JIT warmup on JDK 11 (6/6, the
closest available pre-21 proxy for the JDK 8 CI) and JDK 17 (13/13). JDK 8 was
not available locally; the JDK 8 CI is the final confirmation.

## Known limitation, left unchanged

`testScanFaultOnMapPastEofIsHandledAnyFilesystem` is a separate, pre-existing
flake of the same class and is not touched here. It is green on the JDK 8 CI and
in isolation, but fails in-suite on JDK 11/17 because the sibling methods warm
the JIT and the scan-path async fault then escapes every Java-level catch under
C2 — verified that neither routing through `SegmentRing` nor
`-XX:CompileCommand=dontinline` fixes it. Left as-is because it is green on the
shipping target and production is unaffected. See questdb/java-questdb-client#68
for the analysis.

## Test plan

- `mvn -pl core test -Dtest=MmapSegmentRecoveryFaultTest` — the two
  header-skippable tests exercise the real `SegmentRing` recovery path
- CI on JDK 8 confirms the header flake is gone on the shipping target


glasstiger:
not needed

